PR #6485, opened 2021-02-01
TINY-6942: blocked mutation from happening on resize end

spocke:
Related Ticket:

Description of Changes:
* This PR blocks some mutations on resize end when the RTC plugin is enabled.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/api/dom/ControlSelection.ts
+++ b/modules/tinymce/src/core/main/ts/api/dom/ControlSelection.ts
@@ -10,6 +10,7 @@ import { Selectors, SugarElement } from '@ephox/sugar';
 import * as CefUtils from '../../dom/CefUtils';
 import * as NodeType from '../../dom/NodeType';
 import * as RangePoint from '../../dom/RangePoint';
+import * as Rtc from '../../Rtc';
 import Editor from '../Editor';
 import Env from '../Env';
 import * as Events from '../Events';
@@ -19,7 +20,6 @@ import { EditorEvent } from '../util/EventDispatcher';
 import Tools from '../util/Tools';
 import VK from '../util/VK';
 import EditorSelection from './Selection';
-
 interface ControlSelection {
   isResizable: (elm: Element) => boolean;
   showResizeRect: (elm: Element) => void;
@@ -248,7 +248,7 @@ const ControlSelection = (selection: EditorSelection, editor: Editor): ControlSe
     };
 
     // Set width/height properties
-    if (wasResizeStarted) {
+    if (!Rtc.isRtc(editor) && wasResizeStarted) {
       setSizeProp('width', width);
       setSizeProp('height', height);
     }
@@ -270,7 +270,10 @@ const ControlSelection = (selection: EditorSelection, editor: Editor): ControlSe
 
     if (wasResizeStarted) {
       Events.fireObjectResized(editor, selectedElm, width, height, 'corner-' + selectedHandle.name);
-      dom.setAttrib(selectedElm, 'style', dom.getAttrib(selectedElm, 'style'));
+
+      if (!Rtc.isRtc(editor)) {
+        dom.setAttrib(selectedElm, 'style', dom.getAttrib(selectedElm, 'style'));
+      }
     }
     editor.nodeChanged();
   };
--- a/modules/tinymce/src/core/main/ts/api/dom/ControlSelection.ts
+++ b/modules/tinymce/src/core/main/ts/api/dom/ControlSelection.ts
@@ -20,6 +20,7 @@ import { EditorEvent } from '../util/EventDispatcher';
 import Tools from '../util/Tools';
 import VK from '../util/VK';
 import EditorSelection from './Selection';
+
 interface ControlSelection {
   isResizable: (elm: Element) => boolean;
   showResizeRect: (elm: Element) => void;


modules/tinymce/src/core/main/ts/api/dom/ControlSelection.ts
```diff
@@ -19,7 +20,6 @@ import { EditorEvent } from '../util/EventDispatcher';
 import Tools from '../util/Tools';
 import VK from '../util/VK';
 import EditorSelection from './Selection';
-
```
lnewson, line 22: This line shouldn't have been removed.

lnewson (commented):
I don't understand this change sorry Spocke... doing this will effectively kill the ability to do any resizing, so why not just disable that instead? That would also be less confusing, as atm IIUC it'll look like it'll start to resize, but just not do anything since the dimensions aren't written to the dom.

lnewson:
> I don't understand this change sorry Spocke... doing this will effectively kill the ability to do any resizing, so why not just disable that instead? That would also be less confusing, as atm IIUC it'll look like it'll start to resize, but just not do anything since the dimensions aren't written to the dom.

Oh wait, you might be using the events instead to do this? I have to wonder if RTC would just be better to override the `editor.dom.setAttrib`, `editor.dom.setStyle`, etc... APIs to prevent this sort of mutation though?

modules/tinymce/src/core/main/ts/api/dom/ControlSelection.ts
```diff
@@ -248,7 +248,7 @@ const ControlSelection = (selection: EditorSelection, editor: Editor): ControlSe
     };
 
     // Set width/height properties
-    if (wasResizeStarted) {
+    if (!Rtc.isRtc(editor) && wasResizeStarted) {
```
lnewson, line 251: I'm not sure this can actually be done thinking more, as the current expectation is that by the time the `ObjectResized` event has fired, then the DOM would have been updated, which then allows other logic to run. This breaks that, as the assuming you are using that event, then others also using it may get it called without the content being changed.

This would almost definitely break the table plugin logic, as it definitely needs the table to have been resized when the `ObjectResized` event is fired.

lnewson: Maybe we need to consider adding a `BeforeObjectResized` event?

spocke:
This used to be disabled so that you could only select images. I think it would be an easy win to be able to re-use this resize code in RTC, my initial thought was to rewrite this from scratch. I use the ObjectResized yes, to know when to update the model. We could introduce a new event and I could listen for that instead. However listening for `ObjectResized` would work fine unless someone other listens for that and inspects the model and see that it't not updated with the right size yet. Since the rendering on the RTC side is async it will not update the view until after these events have been fired unless we locally mutate the image or flush the model to the view both things I want to avoid. I guess we could add a `BeforeObjectResized` however since the `ObjectResized` will not really be resized in view either at this point I think it's fine to use the `ObjectResized` for now and maybe change this later if some client X needs fine grain control over resizing of things.

modules/tinymce/src/core/main/ts/api/dom/ControlSelection.ts
spocke: Right eslint was giving me problems I think I moved the line but missed that it also removed the newline. Fixed now.

metricjs:
FYI I'm basing my comment on the assumption this will only change the behaviour of `ObjectResized` in RTC. Pretty sure that's right, but mentioning it just in case.

Are we okay with the fact this may make implementing tables harder in RTC, since the current table logic relies on the current behaviour of this event? Or do you consider that a future problem, since RTC aren't implementing tables yet?

Since you mentioned customers Spocke, I went looking for customers who hook into `ObjectResized` and found a couple related tickets. Though the one use case for this event we found wasn't using the image size after resize so this wouldn't break their code. Still, the fact we know of one customer who is definitely using it makes me wonder how many others are.

Is there any other way to make it so `ObjectResized` fires after the view is updated? I know you said you would rather not do the approaches above, but I do worry about the behaviour change. I guess I'm wary of doing something we know will be a problem down the line, especially since it's a public event. And surely there's other events like this that we're going to hit similar problems with? Maybe we need a discussion about what else we might hit, and how to handle these kinds of things?

spocke:
I guess I could whitelist this mutation in the mutation blocker as long as the model is synced up to the view it should work but that one if getting more and more complex with rules for different mutation scenarios. That way this PR wouldn't even be needed. But in general the view will not be updated directly as we do today. So there will be a massive behavior change all tests that look for things in the view would have to be waiters. This is a performance thing we would just need to live with.

TheSpyder:
> Are we okay with the fact this may make implementing tables harder in RTC, since the current table logic relies on the current behaviour of this event? Or do you consider that a future problem, since RTC aren't implementing tables yet?

I don't think it will make implementing tables harder in RTC. With async rendering the changes made by the event are still available immediately, just in a different place. But I also don't expect to use the current logic as-is once we add tables to RTC. I perhaps haven't made that very clear, we should schedule a conversation soon :)

lnewson:
> I don't think it will make implementing tables harder in RTC. With async rendering the changes made by the event are still available immediately, just in a different place. But I also don't expect to use the current logic as-is once we add tables to RTC. I perhaps haven't made that very clear, we should schedule a conversation soon :)

That statement is very contradictory? You can't have a sync event run where the view is already meant to have been updated before it fires and then have an async operation run that updates the view during that same event.

To be clear, if the view has not updated by the time others handle this event, then that is a **breaking** change and goes against what the event is meant to be used for. As it's a past tense event name because the updates are already meant to be done, so that you can do other post-processing logic immediately. So if the view is updated asynchronously during that event, then the changes won't be available to integrators until the next event loop at the earliest.

TheSpyder:
The point I was trying to make is that code needing to synchronously read the view can change to reading the model instead. In some cases the view is still necessary, so yes RTC will also need to delay broadcasting "view has changed" events until after the next rendering cycle.

PR #6492, opened 2021-02-03
TINY-6964: Ensure br elements are scrolled into view when they are inserted

JamesToohey:
Related Ticket: TINY-6964

Description of Changes:
Previously when hitting `Shift+Enter` repeatedly, the editor would not scroll to the inserted `br` elements as they were pushed down out of the current view.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/changelog.txt
+++ b/modules/tinymce/changelog.txt
@@ -35,7 +35,7 @@ Version 5.7.0 (TBD)
     Fixed incorrect behaviour when editor is destroyed while loading stylesheets #INT-2282
     Fixed figure elements incorrectly splitting from a valid parent element when editing the image within #TINY-6592
     Fixed inserting multiple rows or columns in a table cloning from the incorrect source row or column #TINY-6906
-    Fixed a bug where `br` elements were not scrolled into view upon insertion #TINY-6964
+    Fixed an issue where new lines were not scrolled into view when pressing Shift+Enter #TINY-6964
 Version 5.6.2 (2020-12-08)
     Fixed a UI rendering regression when the document body is using `display: flex` #TINY-6783
 Version 5.6.1 (2020-11-25)
--- a/modules/tinymce/changelog.txt
+++ b/modules/tinymce/changelog.txt
@@ -35,6 +35,7 @@ Version 5.7.0 (TBD)
     Fixed incorrect behaviour when editor is destroyed while loading stylesheets #INT-2282
     Fixed figure elements incorrectly splitting from a valid parent element when editing the image within #TINY-6592
     Fixed inserting multiple rows or columns in a table cloning from the incorrect source row or column #TINY-6906
+    Fixed a bug where `br` elements were not scrolled into view upon insertion #TINY-6964
 Version 5.6.2 (2020-12-08)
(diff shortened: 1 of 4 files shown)


modules/tinymce/changelog.txt
```diff
@@ -35,6 +35,7 @@ Version 5.7.0 (TBD)
     Fixed incorrect behaviour when editor is destroyed while loading stylesheets #INT-2282
     Fixed figure elements incorrectly splitting from a valid parent element when editing the image within #TINY-6592
     Fixed inserting multiple rows or columns in a table cloning from the incorrect source row or column #TINY-6906
+    Fixed a bug where `br` elements were not scrolled into view upon insertion #TINY-6964
```
lnewson, line 38:
```suggestion
    Fixed an issue where new lines were not scrolled into view when pressing Shift+Enter #TINY-6964
```

lnewson (commented):
Changes LGTM, but not approving since I helped with this.

modules/tinymce/changelog.txt
lnewson: Just changing `a bug` to `an issue` for consistency. Also reworded, as it sounded like if I just inserted any BR anywhere it'd scroll to it now, but it's only applicable to inserting new lines where a BR is added to represent the new line.

modules/tinymce/src/core/main/ts/api/dom/DOMUtils.ts
```diff
@@ -218,7 +218,10 @@ interface DOMUtils {
   };
   is: (elm: Node | Node[], selector: string) => boolean;
   add: (parentElm: RunArguments, name: string | Node, attrs?: Record<string, string | boolean | number>, html?: string | Node, create?: boolean) => HTMLElement;
-  create: (name: string, attrs?: Record<string, string | boolean | number>, html?: string | Node) => HTMLElement;
+  create: {
+    <K extends keyof HTMLElementTagNameMap>(name: K, attrs?: Record<string, string | boolean | number>, html?: string | Node): HTMLElementTagNameMap[K];
```
ashfordneil, line 222: Kinda surprised we didn't already have this. Always good to see more types 👍

ashfordneil: Just a question — do we even need the second type signature? Unlike events and settings, I can't see a way for 3rd party code to end up introducing different strings that might be valid here. Does anything break if you set a requirement that the elements we're creating are actual HTML elements?

ashfordneil (approved):

modules/tinymce/src/core/main/ts/api/dom/DOMUtils.ts
lnewson: Yes, you can create custom elements and those won't be in the `HTMLElementTagNameMap` type.

lnewson: Also worth noting you can do `dom.create('BR')` and that's still valid, but `HTMLElementTagNameMap` only contains lower case variants. So removing the second part would be an API breaking change.

ltrouton (approved):

metricjs (approved):

Issue #6493, opened 2021-02-03
Paste plugin with smart_paste not pasting url over highlighted text.


jwcooper:
The 'paste' plugin does not seem to work with links copied from the browser toolbar (tested with firefox and chrome). The 'PowerPaste' plugin does work.

Reproduce here:
https://fiddle.tiny.cloud/d4haab/4

The expected behavior with the paste plugin is that if you highlight text, and paste it in from any url, it will auto-link the text, and not replace the text with the url.

This is with tinymce 5.6.2

PR #6494, opened 2021-02-03
TINY-6493: Paste plugin with smart_paste not pasting url over highlighted text

jwcooper:
Related Ticket:
https://github.com/tinymce/tinymce/issues/6493

Description of Changes:
Fix issue with smart paste not detecting urls, instead replacing the highlighted text with the content of the url. This PR adds a change to detect if the content in the clipboard is an absolute URL, and if so, bypasses the `plainTextMode`.

There may be a better fix, but from what I can understand, `plainTextMode` is used to determine if `smartInsertContent` should be used, or bypassed entirely. Without `smartInsertContent`, the URL's won't be set properly, and will just replace the content highlighted.

Here is where the branching of the `smartPaste` vs text paste:
https://github.com/tinymce/tinymce/blob/develop/modules/tinymce/src/plugins/paste/main/ts/core/Clipboard.ts#L358

With the `pasteAsText` being passed, the `smartInsertContent` won't get called in SmartPaste.ts:
https://github.com/tinymce/tinymce/blob/develop/modules/tinymce/src/plugins/paste/main/ts/core/SmartPaste.ts#L78

Pre-checks:
* [ ] Changelog entry added
* [ ] License headers added on new files (if applicable)

Review:
* [ ] Milestone set

### DIFF
--- a/modules/tinymce/src/plugins/paste/main/ts/core/Clipboard.ts
+++ b/modules/tinymce/src/plugins/paste/main/ts/core/Clipboard.ts
@@ -325,10 +325,11 @@ const registerEventHandlers = (editor: Editor, pasteBin: PasteBin, pasteFormat: 
 
     const isPlainTextHtml = (internal === false && Newlines.isPlainText(content));
     const isImage = SmartPaste.isImageUrl(editor, content);
+    const isAbsoluteUrl = SmartPaste.isAbsoluteUrl(content);
 
     // If we got nothing from clipboard API and pastebin or the content is a plain text (with only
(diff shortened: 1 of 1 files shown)


JamesToohey:
Thanks very much for the bug report and PR @jwcooper.

At a cursory glance this appears to be on the right track. There are a couple of small things missing so we’ll take over from here and merge this into a branch so we can add some tests, changelog entry etc.

Thanks again!

metricjs (approved):

metricjs:
- [ ] Change target branch to Jira branch once we've created it

ltrouton (approved):

PR #6497, opened 2021-02-04
TINY-6009 Added variables to set toolbar backgrounds colors for inline and sticky toolbars

lostkeys:
Related Ticket: TINY-6009, TINY-6903

Description of Changes:
* Added variables to set toolbar backgrounds colors for inline and sticky toolbars which fixes an issue with transparent toolbars in the premium Snow skin

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/oxide/src/less/theme/components/editor/editor.less
```diff
@@ -46,6 +47,7 @@
 
 .tox.tox-tinymce--toolbar-sticky-on {
   .tox-editor-header {
+    background-color: @editor-header-sticky-background-color;
```
lnewson, line 50: This isn't going to work, as inline mode will always enable sticky headers. So the inline background colour will be used while the toolbar is in view, but then when you scroll and this docks the colour will change to the sticky background colour.

I think we'll need the one variable shared across the two `tox-editor-header` background colour.

Note: I tested this locally by making the inline colour red and then the sticky color green. It swaps from red to green when the toolbar docks to the top of the window.

modules/oxide/src/less/theme/globals/global.less
```diff
@@ -13,6 +13,7 @@
 @tinymce-border-radius: 0;
 @tinymce-border-width: 1px;
 @tinymce-box-shadow: none;
+@tinymce-inline-background-color: transparent;
```
lnewson, line 16: If we do keep the 2 variables, should this be `editor-header-inline-background-color` for consistency?

modules/oxide/src/less/theme/components/editor/editor.less
lnewson: I guess another option would be to move the `.tox.tox-tinymce-inline .tox-editor-header` style from `global.less` to below this declaration so the inline rules always have a higher specificity.

lnewson (changes requested):
Marking as needs changes because of the issues mentioned in previous comments.

modules/oxide/src/less/theme/globals/global.less
lostkeys: Fixed!

modules/oxide/src/less/theme/components/editor/editor.less
lostkeys: I made an adjustment to the logic. Now the inline and sticky toolbars are automatically set if the regular toolbar is transparent. I've also kept the variables separate as I see it would be useful to be able to set them individually as a designer.

modules/oxide/src/less/theme/globals/global.less
lnewson: This doesn't look fixed? It's still starting with `tinymce`

modules/oxide/src/less/theme/components/editor/editor.less
lnewson: I think it just complicates things personally, but I don't have enough of an opinion to argue, so marking as resolved.

modules/oxide/src/less/theme/globals/global.less
lostkeys: Now it's fixed.

modules/tinymce/changelog.txt
```diff
@@ -5,8 +5,9 @@ Version 5.7.0 (TBD)
     Added the ability to resize `video` and `iframe` media elements #TINY-6229
     Added a new `font_css` setting for adding fonts to both the editor and the parent document #TINY-6199
     Added a new `ImageUploader` API to simplify uploading image data to the configured `images_upload_url` or `images_upload_handler` #TINY-4601
-    Added an Oxide variable to define the container background color in fullscreen mode #TINY-6903
-    Added a new `AfterProgressState` event that is fired after `editor.setProgressState` calls complete #TINY-6686
+    Added Oxide variable to define the container background color in fullscreen mode #TINY-6903
```
lnewson, line 8: This wasn't merged correctly, so some changes @tylerkelly13 did have been lost.

- [ ] Fix changelog merge issues

lnewson (approved):

modules/tinymce/changelog.txt
lnewson: I've fixed this, however @tylerkelly13 can you please review the new entry to make sure it looks okay?

tylerkelly13: @lnewson,

Added an Oxide variable for defining....

?

lnewson: @tylerkelly13, it's multiple variables so that would be misleading.

tylerkelly13: Added Oxide variables .....

lnewson: I'm confused? It already says that... are you looking at the above, as that's not what's currently in the changelog.

modules/tinymce/changelog.txt
```diff
@@ -6,6 +6,7 @@ Version 5.7.0 (TBD)
     Added a new `font_css` setting for adding fonts to both the editor and the parent document #TINY-6199
     Added a new `ImageUploader` API to simplify uploading image data to the configured `images_upload_url` or `images_upload_handler` #TINY-4601
     Added an Oxide variable to define the container background color in fullscreen mode #TINY-6903
+    Added Oxide variables to set toolbar backgrounds colors for inline and sticky toolbars #TINY-6009
```
tylerkelly13, line 9: @lnewson this make sense?
```suggestion
    Added Oxide variables for setting the toolbar background colors for inline and sticky toolbars #TINY-6009
```
?

lnewson: LGTM :+1:

JamesToohey (approved):

metricjs (approved):

PR #6498, opened 2021-02-04
TINY-5974: Remove list item after outdent

SimonFc:
Related Ticket: TINY-5974

Description of Changes:
* Remove list item after outdent

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


HAFRMO (approved):

modules/tinymce/src/core/test/ts/browser/keyboard/EnterKeyListsTest.ts
```diff
@@ -509,4 +510,56 @@ describe('browser.tinymce.core.keyboard.EnterKeyListsTest', () => {
...
+      '<ul>' +
+        '<li>' +
+          'a' +
+          '<ul>' +
+            '<li>\u00a0</li>' +
+          '</ul>' +
+        '</li>' +
+      '</ul>'
+    );
+    editor.focus();
```
lnewson, line 526: Why are we calling `focus` for each test here, as that shouldn't be necessary?

lnewson: Ahh, many others do that in this file. I thought we had cleaned that up in the conversion, but seems some got missed. We probably should remove them all and just tell `TinyHooks` to focus on init. If it is needed though, we probably should have that in a `beforeEach`.

Anyways, largely out of scope so we can clean that up independently.

modules/tinymce/src/core/main/ts/newline/InsertLi.ts
```diff
@@ -13,6 +13,10 @@ const hasFirstChild = (elm, name) => {
   return elm.firstChild && elm.firstChild.nodeName === name;
 };
 
+const isFirstChild = (elm: HTMLElement) => {
+  return elm.parentNode.firstChild === elm;
```
lnewson, line 17: This may not be safe, as `parentNode` is potentially `null` (though where it's called it should never be null). Maybe do this instead to be safe?
```
return elm.parentNode?.firstChild === elm;
```

lnewson (approved):

modules/tinymce/src/core/test/ts/browser/keyboard/EnterKeyListsTest.ts
SimonFc: That was correct, `focus` wasn't needed for this test so I removed that here.

JamesToohey (approved):

PR #6502, opened 2021-02-05
TINY-6901: Ensure merged cells resize correctly

JamesToohey:
**Related Ticket:**

TINY-6901

**Description of Changes:**

Merging table cells will now resize the new merged cell to the appropriate width.

I also just converted the test to BDD style so it was a bit simpler

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/tinymce/src/plugins/table/test/ts/browser/command/MergeCellCommandTest.ts
```diff
@@ -35,68 +31,153 @@ UnitTest.asynctest('browser.tinymce.plugins.table.command.MergeCellCommandTest',
...
 
-  suite.test('TestCase-TBA: Table: mceTableMergeCells', (editor) => {
-    testCommand(editor, 'mceTableMergeCells', [
-      {
-        message: 'Should merge all cells into one',
-        before: (
-          '<table>' +
+  const getWidthNumber = (widthStyle: string): number => Number(widthStyle.replace('%', ''));
+
+  const hook = TinyHooks.bddSetupLight<Editor>({
```
lnewson, line 50: Can you please put the hook at the top of the file (ie before all the helper functions), as that's where we've put it in every other test so far.

modules/tinymce/src/plugins/table/test/ts/browser/command/MergeCellCommandTest.ts
```diff
@@ -35,68 +31,153 @@ UnitTest.asynctest('browser.tinymce.plugins.table.command.MergeCellCommandTest',
...
-      editor.getBody().innerHTML = test.before;
-      editor.selection.select(editor.dom.select('td[data-mce-selected]')[0], true);
-      editor.selection.collapse(true);
-      editor.execCommand(command);
-      LegacyUnit.equal(cleanTableHtml(editor.getContent()), test.after, test.message);
-      LegacyUnit.equal(modifiedEvents, test.expectedEvents);
-    });
+  const testMerge = (editor: Editor, test: MergeCellTest) => {
+    clearEvents();
+    editor.getBody().innerHTML = test.before;
```
lnewson, line 36:
```suggestion
    editor.setContent(test.before, { format: 'raw' });
```

We shouldn't be doing this, as it should be using the APIs and it's something I've been fixing while doing the conversion.

modules/tinymce/src/plugins/table/test/ts/browser/command/MergeCellCommandTest.ts
```diff
@@ -35,68 +31,153 @@ UnitTest.asynctest('browser.tinymce.plugins.table.command.MergeCellCommandTest',
...
+    editor.selection.select(editor.dom.select('td[data-mce-selected]')[0], true);
+    editor.selection.collapse(true);
+    editor.execCommand('mceTableMergeCells');
+    LegacyUnit.equal(cleanTableHtml(editor.getContent()), test.after);
```
lnewson, line 40:
```suggestion
    assert.equal(cleanTableHtml(editor.getContent()), test.after);
```
Looks like one LegacyUnit test was missed in the conversion.

modules/tinymce/src/plugins/table/test/ts/browser/command/MergeCellCommandTest.ts
```diff
@@ -35,68 +31,153 @@ UnitTest.asynctest('browser.tinymce.plugins.table.command.MergeCellCommandTest',
...
+              '<td colspan="2">b3<br />c3</td>' +
+            '</tr>' +
+          '</tbody>' +
+        '</table>'
+      ),
+      expectedEvents: [ defaultEvent ],
+    });
+  });
+
+  it('TINY-6901: Will apply correct size to cell with colspan after cell merge', () => {
```
lnewson, line 152: This will almost definitely fail on other browsers, as you're going to have issues with precision for floating-point numbers.  I'd suggest using a fixed-width table instead, as it's much less likely to have issues.

lnewson (approved):

modules/tinymce/src/plugins/table/test/ts/browser/command/MergeCellCommandTest.ts
```diff
@@ -35,68 +31,153 @@ UnitTest.asynctest('browser.tinymce.plugins.table.command.MergeCellCommandTest',
...
+            '<td style="width: 42.7414%;"></td>' +
+            '<td style="width: 51.8049%;"></td>' +
+          '</tr>' +
+        '</tbody>' +
+      '</table>'
+    );
+
+    editor.setContent(before);
+    const cols = editor.dom.select('td[data-mce-selected]');
+    const totalColsWidth = Arr.foldl(cols, (acc, col) => acc + getWidthNumber(Css.getRaw(SugarElement.fromDom(col), 'width').getOrDie()), 0);
```
lnewson, line 171:
```suggestion
    const cols = SelectorFilter.descendant(TinyDom.body(editor), 'td[data-mce-selected]');
    const totalColsWidth = Arr.foldl(cols, (acc, col) => acc + getWidthNumber(Css.getRaw(col, 'width').getOrDie()), 0);
```
This just saves having to convert and we normally prefer sugar functions where possible.

modules/tinymce/src/plugins/table/test/ts/browser/command/MergeCellCommandTest.ts
lnewson: Ignore I missed it had custom logic sorry.

metricjs (approved):
Barring Lee's suggestions, LGTM

modules/tinymce/src/plugins/table/test/ts/browser/command/MergeCellCommandTest.ts
```diff
@@ -35,68 +38,146 @@ UnitTest.asynctest('browser.tinymce.plugins.table.command.MergeCellCommandTest',
...
     return html.replace(/<p>(&nbsp;|<br[^>]+>)<\/p>$/, '');
   };
 
-  suite.test('TestCase-TBA: Table: mceTableMergeCells', (editor) => {
-    testCommand(editor, 'mceTableMergeCells', [
-      {
-        message: 'Should merge all cells into one',
-        before: (
-          '<table>' +
+  const getWidthNumber = (widthStyle: string): number => Number(widthStyle.replace('%', ''));
```
ashfordneil, line 55: Sugar dimension thing, or is this okay just being a search?

modules/tinymce/src/plugins/table/test/ts/browser/command/MergeCellCommandTest.ts
```diff
@@ -35,68 +38,146 @@ UnitTest.asynctest('browser.tinymce.plugins.table.command.MergeCellCommandTest',
...
+          '</tr>' +
+          '<tr>' +
+            '<td style="width: 42.7414%;"></td>' +
+            '<td style="width: 51.8049%;"></td>' +
+          '</tr>' +
+        '</tbody>' +
+      '</table>'
+    );
+
+    editor.setContent(before);
```
ashfordneil, line 169: How come this is a regular setContent when elsewhere we're adding `{format: 'raw'}` to the setContent call?

ashfordneil (approved):

modules/tinymce/src/plugins/table/test/ts/browser/command/MergeCellCommandTest.ts
JamesToohey: The Sugar stuff is all computed (which with this particular case, shows a correct width), I need the raw CSS value to ensure that further operations will work i.e. inserting a column.

modules/tinymce/src/plugins/table/test/ts/browser/command/MergeCellCommandTest.ts
JamesToohey: I just ported the old test to BDD and added my new one... I don't think there's any particular reason the existing test needed to use `'raw'` so I can probably change that

modules/tinymce/src/plugins/table/test/ts/browser/command/MergeCellCommandTest.ts
ashfordneil: I was particularly referencing this (potentially over engineered) module I wrote earlier: https://github.com/tinymce/tinymce/blob/develop/modules/sugar/src/main/ts/ephox/sugar/api/view/Dimension.ts

It just takes strings and returns numbers, basically. If you find it's hard to adopt, or overkill, as it's starting to feel in a lot of the places that I suggest it, I might think about redesigning or removing it in the future

JamesToohey: Oh cool I'll look at that

PR #6504, opened 2021-02-05
TINY-6686: Ensure context toolbars also close when the throbber opens

ashfordneil:
Related Ticket: TINY-6686

Description of Changes:
* Same as the last 6686 PR, it's closing popups when the editor is blocked
* This time I made it work for context toolbars

Pre-checks:
* ~[ ] Changelog entry added~
* ~[ ] Branch prefixed with `feature/` for new features (if applicable)~
* ~[ ] License headers added on new files (if applicable)~

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/themes/silver/main/ts/ContextToolbar.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/ContextToolbar.ts
@@ -323,6 +323,13 @@ const register = (editor: Editor, registryContextToolbars, sink: AlloyComponent,
       }
     });
 
+    editor.on('AfterProgressState', (event) => {
+      if (event.state) {
+        lastAnchor.set(Optional.none());
+        InlineView.hide(contextbar);
+      }
+    });
+
     editor.on('NodeChange', (_e) => {
       Focus.search(contextbar.element).fold(
         () => {
--- a/modules/tinymce/src/themes/silver/main/ts/ContextToolbar.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/ContextToolbar.ts
@@ -10,7 +10,7 @@ import {
   Layout, LayoutInside, MaxHeight, MaxWidth, Positioning
 } from '@ephox/alloy';
 import { InlineContent, Toolbar } from '@ephox/bridge';
-import { Arr, Cell, Id, Merger, Obj, Optional, Thunk } from '@ephox/katamari';
+import { Arr, Cell, Fun, Id, Merger, Obj, Optional, Thunk } from '@ephox/katamari';
 import { PlatformDetection } from '@ephox/sand';
 import { Css, Focus, Scroll, SugarBody, SugarElement } from '@ephox/sugar';
 import Editor from 'tinymce/core/api/Editor';
@@ -191,7 +191,12 @@ const register = (editor: Editor, registryContextToolbars, sink: AlloyComponent,
     });
   }));
 
-  type ContextToolbarButtonTypes = Toolbar.ToolbarButtonSpec | Toolbar.ToolbarMenuButtonSpec | Toolbar.ToolbarSplitButtonSpec | Toolbar.ToolbarToggleButtonSpec | Toolbar.GroupToolbarButtonSpec;
+  type ContextToolbarButtonTypes =
+    Toolbar.ToolbarButtonSpec
+    | Toolbar.ToolbarMenuButtonSpec
+    | Toolbar.ToolbarSplitButtonSpec
+    | Toolbar.ToolbarToggleButtonSpec
+    | Toolbar.GroupToolbarButtonSpec;
 
   const buildContextToolbarGroups = (allButtons: Record<string, ContextToolbarButtonTypes>, ctx: InlineContent.ContextToolbar) =>
     identifyButtons(editor, { buttons: allButtons, toolbar: ctx.items, allowToolbarGroups: false }, extras, Optional.some([ 'form:' ]));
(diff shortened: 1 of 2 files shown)


modules/tinymce/src/themes/silver/test/ts/browser/throbber/ThrobberPopupTest.ts
```diff
@@ -19,6 +19,20 @@ describe('browser.tinymce.themes.silver.throbber.ThrobberPopupTest', () => {
...
+      ed.ui.registry.addContextToolbar('test-2', {
+        items: 'bold | italic',
+        scope: 'node',
+        position: 'selection',
+        predicate: (node: Node) => {
+          if (node.nodeType === NodeTypes.ELEMENT) {
+            const elem = node as HTMLElement;
+            if (elem.classList.contains('ctx-menu-me')) {
+              return true;
+            }
```
JamesToohey, line 31:
```suggestion
            return elem.classList.contains('ctx-menu-me');
```

JamesToohey (approved):

modules/tinymce/src/themes/silver/test/ts/browser/throbber/ThrobberPopupTest.ts
```diff
@@ -19,6 +19,20 @@ describe('browser.tinymce.themes.silver.throbber.ThrobberPopupTest', () => {
...
+          }
+          return false;
```
lnewson, line 33:
```suggestion
          } else {
            return false;
          }
```
This is specific "if do this this else do that" logic so we should be capturing that.

modules/tinymce/src/themes/silver/main/ts/ContextToolbar.ts
```diff
@@ -323,6 +323,13 @@ const register = (editor: Editor, registryContextToolbars, sink: AlloyComponent,
       }
     });
 
+    editor.on('AfterProgressState', (event) => {
+      if (event.state) {
+        lastAnchor.set(Optional.none());
```
lnewson, line 328: This is duplicate logic so could we please abstract this out into a single function?

lnewson (approved):

modules/tinymce/src/themes/silver/main/ts/ContextToolbar.ts
ashfordneil: Have done, called it `close` because I *think* that's what clearing the anchor (in combination with hiding the element) does. If you happen to have more wisdom around this and have a better idea I'm quite ready to rename it.

ltrouton (approved):

lnewson (changes requested):
Tested this out locally and it's missing one more thing... when the throbber is disabled, it needs to reshow context toolbars. Without that, you have to manually change the selection, etc... which isn't great UX so we should trigger the toolbars to reshow when the progress state is set to false.

modules/tinymce/src/themes/silver/test/ts/browser/throbber/ThrobberPopupTest.ts
```diff
@@ -43,6 +52,64 @@ describe('browser.tinymce.themes.silver.throbber.ThrobberPopupTest', () => {
...
+
+    editor.setProgressState(true);
+    await pWaitForThrobber();
+
+    await Waiter.pTryUntil('context toolbar is closed', () => {
+      UiFinder.notExists(SugarBody.body(), '.tox-pop');
+    });
+  });
+
+  context('correctly re-opens (or does not re-open) the context toolbar when it closes', () => {
```
JamesToohey, line 71: I feel the `context` shouldn't specify what the tests are doing, but more the... context? It means we can add/change the unit tests without this becoming outdated. Also remember they'll all be printed out together in the test logs so we should avoid duplicating what it's doing

```suggestion
  context('context toolbar', () => {
```

Thoughts?

lnewson: Agreed

ashfordneil: Yeah? Would you also suggest moving the test above (another context toolbar thing) into this block then?

lnewson (approved):
Approving, but we should still look at cleaning up the context as James mentioned. Also thanks for also fixing up the `Fun.noop` case as well :+1:

modules/tinymce/src/themes/silver/test/ts/browser/throbber/ThrobberPopupTest.ts
```diff
@@ -43,6 +52,54 @@ describe('browser.tinymce.themes.silver.throbber.ThrobberPopupTest', () => {
...
+    it('does not re-open when the throbber closes - if there is no focus on the editor', async () => {
+      // blur the editor
+      const input = SugarElement.fromTag('input');
+      Insert.append(SugarBody.body(), input);
+      Focus.focus(input);
+
+      // close the throbber
+      const editor = hook.editor();
+      editor.setProgressState(false);
+      await UiFinder.pWaitForState('throbber is closed', SugarBody.body(), '.tox-throbber', Fun.not(Visibility.isVisible));
```
lnewson, line 91:
```suggestion
      await UiFinder.pWaitForHidden('throbber is closed', SugarBody.body(), '.tox-throbber');
```
No need to over complicate this, as we already have a helper for this.

PR #6505, opened 2021-02-05
Feature/6508: Double-clicking a link also selects br, prevents dialog from getting data

camiloHimura:
Related Ticket: TINY-6508

Description of Changes:
* Add new conditions to handle Double-clicking cases in firefox and IE

Pre-checks:
* [ ] Changelog entry added
* [ ] License headers added on new files (if applicable)

Review:
* [ ] Milestone set


modules/agar/src/main/ts/ephox/agar/mouse/Clicks.ts
```diff
@@ -70,6 +70,10 @@ const click = (settings: Settings) => (element: SugarElement<Node>): void => {
   const dom = element.dom;
   Obj.get(dom as any, 'click').fold(() => event('click', settings)(element), Fun.call);
 };
+const doubleClick = (settings: Settings) => (element: SugarElement<Node>): void => {
+  const dom = element.dom;
+  Obj.get(dom as any, 'dblclick').fold(() => event('dblclick', settings)(element), Fun.call);
```
SimonFc, line 75: Isn't `dblclick` only an event?

modules/tinymce/src/core/demo/ts/demo/IframeDemo.ts
```diff
@@ -39,7 +39,7 @@ export default () => {
     ],
     add_unload_trigger: false,
     autosave_ask_before_unload: false,
-    toolbar: 'undo redo | sidebar1 | print preview media'
+    toolbar: 'undo redo | sidebar1 | print preview media code link unlink'
```
SimonFc, line 42: Accidental commit?

modules/tinymce/src/core/main/ts/selection/ElementSelection.ts
```diff
@@ -93,6 +98,20 @@ const getNode = (root: Element, rng: Range): Element => {
...
+        return endContainer.firstChild;
+      }
+      if (startOffset > endOffset || isAnchorPrevSibling(startContainer)) {
+        return startContainer.nextSibling;
+      }
+    }
+
+    // IE Handle double click an anchor.
+    if (startContainer?.nodeType === 1 && endContainer?.nodeType === 3) {
+      return startContainer.firstChild;
```
SimonFc, line 113: Same here with NodeType and firstChild.

Btw, if you have `|<p>&nbsp;</p><p>a|b</p>`, Selection.getNode will no longer return "_the currently selected element or the common ancestor element for both start and end of the selection_" so looks like you would want to do something else here.

camiloHimura: updated

modules/tinymce/src/core/demo/ts/demo/IframeDemo.ts
camiloHimura: yes, removed

modules/tinymce/src/core/main/ts/selection/ElementSelection.ts
```diff
@@ -46,6 +46,12 @@ const skipEmptyTextNodes = (node: Node, forwards: boolean) => {
   return node || orig;
 };
 
+const isAnchorPrevSibling = (element: Node): boolean =>
+  Traverse.prevSibling(SugarElement.fromDom(element)).fold(
+    Boolean,
+    (node) => node.dom.nodeName === 'A'
+  );
```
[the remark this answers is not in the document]
camiloHimura: Updated

modules/tinymce/src/core/main/ts/selection/ElementSelection.ts
```diff
@@ -93,6 +98,20 @@ const getNode = (root: Element, rng: Range): Element => {
         return startContainer;
       }
     }
+
+    if (startContainer?.nodeType === 3 && endContainer?.nodeType === 1) {
```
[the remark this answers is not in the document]
camiloHimura: updated

modules/tinymce/src/core/main/ts/selection/ElementSelection.ts
```diff
@@ -46,6 +46,10 @@ const skipEmptyTextNodes = (node: Node, forwards: boolean) => {
   return node || orig;
 };
 
+const isAnchorPrevSibling = (element: Node): boolean =>
+  Traverse.prevSibling(SugarElement.fromDom(element))
+    .exists((node) => SugarNode.name(node) === 'a');
```
JamesToohey, line 51: I believe you can use `SugarNode.isTag` here

```suggestion
Traverse.prevSibling(SugarElement.fromDom(element)).exists(SugarNode.isTag('a'));
```

modules/tinymce/src/core/test/ts/browser/dom/SelectionTest.ts
```diff
@@ -641,6 +641,22 @@ describe('browser.tinymce.core.dom.SelectionTest', () => {
     );
   });
 
+  it('dbclick getNode', () => {
```
JamesToohey, line 644: Don't forget to include the JIRA here `TINY-6508: `

modules/tinymce/src/plugins/link/test/ts/browser/ContextToolbarTest.ts
```diff
@@ -54,6 +54,12 @@ UnitTest.asynctest('browser.tinymce.plugins.link.ContextToolbarTest', (success,
           UiFinder.cWaitForState('check link content', '.tox-toolbar input', (ele) => ele.dom.value === 'http://www.google.com/')
         ])
       ]),
+      Log.stepsAsStep('TBA', 'enable button link after double click', [
```
JamesToohey, line 57: Is this test TBA or for this JIRA?

modules/tinymce/src/plugins/link/test/ts/browser/RemoveLinkTest.ts
```diff
@@ -17,6 +17,13 @@ UnitTest.asynctest('browser.tinymce.plugins.link.RemoveLinkTest', (success, fail
     const body = SugarElement.fromDom(editor.getBody());
 
     Pipeline.async({}, [
+      Log.stepsAsStep('TBA', 'Enable Removing link after double click', [
```
JamesToohey, line 20: TBA?

modules/tinymce/src/plugins/link/test/ts/browser/SelectedLinkTest.ts
```diff
@@ -30,6 +32,13 @@ UnitTest.asynctest('browser.tinymce.plugins.link.SelectedLinkTest', (success, fa
         TestLinkUi.sClickCancel,
         TestLinkUi.sClearHistory
       ]),
+      Log.stepsAsStep('TBA', 'Enable Removing link after double click', [
```
JamesToohey, line 35: Another

modules/agar/src/main/ts/ephox/agar/api/Mouse.ts
```diff
@@ -98,12 +99,20 @@ const trueClick = (elem: SugarElement<HTMLElement>): void => {
   mouseUp(elem);
   Clicks.trigger(elem);
 };
+const trueDoubleClick = (elem: SugarElement<HTMLElement>): void => {
+  Focus.focus(elem);
+  mouseDown(elem);
+  mouseUp(elem);
```
JamesToohey, line 105: Should a true double click contain two sets of `mouseDown` and `mouseUp`?

modules/tinymce/src/core/test/ts/browser/dom/SelectionTest.ts
JamesToohey: I'm not quite sure how this is testing double click?

lnewson (changes requested):
Just blocking this for now, as this seems like the wrong way to solve this sorry. What I believe we want is to either fix the initial potentially incorrect double click selection (similar to how we override triple click selection) or update the link plugin to handle a br being in the content. As such, I'll need to really dive into this and discuss with others.

I should add changing this will have quite large implications and it's the type of code that shouldn't have `if anchor do something else` type of branching, as it's behaviour should be the same across all elements.

lnewson:
It's worth noting that when I was testing this, I noticed this issue also occurs for triple-click, so we'll need to take that into account.

lnewson:
We discussed this approach during stand up and we came to the conclusion this would be better solved elsewhere. The primary reasons are that:
- This will cause the `selection.getNode()` core API to return a different value from what's actually in the selection. This has a lot of potential side effects because it's used in many places and therefore could introduce breaking behaviour.
- The core selection logic shouldn't have edge cases like what's been added to handle anchors. It should generally be generic and apply to all elements.

So with that in mind, we discussed the 2 options I mentioned previously: altering the selection or changing the link plugin code. We're concerned what side effects changing the selection could have, as you need to be able to select the br in some cases (eg copy and paste so a new line is carried across). That rules out doing the first option. So changing the link plugin code is probably the best solution here.

Before I jump into suggesting ways for you ways to solve that, one thing we did notice is that the <br> element is only being added due to a previous bug in Firefox, that appears to have since been fixed by Firefox. As such, we'd suggest starting with removing that logic (see the addBrAfterLastLinks function in Quirks.ts) as that will prevent the bogus BR being included in the first place.

Having said that though, we can still run into an issue where if you Shift+Enter after a link, that you can still end up with various BR elements. So we're still going to need some changes to the link plugin to handle that case. One possible solution we thought might work, is to get the selection start and end containers, see if either is a bogus BR element and if so:
- look forwards for the from the start container to get the next node or
- look backwards for the from the end container to get the previous node

Effectively we should ignore any bogus BR elements at the start/end of the selection when trying to find the anchor element. It's possible there are still other solutions we haven't considered, so make sure to investigate other potential solutions as well.
If you need help on how do any of the above, the others in the High Velocity team should be able to help out, otherwise feel free to reach out to us.

lnewson (changes requested):
Forgot to mark as requesting changes in my last comment sorry.

PR #6506, opened 2021-02-07
TINY-6711: Fixed an issue with widths being incorrectly cloned when inserting a new column that had a colspan previously

lnewson:
Related Ticket: TINY-6711

Description of Changes:
* Fixes the width being removed on the source cell, not the cloned cell when a colspan existed.
* Fixes widths incorrectly being cloned for a colspan (it used the width of multiple cols when it should only use the cloned column width).
* Fixed getting column widths not working when a cell in the first row didn't have a width (it'd always use the computed width in that case, not the raw width specified on other cells in the column).

**Note:** I would like to redo how the cell halving is done, as this would still fail in some cases (eg every cell has a colspan) and it's not great we're relying on other code to set the cell width. However, to do that is going to require much larger changes and not something we should be including in 5.7.

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [x] Milestone set

### DIFF
--- a/modules/snooker/src/main/ts/ephox/snooker/api/CellMutations.ts
+++ b/modules/snooker/src/main/ts/ephox/snooker/api/CellMutations.ts
@@ -4,7 +4,7 @@ import * as CellUtils from '../util/CellUtils';
 
 const halve = (main: SugarElement, other: SugarElement): void => {
   const colspan = CellUtils.getSpan(main, 'colspan');
-  // Don't set a width on the new cell if we have a colspan as we can only safely do that for cells
+  // Only set width on the new cell if we have a colspan of 1 (or no colspan) as we can only safely do that for cells
   // that are a single column, since we don't know the individual column widths for a cell with a colspan.
   // Instead, we'll rely on the adjustments/postAction logic to set the widths based on other cells in the column
   if (colspan === 1) {
--- a/modules/snooker/src/main/ts/ephox/snooker/api/CellMutations.ts
+++ b/modules/snooker/src/main/ts/ephox/snooker/api/CellMutations.ts
@@ -1,13 +1,20 @@
 import { SugarElement } from '@ephox/sugar';
 import * as Sizes from '../resize/Sizes';
+import * as CellUtils from '../util/CellUtils';
 
 const halve = (main: SugarElement, other: SugarElement): void => {
-  const width = Sizes.getGenericWidth(main);
-  width.each((w) => {
-    const newWidth = w.value / 2;
-    Sizes.setGenericWidth(main, newWidth, w.unit);
-    Sizes.setGenericWidth(other, newWidth, w.unit);
-  });
+  const colspan = CellUtils.getSpan(main, 'colspan');
+  // Don't set a width on the new cell if we have a colspan as we can only safely do that for cells
+  // that are a single column, since we don't know the individual column widths for a cell with a colspan.
+  // Instead, we'll rely on the adjustments/postAction logic to set the widths based on other cells in the column
+  if (colspan === 1) {
+    const width = Sizes.getGenericWidth(main);
+    width.each((w) => {
+      const newWidth = w.value / 2;
+      Sizes.setGenericWidth(main, newWidth, w.unit);
+      Sizes.setGenericWidth(other, newWidth, w.unit);
+    });
+  }
 };
 
 export {
--- a/modules/snooker/src/main/ts/ephox/snooker/api/TableFill.ts
+++ b/modules/snooker/src/main/ts/ephox/snooker/api/TableFill.ts
@@ -77,7 +77,7 @@ const cellOperations = (mutate: (e1: SugarElement, e2: SugarElement) => void, do
     Css.remove(clone, 'height');
     // dont inherit the width of spanning columns
     if (prev.colspan !== 1) {
-      Css.remove(prev.element, 'width');
+      Css.remove(clone, 'width');
     }
   };
 
--- a/modules/snooker/src/main/ts/ephox/snooker/api/TableOperations.ts
+++ b/modules/snooker/src/main/ts/ephox/snooker/api/TableOperations.ts
@@ -1,9 +1,9 @@
 import { Arr, Fun, Optional } from '@ephox/katamari';
 import { Remove, SugarElement, SugarNode, Width } from '@ephox/sugar';
+import * as Blocks from '../lookup/Blocks';
 import * as DetailsList from '../model/DetailsList';
 import * as GridRow from '../model/GridRow';
 import * as RunOperation from '../model/RunOperation';
-
 import * as TableMerge from '../model/TableMerge';
 import * as Transitions from '../model/Transitions';
 import * as MergingOperations from '../operate/MergingOperations';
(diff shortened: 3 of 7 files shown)


modules/snooker/src/main/ts/ephox/snooker/api/TableFill.ts
```diff
@@ -77,7 +77,7 @@ const cellOperations = (mutate: (e1: SugarElement, e2: SugarElement) => void, do
     Css.remove(clone, 'height');
     // dont inherit the width of spanning columns
     if (prev.colspan !== 1) {
-      Css.remove(prev.element, 'width');
```
lnewson, line 80: This was pretty hacky and the cause of one of the issues, as it would remove the width on the source cell (this should not of been happening, as that will cause other bugs). The reason it did that was to make sure the cell width wasn't set on the new cell during the `mutate` call. So I've moved that logic to the `CellMutations` logic instead.

modules/snooker/src/main/ts/ephox/snooker/api/TableOperations.ts
```diff
@@ -359,9 +359,14 @@ const firstColumnIsLocked = (_warehouse: Warehouse, details: Structs.DetailExt[]
 const lastColumnIsLocked = (warehouse: Warehouse, details: Structs.DetailExt[]) =>
   Arr.exists(details, (detail) => detail.column + detail.colspan >= warehouse.grid.columns && detail.isLocked);
 
-const getColumnsWidth = (details: Structs.DetailExt[]) => {
+const getColumnsWidth = (warehouse: Warehouse, details: Structs.DetailExt[]) => {
+  const columns = Blocks.columns(warehouse);
   const uniqueCols = ColUtils.uniqueColumns(details);
-  return Arr.foldl(uniqueCols, (acc, detail) => acc + Width.getOuter(detail.element), 0);
+  return Arr.foldl(uniqueCols, (acc, detail) => {
```
lnewson, line 365: This was the primary other cause, as it used the "unique" cols width. However, the cell used in the `Width.getOuter` call was the one with the colspan.

So if we had 3 cols, each ~33% and we included the colspan cell in the selection, then the delta would be `66% + 33%` (ie all three columns). However, it should have only been the 2 columns represented by the colspan (eg `33% + 33%`).

As such, this change just makes it use the column index from `uniqueCols` and then uses `Blocks.columns` to find a cell that appropriately represents the column in the table to get the outer width.

modules/snooker/src/main/ts/ephox/snooker/api/CellMutations.ts
```diff
@@ -1,13 +1,20 @@
...
-  width.each((w) => {
-    const newWidth = w.value / 2;
-    Sizes.setGenericWidth(main, newWidth, w.unit);
-    Sizes.setGenericWidth(other, newWidth, w.unit);
-  });
+  const colspan = CellUtils.getSpan(main, 'colspan');
+  // Don't set a width on the new cell if we have a colspan as we can only safely do that for cells
+  // that are a single column, since we don't know the individual column widths for a cell with a colspan.
+  // Instead, we'll rely on the adjustments/postAction logic to set the widths based on other cells in the column
+  if (colspan === 1) {
```
JamesToohey, line 10: Can we use `hasColspan` here?

```suggestion
  if (!hasColspan(main)) {
```

modules/tinymce/src/plugins/table/test/ts/browser/ModifyColumnsTableResizeTest.ts
```diff
@@ -378,6 +448,29 @@ describe('browser.tinymce.plugins.table.ModifyColumnsTableResizeTest', () => {
           mceTableDeleteCol: 0
         });
       });
+
+      it('TINY-6711: should resize table by single column width with a colspan', () => {
```
JamesToohey, line 452: I'm wondering whether the new column should have the width of colspan td if that's where the selection is? I will need to have a play to see

modules/tinymce/src/plugins/table/test/ts/browser/ModifyColumnsTableResizeTest.ts
```diff
@@ -133,6 +133,28 @@ describe('browser.tinymce.plugins.table.ModifyColumnsTableResizeTest', () => {
           mceTableDeleteCol: 0
         });
       });
+
+      it('TINY-6711: will function with a colspan', () => {
```
JamesToohey, line 137: We can remove the comment on line 55 now

JamesToohey (commented):
Looks good. I'll pull this down and have a go with it tomorrow before approving 👍

modules/snooker/src/main/ts/ephox/snooker/api/CellMutations.ts
lnewson: You could have `colspan="1"` in the HTML, so no that wouldn't work.

modules/tinymce/src/plugins/table/test/ts/browser/ModifyColumnsTableResizeTest.ts
lnewson: As far as I'm aware it's not meant to. You'll see stable will make it the width of a single column not multiple columns. There's also a number of tests that assert that.

modules/tinymce/src/plugins/table/test/ts/browser/ModifyColumnsTableResizeTest.ts
lnewson: There's a number of additional more complex colspan tests we should include, so that's why I left that (eg colspan with colgroups).

modules/tinymce/src/plugins/table/test/ts/browser/ModifyColumnsTableResizeTest.ts
lnewson: You'll also see the copy column logic does the same thing, where it works based on a single column when a colspan exists.

lnewson: We could certainly change that behaviour, but it's not something we should be looking to do given the time constraints. If we did do that though, then there's likely to be an expectation multiple columns would be inserted.

modules/snooker/src/main/ts/ephox/snooker/api/CellMutations.ts
lnewson: Oh geez, the implementation does handle that, but it's not clear from the names so I definitely don't like that. I also doesn't handle the case where someone might have `colspan="0"` or something invalid.

modules/snooker/src/main/ts/ephox/snooker/api/CellMutations.ts
```diff
@@ -1,13 +1,20 @@
...
 
 const halve = (main: SugarElement, other: SugarElement): void => {
-  const width = Sizes.getGenericWidth(main);
-  width.each((w) => {
-    const newWidth = w.value / 2;
-    Sizes.setGenericWidth(main, newWidth, w.unit);
-    Sizes.setGenericWidth(other, newWidth, w.unit);
-  });
+  const colspan = CellUtils.getSpan(main, 'colspan');
+  // Don't set a width on the new cell if we have a colspan as we can only safely do that for cells
```
JamesToohey, line 7:
```suggestion
  // Only set width on the new cell if we have a colspan of 1 as we can only safely do that for cells
```

JamesToohey (approved):

metricjs (approved):

modules/snooker/src/main/ts/ephox/snooker/lookup/Blocks.ts
```diff
@@ -9,7 +9,7 @@ import { Warehouse } from '../api/Warehouse';
  * sizes that are only available through the difference of two
  * spanning columns.
  */
-const columns = (warehouse: Warehouse): Optional<SugarElement<HTMLTableCellElement>>[] => {
+const columns = (warehouse: Warehouse, isValidCell: (detail: SugarElement<HTMLTableCellElement>) => boolean = Fun.always): Optional<SugarElement<HTMLTableCellElement>>[] => {
```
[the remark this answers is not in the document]
lnewson: We could remove the inline function type, if that might help?

ltrouton (approved):

modules/snooker/src/main/ts/ephox/snooker/lookup/Blocks.ts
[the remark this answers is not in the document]
lnewson: Have moved it out to a dedicated type.

PR #6510, opened 2021-02-08
DOC-761: Suggested edits to the 5.7 changelogs

tylerkelly13:
Related Ticket: DOC-761

Description of Changes:
* Suggested edits to the 5.7 changelogs

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/changelog.txt
+++ b/modules/tinymce/changelog.txt
@@ -1,6 +1,6 @@
 Version 5.7.0 (TBD)
     Added IPv6 address support to the URI API. Patch contributed by dev7355608 #GH-4409
-    Added new data to the `TableModified` event to describe whether the table `structure`, `style`, or both was modified #TINY-6643
+    Added new `structure` and `style` properties to the `TableModified` event to indicate what kinds of modifications were made #TINY-6643
     Added `video` and `audio` live embed support for the `media` plugin #TINY-6229
     Added the ability to resize `video` and `iframe` media elements #TINY-6229
     Added a new `font_css` setting for adding fonts to both the editor and the parent document #TINY-6199
--- a/modules/tinymce/changelog.txt
+++ b/modules/tinymce/changelog.txt
@@ -39,7 +39,7 @@ Version 5.7.0 (TBD)
     Fixed an issue where new lines were not scrolled into view when pressing Shift+Enter or Shift+Return #TINY-6964
     Fixed an issue where list elements would not be removed when outdenting using the Enter or Return key #TINY-5974
     Fixed an issue where file extensions with uppercase characters were treated as invalid #TINY-6940
-    Fixed dialog block messages were unable to be translated #TINY-6971
+    Fixed dialog block messages were not passed through TinyMCE's translation system #TINY-6971
 Version 5.6.2 (2020-12-08)
     Fixed a UI rendering regression when the document body is using `display: flex` #TINY-6783
 Version 5.6.1 (2020-11-25)
--- a/modules/tinymce/changelog.txt
+++ b/modules/tinymce/changelog.txt
@@ -4,7 +4,7 @@ Version 5.7.0 (TBD)
     Added a new locked column feature to the `table` plugin, blocking column operations on locked columns #TINY-6765
     Added `video` and `audio` live embed support for the `media` plugin #TINY-6229
     Added the ability to resize `video` and `iframe` media elements #TINY-6229
-    Added new `font_css` setting for adding fonts to both the editor and the parent document #TINY-6199
+    Added a new `font_css` setting for adding fonts to both the editor and the parent document #TINY-6199
     Added a new `ImageUploader` API to simplify uploading image data to the configured `images_upload_url` or `images_upload_handler` #TINY-4601
     Added an Oxide variable to define the container background color in fullscreen mode #TINY-6903
     Added a new `AfterProgressState` event that is fired after `editor.setProgressState` calls complete #TINY-6686
--- a/modules/tinymce/changelog.txt
+++ b/modules/tinymce/changelog.txt
@@ -1,7 +1,6 @@
 Version 5.7.0 (TBD)
     Added IPv6 address support to the URI API. Patch contributed by dev7355608 #GH-4409
     Added new data to the `TableModified` event to describe whether the table `structure`, `style`, or both was modified #TINY-6643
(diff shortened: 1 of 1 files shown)


modules/tinymce/changelog.txt
```diff
@@ -1,15 +1,15 @@
 Version 5.7.0 (TBD)
     Added IPv6 address support to the URI API. Patch contributed by dev7355608 #GH-4409
-    Added new data to the `TableModified` event to describe whether the table's `structure`, `style`, or both were modified #TINY-6643
-    Added a new locked column feature to tables that allows column operations to be prevented on certain columns #TINY-6765
+    Added new data to the `TableModified` event to describe whether the table `structure`, `style`, or both was modified #TINY-6643
```
lnewson, line 3: Are you sure about this, reading it aloud sounds really strange?

modules/tinymce/changelog.txt
```diff
@@ -1,15 +1,15 @@
...
+    Added a new locked column feature to the `table` plugin, blocking column operations on locked columns #TINY-6765
```
lnewson, line 4: This line was actually meant to be removed. It seems we forgot to?
```suggestion
```

modules/tinymce/changelog.txt
```diff
@@ -1,15 +1,15 @@
...
     Added `video` and `audio` live embed support for the `media` plugin #TINY-6229
     Added the ability to resize `video` and `iframe` media elements #TINY-6229
-    Added new `font_css` setting for fonts to be added to both containing document and the editor #TINY-6199
+    Added new `font_css` setting for adding fonts to both the editor and the parent document #TINY-6199
```
lnewson, line 7:
```suggestion
    Added a new `font_css` setting for adding fonts to both the editor and the parent document #TINY-6199
```

ashfordneil (approved):

lnewson (approved):

modules/tinymce/changelog.txt
```diff
@@ -34,13 +34,13 @@ Version 5.7.0 (TBD)
...
+    Fixed incorrect behavior when editor is destroyed while loading stylesheets #INT-2282
     Fixed figure elements incorrectly splitting from a valid parent element when editing the image within #TINY-6592
     Fixed inserting multiple rows or columns in a table cloning from the incorrect source row or column #TINY-6906
-    Fixed an issue where new lines were not scrolled into view when pressing Shift+Enter #TINY-6964
-    Fixed an issue where list elements would not be removed when outdenting using enter #TINY-5974
+    Fixed an issue where new lines were not scrolled into view when pressing Shift+Enter or Shift+Return #TINY-6964
+    Fixed an issue where list elements would not be removed when outdenting using the Enter or Return key #TINY-5974
     Fixed an issue where file extensions with uppercase characters were treated as invalid #TINY-6940
-    Fixed dialog block messages unable to be translated #TINY-6971
+    Fixed dialog block messages were unable to be translated #TINY-6971
```
metricjs, line 43: This one reads weird still... `Fixed dialog block message could not be translated`? Not sure that's any better.

metricjs (approved):

modules/tinymce/changelog.txt
tylerkelly13: @metricjs was it that they were hardcoded so we couldn't get them translated?

lnewson: The strings used in the `api.block(message)` call were never passed through the TinyMCE translation system. So not sure hard-coded is the right word, but it'd be equivalent.

lnewson: If not using one of our plugins, you could have technically done something like this to workaround it, which is why I said hard-coded wouldn't be right:
```
api.block(editor.translate(message))
```

modules/tinymce/changelog.txt
tylerkelly13:
```suggestion
    Added two optional return parameters to the `TableModified` event indicating if the `structure` or `style` was modified #TINY-6643
```
?

lnewson: `Optional` is incorrect as they are always provided. They also aren't `return parameters`, they are properties in the event object/type.

lnewson: Also, it's not exclusively either-or, it can be any of the following:
- `{ style: false, structure: true }` (structure change)
- `{ style: true, structure: false }` (style change)
- `{ style: true, structure: true }` (both structure and style change)

lnewson:
```suggestion
    Added new properties to the `TableModified` event to describe whether the table `structure`, `style`, or both were modified #TINY-6643
```

metricjs:
```suggestion
    Added new properties to the `TableModified` event to indicate if the table `structure` or `style` were modified #TINY-6643
```
removed whether so it doesn't imply it's mutually exclusive, so that we maybe don't need the `or both`? wondering if the backticks are needed...

metricjs:
```suggestion
    Added new `structure` and `style` properties to the `TableModified` event to describe how the table was modified #TINY-6643
```
or something maybe? rearranging the sentence...

modules/tinymce/changelog.txt
metricjs: could we just do this?
```suggestion
    Fixed dialog block messages were not translated #TINY-6971
```
maybe implies the translations were missing instead, but it's not incorrect...

lnewson: Yeah, that definitely sounds like they are being translated now, which is wrong as we haven't got them translated yet.

modules/tinymce/changelog.txt
lnewson: LGTM :+1:

To clarify, I was originally asking about changing `were` to `was` as that seemed wrong. So this really just spiralled out of control sorry :disappointed:

modules/tinymce/changelog.txt
metricjs: yeah... 
```suggestion
    Fixed dialog block messages were not passed through TinyMCE's translation system #TINY-6971
```
just be explicit?

modules/tinymce/changelog.txt
metricjs: it read weird to me, I only figured out why when I saw your comments

JamesToohey (approved):

PR #6514, opened 2021-02-08
TINY-6907: Added new mceListUpdate command

TheSpyder:
Related Ticket: TINY-6907

Description of Changes:

The list properties dialog now uses a new `SetListAttributes` command to set the list start attribute. I've also ensured the `mceListProps` command is registered in RTC mode and made use of it in the menu item instead of opening the dialog directly.

I didn't add a changelog for this because the list properties dialog doesn't seem to be documented yet; it's also an internal change that should have no impact on the feature.

Pre-checks:

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/plugins/lists/main/ts/actions/Attributes.ts
+++ b/modules/tinymce/src/plugins/lists/main/ts/actions/Attributes.ts
@@ -1,0 +1,16 @@
+/**
+ * Copyright (c) Tiny Technologies, Inc. All rights reserved.
+ * Licensed under the LGPL or a commercial license.
+ * For LGPL see License.txt in the project root for license information.
+ * For commercial licenses see https://www.tiny.cloud/
+ */
+
+import { Obj } from '@ephox/katamari';
+import { getParentList } from '../core/Selection';
+
+export const setParentListAttributes = (editor, attrs) => {
+  const parentList = getParentList(editor);
+  editor.undoManager.transact(() =>
+    Obj.each(attrs, (v, k) => editor.dom.setAttrib(parentList, k, v))
+  );
+};
--- a/modules/tinymce/src/plugins/lists/main/ts/actions/Attributes.ts
+++ b/modules/tinymce/src/plugins/lists/main/ts/actions/Attributes.ts
@@ -6,9 +6,10 @@
  */
 
 import { Obj } from '@ephox/katamari';
+import Editor from 'tinymce/core/api/Editor';
 import { getParentList } from '../core/Selection';
 
-export const setParentListAttributes = (editor, attrs) => {
+export const setParentListAttributes = (editor: Editor, attrs: Record<string, string>) => {
   const parentList = getParentList(editor);
   editor.undoManager.transact(() =>
     Obj.each(attrs, (v, k) => editor.dom.setAttrib(parentList, k, v))
--- a/modules/tinymce/src/plugins/lists/main/ts/actions/Update.ts
+++ b/modules/tinymce/src/plugins/lists/main/ts/actions/Update.ts
@@ -9,9 +9,15 @@ import { Obj } from '@ephox/katamari';
 import Editor from 'tinymce/core/api/Editor';
 import { getParentList } from '../core/Selection';
 
-export const setParentListAttributes = (editor: Editor, attrs: Record<string, string>) => {
+interface ListUpdate {
+  attrs?: Record<string, string>;
+}
+
+export const updateList = (editor: Editor, update: ListUpdate) => {
   const parentList = getParentList(editor);
-  editor.undoManager.transact(() =>
-    Obj.each(attrs, (v, k) => editor.dom.setAttrib(parentList, k, v))
-  );
+  editor.undoManager.transact(() => {
+    if (update.attrs) {
+      Obj.each(update.attrs, (v, k) => editor.dom.setAttrib(parentList, k, v));
+    }
+  });
 };
--- a/modules/tinymce/src/plugins/lists/main/ts/api/Commands.ts
+++ b/modules/tinymce/src/plugins/lists/main/ts/api/Commands.ts
@@ -5,26 +5,27 @@
  * For commercial licenses see https://www.tiny.cloud/
  */
 
+import Editor from 'tinymce/core/api/Editor';
 import { setParentListAttributes } from '../actions/Attributes';
 import { flattenListSelection, indentListSelection, outdentListSelection } from '../actions/Indendation';
 import * as ToggleList from '../actions/ToggleList';
 import { getParentList } from '../core/Selection';
 import * as Dialog from '../ui/Dialog';
 
-const queryListCommandState = (editor, listName) => {
+const queryListCommandState = (editor: Editor, listName: string) => {
   return () => {
     const parentList = getParentList(editor);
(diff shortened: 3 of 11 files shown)


modules/tinymce/src/plugins/lists/main/ts/actions/Attributes.ts
```diff
@@ -0,0 +1,16 @@
...
+ * Copyright (c) Tiny Technologies, Inc. All rights reserved.
+ * Licensed under the LGPL or a commercial license.
+ * For LGPL see License.txt in the project root for license information.
+ * For commercial licenses see https://www.tiny.cloud/
+ */
+
+import { Obj } from '@ephox/katamari';
+import { getParentList } from '../core/Selection';
+
+export const setParentListAttributes = (editor, attrs) => {
```
lnewson, line 11: Types?

modules/tinymce/src/plugins/lists/main/ts/api/Commands.ts
```diff
@@ -5,17 +5,25 @@
...
 
 const queryListCommandState = (editor, listName) => {
   return () => {
-    const parentList = editor.dom.getParent(editor.selection.getStart(), 'UL,OL,DL');
+    const parentList = getParentList(editor);
     return parentList && parentList.nodeName === listName;
   };
 };
 
+const registerDialog = (editor) => {
```
lnewson, line 21: Types?

modules/tinymce/src/plugins/lists/main/ts/api/Commands.ts
```diff
@@ -43,8 +51,10 @@ const register = (editor) => {
     flattenListSelection(editor);
   });
 
-  editor.addCommand('mceListProps', () => {
-    Dialog.open(editor);
+  registerDialog(editor);
+
+  editor.addCommand('SetListAttributes', (ui, detail) => {
```
lnewson, line 56: This should probably be called `mceSetListAttributes` to keep with the current naming convention (ie custom commands are prefixed with `mce`).

- [ ] Change command name to `mceSetListAttributes`

lnewson: I'm also wondering if this maybe should be called `mceListMutation` (or similar) to go with the event of a similar name? If we do that, we probably need to change the data type to something like `{ attrs: Record<string, string> }`. That would then make this much more flexible for future potential, if say we needed to set styles or classes, etc...

modules/tinymce/src/plugins/lists/test/ts/browser/ListPropertiesTest.ts
```diff
@@ -132,4 +132,52 @@ describe('browser.tinymce.plugins.lists.ListPropertiesTest', () => {
     await TinyUiActions.pWaitForUi(editor, '.tox-collection__item:contains("List properties"):not(.tox-collection__item--state-disabled)');
     TinyUiActions.clickOnMenu(editor, '.tox-mbtn:contains("Custom")');
   });
+
+  it('TINY-6907: List properties command is used to update the DOM', async () => {
+    const editor = hook.editor();
+
+    const blockCommand = (event) => {
```
lnewson, line 139: Types for event?

lnewson (approved):

TheSpyder:
Sigh, I forgot our TypeScript settings don't warn me when I forget to add types. Sorry about that.

modules/tinymce/src/plugins/lists/main/ts/api/Commands.ts
TheSpyder: I have no attachment to the command name, I'm happy to go with whatever the team wants. I am also concerned about the scalability of this to later set CSS attributes; `ApplyUnorderedListStyle` and `ApplyOrderedListStyle` from `advlist` could be replaced with this command if we do that.

A sub-object seemed like overkill but is probably the best approach.

lnewson:
> Sigh, I forgot our TypeScript settings don't warn me when I forget to add types. Sorry about that.

Yeah, I'd love to solve that... we were at ~10k errors in strict mode back in mid-2019 and we're now down to ~7k errors. That's still a long way to go before we can enable it though :cry:

lnewson:
Oops, just remembered we'd only need `noImplicitAny` for that purpose... we have ~3.8k errors still to resolve by just enabling that check.

TheSpyder:
It also didn't help that some of the code I was editing is missing types 😂 I'll add a few more in the files I edited.

modules/tinymce/src/plugins/lists/main/ts/api/Commands.ts
TheSpyder: We don't seem to use `Mutate` in our event names anywhere, how about `mceUpdateList`?

lnewson: The event name is `ListMutation`, but it's used when changing the types mostly. That suggestion seems fine to me, though we do normally try to keep the list name first, eg `mceListUpdate`.

TheSpyder: ah I was only looking at `editor.addCommand` calls, not `editor.fire` as well. oops.

list update goes well with list props. I'll use that.

modules/tinymce/src/plugins/lists/test/ts/browser/ListPropertiesTest.ts
```diff
@@ -132,4 +134,52 @@ describe('browser.tinymce.plugins.lists.ListPropertiesTest', () => {
     await TinyUiActions.pWaitForUi(editor, '.tox-collection__item:contains("List properties"):not(.tox-collection__item--state-disabled)');
     TinyUiActions.clickOnMenu(editor, '.tox-mbtn:contains("Custom")');
   });
+
+  it('TINY-6907: List properties command is used to update the DOM', async () => {
```
JamesToohey, line 138: This test title initially confused me, if I'm understanding correctly, maybe something like:

```suggestion
  it('TINY-6907: List properties dialog uses mceListUpdate command internally to update the DOM', async () => {
```

JamesToohey (approved):

modules/tinymce/src/plugins/lists/main/ts/actions/Update.ts
```diff
@@ -0,0 +1,23 @@
...
+import { getParentList } from '../core/Selection';
+
+interface ListUpdate {
+  attrs?: Record<string, string>;
+}
+
+export const updateList = (editor: Editor, update: ListUpdate) => {
+  const parentList = getParentList(editor);
+  editor.undoManager.transact(() => {
+    if (update.attrs) {
```
ltrouton, line 19: Should we do an `isObject` check here?

ltrouton (approved):

modules/tinymce/src/plugins/lists/main/ts/actions/Update.ts
TheSpyder: Indeed! I added a test for passing null `update` and it fails. Good catch!

TheSpyder:
Looks like the remaining test failure is a safari flake, so it should be good to merge whenever you're ready 🤔

PR #6528, opened 2021-02-11
TINY-4612: Update inline_boundaries to work with home/end/pgup/pgdown keys

camiloHimura:
Related Ticket: TINY-4612

Description of Changes:
* place the caret outside an inline boundary if it's at the start/end

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/tinymce/src/core/test/ts/browser/keyboard/HomeEndKeysTest.ts
```diff
@@ -53,6 +53,25 @@ describe('browser.tinymce.core.keyboard.HomeEndKeysTest', () => {
...
+      it('move caret at the begining on the line', () => {
+        const editor = hook.editor();
+        editor.setContent('<p>test <a href="google.com">link</a> test</p>');
+        TinySelections.setCursor(editor, [ 0, 1, 0 ], 0);
+        TinySelections.setSelection(editor, [ 0, 1, 0 ], 0, [ 0, 1, 0 ], 5);
+        TinyContentActions.keystroke(editor, Keys.home());
+        TinyAssertions.assertCursor(editor, [ 0, 0 ], 0);
+      });
+
+      it('move caret at the beginning on the line (parent) if the first element is an inline element', () => {
```
spocke, line 66: I think you only need to check the cases where there is a link at the start of the element. So hitting home inside plain text after a link should move the caret before the link element. Hitting home inside the middle of a link should move the caret to before the link. Something like that.

modules/tinymce/src/core/test/ts/browser/keyboard/HomeEndKeysTest.ts
```diff
@@ -95,5 +114,25 @@ describe('browser.tinymce.core.keyboard.HomeEndKeysTest', () => {
...
+      it('move caret at the end on the line', () => {
+        const editor = hook.editor();
+        editor.setContent('<p>test <a href="google.com">link</a> test</p>');
+        TinySelections.setCursor(editor, [ 0, 1, 0 ], 0);
+        TinySelections.setSelection(editor, [ 0, 1, 0 ], 0, [ 0, 1, 0 ], 5);
+        TinyContentActions.keystroke(editor, Keys.end());
+        TinyAssertions.assertCursor(editor, [ 0, 2 ], 5);
+      });
+
+      it('move caret at the beginning on the line (parent) if the first element is an inline element', () => {
```
spocke, line 128: Same deal here just add tests for the inline boundaries element override.

modules/tinymce/src/core/test/ts/browser/keyboard/PageUpDownKeysTest.ts
```diff
@@ -0,0 +1,89 @@
...
+describe('browser.tinymce.core.keyboard.PageUpDownKeyTest', () => {
+  const hook = TinyHooks.bddSetupLight<Editor>({
+    add_unload_trigger: false,
+    base_url: '/project/tinymce/js/tinymce',
+    indent: false
+  }, [ Theme ], true);
+  const platform = PlatformDetection.detect();
+
+  context('Page Up', () => {
+    it('caret is at the beginning of an inline element, escape outside the node', () => {
```
spocke, line 18: Since this will depend on browser default behavior for the non inline boundary case it would be hard to automate without say selenium. So instead you could place the caret at `<a href="#">|text</a>` and verify that it gets moved to `|<a href="#">text</a>` on key up since that is what the code will be doing in that case. And you could keypress in the middle of the link and assert that the caret was not moved at all to make sure that the native handling is not overriden.

modules/tinymce/src/core/test/ts/browser/keyboard/PageUpDownKeysTest.ts
```diff
@@ -0,0 +1,89 @@
...
+    });
+
+    it('caret is not at the beginning/end of an inline element, not escape', () => {
+      const editor = hook.editor();
+      editor.setContent('<p>text<a href="google.com">link</a>text</p>');
+      TinySelections.setCursor(editor, [ 0, 1, 0 ], 2);
+      TinyContentActions.keystroke(editor, Keys.pageDown());
+      TinyAssertions.assertCursor(editor, [ 0, 1, 0 ], 2);
+    });
+
```
spocke, line 87: Maybe remove this empty line.

modules/tinymce/src/core/main/ts/keyboard/InlineBoundariesNavigation.ts
```diff
@@ -0,0 +1,47 @@
...
+import { Fun } from '@ephox/katamari';
+import Editor from '../api/Editor';
+import * as Settings from '../api/Settings';
+import CaretPosition from '../caret/CaretPosition';
+import * as BoundaryLocation from './BoundaryLocation';
+import * as BoundarySelection from './BoundarySelection';
+import * as InlineUtils from './InlineUtils';
+import * as NavigationUtils from './NavigationUtils';
+
+const isInlineBoundaries = (editor: Editor) => {
```
spocke, line 17: This function would not be needed if you use the change I provided below.

modules/tinymce/src/core/main/ts/keyboard/PageUpDownKeys.ts
```diff
@@ -0,0 +1,45 @@
...
+ * For LGPL see License.txt in the project root for license information.
+ * For commercial licenses see https://www.tiny.cloud/
+ */
+
+import Editor from '../api/Editor';
+import VK from '../api/util/VK';
+import * as InlineBoundariesNavigation from './InlineBoundariesNavigation';
+import * as MatchKeys from './MatchKeys';
+
+const executeKeydownOverride = (editor: Editor, evt: KeyboardEvent) => {
```
spocke, line 13: This should be named executeKeyupAction or something since it's not really overriding and it's on keyup not keydown.

spocke (changes requested):

modules/tinymce/src/core/main/ts/keyboard/PageUpDownKeys.ts
```diff
@@ -0,0 +1,45 @@
...
+  MatchKeys.execute([
+    { keyCode: VK.PAGE_UP, action: MatchKeys.action(InlineBoundariesNavigation.moveOutInlineBoundaries, editor, false) },
+    { keyCode: VK.PAGE_DOWN, action: MatchKeys.action(InlineBoundariesNavigation.moveOutInlineBoundaries, editor, true) }
+  ], evt).each((_) => {
+    evt.preventDefault();
+  });
+};
+
+const setup = (editor: Editor) => {
+  const blocker = (e) => e.stopImmediatePropagation();
```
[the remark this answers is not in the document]
lnewson: Why are we even listening to `keyup` here instead of `keydown` like all the other overrides, as that's going to cause various other issues such as flickering while it's moved initially and then repositioned? It's also going to cause a number of additional selection change events which isn't great for performance (though unlikely to be a bottleneck).

Additionally I'm very curious why we are blocking `NodeChange` as that's not obvious why that's needed? It also seems like something that's going to cause issues, as many things are hooked up to NodeChange and you're preventing it now, which in turn may block other logic such as Spell Checking running.

Given the above, this really feels like the wrong way to solve this and instead it should be listening and overriding `keydown` instead.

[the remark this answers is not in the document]
spocke: I suggested it since we have no custom page up/down logic so we can't override the default as we do for the other things home/end that we do have custom code for. Also if the node change event is only blocked while the user is holding down the page up/down key it would have limited effects on other things. The alternative would be a flickering UI while caret move in and out of the inline boundary element. We could I guess skip page up/page down and implement that at a later date. Been thinking of how to do that for the content editable false re-implementation in rtc. But it's a lot of work since it need to be able to deal with layouts like tables etc.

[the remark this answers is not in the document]
lnewson: Fair enough. I guess that is okay for now, but If we do stick with this then we definitely need some comments to state why it's different and why we are blocking `NodeChange`. We ideally also should log a follow up JIRA to see if we can implement this in a better way by overriding keydown.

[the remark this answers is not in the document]
spocke: Yes, I know it's a bit hacky but at this point it either that hack or not having the feature until later. Adding more context in comment here makes a lot of sense.

modules/tinymce/src/core/test/ts/browser/keyboard/PageUpDownKeysTest.ts
```diff
@@ -0,0 +1,87 @@
...
+describe('browser.tinymce.core.keyboard.PageUpDownKeyTest', () => {
+  const hook = TinyHooks.bddSetupLight<Editor>({
+    add_unload_trigger: false,
+    base_url: '/project/tinymce/js/tinymce',
+    indent: false
+  }, [ Theme ], true);
+  const platform = PlatformDetection.detect();
+
+  context('Page Up', () => {
+    it('caret is escape outside the node', () => {
```
spocke, line 18: Maybe this should be `Caret should be placed before the inline element if it's at the start of the inline element` or something?

- [ ] Go though the titles of the tests seems like they could be made more clear what they are supposed to be doing

modules/tinymce/src/core/main/ts/keyboard/PageUpDownKeys.ts
```diff
@@ -0,0 +1,45 @@
...
+import Editor from '../api/Editor';
+import VK from '../api/util/VK';
+import * as InlineBoundariesNavigation from './InlineBoundariesNavigation';
+import * as MatchKeys from './MatchKeys';
+
+const executeKeydownOverride = (editor: Editor, evt: KeyboardEvent) => {
+  MatchKeys.execute([
+    { keyCode: VK.PAGE_UP, action: MatchKeys.action(InlineBoundariesNavigation.moveOutInlineBoundaries, editor, false) },
+    { keyCode: VK.PAGE_DOWN, action: MatchKeys.action(InlineBoundariesNavigation.moveOutInlineBoundaries, editor, true) }
+  ], evt).each((_) => {
```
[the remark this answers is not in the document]
camiloHimura: updated

modules/tinymce/src/core/main/ts/keyboard/PageUpDownKeys.ts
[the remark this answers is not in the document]
camiloHimura: updated

spocke (changes requested):

modules/tinymce/src/core/main/ts/keyboard/InlineBoundariesNavigation.ts
```diff
@@ -0,0 +1,47 @@
...
+      Fun.noop
+    );
+  });
+  return true;
+};
+
+const moveToLineEndPoint = (editor: Editor, forward: boolean): boolean => {
+  NavigationUtils.moveToLineEndPoint(editor, forward, isInlineBoundaries(editor));
+  return moveOutInlineBoundaries(editor, forward);
+};
```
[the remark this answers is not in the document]
camiloHimura: updated

modules/tinymce/src/core/test/ts/browser/keyboard/HomeEndKeysTest.ts
```diff
@@ -53,6 +53,25 @@ describe('browser.tinymce.core.keyboard.HomeEndKeysTest', () => {
       TinyAssertions.assertSelection(editor, [ 0, 0 ], 1, [ 0, 0 ], 1);
     });
 
+    context('Inline element', () => {
+      it('move caret at the begining on the line', () => {
+        const editor = hook.editor();
+        editor.setContent('<p>test <a href="google.com">link</a> test</p>');
+        TinySelections.setCursor(editor, [ 0, 1, 0 ], 0);
```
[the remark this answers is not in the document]
camiloHimura: updated

modules/tinymce/src/core/test/ts/browser/keyboard/HomeEndKeysTest.ts
camiloHimura: Done

modules/tinymce/src/core/test/ts/browser/keyboard/HomeEndKeysTest.ts
camiloHimura: updated

modules/tinymce/src/core/test/ts/browser/keyboard/PageUpDownKeysTest.ts
camiloHimura: It's no possible to simulate the behavior only with key up because the changes implemented above requires key down - key up

modules/tinymce/src/core/test/ts/browser/keyboard/PageUpDownKeysTest.ts
camiloHimura: updated

modules/tinymce/src/core/main/ts/keyboard/InlineBoundariesNavigation.ts
camiloHimura: updated

modules/tinymce/src/core/main/ts/keyboard/PageUpDownKeys.ts
camiloHimura: updated

modules/tinymce/src/core/test/ts/browser/keyboard/PageUpDownKeysTest.ts
camiloHimura: updated

modules/tinymce/src/core/main/ts/keyboard/PageUpDownKeys.ts
```diff
@@ -0,0 +1,76 @@
...
+import * as MatchKeys from './MatchKeys';
+
+const executeKeyupAction = (editor: Editor, evt: KeyboardEvent) => {
+  MatchKeys.execute([
+    { keyCode: VK.PAGE_UP, action: MatchKeys.action(InlineBoundariesNavigation.moveOutInlineBoundaries, editor, false) },
+    { keyCode: VK.PAGE_DOWN, action: MatchKeys.action(InlineBoundariesNavigation.moveOutInlineBoundaries, editor, true) }
+  ], evt);
+};
+
+const Blocker = (() => {
```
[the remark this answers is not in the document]
camiloHimura: Updated

modules/tinymce/src/core/test/ts/browser/keyboard/PageUpDownKeysTest.ts
```diff
@@ -50,7 +51,7 @@ describe('browser.tinymce.core.keyboard.PageUpDownKeyTest', () => {
   });
 
   context('Page Down', () => {
-    it('caret is at the beginning of an inline element, escape outside the node', () => {
+    it('Caret should be placed before the inline element if it is at the start of the inline element', () => {
```
spocke, line 54: Shouldn't these be after, Since you are moving down?

- [ ] Double check the test sentences

modules/tinymce/src/core/test/ts/browser/keyboard/PageUpDownKeysTest.ts
```diff
@@ -76,7 +77,7 @@ describe('browser.tinymce.core.keyboard.PageUpDownKeyTest', () => {
       }
     });
 
-    it('caret is not at the beginning/end of an inline element, not escape', () => {
+    it('Caret wont move if it is not at the beginning/end of the inline element', () => {
```
spocke, line 80: This is in the middle? It should move out if it ends up at the end of the link.

modules/tinymce/src/core/test/ts/browser/keyboard/HomeEndKeysTest.ts
```diff
@@ -57,7 +57,7 @@ describe('browser.tinymce.core.keyboard.HomeEndKeysTest', () => {
       it('move caret out and at the beginning of the element', () => {
```
spocke, line 57: Not sure what the consensus is now that we switched to mocha, But should this be uppercase at the beginning of the sensense?

modules/tinymce/src/core/main/ts/keyboard/InlineBoundariesNavigation.ts
```diff
@@ -17,53 +17,49 @@ const isInlineBoundaries = (editor: Editor) => (node: Node) => {
...
-  const newPost = forward ? CaretPosition.after : CaretPosition.before;
-  const move = (inline: HTMLElement) => BoundarySelection.setCaretPosition(editor, newPost(inline));
 
   const location = BoundaryLocation.readLocation(isInlineBoundaries(editor), body, pos);
   location.each((location) => {
     location.fold(
       Fun.noop,
-      move,
-      move,
+      moveOutside(editor, forward),
```
spocke, line 33: Nice cleanup there.

spocke (approved):

modules/tinymce/src/core/test/ts/browser/keyboard/HomeEndKeysTest.ts
lnewson: Our current standard is to always start with the JIRA number, so that really negates the need for it to start with a capital. eg:
```
it('TINY-4612: move caret out and at the beginning of the element',
```

modules/tinymce/src/core/test/ts/browser/keyboard/PageUpDownKeysTest.ts
camiloHimura: Done

modules/tinymce/src/core/test/ts/browser/keyboard/PageUpDownKeysTest.ts
camiloHimura: yes, it's in the middle, so it shouldn't move

modules/tinymce/src/core/test/ts/browser/keyboard/HomeEndKeysTest.ts
camiloHimura: updated

modules/tinymce/src/core/main/ts/keyboard/PageUpDownKeys.ts
```diff
@@ -0,0 +1,69 @@
...
+  if (block && !blocked.get()) {
+    editor.on('NodeChange', stopImmediatePropagation, true);
+  } else if (!block && blocked.get()) {
+    editor.off('NodeChange', stopImmediatePropagation);
+  }
+
+  blocked.set(block);
+};
+
+const setup = (editor: Editor) => {
```
lnewson, line 40: I mentioned this in a previous comment, but can you please add a comment here, stating why we are using `keyup` instead of `keydown` so we know in future why this decision was made (as it's inconsistent with the other overrides).

modules/tinymce/src/core/main/ts/keyboard/InlineBoundariesNavigation.ts
```diff
@@ -0,0 +1,66 @@
...
+  NavigationUtils.getLineEndPoint(editor, forward).exists((pos) => {
+    const location = BoundaryLocation.readLocation(isInlineBoundaries(editor), editor.getBody(), pos)
+      .orThunk(() => {
+        selfMoveOutside(editor, forward);
+        return Optional.none();
+      });
+
+    location.each((loc) => {
+      moveOutside(editor, forward)(BoundaryLocation.getElement(loc));
+    });
```
lnewson, line 58: This would be better done using a fold. eg:
```suggestion
    BoundaryLocation.readLocation(isInlineBoundaries(editor), editor.getBody(), pos).fold(
      () => selfMoveOutside(editor, forward),
      (loc) => moveOutside(editor, forward)(BoundaryLocation.getElement(loc))
    );
```

modules/tinymce/src/core/main/ts/keyboard/InlineBoundariesNavigation.ts
```diff
@@ -0,0 +1,66 @@
...
+
+    return location.isSome();
```
lnewson, line 60: This would be better done using a fold. eg:
```suggestion
    return BoundaryLocation.readLocation(isInlineBoundaries(editor), editor.getBody(), pos).fold(
      () => selfMoveOutside(editor, forward),
      (loc) => moveOutside(editor, forward)(BoundaryLocation.getElement(loc))
    );
```


Note: Doing the above also requires either returning `false` in the none branch above, or making `selfMoveOutside` return a boolean. I think the latter would probably be better, so it's consistent with `moveOutside`.

lnewson: Oops, sorry that's `moveOutInlineBoundaries` that returns a boolean. Either way, we'd be better using a `fold` here to better show the branching.

modules/tinymce/src/core/main/ts/keyboard/InlineBoundariesNavigation.ts
```diff
@@ -0,0 +1,66 @@
...
+import Editor from '../api/Editor';
+import CaretPosition from '../caret/CaretPosition';
+import * as BoundaryLocation from './BoundaryLocation';
+import * as BoundarySelection from './BoundarySelection';
+import * as InlineUtils from './InlineUtils';
+import * as NavigationUtils from './NavigationUtils';
+
+const isInlineBoundaries = (editor: Editor) => (node: Node) => {
+  return InlineUtils.isInlineTarget(editor, node);
+};
```
lnewson, line 18:
```suggestion
const isInlineBoundaries = (editor: Editor) => (node: Node) =>
  InlineUtils.isInlineTarget(editor, node);
```

lnewson: Another option that could be done here is:
```
const isInlineBoundaries = (editor: Editor) => Fun.curry(InlineUtils.isInlineTarget, editor);
```

modules/tinymce/src/core/main/ts/keyboard/PageUpDownKeys.ts
```diff
@@ -0,0 +1,69 @@
...
+    { keyCode: VK.PAGE_UP, action: MatchKeys.action(InlineBoundariesNavigation.moveOutInlineBoundaries, editor, false) },
+    { keyCode: VK.PAGE_DOWN, action: MatchKeys.action(InlineBoundariesNavigation.moveOutInlineBoundaries, editor, true) }
+  ], evt);
+};
+
+const stopImmediatePropagation = (e: EditorEvent<NodeChangeEvent>) => e.stopImmediatePropagation();
+
+const isPageUpDown = (evt: EditorEvent<KeyboardEvent>) => {
+  return evt.keyCode === VK.PAGE_UP || evt.keyCode === VK.PAGE_DOWN;
+};
```
lnewson, line 27:
```suggestion
const isPageUpDown = (evt: EditorEvent<KeyboardEvent>) =>
  evt.keyCode === VK.PAGE_UP || evt.keyCode === VK.PAGE_DOWN;
```
We prefer shorthand arrow functions where possible generally. It's only if doing that makes it harder to read that we use the longhand format.

modules/tinymce/src/core/main/ts/keyboard/PageUpDownKeys.ts
```diff
@@ -0,0 +1,69 @@
...
+  });
+
+  editor.on('keyup', (evt) => {
+    if (isPageUpDown(evt)) {
+      setNodeChangeBlocker(blocked, editor, false);
+      editor.nodeChanged();
+    }
+  });
+
+  editor.on('keyup', (evt) => {
```
lnewson, line 60: This shouldn't be 2 different bindings, so can you please merge them into a single event binding?

modules/tinymce/src/core/main/ts/keyboard/InlineBoundariesNavigation.ts
lnewson: Actually, thinking about this more... One of the following might be a better implementation:
```
const moveToLineEndPoint = (editor: Editor, forward: boolean): boolean => {
  const location = NavigationUtils.getLineEndPoint(editor, forward).bind((pos) =>
    BoundaryLocation.readLocation(isInlineBoundaries(editor), editor.getBody(), pos)
  );
  return location.fold(
    () => selfMoveOutside(editor, forward),
    (loc) => moveOutside(editor, forward)(BoundaryLocation.getElement(loc))
  );
};
```
or
```
const moveToLineEndPoint = (editor: Editor, forward: boolean): boolean => {
  const location = NavigationUtils.getLineEndPoint(editor, forward).bind((pos) =>
    BoundaryLocation.readLocation(isInlineBoundaries(editor), editor.getBody(), pos)
  );
  return location.fold(() => {
    selfMoveOutside(editor, forward);
    return false;
  }, (loc) => { 
    moveOutside(editor, forward)(BoundaryLocation.getElement(loc));
    return true;
  });
};
```

modules/tinymce/src/core/main/ts/keyboard/PageUpDownKeys.ts
```diff
@@ -0,0 +1,69 @@
...
+  editor.on('keydown', (evt) => {
+    if (isPageUpDown(evt)) {
+      setNodeChangeBlocker(blocked, editor, true);
+    }
+  });
+
+  editor.on('keyup', (evt) => {
+    if (isPageUpDown(evt)) {
+      setNodeChangeBlocker(blocked, editor, false);
+      editor.nodeChanged();
```
lnewson, line 56: Is this correct? I believe it should only fire the `NodeChanged` here if it was previously blocking.

modules/tinymce/src/core/test/ts/browser/keyboard/PageUpDownKeysTest.ts
```diff
@@ -0,0 +1,88 @@
...
+      TinyContentActions.keyup(editor, Keys.pageUp());
+
+      if (platform.browser.isIE()) {
+        TinyAssertions.assertCursor(editor, [ 0, 0 ], 4);
+      } else {
+        TinyAssertions.assertCursor(editor, [ 0 ], 1);
+      }
+    });
+
+    it('TINY-4612: caret should be placed before the inline element if it is at the end of it', () => {
```
lnewson, line 31: This text and also the behaviour by the looks isn't correct, as the link is in the middle of the content. So in this case, pageup shouldn't be doing anything and should be falling back to the browser behaviour (ie should be before `text`).

modules/tinymce/src/core/main/ts/keyboard/PageUpDownKeys.ts
```diff
@@ -0,0 +1,69 @@
...
+  ], evt);
+};
+
+const stopImmediatePropagation = (e: EditorEvent<NodeChangeEvent>) => e.stopImmediatePropagation();
+
+const isPageUpDown = (evt: EditorEvent<KeyboardEvent>) => {
+  return evt.keyCode === VK.PAGE_UP || evt.keyCode === VK.PAGE_DOWN;
+};
+
+const setNodeChangeBlocker = (blocked: Cell<boolean>, editor: Editor, block: boolean) => {
```
lnewson, line 29: This logic doesn't appear to be working when testing locally, as when pressing page down I'm still seeing a blue flash whereby the inline boundary applies quickly and then is removed.

I haven't had a chance to debug it, but that's not ideal and we should see if we can fix it.

modules/tinymce/src/core/test/ts/browser/keyboard/PageUpDownKeysTest.ts
```diff
@@ -0,0 +1,88 @@
+import { Keys } from '@ephox/agar';
+import { context, describe, it } from '@ephox/bedrock-client';
+import { TinyAssertions, TinyContentActions, TinyHooks, TinySelections } from '@ephox/mcagar';
+import { PlatformDetection } from '@ephox/sand';
+
+import Editor from 'tinymce/core/api/Editor';
+import Theme from 'tinymce/themes/silver/Theme';
+
+describe('browser.tinymce.core.keyboard.PageUpDownKeyTest', () => {
```
lnewson, line 9: Given that this logic relies on actual browsers doing something and then us overriding that logic, we really shouldn't be using a simulated test here. This instead should be a webdriver test so that actual keyevents are fired and operated by the browser.

I'd suggest asking for help on how to do that, but the gist is that you'll need to:
- move this to the `webdriver` directory
- update the `describe` to say `webdriver` instead of `browser`
- use `RealKeys` from agar instead of using `TinyContentActions` to make webdriver fire real key events

modules/tinymce/src/core/test/ts/browser/keyboard/HomeEndKeysTest.ts
```diff
@@ -53,6 +53,23 @@ describe('browser.tinymce.core.keyboard.HomeEndKeysTest', () => {
       TinyAssertions.assertSelection(editor, [ 0, 0 ], 1, [ 0, 0 ], 1);
     });
 
+    context('Inline element', () => {
```
lnewson, line 56: These aren't really inline elements, they are `Inline boundary elements` so maybe something like this?
```suggestion
    context('Inline boundaries', () => {
```

lnewson (changes requested):
Looking good overall, but there are still issues with the node change activating which is causing flashes of unwanted styling, etc...

Additionally, there are a few test labels that don't match what's being tested and as noted in at least one case the assertions aren't correct. Also as noted in another comment, since we're relying on browser behaviour for page/up down, we'll need to use a webdriver test for that.

lnewson (changes requested):
Another thing I just noticed is that you have no tests for `inline_boundaries: false`, so please make sure to add tests for that as well.

You can test that by just doing `editor.settings.inline_boundaries = false;` inside your test and then deleting the setting in the cleanup.

modules/tinymce/src/core/main/ts/keyboard/PageUpDownKeys.ts
lnewson: Okay, I looked into this and it's because you're calling `editor.nodeChanged()` before moving the selection. If you update the below to be something like this, then it should resolve that:
```
  editor.on('keyup', (evt) => {
    if (evt.isDefaultPrevented() === false) {
      executeKeyupAction(editor, evt);
    }
    if (isPageUpDown(evt)) {
      setNodeChangeBlocker(blocked, editor, false);
      editor.nodeChanged();
    }
  });
```

modules/tinymce/src/core/main/ts/keyboard/PageUpDownKeys.ts
camiloHimura: updated

modules/tinymce/src/core/main/ts/keyboard/PageUpDownKeys.ts
camiloHimura: Done

modules/tinymce/src/core/main/ts/keyboard/PageUpDownKeys.ts
camiloHimura: Updated

modules/tinymce/src/core/main/ts/keyboard/PageUpDownKeys.ts
camiloHimura: updated

modules/tinymce/src/core/main/ts/keyboard/InlineBoundariesNavigation.ts
```diff
@@ -5,65 +5,56 @@
...
-  const newPost = forward ? CaretPosition.after : CaretPosition.before;
-  const move = (inline: HTMLElement) => BoundarySelection.setCaretPosition(editor, newPost(inline));
 
   const location = BoundaryLocation.readLocation(isInlineBoundaries(editor), body, pos);
   location.each((location) => {
     location.fold(
       Fun.noop,
-      move,
-      move,
+      moveOutside(editor, forward),
```
spocke, line 31: You could Fun.curry this on as well maybe have a `const move = Fun.curry(moveOutside, editor, forward);`' that way the moveOutside wouldn't need to be forced curried and you wouldn't need the strange `moveOutside(editor, forward)(node);` call it could just be `moveOutside(editor, forward, node);`.

modules/tinymce/src/core/test/ts/webdriver/PageUpDownKeysTest.ts
```diff
@@ -0,0 +1,98 @@
...
+      TinyContentActions.keyup(editor, Keys.pageUp());
+      await RealKeys.pSendKeysOn('iframe => body', [ RealKeys.combo({}, 'PageUp') ]);
+
+      TinyAssertions.assertCursor(editor, [ 0 ], 0);
+    });
+
+    it('TINY-4612: caret should be placed before the inline element if it is at the end of it', async () => {
+      const editor = hook.editor();
+      editor.setContent('<p><a href="google.com">link</a>text</p>');
+      const offset = platform.browser.isIE() ? 5 : 4;
```
spocke, line 31: Curious why is this platform detection needed is IE rendering the zwnbsp character while other browsers don't? Since the text has only 4 chars I assume that there is extra zwnbsp char before the `link` word. So add a comment why this was needed.

modules/tinymce/src/core/test/ts/webdriver/PageUpDownKeysTest.ts
```diff
@@ -0,0 +1,98 @@
...
+    });
+
+    it('TINY-4612: caret wont move inline_boundaries: false', async () => {
+      const editor = hook.editor();
+      editor.settings.inline_boundaries = false;
+      editor.setContent('<p>text<a href="google.com">link</a>link</p>');
+      TinySelections.setCursor(editor, [ 0, 1, 0 ], 0);
+      await RealKeys.pSendKeysOn('iframe => body', [ RealKeys.combo({}, 'PageUp') ]);
+
+      TinySelections.setCursor(editor, [ 0, 1, 0 ], 0);
```
spocke, line 54: I think this should be an assertCursor.

- [ ] Change to a proper assert

modules/tinymce/src/core/test/ts/webdriver/PageUpDownKeysTest.ts
```diff
@@ -0,0 +1,98 @@
...
+    });
+
+    it('TINY-4612: caret wont move inline_boundaries: false', async () => {
+      const editor = hook.editor();
+      editor.settings.inline_boundaries = false;
+      editor.setContent('<p><a href="google.com">link</a>text</p>');
+      TinySelections.setCursor(editor, [ 0, 0, 0 ], 0);
+      await RealKeys.pSendKeysOn('iframe => body', [ RealKeys.combo({}, 'PageDown') ]);
+
+      TinySelections.setCursor(editor, [ 0, 0, 0 ], 0);
```
spocke, line 94: This should also be an assertCursor.

- [ ] Change to assertCursor

spocke (approved):

modules/tinymce/src/core/test/ts/webdriver/PageUpDownKeysTest.ts
JamesToohey: Un-resolving at this hasn't been changed

modules/tinymce/src/core/test/ts/webdriver/PageUpDownKeysTest.ts
JamesToohey: Un-resolving at this hasn't been changed

modules/tinymce/src/core/test/ts/webdriver/PageUpDownKeysTest.ts
```diff
@@ -0,0 +1,107 @@
...
+      TinyAssertions.assertCursor(editor, [ 0 ], 2);
+    });
+
+    it('TINY-4612: caret wont move if it is not at the beginning/end of the inline element', async () => {
+      const editor = hook.editor();
+      editor.setContent('<p><a href="google.com">link</a>text</p>');
+      TinySelections.setCursor(editor, [ 0, 0, 0 ], 2);
+      await RealKeys.pSendKeysOn('iframe => body', [ RealKeys.combo({}, 'PageDown') ]);
+
+      TinySelections.setCursor(editor, [ 0, 0, 0 ], 2);
```
JamesToohey, line 89: Should this also be `assertCursor`?

modules/tinymce/src/core/test/ts/webdriver/PageUpDownKeysTest.ts
camiloHimura: sorry, I reset back some changes and missed this

modules/tinymce/src/core/test/ts/webdriver/PageUpDownKeysTest.ts
camiloHimura: done

modules/tinymce/src/core/test/ts/webdriver/PageUpDownKeysTest.ts
camiloHimura: Updated

modules/tinymce/src/core/main/ts/keyboard/PageUpDownKeys.ts
lnewson: This has not been done, all you did was move the previous comment which only explains the `NodeChange` prevention. It has nothing about keydown/keyup.

modules/tinymce/src/core/test/ts/browser/keyboard/PageUpDownKeysTest.ts
lnewson: Just leaving another comment to make sure it hasn't been lost. This doesn't appear to have been addressed yet, as the tests still have the wrong assertions. Remember that Page Up/Down should go to the start/end of the line in the viewport not to the start of the anchor, when the anchor is in the middle of a line.

Also worth noting, that @JamesToohey found out yesterday that Page Up/Down on Mac appears to scroll the viewport instead of change the caret position/selection.

camiloHimura: The current behavior for PageUp/PageDown is to move the caret outside the inline element if it is at the start/end of it without overwriting default behavior, not move the caret at the start/end of the line (that it's a browser behavior that only happens on Windows os)

lnewson: And that's what I'm saying is incorrect behaviour. It should **never** move it outside the anchor unless the anchor is at the start/end of the line. Also, the behaviour I mentioned happens on every OS except macOS in our testing, so saying it only applies to Windows would be incorrect.

modules/tinymce/src/core/test/ts/webdriver/PageUpDownKeysTest.ts
```diff
@@ -0,0 +1,115 @@
...
+        this.skip();
+      }
+
+      const editor = hook.editor();
+      editor.settings.inline_boundaries = false;
+      editor.setContent('<p><a href="google.com">link</a>text</p>');
+      TinySelections.setCursor(editor, [ 0, 0, 0 ], 0);
+      await RealKeys.pSendKeysOn('iframe => body', [ RealKeys.combo({}, 'PageDown') ]);
+
+      TinyAssertions.assertCursor(editor, [ 0, 0, 0 ], 0);
```
lnewson, line 111: @camiloHimura here is one part related to what I mentioned in the other comment. This is pressing page down, so it shouldn't be staying in the same location. It should be getting moved to `[ 0, 1 ], 4`. It's possible this may just be webdriver not working, however this assertion is incorrect.

lnewson: To clarify in case the above wasn't clear, it should be at the end of the line, after `text` (with macOS likely being the exception since as mentioned previously it scrolls the viewport).

modules/tinymce/src/core/main/ts/keyboard/PageUpDownKeys.ts
lnewson: This description is not quite correct sorry. See Spocke's previous explanation for why we aren't doing this when I brought it up. eg determining the correct placement on keydown is very complicated and would require handling many edge cases, which we don't have the resources to handle currently. As such we allow the browser to change the selection and then make adjustments later.

You've also now lost the reason we had the `NodeChange` override as well.

modules/tinymce/src/core/test/ts/webdriver/PageUpDownKeysTest.ts
```diff
@@ -0,0 +1,96 @@
...
+      if (!platform.os.isOSX()) {
+        this.skip();
+      }
+
+      const editor = hook.editor();
+      editor.setContent('<p>text<a href="google.com">link2</a></p>');
+      TinySelections.setCursor(editor, [ 0, 0 ], 4);
+      await RealKeys.pSendKeysOn('iframe => body', [ RealKeys.combo({}, 'PageUp') ]);
+
+      TinyAssertions.assertCursor(editor, [ 0, 0 ], 4);
```
lnewson, line 38: This assertion is incorrect, because as mentioned previously this should be at the start of the line.

modules/tinymce/src/core/test/ts/webdriver/PageUpDownKeysTest.ts
```diff
@@ -0,0 +1,96 @@
...
+        this.skip();
+      }
+
+      const editor = hook.editor();
+      editor.settings.inline_boundaries = false;
+      editor.setContent('<p><a href="google.com">link</a>text</p>');
+      TinySelections.setCursor(editor, [ 0, 1 ], 2);
+      await RealKeys.pSendKeysOn('iframe => body', [ RealKeys.combo({}, 'PageUp') ]);
+
+      TinyAssertions.assertCursor(editor, [ 0, 1 ], 2);
```
lnewson, line 52: This assertion is also not correct for the same reason as the previous one.

modules/tinymce/src/core/test/ts/webdriver/PageUpDownKeysTest.ts
```diff
@@ -0,0 +1,96 @@
...
+      const editor = hook.editor();
+      editor.setContent('<p>text<a href="google.com">link</a></p>');
+      TinySelections.setCursor(editor, [ 0, 0 ], 0);
+      await RealKeys.pSendKeysOn('iframe => body', [ RealKeys.combo({}, 'PageDown') ]);
+
+      TinyAssertions.assertCursor(editor, [ 0 ], 2);
+    });
+
+    it('TINY-4612: caret wont move if it is not at inline element at the end of the line', async function () {
+      // Windows os default behavior (regardless the browser) is moving the caret at the end of the line
```
lnewson, line 68: This is exactly the behaviour we want to test, so this shouldn't be getting skipped.

modules/tinymce/src/core/test/ts/webdriver/PageUpDownKeysTest.ts
```diff
@@ -0,0 +1,96 @@
...
+      if (platform.os.isWindows()) {
+        this.skip();
+      }
+
+      const editor = hook.editor();
+      editor.setContent('<p><a href="google.com">link</a>text</p>');
+      TinySelections.setCursor(editor, [ 0, 1 ], 2);
+      await RealKeys.pSendKeysOn('iframe => body', [ RealKeys.combo({}, 'PageDown') ]);
+
+      TinyAssertions.assertCursor(editor, [ 0, 1 ], 2);
```
lnewson, line 78: This is also not an expected/correct assertion, as this should be at the end of the line. This is likely passing again due to it running on macOS which scrolls instead.

modules/tinymce/src/core/test/ts/webdriver/PageUpDownKeysTest.ts
```diff
@@ -0,0 +1,96 @@
...
+        this.skip();
+      }
+
+      const editor = hook.editor();
+      editor.settings.inline_boundaries = false;
+      editor.setContent('<p>text<a href="google.com">link</a></p>');
+      TinySelections.setCursor(editor, [ 0, 0 ], 0);
+      await RealKeys.pSendKeysOn('iframe => body', [ RealKeys.combo({}, 'PageDown') ]);
+
+      TinyAssertions.assertCursor(editor, [ 0, 0 ], 0);
```
lnewson, line 92: Another assertion that's not what should be happening, similar to the above.

modules/tinymce/src/core/main/ts/keyboard/PageUpDownKeys.ts
```diff
@@ -0,0 +1,64 @@
...
+import Editor from '../api/Editor';
+import { NodeChangeEvent } from '../api/EventTypes';
+import { EditorEvent } from '../api/PublicApi';
+import VK from '../api/util/VK';
+import * as InlineBoundariesNavigation from './InlineBoundariesNavigation';
+import * as MatchKeys from './MatchKeys';
+
+const executeKeyupAction = (editor: Editor, evt: KeyboardEvent) => {
+  MatchKeys.execute([
+    { keyCode: VK.PAGE_UP, action: MatchKeys.action(InlineBoundariesNavigation.moveToLineEndPoint, editor, false) },
```
lnewson, line 18: Should this only happen on non macOS platforms? because as mentioned previously macOS causes the page to scroll instead of change the selection. @spocke would be keen to get your thoughts here?

modules/tinymce/src/core/main/ts/keyboard/InlineBoundariesNavigation.ts
```diff
@@ -0,0 +1,51 @@
...
+import * as Settings from '../api/Settings';
+import CaretPosition from '../caret/CaretPosition';
+import * as BoundaryLocation from './BoundaryLocation';
+import * as BoundarySelection from './BoundarySelection';
+import * as InlineUtils from './InlineUtils';
+import * as NavigationUtils from './NavigationUtils';
+
+const isInlineBoundaries = (editor: Editor, node: Node) => {
+  return Settings.isInlineBoundariesEnabled(editor) && InlineUtils.isInlineTarget(editor, node);
+};
```
lnewson, line 19:
```suggestion
const isInlineBoundaries = (editor: Editor, node: Node) =>
  Settings.isInlineBoundariesEnabled(editor) && InlineUtils.isInlineTarget(editor, node);
```

lnewson (changes requested):
Getting much closer @camiloHimura :+1: However, there are still some incorrect assertions that need to be fixed up before I can approve this sorry.

Also, don't forget to include a changelog entry as I can't approve it until that's also included.

modules/tinymce/src/core/test/ts/webdriver/PageUpDownKeysTest.ts
camiloHimura: this assertion applies only to Mac Os, as you mentioned before the caret won move only scroll the page

modules/tinymce/src/core/main/ts/keyboard/PageUpDownKeys.ts
spocke: The home/end keys seems to move to start/end of document from some stackoverflow post I found. So we need to make sure they are platform specific. Also Mac OS seems to do a real home/end of line if you use fn-arrow left/right and fn-arrow up/down is page up.

I'm unsure if webdriver is automatically remapping these or if these are the same key code and modifier just a way to access them.

So I guess we need to investigate what behavior is the user expecting for real home/end/pg down/pg up keys and the fn-arrow combos are they the same thing, are mac people likely to use, them can we detect all combos if they are different?

lnewson: Sorry, you've probably missed some of the previous discussion. Yes, Fn+Up/Down are page up/down (we get keycode 33/34 in the keydown/keyup events). On a real MacBook pro, using contenteditable, the browser will always scroll instead of moving the selection so I'm thinking we need to keep that behaviour here as well (worth noting it does the same thing in the Pages app).

modules/tinymce/src/core/test/ts/webdriver/PageUpDownKeysTest.ts
lnewson: Worth noting it'd probably be best to branch the assertions and assert the relevant state (ie macOS doesn't move, whereas other OS's do). At the moment throughout this test we're just completely skipping instead which isn't ideal.

modules/tinymce/src/core/main/ts/keyboard/PageUpDownKeys.ts
lnewson: Reading and testing https://developer.apple.com/design/human-interface-guidelines/macos/user-interaction/keyboard/#reserved-and-expected-keyboard-shortcuts, it seems Cmd+Up or Cmd+Down is the closest equivalent to Page Up/Down behaviour in other OS's. As such, we might need to handle that key combination instead of Page Up/Down for macOS.

I should not the above goes to the start/end of the document instead of to the start/end of what's inside the viewport.

modules/tinymce/src/core/test/ts/webdriver/PageUpDownKeysTest.ts
camiloHimura: updated

modules/tinymce/src/core/test/ts/webdriver/PageUpDownKeysTest.ts
camiloHimura: updated

modules/tinymce/src/core/test/ts/webdriver/PageUpDownKeysTest.ts
camiloHimura: updated

modules/tinymce/src/core/test/ts/webdriver/PageUpDownKeysTest.ts
camiloHimura: updated

modules/tinymce/src/core/test/ts/webdriver/PageUpDownKeysTest.ts
camiloHimura: updated

modules/tinymce/src/core/main/ts/keyboard/PageUpDownKeys.ts
camiloHimura: updated

modules/tinymce/src/core/main/ts/keyboard/InlineBoundariesNavigation.ts
camiloHimura: updated

modules/tinymce/src/core/main/ts/keyboard/CommandUpDownKeys.ts
```diff
@@ -0,0 +1,77 @@
...
+  MatchKeys.execute([
+    { keyCode: VK.UP, metaKey: platform.os.isOSX(), action: MatchKeys.action(InlineBoundariesNavigation.moveToLineEndPoint, editor, false) },
+    { keyCode: VK.DOWN, metaKey: platform.os.isOSX(), action: MatchKeys.action(InlineBoundariesNavigation.moveToLineEndPoint, editor, true) }
+  ], evt);
+};
+
+const stopImmediatePropagation = (e: EditorEvent<NodeChangeEvent>) => e.stopImmediatePropagation();
+
+const isCommandUpDown = (evt: EditorEvent<KeyboardEvent>) =>
+  (evt.keyCode === VK.UP && evt.metaKey === true) ||
```
spocke, line 30: This could be reduced to:
```
const isCommandUpDown = (evt: EditorEvent<KeyboardEvent>) => evt.metaKey && (evt.keyCode === VK.UP || evt.keyCode === VK.DOWN);
```

modules/tinymce/src/core/test/ts/webdriver/PageUpDownKeysTest.ts
camiloHimura: Updated

modules/tinymce/src/core/main/ts/keyboard/CommandUpDownKeys.ts
camiloHimura: this file was removed

modules/tinymce/src/core/main/ts/keyboard/PageUpDownKeys.ts
lnewson: @camiloHimura in that case we need to not run this logic on Mac OS, otherwise it's potentially going to override the page up/down on Mac OS.

modules/tinymce/src/core/main/ts/keyboard/PageUpDownKeys.ts
```diff
@@ -0,0 +1,72 @@
...
+    { keyCode: VK.UP, metaKey: platform.os.isOSX(), action: MatchKeys.action(InlineBoundariesNavigation.moveToLineEndPoint, editor, false) },
+    { keyCode: VK.DOWN, metaKey: platform.os.isOSX(), action: MatchKeys.action(InlineBoundariesNavigation.moveToLineEndPoint, editor, true) }
+  ], evt);
+};
+
+const stopImmediatePropagation = (e: EditorEvent<NodeChangeEvent>) => e.stopImmediatePropagation();
+
+const isPageUpDown = (evt: EditorEvent<KeyboardEvent>) =>
+  evt.keyCode === VK.PAGE_UP || evt.keyCode === VK.PAGE_DOWN ||
+  (evt.keyCode === VK.UP && evt.metaKey === true) ||
```
lnewson, line 32: It was mentioned in another comment this logic wasn't added? It appears it was half added, so we probably should remove this.

modules/tinymce/src/core/main/ts/keyboard/PageUpDownKeys.ts
lnewson: I think the best option will be to just return immediately in the `setup` function for macOS. Thoughts?

modules/tinymce/src/core/main/ts/keyboard/PageUpDownKeys.ts
camiloHimura: Removed

modules/tinymce/src/core/main/ts/keyboard/PageUpDownKeys.ts
camiloHimura: I've been testing this implementation on Mac, and the default behavior ( scroll up and down) is not overridden. I don't think it is needed to skip the implementation on Mac explicitly.

lnewson: That's not what would break here, it's that it'd scroll **and** change the selection if it was at the start of an inline boundary. We never want this behaviour to run on Mac, as it doesn't have the same page up/down behaviour and as such this code should never be running on Mac.

modules/tinymce/src/core/main/ts/keyboard/InlineBoundariesNavigation.ts
lnewson: This hasn't been updated...

modules/tinymce/src/core/test/ts/webdriver/PageUpDownKeysTest.ts
```diff
@@ -0,0 +1,102 @@
...
+    add_unload_trigger: false,
+    base_url: '/project/tinymce/js/tinymce',
+    indent: false
+  }, [ Theme ], true);
+  const platform = PlatformDetection.detect();
+
+  // It necessary to add a different assertion for Firefox because it does not move the caret start/end on OS windows and Mac,
+  // Linux Firefox could be an exception, nevertheless we does not override the default behavior.
+  context('Page Up', () => {
+    it('TINY-4612: caret should be placed at the start of the line if the first child is an inline element', async () => {
```
lnewson, line 20: This matches what I said in my other comment, but this should **not** be happening on Mac, as it doesn't move the selection when pressing page up/down and as such TinyMCE shouldn't be moving it either.

modules/tinymce/src/core/test/ts/webdriver/PageUpDownKeysTest.ts
```diff
@@ -0,0 +1,102 @@
...
+    });
+
+    it('TINY-4612: macOS wont move the caret if it is not at inline element at the start of the line, whereas other OSs do', async () => {
+      const editor = hook.editor();
+      editor.setContent('<p>text<a href="google.com">link2</a>text</p>');
+      TinySelections.setCursor(editor, [ 0, 2 ], 0);
+      await RealKeys.pSendKeysOn('iframe => body', [ RealKeys.combo({}, 'PageUp') ]);
+      TinyContentActions.keystroke(editor, Keys.pageUp());
+
+      if (platform.os.isOSX() || platform.browser.isFirefox()) {
```
lnewson, line 37: As mentioned via DM, this and the other checks in this file needs to be a "Windows Firefox" check, not a "Firefox" check as this test will fail locally on Linux since it does work the same as other browsers.

modules/tinymce/src/core/test/ts/webdriver/PageUpDownKeysTest.ts
```diff
@@ -0,0 +1,102 @@
...
+        TinyAssertions.assertCursor(editor, [ 0, 2 ], 2);
+      } else {
+        TinyAssertions.assertCursor(editor, [ 0, 0 ], 0);
+      }
+      delete editor.settings.inline_boundaries;
+    });
+  });
+
+  context('Page Down', () => {
+    it('TINY-4612: caret should be placed at the end of the line if the last child is an inline element', async () => {
```
lnewson, line 62: Likewise, this shouldn't be happening on Mac, as page up/down should only be changing the scroll position.

modules/tinymce/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Added
+- Added support for `inline_boundaries` to work with home/end/pgup/pgdown keys
```
lnewson, line 10: Please include the JIRA number at the end of the sentence like the other entries.

modules/tinymce/src/core/test/ts/webdriver/PageUpDownKeysTest.ts
```diff
@@ -0,0 +1,102 @@
...
+      TinyContentActions.keystroke(editor, Keys.pageUp());
+
+      if (platform.os.isOSX() || platform.browser.isFirefox()) {
+        TinyAssertions.assertCursor(editor, [ 0, 2 ], 0);
+      } else {
+        TinyAssertions.assertCursor(editor, [ 0, 0 ], 0);
+      }
+    });
+
+    it('TINY-4612: inline_boundaries: false macOS doesnt move, whereas other OSs do', async () => {
```
lnewson, line 44:
```suggestion
    it('TINY-4612: "inline_boundaries: false" macOS doesnt move, whereas other OSs do', async () => {
```

modules/tinymce/src/core/test/ts/webdriver/PageUpDownKeysTest.ts
```diff
@@ -0,0 +1,102 @@
...
+      TinyContentActions.keystroke(editor, Keys.pageDown());
+
+      if (platform.os.isOSX() || platform.browser.isFirefox()) {
+        TinyAssertions.assertCursor(editor, [ 0, 0 ], 0);
+      } else {
+        TinyAssertions.assertCursor(editor, [ 0, 2 ], 4);
+      }
+    });
+
+    it('TINY-4612: inline_boundaries: false macOS doesnt move, whereas other OSs do', async () => {
```
lnewson, line 86:
```suggestion
    it('TINY-4612: "inline_boundaries: false" macOS doesnt move, whereas other OSs do', async () => {
```

modules/tinymce/src/core/test/ts/webdriver/PageUpDownKeysTest.ts
lnewson: It might also be good to move this check out to a separate variable or function to remove the duplication. eg:
```
const platform = PlatformDetection.detect();
const supportsPageUpDown = !(platform.os.isOSX() || platform.os.isWindows() && platform.browser.isFirefox());

**Review by lnewson — CHANGES_REQUESTED** [review]
You're nearly there, just a few more things that need to be fixed sorry.

File: modules/tinymce/CHANGELOG.md (hunk 46 above)
```suggestion
- The `inline_boundaries` feature now supports the `home`, `end`, `pageup` and `pgdown` keys #TINY-4612
```

modules/tinymce/src/core/main/ts/keyboard/PageUpDownKeys.ts
camiloHimura: Ok, in that case, your idea of immediate return on the `setup` is fine.

modules/tinymce/src/core/test/ts/webdriver/PageUpDownKeysTest.ts
camiloHimura: I'll skip assertions on mac and windows firefox, updated

modules/tinymce/src/core/test/ts/webdriver/PageUpDownKeysTest.ts
camiloHimura: updated

modules/tinymce/CHANGELOG.md

modules/tinymce/src/core/test/ts/webdriver/PageUpDownKeysTest.ts
camiloHimura: All assertions related to "inline boundaries: false" were removed because Windows and Linux browsers always move the caret with PageUp/PageDown, and on Mac, these tests are skipped.

lnewson: These shouldn't have been removed, as they assert the selection isn't moved to the incorrect location by the new logic.

modules/tinymce/src/core/test/ts/webdriver/PageUpDownKeysTest.ts
```diff
@@ -0,0 +1,70 @@
...
+    add_unload_trigger: false,
+    base_url: '/project/tinymce/js/tinymce',
+    indent: false
+  }, [ Theme ], true);
+  const platform = PlatformDetection.detect();
+  const supportsPageUpDown = !(platform.os.isOSX() || platform.os.isWindows() && platform.browser.isFirefox());
+
+  // It necessary to skip tests for Mac and windows Firefox they don't move the selection when pressing page up/down
+  context('Page Up', () => {
+    before(function () {
```
lnewson, line 20: This could be moved outside the `context` blocks and that way you wouldn't need to duplicate it. eg:
```
  // It's necessary to skip tests for Mac and Windows Firefox as they don't move the selection when pressing page up/down
  before(function () {
    if (!supportsPageUpDown) {
      this.skip();
    }
  });

  context('Page Up', () => {
    ...
  });

  context('Page Down', () => {
    ...
  });
```

Note: I also fixed some minor grammar issues in your comment in the above.

lnewson (changes requested):

modules/tinymce/src/core/test/ts/webdriver/PageUpDownKeysTest.ts
camiloHimura: updated

modules/tinymce/src/core/test/ts/webdriver/PageUpDownKeysTest.ts
camiloHimura: updated

modules/tinymce/src/core/test/ts/webdriver/PageUpDownKeysTest.ts
```diff
@@ -0,0 +1,88 @@
...
+    it('TINY-4612: caret should be placed out of the line element', async () => {
+      const editor = hook.editor();
+      editor.setContent('<p><a href="google.com">link</a></p>');
+      TinySelections.setCursor(editor, [ 0, 0, 0 ], 2);
+      await RealKeys.pSendKeysOn('iframe => body', [ RealKeys.combo({}, 'PageDown') ]);
+
+      TinyAssertions.assertCursor(editor, [ 0 ], 1);
+    });
+
+    it('TINY-4612: "inline_boundaries: false" macOS doesnt move, whereas other OSs do', async () => {
```
lnewson, line 76: This test doesn't test what's needed and the title doesn't match what's being tested sorry. What we need to test is that the selection doesn't get moved outside the anchor by our custom code when inline boundaries are disabled.

As such, it probably needs to be something like this:
```
    it('TINY-4612: "inline_boundaries: false" cursor does not move outside anchor', async () => {
      const editor = hook.editor();
      editor.settings.height = 400;
      editor.settings.inline_boundaries = false;
      editor.setContent('<p>test<a href="google.com">link</a></p>');
      TinySelections.setCursor(editor, [ 0, 0 ], 0);
      await RealKeys.pSendKeysOn('iframe => body', [ RealKeys.combo({}, 'PageDown') ]);

      TinyAssertions.assertCursor(editor, [ 0, 1, 0 ], 4);
      delete editor.settings.inline_boundaries;
    });
```

You'll see in the above that we make sure a link is at the end and that the placement is where the browser moved it to (eg inside the anchor).

This issue also exists in the equivalent `Page Up` test, so please be sure to update it to do a similar thing as well.

lnewson (changes requested):
Just one last issue sorry @camiloHimura and then we should be done :+1:

modules/tinymce/src/core/test/ts/webdriver/PageUpDownKeysTest.ts
```diff
@@ -0,0 +1,98 @@
...
+    it('TINY-4612: caret should be placed out of the line element', async () => {
+      const editor = hook.editor();
+      editor.setContent('<p><a href="google.com">link</a></p>');
+      TinySelections.setCursor(editor, [ 0, 0, 0 ], 2);
+      await RealKeys.pSendKeysOn('iframe => body', [ RealKeys.combo({}, 'PageUp') ]);
+
+      TinyAssertions.assertCursor(editor, [ 0 ], 0);
+    });
+
+    it('TINY-4612: "inline_boundaries: false" macOS doesnt move, whereas other OSs do', async () => {
```
lnewson, line 44: This test title also needs updating.
```suggestion
    it('TINY-4612: "inline_boundaries: false" cursor does not move outside anchor', async () => {
```

modules/tinymce/src/core/test/ts/webdriver/PageUpDownKeysTest.ts
```diff
@@ -0,0 +1,98 @@
...
+      const editor = hook.editor();
+      editor.settings.height = 400;
+      editor.settings.inline_boundaries = false;
+      editor.setContent('<p><a href="google.com">link</a>text</p>');
+      TinySelections.setCursor(editor, [ 0, 1 ], 4);
+      await RealKeys.pSendKeysOn('iframe => body', [ RealKeys.combo({}, 'PageUp') ]);
+
+      if (platform.browser.isIE()) {
```
lnewson, line 52: Oh fun, of course IE had to be different :cry:

lnewson (approved):

modules/tinymce/src/core/test/ts/webdriver/PageUpDownKeysTest.ts
camiloHimura: updated

ashfordneil (approved):

modules/agar/src/main/ts/ephox/agar/api/Keys.ts
```diff
@@ -12,5 +12,7 @@ export const Keys = {
   down: Fun.constant(40),
   space: Fun.constant(32),
   home: Fun.constant(36),
-  end: Fun.constant(35)
+  end: Fun.constant(35),
+  pageUp: Fun.constant(33),
```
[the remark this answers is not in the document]
camiloHimura: updated

modules/tinymce/src/core/main/ts/api/util/VK.ts
```diff
@@ -24,6 +24,8 @@ interface VK {
   SPACEBAR: number;
   TAB: number;
   UP: number;
+  PAGE_UP: number;
```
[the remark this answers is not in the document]
camiloHimura: updated

metricjs (approved):

modules/tinymce/CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Added
+- The `inline_boundaries` feature now supports the `home`, `end`, `pageup` and `pagedown` keys #TINY-4612
+- Added a new KeyCodes `PAGE_UP` and `PAGE_DOWN` #TINY-4612
```
```suggestion
- Added new `PAGE_UP` and `PAGE_DOWN` key code constants to the `VK` API #TINY-4612
```

modules/agar/CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+- Added new keys pageUp and pageDown.
```
lnewson, line 8:
```suggestion
- Added `pageUp` and `pageDown` to the `Keys` API.
```

modules/tinymce/src/core/test/ts/webdriver/PageUpDownKeysTest.ts
```diff
@@ -0,0 +1,102 @@
...
+      TinySelections.setCursor(editor, [ 0, 0, 0 ], 2);
+      await RealKeys.pSendKeysOn('iframe => body', [ RealKeys.combo({}, 'PageUp') ]);
+
+      // The caret should be on a ZWSP/fake caret position since we're outside the anchor
+      TinyAssertions.assertCursor(editor, [ 0, 0 ], 0);
+    });
+
+    it('TINY-4612: "inline_boundaries: false" cursor does not move outside anchor', async () => {
+      const editor = hook.editor();
+      editor.settings.height = 400;
```
ltrouton, line 48: Was just wondering if this intentionally included as I don't think it will do anything?

modules/agar/CHANGELOG.md
camiloHimura: updated

modules/tinymce/CHANGELOG.md

PR #6527, opened 2021-02-11
TINY-6980: Convert to new changelog format

lnewson:
Related Ticket: TINY-6980

Description of Changes:
* Converts the TinyMCE and additional changelog to the keep a changelog format
* Updates the Gruntfile to be able to pull the changelog date from the new format
* Fixed up README file name consistency
* Removes some old files that no longer make sense

Review:
* [x] Milestone set
* [x] Review comments resolved

GitHub issues (if applicable):

PR #6529, opened 2021-02-11
TINY-6891: Add alpha list numbering support to the lists plugin

SimonFc:
Related Ticket: TINY-6891

Description of Changes:
* Add alpha list numbering support to the lists plugin

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/tinymce/src/plugins/lists/main/ts/ui/Dialog.ts
```diff
@@ -47,10 +51,17 @@ const open = (editor: Editor) => {
...
+      parseStartValue(data.start).each((detail) => {
+        editor.undoManager.transact(() => {
+          editor.execCommand('mceListUpdate', false, {
+            attrs: {
+              start: detail.start === '1' ? '' : detail.start
+            }
+          });
+          editor.execCommand('InsertOrderedList', false,
+            { 'list-style-type': detail.listStyleType }
+          );
```
SimonFc, line 63: This is how advlist updates the list-style-type of an existing list but I'm not sure if we would want this to use the new `mceListUpdate` command instead.

modules/tinymce/src/plugins/lists/test/ts/atomic/ListNumberingTest.ts
```diff
@@ -0,0 +1,69 @@
...
+describe('atomic.tinymce.plugins.lists.core.ListNumberingTest', () => {
+  const tOptional = OptionalInstances.tOptional;
+
+  const check = (startValue: string, expectedDetail: Optional<ListDetail>) => {
+    const actualDetail = parseStartValue(startValue);
+
+    Assert.eq('Should convert start value to expected detail', expectedDetail, actualDetail, tOptional());
+
+    actualDetail.map(parseDetail).each((initialStartValue) => {
+      Assert.eq('Should convert detail back to initial start value', startValue, initialStartValue);
```
lnewson, line 15: Can you please use a `chai` assertion here instead? as we're trying to remove the bedrock Assert logic now so that we don't need to try and maintain a custom assertion library. Chai also offers a lot more and often clearer assertions.

Note: The one above is okay for now, as we don't have an equivalent way to compare optionals atm.

modules/tinymce/src/plugins/lists/main/ts/ui/Dialog.ts
lnewson: I think this makes sense, as that'll make sure the `ListMutation` event is fired as well.

I'm actually wondering now though if the `mceListUpdate` command should also be firing that event 🤷 If it did fire, then I'd say swap to updating via the single command instead, as that's likely to be simpler for RTC efforts.

modules/tinymce/src/plugins/lists/test/ts/atomic/ListNumberingTest.ts
```diff
@@ -0,0 +1,69 @@
...
+    const arbitrary = Arr.map([
+      fc.mapToConstant({ num: 26, build: (v) => String.fromCharCode(65 + v) }),
+      fc.mapToConstant({ num: 26, build: (v) => String.fromCharCode(97 + v) }),
+      fc.mapToConstant({ num: 10, build: (v) => v.toString() })
+    ], (c) => fc.stringOf(c, 1, 5));
+
+    fc.assert(fc.property(
+      fc.oneof(...arbitrary),
+      (start) => {
+        Assert.eq(
```
lnewson, line 61: Likewise please use chai for this assertion.

modules/tinymce/src/plugins/lists/main/ts/core/ListNumbering.ts
```diff
@@ -0,0 +1,100 @@
+import { Arr, Optional, Strings } from '@ephox/katamari';
+
+export interface ListDetail {
+  start: string;
```
lnewson, line 4: Readonly?

lnewson (approved):

modules/tinymce/src/plugins/lists/main/ts/core/ListNumbering.ts
```diff
@@ -0,0 +1,100 @@
...
+  } else {
+    const remainder = value % 26;
+    const quotient = Math.floor(value / 26);
+
+    const rest = composeAlphabeticBase26(quotient);
+    const char = String.fromCharCode('A'.charCodeAt(0) + remainder);
+
+    return rest + char;
+  }
+};
```
TheSpyder, line 39: This is not far off how we did it in our previous editor that supported updating list start values! I'll note it in JIRA since it's a private repo link.

The code might be worth bookmarking if we later want to support start numbering for roman lists as well.

modules/tinymce/src/plugins/lists/main/ts/ui/Dialog.ts
TheSpyder: The `InsertOrderedList` command would work for both `list-style-type` and `start`; there's a `list-attributes` nested record available on the toggle list details object. RTC doesn't support `list-attributes` yet but I can add that easily enough.

However `mceListUpdate` is the better way to update attributes, because using insert list to do it will update the list element tag and [a bunch of other unnecessary things](https://github.com/tinymce/tinymce/blob/develop/modules/tinymce/src/plugins/lists/main/ts/actions/ToggleList.ts#L273-L279) when all you want is to update two attributes on the parent list element.

Whichever way you decide to go, switching to a single command will indeed help RTC as it will only generate one "update attributes" transform operation instead of two.

SimonFc: Yeah I imagined that just using the `mceListUpdate` command would simplify things for RTC. @TheSpyder do we want `mceListUpdate` take

1. a new `'list-style-type'?: string;` property, similar to the insert list commands?
2. a new `styles?: Record<string, string>;` property?

I'm just guessing that we wouldn't want to use the existing attrs property to update the styles(?)

About the `ListMutation` event, I'm not sure if it should fire if only the attributes are updated: https://www.tiny.cloud/docs/advanced/events/#listsevents I think it was just added for structural DOM changes at least.

TheSpyder: It should be a new `styles` record property, to match `attrs`

modules/tinymce/changelog.txt
```diff
@@ -1,5 +1,5 @@
 Version 5.8.0 (TBD)
-
+    Added alpha list numbering support to the `lists` plugin #TINY-6891
```
ashfordneil, line 2: Should we wait for [the changelog updates](https://github.com/tinymce/tinymce/pull/6527) to go through before we merge this?

modules/tinymce/src/plugins/lists/main/ts/actions/Update.ts
```diff
@@ -11,11 +11,15 @@ import { getParentList } from '../core/Selection';
 
 interface ListUpdate {
   attrs?: Record<string, string>;
+  styles?: Record<string, string>;
 }
 
 export const updateList = (editor: Editor, update: ListUpdate) => {
   const parentList = getParentList(editor);
   editor.undoManager.transact(() => {
+    if (update.styles) {
```
ashfordneil, line 20: Can we make this an explicit `undefined` check instead of going for truthiness?

modules/tinymce/src/plugins/lists/main/ts/core/ListNumbering.ts
```diff
@@ -0,0 +1,100 @@
+import { Arr, Optional, Strings } from '@ephox/katamari';
+
+export interface ListDetail {
+  readonly start: string;
+  readonly listStyleType: string | false;
```
ashfordneil, line 5: Could this be `Optional<string>` or something similar?

modules/tinymce/src/plugins/lists/main/ts/ui/Dialog.ts
```diff
@@ -47,10 +51,17 @@ const open = (editor: Editor) => {
     ],
     onSubmit: (api) => {
       const data = api.getData();
-      editor.execCommand('mceListUpdate', false, {
-        attrs: {
-          start: data.start === '1' ? '' : data.start
-        }
+      parseStartValue(data.start).each((detail) => {
+        editor.undoManager.transact(() => {
```
ashfordneil, line 55: Do we need the transact if we're only doing the one thing inside of the callback?

modules/tinymce/src/plugins/lists/test/ts/browser/ListPropertiesTest.ts
```diff
@@ -191,4 +200,47 @@ describe('browser.tinymce.plugins.lists.ListPropertiesTest', () => {
...
+    TinyAssertions.assertContent(editor, '<ol><li>1</li><li>2</li></ol>');
+  });
+
+  it('TINY-6891: List properties command with invalid value should not update the list', () => {
+    const editor = hook.editor();
+    editor.setContent('<ol style="list-style-type: lower-alpha;" start="1"><li>1</li><li>2</li></ol>');
+    TinySelections.setCursor(editor, [ 0, 0, 0 ], 0);
+    editor.execCommand('mceListProps');
+    updateDialog(editor, 'a', 'abc123');
+    editor.setContent('<ol style="list-style-type: lower-alpha;" start="1"><li>1</li><li>2</li></ol>');
```
ashfordneil, line 244: Should this be an assert?

ashfordneil (approved):

modules/tinymce/CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+### Added
+- Added support for alpha list numbering the `lists` plugin #TINY-6891
```
lnewson, line 9:
```suggestion
## Unreleased

### Added
- Added support for alpha list numbering to the `lists` plugin #TINY-6891
```

Just adding a line break and added `to` .

modules/tinymce/src/plugins/lists/main/ts/ui/Dialog.ts
```diff
@@ -47,10 +52,15 @@ const open = (editor: Editor) => {
     ],
     onSubmit: (api) => {
       const data = api.getData();
-      editor.execCommand('mceListUpdate', false, {
-        attrs: {
-          start: data.start === '1' ? '' : data.start
-        }
+      parseStartValue(data.start).each((detail) => {
```
TheSpyder, line 55: This is the behaviour change, previously an unknown list start was used verbatim. Now it's blocked.

We don't want to post invalid list start values, but we _do_ want to post empty strings. That's remove attribute (this is also what happens in the `detail.start === '1' ? '' : detail.start` code).

PR #6532, opened 2021-02-12
TINY-6869: Menu items no longer squash and distort in certain situations

HAFRMO:
Related Ticket:

Description of Changes:
Added `white-space: pre` to menues to prevent squashing in some situations.

Pre-checks:
* [ ] Changelog entry added

Review:
* [ ] Milestone set

### DIFF
--- a/modules/oxide/src/less/theme/components/menu/menu.less
+++ b/modules/oxide/src/less/theme/components/menu/menu.less
@@ -27,6 +27,7 @@
     display: inline-block;
     overflow: hidden;
     vertical-align: top;
+    white-space: pre;
     z-index: @z-index-menu;
   }
 


Afraithe:
@lostkeys can you have a look at this css solution for this issue? Is it something we can use?

spocke:
Weird that alloy doesn't constrain the menu to the visual viewport in this case. If it was moved inside the view port would that still be OK on mobile @lostkeys or should it wrap on mobile?

lostkeys:
@spocke I don't know how to move it into the viewport without moving the whole menu or make it narrower. I found the older ticket I was refering to and there the issue is about a huge heading fontsize breaking the styleselect: see TINY-4330

lnewson:
@spocke there's no way for it to constrain within the viewport, as adding this CSS causes it to overflow. So something has to give in that it either overflows outside the viewport, overflows within the menu (eg `overflow-x: scroll`), wraps or truncates.

IIRC the current version relies on alloy setting a max-width and wrapping (or similar), but adding this causes the max-width to be overruled and it to overflow.

Maybe we need to look at truncating in that case instead of wrapping (eg `text-overflow: ellipsis`)? That could get confusing with dialogs though. Anyways, that's really a design question though of how we want to render menus when there's no enough width to fit the content. If the decision is to keep it wrapping, then we need to find a way to make this work with Shadow Roots, as there is still room in the viewport, it's just that the sink can't be the size of the viewport, which is what's causing it to collapse at the moment.

spocke:
@lnewson If the menu element was as wide as it's content alloy would be able to contain the menu into the viewport. The old V4 UI was using pre formatted text and constrained the elements to the viewport as I recall so I think it should be possible but will require some additional tweaking. Worst case you measure all children and set the parent to the same width of the children but with modern css I don't think that would be needed. But having forced pre formatting would make it hard to ever have wrapping menu text.

lnewson:
It does that if possible by swapping anchor points. There's nowhere for it to go on mobile though, as it can't go left or right so there's simply nowhere it can render while being full width.

spocke:
Took another stab at this by using viewport values. https://fiddle.tiny.cloud/r7haab with that they do wrap as they supposed to when hitting the viewport but expand beyond the constrains of it's absolute parent.

PR #6545, opened 2021-02-18
TINY-6891_2: Treat empty string as numerical list

SimonFc:
Related Ticket: TINY-6891
Fixes regression caused by https://github.com/tinymce/tinymce/pull/6529

Description of Changes:
* Treat empty string as numerical list

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/tinymce/src/plugins/lists/main/ts/core/ListNumbering.ts
```diff
@@ -45,7 +45,7 @@ const isLowercase = (str: string): boolean => /^[a-z]+$/.test(str);
 const isNumeric = (str: string): boolean => /^[0-9]+$/.test(str);
 
 const deduceListType = (start: string): ListType => {
-  if (isNumeric(start)) {
+  if (isNumeric(start) || start === '') {
```
JamesToohey, line 48: Not for this PR, but I wonder if an empty string check should be a simple function in Katamari?

JamesToohey (approved):

modules/tinymce/src/plugins/lists/main/ts/core/ListNumbering.ts
lnewson: It already exists btw... `Strings.isEmpty`

lnewson, line 48: This really feels like a hack to be honest, I think instead we should consider adding a `None` list type and branch as needed on that instead.

lnewson (changes requested):

modules/tinymce/src/plugins/lists/main/ts/core/ListNumbering.ts
JamesToohey: Ahh I looked in the wrong place!

SimonFc: Yup did this.

JamesToohey (approved):

ltrouton (approved):

modules/tinymce/src/plugins/lists/main/ts/core/ListNumbering.ts
```diff
@@ -83,6 +86,12 @@ const parseStartValue = (start: string): Optional<ListDetail> => {
         start: parseAlphabeticBase26(start).toString()
       });
 
+    case ListType.None:
+      return Optional.some({
+        listStyleType: Optional.none(),
+        start: '1'
```
lnewson, line 92: Should this be `''` instead of `'1'` as I still find it quite strange to be returning a number here for something that's supposed to be "None"?

lnewson (approved):

PR #6550, opened 2021-02-20
Feature/6508: Double-clicking a link also selects br, prevents dialog from getting data

camiloHimura:
Related Ticket:

Description of Changes:
Add new conditions to handle Double-clicking cases in firefox and IE

Pre-checks:
* [ ] Changelog entry added

Review:
* [x] Milestone set


lnewson (changes requested):
I appreciate the effort here, but this is not much different to the "change the selection in core" approach, which we said we shouldn't be doing in the previous PR. It also doesn't handle the triple-click approach I mentioned also applied to this. You'll need to change the lookup logic for the anchor, not the selection to fix this in a safe manner.

As a side note, this also hasn't addressed a few of the comments in the old PR (eg double click should have.

spocke:
I don't think this is really related to double click it's more of a normalization for selections kind of unhanging them. There are multiple ways this bug could appear double clicking, triple clicking, selecting and dropping the mouse at a specific point etc. It will be hard to catch them all. I think it's better to just make a generic solution here and use that before editing the link and don't care where that selection came from.

There are also multiple contents scenarios that for example more nested structures `<p><b><a href="#">[abc</a></b>]<br></p>` we have some code in the core that normalizes ranges not for this particular case but rather between blocks. If the selection is hanging like this `<p><b><a href="#">[abc</a></b></p><p>]a</p>` it becomes `<p><b><a href="#">[abc]</a></b></p><p>a</p>` but that code is unfortunately not exposed and it doesn't handle this BR case.

Doing a proper fix for this would be a pretty big task since it would have to traverse backwards until it finds a valid caret candidate then check if that candidate is the same link as the start link.

So maybe we just fix this simple scenario first.

modules/tinymce/src/plugins/link/main/ts/core/Utils.ts
```diff
@@ -60,11 +61,38 @@ const applyRelTargetRules = (rel: string, isUnsafe: boolean): string => {
...
+    return end;
+  }
+  return null;
+};
+const isTexParagraphtNode = (start: SugarElement<Node>, end: SugarElement<Node>): boolean => getTextParagraphNode(start, end) !== null;
+
 const getAnchorElement = (editor: Editor, selectedElm?: Element): HTMLAnchorElement | null => {
   selectedElm = selectedElm || editor.selection.getNode();
+  const rng = editor.selection.getRng();
+  const start = SugarElement.fromDom(rng.startContainer);
```
spocke, line 80: There is a lot of details here fell free to reach out to me on slack if anything is unclear:

What you want to get here is the BR element. Dom ranges has two modes one is text node based offsets then other one is element based offsets. For this case the endContainer would be the parent paragraph element and the endOffset would be set just before that BR. So for a selection that is `<p><a href="#">[abc</a>]<br></p>` the startContainer would be the `abc` test and the endContainer would be the `P` element and the endOffset would be 1 since it's the index before the start.

So you could resolve that index by doing something like this (dirty pseudo code):
```
if (rng.endContainer.nodeType === 1) {
  return rng.endContainer.childNodes[rng.endOffset];
}
```

So if that is a BR you can then just do a resolvedNode.previousSibling and check if that sibling is the same as the startLink. I don't think you would need to use sugar here since it's pretty simple.

So if this:
editor.dom.getParent(tinymce.activeEditor.selection.getNode(), 'a[href]')

Doesn't return link but this does:
editor.dom.getParent(tinymce.activeEditor.selection.getStart(), 'a[href]')

Then you can resolve the end and see if that is a BR and if that BR has a previous sibling that is the same link element as the start.

This would only solve the simple case not wrapped links for that it would have to traverse. You could then use the tinymce TreeWalker then walk backwards from the resolved endNode to the startNode if you only find inline elements that are not short ended you can continue walking until you hit the link. That means it will not walk over text nodes or IMG elements or other BR elements etc. Checking if a element is a inline could be done using the schema api. So you do something like `editor.schema.getShortEndedElements().hasOwnProperty(traverseNode.nodeName)` to check if it's a IMG or BR then `!editor.schema.getBlockElements().hasOwnProperty(traverseNode.nodeName)` to check if it's a inline if it is not a block then it is an inline unless it's a text node.

So it would be something like in pseudo code:
`const isValidNode = (node) => isElement(node) && isInline(node) && !isShortEnded(node) && node !== startLink`

Also I think this function could be created and tested in isolation then you can feed it different content scenario cases and validate that it all returns the link expected. It would require a the schema and range so maybe something like:
const getSelectedLink = (editor, schema, range): Element? then you could setup a editor in the test set the content to something, feed it a range and see if you get back the expected element.

So maybe pull this out into a separate module called SelectedLink.ts or something add tests for that like SelectedLinkTest.ts that calls the above function on various content pieces. Then you add a specific integration test where you open the dialog etc with a range that is like the BR case to validate that the whole solution works.

spocke (changes requested):

modules/tinymce/src/plugins/link/test/ts/atomic/NormalizeLinkTest.ts
```diff
@@ -0,0 +1,69 @@
...
+      test('<p><b><a href="#">test</a></b><br></p>', makeRange([ 0, 0, 0, 0 ], 0, [ 0 ], 1), Optional.some([ 0, 0, 0 ]));
+    });
+  });
+
+  it('TINY-6508: Return a link if the selection starts in the link and end in the link text', () => {
+    test('<p><a href="#">link</a></p>', makeRange([ 0, 0 ], 0, [ 0, 0, 0 ], 3), Optional.some([ 0, 0 ]));
+  });
+
+  it('TINY-6508: Does not return a link since the caret is after the link', () => {
+    test('<p><a href="#">x</a><br></p>', makeCaretRange([ 0 ], 1), Optional.some(null));
```
spocke, line 67: Optional.none() here

modules/tinymce/src/plugins/link/test/ts/atomic/NormalizeLinkTest.ts
```diff
@@ -0,0 +1,69 @@
...
+    const editor = hook.editor();
+    const scope = SugarElement.fromDom(editor.getBody());
+    editor.setContent(html);
+    const resolvedRange = resolveTestRange(scope, range);
+    editor.selection.setRng(resolvedRange);
+    const link = getSelectedLink(editor);
+    const solvedLink = link ? SugarElement.fromDom(link) : SugarElement.fromHtml('<b><b/>');
+    const path = Hierarchy.path(scope, solvedLink);
+
+    assert.deepEqual(path.getOrNull(), expectedLink.getOrNull());
```
spocke, line 45: Change this to `KAssert.eqOptional('Should be expected path', expectedLink, path);`

modules/tinymce/src/plugins/link/test/ts/atomic/NormalizeLinkTest.ts
```diff
@@ -0,0 +1,69 @@
...
+    return rng;
+  };
+
+  const test = (html, range, expectedLink) => {
+    const editor = hook.editor();
+    const scope = SugarElement.fromDom(editor.getBody());
+    editor.setContent(html);
+    const resolvedRange = resolveTestRange(scope, range);
+    editor.selection.setRng(resolvedRange);
+    const link = getSelectedLink(editor);
```
spocke, line 41: This could take the element as it's fallback so `const link = getSelectedLink(editor, editor.selection.getNode());`

modules/tinymce/src/plugins/link/test/ts/atomic/NormalizeLinkTest.ts
```diff
@@ -0,0 +1,69 @@
...
+import Theme from 'tinymce/themes/silver/Theme';
+
+describe('browser.tinymce.plugins.link.NormalizeLink', () => {
+  const hook = TinyHooks.bddSetupLight<Editor>({
+    add_unload_trigger: false,
+    base_url: '/project/tinymce/js/tinymce',
+    indent: false
+  }, [ Theme ], true);
+
+  const makeRange = (startPath, startOffset, endPath, endOffset) => {
```
spocke, line 17: All :any things should have types in this file.

modules/tinymce/src/plugins/link/test/ts/atomic/NormalizeLinkTest.ts
```diff
@@ -0,0 +1,69 @@
+import { describe, it } from '@ephox/bedrock-client';
+import { Optional } from '@ephox/katamari';
+import { TinyHooks } from '@ephox/mcagar';
+import { Hierarchy, SugarElement } from '@ephox/sugar';
+import { assert } from 'chai';
+import Editor from 'tinymce/core/api/Editor';
+import { getSelectedLink } from 'tinymce/plugins/link/core/NormalizeLink';
+import Theme from 'tinymce/themes/silver/Theme';
+
+describe('browser.tinymce.plugins.link.NormalizeLink', () => {
```
spocke, line 10: This entire file should be in the `browser` directory since it's not an atomic test. Atomic tests are things that you can test without a browser not exactly sure why it's called that though but that is the pattern we have.

modules/tinymce/src/plugins/link/test/ts/atomic/NormalizeLinkTest.ts
```diff
@@ -0,0 +1,69 @@
...
+    assert.deepEqual(path.getOrNull(), expectedLink.getOrNull());
+  };
+
+  describe('Returns a link if the end selection is after the link but before a br', () => {
+    it('TINY-6508: Single link', () => {
+      test('<p><a href="#">test</a><br></p>', makeRange([ 0, 0, 0 ], 0, [ 0 ], 1), Optional.some([ 0, 0 ]));
+    });
+
+    it('TINY-6508: Multiple short ended elements', () => {
+      test('<p><a href="#">test</a><br><img><br></p>', makeRange([ 0, 0, 0 ], 0, [ 0 ], 3), Optional.some([ 0, 0 ]));
```
spocke, line 54: Should this really be passing it looks like it sets the selection to `<p><a href="#">[test</a><br>]<br></p>` so that should not return the link since the end of the selection if after the image.

modules/tinymce/src/plugins/link/main/ts/core/NormalizeLink.ts
```diff
@@ -0,0 +1,43 @@
...
+    elm === sibling.parentNode ||
+    sibling.parentNode && elm === sibling.parentNode.nextSibling
+  );
+  const isValidNode = (node: Node) => !isShortEnded(node) && isAnchor(node);
+
+  const rng = editor.selection.getRng();
+  const start = rng.startContainer;
+  const end = rng.endContainer;
+
+  const endOffsetChild = end.childNodes[rng.endOffset];
```
spocke, line 19: So im guessing this will check childNodes on text nodes as well but that would produce undefined or null then the isBR would pick that up with it's truthy check above. Maybe it would be better to have do the element check first or maybe use options for all this instead of null checking.

modules/tinymce/src/plugins/link/main/ts/core/NormalizeLink.ts
```diff
@@ -0,0 +1,43 @@
...
+  const startOffsetChild = start.childNodes[rng.startOffset];
+
+  if (end.nodeType === 1 && isBR(endOffsetChild)) {
+    const walker = new DomTreeWalker(endOffsetChild, endOffsetChild.parentNode);
+
+    while (walker.current()) {
+      const node = walker.prev();
+      if (isValidNode(node) && isSameSiblingAnchor(node, start)) {
```
spocke, line 27: I think this could instead be a simple eq if you took the start node and did a `editor.dom.getParent(selectedElm, 'a[href]')
` on that. Maybe pulling that out into a `getClosestLink` function or something. So in the if statement you get the link element of the start node then in the while you do `if (isValidNode(node) && node === startLink) {`.

spocke (changes requested):
Starting to look good. Maybe pump out more tests for edge cases that you can think or more complex nested structures.

modules/tinymce/src/plugins/link/test/ts/atomic/NormalizeLinkTest.ts
camiloHimura: Updated

modules/tinymce/src/plugins/link/test/ts/atomic/NormalizeLinkTest.ts
camiloHimura: Updated

modules/tinymce/src/plugins/link/test/ts/atomic/NormalizeLinkTest.ts
camiloHimura: updated

modules/tinymce/src/plugins/link/main/ts/core/NormalizeLink.ts
camiloHimura: updated

modules/tinymce/src/plugins/link/test/ts/atomic/NormalizeLinkTest.ts
camiloHimura: in this specific case the editor.selection.getNode() return a "BR" and thus Util.ts isImageFigure() does not match it as a Img, NormalizeLink get it and return the initial Link

modules/tinymce/src/plugins/link/test/ts/atomic/NormalizeLinkTest.ts
camiloHimura: updated

modules/tinymce/src/plugins/link/test/ts/browser/NormalizeLinkTest.ts
```diff
@@ -0,0 +1,96 @@
...
+      test('<p><a href="tiny"><img src="image.jpg"></a>text<br></p>', makeRange([ 0, 0, 0 ], 0, [ 0 ], 2), Optional.some([ 0, 0 ]));
+    });
+  });
+
+  it('TINY-6508: Return a link if the selection starts in the link and end in the link text', () => {
+    test('<p><a href="tiny">link</a></p>', makeRange([ 0 ], 0, [ 0, 0, 0 ], 4), Optional.some([ 0, 0 ]));
+  });
+
+  it('TINY-6508: Does not return a link since the caret is after the link', () => {
+    test('<p><a href="tiny">x</a><br></p>', makeCaretRange([ 0 ], 1), Optional.none());
```
spocke, line 94: There is a lot more none cases we could check for here. Like a range where the end if in another paragraph. Where the range is between two BRs at the end.

A few cases, there could be more but I guess the most important ones is the ones that deals with BRs since it's checking for that then it traverses:
`<p><a href="tiny">[test</a><br>]<br></p>`
`<p><a href="tiny">[test</a><br></p><p>]<br></p>`

camiloHimura: Code updated and new cases added.

modules/tinymce/src/plugins/link/main/ts/core/NormalizeLink.ts
```diff
@@ -0,0 +1,54 @@
...
+const isValidNode = (editor: Editor, node: Node) => !isShortEnded(editor, node) && isAnchor(node);
+
+const getClosestLink = (editor: Editor, node: Node) => editor.dom.getParent(node, 'a[href]');
+
+const hasSameParent = (editor: Editor, elm: Node, nextNode: Node): elm is HTMLAnchorElement =>
+  editor.dom.getParent(elm, 'p') === editor.dom.getParent(nextNode, 'p');
+
+const getSelectedLink = (editor: Editor, selectedElm?: Element): HTMLAnchorElement | null => {
+  const rng = editor.selection.getRng();
+  const startRng = rng.startContainer;
```
spocke, line 19: It's better to call this startContainer/endContainer since startRng I would think it would be a range not a container node.

modules/tinymce/src/plugins/link/main/ts/core/NormalizeLink.ts
```diff
@@ -0,0 +1,54 @@
...
+  const endRng = rng.endContainer;
+
+  const endOffsetChild = endRng.childNodes[rng.endOffset];
```
spocke, line 22: You could move this into the if statement below then it will check if the startContainer is a element before it tries to get a node out of it. Since the endContainer could be a text node as well.

modules/tinymce/src/plugins/link/main/ts/core/NormalizeLink.ts
```diff
@@ -0,0 +1,54 @@
...
+  const startOffsetChild = startRng.childNodes[rng.startOffset];
```
spocke, line 23: There is a utility function you could use for this to resolve the indexes and it does the checking as well.
It's exposed at public api at:
tinymce.dom.RangeUtils.getNode(container, offset)

This is the function that gets called:
https://github.com/tinymce/tinymce/blob/develop/modules/tinymce/src/core/main/ts/selection/RangeNodes.ts#L19

spocke (changes requested):

modules/tinymce/src/plugins/link/main/ts/core/NormalizeLink.ts
spocke: Added some comments on hot to make this a bit more safe.

modules/tinymce/src/plugins/link/main/ts/core/NormalizeLink.ts
camiloHimura: updated

modules/tinymce/src/plugins/link/main/ts/core/NormalizeLink.ts
camiloHimura: update to RangeUtils.getNode

modules/tinymce/src/plugins/link/main/ts/core/NormalizeLink.ts
camiloHimura: nice utility, updated

spocke (changes requested):

modules/tinymce/src/plugins/link/main/ts/core/NormalizeLink.ts
```diff
@@ -0,0 +1,55 @@
...
+
+const isAnchor = (elm: Node): elm is HTMLAnchorElement => elm && elm.nodeName.toLowerCase() === 'a';
+
+const isShortEnded = (editor: Editor, elm: Node) => elm && editor.schema.getShortEndedElements().hasOwnProperty(elm.nodeName);
+
+const isValidNode = (editor: Editor, node: Node) => !isShortEnded(editor, node) && isAnchor(node);
+
+const getClosestLink = (editor: Editor, node: Node) => editor.dom.getParent(node, 'a[href]');
+
+const hasSameParent = (editor: Editor, elm: Node, nextNode: Node): elm is HTMLAnchorElement =>
```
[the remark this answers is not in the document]
camiloHimura: updated

modules/tinymce/src/plugins/link/main/ts/core/NormalizeLink.ts
```diff
@@ -0,0 +1,55 @@
...
+  const endOffsetChild = RangeUtils.getNode(endContainer, rng.endOffset);
+  const startOffsetChild = RangeUtils.getNode(startContainer, rng.startOffset);
+
+  if (endContainer.nodeType === 1 && isBR(endOffsetChild)) {
+    let startNode: Node = endOffsetChild;
+    let rootNode: Node = endOffsetChild.parentNode;
+    const startLink = getClosestLink(editor, startContainer);
+
+    if (!hasSameParent(editor, startContainer, endOffsetChild) && isAnchor(startLink)) {
+      rootNode = editor.dom.getParent(startLink, 'p');
```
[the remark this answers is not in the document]
camiloHimura: I followed the second approach, updated

modules/tinymce/src/plugins/link/main/ts/core/NormalizeLink.ts
```diff
@@ -0,0 +1,55 @@
...
+const isShortEnded = (editor: Editor, elm: Node) => elm && editor.schema.getShortEndedElements().hasOwnProperty(elm.nodeName);
+
+const isValidNode = (editor: Editor, node: Node) => !isShortEnded(editor, node) && isAnchor(node);
+
+const getClosestLink = (editor: Editor, node: Node) => editor.dom.getParent(node, 'a[href]');
+
+const hasSameParent = (editor: Editor, elm: Node, nextNode: Node): elm is HTMLAnchorElement =>
+  editor.dom.getParent(elm, 'p') === editor.dom.getParent(nextNode, 'p');
+
+const getSelectedLink = (editor: Editor, selectedElm?: Element): HTMLAnchorElement | null => {
```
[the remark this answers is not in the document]
camiloHimura: updated

modules/tinymce/src/plugins/link/main/ts/core/NormalizeLink.ts
```diff
@@ -0,0 +1,65 @@
+import { Fun, Optional } from '@ephox/katamari';
+import RangeUtils from 'tinymce/core/api/dom/RangeUtils';
+import DomTreeWalker from 'tinymce/core/api/dom/TreeWalker';
+import Editor from 'tinymce/core/api/Editor';
+
+const isBR = (elm: Node): Boolean => elm && elm.nodeName.toLowerCase() === 'br';
```
spocke, line 6: These should all return `boolean` not `Boolean` since that is the object type.

camiloHimura: updated

modules/tinymce/src/plugins/link/main/ts/core/NormalizeLink.ts
```diff
@@ -0,0 +1,59 @@
...
+  const startOffsetChild = getNode(startContainer, rng.startOffset);
+  const endOffsetChild = getNode(endContainer, rng.endOffset);
+
+  return getCommonBlockNode(editor, startContainer, endContainer).bind((rootNode) => getClosestLink(editor, startContainer)
+    .bind((startLink) => endOffsetChild.filter(isBR).bind((br) => {
+      const walker = new DomTreeWalker(br, rootNode);
+      while (walker.current()) {
+        const node = walker.prev();
+        if (node === startLink) {
+          return Optional.from(node as HTMLAnchorElement);
```
spocke, line 43: If you return the startLink here you could remove the HTMLAnchorElement cast.

modules/tinymce/src/plugins/link/test/ts/browser/NormalizeLinkTest.ts
```diff
@@ -0,0 +1,120 @@
...
+    const solvedLink = link ? SugarElement.fromDom(link) : SugarElement.fromHtml('<b><b/>');
+    const path = Hierarchy.path(scope, solvedLink);
+    KAssert.eqOptional('Should be expected path', expectedLink, path);
+  };
+  describe('Returns a link if the end selection is after the link but before a br', () => {
+    it('TINY-6508: Single link', () => {
+      test('<p><a href="tiny">test</a><br></p>', makeRange([ 0, 0, 0 ], 0, [ 0 ], 1), Optional.some([ 0, 0 ]));
+    });
+
+    it('TINY-6508: Multiple short ended elements', () => {
```
spocke, line 52: Maybe clarify this by having it be `Should not find a link if there are short ended elements between the link and the br` or something similar.

modules/tinymce/src/plugins/link/test/ts/browser/NormalizeLinkTest.ts
```diff
@@ -0,0 +1,120 @@
...
+      test('<p><a href="tiny">test</a><br><img><br></p>', makeRange([ 0, 0, 0 ], 0, [ 0 ], 3), Optional.none());
+    });
+
+    it('TINY-6508: Nest link', () => {
```
spocke, line 56: Maybe `should find a link if the link is wrapped in another inline element`

modules/tinymce/src/plugins/link/test/ts/browser/NormalizeLinkTest.ts
```diff
@@ -0,0 +1,120 @@
...
+      test('<p><b><a href="tiny">test</a></b><br></p>', makeRange([ 0, 0, 0, 0 ], 0, [ 0 ], 1), Optional.some([ 0, 0, 0 ]));
+    });
+
+    it('TINY-6508: Nest link next to another Nest link', () => {
```
spocke, line 60: Maybe change this to `should not find a link if the start is on one nested link and the end is on a br after another`

modules/tinymce/src/plugins/link/test/ts/browser/NormalizeLinkTest.ts
```diff
@@ -0,0 +1,120 @@
...
+      test('<p><b><a href="tiny">test</a></b><b><a href="tiny">tiny</a></b><br></p>', makeRange([ 0, 0, 0, 0 ], 0, [ 0 ], 2), Optional.none());
+    });
+
+    it('TINY-6508: Single link with one level nest child', () => {
+      test('<p><a href="tiny"><strong>word</strong>other</a><br></p>', makeRange([ 0, 0, 0, 0 ], 0, [ 0 ], 1), Optional.some([ 0, 0 ]));
+    });
+
+    it('TINY-6508: Single link with two level nest child', () => {
+      test('<p><a href="tiny"><strong><em>word<em/></strong>other</a><br></p>', makeRange([ 0, 0, 0, 0, 0 ], 0, [ 0 ], 1), Optional.some([ 0, 0 ]));
```
spocke, line 69: This `<em/>` is incorrect.

modules/tinymce/src/plugins/link/test/ts/browser/NormalizeLinkTest.ts
```diff
@@ -0,0 +1,120 @@
...
+    it('TINY-6508: Nest link in a block element (div)', () => {
+      test('<div><a href="tiny">tiny</a><em>tiny</em><br></div>', makeRange([ 0, 0, 0 ], 0, [ 0 ], 2), Optional.none());
+    });
+
+    it('TINY-6508: Range end in another block element (div)', () => {
+      test('<div><a href="tiny">test</a><br></div><div>info<br></div>', makeRange([ 0, 0, 0 ], 0, [ 1 ], 1), Optional.none());
+    });
+  });
+
+  it('TINY-6508: Return a link if the selection starts in the link and end in the link text', () => {
```
spocke, line 109: Maybe add that this is a IE case here as well. Maybe move it to it's own describe not sure.

modules/tinymce/src/plugins/link/test/ts/browser/NormalizeLinkTest.ts
```diff
@@ -0,0 +1,120 @@
+import { describe, it } from '@ephox/bedrock-client';
+import { Optional } from '@ephox/katamari';
+import { KAssert } from '@ephox/katamari-assertions';
+import { TinyHooks } from '@ephox/mcagar';
+import { Hierarchy, SugarElement } from '@ephox/sugar';
+import Editor from 'tinymce/core/api/Editor';
+import { getSelectedLink } from 'tinymce/plugins/link/core/NormalizeLink';
+import Theme from 'tinymce/themes/silver/Theme';
+interface BaseRange {
```
spocke, line 9: Maybe fluff this up a bit add new lines between each block.

modules/tinymce/src/plugins/link/test/ts/browser/NormalizeLinkTest.ts
```diff
@@ -0,0 +1,120 @@
...
+    rng.setEnd(endContainer.dom, range.endOffset);
+    return rng;
+  };
+  const test = (html: string, range: BaseRange, expectedLink: Optional<number[]>): void => {
+    const editor = hook.editor();
+    const scope = SugarElement.fromDom(editor.getBody());
+    editor.setContent(html);
+    const resolvedRange = resolveTestRange(scope, range);
+    editor.selection.setRng(resolvedRange);
+    const link = getSelectedLink(editor, editor.selection.getNode()).getOrNull();
```
spocke, line 42: Since this is returning an option just bind over `Hierarchy.path` then you wouldn't need the solvedLink part below. So remove getOrNull as well.

modules/tinymce/src/plugins/link/main/ts/core/NormalizeLink.ts
camiloHimura: updated

modules/tinymce/src/plugins/link/test/ts/browser/NormalizeLinkTest.ts
camiloHimura: updated

modules/tinymce/src/plugins/link/test/ts/browser/NormalizeLinkTest.ts
camiloHimura: updated

modules/tinymce/src/plugins/link/test/ts/browser/NormalizeLinkTest.ts
camiloHimura: updated

modules/tinymce/src/plugins/link/test/ts/browser/NormalizeLinkTest.ts
camiloHimura: removed

modules/tinymce/src/plugins/link/test/ts/browser/NormalizeLinkTest.ts
camiloHimura: Add in a new describe for IE cases

modules/tinymce/src/plugins/link/test/ts/browser/NormalizeLinkTest.ts
camiloHimura: updated

modules/tinymce/src/plugins/link/test/ts/browser/NormalizeLinkTest.ts
camiloHimura: updated

PR #6573, opened 2021-03-02
TINY-7074: Fixed a regression with widths/heights not being included on the media wrapper span

lnewson:
Related Ticket: TINY-7074

Description of Changes:
* Fixes the width/height not being copied to the "wrapper" span used in the media plugin.
* Ensure when resizing the width/height are set on both the wrapper span and original media element.

Note: This will convert `%` to `px` when resizing but that's outside the scope of this, as TinyMCE's resizing logic has never supported that outside tables.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/changelog.txt
+++ b/modules/tinymce/changelog.txt
@@ -2,7 +2,7 @@ Version 5.7.1 (TBD)
     Fixed an issue where URLs were not correctly filtered in some cases #TINY-7025
     Fixed a bug where context menu items with names that contained uppercase characters were not displayed #TINY-7072
     Fixed context menu items lacking support for the `disabled` and `shortcut` properties #TINY-7073
-    Fixed a regression whereby the width and height weren't set correctly when embedded content using the `media` dialog #TINY-7074
+    Fixed a regression whereby the width and height weren't set correctly when embedding content using the `media` dialog #TINY-7074
 Version 5.7.0 (2021-02-10)
     Added IPv6 address support to the URI API. Patch contributed by dev7355608 #GH-4409
     Added new `structure` and `style` properties to the `TableModified` event to indicate what kinds of modifications were made #TINY-6643
--- a/modules/tinymce/changelog.txt
+++ b/modules/tinymce/changelog.txt
@@ -1,7 +1,8 @@
 Version 5.7.1 (TBD)
-    Fixed a bug where context menu items with names that contained uppercase characters were not displayed #TINY-7072
     Fixed an issue where URLs were not correctly filtered in some cases #TINY-7025
+    Fixed a bug where context menu items with names that contained uppercase characters were not displayed #TINY-7072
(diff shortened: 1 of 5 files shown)


modules/tinymce/changelog.txt
```diff
@@ -1,7 +1,8 @@
 Version 5.7.1 (TBD)
-    Fixed a bug where context menu items with names that contained uppercase characters were not displayed #TINY-7072
     Fixed an issue where URLs were not correctly filtered in some cases #TINY-7025
+    Fixed a bug where context menu items with names that contained uppercase characters were not displayed #TINY-7072
     Fixed context menu items lacking support for the `disabled` and `shortcut` properties #TINY-7073
+    Fixed a regression whereby the width and height weren't set correctly when embedded content using the `media` dialog #TINY-7074
```
lnewson, line 5:
```suggestion
    Fixed a regression whereby the width and height weren't set correctly when embedding content using the `media` dialog #TINY-7074
```

ltrouton (approved):

modules/tinymce/src/core/test/ts/browser/dom/ControlSelectionTest.ts
```diff
@@ -56,7 +56,7 @@ describe('browser.tinymce.core.dom.ControlSelectionTest', () => {
     const dimension = Css.getRaw(element, name).orThunk(() => Attribute.getOpt(element, name))
       .map((v) => parseInt(v, 10))
       .getOr(0);
-    assert.isBelow(Math.abs(dimension - expectedDimension), 3, label + ` ${dimension}px ~= ${expectedDimension}px`);
+    assert.approximately(dimension, expectedDimension, 3, `${label} ${dimension}px ~= ${expectedDimension}px`);
```
ashfordneil, line 59: Does `assert.approximately` include the `a ~= b` part in the message automatically? Maybe we could take this down to just `label`

ashfordneil (approved):

modules/tinymce/src/core/test/ts/browser/dom/ControlSelectionTest.ts
JamesToohey: https://github.com/chaijs/chai/blob/main/lib/chai/core/assertions.js#L2979 

It looks like it does

JamesToohey (approved):

modules/tinymce/changelog.txt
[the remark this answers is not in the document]
JamesToohey: This was old, ignore

modules/tinymce/src/core/test/ts/browser/dom/ControlSelectionTest.ts
lnewson: We've used this style everywhere else, as while it is included it's more readable in this format. If we want to change it, that's out of scope and something we should do via a separate PR.

PR #6576, opened 2021-03-02
TINY-7098: fixed issue with setup function not always being available

spocke:
Related Ticket:

Description of Changes:
* When rtc is not enabled for a specific api key a sub plugin gets loaded instead of the real plugin. That plugin doesn't have any setup function so the Rtc integration would try to call a setup function on a plugin that doesn't have that function exported. This fix just check if the setup function exists if it doesn't it falls back to the normal editor setup but since the stub plugin will present a notification of that rtc is not enabled the user would be presented with a nice error message.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/Rtc.ts
+++ b/modules/tinymce/src/core/main/ts/Rtc.ts
@@ -251,12 +251,11 @@ const makeNoopAdaptor = (): RtcAdaptor => {
 
 export const isRtc = (editor: Editor) => Obj.has(editor.plugins, 'rtc');
 
-const getRtcSetup = (editor: Editor): Optional<() => Promise<RtcRuntimeApi>> => {
-  return (Obj.get(editor.plugins, 'rtc') as Optional<RtcPluginApi>).bind((rtcPlugin) => {
+const getRtcSetup = (editor: Editor): Optional<() => Promise<RtcRuntimeApi>> =>
+  (Obj.get(editor.plugins, 'rtc') as Optional<RtcPluginApi>).bind((rtcPlugin) =>
     // This might not exist if the stub plugin is loaded on cloud
-    return Optional.from(rtcPlugin.setup);
-  });
-};
+    Optional.from(rtcPlugin.setup)
+  );
 
 export const setup = (editor: Editor): Optional<Promise<boolean>> => {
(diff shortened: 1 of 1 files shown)


TheSpyder:
This probably needs to go in a patch release; the effect of this bug is anyone attempting to use RTC beta without it enabled on their cloud key results in a crash instead of an editor load.

ashfordneil (approved):

modules/tinymce/src/core/main/ts/Rtc.ts
```diff
@@ -251,15 +251,22 @@ const makeNoopAdaptor = (): RtcAdaptor => {
 
 export const isRtc = (editor: Editor) => Obj.has(editor.plugins, 'rtc');
 
+const getRtcSetup = (editor: Editor): Optional<() => Promise<RtcRuntimeApi>> => {
+  return (Obj.get(editor.plugins, 'rtc') as Optional<RtcPluginApi>).bind((rtcPlugin) => {
+    // This might not exist if the stub plugin is loaded on cloud
+    return Optional.from(rtcPlugin.setup);
+  });
+};
```
lnewson, line 259:
```suggestion
const getRtcSetup = (editor: Editor): Optional<() => Promise<RtcRuntimeApi>> =>
  Obj.get(editor.plugins, 'rtc') as Optional<RtcPluginApi>).bind((rtcPlugin) => {
    // This might not exist if the stub plugin is loaded on cloud
    return Optional.from(rtcPlugin.setup);
  });
```

We can use the shorthand for this, which arguably helps make it a little easier to read as well :shrug:

lnewson (approved):

modules/tinymce/src/core/main/ts/Rtc.ts
TheSpyder: why not shorthand both 🤷‍♂️

lnewson: I did consider that, but I assumed that was done purposely :shrug: if that's not the case then yeah we might as well

lnewson: I updated the suggestion, as it probably wasn't done purposely thinking more.

PR #6617, opened 2021-03-12
TINY-7167: Patch the flaky test, and leave a TODO for future investigations

ashfordneil:
Related Ticket: TINY-7167

Description of Changes:
* There was a messy test flake, that's going to need a bit more investigation
* It needs more investigation, but for now this is a workaround

Pre-checks:
* ~[ ] Changelog entry added~
* ~[ ] Branch prefixed with `feature/` for new features (if applicable)~
* ~[ ] License headers added on new files (if applicable)~

Review:
* [x] Milestone set


lnewson (approved):

lnewson (changes requested):
Changing my approval as per discussion this morning. We want to be as safe as possible so we shouldn't be upgrading all these deps, only bedrock itself.

JamesToohey (approved):
- [x] Approving assuming we revert the dependency updates

ltrouton (approved):

lnewson (approved):

PR #6665, opened 2021-03-25
TINY-6706: disabled SelectionOverrides and dom/selection mutating quirks

spocke:
Related Ticket:

Description of Changes:
* Disables SelectionOverrides for content editable false elements since those are now handled by the RTC model.
* Disables some of the mutating quirks when running with RTC.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/util/Quirks.ts
+++ b/modules/tinymce/src/core/main/ts/util/Quirks.ts
@@ -13,6 +13,7 @@ import Delay from '../api/util/Delay';
 import Tools from '../api/util/Tools';
 import VK from '../api/util/VK';
 import * as CaretContainer from '../caret/CaretContainer';
+import * as Rtc from '../Rtc';
 import * as CaretRangeFromPoint from '../selection/CaretRangeFromPoint';
 
 /**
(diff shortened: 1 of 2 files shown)


modules/tinymce/src/core/main/ts/util/Quirks.ts
```diff
@@ -769,58 +770,79 @@ const Quirks = (editor: Editor): Quirks => {
     return (!sel || !sel.rangeCount || sel.rangeCount === 0);
   };
 
-  // All browsers
-  removeBlockQuoteOnBackSpace();
-  emptyEditorWhenDeleting();
+  if (Rtc.isRtc(editor)) {
```
lnewson, line 773: Instead of a massive if else block, would this be better/more readable if we created 2 functions (eg `setupRtc` and `setup`)? We could then do something like this:
```
if (Rtc.isRtc(editor)) {
  setupRtc();
} else {
  setup();
}

lnewson (approved):

ashfordneil (approved):

modules/tinymce/src/core/main/ts/util/Quirks.ts
spocke: Yeah, that is probably easier to read. And would make it easier to yank out later.

ltrouton (approved):

JamesToohey (approved):

PR #6666, opened 2021-03-25
TINY-6567 - RangeWalk.walk is missing parent li elements in nested lists

camiloHimura:
Related Ticket: Feature/tiny 6567

Description of Changes:
This applies the 'formant' to the last selection's item  if it has Block children

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -12,8 +12,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added new `PAGE_UP` and `PAGE_DOWN` key code constants to the `VK` API
 - Added support for alpha list numbering to the `lists` plugin #TINY-6891
 - The editor resize handle can now be controlled using the keyboard #TINY-4823
+- Added a new `fixed_toolbar_container_target` setting which renders the toolbar in the specified `HTMLElement`. Patch contributed by pvrobays
 
 ### Changed
+- Updated the `image` dialog to display the class list dropdown as full-width if the caption checkbox is not present #TINY-6400
 - Renamed the "H Align" and "V Align" input labels in the Table Cell Properties dialog to "Horizontal align" and "Vertical align" respectively #TINY-7285
 
 ### Fixed
@@ -22,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The color picker dialog will now show an alert if it is submitted with an invalid hex color code #TINY-2814
 - Fixed a bug where the `TableModified` event was not fired when adding a table row via the Tab key #TINY-7006
 - Added missing type for `images_file_types` setting #GH-6607
+- The `editor.insertContent` API was incorrectly handling nested `span` elements with matching styles #TINY-6263
 - The HTML5 `small` element could not be removed when clearing text formatting #TINY-6633
 - Added HTML5 `audio` and `video` elements to the default alignment formats #TINY-6633
 - The Oxide button text transform variable was incorrectly using `capitalize` instead of `none`. Patch contributed by dakur #GH-6341
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -7,7 +7,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Added
-- Updated `applyFormat` and `removeRngStyle` to include the last item in the selection #TINY-6567
 - The `inline_boundaries` feature now supports the `home`, `end`, `pageup` and `pagedown` keys #TINY-4612
 - Added new `PAGE_UP` and `PAGE_DOWN` key code constants to the `VK` API
 - Added support for alpha list numbering to the `lists` plugin #TINY-6891
@@ -19,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Renamed the "H Align" and "V Align" input labels in the Table Cell Properties dialog to "Horizontal align" and "Vertical align" respectively #TINY-7285
 
 ### Fixed
+- Fixed a bug where it wasn't possible to align nested list items #TINY-6567
 - The RGB fields in the color picker dialog were not staying in sync with the color palette and hue slider #TINY-6952
 - The color preview box in the color picker dialog was not correctly displaying the saturation and value of the chosen color #TINY-6952
 - The color picker dialog will now show an alert if it is submitted with an invalid hex color code #TINY-2814
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -8,17 +8,29 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - Updated `applyFormat` and `removeRngStyle` to include the last item in the selection #TINY-6567
+- The `inline_boundaries` feature now supports the `home`, `end`, `pageup` and `pagedown` keys #TINY-4612
+- Added new `PAGE_UP` and `PAGE_DOWN` key code constants to the `VK` API
 - Added support for alpha list numbering to the `lists` plugin #TINY-6891
+- The editor resize handle can now be controlled using the keyboard #TINY-4823
+
+### Changed
+- Renamed the "H Align" and "V Align" input labels in the Table Cell Properties dialog to "Horizontal align" and "Vertical align" respectively #TINY-7285
 
 ### Fixed
 - The RGB fields in the color picker dialog were not staying in sync with the color palette and hue slider #TINY-6952
 - The color preview box in the color picker dialog was not correctly displaying the saturation and value of the chosen color #TINY-6952
 - The color picker dialog will now show an alert if it is submitted with an invalid hex color code #TINY-2814
+- Fixed a bug where the `TableModified` event was not fired when adding a table row via the Tab key #TINY-7006
+- Added missing type for `images_file_types` setting #GH-6607
+- The HTML5 `small` element could not be removed when clearing text formatting #TINY-6633
+- Added HTML5 `audio` and `video` elements to the default alignment formats #TINY-6633
+- The Oxide button text transform variable was incorrectly using `capitalize` instead of `none`. Patch contributed by dakur #GH-6341
+- Fix dialog button text that was using title-style capitalization #TINY-6816
 
-## 5.7.1 - TBD
+## 5.7.1 - 2021-03-17
 
 ### Fixed
-- The `help` dialog incorrectly linked to the changelog of TinyMCE 4 instead of TinyMCE 5 #TINY-7031
+- Fixed the `help` dialog incorrectly linking to the changelog of TinyMCE 4 instead of TinyMCE 5 #TINY-7031
 - Fixed a bug where error messages were displayed incorrectly in the image dialog #TINY-7099
 - Fixed an issue where URLs were not correctly filtered in some cases #TINY-7025
 - Fixed a bug where context menu items with names that contained uppercase characters were not displayed #TINY-7072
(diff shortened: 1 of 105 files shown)


modules/tinymce/src/core/main/ts/fmt/ApplyFormat.ts
```diff
@@ -191,6 +192,11 @@ const applyFormat = (ed: Editor, name: string, vars?: FormatVars, node?: Node |
         if (FormatUtils.isSelectorFormat(format)) {
           const found = applyNodeStyle(formatList, node);
 
+          // TINY-6567 Include the last node in the selection
+          if (node.nodeType === 3 && hasBlockChildren(node.parentNode)) {
```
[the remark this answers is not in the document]
camiloHimura: My first approach was to update the RangeWalk.walk to get the last item in the selection (when it had block child), but that broke several tests, and I also had to update ApplyFormat.ts, and even after there was a test, I could make it work https://github.com/tinymce/tinymce/blob/6f090a64c7e3c3629e6df234936a5364f88c67ec/modules/tinymce/src/core/test/ts/browser/FormatterApplyTest.ts#L1229

I'll update removeFormat also.

modules/tinymce/src/core/main/ts/fmt/ApplyFormat.ts
```diff
@@ -132,6 +132,8 @@ const applyFormat = (ed: Editor, name: string, vars?: FormatVars, node?: Node |
     return found;
   };
 
+  const hasBlockChildren = (elm: Node) => Arr.filter(Arr.from(elm.childNodes), dom.isBlock).length > 0;
```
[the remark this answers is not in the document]
camiloHimura: updated

modules/tinymce/src/core/main/ts/fmt/ApplyFormat.ts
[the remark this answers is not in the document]
camiloHimura: feature/TINY-6567

modules/tinymce/src/core/main/ts/fmt/ApplyFormat.ts
```diff
@@ -191,6 +192,11 @@ const applyFormat = (ed: Editor, name: string, vars?: FormatVars, node?: Node |
         if (FormatUtils.isSelectorFormat(format)) {
           const found = applyNodeStyle(formatList, node);
 
+          // TINY-6567 Include the last node in the selection
+          if (!NodeType.isElement(node) && hasBlockChildren(node.parentNode)) {
```
spocke, line 196: Maybe change this if to
`NodeType.isText(node) && hasBlockChildren(node.parentNode) && nodes[node.length - 1] === node`
or
`NodeType.isText(node) && hasBlockChildren(node.parentNode) && Arr.last(nodes).is(node)`

Since it's supposed to check the last node only and if it's not an element it can only be text node, comment or document but we don't want to process the last two. But the RangeWalk.walk will fire the callback multiple times so it's not really the truth either. But it will limit the scope of this apply.

I'm also a bit unsure what impact this would have wouldn't it double apply the node style if a match is made in the middle of the selection. I guess a selector format is always just updating attributes and such an operation is idempotent.

modules/tinymce/src/core/main/ts/fmt/RemoveFormat.ts
```diff
@@ -470,10 +470,17 @@ const remove = (ed: Editor, name: string, vars?: FormatVars, node?: Node | Range
   // as there maybe nested bookmarks
   const isRemoveBookmarkNode = (node: Node) => Bookmarks.isBookmarkNode(node) && NodeType.isElement(node) && (node.id === '_start' || node.id === '_end');
 
+  const hasBlockChildren = (elm: Node) => Arr.exists(Arr.from(elm.childNodes), dom.isBlock);
```
spocke, line 473: Since this is defined multiple times maybe move it to FormatUtils.ts also I think you might not need the Arr.from I think it accepts array like but haven't double checked.

modules/tinymce/src/core/test/ts/browser/FormatterApplyTest.ts
```diff
@@ -2308,4 +2308,182 @@ describe('browser.tinymce.core.FormatterApplyTest', () => {
...
+        '<li>a</li>' +
+        '<li>b<br />' +
+          '<ul>' +
+            '<li>c</li>' +
+            '<li>d</li>' +
+          '</ul>' +
+        '</li>' +
+      '</ul>'
+    );
+    LegacyUnit.setSelection(editor, 'ul', 0, 'ul li li:nth-child(2)', 0);
```
spocke, line 2325: Siwtch to using TinySelections.setSelection here since that is used in the remove tests.

modules/tinymce/src/core/test/ts/browser/fmt/RemoveFormatTest.ts
```diff
@@ -29,6 +30,10 @@ describe('browser.tinymce.core.fmt.RemoveFormatTest', () => {
     editor.formatter.unregister('format');
   };
 
+  const getContent = (editor: Editor) => {
```
spocke, line 33: You can configure tinymce to not indent the source by setting indent: false in the bddSetupLight call.

modules/tinymce/src/core/test/ts/browser/FormatterApplyTest.ts
```diff
@@ -2308,4 +2308,182 @@ describe('browser.tinymce.core.FormatterApplyTest', () => {
...
+        '<div style="text-align: center;">b' +
+          '<div style="text-align: center;">1</div>' +
+        '</div>' +
+        '<div>c</div>' +
+        '<div>d</div>' +
+      '</div>'
+    );
+  });
+
+  it('TINY-6567: Apply inline element with font Size to a div structure with a partial selection', () => {
```
spocke, line 2458: Lower case `Size`

modules/tinymce/src/core/test/ts/browser/FormatterApplyTest.ts
```diff
@@ -2308,4 +2308,182 @@ describe('browser.tinymce.core.FormatterApplyTest', () => {
...
+                '<li><h1>ii</h1></li>' +
+              '</ul>' +
+            '</li>' +
+          '</ul>' +
+        '</li>' +
+      '</ul>'
+    );
+  });
+
+  it('TINY-6567: Apply align center to a div structure', () => {
```
spocke, line 2432: Nice to see some non list cases since this is really a generic problem.

modules/tinymce/src/core/main/ts/fmt/ApplyFormat.ts
camiloHimura: Instead of always check the last element it's better to used Arr.contains because if the user types something like 
(<br/>test) or (test<br/>) the li parent won't be selected, also I update some tests for those cases.

modules/tinymce/src/core/main/ts/fmt/RemoveFormat.ts
camiloHimura: updated

modules/tinymce/src/core/test/ts/browser/FormatterApplyTest.ts
camiloHimura: updated

modules/tinymce/src/core/test/ts/browser/fmt/RemoveFormatTest.ts
camiloHimura: nice to know, updated

modules/tinymce/src/core/test/ts/browser/FormatterApplyTest.ts
camiloHimura: updated

modules/tinymce/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Added
+- Updated `applyFormat` and `removeRngStyle` to include the last item in the selection #TINY-6567
```
spocke, line 10: - [ ] Since this is a fix for a customer reported bug it should be placed under `Fixed` and maybe be describing the bug fix.

So something like `Fixed a bug where it wasn't possible to align nested list items` maybe the editor team has a better suggestion here.

Generally the changelog should not include things that are internal. So if the change is not in the tinymce api directory then it's an internal change. The changelog is to be consumed by our users/customers so they need to be able to understand from their perspective what has been changed.

spocke (approved):

HAFRMO (approved):

lnewson (changes requested):
I agree with what Spocke said that this doesn't appear to fix the root issue and is a workaround. We really need to fix the root issue with the walker at some stage, however as this is a Zendesk issue I'm fine with using a workaround for now provided we log a JIRA to investigate and fix the root issue still. Leave logging that with me and I'll link it to this JIRA.

As such, I'm just marking this as needs changes because of the `isChildText` comment I've left. The rest of it seems reasonable to me and the tests cover a lot of cases, so really nice job on that!

Edit: The JIRA for the root cause has been logged internally as TINY-7393

modules/tinymce/src/core/main/ts/fmt/FormatUtils.ts
```diff
@@ -216,6 +216,10 @@ const isSelectorFormat = (format: ApplyFormat): format is SelectorFormat =>
 const isInlineFormat = (format: ApplyFormat): format is InlineFormat =>
   Obj.hasNonNullableKey(format as any, 'inline');
 
+const hasBlockChildren = (dom: DOMUtils, elm: Node) => Arr.exists(elm.childNodes, dom.isBlock);
+
+const isChildText = (nodes: Node[], node: Node) => NodeType.isText(node) && Arr.contains(nodes, node);
```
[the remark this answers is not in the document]
camiloHimura: updated

modules/tinymce/src/core/main/ts/fmt/RemoveFormat.ts
```diff
@@ -471,9 +471,14 @@ const remove = (ed: Editor, name: string, vars?: FormatVars, node?: Node | Range
   const isRemoveBookmarkNode = (node: Node) => Bookmarks.isBookmarkNode(node) && NodeType.isElement(node) && (node.id === '_start' || node.id === '_end');
 
   // Merges the styles for each node
-  const process = (node: Node) => {
+  const process = (nodes: Node[], node: Node) => {
```
lnewson, line 474: This can be reverted now, as `nodes` is no longer used.

- [ ] Revert back to just passing a single node

modules/tinymce/src/core/demo/ts/demo/FullDemo.ts
```diff
@@ -47,7 +47,7 @@ export default () => {
       { title: 'Some class', value: 'class-name' }
     ],
     importcss_append: true,
-    height: 400,
+    height: 200,
```
lnewson, line 50: Could you please put this back to its original value?
```suggestion
    height: 400,
```

modules/tinymce/src/core/main/ts/fmt/RemoveFormat.ts
```diff
@@ -471,9 +471,14 @@ const remove = (ed: Editor, name: string, vars?: FormatVars, node?: Node | Range
...
     let lastContentEditable: boolean, hasContentEditableState: boolean;
 
+    // TINY-6567 Include the last node in the selection
+    if (NodeType.isText(node) && FormatUtils.hasBlockChildren(dom, node.parentNode)) {
+      removeFormat(ed, format, vars, node.parentNode, node.parentNode);
```
lnewson, line 479:
```suggestion
    const parentNode = node.parentNode;
    if (NodeType.isText(node) && FormatUtils.hasBlockChildren(dom, parentNode)) {
      removeFormat(ed, format, vars, parentNode, parentNode);
```
I missed this last time sorry, but doing this will allow better minification and removes a little duplication.

lnewson (approved):

ashfordneil (approved):

modules/tinymce/src/core/main/ts/fmt/RemoveFormat.ts
camiloHimura: updated

modules/tinymce/src/core/demo/ts/demo/FullDemo.ts
camiloHimura: updated

modules/tinymce/src/core/main/ts/fmt/RemoveFormat.ts
camiloHimura: updated

PR #6678, opened 2021-03-29
TINY-7191: Restricts placement at the right and bottom side in addition to top and left side for the purposes of context toolbar placement.

HAFRMO:
Related Ticket:

Description of Changes:
Previously the area designed for the context toolbar was cropped on the top to prevent the toolbar from moving off-screen during placement, while no such treatment was given to the bottom area. This crops the area appropriately, allowing for the context toolbar to stick to the bottom of the screen instead of always going for the top.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/alloy/src/main/ts/ephox/alloy/positioning/mode/ContentAnchorCommon.ts
+++ b/modules/alloy/src/main/ts/ephox/alloy/positioning/mode/ContentAnchorCommon.ts
@@ -9,8 +9,11 @@ import * as Origins from '../layout/Origins';
 import { Anchoring, NodeAnchor, nu as NuAnchor, SelectionAnchor } from './Anchoring';
 import * as AnchorLayouts from './AnchorLayouts';
 
-const capRect = (left: number, top: number, width: number, height: number): Optional<Boxes.BoxByPoint> => {
-  let newLeft = left, newTop = top, newWidth = width, newHeight = height;
+const capRect = (left: number, top: number, width: number, height: number, cappingViewport: Optional<Boxes.Bounds>): Optional<Boxes.BoxByPoint> => {
+  let newLeft = left;
+  let newTop = top;
+  let newWidth = width;
+  let newHeight = height;
   // Try to prevent the context toolbar from getting above the editor toolbar
   if (left < 0) {
     newLeft = 0;
@@ -20,6 +23,18 @@ const capRect = (left: number, top: number, width: number, height: number): Opti
     newTop = 0;
     newHeight = height + top;
   }
+
+  // Try to prevent the context toolbar from getting below the editor toolbar if we know the bounds of the viewport.
+  cappingViewport.each((cap) => {
+    if ((newLeft + newWidth) > cap.width) {
+      newWidth = cap.width - newLeft;
+    }
+
+    if ((newTop + newHeight) > cap.height) {
+      newHeight = cap.height - newTop;
+    }
+  });
+
   const point = CssPosition.screen(SugarPosition(newLeft, newTop));
   return Optional.some(Boxes.pointed(point, newWidth, newHeight));
 };
--- a/modules/alloy/src/main/ts/ephox/alloy/positioning/mode/NodeAnchor.ts
+++ b/modules/alloy/src/main/ts/ephox/alloy/positioning/mode/NodeAnchor.ts
@@ -1,6 +1,7 @@
 import { FieldSchema } from '@ephox/boulder';
 import { Optional } from '@ephox/katamari';
 import { SugarBody } from '@ephox/sugar';
+import { Bounds } from '../../alien/Boxes';
 
 import { AlloyComponent } from '../../api/component/ComponentApi';
 import * as Fields from '../../data/Fields';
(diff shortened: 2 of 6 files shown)


HAFRMO (commented):
I've verified it manually, but I'm not sure where we'd want to put the tests.

modules/alloy/src/main/ts/ephox/alloy/positioning/mode/ContentAnchorCommon.ts
```diff
@@ -20,6 +23,18 @@ const capRect = (left: number, top: number, width: number, height: number): Opti
     newTop = 0;
     newHeight = height + top;
   }
+
+  // Try to prevent the context toolbar from getting below the editor toolbar if we know the bounds of the viewport.
```
ephox-mogran, line 27: Ideally, this code wouldn't mention "editor toolbar" or 'context toolbar" in a comment (it does above as well). It should only mention the values it understands (parameters passed in).

Also, `capRect` should have some unit tests. If only to give sample input -> output mappings.

modules/alloy/src/main/ts/ephox/alloy/positioning/mode/ContentAnchorCommon.ts
```diff
@@ -9,8 +9,11 @@ import * as Origins from '../layout/Origins';
 import { Anchoring, NodeAnchor, nu as NuAnchor, SelectionAnchor } from './Anchoring';
 import * as AnchorLayouts from './AnchorLayouts';
 
-const capRect = (left: number, top: number, width: number, height: number): Optional<Boxes.BoxByPoint> => {
-  let newLeft = left, newTop = top, newWidth = width, newHeight = height;
+const capRect = (left: number, top: number, width: number, height: number, cappingViewport: Optional<Boxes.Bounds>): Optional<Boxes.BoxByPoint> => {
```
ephox-mogran, line 12: I think this function could do with an overall comment. It seems to be doing something like:

Adjust the left, top, width, and height so that the rectangle is contained *completely* within the cappingViewport. Although at the moment, it is actually capping to (0, 0, cap.width, cap.height). You probably want to mention "why" it does that also.

modules/alloy/src/main/ts/ephox/alloy/positioning/mode/ContentAnchorCommon.ts
```diff
@@ -9,8 +9,11 @@ import * as Origins from '../layout/Origins';
...
+  let newLeft = left;
+  let newTop = top;
+  let newWidth = width;
+  let newHeight = height;
   // Try to prevent the context toolbar from getting above the editor toolbar
   if (left < 0) {
```
ephox-mogran, line 18: Are you intentionally capping to cap.width and cap.height for the bottom, but 0 and 0 for the top left? Should it be cap.left or cap.top or cap.x or cap.y ? And if not, should cappingViewport just be a width/height value rather than a rectangle to avoid confusion?

modules/alloy/src/main/ts/ephox/alloy/positioning/mode/NodeAnchor.ts
```diff
@@ -10,15 +11,16 @@ import * as AnchorLayouts from './AnchorLayouts';
 import * as ContainerOffsets from './ContainerOffsets';
 import * as ContentAnchorCommon from './ContentAnchorCommon';
 
-const placement = (component: AlloyComponent, anchorInfo: NodeAnchor, origin: Origins.OriginAdt): Optional<Anchoring> => {
+const placement = (component: AlloyComponent, anchorInfo: NodeAnchor, origin: Origins.OriginAdt, getBounds: Optional<() => Bounds>): Optional<Anchoring> => {
```
ephox-mogran, line 14: This stuff is notoriously difficult to test, because you feel like you are just asserting a value that could be wrong anyway, but you can write simpler property tests for it. If you can identify some invariant that should hold (e.g. box dimensions can never exceed something), then you can test that invariant without actually having to just calculate where it should go.

I thought I'd done this somewhere, but I can't find it currently.

modules/alloy/src/main/ts/ephox/alloy/positioning/mode/NodeAnchor.ts
```diff
@@ -10,15 +11,16 @@ import * as AnchorLayouts from './AnchorLayouts';
...
   const rootPoint = ContainerOffsets.getRootPoint(component, origin, anchorInfo);
 
   return anchorInfo.node
     // Ensure the node is still attached, otherwise we can't get a valid client rect for a detached node
     .filter(SugarBody.inBody)
     .bind((target) => {
+      const cappingViewport = Origins.viewport(origin, getBounds);
```
ephox-mogran, line 21: There are so many bounds-things everywhere in this code. Given that you are only using `getBounds` to pass it into `Origins.viewport`, would it make it slightly simpler to pass in cappingViewport here rather than getBounds? You'll need to check that the value hasn't changed. It's probably not much difference in readability, anyway.

ephox-mogran (commented):
I can't approve or disapprove. I really don't know enough to know if this is what you want to change or where you should be changing it, sorry. Most of this code seems new to me from when I last saw it (I could be wrong, though), so I'm unfamiliar with what it is trying to do.

What exactly is bounds etc. becomes a bit hard to understand, but that's already the case. I've left some comments. Feel free to ignore all of them.

lnewson (commented):
This seems like the wrong place to be making these changes @HAFRMO, as this will constrain all anchored elements to be within the viewport which is **not** expected behaviour. If that needs to happen, it should be done when passing in the anchor/positioning spec (can't recall which off the top of my head), which for context toolbars is computed here: https://github.com/tinymce/tinymce/blob/develop/modules/tinymce/src/themes/silver/main/ts/ui/context/ContextToolbarBounds.ts

modules/alloy/src/main/ts/ephox/alloy/positioning/mode/NodeAnchor.ts
lnewson: Was it this you're thinking of? https://github.com/tinymce/tinymce/blob/417779b3655c97a21789a23a5c0dcb489c239da2/modules/alloy/src/test/ts/atomic/position/view/BounderCalcRepositionTest.ts#L7

modules/alloy/src/main/ts/ephox/alloy/positioning/mode/ContentAnchorCommon.ts
HAFRMO: Seems like a good idea.

modules/alloy/src/main/ts/ephox/alloy/positioning/mode/ContentAnchorCommon.ts
HAFRMO: The square is provided as relative, and is later put into where exactly the view is. So the top can come as a negative value, which is then how far it is out of the view. So when the code caps it at 0, it caps it at the top of the viewable area.
I'm not sure why this was implemented this way, but it appears to be so as far as I could tell.

modules/alloy/src/main/ts/ephox/alloy/positioning/mode/NodeAnchor.ts
HAFRMO: Not sure if it'd make it simpler as such, but in terms of readability that might be a personal opinion. I'd be fine with this, but if people prefer it otherwise I can change it.

HAFRMO (commented):
I've updated the original issue, which was somewhat outdated, to more clarify the situation.

PR #6621, opened 2021-03-14
Prevent paste_convert_word_fake_lists from removing content and detect roman numerals correctly

aautio:
Hi,

Here are fixes to the MS Word paste cleanup.

- The first commit fixes issues with roman numerals, as documented in #6620 and #2933.
- The second commit fixes old and often reported issue about words disappearing when lists are pasted from MS Word: #2810, #3017, #3480, #5474 (and others).

Please see the full commit messages for rationale and details about the changes.

I'll be happy to elaborate or edit the PR according to your feedback.

Pre-checks:
* [ ] Changelog entry added
* [x] Tests have been added

GitHub issues:

- #6620
- #2810
- #2933
- #3017
- #3480
- #5474

Issue #2933, opened 2016-05-20
paste roman numbered list from MSWord 


PetrBorisov:
Create and copy roman numbered list and paste him in tinyMCE editor
expect:  simple numbered list
1)text
2)text
3)text
4)text ...
really: alternation lists
1)text
2)text
• 3
1)text ....

i know what i can use paste_convert_word_fake_lists: false options, but convertFakeListsToProperLists function work wrong

PR #6690, opened 2021-03-31
Fix disappearing letters and words from lists pasted from Word

### LINKED ISSUE #2810: Pasting from Word: List items starting with single word followed by a dot, will remove that word
Hi,

List items starting with a single word followed by a dot are removed because they are mistakenly being seen as part of the list.
Best to describe it with an example.

**From Word:**
- foo. bar
- foo bar. baz

**Result in TinyMCE:**
- bar
- foo bar. baz

I traced it back to [WordFilter.js](https://github.com/tinymce/tinymce/blob/master/src/plugins/paste/src/main/js/core/WordFilter.js) on (currently) line 184:
`trimListStart(paragraphNode, /^\s*([\u2022\u00b7\u00a7\u25CF]|\w+\.)/);`

The `\w+\.` part of the Regex is the cause of this removal.
I can only assume that this is needed for ordered lists starting with alpha characters (a. b. c.)

It is reproducible with the following Word document containing a small and simple list:
[simplelist.docx](https://github.com/tinymce/tinymce/files/190469/simplelist.docx)

On pasting this content to tinymce, the "foo." part of the first list gets removed.
https://www.tinymce.com/docs/demo/basic-example/

Tested it with Word 2008 and 2011 (mac)
Latest TinyMCE (version 4.3.8)

Let me know if you need more info.

### LINKED ISSUE #3017: Problem with copy & paste Word lists with numbers into editor
We detected the following behavior. When you copy & paste sorted or unsorted lists from Word into the editor, the tinymce removes number(s) before points. 

That happens when the number is placed at the beginning of the list. With double byte characters the problem exist at any position in the list. e.g.23.9 km/L  changed to 9 km/L 

You can see the behavior if you insert the attached example list (word) in the demo editor on https://www.tinymce.com. 
[editor_list_word.docx](https://github.com/tinymce/tinymce/files/327829/editor_list_word.docx)

Finger cross you have an idea for a solution.
Thanks, Kirk

### LINKED ISSUE #3480: Characters stripped when pasted into tinymce
We have found a bug relating to bullet points and decimal point numbers when pasting content into the TinyMCE editor.

If a bullet point is created in MS Word, and the value of 0.25 is written, eg:

- 0.25

Then you copy this, and paste into the TinyMCE editor, the editor strips the "0." from what is pasted.

For an example, see attached image :
image

When you paste this content, it should retain the full content without stripping any characters.

### LINKED ISSUE #5474: [ms word] soft line breaks in number lists
If i have a number list in a word document, which has a soft line break at the end of the first row and after the first word of the next line exists a full stop, this first word will not be appear in TinyMCE when i copy the content from the word document and insert it into the TinyMCE editor.

**Please provide the steps to reproduce and if possible a minimal demo of the problem via [fiddle.tinymce.com](http://fiddle.tinymce.com/) or similar.**
Make a number list with a soft line break at the end of the first row and set a full stop after the first word after the line break. (see screenshot1 point 2)
Then copy the content from word to TinyMCE and the first word with the full stop after the soft line break will be disappear. (see screenshot2 point 2)
For comparison in both screenshots is a soft line break without a full stop after the first word shown, which will be correct inserted by TinyMCE. (point 1 on both screenshots)

screenshot1
screenshot2

That the TinyMCE doesn't drop words of the content when it will be copied and inserted from a word document. Regardless if there is a full stop after the first word after the soft line break or not.

Version: TinyMCE 5.1.6
Browser: Safari on MacOS, Chrome & Firefox (MacOS & Windows)

This issue doesn't exist in TinyMCE with premium word plugin.

aautio:
This pull request replaces #6621 as agreed with @JamesToohey.

Fixes #2810, Fixes #3017, Fixes #3480, Fixes #5474

This commit includes fixes to multiple subtle errors in MS Word list
cleanups. The bugs caused any characters ending with a dot to disappear
from the lists under certain circumstances:

Font/style -changes cause multiple text nodes to appear as siblings in
the AST.

Function `filterStyles` marked only the first node with `_listIgnore` and
thus missed the nodes with roman numerals. This was fixed.

The `removeIgnoredNodes`-function used `node.remove()` to delete nodes
from the AST. The method sets `node.next` to `null` and thus recursion
did not proceed further. Sibling text nodes with `_listIgnore` would
not be deleted. This was fixed by avoiding `node.remove()`.

The `trimListStart` -function did not trim the beginning of the
paragraph. It trimmed all text nodes instead and removed the ones with
words ending with dot. This was fixed by merging the paragraph
text nodes before trimming the first text node.

Regex for cleaning up the beginning is now safe. It no longer removes
any letters (with` \w`). Instead it only looks for bullets. One
regex is enough to filter those as the text nodes have been merged
before matching.

Added a couple of test cases as a proof that the fixes work correctly.

Pre-checks:

Review:
* [ ] Milestone set

- #2933
- #3017
- #3480
- #5474
- possibly others as well

### DIFF
--- a/modules/tinymce/src/plugins/paste/main/ts/core/WordFilter.ts
+++ b/modules/tinymce/src/plugins/paste/main/ts/core/WordFilter.ts
@@ -89,28 +89,39 @@ const convertFakeListsToProperLists = (node) => {
     return txt;
   };
 
-  const trimListStart = (node, regExp) => {
-    if (node.type === 3) {
-      if (regExp.test(node.value)) {
-        node.value = node.value.replace(regExp, '');
-        return false;
+  const joinChildTextNodes = (parent) => {
+    let node = parent.firstChild;
+
+    while (node) {
+      const prev = node;
+      node = prev.next;
+
+      // when two sibling nodes are both text nodes
+      if (prev.type === 3 && node && node.type === 3) {
+        // merge their text content to the next child
+        node.value = prev.value + node.value;
+        // and remove the previous child
+        prev.remove();
       }
     }
+  };
 
-    if ((node = node.firstChild)) {
-      do {
-        if (!trimListStart(node, regExp)) {
-          return false;
-        }
-      } while ((node = node.next));
+  const trimFirstTextNode = (parent, regExp) => {
+    if ((node = parent.firstChild) && node.type === 3) {
+      node.value = node.value.replace(regExp, '');
     }
-
-    return true;
   };
 
   const removeIgnoredNodes = (node) => {
     if (node._listIgnore) {
-      node.remove();
+      // Remove any text content or remove the children from the node.
+      // node.remove() would not work here because it sets node.next and node.prev to null,
+      // thus preventing recursion.
+      if (node.type === 3) {
+        node.value = '';
+      } else {
+        node.empty();
+      }
       return;
     }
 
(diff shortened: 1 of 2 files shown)


metricjs:
Hey @aautio, apologies for not updating you regarding this PR in a while. We want to merge it, but we've noticed there's a breaking change in your changes. As such we also want to do some testing and other work on our side first. We had hoped to fit this in the TinyMCE 5.9 release, but that is looking unlikely. So, unfortunately, I'm going to remove this from the 5.9 milestone and we'll reschedule it for another future release.

TheSpyder:
Closing as discussed via email - I've opened the topic to search for a community maintainer at #7487.

PR #6709, opened 2021-04-06
feature/TINY-6400: Image dialog incorrectly renders class/caption in grid

camiloHimura:
Related Ticket: [TINY-6400](https://ephocks.atlassian.net/secure/RapidBoard.jspa?rapidView=116&modal=detail&selectedIssue=TINY-6400)

Description of Changes:
* Update tap config to forcing it into a 2 column grid if both a class list and caption are enabled

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Added
+- Updated the image dialog to show a grid with two columns only if `hasImageCaption` is true, `image` plugin #TINY-6400
 - Added support for alpha list numbering to the `lists` plugin #TINY-6891
 
 ### Fixed
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -7,12 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Added
-<<<<<<< HEAD
 - Updated the image dialog to display the class list dropdown as full-width if the caption checkbox is not present #TINY-6400
-=======
 - The `inline_boundaries` feature now supports the `home`, `end`, `pageup` and `pagedown` keys #TINY-4612
 - Added new `PAGE_UP` and `PAGE_DOWN` key code constants to the `VK` API
->>>>>>> edf347999fcf9581bfe9ecbc50b1f03eb14f0bba
 - Added support for alpha list numbering to the `lists` plugin #TINY-6891
 - The editor resize handle can now be controlled using the keyboard #TINY-4823
 
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -7,13 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Added
-- Updated the image dialog to display the class list dropdown as full-width if the caption checkbox is not present #TINY-6400
 - The `inline_boundaries` feature now supports the `home`, `end`, `pageup` and `pagedown` keys #TINY-4612
 - Added new `PAGE_UP` and `PAGE_DOWN` key code constants to the `VK` API
 - Added support for alpha list numbering to the `lists` plugin #TINY-6891
 - The editor resize handle can now be controlled using the keyboard #TINY-4823
 
 ### Changed
+- Updated the `image` dialog to display the class list dropdown as full-width if the caption checkbox is not present #TINY-6400
 - Renamed the "H Align" and "V Align" input labels in the Table Cell Properties dialog to "Horizontal align" and "Vertical align" respectively #TINY-7285
 
 ### Fixed
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Added
-- Updated the image dialog to show a grid with two columns only if `hasImageCaption` is true, `image` plugin #TINY-6400
+- Updated the image dialog to display the class list dropdown as full-width if the caption checkbox is not present #TINY-6400
 - Added support for alpha list numbering to the `lists` plugin #TINY-6891
 
 ### Fixed
--- a/modules/tinymce/src/plugins/image/main/ts/ui/MainTab.ts
+++ b/modules/tinymce/src/plugins/image/main/ts/ui/MainTab.ts
@@ -71,6 +71,8 @@ const makeItems = (info: ImageDialogInfo) => {
     ]
   };
 
+  const dialogType = (hasColumns: boolean) => hasColumns ? { type: 'grid', columns: 2 } : { type: 'panel' };
+
   return Arr.flatten<any>([
     [ imageUrl ],
     imageList.toArray(),
@@ -79,8 +81,7 @@ const makeItems = (info: ImageDialogInfo) => {
     info.hasImageTitle ? [ imageTitle ] : [],
     info.hasDimensions ? [ imageDimensions ] : [],
     [{
-      type: 'grid',
-      columns: 2,
+      ...dialogType(info.hasImageCaption),
       items: Arr.flatten([
         classList.toArray(),
         info.hasImageCaption ? [ caption ] : []
--- a/modules/tinymce/src/plugins/image/main/ts/ui/MainTab.ts
+++ b/modules/tinymce/src/plugins/image/main/ts/ui/MainTab.ts
@@ -71,7 +71,7 @@ const makeItems = (info: ImageDialogInfo) => {
     ]
   };
 
-  const dialogType = (hasColumns: boolean) => hasColumns ? { type: 'grid', columns: 2 } : { type: 'panel' };
+  const getDialogContainerType = (useColumns: boolean) => useColumns ? { type: 'grid', columns: 2 } : { type: 'panel' };
 
   return Arr.flatten<any>([
     [ imageUrl ],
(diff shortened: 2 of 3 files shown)


HAFRMO (approved):

SimonFc (approved):

modules/tinymce/src/plugins/image/main/ts/ui/MainTab.ts
```diff
@@ -71,6 +71,8 @@ const makeItems = (info: ImageDialogInfo) => {
     ]
   };
 
+  const dialogType = (hasColumns: boolean) => hasColumns ? { type: 'grid', columns: 2 } : { type: 'panel' };
```
JamesToohey, line 74: Should the parameter here match the value that's coming in?

```suggestion
  const dialogType = (hasImageCaption: boolean) => hasColumns ? { type: 'grid', columns: 2 } : { type: 'panel' };
```

modules/tinymce/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Added
+- Updated the image dialog to show a grid with two columns only if `hasImageCaption` is true, `image` plugin #TINY-6400
```
JamesToohey, line 10: Maybe something like

```suggestion
- Updated the image dialog to show a grid with two columns only if the `hasImageCaption` setting is set to `true` #TINY-6400
```

modules/tinymce/src/plugins/image/main/ts/ui/MainTab.ts
```diff
@@ -79,8 +81,7 @@ const makeItems = (info: ImageDialogInfo) => {
     info.hasImageTitle ? [ imageTitle ] : [],
     info.hasDimensions ? [ imageDimensions ] : [],
     [{
-      type: 'grid',
-      columns: 2,
+      ...dialogType(info.hasImageCaption),
```
ltrouton, line 84: I am wondering if the ternary is simple enough to just inline it?

modules/tinymce/src/plugins/image/test/ts/browser/plugin/MainTabTest.ts
```diff
@@ -47,4 +48,29 @@ describe('browser.tinymce.plugins.image.plugin.MainTabTest', () => {
...
+    const expected = {
+      ['.tox-form__group button[aria-label="Class"]']: 1,
+      ['.tox-form__grid--2col button[aria-label="Class"]']: 0
+    };
+    Assertions.assertPresence('Asserting presence', expected, dialog);
+  });
+
+  it('TINY-6400: render image_class_list and image_caption as Dialog type "group"', async () => {
+    const editor = hook.editor();
+    editor.execCommand('mceImage');
```
JamesToohey, line 67: Should we explicitly set `editor.settings.image_caption = true;`?

```suggestion
    const editor = hook.editor();
    editor.settings.image_caption = false;
    editor.execCommand('mceImage');
```

If so, we should probably set it back to `false` at the end of the test

modules/tinymce/CHANGELOG.md
ltrouton, line 10: I think this should maybe in the `Fixed` section instead of `Added`?
Possible alternative suggestion:
```suggestion
- Updated the image dialog to display the class list dropdown as full-width if the caption checkbox is not present #TINY-6400
```

modules/tinymce/src/plugins/image/main/ts/ui/MainTab.ts
lnewson: Maybe something like this instead, since it's purpose seems to be for general use?
```suggestion
  const getDialogContainerType = (useColumns: boolean) => useColumns ? { type: 'grid', columns: 2 } : { type: 'panel' };
```

modules/tinymce/src/plugins/image/test/ts/browser/plugin/MainTabTest.ts
```diff
@@ -47,4 +48,29 @@ describe('browser.tinymce.plugins.image.plugin.MainTabTest', () => {
...
+    const editor = hook.editor();
+    editor.settings.image_caption = false;
+    editor.execCommand('mceImage');
+
+    const dialog = await TinyUiActions.pWaitForDialog(editor);
+    const expected = {
+      ['.tox-form__group button[aria-label="Class"]']: 1,
+      ['.tox-form__grid--2col button[aria-label="Class"]']: 0
+    };
+    Assertions.assertPresence('Asserting presence', expected, dialog);
```
lnewson, line 62:
```suggestion
    Assertions.assertPresence('Does not have columns for the class', expected, dialog);
```
This really should say what it's asserting the presence of instead of just saying it's asserting something without any context.

modules/tinymce/src/plugins/image/test/ts/browser/plugin/MainTabTest.ts
```diff
@@ -47,4 +48,29 @@ describe('browser.tinymce.plugins.image.plugin.MainTabTest', () => {
       '</figure>'
     ));
   });
+
+  it('TINY-6400: render image_class_list as Dialog type "panel"', async () => {
+    const editor = hook.editor();
+    editor.settings.image_caption = false;
```
lnewson, line 54: This needs to be cleaned up at the end of the test so it doesn't impact other tests in this file. eg:
```
editor.settings.image_caption = true;
```

- [ ] Clean up the setting that was changed in this test case

modules/tinymce/src/plugins/image/test/ts/browser/plugin/MainTabTest.ts
lnewson: It's already set to true in the setup settings, so we shouldn't need to do that here. However, we do need to make sure the override is cleaned up in the previous test case.

modules/tinymce/src/plugins/image/main/ts/ui/MainTab.ts
lnewson, line 84: This isn't quite correct sorry, as it needs to fallback to a panel if either the class list or the caption is disabled. At the moment it'll only do that if the caption is disabled.

To word it differently it should only be using the grid layout with 2 columns when both the classes and caption need to be shown.

lnewson (changes requested):

modules/tinymce/src/plugins/image/main/ts/ui/MainTab.ts
camiloHimura: updated

modules/tinymce/CHANGELOG.md

modules/tinymce/src/plugins/image/test/ts/browser/plugin/MainTabTest.ts
camiloHimura: updated

modules/tinymce/src/plugins/image/main/ts/ui/MainTab.ts
camiloHimura: updated

modules/tinymce/src/plugins/image/test/ts/browser/plugin/MainTabTest.ts
camiloHimura: updated

modules/tinymce/CHANGELOG.md
```diff
@@ -7,8 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Added
+<<<<<<< HEAD
```
spocke, line 10: - [ ] Some merge conflict stuff that needs to be removed here.

spocke (approved):

modules/tinymce/src/plugins/image/test/ts/browser/plugin/MainTabTest.ts
```diff
@@ -47,4 +48,31 @@ describe('browser.tinymce.plugins.image.plugin.MainTabTest', () => {
...
+    editor.execCommand('mceImage');
+    const dialog = await TinyUiActions.pWaitForDialog(editor);
+    const expected = {
+      ['.tox-form__group button[aria-label="Class"]']: 1,
+      ['.tox-form__grid--2col button[aria-label="Class"]']: 0
+    };
+    Assertions.assertPresence('Does not have columns for the class', expected, dialog);
+    TinyUiActions.clickOnUi(editor, 'button.tox-button:contains(Cancel)');
+    delete editor.settings.image_caption;
```
lnewson, line 76:
```suggestion
    editor.settings.image_caption = true;
```

We shouldn't be deleting this, as we want it to go back to the original state, not the default state since we've set `true` in the initial configuration.

modules/tinymce/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Added
+- Updated the image dialog to display the class list dropdown as full-width if the caption checkbox is not present #TINY-6400
```
lnewson, line 10: This isn't an `Added`/New Feature, it's a `Changed` or `Improved` at best, so please make sure to move this to the appropriate heading.

- [ ] Move changelog entry to correct heading

```suggestion
- Updated the `image` dialog to display the class list dropdown as full-width if the caption checkbox is not present #TINY-6400
```
Also just another minor thing, as image is the plugin name it should be in back ticks.

modules/tinymce/src/plugins/image/test/ts/browser/plugin/MainTabTest.ts
```diff
@@ -47,4 +48,31 @@ describe('browser.tinymce.plugins.image.plugin.MainTabTest', () => {
...
+    const expected = {
+      ['.tox-form__grid--2col button[aria-label="Class"]']: 1,
+      ['.tox-form__grid--2col .tox-checkbox']: 1
+    };
+
+    Assertions.assertPresence('Does have columns for the class', expected, dialog);
+    TinyUiActions.clickOnUi(editor, 'button.tox-button:contains(Cancel)');
+  });
+
+  it('TINY-6400: render image_class_list as Dialog type "panel"', async () => {
```
lnewson, line 65: Can you please also add a test for when `image_caption` is true and `image_class_list` is not defined?

- [ ] Add addition test case for the issue I mentioned needed to be fixed in my previous review

lnewson (approved):

modules/tinymce/CHANGELOG.md

PR #6710, opened 2021-04-06
WIP

ephox-mogran:
Sort of the "marked as duplicate task": https://ephocks.atlassian.net/browse/TINY-7191 

So this is an attempt to get part of what Fredrik was wanting. Feel free to ignore all of it ... it's likely the wrong way to go about it. The basic idea is:

* add some new layout types that just use the anchor box (which is the table). They appear at the top or the bottom
* stop capping the anchor box
* get the "docking" behaviour by using the limitX and limitY already in Bounder

The animation will probably be OK for top -> top or bottom -> bottom, but it won't work for top -> bottom etc. There is a comment explaining what might resolve the problem.

There is also an issue with carets and maxWidth / maxHeight calculations. Almost entirely uninvestigated.

You may also want to consider only these changes for a table context menu.

### DIFF
--- a/modules/alloy/src/main/ts/ephox/alloy/positioning/layout/Layout.ts
+++ b/modules/alloy/src/main/ts/ephox/alloy/positioning/layout/Layout.ts
@@ -38,7 +38,9 @@ const southeast: AnchorLayout = (anchor: AnchorBox, element: AnchorElement, bubb
   bubbles.southeast(),
   Direction.southeast(),
   boundsRestriction(anchor, { left: AnchorBoxBounds.LeftEdge, top: AnchorBoxBounds.BottomEdge }),
-  'layout-se');
+  'layout-se',
+  false
+);
 
 const southwest: AnchorLayout = (anchor: AnchorBox, element: AnchorElement, bubbles: Bubble) => NuSpotInfo(
   westX(anchor, element),
@@ -46,7 +48,8 @@ const southwest: AnchorLayout = (anchor: AnchorBox, element: AnchorElement, bubb
   bubbles.southwest(),
   Direction.southwest(),
   boundsRestriction(anchor, { right: AnchorBoxBounds.RightEdge, top: AnchorBoxBounds.BottomEdge }),
-  'layout-sw'
+  'layout-sw',
+  false
 );
 
 const northeast: AnchorLayout = (anchor: AnchorBox, element: AnchorElement, bubbles: Bubble) => NuSpotInfo(
@@ -55,7 +58,8 @@ const northeast: AnchorLayout = (anchor: AnchorBox, element: AnchorElement, bubb
   bubbles.northeast(),
   Direction.northeast(),
   boundsRestriction(anchor, { left: AnchorBoxBounds.LeftEdge, bottom: AnchorBoxBounds.TopEdge }),
-  'layout-ne'
+  'layout-ne',
+  false
 );
 
 const northwest: AnchorLayout = (anchor: AnchorBox, element: AnchorElement, bubbles: Bubble) => NuSpotInfo(
@@ -64,7 +68,8 @@ const northwest: AnchorLayout = (anchor: AnchorBox, element: AnchorElement, bubb
   bubbles.northwest(),
   Direction.northwest(),
   boundsRestriction(anchor, { right: AnchorBoxBounds.RightEdge, bottom: AnchorBoxBounds.TopEdge }),
-  'layout-nw'
+  'layout-nw',
+  false
 );
 
 const north: AnchorLayout = (anchor: AnchorBox, element: AnchorElement, bubbles: Bubble) => NuSpotInfo(
@@ -73,7 +78,8 @@ const north: AnchorLayout = (anchor: AnchorBox, element: AnchorElement, bubbles:
   bubbles.north(),
   Direction.north(),
   boundsRestriction(anchor, { bottom: AnchorBoxBounds.TopEdge }),
-  'layout-n'
+  'layout-n',
+  false
 );
 
 const south: AnchorLayout = (anchor: AnchorBox, element: AnchorElement, bubbles: Bubble) => NuSpotInfo(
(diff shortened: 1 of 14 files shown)


modules/alloy/src/main/ts/ephox/alloy/positioning/layout/Layout.ts
```diff
@@ -38,15 +38,18 @@ const southeast: AnchorLayout = (anchor: AnchorBox, element: AnchorElement, bubb
   bubbles.southeast(),
   Direction.southeast(),
   boundsRestriction(anchor, { left: AnchorBoxBounds.LeftEdge, top: AnchorBoxBounds.BottomEdge }),
-  'layout-se');
+  'layout-se',
+  false
```
ephox-mogran, line 42: Find something better than a boolean.

modules/alloy/src/main/ts/ephox/alloy/positioning/view/Bounder.ts
```diff
@@ -150,7 +150,7 @@ const attempt = (candidate: SpotInfo, width: number, height: number, bounds: Box
 
   // Take special note that we don't use the futz values in the nofit case; whether this position is a good fit is separate
   // to ensuring that if we choose it the popup is actually on screen properly.
-  return originInBounds && sizeInBounds ? adt.fit(reposition) : adt.nofit(reposition, deltaW, deltaH);
+  return (originInBounds && sizeInBounds) || candidate.fitAnyway ? adt.fit(reposition) : adt.nofit(reposition, deltaW, deltaH);
```
ephox-mogran, line 153: I didn't have time to look into whether we could get rid of this hacky flag. Essentially, pinned layouts don't need to fit, because they are just going to be limited to within the bounds. The PinnedLayouts are the only ones that set fitAnyway to true.

modules/alloy/src/main/ts/ephox/alloy/positioning/view/PositionCss.ts
```diff
@@ -26,6 +26,19 @@ const NuPositionCss = (
 const applyPositionCss = (element: SugarElement, position: PositionCss): void => {
   const addPx = (num: number) => num + 'px';
 
+  /*
```
ephox-mogran, line 29: If you want to animate the context toolbar between a state of top and bottom, I think you'll need to create an intermediate state that matches the flow of the destination, and reflow before it. I didn't have time to investigate exactly what you'd need to do this. Origin.reposition might have some approaches.

modules/tinymce/src/themes/silver/main/ts/ContextToolbar.ts
```diff
@@ -52,11 +53,26 @@ const anchorOverrides = {
...
-const desktopAnchorSpecLayouts = {
-  onLtr: () => [ Layout.north, Layout.south, Layout.northeast, Layout.southeast, Layout.northwest, Layout.southwest,
-    LayoutInside.north, LayoutInside.south, LayoutInside.northeast, LayoutInside.southeast, LayoutInside.northwest, LayoutInside.southwest ],
-  onRtl: () => [ Layout.north, Layout.south, Layout.northwest, Layout.southwest, Layout.northeast, Layout.southeast,
-    LayoutInside.north, LayoutInside.south, LayoutInside.northwest, LayoutInside.southwest, LayoutInside.northeast, LayoutInside.southeast ]
+
+const getDesktopAnchorSpecLayouts = (contextbar: AlloyComponent) => {
+  // We try to keep our current pinned layout, so we identify any classes that are on the contextbar,
+  // and prefer that pinned layout to the other pinned layout
+  const getPinnedLayouts = () => {
```
ephox-mogran, line 60: So that it doesn't just jump back to the top as soon as it can, the pinned layouts "remember" their current state and prefer it. They just do that by putting the preference list in the order so that the first preferred pinned state is the current state. Pretty hackily identified.

modules/tinymce/src/themes/silver/main/ts/ContextToolbar.ts
```diff
@@ -79,7 +95,7 @@ const getAnchorLayout = (position: InlineContent.ContextPosition, isTouch: boole
   } else {
     return {
       bubble: Bubble.nu(0, bubbleSize, bubbleAlignments),
-      layouts: isTouch ? mobileAnchorSpecLayouts : desktopAnchorSpecLayouts,
+      layouts: isTouch ? mobileAnchorSpecLayouts : getDesktopAnchorSpecLayouts(contextbar),
```
ephox-mogran, line 98: The `contextbar` is required because we detect what the current layout is on it (for keeping pinned layouts the same). Unfortunately, that information isn't available in the onLtr function ... it just gets the node anchor, not the element being shown.

modules/tinymce/src/themes/silver/main/ts/ContextToolbar.ts
```diff
@@ -310,7 +326,7 @@ const register = (editor: Editor, registryContextToolbars, sink: AlloyComponent,
     editor.on('focusout', (_e) => {
       Delay.setEditorTimeout(editor, () => {
         if (Focus.search(sink.element).isNone() && Focus.search(contextbar.element).isNone()) {
-          close();
+          // close();
```
ephox-mogran, line 329: Reinstate this. You just comment this out when you want a less painful debugging experience.

notes/context-toolbar-changes.md
```diff
@@ -0,0 +1,73 @@
+Changes made
```
ephox-mogran, line 1: This might be out-of-date now. I was just taking notes as I went.

modules/alloy/src/main/ts/ephox/alloy/positioning/view/Bounder.ts
```diff
@@ -150,7 +150,7 @@ const attempt = (candidate: SpotInfo, width: number, height: number, bounds: Box
```
ephox-mogran, line 150: There is some code here for maxHeight and maxWidth that is wrong now, also.

ephox-mogran:
Meant to do this from the fork, sorry. I'll close it once I've copied the comments across.

PR #6717, opened 2021-04-07
TINY-7227: Support formatting multiple classes on same element

camiloHimura:
Related Ticket: [7227](https://ephocks.atlassian.net/secure/RapidBoard.jspa?rapidView=116&modal=detail&selectedIssue=TINY-7227)

Description of Changes:
* Match multiple classes on the same element

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -7,13 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Added
-- Update `MatchFormat` to match multiple values on same element #TINY-7227
 - The `inline_boundaries` feature now supports the `home`, `end`, `pageup` and `pagedown` keys #TINY-4612
 - Added new `PAGE_UP` and `PAGE_DOWN` key code constants to the `VK` API
 - Added support for alpha list numbering to the `lists` plugin #TINY-6891
 - The editor resize handle can now be controlled using the keyboard #TINY-4823
 
 ### Changed
+- Updated `formatter.matchFormat` to support for matching formats with variables in the classes property #TINY-7227
 - Updated the `image` dialog to display the class list dropdown as full-width if the caption checkbox is not present #TINY-6400
 - Renamed the "H Align" and "V Align" input labels in the Table Cell Properties dialog to "Horizontal align" and "Vertical align" respectively #TINY-7285
 
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Added
+- Update `MatchFormat` to match multiple classes on same element #TINY-7227
 - Added support for alpha list numbering to the `lists` plugin #TINY-6891
 - The editor resize handle can now be controlled using the keyboard #TINY-4823
 
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Added
-- Update `MatchFormat` to match multiple classes on same element #TINY-7227
+- Update `MatchFormat` to match multiple values on same element #TINY-7227
 - The `inline_boundaries` feature now supports the `home`, `end`, `pageup` and `pagedown` keys #TINY-4612
 - Added new `PAGE_UP` and `PAGE_DOWN` key code constants to the `VK` API
 - Added support for alpha list numbering to the `lists` plugin #TINY-6891
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -13,8 +13,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The editor resize handle can now be controlled using the keyboard #TINY-4823
 - Added a new `fixed_toolbar_container_target` setting which renders the toolbar in the specified `HTMLElement`. Patch contributed by pvrobays
 
+### Improved
+- Updated the `formatter.matchFormat` API to support matching formats with variables in the `classes` property #TINY-7227
+
 ### Changed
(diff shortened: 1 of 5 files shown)


modules/tinymce/src/core/main/ts/fmt/MatchFormat.ts
```diff
@@ -92,7 +92,11 @@ const matchItems = (dom: DOMUtils, node: Node, format, itemName: string, similar
             return;
           }
 
-          if ((!similar || format.exact) && !isEq(value, FormatUtils.normalizeStyleValue(dom, FormatUtils.replaceVars(items[key], vars), key))) {
+          const normalizedStyle = FormatUtils.normalizeStyleValue(dom, FormatUtils.replaceVars(items[key], vars), key);
+          // TINY-7227 multiple classes on same element
+          const matchAttribute = Arr.exists(value.split(' '), Fun.curry(isEq, normalizedStyle));
+
+          if ((!similar || format.exact) && !matchAttribute) {
```
HAFRMO, line 99: Should this also check if the whole value equals what we're looking for, to preserve the current behavior, in case that it could support values with spaces in them? I don't know where exactly MatchFormat might be used, so if this is never the case, feel free to ignore this.

SimonFc: Is there some attribute where spaces are not to be interpreted as a separator? Perhaps in some custom data-attribute? In that case yeah we should preserve the current behavior.

spocke (changes requested):
I'm really confused about this one. I don't think this one is needed the formatter already has support for this using the classes array. This should either use that or use custom table specific logic. %value should be an exact match so should attribute. The classes will check if all of those classes exists then it matches.

So the test would be something like this and still match:
```
const editor = hook.editor();
editor.formatter.register('formatA', { selector: 'p', classes: ['A']});
editor.formatter.register('formatB', { selector: 'p', classes: ['B']});
editor.formatter.register('formatC', { selector: 'p', classes: ['C']});

editor.setContent('<p class="A B C">test</p>');
LegacyUnit.setSelection(editor, 'p', 0, 'p', 0);

assert.isTrue(editor.formatter.match('formatA'), 'Should match since the onmatch matches on "A" class.');
assert.isTrue(editor.formatter.match('formatB'), 'Should match since the onmatch matches on "B" class.');
assert.isTrue(editor.formatter.match('formatC'), 'Should match since the onmatch matches on "C" class.');
```

Value is more to be used for variable formats like font-size or color where we can't have predefined formats. If the table ui matches or toggles on/off arbitrary classes then that logic should be table specific.

camiloHimura:
@spocke I should haven't used 'class' as a property for that test. I'll change it.

modules/tinymce/src/core/main/ts/fmt/MatchFormat.ts
camiloHimura: updated, also the whole value

spocke:
Ok this just needs one code change in the formatter to support formats like { selector: 'td', classes: [ '%value' ] } it seems that ApplyFormat and RemoveFormat already replaces variables. But to make sure the tests should check so that is the case.

https://github.com/tinymce/tinymce/blob/develop/modules/tinymce/src/core/main/ts/fmt/MatchFormat.ts#L128
```
if (!ed.dom.hasClass(node, FormatUtils.replaceVars(classes[x], vars))) {
````

File: modules/tinymce/src/core/test/ts/browser/FormatterCheckTest.ts [hunk 2]
```diff
@@ -260,4 +260,36 @@ describe('browser.tinymce.core.FormatterCheckTest', () => {
...
+
+    editor.setContent('<p class="a b c">test</p>');
+    LegacyUnit.setSelection(editor, 'p', 0, 'p', 0);
+
+    assert.isTrue(editor.formatter.match('formatA', { value: 'a' }), 'Should match since the onmatch matches on "A" value.');
+    assert.isTrue(editor.formatter.match('formatA', { value: 'b' }), 'Should match since the onmatch matches on "B" value.');
+    assert.isTrue(editor.formatter.match('formatA', { value: 'c' }), 'Should match since the onmatch matches on "C" value.');
+  });
+
+  it('TINY-7227: mach whole value with spaces', () => {
```
spocke, line 286: Change mach to match

modules/tinymce/src/core/test/ts/browser/FormatterCheckTest.ts
```diff
@@ -260,4 +260,36 @@ describe('browser.tinymce.core.FormatterCheckTest', () => {
...
+    const editor = hook.editor();
+    editor.formatter.register('formatA', { selector: 'p', classes: [ '%value' ] });
+
+    editor.setContent('<p class="a b c">test</p>');
+    LegacyUnit.setSelection(editor, 'p', 0, 'p', 0);
+
+    assert.isTrue(editor.formatter.match('formatA', { value: 'b' }), 'Should match since the onmatch matches on "b" class.');
+  });
+
+  it('TINY-7227: match multiple values on an element with multiple classes', () => {
```
spocke, line 274: Also add a test where it checks if match returns false since in theory that function could not always be returning true.

modules/tinymce/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Added
+- Update `MatchFormat` to match multiple values on same element #TINY-7227
```
spocke, line 10: This is more like:
Updated `formatter.matchFormat` to support for matching formats with variables in the classes property
or
Added support for matching formats with variables in the classes property

spocke (approved):

modules/tinymce/CHANGELOG.md

HAFRMO (approved):

lnewson (approved):

modules/tinymce/CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The editor resize handle can now be controlled using the keyboard #TINY-4823
 
 ### Changed
+- Updated `formatter.matchFormat` to support for matching formats with variables in the classes property #TINY-7227
```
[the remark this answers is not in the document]

PR #6724, opened 2021-04-11
TINY-6625: Table plugin tries to perform operations on tables that contain the (inline) editor

vpyshnenko:
Related Ticket: TINY-6625

Description of Changes:
* Placeholder text

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/plugins/table/test/ts/browser/InlineEditorInsideTableTest.ts
+++ b/modules/tinymce/src/plugins/table/test/ts/browser/InlineEditorInsideTableTest.ts
@@ -2,14 +2,14 @@ import { Mouse, UiFinder } from '@ephox/agar';
 import { context, describe, it } from '@ephox/bedrock-client';
 import { TinyDom, TinyHooks, TinyUiActions } from '@ephox/mcagar';
 import { Attribute, Html, Insert, Remove, SelectorFind, SugarBody, SugarElement } from '@ephox/sugar';
-import {assert} from 'chai';
+import { assert } from 'chai';
 
 import Editor from 'tinymce/core/api/Editor';
 import * as EditorFocus from 'tinymce/core/focus/EditorFocus';
 import Plugin from 'tinymce/plugins/table/Plugin';
 import Theme from 'tinymce/themes/silver/Theme';
 
-const count = (selector: string) => UiFinder.findAllIn(SugarBody.body(), selector).length
+const count = (selector: string) => UiFinder.findAllIn(SugarBody.body(), selector).length;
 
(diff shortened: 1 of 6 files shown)


metricjs (approved):
Looks good to me! I do wonder if we should add more test cases, e.g. checking that `mceTableDelete` doesn't work, and maybe that things like adding a column shouldn't work? I'm not sure what would be the best things to test though, and worried the scope might blow out a lot. If you haven't already, it might be good to at least manually test a few things like that and see if anything else breaks?

vpyshnenko:
Good point. I've checked and turns out the outside table is still accessible through commands. I'm gonna fix it.

modules/tinymce/src/plugins/table/test/ts/browser/InlineEditorInsideTableTest.ts
```diff
@@ -35,13 +39,97 @@ describe('browser.tinymce.plugins.table.InlineEditorInsideTableTest', () => {
     inline: true,
     base_url: '/project/tinymce/js/tinymce',
     plugins: 'table',
-    menubar: false,
+    menubar: 'table',
     statusbar: false
   }, setupElement, [ Plugin, Theme ]);
 
+  const focusEditor = (editor: Editor) => {
```
lnewson, line 46: Why is this needed? Can we just use `editor.focus()` instead, as we generally don't like to use internal core modules like this?

Additionally, this seems to be called in every test, so can we move this to a `beforeEach` instead please?

- [ ] Use `editor.focus()` and consider a `beforeEach` hook instead please.

modules/tinymce/src/plugins/table/test/ts/browser/InlineEditorInsideTableTest.ts
```diff
@@ -35,13 +39,97 @@ describe('browser.tinymce.plugins.table.InlineEditorInsideTableTest', () => {
...
+      const td = editor.getBody().parentElement;
+      assert.notEqual(td.style.backgroundColor, 'pink');
+    });
+
+    it('TINY-6625: mceTableInsertRowBefore', () => {
+      const editor = hook.editor();
+      focusEditor(editor);
+      const beforeCount = count('tr');
+      editor.execCommand('mceTableInsertRowBefore');
+      assert.equal(count('tr'), beforeCount);
```
lnewson, line 84: For debugging purposes, it could be better if the count function returned the array of elements and then used `assert.lengthOf` as that means it can print what it found in the error message.

lnewson (approved):

lnewson (approved):

ltrouton (approved):

PR #6737, opened 2021-04-13
TINY-7125: Change the apperance of the table resize handle

SimonFc:
Related Ticket: TINY-7125

Description of Changes:
* Change the apperance of the table resize handle

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/oxide/src/less/theme/content/selection/selection.less
+++ b/modules/oxide/src/less/theme/content/selection/selection.less
@@ -112,7 +112,7 @@
 
 // Table border resizing handles
 .ephox-snooker-resizer-bar {
-  background-color: @object-block-selected-color;
+  background-color: @resize-handle-color;
   opacity: 0;
   -webkit-user-select: none;
   -moz-user-select: none;
@@ -122,10 +122,28 @@
 
 .ephox-snooker-resizer-cols {
   cursor: col-resize;
+
+  // Increase the area of the resize handle
+  &::after {
+    content: "";
+    display: block;
+    height: 100%;
+    margin-left: -3px;
+    width: 8px;
+  }
 }
 
 .ephox-snooker-resizer-rows {
   cursor: row-resize;
+
+  // Increase the area of the resize handle
+  &::after {
+    content: "";
+    display: block;
+    height: 8px;
+    margin-top: -3px;
+    width: 100%;
+  }
 }
 
 .ephox-snooker-resizer-bar.ephox-snooker-resizer-bar-dragging {
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -14,7 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - Renamed the "H Align" and "V Align" input labels in the Table Cell Properties dialog to "Horizontal align" and "Vertical align" respectively #TINY-7285
-- Changed the apperance of the row & column resize handle in the table plugin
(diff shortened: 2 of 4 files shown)


modules/tinymce/CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - Renamed the "H Align" and "V Align" input labels in the Table Cell Properties dialog to "Horizontal align" and "Vertical align" respectively #TINY-7285
+- Changed the apperance of the row & column resize handle in the table plugin
```
ashfordneil, line 17:
```suggestion
- Changed the apperance of the row & column resize handles in the table plugin
```

ashfordneil (approved):

metricjs (approved):
If I'm reading this right, we're moving much of the width of the bars into the CSS, I assume so the styles can be modified more for different skins?

SimonFc:
> If I'm reading this right, we're moving much of the width of the bars into the CSS, I assume so the styles can be modified more for different skins?

The width of the blue handle is still hard-coded in snooker (2px) and the extra CSS in oxide is for retaining the same "clickable" area of the handle so as to make it easier for the user to select the bar (the before/after elements are transparent). It becomes a bit divided between snooker and the skin but I _think_ that we don't want to have these pseudo-elements added by snooker itself.

lnewson:
As screenshots are often more useful, this is what the resizer looks like with this change:
image

modules/oxide/src/less/theme/content/selection/selection.less
```diff
@@ -122,10 +122,28 @@
 
 .ephox-snooker-resizer-cols {
   cursor: col-resize;
+
+  // Increase the area of the resize handle
+  &::after {
+    content: "";
+    display: block;
+    height: 100%;
+    margin-left: -3px;
```
lnewson, line 131: These values don't account for the fact the `after` psuedo selector is placed after the element so it should be `-2px` here instead otherwise it'll be offset 1px to the left of center. eg this is what it currently looks like:

image

modules/tinymce/CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - Renamed the "H Align" and "V Align" input labels in the Table Cell Properties dialog to "Horizontal align" and "Vertical align" respectively #TINY-7285
+- Changed the apperance of the row & column resize handles in the table plugin
```
lnewson, line 17: This is missing the JIRA number at the end of the changelog entry.

lnewson (changes requested):

modules/oxide/src/less/theme/content/selection/selection.less
```diff
@@ -112,7 +112,7 @@
 
 // Table border resizing handles
 .ephox-snooker-resizer-bar {
-  background-color: @object-block-selected-color;
+  background-color: @resize-handle-color;
```
lnewson, line 115: Do we need a changelog entry for this new Oxide variable?

lostkeys: The variable isn't new, it's declared in `resize-handle.less`

lnewson:
I'm closing this, as it's not been re-assigned and doesn't look like it's going to be done anytime soon. When it's ready with the fixes mentioned, lets create a new PR instead.

PR #6759, opened 2021-04-22
TINY-6638: Allow table tab navigation with ranged selections

JamesToohey:
Related Ticket: TINY-6638

Description of Changes:
Allow table tab navigation with ranged selections

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/plugins/table/main/ts/queries/TabContext.ts
+++ b/modules/tinymce/src/plugins/table/main/ts/queries/TabContext.ts
@@ -13,6 +13,7 @@ import Editor from 'tinymce/core/api/Editor';
 import VK from 'tinymce/core/api/util/VK';
 
 import * as Util from '../core/Util';
+import { CellSelectionApi } from '../selection/CellSelection';
 
 const forward = (editor: Editor, isRoot: (e: SugarElement) => boolean, cell: SugarElement<HTMLTableCellElement>) => {
   return go(editor, isRoot, CellNavigation.next(cell));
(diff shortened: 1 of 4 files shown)


vpyshnenko (approved):

modules/tinymce/src/plugins/table/main/ts/queries/TabContext.ts
```diff
@@ -60,17 +61,18 @@ const handle = (event: KeyboardEvent, editor: Editor) => {
...
-        const rng = navigation(editor, isRoot, cell);
-        rng.each((range) => {
-          editor.selection.setRng(range);
-        });
+    // If navigating backwards, use the start of the ranged selection
+    const container = SugarElement.fromDom(event.shiftKey ? rng.startContainer : rng.endContainer);
+    TableLookup.cell(container, isRoot).each((cell) => {
+      event.preventDefault();
+      // Clear fake ranged selection because our new selection will always be collapsed
+      TableLookup.table(cell, isRoot).each(cellSelection.clear);
```
lnewson, line 69: Just wondering, should we only do this when setting the range, as that means it's found a valid navigation location?

modules/tinymce/src/plugins/table/test/ts/browser/TabKeyNavigationTest.ts
```diff
@@ -74,4 +74,114 @@ describe('browser.tinymce.plugins.table.TabKeyNavigationTest', () => {
     assert.equal(events[1].type, 'tablemodified');
     editor.off('TableModified', logEvent);
   });
+
+  context('TINY-6683: Tab navigation with ranged selection', () => {
```
lnewson, line 78: We haven't previously included JIRA numbers in contexts, as it can contain multiple tests with different JIRA numbers, so maybe this instead?
```suggestion
  context('Tab navigation with ranged selection', () => {
```

lnewson (approved):

ltrouton (approved):

modules/tinymce/src/plugins/table/main/ts/queries/TabContext.ts
JamesToohey: Discussed in Slack: Moving this will mean the fake selection is present when adding rows (which uses the `mceTableInsertRowAfter` command) resulting in multiple rows. So we'll leave this where it is.

PR #6613, opened 2021-03-11
TINY-7475 New table related icons...

lostkeys:
and some minor changes to existing ones related to tables.

Related Ticket: TINY-7157

# Description of Changes:
### New icons
* border-style
* cell-background-color
* cell-border-color
* table-caption
* table-cell-classes
* table-classes
* vertical-align

### Changed icons
* border-width – got a minor refinement

# Pre-checks:
* [x] ~~Changelog entry added~~
* [x] ~~Tests have been added (if applicable)~~
* [x] ~~Branch prefixed with `feature/` for new features (if applicable)~~
* [x] ~~License headers added on new files (if applicable)~~

# Review:
* [x] Milestone set
* [x] Review comments resolved

# GitHub issues (if applicable):
_Not applicable_

PR #6761, opened 2021-04-22
Feature/tiny 7163: Added new table buttons and commands.

HAFRMO:
Related Ticket:

Description of Changes:
Added a series of new features and buttons to the table plugin, primarily related to the manipulation of headers and classes. This PR also includes the TINY-7157-branch, to ensure that all icons are included as well as for speed.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


JamesToohey:
- [ ] Fix pull request title

HAFRMO:
Regarding the LGTM comments, as far as I can tell they're referring to this line `canvasContext.fillStyle = '#FFFFFF'; // lgtm[js/useless-assignment-to-property]`, which resets the color of the fillstyle to white before setting it to the intended color. If the intended color is invalid for whatever reason, this white will be used instead.

lostkeys:
We're getting there! A couple of found issues

### Issue 1

* [ ] When the selection looks like below, the row header button should not be enabled

Screen Shot 2021-04-26 at 11 19 58

### Issue 2

* [ ] When the selection spans a header row and a regular row, like in the image above, pressing the button should convert all selected rows into header rows. Currently nothing happens when I press the row header button. 

### Issue 3

* [ ] When **toggling off** a column header, the header rows should still be intact. In the case shown below, when using `table_header_type: "sectionCells"` the cells in the thead should remain <th> and not become <td>. 

Screen Shot 2021-04-26 at 11 29 37

lostkeys:
### Issue 4
* [ ] Remove color should be automatically inserted as the last color swatch, in this case after white. This applies to both cell background and cell border color
Screen Shot 2021-04-26 at 11 37 54

modules/bridge/src/main/ts/ephox/bridge/components/menu/FancyMenuItem.ts
```diff
@@ -4,13 +4,15 @@ import { Result, Fun } from '@ephox/katamari';
...
   fancytype: string;
+  initData?: Record<string, any>;
   onAction: (data: any) => void;
 }
 
 export interface FancyMenuItem {
   type: 'fancymenuitem';
   fancytype: keyof FancyActionArgsMap;
   onAction: <K extends keyof FancyActionArgsMap>(data: FancyActionArgsMap[K]) => void;
+  initData?: Record<string, any>;
```
lnewson, line 15: No optional items in a final spec please. This should default to an empty object instead of `undefined` basically.
```suggestion
  initData: Record<string, any>;
```

lnewson:
I'm changing the base branch on this to use @lostkeys PR from https://github.com/tinymce/tinymce/pull/6613, as that way we can focus on reviewing only new changes.

modules/bridge/src/main/ts/ephox/bridge/components/menu/FancyMenuItem.ts
```diff
@@ -23,7 +25,8 @@ const fancyTypes: (keyof FancyActionArgsMap)[] = [ 'inserttable', 'colorswatch'
 export const fancyMenuItemSchema = ValueSchema.objOf([
   FieldSchema.strictString('type'),
   FieldSchema.strictStringEnum('fancytype', fancyTypes),
-  FieldSchema.defaultedFunction('onAction', Fun.noop)
+  FieldSchema.defaultedFunction('onAction', Fun.noop),
+  FieldSchema.defaulted('initData', undefined)
```
lnewson, line 29:
```suggestion
  FieldSchema.defaulted('initData', {})
```
To match my comment above.

modules/tinymce/CHANGELOG.md
```diff
@@ -10,6 +10,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The `inline_boundaries` feature now supports the `home`, `end`, `pageup` and `pagedown` keys #TINY-4612
 - Added new `PAGE_UP` and `PAGE_DOWN` key code constants to the `VK` API
 - Added support for alpha list numbering to the `lists` plugin #TINY-6891
+- Added new buttons for newly created table actions #TINY-7163
+- Added new icons for the newly added table actions #TINY-7157
+- Added the new command `mceTableToggleColumnHeader` which toggles currently selected column to or from a header #TINY-7163
```
lnewson, line 15: When we were doing the table work in TinyMCE 5.5, we specifically said we did **not** want toggle commands. I'm trying to remember why, but I believe it was because many have more than 2 states and as such it didn't make sense to include them as they'd be confusing.

@metricjs can you recall anything else about why we opted not to provide toggle commands for header types?

modules/tinymce/src/core/demo/ts/demo/FullDemo.ts
```diff
@@ -1,5 +1,5 @@
-/* eslint-disable no-console */
 import { Merger } from '@ephox/katamari';
+/* eslint-disable no-console */
```
lnewson, line 2: Why was this moved?

modules/tinymce/src/plugins/table/main/ts/api/Commands.ts
```diff
@@ -49,6 +51,93 @@ const registerCommands = (editor: Editor, actions: TableActions, cellSelection:
...
+    Arr.each(usedColumns, (value) => {
+      Arr.each(warehouse.all, (row) => {
+        const columnCell = Arr.find(row.cells, (cell) => {
+          return cell.column === value;
+        });
+
+        columnCell.each((cell) => {
+          const theadOpt = SelectorFind.ancestor(cell.element, 'thead');
+
+          if (theadOpt.isNone() && Util.getNodeName(cell.element.dom) !== changeTo) {
```
lnewson, line 94: We should be using `SugarNode.name(cell.element)` instead here.

modules/tinymce/src/plugins/table/main/ts/api/Commands.ts
```diff
@@ -49,6 +51,93 @@ const registerCommands = (editor: Editor, actions: TableActions, cellSelection:
...
+
+        Events.fireTableModified(editor, table.dom, Events.structureModified);
+      });
+    });
+  };
+
+  const toggleCaption = (_ui: boolean, toggleState: boolean, forced?: boolean) => {
+    getSelectionStartCellOrCaption(editor).each((cellOrCaption) => {
+      TableLookup.table(cellOrCaption, isRoot).filter(Fun.not(isRoot)).each((table) => {
+        let captionElm = editor.dom.select('caption', table.dom)[0];
```
lnewson, line 126: Pleaser use the appropriate Sugar APIs here.

modules/tinymce/src/plugins/table/main/ts/api/Commands.ts
```diff
@@ -49,6 +51,93 @@ const registerCommands = (editor: Editor, actions: TableActions, cellSelection:
...
+
+        if (captionElm && (!forced || toggleState)) {
```
lnewson, line 128: No falsy checks please! Use explicit checks instead.

modules/tinymce/src/plugins/table/main/ts/selection/SelectionTargets.ts
```diff
@@ -40,6 +42,98 @@ interface ExtractedSelectionDetails {
   readonly locked: Record<LockedDisableStrs, boolean>;
 }
 
+const isElementHeader = (cell: SugarElement<HTMLTableCellElement>) => {
+  if (cell.dom.nodeName.toLowerCase() === 'th') {
```
lnewson, line 46: Again, please use the appropriate Sugar APIs when dealing with Sugar elements.

modules/tinymce/src/plugins/table/main/ts/ui/ButtonToggleUtils.ts
```diff
@@ -0,0 +1,57 @@
+import { Singleton } from '@ephox/katamari';
```
lnewson, line 1: This import is in the wrong location.

modules/tinymce/src/plugins/table/main/ts/ui/Buttons.ts
```diff
@@ -165,6 +169,303 @@ const addButtons = (editor: Editor, selectionTargets: SelectionTargets, clipboar
...
+  } else {
+    editor.ui.registry.addMenuButton('tableclass', {
+      icon: 'table-classes',
+      tooltip: 'Table styles',
+      fetch: (callback) => {
+        callback(Arr.map(tableClassList, (value) => {
+          const item: Menu.ToggleMenuItemSpec = {
+            text: value.title,
+            type: 'togglemenuitem',
+            onAction: (_api: Menu.MenuItemInstanceApi) => {
```
lnewson, line 280: This type should not be needed here, as it'll be inferred from declaring the entire object a `Menu.ToggleMenuItemSpec`.

modules/tinymce/src/plugins/table/main/ts/ui/Buttons.ts
```diff
@@ -165,6 +169,303 @@ const addButtons = (editor: Editor, selectionTargets: SelectionTargets, clipboar
...
+        },
+      ]);
+    }
+  });
+
+  editor.ui.registry.addMenuButton('colclipboardactions', {
+    icon: 'cut-column',
+    tooltip: 'Column clipboard actions',
+    fetch: (callback) => {
+      callback([
```
lnewson, line 223: Another place that should be using existing items instead of redefining them.

modules/tinymce/src/plugins/table/main/ts/api/Commands.ts
```diff
@@ -49,6 +51,93 @@ const registerCommands = (editor: Editor, actions: TableActions, cellSelection:
...
+        actions.setTableRowType(editor, {
+          type: getNewType(),
+        });
+
+        Events.fireTableModified(editor, table.dom, Events.structureModified);
+      });
+    });
+  };
+
+  const toggleCaption = (_ui: boolean, toggleState: boolean, forced?: boolean) => {
```
lnewson, line 123: Adding a `state`/`forced` property goes against all current toggle based API design in TinyMCE. Can we reconsider how this is being done? Maybe it should be a separate command entirely to clarify that, eg `mceTableAddCaption`?

modules/tinymce/src/plugins/table/main/ts/api/Commands.ts
```diff
@@ -77,6 +166,47 @@ const registerCommands = (editor: Editor, actions: TableActions, cellSelection:
...
+        const allCells = Warehouse.justCells(warehouse);
+
+        const filtered = Arr.filter(allCells, (cellA) =>
+          Arr.exists(cells, (cellB) =>
+            Compare.eq(cellA.element, cellB)
+          )
+        );
+
+        Arr.each(filtered, (value) => {
+          if (Class.has(value.element, requestedClass)) {
```
lnewson, line 198: Please use the `Class.toggle` API here instead.

modules/tinymce/src/plugins/table/main/ts/api/Commands.ts
```diff
@@ -77,6 +166,47 @@ const registerCommands = (editor: Editor, actions: TableActions, cellSelection:
...
+        }
+
+        Events.fireTableModified(editor, table.dom, Events.structureModified);
+      });
+    });
+  };
+
+  const toggleTableCellClass = (_ui: boolean, requestedClass: string) => {
+    getSelectionStartCell(editor).each((startCell) => {
+      TableLookup.table(startCell, isRoot).filter(Fun.not(isRoot)).each((table) => {
```
lnewson, line 185: Why do you have `.filter(Fun.not(isRoot))` here? We're not operating on the table, only the cells within so I don't believe this is needed/makes sense.

modules/tinymce/src/plugins/table/main/ts/api/Commands.ts
```diff
@@ -77,6 +166,47 @@ const registerCommands = (editor: Editor, actions: TableActions, cellSelection:
...
+        const cells = TableSelection.getCellsFromSelection(startCell, selections);
+
+        const warehouse = Warehouse.fromTable(table);
```
lnewson, line 188: `Warehouse` is not meant to be used outside snooker for things that modify the table, it's only exposed to query things. Anything that's modifying it really should be moved to snooker, so could you please look at doing that for this?

modules/tinymce/src/plugins/table/main/ts/api/Commands.ts
```diff
@@ -77,6 +166,47 @@ const registerCommands = (editor: Editor, actions: TableActions, cellSelection:
...
+
+        Arr.each(filtered, (value) => {
+          if (Class.has(value.element, requestedClass)) {
+            Class.remove(value.element, requestedClass);
+          } else {
+            Class.add(value.element, requestedClass);
+          }
+        });
+
+        Events.fireTableModified(editor, table.dom, Events.structureModified);
```
lnewson, line 205: This is definitely the wrong thing to be firing here, as this isn't changing the structure and instead is only changing the styling.

modules/tinymce/src/plugins/table/main/ts/api/Commands.ts
lnewson: Actually looking more, this doesn't even need to use the `Warehouse`. It could just use `TableLookup.cells` instead, so please use that instead here.

modules/tinymce/src/plugins/table/main/ts/actions/TableActions.ts
```diff
@@ -141,6 +143,9 @@ export const TableActions = (editor: Editor, lazyWire: () => ResizeWire, selecti
   const makeColumnsHeader = execute(TableOperations.makeColumnsHeader, Fun.always, Fun.noop, lazyWire, Events.structureModified);
   const unmakeColumnsHeader = execute(TableOperations.unmakeColumnsHeader, Fun.always, Fun.noop, lazyWire, Events.structureModified);
 
+  const makeRowsHeader = execute(TableOperations.makeRowsHeader, Fun.always, Fun.noop, lazyWire, Events.structureModified);
```
lnewson, line 146: I'm sorry, but there's a reason we aren't using the `RowsHeader` operations from snooker for this and instead have the custom `setTableRowType` API. It's because snooker doesn't work with the multiple ways you can define a table row header type and it needs to take into account the `table_header_type` setting.

There's a JIRA in the backlog IIRC to make the snooker API work for that, but @metricjs ran out of time for the 5.5 release and we've yet to get back to it. As such this API cannot be used at this time and you'll need to use the other API instead.

modules/tinymce/src/plugins/table/main/ts/api/Commands.ts
```diff
@@ -49,6 +51,93 @@ const registerCommands = (editor: Editor, actions: TableActions, cellSelection:
     });
   });
 
+  const toggleColumnHeader = () => {
+    getSelectionStartCell(editor).each((startCell) => {
+      TableLookup.table(startCell, isRoot).filter(Fun.not(isRoot)).each((table) => {
```
lnewson, line 56: Another place I don't believe we need the `.filter(Fun.not(isRoot))` check.

modules/tinymce/src/plugins/table/main/ts/api/Commands.ts
```diff
@@ -49,6 +51,93 @@ const registerCommands = (editor: Editor, actions: TableActions, cellSelection:
...
+      });
+    });
+  };
+
+  const editColumnsHeader = (table: SugarElement<HTMLTableElement>, startCell: SugarElement<HTMLTableCellElement>, changeTo: 'th' | 'td'): void => {
+    const warehouse = Warehouse.fromTable(table);
+    const targets = TableTargets.forMenu(selections, table, startCell);
+
+    const usedColumns: number[] = [];
+    Arr.each(warehouse.all, (row) => {
```
lnewson, line 73: As mentioned elsewhere, using the warehouse to modify the table shouldn't be done here. Please create an appropriate snooker operation instead and use it via `TableActions`.

lnewson: Actually, correct me if I'm wrong but why do we even need all this logic? I believe this just tries to reimplement the `makeColumnsHeader` and `unmakeColumnsHeader` APIs.

lnewson (dismissed):

lnewson:
I've not yet finished my review, as I've only glanced over most of it and need to review in depth. However, there's a number of things that are going to need to change sorry @HAFRMO so I'll hold off reviewing anymore for now until more has been fixed.

modules/tinymce/src/plugins/table/main/ts/ui/Buttons.ts
```diff
@@ -165,6 +169,303 @@ const addButtons = (editor: Editor, selectionTargets: SelectionTargets, clipboar
...
+          };
+
+          return item;
+        }));
+      },
+      onSetup: selectionTargets.onSetupCellOrRow
+    });
+  }
+
+  const alignTableButtons = [
```
lnewson, line 317: These are meant to be the same as what's shown in the dialog IIUC. As such, please don't duplicate them and instead make them use a shared definition, otherwise if these need to be changed at any point it's highly likely one or the other will be missed.

modules/tinymce/src/plugins/table/main/ts/ui/CustomColorSwatch.ts
```diff
@@ -0,0 +1,69 @@
...
+ */
+
+import Editor from 'tinymce/core/api/Editor';
+import { Menu } from 'tinymce/core/api/ui/Ui';
+
+interface ColorSwatchDialogData {
+  colorpicker: string;
+}
+
+const mapColors = (colorMap: string[]): Menu.ChoiceMenuItemSpec[] => {
```
lnewson, line 15: This is duplicated with the Silver logic... we want to avoid duplication if possible, so can this be moved and made a little more generic? The `acid` library would be a good place for that I believe.

modules/tinymce/src/themes/silver/main/ts/ui/menus/item/build/ColorSwatchItem.ts
```diff
@@ -53,3 +53,15 @@ export const renderColorSwatchItem = (spec: Menu.FancyMenuItem, backstage: UiFac
     ]
   };
 };
+
+const getcolorItems = (spec: Menu.FancyMenuItem, backstage: UiFactoryBackstage): Menu.ChoiceMenuItemSpec[] => {
```
lnewson, line 57: Typo in the function name here. It should be `getColorItems`

modules/tinymce/src/themes/silver/main/ts/ui/menus/item/build/ColorSwatchItem.ts
```diff
@@ -53,3 +53,15 @@ export const renderColorSwatchItem = (spec: Menu.FancyMenuItem, backstage: UiFac
...
+  if (Type.isNonNullable(spec.initData)) {
+    if (spec.initData.colorselection) {
```
lnewson, line 59: No falsy checks please! Use something explicit like`Type.isObject` instead to make sure it's what we want.

Additionally, there's a typo here as this should be `colourSelection`. In saying that though, `selection` is confusing and not needed I believe. Instead, might I recommend just calling this `colors`?

modules/tinymce/src/themes/silver/main/ts/ui/menus/item/build/ColorSwatchItem.ts
```diff
@@ -53,3 +53,15 @@ export const renderColorSwatchItem = (spec: Menu.FancyMenuItem, backstage: UiFac
...
+      return spec.initData.colorselection.concat(ColorSwatch.getAdditionalColors(false));
+    } else {
+      return ColorSwatch.getColors(backstage.colorinput.getColors(), false);
```
lnewson, line 62: Why is this different? This should either be using the provided colors configuration, or always falling back to the same default. As such, I'd suggest rewriting this function to something like this:
```
const getColorItems = (spec: Menu.FancyMenuItem, backstage: UiFactoryBackstage): Menu.ChoiceMenuItemSpec[] =>
  Optional.from(spec.initData.colors)
    .filter(Type.isArray)
    .map((colors) => colors.concat(ColorSwatch.getAdditionalColors(false)))
    .getOrThunk(() => ColorSwatch.getColors(backstage.colorinput.getColors(), backstage.colorinput.hasCustomColors()));
```

modules/tinymce/src/themes/silver/main/ts/ui/menus/item/build/ColorSwatchItem.ts
```diff
@@ -53,3 +53,15 @@ export const renderColorSwatchItem = (spec: Menu.FancyMenuItem, backstage: UiFac
     ]
   };
 };
+
+const getcolorItems = (spec: Menu.FancyMenuItem, backstage: UiFactoryBackstage): Menu.ChoiceMenuItemSpec[] => {
+  if (Type.isNonNullable(spec.initData)) {
+    if (spec.initData.colorselection) {
+      return spec.initData.colorselection.concat(ColorSwatch.getAdditionalColors(false));
```
lnewson, line 60: Why are we not using the custom colors here? That definitely seems wrong, as it's supposed to be used for all color swatches if enabled. If you need it to be disabled for specific cases, that should be another `initData` parameter that defaults to adding the custom colors if not set to false.

lnewson:
For future reference, this would have been **a lot** better if it had been multiple PR's, as it'd make it a lot easier to review the individual pieces.

As a side note, I see you changed the title as @JamesToohey asked, but it's not very descriptive still so please use a descriptive title that explains what's been changed as it'll be used in the final squash commit.

- [ ] Add descriptive PR title please

modules/tinymce/src/plugins/table/main/ts/ui/CustomColorSwatch.ts
HAFRMO: Added to the external library.

modules/tinymce/src/plugins/table/main/ts/api/Commands.ts
lnewson: The event is correct, it's the data that's wrong. It should be `Events.styleModified` instead.

PR #6772, opened 2021-04-29
TINY-7337: Add toolbar_sticky_offset property to update Sticky toolbar position

camiloHimura:
Related Ticket: TINY-7337

Description of Changes:
* Added new `toolbar_sticky_offset ` setting to control sticky header offset

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


lnewson:
Moving the milestone to 5.9, as 5.8 feature code freeze is today and there's no way this can make that in time given it's only been logged today sorry.

lnewson (changes requested):
This was a nice attempt, but I'm afraid it's the wrong approach sorry @camiloHimura.

Here's the issues I've noted:
- This will only work for iframe mode, whereas it needs to work with both inline and iframe.
- This will more than likely break docking behaviour in unexpected ways, as it's changing the position set by docking. If nothing else it's not an ideal design pattern to be doing the positioning logic in multiple places.
- The setting name is inconsistent and should be `toolbar_sticky_offset` to match the `toolbar_sticky` setting. This was incorrect in the JIRA description sorry.
- The setting is being dynamically altered using the new API/Command you created. We generally don't like to do that (or anything that alters how things are rendered after init). I know it mentioned that as a bonus in the description, however that wasn't approved from what I can tell and when talking to the editor team we didn't like the idea of that. As such, if we want to go with that it'll need more discussion and it likely better done in another JIRA/PR. You could do the enable/disable API, but I'm not sure that'll have an actual use case, so it's probably better to leave that out for now as well.

So with that in mind, how should this be done? What we want to do is adjust the bounds we're passing to the docking behaviour so that it knows to use a viewport with the offset taken into account. In this case that means providing a custom `lazyViewport` configuration to docking such as:
```
lazyViewport: (comp) => {
  const win = Boxes.win();
  const offset = 50;
  const top = win.y + (isDockedMode(comp, 'top') ? offset : 0);
  const height = win.height - (isDockedMode(comp, 'bottom') ? offset : 0);
  return Boxes.bounds(win.x, top, win.width, height);
},
```
Note: In the above I've just hard coded it to 50px, but that'd need to be looked up from the setting.

lnewson:
Another minor thing sorry, can you please improve the PR title as remember it's used as the final commit message?

- [ ] Improve PR title to be more descriptive

modules/tinymce/src/themes/silver/main/ts/ui/header/StickyHeader.ts
```diff
@@ -65,7 +66,6 @@ const updateIframeContentFlow = (header: AlloyComponent): void => {
   const elm = header.element;
   Traverse.parent(elm).each((parentElem: SugarElement<HTMLElement>) => {
     const padding = 'padding-' + Docking.getModes(header)[0];
-
```
lnewson, line 68: Should this line have been removed?

modules/tinymce/CHANGELOG.md
```diff
@@ -249,6 +249,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## 5.4.0 - 2020-06-30
 
 ### Added
+- Added new `toolbar_sticky_offset` setting to control sticky header offset #TINY-7337
```
lnewson, line 252:
```suggestion
- Added a new `toolbar_sticky_offset` setting to allow the toolbar to be offset from the top of the page #TINY-7337
```
Maybe? Someone else might have a better suggestion yet though.

modules/tinymce/src/themes/silver/test/ts/module/StickyHeaderUtils.ts
```diff
@@ -184,6 +185,18 @@ const assertEditorClasses = (docked: boolean) => {
   })), container);
 };
 
+const assertHeaderPosition = async (top: boolean, location: ToolbarLocation, value: number) => {
```
lnewson, line 188: Can we please change `top` -> `isToolbarTop` to keep it consistent with how it's used elsewhere here?

Alternatively, we could just pass `location` and then determine if we're positioning at the top using that.

lnewson (approved):

modules/tinymce/src/themes/silver/main/ts/ui/header/StickyHeader.ts
camiloHimura: put it back

modules/tinymce/CHANGELOG.md

modules/tinymce/src/themes/silver/test/ts/module/StickyHeaderUtils.ts
camiloHimura: updated

JamesToohey (approved):

HAFRMO (approved):

modules/tinymce/CHANGELOG.md
```diff
@@ -253,6 +253,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## 5.4.0 - 2020-06-30
 
 ### Added
+- Added a new `toolbar_sticky_offset` setting to allow the toolbar to be offset from the top of the page #TINY-7337
```
lnewson, line 256: Don't forget to merge develop and move this to the `Unreleased` section.

- [ ] Move to correct changelog location

modules/tinymce/src/themes/silver/test/ts/module/StickyHeaderUtils.ts
```diff
@@ -184,6 +185,19 @@ const assertEditorClasses = (docked: boolean) => {
   })), container);
 };
 
+const assertHeaderPosition = async (toolbarLocation: ToolbarLocation, value: number) => {
```
lnewson, line 188: I missed this previously sorry, but as this is async it should use the `p` prefix.

- [ ] rename this to `pAssertHeaderPosition` to keep consistent

lnewson (approved):

modules/tinymce/CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+- Added a new `toolbar_sticky_offset` setting to allow the toolbar to be offset from the top of the page #TINY-7337
```
lnewson, line 8: @camiloHimura this doesn't follow the keep a changelog format sorry, it needs to be something like this instead:
```suggestion

### Added
- Added a new `toolbar_sticky_offset` setting to allow the toolbar to be offset from the top of the page #TINY-7337
```

modules/tinymce/CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## 5.8.0 - 2021-05-06
 
 ### Added
+- Added a new `toolbar_sticky_offset` setting to allow the toolbar to be offset from the top of the page #TINY-7337
```
lnewson, line 18: Sorry we've let this stall @camiloHimura 😞 Anyways, this changelog was added to the incorrect location originally so could you please merge in develop and move this to the 5.9/Unreleased section?

- [ ] Move changelog entry

camiloHimura: Updated

modules/tinymce/CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### 5.9.0
+
```
lnewson, line 10: This should not be here sorry @camiloHimura, it should only be under `Unreleased`.
```suggestion
```

modules/tinymce/CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+### Added
```
lnewson, line 8:
```suggestion
## Unreleased

### Added
```

PR #6773, opened 2021-04-29
fix/TINY-4612-inline-boundaries

camiloHimura:
Related Ticket: [TINY-4612](https://ephocks.atlassian.net/browse/TINY-4612)

Description of Changes:
* "moveToLineEndPoint" updated to check the selected node if the last element in the line is not inline

Pre-checks:
~*  [ ] Changelog entry added~
*  [ ] Tests have been added (if applicable)
*  [x] Branch prefixed with `feature/` for new features (if applicable)
*  [x] License headers added on new files (if applicable)

Review:
* [x] Milestone set


lnewson (changes requested):
I've investigated this more and this isn't the right approach sorry. The bug is the `NavigationUtils.getLineEndPoint` function, as it's returning the start of the **next** paragraph when it should be returning the end of the current paragraph. Additionally, using `selection.getNode()` here isn't going to work with nested content.

So given, both will result in the wrong placement in certain cases, we need to fix the root cause instead. I'm trying to investigate more now to see if I can find the root cause, however this might be rather risky and something we should push out until the next release.

I should add no matter what the fix is, we'll also need tests.

PR #6775, opened 2021-04-30
DOC-795: Workaround for EditorManager API docs marked as private

tylerkelly13:
Related Ticket: DOC-795

Description of Changes:
* Workaround for EditorManager API docs marked as private
* Remove tinymce.data namespace from api docs build (APIs removed in 4.7)

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/tinymce/src/core/main/ts/api/EditorManager.ts
```diff
@@ -24,12 +24,15 @@ import URI from './util/URI';
 declare const window: Window & { tinymce: any; tinyMCEPreInit: any };
 
 /**
+ * NOTE: the class tag is commented out for the include in
+ * `modules/tinymce/tools/docs/tinymce.js`
+ *
```
lnewson, line 29: Can we move this note outside the `EditorManager` class comment block? eg:
```suggestion
// NOTE: the class tag is commented out for the include in `modules/tinymce/tools/docs/tinymce.js`
/**
```

modules/tinymce/tools/docs/tinymce.js
```diff
@@ -14,6 +14,12 @@
  * @property {tinymce.AddOnManager} ThemeManager Global ThemeManager instance. Instance of AddOnManager.
  */
 
+/**
+ * Include EditorManager API docs.
+ *
+ * @include ../../src/core/main/ts/api/EditorManager.ts
```
lnewson, line 20: We should also comment out/remove the `borrows-member` part above as well, as we don't want that to be used if we're including it.

lnewson (approved):
This turned out better than I expected... I was thinking we'd need a separate file and then include in both locations. However, this keeps the code comments and all lets the API docs work :+1:

modules/tinymce/tools/docs/tinymce.js
tylerkelly13: Deleted in [`b995397` (#6775)](https://github.com/tinymce/tinymce/pull/6775/commits/b99539748bc82c00e104582adc37a9b0db558e01)

modules/tinymce/src/core/main/ts/api/EditorManager.ts
tylerkelly13: Fixed in [`b995397` (#6775)](https://github.com/tinymce/tinymce/pull/6775/commits/b99539748bc82c00e104582adc37a9b0db558e01)

ashfordneil (approved):

PR #6795, opened 2021-05-06
Slate-sugar package for traversing the rtc model

vpyshnenko:
Related Ticket:
TINY-7329
TINY-7381

Description of Changes:
* Placeholder text

Pre-checks:
* [ ] Changelog entry added

Review:
* [ ] Milestone set


modules/slate-sugar/src/main/ts/ephox/slate-sugar/api/PredicateFilter.ts
```diff
@@ -0,0 +1,9 @@
+import { Maybes, Maybe } from '@ephox/katamari';
+import { ModelApi, SlateLoc } from './Api';
+
+const descendants = (api: ModelApi, scope: SlateLoc, predicate: (loc: SlateLoc) => boolean): Maybe<SlateLoc[]> =>
```
TheSpyder, line 4: If you look at Sugar `PredicateFilter` vs `PredicateFind`, `*Find` modules are intended to return a maybe single element; `*Filter` methods must always return an array (even if it's empty). This is true in the RTC methods you're binding to, as well, so the `Maybe<>` type here is unnecessary.

TheSpyder (changes requested):
Nice work, overall!

modules/slate-sugar/src/main/ts/ephox/slate-sugar/api/PredicateFilter.ts
vpyshnenko: I've removed the PredicateFilter.descendants result to be wrapped in Maybe

TheSpyder (approved):

modules/slate-sugar/src/main/ts/ephox/slate-sugar/api/Api.ts
```diff
@@ -0,0 +1,79 @@
...
+  node: SlateNode;
+  parents: List<SlateLoc>;
+}
+
+interface SlateEditor {}
+
+interface ModelApi {
+  body: {
+    getBody: () => SlateLoc;
+    getEditor: () => SlateEditor;
```
TheSpyder, line 45: This was removed

modules/slate-sugar/src/main/ts/ephox/slate-sugar/api/Api.ts
```diff
@@ -0,0 +1,79 @@
...
+interface SlateText {
+  text: string;
+}
+
+type SlateNode = SlateElement | SlateText;
+
+interface List<T>{
+  hd: T;
+  tl: List<T> | 0;
+}
```
TheSpyder, line 33: We can't have this in the API, it's not a shared type, but we already had some plans to make `SlateLoc` use arrays so maybe that can happen separately

spocke: I wonder what impact it will have on performance switching to arrays but I guess the depth of documents will never be that high that memory reallocation becomes a major bottle neck it should be faster than traversing from the root for every iteration like slate does for its positions. I would really want to hide these implementation details maybe we can prefix them as _ in the reason side then not add any types for it. But opaque types is hard to have in a structural type system so not sure how well this will work. The only thing that is nominal in ts that I know of is enums but tagging everything will be annoying.

modules/slate-sugar/src/main/ts/ephox/slate-sugar/api/Api.ts
```diff
@@ -0,0 +1,79 @@
...
+
+interface SlateLoc {
```
spocke, line 35: Do we have any ideas on what to call this I think all references to slate should be removed on the external api. Maybe we should just call it Location but maybe that clashes with dom interfaces or ModelLocation or ModelLoc or just Loc naming things are hard?

lnewson: Can confirm `Location` is in the native DOM types. `ModelLocation` probably makes the most sense to me, as it's still readable/clear what it's applicable to.

modules/slate-sugar/src/main/ts/ephox/slate-sugar/api/Api.ts
vpyshnenko: fixed

modules/slate-sugar/src/main/ts/ephox/slate-sugar/api/Api.ts
vpyshnenko: renamed Slate references to Model

modules/slate-sugar/src/main/ts/ephox/slate-sugar/api/Api.ts
vpyshnenko: removed List in a favor of Array (not currently used in Demo)

modules/slate-sugar/src/main/ts/ephox/slate-sugar/api/Api.ts
```diff
@@ -0,0 +1,69 @@
...
+  nodeTransforms: {
+    setPropsAtPath: (path: path, props: Record<string, string>) => void;
+  };
+  predicateFind: {
+    closest: (loc: ModelLocation, predicate: (loc: ModelLocation) => boolean, isRoot: (loc: ModelLocation) => boolean) => ModelLocation;
+  };
+  predicateFilter: {
+    descendants: (loc: ModelLocation, predicate: (loc: ModelLocation) => boolean) => ModelLocation[];
+  };
+  slateLoc: {
```
spocke, line 54: I think this needs to be renamed to modelLocation as well. Or be exposed in a new path module as modelLocationToPath we will probably expose more path utility functions from the slate apis since they are also handy.

vpyshnenko: done

spocke: I think I was a bit unclear it could be in a Path module as a function called moduleLocationToPath since we are likely to add more path utility functions later like Path.equals etc.

vpyshnenko: done

TheSpyder:
The overall idea is good, so I'm not going to delete the branch, but we're going to have a long think about how to expose the model to integrators before we look at this again.

PR #6799, opened 2021-05-07
TINY-6264: Ensure list-style-type: none is not removed from nested lists when clearing formatting

JamesToohey:
Related Ticket: TINY-6264

Description of Changes:
Ensure that nested list items do not have their `list-style-type: none;` style removed when clearing formatting. This resulted in a bug where the markers for each li would be revealed.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/fmt/RemoveFormat.ts
+++ b/modules/tinymce/src/core/main/ts/fmt/RemoveFormat.ts
@@ -306,6 +306,11 @@ const removeFormatInternal = (ed: Editor, format: RemoveFormatPartial, vars?: Fo
           }
         }
 
+        // Remove mce prefixed attributes (must clean before short circuit operations)
+        if (MCE_ATTR_RE.test(name)) {
+          elm.removeAttribute('data-mce-' + name);
+        }
+
         // keep style="list-style-type: none" on <li>s
         if (name === 'style' && NodeType.matchNodeNames([ 'li' ])(elm) && dom.getStyle(elm, 'list-style-type') === 'none') {
           elm.removeAttribute(name);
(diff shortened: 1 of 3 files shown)


HAFRMO (approved):

metricjs (approved):

modules/tinymce/src/core/test/ts/browser/fmt/RemoveFormatTest.ts
```diff
@@ -30,6 +30,25 @@ describe('browser.tinymce.core.fmt.RemoveFormatTest', () => {
     editor.formatter.unregister('format');
   };
 
+  context('DefaultFormats remove format behavior', () => {
+    it('TINY-6264: Will not remove style="list-style-type: none" from list item elements', () => {
+      const editor = hook.editor();
+      const nestedListHtml = '<ul><li style="list-style-type: none;"><ul><li>hello</li></ul></li></ul>';
```
lnewson, line 36: Maybe use an `ol` here instead to validate this works with ordered and unordered lists? as then this would do ordered and the below would do unordered.

modules/tinymce/src/core/main/ts/fmt/RemoveFormat.ts
```diff
@@ -306,6 +306,13 @@ const removeFormatInternal = (ed: Editor, format: RemoveFormatPartial, vars?: Fo
           }
         }
 
+        // keep style="list-style-type: none" on <li>s
+        if (name === 'style' && NodeType.matchNodeNames([ 'li' ])(elm) && dom.getStyle(elm, 'list-style-type') === 'none') {
```
lnewson, line 310: I realised this is going to cause the logic on line 322 to not run, which might leave a case where the `data-mce-style` attribute would remain. As such we probably need to remove both the `style` and `data-mce-style` attribute in this short-circuit case or alternatively move this short circuit case to below line 322

Note: I suspect the `dom.setStyle` call will clean this up anyways, but I'm not confident on relying on that and as such I think we should explicitly clean it up.

- [ ] Ensure the `data-mce-style` attribute is also cleaned up properly.

lnewson (approved):

PR #6789, opened 2021-05-05
TINY-6383: Fixed an exception thrown when switching to readonly mode and resizing the editor

lnewson:
Related Ticket: TINY-6383

Description of Changes:
* Fixed an exception thrown when switching to readonly mode and resizing the editor
* Cleaned up the bespoke logic, as it was overly complicated

Note: This is the low-risk version of this fix, however I'm going to create a more versatile fix for 5.9 that will follow this.

Pre-checks:
* [x] Changelog entry added
* [x] Tests have been added (if applicable)
* [x] Branch prefixed with `feature/` for new features (if applicable)

Review:
* [x] Milestone set
* [x] Review comments resolved

GitHub issues (if applicable):

PR #6805, opened 2021-05-09
TINY-7436: Fixed issues with event handlers still executing even after being unbound

lnewson:
Related Ticket: TINY-7436

Description of Changes:
* Fixed an issue with events bound using `editor.on` still running the callback if it was unbound in an earlier handler for the same event.
* Changed the Alloy `NoContextApi` (what's used when a component is detached) to log a console warning instead of throw an exception for APIs that trigger events.
* Reverted previous `isConnected` checks added because of this bug.

Note: This is the more foolproof solution that was a little too risky to put into 5.8.1, as mentioned in https://github.com/tinymce/tinymce/pull/6789

I should add this also makes the event dispatcher work in the same way as the DOM event system. Here are extracts of the spec thanks to @ashfordneil:
> If an `EventListener` is added to an `EventTarget` while it is processing an event, it will not be triggered by the current actions but may be triggered during a later stage of event flow, such as the bubbling phase.

> If an `EventListener` is removed from an `EventTarget` while it is processing an event, it will not be triggered by the current actions.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set
* [x] Don't merge until interim 5.8.1 changes have been QA'd


modules/alloy/src/main/ts/ephox/alloy/api/system/NoContextApi.ts
```diff
@@ -5,27 +5,33 @@ import { AlloyComponent } from '../component/ComponentApi';
 import { AlloySystemApi } from './SystemApi';
 
 const NoContextApi = (getComp?: () => AlloyComponent): AlloySystemApi => {
-  const fail = (event: string) => () => {
-    throw new Error('The component must be in a context to send: ' + event +
-        (getComp ? '\n' + AlloyLogger.element(getComp().element) + ' is not in context.' : '')
-    );
+  const getMessage = (event: string) => `The component must be in a context to execute: ${event}` +
```
lnewson, line 8: This was intentionally changed from `send` -> `execute` as send was a little confusing.

ephox-mogran (approved):

modules/tinymce/src/core/main/ts/api/util/EventDispatcher.ts
```diff
@@ -296,7 +305,10 @@ class EventDispatcher<T> {
             // Unbind specific ones
             let hi = handlers.length;
             while (hi--) {
-              if (handlers[hi].func === callback) {
+              const handler = handlers[hi];
```
ashfordneil, line 308: [nitpick / optional] Theoretically, now that we have this `removed` flag, we could make this look more "functional"-ish, if you want to. Something like:

```typescript
Arr.forEach(handlers, (handler) => {
  if (handler.func === callback) {
    // do removing in two parts (flag the things to remove, and then remove them all later) to make iterating over a changing list simpler
    // not just here, but also if we happen to be in the middle of iterating over these handlers to fire this event right now
    handler.removed = true;
  }
});
this.bindings[currentName] = Arr.filter(handlers, (handler) => !handler.removed);
```

modules/tinymce/src/core/main/ts/api/util/EventDispatcher.ts
```diff
@@ -183,6 +184,13 @@ class EventDispatcher<T> {
       for (let i = 0, l = handlers.length; i < l; i++) {
         const callback = handlers[i];
 
+        // The handler was removed by an earlier handler in this loop so skip it. This is needed as
+        // the `handlers` array is not a "live" list, so when the handler is removed while we are
```
ashfordneil, line 188: I'm not a fan of saying that this isn't a "live" list, let's find a better alternative.

Firstly, if someone _adds_ a new event handler while we're in the middle of this loop, it will be included in this loop as though it were a "live" list, because we just call `.push` on the same underlying list.

Secondly, I feel like we could be setting ourselves up for trouble by not stressing how important it is that this section of the code, and the matching section of the `off` function 120 lines below, are always kept in sync. I can see myself:

1. looking at a comment explaining how this list isn't "live"
2. looking at how we pulled a (super-duper-shallow) copy of the handlers list out of the big bindings object a few lines earlier
3. thinking "that (super-duper-shallow) copy of the handlers list stops it from being live", and then
4. messing something up if I tried to change this code (eg mutating on `this.bindings[name]` in another function and expecting it not to interfere with this loop)

The best suggestion I can think of right now is something like:

> This is needed as we remove callbacks from the handlers list by creating a new list and updating the reference in `this.bindings[name]`, which won't affect our `handlers` array.

But I'm not super attached to that wording, as long as the message gets through (that it's linked super closely to the removal process)

ashfordneil (changes requested):

ashfordneil (approved):

modules/tinymce/src/core/main/ts/api/util/EventDispatcher.ts
lnewson: Have changed this to be more functional, it's a bit different to what what was above though just to reduce the amount of looping being done.

modules/tinymce/src/core/main/ts/api/util/EventDispatcher.ts
lnewson: As discussed I've made the handlers a clone instead to make it clearer and done some other minor rework.

modules/tinymce/CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+- Unbound event handlers were incorrectly running when unbound in a previous event handler #TINY-7436
```
ashfordneil, line 10:
```suggestion
- Unbinding an event handler while that event is firing now takes effect immediately #TINY-7436
```

Definitely [nitpick], I just saw `event handler` and `unbound` too many times in that sentence and it made it harder for me to parse.

modules/tinymce/CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
```
ashfordneil, line 11: Should we add a `### Changed` and discuss how _adding_ an event handler while the event is firing no longer takes effect immediately?

ashfordneil (approved):

modules/tinymce/CHANGELOG.md
lnewson: I see your point, however that suggestion sounds more like a `Changed` when we want this to definitely be a `Fixed`. Paging @tylerkelly13 for his thoughts.

modules/tinymce/CHANGELOG.md
lnewson: That was never meant to be supported as far as I'm aware, so we don't want a `Changed`. Maybe a `Fixed` but I'm also not convinced it needs to be mentioned here.

modules/tinymce/CHANGELOG.md
tylerkelly13:
```suggestion
- Event handlers could not unbind while the event was firing #TINY-7436
```
Maybe? Don't know if I understand the problem being fixed here.

```suggestion
- Unbinding an event handler, while that event was firing, did not take effect immediately #TINY-7436
- Binding an event handler, while that event was firing, incorrect took effect immediately #TINY-7436
```
@ashfordneil how does that look?

ltrouton (approved):

spocke (approved):

modules/tinymce/src/core/main/ts/api/util/EventDispatcher.ts
```diff
@@ -178,27 +179,34 @@ class EventDispatcher<T> {
       this.settings.beforeFire(args);
     }
 
-    const handlers = this.bindings[name];
-    if (handlers) {
-      for (let i = 0, l = handlers.length; i < l; i++) {
-        const callback = handlers[i];
+    // Lookup the handlers and make a clone to ensure the array won't be mutated while we're iterating
+    // otherwise some handlers could be skipped due to elements being prepended or removed
+    const handlers = Obj.get(this.bindings, name).map(Arr.from).getOr([]);
```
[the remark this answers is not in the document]
lnewson: Yeah, that is a valid point. I had figured it was likely going to be okay as they generally are small arrays. To be safe though I guess we should clone when doing the `on` addition instead, which would prevent the performance issue from cloning.

- [ ] Move the clone to the `on` function and update comments.

metricjs (commented):
Feel like I was a bit too involved in discussions about how to solve this for my tick to really count, but LGTM with the change from Spocke's comment and the updated changelog entries.

modules/tinymce/src/core/main/ts/api/util/EventDispatcher.ts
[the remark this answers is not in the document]
lnewson: Okay, have reworked this and updated the comment to take into account @ashfordneil's original concern.

PR #6806, opened 2021-05-09
Added return types and fixed no implicit any errors for modules in the caret directory

lnewson:
Related Ticket: N/A

Description of Changes:
This is an attempt to try and get more types in the core so that we can one day get to enabling `noImplicitAny` or even `strict` types. This is by no means an exhaustive conversion but it attempts to fix a number of issues in the `caret` directory (non public APIs) for:
- the `noImplicitAny` tsconfig rule
- the `@typescript-eslint/explicit-module-boundary-types` eslint rule

**Reminder:** This is only about trying to add/improve/fix the existing types without changing functionality as much as possible, so please be mindful about the scope of this PR when reviewing (ie comments about code design/existing code issues are out of scope).

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [x] Milestone set


modules/tinymce/src/core/main/ts/caret/CaretPosition.ts
```diff
@@ -74,7 +73,7 @@ const getBrClientRect = (brNode: Element): ClientRect => {
 };
 
 // Safari will not return a rect for <p>a<br>|b</p> for some odd reason
-const getBoundingClientRectWebKitText = (rng: Range): ClientRect => {
+const getBoundingClientRectWebKitText = (rng: Range): GeomClientRect.ClientRect | null => {
```
ashfordneil, line 76: Should we Jira up renaming this (if it's not public) so it doesn't conflict with the DOM type?

modules/tinymce/src/core/main/ts/caret/FakeCaret.ts
```diff
@@ -118,10 +120,10 @@ export const FakeCaret = (editor: Editor, root: HTMLElement, isBlock: (node: Nod
 
     if (isBlock(element)) {
       caretContainerNode = CaretContainer.insertBlock(caretBlock, element, before);
-      clientRect = getAbsoluteClientRect(root, element, before);
+      const clientRect = getAbsoluteClientRect(root, element, before);
       DomQuery(caretContainerNode).css('top', clientRect.top);
 
-      const caret = DomQuery<HTMLElement>('<div class="mce-visual-caret" data-mce-bogus="all"></div>').css(clientRect).appendTo(root)[0];
+      const caret = DomQuery<HTMLElement>('<div class="mce-visual-caret" data-mce-bogus="all"></div>').css({ ...clientRect }).appendTo(root)[0];
```
ashfordneil, line 126: What functional changes does this clone lead to?

modules/tinymce/src/core/main/ts/caret/LineWalker.ts
```diff
@@ -84,17 +83,23 @@ const walkUntil = (direction: VDirection, isAboveFn: PosPredicate, isBeflowFn: P
...
 const upUntil = Fun.curry(walkUntil, VDirection.Up, ClientRect.isAbove, ClientRect.isBelow) as (root: Node, predicateFn: RectPredicate, caretPosition: CaretPosition) => LineNodeClientRect[];
 const downUntil = Fun.curry(walkUntil, VDirection.Down, ClientRect.isBelow, ClientRect.isAbove) as (root: Node, predicateFn: RectPredicate, caretPosition: CaretPosition) => LineNodeClientRect[];
 
 const positionsUntil = (direction: VDirection, root: Node, predicateFn: RectPredicate, node: Node): LinePosClientRect[] => {
   const caretWalker = CaretWalker(root);
-  let walkFn, isBelowFn, isAboveFn,
-    caretPosition;
-  const result = [];
-  let line = 0, clientRect;
+  let walkFn;
```
ashfordneil, line 97: You might have forgotten one here. Or is this deliberate due to complexities in the type of the thing we're assigning this to?

ashfordneil (approved):

modules/tinymce/src/core/main/ts/caret/TableCells.ts
```diff
@@ -73,8 +73,8 @@ const getTopValue = (rect: ClientRect) => rect.top;
 const isAbove = (corner: Corner, y: number) => corner.y < y;
 const isBelow = (corner: Corner, y: number) => corner.y > y;
 
-const getClosestCellAbove = Fun.curry(getClosestCell, getBottomValue, isAbove);
-const getClosestCellBelow = Fun.curry(getClosestCell, getTopValue, isBelow);
+const getClosestCellAbove = Fun.curry(getClosestCell, getBottomValue, isAbove) as (table: HTMLElement, x: number, y: number) => Optional<CellOrCaption>;
```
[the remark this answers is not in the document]
lnewson: Actually, if you look everywhere else this is already used in the caret code it uses the above style...

modules/tinymce/src/core/main/ts/caret/CaretPosition.ts
lnewson: It is public unfortunately...

modules/tinymce/src/core/main/ts/caret/TableCells.ts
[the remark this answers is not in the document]
lnewson: Having said that though, you are right we normally use the other style. Will consider fixing all of them here, otherwise I'll log a JIRA to go through the entire code base to standardise them.

modules/tinymce/src/core/main/ts/caret/LineWalker.ts
lnewson: I did miss this one, thanks!

modules/tinymce/src/core/main/ts/caret/CaretUtils.ts
```diff
@@ -170,19 +167,21 @@ const before = Fun.curry(beforeAfter, true) as (node: Node) => Range;
...
   const offset = range.startOffset;
 
   if (CaretContainer.isCaretContainerBlock(container)) {
     if (!isElement(container)) {
       container = container.parentNode;
     }
 
-    location = container.getAttribute('data-mce-caret');
+    // TODO: The `isCaretContainerBlock` function already asserts container is an element, so
+    //  the above isElement check likely isn't needed which would remove the need for the cast.
```
JamesToohey, line 183: Tiny thing but I don't think this extra space was deliberate?

```suggestion
    // TODO: The `isCaretContainerBlock` function already asserts container is an element, so
    // the above isElement check likely isn't needed which would remove the need for the cast.
```

JamesToohey (approved):

modules/tinymce/src/core/main/ts/caret/CaretUtils.ts
lnewson: Yes, it's deliberate otherwise IDEs, etc... assume it's a newline and not part of the TODO statement. We could move it over further so it's inline with the TODO if you prefer? eg:
```suggestion
    // TODO: The `isCaretContainerBlock` function already asserts container is an element, so
    //       the above isElement check likely isn't needed which would remove the need for the cast.
```

HAFRMO (approved):

PR #6827, opened 2021-05-14
Migrate some plugins to use the Maybe API

ashfordneil:
Related Ticket: TINY-7444

Description of Changes:
* Each plugin I migrated was a separate commit, so
  * It should be easier to review
  * We can revert any one plugin individually
* The plugins themselves that have been converted so far
  * advlist
  * imagetools
  * fullscreen
  * codesample
  * emoticons
* Images was considered, but it ended up being too messy at this point. It'll need to wait until more of the underlying libraries have been migrated.

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [ ] Milestone set

### DIFF
--- a/modules/tinymce/src/plugins/codesample/main/ts/core/CodeSample.ts
+++ b/modules/tinymce/src/plugins/codesample/main/ts/core/CodeSample.ts
@@ -5,7 +5,7 @@
  * For commercial licenses see https://www.tiny.cloud/
  */
 
-import { Optional } from '@ephox/katamari';
+import { Fun, Maybes } from '@ephox/katamari';
 import DOMUtils from 'tinymce/core/api/dom/DOMUtils';
 import Editor from 'tinymce/core/api/Editor';
 import * as Utils from '../util/Utils';
@@ -15,10 +15,10 @@ const getSelectedCodeSample = (editor: Editor) => {
   const node = editor.selection ? editor.selection.getNode() : null;
 
   if (Utils.isCodeSample(node)) {
-    return Optional.some(node);
+    return Maybes.just(node);
   }
 
-  return Optional.none<Element>();
+  return Maybes.nothing<Element>();
 };
 
 const insertCodeSample = (editor: Editor, language: string, code: string) => {
@@ -27,21 +27,24 @@ const insertCodeSample = (editor: Editor, language: string, code: string) => {
 
     code = DOMUtils.DOM.encode(code);
 
-    return node.fold(() => {
+    if (Maybes.isNothing(node)) {
       editor.insertContent('<pre id="__new" class="language-' + language + '">' + code + '</pre>');
       editor.selection.select(editor.$('#__new').removeAttr('id')[0]);
-    }, (n) => {
+    } else {
+      const n = node.value;
       editor.dom.setAttrib(n, 'class', 'language-' + language);
       n.innerHTML = code;
       Prism.get(editor).highlightElement(n);
       editor.selection.select(n);
-    });
+    }
   });
 };
 
 const getCurrentCode = (editor: Editor): string => {
-  const node = getSelectedCodeSample(editor);
-  return node.fold(() => '', (n) => n.textContent);
+  return Fun.pipe(
+    getSelectedCodeSample(editor),
+    Maybes.fold(Fun.constant(''), (n) => n.textContent)
+  );
 };
 
 export {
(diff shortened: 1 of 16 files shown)


modules/tinymce/src/plugins/codesample/main/ts/core/CodeSample.ts
```diff
@@ -27,21 +27,24 @@ const insertCodeSample = (editor: Editor, language: string, code: string) => {
 
     code = DOMUtils.DOM.encode(code);
 
-    return node.fold(() => {
+    if (Maybes.isNothing(node)) {
```
lnewson, line 30: Should this have used `Maybes.fold`? As that was why that was added I thought.

modules/tinymce/src/plugins/emoticons/main/ts/core/EmojiDatabase.ts
```diff
@@ -98,23 +98,27 @@ const initDatabase = (editor: Editor, databaseUrl: string, databaseId: string):
...
-    return categories.get().bind((cats) => Optional.from(cats[category])).getOr([]);
+    return Fun.pipe(
+      categories.get(),
+      Maybes.bindNullable((cats) => cats[category]),
+      Maybes.getOr([])
+    );
   };
 
-  const listAll = (): EmojiEntry[] => all.get().getOr([]);
+  const listAll = (): EmojiEntry[] => Maybes.getOr<EmojiEntry[]>([])(all.get());
```
lnewson, line 117: Should this be using `Fun.pipe`? The use of it in this file is rather inconsistent which I'm not a fan of.

modules/tinymce/src/plugins/codesample/main/ts/core/CodeSample.ts
ashfordneil: I'm not gonna try and convince you that I've been 100% consistent here, but I would prefer to think of fold as a ternary equivalent. Not in the sense that it makes code convoluted and should only be used in simple cases, but more in the sense that it's meant to return something. As I didn't need to return something here, I didn't use it.

(Also, unless you've already bit the bullet with `Fun.pipe`, calling curried functions is painful, both for parenthesis soup and type annotation reasons)

spocke: Could we have curried and uncurried versions. Sometimes you might just want to  call a single function going though a pipe just to avoid the uncurry seems a bit strange. Maybe have things like Maybes.mapU that is uncurried.

modules/tinymce/src/plugins/emoticons/main/ts/core/EmojiDatabase.ts
spocke: Could we have curried and uncurried versions. Sometimes you might just want to  call a single function going though a pipe just to avoid the uncurry seems a bit strange. Maybe have things like Maybes.mapU that is uncurried.

modules/tinymce/src/plugins/fullscreen/main/ts/core/Actions.ts
```diff
@@ -132,7 +132,10 @@ const toggleFullscreen = (editor: Editor, fullscreenState: Cell<ScrollInfo | nul
     handleClasses(DOM.removeClass);
 
     viewportUpdate.unbind();
-    Optional.from(fullscreenState.get()).each((info) => info.fullscreenChangeHandler.unbind());
+    const info = Maybes.from(fullscreenState.get());
+    if (Maybes.isJust(info)) {
```
metricjs, line 136: Hm. I can't remember from the discussions - did we implement an `each` equivalent for Maybes? Might be that I'm just used to the old style, but switching from `each` to an if here feels really odd to me.

modules/tinymce/src/plugins/imagetools/main/ts/core/Actions.ts
```diff
@@ -196,29 +202,32 @@ const updateSelectedImage = (editor: Editor, origBlob: Blob, ir: ImageResult, up
...
-    const flippedSize = imgOpt.fold(() => null, (img) => {
-      const size = ImageSize.getImageSize(img.dom);
-      return size ? { w: size.h, h: size.w } : null;
-    });
+    const flippedSize = Fun.pipe(
+      getSelectedImage(editor),
+      Maybes.bindNullable((img) => ImageSize.getImageSize(img.dom)),
+      Maybes.map((size) => ({ w: size.h, h: size.w })),
+      Maybes.getOrNull
+    );
```
metricjs, line 230: I like this :D

PR #6831, opened 2021-05-17
TINY-7249: changed loading order so content css gets loaded before loading contents

spocke:
Related Ticket:

Description of Changes:
* This changes the load order of things so that the content css gets loaded before the editor is filled with contents. This is to prevent a FOUC happening while loading the editor. It's extra visible with RTC but I think this happens on the non RTC version of tinymce as well. Not sure how to make a unit test for this though since just waiting for the events to fire is not enough to ensure that a flash doesn't occur.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -6,8 +6,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Changed
+- Changed the load order so that the content css gets loaded before the editor gets populated with contents. #TINY-7249
+
 ### Fixed
-- Flash of unstyled content while loading the content css was shown. It will now load the content css first then populate the editor #TINY-7249
+- Flash of unstyled content while loading the editor because the content css was loaded after the editor content was rendered #TINY-7249
 - Unbinding an event handler did not take effect immediately while the event was firing #TINY-7436
 - Binding an event handler incorrectly took effect immediately while the event was firing #TINY-7436
 
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Fixed
+- Flash of unstyled content while loading the content css was shown. It will now load the content css first then populate the editor #TINY-7249
 - Unbinding an event handler did not take effect immediately while the event was firing #TINY-7436
 - Binding an event handler incorrectly took effect immediately while the event was firing #TINY-7436
 
--- a/modules/tinymce/src/core/main/ts/Rtc.ts
+++ b/modules/tinymce/src/core/main/ts/Rtc.ts
@@ -8,7 +8,6 @@
 import { Cell, Fun, Obj, Optional, Type } from '@ephox/katamari';
 import Editor from './api/Editor';
 import Formatter from './api/Formatter';
-import Promise from './api/util/Promise';
 import { Content, ContentFormat, GetContentArgs, SetContentArgs } from './content/ContentTypes';
 import { getContentInternal } from './content/GetContentImpl';
 import { insertHtmlAtCaret } from './content/InsertContentImpl';
@@ -257,23 +256,24 @@ const getRtcSetup = (editor: Editor): Optional<() => Promise<RtcRuntimeApi>> =>
     Optional.from(rtcPlugin.setup)
   );
 
-export const setup = (editor: Editor): Optional<Promise<boolean>> => {
+export const setup = (editor: Editor): Optional<() => Promise<boolean>> => {
   const editorCast = editor as RtcEditor;
   return getRtcSetup(editor).fold(
     () => {
       editorCast.rtcInstance = makePlainAdaptor(editor);
       return Optional.none();
     },
-    (setup) => Optional.some(
-      setup().then((rtcEditor) => {
-        editorCast.rtcInstance = makeRtcAdaptor(rtcEditor);
-        return rtcEditor.rtc.isRemote;
-      }, (err) => {
-        // We need to provide a noop adaptor on init failure since otherwise calls to hasUndo etc will continue to throw errors
-        editorCast.rtcInstance = makeNoopAdaptor();
-        return Promise.reject<boolean>(err);
-      })
-    )
+    (setup) => {
+      // We need to provide a noop adaptor while initializing since any call by the theme or plugins to say undoManager.hasUndo would throw errors
+      editorCast.rtcInstance = makeNoopAdaptor();
+
+      return Optional.some(
+        () => setup().then((rtcEditor) => {
+          editorCast.rtcInstance = makeRtcAdaptor(rtcEditor);
+          return rtcEditor.rtc.isRemote;
+        })
+      );
+    }
   );
 };
 
--- a/modules/tinymce/src/core/main/ts/init/InitContentBody.ts
+++ b/modules/tinymce/src/core/main/ts/init/InitContentBody.ts
@@ -330,12 +330,16 @@ const loadContentCss = (editor: Editor) => {
 const preInit = (editor: Editor) => {
   const settings = editor.settings, doc = editor.getDoc(), body = editor.getBody();
 
+  Events.firePreInit(editor);
+
   if (!settings.browser_spellcheck && !settings.gecko_spellcheck) {
     doc.body.spellcheck = false; // Gecko
     DOM.setAttrib(body, 'spellcheck', 'false');
   }
 
   editor.quirks = Quirks(editor);
+
+  Events.firePostRender(editor);
 
   const directionality = Settings.getDirectionality(editor);
   if (directionality !== undefined) {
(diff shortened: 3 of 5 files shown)


modules/tinymce/src/core/main/ts/init/InitContentBody.ts
```diff
@@ -453,17 +460,23 @@ const initContentBody = (editor: Editor, skipWrite?: boolean) => {
   Placeholder.setup(editor);
 
   Events.firePreInit(editor);
+  preInit(editor);
 
   Rtc.setup(editor).fold(() => {
-    preInit(editor);
-  }, (loadingRtc) => {
+    Events.firePostRender(editor);
```
lnewson, line 466: Changing this event like this is pretty much a breaking change. I don't think we can do this as it's moved a reasonable amount later in the process now and we have no idea how many people were relying on it firing at the old location sorry.

modules/tinymce/src/core/test/ts/browser/EditorRemoveTest.ts
```diff
@@ -38,7 +38,7 @@ describe('browser.tinymce.core.EditorRemoveTest', () => {
     await McEditor.pFromHtml('<textarea></textarea>', {
       ...settings,
       setup: (editor: Editor) => {
-        editor.on('LoadContent', () => {
+        editor.on('PreInit', () => {
```
lnewson, line 41: Why was this changed?

lnewson (changes requested):

modules/tinymce/src/core/main/ts/init/InitContentBody.ts
lnewson: For context, a simple example of how this may break an integration is if an integration changed the `directionality` in the `PostRender`. Previously that would have taken effect, but it now no longer will since the directionality is read before `PostRender` fires.

modules/tinymce/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Fixed
+- Flash of unstyled content while loading the content css was shown. It will now load the content css first then populate the editor #TINY-7249
```
lnewson, line 10: If we're going ahead with this, then this really needs to be a `Changed` entry as well. We really need to call out/detail that we're changing the initialization sequence here as that could break some people integrations.

lnewson:
Thinking about this more, I think the better/safer/least non-breaking option would be to do something similar to what we do in the theme. Whereby we set `visibility: hidden` while the editor is loading and then remove that once the content CSS has loaded.

modules/tinymce/src/core/test/ts/browser/EditorRemoveTest.ts
spocke: Since it patches over the style sheet loader it needs to be done before the actual css starts loading.

modules/tinymce/src/core/main/ts/init/InitContentBody.ts
spocke: Good point let me see if I can move things around so that the rtc instance is setup before preInit and so the we can do all the steps in the exact same order. You can pull the "breaking change card" for any change it's all about how much impact it might have on users. The 5.2, 5.3 change from sync to async init of the content of the editor was far more "breaking" than this re-ordering and it doesn't seem to have broken any integrations. We really need to fix this rendering look at the attached video of the editor blinking it looks really unprofessional. The firePostRender needs to happen early on other wise you get an editor without a toolbar then after a while you get the toolbar and the content so it produces a new type of flash.

lnewson: > You can pull the "breaking change card" for any change it's all about how much impact it might have on users.

Yeah, it has a low risk of breaking someone's integration, however I don't believe we need to do it in this case. We can either change things to make sure the event fires in the same order as you mentioned or look at doing the `hidden` trick I mentioned. If neither of those work though, then yeah we'll have to sort it out and potentially end up changing where things fire.

> The 5.2, 5.3 change from sync to async init of the content of the editor was far more "breaking" than this re-ordering and it doesn't seem to have broken any integrations

I'm going to have to disagree here, as integrations had no hooks into the part of the init sequence that was changed and instead they could only listen for the events we fire. Additionally, the whole init sequence has always been async, since it needs to load stylesheets/javascript files.

spocke: > I'm going to have to disagree here, as integrations had no hooks into the part of the init sequence that was changed and instead they could only listen for the events we fire. Additionally, the whole init sequence has always been async, since it needs to load stylesheets/javascript files.

Right it was only the RTC part that got converted into an async operation I thought it was both of the branches.

I moved things around now so the postRender event should now fire where it used to fire. Not sure what makes more sense here using a thunk or having a separate function that setups the noop rtcInstance.

modules/tinymce/CHANGELOG.md
spocke: Added a change item as well since it's a behavior change for sure. Who knows what people might be doing out there.

spocke:
> Thinking about this more, I think the better/safer/least non-breaking option would be to do something similar to what we do in the theme. Whereby we set `visibility: hidden` while the editor is loading and then remove that once the content CSS has loaded.

That would not show a throbber while the content is loading. The RTC side could take a while to initialize so the UI needs to render before content and before content css.

lnewson:
> That would not show a throbber while the content is loading. The RTC side could take a while to initialize so the UI needs to render before content and before content css.

Wait why not? The throbber is UI based so it relies on the skin styles. It should not be impacted by anything we do on the editor body/content css.

modules/tinymce/src/core/main/ts/init/InitContentBody.ts
lnewson: Thanks Spocke! At a glance that should be fine now, but I'll review it properly in the morning.

spocke:
> Wait why not? The throbber is UI based so it relies on the skin styles. It should not be impacted by anything we do on the editor body/content css.

Doesn't the whole thing become a race condition if we stick hidden on the UI then the UI will not show until the content css is loaded but what happens if the UI css gets loaded before content CSS then that will flag the hidden state off by it's !important css rule. We would have to wait the the UI css and the content css to both resolve then we could lift the hidden state by using code then so we would have to tweak the oxide css. This all feels like a refactoring for v6 content css should really be loaded by the silver theme in my opinion also the visibility state should be handled by code not css and !important rules that way it could wait for all css to be loaded in parallel then render the UI. It was handled this way since there was no reliable way to have a load event on CSS on WebKit so the display part of the UI had to be in the css not in code this is no longer the case it been fixed years ago.

modules/tinymce/src/core/main/ts/init/InitContentBody.ts
spocke: This would still need the firePostRender before content css loading change since otherwise we still get that half rendered UI while the RTC content is loading I want the full editor UI in readonly mode with a throbber while the content is loading and that is what is happening now. We could change this to use a hidden state on the body but that would still require the other changes to the loading to avoid that for RTC and it would be a bit more involved that for example in inline mode we don't want to toggle the hidden state on/off since that would flash the editor contents while loading the css in and the page css might already have all the styles it need to render the content. So I think we are good with the things as is for now.

HAFRMO (approved):

TheSpyder (approved):

modules/tinymce/src/core/main/ts/init/InitContentBody.ts
TheSpyder: I think the thunk is fine.

Whether the UI displays with a toolbar or not is less important than the throbber - that must display immediately. If we can order the events (or safely change other parts of the code) such that the toolbar shows, that's cool, if not that's fine too.

lnewson (approved):

lnewson:
> Now I think I understand you are talking about setting a visibility: hidden on the body and flag that on/off while rendering. I thought you where talking about this: https://github.com/tinymce/tinymce/blob/develop/modules/oxide/src/less/theme/globals/global.less#L28 having the whole UI switched off then switched on when the content css is loaded that would obviously not render a UI.

Doesn't really matter now, as these changes should work fine, but yes this is what I was talking about.

modules/tinymce/src/core/main/ts/Rtc.ts
```diff
@@ -257,23 +256,24 @@ const getRtcSetup = (editor: Editor): Optional<() => Promise<RtcRuntimeApi>> =>
...
-        editorCast.rtcInstance = makeNoopAdaptor();
-        return Promise.reject<boolean>(err);
-      })
-    )
+    (setup) => {
+      // We need to provide a noop adaptor while initializing since any call by the theme or plugins to say undoManager.hasUndo would throw errors
+      editorCast.rtcInstance = makeNoopAdaptor();
+
+      return Optional.some(
+        () => setup().then((rtcEditor) => {
```
metricjs, line 271: Just checking: what happened to the `err` here? I assume we no longer need it for some reason, but I couldn't figure out the logic.

metricjs (approved):

modules/tinymce/src/core/main/ts/Rtc.ts
spocke: The errors gets handled later on. This specific error branch was added there to provide a noop rtc instance if an error was produced that is now replaced with the noop rtc instance that is always there while initializing but replaced with a real rtc instance once the editor has started.

PR #6841, opened 2021-05-18
TINY-7476: Added table cell class and table class.

HAFRMO:
Related Ticket:

Description of Changes:
Toggle classes on tables and cells

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/tinymce/CHANGELOG.md
```diff
@@ -12,6 +12,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### 5.8.1 - TBD
 
+### Added
```
lnewson, line 15: This should be in the `Unreleased` section which is what will be used for 5.9.

- [ ] Move to correct section

modules/tinymce/CHANGELOG.md
```diff
@@ -12,6 +12,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- Added the new command `mceTableCellToggleClass` which toggles currently selected cells to have or not have the provided class #TINY-7476
+- Added the new command `mceTableToggleClass` which toggles currently selected table to have or not have the provided class #TINY-7476
```
lnewson, line 17:
```suggestion
- Added a new `mceTableToggleClass` command which toggles the provided class on the currently selected table #TINY-7476
```

modules/tinymce/CHANGELOG.md
```diff
@@ -12,6 +12,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### 5.8.1 - TBD
 
+### Added
+- Added the new command `mceTableCellToggleClass` which toggles currently selected cells to have or not have the provided class #TINY-7476
```
lnewson, line 16:
```suggestion
- Added a new `mceTableCellToggleClass` command which toggles the provided class on the currently selected cells #TINY-7476
```

modules/tinymce/src/plugins/table/main/ts/api/Commands.ts
```diff
@@ -69,6 +69,30 @@ const registerCommands = (editor: Editor, actions: TableActions, cellSelection:
...
+        const cells = TableSelection.getCellsFromSelection(startCell, selections);
+
+        Arr.each(cells, (value) => {
+          Class.toggle(value, requestedClass);
+        });
+
+        Events.fireTableModified(editor, table.dom, Events.styleModified);
+      });
+    });
+  };
```
lnewson, line 94: There's a reasonable amount of duplication here so we can simplify this (we also don't need the `Fun.not(isRoot)` from what I can see). eg:
```suggestion
  const toggleClass = (clazz: string, lookup: (table: SugarElement<HTMLTableElement>, startCell: SugarElement<HTMLTableCellElement>) => SugarElement<HTMLElement>[]) => {
    getSelectionStartCell(editor).each((startCell) => {
      TableLookup.table(startCell, isRoot).each((table) => {
        const elements = lookup(table, startCell);

        Arr.each(elements, (value) => {
          Class.toggle(value, clazz);
        });

        Events.fireTableModified(editor, table.dom, Events.styleModified);
      });
    });
  };

  const toggleTableClass = (_ui: boolean, clazz: string) =>
    toggleClass(clazz, (table, _) => [ table ]);

  const toggleTableCellClass = (_ui: boolean, clazz: string) =>
    toggleClass(clazz, (_, startCell) => TableSelection.getCellsFromSelection(startCell, selections, isRoot));
```

Note: You'll also see I added the missing `isRoot` argument to `getCellsFromSelection`.

modules/tinymce/src/plugins/table/main/ts/ui/ButtonToggleUtils.ts
```diff
@@ -0,0 +1,35 @@
...
+const onSetupToggle = (editor: Editor, styleName: string, styleValue: string) => {
+  return (api: Toolbar.ToolbarToggleButtonInstanceApi) => {
+    const boundCallback = Singleton.unbindable();
+
+    const init = () => {
+      const value = {
+        value: styleValue
+      };
+
+      api.setActive(editor.formatter.match(styleName, value));
```
lnewson, line 21: This is overly verbose and can be inlined:
```suggestion
      api.setActive(editor.formatter.match(styleName, { value: styleValue }));
```

modules/tinymce/src/plugins/table/main/ts/ui/ButtonToggleUtils.ts
```diff
@@ -0,0 +1,35 @@
...
+ * For LGPL see License.txt in the project root for license information.
+ * For commercial licenses see https://www.tiny.cloud/
+ */
+
+import { Singleton } from '@ephox/katamari';
+import Editor from 'tinymce/core/api/Editor';
+import { Toolbar } from 'tinymce/core/api/ui/Ui';
+
+const onSetupToggle = (editor: Editor, styleName: string, styleValue: string) => {
+  return (api: Toolbar.ToolbarToggleButtonInstanceApi) => {
```
lnewson, line 13: This type is wrong for where this is being used. It's being used in a menu button, not a toggle button.
```suggestion
  return (api: Toolbar.ToolbarMenuButtonInstanceApi) => {
```

modules/tinymce/src/plugins/table/main/ts/ui/Buttons.ts
```diff
@@ -165,6 +168,56 @@ const addButtons = (editor: Editor, selectionTargets: SelectionTargets, clipboar
     icon: 'table'
   });
 
+  const tableClassList = getTableClassList(editor);
+
+  if (tableClassList.length === 0) {
+    // eslint-disable-next-line no-console
+    console.error('Missing table class list');
```
lnewson, line 175: I'm not sure what this is attempting to do, but this is definitely the wrong thing to be doing here.

modules/tinymce/src/plugins/table/main/ts/ui/Buttons.ts
```diff
@@ -165,6 +168,56 @@ const addButtons = (editor: Editor, selectionTargets: SelectionTargets, clipboar
...
+  } else {
+    editor.ui.registry.addMenuButton('tableclass', {
+      icon: 'table-classes',
+      tooltip: 'Table styles',
+      fetch: (callback) => {
+        callback(Arr.map(tableClassList, (value) => {
```
lnewson, line 181: This logic is all duplicated below with the exception of the class list, command name and format name. Please pull this out into a helper function to generate the `fetch` callback and just pass those three things in.

- [ ] Remove duplication

modules/tinymce/src/plugins/table/main/ts/ui/Buttons.ts
```diff
@@ -165,6 +168,56 @@ const addButtons = (editor: Editor, selectionTargets: SelectionTargets, clipboar
...
   });
 
+  const tableClassList = getTableClassList(editor);
+
+  if (tableClassList.length === 0) {
+    // eslint-disable-next-line no-console
+    console.error('Missing table class list');
+  } else {
+    editor.ui.registry.addMenuButton('tableclass', {
+      icon: 'table-classes',
```
lnewson, line 178: This relies on the icons being merged first, so we'll need that merged before we can merge this PR.

- [ ] Don't merge until https://github.com/tinymce/tinymce/pull/6613 has been merged

modules/tinymce/src/plugins/table/test/ts/browser/command/ModifyClassesCommandsTest.ts
```diff
@@ -0,0 +1,249 @@
...
+    assert.lengthOf(events, 0, label + ', events not zero');
+    editor.execCommand(cmdName, ui, data);
+    assert.lengthOf(events, 1, label + ', events not one');
+    assert.equal(events[0].type, 'tablemodified', label + ', event not tablemodified');
+    assert.isTrue(events[0].structure, label + ', structure is not true');
+    assert.isFalse(events[0].style, label + ', style is not false');
+    events = [];
+  };
+
+  context('TINY-7163: Modify cell class', () => {
```
lnewson, line 44: JIRA numbers should not be used on contexts, they should be in each `it` block as we may add tests later and they will have a different JIRA number.

modules/tinymce/src/plugins/table/test/ts/browser/command/ModifyClassesCommandsTest.ts
```diff
@@ -0,0 +1,249 @@
...
+    context('Can use mceTableCellToggleClass to toggle the cell class', () => {
+      context('When there is no class already on the cell', () => {
+        const getContentWithoutClass = () => {
```
lnewson, line 47: This doesn't need to be thunked. It can be used as a string.

modules/tinymce/src/plugins/table/test/ts/browser/command/ModifyClassesCommandsTest.ts
```diff
@@ -0,0 +1,249 @@
...
+import { TinyHooks, TinySelections } from '@ephox/mcagar';
+import { PlatformDetection } from '@ephox/sand';
+import { assert } from 'chai';
+import Editor from 'tinymce/core/api/Editor';
+import { EditorEvent } from 'tinymce/core/api/util/EventDispatcher';
+import { TableModifiedEvent } from 'tinymce/plugins/table/api/Events';
+import Plugin from 'tinymce/plugins/table/Plugin';
+import Theme from 'tinymce/themes/silver/Theme';
+
+const platform = PlatformDetection.detect();
```
lnewson, line 11: This should be inside the `describe` block, can you please move it there?

modules/tinymce/src/plugins/table/test/ts/browser/command/ModifyClassesCommandsTest.ts
```diff
@@ -0,0 +1,249 @@
...
+
+describe('browser.tinymce.plugins.table.command.ModifiyClassesCommandsTest', () => {
```
lnewson, line 13:
```suggestion
describe('browser.tinymce.plugins.table.command.ModifyClassesCommandsTest', () => {
```

modules/tinymce/src/plugins/table/test/ts/browser/command/ModifyClassesCommandsTest.ts
```diff
@@ -0,0 +1,249 @@
...
+
+        it('Can be toggled on', () => {
+          const editor = hook.editor();
+          editor.setContent(getContentWithoutClass());
+
+          TinySelections.setSelection(editor, [ 0, 0, 0 ], 0, [ 0, 0, 0 ], 1, true);
+
+          execCmdAndAssertEvent('TINY-7163', editor, 'mceTableCellToggleClass', false, 'a');
+
+          assert.equal(cleanTableHtml(editor.getContent()), getContentWithClass());
```
lnewson, line 81: The `cleanTableHtml` should not be needed here as it's not actually doing anything, in which case this should just be using the `TinyAssertions.assertContent` assertion instead. Please make sure to fix all cases, not just this single one.

modules/tinymce/src/plugins/table/test/ts/browser/command/ModifyClassesCommandsTest.ts
```diff
@@ -0,0 +1,249 @@
...
+                '<tr>' +
+                  (platform.browser.isIE() ? '<td>1</td>' : '<td class="">1</td>') +
+                  '<td>2</td>' +
+                '</tr>' +
+              '</tbody>' +
+            '</table>'
+          );
+        };
+
+        const getContentWithClass = () => {
```
lnewson, line 60: This doesn't need to be thunked. It can be used as a string.

modules/tinymce/src/plugins/table/test/ts/browser/command/ModifyClassesCommandsTest.ts
```diff
@@ -0,0 +1,249 @@
...
+              '</tbody>' +
+            '</table>'
+          );
+        };
+
+        it('Can be toggled on', () => {
+          const editor = hook.editor();
+          editor.setContent(getContentWithoutClass());
+
+          TinySelections.setSelection(editor, [ 0, 0, 0 ], 0, [ 0, 0, 0 ], 1, true);
```
lnewson, line 77: We shouldn't be passing `true` here as it's the default. It should only be used when we need to pass `false` which should be very rare.

modules/tinymce/src/plugins/table/test/ts/browser/command/ModifyClassesCommandsTest.ts
```diff
@@ -0,0 +1,249 @@
...
+
+          execCmdAndAssertEvent('TINY-7163', editor, 'mceTableCellToggleClass', false, 'a');
+
+          assert.equal(cleanTableHtml(editor.getContent()), getContentWithoutClass());
+        });
+      });
+
+      context('When there is already a class on the cell, which should not be deleted', () => {
+        const getContentWithoutClass = () => {
```
lnewson, line 97: More that don't need to be thunked. I'm not going to comment on anymore at this stage, as there's quite a few ahead.

modules/tinymce/src/plugins/table/test/ts/browser/command/ModifyClassesCommandsTest.ts
```diff
@@ -0,0 +1,249 @@
...
+  const cleanTableHtml = (html: string) =>
+    html.replace(/<p>(&nbsp;|<br[^>]+>)<\/p>$/, '');
+
+  let events: Array<EditorEvent<TableModifiedEvent>> = [];
+  const logEvent = (event: EditorEvent<TableModifiedEvent>) => {
+    events.push(event);
+  };
+
+  const execCmdAndAssertEvent = (label: string, editor: Editor, cmdName: string, ui?: boolean, data?: string) => {
+    assert.lengthOf(events, 0, label + ', events not zero');
```
lnewson, line 35:
```suggestion
    assert.lengthOf(events, 0, `${label}: before executing the command`);
```

modules/tinymce/src/plugins/table/test/ts/browser/command/ModifyClassesCommandsTest.ts
```diff
@@ -0,0 +1,249 @@
...
+    editor.execCommand(cmdName, ui, data);
+    assert.lengthOf(events, 1, label + ', events not one');
```
lnewson, line 37:
```suggestion
    assert.lengthOf(events, 1, `${label}: after executing the command`);
```

modules/tinymce/src/plugins/table/test/ts/browser/command/ModifyClassesCommandsTest.ts
lnewson: I also don't think we need the `label` here as it's doesn't hold any meaning. It's only currently being used for the JIRA number which should be in the `it` definitions instead.

modules/tinymce/src/plugins/table/test/ts/browser/command/ModifyClassesCommandsTest.ts
lnewson: Not going to do anymore, but you get the poin. ie if adding a message, don't say what the assertion is doing and also use template strings if possible.

modules/tinymce/src/plugins/table/test/ts/browser/command/ModifyClassesCommandsTest.ts
```diff
@@ -0,0 +1,249 @@
...
+          TinySelections.setSelection(editor, [ 0, 0, 0 ], 0, [ 0, 0, 0 ], 1, true);
+
+          execCmdAndAssertEvent('TINY-7163', editor, 'mceTableCellToggleClass', false, 'a');
+
+          assert.equal(cleanTableHtml(editor.getContent()), getContentWithoutClass());
+        });
+      });
+    });
+
+    context('Can use mceTableToggleClass to toggle the table class', () => {
```
lnewson, line 147: This is not part of `Modify cell class` this is now modifying table classes, so either remove the `Modify cell class` context or create a new context and move this inside please.

modules/tinymce/src/plugins/table/test/ts/browser/command/ModifyClassesCommandsTest.ts
```diff
@@ -0,0 +1,249 @@
...
+      context('When there is no class already on the cell', () => {
```
lnewson, line 148: This description doesn't match, it's not doing operations on the cell and instead is doing it on the table.

modules/tinymce/src/plugins/table/test/ts/browser/command/ModifyClassesCommandsTest.ts
```diff
@@ -0,0 +1,249 @@
...
+
+  context('TINY-7163: Modify cell class', () => {
+    context('Can use mceTableCellToggleClass to toggle the cell class', () => {
+      context('When there is no class already on the cell', () => {
+        const getContentWithoutClass = () => {
+          return (
+            '<table>' +
+              '<tbody>' +
+                '<tr>' +
+                  (platform.browser.isIE() ? '<td>1</td>' : '<td class="">1</td>') +
```
lnewson, line 52: Why is this needed? This smells of a bug being somewhere, as we should have `class=""` if we're testing without a class. Additionally, if we do want it then it should be the same on all browsers for something like this.

modules/tinymce/src/plugins/table/test/ts/browser/command/ModifyClassesCommandsTest.ts
```diff
@@ -0,0 +1,249 @@
...
+          execCmdAndAssertEvent('TINY-7163', editor, 'mceTableToggleClass', false, 'a');
+
+          assert.equal(cleanTableHtml(editor.getContent()), getContentWithClass());
+        });
+
+        it('Can be toggled off', () => {
+          const editor = hook.editor();
+          editor.setContent(getContentWithClass());
+
+          TinySelections.setSelection(editor, [ 0, 0, 0 ], 0, [ 0, 0, 0 ], 1, true);
```
lnewson, line 240: Can you use some different selections where it makes sense throughout this test please? That makes sure it works for things that aren't just in the first cell.

modules/tinymce/src/plugins/table/test/ts/browser/command/ModifyClassesCommandsTest.ts
```diff
@@ -0,0 +1,249 @@
...
+
+          execCmdAndAssertEvent('TINY-7163', editor, 'mceTableToggleClass', false, 'a');
+
+          assert.equal(cleanTableHtml(editor.getContent()), getContentWithoutClass());
+        });
+      });
+
+      context('When there is already a class on the cell, which should not be deleted', () => {
```
lnewson, line 198: Again this context doesn't match what's actually happening here.

lnewson (changes requested):
The core of the logic seems sound, so nice work there :+1: Unfortunately, there's some cleanup work that needs to be done for this sorry @HAFRMO.

modules/tinymce/src/plugins/table/main/ts/api/Commands.ts
lnewson: Also not entirely related to your changes, but we are getting this pattern duplicated a lot. As such, we should probably look at pulling that duplication out as well. eg:
```
const performActionOnSelection = (f: (table: SugarElement<HTMLTableElement>, startCell: SugarElement<HTMLTableCellElement>) => void) =>
  getSelectionStartCell(editor).each((cell) => {
    getTableFromCell(cell).each((table) => {
      f(table, cell);
    });
  });
```

modules/tinymce/src/plugins/table/main/ts/ui/Buttons.ts
HAFRMO: This is as per the spec.

lnewson: I haven't seen an official spec for this, so it's likely only a draft version. This is not how we do things in TinyMCE for an optional setting, so where this is mentioned it needs correcteing.

modules/tinymce/src/plugins/table/test/ts/browser/command/ModifyClassesCommandsTest.ts
HAFRMO: If you remember I do believe I asked you specifically about this. Short version, `class.toggle` leaves behind the `class=""` even though it shouldn't (As `class.remove` does not leave behind this empty string), but IE used a different method which doesn't. It's unrelated to this issue.

lnewson: Ahh, yep I do recall that now sorry. We should fix that at some stage, but as we said for now it's fine/out of scope 👍

modules/tinymce/src/plugins/table/main/ts/ui/Buttons.ts
HAFRMO: Very well. I'll remove the error-logging.

metricjs (approved):

modules/tinymce/CHANGELOG.md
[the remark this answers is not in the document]
metricjs: Oops, old comments. Ignore.

modules/tinymce/CHANGELOG.md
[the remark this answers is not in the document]
metricjs: Oops, old comments. Ignore.

metricjs:
Saw a couple nitpicky style things we could clean up, but they weren't directly your code so I figure they're better done later, once all your changes are through.

modules/tinymce/src/plugins/table/main/ts/api/Commands.ts
```diff
@@ -69,6 +72,34 @@ const registerCommands = (editor: Editor, actions: TableActions, cellSelection:
 
   const getTableFromCell = (cell: SugarElement<HTMLTableCellElement>) => TableLookup.table(cell, isRoot);
 
+  const performActionOnSelection = (action: ExecuteAction) => {
+    getSelectionStartCell(editor).each((cell) => {
```
metricjs, line 76: I guess some of the `{` and `}` could be removed here, since these are all one line in each indented function?

lnewson:
Another issue I've realised from playing with this more is it doesn't handle the table/cell class options properly. The issue is the setting was designed to be exclusive, as such it includes a `None` option. This isn't going to work with a multi-select setup.

As such, we'll either need to filter that option out for the new buttons or add special handling to make it remove all the classes. The latter would be confusing from a UX perspective, so personally I'd opt for filtering it out.

- [ ] Handle the none option

modules/tinymce/src/plugins/table/main/ts/api/Commands.ts
```diff
@@ -69,6 +72,34 @@ const registerCommands = (editor: Editor, actions: TableActions, cellSelection:
...
+      getTableFromCell(cell).each((table) => {
+        action(table, cell);
+      });
+    });
+  };
+
+  const toggleTableClass = (_ui: boolean, requestedClass: string) => {
```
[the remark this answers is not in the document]
lnewson: I should also add having requested in the name can sound like it's something that may not exist and needs to be looked up. That's an additional reason I'm not a fan and would prefer to stick to conventions.

modules/tinymce/src/plugins/table/main/ts/api/Commands.ts
```diff
@@ -69,6 +72,34 @@ const registerCommands = (editor: Editor, actions: TableActions, cellSelection:
...
+    performActionOnSelection((table) => {
+      Class.toggle(table, requestedClass);
+
+      Events.fireTableModified(editor, table.dom, Events.styleModified);
```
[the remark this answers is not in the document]
HAFRMO: I disagree in the claim that it makes it easier to read the code, that really is opinion-based. But I'll change it.

modules/tinymce/src/plugins/table/main/ts/api/Commands.ts
HAFRMO: I was going to, looked away, and then thought I'd done it. My bad.

HAFRMO:
> Another issue I've realised from playing with this more is it doesn't handle the table/cell class options properly. The issue is the setting was designed to be exclusive, as such it includes a `None` option. This isn't going to work with a multi-select setup.

It can include a none-option, but does not seem to have it by default ( it does not seem to have anything as a default ). Do we really want to filter out None, if it can be included in different languages?

> As such, we'll either need to filter that option out for the new buttons or add special handling to make it remove all the classes. The latter would be confusing from a UX perspective, so personally I'd opt for filtering it out. We should confirm with Design though.

It's already been decided that it won't be the later, that was one of the things that were tried and rejected.

modules/tinymce/src/plugins/table/main/ts/ui/ButtonToggleUtils.ts
```diff
@@ -0,0 +1,31 @@
...
+import { Singleton } from '@ephox/katamari';
+import Editor from 'tinymce/core/api/Editor';
+import { Toolbar } from 'tinymce/core/api/ui/Ui';
+
+const onSetupToggle = (editor: Editor, styleName: string, styleValue: string) => {
+  return (api: Toolbar.ToolbarMenuButtonInstanceApi) => {
+    const boundCallback = Singleton.unbindable();
+
+    const init = () => {
+      api.setActive(editor.formatter.match(styleName, { value: styleValue }));
```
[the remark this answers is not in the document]
HAFRMO: In my rush to get out the PR I forgot to include them.

lnewson (commented):
Oops sorry, didn't submit this last night sorry 😞

lnewson:
> It can include a none-option, but does not seem to have it by default ( it does not seem to have anything as a default ). Do we really want to filter out None, if it can be included in different languages?

We shouldn't be looking at the "None" text, but instead the none value, as that's something supported for this option. So we need to exclude anything with an empty value.

modules/tinymce/src/plugins/table/main/ts/api/Commands.ts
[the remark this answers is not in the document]
lnewson: I had discussed this one with the editor team before commenting fwiw and they agreed it was overly verbosely laid out. If you look at the rest of the TinyMCE code style it follows a similar pattern, so we want to keep it fairly similar.

modules/tinymce/src/plugins/table/main/ts/ui/ButtonToggleUtils.ts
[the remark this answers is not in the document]
lnewson: This raises a whole new set of questions then sorry... why aren't we using the `editor.formatter.toggle` API inside the commands then and instead are trying to do the toggling ourselves? The problem is that the button match logic is not the same now for how it actually gets applied. As an example, someone could tweak the `tableclass` format and that would then cause things to break.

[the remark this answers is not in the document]
lnewson: As a side note, remember we can pass a specific node to `editor.formatter.toggle` like so:
```
editor.formatter.toggle('tableclass', { value: styleValue }, tableNode); 
```
Just mentioning that in case it was done manually since most times the formatter works on the selection instead.

modules/tinymce/src/plugins/table/main/ts/ui/ButtonToggleUtils.ts
```diff
@@ -0,0 +1,31 @@
...
+ * Licensed under the LGPL or a commercial license.
+ * For LGPL see License.txt in the project root for license information.
+ * For commercial licenses see https://www.tiny.cloud/
+ */
+
+import { Singleton } from '@ephox/katamari';
+import Editor from 'tinymce/core/api/Editor';
+import { Toolbar } from 'tinymce/core/api/ui/Ui';
+
+const onSetupToggle = (editor: Editor, styleName: string, styleValue: string) => {
```
lnewson, line 12: `formatName` would probably be more accurate than `styleName` here maybe?

modules/tinymce/src/plugins/table/main/ts/ui/Buttons.ts
```diff
@@ -5,10 +5,35 @@
...
+import { Menu } from 'tinymce/core/api/ui/Ui';
+import { ClassList, getCellClassList, getTableClassList, getToolbar } from '../api/Settings';
 import { Clipboard } from '../core/Clipboard';
 import { SelectionTargets, LockedDisable } from '../selection/SelectionTargets';
+import { onSetupToggle } from './ButtonToggleUtils';
+
+const filterNoneItem = (list: ClassList) =>
+  Arr.filter(list, (item) =>
+    item.value !== ''
+  );
```
lnewson, line 19:
```suggestion
  Arr.filter(list, (item) => item.value !== '');
```
Also, this should really be using `Strings.isNotEmpty(item.value)` instead, however I can't easily make that suggestion.

- [ ] Formatting and use `Strings.isNotEmpty` please

modules/tinymce/src/plugins/table/test/ts/browser/ui/TableClassListButtonsTest.ts
```diff
@@ -0,0 +1,126 @@
...
+
+  let events: Array<EditorEvent<TableModifiedEvent>> = [];
+  const logEvent = (event: EditorEvent<TableModifiedEvent>) => {
+    events.push(event);
+  };
+
+  afterEach(() => {
+    events = [];
+
+    try { // Check if the menu is still open. If it is, close it.
```
lnewson, line 41: Why is this in a try-catch loop? This has a massive code smell about it and it would appear to be done because individual tests aren't cleaning up.

- [ ] Ideally remove this and make sure menus are closed in the tests where they are opened

modules/tinymce/src/plugins/table/test/ts/browser/ui/TableClassListButtonsTest.ts
```diff
@@ -0,0 +1,126 @@
...
+      UiFinder.exists(sugarContainer, '.tox-collection__item[aria-checked="false"]'); // No checked ones exist.
+      UiFinder.notExists(sugarContainer, '.tox-collection__item[aria-checked="true"]');
+    });
+
+    editor.execCommand('mceTableToggleClass', false, 'a');
+    assert.lengthOf(events, 1, 'Command executed successfully.');
+
+    // Cycle the menu to let the checkmark load.
+    TinyUiActions.clickOnToolbar(editor, 'button[title="Table styles"]');
+    TinyUiActions.clickOnToolbar(editor, 'button[title="Table styles"]');
```
lnewson, line 78: This is rather confusing... a better/clearer way to handle closing menus which is what's used elsewhere is to send the `escape` key event. I'd also recommend splitting these out into `openMenu` and `closeMenu` functions which will improve readability greatly. Another suggestion I'd have is to move the close part to immediately after we've asserted the menu state.

- [ ] Improve this please

modules/tinymce/src/plugins/table/test/ts/browser/ui/TableClassListButtonsTest.ts
```diff
@@ -0,0 +1,126 @@
...
+
+    const sugarContainer = SugarBody.body();
+
+    TinyUiActions.clickOnToolbar(editor, 'button[title="Table styles"]');
+
+    await Waiter.pTryUntil('There should not be a checkmark', () => {
+      UiFinder.exists(sugarContainer, '.tox-menu');
+      UiFinder.exists(sugarContainer, '.tox-collection__item[aria-checked="false"]'); // No checked ones exist.
+      UiFinder.notExists(sugarContainer, '.tox-collection__item[aria-checked="true"]');
+    });
```
lnewson, line 71: This logic is duplicated in tests below, please pull this out into a separate function. eg:
```
const pTestMenuState = async (editor: Editor) => {
   // Open menu
   TinyUiActions.clickOnToolbar(editor, 'button[title="Table styles"]');
   // Check the menu state
  const sugarContainer = SugarBody.body();
  await Waiter.pTryUntil('There should not be a checkmark', () => {
    UiFinder.exists(sugarContainer, '.tox-menu');
    UiFinder.exists(sugarContainer, '.tox-collection__item[aria-checked="false"]'); // No checked ones exist.
    UiFinder.notExists(sugarContainer, '.tox-collection__item[aria-checked="true"]');
  });
  // Close the menu
  TinyUiActions.clickOnToolbar(editor, 'button[title="Table styles"]');
};
```

- [ ] Cleanup duplication

modules/tinymce/src/plugins/table/test/ts/browser/ui/TableClassListButtonsTest.ts
```diff
@@ -0,0 +1,126 @@
...
+
+    editor.execCommand('mceTableToggleClass', false, 'a');
+    editor.execCommand('mceTableToggleClass', false, 'b');
+    assert.lengthOf(events, 2, 'Command executed successfully.');
+
+    // Cycle the menu to let the checkmark load.
+    TinyUiActions.clickOnToolbar(editor, 'button[title="Table styles"]');
+    TinyUiActions.clickOnToolbar(editor, 'button[title="Table styles"]');
+
+    await Waiter.pTryUntil('There should be a checkmark', () => {
```
lnewson, line 120: This doesn't really check what's mentioned in the description. It's only checking that 1 checkmark has appeared and that there's no others, it's possible that only 1 menu item rendered. It might be a good idea to use the `Assertions.assertPresence` API here instead which would allow you to assert there's 2 checkmarks.

- [ ] Make sure this asserts there's 2 checked menu items

lnewson (approved):

modules/tinymce/src/plugins/table/test/ts/browser/ui/TableClassListButtonsTest.ts
```diff
@@ -0,0 +1,121 @@
...
+    TinySelections.setSelection(editor, [ 0, 0, 0 ], 0, [ 0, 0, 0 ], 1);
+  };
+
+  const openMenu = (editor: Editor) =>
+    TinyUiActions.clickOnToolbar(editor, 'button[title="Table styles"]');
+
+  const closeMenu = (editor) =>
+    TinyUiActions.keydown(editor, Keys.escape());
+
+  const checkMenuPresence = async (editor: Editor, label: string, expected: Record<string, number>, container: SugarElement<HTMLElement>) => {
```
lnewson, line 63:
```suggestion
  const pCheckMenuPresence = async (editor: Editor, label: string, expected: Record<string, number>, container: SugarElement<HTMLElement>) => {
```
Helpers in tests currently use the p prefix when it's async, so we should keep that convention for now. There also other helper functions here, so please make sure to update all of them.

lnewson (approved):

ltrouton (approved):

lnewson:
As mentioned on Slack and the newer PR, I missed that this only adds toolbar buttons. We have a rule that if there's a toolbar button, then there should be an equivalent menu item (or vice versa) to help with the customizability of the editor. As such, we'll need 2 equivalent menu items added for these new toolbar buttons as well.

- [ ] Add menu items to match toolbar functionality

lnewson:
Thanks for adding the menu items @HAFRMO, however I think we can do this a better way. We don't normally like to register menu items for dynamic menu items like this and generating the names like this could lead to some weird issues or the menu items being misused. The other issue is it makes it hard to see where the implementation is for the buttons since it's relying on the dynamically generated items in `MenuItems.ts`. So with that in mind, here's a way that would solve those concerns by moving the dynamic generation to `UiUtils`: https://github.com/tinymce/tinymce/compare/feature/TINY-7476...feature/TINY-7476-3

P.S. Feel free to just cherry-pick/merge those changes directly if you want.

lnewson (approved):

PR #6880, opened 2021-05-30
Treat the default/undefined `schema` as 'html5', as specified by docs.

mnoorenberghe:
Description of Changes:
The documentation claims that [the default schema is 'html5'](https://www.tiny.cloud/docs/configure/content-filtering/#schema) but nowhere is this default actually specified so the default value ends up being `undefined`. This is fine for code like `schema !== 'html4'` but doesn't work for `schema !== 'html5'` since it will evaluate to true when the schema is undefined.

The impact of this problem is that `schema: 'html5'` actually differs from omitting the schema. Only an explicit 'html5' actually disables the conversion from <b> => <strong> and <i> => <em>.

This change defaults the schema to 'html5' when `undefined`.

Pre-checks:
* [x] Changelog entry added
* [x] Tests have been added (if applicable) => Tests corrected.

Review:
* [ ] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Changed the load order so that the content css gets loaded before the editor gets populated with contents. #TINY-7249
 
 ### Fixed
+- The `schema` correctly defaults to 'html5', as documented. <b> and <i> aren't converted.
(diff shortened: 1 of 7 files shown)


metricjs:
Hi @mnoorenberghe, thanks for this PR. We are aware of this issue, but unfortunately this change needs to happen in a major release as TinyMCE follows [semantic versioning](https://semver.org/) and this would be a breaking change. Solving this issue is being considered internally for TinyMCE 6.0, along with some other related issues.

That said, we will look into clarifying the docs regarding this setting, as that doesn't need to wait for a release.

PR #6887, opened 2021-06-01
Exported more types from robin

ashfordneil:
Related Ticket: N/A

Description of Changes:
* Export some types from Robin so that we're able to import them directly instead of using weird `ReturnType` shenanigans

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/robin/CHANGELOG.md
+++ b/modules/robin/CHANGELOG.md
@@ -1,0 +1,13 @@
+# Changelog
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
+and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
+
+## Unreleased
+
+### Added
+* Made `TextZones.Zone` type public
+* Made `Words.WordScope` type public
+
+## 8.0.8
--- a/modules/robin/CHANGELOG.md
+++ b/modules/robin/CHANGELOG.md
@@ -7,7 +7,5 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## Unreleased
 
 ### Added
-* Made `TextZones.Zone` type public
-* Made `Words.WordScope` type public
-
-## 8.0.8
+- Made `TextZones.Zone` type public.
+- Made `Words.WordScope` type public.
--- a/modules/robin/CHANGELOG.md
+++ b/modules/robin/CHANGELOG.md
@@ -7,5 +7,4 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## Unreleased
 
 ### Added
-- Made `TextZones.Zone` type public.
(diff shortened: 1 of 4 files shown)


modules/robin/CHANGELOG.md
```diff
@@ -0,0 +1,13 @@
+# Changelog
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
+and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
+
+## Unreleased
+
+### Added
+* Made `TextZones.Zone` type public
```
lnewson, line 10: These should be hyphens, not asterisks. Also please either add a fullstop at the end, or add a JIRA number.

modules/robin/CHANGELOG.md
```diff
@@ -0,0 +1,13 @@
...
+* Made `Words.WordScope` type public
+
+## 8.0.8
```
lnewson, line 13: This isn't valid. Either add the date and entries for 8.0.8 or just remove this please.

- [ ] Fix invalid changelog

versions.txt
```diff
@@ -5,3 +5,4 @@ acid@3.1.0
 agar@5.4.0
 alloy@8.3.0
 mcagar@6.2.0
+robin@8.1.0
```
lnewson, line 8: Why is this a minor version bump? There's nothing about the API that's changed and instead we're just fixing types that should have been exposed originally as far as I can see.

- [ ] Remove minor version bump unless we have a good reason

lnewson (approved):

HAFRMO (approved):

versions.txt
ashfordneil: We're exporting a thing that wasn't previously exported — even if it maybe should have been exported before, that sounds like a minor bump to me.

Worth noting, this will probably get promoted to a major bump as part of the Optionals work in the other PR, so it's probably a moot point.

lnewson: As discussed via Slack, I disagree with that statement but do agree this will end up being a minor anyways. As such, I'm going to resolve this.

modules/robin/CHANGELOG.md
```diff
@@ -0,0 +1,11 @@
...
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
+and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
+
+## Unreleased
+
+### Added
+- Made `TextZones.Zone` type public.
+- Made `Words.WordScope` type public.
```
lnewson, line 11:
```suggestion
- Made `TextZones.Zone` and `Words.WordScope` types public.
```

lnewson: Actually, we can probably reword that further to something like this:
```suggestion
- Exposed the `TextZones.Zone` and `Words.WordScope` types.
```

PR #6892, opened 2021-06-02
TINY-3254: Fixed EventDispatcher mutating the original event data

lnewson:
Related Ticket: TINY-3254

Description of Changes:
* Fixed `EventDispatcher` mutating the data provided
* Reworked the logic to share the clone/normalize functionality between `EventUtils` and `EventDispatcher`
* Fixed a small bug in `SetSelectionContent` I found when cleaning up the incorrect reliance on mutation when getting and setting content.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/api/util/EventDispatcher.ts
+++ b/modules/tinymce/src/core/main/ts/api/util/EventDispatcher.ts
@@ -6,6 +6,7 @@
  */
 
 import { Arr, Fun, Obj } from '@ephox/katamari';
+import * as EventUtils from '../../events/EventUtils';
 import Tools from './Tools';
 
 export type MappedEvent<T, K extends string> = K extends keyof T ? T[K] : any;
@@ -53,16 +54,7 @@ export interface NativeEventMap {
   'wheel': WheelEvent;
 }
 
-export type EditorEvent<T> = T & {
-  target: any;
-  type: string;
-  preventDefault: () => void;
-  isDefaultPrevented: () => boolean;
-  stopPropagation: () => void;
-  isPropagationStopped: () => boolean;
-  stopImmediatePropagation: () => void;
-  isImmediatePropagationStopped: () => boolean;
-};
+export type EditorEvent<T> = EventUtils.NormalizedEvent<T>;
 
 export interface EventDispatcherSettings {
   scope?: any;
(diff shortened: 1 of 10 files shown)


modules/tinymce/src/core/main/ts/api/util/EventDispatcher.ts
```diff
@@ -195,24 +158,23 @@ class EventDispatcher<T> {
 
         // Unbind handlers marked with "once"
         if (callback.once) {
-          this.off(name, callback.func);
+          this.off(lcName, callback.func);
         }
 
         // Stop immediate propagation if needed
-        if (args.isImmediatePropagationStopped()) {
-          args.stopPropagation();
```
lnewson, line 203: Note: This was intentionally removed, as it's no longer needed since the normalized function calls this immediately.

ashfordneil (approved):

modules/tinymce/src/core/main/ts/content/GetContentImpl.ts
```diff
@@ -23,38 +23,36 @@ const trimEmptyContents = (editor: Editor, html: string): string => {
...
-  args.format = format;
-  args.get = true;
-  args.getInner = true;
-
-  if (!args.no_events) {
-    editor.fire('BeforeGetContent', args);
-  }
-
-  if (args.format === 'raw') {
+  const updatedArgs = args.no_events ? args : editor.fire('BeforeGetContent', {
```
spocke, line 26: Yeah, this mutation is something I wanted to get rid of for a long time but I've been afraid that implementations out there would break. Is this something we could schedule for 6.0 cloning these instead of letting the event handlers mutate things.

spocke (approved):

modules/tinymce/src/core/main/ts/content/GetContentImpl.ts
lnewson: Yeah, I had the same concern. When we discussed this at standup though with Product, QA, and the rest of the editor team, we decided the implications of it mutating were worse than things potentially breaking for something that's not documented, nor really expected. Additionally, it's not that common for people to be firing custom events, they'd generally normally listen instead for existing events. It's also a relatively simple fix if someone was incorrectly relying on the mutation :shrug: 

I should add from that meeting, we did mention this would definitely have to be called out in the release notes with the workaround for anyone relying on it. That said, I'll bring it up again at standup today and make 100% sure we are still happy to go along with this.

ltrouton (approved):

modules/tinymce/src/core/main/ts/content/GetContentImpl.ts
EkimChau: I'm happy to go with a fix in 5.9 given the the fix for anyone who is relying on Mutation is small and the benefits it provides.  We will do our best to alert to the community of the potential breaking changes if they rely on event mutation, we will pin a message on github issues as an early warning followed up by Docs.

JamesToohey (approved):

PR #6905, opened 2021-06-05
Added return types and fixed no implicit any errors for modules in the commands and content directories

lnewson:
Related Ticket: N/A

Description of Changes:
This is a follow on from https://github.com/tinymce/tinymce/pull/6806, so same situation as then whereby this is just trying to get more types declared in core. This isn't meant to be an exhaustive conversion and certainly somethings aren't typed still (should only be variables inside functions).

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
+++ b/modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
@@ -9,9 +9,9 @@ import { Optional, Type } from '@ephox/katamari';
 import { SugarElement } from '@ephox/sugar';
 import DOMUtils from '../api/dom/DOMUtils';
 import ElementUtils from '../api/dom/ElementUtils';
-import EditorSelection from '../api/dom/Selection';
 import Editor from '../api/Editor';
 import Env from '../api/Env';
+import { ParserArgs } from '../api/html/DomParser';
 import AstNode from '../api/html/Node';
 import HtmlSerializer from '../api/html/Serializer';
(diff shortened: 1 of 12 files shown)


JamesToohey (approved):

modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
```diff
@@ -213,10 +214,11 @@ const deleteSelectedContent = (editor: Editor) => {
   }
 };
 
-export const insertHtmlAtCaret = (editor: Editor, value: string, details) => {
-  let parentNode, rootNode, args;
-  let marker, rng, node;
-  const selection: EditorSelection = editor.selection, dom = editor.dom;
+export const insertHtmlAtCaret = (editor: Editor, value: string, details: InsertContentDetails): void => {
+  let parentNode;
+  let rng, node;
```
ltrouton, line 219: Wondering if these can be given types or do they need to be `any`?

ltrouton (approved):

modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
lnewson: The types for these are complicated, partly because there's still IE 8 code in here. As I said in the description, it's known that some of these won't have types as the intention at this stage is to only fix the APIs not the internals.

PR #6908, opened 2021-06-08
TINY-7168: update word count on ViewUpdate

vpyshnenko:
Related Ticket: TINY-7168

Description of Changes:
* Placeholder text

Pre-checks:
* [ ] Changelog entry added

Review:
* [ ] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/api/EventTypes.ts
+++ b/modules/tinymce/src/core/main/ts/api/EventTypes.ts
@@ -101,6 +101,7 @@ export interface EditorEventMap extends Omit<NativeEventMap, 'blur' | 'focus'> {
   'ProgressState': ProgressStateEvent;
   'AfterProgressState': AfterProgressStateEvent;
   'PlaceholderToggle': PlaceholderToggleEvent;
+  'ViewUpdate': { };
   'tap': TouchEvent;
   'longpress': TouchEvent;
   'longpresscancel': { };
--- a/modules/tinymce/src/plugins/wordcount/main/ts/core/WordCounter.ts
+++ b/modules/tinymce/src/plugins/wordcount/main/ts/core/WordCounter.ts
@@ -20,7 +20,7 @@ const setup = (editor: Editor, api: WordCountApi, delay: number) => {
   editor.on('init', () => {
     updateCount(editor, api);
     Delay.setEditorTimeout(editor, () => {
-      editor.on('SetContent BeforeAddUndo Undo Redo keyup', debouncedUpdate);
+      editor.on('SetContent BeforeAddUndo Undo Redo ViewUpdate keyup', debouncedUpdate);
     }, 0);
   });
 };


modules/tinymce/CHANGELOG.md
```diff
@@ -29,6 +29,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Binding an event handler incorrectly took effect immediately while the event was firing #TINY-7436
 - Partially transparent RGBA values provided in the `color_map` setting were given the wrong hex value #TINY-7163
 - Fixed an issue when pasting cells from tables containing `colgroup`s into tables without `colgroup`s #TINY-6675
+- Word count was not updated when editor content has been changed by remote part #TINY-7168
```
lnewson, line 32: This is for RTC so we shouldn't have a changelog or types for this yet.
```suggestion
```

modules/tinymce/src/core/main/ts/api/EventTypes.ts
```diff
@@ -101,6 +101,7 @@ export interface EditorEventMap extends Omit<NativeEventMap, 'blur' | 'focus'> {
   'ProgressState': ProgressStateEvent;
   'AfterProgressState': AfterProgressStateEvent;
   'PlaceholderToggle': PlaceholderToggleEvent;
+  'ViewUpdate': { };
```
lnewson, line 104:
```suggestion
```
As mentioned in my previous comment, this isn't part of the current core so we shouldn't have types, etc... if we were to add this then it should also have an implementation.

lnewson (approved):

modules/tinymce/src/core/main/ts/api/EventTypes.ts
TheSpyder: I was half expecting to do the implementation now, and have word count listen for _only_ `ViewUpdate`, but maybe that should wait for Tiny 6.

spocke: I guess if we hide it now and don't expose it or document it we can more easily change it if needed then we more the alt model into core. I thought it would be nice to expose it since it's now used by a core plugin.

modules/tinymce/CHANGELOG.md
spocke: A rule of thumb is if the PR is not changing editor behavior or editor apis then there is no need for a changelog item. For example internal refactoring to get rtc working better is only effecting rtc users like this case. But it's a bit of a gray area since we are adding a event but it will only be emitted when the rtc plugin is enabled.

vpyshnenko: removed changelog update

modules/tinymce/src/core/main/ts/api/EventTypes.ts
vpyshnenko: done

lnewson (approved):

PR #6910, opened 2021-06-08
TINY-7545: Fixed a number of positioning bugs found in Bounder

lnewson:
Related Ticket: TINY-7545

Description of Changes:
This is part 1 of 3 PRs to get this working. This fixes a number of bugs found in the alloy bounder and layout bounds code largely around `LayoutInside` layouts and also does a little code cleanup (eg. deleting the unused `Styles.ts` and `Anchors.ts` files).

**Note:** There are no tests for layout inside at all, so this is largely new tests for `Bounder` and `LayoutBounds`. The tests for `LayoutInside` will be in the next PR, as there were bugs/inconsistencies found in them as well.

Pre-checks:
* [x] Changelog entry added (Note: No TinyMCE changelog yet, that'll be in the last PR).

Review:
* [x] Milestone set


modules/alloy/src/main/ts/ephox/alloy/positioning/mode/ContentAnchorCommon.ts
```diff
@@ -9,19 +9,9 @@ import * as Origins from '../layout/Origins';
 import { Anchoring, NodeAnchor, nu as NuAnchor, SelectionAnchor } from './Anchoring';
 import * as AnchorLayouts from './AnchorLayouts';
 
-const capRect = (left: number, top: number, width: number, height: number): Optional<Boxes.BoxByPoint> => {
```
lnewson, line 12: This logic here was only a workaround to a larger issue and only resolved it for things rendered near the top of the screen, not the bottom. It's now better handled within Bounder by its constraints/restrictions.

modules/alloy/src/main/ts/ephox/alloy/positioning/view/Bounder.ts
```diff
@@ -165,34 +197,27 @@ const attempt = (candidate: SpotInfo, width: number, height: number, bounds: Box
...
     const next: SpotInfo = layout(anchorBox, elementBox, bubbles);
     const attemptLayout = attempt(next, panelWidth, panelHeight, bounds);
 
-    // unwrapping fit only to rewrap seems... silly
-    return attemptLayout.fold(adt.fit, (newReposition, newDeltaW, newDeltaH) => {
-      // console.log(`label: ${next.label()}, newDeltaW: ${newDeltaW}, deltaW: ${deltaW}, newDeltaH: ${newDeltaH}, deltaH: ${deltaH}`);
-      const improved = newDeltaH > deltaH || newDeltaW > deltaW;
+    return attemptLayout.fold(Fun.constant(attemptLayout), (newReposition, newVisibleW, newVisibleH, newIsVisible) => {
+      // console.log(`label: ${next.label}, newVisibleW: ${newVisibleW}, visibleW: ${visibleW}, newVisibleH: ${newVisibleH}, visibleH: ${visibleH}, newIsVisible: ${newIsVisible}, isVisible: ${isVisible}`);
+      const improved = isVisible === newIsVisible ? (newVisibleH > visibleH || newVisibleW > visibleW) : (!isVisible && newIsVisible);
```
ashfordneil, line 206: This is something from the old code, so feel free to ignore, but I'm wondering if we have a bug here?

Is `newVisibleH > visibleH || newVisibleW > visibleW` the right heuristic? Imagine two attempts: one that is horrible vertically, but near perfect horizontally, and the other that's near perfect vertically, and horrible horizontally. The only thing that determines which is better* here is the order they're listed — and we're setting things up to prefer the one that comes last in the list, whereas normally when we're giving a list of things to try to a function we list them with the ones we want first.

*assuming they're either both invisible or both visible

modules/alloy/src/main/ts/ephox/alloy/positioning/view/Bounder.ts
```diff
@@ -11,146 +11,178 @@ import { SpotInfo } from './SpotInfo';
...
-  const boundsX = bounds.x;
-  const boundsY = bounds.y;
-  const boundsWidth = bounds.width;
-  const boundsHeight = bounds.height;
+/**
+ * This will attempt to determine if the box will fit within the specified bounds or if it needs to be repositioned.
+ * It will return the following details:
+ *  - if the original rect was in bounds (originInBounds & sizeInBounds). This is used to determine if we fitted
+ *    without having to make adjustments.
+ *  - the height and width deltas in relation to how much height/width would be visible in the original location.
```
ashfordneil, line 48:
```suggestion
 *  - the width and height of the portion of the original rect that fits inside the bounds (may be negative if there is no overlap between the original rect and the bounds)
```

modules/alloy/src/test/ts/atomic/position/layout/LayoutBoundsTest.ts
```diff
@@ -0,0 +1,87 @@
...
+            return Optional.some(anchorBox.right);
+          case 'top':
+            return Optional.some(anchorBox.y);
+          case 'bottom':
+            return Optional.some(anchorBox.bottom);
+        }
+      } else {
+        return Optional.none();
+      }
+    }))
```
ashfordneil, line 29:
```suggestion
    fc.constantFrom(...dimensions).map((dim) => ({
      left: Optionals.someIf(dim === 'left', anchorBox.x),
      right: Optionals.someIf(dim === 'right', anchorBox.right),
      top: Optionals.someIf(dim === 'top', anchorBox.y),
      bottom: Optionals.someIf(dim === 'bottom', anchorBox.bottom)
    }))
```

modules/alloy/src/test/ts/atomic/position/layout/LayoutBoundsTest.ts
```diff
@@ -0,0 +1,87 @@
...
+      const newBounds = LayoutBounds.adjustBounds(bounds, restriction, SugarPosition(bubbleLeft, bubbleRight));
+      assertInBounds(newBounds, bounds);
+    }));
+  });
+
+  it('TINY-7545: the left coord should be the anchor box/bounds plus the bubble offset', () => {
+    const bounds = Boxes.bounds(0, 0, 100, 100);
+    const inBoundsRestriction = LayoutBounds.boundsRestriction(Boxes.bounds(0, 0, 10, 100), { left: LayoutBounds.AnchorBoxBounds.RightEdge });
+    const outBoundsRestriction = LayoutBounds.boundsRestriction(Boxes.bounds(-20, 0, 10, 100), { left: LayoutBounds.AnchorBoxBounds.RightEdge });
+    fc.assert(fc.property(fc.integer(0, 90), fc.integer(0, 100), (bubbleLeft, bubbleTop) => {
```
ashfordneil, line 44: Can we leave something like this for when we forget why that's 90?
```suggestion
    // Make sure bubble-left is at least 10 less than the bounds so that inBoundsRestriction will always fit
    fc.assert(fc.property(fc.integer(0, 90), fc.integer(0, 100), (bubbleLeft, bubbleTop) => {
```

modules/alloy/src/test/ts/atomic/position/view/BounderDeterminePositionTest.ts
```diff
@@ -0,0 +1,77 @@
+import { describe, it } from '@ephox/bedrock-client';
+import { assert } from 'chai';
+
+import * as Boxes from 'ephox/alloy/alien/Boxes';
+import * as Bounder from 'ephox/alloy/positioning/view/Bounder';
+
+describe('BounderDeterminePositionTest', () => {
+  it('inside the bounds', () => {
```
ashfordneil, line 8: Should these have Jira numbers / TBAs?

ashfordneil (approved):

modules/alloy/src/test/ts/atomic/position/view/BounderDeterminePositionTest.ts
lnewson: Yeah, this should have a JIRA number thanks.

modules/alloy/src/test/ts/atomic/position/layout/LayoutBoundsTest.ts
lnewson: Oh, nice suggestion thanks!

modules/alloy/src/main/ts/ephox/alloy/positioning/view/Bounder.ts
lnewson: That's not correct, it's literally the height/width delta not a portion of it.

ashfordneil: Isn't it just the size of the overlap?

lnewson: Yeah it is, but `portion` is the part I don't like/don't consider correct. The original wording is clearer for me at least.

ashfordneil: To me the original wording didn't really make sense, and I had to dig around to find out what these fields were. Most of that I think comes from calling them deltas, because it's not quite a difference between two things. Can we just describe it as:

```suggestion
 *  - the height and width that would be visible in the original location. (ie the overlap between the rect and the bounds)
```

modules/alloy/src/main/ts/ephox/alloy/positioning/view/Bounder.ts
lnewson: Yeah I'm not sure on that tbh. I'll log a follow up JIRA to investigate that, as we've already spent a considerable amount of additional time on getting this far.

modules/alloy/src/main/ts/ephox/alloy/positioning/view/Bounder.ts
lnewson: I really don't think describes the negative case well. It handles the positive case nicely though 🤷

modules/alloy/src/test/ts/atomic/position/layout/LayoutBoundsTest.ts
lnewson: How about this, as I don't like starting with `Make sure`:
> // Note: bubbleLeft needs to be at least 10 less than the bounds width to make sure it fits

modules/alloy/src/main/ts/ephox/alloy/positioning/view/Bounder.ts
ashfordneil:
```suggestion
 *  - the height and width that would be visible in the original location. (ie the overlap between the rect and the bounds, may be negative if there is no overlap)
```

lnewson: That's still not ideal, but it's better than the original and conveys the negative value so lets go with that. Thanks!

lnewson: Done

modules/alloy/src/test/ts/atomic/position/layout/LayoutBoundsTest.ts
lnewson: Done

modules/alloy/src/test/ts/atomic/position/view/BounderDeterminePositionTest.ts
lnewson: Done

ltrouton (approved):

modules/alloy/src/main/ts/ephox/alloy/positioning/view/Bounder.ts
lnewson: Logged as TINY-7581

PR #6914, opened 2021-06-09
TINY-7479: Added table caption functionality.

HAFRMO:
Related Ticket:

Description of Changes:
Added table's ability to toggle captions

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -9,17 +9,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Added a new `toolbar_sticky_offset` setting to allow the toolbar to be offset from the top of the page #TINY-7337
 - Added a new `mceFocus` command that focuses the editor. Equivalent to using `editor.focus()` #TINY-7373
+- Added a new `mceTableToggleClass` command which toggles the provided class on the currently selected table #TINY-7476
+- Added a new `mceTableCellToggleClass` command which toggles the provided class on the currently selected table cells #TINY-7476
 - Added a new `tablecellvalign` toolbar button and menu item for vertical table cell alignment #TINY-7477
 - Added a new `tablecellborderwidth` toolbar button and menu item to change table cell border width #TINY-7478
 - Added a new `tablecellborderstyle` toolbar button and menu item to change table cell border style #TINY-7478
+- Added a new `tablecaption` toolbar button and menu item to toggle captions on tables #TINY-7479
 
 ### Improved
 - Improved the load time of the `fullpage` plugin by using the existing editor schema rather than creating a new one #TINY-6504
 
 ### Changed
 - Changed the load order so that the content css gets loaded before the editor gets populated with contents. #TINY-7249
-- Added a new `mceTableToggleClass` command which toggles the provided class on the currently selected table #TINY-7476
-- Added a new `mceTableCellToggleClass` command which toggles the provided class on the currently selected table cells #TINY-7476
 
 ### Fixed
 - `editor.fire` was incorrectly mutating the original `args` provided #TINY-3254
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added a new `tablecellborderwidth` toolbar button and menu item to change table cell border width #TINY-7478
 - Added a new `tablecellborderstyle` toolbar button and menu item to change table cell border style #TINY-7478
 - Added a new `tablecaption` toolbar button and menu item to toggle captions on tables #TINY-7479
+- Added a new `mceTableToggleCaption` command that toggles caption on selected table #TINY-7479
 
 ### Improved
 - Improved the load time of the `fullpage` plugin by using the existing editor schema rather than creating a new one #TINY-6504
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added a new `tablecellborderwidth` toolbar button and menu item to change table cell border width #TINY-7478
 - Added a new `tablecellborderstyle` toolbar button and menu item to change table cell border style #TINY-7478
 - Added a new `tablecaption` toolbar button and menu item to toggle captions on tables #TINY-7479
-- Added a new `mceTableToggleCaption` command that toggles caption on selected table #TINY-7479
+- Added a new `mceTableToggleCaption` command that toggles captions on a selected table #TINY-7479
 
 ### Improved
 - Improved the load time of the `fullpage` plugin by using the existing editor schema rather than creating a new one #TINY-6504
--- a/modules/tinymce/src/plugins/table/main/ts/api/Commands.ts
+++ b/modules/tinymce/src/plugins/table/main/ts/api/Commands.ts
@@ -94,9 +94,7 @@ const registerCommands = (editor: Editor, actions: TableActions, cellSelection: 
   const toggleCaption = () => {
     getSelectionStartCellOrCaption(editor).each((cellOrCaption) => {
       TableLookup.table(cellOrCaption, isRoot).each((table) => {
-        const captionElementOpt = Selectors.one('caption', table);
-
-        captionElementOpt.fold(
+        Selectors.one('caption', table).fold(
           () => {
             const caption = SugarElement.fromTag('caption');
             Insert.append(caption, SugarElement.fromText('Caption'));
--- a/modules/tinymce/src/plugins/table/main/ts/api/Commands.ts
+++ b/modules/tinymce/src/plugins/table/main/ts/api/Commands.ts
@@ -8,7 +8,7 @@
 import { Selections } from '@ephox/darwin';
 import { Arr, Fun, Obj, Optional, Type } from '@ephox/katamari';
 import { CopyCols, CopyRows, Sizes, TableFill, TableLookup } from '@ephox/snooker';
-import { Insert, Remove, Replication, Selectors, SugarElement } from '@ephox/sugar';
+import { Insert, Remove, Replication, SelectorFind, SugarElement } from '@ephox/sugar';
 import Editor from 'tinymce/core/api/Editor';
 import { enforceNone, enforcePercentage, enforcePixels } from '../actions/EnforceUnit';
 import { insertTableWithDataValidation } from '../actions/InsertTable';
@@ -94,7 +94,7 @@ const registerCommands = (editor: Editor, actions: TableActions, cellSelection: 
   const toggleCaption = () => {
     getSelectionStartCellOrCaption(editor).each((cellOrCaption) => {
       TableLookup.table(cellOrCaption, isRoot).each((table) => {
-        Selectors.one('caption', table).fold(
+        SelectorFind.child(table, 'caption').fold(
           () => {
             const caption = SugarElement.fromTag('caption');
             Insert.append(caption, SugarElement.fromText('Caption'));
--- a/modules/tinymce/src/plugins/table/main/ts/api/Commands.ts
+++ b/modules/tinymce/src/plugins/table/main/ts/api/Commands.ts
@@ -8,7 +8,7 @@
 import { Selections } from '@ephox/darwin';
 import { Arr, Fun, Obj, Optional, Type } from '@ephox/katamari';
 import { CopyCols, CopyRows, Sizes, TableFill, TableLookup } from '@ephox/snooker';
-import { Insert, Remove, Replication, SelectorFind, SugarElement } from '@ephox/sugar';
+import { Insert, Remove, Replication, SelectorFind, Selectors, SugarElement } from '@ephox/sugar';
 import Editor from 'tinymce/core/api/Editor';
 import { enforceNone, enforcePercentage, enforcePixels } from '../actions/EnforceUnit';
 import { insertTableWithDataValidation } from '../actions/InsertTable';
@@ -100,7 +100,12 @@ const registerCommands = (editor: Editor, actions: TableActions, cellSelection: 
             Insert.append(caption, SugarElement.fromText('Caption'));
             Insert.appendAt(table, caption, 0);
           },
-          Remove.remove
+          (caption) => {
+            Remove.remove(caption);
+            Selectors.one('td', table).each((td) =>
+              editor.selection.setCursorLocation(td.dom, 0)
+            );
+          }
         );
 
         Events.fireTableModified(editor, table.dom, Events.structureModified);
--- a/modules/tinymce/src/plugins/table/main/ts/api/Commands.ts
+++ b/modules/tinymce/src/plugins/table/main/ts/api/Commands.ts
@@ -8,7 +8,7 @@
 import { Selections } from '@ephox/darwin';
 import { Arr, Fun, Obj, Optional, Type } from '@ephox/katamari';
 import { CopyCols, CopyRows, Sizes, TableFill, TableLookup } from '@ephox/snooker';
-import { Insert, Remove, Replication, SugarElement } from '@ephox/sugar';
+import { Insert, Remove, Replication, Selectors, SugarElement } from '@ephox/sugar';
 import Editor from 'tinymce/core/api/Editor';
 import { enforceNone, enforcePercentage, enforcePixels } from '../actions/EnforceUnit';
 import { insertTableWithDataValidation } from '../actions/InsertTable';
@@ -91,6 +91,27 @@ const registerCommands = (editor: Editor, actions: TableActions, cellSelection: 
     });
   };
 
+  const toggleCaption = () => {
+    getSelectionStartCellOrCaption(editor).each((cellOrCaption) => {
+      TableLookup.table(cellOrCaption, isRoot).filter(Fun.not(isRoot)).each((table) => {
+        const captionElementOpt = Selectors.one('caption', table);
+
+        captionElementOpt.fold(
+          () => {
+            const caption = SugarElement.fromTag('caption');
+            Insert.append(caption, SugarElement.fromText('Caption'));
+            Insert.appendAt(table, caption, 0);
+          },
+          (caption) => {
+            Remove.remove(caption);
+          }
+        );
+
+        Events.fireTableModified(editor, table.dom, Events.structureModified);
+      });
+    });
+  };
+
   const postExecute = (table: SugarElement<HTMLTableElement>) => (data: TableActionResult): void => {
     editor.selection.setRng(data.rng);
     editor.focus();
@@ -156,6 +177,7 @@ const registerCommands = (editor: Editor, actions: TableActions, cellSelection: 
     mceTableDelete: eraseTable,
     mceTableCellToggleClass: toggleTableCellClass,
     mceTableToggleClass: toggleTableClass,
+    mceTableToggleCaption: toggleCaption,
     mceTableSizingMode: (_ui: boolean, sizing: string) => setSizingMode(sizing)
   }, (func, name) => editor.addCommand(name, func));
 
--- a/modules/tinymce/src/plugins/table/main/ts/api/Commands.ts
+++ b/modules/tinymce/src/plugins/table/main/ts/api/Commands.ts
@@ -99,6 +99,7 @@ const registerCommands = (editor: Editor, actions: TableActions, cellSelection: 
             const caption = SugarElement.fromTag('caption');
             Insert.append(caption, SugarElement.fromText('Caption'));
             Insert.appendAt(table, caption, 0);
+            editor.selection.setCursorLocation(caption.dom, 0);
           },
           (caption) => {
             if (SugarNode.isTag('caption')(cellOrCaption)) {
--- a/modules/tinymce/src/plugins/table/main/ts/api/Commands.ts
+++ b/modules/tinymce/src/plugins/table/main/ts/api/Commands.ts
@@ -93,7 +93,7 @@ const registerCommands = (editor: Editor, actions: TableActions, cellSelection: 
 
   const toggleCaption = () => {
     getSelectionStartCellOrCaption(editor).each((cellOrCaption) => {
-      TableLookup.table(cellOrCaption, isRoot).filter(Fun.not(isRoot)).each((table) => {
+      TableLookup.table(cellOrCaption, isRoot).each((table) => {
         const captionElementOpt = Selectors.one('caption', table);
 
         captionElementOpt.fold(
@@ -102,9 +102,7 @@ const registerCommands = (editor: Editor, actions: TableActions, cellSelection: 
             Insert.append(caption, SugarElement.fromText('Caption'));
             Insert.appendAt(table, caption, 0);
           },
-          (caption) => {
-            Remove.remove(caption);
-          }
+          Remove.remove
         );
 
         Events.fireTableModified(editor, table.dom, Events.structureModified);
--- a/modules/tinymce/src/plugins/table/main/ts/api/Commands.ts
+++ b/modules/tinymce/src/plugins/table/main/ts/api/Commands.ts
@@ -8,7 +8,7 @@
 import { Selections } from '@ephox/darwin';
 import { Arr, Fun, Obj, Optional, Type } from '@ephox/katamari';
 import { CopyCols, CopyRows, Sizes, TableFill, TableLookup } from '@ephox/snooker';
-import { Insert, Remove, Replication, SelectorFind, Selectors, SugarElement } from '@ephox/sugar';
+import { Insert, Remove, Replication, SelectorFind, Selectors, SugarElement, SugarNode } from '@ephox/sugar';
 import Editor from 'tinymce/core/api/Editor';
 import { enforceNone, enforcePercentage, enforcePixels } from '../actions/EnforceUnit';
 import { insertTableWithDataValidation } from '../actions/InsertTable';
@@ -101,10 +101,12 @@ const registerCommands = (editor: Editor, actions: TableActions, cellSelection: 
             Insert.appendAt(table, caption, 0);
           },
           (caption) => {
+            if (SugarNode.isTag('caption')(cellOrCaption)) {
+              Selectors.one('td', table).each((td) =>
+                editor.selection.setCursorLocation(td.dom, 0)
+              );
+            }
             Remove.remove(caption);
-            Selectors.one('td', table).each((td) =>
-              editor.selection.setCursorLocation(td.dom, 0)
-            );
           }
         );
 
--- a/modules/tinymce/src/plugins/table/main/ts/selection/SelectionTargets.ts
+++ b/modules/tinymce/src/plugins/table/main/ts/selection/SelectionTargets.ts
@@ -6,9 +6,9 @@
  */
 
 import { Selections } from '@ephox/darwin';
-import { Arr, Cell, Fun, Optional, Thunk, Type } from '@ephox/katamari';
+import { Arr, Cell, Fun, Optional, Thunk } from '@ephox/katamari';
 import { RunOperation, Structs, TableLookup, Warehouse } from '@ephox/snooker';
-import { SugarElement, SugarNode } from '@ephox/sugar';
+import { SelectorExists, SugarElement, SugarNode } from '@ephox/sugar';
 import Editor from 'tinymce/core/api/Editor';
 import { Menu, Toolbar } from 'tinymce/core/api/ui/Ui';
 import * as Util from '../core/Util';
@@ -16,6 +16,7 @@ import * as TableTargets from '../queries/TableTargets';
 import * as TableSelection from './TableSelection';
 
 type UiApi = Menu.MenuItemInstanceApi | Toolbar.ToolbarButtonInstanceApi;
+type UiToggleApi = Menu.ToggleMenuItemInstanceApi | Toolbar.ToolbarToggleButtonInstanceApi;
 
 /*
 onAny - disable if any column in the selection is locked
@@ -38,7 +39,7 @@ export interface SelectionTargets {
   readonly onSetupMergeable: (api: UiApi) => () => void;
   readonly onSetupUnmergeable: (api: UiApi) => () => void;
   readonly resetTargets: () => void;
-  readonly onSetupTableWithCaption: (api: Toolbar.ToolbarToggleButtonInstanceApi) => () => void;
+  readonly onSetupTableWithCaption: (api: UiToggleApi) => () => void;
   readonly targets: () => Optional<RunOperation.CombinedTargets>;
 }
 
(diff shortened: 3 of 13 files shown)


modules/tinymce/src/plugins/table/main/ts/api/Commands.ts
```diff
@@ -91,6 +91,25 @@ const registerCommands = (editor: Editor, actions: TableActions, cellSelection:
     });
   };
 
+  const toggleCaption = () => {
+    getSelectionStartCellOrCaption(editor).each((cellOrCaption) => {
+      TableLookup.table(cellOrCaption, isRoot).each((table) => {
+        const captionElementOpt = Selectors.one('caption', table);
+
+        captionElementOpt.fold(
```
lnewson, line 99:
```suggestion
        Selectors.one('caption', table).fold(
```

modules/tinymce/src/plugins/table/main/ts/api/Commands.ts
```diff
@@ -91,6 +91,25 @@ const registerCommands = (editor: Editor, actions: TableActions, cellSelection:
...
+          () => {
+            const caption = SugarElement.fromTag('caption');
+            Insert.append(caption, SugarElement.fromText('Caption'));
+            Insert.appendAt(table, caption, 0);
```
lnewson, line 103: This is inconsistent behaviour with how the dialog works, as it does not insert the "Caption" text. It should be the same and really the dialog should be using this command.

- [ ] Verify with design if the "Caption" text should be getting inserted
- [ ] Make the table dialog use the command so all behaviour is consitent

modules/tinymce/src/plugins/table/main/ts/selection/SelectionTargets.ts
```diff
@@ -134,7 +155,17 @@ export const getSelectionTargets = (editor: Editor, selections: Selections): Sel
...
   const onSetupMergeable = (api: UiApi) => onSetup(api, (_targets) => isDisabledForSelection('mergeable'));
   const onSetupUnmergeable = (api: UiApi) => onSetup(api, (_targets) => isDisabledForSelection('unmergeable'));
-
+  const onSetupTableWithCaption = (api: Toolbar.ToolbarToggleButtonInstanceApi) => {
+    return onSetupWithToggle(api, Fun.never, (targets) => {
+      const potentialTables = TableLookup.table(targets.element, Util.getIsRoot(editor));
+      const tableOpt = potentialTables.filter(Fun.not(Util.getIsRoot(editor)));
+
+      return tableOpt.map((table) => {
+        const captionElm = editor.dom.select('caption', table.dom)[0];
```
lnewson, line 164: Please use Sugar here, not DOMUtils as we're using Sugar everywhere else. eg:
```
return tableOpt.exists((table) => SelectorExists.child(table, 'caption'));
```

- [ ] Convert to using sugar functions

modules/tinymce/src/plugins/table/main/ts/selection/SelectionTargets.ts
```diff
@@ -134,7 +155,17 @@ export const getSelectionTargets = (editor: Editor, selections: Selections): Sel
     onSetup(api, (targets) => isCaption(targets.element) || getClipboardData().isNone() || isDisabledFromLocked(lockedDisable));
   const onSetupMergeable = (api: UiApi) => onSetup(api, (_targets) => isDisabledForSelection('mergeable'));
   const onSetupUnmergeable = (api: UiApi) => onSetup(api, (_targets) => isDisabledForSelection('unmergeable'));
-
+  const onSetupTableWithCaption = (api: Toolbar.ToolbarToggleButtonInstanceApi) => {
+    return onSetupWithToggle(api, Fun.never, (targets) => {
+      const potentialTables = TableLookup.table(targets.element, Util.getIsRoot(editor));
```
lnewson, line 160: This will only ever return a single table, so this shouldn't be a plural. Additionally, the filter is not needed below, so please remove that as well.

modules/tinymce/src/plugins/table/main/ts/selection/SelectionTargets.ts
```diff
@@ -134,7 +155,17 @@ export const getSelectionTargets = (editor: Editor, selections: Selections): Sel
     onSetup(api, (targets) => isCaption(targets.element) || getClipboardData().isNone() || isDisabledFromLocked(lockedDisable));
   const onSetupMergeable = (api: UiApi) => onSetup(api, (_targets) => isDisabledForSelection('mergeable'));
   const onSetupUnmergeable = (api: UiApi) => onSetup(api, (_targets) => isDisabledForSelection('unmergeable'));
-
+  const onSetupTableWithCaption = (api: Toolbar.ToolbarToggleButtonInstanceApi) => {
```
lnewson, line 158: This needs to work for both menu items and toolbar buttons, so please create a new `UiToggleApi` and use that where relevant. eg:
```
type UiApi = Menu.MenuItemInstanceApi | Toolbar.ToolbarButtonInstanceApi;
type UiToggleApi = Menu.ToggleMenuItemInstanceApi | Toolbar.ToolbarToggleButtonInstanceApi;
```

- [ ] Fix incorrect types

modules/tinymce/src/plugins/table/main/ts/selection/SelectionTargets.ts
```diff
@@ -134,7 +155,17 @@ export const getSelectionTargets = (editor: Editor, selections: Selections): Sel
...
+    return onSetupWithToggle(api, Fun.never, (targets) => {
+      const potentialTables = TableLookup.table(targets.element, Util.getIsRoot(editor));
+      const tableOpt = potentialTables.filter(Fun.not(Util.getIsRoot(editor)));
+
+      return tableOpt.map((table) => {
+        const captionElm = editor.dom.select('caption', table.dom)[0];
+        return Type.isNonNullable(captionElm);
+      }).getOr(false);
+    });
+  };
```
lnewson, line 168: This can also be shortened to keep with the existing code style used here. eg:
```suggestion
  const onSetupTableWithCaption = (api: UiToggleApi) => onSetupWithToggle(api, Fun.never, (targets) => {
    const tableOpt = TableLookup.table(targets.element, Util.getIsRoot(editor));
    return tableOpt.exists((table) => SelectorExists.child(table, 'caption'));
  });
```

modules/tinymce/src/plugins/table/main/ts/ui/Buttons.ts
```diff
@@ -239,6 +239,15 @@ const addButtons = (editor: Editor, selectionTargets: SelectionTargets, clipboar
     ),
     onSetup: selectionTargets.onSetupCellOrRow
   });
+
+  editor.ui.registry.addToggleButton('tablecaption', {
+    tooltip: 'Table caption',
+    onAction: () => {
+      editor.execCommand('mceTableToggleCaption');
+    },
```
lnewson, line 247: There's already a helper here to do this, so please reuse it.
```suggestion
    onAction: cmd('mceTableToggleCaption'),
```

modules/tinymce/src/plugins/table/main/ts/ui/MenuItems.ts
```diff
@@ -280,6 +280,14 @@ const addMenuItems = (editor: Editor, selectionTargets: SelectionTargets, clipbo
     ),
     onSetup: selectionTargets.onSetupCellOrRow
   });
+
+  editor.ui.registry.addMenuItem('tablecaption', {
+    onAction: () => {
+      editor.execCommand('mceTableToggleCaption');
+    },
```
lnewson, line 287:
```suggestion
    onAction: cmd('mceTableToggleCaption'),
```

modules/tinymce/src/plugins/table/main/ts/ui/MenuItems.ts
```diff
@@ -280,6 +280,14 @@ const addMenuItems = (editor: Editor, selectionTargets: SelectionTargets, clipbo
     ),
     onSetup: selectionTargets.onSetupCellOrRow
   });
+
+  editor.ui.registry.addMenuItem('tablecaption', {
```
lnewson, line 284: This is missing the `text` property and looking back we missed this in all your other PRs as well :facepalm: Please fix this and the other menu items (feel free to do the others in another PR). Here's what it currently looks like, so it also shows this likely wasn't tested:
image

```suggestion
  editor.ui.registry.addToggleMenuItem('tablecaption', {
    text: 'Table caption',
```

- [ ] Fix missing `text` field on menu items
- [ ] Ensure tests exist for this menu item

lnewson: Additionally, this needs to be a toggle menu item otherwise the toggling won't work.

- [ ] Fix incorrect registration of toggle menu item

modules/tinymce/src/plugins/table/test/ts/browser/ui/TableCaptionTest.ts
```diff
@@ -0,0 +1,97 @@
...
+          '<td>Filler</td>' +
+        '</tr>' +
+      '</tbody>' +
+    '</table>'
+  );
+
+  const toggleCaption = (editor: Editor) => {
+    editor.execCommand('mceTableToggleCaption');
+
+    assert.lengthOf(events, 1, 'Command executed successfully.');
```
lnewson, line 45:
```suggestion
    assert.lengthOf(events, 1, 'Command executed successfully');
```

modules/tinymce/src/plugins/table/test/ts/browser/ui/TableCaptionTest.ts
```diff
@@ -0,0 +1,97 @@
+import { afterEach, context, describe, it } from '@ephox/bedrock-client';
+import { TinyAssertions, TinyHooks, TinySelections } from '@ephox/mcagar';
+import { assert } from 'chai';
+import Editor from 'tinymce/core/api/Editor';
```
lnewson, line 4:
```suggestion
import { assert } from 'chai';

import Editor from 'tinymce/core/api/Editor';
```

modules/tinymce/src/plugins/table/test/ts/browser/ui/TableCaptionTest.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  context('Test when using the command directly', () => {
+    it('TINY-7476: The caption should appear', () => {
+      const editor = hook.editor();
+
+      setEditorContentTableAndSelection(editor, 1, 1);
+      toggleCaption(editor);
+
+      assertTableContainsCaption(editor);
+    });
```
lnewson, line 66: As I mentioned in a previous PR, this isn't consistent with how tests are normally written. It's normally setup and then do the action/assert the result. So this should be something like this instead:
```suggestion
    it('TINY-7476: The caption should appear', () => {
      const editor = hook.editor();
      setEditorContentTableAndSelection(editor, 1, 1);

      toggleCaption(editor);
      assertTableContainsCaption(editor);
    });
```

- [ ] Fix formatting

modules/tinymce/src/plugins/table/test/ts/browser/ui/TableCaptionTest.ts
```diff
@@ -0,0 +1,97 @@
...
+      const editor = hook.editor();
+
+      setTableCaptionStructureAndSelection(editor);
+      toggleCaption(editor);
+
+      assertStructureIsRestoredToDefault(editor, 1, 1);
+    });
+  });
+
+  context('Test when using the button', () => {
```
lnewson, line 78: Using `Test` in a test name is a little weird here... could this just be `Using the toolbar button` or something similar instead?

modules/tinymce/src/plugins/table/test/ts/browser/ui/TableCaptionTest.ts
```diff
@@ -0,0 +1,97 @@
...
+    TinyAssertions.assertContent(editor, tableCaptionStructure);
+  };
+
+  const setTableCaptionStructureAndSelection = (editor: Editor) => {
+    editor.setContent(tableCaptionStructure);
+
+    TinySelections.setSelection(editor, [ 0, 1, 0, 0 ], 0, [ 0, 1, 0, 0 ], 1);
+  };
+
+  context('Test when using the command directly', () => {
```
lnewson, line 58: Likewise here, can we come up with something else that doesn't include `Test` please?

lnewson (changes requested):
Overall looks good, however there are still some core things that need work here so flagging as needs changes sorry.

modules/tinymce/src/plugins/table/main/ts/selection/SelectionTargets.ts
```diff
@@ -122,6 +123,26 @@ export const getSelectionTargets = (editor: Editor, selections: Selections): Sel
     };
   };
 
+  const onSetupWithToggle = (api: Toolbar.ToolbarToggleButtonInstanceApi, isDisabled: (targets: RunOperation.CombinedTargets) => boolean, isActive: (targets: RunOperation.CombinedTargets) => boolean) => {
```
lnewson, line 126: Forgot to address this sorry... this duplicates ~50% of the `onSetup` function so please restructure this to remove the duplication.

- [ ] Fix duplication

modules/tinymce/src/plugins/table/main/ts/selection/SelectionTargets.ts
lnewson: I should add this also uses `exists` as I mentioned in the call last night, as if you see a `getOr(false)` it should pretty much always be using `exists` instead.

modules/tinymce/src/plugins/table/main/ts/ui/MenuItems.ts
metricjs: If there's any new strings for the `text` properties, we'll need to make a note for translations too. I don't know if we have strings for these things yet, so thought I'd mention.

lnewson: There should not be as they are already being used in the tooltips for buttons. The only case that should apply to is if we added new strings for both.

modules/tinymce/src/plugins/table/main/ts/api/Commands.ts
lostkeys: The word "Caption" should be inserted. Ideally that shows up when using the dialog as well. This is to make it consistent with image captions, and also to show the user where the table captions are 👀.

HAFRMO: It should definitively come with the "Caption"-text as there is an issue to add that...

modules/tinymce/src/plugins/table/test/ts/browser/ui/TableCaptionTest.ts
```diff
@@ -0,0 +1,116 @@
...
+    toolbar: 'tablecaption',
+    base_url: '/project/tinymce/js/tinymce',
+    setup: (editor: Editor) => {
+      editor.on('tablemodified', logEvent);
+    },
+    menu: {
+      table: {
+        title: 'Table',
+        items: 'tablecaption'
+      },
```
lnewson, line 25:
```suggestion
      table: { title: 'Table', items: 'tablecaption' },
```

modules/tinymce/src/plugins/table/test/ts/browser/ui/TableClassListButtonsTest.ts
```diff
@@ -14,6 +14,15 @@ describe('browser.tinymce.plugins.table.ui.TableClassListButtonsTest', () => {
   const hook = TinyHooks.bddSetupLight<Editor>({
     plugins: 'table',
     indent: false,
+    toolbar: 'tableclass',
+    menu: {
+      table: {
+        title: 'Table',
+        items: 'tableclass'
+      },
```
lnewson, line 22:
```suggestion
      table: { title: 'Table', items: 'tableclass' },
```

modules/tinymce/src/plugins/table/test/ts/browser/ui/TableCaptionTest.ts
```diff
@@ -0,0 +1,116 @@
...
+    it('TINY-7476: The caption should be removed', () => {
+      const editor = hook.editor();
+      setTableCaptionStructureAndSelection(editor);
+
+      toggleCaption(editor);
+      assertStructureIsRestoredToDefault(editor, 1, 1);
+    });
+  });
+
+  context('Using the button', () => {
```
lnewson, line 100: This is using both the menu and toolbar button. Please break this up into 2 separate context blocks, as "using the button" is incorrect when it does both.

- [ ] Break this up into a different context for the toolbar button and menu item

modules/tinymce/src/plugins/table/test/ts/browser/ui/TableValignButtonsTest.ts
```diff
@@ -10,6 +10,13 @@ describe('browser.tinymce.plugins.table.ui.TableValignButtonsTest', () => {
     plugins: 'table',
     indent: false,
     toolbar: 'tablecellvalign',
+    menu: {
+      table: {
+        title: 'Table',
+        items: 'tablecellvalign'
+      },
```
lnewson, line 17:
```suggestion
      table: { title: 'Table', items: 'tablecellvalign' },
```

modules/tinymce/src/plugins/table/test/ts/module/test/TableModifiersTestUtils.ts
```diff
@@ -78,64 +78,80 @@ const assertStructureHasCustomStyle = (editor: Editor, rows: number, columns: nu
...
 
-const closeMenu = (editor) =>
+const closeMenu = (editor) => {
   TinyUiActions.keydown(editor, Keys.escape());
+  TinyUiActions.keydown(editor, Keys.escape());
+};
 
-const openMenu = (editor: Editor, title: string) =>
-  TinyUiActions.clickOnToolbar(editor, `button[title="${title}"]`);
+const clickOnButton = async (editor: Editor, title: string, toolbar: boolean) => {
```
lnewson, line 86: Please make this 2 separate functions instead of trying to overload it like this. It makes it hard to read/understand what's happening with just a bunch of booleans. It's also confusing, as click on button has nothing to do with menu items.

- [ ] Break this up into a `clickOnButton` and `clickOnMenuItem` or similar functions

modules/tinymce/src/plugins/table/test/ts/module/test/TableModifiersTestUtils.ts
```diff
@@ -78,64 +78,80 @@ const assertStructureHasCustomStyle = (editor: Editor, rows: number, columns: nu
...
+    TinyUiActions.clickOnMenu(editor, 'span:contains("Table")');
+    await TinyUiActions.pWaitForUi(editor, `div[title="${title}"]`);
+    TinyUiActions.clickOnUi(editor, `div[title="${title}"]`);
+  }
+};
 
-const pAssertMenuPresence = async (editor: Editor, label: string, menuTitle: string, expected: Record<string, number>, container: SugarElement<HTMLElement>) => {
-  openMenu(editor, menuTitle);
+const pAssertMenuPresence = async (editor: Editor, label: string, menuTitle: string, expected: Record<string, number>, container: SugarElement<HTMLElement>, toolbar: boolean) => {
+  await clickOnButton(editor, menuTitle, toolbar);
```
lnewson, line 97: As this is being overloaded and it really shouldn't be, you'll need to either pass in an open function or just do it separately before calling this.

- [ ] Don't overload functions like this please, keep it specific to a single action

modules/tinymce/src/plugins/table/test/ts/browser/ui/TableCaptionTest.ts
```diff
@@ -0,0 +1,116 @@
...
+    TinyAssertions.assertContent(editor, tableCaptionStructure);
+  };
+
+  const setTableCaptionStructureAndSelection = (editor: Editor) => {
+    editor.setContent(tableCaptionStructure);
+
+    TinySelections.setSelection(editor, [ 0, 1, 0, 0 ], 0, [ 0, 1, 0, 0 ], 1);
+  };
+
+  const pAssertTableButton = async (toolbar: boolean, addCaption: boolean) => {
```
lnewson, line 66: I've said this in a few places, but please use separate functions and abstract duplicate logic instead of creating overloaded functions like this.

lnewson (changes requested):
Looks a lot better thanks! However, the test function overloading is a design pattern we don't like to use as it makes for messy and hard to read code. It's better to create separate functions and abstract the duplicate code, so marking as needs changes sorry.

modules/tinymce/src/plugins/table/main/ts/ui/MenuItems.ts
```diff
@@ -256,8 +259,9 @@ const addMenuItems = (editor: Editor, selectionTargets: SelectionTargets, clipbo
   });
 
   const tableCellBorderWidthsList = getTableBorderWidths(editor);
-  editor.ui.registry.addNestedMenuItem('tablecellvalign', {
+  editor.ui.registry.addNestedMenuItem('tablecellborderwidth', {
```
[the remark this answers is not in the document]
HAFRMO: Yeah. Let's fix them here, and pretend it was never wrong.

modules/tinymce/src/plugins/table/main/ts/api/Commands.ts
```diff
@@ -91,6 +91,23 @@ const registerCommands = (editor: Editor, actions: TableActions, cellSelection:
     });
   };
 
+  const toggleCaption = () => {
+    getSelectionStartCellOrCaption(editor).each((cellOrCaption) => {
+      TableLookup.table(cellOrCaption, isRoot).each((table) => {
+        Selectors.one('caption', table).fold(
```
lnewson, line 97: I just realised this won't work as expected because it looks all the way down the dom tree. So if you have a tested table it'll find the caption in that potentially. As such this should be using `SelectorFind.child` instead which will only look at the direct children.

- [ ] Fix this so it won't find captions in nested tables
- [ ] Consider adding a test for this case please.

modules/tinymce/src/plugins/table/test/ts/module/test/TableModifiersTestUtils.ts
```diff
@@ -78,64 +78,89 @@ const assertStructureHasCustomStyle = (editor: Editor, rows: number, columns: nu
...
+};
+
 export {
   pAssertStyleCanBeToggledOnAndOff,
   setEditorContentTableAndSelection,
   pAssertNoCheckmarksInMenu,
-  pAssertMenuPresence
+  pAssertMenuPresence,
+  clickOnButton,
+  pClickOnMenuItem as clickOnMenuItem,
```
lnewson, line 164: This wasn't entirely correctly renamed. It was only renamed internally, so please make sure to remove the alias and rename it outside this module as well.

lnewson (approved):

modules/tinymce/src/plugins/table/test/ts/module/test/TableModifiersTestUtils.ts
HAFRMO: Dammit.

modules/tinymce/CHANGELOG.md
```diff
@@ -9,17 +9,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Added a new `toolbar_sticky_offset` setting to allow the toolbar to be offset from the top of the page #TINY-7337
 - Added a new `mceFocus` command that focuses the editor. Equivalent to using `editor.focus()` #TINY-7373
+- Added a new `mceTableToggleClass` command which toggles the provided class on the currently selected table #TINY-7476
+- Added a new `mceTableCellToggleClass` command which toggles the provided class on the currently selected table cells #TINY-7476
 - Added a new `tablecellvalign` toolbar button and menu item for vertical table cell alignment #TINY-7477
 - Added a new `tablecellborderwidth` toolbar button and menu item to change table cell border width #TINY-7478
 - Added a new `tablecellborderstyle` toolbar button and menu item to change table cell border style #TINY-7478
+- Added a new `tablecaption` toolbar button and menu item to toggle captions on tables #TINY-7479
```
metricjs, line 17: Is there a new command for this too, or did that already exist?

modules/tinymce/src/plugins/table/main/ts/ui/MenuItems.ts
```diff
@@ -256,8 +259,9 @@ const addMenuItems = (editor: Editor, selectionTargets: SelectionTargets, clipbo
...
     icon: 'vertical-align',
```
metricjs, line 263: The icon is still `vertical-align`? Here and the `tablecellborderstyle` one too

lnewson: - [ ] Fix incorrect icons

modules/tinymce/src/plugins/table/test/ts/browser/ui/TableClassListButtonsTest.ts
```diff
@@ -48,36 +52,35 @@ describe('browser.tinymce.plugins.table.ui.TableClassListButtonsTest', () => {
...
     };
-    await pAssertMenuPresence(editor, 'There should be a checkmark', 'Table styles', expected, sugarContainer);
-  });
+    await pAssertMenuPresence(editor, `There should be ${trueMarked} checkmark(s)`, 'Table styles', expected, sugarContainer, toolbar);
+  };
 
-  it('TINY-7476: Ensure that the checkmark appears for two classes', async () => {
-    const editor = hook.editor();
-    const sugarContainer = SugarBody.body();
+  it('TINY-7476: Ensure that the checkmark appears for a single class', async () => {
```
metricjs, line 71:
```suggestion
  it('TINY-7476: Ensure that the checkmark appears for a single class in toolbar', async () => {
```
just to be super clear? I had to track back the boolean to figure this out, so

modules/tinymce/src/plugins/table/test/ts/browser/ui/TableClassListButtonsTest.ts
```diff
@@ -48,36 +52,35 @@ describe('browser.tinymce.plugins.table.ui.TableClassListButtonsTest', () => {
...
+  it('TINY-7476: Ensure that the checkmark appears for a single class in menu', async () => {
+    await pPerformToggleClassSetup([ 'a' ], 1, 1, false);
+  });
 
-    const expected = {
-      '.tox-menu': 1,
-      '.tox-collection__item[aria-checked="true"]': 2,
-      '.tox-collection__item[aria-checked="false"]': 0
-    };
+  it('TINY-7476: Ensure that the checkmark appears for two classes', async () => {
```
metricjs, line 79:
```suggestion
  it('TINY-7476: Ensure that the checkmark appears for two classes in toolbar', async () => {
```
as above

modules/tinymce/CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added a new `tablecellborderwidth` toolbar button and menu item to change table cell border width #TINY-7478
 - Added a new `tablecellborderstyle` toolbar button and menu item to change table cell border style #TINY-7478
 - Added a new `tablecaption` toolbar button and menu item to toggle captions on tables #TINY-7479
+- Added a new `mceTableToggleCaption` command that toggles caption on selected table #TINY-7479
```
metricjs, line 18:
```suggestion
- Added a new `mceTableToggleCaption` command that toggles captions on selected table #TINY-7479
```
to match the line above

modules/tinymce/src/plugins/table/main/ts/api/Commands.ts
```diff
@@ -91,6 +91,23 @@ const registerCommands = (editor: Editor, actions: TableActions, cellSelection:
...
+  const toggleCaption = () => {
+    getSelectionStartCellOrCaption(editor).each((cellOrCaption) => {
+      TableLookup.table(cellOrCaption, isRoot).each((table) => {
+        SelectorFind.child(table, 'caption').fold(
+          () => {
+            const caption = SugarElement.fromTag('caption');
+            Insert.append(caption, SugarElement.fromText('Caption'));
+            Insert.appendAt(table, caption, 0);
+          },
+          Remove.remove
```
metricjs, line 103: Discussed elsewhere but reminder: cursor/selection goes somewhere weird after this because we've removed the element it was in, so we should purposefully put it somewhere. I reckon start of the first cell works.

metricjs: - [ ] fix

lnewson: I suspect this was actually a bug with the dialog before, as it also deleted without updating the selection :thinking: One thing to be careful of here, is that we only should move the selection if it was within the caption.

modules/tinymce/CHANGELOG.md
lnewson:
```suggestion
- Added a new `mceTableToggleCaption` command that toggles captions on a selected table #TINY-7479
```
Maybe? It just reads a little weirdly otherwise.

metricjs (approved):
Approving with the things I've previously commented on

modules/tinymce/src/plugins/table/test/ts/browser/ui/TableCaptionTest.ts
```diff
@@ -0,0 +1,178 @@
...
+    TinyAssertions.assertContent(editor, tableStructure);
+  };
+
+  const setTableCaptionStructureAndSelection = (editor: Editor, structure: string, indexOftbody: number) => {
+    editor.setContent(structure);
+
+    TinySelections.setSelection(editor, [ 0, indexOftbody, 0, 0 ], 0, [ 0, indexOftbody, 0, 0 ], 1);
+  };
+
+  const pAssertTableCaption = async (toolbar: boolean, addCaption: boolean) => {
```
JamesToohey, line 102: Rather than using a boolean for toolbar, could we do something like this

```suggestion
  const pAssertTableCaption = async (action: 'toolbar' | 'menuitem', addCaption: boolean) => {
```

so that it's a bit easier to follow in the tests?

modules/tinymce/src/plugins/table/test/ts/module/test/TableModifiersTestUtils.ts
```diff
@@ -78,64 +78,89 @@ const assertStructureHasCustomStyle = (editor: Editor, rows: number, columns: nu
     '</table>');
 };
 
-const closeMenu = (editor) =>
+const closeMenu = (editor) => {
```
JamesToohey, line 81: Can we add the type to this while we're in here

modules/tinymce/src/plugins/table/test/ts/module/test/TableModifiersTestUtils.ts
```diff
@@ -78,64 +78,89 @@ const assertStructureHasCustomStyle = (editor: Editor, rows: number, columns: nu
...
   TinyUiActions.keydown(editor, Keys.escape());
+  TinyUiActions.keydown(editor, Keys.escape());
```
JamesToohey, line 83: Is this needed?

JamesToohey (approved):

modules/tinymce/src/plugins/table/test/ts/module/test/TableModifiersTestUtils.ts
HAFRMO: Yes, when you open the menuitems it requires an extra escape.

modules/tinymce/src/plugins/table/test/ts/browser/ui/TableCaptionTest.ts
HAFRMO: Not sure it'll do much good since one of the values will never be seen in the actual code, but sure.

PR #6916, opened 2021-06-10
TINY-6297: Push editor rendering on focus into another event loop

### LINKED ISSUE #5947: I can't find what I'm looking for: general-configuration-guide/multiple-editors.md
When I create multiple inline Editor instances then if I perform any editor action on text field like making characters bold or italic then the editor corresponding button will get enabled which is obvious and correct behaviour. But after it when I put my cursor on next text field then still the button in editor **remains enabled**. After inspecting it I found that class "tox-tbtn--enabled" is present there which should not be present.

I am also attaching the image of the issue. Here BOLD button **should not be enabled**. Also, all fields toolbar is attached to a div above all text fields as shown in image

Screenshot from 2020-08-06 18-00-44

I have done some more investigation and found that first toolbar creation is perfect but when others get initialised they copy properties like classes and all from the first toolbar. That's why the button gets enabled. I am facing this issue in our product.

I have just shown a demo here. I will really appreciate if I get a solution for it. Thanks in advance.

JamesToohey:
Related Ticket: TINY-6297

Description of Changes:
This fixes certain operations performing on a selection that may not be valid yet (e.g. the selection could be in another editor on the same page).

The bug reported was when two inline editors are on the same page, the toolbar state does not update correctly when you move your selection to another editor.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

GitHub issues (if applicable): Fixes #5947

### DIFF
--- a/modules/tinymce/src/themes/silver/main/ts/modes/Inline.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/modes/Inline.ts
@@ -126,7 +126,7 @@ const render = (editor: Editor, uiComponents: RenderUiComponents, rawUiConfig: R
 
   editor.on('init', () => {
     if (editor.hasFocus() || toolbarPersist) {
-      render();
+      delayedRender();
     }
   });
 
--- a/modules/tinymce/src/themes/silver/main/ts/modes/Inline.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/modes/Inline.ts
@@ -112,11 +112,15 @@ const render = (editor: Editor, uiComponents: RenderUiComponents, rawUiConfig: R
     editor.nodeChanged();
   };
 
+  // In certain circumstances we need to delay the render function until the next
+  // event loop to ensure things like the current selection are correct.
+  const delayedRender = () => Delay.setEditorTimeout(editor, render, 0);
+
   editor.on('show', render);
   editor.on('hide', ui.hide);
 
   if (!toolbarPersist) {
-    editor.on('focus', render);
+    editor.on('focus', delayedRender);
     editor.on('blur', ui.hide);
   }
 
--- a/modules/tinymce/src/themes/silver/test/ts/browser/editor/ShowHideTest.ts
+++ b/modules/tinymce/src/themes/silver/test/ts/browser/editor/ShowHideTest.ts
@@ -6,7 +6,7 @@ import { SugarBody } from '@ephox/sugar';
 import Editor from 'tinymce/core/api/Editor';
 import Theme from 'tinymce/themes/silver/Theme';
 
-import * as TestUtils from '../../module/TestUtils';
+import * as TestUtils from '../../module/UiUtils';
 
 describe('browser.tinymce.themes.silver.editor.ShowHideTest', () => {
   const base_url = '/project/tinymce/js/tinymce';
@@ -37,7 +37,7 @@ describe('browser.tinymce.themes.silver.editor.ShowHideTest', () => {
     ]
   });
 
-  const pFocusEditor = async (editor: Editor) => {
+  const pActivateEditor = async (editor: Editor) => {
     editor.focus();
     await TestUtils.pWaitForEditorToRender();
   };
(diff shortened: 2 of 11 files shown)


modules/tinymce/src/themes/silver/main/ts/modes/Inline.ts
```diff
@@ -112,11 +112,15 @@ const render = (editor: Editor, uiComponents: RenderUiComponents, rawUiConfig: R
...
+  // In certain circumstances we need to delay the render function until the next
+  // event loop to ensure things like the current selection are correct.
+  const delayedRender = () => Delay.setEditorTimeout(editor, render, 0);
+
   editor.on('show', render);
   editor.on('hide', ui.hide);
 
   if (!toolbarPersist) {
-    editor.on('focus', render);
+    editor.on('focus', delayedRender);
```
lnewson, line 123: This needs changing on the init/hasFocus call below as well.

modules/tinymce/src/themes/silver/test/ts/browser/editor/ShowHideTest.ts
```diff
@@ -35,10 +37,15 @@ describe('browser.tinymce.themes.silver.editor.ShowHideTest', () => {
     ]
   });
 
+  const pFocusEditor = async (editor: Editor) => {
```
lnewson, line 40: This name is very misleading, as it's doing much more than just focusing the editor. Maybe `pActivateEditor`, `pShowToolbar` or something like that?

modules/tinymce/src/themes/silver/test/ts/module/TestUtils.ts
```diff
@@ -0,0 +1,9 @@
+import { UiFinder, Waiter } from '@ephox/agar';
+import { SugarBody } from '@ephox/sugar';
+
+const pWaitForEditorToRender = () =>
```
lnewson, line 4: This would make more sense in `UiUtils` so can we please move it there?

modules/tinymce/src/themes/silver/test/ts/webdriver/editor/SimpleControlsInlineTest.ts
```diff
@@ -0,0 +1,32 @@
+import { RealMouse, UiFinder, Waiter } from '@ephox/agar';
+import { before, describe, it } from '@ephox/bedrock-client';
+import { McEditor } from '@ephox/mcagar';
+import { SugarBody } from '@ephox/sugar';
+
+import Editor from 'tinymce/core/api/Editor';
+import Theme from 'tinymce/themes/silver/Theme';
+
+describe('browser.tinymce.themes.silver.editor.core.SimpleControlsInlineTest', () => {
+  before(() => Theme());
```
lnewson, line 10:
```suggestion
  before(() => {
    Theme()
  });
```
We should never use shorthand for hooks, as the return value from the function executed will be passed to the hook.

modules/tinymce/src/themes/silver/test/ts/webdriver/editor/SimpleControlsInlineTest.ts
```diff
@@ -0,0 +1,32 @@
...
+  const pAssertToolbarButtonPressed = (title: string, pressed: boolean = true) =>
+    Waiter.pTryUntil('button pressed', () => UiFinder.exists(SugarBody.body(), `button[title="${title}"][aria-pressed="${pressed ? 'true' : 'false'}"]`));
+
+  it('TINY-6675: Button state on multiple inline editors is correct', async () => {
+    const editorOne = await McEditor.pFromSettings<Editor>(settings);
+    const editorTwo = await McEditor.pFromSettings<Editor>(settings);
+    editorOne.setContent('<p id="number1"><strong>blarg</strong></p>');
+    editorTwo.setContent('<p id="number2">blarg</p>');
+    await RealMouse.pClickOn('#number1');
+    await pAssertToolbarButtonPressed('Bold');
```
lnewson, line 26: I find this confusing and inconsistent, so can you please remove the default on `pressed` and make it explicit here?

modules/tinymce/src/themes/silver/test/ts/webdriver/throbber/ThrobberTabbingTest.ts
```diff
@@ -56,17 +58,22 @@ describe('webdriver.tinymce.themes.silver.throbber.ThrobberTabbingTest', () => {
     await pAssertFocus();
   };
 
+  const pSetFocus = async (selector: string) => {
```
lnewson, line 61: Again, this is confusing/misleading as it's doing more than just focusing Can we please come up with another name?

lnewson (approved):

ltrouton (approved):

PR #6923, opened 2021-06-11
Optimise the Optional and Result types to use classes and therefore be easier to construct

ashfordneil:
Related Ticket: N/A
 
Description of Changes:
* **We have COMMENTS!!**
* They're classes now
* .tag and .value (or .item for Result) exist as private properties that you can't use in code but you can:
  * see them when logging the objects
  * view them in the debugger
  * use them to make chai assertions play nicely with Optional and Result types
* One single test had issues when I made this conversion (it was trying to stringily a cyclic structure, but because stringifying Optionals doesn't really work it wasn't breaking)
* Deep merge had to be updated to skip over classes so that it didn't mess with the prototype of the Optional / Result (or any other class going through)
  * This involved adding a `Type.isPlainObject` function to tell the difference between classes and normal json

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


ltrouton (approved):

tiny-james (approved):
Looks good. My only worry is that the HCL Connections minifier will choke on it as it has had problems with classes in the past. I haven't checked in the last couple of years though so they might have fixed that bug and/or it might have been something specific about that class. I guess the only way we're going to find out is to try again.

lnewson:
I'm converting this to a draft, as it has failing tests and has been stagnant for a while.

ashfordneil:
Re-requesting reviews now that `Result` has also been converted

tiny-james (approved):

lnewson (changes requested):
Requesting changes as mentioned on slack, as `Result` was not given the go ahead to be converted to a class. We only marginally gave the go ahead for `Optional` as the pros vs cons were marginally in favour of doing so. We don't have that for `Result` so it's not worth the risk involved in using classes.

modules/katamari/src/main/ts/ephox/katamari/api/Type.ts
```diff
@@ -26,6 +26,21 @@ export const isString: (value: any) => value is string =
...
+  if (!isObject(value)) {
+    return false;
+  }
+
+  const proto = getPrototypeOf(value);
+  if (proto === Object.prototype) {
+    return true;
+  } else {
+    return proto.toString() === '[object Object]';
+  }
```
lnewson, line 41:
```suggestion
  return proto === Object.prototype || proto.toString() === '[object Object]';
```

modules/katamari/src/main/ts/ephox/katamari/api/Type.ts
```diff
@@ -26,6 +26,21 @@ export const isString: (value: any) => value is string =
 export const isObject: (value: any) => value is Object =
   isType('object');
 
+const getPrototypeOf = Object.getPrototypeOf;
```
lnewson, line 29: Can we please put this at the top of this file?

modules/katamari/CHANGELOG.md
```diff
@@ -19,6 +20,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   - In particular, this means that if all `Cat`s are `Animal`s, then all `Optional<Cat>`s are now `Optional<Animal>`s.
 - The `Result` type is now covariant with respect to its type argument.
 - All singletons now have a `get` function that returns an `Optional` value.
+- Made constructing `Optional`s faster by re-implementing them as a class.
+- Made constructing `Result`s faster by re-implementing them as a class.
```
lnewson, line 24: This needs reverting

modules/katamari/src/main/ts/ephox/katamari/api/Optional.ts
```diff
@@ -1,141 +1,277 @@
...
+ * - Try to use `Optional` instead of null or undefined where you can
  */
-export interface Optional<T> {
-  /** If none, run whenNone; if some(a) run whenSome(a) */
-  readonly fold: <U> (whenNone: () => U, whenSome: (v: T) => U) => U;
 
-  readonly isSome: () => boolean;
-  readonly isNone: () => boolean;
+export class Optional<T> {
+  private readonly tag: boolean;
```
lnewson, line 23: Can we use a string/enum here instead of a boolean? It'll keep it consistent with other discriminate unions we have and make it clearer when debugging.

modules/katamari/src/main/ts/ephox/katamari/api/Optional.ts
```diff
@@ -1,141 +1,277 @@
...
+   */
+  public static none<T = never>(this: void): Optional<T> {
+    return Optional.singletonNone;
+  }
 
-  /** Run a function over the 'some' value.
-   *  "map" operation on the Optional functor.
+  /**
+   * Perform a transform on an Optional type. Similar to the fold method on a
+   * number of ADTs in the TinyMCE codebase, you pass two functions into this -
```
lnewson, line 56: No mention to TinyMCE inside a separate library please.

modules/katamari/src/main/ts/ephox/katamari/api/Optional.ts
```diff
@@ -1,141 +1,277 @@
...
-  /** Do all values contained in this option match this predicate? */
-  readonly forall: (f: (x: T) => boolean) => boolean;
+  /**
+   * Perform a transform on an Optional object, **if** there is a value. If you
+   * provide a function to turn a T into a U, this is the function you use to
+   * turn an `Optional<T>` into an `Optional<U>`. If this **does** contain a
+   * value then the output will also contain a value (that value being the
+   * output of `mapper(this.value)`), and if this **does not** contain a value
+   * then neither will the output.
+   */
```
lnewson, line 96: This does not read well. It seems to repeat the same thing over and over?

modules/katamari/src/main/ts/ephox/katamari/api/Optional.ts
```diff
@@ -1,141 +1,277 @@
...
+    }
+  }
+
+  // --- Monad ---
 
-  /** Return all values in this Optional that match the predicate.
-   *  The predicate may refine the constituent type using TypeScript type predicates.
+  /**
+   * Perform a transform on an Optional object, **if** there is a value. Unlike
+   * the map function earlier in the piece, here the transform itself also
```
lnewson, line 109: The "earlier in the piece" doesn't make sense when looking at the doc string for this alone. Each comment should be independent, not like a walkthrough where you read it in order.

- [ ] Ensure comments are independent/not relying on reading in order.

modules/katamari/src/main/ts/ephox/katamari/api/Optional.ts
lnewson: Looking more I can see this was done to be able to use truthy checks later on, so it's probably worth the payoff of using booleans.

modules/katamari/src/main/ts/ephox/katamari/api/Type.ts
```diff
@@ -26,6 +26,21 @@ export const isString: (value: any) => value is string =
...
+export const isPlainObject = (value: unknown): value is Object => {
+  if (!isObject(value)) {
+    return false;
+  }
+
+  const proto = getPrototypeOf(value);
+  if (proto === Object.prototype) {
+    return true;
+  } else {
+    return proto.toString() === '[object Object]';
```
TheSpyder, line 40: How often is the toString branch triggered? Slate uses `isPlainObject` quite a lot and it becomes quite a performance bottleneck if used too frequently.

modules/katamari/src/main/ts/ephox/katamari/api/Optional.ts
```diff
@@ -1,141 +1,277 @@
...
+ * strict-null-checks
+ * - Try to use `Optional` instead of null or undefined where you can
  */
-export interface Optional<T> {
-  /** If none, run whenNone; if some(a) run whenSome(a) */
-  readonly fold: <U> (whenNone: () => U, whenSome: (v: T) => U) => U;
 
-  readonly isSome: () => boolean;
-  readonly isNone: () => boolean;
+export class Optional<T> {
```
TheSpyder, line 22: Since we still emit ES5 for IE11 I'm curious what JS this actually outputs. Classes are ES6.

lnewson: It just outputs the good old prototype variant. We're already using classes for other older things in TinyMCE fwiw (eg the editor).

ashfordneil:
Replaced by #7191

PR #6941, opened 2021-06-16
TINY-7593: Normalize table border styling

spocke:
Related Ticket:

Description of Changes:
* This adds some normalization behavior when applying border style/color/width using the new table ui components. More details on this is available in the JIRA ticket.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/plugins/table/main/ts/core/BorderNormalizer.ts
+++ b/modules/tinymce/src/plugins/table/main/ts/core/BorderNormalizer.ts
@@ -1,0 +1,73 @@
+/**
+ * Copyright (c) Tiny Technologies, Inc. All rights reserved.
+ * Licensed under the LGPL or a commercial license.
+ * For LGPL see License.txt in the project root for license information.
+ * For commercial licenses see https://www.tiny.cloud/
+ */
+
+import { Arr, Optional } from '@ephox/katamari';
+import { Css, SugarElement } from '@ephox/sugar';
+import { DomModifier } from '../ui/DomModifier';
+
+const getComputedCombinedBorderProperty = (el: SugarElement<HTMLTableCellElement>, property: string) => {
+  const getSideValue = (side: string) => {
+    const propName = `border-${side}-${property}`;
+
+    // IE doesn't return the right computed style value even if it's explicitly set in the style
+    // attribute so we need to first check the raw value since that is the common case anyway
+    return Css.getRaw(el, propName).getOr(Css.get(el, propName));
+  };
+
+  const values = Arr.map([ 'top', 'left', 'bottom', 'right' ], getSideValue);
+  return values[0] !== '' && Arr.forall(values.slice(1), (v) => v === values[0]) ? Optional.some(values[0]) : Optional.none();
+};
+
+// IE returns subpixel border values even if the css is set to 1px so lets normalize that to the closest pixel
+const parsePxValue = (w: string) => Math.round(parseFloat(w));
+
+const getComputedBorderWidth = (el: SugarElement<HTMLTableCellElement>) => getComputedCombinedBorderProperty(el, 'width').map(parsePxValue).getOr(1);
+const getComputedBorderStyle = (el: SugarElement<HTMLTableCellElement>) => getComputedCombinedBorderProperty(el, 'style').getOr('solid');
+
+const setBorderStyle = (modifier: DomModifier, style: string) => modifier.setFormat('tablecellborderstyle', style);
+const setBorderWidth = (modifier: DomModifier, width: string) => modifier.setFormat('tablecellborderwidth', width);
+
+const isVisibleBorderStyle = (value: string) => value !== 'none' && value !== 'hidden' && value !== '';
+const hasVisibleBorderStyle = (el: SugarElement<HTMLTableCellElement>) => isVisibleBorderStyle(getComputedBorderStyle(el));
+const hasBorderColor = (el: SugarElement<HTMLTableCellElement>) => Css.getRaw(el, 'border-color').isSome();
+const hasOnePxBorderWidth = (el: SugarElement<HTMLTableCellElement>) => getComputedBorderWidth(el) === 1;
+
+export const normalizeSetCellBorderColor = (modifier: DomModifier, cell: SugarElement<HTMLTableCellElement>, value: string) => {
+  if (hasOnePxBorderWidth(cell) && hasVisibleBorderStyle(cell)) {
+    // We need to set it to unset the border-style when we are clearing the border-color to prevent the border from overlapping to adjacent cells
+    setBorderStyle(modifier, value === '' ? '' : 'double');
+  }
+};
+
+export const normalizeSetCellBorderWidth = (modifier: DomModifier, cell: SugarElement<HTMLTableCellElement>, value: string) => {
+  const width = value === '' ? 0 : parseFloat(value);
+
+  if (hasVisibleBorderStyle(cell)) {
+    if (width === 1) {
+      // We need to set border-style to `double` in order render the border-color
+      setBorderStyle(modifier, hasBorderColor(cell) ? 'double' : '');
+    } else if (width === 0) {
+      setBorderStyle(modifier, '');
+    }
+  }
+};
+
+export const normalizeSetCellBorderStyle = (modifier: DomModifier, cell: SugarElement<HTMLTableCellElement>, value: string) => {
+  if (hasOnePxBorderWidth(cell)) {
+    if (value === '') {
+      setBorderWidth(modifier, '');
+    } else if (value === 'solid') {
+      // We need to set the border-width to 2px if the cell has border-color in order to render that color
+      if (hasBorderColor(cell)) {
+        setBorderWidth(modifier, '2px');
+      }
+    } else if (isVisibleBorderStyle(value)) {
+      // Any visible borders except solid needs a 2px border-width to render
+      setBorderWidth(modifier, '2px');
+    }
+  }
+};
--- a/modules/tinymce/src/plugins/table/main/ts/core/BorderNormalizer.ts
+++ b/modules/tinymce/src/plugins/table/main/ts/core/BorderNormalizer.ts
@@ -5,7 +5,7 @@
  * For commercial licenses see https://www.tiny.cloud/
  */
 
-import { Arr, Optional } from '@ephox/katamari';
+import { Arr, Optionals } from '@ephox/katamari';
 import { Css, SugarElement } from '@ephox/sugar';
 import { DomModifier } from '../ui/DomModifier';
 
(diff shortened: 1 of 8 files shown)


modules/tinymce/src/plugins/table/test/ts/browser/BorderNormalizationTest.ts
```diff
@@ -0,0 +1,137 @@
+import { describe, context, it } from '@ephox/bedrock-client';
+import { Arr } from '@ephox/katamari';
+import { Attribute, SugarElement } from '@ephox/sugar';
+import { assert } from 'chai';
+import * as BorderNormalizer from 'tinymce/plugins/table//core/BorderNormalizer';
+import { DomModifier } from 'tinymce/plugins/table/ui/DomModifier';
+
+type normalizer = (modifier: DomModifier, cell: SugarElement<HTMLTableCellElement>, color: string) => void;
```
ashfordneil, line 8:
```suggestion
type Normalizer = (modifier: DomModifier, cell: SugarElement<HTMLTableCellElement>, color: string) => void;
```

modules/tinymce/src/plugins/table/test/ts/browser/command/ApplyCellStyleCommandTest.ts
```diff
@@ -208,5 +208,70 @@ describe('browser.tinymce.plugins.table.command.ApplyCellStyleCommandTest', () =
     ]);
   });
 
+  context('border normalization', () => {
+    context('border normalization single cell', () => {
+      const testSingleCellNormalization = (initialStyles: string, styles: Record<string, string>, expectedStyles: string) => {
```
ashfordneil, line 213: A pattern that I've seen in a lot of other tests is something like this:
```typescript
Arr.each([
  {
    label: 'TINY-7593: should switch border-style to double when setting a border-colour on a 1px border-width cell',
    initialStyles: 'border: 1px solid red',
    styles: { borderColor: 'blue' },
    expectedStyles: 'border: 1px double blue'
  }, {
    // ..
  }, {
    // ...
  }
], (spec) => {
  it(spec.label, () => {
    const editor = hook.editor();
    const html = ...
  });
});
```

It's personal preference, but if you wanted to avoid the one-line tests it's an option.

ashfordneil (approved):

lnewson (changes requested):
Marking as needs changes for now as this doesn't address the 2 issues the editor team rejected the previous approach for sorry:
- Doesn't work with the dialogs (it needs to, as otherwise the dialogs and toolbars will show completely different values and behave differently, which is a massive UX issue)
- The action should ideally be a single operation (ie the format should be tweaked before setting the styles, not set the styles and then change it immediately).

I'll need to review this properly and discuss it more with the editor team again, but I just don't want this getting merged yet. I'm sorry that it seems that feedback never made it back to the HV team. I think last time we discussed this when Mike C brought it up, we said it'd need to be behind a setting that integrators would enable if they want the normalization.

Edit: A third issue I can see, is that this will cause the incorrect value to be shown in the dropdowns. Say I select `1px` for the cell border width and "dotted" for the style then open the menu again it'll show `2px` for the border width which is unexpected behaviour. As such, any normalization that's happening also needs to apply when reading the values, not just writing them.

modules/tinymce/src/plugins/table/main/ts/core/BorderNormalizer.ts
```diff
@@ -0,0 +1,73 @@
...
+  }
+};
+
+export const normalizeSetCellBorderWidth = (modifier: DomModifier, cell: SugarElement<HTMLTableCellElement>, value: string) => {
+  const width = value === '' ? 0 : parseFloat(value);
+
+  if (hasVisibleBorderStyle(cell)) {
+    if (width === 1) {
+      // We need to set border-style to `double` in order render the border-color
+      setBorderStyle(modifier, hasBorderColor(cell) ? 'double' : '');
```
lnewson, line 52: This also breaks things if the previous border style was not single/double. Eg if it was dotted and then I change the width all of a sudden the style disappears.

modules/tinymce/src/plugins/table/test/ts/browser/BorderNormalizationTest.ts
spocke: Oops. Been writing reason for too long there everything is lower case first.

modules/tinymce/src/plugins/table/test/ts/browser/command/ApplyCellStyleCommandTest.ts
spocke: Good to know. It would make the test way more verbose though. But that seems like a good pattern to have. I think I stick with just looping the test guts for now.

modules/tinymce/src/plugins/table/main/ts/core/BorderNormalizer.ts
spocke: That is by design yes. All border-styles except solid will not render properly in 1px and our default is solid for tables so it makes sense to me to just clear it then.

lnewson: The editor team were in a call and we found that very confusing. I think this whole concept needs more thought and planning before implementation honestly. When we had the call today, we are planning to have a call with Product and HV to determine how to move forward with this.

lnewson: I should also add that behaviour is not how other editors work (eg google docs or word). They will keep the correct styling when changing the widths.

spocke:
> * Doesn't work with the dialogs (it needs to, as otherwise the dialogs and toolbars will show completely different values and behave differently, which is a massive UX issue)

The instructions I got from Fredrik was that the dialogs should remain unchanged since that is a more low level editing. Last time I heard he is our DX guy. Personally I'm not sure what is best here. Should be fairly easy to hook up to the dialogs as well unless we want the UI options to update based on the changes in the dialog that would be tricky but as a post submit normalizer then it would be easy.

> * The action should ideally be a single operation (ie the format should be tweaked before setting the styles, not set the styles and then change it immediately).

I think we shouldn't even use formats and do this all by setting styles. But that decision was made before this PR and unrevealing that at this point seems like a much bigger task. If you have the resources to do that before release go ahead.

> I'll need to review this properly and discuss it more with the editor team again, but I just don't want this getting merged yet. I'm sorry that it seems that feedback never made it back to the HV team. I think last time we discussed this when Mike C brought it up, we said it'd need to be behind a setting that integrators would enable if they want the normalization.

Please do. Fredik also told me that it shouldn't be optional and that could be something we add later if needed. If that has changed then sure I can add some form of normalizer option that can be disabled.

> Edit: A third issue I can see, is that this will cause the incorrect value to be shown in the dropdowns. Say I select `1px` for the cell border width and "dotted" for the style then open the menu again it'll show `2px` for the border width which is unexpected behaviour. As such, any normalization that's happening also needs to apply when reading the values, not just writing them.

Yes, that is a side effect of this. But that is the only sane way we could do this or not do it at all. We shouldn't lie about states in the UI that would be even more confusing. Switching to a dotted line on a 1px border will update the border to 2px that is a visual change that the border became wider and the UI just reflect what happened visually. I will double check with Fredrik on that.

This entire feature is pretty useless unless we figure out how to get browser to render 1px styled borders. The only way that seems possible is though this normalization. I did various experiments with trying to do this more with just css but the browsers all rendered that differently table border rendering is a mess.

modules/tinymce/src/plugins/table/main/ts/core/BorderNormalizer.ts
```diff
@@ -0,0 +1,73 @@
...
+  const getSideValue = (side: string) => {
+    const propName = `border-${side}-${property}`;
+
+    // IE doesn't return the right computed style value even if it's explicitly set in the style
+    // attribute so we need to first check the raw value since that is the common case anyway
+    return Css.getRaw(el, propName).getOr(Css.get(el, propName));
+  };
+
+  const values = Arr.map([ 'top', 'left', 'bottom', 'right' ], getSideValue);
+  return values[0] !== '' && Arr.forall(values.slice(1), (v) => v === values[0]) ? Optional.some(values[0]) : Optional.none();
```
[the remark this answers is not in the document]
spocke: That spares the Optional import as well. Nice catch.

modules/tinymce/src/plugins/table/main/ts/core/BorderNormalizer.ts
spocke: Yes, but they have a dedicated rendering engine not html as their base. So we can't compare apples and oranges here in this case. The logic here might be confusing yes, that is why I tried to back it up with comments, tests and details in the ticket for these different cases.

spocke: One thing that google docs does though is that it doesn't style the TD:s with borders that makes it possible to apply 1px styles to the cells and render them. However I remember asking Fredrik about why our content skin is setting styles on TD that and got the answer that that is a common way to do that. But maybe we should just alter the css to not do that then we wouldn't need this normalizer at all and instruct people that have their own css to not do that either. I will double check that it works correctly first on all platforms we support.

spocke: I immediately hit an issue with that google docs renders the table differently in FF vs Chrome. It will put border: 1px solid black on all cells on FF but on Chrome it will fallback to the default border colors of the table since that is how Chrome renders borders. We could in theory do this once we have a separate rendering of the view with the alt model but until then this would not work or produce super verbose html since we then would have to normalize the whole table.

lostkeys:
I hope I can clarify some things. 

This feature is to normalize table border styling behavior across browsers and is focused on user intent, not html output. You can clearly see the distinction where the dialogs are mapped 1-1 with their css properties, whereas the toolbar buttons give us more freedom to focus on getting the expected result. This is backed by testing, where test subjects clearly see our current implementation as broken.

The feature @spocke built works as expected except one "rule" that is missing from the logic which I will describe in another comment.

The only issue with dialogs that I remember were about the border styles not being the same between the toolbar button menu and the dialog select menu. I believe that has been resolved in the other PR related to the border-styles button. I don't see that a user can get into an "incompatible" state between the dialogs and buttons, were one of them simply stops working. If that is the case, let us know.

We also took the decision to not make this optional for the toolbar buttons as the normalization is isolated to the buttons only. If the developer needs web-developer-control over borders they can use the dialogs. We're not changing the default table toolbar, so adding the buttons is a deliberate action by the developer in it-self. So in effect this feature

I'm ok with the border-width changing automatically, that was expected and to be honest the point of this feature. If we need to abstract this further, we'll address that later, but this is good enough.

This issue has been thorouhly discussed for a very long time and this is the best we are going to get it at this time. So if there aren't any major issues with the code itself like bugs or major maintenance issues, the feature works as designed and is approved.

lostkeys (changes requested):
The missing rule is that when setting border-width to more than 2, the border-style should switch back to solid if it were set to double. And yes, if the end-user explicitly wanted to use double, they would have to "reassign that", but that is an accepted trade-off.

- [x] Change "double" become "solid" if border-width is above 2 px.

Masking "double" as "solid" is one of the few improvement that I wished we could implement but it did not fit the scope.

lostkeys (approved):

spocke:
I'm converting this to draft if anyone needs to check the code feel free to delete if it needed.

lnewson:
Closing this based on recent discussions. We can reopen still if we decide to come back to this approach.

PR #6945, opened 2021-06-17
TINY-7602: Fix throbber focus logic not working with pastebins

ltrouton:
Related Ticket: TINY-7602, TINY-7373

Description of Changes:
* Update the throbber focus code (TINY-7373) to not steal focus from the pastebin when pasting in IE11

Pre-checks:
* [X] Changelog entry added (Haven't added a changelog since the code causing the issue hasn't be released)

Review:
* [X] Milestone set


modules/tinymce/src/themes/silver/main/ts/ui/throbber/Throbber.ts
```diff
@@ -103,12 +106,19 @@ const renderThrobber = (spec): AlloySpec => ({
   components: [ ]
 });
 
+const isPasteBinTarget = (target: unknown) =>
+  Optional.from(target)
+    .map(SugarElement.fromDom)
```
lnewson, line 111: This isn't going to work nicely when the target is an editor. Basically, we can't be converting to a SugarElement, without first knowing we're dealing with a `Node`.

modules/tinymce/src/themes/silver/test/ts/browser/throbber/ThrobberFocusTest.ts
```diff
@@ -189,4 +191,30 @@ describe('browser.tinymce.themes.silver.throbber.ThrobberFocusTest', () => {
...
+    const normalDiv = SugarElement.fromHtml<HTMLElement>('<div class="not-pastebin" >&nbsp;</div>');
+    const fakePasteBin = SugarElement.fromHtml<HTMLElement>('<div class="mce-pastebin">&nbsp;</div>');
+
+    Insert.append(editorBody, normalDiv);
+    Insert.append(editorBody, fakePasteBin);
+    await pEnableThrobber(editor, pAssertThrobberFocus);
+    FocusTools.setFocus(SugarBody.body(), '#tempInput');
+
+    // Make sure throbber doesn't take focus when the pastebin is the target
+    editor.dom.fire(fakePasteBin.dom, 'focusin');
```
lnewson, line 207: Can we please not use fake events here and instead actually focus the element using `Focus.focus(fakePasteBin)`?

ltrouton: I had tried that but was unable to get it to trigger the listeners set up in Throbber.ts

ltrouton: I thought I had tried this before, but adding `tabindex` to the divs seems to allow `Focus.focus` to be used.

lnewson: The problem isn't really that from what I can see, it's that we're not setting up the pastebin correctly. If you look at https://github.com/tinymce/tinymce/blob/develop/modules/tinymce/src/plugins/paste/main/ts/core/PasteBin.ts#L40 you'll see we add a number of other attributes (the contenteditable one specifically). To properly test this, we should be setting up a more realistic paste bin.

lnewson: In this case, it's likely the `contenteditable="true"` attribute that's needed to handle the focus call.

modules/tinymce/src/themes/silver/main/ts/ui/throbber/Throbber.ts
lnewson: To clarify more, the type for `target` here should be `Editor | Node`

modules/tinymce/src/themes/silver/test/ts/browser/throbber/ThrobberFocusTest.ts
ltrouton: Have added a note as to why we can't use `Focus.focus` here

modules/tinymce/src/themes/silver/main/ts/ui/throbber/Throbber.ts
ltrouton: Thanks for pointing this out. This has been reworked now with your provided snippets

lnewson (approved):

ashfordneil (approved):

metricjs (approved):

PR #6946, opened 2021-06-17
TINY-7588: Upgrade to latest eslint-plugin and auto fix import/order rules

lnewson:
Related Ticket: TINY-7588

Description of Changes:
* Upgraded to latest eslint-plugin and upgraded other deps (typescript and tslib)
* Fixed issues from the new import/order eslint rule configuration (done via the fixer rule, eg `yarn eslint --fix`)

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


JamesToohey (approved):

tiny-james (approved):
I checked a few files at random and didn't see any problems.

modules/tinymce/src/plugins/table/main/ts/core/Util.ts
```diff
@@ -8,6 +8,7 @@
 import { Arr, Optional, Strings } from '@ephox/katamari';
 import { TableLookup } from '@ephox/snooker';
 import { Attribute, Compare, SugarElement } from '@ephox/sugar';
+
 import Editor from 'tinymce/core/api/Editor';
```
ashfordneil, line 12: Just noticed that for some reason we don't have a lint rule enforcing that there should be a blank line between imports and the start of actual code. We should (separately) fix that

yarn.lock
```diff
@@ -1287,9 +1287,9 @@
     defer-to-connect "^2.0.0"
 
 "@tinymce/eslint-plugin@^1.8.1":
-  version "1.8.2"
-  resolved "https://registry.yarnpkg.com/@tinymce/eslint-plugin/-/eslint-plugin-1.8.2.tgz#5e70f59196f29b10363585f32a0b69d28f663205"
-  integrity sha512-18s/lxY6knDLsW/xICS/qjJSESBHiIJU4nzJkEkTquakENoHjWJonP8dm81OhfjYqyS2h6DRVtyIpoSLZrLVlA==
+  version "1.9.1"
```
ashfordneil, line 1290: What are your thoughts on also bumping the version in the package.json from ^1.8.1 to ^1.9.2?

ashfordneil (approved):

modules/tinymce/src/plugins/table/main/ts/core/Util.ts
lnewson: Yeah, we should. I've fixed this specific case for now. If anyone wants to try this, it's a separate linter rule: https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/newline-after-import.md

yarn.lock
lnewson: Point. I've converted it to use `^1.9.1`.

modules/boss/src/main/ts/ephox/boss/api/Main.ts
```diff
@@ -6,7 +6,6 @@ import { SpecialGene } from './SpecialGene';
 import { TestUniverse } from './TestUniverse';
 import { TextGene } from './TextGene';
 import { Universe } from './Universe';
-
 /* eslint-disable import/order */
```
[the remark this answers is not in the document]
lnewson: Unfortunately no as it turns out. The disable only applies to the lines within the "ignore" block. So since `import { Universe } from './Universe';` is outside that, it'll still get flagged as it sees an import on the lines below and it doesn't match the spacing rules.

I mentioned a few alternatives on slack, but all are gross so I believe this is still the best solution.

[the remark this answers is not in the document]
ashfordneil: Yeah, I'm happy with this given the alternatives.

PR #6952, opened 2021-06-19
TINY-7572: Minor toolbar button setup performance improvement and cleanup

lnewson:
Related Ticket: TINY-7572

Description of Changes:
The bespoke toolbar buttons where running a number of times before init and this didn't do anything meaningful as they need the initial location set during init. Additionally, these aren't the fastest lookups so this just makes the bespoke and other core buttons only start listening to events once the editor is initialized. I also did some code cleanup to remove duplicate code which made this easier to implement. Doing all this ends up saving another ~20-25ms during init time for the full feature demo.

Pre-checks:
* [x] ~Changelog entry added~ (Already have a changelog from the previous PR)
* [x] ~Tests have been added (if applicable)~ (Covered by existing tests)

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/themes/silver/main/ts/ui/core/ControlUtils.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/ui/core/ControlUtils.ts
@@ -10,7 +10,7 @@ import { Singleton } from '@ephox/katamari';
 import Editor from 'tinymce/core/api/Editor';
 import { Toolbar } from 'tinymce/core/api/ui/Ui';
 
-import { FormatterFormatItem } from '../BespokeSelect';
+import { FormatterFormatItem } from './complex/BespokeSelect';
 
 const onSetupFormatToggle = (editor: Editor, name: string) => (api: Toolbar.ToolbarToggleButtonInstanceApi) => {
   const boundCallback = Singleton.unbindable();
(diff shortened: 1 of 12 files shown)


modules/tinymce/src/themes/silver/main/ts/ui/core/ControlUtils.ts
```diff
@@ -22,9 +22,29 @@ const onSetupFormatToggle = (editor: Editor, name: string) => (api: Toolbar.Tool
   };
 
   // The editor may or may not have been setup yet, so check for that
-  editor.initialized ? init() : editor.on('init', init);
+  editor.initialized ? init() : editor.once('init', init);
 
-  return boundCallback.clear;
+  return () => {
+    editor.off('init', init);
```
lnewson, line 28: This was actually a bug, as due to refreshes happening the button would have been setup and destroyed at least once before init. This meant that the `init` code actually ran twice since it wasn't unbound.

ashfordneil: Is this still necessary if you changed it above to use `editor.once`?

modules/tinymce/src/themes/silver/main/ts/ui/core/ControlUtils.ts
```diff
@@ -22,9 +22,29 @@ const onSetupFormatToggle = (editor: Editor, name: string) => (api: Toolbar.Tool
...
+  const init = () => {
+    f(api);
+    editor.on(event, handleEvent);
+  };
+
+  // The editor may or may not have been setup yet, so check for that
+  editor.initialized ? init() : editor.once('init', init);
+
+  return () => {
+    editor.off('init', init);
```
ashfordneil, line 45: Here also, do we need `.off` and `.once`? Or is it just a "to be safe" sorta thing

ashfordneil (approved):

modules/tinymce/src/themes/silver/main/ts/ui/core/ControlUtils.ts
lnewson: Yes, as once only removes it when it's actually called. If you read the above, I mentioned it's binding but not ever running. As such even using once it would still bind multiple times if the toolbar button is setup and destroyed before the `init` event is fired.

lnewson: I should add the reason I changed it to `once` is so that we don't keep a reference for the entire life of the editor once it does actually run.

modules/tinymce/src/themes/silver/main/ts/ui/core/UndoRedo.ts
```diff
@@ -6,52 +6,50 @@
...
     shortcut: 'Meta+Y',
-    onSetup: (api) => toggleUndoRedoState(api, editor, 'hasRedo'),
-    onAction: () => editor.execCommand('redo')
+    onSetup: onSetupUndoRedoState(editor, 'hasRedo'),
+    onAction: onActionExecCommand(editor, 'redo')
   });
 };
 
+// Note: The undo/redo buttons are disabled by default here, as they'll be rendered
+// on init generally and it won't haven't any undo levels at that stage.
```
ltrouton, line 37:
```suggestion
// on init generally and it won't have any undo levels at that stage.
```

ltrouton (approved):

JamesToohey (approved):

PR #6959, opened 2021-06-22
TINY-6149: Add a "language" button / menu for setting "lang" attributes as a format

ashfordneil:
Related Ticket: TINY-6149

# Description of Changes:
* Add a `lang` format
* Add a `language` menu item / toolbar button for applying that format (see details below)
* Add a `content_langs` setting item for specifying which languages can be used
* Fix a formatter bug (see details below) that was getting in the way

## Fix a formatter bug
Previously, if your format had something like `{ ..., attributes: { 'data-my-attribute': null } }` you would never be able to correctly "match" that format in some content. The `matchNode` function gets as far as "the spec says we need a `data-my-attribute` attribute but the element doesn't have one" and bails. I've corrected that, so that if the spec asks for a falsey attribute (or style) and the element doesn't have that attribute (or style) we still consider it a candidate for a match.

## Add a `language` menu item / toolbar button
Turns out this shared a lot of code with the LineHeight UI (and not just because that was the last time I looked in this folder, and I took inspiration from it). The concept of "we have a list of allowed values for this format, and we need a dropdown to let people choose from them" is something that's surprisingly difficult to handle, mainly because the formatter's event handling system doesn't support it, so you need to do the nodechange handlers yourself. As such, I've pulled it out into `ChoiceControls.ts`.

This also included fixing some type errors / implicit `any`s within some areas of the formatter. @ RTC people I found a type error in RTC.ts, which will need investigation on your end before we merge this.
- [x] Make sure this type change doesn't break RTC

# Pre-checks:
* [x] Changelog entry added

# Before merging:
* [x] Milestone set
* [x] Find out what the default value of the content_langs setting is meant to be, as that wasn't specified initially


modules/tinymce/src/core/main/ts/fmt/MatchFormat.ts
```diff
@@ -67,65 +67,65 @@ const matchName = (dom: DOMUtils, node: Node, format) => {
...
-  let key, value;
+const matchItems = (dom: DOMUtils, node: Node, format: Format, itemName: string, similar: boolean, vars: FormatVars): boolean => {
   const items = format[itemName];
-  let i;
 
   // Custom match
   if (format.onmatch) {
-    return format.onmatch(node, format, itemName);
+    const onMatch: (node: Node, format: Format, name: string) => boolean = format.onmatch;
+    return onMatch(node, format, itemName);
```
lnewson, line 76: Why was this changed? This is very ugly and doesn't add anything, so I'd like to see this reverted please.
```suggestion
    return format.onmatch(node, format, itemName);
```

modules/tinymce/src/core/main/ts/fmt/MatchFormat.ts
```diff
@@ -47,7 +47,7 @@ const matchParents = (editor: Editor, node: Node, name: string, vars: FormatVars
   });
 
   // Do an exact check on the similar format element
-  return matchNode(editor, node, name, vars);
+  return !!matchNode(editor, node, name, vars);
```
ashfordneil, line 50: This should be harmless, as we were only looking at the output of matchParents as a boolean.

That said, probably could use:
```suggestion
  return matchNode(editor, node, name, vars) !== undefined;
```

modules/katamari/src/main/ts/ephox/katamari/api/Arr.ts
```diff
@@ -182,7 +182,7 @@ export const findUntil: {
 
 export const find: {
   <T, U extends T>(xs: ArrayLike<T>, pred: ArrayGuardPredicate<T, U>): Optional<U>;
-  <T = any>(xs: ArrayLike<T>, pred: ArrayPredicate<T>): Optional<T>;
+  <T>(xs: ArrayLike<T>, pred: ArrayPredicate<T>): Optional<T>;
```
lnewson, line 185: This is a breaking change to the types which could cause a number of things to fail to compile (which is probably a good thing). It however doesn't seem like it's needed for this PR, so if you want to make this change, please log a separate PR.

- [ ] Break this out into a separate PR if needed, including a changelog entry to state it no longer defaults to any.

modules/tinymce/src/core/main/ts/fmt/MatchFormat.ts
```diff
@@ -67,65 +67,65 @@ const matchName = (dom: DOMUtils, node: Node, format) => {
...
-const matchItems = (dom: DOMUtils, node: Node, format, itemName: string, similar: boolean, vars: FormatVars) => {
-  let key, value;
+const matchItems = (dom: DOMUtils, node: Node, format: Format, itemName: string, similar: boolean, vars: FormatVars): boolean => {
   const items = format[itemName];
-  let i;
 
   // Custom match
   if (format.onmatch) {
-    return format.onmatch(node, format, itemName);
+    const onMatch: (node: Node, format: Format, name: string) => boolean = format.onmatch;
```
ashfordneil, line 75: The reason this cast is here is because `format.onmatch` technically can't take `Format` as its second argument. The `.onmatch` of a `SelectorFormat` can only accept a `SelectorFormat`, the `.onmatch` of a `BlockFormat` can only accept a `BlockFormat`, etc etc. The cast is safe, because we only pass the format itself into this `onmatch` function, but unfortunately I'm not aware of any way to skip the cast.

modules/tinymce/src/core/main/ts/fmt/MatchFormat.ts
```diff
@@ -67,65 +67,65 @@ const matchName = (dom: DOMUtils, node: Node, format) => {
...
+      for (const key in items) {
         if (Obj.has(items, key)) {
-          if (itemName === 'attributes') {
-            value = dom.getAttrib(node, key);
-          } else {
-            value = FormatUtils.getStyle(dom, node, key);
+          const value = itemName === 'attributes' ? dom.getAttrib(node, key) : FormatUtils.getStyle(dom, node, key);
+          const expectedValue = FormatUtils.replaceVars(items[key], vars);
+
+          if (!value && !expectedValue) {
```
ashfordneil, line 88: This here is the actual bug fix for the formatter change. If the expected value is falsey, and the value itself is falsey, then for this key we have a match (thus move onto the next key).

If this condition needs tweaking at all to account for `similar` or `format.exact` (I don't know much about either value) please let me know.

modules/tinymce/CHANGELOG.md
```diff
@@ -16,6 +16,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added a new `tablecellborderstyle` toolbar button and menu item to change table cell border style #TINY-7478
 - Added a new `tablecaption` toolbar button and menu item to toggle captions on tables #TINY-7479
 - Added a new `mceTableToggleCaption` command that toggles captions on a selected table #TINY-7479
+- Added a new `language` menu item and toolbar button to set the `lang` attribute on content #TINY-6149
+- Added a new `content_langs` setting to specify the languages available for the `language` menu item and toolbar button #TINY-6149
```
lnewson, line 20: This is all part of the one feature and has duplicate information, so please merge this into a single changelog entry.

- [ ] Merge into a single changelog entry

modules/tinymce/src/core/main/ts/api/Formatter.ts
```diff
@@ -39,7 +39,7 @@ interface Formatter extends FormatRegistry {
   match: (name: string, vars?: FormatVars, node?: Node) => boolean;
   closest: (names) => string | null;
   matchAll: (names: string[], vars?: FormatVars) => string[];
-  matchNode: (node: Node, name: string, vars?: FormatVars, similar?: boolean) => boolean;
+  matchNode: (node: Node, name: string, vars?: FormatVars, similar?: boolean) => Format | undefined;
```
lnewson, line 42: Please include a changelog entry for this fix

- [ ] Add a fixed changelog entry

modules/tinymce/src/plugins/table/main/ts/ui/Helpers.ts
```diff
@@ -148,7 +148,7 @@ const getSharedValues = <T>(data: Array<T>) => {
 // because some of these are crazy complicated
 
 const getAlignment = (formats: string[], formatName: string, editor: Editor, elm: Node) =>
-  Arr.find(formats, (name) => editor.formatter.matchNode(elm, formatName + name)).getOr('');
+  Arr.find(formats, (name) => !!editor.formatter.matchNode(elm, formatName + name)).getOr('');
```
ashfordneil, line 151:
```suggestion
  Arr.find(formats, (name) => editor.formatter.matchNode(elm, formatName + name) !== undefined).getOr('');
```
I think this works, but without strict typing turned on it's hard to be sure.

modules/tinymce/src/core/main/ts/fmt/MatchFormat.ts
lnewson: I think this is fine for now, as it's doing the same a couple lines above. If we do make apply the suggestion, please make sure to do it in both places.

modules/tinymce/src/core/main/ts/fmt/MatchFormat.ts
lnewson, line 88: As this is new code, can we please not use falsy checks here and instead use the relevant `Type` functions. If you're keen we can probably change the other could below as well I guess?

- [ ] Avoid falsy checks please

modules/tinymce/src/themes/silver/main/ts/api/Settings.ts
```diff
@@ -50,6 +51,25 @@ const isMergeStyleFormats = (editor: Editor): boolean => editor.getParam('style_
 const getLineHeightFormats = (editor: Editor): string[] =>
   editor.getParam('lineheight_formats', '1 1.1 1.2 1.3 1.4 1.5 2', 'string').split(' ');
 
+const getContentLangugaes = (editor: Editor): ContentLanguage[] => {
+  const defaultSettings = [
+    { code: 'en', title: 'English (US)' },
```
lnewson, line 56: Can we please swap the order of these, as this goes against existing code as the `title` is always first generally.

modules/tinymce/src/themes/silver/main/ts/api/Settings.ts
```diff
@@ -50,6 +51,25 @@ const isMergeStyleFormats = (editor: Editor): boolean => editor.getParam('style_
 const getLineHeightFormats = (editor: Editor): string[] =>
   editor.getParam('lineheight_formats', '1 1.1 1.2 1.3 1.4 1.5 2', 'string').split(' ');
 
+const getContentLangugaes = (editor: Editor): ContentLanguage[] => {
```
lnewson, line 54: This doesn't match the requirements, as it's supposed to use the same languages as the existing spellchecker language menu which is:
```
const selectContentLanguages = 'en,es,fr,de,pt,zh';
```

lnewson (changes requested):
Haven't finished reviewing, but found an issue that doesn't meet the requirements so marking as needs changes.

modules/tinymce/src/plugins/table/main/ts/ui/Helpers.ts
lnewson: Please use `!Type.isUndefined` instead of this.

modules/tinymce/src/core/main/ts/fmt/MatchFormat.ts
lnewson: We also need a changelog entry for this fix.

- [ ] Changelog entry

modules/tinymce/src/core/main/ts/fmt/DefaultFormats.ts
```diff
@@ -175,6 +175,16 @@ const get = (dom: DOMUtils) => {
       }
     },
 
+    lang: {
+      inline: 'span',
+      clear_child_styles: true,
+      remove_similar: true,
+      attributes: {
+        'lang': '%value',
+        'data-mce-lang': (vars) => vars?.customValue || null
```
lnewson, line 184: Should this be this instead? That way we're not using a falsy check:
```suggestion
        'data-mce-lang': (vars) => vars?.customValue ?? null
```

modules/tinymce/src/core/main/ts/fmt/MatchFormat.ts
lnewson: I see you said why after I'd added this in my review. That seems like a bad reason to be doing this and instead I'd prefer the `format: Format` type be removed above instead until we can resolve what's going on.

modules/tinymce/src/core/main/ts/fmt/MatchFormat.ts
```diff
@@ -67,65 +67,65 @@ const matchName = (dom: DOMUtils, node: Node, format) => {
...
+            continue;
           }
 
           if (similar && !value && !format.exact) {
-            return;
+            return undefined;
```
lnewson, line 93: This is incorrect, it should be returning `false`. The same issue exists below as well.

modules/tinymce/src/themes/silver/main/ts/ui/core/ChoiceControls.ts
```diff
@@ -0,0 +1,150 @@
...
+
+  readonly getOptions: (editor: Editor) => Array<T>;
+  readonly normalise: (item: T) => string;
+  readonly display: (item: T) => string;
+
+  readonly getCurrent: (editor: Editor) => Optional<T>;
+  readonly setCurrent: (editor: Editor, value: T) => void;
+}
+
+const registerController = <T>(editor: Editor, spec: ControlSpec<T>) => {
```
lnewson, line 30: This is really looking the same as a bespoke select now... we should log a follow up JIRA to look at merging the two, as I believe it could be done.

modules/tinymce/src/themes/silver/main/ts/ui/core/ChoiceControls.ts
```diff
@@ -0,0 +1,150 @@
...
+  setCurrent: (editor, value) => editor.execCommand('LineHeight', false, value)
+};
+
+const languageSpec: ControlSpec<ContentLanguage> = {
+  name: 'language',
+  text: 'Language',
+  icon: 'translate',
+
+  getOptions: Settings.getContentLangugaes,
+  normalise: (input) => Type.isUndefined(input.customCode) ? input.code : `${input.code}/${input.customCode}`,
```
lnewson, line 115: This really isn't normalising anything anymore, it's generating a unique lookup key so can we please rename this to reflect that?

modules/tinymce/src/themes/silver/main/ts/ui/core/ChoiceControls.ts
```diff
@@ -0,0 +1,150 @@
...
+
+      return codeOpt.map((code): ContentLanguage => ({ code, customCode, title }));
+    });
+  },
+  setCurrent: (editor, lang) => {
+    editor.formatter.toggle('lang', {
+      value: lang.code,
+      customValue: lang.customCode
+    });
+    editor.nodeChanged();
```
lnewson, line 139: This needs to be in an undomanager transaction, otherwise you won't be able to undo the changes.

modules/tinymce/src/themes/silver/test/ts/browser/editor/core/ChoiceControlsTest.ts
```diff
@@ -0,0 +1,303 @@
...
+import Theme from 'tinymce/themes/silver/Theme';
+
+interface ToolbarOrMenuSpec {
+  readonly name: string;
+  readonly pOpen: (editor: Editor, title: string) => Promise<void>;
+  readonly close: (editor: Editor, title: string) => void;
+  readonly menuSelector: string;
+}
+
+describe('browser.tinymce.themes.silver.editor.core.ChoiceControlsTest', () => {
```
lnewson, line 18: This is a pretty messy test now... I think it would be better to have 2 separate test files for line height and content language. If you're really against that, then we should have a `context` per type instead to properly abstract the 2 different types of tests happening here and the prefixes on each separate context should be removed.

modules/tinymce/src/themes/silver/main/ts/api/Settings.ts
tiny-james: Also there is a typo in the spelling.
```
getContentLangugaes
               ||
getContentLanguages
```

modules/tinymce/src/core/main/ts/fmt/DefaultFormats.ts
```diff
@@ -175,6 +175,16 @@ const get = (dom: DOMUtils) => {
       }
     },
 
+    lang: {
+      inline: 'span',
+      clear_child_styles: true,
+      remove_similar: true,
+      attributes: {
+        'lang': '%value',
+        'data-mce-lang': (vars) => vars?.customValue ?? null
```
spocke, line 184: Right, this has impact on RTC we only support a smaller set of formatter api things and functions on these values is not one of them. I'm adding as separate ticket to expand the rtc formatter to include this as well. Curious why is this data attribute needed?

modules/tinymce/src/core/main/ts/Rtc.ts
```diff
@@ -226,7 +227,7 @@ const makeNoopAdaptor = (): RtcAdaptor => {
     formatter: {
       match: Fun.never,
       matchAll: Fun.constant([]),
-      matchNode: Fun.never,
+      matchNode: Fun.constant(undefined),
```
spocke, line 230: Since we are currently throwing on the RTC side for this function call and this noop interface is only used when the editor is in the initializing state the impact should be low. But good to get the right type here.

modules/tinymce/src/core/main/ts/fmt/MatchFormat.ts
```diff
@@ -67,65 +67,65 @@ const matchName = (dom: DOMUtils, node: Node, format) => {
...
+          const value = itemName === 'attributes' ? dom.getAttrib(node, key) : FormatUtils.getStyle(dom, node, key);
+          const expectedValue = FormatUtils.replaceVars(items[key], vars);
+
+          if (Type.isNullable(value) && Type.isNullable(expectedValue)) {
+            continue;
           }
 
           if (similar && !value && !format.exact) {
-            return;
+            return false;
```
spocke, line 93: Yeah, I used to be pretty obsessed about size. So true/false was usually falsy things that doesn't work well with TS.

modules/tinymce/src/core/main/ts/fmt/MatchFormat.ts
spocke: Should this be:
`return !Type.isUndefined(matchNode(editor, node, name, vars));`
or
`return Type.isUndefined(matchNode(editor, node, name, vars)) === false;`

modules/tinymce/src/themes/silver/main/ts/ui/core/ChoiceControls.ts
spocke: Maybe this should be a command I think most other UI things can be done though commands. Like a `mceToggleLanguage` or just `ToggleLanguage` if we are getting rid of the `mce` prefixes?

spocke (approved):

modules/tinymce/src/core/main/ts/fmt/MatchFormat.ts
ashfordneil: Have changed this to a different cast that's less ugly but still more correct than just leaving format to be an implicit any.

modules/katamari/src/main/ts/ephox/katamari/api/Arr.ts
ashfordneil: Reverted

modules/tinymce/src/core/main/ts/fmt/MatchFormat.ts
ashfordneil: Added

modules/tinymce/CHANGELOG.md
ashfordneil: Merged, as per discussion in slack. Would be nice to have some more opinions on whether the nested entry is good or not.

modules/tinymce/src/core/main/ts/fmt/MatchFormat.ts
ashfordneil: Done — I'm slightly worried there will be something here that depended on empty strings being falsey, but that's what tests are for I guess.

modules/tinymce/src/themes/silver/main/ts/ui/core/ChoiceControls.ts
ashfordneil: Added it to a transaction for now, will have a chat among the team about whether or not to add a command.

modules/tinymce/src/core/main/ts/fmt/DefaultFormats.ts
ashfordneil: The data attribute is there for languages like "English (US Medical)" that are useful variants to be able to tell the difference between in your HTML, but don't have valid ISO language codes that could be put into a `lang` attribute.

For what it's worth, the main reason we want to note things like "this content uses medical terms" in the HTML is so that spellchecker pro can know to treat them differently.

modules/tinymce/src/core/main/ts/api/Formatter.ts
ashfordneil: Also added.

modules/tinymce/src/themes/silver/test/ts/browser/editor/core/ChoiceControlsTest.ts
ashfordneil: Because so much of the logic is in the choice menu abstraction, rather than the specific implementations of line height or language buttons, I'd rather test it this way. I've added the contexts as you've asked

modules/tinymce/src/themes/silver/main/ts/ui/core/ChoiceControls.ts
ashfordneil: Created TINY-7642, if you want to update the description beyond my guesses for what this would entail.

modules/tinymce/src/themes/silver/test/ts/browser/editor/core/ChoiceControlsTest.ts
```diff
@@ -0,0 +1,301 @@
...
+        it(`TINY-4843: ${spec.name} lists specified line heights`, async () => {
+          const editor = hook.editor();
+          await spec.pOpen(editor, 'Line height');
+          await pAssertOptions(editor, spec.menuSelector, [ '1', '1.1', '1.11', '1.111' ], Optional.none());
+          spec.close(editor, 'Line height');
+        });
+      });
+    });
+
+    context('Line height normalisation tests', () => {
```
lnewson, line 130:
```suggestion
    context('normalisation tests', () => {
```
Otherwise this is going to read "Line height Line height normalisation tests"

modules/tinymce/src/core/main/ts/fmt/MatchFormat.ts
lnewson: Yeah, if we do change it, it should be one of those.

lnewson (approved):

modules/tinymce/CHANGELOG.md
lnewson: We did discuss the other day already, but it seems you may have forgotten about it? I've added a suggestion anyways.

lnewson (changes requested):
Actually, I just noticed CI is failing on this so removing my approval as that needs to be fixed.

modules/tinymce/src/themes/silver/main/ts/ui/core/ChoiceControls.ts
ashfordneil: Now a command, currently just called `Lang` for parity with the other commands that are just wrappers around a format.

modules/tinymce/src/themes/silver/main/ts/ui/core/ChoiceControls.ts
```diff
@@ -0,0 +1,131 @@
...
+ * For commercial licenses see https://www.tiny.cloud/
+ */
+
+import { Arr, Fun, Obj, Optional, Singleton, Type } from '@ephox/katamari';
+import { Dimension } from '@ephox/sugar';
+
+import Editor from 'tinymce/core/api/Editor';
+import { ContentLanguage } from 'tinymce/core/api/SettingsTypes';
+import { Menu } from 'tinymce/core/api/ui/Ui';
+import { getContentLanguage } from 'tinymce/core/commands/ContentLanguage';
```
lnewson, line 14: This is not allowed... you can only import from `api` paths in core. Please move this logic into Silver instead.

modules/tinymce/src/core/main/ts/api/EditorCommands.ts
```diff
@@ -407,6 +408,10 @@ class EditorCommands {
         LineHeightCommands.lineHeightAction(editor, value);
       },
 
+      'Lang': (command, ui, value) => {
+        ContentLanguageCommands.contentLanguageAction(editor, value);
```
lnewson, line 412: As I mentioned the other day, this should be using the existing the existing `toggleFormat` function like `Bold`, etc... as well. So this should be like this:
```suggestion
        self.toggleFormat(command, {
          value: value.code,
          customValue: value.customCode
        });
```

lnewson (changes requested):

modules/tinymce/src/core/main/ts/api/EditorCommands.ts
lnewson: I should add I'm guessing this may have been done since you went to do the query command and were trying to keep it the same as LineHeight and FontSize. However, since we're not doing the query command anymore it should be consistent with the others.

modules/tinymce/CHANGELOG.md
```diff
@@ -19,6 +19,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added a new `tablecellbordercolor` toolbar button and menu item to select table cell border colors, with an accompanying setting `table_cell_border_color_map` to customize the available values #TINY-7480
 - Added a new `tablecellbackgroundcolor` toolbar button and menu item to select table cell background colors, with an accompanying setting `table_cell_background_color_map` to customize the available values #TINY-7480
 - Added a new `initData` property to `fancymenuitem` to allow custom initialization data #TINY-7480
+- Added a new `language` menu item and toolbar button to add `lang` attributes to content, with an accompanying `content_langs` setting to specify the languages available #TINY-6149
+- A new `lang` format is now available that be used with `editor.formatter`, or applied with the `Lang` editor command #TINY-6149
```
lnewson, line 23:
```suggestion
- A new `lang` format is now available that can be used with `editor.formatter`, or applied with the `Lang` editor command #TINY-6149
```

modules/tinymce/CHANGELOG.md
```diff
@@ -38,6 +40,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Unbinding an event handler did not take effect immediately while the event was firing #TINY-7436
 - Binding an event handler incorrectly took effect immediately while the event was firing #TINY-7436
 - Partially transparent RGBA values provided in the `color_map` setting were given the wrong hex value #TINY-7163
+- The formatter match APIs was incorrectly returning false for formats that specified an attribute or style should be removed from an element #TINY-6149
```
lnewson, line 43:
```suggestion
- The formatter match APIs were incorrectly returning false for formats that specified an attribute or style should be removed from an element #TINY-6149
```
Sorry, this was something I messed up in my previous suggestion :disappointed:

modules/tinymce/src/themes/silver/main/ts/ui/core/ChoiceControls.ts
```diff
@@ -0,0 +1,145 @@
...
+      Optional.some(n)
+        .filter(SugarNode.isElement)
+        .bind((ele) => {
+          const codeOpt = Attribute.getOpt(ele, 'lang');
+          const customCode = Attribute.getOpt(ele, 'data-mce-lang').getOrUndefined();
+          return codeOpt.map((code): ContentLanguage => ({
+            code,
+            customCode,
+            title: ''
+          }));
```
lnewson, line 131:
```suggestion
          return codeOpt.map((code): ContentLanguage => ({
            code,
            customCode: Attribute.getOpt(ele, 'data-mce-lang').getOrUndefined(),
            title: ''
          }));
```
This way we're not creating an Optional and looking up the attribute when we don't need to be.

lnewson: If you prefer, we could also do something like this:
```suggestion
          return codeOpt.map((code): ContentLanguage => {
            const customCode = Attribute.getOpt(ele, 'data-mce-lang').getOrUndefined();
            return { code, customCode, title: '' };
          });
```

lnewson (approved):

modules/tinymce/src/core/test/ts/browser/commands/ContentLangTest.ts
```diff
@@ -0,0 +1,84 @@
...
+  }, [ Theme ]);
+
+  context('lang format plays nicely with other formats', () => {
+    beforeEach(() => {
+      const editor = hook.editor();
+      editor.setContent('<p>Hello world</p>');
+      TinySelections.setSelection(editor, [ 0, 0 ], 0, [ 0, 0 ], 'Hello world'.length);
+    });
+
+    it('TINY-6149: supports setting language then setting color', () => {
```
ltrouton, line 19: Could be misunderstanding but should the labels for this test and the below test be reversed?

ltrouton (approved):

modules/tinymce/src/core/test/ts/browser/commands/ContentLangTest.ts
ashfordneil: Good catch

PR #6963, opened 2021-06-23
TINY-7481: Added table header row buttons

HAFRMO:
Related Ticket:

Description of Changes:
Added table row header toggle button and command.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/tinymce/src/plugins/table/main/ts/selection/SelectionTargets.ts
```diff
@@ -53,6 +54,46 @@ interface ExtractedSelectionDetails {
...
+      });
+
+      return Arr.forall(usedRows, (rowIndex) =>
+        Arr.forall(warehouse.all[rowIndex].cells, (cell) =>
+          isElementHeader(cell.element)
+        )
+      );
+    });
+  });
+};
```
JamesToohey, line 95: Does this function need to be this complex? I might be misunderstanding but I believe it's checking if all cells within the selection exist within a header row or not? Do we need to check every cell in that row or can we safely assume that the state of the table is correct and than if one cell in a row is in a `thead` then they all are?

```suggestion
export const isEntireRowsHeaders = (editor: Editor, selections: Selections) => {
  const selectedCells = TableSelection.getCellsFromSelection(Util.getSelectionStart(editor), selections, Util.getIsRoot(editor));
  return Arr.forall(selectedCells, isElementHeader);
};
```

JamesToohey (commented):
If you create a 2x2 table and convert the first row into a header, the cursor moves to the second row (is this intended?) and now the button is disabled. If you then click out and back into that same cell the button is enabled

Screen Shot 2021-06-25 at 9 01 13 am

lnewson:
> If you create a 2x2 table and convert the first row into a header, the cursor moves to the second row (is this intended?)

This is a known bug with the row type and cell type action, see TINY-6666.

modules/tinymce/src/plugins/table/main/ts/api/Commands.ts
```diff
@@ -118,6 +119,26 @@ const registerCommands = (editor: Editor, actions: TableActions, cellSelection:
     });
   };
 
+  const toggleRowHeader = () => {
+    getSelectionStartCell(editor).each((startCell) => {
```
lnewson, line 123: This should be using the `performActionOnSelection` helper function instead of implementing duplicate code.

modules/tinymce/src/plugins/table/main/ts/selection/SelectionTargets.ts
```diff
@@ -42,6 +42,7 @@ export interface SelectionTargets {
   readonly onSetupUnmergeable: (api: UiApi) => () => void;
   readonly resetTargets: () => void;
   readonly onSetupTableWithCaption: (api: UiToggleApi) => () => void;
+  readonly onSetupTableHeaders: () => (api: Toolbar.ToolbarToggleButtonInstanceApi) => () => void;
```
lnewson, line 45: This should not be thunked and the type is incorrect here, as it should be `UiToggleApi`. eg:
```suggestion
  readonly onSetupTableHeaders: (api: UiToggleApi) => () => void;
```

modules/tinymce/src/plugins/table/main/ts/selection/SelectionTargets.ts
lnewson: As mentioned in my overall comment, we shouldn't even have a function for this as the logic already exists, so this is duplicate code.

modules/tinymce/src/plugins/table/main/ts/selection/SelectionTargets.ts
```diff
@@ -161,6 +202,17 @@ export const getSelectionTargets = (editor: Editor, selections: Selections): Sel
...
+
+  const onSetupTableHeaders = () => {
+    return (api: Toolbar.ToolbarToggleButtonInstanceApi): () => void => {
+      return onSetupWithToggle(
+        api,
+        (targets) => isCaption(targets.element),
+        () => isEntireRowsHeaders(editor, selections)
+      );
+    };
+  };
```
lnewson, line 214:
```suggestion
  const onSetupTableHeaders = (api: UiToggleApi) => {
    return onSetupWithToggle(api,
      (targets) => isCaption(targets.element),
      () => editor.queryCommandValue('mceTableRowType') === 'header'
    );
  };
```

lnewson (changes requested):
In general nice work here glancing over the changes, however I'm marking as needs changes sorry, as the issues I'd mentioned in the past PR/discussions haven't been resolved here.

So firstly, we shouldn't have the `isEntireRowsHeaders` function. As mentioned last time we already have a command/API for this (see the `mceTableRowType` query command/action) so we should be using that instead of duplicating logic.

The second issue is that in 5.5 we'd erred against introducing toggle APIs as there's a few different states here and then if people want to override commands it gets trickier as well. I discussed this with the team this morning at stand up and the decision is that we shouldn't have the `mceTableToggleRowHeader` command. Reason being this can be implemented in 3 lines of code outside instead, which avoids confusion and saves having to support yet another API/command. Here's what's required to implement it:
```
const currentType = editor.queryCommandValue('mceTableRowType');
const newType = currentType === 'header' ? 'body' : 'header';
editor.execCommand('mceTableRowType', false, { type: newType });
```

HAFRMO:
Tests not working yet, due to discrepancies between previous and new behavior. Setting up a meeting to discuss this.

modules/tinymce/src/plugins/table/test/ts/browser/command/TableRowHeaderCommandCellsTest.ts
```diff
@@ -0,0 +1,515 @@
...
+import { context, describe, it } from '@ephox/bedrock-client';
+import { TinyAssertions, TinyHooks, TinySelections } from '@ephox/mcagar';
+
+import Editor from 'tinymce/core/api/Editor';
+import Plugin from 'tinymce/plugins/table/Plugin';
+import Theme from 'tinymce/themes/silver/Theme';
+
+import { clickOnButton, makeCell } from '../../module/test/TableModifiersTestUtils';
+
+describe('browser.tinymce.plugins.table.command.TableRowHeaderCommandCellsTest', () => {
```
JamesToohey, line 11: Are these test names still accurate given we no longer have a new command?

modules/tinymce/src/plugins/table/test/ts/browser/command/TableRowHeaderCommandSectionCellsTest.ts
```diff
@@ -0,0 +1,517 @@
...
+import { context, describe, it } from '@ephox/bedrock-client';
+import { TinyAssertions, TinyHooks, TinySelections } from '@ephox/mcagar';
+
+import Editor from 'tinymce/core/api/Editor';
+import Plugin from 'tinymce/plugins/table/Plugin';
+import Theme from 'tinymce/themes/silver/Theme';
+
+import { clickOnButton, makeCell } from '../../module/test/TableModifiersTestUtils';
+
+describe('browser.tinymce.plugins.table.command.TableRowHeaderCommandSectionCellsTest', () => {
```
JamesToohey, line 11: Are these three test files identical except for the `table_header_type` setting in the config? 

It should be one test with something like this:

```
Arr.each([ 'sectionCells', 'cells', 'section' ], (headerType) => {
  context(`table_header_type: ${headerType}`, () => {
      plugins: 'table',
      indent: false,
      toolbar: 'tablerowheader',
      table_header_type: headerType,
    }, [ Plugin, Theme ], true);
  });
})
```

- [ ] Fix test duplication

JamesToohey (changes requested):
Requesting changes due to test duplication

modules/tinymce/src/plugins/table/test/ts/browser/command/TableRowHeaderCommandSectionCellsTest.ts
JamesToohey: I've just noticed the slight differences in behaviour, so the files aren't identical, but this still should be improved to reduce the significant duplication.

modules/tinymce/CHANGELOG.md
```diff
@@ -19,6 +19,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added a new `tablecellbordercolor` toolbar button and menu item to select table cell border colors, with an accompanying setting `table_cell_border_color_map` to customize the available values #TINY-7480
 - Added a new `tablecellbackgroundcolor` toolbar button and menu item to select table cell background colors, with an accompanying setting `table_cell_background_color_map` to customize the available values #TINY-7480
 - Added a new `initData` property to `fancymenuitem` to allow custom initialization data #TINY-7480
+- Added a new `tablerowheader` toolbar button and menu item to toggle row cells to and from header #TINY-7478
+- Added a new `mceTableToggleRowHeader` command that toggles selected rows to and from headers #TINY-7479
```
JamesToohey, line 23: And this shouldn't be here anymore

modules/tinymce/src/plugins/table/test/ts/browser/command/TableRowHeaderCommandSectionCellsTest.ts
JamesToohey: `table_header_type` is extensively covered in `SwitchTableSectionTest.ts`, `TableSectionApiTest.ts`, and others. Are these tests covering anything new?

HAFRMO: I don't see any really easy way to reduce duplication, unfortunately. The setups are different for the different tests, and the outcome is further different depending on the setting. It might be possible to make some array-arrangement, but I don't think it'd be any easier to read.

As for testing it, this does go through the query `mceTableRowType` and the command `mceTableRowType`, so it's not simply testing the `table_header_type`. Deleting the different variants aside, this would still test other things regarding selection and switching. We can probably see if it tests something the others does not to see if we still want the variants.

Name can probably be changed though.

JamesToohey: `SwitchTableSectionTest` does a lot of tests for switching, I suspect the internal logic for the commands is quite well tested, so we likely only need your new `TableRowHeaderUiTest.ts` to test the new buttons work

modules/tinymce/src/plugins/table/main/ts/ui/UiUtils.ts
```diff
@@ -74,11 +74,19 @@ const generateColorSelector = (editor: Editor, colorList: Menu.ChoiceMenuItemSpe
   }
 }];
 
+const changeRowHeader = (editor: Editor) =>
+  () => {
```
JamesToohey, line 78: I'm not sure about this being on a new line, interested to hear others' thoughts.

modules/tinymce/src/plugins/table/test/ts/module/test/TableModifiersTestUtils.ts
```diff
@@ -172,6 +172,36 @@ const pAssertStyleCanBeToggledOnAndOffWithoutCheckmarks = async (editor: Editor,
...
+    return attributes.join(' ');
+  };
+
+  const getScope = () => {
+    if (hasScope) {
+      return ' scope="col"';
+    } else {
+      return '';
+    }
+  };
```
JamesToohey, line 200: Can this just be a ternary in the return statement?

modules/tinymce/src/plugins/table/test/ts/module/test/TableModifiersTestUtils.ts
```diff
@@ -172,6 +172,36 @@ const pAssertStyleCanBeToggledOnAndOffWithoutCheckmarks = async (editor: Editor,
   await pAssertStyleCanBeToggledWithoutCheckmarks(editor, options, 'menuitem');
 };
 
+const makeCell = (type: string, content: string, hasScope?: boolean, isSelected?: boolean, isStart?: boolean, isEnd?: boolean) => {
```
JamesToohey, line 175: I just thought I'd note that I felt this made the tests more difficult to read than just inlining the selected attributes like we do elsewhere.

This is just so difficult to read:
`makeCell('td', '1-1', false, true, false, true)`

modules/tinymce/src/plugins/table/test/ts/browser/ui/TableRowHeaderUiTest.ts
```diff
@@ -0,0 +1,106 @@
...
+    base_url: '/project/tinymce/js/tinymce',
+    menu: {
+      table: { title: 'Table', items: 'tablerowheader' },
+    },
+    menubar: 'table',
+    toolbar: 'tablerowheader',
+    indent: false,
+  }, [ Plugin, Theme ], true);
+
+  const setEditorContentAndSelection = (editor: Editor, headBody: string) => {
```
JamesToohey, line 22: Can we type `headBody`? I also don't think we need to make it confusing by leaving the 't' out.

maybe `tableSection: 'thead' | 'tbody'`

JamesToohey (approved):

modules/tinymce/src/plugins/table/test/ts/module/test/TableModifiersTestUtils.ts
HAFRMO: I found it a lot easier to read and write that, rather than having the long pieces of attributes everywhere myself, not to mention having to put them everywhere.
But if people thinks it's easier to read all the selection-stuff, I'll put it there instead.

modules/tinymce/src/plugins/table/test/ts/browser/ui/TableRowHeaderUiTest.ts
HAFRMO: I did actually have it typed, but I wasn't sure people would like it, so I put it as a string. My bad.

modules/tinymce/src/plugins/table/test/ts/module/test/TableModifiersTestUtils.ts
lnewson: I'm with James on this one, this is hard to read and either passing an object (so you can see the names) or just passing the values would be better.

modules/tinymce/src/plugins/table/test/ts/module/test/TableModifiersTestUtils.ts
lnewson: This also doesn't need to be a function, it can just be a variable like so:
```
const scope = hasScope ? ` scope="col"` : '';
```

modules/tinymce/src/plugins/table/main/ts/ui/UiUtils.ts
lnewson: Yeah, normally we'd keep this on a single line.

lnewson (approved):

PR #6964, opened 2021-06-24
TINY-7589: Fix a parser bug when with mismatched quotes inside comments

### LINKED ISSUE #6758: noneditable plugin causing content duplication
With the `noneditable` plugin enabled, scripting DOM content changes inside a non-editable space can trigger duplication of that content outside the non-editable space.  In the example provided, the scripting is pared down to just deleting existing blocks and the deleted block is what gets duplicated, but in larger cases the duplication can be several blocks together.

Expected behavior: only the scripted DOM changes occur.
Actual behavior: additional content is injected/duplicated

Some details known to be required to reproduce the bug:
 - a non-editable block that is not the body
 - an apostrophe somewhere in text content within the document

And some details that are likely required:
 - A control element the user must click that is within the fragment getting manipulated

[Demo](https://fiddle.tiny.cloud/rAhaab)

TinyMCE: 5.4.2 and 5.7.1 (the bug was absent in 4.5.7)
Browsers: Edge 89, Chrome 90, Opera 75 and others (no browser found without the bug)

There is a workaround, which involves explicitly selecting the element being manipulated before invoking the scripted change.  I don't know if that _should_ be required, but the fact that an unrelated apostrophe can trigger the altered behavior seems to me like indication of an obscure internal issue.

ashfordneil:
Related Ticket: TINY-7589

Description of Changes:
* When encountering `data-mce-bogus="all"` elements, the SaxParser skips over them entirely instead of trying to parse them
* This skipping logic tried to match up quote characters ('/") inside of HTML comments
* This means that HTML comments with an odd number of quote characters wouldn't end at the right spot
We now skip over comments more correctly.

Pre-checks:
* [x] Changelog entry added

Before merging:
* [x] Milestone set
* [x] In looking at the code, I found that we'll have some similar bugs with CDATA elements — make sure we Jira this up and add a TODO before merging

GitHub issues (if applicable): Fixes #6758


modules/tinymce/src/core/main/ts/api/html/SaxParser.ts
```diff
@@ -118,40 +118,67 @@ const isInvalidUri = (settings: SaxParserSettings, uri: string, tagName: string)
   }
 };
 
+interface LightweightTag {
```
ashfordneil, line 121: Anyone want to help me find better names for... all of this? I'd prefer to have not split this out into a function in the first place, because it's really just an implementation detail for `findMatchingEndTagIndex`, but unfortunately JS won't let me assign to the results of loops or complex if statements.

modules/tinymce/src/core/main/ts/api/html/SaxParser.ts
```diff
@@ -118,40 +118,67 @@ const isInvalidUri = (settings: SaxParserSettings, uri: string, tagName: string)
...
+ * @method findMatchingEndTagIndex
  * @param {tinymce.html.Schema} schema Schema instance to use to match short ended elements.
  * @param {String} html HTML string to find the end tag in.
  * @param {Number} startIndex Index to start searching at should be after the start tag.
  * @return {Number} Index of the end tag.
  */
-const findEndTagIndex = (schema: Schema, html: string, startIndex: number): number => {
-  let count = 1, index, matches;
-
+const findMatchingEndTagIndex = (schema: Schema, html: string, startIndex: number): number => {
```
ashfordneil, line 174: Happy to revert this, but I thought it might be useful to stress that we're looking for matching tags here. The name it's exported with hasn't changed, and the old name wasn't equal to the exported name anyway.

modules/tinymce/src/core/main/ts/api/html/SaxParser.ts
```diff
@@ -118,40 +118,70 @@ const isInvalidUri = (settings: SaxParserSettings, uri: string, tagName: string)
...
+
+// This finds a tag, but with very small amounts of detail - used only in findMatchingEndTagIndex
+// Instead of returning what type of tag it is, we just say "is it an open tag, a close tag, or a self-closing tag (like a br, or comment)"
+// And we also don't bother returning the start index of the tag, just the end
+const findLightweightTag = (html: string, startIndex: number, shortEndedElements: SchemaMap): LightweightTag | null => {
+  const startTagRegExp = /<([!?\/])?([A-Za-z0-9\-_:.]+)/;
+  const endTagRegExp = /^(?:\s(?:[^'">]+(?:"[^"]*"|'[^']*'))*[^"'>]*(?:"[^">]*|'[^'>]*)?|\s*|\/)>/;
+
+  while (true) {
+    const startMatch = startTagRegExp.exec(html.slice(startIndex));
```
lnewson, line 134: As discussed on slack yesterday, this should **not** be using slice. Use the old `lastIndex` instead please.

lnewson (changes requested):
Marking as needs changes as I mentioned in the call we explicitly use lastIndex for both performance and memory. Please don't just skip over things like that, as I'd already said that needed fixing.

modules/tinymce/src/core/main/ts/api/html/SaxParser.ts
```diff
@@ -118,40 +118,70 @@ const isInvalidUri = (settings: SaxParserSettings, uri: string, tagName: string)
...
+    if (startMatch === null) {
+      return null;
+    }
+    const endOfStart = startIndex + startMatch.index + startMatch[0].length;
+
+    if (startMatch[0] === '<!--') { // it's a comment
```
lnewson, line 140: This won't work for a number of cases... it needs to be `<!` to handle bogus comments. From there we need to see if `--` follows and if it does, then it's not bogus, however if that is missing then it is bogus.

modules/tinymce/src/core/main/ts/api/html/SaxParser.ts
```diff
@@ -118,40 +118,70 @@ const isInvalidUri = (settings: SaxParserSettings, uri: string, tagName: string)
...
 
+interface LightweightTag {
+  readonly tagDepth: number;
+  readonly end: number;
+}
+
+// This finds a tag, but with very small amounts of detail - used only in findMatchingEndTagIndex
+// Instead of returning what type of tag it is, we just say "is it an open tag, a close tag, or a self-closing tag (like a br, or comment)"
+// And we also don't bother returning the start index of the tag, just the end
+const findLightweightTag = (html: string, startIndex: number, shortEndedElements: SchemaMap): LightweightTag | null => {
```
lnewson, line 129: Another minor issue here is going to be less performance than the old code and parsing is a hot code path. It does a lot of additional scope switching and object creation for the return values. As shown recently by the Boulder work that does make a difference for hot code paths, especially when it starts triggering GC collections (which is also related to the memory issue and why why shouldn't be using slice).

As such, did this really need to be split out? I don't believe so and am assuming it was only done for readability.

modules/tinymce/src/core/main/ts/api/html/SaxParser.ts
```diff
@@ -119,38 +119,55 @@ const isInvalidUri = (settings: SaxParserSettings, uri: string, tagName: string)
...
+    // ideally, we only want to run through this the once - but sometimes the startTagRegExp will give us false positives (things that begin
+    // like tags, but don't end like them) and so we might need to bump up its lastIndex and try again.
+    while (true) {
+      const startMatch = startTagRegExp.exec(html);
+      if (startMatch === null) {
+        // doesn't matter what count is, we've run out of HTML tags
+        return index;
       }
+      if (startMatch[1] === '!') {
+        index = findCommentEndIndex(html, startMatch[2] === '--', startTagRegExp.lastIndex);
```
lnewson, line 151: This logic is incorrect as the expression is back to front. It'll treat valid comments as being bogus currently.
```suggestion
        index = findCommentEndIndex(html, startMatch[2] !== '--', startTagRegExp.lastIndex);
```

- [ ] Add a test to make sure this is working correctly

modules/tinymce/src/core/main/ts/api/html/SaxParser.ts
```diff
@@ -119,38 +119,55 @@ const isInvalidUri = (settings: SaxParserSettings, uri: string, tagName: string)
...
+
+    // ideally, we only want to run through this the once - but sometimes the startTagRegExp will give us false positives (things that begin
+    // like tags, but don't end like them) and so we might need to bump up its lastIndex and try again.
+    while (true) {
+      const startMatch = startTagRegExp.exec(html);
+      if (startMatch === null) {
+        // doesn't matter what count is, we've run out of HTML tags
+        return index;
       }
+      if (startMatch[1] === '!') {
```
lnewson, line 150: Else if here please, as this is a continuation of the branching, not something new.

lnewson: Also, do we care that this'll incorrectly run for `<!DOCTYPE`? We shouldn't ever run into that when trying to find an end tag though.

modules/tinymce/src/core/main/ts/api/html/SaxParser.ts
```diff
@@ -119,38 +119,55 @@ const isInvalidUri = (settings: SaxParserSettings, uri: string, tagName: string)
...
+  const startTagRegExp = /<([!?\/])?([A-Za-z0-9\-_:.]+)/g;
+  const endTagRegExp = /(?:\s(?:[^'">]+(?:"[^"]*"|'[^']*'))*[^"'>]*(?:"[^">]*|'[^'>]*)?|\s*|\/)>/g;
   const shortEndedElements = schema.getShortEndedElements();
-  const tokenRegExp = /<([!?\/])?([A-Za-z0-9\-_:.]+)(\s(?:[^'">]+(?:"[^"]*"|'[^']*'))*[^"'>]*(?:"[^">]*|'[^'>]*)?|\s*|\/)>/g;
-  tokenRegExp.lastIndex = index = startIndex;
-
-  while ((matches = tokenRegExp.exec(html))) {
-    index = tokenRegExp.lastIndex;
-
-    if (matches[1] === '/') { // End element
```
lnewson, line 142: Please put these comments back, as they help to make it clearer what was happening.

- [ ] Add comments back about what the various branches handle

modules/tinymce/src/core/main/ts/api/html/SaxParser.ts
lnewson: Another issue here is what we discussed yesterday whereby this won't be supporting CDATA. Please do what we said there and log a Jira and add a TODO comment here.

modules/tinymce/src/core/main/ts/api/html/SaxParser.ts
lnewson: Looking more this also really needs to be a startsWith check as well instead, as you could have something like `<!--test-->`. In which case you'd have more than just -- in the match.

modules/tinymce/src/core/main/ts/api/html/SaxParser.ts
```diff
@@ -119,38 +119,60 @@ const isInvalidUri = (settings: SaxParserSettings, uri: string, tagName: string)
...
+const findMatchingEndTagIndex = (schema: Schema, html: string, startIndex: number): number => {
+  // TODO: TINY-7658: this regex does not support CDATA
+  const startTagRegExp = /<([!?\/])?([A-Za-z0-9\-_:.]+)/g;
+  const endTagRegExp = /(?:\s(?:[^'">]+(?:"[^"]*"|'[^']*'))*[^"'>]*(?:"[^">]*|'[^'>]*)?|\s*|\/)>/g;
   const shortEndedElements = schema.getShortEndedElements();
-  const tokenRegExp = /<([!?\/])?([A-Za-z0-9\-_:.]+)(\s(?:[^'">]+(?:"[^"]*"|'[^']*'))*[^"'>]*(?:"[^">]*|'[^'>]*)?|\s*|\/)>/g;
-  tokenRegExp.lastIndex = index = startIndex;
+  let count = 1, index = startIndex;
+
+  // keep finding HTML tags (opening, closing, or neither like comments or <br>s
```
lnewson, line 139:
```suggestion
  // keep finding HTML tags (opening, closing, or neither like comments or <br>s)
```
There's a missing end parenthesis here. I'm assuming this is where you meant it to be.

lnewson (approved):
Looks good now, thanks for fixing up those issues and adding further tests :+1:

ltrouton (approved):

PR #6884, opened 2021-06-01
Moved some APIs around to make `Optional<T>` covariant with respect to T

ashfordneil:
Related Ticket: N/A

Description of Changes:
* Removed .is and .equals and .equals_
* Added .is and .equals to Optionals.ts (note the s)
* Added a test to make sure that optionals are now covariant

Pre-checks:
* [x] Changelog entry added
  * [x] Major version bump of every library we have?
* [x] Tests have been added (if applicable)
* [x] Branch prefixed with `feature/` for new features (if applicable)

Review:
* [x] Milestone set
* [x] Review comments resolved

GitHub issues (if applicable): N/A

PR #6968, opened 2021-06-27
Moved some APIs around to make Result<T, E> covariant

lnewson:
Related Ticket: N/A

Description of Changes:
* Removed .is from Result.ts
* Added .is to Results.ts (note the s)
* Added a test to make sure that results are now covariant
* Enabled `strictFunctionTypes` for katamari

Note: This is a follow up to the work done for Optionals in https://github.com/tinymce/tinymce/pull/6884 by @ashfordneil.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


tiny-james (approved):

JamesToohey (approved):

ltrouton (approved):

modules/katamari/src/test/ts/atomic/api/data/ResultCovarianceAssertionTest.ts
```diff
@@ -0,0 +1,29 @@
...
+
+interface Cat extends Animal {
+  readonly length: number;
+}
+
+// https://en.wikipedia.org/wiki/Covariance_and_contravariance_(computer_science)
+//
+// We need to make sure that Result<T, E> is, and remains, covariant with
+// respect to T and E.
+describe('atomic.katamari.api.data.ResultCovarianceAssertionTest', () => {
```
ashfordneil, line 19: Might as well merge the pair of these into a single `CovarianceAssertionTest` and get Optional and Result in the same group — 99% of the code is the same.

Also, does this test still pass with strict mode on (I know everything else in katamari breaks, just want to make sure this file in particular doesn't have any errors so we don't get bitten like last time)

modules/katamari/src/test/ts/module/ephox/katamari/test/arb/ArbDataTypes.ts
```diff
@@ -7,10 +7,10 @@ import { Result } from 'ephox/katamari/api/Result';
 
 type Arbitrary<A> = fc.Arbitrary<A>;
 
-export const arbResultError = <A, E> (arbE: Arbitrary<E>): Arbitrary<Result<A, E>> =>
+export const arbResultError = <A = never, E = any> (arbE: Arbitrary<E>): Arbitrary<Result<A, E>> =>
```
ashfordneil, line 10: No need for an `any` here
```suggestion
export const arbResultError = <E, A = never> (arbE: Arbitrary<E>): Arbitrary<Result<A, E>> =>
```

modules/katamari/src/main/ts/ephox/katamari/api/Result.ts
```diff
@@ -103,7 +97,7 @@ const value = <T, E = any>(o: T): Result<T, E> => {
   };
 };
 
-const error = <T = any, E = any>(message: E): Result<T, E> => {
+const error = <T = never, E = any>(message: E): Result<T, E> => {
```
ashfordneil, line 100: Awesome to have this `any` removed — we should try to get the other one on this line in the next slack time PR

ashfordneil (approved):

HAFRMO (approved):

modules/katamari/src/test/ts/module/ephox/katamari/test/arb/ArbDataTypes.ts
lnewson: I'm not doing this, as it requires switching the order which is going to be confusing as the order for results is always `<Value, Error>`.

modules/katamari/src/test/ts/atomic/api/data/ResultCovarianceAssertionTest.ts
lnewson: Have merged the 2 tests into a single file. Also yes, I double checked and there's no errors in this test when strict mode is enabled.

PR #6972, opened 2021-06-30
TINY-4589: Rtl/ltl attributes sometimes gets applied to list items

diogo405:
Related Ticket:

Description of Changes:
* when `rtl` or `ltr` is clicked on a list item (li) it applies `dir` to parent (`ul` or `ol`).

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -32,6 +32,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Changed the load order so that the content css gets loaded before the editor gets populated with contents. #TINY-7249
 
 ### Fixed
+- Rtl/ltl attributes sometimes gets applied to list items #TINY-4589
 - `editor.fire` was incorrectly mutating the original `args` provided #TINY-3254
 - The `SetContent` event contained the incorrect `content` when using the `editor.selection.setContent()` API #TINY-3254
 - The editor content could be edited after calling `setProgressState(true)` in iframe mode #TINY-7373
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -32,7 +32,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Changed the load order so that the content css gets loaded before the editor gets populated with contents. #TINY-7249
 
 ### Fixed
-- Rtl/ltl attributes sometimes gets applied to list items #TINY-4589
+- The `dir` attribute was being incorrectly applied to list items #TINY-4589
 - `editor.fire` was incorrectly mutating the original `args` provided #TINY-3254
 - The `SetContent` event contained the incorrect `content` when using the `editor.selection.setContent()` API #TINY-3254
 - The editor content could be edited after calling `setProgressState(true)` in iframe mode #TINY-7373
--- a/modules/tinymce/src/plugins/directionality/main/ts/api/Commands.ts
+++ b/modules/tinymce/src/plugins/directionality/main/ts/api/Commands.ts
@@ -9,11 +9,11 @@ import * as Direction from '../core/Direction';
 
 const register = (editor) => {
   editor.addCommand('mceDirectionLTR', () => {
-    Direction.setDir(editor, 'ltr');
+    Direction.setDir(editor, 'ltr' as Direction.Dir);
   });
 
   editor.addCommand('mceDirectionRTL', () => {
-    Direction.setDir(editor, 'rtl');
+    Direction.setDir(editor, 'rtl' as Direction.Dir);
   });
 };
 
--- a/modules/tinymce/src/plugins/directionality/main/ts/api/Commands.ts
+++ b/modules/tinymce/src/plugins/directionality/main/ts/api/Commands.ts
@@ -9,11 +9,11 @@ import * as Direction from '../core/Direction';
 
 const register = (editor) => {
   editor.addCommand('mceDirectionLTR', () => {
-    Direction.setDir(editor, 'ltr' as Direction.Dir);
+    Direction.setDir(editor, 'ltr');
   });
 
   editor.addCommand('mceDirectionRTL', () => {
-    Direction.setDir(editor, 'rtl' as Direction.Dir);
+    Direction.setDir(editor, 'rtl');
   });
 };
 
--- a/modules/tinymce/src/plugins/directionality/main/ts/core/Direction.ts
+++ b/modules/tinymce/src/plugins/directionality/main/ts/core/Direction.ts
@@ -10,7 +10,7 @@ import { Traverse, Attribute, SugarElement } from '@ephox/sugar';
 
 import Editor from 'tinymce/core/api/Editor';
 
-type Dir = 'rtl' | 'ltl';
+type Dir = 'rtl' | 'ltr';
 
 const setDir = (editor: Editor, dir: Dir) => {
   const selectedBlocks = editor.selection.getSelectedBlocks();
@@ -43,5 +43,5 @@ const setDirAttr = (editor: Editor, element: SugarElement, dir: Dir): void => {
 const isListItem = (element: SugarElement<Element>): boolean => element.dom.nodeName === 'LI';
 
 export {
-  setDir, Dir
+  setDir
 };
--- a/modules/tinymce/src/plugins/directionality/main/ts/core/Direction.ts
+++ b/modules/tinymce/src/plugins/directionality/main/ts/core/Direction.ts
@@ -5,21 +5,26 @@
  * For commercial licenses see https://www.tiny.cloud/
  */
 
-import { Arr } from '@ephox/katamari';
-import { Traverse, Attribute, SugarElement } from '@ephox/sugar';
+import { Arr, Optional, Type } from '@ephox/katamari';
+import { Traverse, Attribute, SugarElement, SugarNode } from '@ephox/sugar';
 
 import Editor from 'tinymce/core/api/Editor';
 
 type Dir = 'rtl' | 'ltr';
 
+const getParentElement = (element: SugarElement<Element>): Optional<SugarElement<Element>> => Traverse.parent(element).filter(SugarNode.isElement);
+
 const setDir = (editor: Editor, dir: Dir) => {
   const selectedBlocks = editor.selection.getSelectedBlocks();
   Arr.each(selectedBlocks, (block) => {
     const sugarBlock = SugarElement.fromDom(block);
-    const blockParent = Traverse.parent(sugarBlock);
-    blockParent.each((blockParent: SugarElement<Element>) => {
+    const blockParent = getParentElement(sugarBlock);
+    blockParent.each((blockParent) => {
       const blockParentDirection = Attribute.get(blockParent, 'dir');
-      if ((blockParentDirection === undefined) || (blockParentDirection !== dir)) {
+      if (Type.isUndefined(blockParentDirection) ||
+        Type.isNull(blockParentDirection) ||
+        blockParentDirection.trim() === '' ||
+        blockParentDirection !== dir) {
         setDirAttr(editor, sugarBlock, dir);
       } else { // if parent and child dir are going to be the same then remove it from child
         Attribute.remove(sugarBlock, 'dir');
@@ -29,10 +34,12 @@ const setDir = (editor: Editor, dir: Dir) => {
   });
 };
 
+const isListItem = SugarNode.isTag('li');
+
 const setDirAttr = (editor: Editor, element: SugarElement, dir: Dir): void => {
   if (isListItem(element)) {
-    const list = Traverse.parent(element);
-    list.each((l: SugarElement<Element>) => Attribute.set(l, 'dir', dir));
+    const list = getParentElement(element);
+    list.each((l) => Attribute.set(l, 'dir', dir));
   } else {
     Attribute.set(element, 'dir', dir);
   }
@@ -40,7 +47,6 @@ const setDirAttr = (editor: Editor, element: SugarElement, dir: Dir): void => {
   editor.nodeChanged();
 };
 
-const isListItem = (element: SugarElement<Element>): boolean => element.dom.nodeName === 'LI';
 
 export {
   setDir
--- a/modules/tinymce/src/plugins/directionality/main/ts/core/Direction.ts
+++ b/modules/tinymce/src/plugins/directionality/main/ts/core/Direction.ts
@@ -5,40 +5,65 @@
  * For commercial licenses see https://www.tiny.cloud/
  */
 
-import { Arr, Optional, Type } from '@ephox/katamari';
+import { Arr, Optional } from '@ephox/katamari';
 import { Traverse, Attribute, SugarElement, SugarNode } from '@ephox/sugar';
 
 import Editor from 'tinymce/core/api/Editor';
 
 type Dir = 'rtl' | 'ltr';
 
-const getParentElement = (element: SugarElement<Element>): Optional<SugarElement<Element>> => Traverse.parent(element).filter(SugarNode.isElement);
+const isListItem = SugarNode.isTag('li');
+
+const getParentElementWithDir = (element: SugarElement<Element>): Optional<SugarElement<Element>> => {
+  const parents = Traverse.parents(element).filter((e) => {
+    return SugarNode.isHTMLElement(e) &&
+      !isListItem(e) &&
+      (Attribute.get(e, 'dir') === 'ltr' || Attribute.get(e, 'dir') === 'rtl');
+  });
+
+  return parents.length > 0 ? Optional.from(parents[0] as SugarElement<Element>) : Optional.none();
+};
 
 const setDir = (editor: Editor, dir: Dir) => {
   const selectedBlocks = editor.selection.getSelectedBlocks();
-  Arr.each(selectedBlocks, (block) => {
-    const sugarBlock = SugarElement.fromDom(block);
-    const blockParent = getParentElement(sugarBlock);
-    blockParent.each((blockParent) => {
-      const blockParentDirection = Attribute.get(blockParent, 'dir');
-      if (Type.isUndefined(blockParentDirection) ||
-        Type.isNull(blockParentDirection) ||
-        blockParentDirection.trim() === '' ||
-        blockParentDirection !== dir) {
-        setDirAttr(editor, sugarBlock, blockParent, dir);
-      } else { // if parent and child dir are going to be the same then remove it from child
-        Attribute.remove(sugarBlock, 'dir');
-        editor.nodeChanged();
+  if (selectedBlocks.length > 0) {
+    Arr.each(selectedBlocks, (block) => {
+      const sugarBlock = SugarElement.fromDom(block);
+      if (isListItem(sugarBlock)) {
+        const sugarBlockParent = Traverse.parent(sugarBlock).filter(SugarNode.isHTMLElement);
+        sugarBlockParent.each((parent) => {
+          const sugarBlockParentWithDir = getParentElementWithDir(parent);
+          sugarBlockParentWithDir.fold(
+            () => {
+              Attribute.remove(sugarBlock, 'dir'); // li should not have dir
+              Attribute.set(parent, 'dir', dir);
+            },
+            (parentWithDir) => {
+              if (Attribute.get(parentWithDir, 'dir') === dir) {
+                Attribute.remove(parent, 'dir');
+              } else {
+                Attribute.set(parent, 'dir', dir);
+              }
+            }
+          );
+        });
+      } else {
+        const sugarBlockParentWithDir = getParentElementWithDir(sugarBlock);
+        sugarBlockParentWithDir.fold(
+          () => Attribute.set(sugarBlock, 'dir', dir),
+          (parentWithDir) => {
+            if (Attribute.get(parentWithDir, 'dir') === dir) {
+              Attribute.remove(sugarBlock, 'dir');
+            } else {
+              Attribute.set(sugarBlock, 'dir', dir);
+            }
+          }
+        );
       }
     });
-  });
-};
 
-const isListItem = SugarNode.isTag('li');
-
-const setDirAttr = (editor: Editor, element: SugarElement<Element>, parent: SugarElement<Element>, dir: Dir): void => {
-  Attribute.set(isListItem(element) ? parent : element, 'dir', dir);
-  editor.nodeChanged();
+    editor.nodeChanged();
+  }
 };
 
 export {
--- a/modules/tinymce/src/plugins/directionality/main/ts/core/Direction.ts
+++ b/modules/tinymce/src/plugins/directionality/main/ts/core/Direction.ts
@@ -6,62 +6,47 @@
  */
 
 import { Arr, Optional } from '@ephox/katamari';
-import { Traverse, Attribute, SugarElement, SugarNode } from '@ephox/sugar';
+import { Traverse, Attribute, SugarElement, SugarNode, SelectorFind, Direction, SelectorFilter } from '@ephox/sugar';
 
 import Editor from 'tinymce/core/api/Editor';
 
 type Dir = 'rtl' | 'ltr';
 
-const isListItem = SugarNode.isTag('li');
-
-const getParentElementWithDir = (element: SugarElement<Element>): Optional<SugarElement<Element>> => {
-  const parents = Traverse.parents(element).filter((e) => {
-    return SugarNode.isHTMLElement(e) &&
-      !isListItem(e) &&
-      (Attribute.get(e, 'dir') === 'ltr' || Attribute.get(e, 'dir') === 'rtl');
-  });
-
-  return parents.length > 0 ? Optional.from(parents[0] as SugarElement<Element>) : Optional.none();
+const getParentElement = (element: SugarElement<Element>): Optional<SugarElement<Element>> => {
+  return Traverse.parent(element).filter(SugarNode.isElement);
 };
 
-const setDir = (editor: Editor, dir: Dir) => {
+// if the block is a list item, we need to get the parent of the list itself
+const getNormalizedBlock = (element: SugarElement<Element>, isListItem: boolean): SugarElement<Element> => {
+  const normalizedElement = isListItem ? SelectorFind.ancestor(element, 'ol,ul') : Optional.some(element);
+  return normalizedElement.getOr(element);
+};
+
+const isListItem = SugarNode.isTag('li');
+
+const setDir = (editor: Editor, dir: Dir): void => {
   const selectedBlocks = editor.selection.getSelectedBlocks();
   if (selectedBlocks.length > 0) {
     Arr.each(selectedBlocks, (block) => {
       const sugarBlock = SugarElement.fromDom(block);
-      if (isListItem(sugarBlock)) {
-        const sugarBlockParent = Traverse.parent(sugarBlock).filter(SugarNode.isHTMLElement);
-        sugarBlockParent.each((parent) => {
-          const sugarBlockParentWithDir = getParentElementWithDir(parent);
-          sugarBlockParentWithDir.fold(
-            () => {
-              Attribute.remove(sugarBlock, 'dir'); // li should not have dir
-              Attribute.set(parent, 'dir', dir);
-            },
-            (parentWithDir) => {
-              if (Attribute.get(parentWithDir, 'dir') === dir) {
-                Attribute.remove(parent, 'dir');
-              } else {
-                Attribute.set(parent, 'dir', dir);
-              }
-            }
-          );
-        });
-      } else {
-        const sugarBlockParentWithDir = getParentElementWithDir(sugarBlock);
-        sugarBlockParentWithDir.fold(
-          () => Attribute.set(sugarBlock, 'dir', dir),
-          (parentWithDir) => {
-            if (Attribute.get(parentWithDir, 'dir') === dir) {
-              Attribute.remove(sugarBlock, 'dir');
-            } else {
-              Attribute.set(sugarBlock, 'dir', dir);
-            }
-          }
-        );
-      }
-    });
+      const isSugarBlockListItem = isListItem(sugarBlock);
+      const normalizedBlock = getNormalizedBlock(sugarBlock, isSugarBlockListItem);
+      const normalizedBlockParent = getParentElement(normalizedBlock);
+      normalizedBlockParent.each((parent) => {
+        const parentDirection = Direction.getDirection(parent);
+        if (parentDirection !== dir) {
+          Attribute.set(normalizedBlock, 'dir', dir);
+        } else if (Direction.getDirection(normalizedBlock) !== dir) {
+          Attribute.remove(normalizedBlock, 'dir');
+        }
 
+        // remove dir attr from list children
+        if (isSugarBlockListItem) {
+          const listItems = SelectorFilter.children(normalizedBlock, 'li[dir]');
+          Arr.each(listItems, (listItem) => Attribute.remove(listItem, 'dir'));
+        }
+      });
+    });
     editor.nodeChanged();
   }
 };
--- a/modules/tinymce/src/plugins/directionality/main/ts/core/Direction.ts
+++ b/modules/tinymce/src/plugins/directionality/main/ts/core/Direction.ts
@@ -36,7 +36,7 @@ const setDir = (editor: Editor, dir: Dir) => {
 
 const isListItem = SugarNode.isTag('li');
 
-const setDirAttr = (editor: Editor, element: SugarElement, dir: Dir): void => {
+const setDirAttr = (editor: Editor, element: SugarElement<Element>, dir: Dir): void => {
   if (isListItem(element)) {
     const list = getParentElement(element);
     list.each((l) => Attribute.set(l, 'dir', dir));
--- a/modules/tinymce/src/plugins/directionality/main/ts/core/Direction.ts
+++ b/modules/tinymce/src/plugins/directionality/main/ts/core/Direction.ts
@@ -17,7 +17,7 @@ const setDir = (editor: Editor, dir: Dir) => {
   Arr.each(selectedBlocks, (block) => {
     const sugarBlock = SugarElement.fromDom(block);
     const blockParent = Traverse.parent(sugarBlock);
-    blockParent.map((blockParent: SugarElement<Element>) => {
+    blockParent.each((blockParent: SugarElement<Element>) => {
       const blockParentDirection = Attribute.get(blockParent, 'dir');
       if ((blockParentDirection === undefined) || (blockParentDirection !== dir)) {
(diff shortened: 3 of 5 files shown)


modules/tinymce/src/plugins/directionality/main/ts/core/Direction.ts
```diff
@@ -5,28 +5,43 @@
...
       }
     });
+  });
+};
 
-    editor.nodeChanged();
+const setDirAttr = (editor: Editor, element: SugarElement, dir: Dir): void => {
+  if (isListItem(element)) {
+    const list = Traverse.parent(element);
+    list.map((l: SugarElement<Element>) => Attribute.set(l, 'dir', dir));
```
JamesToohey, line 35: Like earlier, this can be `.each` and we should be filtering

modules/tinymce/src/plugins/directionality/main/ts/core/Direction.ts
```diff
@@ -5,28 +5,43 @@
...
+  } else {
+    Attribute.set(element, 'dir', dir);
   }
+
+  editor.nodeChanged();
 };
 
+const isListItem = (element: SugarElement<Element>): boolean => element.dom.nodeName === 'LI';
```
JamesToohey, line 43: We normally like to put helper functions above where they are used.

Also this can be simplified using `SugarNode.isTag()`

```suggestion
const isListItem = SugarNode.isTag('li');
```

modules/tinymce/src/plugins/directionality/main/ts/core/Direction.ts
```diff
@@ -5,28 +5,43 @@
...
 
-  if (blocks.length) {
-    curDir = dom.getAttrib(blocks[0], 'dir');
+type Dir = 'rtl' | 'ltl';
 
-    Tools.each(blocks, (block) => {
-      // Add dir to block if the parent block doesn't already have that dir
-      if (!dom.getParent(block.parentNode, '*[dir="' + dir + '"]', dom.getRoot())) {
-        dom.setAttrib(block, 'dir', curDir !== dir ? dir : null);
+const setDir = (editor: Editor, dir: Dir) => {
```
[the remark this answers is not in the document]
JamesToohey: Either way I probably wouldn't export it

modules/tinymce/src/plugins/directionality/main/ts/core/Direction.ts
```diff
@@ -5,28 +5,43 @@
...
+  const selectedBlocks = editor.selection.getSelectedBlocks();
+  Arr.each(selectedBlocks, (block) => {
+    const sugarBlock = SugarElement.fromDom(block);
+    const blockParent = Traverse.parent(sugarBlock);
+    blockParent.map((blockParent: SugarElement<Element>) => {
+      const blockParentDirection = Attribute.get(blockParent, 'dir');
+      if ((blockParentDirection === undefined) || (blockParentDirection !== dir)) {
+        setDirAttr(editor, sugarBlock, dir);
+      } else { // if parent and child dir are going to be the same then remove it from child
```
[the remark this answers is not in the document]
JamesToohey: I suspect this may be causing a bug in its current state, if so we should probably add a test for it (what happens if the parent does not have a `dir` attribute?)

modules/tinymce/src/plugins/directionality/main/ts/api/Commands.ts
```diff
@@ -9,11 +9,11 @@ import * as Direction from '../core/Direction';
 
 const register = (editor) => {
   editor.addCommand('mceDirectionLTR', () => {
-    Direction.setDir(editor, 'ltr');
+    Direction.setDir(editor, 'ltr' as Direction.Dir);
```
JamesToohey, line 12: We shouldn't need to type cast here

modules/tinymce/CHANGELOG.md
```diff
@@ -32,6 +32,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Changed the load order so that the content css gets loaded before the editor gets populated with contents. #TINY-7249
 
 ### Fixed
+- Rtl/ltl attributes sometimes gets applied to list items #TINY-4589
```
JamesToohey, line 35: Maybe

```suggestion
- The `dir` attribute was being incorrectly applied to list items #TINY-4589
```

modules/tinymce/src/plugins/directionality/main/ts/core/Direction.ts
```diff
@@ -5,28 +5,43 @@
...
 
-const setDir = (editor: Editor, dir: string) => {
-  const dom = editor.dom;
-  let curDir;
-  const blocks = editor.selection.getSelectedBlocks();
+import Editor from 'tinymce/core/api/Editor';
 
-  if (blocks.length) {
-    curDir = dom.getAttrib(blocks[0], 'dir');
+type Dir = 'rtl' | 'ltl';
```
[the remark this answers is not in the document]
JamesToohey: (also there is a typo here, it should be 'ltr')

modules/tinymce/src/plugins/directionality/main/ts/api/Commands.ts
JamesToohey: This type casting was also hiding a typo, there's normally very few reasons we need to cast

modules/tinymce/src/plugins/directionality/main/ts/core/Direction.ts
JamesToohey: I suggest using a helper function for this because we will be doing it twice

JamesToohey (commented):
Well done! Congrats on your first TinyMCE PR

modules/tinymce/src/plugins/directionality/main/ts/core/Direction.ts
```diff
@@ -5,28 +5,43 @@
...
+  Arr.each(selectedBlocks, (block) => {
+    const sugarBlock = SugarElement.fromDom(block);
+    const blockParent = Traverse.parent(sugarBlock);
+    blockParent.map((blockParent: SugarElement<Element>) => {
+      const blockParentDirection = Attribute.get(blockParent, 'dir');
+      if ((blockParentDirection === undefined) || (blockParentDirection !== dir)) {
+        setDirAttr(editor, sugarBlock, dir);
+      } else { // if parent and child dir are going to be the same then remove it from child
+        Attribute.remove(sugarBlock, 'dir');
+        editor.nodeChanged();
```
lnewson, line 26: This is going to cause incorrect behaviour. The `NodeChange` event should only be fired once. This current change will cause it to fire once per selected block element. As such, we really need to be doing something like it was previous, i.e.:
```
if (selectedBlocks.length >= 0) {
  Arr.each(selectedBlocks, ...);
  editor.nodeChanged();
}
```

modules/tinymce/src/plugins/directionality/main/ts/core/Direction.ts
```diff
@@ -5,28 +5,43 @@
...
-      // Add dir to block if the parent block doesn't already have that dir
-      if (!dom.getParent(block.parentNode, '*[dir="' + dir + '"]', dom.getRoot())) {
-        dom.setAttrib(block, 'dir', curDir !== dir ? dir : null);
+const setDir = (editor: Editor, dir: Dir) => {
+  const selectedBlocks = editor.selection.getSelectedBlocks();
+  Arr.each(selectedBlocks, (block) => {
+    const sugarBlock = SugarElement.fromDom(block);
+    const blockParent = Traverse.parent(sugarBlock);
+    blockParent.map((blockParent: SugarElement<Element>) => {
+      const blockParentDirection = Attribute.get(blockParent, 'dir');
```
lnewson, line 21: This is a regression from the previous behaviour (though the previous behaviour also had a bug). The `editor.dom.getParent` function will get any ancestor that matches the specified selector. As such, this needed to check multiple parent elements, not just the immediate parent. So the equivalent to the old logic would have been `SelectorFind.closest(sugarBlock, '*[dir="' + dir + '"]', (elem) => Compare.eq(elem, SugarElement.fromDom(editor.dom.getRoot()))`.

We can do this more easily though, as we can use computed CSS values to get the directionality. As such, we can fix the regression by just changing this to the following:
```suggestion
      const blockParentDirection = Direction.get(blockParent);
```
This uses the Sugar `Direction` module (which isn't currently imported) to determine the current computed directionality.

modules/tinymce/src/plugins/directionality/main/ts/core/Direction.ts
```diff
@@ -5,28 +5,43 @@
...
 
-    Tools.each(blocks, (block) => {
-      // Add dir to block if the parent block doesn't already have that dir
-      if (!dom.getParent(block.parentNode, '*[dir="' + dir + '"]', dom.getRoot())) {
-        dom.setAttrib(block, 'dir', curDir !== dir ? dir : null);
+const setDir = (editor: Editor, dir: Dir) => {
+  const selectedBlocks = editor.selection.getSelectedBlocks();
+  Arr.each(selectedBlocks, (block) => {
+    const sugarBlock = SugarElement.fromDom(block);
+    const blockParent = Traverse.parent(sugarBlock);
```
lnewson, line 19:
```suggestion
    const blockParent = Traverse.parent(SugarElement.fromDom(block));
```
Just a style thing here, as we don't need to use `sugarBlock` again we can just inline this.

modules/tinymce/src/plugins/directionality/main/ts/core/Direction.ts
```diff
@@ -5,28 +5,43 @@
...
+      } else { // if parent and child dir are going to be the same then remove it from child
+        Attribute.remove(sugarBlock, 'dir');
+        editor.nodeChanged();
       }
     });
+  });
+};
 
-    editor.nodeChanged();
+const setDirAttr = (editor: Editor, element: SugarElement, dir: Dir): void => {
```
lnewson, line 32:
```suggestion
const setDirAttr = (editor: Editor, element: SugarElement<Element>, dir: Dir): void => {
```
Missing type here for what the sugar element contains.

modules/tinymce/src/plugins/directionality/main/ts/core/Direction.ts
```diff
@@ -5,28 +5,43 @@
...
+  if (isListItem(element)) {
```
lnewson, line 33: This branch has a bug that will only be apparent with existing content. The problem is that it doesn't remove the directionality on the `li` elements. As such, if the `li` element already has a directionality it will never get changed and will be forever broken. As such we have 2 options here:
1. Ensure the `dir` attribute is removed on all child `li` elements of the list.
2. Remove the `dir` attribute for the element that's passed in. This is a little more brittle, as it will only remove the `dir` attribute from the `li` elements in the selection.

Also please make sure to add a test for trying to convert something like this (which is possible to do with the current version of TinyMCE):
```
<ul>
  <li dir="rtl">foo</li>
  <li>bar</li>
</ul>
```

modules/tinymce/src/plugins/directionality/main/ts/core/Direction.ts
```diff
@@ -5,28 +5,43 @@
...
+    const list = Traverse.parent(element);
+    list.map((l: SugarElement<Element>) => Attribute.set(l, 'dir', dir));
+  } else {
+    Attribute.set(element, 'dir', dir);
   }
+
+  editor.nodeChanged();
```
lnewson, line 40: This is also part of the node change issue I mentioned earlier.

modules/tinymce/src/plugins/directionality/main/ts/core/Direction.ts
```diff
@@ -5,28 +5,43 @@
...
 };
 
+const isListItem = (element: SugarElement<Element>): boolean => element.dom.nodeName === 'LI';
+
 export {
-  setDir
+  setDir, Dir
```
lnewson, line 46: Just a small style thing here (though largely irrelevant given James' comments), but we try to keep the exports on separate lines.

modules/tinymce/src/plugins/directionality/test/ts/browser/DirectionalitySanityTest.ts
```diff
@@ -47,4 +47,139 @@ describe('browser.tinymce.plugins.directionality.DirectionalitySanityTest', () =
       });
     }));
   });
+
+  it('should set two paragraphs to rtl and ltl', () => {
```
lnewson, line 51: Can you please add the JIRA number as a prefix here? We always try to include the JIRA number so we can trace why a test was added. eg:
```suggestion
  it('TINY-4589: should set two paragraphs to rtl and ltl', () => {
```

modules/tinymce/src/plugins/directionality/test/ts/browser/DirectionalitySanityTest.ts
```diff
@@ -47,4 +47,139 @@ describe('browser.tinymce.plugins.directionality.DirectionalitySanityTest', () =
...
+            attrs: {
+              dir: str.is('ltr')
+            }
+          })
+        ]
+      });
+    }));
+  });
+
+  it('should set parent dir when element is a list item', () => {
```
lnewson, line 93:
```suggestion
  it('TINY-4589: should set parent dir when element is a list item', () => {
```

modules/tinymce/src/plugins/directionality/test/ts/browser/DirectionalitySanityTest.ts
```diff
@@ -47,4 +47,139 @@ describe('browser.tinymce.plugins.directionality.DirectionalitySanityTest', () =
...
+              s.element('li', {}),
+              s.element('li', {}),
+            ]
+          }),
+        ]
+      });
+    }));
+  });
+
+  it('should remove dir attr if parent has same dir', () => {
```
lnewson, line 137:
```suggestion
  it('TINY-4589: should remove dir attr if parent has same dir', () => {
```

modules/tinymce/src/plugins/directionality/main/ts/core/Direction.ts
lnewson: Another bug I found while reviewing is this doesn't take into account where we need to apply the directionality. In the `li` case, the parent here will be the `ul` or `ol`. So checking the current directionality on that would be incorrect and instead we need to look at the `ul` or `ol` parent instead since that's where we apply the `dir` attribute.

modules/tinymce/src/plugins/directionality/main/ts/core/Direction.ts
```diff
@@ -5,28 +5,43 @@
...
+        editor.nodeChanged();
       }
     });
+  });
+};
 
-    editor.nodeChanged();
+const setDirAttr = (editor: Editor, element: SugarElement, dir: Dir): void => {
+  if (isListItem(element)) {
+    const list = Traverse.parent(element);
```
lnewson, line 34: We already have the parent in the previous function, so we could just pass it through here instead of looking it up again.

lnewson (changes requested):
Very nice work for a first attempt! There's a few complexities here though that we need to account for that aren't immediately obvious.

modules/tinymce/src/plugins/directionality/test/ts/browser/DirectionalitySanityTest.ts
```diff
@@ -47,4 +47,139 @@ describe('browser.tinymce.plugins.directionality.DirectionalitySanityTest', () =
...
     }));
   });
+
+  it('should set two paragraphs to rtl and ltl', () => {
+    const editor = hook.editor();
+    editor.setContent('<p>foo</p><p>bar</p>');
+    TinySelections.setSelection(editor, [ 0 ], 0, [ 1 ], 1);
+    TinyUiActions.clickOnToolbar(editor, 'button[title="Right to left"]');
+    TinyAssertions.assertContentStructure(editor, ApproxStructure.build((s, str) => {
+      return s.element('body', {
```
JamesToohey, line 57: I know the existing tests were using `assertContentStructure` but I think a plain `assertContent` here will be nicer and easier to read. I think we should assert the entire editor content given how simple the output would be in this case

`TinyAssertions.assertContent(editor, '<p dir="rtl">foo</p><p dir="rtl">bar</p>');`

You may want to set `indent: false` in the editor settings so that whitespace isn't added during serialization when `getContent` is called in these assertions.

modules/tinymce/src/plugins/directionality/main/ts/core/Direction.ts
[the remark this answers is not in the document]
diogo405: Didn't export the type but didn't inline as it's used more than once.

modules/tinymce/src/plugins/directionality/main/ts/core/Direction.ts
```diff
@@ -5,28 +5,43 @@
...
-    Tools.each(blocks, (block) => {
-      // Add dir to block if the parent block doesn't already have that dir
-      if (!dom.getParent(block.parentNode, '*[dir="' + dir + '"]', dom.getRoot())) {
-        dom.setAttrib(block, 'dir', curDir !== dir ? dir : null);
+const setDir = (editor: Editor, dir: Dir) => {
+  const selectedBlocks = editor.selection.getSelectedBlocks();
+  Arr.each(selectedBlocks, (block) => {
+    const sugarBlock = SugarElement.fromDom(block);
+    const blockParent = Traverse.parent(sugarBlock);
+    blockParent.map((blockParent: SugarElement<Element>) => {
```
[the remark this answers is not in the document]
diogo405: Didn't know Optional had `.each`. Changed 👍

modules/tinymce/src/plugins/directionality/main/ts/core/Direction.ts
```diff
@@ -5,28 +5,43 @@
...
+      const blockParentDirection = Attribute.get(blockParent, 'dir');
+      if ((blockParentDirection === undefined) || (blockParentDirection !== dir)) {
```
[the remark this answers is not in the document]
diogo405: If `dir` is empty it returns an empty string. I've added the condition to the if clause.

modules/tinymce/src/plugins/directionality/main/ts/core/Direction.ts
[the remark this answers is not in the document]
diogo405: Added more conditions to the if clause: `isUndefined` || `isNull` || `isEmpty`.

modules/tinymce/src/plugins/directionality/main/ts/core/Direction.ts
diogo405: Changed!

modules/tinymce/src/plugins/directionality/main/ts/core/Direction.ts
diogo405: Why `>= 0` ? Shouldn't it be `> 0` ?

modules/tinymce/src/plugins/directionality/main/ts/core/Direction.ts
diogo405: Hmm `sugarBlock` is being used in `setAttrDir` (~line 28) and `Attribute.remove` (else).

modules/tinymce/src/plugins/directionality/main/ts/core/Direction.ts
diogo405: Fixed 👍

modules/tinymce/src/plugins/directionality/main/ts/core/Direction.ts
lnewson: Yeah it should be `> 0` sorry.

modules/tinymce/src/plugins/directionality/main/ts/core/Direction.ts
lnewson: Ahh, I missed that somehow sorry. Ignore this then.

modules/tinymce/src/plugins/directionality/main/ts/core/Direction.ts
diogo405: Fixed. Although this function will probably die when I fix the `editor.nodeChanged()`. It will have only one line and it could be inside `setDir`. Something like: 

`Attribute.set(isListItem(sugarBlock) ? blockParent : sugarBlock, 'dir', dir);`.

modules/tinymce/src/plugins/directionality/main/ts/core/Direction.ts
diogo405: Yeah... I was thinking about the logic, we would have something like this, right?

```
each block
  if li
    li remove dir
    check parent of ul/ol (p2) that's not a li and has dir // dunno how we gonna do this
    if p2 dir = clickedDir
      p2 remove dir
    else
      set dir on parent of li 
  else
    if parent dir = clickedDir
      block remove dir
    else
      set dir on block
```

This line `check parent of ul/ol (p2) that's not a li and has dir` because we can have ul that has a li as a parent with the dir attribute (incorrectly):

```
<ul>
  <li dir="rtl">
    Item 1
    <ul> <!-- this one -->
      <li> 
        Item 2
      </li>
    </ul>
  </li>
</ul>
```

modules/tinymce/src/plugins/directionality/main/ts/core/Direction.ts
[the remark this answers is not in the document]
diogo405: Ended up changing to follow your logic. Seems easier to understand :)

modules/tinymce/src/plugins/directionality/test/ts/browser/DirectionalitySanityTest.ts
diogo405: For simple tests I changed to `assertContent` for long content I'm still using `assertContentStructure`.

modules/tinymce/src/plugins/directionality/main/ts/core/Direction.ts
diogo405: I'm requesting a new review as I changed the logic to find the closest parent which has `dir` and is not a li.

modules/tinymce/src/plugins/directionality/main/ts/core/Direction.ts
diogo405: I fixed that with the last commit 👍

modules/tinymce/src/plugins/directionality/test/ts/browser/DirectionalitySanityTest.ts
```diff
@@ -17,31 +17,272 @@ describe('browser.tinymce.plugins.directionality.DirectionalitySanityTest', () =
...
+    TinyUiActions.clickOnToolbar(editor, 'button[title="Right to left"]');
+    TinyAssertions.assertContent(editor, `<div dir="">\n<p dir="rtl">foo</p>\n<p>bar</p>\n</div>`);
+
+    TinyUiActions.clickOnToolbar(editor, 'button[title="Left to right"]');
+    TinyAssertions.assertContent(editor, `<div dir="">\n<p dir="ltr">foo</p>\n<p>bar</p>\n</div>`);
+  });
+
+  it('TINY-4589: should use first correct parent dir', () => {
+    const editor = hook.editor();
+    editor.setContent(`
```
ashfordneil, line 182: I'm pretty sure with `editor.setContent` this won't make a difference, but the extra newlines and whitespace in the HTML that you get from putting it all into a multiline template string like this can sometimes lead to extra text nodes, which throws off all of your `setSelection` API call numbers.

Shouldn't be an issue with `editor.setContent`, but keep it in mind if you ever need to use something like `SugarElement.fromHTML` in a test.

ashfordneil (approved):

modules/tinymce/src/plugins/directionality/test/ts/browser/DirectionalitySanityTest.ts
```diff
@@ -17,31 +17,272 @@ describe('browser.tinymce.plugins.directionality.DirectionalitySanityTest', () =
...
+    TinyAssertions.assertContent(editor, `<p dir="rtl">foo</p>\n<p dir="rtl">bar</p>`);
+
+    TinySelections.setSelection(editor, [ 0 ], 0, [ 1 ], 1);
+    TinyUiActions.clickOnToolbar(editor, 'button[title="Left to right"]');
+    TinyAssertions.assertContent(editor, `<p dir="ltr">foo</p>\n<p dir="ltr">bar</p>`);
+  });
+
+  it('TINY-4589: should set parent dir when element is a list item', () => {
+    const editor = hook.editor();
+    editor.setContent(`
```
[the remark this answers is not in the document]
diogo405: Changed to one line.

modules/tinymce/src/plugins/directionality/test/ts/browser/DirectionalitySanityTest.ts
```diff
@@ -17,31 +17,272 @@ describe('browser.tinymce.plugins.directionality.DirectionalitySanityTest', () =
...
+    <ul>
+      <li>foo</li>
+      <li>bar</li>
+    </ul>
+    `);
+    TinySelections.setSelection(editor, [ 0, 0 ], 0, [ 0, 0 ], 1);
+    TinyUiActions.clickOnToolbar(editor, 'button[title="Right to left"]');
+    TinyAssertions.assertContent(editor, `<ul dir="rtl">\n<li>foo</li>\n<li>bar</li>\n</ul>`);
```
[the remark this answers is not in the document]
diogo405: Remove this one and all other unnecessary ` 👍

modules/tinymce/src/plugins/directionality/test/ts/browser/DirectionalitySanityTest.ts
```diff
@@ -17,31 +17,272 @@ describe('browser.tinymce.plugins.directionality.DirectionalitySanityTest', () =
...
+                  dir: str.is('ltr')
+                },
+                children: [
+                  s.anything(),
+                  s.element('ul', {
+                    attrs: {
+                      dir: str.is('ltr')
+                    },
+                    children: [
+                      s.element('li', {}),
```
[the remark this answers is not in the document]
diogo405: Oh didn't know that, thanks. Changed some tests to `dir: str.none()`.

modules/tinymce/src/plugins/directionality/test/ts/browser/DirectionalitySanityTest.ts
diogo405: Roight :)

modules/tinymce/src/plugins/directionality/test/ts/browser/DirectionalitySanityTest.ts
```diff
@@ -17,31 +17,308 @@ describe('browser.tinymce.plugins.directionality.DirectionalitySanityTest', () =
...
+          <li dir="rtl">b</li>
+          <li>c</li>
+        </ul>
+      </li>
+      <li dir="xyz">bar</li>
+    </ul>
+    `);
+    TinySelections.setSelection(editor, [ 0, 0 ], 0, [ 0, 0 ], 1);
+    TinyUiActions.clickOnToolbar(editor, 'button[title="Right to left"]');
     TinyAssertions.assertContentStructure(editor, ApproxStructure.build((s, str) => {
```
metricjs, line 71: I'm curious - could this be a string assertion instead of an ApproxStructure, and do we care?

metricjs (approved):

modules/tinymce/src/plugins/directionality/main/ts/core/Direction.ts
```diff
@@ -5,24 +5,48 @@
...
-  const dom = editor.dom;
-  let curDir;
-  const blocks = editor.selection.getSelectedBlocks();
+type Dir = 'rtl' | 'ltr';
 
-  if (blocks.length) {
-    curDir = dom.getAttrib(blocks[0], 'dir');
+const getParentElement = (element: SugarElement<Element>): Optional<SugarElement<Element>> => {
+  return Traverse.parent(element).filter(SugarNode.isElement);
+};
```
lnewson, line 17:
```suggestion
const getParentElement = (element: SugarElement<Element>): Optional<SugarElement<Element>> =>
  Traverse.parent(element).filter(SugarNode.isElement);
```
We normally try to use shorthand if possible, unless it causes issues with readability.

modules/tinymce/src/plugins/directionality/main/ts/core/Direction.ts
```diff
@@ -5,24 +5,48 @@
...
 
+const isListItem = SugarNode.isTag('li');
+
+const setDir = (editor: Editor, dir: Dir): void => {
+  const selectedBlocks = editor.selection.getSelectedBlocks();
+  if (selectedBlocks.length > 0) {
+    Arr.each(selectedBlocks, (block) => {
+      const sugarBlock = SugarElement.fromDom(block);
+      const isSugarBlockListItem = isListItem(sugarBlock);
```
lnewson, line 32: Try to avoid calling this `sugar` please, as we don't generally do this anywhere else since most times we use Sugar wherever possible. Instead this could just be `blockElem` and `isBlockListItem`.

- [ ] Rename this variables please

modules/tinymce/src/plugins/directionality/test/ts/browser/DirectionalitySanityTest.ts
lnewson: I suspect this is because of the indentation. This can be solved by adding `indent: false` to the editor settings, so yes we should use that instead.

- [ ] Please convert all these to a string based assertion instead

lnewson (changes requested):
Marking as request changes just because of the tests sorry, as they are largely unreadable atm with the approx structures.

modules/tinymce/src/plugins/directionality/main/ts/core/Direction.ts
diogo405: Changed.

lnewson: Not fixed sorry, as you squashed this all onto a single line which is too long and not what I had suggested.

modules/tinymce/src/plugins/directionality/test/ts/browser/DirectionalitySanityTest.ts
```diff
@@ -9,42 +8,175 @@ describe('browser.tinymce.plugins.directionality.DirectionalitySanityTest', () =
...
+        '<li dir="ltr">ini' +
+          '<ul>' +
+            '<li>foo</li>' +
+            '<li>bar</li>' +
+          '</ul>' +
+        '</li>' +
+      '</ul>'
+    );
+    TinySelections.setSelection(editor, [ 0, 0, 1, 0 ], 0, [ 0, 0, 1, 0 ], 1); // foo
+    TinyUiActions.clickOnToolbar(editor, 'button[title="Left to right"]');
```
lnewson, line 97: This test isn't really doing anything. Should this have been placed after the `rtl` change so we can check it moves back to `ltr`?

modules/tinymce/src/plugins/directionality/test/ts/browser/DirectionalitySanityTest.ts
```diff
@@ -9,42 +8,175 @@ describe('browser.tinymce.plugins.directionality.DirectionalitySanityTest', () =
...
+    editor.setContent('<div dir="ltr"><p>foo</p><p>bar</p></div>');
+    TinySelections.setSelection(editor, [ 0, 0 ], 0, [ 0, 0 ], 1);
+    TinyUiActions.clickOnToolbar(editor, 'button[title="Right to left"]');
+    TinyAssertions.assertContent(editor, '<div dir="ltr"><p dir="rtl">foo</p><p>bar</p></div>');
+
+    TinyUiActions.clickOnToolbar(editor, 'button[title="Left to right"]');
+    TinyAssertions.assertContent(editor, '<div dir="ltr"><p>foo</p><p>bar</p></div>');
+  });
+
+  it('TINY-4589: should get computed dir from #target', () => {
```
lnewson, line 133: What is going on here? There's a whole heap of invalid data here which doesn't make sense to me. Fair enough if you had one, but having 3 all doing the same thing is just complicating this.

modules/tinymce/src/plugins/directionality/main/ts/core/Direction.ts
diogo405: Changed to two lines. What's the line width?

modules/tinymce/src/plugins/directionality/test/ts/browser/DirectionalitySanityTest.ts
diogo405: The idea of this one is to assert that clicking on LTR when foo is selected won't change the ul direction as its parent has dir="ltr" already.

modules/tinymce/src/plugins/directionality/test/ts/browser/DirectionalitySanityTest.ts
diogo405: I was making sure it goes up till it gets a valid `dir`, that's why I tested empty and invalid ('x' and ' ').

modules/tinymce/src/plugins/directionality/test/ts/browser/DirectionalitySanityTest.ts
lnewson: That should be a separate case then, as that's testing something different to what's in the title (it's testing that applying the same dir makes no changes).

diogo405: Broke into two test cases: `should consider list item dir` and `applying the same dir makes no changes`.

modules/tinymce/src/plugins/directionality/test/ts/browser/DirectionalitySanityTest.ts
lnewson: You don't need that though, it just needs one invalid case not three.

modules/tinymce/src/plugins/directionality/test/ts/browser/DirectionalitySanityTest.ts
```diff
@@ -9,42 +8,189 @@ describe('browser.tinymce.plugins.directionality.DirectionalitySanityTest', () =
...
+          '</ul>' +
+        '</li>' +
+        '<li>bar</li>' +
+      '</ul>'
+    );
+  });
+
+  it('TINY-4589: applying the same dir makes no changes', () => {
+    const editor = hook.editor();
+    editor.setContent(
```
lnewson, line 86: This uses the same HTML twice, so can we pull this out into a separate variable please? That'll also guarantee it's 100% the same.

lnewson (approved):

modules/tinymce/src/plugins/directionality/test/ts/browser/DirectionalitySanityTest.ts
diogo405: Yeap, agree.

lnewson (approved):

PR #6982, opened 2021-07-03
Move caret behind newly inserted links

### LINKED ISSUE #6711: Collapse selection to end after inserting hyperlink on selected text
tl;dr: Just read the bold-printed text below.

### What is the current behavior? How to reproduce

* https://fiddle.tiny.cloud/
* Write some text (e.g. "Test") and **select it**.
* Press Ctrl+K to **insert a hyperlink**. Give it an arbitrary URL (eg. `http://test.com`) and click "Save".
* Observe that "Test" is now selected, i.e. this is the selection:
  ```bash
  Sometext |<a href="...">Linktext</a>| after
           |--------------------------|    <---this is where the selection is
  ```
* Typing now removes the link has just been inserted, e.g. if I type `B` I get: `Sometext b| after`.
* In order to continue typing behind the newly inserted link (as one would normally want when typing texts), the user has to press the right arrow key twice (once to collapse the selection to the end resulting in `Sometext <a href="...">Linktext|</a> after`, and one to step outside the link text resulting in `Sometext <a href="...">Linktext</a>| after`).

### What is the expected behavior?

**After inserting a link, the cursor should be placed _behind_ the link**, so that the next typed character is not added to the link or replaces it. I.e.

```bash
Sometext <a href="...">Linktext</a>| after
                                   ^---- this is where the caret should be after inserting the link
```

To me this behavior makes more sense because the user likely wants to continue typing. Given that they just had the opportunity to edit the link text in the dialog, it's unlikely that they want to modify it again in the editor after closing the dialog.

The desired behavior is also the behavior of Google Docs and Microsoft Word (at least their web version, I don't have it installed).

**Does it matter? I think so.** I have muscle memory to paste links on selected text and continue typing behind it (Ctrl+K, Ctrl+V, Enter, Space, ...). It works in all applications I use, except with TinyMCE I end up erasing my inserted links.

### Which versions of TinyMCE, and which browser / OS are affected by this issue? Did this work in previous versions of TinyMCE?

I'm using the latest Chrome version on Windows and tinymce5-cloud-stable. This also reproduces e.g. in 5.0.0.

Philipp91:
Fixes #6711.

Description of Changes:
* Add unit test that succeeds _before_ the changes, for documentation purposes.
* Change behavior to move caret behind newly inserted links.
* Update the unit tests accordingly. Add a new test case to illustrate with an extreme case (adding a link on `b` within `abc`, then typing `d`, which should result in `a<a>b</a>dc` -- note how the `d` is _not_ inside the link).

Pre-checks:
* [x] Changelog entry added
* [ ] ~~Branch prefixed with `feature/` for new features (if applicable)~~

Review:
* [ ] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -27,6 +27,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Improved the performance when rendering UI components #TINY-7572
 - When scrolling, the context toolbar will stick to where it was previously for large elements, such as tables #TINY-7545
 - The context toolbar will now move out of the way when it overlaps with the selection, such as in table cells #TINY-7192
+- The caret is now positioned behind newly inserted links, so that the user can carry on typing #6711
 
 ### Changed
 - Changed the load order so that the content css gets loaded before the editor gets populated with contents. #TINY-7249
--- a/modules/tinymce/src/core/main/ts/api/EditorCommands.ts
+++ b/modules/tinymce/src/core/main/ts/api/EditorCommands.ts
@@ -518,6 +518,9 @@ class EditorCommands {
         if (value.href) {
           editor.formatter.apply('link', value, anchor);
         }
+
+        // Move the selection behind the new link, so the user can go on typing.
+        editor.selection.setStart(editor.selection.getEnd());
       },
 
       'selectAll': () => {
--- a/modules/tinymce/src/core/main/ts/api/dom/Selection.ts
+++ b/modules/tinymce/src/core/main/ts/api/dom/Selection.ts
@@ -85,6 +85,7 @@ interface EditorSelection {
   setRng: (rng: Range, forward?: boolean) => void;
   getRng: () => Range;
   getStart: (real?: boolean) => Element;
+  setStart: (elm: Element) => void;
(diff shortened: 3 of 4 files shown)


lnewson:
Hi @Philipp91,

Thanks for logging this PR, however without the CLA being signed we're unable to accept this. Additionally, tests are currently failing with these changes. As it's been over a week and this hasn't been resolved we're going to close this.

If you can fix both of those issues we're happy to reopen this. However, I should also add that we have discussed this internally and we would like to see some additional changes if you'd like to continue this. Firstly, the setStart API  is confusing/inconsistent as it doesn't actually do what most of the other selection APIs do (they try and place it within the node, not after). We'd like to see that removed and just use the setRng API instead. Secondly, we'd like to see the selection change only happen when links are used via the link plugin, not in core. The reason being doing it in core would cause selection issues in some places this is used (as noted by the test failures). Lastly, we'd only want the selection to be placed after the link when inserting a new link. Editing a link should not change the selection.

Let us know if you want to attempt these changes. If not, we're happy to make an internal ticket for this behaviour, however at this stage we don't have a time frame for when this would be implemented.

Philipp91:
> If not, we're happy to make an internal ticket for this behaviour.

Please do that then, assuming #6711 isn't equivalent.

PR #6984, opened 2021-07-04
TINY-7678: Replace Cell<Optional<T>> with Singleton<T>

lnewson:
Related Ticket: TINY-7678

Description of Changes:
* Replace `Cell<Optional<T>>` uses with `Singleton<T>`
* Clean up `Singleton` to remove some duplication

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [x] Milestone set

### DIFF
--- a/modules/alloy/src/main/ts/ephox/alloy/behaviour/reflecting/ReflectingState.ts
+++ b/modules/alloy/src/main/ts/ephox/alloy/behaviour/reflecting/ReflectingState.ts
@@ -1,20 +1,18 @@
-import { Cell, Fun, Optional } from '@ephox/katamari';
+import { Cell, Optional } from '@ephox/katamari';
 
 import { ReflectingState } from './ReflectingTypes';
 
 const init = <S>(): ReflectingState<S> => {
-  const cell: Cell<Optional<S>> = Cell(Optional.none<S>());
+  const cell = Cell(Optional.none<S>());
 
-  const set = (optS: Optional<S>) => cell.set(optS);
   const clear = () => cell.set(Optional.none<S>());
-  const get = () => cell.get();
 
-  const readState = (): any => cell.get().fold<any>(Fun.constant('none'), Fun.identity);
+  const readState = (): any => cell.get().getOr('none');
 
   return {
     readState,
-    get,
-    set,
+    get: cell.get,
+    set: cell.set,
     clear
   };
 };
--- a/modules/katamari/src/main/ts/ephox/katamari/api/Singleton.ts
+++ b/modules/katamari/src/main/ts/ephox/katamari/api/Singleton.ts
@@ -1,4 +1,5 @@
 import { Cell } from './Cell';
+import * as Fun from './Fun';
 import { Optional } from './Optional';
 
 interface Singleton<T> {
@@ -18,7 +19,7 @@ export interface Value<T> extends Singleton<T> {
   readonly on: (fn: (data: T) => void) => void;
 }
 
-const revocable = <T> (doRevoke: (data: T) => void): Singleton<T> => {
+const singleton = <T> (doRevoke: (data: T) => void): Singleton<T> => {
   const subject = Cell(Optional.none<T>());
 
   const revoke = (): void => subject.get().each(doRevoke);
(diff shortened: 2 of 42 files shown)


modules/tinymce/src/core/main/ts/undo/UndoManagerTypes.ts
```diff
@@ -43,4 +43,4 @@ export type Index = Cell<number>;
 
 export type Locks = Cell<number>;
 
-export type UndoBookmark = Cell<Optional<Bookmark>>;
+export type UndoBookmark = Singleton.Value<Bookmark>;
```
lnewson, line 46: FYI @TheSpyder this will change the type of object being passed to RTC. However, I had a look and RTC doesn't implement `add` or `beforeChange` yet so it shouldn't be impacted by this. If it will cause issues let me know and I'll revert it back.

modules/tinymce/src/core/main/ts/undo/Levels.ts
```diff
@@ -5,23 +5,17 @@
...
-import { Arr, Cell, Optional } from '@ephox/katamari';
+import { Arr, Thunk } from '@ephox/katamari';
 import { Html, Remove, SelectorFilter, SugarElement } from '@ephox/sugar';
 
 import Editor from '../api/Editor';
 import * as TrimHtml from '../dom/TrimHtml';
 import * as Fragments from './Fragments';
 import { UndoLevel, UndoLevelType } from './UndoManagerTypes';
 
-const undoLevelDocument = Cell<Optional<Document>>(Optional.none());
```
lnewson, line 16: This only needed a `Thunk.cached` and didn't need a cell/singleton so I've replaced it.

HAFRMO (approved):

modules/alloy/src/main/ts/ephox/alloy/behaviour/reflecting/ReflectingState.ts
```diff
@@ -1,20 +1,18 @@
-import { Cell, Fun, Optional } from '@ephox/katamari';
+import { Cell, Optional } from '@ephox/katamari';
 
 import { ReflectingState } from './ReflectingTypes';
 
 const init = <S>(): ReflectingState<S> => {
-  const cell: Cell<Optional<S>> = Cell(Optional.none<S>());
+  const cell = Cell(Optional.none<S>());
```
tiny-james, line 6: Why did this one not become a Singleton?

lnewson: Because the input data is `Optional` not an actual value, so `Cell` is a more appropriate data structure for this.

tiny-james (approved):

ltrouton (approved):

modules/katamari/src/main/ts/ephox/katamari/api/Singleton.ts
```diff
@@ -18,7 +19,7 @@ export interface Value<T> extends Singleton<T> {
   readonly on: (fn: (data: T) => void) => void;
 }
 
-const revocable = <T> (doRevoke: (data: T) => void): Singleton<T> => {
+const singleton = <T> (doRevoke: (data: T) => void): Singleton<T> => {
```
ashfordneil, line 22: Is there a reason this was changed? Its still specifically a function to build a Singleton that can "revoke" the thing inside it, we just sometimes use it with `Fun.noop` to create a more generic singleton.

ashfordneil (approved):

modules/katamari/src/main/ts/ephox/katamari/api/Singleton.ts
lnewson: It's the common interface between all now and not specific to revocable singletons. We could make `doRevoke` optional or defaulted but it didn't seem neccessary.

TheSpyder: Singleton seems like the wrong name for this now... but I don't have a better name to offer. In hindsight I do wonder if `Singleton.value()` should have been called `Singleton.empty()` for clarity.

modules/tinymce/src/core/main/ts/undo/Levels.ts
TheSpyder: oh wow I guess when we converted to katamari the person doing this file didn't know about cached thunks :)

modules/tinymce/src/core/main/ts/undo/UndoManagerTypes.ts
TheSpyder: If you look closely the type is only used in `RtcAdaptor`, which is a copy of the API types for the purposes of making runtime adaptors. The actual RTC api is interface `RtcRuntimeApi` so yes the value is never passed to the plugin (because, as you found, we haven't implemented the entire undo manager API in the alt-model).

TheSpyder (approved):

modules/tinymce/src/core/main/ts/undo/UndoManagerTypes.ts
lnewson: Right, I was looking at the adaptor instead. Thanks for the clarification!

modules/katamari/src/main/ts/ephox/katamari/api/Singleton.ts
lnewson: I think `empty` could be just as confusing, as it sounds like it should never have a value? Anyways, for now lets just leave it as is. If someone wants to desperately rename it, then feel free to do so in another PR 🤷

Issue #5013, opened 2019-05-24
TinyMCE 5: Focus trapped in IE11


jburghardt:
**Do you want to request a *feature* or report a *bug*?**
 bug

**What is the current behavior?**

In IE11, the focus is trapped in the Menu.
The Menu should not be in the Tab order, due to tabindex -1

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via [fiddle.tinymce.com](http://fiddle.tinymce.com/) or similar.**

Open https://www.tiny.cloud/docs/demo/full-featured/ in IE11
and tab trough the page

**What is the expected behavior?**

Tab from outside will focus the Editor edit area, the menu is only focusable trough hotkey.

**Which versions of TinyMCE, and which browser / OS are affected by this issue? Did this work in previous versions of TinyMCE?**

5.0.6

Tjerk-Haaye-Henricus:
Hey w👋 

The reason is.. IE11 makes each svg tabbable with the `focusable` attribute. To solve this issue you guys have to add the attribute `focusable="false"` to the svgicons.

With Love from Tjerk ❤️

PR #6990, opened 2021-07-06
TINY-3707: Tabbing from outside the editor into the editor focus on icon on IE 11 

diogo405:
Related Ticket:

Description of Changes:
* Set `focusable` attribute on SVG to false.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

Fixes https://github.com/tinymce/tinymce/issues/5013


modules/tinymce/CHANGELOG.md
```diff
@@ -45,6 +45,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Links in notification text did not show the correct mouse pointer #TINY-7661
 - The formatter match APIs were incorrectly returning false for formats that specified an attribute or style should be removed from an element #TINY-6149
 - The type signature on `editor.formatter.matchNode` had the wrong return type (was `boolean` but should have been `Formatter | undefined`) #TINY-6149
+- Tabbing from outside the editor into it focus on toolbar icon rather than editor on IE 11 #TINY-3707
```
JamesToohey, line 48: Maybe something like this?
```suggestion
- Using the Tab key to navigate into the editor on IE 11 would incorrectly focus the toolbar #TINY-3707
```

modules/tinymce/src/themes/silver/main/ts/ui/icons/Icons.ts
```diff
@@ -17,8 +19,26 @@ const getOr = (name: string, icons: IconProvider, fallback: Optional<string>): s
...
+const render = (tagName: string, iconHtml: string, classes: string[], behaviours: Partial<SimpleOrSketchSpec> = {}): SimpleOrSketchSpec => {
+  return ({
+    dom: {
+      tag: tagName,
+      innerHtml: iconHtml,
+      classes
+    },
+    events: AlloyEvents.derive([
+      AlloyEvents.runOnAttached((comp) => {
+        // ie 11 focus on svg without focusable = false attr
```
JamesToohey, line 31: Maybe?
```suggestion
        // set focusable=false on SVGs to prevent IE11 from focusing the toolbar when tabbing into the editor
```

modules/tinymce/src/themes/silver/main/ts/ui/button/ButtonSlices.ts
```diff
@@ -8,18 +8,11 @@
...
-  ({
-    dom: {
-      tag: 'span',
-      innerHtml: iconHtml,
-      classes: [ ToolbarButtonClasses.Icon, ToolbarButtonClasses.IconWrap ]
-    },
-    ...behaviours
-  });
+const renderIcon = (iconHtml: string, behaviours: Partial<SimpleOrSketchSpec>): SimpleOrSketchSpec =>
+  renderIconElement('span', iconHtml, [ ToolbarButtonClasses.Icon, ToolbarButtonClasses.IconWrap ], behaviours);
```
JamesToohey, line 15: I know this was previously named this way, but am I right in thinking that `behaviours` could really be any partial of the `ComponentSpec`? That means you could pass in `{ events: ...something }` and it would override the `AlloyEvents` you've specified in the `render` in `Icons.ts`.

If this is only ever actual behaviours, maybe it should be `AlloyBehaviourRecord` and not be spread into the object in `Icons.ts`.

This may not be something we care about (or I may be wrong) I just wanted to raise it

JamesToohey (approved):

modules/tinymce/src/themes/silver/main/ts/ui/icons/Icons.ts
diogo405: Better 👍

modules/tinymce/src/themes/silver/main/ts/ui/button/ButtonSlices.ts
diogo405: Hmm I know what you mean (after reading it 5x hehe) but I don't think I have knowledge of the codebase to answer. Maybe @lnewson ?

modules/tinymce/src/themes/silver/main/ts/ui/icons/Icons.ts
```diff
@@ -17,8 +19,26 @@ const getOr = (name: string, icons: IconProvider, fallback: Optional<string>): s
...
+  return ({
+    dom: {
+      tag: tagName,
+      innerHtml: iconHtml,
+      classes
+    },
+    events: AlloyEvents.derive([
+      AlloyEvents.runOnAttached((comp) => {
+        // set focusable=false on SVGs to prevent IE 11 from focusing the toolbar when tabbing into the editor
+        SelectorFind.child(comp.element, 'svg').each((svg) => Attribute.set(svg, 'focusable', 'false'));
```
ashfordneil, line 32: I don't think there's any scenario where we could end up with multiple SVGs in the same icon element, but is it worth doing this just to be defensive?
```suggestion
        const svgs = SelectorFilter.children(comp.element, 'svg');
        Arr.each(svgs, (svg) => Attribute.set(svg, 'focusable', 'false'));
```

ashfordneil (approved):

modules/tinymce/src/themes/silver/main/ts/ui/icons/Icons.ts
lnewson: We only support a single icon, so I don't want to be encouraging or looking like we support this.

modules/tinymce/src/themes/silver/main/ts/ui/button/ButtonSlices.ts
diogo405: Looking at the code, only `renderIconFromPack` and `renderReplacableIconFromPack` from `ButtonSlices` call `render` from `Icon` passing behaviours. `renderIconFromPack` passes an empty object and `renderReplacableIconFromPack` passes behaviours (no override). That said, I think you are correct, we don't need the spread and can change the type.

lnewson: Yeah, this is what I mentioned in our call the other day. We need to either rename `behaviours` to something more accurate, or make it `Optional<AlloyBehaviourRecord>`. The latter is tricker since it's not always passed, so the `behaviours` should only be in the spec if it's `some`.

diogo405: Change the type to `AlloyBehaviourRecord` and remove the spread.

lnewson: As I mentioned above, it needs to be `Optional` as we shouldn't be passing `undefined` into the spec, so this needs changing please.

- [ ] Fix as per comments above

diogo405: ah right, "we shouldn't be passing undefined" triggered the optional in my head :)

modules/tinymce/src/themes/silver/main/ts/ui/icons/Icons.ts
```diff
@@ -17,8 +19,30 @@ const getOr = (name: string, icons: IconProvider, fallback: Optional<string>): s
...
+      })
+    ])
+  };
+
+  return behaviours.fold(
+    Fun.constant(iconSpec),
+    (behaviour) => ({ ...iconSpec, behaviour })
```
lnewson, line 39: This is failing tests, as this needs to be `behaviours`, not `behaviour` (it's missing the trailing `s`).

diogo405: Fixed.

modules/tinymce/CHANGELOG.md
```diff
@@ -45,6 +45,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Links in notification text did not show the correct mouse pointer #TINY-7661
 - The formatter match APIs were incorrectly returning false for formats that specified an attribute or style should be removed from an element #TINY-6149
 - The type signature on `editor.formatter.matchNode` had the wrong return type (was `boolean` but should have been `Formatter | undefined`) #TINY-6149
+- Using the Tab key to navigate into the editor on IE 11 would incorrectly focus the toolbar #TINY-3707
```
JamesToohey, line 48: Do we want (and is it easy) to add a test for this specifically for IE?

modules/tinymce/src/themes/silver/main/ts/ui/icons/Icons.ts
```diff
@@ -17,8 +19,32 @@ const getOr = (name: string, icons: IconProvider, fallback: Optional<string>): s
...
+    ])
+  };
+
+  const behavioursSpec = behavioursOpt.map((behaviours) => ({ behaviours })).getOr({});
+
+  return {
+    ...iconSpec,
+    ...behavioursSpec
+  };
+};
```
JamesToohey, line 43: This is done this way to prevent `behaviours` from being overwritten during the spread, but in its current state it looks very "refactorable" -- that is, someone might simplify this later and just return behaviours in this response like `behaviours: behavioursOpt.getOr({})`. Was there a specific reason you removed the `fold` from earlier?

lnewson: I asked Diogo to remove the fold, as it's a bad pattern to be using for this case as it prevents multiple optional arguments being folded into the spec (even if we only have one now).

lnewson: I should also note that this pattern is what's used in other places in Silver. You certainly could do `behaviours: behavioursOpt.getOr({})` though, as that would almost be identical (the current one means the key/value isn't included at all, while the latter includes an empty object).

modules/tinymce/CHANGELOG.md
diogo405: The test would be a text input next to Tiny Editor, the user tabs from the input and it should focus on the editor, right?!

lnewson: Agreed we should add a test here... so yes, this would need to be a webdriver test with an element outside the editor that uses `RealKeys` to tab inside the editor IIUC.

diogo405: Test added.

modules/tinymce/src/themes/silver/test/ts/webdriver/editor/TabbingTest.ts
```diff
@@ -0,0 +1,24 @@
...
+import Theme from 'tinymce/themes/silver/Theme';
+
+describe('webdriver.tinymce.themes.silver.editor.TabbingTest', () => {
+  const hook = TinyHooks.bddSetup<Editor>({
+    base_url: '/project/tinymce/js/tinymce',
+  }, [ Theme ]);
+
+  it('TINY-3707: Should focus on text editor when tabbing into it', async () => {
+    const editor = hook.editor();
+    const textInput = SugarElement.fromHtml('<input>') as SugarElement<HTMLElement>;
```
lnewson, line 16:
```suggestion
    const textInput = SugarElement.fromTag('input');
```

modules/tinymce/src/themes/silver/test/ts/webdriver/editor/TabbingTest.ts
```diff
@@ -0,0 +1,24 @@
...
+    const editorElement = SugarElement.fromDom(editor.getElement());
+    Insert.before(editorElement, textInput);
```
lnewson, line 18: This `textInput` element also needs to be removed from the dom when the test completes.

- [ ] Ensure input element is cleaned up

modules/tinymce/src/themes/silver/test/ts/webdriver/editor/TabbingTest.ts
```diff
@@ -0,0 +1,24 @@
...
+
+describe('webdriver.tinymce.themes.silver.editor.TabbingTest', () => {
+  const hook = TinyHooks.bddSetup<Editor>({
+    base_url: '/project/tinymce/js/tinymce',
+  }, [ Theme ]);
+
+  it('TINY-3707: Should focus on text editor when tabbing into it', async () => {
+    const editor = hook.editor();
+    const textInput = SugarElement.fromHtml('<input>') as SugarElement<HTMLElement>;
+    const editorElement = SugarElement.fromDom(editor.getElement());
```
lnewson, line 17: This should be using `TinyDom.targetElement(editor)` instead.

modules/tinymce/src/themes/silver/test/ts/webdriver/editor/TabbingTest.ts
```diff
@@ -0,0 +1,24 @@
...
+    Insert.before(editorElement, textInput);
+
+    Focus.focus(textInput);
+    await RealKeys.pSendKeysOn('input', [ RealKeys.text('\t') ]);
+    Assertions.assertEq('Editor has focus', true, editor.hasFocus());
```
lnewson, line 22: This is using the older assertions API, where it should be using chai. eg:
```
assert.isTrue(editor.hasFocus());
```

lnewson (approved):
Looks good, just needs some fixes in the new test.

modules/tinymce/src/themes/silver/test/ts/webdriver/editor/TabbingTest.ts
lnewson: Worth noting you can also do this to avoid the cast when using `fromHtml`:
```
SugarElement.fromHtml<HTMLInputElement>('<input>');
```

modules/tinymce/src/themes/silver/test/ts/webdriver/editor/TabbingTest.ts
diogo405: Just curious, why the element needs to be removed?

lnewson: Because otherwise it'll stay on the page indefinitely, which could cause other tests to fail.

metricjs (approved):

diogo405 (commented):
Some SVGs didn't have the `focusable=false` (e.g. dropdown arrow icon) and that failed IE 11 test. So we changed the `renderIcon` function and few files to render icons using Icons.ts function. That way we know the icon will include the `focusable=false` attribute.

When the icon cannot use the `renderIcon` function we are using `addFocusableBehaviour` inside the behaviours field.

lnewson:
Re-requesting reviews due to some larger changes done as mentioned above.

modules/tinymce/src/themes/silver/main/ts/ui/general/Notification.ts
```diff
@@ -14,7 +14,7 @@ import { Arr, Optional } from '@ephox/katamari';
 
 import { TranslatedString, Untranslated } from 'tinymce/core/api/util/I18n';
 
-import { get as getIcon, getFirst, IconProvider } from '../icons/Icons';
+import { addFocusableBehaviour, get as getIcon, getFirst, IconProvider } from '../icons/Icons';
```
lnewson, line 17: Can this please be changed to use `* as Icons` and update the references below? It's more consistent and helps keep the context for where it's used below.

Example:
```
import * as Icons from '../icons/Icons';
```

modules/tinymce/src/themes/silver/main/ts/ui/general/Notification.ts
```diff
@@ -14,7 +14,7 @@ import { Arr, Optional } from '@ephox/katamari';
```
lnewson, line 14: There's been an icon missed in this file, see line 180 where it uses `getFirst`.

- [ ] Fix missed icon

modules/tinymce/src/themes/silver/main/ts/ui/menus/item/structure/ItemSlices.ts
```diff
@@ -11,17 +11,12 @@ import { Optional } from '@ephox/katamari';
 
 import I18n from 'tinymce/core/api/util/I18n';
 
-import { get as getIcon, IconProvider } from '../../../icons/Icons';
+import { get as getIcon, IconProvider, render as renderIconElement } from '../../../icons/Icons';
```
lnewson, line 14: Can we also change this to `* as Icons` please? That'll also avoid having to come up with different names like we are now.

lnewson (approved):

ltrouton (approved):

ashfordneil (approved):

PR #7001, opened 2021-07-08
TINY-7570: Allow overrides to language extraction in robin's Zones APIs

ashfordneil:
Related Ticket: TINY-7570

Description of Changes:
* Add a new `getLanguage` function to boss' Universe
* Use that function everywhere we were previously looking at `.attr('lang')`

I'm targeting master here, so that we can get the changes released before 5.9 so that we can start using them elsewhere.
* This is a minor version bump of boss and robin, but that's okay because they've both received major version bumps on the develop branch

Pre-checks:
* [x] Changelog entry added
* [x] ~Branch prefixed with `feature/` for new features (if applicable)~ Technically it is a "new feature" but it's also targeting master so I kept this name.

Review:
* [x] Milestone set

### DIFF
--- a/modules/boss/CHANGELOG.md
+++ b/modules/boss/CHANGELOG.md
@@ -6,7 +6,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
 
-## 4.2.0 - 2021-07-09
+## 4.2.0 - 2021-07-12
 
 ### Added
 - Added `property().getLanguage` function to Universe #TINY-7570
--- a/modules/boss/CHANGELOG.md
+++ b/modules/boss/CHANGELOG.md
@@ -1,0 +1,12 @@
+# Changelog
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
+and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
+
+## Unreleased
+
+## 4.2.0 - 2021-07-09
+
+### Added
+- Added `property().getLanguage` method to Universe
--- a/modules/boss/CHANGELOG.md
+++ b/modules/boss/CHANGELOG.md
@@ -9,4 +9,4 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## 4.2.0 - 2021-07-09
 
 ### Added
-- Added `property().getLanguage` method to Universe
+- Added `property().getLanguage` function to Universe #TINY-7570
--- a/modules/boss/src/main/ts/ephox/boss/api/DomUniverse.ts
+++ b/modules/boss/src/main/ts/ephox/boss/api/DomUniverse.ts
@@ -48,13 +48,8 @@ export default (): Universe<SugarElement, Document> => {
     ], tag);
   };
 
-  const getLanguage = (element: SugarElement<Node>): Optional<string> => {
-    if (SugarNode.isElement(element)) {
-      return Attribute.getOpt(element, 'lang');
-    } else {
-      return Optional.none();
-    }
-  };
+  const getLanguage = (element: SugarElement<Node>): Optional<string> =>
+    SugarNode.isElement(element) ? Attribute.getOpt(element, 'lang') : Optional.none();
 
   return {
     up: Fun.constant({
--- a/modules/boss/src/main/ts/ephox/boss/api/DomUniverse.ts
+++ b/modules/boss/src/main/ts/ephox/boss/api/DomUniverse.ts
@@ -1,4 +1,4 @@
-import { Arr, Fun } from '@ephox/katamari';
+import { Arr, Fun, Optional } from '@ephox/katamari';
 import {
   Attribute, Compare, Css, Insert, InsertAll, PredicateFilter, PredicateFind, Remove, SelectorFilter, SelectorFind, SugarElement, SugarNode,
   SugarText, Traverse
@@ -46,6 +46,14 @@ export default (): Universe<SugarElement, Document> => {
     return Arr.contains([
       'script', 'noscript', 'iframe', 'noframes', 'noembed', 'title', 'style', 'textarea', 'xmp'
     ], tag);
+  };
+
+  const getLanguage = (element: SugarElement<Node>): Optional<string> => {
+    if (SugarNode.isElement(element)) {
+      return Attribute.getOpt(element, 'lang');
+    } else {
+      return Optional.none();
+    }
   };
 
   return {
@@ -103,6 +111,7 @@ export default (): Universe<SugarElement, Document> => {
       isComment: SugarNode.isComment,
       isElement: SugarNode.isElement,
       isSpecial,
+      getLanguage,
       getText: SugarText.get,
       setText: SugarText.set,
       isBoundary,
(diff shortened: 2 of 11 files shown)


modules/boss/src/main/ts/ephox/boss/api/DomUniverse.ts
```diff
@@ -48,6 +48,14 @@ export default (): Universe<SugarElement, Document> => {
     ], tag);
   };
 
+  const getLanguage = (element: SugarElement<Node>): Optional<string> => {
```
diogo405, line 51: I'd use a ternary here but it's just code style :)

`return SugarNode.isElement(element) ? Attribute.getOpt(element, 'lang') : Optional.none();`

modules/boss/src/main/ts/ephox/boss/mutant/Properties.ts
```diff
@@ -41,6 +41,10 @@ const isSpecial = (item: Gene): boolean => {
   return item.name === GeneTypes.Special;
 };
 
+const getLanguage = (item: Gene): Optional<string> => {
```
diogo405, line 44: Again: just code style, don't hate me: 

`const getLanguage = (item: Gene): Optional<string> => Obj.get(item.attrs, 'lang');`

diogo405 (approved):

modules/boss/src/main/ts/ephox/boss/api/DomUniverse.ts
```diff
@@ -48,6 +48,14 @@ export default (): Universe<SugarElement, Document> => {
...
+    if (SugarNode.isElement(element)) {
+      return Attribute.getOpt(element, 'lang');
+    } else {
+      return Optional.none();
+    }
+  };
```
lnewson, line 57:
```suggestion
  const getLanguage = (element: SugarElement<Node>): Optional<string> =>
    SugarNode.isElement(element) ? Attribute.getOpt(element, 'lang') : Optional.none();
```
I agree with Diogo, so just making the actual suggestion here.

modules/boss/CHANGELOG.md
```diff
@@ -0,0 +1,12 @@
...
+
+The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
+and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
+
+## Unreleased
+
+## 4.2.0 - 2021-07-09
+
+### Added
+- Added `property().getLanguage` method to Universe
```
lnewson, line 12:
```suggestion
- Added `property().getLanguage` function to Universe.
```
Changelogs should either have a JIRA reference or a fullstop at the end. Also method isn't used very often in the JS world, so function is probably more accurate.

modules/boss/src/main/ts/ephox/boss/mutant/Properties.ts
```diff
@@ -41,6 +41,10 @@ const isSpecial = (item: Gene): boolean => {
...
+  return Obj.get(item.attrs, 'lang');
+};
```
lnewson, line 46:
```suggestion
const getLanguage = (item: Gene): Optional<string> =>
  Obj.get(item.attrs, 'lang');
```
Just giving an actual suggestion for Diogo's comment, as he's right that we do prefer shorthand (though it should be over 2 lines).

modules/robin/CHANGELOG.md
```diff
@@ -0,0 +1,12 @@
...
+
+The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
+and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
+
+## Unreleased
+
+## 8.1.0 - 2021-07-09
+
+### Changed
+- TextZones APIs now use `universe.property().getLanguage` method instead of looking at `lang` attribute directly
```
lnewson, line 12:
```suggestion
- `TextZones` APIs now use `universe.property().getLanguage` function instead of looking at `lang` attribute directly.
```

modules/robin/src/main/ts/ephox/robin/zone/LanguageZones.ts
```diff
@@ -122,9 +122,10 @@ const nu = <E>(defaultLang: string): LanguageZones<E> => {
 //    (regardless of 'classic'/iframe or 'inline'/div mode).
 // Note: there may be descendant elements with a different language
 const calculate = <E, D>(universe: Universe<E, D>, item: E): Optional<string> => {
-  return universe.up().closest(item, '[lang]', Fun.never).bind((el) => {
-    const lang = universe.attrs().get(el, 'lang');
-    return lang === undefined ? Optional.none<string>() : Optional.some(lang);
+  const getLanguage = universe.property().getLanguage;
+  return getLanguage(item).orThunk(() => {
+    const ancestors = universe.up().all(item, Fun.never);
+    return Arr.findMap(ancestors, getLanguage);
```
lnewson, line 128:
```suggestion
  const props = universe.property();
  return props.getLanguage(item).orThunk(() => {
    const ancestors = universe.up().all(item, Fun.never);
    return Arr.findMap(ancestors, props.getLanguage);
```
This is rather strange to be storing a variable for a function and not helpful for debugging either. So just trying to change it to be more common with what might be done elsewhere, though I know it won't minify quite as well so not entirely sure about this...

modules/robin/src/test/ts/browser/LanguageOverrideTest.ts
```diff
@@ -0,0 +1,72 @@
...
+}
+
+describe('browser.robin.LanguageOverrideTest', () => {
+  const top = SugarElement.fromHtml('<div>' +
+    '<p lang="en">Hello <span>world</span></p>' +
+    '<p>This is <span lang="pt">multi-lingual content</span></p>' +
+    '</div>'
+  );
+
+  const walk = (universe: Universe<SugarElement, Document>): Array<AssertableZone> => {
```
lnewson, line 22:
```suggestion
  const walk = (universe: Universe<SugarElement, Document>): AssertableZone[] => {
```
`Array` should only be used for complex types (I miss the old tslint rule that could enforce this 😢)

versions.txt
```diff
@@ -1,2 +1,5 @@
 # List of packages to bump:
 # Format: [package_name]@[new_version]
+
+boss@4.2.0
```
lnewson, line 4: These should never be added when merging to master, as changes to master will be released in the same day. So please make sure to revert all changes to this file.

lnewson (approved):

modules/boss/CHANGELOG.md
ashfordneil:
```suggestion
- Added `property().getLanguage` function to Universe #TINY-7570
```

modules/robin/CHANGELOG.md
ashfordneil:
```suggestion
- `TextZones` APIs now use `universe.property().getLanguage` function instead of looking at `lang` attribute directly #TINY-7570
```

versions.txt
ashfordneil: Happy to revert, but how do I make sure we get minor bumps instead of just patch fixes? Do I edit the package.json versions manually?

lnewson: See the contributing docs... https://github.com/tinymce/tinymce/blob/develop/CONTRIBUTING.md#publishing-process. So in this case you'd run `yarn lerna version` after this has been merged to do the minor version bumps.

ltrouton (approved):

PR #7007, opened 2021-07-12
TINY-7663: Fix selection location when undoing or redoing in a document with CEF elements

JamesToohey:
**Related Ticket:**

TINY-7663

**Description of Changes:**

When setting a bookmark, we now check to see if the selection is within a fake caret. If it _isn't_, we need to ensure that when we call `setContent` when restoring the content that a fake caret is not created (this will happen if the first element in the content is `contenteditable="false"`) by `moveSelection` in `SetContentImpl.ts`. 

If a fake caret is created - as it was previously - as this will mean when the selection is restored in `Levels.ts` that the path will be off by one.

**Pre-checks:**
* [x] Changelog entry added

**Review:**
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/content/SetContentImpl.ts
+++ b/modules/tinymce/src/core/main/ts/content/SetContentImpl.ts
@@ -35,9 +35,11 @@ const moveSelection = (editor: Editor): void => {
   }
 };
 
-const setEditorHtml = (editor: Editor, html: string): void => {
+const setEditorHtml = (editor: Editor, html: string, noSelection: boolean = false): void => {
   editor.dom.setHTML(editor.getBody(), html);
-  moveSelection(editor);
+  if (!noSelection) {
+    moveSelection(editor);
+  }
 };
 
 const setContentString = (editor: Editor, body: HTMLElement, content: string, args: SetContentArgs): string => {
@@ -65,7 +67,7 @@ const setContentString = (editor: Editor, body: HTMLElement, content: string, ar
       content = '<br data-mce-bogus="1">';
     }
 
-    setEditorHtml(editor, content);
+    setEditorHtml(editor, content, args.no_selection);
 
     editor.fire('SetContent', args);
   } else {
@@ -78,7 +80,7 @@ const setContentString = (editor: Editor, body: HTMLElement, content: string, ar
     }
 
     args.content = isWsPreserveElement(SugarElement.fromDom(body)) ? content : Tools.trim(content);
-    setEditorHtml(editor, args.content);
+    setEditorHtml(editor, args.content, args.no_selection);
 
     if (!args.no_events) {
       editor.fire('SetContent', args);
--- a/modules/tinymce/src/core/main/ts/content/SetContentImpl.ts
+++ b/modules/tinymce/src/core/main/ts/content/SetContentImpl.ts
@@ -35,9 +35,9 @@ const moveSelection = (editor: Editor): void => {
   }
 };
 
-const setEditorHtml = (editor: Editor, html: string, noSelection: boolean = false): void => {
+const setEditorHtml = (editor: Editor, html: string, noSelection: boolean | undefined): void => {
   editor.dom.setHTML(editor.getBody(), html);
-  if (!noSelection) {
+  if (noSelection !== true) {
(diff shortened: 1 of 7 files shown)


modules/tinymce/src/core/main/ts/content/SetContentImpl.ts
```diff
@@ -35,9 +35,11 @@ const moveSelection = (editor: Editor): void => {
   }
 };
 
-const setEditorHtml = (editor: Editor, html: string): void => {
+const setEditorHtml = (editor: Editor, html: string, noSelection: boolean = false): void => {
```
lnewson, line 38: This isn't really accurate, as we're always passing a value here. So can we please update the types to reflect that?
```suggestion
const setEditorHtml = (editor: Editor, html: string, noSelection: boolean | undefined): void => {
```

modules/tinymce/src/core/main/ts/content/SetContentImpl.ts
```diff
@@ -35,9 +35,11 @@ const moveSelection = (editor: Editor): void => {
...
   editor.dom.setHTML(editor.getBody(), html);
-  moveSelection(editor);
+  if (!noSelection) {
```
lnewson, line 40: This will also need changing if we do the above, e.g.:
```suggestion
  if (noSelection !== true) {
```

modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
```diff
@@ -564,4 +565,17 @@ describe('browser.tinymce.core.UndoManager', () => {
     editor.execCommand('mceFocus');
     assert.isFalse(editor.undoManager.hasUndo());
   });
+
+  it('TINY-7663: Undo when first element is contenteditable="false" should restore correct cursor location', () => {
```
lnewson, line 569: Can we add a test where the cursor is also in the fake caret when the undo bookmark is created please? or does that already exist in this file already (I couldn't see one)?

modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
```diff
@@ -564,4 +565,41 @@ describe('browser.tinymce.core.UndoManager', () => {
     editor.execCommand('mceFocus');
     assert.isFalse(editor.undoManager.hasUndo());
   });
+
+  context('Undo when first element is contenteditable="false"', () => {
+    beforeEach(() => {
+      const editor = hook.editor();
+      editor.undoManager.clear();
+      editor.setDirty(false);
+      editor.setContent('<div contenteditable="false"><p>CEF</p></div><p>something</p><p>something else</p>', { format: 'raw' });
```
lnewson, line 574:
```suggestion
      editor.resetContent('<div contenteditable="false"><p>CEF</p></div><p>something</p><p>something else</p>');
```
That should be the same? `resetContent` won't allow us to pass as raw, however there's nothing we are including here where raw is required.

JamesToohey: Looks like this works, nice

ashfordneil (approved):

ltrouton (approved):

modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
```diff
@@ -564,4 +565,39 @@ describe('browser.tinymce.core.UndoManager', () => {
...
+      editor.resetContent('<div contenteditable="false"><p>CEF</p></div><p>something</p><p>something else</p>');
+      editor.focus();
+    });
+
+    it('TINY-7663: No fake caret - should restore correct cursor location', () => {
+      const editor = hook.editor();
+      TinyAssertions.assertContentPresence(editor, { 'p[data-mce-caret=before]': 1 });
+      // selection path must include fake caret which is before the CEF div
+      TinySelections.setCursor(editor, [ 3, 0 ], 14);
+      // moving the selection removed the CEF div
```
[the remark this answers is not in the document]
JamesToohey: Sorry this comment is bad, it removes the fake caret - good catch, I'll fix it

lnewson (approved):

PR #7008, opened 2021-07-12
TINY-7690: None-entries are now selected if none other is.

HAFRMO:
Related Ticket:

Description of Changes:
For table lists with a None-entry, it now selects that entry if no other entry could be selected.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/plugins/table/main/ts/ui/UiUtils.ts
+++ b/modules/tinymce/src/plugins/table/main/ts/ui/UiUtils.ts
@@ -5,7 +5,7 @@
  * For commercial licenses see https://www.tiny.cloud/
  */
 
-import { Arr, Singleton, Strings } from '@ephox/katamari';
+import { Arr, Optional, Singleton, Strings } from '@ephox/katamari';
 
 import Editor from 'tinymce/core/api/Editor';
 import { Menu, Toolbar } from 'tinymce/core/api/ui/Ui';
@@ -14,12 +14,12 @@ interface Item {
   readonly value: string;
 }
 
-const onSetupToggle = (editor: Editor, formatName: string, formatValue: string) => {
+const onSetupToggle = (editor: Editor, formatName: string, active: () => boolean) => {
   return (api: Toolbar.ToolbarMenuButtonInstanceApi) => {
     const boundCallback = Singleton.unbindable();
 
     const init = () => {
-      api.setActive(editor.formatter.match(formatName, { value: formatValue }));
+      api.setActive(active());
       const binding = editor.formatter.formatChanged(formatName, api.setActive);
       boundCallback.set(binding);
     };
@@ -31,6 +31,14 @@ const onSetupToggle = (editor: Editor, formatName: string, formatValue: string) 
   };
 };
 
+const onSetupToggleSingle = (editor: Editor, formatName: string, formatValue: string) => {
+  return onSetupToggle(editor, formatName, () => editor.formatter.match(formatName, { value: formatValue }));
+};
+
+const onSetupToggleInverse = (editor: Editor, formatName: string, formatValues: string[]) => {
+  return onSetupToggle(editor, formatName, () => !Arr.exists(formatValues, (value) => editor.formatter.match(formatName, { value })));
+};
+
 const applyTableCellStyle = <T extends Item>(editor: Editor, style: string) =>
   (item: T) =>
     editor.execCommand('mceTableApplyCellStyle', false, { [style]: item.value });
@@ -38,15 +46,25 @@ const applyTableCellStyle = <T extends Item>(editor: Editor, style: string) =>
 const filterNoneItem = <T extends Item>(list: T[]) =>
   Arr.filter(list, (item) => Strings.isNotEmpty(item.value));
 
-const generateItem = <T extends Item>(editor: Editor, item: T, format: string, extractText: (item: T) => string, onAction: (item: T) => void): Menu.ToggleMenuItemSpec => ({
+const generateItem = <T extends Item>(editor: Editor, item: T, otherValues: Optional<string[]>, format: string, extractText: (item: T) => string, onAction: (item: T) => void): Menu.ToggleMenuItemSpec => ({
   text: extractText(item),
   type: 'togglemenuitem',
   onAction: () => onAction(item),
-  onSetup: onSetupToggle(editor, format, item.value)
+  onSetup: otherValues.fold(
+    () => onSetupToggleSingle(editor, format, item.value),
+    (values) => onSetupToggleInverse(editor, format, values)
+  )
 });
 
 const generateItems = <T extends Item>(editor: Editor, items: T[], format: string, extractText: (item: T) => string, onAction: (item: T) => void): Menu.ToggleMenuItemSpec[] =>
-  Arr.map(items, (item) => generateItem(editor, item, format, extractText, onAction));
+  Arr.map(items, (item) => {
+    if (item.value === '') {
+      const otherValues = Optional.some(Arr.map(filterNoneItem(items), (itemValue) => itemValue.value));
+      return generateItem(editor, item, otherValues, format, extractText, onAction);
+    } else {
+      return generateItem(editor, item, Optional.none(), format, extractText, onAction);
+    }
+  });
 
 const generateItemsCallback = <T extends Item>(editor: Editor, items: T[], format: string, extractText: (item: T) => string, onAction: (item: T) => void) =>
   (callback: (items: Menu.ToggleMenuItemSpec[]) => void) =>
--- a/modules/tinymce/src/plugins/table/main/ts/ui/UiUtils.ts
+++ b/modules/tinymce/src/plugins/table/main/ts/ui/UiUtils.ts
@@ -31,13 +31,11 @@ const onSetupToggle = (editor: Editor, formatName: string, active: () => boolean
   };
 };
 
-const onSetupToggleSingle = (editor: Editor, formatName: string, formatValue: string) => {
-  return onSetupToggle(editor, formatName, () => editor.formatter.match(formatName, { value: formatValue }));
-};
+const onSetupToggleSingle = (editor: Editor, formatName: string, formatValue: string) =>
+  onSetupToggle(editor, formatName, () => editor.formatter.match(formatName, { value: formatValue }));
 
-const onSetupToggleInverse = (editor: Editor, formatName: string, formatValues: string[]) => {
-  return onSetupToggle(editor, formatName, () => !Arr.exists(formatValues, (value) => editor.formatter.match(formatName, { value })));
-};
+const onSetupToggleInverse = (editor: Editor, formatName: string, formatValues: string[]) =>
+  onSetupToggle(editor, formatName, () => !Arr.exists(formatValues, (value) => editor.formatter.match(formatName, { value })));
 
 const applyTableCellStyle = <T extends Item>(editor: Editor, style: string) =>
   (item: T) =>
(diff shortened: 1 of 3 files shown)


modules/tinymce/src/plugins/table/main/ts/ui/UiUtils.ts
```diff
@@ -31,22 +31,40 @@ const onSetupToggle = (editor: Editor, formatName: string, formatValue: string)
...
+  onSetup: otherValues.fold(
+    () => onSetupToggleSingle(editor, format, item.value),
+    (values) => onSetupToggleInverse(editor, format, values)
+  )
 });
 
 const generateItems = <T extends Item>(editor: Editor, items: T[], format: string, extractText: (item: T) => string, onAction: (item: T) => void): Menu.ToggleMenuItemSpec[] =>
-  Arr.map(items, (item) => generateItem(editor, item, format, extractText, onAction));
+  Arr.map(items, (item) => {
+    if (item.value === '') {
```
lnewson, line 61: This should be using the `Strings.isEmpty` module from Katamari instead.

modules/tinymce/src/plugins/table/main/ts/ui/UiUtils.ts
```diff
@@ -14,12 +14,12 @@ interface Item {
...
 
-const onSetupToggle = (editor: Editor, formatName: string, formatValue: string) => {
+const onSetupToggle = (editor: Editor, formatName: string, active: () => boolean) => {
   return (api: Toolbar.ToolbarMenuButtonInstanceApi) => {
     const boundCallback = Singleton.unbindable();
 
     const init = () => {
-      api.setActive(editor.formatter.match(formatName, { value: formatValue }));
+      api.setActive(active());
       const binding = editor.formatter.formatChanged(formatName, api.setActive);
```
lnewson, line 23: This change doesn't handle the `formatChanged` case, as that needs to call `api.setActive(true)` when nothing matches as well, as the above call only handles the very first init and doesn't handle cases after that.

Worth noting `formatChanged` has a `similar` argument that you could pass to be `true` when the format value is empty. You can probably then check if that returns false, then it means nothing that uses the same format with a different value matches. If not, you might be able to use your `active` callback here, however I do have concerns about the performance impacts of that, as it needs to check every value is not set. What we nearly need to be able to do this in a performant way is to be able to pass `similar` to the match function as well.

modules/tinymce/src/plugins/table/main/ts/ui/UiUtils.ts
```diff
@@ -31,22 +31,40 @@ const onSetupToggle = (editor: Editor, formatName: string, formatValue: string)
   };
 };
 
+const onSetupToggleSingle = (editor: Editor, formatName: string, formatValue: string) => {
+  return onSetupToggle(editor, formatName, () => editor.formatter.match(formatName, { value: formatValue }));
+};
```
lnewson, line 36:
```suggestion
const onSetupToggleSingle = (editor: Editor, formatName: string, formatValue: string) =>
  onSetupToggle(editor, formatName, () => editor.formatter.match(formatName, { value: formatValue }));
```
Shorthand functions please

modules/tinymce/src/plugins/table/main/ts/ui/UiUtils.ts
```diff
@@ -31,22 +31,40 @@ const onSetupToggle = (editor: Editor, formatName: string, formatValue: string)
...
+
+const onSetupToggleInverse = (editor: Editor, formatName: string, formatValues: string[]) => {
+  return onSetupToggle(editor, formatName, () => !Arr.exists(formatValues, (value) => editor.formatter.match(formatName, { value })));
+};
```
lnewson, line 40:
```suggestion
const onSetupToggleInverse = (editor: Editor, formatName: string, formatValues: string[]) =>
  onSetupToggle(editor, formatName, () => !Arr.exists(formatValues, (value) => editor.formatter.match(formatName, { value })));
```

lnewson (changes requested):
Good work getting this far, however some parts are still missing so requesting changes.

modules/tinymce/src/plugins/table/main/ts/ui/UiUtils.ts
```diff
@@ -17,10 +17,14 @@ interface Item {
...
     const boundCallback = Singleton.unbindable();
+    const isNone = Strings.isEmpty(formatValue);
 
     const init = () => {
-      api.setActive(editor.formatter.match(formatName, { value: formatValue }));
-      const binding = editor.formatter.formatChanged(formatName, api.setActive);
+      const setActive = (matched) =>
+        api.setActive(isNone ? !matched : matched);
+
+      setActive(editor.formatter.match(formatName, { value: formatValue }, undefined, isNone));
```
ashfordneil, line 26:
```suggestion
      // For most checkboxes, we want to make sure that the formatter _does_ match, exactly
      // For the None checkbox, we want to make sure that the formatter _does not_ match, even for similar values
      setActive(editor.formatter.match(formatName, { value: formatValue }, undefined, isNone));
```

Can we add a comment like this (or something else) to explain the logic here? It feels like the sort of thing that's going to be very confusing without the current context

ashfordneil (approved):

modules/tinymce/src/plugins/table/main/ts/ui/UiUtils.ts
```diff
@@ -17,10 +17,15 @@ interface Item {
 const onSetupToggle = (editor: Editor, formatName: string, formatValue: string) => {
   return (api: Toolbar.ToolbarMenuButtonInstanceApi) => {
     const boundCallback = Singleton.unbindable();
+    const isNone = Strings.isEmpty(formatValue);
 
     const init = () => {
-      api.setActive(editor.formatter.match(formatName, { value: formatValue }));
-      const binding = editor.formatter.formatChanged(formatName, api.setActive);
+      // If value is empty (A None-entry in the list), check it only if the format is not set at all. Otherwise, check it if the format is set to the correct value.
+      const setActive = (matched) =>
```
lnewson, line 24:
```suggestion
      const setActive = (matched: boolean) =>
```

modules/tinymce/src/plugins/table/main/ts/ui/UiUtils.ts
```diff
@@ -17,10 +17,15 @@ interface Item {
 const onSetupToggle = (editor: Editor, formatName: string, formatValue: string) => {
   return (api: Toolbar.ToolbarMenuButtonInstanceApi) => {
     const boundCallback = Singleton.unbindable();
+    const isNone = Strings.isEmpty(formatValue);
 
     const init = () => {
-      api.setActive(editor.formatter.match(formatName, { value: formatValue }));
-      const binding = editor.formatter.formatChanged(formatName, api.setActive);
+      // If value is empty (A None-entry in the list), check it only if the format is not set at all. Otherwise, check it if the format is set to the correct value.
```
lnewson, line 23:
```suggestion
      // If value is empty (A None-entry in the list), check if the format is not set at all. Otherwise, check if the format is set to the correct value.
```

modules/tinymce/src/plugins/table/main/ts/ui/UiUtils.ts
```diff
@@ -17,10 +17,15 @@ interface Item {
...
+      const setActive = (matched) =>
+        api.setActive(isNone ? !matched : matched);
+
+      setActive(editor.formatter.match(formatName, { value: formatValue }, undefined, isNone));
+      const binding = editor.formatter.formatChanged(formatName, setActive, isNone);
```
lnewson, line 28: Can you please add a note about this highlighting all items and add a TODO to TINY-7713?

lnewson (approved):

modules/tinymce/src/plugins/table/main/ts/ui/UiUtils.ts
```diff
@@ -46,7 +51,8 @@ const generateItem = <T extends Item>(editor: Editor, item: T, format: string, e
 });
 
 const generateItems = <T extends Item>(editor: Editor, items: T[], format: string, extractText: (item: T) => string, onAction: (item: T) => void): Menu.ToggleMenuItemSpec[] =>
-  Arr.map(items, (item) => generateItem(editor, item, format, extractText, onAction));
+  Arr.map(items, (item) =>
+    generateItem(editor, item, format, extractText, onAction));
```
JamesToohey, line 55: Can we put this back on one line?

```suggestion
  Arr.map(items, (item) => generateItem(editor, item, format, extractText, onAction));
```

modules/tinymce/src/plugins/table/test/ts/module/test/TableModifiersTestUtils.ts
```diff
@@ -151,14 +151,14 @@ const pAssertStyleCanBeToggledWithoutCheckmarks = async (editor: Editor, options
 const pAssertStyleCanBeToggled = async (editor: Editor, options: AssertStyleOptionsWithCheckmarks, useMenuOrToolbar: 'toolbar' | 'menuitem') => {
   const sugarContainer = SugarBody.body();
   setEditorContentTableAndSelection(editor, options.rows, options.columns);
-  await pAssertNoCheckmarksInMenu(editor, options.menuTitle, options.checkMarkEntries, sugarContainer, useMenuOrToolbar);
+  await pAssertCheckmarkOn(editor, options.menuTitle, options.subMenuRemoveTitle, options.checkMarkEntries - 1, sugarContainer, useMenuOrToolbar);
```
JamesToohey, line 154: Comment for `- 1` magic number?

JamesToohey (approved):

modules/tinymce/src/plugins/table/test/ts/module/test/TableModifiersTestUtils.ts
HAFRMO: It was repeated in a few places so I moved it and changed the variable name. It should be clear what it is now.

modules/tinymce/src/plugins/table/main/ts/ui/UiUtils.ts
HAFRMO: I'd find it better on two, but sure.

modules/tinymce/src/plugins/table/main/ts/ui/UiUtils.ts
```diff
@@ -17,10 +17,16 @@ interface Item {
...
 
     const init = () => {
-      api.setActive(editor.formatter.match(formatName, { value: formatValue }));
-      const binding = editor.formatter.formatChanged(formatName, api.setActive);
+      // If value is empty (A None-entry in the list), check if the format is not set at all. Otherwise, check if the format is set to the correct value.
+      const setActive = (matched: boolean) =>
+        api.setActive(isNone ? !matched : matched);
+
+      setActive(editor.formatter.match(formatName, { value: formatValue }, undefined, isNone));
+      // formatChanged currently incorrectly highlights items. TODO TINY-7713
```
lnewson, line 28:
```suggestion
      // TODO: TINY-7713: formatChanged doesn't currently handle formats with dynamic values so this will currently cause all items to show as active
```
TODO should always be at the start, not at the end.

Also, I wonder if we should just have this commented out for now? As what's better not updating or showing all items as active?

HAFRMO: I commented it out. It's not likely to occur to begin with, so "Not updating" is probably better than "Updating incorrectly", as you said.

PR #6113, opened 2020-10-16
TINY-6479: Bold/Italic buttons not enabled when toggling caret format in rtc mode

spocke:
Related Ticket:

Description of Changes:
* Adds a new public closest api to formatter that isn't dom specific and redirects that to rtc.
* Redirects formatter.match to rtc.
* Redirects formatter.formatChanged to rtc.
* Contains rtc ignore for addVisual since it updates the dom view did some minor cleanup on that as well.

Pre-checks:
* [x] Changelog entry added
* [x] Tests have been added (if applicable)
* [x] Branch prefixed with `feature/` for new features (if applicable)
* [x] License headers added on new files (if applicable)

Review:
* [x] Milestone set
* [x] Review comments resolved

GitHub issues (if applicable):

PR #7009, opened 2021-07-13
TINY-7393: Fixed formats not applied correctly to nested list items

lnewson:
**Related Ticket:** TINY-7393

**Description of Changes:**
* Fixes an issue where selector formats (eg center alignment) wouldn't work due to how the ranges are walked (only nodes that are fully selected are returned). This expands on the original fix done in: https://github.com/tinymce/tinymce/pull/6666
* As part of this, I tried a number of solutions, one of which was making range walker return all the leaf nodes (which did work but ran into other edge case issues). As part of that I did a fair amount of clean up and wrote some additional tests, as `RangeWalk` has no tests currently, so this includes all of those clean ups/improvements as well (as discussed at standup).

Note: There's 2 commits to help with reviewing the different parts above.

**Additional information:**
So for those not familiar with formats, TinyMCE has 3 types of formats: block, inline and selector. Both block and inline formats will wrap/replace the relevant nodes within the selection. Selector formats are different, as they match existing nodes and mutate them instead.

Now, the most useful background information to know here is that when applying the format, it'll take the current selection and expand it to match the format being applied for block and selector formats (both are expanded differently). So in the selector case, it'll expand the start and end container until it find a nodes that matches the selector provided in the format. So for lists, it'll expand out to the closest `li` elements. So the problem here, is since the selection expands to the closest node that matches, the entire list item isn't in the selection so only the nodes that are fully in the selection are processed when walking.

So this solution expands on the previous one, which takes each node processed and firstly tries to apply the format directly to the node. If the format matches, then nothing more is done. However, if it's not matched then we walk up to the parent and also try to apply the selection there (which in this case will be the `li` elements). We need to make sure we don't do this however, if the selector format is configured to not expand.

**Pre-checks:**
* [x] Changelog entry added

**Review:**
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -48,7 +48,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Partially transparent RGBA values provided in the `color_map` setting were given the wrong hex value #TINY-7163
 - HTML comments with mismatched quotes were parsed incorrectly under certain circumstances #TINY-7589
 - Links in notification text did not show the correct mouse pointer #TINY-7661
-- Applying selector formats would sometimes not apply the format correctly on lists #TINY-7393
+- Applying selector formats would sometimes not apply the format correctly to elements in a list #TINY-7393
 - The formatter match APIs were incorrectly returning false for formats that specified an attribute or style should be removed from an element #TINY-6149
 - The type signature on `editor.formatter.matchNode` had the wrong return type (was `boolean` but should have been `Formatter | undefined`) #TINY-6149
 
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -48,6 +48,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Partially transparent RGBA values provided in the `color_map` setting were given the wrong hex value #TINY-7163
 - HTML comments with mismatched quotes were parsed incorrectly under certain circumstances #TINY-7589
 - Links in notification text did not show the correct mouse pointer #TINY-7661
+- Applying selector formats would sometimes not apply the format correctly on lists #TINY-7393
 - The formatter match APIs were incorrectly returning false for formats that specified an attribute or style should be removed from an element #TINY-6149
 - The type signature on `editor.formatter.matchNode` had the wrong return type (was `boolean` but should have been `Formatter | undefined`) #TINY-6149
 
--- a/modules/tinymce/src/core/main/ts/Rtc.ts
+++ b/modules/tinymce/src/core/main/ts/Rtc.ts
@@ -45,7 +45,7 @@ interface RtcRuntimeApi {
     match: (format: string, vars: Record<string, string>) => boolean;
     matchAll: () => string[];
     matchNode: () => Format | undefined;
-    closest: (formats: string) => string;
+    closest: (formats: string[]) => string;
     apply: (format: string, vars: Record<string, string>) => void;
     remove: (format: string, vars: Record<string, string>) => void;
     toggle: (format: string, vars: Record<string, string>) => void;
@@ -372,7 +372,7 @@ export const canApplyFormat = (
 
 export const closestFormat = (
   editor: Editor,
-  names: string): string => getRtcInstanceWithError(editor).formatter.closest(names);
+  names: string[]): string => getRtcInstanceWithError(editor).formatter.closest(names);
 
 export const applyFormat = (
   editor: Editor,
--- a/modules/tinymce/src/core/main/ts/api/dom/DOMUtils.ts
+++ b/modules/tinymce/src/core/main/ts/api/dom/DOMUtils.ts
@@ -1205,15 +1205,21 @@ const DOMUtils = (doc: Document, settings: Partial<DOMUtilsSettings> = {}): DOMU
   };
 
   const isChildOf = (node: Node, parent: Node) => {
-    while (node) {
-      if (parent === node) {
-        return true;
+    // IE only supports the native `contains` functions on HTMLElements, but not Nodes or Elements
+    // so we have to use the slow walking path for IE unfortunately.
+    if (!isIE) {
+      return node === parent || parent.contains(node);
+    } else {
+      while (node) {
+        if (parent === node) {
+          return true;
+        }
+
+        node = node.parentNode;
       }
 
-      node = node.parentNode;
+      return false;
     }
-
-    return false;
   };
 
   const dumpRng = (r: Range) => (
--- a/modules/tinymce/src/core/main/ts/fmt/RemoveFormat.ts
+++ b/modules/tinymce/src/core/main/ts/fmt/RemoveFormat.ts
@@ -520,7 +520,7 @@ const remove = (ed: Editor, name: string, vars?: FormatVars, node?: Node | Range
     }
 
     // Note: Assists with cleaning up any stray text decorations that may been applied when text decorations
-    // and text colors were merged together from a applied format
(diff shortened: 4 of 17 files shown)


modules/tinymce/src/core/main/ts/Rtc.ts
```diff
@@ -372,7 +372,7 @@ export const canApplyFormat = (
 
 export const closestFormat = (
   editor: Editor,
-  names: string): string => getRtcInstanceWithError(editor).formatter.closest(names);
+  names: string[]): string => getRtcInstanceWithError(editor).formatter.closest(names);
```
lnewson, line 375: This incorrect type was something we missed when originally reviewing this in https://github.com/tinymce/tinymce/pull/6113/. I noticed it when I added the missing type in `Formatter.ts`.

modules/tinymce/src/core/main/ts/api/dom/DOMUtils.ts
```diff
@@ -1205,15 +1205,20 @@ const DOMUtils = (doc: Document, settings: Partial<DOMUtilsSettings> = {}): DOMU
   };
 
   const isChildOf = (node: Node, parent: Node) => {
-    while (node) {
-      if (parent === node) {
-        return true;
+    // IE only supports the native `contains` functions on Elements, not Nodes
+    if (!isIE || NodeType.isElement(parent)) {
```
lnewson, line 1209: This just makes sure we use the native `Node.contains` function if it's available/viable as it's ~5x faster on Chrome.

Note: I made this change to account for the fact I switched to using `isChildOf` in `RangeWalk` which used to cap how far it'd walk, so using `isChildOf` would have had worse performance (it's also just a good modernisation change).

modules/tinymce/src/core/main/ts/fmt/RemoveFormat.ts
```diff
@@ -517,6 +518,24 @@ const remove = (ed: Editor, name: string, vars?: FormatVars, node?: Node | Range
         }
       }
     }
+
+    // Note: Assists with cleaning up any stray text decorations that may been applied when text decorations
+    // and text colors were merged together from a applied format
+    // Remove child span if it only contains text-decoration and a parent node also has the same text decoration.
+    const textDecorations = [ 'underline', 'line-through', 'overline' ];
```
lnewson, line 525: I moved this, as it wasn't running in some cases when it really should have been from what I could tell.

modules/tinymce/src/core/test/ts/browser/selection/RangeWalkTest.ts
```diff
@@ -0,0 +1,113 @@
...
+
+    const rng = editor.selection.getRng();
+    RangeWalk.walk(editor.dom, rng, (nodes) => {
+      nodeNames.push(Arr.map(nodes, (node) => node.nodeName.toLowerCase()));
+    });
+
+    assert.deepEqual(nodeNames, expected);
+  };
+
+  it('TBD: should only include itself if the full node is selected', () => {
```
lnewson, line 42: These are `TBD`s as it's not applicable to this change and just tests the existing functionality.

modules/tinymce/src/core/main/ts/api/dom/DOMUtils.ts
ashfordneil: Are we also able to short-circuit in the case where parent isn't an `Element`? If not, still a good win

lnewson: Not sure what you mean there sorry? So on every browser except IE 11 it'll use the `Node.contains` path and for IE 11 it'll use `Element.contains` so it'd only do the slower walking approach if the passed in parent is not an `Element`.

modules/tinymce/CHANGELOG.md
```diff
@@ -48,6 +48,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Partially transparent RGBA values provided in the `color_map` setting were given the wrong hex value #TINY-7163
 - HTML comments with mismatched quotes were parsed incorrectly under certain circumstances #TINY-7589
 - Links in notification text did not show the correct mouse pointer #TINY-7661
+- Applying selector formats would sometimes not apply the format correctly on lists #TINY-7393
```
JamesToohey, line 51: Maybe?

```suggestion
- Applying selector formats would sometimes not apply the format correctly to lists #TINY-7393
```

or even

```suggestion
- Applying selector formats would sometimes not apply the format correctly to elements in a list #TINY-7393
```

modules/tinymce/src/core/main/ts/fmt/RemoveFormat.ts
```diff
@@ -517,6 +518,24 @@ const remove = (ed: Editor, name: string, vars?: FormatVars, node?: Node | Range
         }
       }
     }
+
+    // Note: Assists with cleaning up any stray text decorations that may been applied when text decorations
+    // and text colors were merged together from a applied format
```
JamesToohey, line 523:
```suggestion
    // and text colors were merged together from an applied format
```

JamesToohey (approved):

modules/tinymce/src/core/main/ts/api/dom/DOMUtils.ts
lnewson: Turned out this is why my tests were failing... IE 11 doesn't handle `node` being a text node, or similar. As such I've just changed it to always use the slow path on IE since it also turns out it doesn't work with non HTML elements (eg SVGs).

modules/tinymce/src/core/test/ts/browser/selection/RangeWalkTest.ts
ashfordneil: Elsewhere I think we use `TBA`, but either way didn't we agree we would stop adding those for tests that we didn't know what Jira they were associated with?

lnewson: You're correct on both accounts... I shall remove them.

- [ ] Remove `TBD` prefixes

ashfordneil (approved):

modules/tinymce/src/core/main/ts/selection/RangeNodes.ts
```diff
@@ -17,18 +21,27 @@ const getSelectedNode = (range: Range): Node => {
...
+    const childNodes = container.childNodes;
+    const safeOffset = Num.clamp(offset, 0, childNodes.length - 1);
+    return childNodes[safeOffset];
+  } else {
+    return container;
   }
+};
 
-  return container;
+/** @deprecated Use getNode instead */
```
metricjs, line 33: We're deprecating this but also using it in CaretPosition above?

metricjs (approved):
Admittedly I'm not clear on every nuance of this, but it makes enough sense that I'm comfortable ticking it.

modules/tinymce/src/core/main/ts/selection/RangeNodes.ts
lnewson: Yes, it's because `CaretPosition` has always needed this logic and we can't change it without breaking a heap of things. I merged the safer option from `RangeWalk` which clamps the offset, which is now the replacement `getNode` since it's safer. So this just makes sure we don't attempt to use the unsafe version in future.

modules/tinymce/src/core/test/ts/browser/selection/RangeWalkTest.ts
lnewson: Fixed

modules/tinymce/CHANGELOG.md
lnewson: Went with the second suggestion.

PR #7012, opened 2021-07-14
TINY-7665: Update the context toolbar to hide the bubble arrow when using an inset layout

lnewson:
Related Ticket: TINY-7665

Description of Changes:
* Renamed `LayoutInside` to `LayoutInset` since that's the relevant CSS term and what we've been calling it more lately.
* Make `Bubble` include a new inset class when generating a bubble to be used for an inset layout.
* Changed context toolbars to use the new `tox-pop--inset` class when using an inset layout. This class will hide the bubble arrow. Additionally to match the gif provided I've halved the bubble for insets.
  * Note: I've intentionally used `opacity` here instead of `display: none` as there was some talk about making it transition as it disappears.
* Updated the tests to properly check if the toolbar is positioned so that it's overlapping (this couldn't be done previously since we didn't have a selector that was different). Also took `6px` away for the positioning assertions of the toolbars due to the reduced bubble.

image

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/alloy/src/main/ts/ephox/alloy/positioning/layout/Bubble.ts
+++ b/modules/alloy/src/main/ts/ephox/alloy/positioning/layout/Bubble.ts
@@ -16,14 +16,14 @@ export interface Bubble {
   north: () => BubbleInstance;
   east: () => BubbleInstance;
   west: () => BubbleInstance;
-  innerSoutheast: () => BubbleInstance;
-  innerSouthwest: () => BubbleInstance;
-  innerNorthwest: () => BubbleInstance;
-  innerNortheast: () => BubbleInstance;
-  innerSouth: () => BubbleInstance;
-  innerNorth: () => BubbleInstance;
-  innerEast: () => BubbleInstance;
-  innerWest: () => BubbleInstance;
+  insetSoutheast: () => BubbleInstance;
+  insetSouthwest: () => BubbleInstance;
+  insetNorthwest: () => BubbleInstance;
+  insetNortheast: () => BubbleInstance;
+  insetSouth: () => BubbleInstance;
+  insetNorth: () => BubbleInstance;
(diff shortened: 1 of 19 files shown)


modules/alloy/src/main/ts/ephox/alloy/positioning/layout/Bubble.ts
```diff
@@ -56,10 +59,15 @@ const allAlignments: Array<keyof BubbleAlignments> = [
...
-  'right'
+  'right',
+
+  'inset'
 ];
 
-const nu = (width: number, yoffset: number, classes: BubbleAlignments): Bubble => {
+const nu = (width: number, yoffset: number, classes: BubbleAlignments, insetModifier: number = 1): Bubble => {
+  const insetWidth = width * insetModifier;
+  const insetYoffset = yoffset * insetModifier;
```
diogo405, line 69: I'd name the variable `yOffset` rather than `yoffset`. Just easier to read.

diogo405 (approved):

ltrouton (approved):

modules/alloy/src/main/ts/ephox/alloy/positioning/layout/Bubble.ts
lnewson: Yeah, I don't like the names of either of these variables to be honest. I just kept it the same as it already was for consistency.

ashfordneil (approved):

modules/alloy/src/main/ts/ephox/alloy/positioning/layout/Bubble.ts
lnewson: Renamed them to `xOffset` and `yOffset` (as well as doing the same with the inset variables).

PR #7024, opened 2021-07-20
TINY-7729: Use Utils.link() to both create and edit links from the quicklink bar

TheSpyder:
Use `Utils.link()` to both create and edit links from the quicklink bar.

Related Ticket: TINY-7729

Description of Changes:
* No user-visible changes, so no changelog, just redirecting quicklink code into existing code for create/edit link that knows about RTC.

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/plugins/link/main/ts/core/Utils.ts
+++ b/modules/tinymce/src/plugins/link/main/ts/core/Utils.ts
@@ -209,7 +209,7 @@ const unlinkDomMutation = (editor: Editor) => {
   });
 };
 
-/**
+/*
  * RTC uses unwrapped options.
  *
  * To best simulate this, we unwrap to null and filter out empty values.
--- a/modules/tinymce/src/plugins/link/main/ts/core/Utils.ts
+++ b/modules/tinymce/src/plugins/link/main/ts/core/Utils.ts
@@ -209,6 +209,11 @@ const unlinkDomMutation = (editor: Editor) => {
   });
 };
 
+/**
+ * RTC uses unwrapped options.
+ *
+ * To best simulate this, we unwrap to null and filter out empty values.
+ */
 const unwrapOptions = (data: LinkDialogOutput) => {
   const { class: cls, href, rel, target, text, title } = data;
 
--- a/modules/tinymce/src/plugins/link/main/ts/ui/Controls.ts
+++ b/modules/tinymce/src/plugins/link/main/ts/ui/Controls.ts
@@ -79,7 +79,7 @@ const setupContextToolbars = (editor: Editor) => {
     return Fun.noop;
   };
 
-  /**
+  /*
    * if we're editing a link, don't change the text.
    * if anything other than text is selected, don't change the text.
    */
--- a/modules/tinymce/src/plugins/link/main/ts/ui/Controls.ts
+++ b/modules/tinymce/src/plugins/link/main/ts/ui/Controls.ts
@@ -105,28 +105,25 @@ const setupContextToolbars = (editor: Editor) => {
           return Actions.toggleActiveState(editor)(buttonApi);
         },
         onAction: (formApi) => {
-          const anchor = Utils.getAnchorElement(editor);
           const value = formApi.getValue();
-          if (!anchor) {
-            const attachState = { href: value, attach: Fun.noop };
-            const onlyText = Utils.isOnlyTextSelected(editor);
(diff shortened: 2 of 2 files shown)


modules/tinymce/src/plugins/link/main/ts/ui/Controls.ts
```diff
@@ -105,28 +105,25 @@ const setupContextToolbars = (editor: Editor) => {
...
-              collapseSelectionToEnd(editor);
-              formApi.hide();
-            });
-          }
+
+          // if we're editing a link, don't change the text.
+          // if anything other than text is selected, don't change the text.
+          const anchor = Utils.getAnchorElement(editor);
+          const onlyText = Utils.isOnlyTextSelected(editor);
+          const text: Optional<string> = !anchor && onlyText ? Optional.some(Utils.getAnchorText(editor.selection, anchor)).filter((t) => t.length > 0).or(Optional.from(value)) : Optional.none();
```
ashfordneil, line 114: Can we break this up a little bit? It's a very complicated ternary expression

ashfordneil (approved):
- [ ] Remember to give the PR a more meaningful name, as that's going to be the commit message when it's merged

modules/tinymce/src/plugins/link/main/ts/ui/Controls.ts
lnewson: This is probably a good use case for `Optionals.someIf`. eg:
```suggestion
          const text = Optionals.someIf(!anchor && onlyText, Utils.getAnchorText(editor.selection, anchor))
            .filter((t) => t.length > 0)
            .or(Optional.from(value));
```

TheSpyder: Will do. I tend to avoid changing more than I need to (this code is almost entirely the same as it was before) but I'm happy to make further changes.

TheSpyder: > Remember to give the PR a more meaningful title, as that's going to be the commit message when it's merged

I always forget this repo changed to squash merge.

TheSpyder:
> Remember to give the PR a more meaningful title, as that's going to be the commit message when it's merged

I always forget this repo changed to squash merge.

modules/tinymce/src/plugins/link/main/ts/ui/Controls.ts
TheSpyder: It's not quite as simple as `someIf` - the `.or()` only runs inside the `if` condition of the ternary. But I think I have a decent refactoring for clarity.

lnewson: Actually the above wouldn't be the same because of the `or` at the end :cry:

TheSpyder:
The tests are failing consistently but for a reason that seems unrelated to my change and I can't reproduce it locally.

modules/tinymce/src/plugins/link/main/ts/ui/Controls.ts
```diff
@@ -79,6 +79,21 @@ const setupContextToolbars = (editor: Editor) => {
     return Fun.noop;
   };
 
+  /**
```
lnewson, line 82:
```suggestion
  /*
```
This just seems to be a regular comment and not a API type doc string so we should only have a single asterisk.

modules/tinymce/src/plugins/link/main/ts/core/Utils.ts
```diff
@@ -209,6 +209,11 @@ const unlinkDomMutation = (editor: Editor) => {
   });
 };
 
+/**
```
lnewson, line 212:
```suggestion
/*
```

lnewson (approved):

modules/tinymce/src/plugins/link/main/ts/ui/Controls.ts
lnewson: This could also just be converted to using single line comments, which is what is more commonly used for this in the current code base 🤷

HAFRMO (approved):

modules/tinymce/src/plugins/link/main/ts/ui/Controls.ts
TheSpyder: Just habit, I guess. Even if a method doc comment doesn't include details of function arguments and return type it will still show up as autocomplete documentation.

PR #7032, opened 2021-07-21
TINY-6533: Turn off parts of image upload that RTC handles itself

TheSpyder:
Related Ticket: TINY-6533

Description of Changes:
* No user-visible changes, so no changelog, just turning off code that mutates the DOM when RTC is active.
* RTC already handles image upload updates; we just had special code in our mutation blocker to allow these modifications to the view. They had no effect on the model.
* Turning off the unnecessary modifications allows us to remove those checks from our mutation blocker and prevent a class of mutations that could be performed without an error.

From what I can tell the image scanner should _never_ need to convert base64 to blob anymore - Spocke updated the parser to do it - but I'm not prepared to verify that right now.

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/api/EditorUpload.ts
+++ b/modules/tinymce/src/core/main/ts/api/EditorUpload.ts
@@ -155,7 +155,11 @@ const EditorUpload = (editor: Editor): EditorUpload => {
 
           if (uploadInfo.status && Settings.shouldReplaceBlobUris(editor)) {
             blobCache.removeByUri(image.src);
-            replaceImageUriInView(image, uploadInfo.url);
+            if (Rtc.isRtc(editor)) {
+              // RTC handles replacing the image URL through callback events
+            } else {
+              replaceImageUriInView(image, uploadInfo.url);
+            }
           } else if (uploadInfo.error) {
             if (uploadInfo.error.options.remove) {
               replaceUrlInUndoStack(image.getAttribute('src'), Env.transparentSrc);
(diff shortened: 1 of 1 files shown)


ashfordneil (approved):
Just want to make sure that there's testing on the RTC side of things to make sure this works?

modules/tinymce/src/core/main/ts/api/EditorUpload.ts
```diff
@@ -239,11 +243,14 @@ const EditorUpload = (editor: Editor): EditorUpload => {
         return true;
       });
 
-      Arr.each(result, (resultItem) => {
-        replaceUrlInUndoStack(resultItem.image.src, resultItem.blobInfo.blobUri());
-        resultItem.image.src = resultItem.blobInfo.blobUri();
-        resultItem.image.removeAttribute('data-mce-src');
-      });
+      if (!Rtc.isRtc(editor)) {
```
JamesToohey, line 246: Is there any reason this one does a not condition and the others have the if/else? FWIW I'd prefer it this way

TheSpyder: because I did them at different times 🤷‍♂️ will fix

TheSpyder:
> Just want to make sure that there's testing on the RTC side of things to make sure this works?

yes, I'll link it in slack since I know we don't like private links in these discussions

lnewson (approved):

HAFRMO (approved):

PR #7033, opened 2021-07-22
TINY-7691: Table cell class changing now as a group, not individually

HAFRMO:
Related Ticket:

Description of Changes:
Classes are now toggled as a group. Buttons also now consider the whole selection, not just the first.
No changelog since it's a bug-fix for unreleased feature.

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/plugins/table/main/ts/api/Commands.ts
+++ b/modules/tinymce/src/plugins/table/main/ts/api/Commands.ts
@@ -84,11 +84,14 @@ const registerCommands = (editor: Editor, actions: TableActions, cellSelection: 
   };
 
   const toggleTableCellClass = (_ui: boolean, clazz: string) => {
-    performActionOnSelection((table, startCell) => {
-      const cells = TableSelection.getCellsFromSelection(startCell, selections, isRoot);
-      Arr.each(cells, (value) => {
-        editor.formatter.toggle('tablecellclass', { value: clazz }, value.dom);
-      });
+    performActionOnSelection((table) => {
+      const selectedCells = TableSelection.getCellsFromSelection(Util.getSelectionStart(editor), selections, Util.getIsRoot(editor));
+      const allHaveClass = Arr.forall(selectedCells, (cell) => editor.formatter.match('tablecellclass', { value: clazz }, cell.dom));
+      const formatterAction = allHaveClass ? editor.formatter.remove : editor.formatter.apply;
+
+      Arr.each(selectedCells, (value) =>
+        formatterAction('tablecellclass', { value: clazz }, value.dom)
+      );
       Events.fireTableModified(editor, table.dom, Events.styleModified);
     });
   };
--- a/modules/tinymce/src/plugins/table/main/ts/api/Commands.ts
+++ b/modules/tinymce/src/plugins/table/main/ts/api/Commands.ts
@@ -89,8 +89,8 @@ const registerCommands = (editor: Editor, actions: TableActions, cellSelection: 
       const allHaveClass = Arr.forall(selectedCells, (cell) => editor.formatter.match('tablecellclass', { value: clazz }, cell.dom));
       const formatterAction = allHaveClass ? editor.formatter.remove : editor.formatter.apply;
 
-      Arr.each(selectedCells, (value) =>
-        formatterAction('tablecellclass', { value: clazz }, value.dom)
+      Arr.each(selectedCells, (cell) =>
+        formatterAction('tablecellclass', { value: clazz }, cell.dom)
       );
       Events.fireTableModified(editor, table.dom, Events.styleModified);
     });
--- a/modules/tinymce/src/plugins/table/main/ts/core/TableFormats.ts
+++ b/modules/tinymce/src/plugins/table/main/ts/core/TableFormats.ts
@@ -11,39 +11,46 @@ const cellFormats = {
   tablecellbackgroundcolor: {
     selector: 'td,th',
     styles: { backgroundColor: '%value' },
-    remove_similar: true
+    remove_similar: true,
+    inherit: false
   },
   tablecellverticalalign: {
     selector: 'td,th',
     styles: {
       'vertical-align': '%value'
     },
-    remove_similar: true
+    remove_similar: true,
+    inherit: false
   },
   tablecellbordercolor: {
     selector: 'td,th',
     styles: { borderColor: '%value' },
-    remove_similar: true
+    remove_similar: true,
+    inherit: false
   },
   tablecellclass: {
     selector: 'td,th',
     classes: [ '%value' ],
-    remove_similar: true
+    remove_similar: true,
+    inherit: false
   },
   tableclass: {
     selector: 'table',
     classes: [ '%value' ],
-    remove_similar: true
+    remove_similar: true,
+    inherit: false
   },
   tablecellborderstyle: {
     selector: 'td,th',
     styles: { borderStyle: '%value' },
-    remove_similar: true
+    remove_similar: true,
+    inherit: false
   },
   tablecellborderwidth: {
     selector: 'td,th',
     styles: { borderWidth: '%value' },
-    remove_similar: true
+    remove_similar: true,
+    inherit: false
   }
 };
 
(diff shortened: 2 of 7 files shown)


modules/tinymce/src/plugins/table/main/ts/api/Commands.ts
```diff
@@ -84,11 +84,14 @@ const registerCommands = (editor: Editor, actions: TableActions, cellSelection:
...
-      const cells = TableSelection.getCellsFromSelection(startCell, selections, isRoot);
-      Arr.each(cells, (value) => {
-        editor.formatter.toggle('tablecellclass', { value: clazz }, value.dom);
-      });
+    performActionOnSelection((table) => {
+      const selectedCells = TableSelection.getCellsFromSelection(Util.getSelectionStart(editor), selections, Util.getIsRoot(editor));
+      const allHaveClass = Arr.forall(selectedCells, (cell) => editor.formatter.match('tablecellclass', { value: clazz }, cell.dom));
+      const formatterAction = allHaveClass ? editor.formatter.remove : editor.formatter.apply;
+
+      Arr.each(selectedCells, (value) =>
```
diogo405, line 92: I'd change `value` to `cell`.
```
Arr.each(selectedCells, (cell) =>
```

diogo405 (approved):

modules/tinymce/src/plugins/table/main/ts/ui/UiUtils.ts
```diff
@@ -24,7 +28,10 @@ const onSetupToggle = (editor: Editor, formatName: string, formatValue: string)
       const setActive = (matched: boolean) =>
         api.setActive(isNone ? !matched : matched);
 
-      setActive(editor.formatter.match(formatName, { value: formatValue }, undefined, isNone));
+      const selectedCells = TableSelection.getCellsFromSelection(Util.getSelectionStart(editor), selections, Util.getIsRoot(editor));
+      const forAll = () => Arr.forall(selectedCells, (cell) => editor.formatter.match(formatName, { value: formatValue }, cell.dom));
+      const forAny = () => Arr.exists(selectedCells, (cell) => editor.formatter.match(formatName, { value: formatValue }, cell.dom, isNone));
+      setActive(isNone ? forAny() : forAll());
```
ashfordneil, line 34: Maybe breaking it up into an if statement could simplify things?
```suggestion
      if (isNone) {
        setActive(Arr.exists(selectedCells, (cell) => editor.formatter.match(formatName, { value: formatValue }, cell.dom, true)));
      } else {
        setActive(Arr.forall(selectedCells, (cell) => editor.formatter.match(formatName, { value: formatValue }, cell.dom)));
      }
```

ashfordneil (approved):

modules/tinymce/src/plugins/table/main/ts/ui/UiUtils.ts
lnewson: This also should be using `matchNode` not `match` as we only want to look at this specific node and not it's parent nodes.

modules/tinymce/src/plugins/table/main/ts/core/TableFormats.ts
```diff
@@ -11,39 +11,46 @@ const cellFormats = {
   tablecellbackgroundcolor: {
     selector: 'td,th',
     styles: { backgroundColor: '%value' },
-    remove_similar: true
+    remove_similar: true,
```
lnewson, line 14: There is a lot of duplication here now, can we remove some by making a function or "base" object to set the common items (e.g. selector, remove_similar and inherit)?

modules/tinymce/src/plugins/table/main/ts/ui/UiUtils.ts
```diff
@@ -24,7 +29,16 @@ const onSetupToggle = (editor: Editor, formatName: string, formatValue: string)
...
         api.setActive(isNone ? !matched : matched);
 
-      setActive(editor.formatter.match(formatName, { value: formatValue }, undefined, isNone));
+      const selectedCells = TableSelection.getCellsFromSelection(Util.getSelectionStart(editor), selections, Util.getIsRoot(editor));
+
+      const checkNode = (cell: SugarElement<Element>) =>
+        editor.formatter.match(formatName, { value: formatValue }, cell.dom, isNone);
+
+      if (isNone) {
+        setActive(Arr.exists(selectedCells, checkNode));
```
lnewson, line 38: This logic is confusing as it's doing a double negation that's not immediately obvious so can we please rework it and add some comments? e.g.:
```
// For the None-entry we need to check that no format matches, whereas for others we need every cell to match the format
const predicate = isNone ? Arr.exists : Arr.forall;
const matches = predicate(selectedCells, (cell) => editor.formatter.match(formatName, { value: formatValue }, cell.dom, isNone));
setActive(matches);
```

Another option is to move the `setActive` function closer since it has a comment already.

modules/tinymce/src/plugins/table/test/ts/browser/command/ModifyClassesCommandsTest.ts
```diff
@@ -60,22 +65,92 @@ describe('browser.tinymce.plugins.table.command.ModifyClassesCommandsTest', () =
...
+                '<td data-mce-first-selected="1" data-mce-selected="1" class="a">1</td>' +
+                '<td data-mce-last-selected="1" data-mce-selected="1" class="a">2</td>' +
+              '</tr>' +
+            '</tbody>' +
+          '</table>'
+        );
+
+        it('TINY-7691: Can be toggled on', () => {
+          const editor = hook.editor();
+          editor.setContent(contentWithoutClassDoubleSelection);
```
lnewson, line 133: Can we please use your `setContentAndSelection` function here, as this should be setting a selection otherwise it's not a realistic test. Remember that normally we have the attributes and the selection sitting inside one of the cells.

- [ ] Ensure the native selection is set as well for these tests

modules/tinymce/src/plugins/table/test/ts/browser/command/ModifyClassesCommandsTest.ts
```diff
@@ -26,6 +26,11 @@ describe('browser.tinymce.plugins.table.command.ModifyClassesCommandsTest', () =
     events.push(event);
   };
 
+  const setContentAndSelection = (editor: Editor, content: string, selectionStart: number, selectionEnd: number) => {
+    editor.setContent(content);
+    TinySelections.setSelection(editor, [ 0, 0, 0 ], selectionStart, [ 0, 0, 0 ], selectionEnd);
```
lnewson, line 31:
```suggestion
  const setContentAndSelection = (editor: Editor, content: string, startCellOffset: number, endCellOffset: number) => {
    editor.setContent(content);
    TinySelections.setSelection(editor, [ 0, 0, 0 ], startCellOffset, [ 0, 0, 0 ], endCellOffset);
```
This is more accurate to what's happening here, as atm it's not obvious it must be a cell offset and not something else.

lnewson (approved):

modules/tinymce/src/plugins/table/main/ts/ui/UiUtils.ts
HAFRMO: Moved setActive closer.

lnewson: That only solves the comment part, not the confusing double negation I mentioned. Additionally, the if branches should always call the same thing, so having them operate separately will likely lead to maintenance issues in the future.

HAFRMO: Made some changes, not sure the comment applies as much anymore.

lnewson: This goes the opposite direction I wanted... it's now removed the `setActive` function which was shared code with the commented out `formatChanged` logic. It also doesn't solve the concern of having 2 branches that are largely trying to do the same thing.

lnewson: On second thoughts this is probably nitpicking a little too much. My point is still accurate, but it's not likely to happen when both are side-by-side. There is still the issue of the logic being inlined that was shared code, but we can cross that bridge when we get to fixing the other issue.

lnewson (approved):

PR #7039, opened 2021-07-22
TINY-7688: Android context menus are not showing consistently

diogo405:
Description of Changes:
* Mobile context menus not using NodeAnchor anymore due to problem anchoring to the body when there's more than one element selected.
* Removed selection workaround (`editor.selection.setCursorLocation(e.target, 0)`) as it isn't needed for Android 10 & 11. On Android 11 it wasn't creating the selection and was moving the cursor to the beginning of the node.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/tinymce/src/themes/silver/main/ts/ui/menus/contextmenu/platform/MobileContextMenu.ts
```diff
@@ -152,12 +121,6 @@ export const initAndShow = (editor: Editor, e: EditorEvent<TouchEvent>, buildMen
       editor.once('touchend', () => editor.off('selectionchange', openiOS));
     }
   } else {
-    // On Android editor.selection hasn't updated yet at this point, so need to do it manually
-    // Without this longpress causes drag-n-drop duplication of code on Android
-    if (isAndroid && !useNodeAnchor) {
-      editor.selection.setCursorLocation(e.target, 0);
-    }
-
```
diogo405, line 160: Removed this part cause it was removing the selection and moving the cursor to the beginning of the node when longpressing on Android.

modules/tinymce/src/themes/silver/main/ts/ui/menus/contextmenu/platform/MobileContextMenu.ts
```diff
@@ -21,29 +21,9 @@ import ItemResponse from '../../item/ItemResponse';
...
-  valignCentre: [],
-  alignCentre: [],
-  alignLeft: [ 'tox-pop--align-left' ],
-  alignRight: [ 'tox-pop--align-right' ],
-  right: [ 'tox-pop--right' ],
-  left: [ 'tox-pop--left' ],
-  bottom: [ 'tox-pop--bottom' ],
-  top: [ 'tox-pop--top' ]
-};
-
```
diogo405, line 46: Moved to SilverContextMenu.

modules/tinymce/src/themes/silver/main/ts/ui/menus/contextmenu/platform/MobileContextMenu.ts
```diff
@@ -60,16 +40,6 @@ const isTouchWithinSelection = (editor: Editor, e: EditorEvent<TouchEvent>) => {
...
-const getPointAnchorSpec = (editor: Editor, e: EditorEvent<TouchEvent>) => ({
-  bubble: Bubble.nu(0, bubbleSize, bubbleAlignments),
-  layouts,
-  overrides: {
-    maxWidthFunction: MaxWidth.expandable(),
-    maxHeightFunction: MaxHeight.expandable()
-  },
-  ...getPointAnchor(editor, e)
-});
-
```
diogo405, line 72: Moved to SilverContextMenu.

diogo405 (commented):
Just one thing, when using the SelectionAnchor on mobile it covers the text if the selection has more than one line. See below. 

Screenshot_20210722-154902

modules/tinymce/src/themes/silver/main/ts/ui/menus/contextmenu/SilverContextMenu.ts
```diff
@@ -119,22 +142,37 @@ export const isTriggeredByKeyboard = (editor: Editor, e: PointerEvent) =>
...
+  const pointAnchor = getPointAnchor(editor, e);
+  if (isTouch()) {
+    return {
+      bubble: Bubble.nu(0, bubbleSize, bubbleAlignments),
+      layouts,
+      overrides: {
+        maxWidthFunction: MaxWidth.expandable(),
+        maxHeightFunction: MaxHeight.expandable()
+      },
+      ...pointAnchor
```
ashfordneil, line 155: Just confirming: which object do we want to take precedence if `pointAnchor` has bubble/layouts/overrides fields? I would have thought (without much context, so I could be wrong here) that we'd want to put `...pointAnchor` at the start of this object, instead of the end.

ashfordneil (approved):

modules/tinymce/src/themes/silver/main/ts/ui/menus/contextmenu/SilverContextMenu.ts
lnewson: Yeah, this all is incorrect. Mobile and Desktop have different requirements, so the bubble, layouts and anchor definitions need to remain in those files, not here.

lnewson (changes requested):
Requesting changes as the bubble, layout and anchor should not be in `SilverContextMenu`. They change based on the type of menu being shown, so they should at all times remain in the `MobileContextMenu` or `DesktopContextMenu` files instead. What we should be doing, as I had noted in standup, is passing an enum/string about which type of anchor should be used. Sorry if that wasn't clear :cry:

lnewson:
> Just one thing, when using the SelectionAnchor on mobile it covers the text if the selection has more than one line. See below.

As discussed on slack, remember that this is a bug. So please also make sure a Jira is logged and linked to this so we know to fix it before 5.9 release.

- [ ] Ensure a follow up Jira is included in 5.9 to fix selection anchors overlapping like this

modules/tinymce/src/themes/silver/main/ts/ui/menus/contextmenu/platform/MobileContextMenu.ts
lnewson: This can't be removed as it was needed on older versions of Android. We'll need a way/some heuristic to determine when it's needed.

modules/tinymce/src/themes/silver/main/ts/ui/menus/contextmenu/SilverContextMenu.ts
```diff
@@ -119,22 +142,37 @@ export const isTriggeredByKeyboard = (editor: Editor, e: PointerEvent) =>
...
+    return pointAnchor;
+  }
+};
+
+const getAnchorSpec = (editor: Editor, e): AnchorSpec => {
   const selector = Settings.getAvoidOverlapSelector(editor);
-  if (isTriggeredByKeyboard(editor, e)) {
-    return true;
-  } else if (selector) {
+  if (Type.isString(selector) && selector !== '') {
```
lnewson, line 164: This is a behavioural change. Are we sure we want that? If so it'll need to be included in the changelog.

lnewson: We also should be using `Strings.isNotEmpty(selector)` here instead of `selector !== ''`.

modules/tinymce/src/themes/silver/main/ts/ui/menus/contextmenu/SilverContextMenu.ts
diogo405: Hmm I moved and added an `if (isTouch())` so the code doesn't change - I mean from what was doing previously. But I agree that all those mobile config (layouts, bubble, ...) should be in the MobileContextMenu. I'll move it back.

diogo405:
> Requesting changes as the bubble, layout and anchor should not be in `SilverContextMenu`. They change based on the type of menu being shown, so they should at all times remain in the `MobileContextMenu` or `DesktopContextMenu` files instead. What we should be doing, as I had noted in standup, is passing an enum/string about which type of anchor should be used. Sorry if that wasn't clear 😢

You mean to pass an enum/string in the `initAndShow`  method rather than a `useNodeAnchor: boolean`? And inside `MobileContextMenu` and `DesktopContextMenu` create the AnchorSpec based on the enum/string? If so, it's not better to do that in `SilverContextMenu` and pass an `AnchorSpec` to `initAndShow`?

modules/tinymce/src/themes/silver/main/ts/ui/menus/contextmenu/SilverContextMenu.ts
diogo405: Did that to avoid the truthy condition.

diogo405 (commented):
Basically, reverted everything and only changed `shouldUseNodeAnchor` in `SilverContextMenu` and the workaround condition to `isAndroid10 && !useNodeAnchor` in `MobileContextMenu`.

modules/tinymce/src/themes/silver/main/ts/ui/menus/contextmenu/platform/MobileContextMenu.ts
```diff
@@ -152,11 +153,14 @@ export const initAndShow = (editor: Editor, e: EditorEvent<TouchEvent>, buildMen
       editor.once('touchend', () => editor.off('selectionchange', openiOS));
     }
   } else {
-    // On Android editor.selection hasn't updated yet at this point, so need to do it manually
-    // Without this longpress causes drag-n-drop duplication of code on Android
-    if (isAndroid && !useNodeAnchor) {
-      editor.selection.setCursorLocation(e.target, 0);
-    }
+    // TINY-7688: Commented out cause when longpressing doesn't select and
```
ashfordneil, line 156: Can we avoid merging in commented out code? Either
* Remove it
* Fix it up and uncomment before merging
* Remove it and make a follow-up Jira to fix it, and leave the code in a comment on that Jira as a hint to whoever gets to implement it

modules/tinymce/src/themes/silver/main/ts/ui/menus/contextmenu/platform/MobileContextMenu.ts
```diff
@@ -103,6 +103,7 @@ const setupiOSOverrides = (editor: Editor) => {
 
 const show = (editor: Editor, e: EditorEvent<TouchEvent>, items: MenuItems, backstage: UiFactoryBackstage, contextmenu: AlloyComponent, useNodeAnchor: boolean, highlightImmediately: boolean) => {
   const anchorSpec = useNodeAnchor ? getNodeAnchor(editor) : getPointAnchorSpec(editor, e);
+  console.log(anchorSpec.anchor);
```
ashfordneil, line 106: Make sure this gets removed before we merge

diogo405: 😅  Removed

modules/tinymce/src/themes/silver/main/ts/ui/menus/contextmenu/platform/MobileContextMenu.ts
diogo405: Did that cause I saw it somewhere hehe. I've removed it.

modules/tinymce/src/themes/silver/main/ts/ui/menus/contextmenu/SilverContextMenu.ts
```diff
@@ -122,19 +123,16 @@ const getSelectedElement = (editor: Editor, e: PointerEvent) =>
 const shouldUseNodeAnchor = (editor: Editor, e: PointerEvent) => {
   const selector = Settings.getAvoidOverlapSelector(editor);
   if (isTriggeredByKeyboard(editor, e)) {
-    return true;
+    return false;
```
lnewson, line 126: This can't be done and doesn't do what we had said. With this change it'll use a point anchor instead which is incorrect. So as we've discussed previously, it should be using a selection anchor instead for keyboard activation. It should **always** be using a node anchor for the case below. Lastly it should use the point anchor for clicks/touches.

lnewson (changes requested):
This goes against everything that's been discussed and definitely causes the wrong anchors to be used. See my other comment for more details.

modules/tinymce/src/themes/silver/main/ts/ui/menus/contextmenu/SilverContextMenu.ts
lnewson: I also still stand by what I had told you originally, that there's possibly a bug in the `isTriggeredByKeyboard` logic if it's returning true for Android devices.

diogo405: I've made the change passing the anchor type rather than `useAnchorNode: boolean`. Haven't figured out how to fix Android returning true for `isTriggeredByKeyboard`. Did some tests and it seems when you `longpress` it triggers `longpress` twice and it goes to `getPointAnchorSpec`. Then fires `contextmenu` event twice with `e.button === -1` and returns a SelectionAnchor. See below.

longpress-android

modules/tinymce/src/themes/silver/main/ts/ui/menus/contextmenu/platform/MobileContextMenu.ts
```diff
@@ -101,28 +102,42 @@ const setupiOSOverrides = (editor: Editor) => {
   };
 };
 
-const show = (editor: Editor, e: EditorEvent<TouchEvent>, items: MenuItems, backstage: UiFactoryBackstage, contextmenu: AlloyComponent, useNodeAnchor: boolean, highlightImmediately: boolean) => {
-  const anchorSpec = useNodeAnchor ? getNodeAnchor(editor) : getPointAnchorSpec(editor, e);
+const getAnchorSpec = (editor: Editor, e: EditorEvent<TouchEvent>, anchorType: ContextMenuAnchorType) => {
```
ashfordneil, line 105: Seeing as this is duplicated, I'm wondering if it might be worth moving this (as well as the `ContextMenuAnchorType` definition) into a single shared file? Possibly even `../Coords.ts`, seeing as that's where the individual get functions are

lnewson: Agreed, I hadn't got around to commenting yet but this was going to be one of the changes I was going to request.

modules/tinymce/src/themes/silver/main/ts/ui/menus/contextmenu/SilverContextMenu.ts
```diff
@@ -119,15 +120,16 @@ export const isTriggeredByKeyboard = (editor: Editor, e: PointerEvent) =>
...
+const getAnchorType = (editor: Editor, e: PointerEvent): ContextMenuAnchorType => {
   const selector = Settings.getAvoidOverlapSelector(editor);
-  if (isTriggeredByKeyboard(editor, e)) {
-    return true;
-  } else if (selector) {
+  if (selector) {
     const target = getSelectedElement(editor, e);
-    return SelectorExists.closest(SugarElement.fromDom(target), selector);
+    const selectorExists = SelectorExists.closest(SugarElement.fromDom(target), selector);
+    return selectorExists ? 'node' : 'point';
```
lnewson, line 128: This fallback is incorrect. It needs to continue on to the logic below instead when the selector doesn't exist.

modules/tinymce/src/themes/silver/main/ts/ui/menus/contextmenu/platform/MobileContextMenu.ts
lnewson: Putting it in `Coords` would probably be a reasonable place given that's where the current anchors are generated.

diogo405: Moved to Coords.

modules/tinymce/src/themes/silver/main/ts/ui/menus/contextmenu/SilverContextMenu.ts
diogo405: Changed.

modules/tinymce/src/themes/silver/main/ts/ui/menus/contextmenu/platform/MobileContextMenu.ts
```diff
@@ -101,28 +91,47 @@ const setupiOSOverrides = (editor: Editor) => {
...
+      bubble: Bubble.nu(0, bubbleSize, bubbleAlignments),
+      layouts,
+      overrides: {
+        maxWidthFunction: MaxWidth.expandable(),
+        maxHeightFunction: MaxHeight.expandable()
+      },
+      ...Coords.getPointAnchor(editor, e)
+    };
+  } else {
+    return Coords.getAnchorSpec(editor, e, anchorType);
```
lnewson, line 106: Again, there's duplication here which we don't want... please use the example I gave in slack if you're taking this approach.

- [ ] Fix this to use the example given on slack whereby the `getAnchorSpec` call is not duplicated

modules/tinymce/src/themes/silver/main/ts/ui/menus/contextmenu/platform/MobileContextMenu.ts
```diff
@@ -101,28 +91,47 @@ const setupiOSOverrides = (editor: Editor) => {
...
+const show = (editor: Editor, e: EditorEvent<TouchEvent>, items: MenuItems, backstage: UiFactoryBackstage, contextmenu: AlloyComponent, anchorType: Coords.ContextMenuAnchorType, highlightImmediately: boolean) => {
+
+  const anchorSpec = getAnchorSpec(editor, e, anchorType);
 
   NestedMenus.build(items, ItemResponse.CLOSE_ON_EXECUTE, backstage, true).map((menuData) => {
     e.preventDefault();
 
     // Show the context menu, with items set to close on click
+    const toolbarType = anchorType === 'node' ? 'node' : 'selection';
+    const contextToolbarBounds = getContextToolbarBounds(editor, backstage.shared, toolbarType);
```
lnewson, line 119: This change is incorrect... the bounds should be lazily computed (you'll note previously it was inside a thunk).

- [ ] Revert this change

lnewson (changes requested):

modules/tinymce/src/themes/silver/main/ts/ui/menus/contextmenu/platform/MobileContextMenu.ts
lnewson: To add more as to why this is necessary, if the UI is repositioned (possibly due to a scroll or window resize) then it needs to check the bounds again which may have changed. As such the bounds can't be eagerly evaluated like this.

modules/tinymce/CHANGELOG.md
```diff
@@ -59,6 +59,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Menus and context menus were not closed when clicking into a different editor #TINY-7399
 - Using the Tab key to navigate into the editor on IE 11 would incorrectly focus the toolbar #TINY-3707
 - The editor selection could be placed in an incorrect location when undoing or redoing changes in a document containing `contenteditable="false"` elements #TINY-7663
+- Context menus on Android were not being displayed when more than one HTML element was selected #TINY-7688
```
lnewson, line 62: `being` is a banned word so we can't use it sorry. Maybe something like this?
```suggestion
- Context menus on Android were not displayed when more than one HTML element was selected #TINY-7688
```

modules/tinymce/src/themes/silver/main/ts/ui/menus/contextmenu/Coords.ts
```diff
@@ -5,13 +5,15 @@
...
 
-import { MakeshiftAnchorSpec, NodeAnchorSpec, SelectionAnchorSpec } from '@ephox/alloy';
+import { AnchorSpec, MakeshiftAnchorSpec, NodeAnchorSpec, SelectionAnchorSpec } from '@ephox/alloy';
 import { Optional } from '@ephox/katamari';
 import { SugarElement } from '@ephox/sugar';
 
 import DOMUtils from 'tinymce/core/api/dom/DOMUtils';
 import Editor from 'tinymce/core/api/Editor';
 
+export type ContextMenuAnchorType = 'node' | 'selection' | 'point';
```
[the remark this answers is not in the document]
diogo405: Yes!

modules/tinymce/src/themes/silver/main/ts/ui/menus/contextmenu/platform/MobileContextMenu.ts
diogo405: I thought it would be better than calling the method when is not needed. But if it's how we do it, I'm gonna change 👍

diogo405: Actually they are different: `getPointAnchor` and `getAnchorSpec`.

lnewson: No, they are the same as the intention is to lookup the anchor based on it's type and the augment it. So this ends up duplicating the lookup logic which we don't want.

lnewson: I should add this is still calling the same logic (so it's not calling anything that's unneeded), it's just you're manually short circuiting it, which doesn't convey at the core it should be the same just with some additional defaults.

modules/tinymce/src/themes/silver/main/ts/ui/menus/contextmenu/SilverContextMenu.ts
```diff
@@ -119,15 +120,15 @@ export const isTriggeredByKeyboard = (editor: Editor, e: PointerEvent) =>
 const getSelectedElement = (editor: Editor, e: PointerEvent) =>
   isTriggeredByKeyboard(editor, e) ? editor.selection.getStart(true) : e.target as Element;
 
-const shouldUseNodeAnchor = (editor: Editor, e: PointerEvent) => {
+const getAnchorType = (editor: Editor, e: PointerEvent): AnchorType => {
   const selector = Settings.getAvoidOverlapSelector(editor);
-  if (isTriggeredByKeyboard(editor, e)) {
-    return true;
-  } else if (selector) {
+  const triggeredByKeyboard = isTriggeredByKeyboard(editor, e) ? 'selection' : 'point';
```
lnewson, line 125: This sounds like a boolean value, not that it holds the anchor type. Can we please rename this? Maybe `type` or `anchorType`?

modules/tinymce/src/themes/silver/main/ts/ui/menus/contextmenu/platform/MobileContextMenu.ts
diogo405: Calling `getPointAnchor` avoids the switch, that's the difference. Changed anyways.

modules/tinymce/src/themes/silver/main/ts/ui/menus/contextmenu/SilverContextMenu.ts
diogo405: Sure.

modules/tinymce/src/themes/silver/main/ts/ui/menus/contextmenu/platform/MobileContextMenu.ts
```diff
@@ -101,8 +91,26 @@ const setupiOSOverrides = (editor: Editor) => {
...
+      ...anchorSpec
+    };
+  } else {
+    return anchorSpec;
+  }
+};
+
+const show = (editor: Editor, e: EditorEvent<TouchEvent>, items: MenuItems, backstage: UiFactoryBackstage, contextmenu: AlloyComponent, anchorType: Coords.AnchorType, highlightImmediately: boolean) => {
+
+  const anchorSpec = getAnchorSpec(editor, e, anchorType);
```
lnewson, line 113:
```suggestion
const show = (editor: Editor, e: EditorEvent<TouchEvent>, items: MenuItems, backstage: UiFactoryBackstage, contextmenu: AlloyComponent, anchorType: Coords.AnchorType, highlightImmediately: boolean) => {
  const anchorSpec = getAnchorSpec(editor, e, anchorType);
```

modules/tinymce/src/themes/silver/main/ts/ui/menus/contextmenu/platform/MobileContextMenu.ts
```diff
@@ -101,8 +91,26 @@ const setupiOSOverrides = (editor: Editor) => {
...
+      overrides: {
+        maxWidthFunction: MaxWidth.expandable(),
+        maxHeightFunction: MaxHeight.expandable()
+      },
+      ...anchorSpec
+    };
+  } else {
+    return anchorSpec;
+  }
+};
```
lnewson, line 109:
```suggestion
const getAnchorSpec = (editor: Editor, e: EditorEvent<TouchEvent>, anchorType: Coords.AnchorType) => {
  const anchorSpec = Coords.getAnchorSpec(editor, e, anchorType);
  return {
    bubble: Bubble.nu(0, 0, bubbleAlignments),
    layouts,
    overrides: {
      maxWidthFunction: MaxWidth.expandable(),
      maxHeightFunction: MaxHeight.expandable()
    },
    ...anchorSpec
  };
};
```

As mentioned on slack, after looking at this more and playing around with the branch locally I believe this is what we want. At the very least the overrides should apply to `selection` and `point`.

Note: I've also removed the `bubbleSize` variable in the above as we don't render the "bubble arrow" on mobile so I don't think we want the context toolbar to be sitting 12px below the selection/touch point.

modules/tinymce/src/themes/silver/main/ts/ui/menus/contextmenu/platform/MobileContextMenu.ts
diogo405: Can't spot the difference 😅  line break?!

lnewson (approved):

modules/tinymce/src/themes/silver/main/ts/ui/menus/contextmenu/SilverContextMenu.ts
```diff
@@ -119,15 +120,15 @@ export const isTriggeredByKeyboard = (editor: Editor, e: PointerEvent) =>
...
   isTriggeredByKeyboard(editor, e) ? editor.selection.getStart(true) : e.target as Element;
 
-const shouldUseNodeAnchor = (editor: Editor, e: PointerEvent) => {
+const getAnchorType = (editor: Editor, e: PointerEvent): AnchorType => {
   const selector = Settings.getAvoidOverlapSelector(editor);
-  if (isTriggeredByKeyboard(editor, e)) {
-    return true;
-  } else if (selector) {
+  const anchorType = isTriggeredByKeyboard(editor, e) ? 'selection' : 'point';
+  if (selector) {
```
JamesToohey, line 126: I realise this was already here, but given we are modifying this code maybe we can use `Strings.isNotEmpty(selector)`?

JamesToohey (approved):

ashfordneil (approved):

ltrouton (approved):

modules/tinymce/src/themes/silver/main/ts/ui/menus/contextmenu/SilverContextMenu.ts
diogo405: I had done that but reverted cause @lnewson said it was going to be a behaviour change?! Can't remember actually...

lnewson: That was different as you had reordered the ifs previously and that was the behaviour change I mentioned.

diogo405: ah right, changed 👍

PR #7042, opened 2021-07-23
TINY-7737: Table buttons enable only when whole selection is within the same table

HAFRMO:
Description of Changes:
Now check that the start and end are both within the same table.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/plugins/table/main/ts/selection/SelectionTargets.ts
+++ b/modules/tinymce/src/plugins/table/main/ts/selection/SelectionTargets.ts
@@ -8,7 +8,7 @@
 import { Selections } from '@ephox/darwin';
 import { Arr, Cell, Fun, Optional, Thunk } from '@ephox/katamari';
 import { RunOperation, Structs, TableLookup, Warehouse } from '@ephox/snooker';
-import { SelectorExists, SugarElement, SugarNode } from '@ephox/sugar';
+import { Compare, SelectorExists, SugarElement, SugarNode } from '@ephox/sugar';
 
 import Editor from 'tinymce/core/api/Editor';
 import { Menu, Toolbar } from 'tinymce/core/api/ui/Ui';
@@ -62,18 +62,25 @@ export const getSelectionTargets = (editor: Editor, selections: Selections): Sel
 
   const isCaption = SugarNode.isTag('caption');
   const isDisabledForSelection = (key: keyof ExtractedSelectionDetails) => selectionDetails.forall((details) => !details[key]);
+  const getStart = () => TableSelection.getSelectionStartCellOrCaption(Util.getSelectionStart(editor), Util.getIsRoot(editor));
+  const getEnd = () => TableSelection.getSelectionEndCellOrCaption(Util.getSelectionEnd(editor), Util.getIsRoot(editor));
 
-  const findTargets = (): Optional<RunOperation.CombinedTargets> => TableSelection.getSelectionStartCellOrCaption(Util.getSelectionStart(editor), Util.getIsRoot(editor))
-    .bind((cellOrCaption) => {
-      const table = TableLookup.table(cellOrCaption);
-      return table.map((table) => {
-        if (isCaption(cellOrCaption)) {
-          return TableTargets.noMenu(cellOrCaption);
-        } else {
-          return TableTargets.forMenu(selections, table, cellOrCaption);
-        }
-      });
-    });
+  const findTargets = (): Optional<RunOperation.CombinedTargets> =>
+    getStart().bind((startCellOrCaption) =>
+      getEnd().bind(TableLookup.table).bind((endTable) =>
+        TableLookup.table(startCellOrCaption).bind((startTable) => {
+          if (Compare.eq(startTable, endTable)) {
+            if (isCaption(startCellOrCaption)) {
+              return Optional.some(TableTargets.noMenu(startCellOrCaption));
+            } else {
+              return Optional.some(TableTargets.forMenu(selections, startTable, startCellOrCaption));
+            }
+          }
+
+          return Optional.none();
+        })
+      )
+    );
 
   const getExtractedDetails = (targets: RunOperation.CombinedTargets): Optional<ExtractedSelectionDetails> => {
     const tableOpt = TableLookup.table(targets.element);
--- a/modules/tinymce/src/plugins/table/main/ts/selection/SelectionTargets.ts
+++ b/modules/tinymce/src/plugins/table/main/ts/selection/SelectionTargets.ts
@@ -6,7 +6,7 @@
  */
 
 import { Selections } from '@ephox/darwin';
-import { Arr, Cell, Fun, Optional, Thunk } from '@ephox/katamari';
+import { Arr, Cell, Fun, Optional, Optionals, Thunk } from '@ephox/katamari';
 import { RunOperation, Structs, TableLookup, Warehouse } from '@ephox/snooker';
 import { Compare, SelectorExists, SugarElement, SugarNode } from '@ephox/sugar';
 
@@ -67,8 +67,8 @@ export const getSelectionTargets = (editor: Editor, selections: Selections): Sel
 
   const findTargets = (): Optional<RunOperation.CombinedTargets> =>
     getStart().bind((startCellOrCaption) =>
-      getEnd().bind(TableLookup.table).bind((endTable) =>
-        TableLookup.table(startCellOrCaption).bind((startTable) => {
+      Optionals.flatten(
+        Optionals.lift2(TableLookup.table(startCellOrCaption), getEnd().bind(TableLookup.table), (startTable, endTable) => {
           if (Compare.eq(startTable, endTable)) {
             if (isCaption(startCellOrCaption)) {
               return Optional.some(TableTargets.noMenu(startCellOrCaption));
--- a/modules/tinymce/src/plugins/table/main/ts/selection/SelectionTargets.ts
+++ b/modules/tinymce/src/plugins/table/main/ts/selection/SelectionTargets.ts
@@ -62,8 +62,8 @@ export const getSelectionTargets = (editor: Editor, selections: Selections): Sel
 
   const isCaption = SugarNode.isTag('caption');
   const isDisabledForSelection = (key: keyof ExtractedSelectionDetails) => selectionDetails.forall((details) => !details[key]);
-  const getStart = () => TableSelection.getSelectionStartCellOrCaption(Util.getSelectionStart(editor), Util.getIsRoot(editor));
-  const getEnd = () => TableSelection.getSelectionEndCellOrCaption(Util.getSelectionEnd(editor), Util.getIsRoot(editor));
+  const getStart = () => TableSelection.getSelectionCellOrCaption(Util.getSelectionStart(editor), Util.getIsRoot(editor));
+  const getEnd = () => TableSelection.getSelectionCellOrCaption(Util.getSelectionEnd(editor), Util.getIsRoot(editor));
 
   const findTargets = (): Optional<RunOperation.CombinedTargets> =>
     getStart().bind((startCellOrCaption) =>
--- a/modules/tinymce/src/plugins/table/main/ts/selection/TableSelection.ts
+++ b/modules/tinymce/src/plugins/table/main/ts/selection/TableSelection.ts
@@ -12,22 +12,27 @@ import { Attribute, Compare, SelectorFind, SugarElement, SugarElements, SugarNod
 
 import { ephemera } from './Ephemera';
 
-const getSelectionStartCellFallback = (start: SugarElement<Node>) =>
-  TableLookup.table(start).bind((table) =>
+const getSelectionCellFallback = (element: SugarElement<Node>, takeLast: boolean) =>
+  TableLookup.table(element).bind((table) =>
     TableSelection.retrieve(table, ephemera.firstSelectedSelector)
-  ).fold(() => start, (cells) => cells[0]);
+  ).fold(
+    () => element,
+    (cells) => takeLast ? cells[cells.length - 1] : cells[0]);
 
-const getSelectionStartFromSelector = <T extends Element>(selector: string) => (start: SugarElement<Node>, isRoot?: (el: SugarElement<Node>) => boolean) => {
-  const startCellName = SugarNode.name(start);
-  const startCell = startCellName === 'col' || startCellName === 'colgroup' ? getSelectionStartCellFallback(start) : start;
-  return SelectorFind.closest<T>(startCell, selector, isRoot);
-};
+const getSelectionFromSelector = <T extends Element>(selector: string, takeLast: boolean) =>
+  (initCell: SugarElement<Node>, isRoot?: (el: SugarElement<Node>) => boolean) => {
+    const cellName = SugarNode.name(initCell);
+    const cell = cellName === 'col' || cellName === 'colgroup' ? getSelectionCellFallback(initCell, takeLast) : initCell;
+    return SelectorFind.closest<T>(cell, selector, isRoot);
+  };
 
-const getSelectionStartCaption = getSelectionStartFromSelector<HTMLTableCaptionElement>('caption');
+const getSelectionStartCaption = getSelectionFromSelector<HTMLTableCaptionElement>('caption', false);
 
-const getSelectionStartCell = getSelectionStartFromSelector<HTMLTableCellElement>('th,td');
+const getSelectionStartCell = getSelectionFromSelector<HTMLTableCellElement>('th,td', false);
 
-const getSelectionStartCellOrCaption = getSelectionStartFromSelector<HTMLTableCellElement | HTMLTableCaptionElement>('th,td,caption');
+const getSelectionStartCellOrCaption = getSelectionFromSelector<HTMLTableCellElement | HTMLTableCaptionElement>('th,td,caption', false);
+
+const getSelectionEndCellOrCaption = getSelectionFromSelector<HTMLTableCellElement | HTMLTableCaptionElement>('th,td,caption', false);
 
 const getCellsFromSelection = (start: SugarElement<Node>, selections: Selections, isRoot?: (el: SugarElement<Node>) => boolean): SugarElement<HTMLTableCellElement>[] =>
   getSelectionStartCell(start, isRoot)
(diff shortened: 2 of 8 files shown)


HAFRMO:
Tests to be added.

modules/tinymce/src/plugins/table/test/ts/browser/settings/TableToolbarTest.ts
```diff
@@ -50,4 +52,55 @@ describe('browser.tinymce.plugins.table.TableToolbarTest', () => {
...
+
+  context('Ensure table buttons are enabled when appropriate', () => {
+    it('TINY-7737: When the whole selection is in the table', async () => {
+      const editor = await pCreateEditor('tablecaption');
+      editor.focus();
+      editor.setContent(tableWithAdditionalHtml);
+      await TinyUiActions.pWaitForUi(editor, '.tox-tbtn--disabled[title="Table caption"]');
+      TinySelections.setSelection(editor, [ 1, 0, 0, 0 ], 0, [ 1, 0, 0, 0 ], 1);
+      // Wait for a while to allow the toolbar a chance to render
+      await Waiter.pWait(100);
```
lnewson, line 64: This wait should not be here and in the other tests, as the toolbar will always be rendered.

modules/tinymce/src/plugins/table/test/ts/browser/settings/TableToolbarTest.ts
```diff
@@ -50,4 +52,55 @@ describe('browser.tinymce.plugins.table.TableToolbarTest', () => {
     await TinyUiActions.pWaitForUi(editor, '.tox-pop .tox-tbtn[title="Paste row before"]:not(.tox-tbtn--disabled)');
     McEditor.remove(editor);
   });
+
+  context('Ensure table buttons are enabled when appropriate', () => {
```
lnewson, line 56: This isn't a great place for this... this test is checking the context toolbar rendering logic and the `table_toolbar` setting. Your new tests don't use any of that, so please move them to a separate file. You also don't need to be initializing an editor per test, as you aren't changing any settings.

modules/tinymce/src/plugins/table/test/ts/browser/settings/ToolbarButtonDisplayTest.ts
```diff
@@ -0,0 +1,50 @@
...
+describe('browser.tinymce.plugins.table.ToolbarButtonDisplayTest', () => {
+  const hook = TinyHooks.bddSetupLight<Editor>({
+    plugins: 'table',
+    toolbar: 'tablecaption',
+    base_url: '/project/tinymce/js/tinymce'
+  }, [ Plugin, Theme ], true);
+
+  const editorContent = '<p>A</p><table><tbody><tr><td>B</td></tr></tbody></table><p>C</p>';
+
+  const displayTest = async (startPath: number[], startOffset: number, endPath: number[], endOffset: number, shouldBeEnabled: boolean) => {
```
lnewson, line 19: This should start with `p` as it's returning a promise. e.g. `pDisplayTest`

modules/tinymce/src/plugins/table/test/ts/browser/settings/ToolbarButtonDisplayTest.ts
```diff
@@ -0,0 +1,50 @@
...
+
+    if (shouldBeEnabled) {
+      await TinyUiActions.pWaitForUi(editor, '.tox-tbtn[title="Table caption"]');
+      UiFinder.notExists(SugarBody.body(), '.tox-tbtn--disabled[title="Table caption"]');
+    } else {
+      await TinyUiActions.pWaitForUi(editor, '.tox-tbtn--disabled[title="Table caption"]');
+    }
+  };
+
+  context('Ensure table buttons are enabled when appropriate', () => {
```
lnewson, line 33:
```suggestion
  context('Ensure table caption buttons are enabled when appropriate', () => {
```
This is very specifically targeted at the `tablecaption` button.

modules/tinymce/src/plugins/table/test/ts/browser/settings/ToolbarButtonDisplayTest.ts
```diff
@@ -0,0 +1,50 @@
...
+    it('TINY-7737: When the whole selection is in the table', async () => {
+      await displayTest([ 1, 0, 0, 0 ], 0, [ 1, 0, 0, 0 ], 1, true);
+    });
```
lnewson, line 36:
```
    it('TINY-7737: When the whole selection is in the table', () => displayTest([ 1, 0, 0, 0 ], 0, [ 1, 0, 0, 0 ], 1, true));
```
```
    it('TINY-7737: When the whole selection is in the table', () =>
      displayTest([ 1, 0, 0, 0 ], 0, [ 1, 0, 0, 0 ], 1, true)
    );
```
This and others could be shortened to something like one of these 2.

modules/tinymce/src/plugins/table/test/ts/browser/settings/ToolbarButtonDisplayTest.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  const displayTest = async (startPath: number[], startOffset: number, endPath: number[], endOffset: number, shouldBeEnabled: boolean) => {
+    const editor = hook.editor();
+    editor.setContent(editorContent);
+    await TinyUiActions.pWaitForUi(editor, '.tox-tbtn--disabled[title="Table caption"]');
+    TinySelections.setSelection(editor, startPath, startOffset, endPath, endOffset);
+
+    if (shouldBeEnabled) {
+      await TinyUiActions.pWaitForUi(editor, '.tox-tbtn[title="Table caption"]');
+      UiFinder.notExists(SugarBody.body(), '.tox-tbtn--disabled[title="Table caption"]');
```
lnewson, line 27:
```suggestion
      await TinyUiActions.pWaitForUi(editor, '.tox-tbtn[title="Table caption"]:not(.tox-tbtn-disabled)');
```
This can be shortened to this I believe.

modules/tinymce/src/plugins/table/test/ts/browser/settings/ToolbarButtonDisplayTest.ts
HAFRMO: Not sure what you mean here.

lnewson: You can just return the result of the function, so we don't need to do an async/await.

HAFRMO: Oh, you mean so. Yeah, that's true.

modules/tinymce/src/plugins/table/main/ts/selection/SelectionTargets.ts
```diff
@@ -62,18 +62,25 @@ export const getSelectionTargets = (editor: Editor, selections: Selections): Sel
 
   const isCaption = SugarNode.isTag('caption');
   const isDisabledForSelection = (key: keyof ExtractedSelectionDetails) => selectionDetails.forall((details) => !details[key]);
+  const getStart = () => TableSelection.getSelectionStartCellOrCaption(Util.getSelectionStart(editor), Util.getIsRoot(editor));
+  const getEnd = () => TableSelection.getSelectionEndCellOrCaption(Util.getSelectionEnd(editor), Util.getIsRoot(editor));
+
+  const findTargets = (): Optional<RunOperation.CombinedTargets> =>
```
ashfordneil, line 68: Could this be simpler with `Optionals.lift2`?

Something like

```typescript
const start = getStart();
const end = getEnd().bind(TableLookup.table);

return Optionals.lift2(start, end, (startCellOrCaption, endTable) => {
 ...
});
```

That many nested binds looks like an anti pattern to me

modules/tinymce/src/plugins/table/main/ts/selection/TableSelection.ts
```diff
@@ -12,22 +12,27 @@ import { Attribute, Compare, SelectorFind, SugarElement, SugarElements, SugarNod
...
 import { ephemera } from './Ephemera';
 
-const getSelectionStartCellFallback = (start: SugarElement<Node>) =>
-  TableLookup.table(start).bind((table) =>
+const getSelectionCellFallback = (element: SugarElement<Node>, takeLast: boolean) =>
+  TableLookup.table(element).bind((table) =>
     TableSelection.retrieve(table, ephemera.firstSelectedSelector)
-  ).fold(() => start, (cells) => cells[0]);
+  ).fold(
+    () => element,
```
ashfordneil, line 19:
```suggestion
    Fun.constant(element),
```

ashfordneil (approved):

modules/tinymce/src/plugins/table/main/ts/selection/SelectionTargets.ts
lnewson: I did try to do this fwiw, but because we need to return an `Optional` inside we'd end up with the type being `Optional<Optional<RunOperation.CombinedTargets>>` so it still ends up fairly messy to handle that.

ashfordneil: `.bind(Fun.identity)` once at the end might not be a bad trade-off

lnewson: There's `Optionals.flatten` for that instead, but it's still not pretty.

ashfordneil: This feels cleaner to me — it separates the "is this in the same table as the end of the selection" out from the other logic with `.filter`
```typescript
  const findTargets = (): Optional<RunOperation.CombinedTargets> =>
    getStart().filter((startCellOrCaption) => {
      const startTableOpt = TableLookup.table(startCellOrCaption);
      const endTableOpt = getEnd().bind(TableLookup.table);
      return Optionals.lift2(startTableOpt, endTableOpt, Compare.eq).getOr(false)
    }).map((startCellOrCaption) => {
      if (isCaption(startCellOrCaption)) {
        return TableTargets.noMenu(startCellOrCaption);
      } else {
        // We know this is okay, because it has to be Some for the previous filter to pass
        const startTable = TableLookup.table(startCellOrCaption).getOrDie();
        return TableTargets.forMenu(selections, startTable, startCellOrCaption);
      }
    });
```

modules/tinymce/src/plugins/table/main/ts/selection/TableSelection.ts
```diff
@@ -6,28 +6,33 @@
...
-import { Arr, Optionals } from '@ephox/katamari';
+import { Arr, Fun, Optionals } from '@ephox/katamari';
 import { TableLookup } from '@ephox/snooker';
 import { Attribute, Compare, SelectorFind, SugarElement, SugarElements, SugarNode } from '@ephox/sugar';
 
 import { ephemera } from './Ephemera';
 
-const getSelectionStartCellFallback = (start: SugarElement<Node>) =>
-  TableLookup.table(start).bind((table) =>
+const getSelectionCellFallback = (element: SugarElement<Node>, takeLast: boolean) =>
```
lnewson, line 15: The `takeLast` argument here is never used as it's value is always `false` and doesn't really make sense either as the cell fallback logic is used to move from a col element to a cell element, so it should always be the first row/cell we want.

- [ ] Revert `takeLast` logic

modules/tinymce/src/plugins/table/test/ts/browser/settings/ToolbarButtonDisplayTest.ts
```diff
@@ -0,0 +1,47 @@
...
+      await TinyUiActions.pWaitForUi(editor, '.tox-tbtn--disabled[title="Table caption"]');
+    }
+  };
+
+  context('Ensure table caption buttons are enabled when appropriate', () => {
+    it('TINY-7737: When the whole selection is in the table', async () =>
+      pDisplayTest([ 1, 0, 0, 0 ], 0, [ 1, 0, 0, 0 ], 1, true)
+    );
+
+    it('TINY-7737: When the selection starts outside the table, but ends in the table', async () =>
```
lnewson, line 35: This should not be `async` as there is no `await` called.

- [ ] Fix incorrect async awaits

modules/tinymce/src/plugins/table/test/ts/browser/settings/ToolbarButtonDisplayTest.ts
```diff
@@ -0,0 +1,47 @@
...
+import Theme from 'tinymce/themes/silver/Theme';
+
+describe('browser.tinymce.plugins.table.ToolbarButtonDisplayTest', () => {
+  const hook = TinyHooks.bddSetupLight<Editor>({
+    plugins: 'table',
+    toolbar: 'tablecaption',
+    base_url: '/project/tinymce/js/tinymce'
+  }, [ Plugin, Theme ], true);
+
+  const editorContent = '<p>A</p><table><tbody><tr><td>B</td></tr></tbody></table><p>C</p>';
```
lnewson, line 15: This is never used elsewhere, so maybe we should inline it?

modules/tinymce/src/plugins/table/test/ts/browser/settings/TableToolbarTest.ts
```diff
@@ -18,6 +18,7 @@ describe('browser.tinymce.plugins.table.TableToolbarTest', () => {
 
   const pCreateEditor = (toolbar: string) => McEditor.pFromSettings<Editor>({
     plugins: 'table',
+    toolbar: 'tablecaption',
```
lnewson, line 21: Why was this added here? It looks like this wasn't reverted properly when the test logic was moved.

- [ ] Revert this change please

lnewson (approved):

modules/tinymce/src/plugins/table/main/ts/selection/TableSelection.ts
```diff
@@ -6,28 +6,33 @@
...
+  TableLookup.table(element).bind((table) =>
     TableSelection.retrieve(table, ephemera.firstSelectedSelector)
-  ).fold(() => start, (cells) => cells[0]);
+  ).fold(
+    Fun.constant(element),
+    (cells) => takeLast ? cells[cells.length - 1] : cells[0]);
```
lnewson, line 20: This formatting is incorrect, as the `);` should be on a newline. However, this shouldn't matter when this is reverted, as it it should go back to a single line.

modules/tinymce/src/plugins/table/main/ts/selection/TableSelection.ts
```diff
@@ -6,28 +6,33 @@
...
+const getSelectionStartCaption = getSelectionFromSelector<HTMLTableCaptionElement>('caption', false);
 
-const getSelectionStartCaption = getSelectionStartFromSelector<HTMLTableCaptionElement>('caption');
+const getSelectionStartCell = getSelectionFromSelector<HTMLTableCellElement>('th,td', false);
 
-const getSelectionStartCell = getSelectionStartFromSelector<HTMLTableCellElement>('th,td');
+const getSelectionStartCellOrCaption = getSelectionFromSelector<HTMLTableCellElement | HTMLTableCaptionElement>('th,td,caption', false);
 
-const getSelectionStartCellOrCaption = getSelectionStartFromSelector<HTMLTableCellElement | HTMLTableCaptionElement>('th,td,caption');
+const getSelectionEndCellOrCaption = getSelectionFromSelector<HTMLTableCellElement | HTMLTableCaptionElement>('th,td,caption', false);
```
lnewson, line 35: The implementation of these 2 are identical. This could be changed to just a single function and renamed to `getSelectionCellOrCaption` or similar.

The same could also be done with the other 2 functions, though honestly this whole file confuses me as it's never doing anything with the actual selection. I'm not sure if you want to go that far, but at the very least please just remove the references to `Start` in this file.

- [ ] Cleanup this logic a little

modules/tinymce/src/plugins/table/main/ts/selection/TableSelection.ts
HAFRMO: You are right, it should have been true at one point. My bad, fixed that.

modules/tinymce/src/plugins/table/main/ts/selection/TableSelection.ts
HAFRMO: The error was a boolean that should have been true, to make it pick the last selected cell instead of the first when using the later function. Fixed that.

modules/tinymce/src/plugins/table/main/ts/selection/TableSelection.ts
HAFRMO: It shouldn't be rewerted, since there should be situations where it's used, now when I've added the correct variables. Not quite sure what the situation is, but it should be there now.

modules/tinymce/src/plugins/table/test/ts/browser/settings/ToolbarButtonDisplayTest.ts
HAFRMO: Sure, if you want.

modules/tinymce/src/plugins/table/main/ts/selection/SelectionTargets.ts
HAFRMO: Then we're looking up the table twice, and we do the `getOrDie`. That there needs to be a comment saying it's fine speaks against it being an improvement to me.

modules/tinymce/src/plugins/table/main/ts/selection/TableSelection.ts
lnewson: Did you read what I said above? This logic is **incorrect** so it shouldn't have been fixed to use it and should have been removed. To re-iterate this handles the case where the selection ends up in an invalid location within a `<col>` element (Firefox IIRC was the only browser that did this). So when it gets in that state it should **always** be using the first cell as that's always the closest valid position to use. Using the last cell within a selection is definitely the wrong thing to be doing.

HAFRMO: I read it, but misunderstood what you referred to. My apologizes.

modules/tinymce/src/plugins/table/main/ts/selection/SelectionTargets.ts
ashfordneil: Fair enough, let's ship it as is then

PR #7048, opened 2021-07-27
TINY-7374: Fixed SandHTMLElement not working across different windows

lnewson:
Related Ticket: TINY-7374

Description of Changes:
* Fix the `SandHTMLElement.isPrototypeOf` function not working when an element is constructed in another window.

**Note:** Tests were failing because of the `ColumnSizesTest` issue on Chrome 92, so I've also cherry-picked that commit as well.

Pre-checks:
* [x] ~Changelog entry added~ No changelogs on master

Review:
* [x] Milestone set

### DIFF
--- a/modules/sand/src/main/ts/ephox/sand/api/SandHTMLElement.ts
+++ b/modules/sand/src/main/ts/ephox/sand/api/SandHTMLElement.ts
@@ -1,5 +1,7 @@
-import { Resolve } from '@ephox/katamari';
+import { Resolve, Type } from '@ephox/katamari';
 import * as Global from '../util/Global';
+
+const getPrototypeOf = Object.getPrototypeOf;
 
 /*
  * IE9 and above
@@ -16,7 +18,9 @@ const isPrototypeOf = (x: any): x is HTMLElement => {
   // undefined scope later triggers using the global window.
   const scope: Window | undefined = Resolve.resolve('ownerDocument.defaultView', x);
 
-  return sandHTMLElement(scope).prototype.isPrototypeOf(x);
+  // TINY-7374: We can't rely on looking at the owner window HTMLElement as the element may have
+  // been constructed in a different window and then appended to the current window document.
+  return Type.isObject(x) && (sandHTMLElement(scope).prototype.isPrototypeOf(x) || /^\[object HTML\w*Element\]$/.test(getPrototypeOf(x).toString()));
 };
 
 export {
--- a/modules/sand/src/test/ts/browser/HtmlElementTest.ts
+++ b/modules/sand/src/test/ts/browser/HtmlElementTest.ts
@@ -1,11 +1,53 @@
-import { assert, UnitTest } from '@ephox/bedrock-client';
-import { SandHTMLElement } from 'ephox/sand/api/Main';
+import { context, describe, it } from '@ephox/bedrock-client';
+import { assert } from 'chai';
 
-UnitTest.test('HtmlElementTest', () => {
-  const span = document.createElement('div');
-  assert.eq(false, SandHTMLElement.isPrototypeOf(null));
-  assert.eq(false, SandHTMLElement.isPrototypeOf(undefined));
-  assert.eq(false, SandHTMLElement.isPrototypeOf('a string'));
-  assert.eq(false, SandHTMLElement.isPrototypeOf({}));
-  assert.eq(true, SandHTMLElement.isPrototypeOf(span));
+import * as SandHTMLElement from 'ephox/sand/api/SandHTMLElement';
+
+describe('HtmlElementTest', () => {
+  context('isPrototypeOf', () => {
+    it('non elements should be false', () => {
+      assert.isFalse(SandHTMLElement.isPrototypeOf(null));
+      assert.isFalse(SandHTMLElement.isPrototypeOf(undefined));
+      assert.isFalse(SandHTMLElement.isPrototypeOf('a string'));
+      assert.isFalse(SandHTMLElement.isPrototypeOf({}));
+    });
+
+    it('nodes should be false', () => {
+      const text = document.createTextNode('text');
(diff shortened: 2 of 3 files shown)


modules/sand/src/main/ts/ephox/sand/api/SandHTMLElement.ts
```diff
@@ -1,6 +1,8 @@
-import { Resolve } from '@ephox/katamari';
+import { Resolve, Type } from '@ephox/katamari';
 import * as Global from '../util/Global';
 
+const getPrototypeOf = Object.getPrototypeOf;
```
lnewson, line 4: I stored this to ensure the API isn't messed with by polyfills. We currently do the same thing in other places for this reason, e.g. `Obj`.

modules/sand/src/main/ts/ephox/sand/api/SandHTMLElement.ts
```diff
@@ -16,7 +18,9 @@ const isPrototypeOf = (x: any): x is HTMLElement => {
   // undefined scope later triggers using the global window.
   const scope: Window | undefined = Resolve.resolve('ownerDocument.defaultView', x);
```
diogo405, line 19: I'd rename `x` to something more meaningful. Element?!

modules/sand/src/test/ts/browser/HtmlElementTest.ts
```diff
@@ -1,11 +1,53 @@
...
-  assert.eq(false, SandHTMLElement.isPrototypeOf({}));
-  assert.eq(true, SandHTMLElement.isPrototypeOf(span));
+import { context, describe, it } from '@ephox/bedrock-client';
+import { assert } from 'chai';
+
+import * as SandHTMLElement from 'ephox/sand/api/SandHTMLElement';
+
+describe('HtmlElementTest', () => {
+  context('isPrototypeOf', () => {
+    it('non elements should be false', () => {
```
diogo405, line 8: Shouldn't it have `TINY-7374` in the test label?

diogo405 (approved):

modules/sand/src/main/ts/ephox/sand/api/SandHTMLElement.ts
lnewson: Not something changed in this PR.

modules/sand/src/test/ts/browser/HtmlElementTest.ts
lnewson: No, this is the existing functionality and not related to this change. There's only the one test that's related to this change.

ltrouton (approved):

JamesToohey (approved):

HAFRMO (approved):

modules/sand/src/main/ts/ephox/sand/api/SandHTMLElement.ts
```diff
@@ -16,7 +18,9 @@ const isPrototypeOf = (x: any): x is HTMLElement => {
...
 
-  return sandHTMLElement(scope).prototype.isPrototypeOf(x);
+  // TINY-7374: We can't rely on looking at the owner window HTMLElement as the element may have
+  // been constructed in a different window and then appended to the current window document.
+  return Type.isObject(x) && (sandHTMLElement(scope).prototype.isPrototypeOf(x) || /^\[object HTML\w*Element\]$/.test(getPrototypeOf(x).toString()));
```
ashfordneil, line 23: Instead of doing the main test, and then falling back to the regex, why not only do the regex?

```suggestion
  return Type.isObject(x) && /^\[object HTML\w*Element\]$/.test(getPrototypeOf(x).toString());
```

modules/sand/src/test/ts/browser/HtmlElementTest.ts
```diff
@@ -1,11 +1,53 @@
...
+
+      const iframeDoc = iframe.contentDocument;
+      iframeDoc.open();
+      iframeDoc.write('<html><body></body></html>');
+      iframeDoc.close();
+
+      iframeDoc.body.appendChild(span);
+      iframeDoc.body.appendChild(strong);
+      assert.isTrue(SandHTMLElement.isPrototypeOf(span));
+      assert.isTrue(SandHTMLElement.isPrototypeOf(strong));
```
ashfordneil, line 48: Is this asserting what we want it to? The `span` and `strong` were still created in the parent document. Would it be better to do something like

```typescript
const span = iframeDoc.createElement('span');
const strong = iframeDoc.createElement('strong');

assert.isTrue(SandHTMLElement.isPrototypeOf(span));
assert.isTrue(SandHTMLElement.isPrototypeOf(strong));
```

ashfordneil (approved):

modules/sand/src/main/ts/ephox/sand/api/SandHTMLElement.ts
lnewson: The other case is a guaranteed version so it 100% should not be removed. This also keeps it consistent with the pattern used in `Type`.

modules/sand/src/test/ts/browser/HtmlElementTest.ts
lnewson: No... that's the exact opposite of what we want to check and would have passed before this change. As it gets the owner document -> owner window -> HTML element currently so that would already work for the above. The case it fails is the one in a test, whereby it's created with a document in window x and then added to window y. Since it'll use the `HTMLElement` from window y it fails since the element prototype is from window x.

modules/sand/src/main/ts/ephox/sand/api/SandHTMLElement.ts
lnewson: I should add this also gives us some protection in case something isn't caught by the prototype name comparison.

PR #7049, opened 2021-07-28
TINY-7700: Disabled menu items with children can still be opened

diogo405:
Related Ticket:

Description of Changes:
* Do not open disabled menu items

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/alloy/CHANGELOG.md
+++ b/modules/alloy/CHANGELOG.md
@@ -28,6 +28,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed a number of positioning bugs when using an "inset" layout, especially any that renders at the bottom.
 - `LayoutInset` bubble classes were applied incorrectly, causing the bubble arrows to show on the opposite side.
 - `MakeshiftAnchorSpec` did not extend `HasLayoutAnchorSpec` which meant layouts couldn't be used for makeshift anchors.
+- Disabled nested menu items could still be opened.
 
 ## 8.2.0 - 2021-05-06
 
--- a/modules/alloy/CHANGELOG.md
+++ b/modules/alloy/CHANGELOG.md
@@ -28,7 +28,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed a number of positioning bugs when using an "inset" layout, especially any that renders at the bottom.
 - `LayoutInset` bubble classes were applied incorrectly, causing the bubble arrows to show on the opposite side.
 - `MakeshiftAnchorSpec` did not extend `HasLayoutAnchorSpec` which meant layouts couldn't be used for makeshift anchors.
-- Disabled nested menu items could still be opened.
+- `TieredMenu` child menus were incorrectly opened when the parent menu item was disabled.
 
 ## 8.2.0 - 2021-05-06
 
--- a/modules/alloy/src/main/ts/ephox/alloy/ui/single/TieredMenuSpec.ts
+++ b/modules/alloy/src/main/ts/ephox/alloy/ui/single/TieredMenuSpec.ts
@@ -3,6 +3,7 @@ import { Attribute, Class, Classes, SelectorFilter, SelectorFind, SugarBody } fr
 
 import * as EditableFields from '../../alien/EditableFields';
 import { Composing } from '../../api/behaviour/Composing';
+import { Disabling } from '../../api/behaviour/Disabling';
 import { Highlighting } from '../../api/behaviour/Highlighting';
 import { Keying } from '../../api/behaviour/Keying';
 import { Replacing } from '../../api/behaviour/Replacing';
@@ -170,31 +171,35 @@ const make: SingleSketchFactory<TieredMenuDetail, TieredMenuSpec> = (detail, _ra
   };
 
   const expandRight = (container: AlloyComponent, item: AlloyComponent, decision: ExpandHighlightDecision = ExpandHighlightDecision.HighlightSubmenu): Optional<AlloyComponent> => {
-    const value = getItemValue(item);
-    return layeredState.expand(value).bind((path) => {
-      // Called when submenus are opened by keyboard AND hovering navigation
-      updateAriaExpansions(container, path);
-      // When expanding, always select the first.
-      return Optional.from(path[0]).bind((menuName) => layeredState.lookupMenu(menuName).bind((activeMenuPrep) => {
-        const activeMenu = buildIfRequired(container, menuName, activeMenuPrep);
+    if (item.hasConfigured(Disabling) && Disabling.isDisabled(item)) {
+      return Optional.none();
+    } else {
+      const value = getItemValue(item);
+      return layeredState.expand(value).bind((path) => {
+        // Called when submenus are opened by keyboard AND hovering navigation
+        updateAriaExpansions(container, path);
+        // When expanding, always select the first.
+        return Optional.from(path[0]).bind((menuName) => layeredState.lookupMenu(menuName).bind((activeMenuPrep) => {
+          const activeMenu = buildIfRequired(container, menuName, activeMenuPrep);
 
-        // DUPE with above. Fix later.
-        if (!SugarBody.inBody(activeMenu.element)) {
-          Replacing.append(container, GuiFactory.premade(activeMenu));
-        }
+          // DUPE with above. Fix later.
+          if (!SugarBody.inBody(activeMenu.element)) {
+            Replacing.append(container, GuiFactory.premade(activeMenu));
+          }
 
-        // updateMenuPath is the code which changes the active menu. We don't always
-        // want to change the active menu. Sometimes, we just want to show it (e.g. hover)
-        detail.onOpenSubmenu(container, item, activeMenu, Arr.reverse(path));
-        if (decision === ExpandHighlightDecision.HighlightSubmenu) {
-          Highlighting.highlightFirst(activeMenu);
-          return updateMenuPath(container, layeredState, path);
-        } else {
-          Highlighting.dehighlightAll(activeMenu);
-          return Optional.some(item);
-        }
-      }));
-    });
+          // updateMenuPath is the code which changes the active menu. We don't always
+          // want to change the active menu. Sometimes, we just want to show it (e.g. hover)
+          detail.onOpenSubmenu(container, item, activeMenu, Arr.reverse(path));
+          if (decision === ExpandHighlightDecision.HighlightSubmenu) {
+            Highlighting.highlightFirst(activeMenu);
+            return updateMenuPath(container, layeredState, path);
+          } else {
+            Highlighting.dehighlightAll(activeMenu);
+            return Optional.some(item);
+          }
+        }));
+      });
+    }
   };
 
   const collapseLeft = (container: AlloyComponent, item: AlloyComponent): Optional<AlloyComponent> => {
--- a/modules/alloy/src/test/ts/browser/ui/dropdown/TieredMenuTest.ts
+++ b/modules/alloy/src/test/ts/browser/ui/dropdown/TieredMenuTest.ts
@@ -37,18 +37,25 @@ UnitTest.asynctest('TieredMenuTest', (success, failure) => {
             items: Arr.map([
               { type: 'item', data: { value: 'a-alpha', meta: { text: 'a-Alpha' }}, hasSubmenu: false },
               { type: 'item', data: { value: 'a-beta', meta: { text: 'a-Beta' }}, hasSubmenu: true },
-              { type: 'item', data: { value: 'a-gamma', meta: { text: 'a-Gamma' }}, hasSubmenu: false }
+              { type: 'item', data: { value: 'a-gamma', meta: { text: 'a-Gamma', disabled: true }}, hasSubmenu: true }
             ], TestDropdownMenu.renderItem)
           },
           'a-beta': { // menu name should be triggering parent item so TieredMenuSpec path works
             value: 'menu-b',
             items: Arr.map([
-              { type: 'item', data: { value: 'b-alpha', meta: { text: 'b-Alpha' }}, hasSubmenu: false }
+              { type: 'item', data: { value: 'b-alpha', meta: { text: 'b-Alpha' }}, hasSubmenu: false },
+            ], TestDropdownMenu.renderItem)
+          },
+          'a-gamma': {
+            value: 'menu-c',
+            items: Arr.map([
+              { type: 'item', data: { value: 'c-alpha', meta: { text: 'c-Alpha' }}, hasSubmenu: false },
             ], TestDropdownMenu.renderItem)
           }
         }, TestDropdownMenu.renderMenu),
         expansions: {
-          'a-beta': 'a-beta'
+          'a-beta': 'a-beta',
+          'a-gamma': 'a-gamma',
         }
       },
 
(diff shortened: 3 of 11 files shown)


diogo405 (commented):
The diff doesn't help much so this is what I added to `expandRight`:
```
if (item.hasConfigured(Disabling) && Disabling.isDisabled(item)) {
      return Optional.none();
    } else {
...
```

modules/tinymce/src/themes/silver/test/ts/webdriver/editor/DisabledMenuItemTest.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  const assertVerticalAlignMenuIsNotOpen = () => {
+    UiFinder.notExists(SugarBody.body(), '[role="menuitemcheckbox"]:contains("None")');
+    UiFinder.notExists(SugarBody.body(), '[role="menuitemcheckbox"]:contains("Top")');
+    UiFinder.notExists(SugarBody.body(), '[role="menuitemcheckbox"]:contains("Middle")');
+    UiFinder.notExists(SugarBody.body(), '[role="menuitemcheckbox"]:contains("Bottom")');
+  };
+
+  it('TINY-7700: Disabled menu item with children should not open on mouse hover', async () => {
+    const editor = await McEditor.pFromSettings(settings);
```
ashfordneil, line 39: Should this be done with `TinyHooks`? It looks like you're using the same editor config for all the tests

ashfordneil (approved):

modules/tinymce/src/themes/silver/test/ts/webdriver/editor/DisabledMenuItemTest.ts
diogo405: My first go was using `TinyHooks`. I changed it cause for some reason (dunno) it was breaking `pSendKeysOn` from the last test. It was throwing an error about the selector `'[title="Vertical align"]'`. The only thing I can think of is that the previous editor was causing that. So I used the `McEditor` to add/remove the editor in each test and it worked. Dunno if that makes sense 🤷

modules/tinymce/src/themes/silver/test/ts/webdriver/editor/DisabledMenuItemTest.ts
```diff
@@ -0,0 +1,55 @@
...
+    menu: {
+      table: { title: 'Table', items: 'inserttable | cell row column | tableclass tablecellclass tablecellvalign tablecellborderwidth tablecellborderstyle tablecaption tablecellbackgroundcolor tablecellbordercolor | advtablesort | tableprops deletetable' }
+    },
+    menubar: 'table',
+  };
+
+  const pOpenTableMenu = (editor) => {
+    const tableMenuItemSelector = '[role="menuitem"]:contains("Table")';
+    TinyUiActions.clickOnMenu(editor, tableMenuItemSelector);
+    return Waiter.pTryUntil('Wait for a specific menu to open', () => UiFinder.exists(SugarBody.body(), tableMenuItemSelector));
```
JamesToohey, line 28: This will always be present as you're using the same selector to open the menu.

That is, `tableMenuItemSelector` is always in the menubar. You need to use a selector to look for the vertical align button for this assertion instead.

`const verticalAlignMenuItemSelector = 'div[title="Vertical align"]';`

I suspect this is the root cause of your issues

modules/tinymce/src/themes/silver/test/ts/webdriver/editor/DisabledMenuItemTest.ts
JamesToohey: We want to avoid instantiating editors whenever we can, so using `McEditor` sparingly. Often when using `TinyHooks` we need to tidy up between each test (closing dialogs, menus, clearing editor content). This can sometimes be done in the `beforeEach` and `afterEach` hooks.

See my other comments as to why your test wasn't working

modules/tinymce/src/themes/silver/test/ts/webdriver/editor/DisabledMenuItemTest.ts
```diff
@@ -0,0 +1,55 @@
...
+    base_url: '/project/tinymce/js/tinymce',
+    plugins: 'table',
+    menu: {
+      table: { title: 'Table', items: 'inserttable | cell row column | tableclass tablecellclass tablecellvalign tablecellborderwidth tablecellborderstyle tablecaption tablecellbackgroundcolor tablecellbordercolor | advtablesort | tableprops deletetable' }
+    },
+    menubar: 'table',
+  };
+
+  const pOpenTableMenu = (editor) => {
+    const tableMenuItemSelector = '[role="menuitem"]:contains("Table")';
```
JamesToohey, line 26: You can move this selector out of this function

modules/tinymce/src/themes/silver/test/ts/webdriver/editor/DisabledMenuItemTest.ts
```diff
@@ -0,0 +1,55 @@
...
+  const settings = {
+    base_url: '/project/tinymce/js/tinymce',
+    plugins: 'table',
+    menu: {
+      table: { title: 'Table', items: 'inserttable | cell row column | tableclass tablecellclass tablecellvalign tablecellborderwidth tablecellborderstyle tablecaption tablecellbackgroundcolor tablecellbordercolor | advtablesort | tableprops deletetable' }
+    },
+    menubar: 'table',
+  };
+
+  const pOpenTableMenu = (editor) => {
```
JamesToohey, line 25: Missing a type here 
```suggestion
  const pOpenTableMenu = (editor: Editor) => {
```

modules/tinymce/src/themes/silver/test/ts/webdriver/editor/DisabledMenuItemTest.ts
JamesToohey: I'm also not sure this label makes sense? Should it be something like "Wait for table menu to open"?

JamesToohey (changes requested):
Requesting changes so we can tidy up the `DisabledMenuItemTest`. This will likely mean cleaning up the open menus with a helper function between the tests.

- [x] Use TinyHooks for `DisabledMenuItemTest`
- [ ] Fix selectors in test

modules/tinymce/src/themes/silver/test/ts/webdriver/editor/DisabledMenuItemTest.ts
```diff
@@ -0,0 +1,53 @@
...
+    UiFinder.notExists(SugarBody.body(), '[role="menuitemcheckbox"]:contains("Middle")');
+    UiFinder.notExists(SugarBody.body(), '[role="menuitemcheckbox"]:contains("Bottom")');
+  };
+
+  beforeEach(async () => {
+    await pOpenTableMenu();
+  });
+
+  afterEach(() => {
+    assertVerticalAlignMenuIsNotOpen();
```
lnewson, line 43: Assertions should never be done in hooks, they should only contain setup and teardown logic.

- [ ] Revert this change please

modules/tinymce/src/themes/silver/test/ts/webdriver/editor/DisabledMenuItemTest.ts
```diff
@@ -0,0 +1,53 @@
...
+import Theme from 'tinymce/themes/silver/Theme';
+
+describe('webdriver.tinymce.themes.silver.editor.menubar.DisabledMenuItemTest', () => {
+
+  const verticalAlignMenuItemSelector = '[title="Vertical align"]';
+  const tableMenuItemSelector = '[role="menuitem"]:contains("Table")';
+
+  const hook = TinyHooks.bddSetupLight<Editor>({
+    base_url: '/project/tinymce/js/tinymce',
+    plugins: 'table',
```
lnewson, line 17: We generally try to avoid pulling in plugins in silver tests unless absolutely necessary. For this case it's not necessary, so can we please either move this test to the table plugin or just register our own nested menu and do assertions against that (preferably the latter option).

lnewson: To provide more context, it's not the table plugin logic we want to test from the theme perspective but instead the fact that a nested menu won't open when it's disabled. This can also help with maintainability, as if we refactor how the table plugin works then we have less places we have to make updates to (ie it's largely all isolated inside the table plugin).

modules/tinymce/src/themes/silver/test/ts/webdriver/editor/DisabledMenuItemTest.ts
```diff
@@ -0,0 +1,53 @@
+import { Keys, Mouse, RealKeys, UiFinder, Waiter } from '@ephox/agar';
+import { afterEach, beforeEach, describe, it } from '@ephox/bedrock-client';
+import { TinyHooks, TinyUiActions } from '@ephox/mcagar';
+import { SugarBody } from '@ephox/sugar';
+
+import { Editor } from 'tinymce/core/api/PublicApi';
+import TablePlugin from 'tinymce/plugins/table/Plugin';
+import Theme from 'tinymce/themes/silver/Theme';
+
+describe('webdriver.tinymce.themes.silver.editor.menubar.DisabledMenuItemTest', () => {
```
lnewson, line 10: Should this be called `DisabledNestedMenuItemTest`? As it's not testing a regular menu item, but instead a nested menu item.

modules/alloy/src/test/ts/browser/ui/dropdown/TieredMenuTest.ts
```diff
@@ -37,18 +37,25 @@ UnitTest.asynctest('TieredMenuTest', (success, failure) => {
             items: Arr.map([
               { type: 'item', data: { value: 'a-alpha', meta: { text: 'a-Alpha' }}, hasSubmenu: false },
               { type: 'item', data: { value: 'a-beta', meta: { text: 'a-Beta' }}, hasSubmenu: true },
-              { type: 'item', data: { value: 'a-gamma', meta: { text: 'a-Gamma' }}, hasSubmenu: false }
+              { type: 'item', data: { value: 'a-gamma', meta: { text: 'a-Gamma', disabled: true }}, hasSubmenu: true }
```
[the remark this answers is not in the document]
lnewson: This seems fine to me :shrug: This data structure is only used for tests anyways, so this seems as good of place as any given it doesn't really matter how it's passed in this instance.

modules/alloy/src/main/ts/ephox/alloy/ui/single/TieredMenuSpec.ts
```diff
@@ -170,31 +171,35 @@ const make: SingleSketchFactory<TieredMenuDetail, TieredMenuSpec> = (detail, _ra
...
-      updateAriaExpansions(container, path);
-      // When expanding, always select the first.
-      return Optional.from(path[0]).bind((menuName) => layeredState.lookupMenu(menuName).bind((activeMenuPrep) => {
-        const activeMenu = buildIfRequired(container, menuName, activeMenuPrep);
-
-        // DUPE with above. Fix later.
-        if (!SugarBody.inBody(activeMenu.element)) {
-          Replacing.append(container, GuiFactory.premade(activeMenu));
-        }
+    if (item.hasConfigured(Disabling) && Disabling.isDisabled(item)) {
```
lnewson, line 174: This needs a changelog entry for Alloy as well.

modules/tinymce/CHANGELOG.md
```diff
@@ -60,6 +60,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Using the Tab key to navigate into the editor on IE 11 would incorrectly focus the toolbar #TINY-3707
 - The editor selection could be placed in an incorrect location when undoing or redoing changes in a document containing `contenteditable="false"` elements #TINY-7663
 - Unbinding a native event handler inside the `remove` event caused an exception that blocked editor removal #TINY-7730
+- Disabled menu items could still be opened #TINY-7700
```
lnewson, line 63:
```suggestion
- Disabled nested menu items could still be opened #TINY-7700
```

modules/tinymce/src/themes/silver/test/ts/webdriver/editor/DisabledMenuItemTest.ts
diogo405: Yep, makes sense 👍

modules/tinymce/src/themes/silver/test/ts/webdriver/editor/DisabledMenuItemTest.ts
```diff
@@ -0,0 +1,53 @@
...
+    plugins: 'table',
+    menu: {
+      table: { title: 'Table', items: 'inserttable | cell row column | tableclass tablecellclass tablecellvalign tablecellborderwidth tablecellborderstyle tablecaption tablecellbackgroundcolor tablecellbordercolor | advtablesort | tableprops deletetable' }
+    },
+    menubar: 'table',
+  }, [ Theme, TablePlugin ]);
+
+  const pOpenTableMenu = () => {
+    const editor = hook.editor();
+    TinyUiActions.keydown(editor, Keys.escape()); // to close table menu before opening
```
JamesToohey, line 26: Could we move this into an `afterEach` rather than it being in the helper function for opening the menu? I also think we should have `pOpenTableMenu` called in each test rather than in the `beforeEach`, otherwise it's difficult to discern what the tests are doing

diogo405: Done.

modules/alloy/CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Fixed a number of positioning bugs when using an "inset" layout, especially any that renders at the bottom.
 - `LayoutInset` bubble classes were applied incorrectly, causing the bubble arrows to show on the opposite side.
 - `MakeshiftAnchorSpec` did not extend `HasLayoutAnchorSpec` which meant layouts couldn't be used for makeshift anchors.
+- Disabled nested menu items could still be opened.
```
lnewson, line 31: This needs to be a little different to the TinyMCE changelog entry... so it probably needs to be something more like this:
```suggestion
- `TieredMenu` child menus were incorrectly opened when the parent menu item was disabled.
```

modules/tinymce/src/themes/silver/test/ts/webdriver/editor/DisabledNestedMenuItemTest.ts
```diff
@@ -0,0 +1,72 @@
...
+        text: 'Preferences',
+        disabled: true,
+        getSubmenuItems: () => {
+          return [
+            {
+              type: 'menuitem',
+              text: 'Settings',
+            }
+          ];
+        }
```
lnewson, line 39:
```suggestion
        getSubmenuItems: () => [{
          type: 'menuitem',
          text: 'Settings',
        }]
```
This can be slimmed down by using shorthand syntax.

lnewson (approved):

modules/tinymce/src/themes/silver/test/ts/webdriver/editor/DisabledNestedMenuItemTest.ts
```diff
@@ -0,0 +1,68 @@
...
+    UiFinder.notExists(SugarBody.body(), '[role="menuitem"]:contains("Settings")');
+  };
+
+  afterEach(() => {
+    closeCodeMenu();
+  });
+
+  it('TINY-7700: Disabled menu item with children should not open on mouse hover', async () => {
+    await pOpenCodeMenu();
+    Mouse.hoverOn(SugarBody.body(), '[role="menuitem"]:contains("Preferences")');
```
JamesToohey, line 59: Can we use `preferencesMenuItemSelector` here?

JamesToohey (approved):

ashfordneil (approved):

PR #7051, opened 2021-07-28
TINY-7756: Fixed attributes clobbering native document properties causing the editor to crash

lnewson:
Related Ticket: TINY-7756

Description of Changes:
This updates the parser to mark attributes as unsafe if they would clobber a document/form property. For those that aren't aware, using an id/name can cause the browser to store a reference to that element on the owner `document` or `form`. As such, using certain names/ids will cause the browser to overwrite the existing properties/functions.

Note: This is only part 1 of the issues for this Jira and I'm just creating the PR now to keep the PR size smaller as discussed earlier in the week.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/test/ts/browser/dom/SerializerTest.ts
+++ b/modules/tinymce/src/core/test/ts/browser/dom/SerializerTest.ts
@@ -163,8 +163,8 @@ describe('browser.tinymce.core.dom.SerializerTest', () => {
     DOM.setHTML('test', '<input type="text" value="text" length="128" maxlength="129" />');
     assert.equal(ser.serialize(DOM.get('test')), '<input type="text" value="text" length="128" maxlength="129" />');
 
-    DOM.setHTML('test', '<form method="post"><input type="hidden" name="method" value="get" /></form>');
-    assert.equal(ser.serialize(DOM.get('test')), '<form method="post"><input type="hidden" name="method" value="get" /></form>');
+    DOM.setHTML('test', '<form method="post"><input type="hidden" name="formmethod" value="get" /></form>');
+    assert.equal(ser.serialize(DOM.get('test')), '<form method="post"><input type="hidden" name="formmethod" value="get" /></form>');
 
     DOM.setHTML('test', '<label for="test">label</label>');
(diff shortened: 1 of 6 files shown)


modules/tinymce/src/core/test/ts/browser/dom/SerializerTest.ts
```diff
@@ -163,8 +163,8 @@ describe('browser.tinymce.core.dom.SerializerTest', () => {
     DOM.setHTML('test', '<input type="text" value="text" length="128" maxlength="129" />');
     assert.equal(ser.serialize(DOM.get('test')), '<input type="text" value="text" length="128" maxlength="129" />');
 
-    DOM.setHTML('test', '<form method="post"><input type="hidden" name="method" value="get" /></form>');
-    assert.equal(ser.serialize(DOM.get('test')), '<form method="post"><input type="hidden" name="method" value="get" /></form>');
+    DOM.setHTML('test', '<form method="post"><input type="hidden" name="formmethod" value="get" /></form>');
```
lnewson, line 166: This was changed as it clobbers and as such was being removed. Also interesting to note the spec specifically mentions use "method" as a name should be avoided because it does clobber: https://html.spec.whatwg.org/#naming-form-controls:-the-name-attribute

modules/tinymce/src/core/test/ts/browser/dom/SerializerTest.ts
```diff
@@ -163,8 +163,8 @@ describe('browser.tinymce.core.dom.SerializerTest', () => {
...
+    assert.equal(ser.serialize(DOM.get('test')), '<form method="post"><input type="hidden" name="formmethod" value="get" /></form>');
```
diogo405, line 167: I'd rename `ser` to `serializer` :)

diogo405 (approved):

modules/tinymce/src/core/test/ts/browser/dom/SerializerTest.ts
lnewson: Again, not something related to this change.

HAFRMO (approved):

ashfordneil (approved):

PR #7053, opened 2021-07-28
Fix the color contrast issues for axe a11y audit

SiarheiBobryk:
Elements with the following CSS classes fire the "Text elements must have sufficient color contrast against the background" a11y violation in the axe DevTools:

- [`tox-mbtn__select-label`](https://axe.deque.com/issues/7babe2d7-8d6f-4f82-a3f9-9f38e3d6ab09)
Screen Shot 2021-07-29 at 12 40 26 am

- [`tox-tbtn__select-label`](https://axe.deque.com/issues/a190e94d-b8fe-41f5-aa4c-ad14ef6fef4d)
Screen Shot 2021-07-29 at 12 41 26 am

https://dequeuniversity.com/rules/axe/4.3/color-contrast?application=AxeChrome

As far as I understand the only purpose of `@toolbar-background` and `@menubar-background` is to dynamically generate a transparent SVG background with a 1px bottom line.

So I'd suggest achieving absolutely the same result by setting the `border-bottom` property and as a benefit improve accessibility.

Screen Shot 2021-07-29 at 12 46 27 am

### DIFF
--- a/modules/oxide/src/less/theme/components/menubar/menubar.less
+++ b/modules/oxide/src/less/theme/components/menubar/menubar.less
@@ -10,7 +10,6 @@
 //
 
 @menubar-background-color: @background-color;
-@menubar-background: url("data:image/svg+xml;charset=utf8,%3Csvg height='@{_menubar-height}' viewBox='0 0 40 @{_menubar-height}' width='40' xmlns='http://www.w3.org/2000/svg'%3E%3Crect x='0' y='@{_menubar-divider-height}' width='100' height='1' fill='@{_menubar-row-separator-color}'/%3E%3C/svg%3E") left 0 top 0 @menubar-background-color;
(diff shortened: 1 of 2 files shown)


lostkeys:
Hi @SiarheiBobryk and thanks for your suggestion. The SVG solution is to achieve horizontal separator lines between each toolbar row when the toolbar collapses into multiple rows. If you need to change that you can [create a custom skin](https://www.tiny.cloud/docs/advanced/creating-a-skin/) to add a css border bottom.
We will consider a change in a future major version of TinyMCE.

lnewson:
If it helps the background images aren't causing color/contrast issues, it's just that axe cannot detect the color of the background image and as such is a false-positive.

Additionally, this is the issue @lostkeys mentioned that this PR causes:
image

PR #7055, opened 2021-07-29
TINY-6666: Ensure the existing row is re-used where possible when doing operations

lnewson:
Related Ticket: TINY-6666

Description of Changes:
This is part 1 of this issue and unfortunately was something that we hadn't forseen. The problem was that due to how the models worked internally it would try and lookup the row by looking at the cells and then trying to find the parent row from that. However, this did not work when all cells in a row are replaced.

As such, this makes a change so that the row is maintained in the model state and passed around as required. This also then updates the relevant modification operations to generate a new row and flag it as being new, instead of trying to do this later when converting between the older/newer models.

I've also done a little bit of clean up here and removed some logic, as I found some unused functionality that was just broken. Additionally some of the `Structs` ended up being nearly identical so I merged it to avoid some confusion.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/snooker/CHANGELOG.md
+++ b/modules/snooker/CHANGELOG.md
@@ -6,8 +6,23 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
 
+### Added
+- `Warehouse` now contains the colgroup elements from the table.
+
+### Improved
+- Rows are now stored and tracked in the table model structures.
+
 ### Changed
+- `RowDetails` has been renamed to `RowDetail` and now takes a generic argument to define what is the type of cells stored.
+- `row` and `colgroup` generators are now passed the previous element during transform or modification operations.
 - Upgraded to Katamari 8.0, which includes breaking changes to the `Optional` API used in this module.
+
+### Fixed
+- Table operations that replace all cells in a row now re-use the existing row instead of creating a new row.
+
+### Removed
+- `Structs.RowData` has been merged into and replaced by `Structs.RowDetail` to remove some duplication.
+- The `cursor` function has been removed from `Generators` as it was unused and incorrectly implemented.
 
 ## 8.0.4 - 2021-06-23
 
--- a/modules/snooker/CHANGELOG.md
+++ b/modules/snooker/CHANGELOG.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Rows are now stored and tracked in the table model structures.
 
 ### Changed
-- `RowDetails` has been renamed to `RowDetail` and now takes a generic argument to define what is the type of cells stored.
+- `RowDetails` has been renamed to `RowDetail` and now takes a generic argument to define the type of cells stored.
 - `row` and `colgroup` generators are now passed the previous element during transform or modification operations.
 - Upgraded to Katamari 8.0, which includes breaking changes to the `Optional` API used in this module.
 
--- a/modules/snooker/src/main/ts/ephox/snooker/api/CopyRows.ts
+++ b/modules/snooker/src/main/ts/ephox/snooker/api/CopyRows.ts
@@ -8,7 +8,7 @@ import * as Redraw from '../operate/Redraw';
 import { Generators } from './Generators';
 import { Warehouse } from './Warehouse';
 
-const copyRows = (table: SugarElement<HTMLTableElement>, target: TargetSelection, generators: Generators): Optional<SugarElement<HTMLTableRowElement>[]> => {
+const copyRows = (table: SugarElement<HTMLTableElement>, target: TargetSelection, generators: Generators): Optional<SugarElement<HTMLTableRowElement | HTMLTableColElement>[]> => {
   const warehouse = Warehouse.fromTable(table);
   // Cannot use onUnlockedCells like extractor here as if only cells in a locked column are selected, then this will be Optional.none and
   // there is now no way of knowing which rows are selected
@@ -23,7 +23,7 @@ const copyRows = (table: SugarElement<HTMLTableElement>, target: TargetSelection
       const newCells = Arr.filter(row.cells, (cell) => !cell.isLocked);
       return newCells.length > 0 ? [{ ...row, cells: newCells }] : [];
     });
-    const slicedDetails = toDetailList(filteredGrid, generators);
+    const slicedDetails = toDetailList(filteredGrid);
     return Optionals.someIf(slicedDetails.length > 0, slicedDetails);
   }).map((slicedDetails) => Redraw.copy(slicedDetails));
 };
--- a/modules/snooker/src/main/ts/ephox/snooker/api/Generators.ts
+++ b/modules/snooker/src/main/ts/ephox/snooker/api/Generators.ts
@@ -1,21 +1,25 @@
-import { Arr, Fun, Optional, Optionals, Singleton } from '@ephox/katamari';
+import { Arr, Fun, Optional, Optionals } from '@ephox/katamari';
 import { Attribute, Css, SugarElement, SugarNode } from '@ephox/sugar';
 
 import { getAttrValue } from '../util/CellUtils';
 
-export interface CellSpan {
+export interface RowData {
+  readonly element: SugarElement<HTMLTableRowElement | HTMLTableColElement>;
+}
+
+export interface CellData {
   readonly element: SugarElement<HTMLTableCellElement | HTMLTableColElement>;
   readonly colspan: number;
   readonly rowspan: number;
 }
 
 export interface Generators {
-  readonly cell: (cellSpan: CellSpan) => SugarElement<HTMLTableCellElement>;
-  readonly row: () => SugarElement<HTMLTableRowElement>;
+  readonly cell: (cellData: CellData) => SugarElement<HTMLTableCellElement>;
+  readonly row: (rowData: RowData) => SugarElement<HTMLTableRowElement>;
   readonly replace: <K extends keyof HTMLElementTagNameMap>(cell: SugarElement<HTMLTableCellElement>, tag: K, attrs: Record<string, string | number | boolean | null>) => SugarElement<HTMLElementTagNameMap[K]>;
   readonly gap: () => SugarElement<HTMLTableCellElement>;
-  readonly col: (prev: CellSpan) => SugarElement<HTMLTableColElement>;
-  readonly colgroup: () => SugarElement<HTMLTableColElement>;
+  readonly col: (prev: CellData) => SugarElement<HTMLTableColElement>;
+  readonly colgroup: (prev: RowData) => SugarElement<HTMLTableColElement>;
 }
 
 export interface SimpleGenerators extends Generators {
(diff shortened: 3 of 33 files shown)


modules/snooker/src/main/ts/ephox/snooker/api/Generators.ts
```diff
@@ -1,21 +1,25 @@
...
   readonly element: SugarElement<HTMLTableCellElement | HTMLTableColElement>;
   readonly colspan: number;
   readonly rowspan: number;
 }
 
 export interface Generators {
-  readonly cell: (cellSpan: CellSpan) => SugarElement<HTMLTableCellElement>;
-  readonly row: () => SugarElement<HTMLTableRowElement>;
+  readonly cell: (cellData: CellData) => SugarElement<HTMLTableCellElement>;
+  readonly row: (rowData: RowData) => SugarElement<HTMLTableRowElement>;
```
lnewson, line 18: Worth noting this row data is actually not used currently, as when cloning a row we've always just ignored any styles, classes, etc... That is likely something we'll want to revisit at some stage  (e.g. there's a regression in the GitHub backlog for classes being lost when inserting new rows) and it was a good time to include this while refactoring.

modules/tinymce/src/plugins/table/test/ts/browser/TableSectionApiTest.ts
```diff
@@ -110,11 +110,11 @@ describe('browser.tinymce.plugins.table.TableSectionApiTest', () => {
 
   const bodyMultipleChangesColumnContent = `<table>
 <tbody>
-<tr>
+<tr id="one">
```
lnewson, line 113: This is the only test here that didn't have id's on the row, so I suspect we hit this bug previously and forgot to log a Jira. For reference, this was the root issue this PR is attempting to fix as previously the id's would have been lost when changing the column type.

modules/snooker/src/main/ts/ephox/snooker/model/RunOperation.ts
```diff
@@ -74,27 +74,8 @@ export interface ExtractPasteRows {
 const fromWarehouse = (warehouse: Warehouse, generators: Generators) =>
   Transitions.toGrid(warehouse, generators, false);
 
-const deriveRows = (rendered: Structs.RowDetails[], generators: Generators) => {
```
lnewson, line 77: This was the core of the issue, as it's the part that would generate a new row whenever all the cells in a row were modified. The rest is largely just flow on effects from having to pass the row details around. It did end up making this logic much simpler :shrug:

modules/snooker/CHANGELOG.md
```diff
@@ -6,9 +6,24 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
 ## Unreleased
 
+### Added
+- `Warehouse` now contains the colgroup elements from the table.
+
+### Improved
+- Rows are now stored and tracked in the table model structures.
+
 ### Changed
+- `RowDetails` has been renamed to `RowDetail` and now takes a generic argument to define what is the type of cells stored.
```
JamesToohey, line 16:
```suggestion
- `RowDetails` has been renamed to `RowDetail` and now takes a generic argument to define the type of cells stored.
```

modules/snooker/src/main/ts/ephox/snooker/operate/ModificationOperations.ts
```diff
@@ -15,13 +19,13 @@ const insertRowAt = (grid: Structs.RowCells[], index: number, example: number, c
   const before = rows.slice(0, index);
   const after = rows.slice(index);
 
-  const between = GridRow.mapCells(rows[example], (ex, c) => {
```
JamesToohey, line 18: This is a fair bit cleaner now, wow

JamesToohey (approved):

modules/snooker/src/main/ts/ephox/snooker/api/CopyRows.ts
```diff
@@ -8,7 +8,7 @@ import * as Redraw from '../operate/Redraw';
 import { Generators } from './Generators';
 import { Warehouse } from './Warehouse';
 
-const copyRows = (table: SugarElement<HTMLTableElement>, target: TargetSelection, generators: Generators): Optional<SugarElement<HTMLTableRowElement>[]> => {
+const copyRows = (table: SugarElement<HTMLTableElement>, target: TargetSelection, generators: Generators): Optional<SugarElement<HTMLTableRowElement | HTMLTableColElement>[]> => {
```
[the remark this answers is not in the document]
lnewson: It's used in far too many places to be leaving comments unfortunately. I think this will just need to be one of those things people need to know since it's a confusing part of the HTML spec.

modules/snooker/src/main/ts/ephox/snooker/api/Generators.ts
```diff
@@ -77,18 +85,23 @@ const modification = (generators: Generators, toData = elementToData): Generator
...
-    const data = toData(element);
-    return nu(data);
+  const nuRow = (data: RowData) => {
+    switch (SugarNode.name(data.element)) {
+      case 'colgroup':
+        return generators.colgroup(data);
+      default:
+        return generators.row(data);
+    }
   };
```
diogo405, line 95: For two options I think a ternary would be better here:
```
const nuRow = (data: RowData) => {
    const name = SugarNode.name(data.element);
    return name === 'colgroup' ? generators.colgroup(data) : generators.row(data);
};
```

diogo405 (approved):

modules/snooker/src/main/ts/ephox/snooker/api/Generators.ts
lnewson: I honestly just copied this from `nuCell` above. I agree both would be better as a ternary.

modules/snooker/src/main/ts/ephox/snooker/api/Generators.ts
```diff
@@ -1,21 +1,25 @@
...
 import { Attribute, Css, SugarElement, SugarNode } from '@ephox/sugar';
 
 import { getAttrValue } from '../util/CellUtils';
 
-export interface CellSpan {
+export interface RowData {
+  readonly element: SugarElement<HTMLTableRowElement | HTMLTableColElement>;
+}
+
+export interface CellData {
```
ashfordneil, line 10: Should this rename get a changelog entry?

modules/snooker/src/main/ts/ephox/snooker/api/Generators.ts
```diff
@@ -27,9 +31,7 @@ export interface SimpleGenerators extends Generators {
   readonly colgroup: () => SugarElement<HTMLTableColElement>;
 }
 
-export interface GeneratorsWrapper {
-  readonly cursor: () => Optional<SugarElement>;
-}
+export interface GeneratorsWrapper {}
```
ashfordneil, line 34: Is this still needed, if it's empty?

modules/snooker/src/main/ts/ephox/snooker/model/GridRow.ts
```diff
@@ -3,6 +3,9 @@ import { SugarElement } from '@ephox/sugar';
 
 import * as Structs from '../api/Structs';
 
+type RowMorphism = (element: SugarElement<HTMLTableRowElement | HTMLTableColElement>) => SugarElement<HTMLTableRowElement | HTMLTableColElement>;
```
ashfordneil, line 6: Should `RowMorphism` have the `index: number` argument as well for parity?

modules/snooker/src/main/ts/ephox/snooker/model/Fitment.ts
```diff
@@ -65,12 +65,15 @@ const generateElements = (amount: number, row: Structs.RowCells, generators: Sim
   return Arr.range(amount, (idx) => Structs.elementnew(generator(), true, isLocked(idx)));
 };
 
-const rowFill = (grid: Structs.RowCells[], amount: number, generators: SimpleGenerators, lockedColumns: Record<string, boolean>): Structs.RowCells[] =>
-  grid.concat(Arr.range(amount, () => {
-    const row = grid[grid.length - 1];
+const rowFill = (grid: Structs.RowCells[], amount: number, generators: SimpleGenerators, lockedColumns: Record<string, boolean>): Structs.RowCells[] => {
+  const exampleRow = grid[grid.length - 1];
```
ashfordneil, line 69: If this is fixing a bug that's visible at the editor level, should it get a TinyMCE changelog entry?

(to me who doesn't know snooker, it _looks like_ this is where the core of the fix is for that row information being lost bug is, if not then pretend I made the comment in the right place)

modules/snooker/src/main/ts/ephox/snooker/api/Generators.ts
lnewson: It's not publicly exposed outside snooker which is why I didn't include it.

modules/snooker/src/main/ts/ephox/snooker/api/Generators.ts
lnewson: It's still required, as it's a "base" interface.

modules/snooker/src/main/ts/ephox/snooker/model/GridRow.ts
lnewson: No, as within the context of this module it'll only ever convert one row and not an array like is the case with cells. Maybe we need to rename it to be something else? I'm open to suggestions as I really didn't know what to call these 2.

modules/snooker/src/main/ts/ephox/snooker/model/Fitment.ts
lnewson: You're correct. I'll add an editor changelog, thanks!

modules/snooker/src/main/ts/ephox/snooker/api/Generators.ts
ashfordneil: Oh I see.

Out of scope, but is it a problem that something exported from `/api/*` isn't then exported from the `/api/Main`?

lnewson: I think you've misunderstood... `Generators` is exported, but only parts of it. Here's what's exported:
```
import { Generators, SimpleGenerators } from './Generators';
```

ashfordneil: Yeah no that's what I mean. My understanding was that normally for things like that we'd have `api/Generators` **only** export those two interfaces, and make a `core/Generators` or similar that has the rest of those interfaces (maybe even putting `Generators` and `SimpleGenerators` into that file and re-exporting them from api).

To be super clear, I'm not at all asking for a change here, just trying to clarify our own code style

lnewson: That understanding is incorrect. Anything in `api` just means that something from that module is used publicly, not everything.

modules/snooker/CHANGELOG.md
lnewson: Fixed

modules/snooker/src/main/ts/ephox/snooker/api/Generators.ts
lnewson: Fixed

modules/snooker/src/main/ts/ephox/snooker/model/Fitment.ts
lnewson: Fixed

modules/snooker/src/main/ts/ephox/snooker/model/GridRow.ts
ashfordneil: I'm okay with this name I think — if someone else has a suggestion though that's fair

PR #7063, opened 2021-08-02
TINY-7772: When no imagetools_proxy is provided buttons should be disabled for remote images

diogo405:
Related Ticket:

Description of Changes:
* Disable buttons - Edit, Flip h, Flip v, Rotate counterclockwise, Rotate clockwise - when setting `imagetools_proxy` is not provided and image is remote

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -62,7 +62,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Certain HTML content when inserted could cause the editor to crash #TINY-7756
 - Unbinding a native event handler inside the `remove` event caused an exception that blocked editor removal #TINY-7730
 - Disabled nested menu items could still be opened #TINY-7700
-- `Imagetools` buttons were enabled for remote images without `imagetools_proxy` set #TINY-7772
+- `imagetools` buttons were incorrectly enabled for remote images without `imagetools_proxy` set #TINY-7772
 - Some table operations would incorrectly cause table row attributes and styles to be lost #TINY-6666
 
 ### Deprecated
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -62,6 +62,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Certain HTML content when inserted could cause the editor to crash #TINY-7756
 - Unbinding a native event handler inside the `remove` event caused an exception that blocked editor removal #TINY-7730
 - Disabled nested menu items could still be opened #TINY-7700
+- `Imagetools` buttons were enabled for remote images without `imagetools_proxy` set #TINY-7772
 
 ### Deprecated
 - The `bbcode`, `fullpage`, `legacyoutput` and `spellchecker` plugins have been deprecated and marked for removal in the next major release #TINY-7260
--- a/modules/tinymce/src/plugins/imagetools/main/ts/ui/Buttons.ts
+++ b/modules/tinymce/src/plugins/imagetools/main/ts/ui/Buttons.ts
@@ -5,16 +5,18 @@
  * For commercial licenses see https://www.tiny.cloud/
  */
 
-import { Fun } from '@ephox/katamari';
+import { Fun, Obj } from '@ephox/katamari';
 
 import Editor from 'tinymce/core/api/Editor';
 import { Toolbar } from 'tinymce/core/api/ui/Ui';
 
 import * as Actions from '../core/Actions';
 
+type ButtonApiRecord = Record<string, Toolbar.ToolbarButtonInstanceApi>;
+
 const register = (editor: Editor) => {
 
-  const buttonApiMap = new Map<string, Toolbar.ToolbarButtonInstanceApi>();
+  const buttonApiRecord: ButtonApiRecord = {};
 
   const cmd = (command: string) => () => editor.execCommand(command);
 
@@ -23,14 +25,14 @@ const register = (editor: Editor) => {
       return Actions.getEditableImage(editor, element.dom).isNone();
     });
 
-    buttonApiMap.forEach((api: Toolbar.ToolbarButtonInstanceApi) => {
+    Obj.each(buttonApiRecord, (api: Toolbar.ToolbarButtonInstanceApi) => {
       api.setDisabled(disabled);
     });
   };
 
-  const addToButtonApiMap = (key: string, api: Toolbar.ToolbarButtonInstanceApi) => {
-    if (!buttonApiMap.has(key)) {
-      buttonApiMap.set(key, api);
+  const addToButtonApiRecord = (key: string, api: Toolbar.ToolbarButtonInstanceApi) => {
+    if (!Obj.has(buttonApiRecord, key)) {
+      buttonApiRecord[key] = api;
     }
   };
 
@@ -41,7 +43,7 @@ const register = (editor: Editor) => {
     icon: 'rotate-left',
     onAction: cmd('mceImageRotateLeft'),
     onSetup: (buttonApi: Toolbar.ToolbarButtonInstanceApi) => {
-      addToButtonApiMap('rotateleft', buttonApi);
+      addToButtonApiRecord('rotateleft', buttonApi);
       return Fun.noop;
     }
   });
@@ -51,7 +53,7 @@ const register = (editor: Editor) => {
     icon: 'rotate-right',
     onAction: cmd('mceImageRotateRight'),
     onSetup: (buttonApi: Toolbar.ToolbarButtonInstanceApi) => {
-      addToButtonApiMap('rotateright', buttonApi);
+      addToButtonApiRecord('rotateright', buttonApi);
       return Fun.noop;
     }
   });
@@ -61,7 +63,7 @@ const register = (editor: Editor) => {
     icon: 'flip-vertically',
     onAction: cmd('mceImageFlipVertical'),
     onSetup: (buttonApi: Toolbar.ToolbarButtonInstanceApi) => {
-      addToButtonApiMap('flipv', buttonApi);
+      addToButtonApiRecord('flipv', buttonApi);
       return Fun.noop;
     }
   });
@@ -71,7 +73,7 @@ const register = (editor: Editor) => {
     icon: 'flip-horizontally',
     onAction: cmd('mceImageFlipHorizontal'),
     onSetup: (buttonApi: Toolbar.ToolbarButtonInstanceApi) => {
-      addToButtonApiMap('fliph', buttonApi);
+      addToButtonApiRecord('fliph', buttonApi);
       return Fun.noop;
     }
   });
@@ -81,7 +83,7 @@ const register = (editor: Editor) => {
     icon: 'edit-image',
     onAction: cmd('mceEditImage'),
     onSetup: (buttonApi: Toolbar.ToolbarButtonInstanceApi) => {
-      addToButtonApiMap('editimage', buttonApi);
+      addToButtonApiRecord('editimage', buttonApi);
       return Fun.noop;
     }
   });
--- a/modules/tinymce/src/plugins/imagetools/main/ts/ui/Buttons.ts
+++ b/modules/tinymce/src/plugins/imagetools/main/ts/ui/Buttons.ts
@@ -5,6 +5,8 @@
  * For commercial licenses see https://www.tiny.cloud/
  */
 
+import { Toolbar } from '@ephox/bridge';
+
 import Editor from 'tinymce/core/api/Editor';
 
 import * as Actions from '../core/Actions';
@@ -12,7 +14,7 @@ import * as Actions from '../core/Actions';
 const register = (editor: Editor) => {
   const cmd = (command: string) => () => editor.execCommand(command);
 
-  const onSetup = (buttonApi) => {
+  const onSetup = (buttonApi: Toolbar.ToolbarButtonInstanceApi) => {
     const setDisabled = () => {
       const disabled = Actions.getSelectedImage(editor).forall((element) => {
         return Actions.getEditableImage(editor, element.dom).isNone();
--- a/modules/tinymce/src/plugins/imagetools/main/ts/ui/Buttons.ts
+++ b/modules/tinymce/src/plugins/imagetools/main/ts/ui/Buttons.ts
@@ -5,9 +5,8 @@
  * For commercial licenses see https://www.tiny.cloud/
  */
 
-import { Toolbar } from '@ephox/bridge';
-
 import Editor from 'tinymce/core/api/Editor';
+import { Toolbar } from 'tinymce/core/api/ui/Ui';
 
 import * as Actions from '../core/Actions';
 
--- a/modules/tinymce/src/plugins/imagetools/main/ts/ui/Buttons.ts
+++ b/modules/tinymce/src/plugins/imagetools/main/ts/ui/Buttons.ts
@@ -12,48 +12,54 @@ import * as Actions from '../core/Actions';
 const register = (editor: Editor) => {
   const cmd = (command: string) => () => editor.execCommand(command);
 
+  const onSetup = (buttonApi) => {
+    const setDisabled = () => {
+      const disabled = Actions.getSelectedImage(editor).forall((element) => {
+        return Actions.getEditableImage(editor, element.dom).isNone();
+      });
+      buttonApi.setDisabled(disabled);
+    };
+
+    editor.on('NodeChange', setDisabled);
+
+    return () => {
+      editor.off('NodeChange', setDisabled);
+    };
+  };
+
   editor.ui.registry.addButton('rotateleft', {
     tooltip: 'Rotate counterclockwise',
     icon: 'rotate-left',
-    onAction: cmd('mceImageRotateLeft')
+    onAction: cmd('mceImageRotateLeft'),
+    onSetup
   });
 
   editor.ui.registry.addButton('rotateright', {
     tooltip: 'Rotate clockwise',
     icon: 'rotate-right',
-    onAction: cmd('mceImageRotateRight')
+    onAction: cmd('mceImageRotateRight'),
+    onSetup
   });
 
   editor.ui.registry.addButton('flipv', {
     tooltip: 'Flip vertically',
     icon: 'flip-vertically',
-    onAction: cmd('mceImageFlipVertical')
+    onAction: cmd('mceImageFlipVertical'),
+    onSetup
   });
 
   editor.ui.registry.addButton('fliph', {
     tooltip: 'Flip horizontally',
     icon: 'flip-horizontally',
-    onAction: cmd('mceImageFlipHorizontal')
+    onAction: cmd('mceImageFlipHorizontal'),
+    onSetup
   });
 
   editor.ui.registry.addButton('editimage', {
     tooltip: 'Edit image',
     icon: 'edit-image',
     onAction: cmd('mceEditImage'),
-    onSetup: (buttonApi) => {
-      const setDisabled = () => {
-        const disabled = Actions.getSelectedImage(editor).forall((element) => {
-          return Actions.getEditableImage(editor, element.dom).isNone();
-        });
-        buttonApi.setDisabled(disabled);
-      };
-
-      editor.on('NodeChange', setDisabled);
-
-      return () => {
-        editor.off('NodeChange', setDisabled);
-      };
-    }
+    onSetup
   });
 
   editor.ui.registry.addButton('imageoptions', {
--- a/modules/tinymce/src/plugins/imagetools/main/ts/ui/Buttons.ts
+++ b/modules/tinymce/src/plugins/imagetools/main/ts/ui/Buttons.ts
@@ -5,87 +5,72 @@
  * For commercial licenses see https://www.tiny.cloud/
  */
 
-import { Fun, Obj } from '@ephox/katamari';
+import { Arr } from '@ephox/katamari';
 
 import Editor from 'tinymce/core/api/Editor';
 import { Toolbar } from 'tinymce/core/api/ui/Ui';
 
 import * as Actions from '../core/Actions';
 
-type ButtonApiRecord = Record<string, Toolbar.ToolbarButtonInstanceApi>;
-
 const register = (editor: Editor) => {
-
-  const buttonApiRecord: ButtonApiRecord = {};
+  let changeHandlers: Array<(isEditableImage: boolean) => void> = [];
 
   const cmd = (command: string) => () => editor.execCommand(command);
 
-  const setDisabled = () => {
-    const disabled = Actions.getSelectedImage(editor).forall((element) => {
-      return Actions.getEditableImage(editor, element.dom).isNone();
-    });
+  const isEditableImage = () => Actions.getSelectedImage(editor).exists((element) => {
+    return Actions.getEditableImage(editor, element.dom).isSome();
+  });
 
-    Obj.each(buttonApiRecord, (api: Toolbar.ToolbarButtonInstanceApi) => {
-      api.setDisabled(disabled);
-    });
+  const onSetup = (api: Toolbar.ToolbarButtonInstanceApi) => {
+    const handler = (isEditableImage: boolean) => api.setDisabled(!isEditableImage);
+    // Execute the handler to set the initial state
+    handler(isEditableImage());
+    // Register the handler so we can update the state when the selected node changes
+    changeHandlers = changeHandlers.concat([ handler ]);
+    return () => {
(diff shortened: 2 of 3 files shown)


modules/tinymce/CHANGELOG.md
```diff
@@ -62,6 +62,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Certain HTML content when inserted could cause the editor to crash #TINY-7756
 - Unbinding a native event handler inside the `remove` event caused an exception that blocked editor removal #TINY-7730
 - Disabled nested menu items could still be opened #TINY-7700
+- `Imagetools` buttons were enabled for remote images without `imagetools_proxy` set #TINY-7772
```
diogo405, line 65: Should we have a entry in `imagetools` changelog?

modules/tinymce/src/plugins/imagetools/main/ts/ui/Buttons.ts
```diff
@@ -12,48 +12,54 @@ import * as Actions from '../core/Actions';
 const register = (editor: Editor) => {
   const cmd = (command: string) => () => editor.execCommand(command);
 
+  const onSetup = (buttonApi) => {
```
ashfordneil, line 15: Make sure you put a type in for `buttonApi`

modules/tinymce/src/plugins/imagetools/test/ts/browser/ImageToolsDisabledButtonsTest.ts
```diff
@@ -0,0 +1,104 @@
...
+      assertImageToolsButtonsExist(editor);
+      const enabled = true;
+      assertButtonStatus(editor, editImageButtonSelector, enabled);
+      assertButtonStatus(editor, flipHorizontallyImageButtonSelector, enabled);
+      assertButtonStatus(editor, flipVerticallyImageButtonSelector, enabled);
+      assertButtonStatus(editor, rotateCounterClockwiseImageButtonSelector, enabled);
+      assertButtonStatus(editor, rotateClockwiseImageButtonSelector, enabled);
+      McEditor.remove(editor);
+    } catch (e) {
+      assert.fail(`Test failed: ${e.message}`);
```
ashfordneil, line 81: These whole try/catch blocks shouldn't be necessary. That sort of behaviour is built into the bedrock client

ashfordneil (approved):

modules/tinymce/CHANGELOG.md
ashfordneil: It doesn't look like anything has changed in the `imagetools` module, you've only edited things in the plugin. So I'd say no

modules/tinymce/src/plugins/imagetools/main/ts/ui/Buttons.ts
diogo405: FIxed.

modules/tinymce/src/plugins/imagetools/test/ts/browser/ImageToolsDisabledButtonsTest.ts
diogo405: Removed 👍

HAFRMO (approved):

modules/tinymce/CHANGELOG.md
lnewson:
```suggestion
- `imagetools` buttons were incorrectly enabled for remote images without `imagetools_proxy` set #TINY-7772
```
If we're using backticks then we need to use the actual name.

modules/tinymce/src/plugins/imagetools/main/ts/ui/Buttons.ts
```diff
@@ -5,55 +5,63 @@
...
+import { Toolbar } from '@ephox/bridge';
+
 import Editor from 'tinymce/core/api/Editor';
 
 import * as Actions from '../core/Actions';
 
 const register = (editor: Editor) => {
   const cmd = (command: string) => () => editor.execCommand(command);
 
+  const onSetup = (buttonApi: Toolbar.ToolbarButtonInstanceApi) => {
```
lnewson, line 17: This is pretty inefficient, as is has to do this lookup at least 5 times for every `NodeChange` event. Can we please look at rewriting this so the lookup is only done once and then we notify any buttons that are setup?

The best way to do that is probably going to be to register the NodeChange event once and then have an array of listeners to call with the state. Let me know if you need more details or clarification on doing that.

- [ ] Optimize this lookup so it only needs to be done once

modules/tinymce/src/plugins/imagetools/test/ts/browser/ImageToolsDisabledButtonsTest.ts
```diff
@@ -0,0 +1,96 @@
...
+    height: 900
+  };
+
+  const editImageButtonSelector = '[role="toolbar"] button[title="Edit image"]';
+  const flipHorizontallyImageButtonSelector = '[role="toolbar"] button[title="Flip horizontally"]';
+  const flipVerticallyImageButtonSelector = '[role="toolbar"] button[title="Flip vertically"]';
+  const rotateCounterClockwiseImageButtonSelector = '[role="toolbar"] button[title="Rotate counterclockwise"]';
+  const rotateClockwiseImageButtonSelector = '[role="toolbar"] button[title="Rotate clockwise"]';
+
+  const assertImageToolsButtonsExist = (editor) => {
```
lnewson, line 27: Types please

modules/tinymce/src/plugins/imagetools/test/ts/browser/ImageToolsDisabledButtonsTest.ts
```diff
@@ -0,0 +1,96 @@
...
+    UiFinder.exists(TinyDom.container(editor), editImageButtonSelector);
+    UiFinder.exists(TinyDom.container(editor), flipHorizontallyImageButtonSelector);
+    UiFinder.exists(TinyDom.container(editor), flipVerticallyImageButtonSelector);
+    UiFinder.exists(TinyDom.container(editor), rotateCounterClockwiseImageButtonSelector);
+    UiFinder.exists(TinyDom.container(editor), rotateClockwiseImageButtonSelector);
+  };
+
+  const assertButtonStatus = (editor, selector: string, enabled: boolean) => {
```
lnewson, line 35: Types here as well please.

modules/tinymce/src/plugins/imagetools/test/ts/browser/ImageToolsDisabledButtonsTest.ts
```diff
@@ -0,0 +1,96 @@
...
+  };
+
+  const editImageButtonSelector = '[role="toolbar"] button[title="Edit image"]';
+  const flipHorizontallyImageButtonSelector = '[role="toolbar"] button[title="Flip horizontally"]';
+  const flipVerticallyImageButtonSelector = '[role="toolbar"] button[title="Flip vertically"]';
+  const rotateCounterClockwiseImageButtonSelector = '[role="toolbar"] button[title="Rotate counterclockwise"]';
+  const rotateClockwiseImageButtonSelector = '[role="toolbar"] button[title="Rotate clockwise"]';
+
+  const assertImageToolsButtonsExist = (editor) => {
+    UiFinder.exists(TinyDom.container(editor), editImageButtonSelector);
```
lnewson, line 28: The container is created many times here... can we please pull this out into a variable so it's only done once?

modules/tinymce/src/plugins/imagetools/test/ts/browser/ImageToolsDisabledButtonsTest.ts
```diff
@@ -0,0 +1,96 @@
...
+    UiFinder.exists(TinyDom.container(editor), rotateClockwiseImageButtonSelector);
+  };
+
+  const assertButtonStatus = (editor, selector: string, enabled: boolean) => {
+    UiFinder.findIn(TinyDom.container(editor), selector).map((button) => {
+      assert.equal(Attribute.get(button, 'aria-disabled'), Boolean(enabled).toString());
+    });
+  };
+
+  const pLoadRemoteImage = (editor): Promise<void> => {
```
lnewson, line 41: Types here as well please.

modules/tinymce/src/plugins/imagetools/test/ts/browser/ImageToolsDisabledButtonsTest.ts
```diff
@@ -0,0 +1,96 @@
...
+    return new PromisePolyfill((resolve, reject) => {
+      const remoteImageSrc = 'https://www.google.com/images/branding/googlelogo/1x/googlelogo_color_272x92dp.png';
+      const img = new Image();
+      img.src = remoteImageSrc;
+
+      img.onload = () => {
```
lnewson, line 47: You have a potential memory leak here, as you're not clearing the handlers.

- [ ] Unbind handlers when loading is complete

modules/tinymce/src/plugins/imagetools/test/ts/browser/ImageToolsDisabledButtonsTest.ts
```diff
@@ -0,0 +1,96 @@
...
+      };
+
+      img.onerror = (e) => reject(e);
+    });
+  };
+
+  const pWaitImageToolsContextMenu = () =>
+    Waiter.pTryUntil('Wait for Image Tools context menu to open', () => UiFinder.exists(SugarBody.body(), '.tox-pop'));
+
+  before(() => {
```
lnewson, line 60: This should be above any functions, not below. So to be clear, the order should be: variables -> hooks -> helper functions -> tests.

- [ ] Move before hook to correct place

modules/tinymce/src/plugins/imagetools/test/ts/browser/ImageToolsDisabledButtonsTest.ts
lnewson: Actually... why are we even doing this? It serves no purpose and we could just be inserting the image directly without trying to load it first.

modules/tinymce/src/plugins/imagetools/test/ts/browser/ImageToolsDisabledButtonsTest.ts
```diff
@@ -0,0 +1,96 @@
...
+        editor.setContent(`<p><img src="${remoteImageSrc}" width="50" height="50" /></p>`);
+        editor.focus();
+        resolve();
+      };
+
+      img.onerror = (e) => reject(e);
+    });
+  };
+
+  const pWaitImageToolsContextMenu = () =>
```
lnewson, line 57: This is incorrectly named, this is waiting for the context toolbar, not the context menu.

modules/tinymce/src/plugins/imagetools/test/ts/browser/ImageToolsDisabledButtonsTest.ts
```diff
@@ -0,0 +1,96 @@
...
+    const enabled = true;
+    assertButtonStatus(editor, editImageButtonSelector, enabled);
+    assertButtonStatus(editor, flipHorizontallyImageButtonSelector, enabled);
+    assertButtonStatus(editor, flipVerticallyImageButtonSelector, enabled);
+    assertButtonStatus(editor, rotateCounterClockwiseImageButtonSelector, enabled);
+    assertButtonStatus(editor, rotateClockwiseImageButtonSelector, enabled);
+    McEditor.remove(editor);
+  });
+
+  it('TINY-7772: Edit, flip and rotate should be enabled for remote images with imagetools_proxy set', async () => {
```
lnewson, line 81: We're missing one test case here, as we need to test these are disabled when no image is selected as well.

- [ ] Add missing test case

lnewson (approved):

modules/tinymce/src/plugins/imagetools/main/ts/ui/Buttons.ts
```diff
@@ -5,55 +5,63 @@
  * For commercial licenses see https://www.tiny.cloud/
  */
 
+import { Toolbar } from '@ephox/bridge';
```
lnewson, line 8: This is using the wrong type. It should be using the `Toolbar` type from `tinymce/core/api/ui/Ui` instead.

modules/tinymce/src/plugins/imagetools/test/ts/browser/ImageToolsDisabledButtonsTest.ts
JamesToohey: I don't think we need this helper function, the buttons can be assumed to be present because of the editor config (and we assert their enabled/disabled state anyway, which also asserts they are present).

modules/tinymce/src/plugins/imagetools/test/ts/browser/ImageToolsDisabledButtonsTest.ts
```diff
@@ -0,0 +1,96 @@
...
+  const assertImageToolsButtonsExist = (editor) => {
+    UiFinder.exists(TinyDom.container(editor), editImageButtonSelector);
+    UiFinder.exists(TinyDom.container(editor), flipHorizontallyImageButtonSelector);
+    UiFinder.exists(TinyDom.container(editor), flipVerticallyImageButtonSelector);
+    UiFinder.exists(TinyDom.container(editor), rotateCounterClockwiseImageButtonSelector);
+    UiFinder.exists(TinyDom.container(editor), rotateClockwiseImageButtonSelector);
+  };
+
+  const assertButtonStatus = (editor, selector: string, enabled: boolean) => {
+    UiFinder.findIn(TinyDom.container(editor), selector).map((button) => {
```
lnewson, line 36: This should just be a `getOrDie`, not a `map` as otherwise the assertion may not be working. Even better, this should be using `UiFinder.exists` or similar as we don't need to lookup the `aria-disabled` attribute and instead can use a selector.

- [ ] Use appropriate `UiFinder` assertion here instead

modules/tinymce/src/plugins/imagetools/test/ts/browser/ImageToolsDisabledButtonsTest.ts
lnewson: Good point and if they aren't visible then the assertions to check the state should fail anyways.

- [ ] Remove this assertion as it's unnecessary

modules/tinymce/src/plugins/imagetools/main/ts/ui/Buttons.ts
diogo405: I didn't write this part, I just moved what we already had. I'm happy to refactor but shouldn't it be another task?

lnewson: No, because previously it was only used for a single button so the logic was okay. You're now trying to use that same logic across multiple buttons.

modules/tinymce/src/plugins/imagetools/test/ts/browser/ImageToolsDisabledButtonsTest.ts
```diff
@@ -0,0 +1,74 @@
...
+    editor.setContent(`<p><img src="${remoteImageSrc}" width="50" height="50" /></p>`);
+  };
+
+  it('TINY-7772: Edit, flip and rotate should be disabled for remote images without imagetools_proxy set', async () => {
+    const editor = await McEditor.pFromSettings<Editor>(settings);
+    insertRemoteImage(editor);
+    Mouse.clickOn(TinyDom.body(editor), 'img');
+
+    const disabled = true;
+    assertButtonsStatus(editor, disabled);
```
lnewson, line 51: You don't need this variable any more as you've moved the logic out into a separate function
```suggestion
    assertButtonsStatus(editor, true);
```

modules/tinymce/src/plugins/imagetools/test/ts/browser/ImageToolsDisabledButtonsTest.ts
```diff
@@ -0,0 +1,74 @@
...
+  };
+
+  const insertRemoteImage = (editor: Editor) => {
+    const remoteImageSrc = 'https://www.google.com/images/branding/googlelogo/1x/googlelogo_color_272x92dp.png';
+    editor.setContent(`<p><img src="${remoteImageSrc}" width="50" height="50" /></p>`);
+  };
+
+  it('TINY-7772: Edit, flip and rotate should be disabled for remote images without imagetools_proxy set', async () => {
+    const editor = await McEditor.pFromSettings<Editor>(settings);
+    insertRemoteImage(editor);
```
lnewson, line 47: This could be done in a `beforeEach` as it's setup logic.

modules/tinymce/src/plugins/imagetools/test/ts/browser/ImageToolsDisabledButtonsTest.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  });
+
+  it('TINY-7772: Edit, flip and rotate should be enabled for remote images with imagetools_proxy set', async () => {
+    const editor = await McEditor.pFromSettings<Editor>({ ...settings, imagetools_proxy: 'foo.php' });
+    insertRemoteImage(editor);
+    Mouse.clickOn(TinyDom.body(editor), 'img');
+
+    const disabled = false;
+    assertButtonsStatus(editor, disabled);
```
lnewson, line 62:
```suggestion
    assertButtonsStatus(editor, false);
```

modules/tinymce/src/plugins/imagetools/test/ts/browser/ImageToolsDisabledButtonsTest.ts
```diff
@@ -0,0 +1,74 @@
...
+    McEditor.remove(editor);
+  });
+
+  it('TINY-7772: Edit, flip and rotate should be disabled when no images are selected', async () => {
+    const editor = await McEditor.pFromSettings<Editor>(settings);
+    insertRemoteImage(editor);
+
+    const disabled = true;
+    assertButtonsStatus(editor, disabled);
```
lnewson, line 71:
```suggestion
    assertButtonsStatus(editor, true);
```

modules/tinymce/src/plugins/imagetools/test/ts/browser/ImageToolsDisabledButtonsTest.ts
diogo405: I use the variable to be easier to understand. When you pass `true` it doesn't mean much...

modules/tinymce/src/plugins/imagetools/test/ts/browser/ImageToolsDisabledButtonsTest.ts
diogo405: There's one test that uses different `settings`:
`const editor = await McEditor.pFromSettings<Editor>({ ...settings, imagetools_proxy: 'foo.php' });`

If they had the same settings how I'd remove the `editor` at the end of each test?

lnewson: I didn't comment on that line? I commented on the part about setting the initial content so it contains the remote image.

lnewson: Oh wait I see what you mean. You're right we don't have an editor instance at that point so we can't. In that case ignore this sorry.

modules/tinymce/src/plugins/imagetools/test/ts/browser/ImageToolsDisabledButtonsTest.ts
lnewson: I understand, but this isn't something we do anywhere else and I'd like us to be consistent. If you're worried about readability then we'd normally pass a "spec" (object) instead so you can do something like `assertButtonStatus(editor, { disabled: true })`

diogo405: Yeah, I was going to ask that `{disabled: boolean}`. Great, I'm changing 👍

modules/tinymce/src/plugins/imagetools/main/ts/ui/Buttons.ts
diogo405: I tried to make the lookup more efficient using a map.

modules/tinymce/src/plugins/imagetools/main/ts/ui/Buttons.ts
```diff
@@ -5,62 +5,85 @@
...
-      });
-      buttonApi.setDisabled(disabled);
-    };
+  const setDisabled = () => {
+    const disabled = Actions.getSelectedImage(editor).forall((element) => {
+      return Actions.getEditableImage(editor, element.dom).isNone();
+    });
 
-    editor.on('NodeChange', setDisabled);
+    buttonApiMap.forEach((api: Toolbar.ToolbarButtonInstanceApi) => {
```
diogo405, line 26: Couldn't figure out how to use `Obj.map` here.

modules/tinymce/src/plugins/imagetools/main/ts/ui/Buttons.ts
```diff
@@ -5,62 +5,85 @@
...
+      api.setDisabled(disabled);
+    });
+  };
 
-    return () => {
-      editor.off('NodeChange', setDisabled);
-    };
+  const addToButtonApiMap = (key: string, api: Toolbar.ToolbarButtonInstanceApi) => {
+    if (!buttonApiMap.has(key)) {
```
diogo405, line 32: Same, couldn't use `Obj.has(buttonApiMap, key)`, might be missing something...

modules/tinymce/src/plugins/imagetools/main/ts/ui/Buttons.ts
```diff
@@ -5,62 +5,85 @@
...
+      buttonApiMap.set(key, api);
+    }
   };
 
+  editor.on('NodeChange', setDisabled);
```
diogo405, line 37: Should we have an `editor.off()` somewhere?! If I add it here it doesn't trigger the `NodeChange` at all.

lnewson (changes requested):
I'm sorry but this change isn't going to work... you have the following issues:

1. You aren't cleaning up the handlers when the button is destroyed (ie has a leak)
2. Your logic won't work if the button is rendered multiple times (eg in the toolbar and in the context toolbar).
3. This was also an issue in the original version, but it doesn't set the initial button state on setup.

I've gone ahead and created a branch (see https://github.com/tinymce/tinymce/compare/feature/TINY-7772...feature/TINY-7772_2) to show what I was expecting (this follows the design of `SelectionTargets` like I mentioned via DMs). Feel free to just merge/cherry-pick this into your branch.

modules/tinymce/src/plugins/imagetools/main/ts/ui/Buttons.ts
lnewson: It should be fine to not include that, as since it's an editor managed event it will be cleaned up when the editor is removed.

diogo405:
> I'm sorry but this change isn't going to work... you have the following issues:
> 
> 1. You aren't cleaning up the handlers when the button is destroyed (ie has a leak).
> 2. Your logic won't work if the button is rendered multiple times (eg in the toolbar and in the context toolbar).
> 3. This was also an issue in the original version, but it doesn't set the initial button state on setup.
> 
> I've gone ahead and created a branch (see [feature/TINY-7772...feature/TINY-7772_2](https://github.com/tinymce/tinymce/compare/feature/TINY-7772...feature/TINY-7772_2)) to show what I was expecting (this follows the design of `SelectionTargets` like I mentioned via DMs). Feel free to just merge/cherry-pick this into your branch if you want.

Merged.

lnewson (approved):

modules/tinymce/src/plugins/imagetools/test/ts/browser/ImageToolsDisabledButtonsTest.ts
```diff
@@ -0,0 +1,71 @@
...
+    McEditor.remove(editor);
+  });
+
+  it('TINY-7772: Edit, flip and rotate should be disabled when no images are selected', async () => {
+    const editor = await McEditor.pFromSettings<Editor>(settings);
+    insertRemoteImage(editor);
+
+    assertButtonsStatus(editor, { disabled: true });
+    McEditor.remove(editor);
+  });
```
JamesToohey, line 70: - [ ] This test should have an `imagetools_proxy` set otherwise it could be `disabled` for either reason

JamesToohey (changes requested):
Requesting changes just for one test tweak

JamesToohey (approved):

ashfordneil (approved):

PR #7074, opened 2021-08-04
TINY-7739: Fix inset layouts incorrectly used for selection context toolbars

lnewson:
Related Ticket: TINY-7739

Description of Changes:
So there were 3 parts to this in the end:

1. There was an issue with using the wrong anchor position lookup when trying to see if the toolbar should hide. This was because it was using the element position (e.g. the `p` element) for selection toolbars which isn't the same as the selection and in most cases was off by 1px. That off by 1 is what caused the "north" layout to say it didn't fit as part of it was deemed to be offscreen.
2. The was an issue with the overlap detection logic as if it was the same value (e.g. the anchor was `y` 400 and the `bottom` bounds was 400) than nothing was visible so it should have been hidden. It also had an issue when one of the bounds had a height of zero.
3. I've also improved the context toolbar layout decisions to not use an inset layout for a context toolbar placed on a selection. We don't want that as it'd cover the selection at all times and we wouldn't be able to see the selection content like we can with say tables.

Pre-checks:
* [x] ~Changelog entry added~ No changelog as this was caused by changes in 5.9

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/themes/silver/main/ts/ui/context/ContextToolbarBounds.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/ui/context/ContextToolbarBounds.ts
@@ -15,8 +15,9 @@ import Editor from 'tinymce/core/api/Editor';
 import * as Settings from '../../api/Settings';
 import { UiFactoryBackstageShared } from '../../backstage/Backstage';
 
+// Note: We round the values here as we are looking for changes that aren't just 0.001px difference
 const isVerticalOverlap = (a: Bounds, b: Bounds): boolean =>
-  a.y < b.bottom && a.bottom > b.y;
+  Math.round(a.y) < Math.round(b.bottom) && Math.round(a.bottom) > Math.round(b.y);
 
 const getRangeRect = (rng: Range): DOMRect => {
   const rect = rng.getBoundingClientRect();
--- a/modules/tinymce/src/themes/silver/main/ts/ui/context/ContextToolbarBounds.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/ui/context/ContextToolbarBounds.ts
@@ -16,7 +16,7 @@ import * as Settings from '../../api/Settings';
 import { UiFactoryBackstageShared } from '../../backstage/Backstage';
 
 const isVerticalOverlap = (a: Bounds, b: Bounds): boolean =>
-  Math.max(a.y, b.y) <= Math.min(a.bottom, b.bottom);
+  a.y < b.bottom && a.bottom > b.y;
 
 const getRangeRect = (rng: Range): DOMRect => {
   const rect = rng.getBoundingClientRect();
--- a/modules/tinymce/src/themes/silver/main/ts/ui/context/ContextToolbarBounds.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/ui/context/ContextToolbarBounds.ts
@@ -15,9 +15,9 @@ import Editor from 'tinymce/core/api/Editor';
 import * as Settings from '../../api/Settings';
 import { UiFactoryBackstageShared } from '../../backstage/Backstage';
 
-// Note: We round the values here as we are looking for changes that aren't just 0.001px difference
+// Note: We want to avoid a small difference such as 0.001px so we ensure at least 1px is visible
 const isVerticalOverlap = (a: Bounds, b: Bounds): boolean =>
-  Math.round(a.y) < Math.round(b.bottom) && Math.round(a.bottom) > Math.round(b.y);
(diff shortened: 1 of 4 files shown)


JamesToohey (approved):

ltrouton (approved):

modules/tinymce/src/themes/silver/main/ts/ui/context/ContextToolbarBounds.ts
```diff
@@ -15,8 +15,9 @@ import Editor from 'tinymce/core/api/Editor';
 import * as Settings from '../../api/Settings';
 import { UiFactoryBackstageShared } from '../../backstage/Backstage';
 
+// Note: We round the values here as we are looking for changes that aren't just 0.001px difference
 const isVerticalOverlap = (a: Bounds, b: Bounds): boolean =>
-  Math.max(a.y, b.y) <= Math.min(a.bottom, b.bottom);
+  Math.round(a.y) < Math.round(b.bottom) && Math.round(a.bottom) > Math.round(b.y);
```
ashfordneil, line 20: This is still going to break if for example `a.y = 0.49` and `b.bottom = 0.51`

Could we try something like
```suggestion
  b.bottom - a.y >= 1.0 && a.bottom - b.y >= 1.0
```

Alternatively, could we do a rounding / number normalising process higher up the call tree, and make sure it applies to everything. That way `0.49` and `0.51` isn't a problem, because they're treated as `0` and `1` everywhere (even on-screen)

modules/tinymce/src/themes/silver/test/ts/browser/editor/contexttoolbar/ContextToolbarIframePositionTest.ts
```diff
@@ -37,7 +39,7 @@ describe('browser.tinymce.themes.silver.editor.contexttoolbar.ContextToolbarIFra
     toolbar: 'fullscreen',
     height: 400,
     base_url: '/project/tinymce/js/tinymce',
-    content_style: 'body, p { margin: 0; } tr { height: 36px; }',
+    content_style: 'body, p { margin: 0; line-height: 22px; font-family: sans-serif; } tr { height: 36px; }',
```
ashfordneil, line 42: Why was the font-family changed?

ashfordneil (approved):

modules/tinymce/src/themes/silver/test/ts/browser/editor/contexttoolbar/ContextToolbarIframePositionTest.ts
lnewson: I had mentioned this in stand up, but it's because the default font family tries to use "System default" which varies between browsers. As such, it causes the selection heights to be different and causes differing behaviour so we need to make sure we're using the same font on all OSs.

modules/tinymce/src/themes/silver/main/ts/ui/context/ContextToolbarBounds.ts
lnewson: Yep, that we're fine with as discussed in calls/slack. We just don't want it to be picking up minimal changes such as `0.001px`.

ashfordneil: When I gave the thumbs up I was thinking more about the behaviour of not counting minimal changes, and not thinking so much about the particular `round` implementation. I still agree that we want to do something to avoid minimal changes, I just worry that in some cases this code won't be able to filter out minimal changes.

I don't know where we source the numbers for the two bounding boxes we're comparing, so this might be super unlikely to happen in real life, but the issue with comparing things that have come out of a `round` function is that there's always going to be a threshold where a small difference gets treated like a large difference. For example, if the boxes have a tiny overlap, but one of them is `1005.499999` and the other is `1005.500001 `, we might still end up accidentally returning true here, because one value will round down and the other will round up.

Maybe it'd be worth changing the `>= 1.0` to a different threshold, if we have a clear idea of where the line is between "this is too small a change to consider" and "this is a change worth considering as a real overlap", but I do think we need to adjust this to look at the difference between the two numbers instead of rounding them.

lnewson: Yeah fair point... okay I'll change it to ensure we meet at least the full 1px visible threshold. For reference the numbers just come from the browsers `getBoundingClientRect()` function. I won't be using `1.0` though as it can just be `1` in this case.

lnewson: Changed as discussed.

PR #7079, opened 2021-08-05
TINY-7695: Fixed table operations placing the cursor in an invalid position

lnewson:
Related Ticket: TINY-7695

Description of Changes:
* Ensure that when erasing columns/rows we try to ensure the cursor position used is within the grid. This means that when deleting the last row/column it'll move to the previous row/column instead of the first cell in the table.
* Update snooker to attempt to find a different cursor position that's editable if the one provided by the operation isn't valid/editable.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/snooker/src/main/ts/ephox/snooker/api/TableOperations.ts
+++ b/modules/snooker/src/main/ts/ephox/snooker/api/TableOperations.ts
@@ -1,7 +1,7 @@
-import { Arr, Fun, Optional } from '@ephox/katamari';
+import { Arr, Fun, Optional, Optionals } from '@ephox/katamari';
 import { Remove, SugarElement, Width } from '@ephox/sugar';
-import { TableSection } from 'src/main/ts/ephox/snooker/api/TableSection';
 
+import { TableSection } from '../api/TableSection';
 import * as Blocks from '../lookup/Blocks';
 import { findCommonCellType, findCommonRowType } from '../lookup/Type';
 import * as DetailsList from '../model/DetailsList';
@@ -42,7 +42,10 @@ interface ExtractColsDetail {
 
 type CompElm = (e1: SugarElement, e2: SugarElement) => boolean;
 
-const prune = (table: SugarElement) => {
+const isEditable = (elem: SugarElement<HTMLElement>) =>
+  elem.dom.contentEditable !== 'false';
+
+const prune = (table: SugarElement<HTMLTableElement>) => {
   const cells = TableLookup.cells(table);
   if (cells.length === 0) {
     Remove.remove(table);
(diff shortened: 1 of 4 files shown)


diogo405 (approved):

modules/snooker/src/main/ts/ephox/snooker/api/TableOperations.ts
```diff
@@ -54,23 +57,25 @@ const outcome = (grid: Structs.RowCells[], cursor: Optional<SugarElement>): Tabl
...
 
+const findEditableCursorPosition = (rows: Structs.RowCells[]) =>
+  Arr.findMap(rows, (row) =>
+    Arr.findMap(row.cells, (cell) => {
+      const elem = cell.element;
+      return Optionals.someIf(isEditable(elem), elem);
+    })
+  );
+
 const elementFromGrid = (grid: Structs.RowCells[], row: number, column: number) => {
```
metricjs, line 68: Is this function name clear enough now, regarding what this should be used for?

modules/snooker/src/test/ts/browser/EraseOperationsTest.ts
```diff
@@ -580,7 +580,7 @@ UnitTest.test('EraseOperationsTest', () => {
   };
   Assertions.checkDelete(
     'TBA',
-    Optional.some({ section: 0, row: 0, column: 0 }),
+    Optional.some({ section: 0, row: 2, column: 0 }),
```
metricjs, line 583: Sanity check: based on the two points in your description, this tests the first (going to previous instead of first cell), and the test cases further down deal with the second point (cef cells)?

If yes, do we care to add a deleting rows case too for the first point? Or is that already covered in another case? I looked up and down a bit, but can't see a specifically relevant case for it.

metricjs (approved):

modules/snooker/src/test/ts/browser/EraseOperationsTest.ts
lnewson: It's covered in another case (you can see I had to change one test from row 0 to row 2).

lnewson: Discussed in a call and clarified. I'll make sure we have a case for both row/column deletion.

lnewson: Have added the extra test case

ltrouton (approved):

modules/snooker/src/main/ts/ephox/snooker/api/TableOperations.ts
```diff
@@ -54,23 +57,25 @@ const outcome = (grid: Structs.RowCells[], cursor: Optional<SugarElement>): Tabl
...
   const rows = GridRow.extractGridDetails(grid).rows;
-  return findIn(rows, row, column).orThunk(() =>
-    findIn(rows, 0, 0)
-  );
+  return Optional.from(rows[row]?.cells[column]?.element)
+    .filter(isEditable)
+    // Fallback to the first valid position in the table
+    .orThunk(() => findEditableCursorPosition(rows));
```
ashfordneil, line 73: Do we normally have to put more heuristics in to try to put the selection somewhere _close_ to where it was previously, instead of just in the first content editable cell?

ashfordneil (approved):

modules/snooker/src/main/ts/ephox/snooker/api/TableOperations.ts
lnewson: This will try and place it close to where it was previously (which is what row and column is here). However, after that it's always just defaulted to the first viable position. Changing that is outside the scope of this PR so if you'd like to see that changed discuss it with Product/log a Jira please.

modules/snooker/src/main/ts/ephox/snooker/api/TableOperations.ts
lnewson: No one else has commented regarding this and it's still not any different to what it was previously so marking as resolved.

PR #7089, opened 2021-08-09
TINY-7756: Fix DomParser not checking validity of children before unwrapping nodes

JamesToohey:
Related Ticket: TINY-7756

Description of Changes:
* Add a new `.children()` method to AstNode
* Tidy up some Docstrings for clarity
* Ensure that when we unwrap nodes that are invalid children of other nodes, we also check that the resulting 'unwrapped' nodes will be valid children. If they are not, we need to treat them the same way

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/tinymce/src/core/test/ts/browser/content/InsertContentTest.ts
```diff
@@ -539,4 +539,37 @@ describe('browser.tinymce.core.content.InsertContentTest', () => {
...
+      editor,
+      '<table><button><a><meta>'
+    );
+    TinyAssertions.assertContent(editor, '');
+
+    editor.setContent('');
+    TinySelections.setCursor(editor, [ 0 ], 0);
+    InsertContent.insertAtCaret(
+      editor,
+      '<table><tbody><tr><td><meta><button><img/><button><a><meta/></a></button><img/></button>'
```
JamesToohey, line 557: - [ ] link to new Jira next to these tests with an explanation

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
```diff
@@ -168,9 +182,17 @@ const DomParser = (settings?: DomParserSettings, schema = Schema()): DomParser =
...
           parent.insert(newParent, parents[0], true);
-          parent.insert(node, newParent);
+          if (schema.isValidChild(parent.name, node.name)) {
+            parent.insert(node, newParent);
+          } else {
+            unwrapInvalidNode(node);
+          }
         } else {
-          parent.insert(node, parents[0], true);
+          if (schema.isValidChild(parent.name, node.name)) {
```
lnewson, line 191: This doesn't look correct? I believe it should be `parents[0]` instead as that's where it's attempting to insert it into.

modules/tinymce/CHANGELOG.md
```diff
@@ -70,6 +71,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The selection was incorrectly lost when using the `mceTableCellType` and `mceTableRowType` commands #TINY-6666
 - The `mceTableRowType` was reversing the order of the rows when converting multiple header rows back to body rows #TINY-6666
 - The nested menu item chevron icon was not fading when the menu item was disabled #TINY-7700
+- Inserting certain HTML content into the editor could result in invalid HTML once parsed #TINY-7756
```
lnewson, line 74: Maybe move this beside the existing `TINY-7756` (ie on line 65)

modules/tinymce/src/core/test/ts/browser/html/NodeTest.ts
```diff
@@ -394,4 +394,48 @@ describe('browser.tinymce.core.html.NodeTest', () => {
...
+      assert.lengthOf(root.children(), 0);
+    });
+
+    it('TINY-7756: Returns single child', () => {
+      const root = new AstNode('p', 1);
+      const child = root.append(new AstNode('span', 1));
+
+      const children = root.children();
+      assert.lengthOf(children, 1);
+      assert.deepEqual(children[0], child);
```
lnewson, line 410: Should this and the below cases be `strictEquals` instead of `deepEquals`? As we aren't checking a nested object and instead want to make sure this is the same reference.

lnewson:
Not approving as I've had a little too much to do with this during discussions, etc... I've left comments on somethings, but otherwise this looks good to me.

ashfordneil (commented):
The things Lee has said all seem like they could be real bugs, so I don't want to approve until you've fixed or replied to those.

Also, would it be possible to add closing tags to the HTML you're using in these tests? It's already invalid HTML by design which makes it pretty hard to reason about, but omitting end tags just adds to the confusion when trying to step through it. Of course, if the omission of the end tags is part of what you're testing then never mind.

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
JamesToohey: We are calling `.insert` on `parent` not on `parents[0]`, `parents[0]` is the `refNode` used by `insert` (i.e. the sibling we will be placing `node` next to). 

```
<parent>
  <------------ // where we want to put `<node></node>` (before `refNode` because we pass `true` as the 3rd argument)
  <parents[0]> // refNode
  </parents[0]>
</parent>
```

So `node` needs to be a valid child of `parent`

lnewson: Ahh right, that damn API always confuses me 😔

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
```diff
@@ -141,42 +155,48 @@ const DomParser = (settings?: DomParserSettings, schema = Schema()): DomParser =
...
-        // Start cloning and moving children on the left side of the target node
-        let currentNode = newParent;
-        for (let i = 0; i < parents.length - 1; i++) {
-          if (schema.isValidChild(currentNode.name, parents[i].name)) {
-            tempNode = filterNode(parents[i].clone());
-            currentNode.append(tempNode);
-          } else {
-            tempNode = currentNode;
-          }
+        // If the node won't be a valid child of the parent then just unwrap
```
metricjs, line 158: Just re: understanding this, it lost me for a second that this comment is more about the else case. Could we word it more like "If it's a valid child then try to move it, else just unwrap it"? The comment reflects the logic a bit better that way.

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
```diff
@@ -101,6 +101,20 @@ const DomParser = (settings?: DomParserSettings, schema = Schema()): DomParser =
     const textBlockElements = schema.getTextBlockElements();
     const specialElements = schema.getSpecialElements();
 
+    const unwrapInvalidNode = (node: AstNode, originalNodeParent: AstNode = node.parent): void => {
+      if (specialElements[node.name]) {
```
metricjs, line 105: Do we care that this is called "unwrap" but does both unwrap and remove depending?

modules/tinymce/src/core/main/ts/api/html/Node.ts
```diff
@@ -433,11 +433,11 @@ class AstNode {
   }
 
   /**
-   * Get all children by name.
+   * Get all descendents by name.
```
metricjs, line 436:
```suggestion
   * Get all descendants by name.
```

modules/tinymce/src/core/main/ts/api/html/Node.ts
```diff
@@ -433,11 +433,11 @@ class AstNode {
...
    *
    * @method getAll
-   * @param {String} name Name of the child nodes to collect.
-   * @return {Array} Array with child nodes matchin the specified name.
+   * @param {String} name Name of the descendent nodes to collect.
```
metricjs, line 439:
```suggestion
   * @param {String} name Name of the descendant nodes to collect.
```

modules/tinymce/src/core/main/ts/api/html/Node.ts
```diff
@@ -433,11 +433,11 @@ class AstNode {
...
+   * @return {Array} Array with descendent nodes matching the specified name.
```
metricjs, line 440:
```suggestion
   * @return {Array} Array with descendant nodes matching the specified name.
```

modules/tinymce/src/core/main/ts/api/html/Node.ts
```diff
@@ -452,6 +452,23 @@ class AstNode {
     return collection;
   }
 
+  /**
+   * Get all children of the current node.
```
metricjs, line 456:
```suggestion
   * Get all children of this node.
```
We changed current to this above, so I assume we should do the same here?

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
metricjs: Maybe add a comment for remove case to match the unwrap case? Idk

metricjs (approved):

modules/tinymce/src/core/test/ts/browser/html/NodeTest.ts
```diff
@@ -394,4 +394,48 @@ describe('browser.tinymce.core.html.NodeTest', () => {
...
+      const child3 = root.append(new AstNode('em', 1));
+
+      const children = root.children();
+      assert.lengthOf(children, 3);
+      assert.strictEqual(children[0], child1);
+      assert.strictEqual(children[1], child2);
+      assert.strictEqual(children[2], child3);
+    });
+
+    it('TINY-7756: Does not return descendents', () => {
```
JamesToohey, line 426:
```suggestion
    it('TINY-7756: Does not return descendants', () => {
```

ashfordneil (approved):

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
JamesToohey: `removeOrUnwrapInvalidNode`?

metricjs: If we don't find that too wordy, I'd rather that, yes please.

JamesToohey: I think it's okay

PR #7090, opened 2021-08-09
TINY-7485: Fixed pasting into the wrong cell when multiple cells were selected

lnewson:
Related Ticket: TINY-7485

Description of Changes:
The paste logic incorrectly looked up the cell selection as it only ever considered the native selection, not the fake selection. So this swaps it over to use the proper way to lookup the selection in for the tables plugin. I also cleaned up a little by using the `SugarNode.isTag` function.

Additionally, I discovered that `CellOpSelection` was returning a `caption` element in some cases which was incorrect, so I've corrected the types in Darwin and feed that through. This hadn't caused an issue previously thankfully because the only place it's used is for `TableSelection` and `RunOperation`. The former had an additional guard to workaround this bug and the latter does `TableLookup.cell` on the single element passed, which in turn returns `Optional.none`.

Note: To use the table clipboard functionality the paste plugin is required, so I've also added that to the `ClipboardTest`.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/plugins/table/test/ts/browser/ClipboardTest.ts
+++ b/modules/tinymce/src/plugins/table/test/ts/browser/ClipboardTest.ts
@@ -1,21 +1,23 @@
+import { Clipboard } from '@ephox/agar';
 import { describe, it } from '@ephox/bedrock-client';
-import { LegacyUnit, TinyHooks } from '@ephox/mcagar';
+import { LegacyUnit, TinyAssertions, TinyDom, TinyHooks, TinySelections } from '@ephox/mcagar';
 import { assert } from 'chai';
 
 import Editor from 'tinymce/core/api/Editor';
 import Tools from 'tinymce/core/api/util/Tools';
-import Plugin from 'tinymce/plugins/table/Plugin';
+import PastePlugin from 'tinymce/plugins/paste/Plugin';
+import TablePlugin from 'tinymce/plugins/table/Plugin';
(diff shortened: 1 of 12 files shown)


modules/tinymce/src/plugins/table/test/ts/browser/ClipboardTest.ts
```diff
@@ -678,4 +680,56 @@ describe('browser.tinymce.plugins.table.ClipboardTest', () => {
       '</table>'
     );
   });
+
+  it('TINY-7485: Pasting into a table with a single cells contents selected', () => {
```
diogo405, line 684: This test label seems wrong. Maybe `Pasting into a table with single cells contents selected` (without `a`)?!

diogo405 (approved):

modules/tinymce/src/plugins/table/test/ts/browser/ClipboardTest.ts
lnewson: The a is correct

JamesToohey: This might help

```suggestion
  it('TINY-7485: Pasting into a table with a single cell's contents selected', () => {
```

ltrouton (approved):

ashfordneil (approved):

PR #7099, opened 2021-08-10
TINY-7713: Add variable support to formatchanged

ashfordneil:
Related Ticket: TINY-7713

Description of Changes:
* Fix `formatChanged` incorrectly handling multiple listeners on the same format where some were `similar` and others weren't
* Fix `formatChanged` incorrectly assuming that the initial state of any format would be `false` when first registered
* Let `formatChanged` take variables as arguments, and filter down on them
* Basically rewrote all of `FormatChanged.ts`

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/tinymce/src/core/test/ts/browser/fmt/FormatChangeVarsTest.ts
```diff
@@ -0,0 +1,194 @@
...
+import { Type } from '@ephox/katamari';
+import { TinyHooks, TinySelections } from '@ephox/mcagar';
+import { assert } from 'chai';
+
+import Editor from 'tinymce/core/api/Editor';
+
+describe('browser.tinymce.core.fmt.FormatChangeVarsTest', () => {
+  const hook = TinyHooks.bddSetupLight<Editor>({
+    base_url: '/project/tinymce/js/tinymce'
+  }, []);
```
lnewson, line 12: You've forgotten to import and pass the `Theme` here btw.

modules/tinymce/src/core/test/ts/browser/fmt/FormatChangeVarsTest.ts
```diff
@@ -0,0 +1,194 @@
...
+    events.helvetica = [];
+    events.comicSans = [];
+  };
+
+  const helveticaFont = 'helvetica, arial, sans-serif';
+  const comicSansFont = '\'comic sans ms\', sans-serif';
+
+  before(() => {
+    const editor = hook.editor();
+    const general = editor.formatter.formatChanged('fontname', (evt) => events.general.push(evt), true);
```
lnewson, line 33: This would be better suited to using `Singleton.unbindable`.

modules/tinymce/src/core/test/ts/browser/fmt/FormatChangeVarsTest.ts
ashfordneil: Thanks, fixed now

modules/tinymce/src/core/main/ts/fmt/FormatChanged.ts
```diff
@@ -5,101 +5,167 @@
...
+    (node) => { // Can we guarantee that no parent of this will be the node we want?
+      if (MatchFormat.matchesUnInheritedFormatSelector(editor, node, format)) {
+        return true;
+      } else {
+        if (!similar) {
+          // If we want to find an exact match, then finding a similar match halfway up the parents tree is bad
+          return Type.isNonNullable(editor.formatter.matchNode(node, format, vars, true));
+        } else {
+          return false;
+        }
```
diogo405, line 65: I'd have a ternary here:
```
return similar ? 
    false : 
    Type.isNonNullable(editor.formatter.matchNode(node, format, vars, true));
```

modules/tinymce/src/core/main/ts/fmt/FormatChanged.ts
```diff
@@ -5,101 +5,167 @@
...
+  );
+};
+
+const hasChanged = (state: Cell<boolean>, newState: boolean): boolean => {
+  if (state.get() !== newState) {
+    state.set(newState);
+    return true;
+  } else {
+    return false;
+  }
```
diogo405, line 83: I'd just invert the logic to check if it equals, it's easier to read for me. But it's just personal preference :)
```
if (state.get() === newState) {
    return false;
} else {
    state.set(newState);
    return true;
}
```

modules/tinymce/src/core/main/ts/fmt/FormatChanged.ts
```diff
@@ -5,101 +5,167 @@
...
 
-        matchedFormats[format] = data.callbacks;
-        return false;
+    if (data.similarFalseCallbacks.length > 0) {
+      const match = matchingNode(editor, parents, format, false);
+      if (hasChanged(data.similarFalseState, match.isSome())) {
+        const node = match.getOr(elm);
+        Arr.each(data.similarFalseCallbacks, (callback) => callback(match.isSome(), { node, format, parents }));
       }
+    }
```
diogo405, line 109: Maybe a function here as we have similar code?!

modules/tinymce/src/core/test/ts/browser/fmt/FormatChangeVarsTest.ts
```diff
@@ -0,0 +1,194 @@
...
+    beforeEach(() => {
+      const editor = hook.editor();
+      editor.setContent(`<p>Unstyled content <span style="font-family: ${comicSansFont}">Styled content</span></p>`);
+      TinySelections.setCursor(editor, [ 0, 0 ], 0);
+      clearEvents();
+    });
+
+    it('TINY-7713: Fires when entering styled content', () => {
+      const editor = hook.editor();
+      // Inside the span
```
diogo405, line 65: I like this comment 👍  much easier than 0s and 1s

diogo405 (approved):

modules/tinymce/CHANGELOG.md
```diff
@@ -75,6 +76,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The nested menu item chevron icon was not fading when the menu item was disabled #TINY-7700
 - The table dialog did not always respect the `table_style_with_css` option #TINY-4926
 - Pasting into a table with multiple cells selected could cause the content to be pasted in the wrong location #TINY-7485
+- `editor.formatter.formatChanged` would ignore the `similar` parameter if another callback had alraedy been registered for the same format #TINY-7713
```
JamesToohey, line 79:
```suggestion
- `editor.formatter.formatChanged` would ignore the `similar` parameter if another callback had already been registered for the same format #TINY-7713
```

modules/tinymce/CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added a new `tablecolheader` toolbar button and menu item to toggle the header state of column cells #TINY-7482
 - Added a new `table-row-numbering` icon #TINY-7327
 - Added a new `iframe_aria_text` setting to set the iframe title attribute #TINY-1264
+- `editor.formatter.formatChanged` now supports filtering with format variables #TINY-7713
```
JamesToohey, line 30: It might just be me but this entry doesn't really explain what has been added

JamesToohey (commented):
I just did a quick skim, will review properly later today

modules/tinymce/CHANGELOG.md
lnewson: Agreed and I'd also say `filtering` is the wrong term. It's more like listening to specific format vars changing.

modules/tinymce/src/core/main/ts/fmt/FormatChanged.ts
lnewson: This suggestion is harder to read for me and would make including the comment harder to read as well, so I prefer the original here. We should generally only be using ternaries for fairly simple expressions (it'll still get minified to a ternary in this case anyways).

modules/tinymce/src/core/main/ts/fmt/FormatChanged.ts
```diff
@@ -5,101 +5,167 @@
...
+          return false;
+        }
+      }
+    }
+  );
+
+const getParents = (editor: Editor, elm?: Element): Element[] => {
+  const element = elm ?? editor.selection.getNode();
+  return Arr.filter(FormatUtils.getParents(editor.dom, element), (node) =>
+    node.nodeType === 1 && !node.getAttribute('data-mce-bogus')
```
lnewson, line 73: I know this was like this previously, but since you've done a lot of refactoring/cleaning up here, can we please use the `NodeType.isElement` function here instead of the `nodeType` comparision?

modules/tinymce/src/core/main/ts/fmt/FormatChanged.ts
```diff
@@ -5,101 +5,167 @@
...
-  unbind: () => void;
+interface CallbackGroup {
+  // Bundle all callbacks with similar=true and unspecified vars
+  readonly similarState: Cell<boolean>;
+  readonly similarCallbacks: FormatChangeCallback[];
+  // Also bundle all callbacks with similar=false and unspecified vars
+  readonly similarFalseState: Cell<boolean>;
+  readonly similarFalseCallbacks: FormatChangeCallback[];
+  // Then for callbacks that have variables, handle everything separately
+  readonly others: Array<CustomCallback>;
```
lnewson, line 38:
```suggestion
  readonly others: CustomCallback[];
```

modules/tinymce/src/core/main/ts/fmt/FormatChanged.ts
```diff
@@ -5,101 +5,167 @@
...
 };
 
-const updateAndFireChangeCallbacks = (editor: Editor, elm: Element, currentFormats: Cell<FormatCallbacks>, formatChangeData: RegisteredFormats) => {
-  const formatsList = Obj.keys(currentFormats.get());
-  const newFormats: FormatCallbacks = { };
-  const matchedFormats: FormatCallbacks = { };
+const matchingNode = (editor: Editor, parents: Element[], format: string, similar: boolean, vars?: FormatVars): Optional<Element> =>
+  Arr.findUntil(
+    parents,
+    (node) => { // is this the node we want?
```
lnewson, line 52: Are these TODOs or what? They aren't very clear comments, so can we please improve them

modules/tinymce/src/core/main/ts/fmt/FormatChanged.ts
```diff
@@ -5,101 +5,167 @@
...
-          // Execute callbacks
-          Arr.each(data.callbacks, (callback: FormatChangeCallback) => {
-            callback(true, { node, format, parents });
-          });
-
-          newFormats[format] = data.callbacks;
-        }
+  const parents = getParents(editor, elm);
+
+  Obj.each(registeredCallbacks, (data: CallbackGroup, format: string) => {
```
lnewson, line 94:
```suggestion
  Obj.each(registeredCallbacks, (data, format) => {
```
The types can be inferred here so we shouldn't be including them.

modules/tinymce/src/core/main/ts/fmt/FormatChanged.ts
```diff
@@ -5,101 +5,167 @@
...
+        group.similarCallbacks.push(callback);
+        if (group.similarCallbacks.length === 1) {
+          group.similarState.set(getCurrent());
+        }
+      } else {
+        group.similarFalseCallbacks.push(callback);
+        if (group.similarFalseCallbacks.length === 1) {
+          group.similarFalseState.set(getCurrent());
+        }
+      }
```
lnewson, line 160: To go with Diogo's earlier comment, this logic is also duplicated so it would be good to refactor that.

modules/tinymce/src/core/main/ts/fmt/FormatChanged.ts
```diff
@@ -5,101 +5,167 @@
...
+interface CustomCallback {
+  readonly state: Cell<boolean>;
+  readonly similar: boolean;
+  readonly vars: FormatVars;
+  readonly callback: FormatChangeCallback;
+}
 
-export interface UnbindFormatChanged {
-  unbind: () => void;
+interface CallbackGroup {
```
lnewson, line 30: I'm not really a fan of this data structure, as it requires a lot of looping to do the various operations (remember we this is for NodeChange so performance does matter somewhat). I'll check this out and have a play locally to see if I can come up with any other suggestions.

ashfordneil: What it comes down to to me isn't so much the looping through the data structure that we have to do, but the number of times we need to walk through the parents array calling `formatter.matchNode`. Right now, that's set at one walk operation for each unique `(format, similar, vars)` combination that's been registered, which means we aren't duplicating any work at least.

I we could get different information out of `formatter.matchNode` (like getting it to tell us which variables are applied, instead of giving us a yes/no for a particular set of variables) it might be possible to cut that down to now walk operation for each unique `(format, similar)` combination, or maybe even just one walk operation per format — but with how the formatter is set up I don't think that sort of path is possible.

Anyway, if you've got time, and you can think of a way to cut down the number of walk operations that I've missed, let me know and I'll do what I can to fix it. This is 100% a hot code path, it'd be nice to make it snappy

modules/tinymce/src/core/main/ts/fmt/FormatChanged.ts
ashfordneil: It was more just there as a reminder of what the two callbacks passed to `Arr.findUntil` are. It seems like a pretty uncommon function, so I thought I'd make it clearer what cases both callbacks are trying to handle. Happy to remove or change them if you have any ideas.

lnewson: The problem is the comments aren't saying what it's doing, as they are questions, not statements. If we can't come up with a comment/statement to accurately describe what it's doing I'd prefer to see them removed.

modules/tinymce/CHANGELOG.md
lnewson: I've also just noticed this should be in `Improved` not `Added` as it's not really a new feature, it's an improvement to an existing feature.

modules/tinymce/src/core/test/ts/browser/fmt/FormatChangeVarsTest.ts
lnewson: That wasn't what I meant sorry... I meant have 3 variables for each instance, as I want to avoid having the custom cleanup function.

modules/tinymce/src/core/main/ts/fmt/FormatChanged.ts
lnewson: Yeah, it may not be possible to do anything here. I'm fine with merging this as is if we can't find anything.

TheSpyder:
I'm not sure if RTC can support these vars easily, we [don't support matchnode](https://github.com/tinymce/rockfest/blob/6f8dc35d9084f77a3cca829f1855d1508db38c0f/modules/rtc-plugin/src/main/re/api/PrivateApi.re#L247-L248) but we do have a custom implementation that drives `formatChanged` events (although [`similar` is not supported either](https://github.com/tinymce/rockfest/blob/6f8dc35d9084f77a3cca829f1855d1508db38c0f/modules/rtc-plugin/src/main/re/api/PublicApi.re#L321-L324)).

But in the meantime, can the new vars param please be added to the RTC runtime API? We already log a note when similar is used, I'd like to add one for vars as well just in case we don't find the time to implement this.

modules/tinymce/src/core/main/ts/fmt/FormatChanged.ts
spocke: Can it be pulled out into a local function instead but in the matchingNode function scope. Like `const isMatchingFormatNode = ...` or something. I tend to favor using code to describe what things are instead of comments. It might add extra bundle size though unsure if the minifier will remove the variable declaration and inline it in place like it's currently being used.

modules/tinymce/src/core/main/ts/fmt/FormatChanged.ts
```diff
@@ -5,126 +5,203 @@
...
+  );
 
-  // Update the current formats
-  currentFormats.set({
-    ...newFormats,
-    ...remainingFormats
-  });
+const getParents = (editor: Editor, elm?: Element): Element[] => {
+  const element = elm ?? editor.selection.getNode();
+  return Arr.filter(FormatUtils.getParents(editor.dom, element), (node) =>
```
spocke, line 73: Could `NodeType.isBogus` be used here? It will match on `data-mce-bogus=""` though but I think that is correct since it's an internal attribute. Anything marked with a bogus attribute should be bogus.

spocke (approved):

HAFRMO (approved):

TheSpyder:
> I'm not sure if RTC can support these vars easily

I have been working on implementing support for function formatter variables, and I now realise this can be done. It shouldn't be too difficult, by storing the vars alongside the callback as this PR does.

modules/tinymce/src/core/main/ts/fmt/FormatChanged.ts
ashfordneil: This one should be better now

modules/tinymce/src/core/main/ts/fmt/FormatChanged.ts
ashfordneil: Good idea spocke, thanks

modules/tinymce/src/core/main/ts/fmt/FormatChanged.ts
ashfordneil: Have refactored this, but I'm not sure if the way I did it was good or not. Can someone take another look?

modules/tinymce/src/core/test/ts/browser/fmt/FormatChangeVarsTest.ts
```diff
@@ -0,0 +1,199 @@
...
+    it('TINY-7713: Fires when changing formats', () => {
+      const editor = hook.editor();
+      // Select the entire styled content
+      TinySelections.setSelection(editor, [ 0, 1, 0 ], 0, [ 0, 1, 0 ], 'Styled content'.length);
+      clearEvents();
+      // Change it
+      editor.formatter.apply('fontname', { value: helveticaFont });
+
+      if (browser.isIE) {
+        assert.isTrue(Arr.last(events.general).forall(Fun.identity), 'font-family was either unchanged, or ended up in the on state');
```
ashfordneil, line 92: - [ ] See discussion in slack, can we fix whatever is causing IE to behave like this?

modules/tinymce/src/core/main/ts/api/EventTypes.ts
```diff
@@ -23,6 +24,8 @@ export interface NewBlockEvent { newBlock: Element }
 
 export interface NodeChangeEvent { element: Element; parents: Node[]; selectionChange?: boolean; initial?: boolean }
 
+export interface FormatEvent { format: string; vars: FormatVars }
```
lnewson, line 27: As discussed the other day, this should also be including the optional node/range passed to the apply/remove functions. Additionally, this type is incorrect as `vars` is optional.

- [ ] Fix to include the optional `node` as discussed
- [ ] Fix incorrect types

modules/tinymce/src/core/main/ts/fmt/ApplyFormat.ts
```diff
@@ -312,10 +312,12 @@ const applyFormat = (ed: Editor, name: string, vars?: FormatVars, node?: Node |
       const formatItem = formatList[i];
       if (formatItem.ceFalseOverride && FormatUtils.isSelectorFormat(formatItem) && dom.is(node, formatItem.selector)) {
         setElementFormat(node, formatItem);
+        ed.fire('FormatApply', { format: name, vars });
```
lnewson, line 315: All `fire` calls should be in the `Events.ts` file, so please move these there.

- [ ] Move all `ed.fire` calls to `Events.ts`

modules/tinymce/src/core/main/ts/api/Formatter.ts
```diff
@@ -41,7 +41,7 @@ interface Formatter extends FormatRegistry {
   matchAll: (names: string[], vars?: FormatVars) => string[];
   matchNode: (node: Node, name: string, vars?: FormatVars, similar?: boolean) => Format | undefined;
   canApply: (name: string) => boolean;
-  formatChanged: (names: string, callback: FormatChanged.FormatChangeCallback, similar?: boolean) => { unbind: () => void };
+  formatChanged: (names: string, callback: FormatChanged.FormatChangeCallback, similar?: boolean, value?: FormatVars) => { unbind: () => void };
```
lnewson, line 44: Why is this called `value`? It should be `vars` for consistency.
```suggestion
  formatChanged: (names: string, callback: FormatChanged.FormatChangeCallback, similar?: boolean, vars?: FormatVars) => { unbind: () => void };
```

modules/tinymce/src/core/main/ts/fmt/ApplyFormat.ts
lnewson: Additionally this duplicates the fire call, so this could be improved by removing this and just changing the below line to be `break` instead.

lnewson:
We've missed updating a case this was needed/needed for here: https://github.com/tinymce/tinymce/blob/88094dd33a970680c1b7ed46ed44679040091d58/modules/tinymce/src/plugins/table/main/ts/ui/UiUtils.ts#L38

- [ ] Update table plugin to also use this new API

modules/tinymce/src/core/main/ts/fmt/FormatChanged.ts
```diff
@@ -5,101 +5,167 @@
...
-            callback(true, { node, format, parents });
-          });
-
-          newFormats[format] = data.callbacks;
-        }
 
-        matchedFormats[format] = data.callbacks;
-        return false;
-      }
+  editor.on('FormatApply FormatRemove', (e: EditorEvent<FormatEvent>) => {
```
lnewson, line 51: This logic is incorrect... it won't work when you pass in `node` to apply/remove, which is also why I had said in the call the other day that needs to be passed through.

- [ ] Fix this to take into account the node/range like object passed to the relevant format functions

lnewson (changes requested):
Marking as needs changes, as there's been a few too many things missed/not accounted for here.

modules/tinymce/src/core/main/ts/fmt/FormatChanged.ts
```diff
@@ -5,101 +5,167 @@
...
+  readonly similar: boolean;
+  readonly vars: FormatVars;
+  readonly callback: FormatChangeCallback;
+}
 
-export interface UnbindFormatChanged {
-  unbind: () => void;
+interface CallbackGroup {
+  // Bundle all callbacks with similar=true and unspecified vars
+  readonly similarState: Cell<boolean>;
```
lnewson, line 35: Can we please change this to use something like this structure instead?
```
interface CallbackGroup {
  // Bundle all callbacks with similar=true and unspecified vars
  readonly similar: {
    state: Cell<boolean>;
    callbacks: FormatChangeCallback[];
  };
  // Also bundle all callbacks with similar=false and unspecified vars
  readonly similarFalse: {
    readonly state: Cell<boolean>;
    readonly callbacks: FormatChangeCallback[];
  };
  // Then for callbacks that have variables, handle everything separately
  readonly others: CustomCallback[];
}
```
That way it lets us run the same functions over the `simliar` and `similarFalse` states.

TheSpyder:
> Marking as needs changes, as there's been a few too many things missed/not accounted for here.

I also added a task to one of my earlier comments that may have been missed. I'm a little surprised there wasn't a compiler error, actually, but I guess we did mark the `vars` parameter as optional.

modules/tinymce/src/core/main/ts/fmt/FormatChanged.ts
```diff
@@ -5,101 +5,167 @@
...
+  readonly similarCallbacks: FormatChangeCallback[];
+  // Also bundle all callbacks with similar=false and unspecified vars
+  readonly similarFalseState: Cell<boolean>;
+  readonly similarFalseCallbacks: FormatChangeCallback[];
+  // Then for callbacks that have variables, handle everything separately
+  readonly others: CustomCallback[];
```
lnewson, line 41: This is called 2 different things on this line alone: `others` and `custom`. Can we please use the same name throughout? Also IIUC, this is for variable based matches, so maybe we should call it something related to that instead?

modules/tinymce/src/core/main/ts/fmt/FormatChanged.ts
```diff
@@ -5,101 +5,167 @@
...
     return true;
   }
-}).t;
+};
+
+const updateAndFireChangeCallbacks = (
+  editor: Editor,
+  elm: Element,
+  registeredCallbacks: RegisteredFormats
+) => {
```
lnewson, line 97:
```suggestion
const updateAndFireChangeCallbacks = (editor: Editor, elm: Element, registeredCallbacks: RegisteredFormats) => {
```
This shouldn't be wrapped, as it's well below the 160 char limit.

modules/tinymce/src/core/main/ts/fmt/FormatChanged.ts
```diff
@@ -108,23 +174,33 @@ const addListeners = (registeredFormatListeners: Cell<RegisteredFormats>, format
...
-  Arr.each(formats.split(','), (format) => {
-    formatChangeItems[format].callbacks = Arr.filter(formatChangeItems[format].callbacks, (c) => c !== callback);
-
-    if (formatChangeItems[format].callbacks.length === 0) {
-      delete formatChangeItems[format];
-    }
-  });
+  Arr.each(formats.split(','), (format) => Obj.get(formatChangeItems, format).each((group) => {
+    formatChangeItems[format] = {
+      similarState: group.similarState,
```
lnewson, line 179: Why aren't we removing the callback here? That seems like it'd be a bug.

modules/tinymce/src/core/main/ts/fmt/FormatChanged.ts
```diff
@@ -5,101 +5,172 @@
...
+  // Ignore bogus nodes like the <a> tag created by moveStart()
+  const parents = getParents(editor, elm);
+
+  Obj.each(registeredCallbacks, (data, format) => {
+    Arr.each([ data.withSimilar, data.withoutSimilar ], (spec) => {
+      if (spec.callbacks.length > 0) {
+        const match = matchingNode(editor, parents, format, spec.similar);
+        if (hasChanged(spec.state, match.isSome())) {
+          const node = match.getOr(elm);
+          Arr.each(spec.callbacks, (callback) => callback(match.isSome(), { node, format, parents }));
```
diogo405, line 111: I think we can still create a function for this part... The only difference with the code below would be the array of callbacks?!

modules/tinymce/src/core/main/ts/fmt/FormatChanged.ts
```diff
@@ -5,101 +5,172 @@
...
-          Arr.each(data.callbacks, (callback: FormatChangeCallback) => {
-            callback(true, { node, format, parents });
-          });
-
-          newFormats[format] = data.callbacks;
-        }
+  editor.on('FormatApply FormatRemove', (e: EditorEvent<FormatEvent>) => {
+    const element = Optional.from(e.node)
+      .map((nodeOrRange) => Obj.get(nodeOrRange as RangeLikeObject, 'startContainer').getOr(nodeOrRange as Node))
+      .bind((node) => NodeType.isElement(node) ? Optional.some(node) : Optional.from(node.parentElement))
```
lnewson, line 58: This likely is going to be the wrong element in some cases, as `startContainer` could be a text node or an element itself. So if I'm understanding this logic correctly, we should only be using `parentElement` if it's a text node.

modules/tinymce/src/core/main/ts/fmt/FormatChanged.ts
```diff
@@ -108,23 +179,38 @@ const addListeners = (registeredFormatListeners: Cell<RegisteredFormats>, format
...
+      withSimilar: {
+        state: group.withSimilar.state,
+        similar: group.withSimilar.similar,
+        callbacks: Arr.filter(group.withSimilar.callbacks, (cb) => cb !== callback),
+      },
+      withoutSimilar: {
+        state: group.withoutSimilar.state,
+        similar: group.withoutSimilar.similar,
+        callbacks: Arr.filter(group.withoutSimilar.callbacks, (cb) => cb !== callback),
+      },
```
lnewson, line 193:
```suggestion
      withSimilar: {
        ...group.withSimilar
        callbacks: Arr.filter(group.withSimilar.callbacks, (cb) => cb !== callback),
      },
      withoutSimilar: {
        ...group.withoutSimilar,
        callbacks: Arr.filter(group.withoutSimilar.callbacks, (cb) => cb !== callback),
      },
```

modules/tinymce/src/core/main/ts/fmt/FormatChanged.ts
```diff
@@ -5,101 +5,172 @@
...
-            callback(true, { node, format, parents });
-          });
-
-          newFormats[format] = data.callbacks;
-        }
+  editor.on('FormatApply FormatRemove', (e: EditorEvent<FormatEvent>) => {
+    const element = Optional.from(e.node)
+      .map((nodeOrRange) => Obj.get(nodeOrRange as RangeLikeObject, 'startContainer').getOr(nodeOrRange as Node))
+      .bind((node) => NodeType.isElement(node) ? Optional.some(node) : Optional.from(node.parentElement))
+      .getOrThunk(() => editor.selection.getNode());
```
lnewson, line 59: Should this be using `getStart` instead of `getNode` to match how `NodeChange` generates it's `element` property?

modules/tinymce/src/core/main/ts/fmt/FormatChanged.ts
```diff
@@ -5,101 +5,172 @@
...
 };
 
-const filterRemainingFormats = (currentFormats: FormatCallbacks, matchedFormats: FormatCallbacks, elm: Element, parents: Node[]) => Obj.bifilter(currentFormats, (callbacks: FormatChangeCallback[], format: string) => {
-  if (!Obj.has(matchedFormats, format)) {
-    // Execute callbacks
-    Arr.each(callbacks, (callback: FormatChangeCallback) => {
-      callback(false, { node: elm, format, parents });
-    });
+const getParents = (editor: Editor, elm?: Element): Element[] => {
+  const element = elm ?? editor.selection.getNode();
```
lnewson, line 86: Similar to the above, I believe using `getNode` is incorrect here and we want to be using `getStart` instead since the old logic also always worked off that instead of the common ancestor.

lnewson: Also maybe this fallback should be a shared function between the `FormatApply` lookup and here?

modules/tinymce/src/core/main/ts/fmt/FormatChanged.ts
```diff
@@ -5,101 +5,172 @@
...
-          // Execute callbacks
-          Arr.each(data.callbacks, (callback: FormatChangeCallback) => {
-            callback(true, { node, format, parents });
-          });
-
-          newFormats[format] = data.callbacks;
-        }
+  editor.on('FormatApply FormatRemove', (e: EditorEvent<FormatEvent>) => {
+    const element = Optional.from(e.node)
+      .map((nodeOrRange) => Obj.get(nodeOrRange as RangeLikeObject, 'startContainer').getOr(nodeOrRange as Node))
```
lnewson, line 57: Be aware this may not be working as intended, as `Obj.get` only gets properties owned by the object. Depending on how a native `Range` is implemented, that check may return false and not return the start container. I'd **strongly** suggest changing this to something like this:
```suggestion
      .map((nodeOrRange) => FormatUtils.isNode(node) ? node : node.startContainer)
```
as that is how its handled elsewhere in formatter, so would keep the processing logic consistent.

lnewson: Just confirmed this logic wouldn't have been working with native ranges, which can be passed in `RemoveFormat`:
```
> r = document.createRange()
> Object.hasOwnProperty.call(r, 'startContainer')
false
```

modules/tinymce/src/core/test/ts/browser/fmt/FormatChangeVarsTest.ts
ashfordneil: Not going to fix this, as it always ends up in the right state anyway and was an existing issue

modules/tinymce/src/core/main/ts/fmt/FormatChanged.ts
ashfordneil: Changed to use getStart, but I don't think it's worth a shared function for just the one line of code

lnewson: That's not why I wanted this as a shared function, it's that this behaviour should always be the same. So if someone comes along later and changes one they are quite likely to miss the second case.

modules/tinymce/src/core/main/ts/fmt/FormatChanged.ts
```diff
@@ -5,101 +5,171 @@
...
-      callback(false, { node: elm, format, parents });
-    });
+const getParents = (editor: Editor, elm?: Element): Element[] => {
+  const element = elm ?? editor.selection.getStart();
+  return Arr.filter(FormatUtils.getParents(editor.dom, element), (node) =>
+    NodeType.isElement(node) && !NodeType.isBogus(node)
+  );
+};
 
+const hasChanged = (state: Cell<boolean>, newState: boolean): boolean => {
```
lnewson, line 91: I missed this earlier, but I'm not very fond of this mutating state while looking it up (I'd never expect a check function to change the state). Can we move the `state.set` call to where this returns true instead please?

modules/tinymce/src/core/main/ts/fmt/FormatChanged.ts
ashfordneil: Done

modules/tinymce/src/core/main/ts/fmt/FormatChanged.ts
ashfordneil: Done

lnewson (approved):

ltrouton (approved):

PR #7102, opened 2021-08-11
TINY-7721: Fix broken typescript types in Jax

diogo405:
Related Ticket:

Description of Changes:
* Change HttpError responseText type from `string` to `ResponseTypeMap[T]` - `HttpError<T extends keyof ResponseTypeMap>`

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/jax/CHANGELOG.md
+++ b/modules/jax/CHANGELOG.md
@@ -10,4 +10,4 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Upgraded to Katamari 8.0, which includes breaking changes to the `Optional` API used in this module.
 
 ### Improved
-- Changed `HttpError.responseText` type from `string` to `ResponseTypeMap[T]`, `T extends keyof ResponseTypeMap`.
+- Improved the `HttpError.responseText` type to track the actual `ResponseType`, instead of just `string`. This means that `HttpError` now requires a generic type that extends `keyof ResponseTypeMap`.
--- a/modules/jax/CHANGELOG.md
+++ b/modules/jax/CHANGELOG.md
@@ -6,8 +6,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
 
+### Improved
+- Improved the `HttpError.responseText` type to track the actual `ResponseType`, instead of just `string`. This means that `HttpError` now requires a generic type that extends `keyof ResponseTypeMap`.
+
 ### Changed
 - Upgraded to Katamari 8.0, which includes breaking changes to the `Optional` API used in this module.
 
-### Improved
-- Improved the `HttpError.responseText` type to track the actual `ResponseType`, instead of just `string`. This means that `HttpError` now requires a generic type that extends `keyof ResponseTypeMap`.
--- a/modules/jax/CHANGELOG.md
+++ b/modules/jax/CHANGELOG.md
@@ -8,3 +8,6 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Changed
 - Upgraded to Katamari 8.0, which includes breaking changes to the `Optional` API used in this module.
+
+### Improved
+- Changed `HttpError.responseText` type from `string` to `ResponseTypeMap[T]`, `T extends keyof ResponseTypeMap`.
--- a/modules/jax/src/main/ts/ephox/jax/core/Http.ts
+++ b/modules/jax/src/main/ts/ephox/jax/core/Http.ts
@@ -84,7 +84,7 @@ const getData = (body: RequestBody): Optional<string | FormData | Blob> =>
     }
   });
 
-const send = <T extends keyof ResponseTypeMap>(init: HttpTypes.HttpRequest<T>): FutureResult<ResponseTypeMap[T], HttpError> => FutureResult.nu((callback) => {
+const send = <T extends keyof ResponseTypeMap>(init: HttpTypes.HttpRequest<T>): FutureResult<ResponseTypeMap[T], HttpError<T>> => FutureResult.nu((callback) => {
   const request = new XMLHttpRequest();
   request.open(init.method, buildUrl(init.url, Optional.from(init.query)), true); // enforced async! enforced type as String!
 
@@ -92,7 +92,7 @@ const send = <T extends keyof ResponseTypeMap>(init: HttpTypes.HttpRequest<T>): 
   applyOptions(request, options);
 
   const onError = () => {
-    ResponseError.handle(init.url, init.responseType, request).get((err) => callback(Result.error(err)));
+    ResponseError.handle<T>(init.url, init.responseType, request).get((err) => callback(Result.error(err)));
   };
 
   const onLoad = () => {
@@ -103,7 +103,7 @@ const send = <T extends keyof ResponseTypeMap>(init: HttpTypes.HttpRequest<T>): 
     } else if (request.status < 100 || request.status >= 400) {
       onError();
     } else {
-      ResponseSuccess.validate(init.responseType, request).get(callback);
+      ResponseSuccess.validate<T>(init.responseType, request).get(callback);
     }
   };
 
@@ -120,16 +120,16 @@ const send = <T extends keyof ResponseTypeMap>(init: HttpTypes.HttpRequest<T>): 
 
 const empty = () => textData('');
 
-const post = <T extends keyof ResponseTypeMap>(init: HttpTypes.PostPutInit<T>): FutureResult<ResponseTypeMap[T], HttpError> =>
+const post = <T extends keyof ResponseTypeMap>(init: HttpTypes.PostPutInit<T>): FutureResult<ResponseTypeMap[T], HttpError<T>> =>
   send({ ...init, method: HttpTypes.HttpMethod.Post });
 
-const put = <T extends keyof ResponseTypeMap>(init: HttpTypes.PostPutInit<T>): FutureResult<ResponseTypeMap[T], HttpError> =>
+const put = <T extends keyof ResponseTypeMap>(init: HttpTypes.PostPutInit<T>): FutureResult<ResponseTypeMap[T], HttpError<T>> =>
   send({ ...init, method: HttpTypes.HttpMethod.Put });
 
-const get = <T extends keyof ResponseTypeMap>(init: HttpTypes.GetDelInit<T>): FutureResult<ResponseTypeMap[T], HttpError> =>
+const get = <T extends keyof ResponseTypeMap>(init: HttpTypes.GetDelInit<T>): FutureResult<ResponseTypeMap[T], HttpError<T>> =>
   send({ ...init, method: HttpTypes.HttpMethod.Get, body: empty() });
 
-const del = <T extends keyof ResponseTypeMap>(init: HttpTypes.GetDelInit<T>): FutureResult<ResponseTypeMap[T], HttpError> =>
+const del = <T extends keyof ResponseTypeMap>(init: HttpTypes.GetDelInit<T>): FutureResult<ResponseTypeMap[T], HttpError<T>> =>
   send({ ...init, method: HttpTypes.HttpMethod.Delete, body: empty() });
 
 const sendProgress = (init: HttpTypes.DownloadHttpRequest, loaded: number) => {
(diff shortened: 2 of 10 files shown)


diogo405:
Converting to draft and wearing a helmet before I get smashed 😅 
Maybe start looking at `HttpError` interface in `HttpError.ts`. This change cascade to a massive type change inside `jax`.
Also, left some comments where I'm sure/dunno what to do about typing.

modules/jax/src/main/ts/ephox/jax/core/Http.ts
```diff
@@ -140,12 +140,12 @@ const sendProgress = (init: HttpTypes.DownloadHttpRequest, loaded: number) => {
...
 
-const fetchDownload = (init: HttpTypes.DownloadHttpRequest): FutureResult<Blob, HttpError> => FutureResult.nu((resolve) => {
+const fetchDownload = (init: HttpTypes.DownloadHttpRequest): FutureResult<Blob, HttpError<DataType.Blob>> => FutureResult.nu((resolve) => {
   const fail = (message: string, status: number) => {
-    resolve(Result.error({
+    resolve(Result.error<Blob, HttpError<DataType.Blob>>({
       message,
       status,
-      responseText: ''
+      responseText: new Blob()
```
diogo405, line 148: Not sure about this. Changed to `Blob` to avoid different return types for `fetchDownload` and `fallbackDownload` and end up with something like this:

`const download = (init: HttpTypes.DownloadHttpRequest): FutureResult<Blob, HttpError<DataType.Blob>> | FutureResult<Blob, HttpError<DataType.Text>>`

modules/jax/src/main/ts/ephox/jax/core/HttpError.ts
```diff
@@ -10,10 +12,10 @@ export const enum HttpErrorCode {
   InternalServerError = 500
 }
 
-export interface HttpError {
+export interface HttpError<T extends keyof ResponseTypeMap> {
   message: string;
   status: HttpErrorCode;
-  responseText: string;
+  responseText: ResponseTypeMap[T];
 }
```
diogo405, line 19: This was my starting point.

modules/jax/src/main/ts/ephox/jax/core/HttpJwt.ts
```diff
@@ -10,24 +10,25 @@ const headers = (headersInput: HttpRequest<ResponseBodyDataTypes>['headers'], to
...
+const requestFreshToken = <T extends keyof ResponseTypeMap>(tokenFactory: JwtTokenFactory): FutureResult<JwtToken, HttpError<T>> => tokenFactory(true);
+const requestCachedToken = <T extends keyof ResponseTypeMap>(tokenFactory: JwtTokenFactory): FutureResult<JwtToken, HttpError<T>> => tokenFactory(false);
 
-const tryAgain = <T>(tokenFactory: JwtTokenFactory, runMethod: RunMethod<T>) => (error: HttpError) => error.status === HttpErrorCode.Unauthorized ? requestFreshToken(tokenFactory).bindFuture(runMethod) : FutureResult.error(error);
+const tryAgain = <T, U extends keyof ResponseTypeMap>(tokenFactory: JwtTokenFactory, runMethod: RunMethod<T, U>) =>
+  (error: HttpError<U>) => error.status === HttpErrorCode.Unauthorized ? requestFreshToken(tokenFactory).bindFuture(runMethod) : FutureResult.error(error);
 
-const runWithToken = <T extends keyof ResponseTypeMap>(runMethod: RunMethod<ResponseTypeMap[T]>, tokenFactory: JwtTokenFactory): FutureResult<T, HttpError> =>
+const runWithToken = <T extends keyof ResponseTypeMap>(runMethod: RunMethod<ResponseTypeMap[T], T>, tokenFactory: JwtTokenFactory): FutureResult<T, HttpError<any>> =>
```
diogo405, line 21: Haven't figured out how to get rid of `any` :/

modules/jax/src/main/ts/ephox/jax/core/Http.ts
ashfordneil: That looks great to me, good thinking

modules/jax/src/main/ts/ephox/jax/core/HttpJwt.ts
lnewson: Just dumping this here from DM's, but I believe this should fix the `any`:
```
const tryAgain = <T, U extends keyof ResponseTypeMap>(tokenFactory: JwtTokenFactory, runMethod: RunMethod<T, U>) =>
  (error: HttpError<U>) => error.status === HttpErrorCode.Unauthorized ? requestFreshToken<U>(tokenFactory).bindFuture(runMethod) : FutureResult.error(error);

const runWithToken = <T extends keyof ResponseTypeMap>(runMethod: RunMethod<ResponseTypeMap[T], T>, tokenFactory: JwtTokenFactory): FutureResult<T, HttpError<T>> =>
  requestCachedToken<T>(tokenFactory)
    .bindFuture<ResponseTypeMap[T]>(
    (token) => runMethod(token).bind((result) => result.fold(tryAgain<T, T>(tokenFactory, runMethod), FutureResult.pure))
  );
```

diogo405: I've `Ctrl+c` `Ctrl+v` ed thanks.

modules/jax/CHANGELOG.md
```diff
@@ -8,3 +8,6 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Changed
 - Upgraded to Katamari 8.0, which includes breaking changes to the `Optional` API used in this module.
+
+### Improved
```
lnewson, line 12: Can we please put this section above Changed please?

modules/jax/CHANGELOG.md
```diff
@@ -8,3 +8,6 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
+- Changed `HttpError.responseText` type from `string` to `ResponseTypeMap[T]`, `T extends keyof ResponseTypeMap`.
```
lnewson, line 13:
```suggestion
- Improved the `HttpError.responseText` type to track the actual `ResponseType`, instead of just `string`. This means that `HttpError` now requires a generic type that extends `keyof ResponseTypeMap`.
```
Maybe?

modules/jax/src/main/ts/ephox/jax/core/HttpError.ts
```diff
@@ -10,10 +12,10 @@ export const enum HttpErrorCode {
   InternalServerError = 500
 }
 
-export interface HttpError {
+export interface HttpError<T extends ResponseType> {
```
lnewson, line 15: Not related to this change, but if you don't mind can we please add `readonly` to these 3 properties as well while we're making changes?

modules/jax/src/main/ts/ephox/jax/core/HttpJwt.ts
```diff
@@ -10,24 +10,25 @@ const headers = (headersInput: HttpRequest<ResponseBodyDataTypes>['headers'], to
   return headersInput ? { ...headersInput, ...authHeader } : authHeader;
 };
 
-type RunMethod<T> = (token: JwtToken) => FutureResult<T, HttpError>;
+type RunMethod<T, U extends ResponseType> = (token: JwtToken) => FutureResult<T, HttpError<U>>;
```
lnewson, line 13: Maybe use `E` instead of `U` here and below as this is capturing the "Error" response type?

lnewson: Also do we want to allow different response types here between a success and failure? We don't allow it in `put`, `post`, etc... In which case this might be better as something like this:
```suggestion
type RunMethod<T extends ResponseType> = (token: JwtToken) => FutureResult<ResponseTypeMap[T], HttpError<T>>;
```
and then we can update the places below to just pass `T extends ResponseType`.

lnewson (approved):

modules/jax/src/main/ts/ephox/jax/core/HttpJwt.ts
```diff
@@ -10,24 +10,25 @@ const headers = (headersInput: HttpRequest<ResponseBodyDataTypes>['headers'], to
...
+  (error: HttpError<U>) => error.status === HttpErrorCode.Unauthorized ? requestFreshToken<U>(tokenFactory).bindFuture(runMethod) : FutureResult.error(error);
 
-const runWithToken = <T extends keyof ResponseTypeMap>(runMethod: RunMethod<ResponseTypeMap[T]>, tokenFactory: JwtTokenFactory): FutureResult<T, HttpError> =>
-  requestCachedToken(tokenFactory)
+const runWithToken = <T extends ResponseType>(runMethod: RunMethod<ResponseTypeMap[T], T>, tokenFactory: JwtTokenFactory): FutureResult<T, HttpError<T>> =>
+  requestCachedToken<T>(tokenFactory)
     .bindFuture<ResponseTypeMap[T]>(
-    (token) => runMethod(token).bind((result) => result.fold(tryAgain(tokenFactory, runMethod), FutureResult.pure))
+    (token) => runMethod(token).bind((result) => result.fold(tryAgain<T, T>(tokenFactory, runMethod), FutureResult.pure))
   );
```
lnewson, line 25: I realised we also don't need the `ResponseTypeMap[T]` here now either, so this could be something like this:
```
  requestCachedToken<T>(tokenFactory).bindFuture(
    (token) => runMethod(token).bind((result) => result.fold(tryAgain<T, T>(tokenFactory, runMethod), FutureResult.pure))
  );
```

HAFRMO (approved):

modules/jax/src/main/ts/ephox/jax/core/HttpError.ts
diogo405: Done.

modules/jax/src/main/ts/ephox/jax/core/HttpJwt.ts
diogo405: Fixed.

modules/jax/src/main/ts/ephox/jax/core/HttpError.ts
```diff
@@ -10,10 +12,10 @@ export const enum HttpErrorCode {
...
-  status: HttpErrorCode;
-  responseText: string;
+export interface HttpError<T extends ResponseType> {
+  readonly message: string;
+  readonly status: HttpErrorCode;
+  readonly responseText: ResponseTypeMap[T];
 }
 
-export const httpError = <T>(status: number, message: string, responseText: string): Result<T, HttpError> => Result.error<T>({ message, status, responseText });
+export const httpError = <T extends ResponseType>(status: number, message: string, responseText: T): Result<T, HttpError<T>> => Result.error<T>({ message, status, responseText });
```
ashfordneil, line 21:
```suggestion
export const httpError = <T extends ResponseType>(status: number, message: string, responseText: ResponseTypeMap[T]): Result<T, HttpError<T>> => Result.error<T>({ message, status, responseText });
```
Should it be this instead?

modules/jax/src/main/ts/ephox/jax/core/HttpTypes.ts
```diff
@@ -32,7 +32,7 @@ export interface HttpResponse<T extends ResponseBody> {
 }
 
 export type JwtToken = string;
-export type JwtTokenFactory = (fresh: boolean) => FutureResult<JwtToken, HttpError>;
+export type JwtTokenFactory = <T extends ResponseType>(fresh: boolean) => FutureResult<JwtToken, HttpError<T>>;
```
ashfordneil, line 35:
```suggestion
export type JwtTokenFactory<T extends ResponseType> = (fresh: boolean) => FutureResult<JwtToken, HttpError<T>>;
```
I know this means we are going to get more generics propagating everywhere, but I'm pretty sure it has to be this unfortunately. For any given JwtTokenFactory, it's only going to have one type of error it actually returns (based on the response type of the request its based on), so you can't (safely) cast it to a generic function that will return whichever error type the caller wants.

ashfordneil (commented):
Really good improvements here, just one or two types that might need tweaking. Hopefully they don't cascade too much

modules/jax/src/main/ts/ephox/jax/core/HttpTypes.ts
lnewson: Good catch :+1:

ashfordneil (approved):

PR #7107, opened 2021-08-11
TINY-7714: Changed autolink regex.

HAFRMO:
Related Ticket:

Description of Changes:
Changed autolink regex.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/plugins/autolink/main/ts/api/Settings.ts
+++ b/modules/tinymce/src/plugins/autolink/main/ts/api/Settings.ts
@@ -7,7 +7,11 @@
 
 import Editor from 'tinymce/core/api/Editor';
 
-const getAutoLinkPattern = (editor: Editor) => editor.getParam('autolink_pattern', /^(https?:\/\/|ssh:\/\/|ftp:\/\/|file:\/|www\.|(?:mailto:)?[A-Z0-9._%+\-]+@(?!.*@))(.+)$/i);
+const protocol = '[A-Za-z][A-Za-z\\d.+-]{0,14}:\\/\\/(?:[-.~*+=!&;:\'%@?^${}(),\\w]+@)?|www\\.|[-;:&=+$,.\\w]+@';
+const link = '[A-Za-z\\d-]+(?:\\.[A-Za-z\\d-]+)*(?::\\d+)?(?:\\/(?:[-+~=.,%()\\/\\w]*[-+~=%()\\/\\w])?)?(?:\\?(?:[-.~*+=!&;:\'%@?^${}(),\\/\\w]+))?(?:#(?:[-.~*+=!&;:\'%@?^${}(),\\/\\w]+))?';
+const defaultLinkPattern = () => new RegExp('^(' + protocol + ')(' + link + ')([-.~*+=!&;:\'%@?#^${}(),]*)$');
+
+const getAutoLinkPattern = (editor: Editor) => editor.getParam('autolink_pattern', defaultLinkPattern());
 
 const getDefaultLinkTarget = (editor: Editor) => {
   return editor.getParam('default_link_target', false);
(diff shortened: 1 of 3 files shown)


modules/tinymce/src/plugins/autolink/main/ts/api/Settings.ts
```diff
@@ -7,7 +7,11 @@
 
 import Editor from 'tinymce/core/api/Editor';
 
-const getAutoLinkPattern = (editor: Editor) => editor.getParam('autolink_pattern', /^(https?:\/\/|ssh:\/\/|ftp:\/\/|file:\/|www\.|(?:mailto:)?[A-Z0-9._%+\-]+@(?!.*@))(.+)$/i);
+const protocol = '[A-Za-z][A-Za-z\\d.+-]{0,14}:\\/\\/(?:[-.~*+=!&;:\'%@?^${}(),\\w]+@)?|www\\.|[-;:&=+$,.\\w]+@';
+const link = '[A-Za-z\\d-]+(?:\\.[A-Za-z\\d-]+)*(?::\\d+)?(?:\\/(?:[-+~=.,%()\\/\\w]*[-+~=%()\\/\\w])?)?(?:\\?(?:[-.~*+=!&;:\'%@?^${}(),\\/\\w]+))?(?:#(?:[-.~*+=!&;:\'%@?^${}(),\\/\\w]+))?';
+const defaultLinkPattern = () => new RegExp('^(' + protocol + ')(' + link + ')([-.~*+=!&;:\'%@?#^${}(),]*)$');
```
lnewson, line 12: Why wasn't the regex from polaris used here? As the Jira says it was supposed to be used and we definitely don't want this duplicated.

I recall you mentioning something about needing the groups to detect the protocol? We can likely add that to polaris if needed, but this should at it's core be using `Regexes.autolink()` by default.

HAFRMO: Alright, I'll change that up.

modules/tinymce/src/plugins/autolink/test/ts/browser/AutoLinkPluginTest.ts
```diff
@@ -72,8 +74,6 @@ describe('browser.tinymce.plugins.autolink.AutoLinkPluginTest', () => {
   it('TINY-4773: AutoLink: Unexpected urls ended with space', () => {
     const editor = hook.editor();
     assertIsLink(editor, 'first-last@domain', 'mailto:first-last@domain'); // No .com or similar needed.
-    assertIsLink(editor, 'first-last@()', 'mailto:first-last@()'); // Anything goes after the @.
```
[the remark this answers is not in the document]
HAFRMO: That is a fair point. I'll change them to that.

diogo405 (approved):

Afraithe:
If this ends up a bigger ticket than just switching that regexp, lets re-work this for post 5.9. I believe this kind of qualifies as a bug that we could get into a 5.9.1 if we wanted. Lets not try and squeeze more into 5.9.

lnewson:
> If this ends up a bigger ticket than just switching that regexp, lets re-work this for post 5.9. I believe this kind of qualifies as a bug that we could get into a 5.9.1 if we wanted. Lets not try and squeeze more into 5.9.

This has already been marked a blocker for 5.9.0 due to it's priority (top 10 ranked customer ticket) and the fact the other part of this Zendesk issue is also in 5.9.0 already. It's also flagged as an improvement since we only ever supported common protocols before, so this is adding support for custom protocols.

This should be a relatively straightforward thing to solve, so we're happy to take this over if you guys aren't going to have time.

lnewson:
I should also add this definitely can't go in a patch, as it's a `Changed` entry at the minimum since it changes the default regex used by the autolink plugin.

PR #7116, opened 2021-08-13
TINY-6939: The `TableModified` event was not fired when pasting cells into a table

diogo405:
Related Ticket:

Description of Changes:
* Fire TableModified when Ctrl+V
* Sending `{ structure: true, style: true }` when pasting cells

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -76,7 +76,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The `mceTableRowType` was reversing the order of the rows when converting multiple header rows back to body rows #TINY-6666
 - The nested menu item chevron icon was not fading when the menu item was disabled #TINY-7700
 - The table dialog did not always respect the `table_style_with_css` option #TINY-4926
-- Pasting into a table with multiple cells selected could cause the content to be pasted in the wrong location #TINY-748
+- Pasting into a table with multiple cells selected could cause the content to be pasted in the wrong location #TINY-7485
 - The `TableModified` event was not fired when pasting cells into a table #TINY-6939
 - Base64 encoded images with spaces or line breaks in the data URI were not displayed correctly. Patch contributed by RoboBurned
 
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -76,6 +76,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The nested menu item chevron icon was not fading when the menu item was disabled #TINY-7700
 - The table dialog did not always respect the `table_style_with_css` option #TINY-4926
 - Pasting into a table with multiple cells selected could cause the content to be pasted in the wrong location #TINY-7485
+- `TableModified` event was not fired when pasting cells into table via `Ctrl+V` #TINY-6939
 
 ### Deprecated
 - The `bbcode`, `fullpage`, `legacyoutput` and `spellchecker` plugins have been deprecated and marked for removal in the next major release #TINY-7260
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -35,6 +35,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - When scrolling, the context toolbar will stick to where it was previously for large elements, such as tables #TINY-7545
 - The context toolbar will now move out of the way when it overlaps with the selection, such as in table cells #TINY-7192
 - The `formatter.match` API can now take an optional `similar` parameter to check if the format partial matches #TINY-7712
+- Icons with a `-rtl` suffix in their name will now automatically be used when the UI is rendered in right-to-left mode #TINY-7782
 
 ### Changed
 - Changed the load order so that the content css gets loaded before the editor gets populated with contents #TINY-7249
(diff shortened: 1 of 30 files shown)


modules/tinymce/CHANGELOG.md
```diff
@@ -76,6 +76,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The nested menu item chevron icon was not fading when the menu item was disabled #TINY-7700
 - The table dialog did not always respect the `table_style_with_css` option #TINY-4926
 - Pasting into a table with multiple cells selected could cause the content to be pasted in the wrong location #TINY-7485
+- `TableModified` event was not fired when pasting cells into table via `Ctrl+V` #TINY-6939
```
JamesToohey, line 79: Will this be any paste action or just `Ctrl+V`? Should this just be something like

```suggestion
- The `TableModified` event was not fired when pasting cells into a table #TINY-6939
```

modules/tinymce/src/plugins/table/main/ts/actions/Clipboard.ts
```diff
@@ -68,6 +68,7 @@ const registerEvents = (editor: Editor, selections: Selections, actions: TableAc
               editor.selection.setRng(data.rng);
               editor.focus();
               cellSelection.clear(table);
+              editor.fire('TableModified', { ...data, table });
```
JamesToohey, line 71: I'm not sure this will work as intended, also `table` is a SugarElement in this context.

Can we use `Events.fireTableModified` like everywhere else?

`import * as Events from '../api/Events';`

```suggestion
              Events.fireTableModified(editor, table.dom, Events.styleAndStructureModified);
```

JamesToohey (changes requested):
Requesting changes as I don't believe this is firing with the correct event data

modules/tinymce/src/plugins/table/main/ts/actions/Clipboard.ts
JamesToohey: Note: I haven't tested this

lnewson: For additional context, we should pretty much never be calling `editor.fire` from code and it should always be calling a function in the `Events` module.

lnewson: This could also be using `data.effect`, eg:
```suggestion
              Events.fireTableModified(editor, table.dom, data.effect);
```

lnewson (changes requested):
This also needs tests, to please make sure to add them as well. See `ClipboardTest.ts` for the existing tests for this logic.

HAFRMO (approved):

JamesToohey:
We also need to update the PR title because that's the default commit message when squash merging

diogo405:
> We also need to update the PR title because that's the default commit message when squash merging

Done :)

modules/tinymce/src/plugins/table/test/ts/browser/ClipboardTest.ts
```diff
@@ -732,4 +733,33 @@ describe('browser.tinymce.plugins.table.ClipboardTest', () => {
       '</tbody></table>'
     );
   });
+
+  it('TINY-6939: Pasting into a table should fire TableModified event', () => {
+    const editor = hook.editor();
+
+    let callbackCalled = false;
+    let event: TableEventData = { structure: false, style: false };
```
lnewson, line 741: This doesn't accurately test the event is firing correctly, as it only looks at the latest event. This should do what we do elsewhere, whereby we log all the event names/state and then compare them.

Here is an example of one way to do this: https://github.com/tinymce/tinymce/blob/8ee937c891eabef49cb86cce605680b913915346/modules/tinymce/src/plugins/table/test/ts/module/test/TableSizingModeCommandUtil.ts#L74

- [ ] Improve test to ensure only 1 event is fired on paste

lnewson (approved):

modules/tinymce/src/plugins/table/test/ts/browser/ClipboardTest.ts
```diff
@@ -732,4 +733,37 @@ describe('browser.tinymce.plugins.table.ClipboardTest', () => {
...
+
+    const callback = (e: TableEventData) => {
+      events.push({
+        structure: e.structure,
+        style: e.style,
+      });
+    };
+
+    const clearEvents = () => events = [];
+    clearEvents();
```
lnewson, line 749:
```suggestion
```
This is unnecessary here as we've only just initialised it so it's definitely going to be empty.

modules/tinymce/src/plugins/table/test/ts/browser/ClipboardTest.ts
```diff
@@ -732,4 +733,37 @@ describe('browser.tinymce.plugins.table.ClipboardTest', () => {
       '</tbody></table>'
     );
   });
+
+  it('TINY-6939: Pasting into a table should fire TableModified event', () => {
+    const editor = hook.editor();
+    let events: Array<TableEventData> = [];
```
lnewson, line 739:
```suggestion
    const events: TableEventData[] = [];
```

modules/tinymce/src/plugins/table/test/ts/browser/ClipboardTest.ts
```diff
@@ -732,4 +733,37 @@ describe('browser.tinymce.plugins.table.ClipboardTest', () => {
...
+      events.push({
+        structure: e.structure,
+        style: e.style,
+      });
+    };
+
+    const clearEvents = () => events = [];
+    clearEvents();
+
+    editor.on('TableModified', callback);
```
lnewson, line 751: You're not cleaning up this listener at the end of the test.

modules/tinymce/src/plugins/table/test/ts/browser/ClipboardTest.ts
```diff
@@ -732,4 +733,37 @@ describe('browser.tinymce.plugins.table.ClipboardTest', () => {
...
+    selectRangeXY(editor, 'table tr td:nth-child(1)', 'table tr td:nth-child(2)');
+    Clipboard.pasteItems(TinyDom.body(editor), {
+      'text/html': '<table><tbody>' +
+        '<tr><td>1</td><td>2</td></tr>' +
+        '</tbody></table>'
+    });
+
+    assert.deepEqual(events, [{ structure: true, style: true }]);
+
+    clearEvents();
```
lnewson, line 767:
```suggestion
    editor.off('TableModified', callback);
```

lnewson (changes requested):
This was just copied without consideration of use. So I've left some suggestions to fix it given the current use. We should either make sure it's used in isolation for a single test correctly, or make it global so it's applicable to all, as atm it's a mix of the two.

modules/tinymce/src/plugins/table/test/ts/browser/ClipboardTest.ts
diogo405: Fixed.

lnewson (approved):

diogo405:
> Requesting changes as I don't believe this is firing with the correct event data

Can you re-review?

ashfordneil (approved):

JamesToohey (approved):

PR #7118, opened 2021-08-15
TINY-7740: Added transitioning to context toolbars when changing placements

lnewson:
Related Ticket: TINY-7740

Description of Changes:

Alloy:
* Update alloy positioning to support transitioning when placing an object. It supports 3 modes:
  * `all` - This will attempt to transition at all times when moving a component.
  * `layout` - This will attempt to transition whenever the layout changes.
  * `placement` - Similar to `layout` except it won't transition when swapping from say `layout-north` to `layout-inset-north`.
* Expose each components unique identifier as the `uid` property on an `AlloyComponent`.
* Changed the alloy positioning behaviour to accept a `PlacementSpec` instead of a `AnchorSpec` (since we now need to be able to pass the transition configuration as well).
  * Note: A reasonable amount of the changed files in Alloy are because of this. 

Silver:
* Update context toolbars to transition on placement changes (i.e. when going from `north` -> `south`).
* Updated the bounds and bounds overlap logic a little to account for the transitioning requirements.
* Update a number of existing tests to account for the transition delay (the old logic didn't work since things are delayed a little more and the elements changed).

Katamari:
* Updated the `Arr.foldl` and `Arr.foldr` to pass the item index.

Pre-checks:
* [x] Changelog entry added
* [x] Tests have been added (if applicable) - Note: Testing this via automation is difficult as it's largely a visual change, but I've tried to cover what I can.

Review:
* [x] Milestone set


modules/alloy/src/main/ts/ephox/alloy/api/events/NativeEvents.ts
```diff
@@ -2,36 +2,38 @@ import { Fun } from '@ephox/katamari';
...
-const touchmove = constant('touchmove');
-const touchend = constant('touchend');
-const touchcancel = constant('touchcancel');
-const gesturestart = constant('gesturestart');
-const mousedown = constant('mousedown');
-const mousemove = constant('mousemove');
-const mouseout = constant('mouseout');
-const mouseup = constant('mouseup');
-const mouseover = constant('mouseover');
+const contextmenu = constant('contextmenu' as const);
```
lnewson, line 5: I made this change so that the value returned can be used to infer other things (e.g. the event type when used in a callback).

modules/tinymce/src/themes/silver/main/ts/ui/context/ContextToolbarBounds.ts
```diff
@@ -7,16 +7,17 @@
...
 import Editor from 'tinymce/core/api/Editor';
 
 import * as Settings from '../../api/Settings';
 import { UiFactoryBackstageShared } from '../../backstage/Backstage';
 
-const isVerticalOverlap = (a: Bounds, b: Bounds, threshold: number = 0): boolean =>
-  b.bottom - a.y >= threshold && a.bottom - b.y >= threshold;
+// Note: We want to avoid including with small difference such as 0.001px
+const isVerticalOverlap = (a: Bounds, b: Bounds, threshold: number = 0, precision: number = 2): boolean =>
+  Num.toPrecision(b.bottom - a.y, precision) >= threshold && Num.toPrecision(a.bottom - b.y, precision) >= threshold;
```
tiny-james, line 20: Why not just make the threshold 0.01? How is this rounding method different?

I worry that there will be weird edge cases with Num.toPrecision given that we are dealing with floating point numbers which are notorious for not being able to exactly represent some decimal numbers.

tiny-james (approved):

modules/tinymce/src/themes/silver/main/ts/ui/context/ContextToolbarBounds.ts
lnewson: That's a very good point, thanks! I've changed it and removed the `toPrecision` function/tests.

ltrouton (commented):
I have looked through all of the files and couldn't find anything that stood out to me. Admittedly, I don't really have any prior experience with transitions so just leaving a comment instead of a tick for now.

JamesToohey (commented):
I read through what you've done and I also pulled this locally to play with the feature.

I couldn't find any issues, but I'm also with Luke in that I don't know this code very well... so commenting for now.

ltrouton (approved):
Ticking now as we have had a team call going through this PR. Have also manually tested a bit and couldn't find anything standout

PR #7133, opened 2021-08-18
TINY-7697: Update cell-border-color icon

PaulKutuzov:
Related Ticket:
TINY-7697

Description of Changes:
Update cell-border-color icon

Pre-checks:
* [x] ~Changelog entry added~ N/A as this is a new icon for 5.9

Review:
* [x] Milestone set


lnewson (approved):

metricjs (approved):

JamesToohey (commented):
I noticed that the corner of the table is no longer rounded, and also the style is quite different to other icons like https://github.com/tinymce/tinymce/blob/develop/modules/oxide-icons-default/src/svg/table-cell-classes.svg (here the table is greyed out and the cell is colored).

Just checking if that's intentionally different

lostkeys (approved):

lostkeys:
I would probably add a slight rounded outer edge too as @JamesToohey mentioned above, but ultimately I leave @PaulKutuzov to decide.

PaulKutuzov:
@lostkeys @JamesToohey Hey guys! The edges were intentionally unrounded to avoid original confusion with 'Is it button for changing tables border?'. Now it shows literally cell borders. So I'd prefer to leave it as it is

lnewson:
@PaulKutuzov then we'll need to change the other cell icons to also include the shaded outer edge, since atm it's very inconsistent.

PaulKutuzov:
@lnewson Nope, no need to change all the icons referring to Table as an object (e.g. 'table-cell-select', 'table-insert..' etc.) as they use rounded corners to reflect Table metaphor. In this case we refer to cells only, so it's totally fine to use straight corners here.

lnewson:
@PaulKutuzov that's not the issue here... the issue is the black outer edge, not the rounded vs sharp corners. Here's the two side by side that James mentioned:
image
image
You can see the classes one does **not** have sharp edges and also does **not** have a black edge around the whole icon. These 2 icons should be fairly similar as they are both for applying an action to a table cell. Neither will apply the style to the table.

lnewson:
I'm not sure which one is correct, but atm they are different. So one is incorrect, and if you're saying this one is correct, then the other is wrong as well which is why I had mentioned about it needed to be changed

PaulKutuzov:
@lnewson I clearly see your thinking here. And I've tried to explain exactly this issue. 
Original icon was consistent but as I understood was confusing also, that's why we updated it here
image

lnewson (changes requested):
<strike>@PaulKutuzov the original icon wasn't consistent which was the root problem, since it made it look like the color is applied to the table, not the cell. If you look at the classes one the inside cell is shaded and not the outside to indicate it only applies to the cell, not the table.

Effectively, both `cell-border-icon` and `table-cell-classes` have inconsistencies when they should be largely the same. This fixes the issue with the former to make it similar to the the other `table-cell-*` variants (which I believe is what you've been trying to say), but we still have an inconsistency issue with `table-cell-classes` then and that needs fixing if we're deeming it incorrect.

I'm changing my review to request changes until this is resolved, as the entire premise we've sold to our customers about why we don't allow customization is to ensure the UI is consistent. This breaks that promise and we need to ensure our icons are consistent where they apply to the same thing (in this case applying actions to table cells).</strike>

Edit: GitHub somehow submitted my reply twice :cry:

lnewson (changes requested):
@PaulKutuzov the original icon wasn't consistent which was the root problem, since it made it look like the color is applied to the table, not the cell. If you look at the classes one the inside cell is shaded and not the outside to indicate it only applies to the cell, not the table.

Effectively, both `cell-border-icon` and `table-cell-classes` have inconsistencies when they should be largely the same. This fixes the issue with the former to make it similar to the the other `table-cell-*` variants (which I believe is what you've been trying to say), but we still have an inconsistency issue with `table-cell-classes` then and that needs fixing if we're deeming it incorrect.

I'm changing my review to request changes until this is resolved, as the entire premise we've sold to our customers about why we don't allow customization is to ensure the UI is consistent. This breaks that promise and we need to ensure our icons are consistent where they apply to the same thing (in this case applying actions to table cells).

lnewson:
I should add we do get customers raising support cases for design inconsistencies such as the issues James pointed out. It can either be UI, text, etc... issues so customers do expect some level of consistent and easy to understand UI (even more so since that's what we've been promising/selling TinyMCE 5 as).

PaulKutuzov:
Yeah, we should abandon this icon for now, as no agreement on that.

lnewson (approved):

PaulKutuzov:
Okay then, updated icon is here

JamesToohey (approved):

HAFRMO (approved):

PR #6965, opened 2021-06-26
TINY-7192: Update the context toolbar to move out the way when the selection overlaps

lnewson:
Related Ticket: TINY-7192

Description of Changes:
* Added new `flip` inside layout to move the anchor to the position on the opposite axis (eg north -> south).
* Added new `get` functions to `Singleton`
* Split out the context toolbar anchoring and bounds logic more to keep concepts isolated better and reduce the size of `ContextToolbar.ts`
* Cleaned up `ContextToolbar.ts` logic somewhat (use the `InlineView.reposition` API, ensure singletons/cells are cleared on remove, added some missing types and moved a couple things).
* Update the anchoring calculation logic to "flip" the anchor placement when the selection overlaps on toolbar "launch" operations (click, keyup, focus, SetContent and NodeChange)

Pre-checks:
* [x] Changelog entry added
* [x] Tests have been added (if applicable)
* [x] Branch prefixed with `feature/` for new features (if applicable)

Review:
* [x] Milestone set
* [x] Review comments resolved

GitHub issues (if applicable):

PR #7141, opened 2021-08-21
TINY-7192: Simplify get selection bounds logic

lnewson:
Related Ticket: TINY-7192

Description of Changes:
This is a small optimisation to the implementation done in https://github.com/tinymce/tinymce/pull/6965, as while fixing the 4.4 types I found an API in core I wasn't aware of that does what I'd implemented in `ContextToolbarBounds`. So this just switches to that to remove duplication.

Pre-checks:
* [x] ~Changelog entry added~
* [x] ~Tests have been added (if applicable)~ Covered by existing tests

Review:
* [x] Milestone set


lnewson:
Turns out it's the same on every browser except IE 11, as it's somehow still returning a 0x0 rect :cry: That's potentially a bug in the core `CaretPosition` logic, but for now we shouldn't be touching that in 5.9.0.

PR #7146, opened 2021-08-23
TINY-7713: Fix computed line height bug

ashfordneil:
Related Ticket: TINY-7713

Description of Changes:
* Don't let two submenus be active at once

Pre-checks:
* [x] ~Changelog entry added~
* [x] xBranch prefixed with `feature/` for new features (if applicable)

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/themes/silver/main/ts/ui/core/ChoiceControls.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/ui/core/ChoiceControls.ts
@@ -37,27 +37,25 @@ const registerController = <T>(editor: Editor, spec: ControlSpec<T>) => {
     const options = spec.getOptions(editor);
     const initial = spec.getCurrent(editor).map(spec.hash);
 
-    const current = Singleton.destroyable();
+    const current = Singleton.value<Menu.ToggleMenuItemInstanceApi>();
 
-    return Arr.map(
-      options,
-      (value) => ({
-        type: 'togglemenuitem',
-        text: spec.display(value),
-        onSetup: (api) => {
-          const setActive = (active: boolean) => {
-            api.setActive(active);
-            if (active) {
-              current.set({ destroy: () => api.setActive(false) });
-            }
-          };
+    return Arr.map(options, (value) => ({
+      type: 'togglemenuitem',
+      text: spec.display(value),
+      onSetup: (api) => {
+        const setActive = (active: boolean) => {
+          api.setActive(active);
+          if (active) {
+            current.on((oldApi) => oldApi.setActive(false));
+            current.set(api);
+          }
+        };
 
-          setActive(Optionals.is(initial, spec.hash(value)));
-          return spec.watcher(editor, value, setActive);
-        },
-        onAction: () => spec.setCurrent(editor, value)
-      })
-    );
+        setActive(Optionals.is(initial, spec.hash(value)));
+        return spec.watcher(editor, value, setActive);
+      },
+      onAction: () => spec.setCurrent(editor, value)
+    }));
   };
 
   editor.ui.registry.addMenuButton(spec.name, {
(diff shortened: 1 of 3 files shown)


modules/tinymce/src/themes/silver/main/ts/ui/core/ChoiceControls.ts
```diff
@@ -37,14 +37,23 @@ const registerController = <T>(editor: Editor, spec: ControlSpec<T>) => {
...
-          return spec.watcher(editor, value, api.setActive);
+          const setActive = (active: boolean) => {
+            api.setActive(active);
+            if (active) {
+              current.set({ destroy: () => api.setActive(false) });
+            }
+          };
+
+          setActive(Optionals.is(initial, spec.hash(value)));
+          return spec.watcher(editor, value, setActive);
```
lnewson, line 56: Does the singleton need to be destroyed here when the button is destroyed, otherwise this will potentially keep the API reference in memory indefinitely?

modules/tinymce/src/themes/silver/main/ts/ui/core/ChoiceControls.ts
```diff
@@ -37,14 +37,23 @@ const registerController = <T>(editor: Editor, spec: ControlSpec<T>) => {
     const options = spec.getOptions(editor);
     const initial = spec.getCurrent(editor).map(spec.hash);
 
+    const current = Singleton.destroyable();
+
     return Arr.map(
```
lnewson, line 42: Can we flatten this to a single line, as it'd make it more readable?
```
    return Arr.map(options, (value) => ({
```

modules/tinymce/src/themes/silver/main/ts/ui/core/ChoiceControls.ts
```diff
@@ -37,14 +37,23 @@ const registerController = <T>(editor: Editor, spec: ControlSpec<T>) => {
...
+      (value) => ({
         type: 'togglemenuitem',
         text: spec.display(value),
         onSetup: (api) => {
-          api.setActive(Optionals.is(initial, spec.hash(value)));
-          return spec.watcher(editor, value, api.setActive);
+          const setActive = (active: boolean) => {
+            api.setActive(active);
+            if (active) {
+              current.set({ destroy: () => api.setActive(false) });
```
JamesToohey, line 51: After a _significant_ amount of time... I now understand what this is doing. Singletons have always been a bit odd for me, but do you think we could do it a slightly more verbose but easier to understand way using `Singleton.value`?

```
current.on((oldApi) => oldApi.setActive(false));
current.set(api);
```

modules/tinymce/src/themes/silver/test/ts/browser/editor/core/ChoiceControlsTest.ts
```diff
@@ -108,6 +108,18 @@ describe('browser.tinymce.themes.silver.editor.core.ChoiceControlsTest', () => {
           await pAssertOptions(editor, spec.menuSelector, [ '1', '1.1', '1.2', '1.3', '1.4', '1.5', '2' ], Optional.some('1.1'));
           spec.close(editor, 'Line height');
         });
+
+        it(`TINY-7713: ${spec.name} updates if computed line height changes`, async () => {
```
lnewson, line 112: This test is failing on Safari, likely due to the same reason as https://github.com/tinymce/tinymce/blob/837146ecbec7f4508258590d039823fa17322f3c/modules/tinymce/src/core/test/ts/browser/commands/LineHeightTest.ts#L28

modules/tinymce/src/themes/silver/main/ts/ui/core/ChoiceControls.ts
ashfordneil: The singleton should get garbage collected the same way all of the other state here does — the only things that hold a reference to it are the `setActive` functions we pass (through `spec.watcher`) into `FormatChanged`, and those call-backs gets removed automatically.

ltrouton (approved):

lnewson (approved):

PR #7170, opened 2021-09-02
TINY-7875: Added content formatter function

HAFRMO:
Related Ticket:

Description of Changes:
Added formatter for content

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/mcagar/src/main/ts/ephox/mcagar/alien/EditorTypes.ts
```diff
@@ -1,5 +1,8 @@
 type EventCallback = (event: any) => void;
 
+export type GetContentFormatter = (editor: Editor, args: GetContentArgs) => string;
+export type SetContentFormatter = (editor: Editor, content: any, args: SetContentArgs) => string;
```
HAFRMO, line 4: The `content: any` is because we don't seem to expose AstNode and I didn't know how much of that we wanted to drag out. If anyone know a good solution please, do tell.

lnewson (dismissed):
Flagging this as needs changes just because planning things I'd requested for this haven't been sorted out still, so this can't be reviewed/merged until that is resolved.

lnewson:
I had a quick skim over this, and will the above still applies (e.g. we need the public API planned/discussed), there's also been a number of get/set content APIs that have been missed in this implementation. e.g.
- `editor.selection.getContent`
- `editor.selection.setContent`
- `editor.insertContent` (and the equivalent `mceInsertContent` command)

lnewson:
Another large issue I see is that the content formatters are registered in a module scoped object, which won't work when using multiple editors.

HAFRMO:
> I had a quick skim over this, and will the above still applies (e.g. we need the public API planned/discussed), there's also been a number of get/set content APIs that have been missed in this implementation. e.g.
> 
>     * `editor.selection.getContent`
> 
>     * `editor.selection.setContent`
> 
>     * `editor.insertContent` (and the equivalent `mceInsertContent` command)

This is by design, and such changes would be implemented at a later date if needed. The primary need is for set/getContent, which is all that is implemented here.

lnewson:
> This is by design, and such changes would be implemented at a later date if needed. The primary need is for set/getContent, which is all that is implemented here.

I'm sorry but that's an unacceptable design and this would have been caught if processes were followed, such as this "plan" being reviewed by the editor team before implementing. This needs to be a full implementation, as the various APIs are used throughout in varying ways, so just picking one or two and doing that isn't acceptable for a core change.

If you want to do this in chunks, that's fine but this **cannot** be merged into a main/release branch until it's complete and the APIs I've mentioned have also been updated to work with this new feature.

lnewson:
> Flagging this as needs changes just because planning things I'd requested for this haven't been sorted out still, so this can't be reviewed/merged until that is resolved.

This has now been resolved, as per our discussion today. I'm going to leave this as needs changes though until the changes are done and the module scope object for the custom formats are resolved though.

modules/tinymce/src/core/main/ts/content/GetContentImpl.ts
```diff
@@ -15,52 +15,91 @@ import Tools from '../api/util/Tools';
...
-import { Content, ContentFormat, GetContentArgs } from './ContentTypes';
+import { Content, GetContentArgs } from './ContentTypes';
+import { GetContentFormatter } from './EditorContent';
+
+const defaultContentFormatter = (editor: Editor, args: GetContentArgs): Content =>
+  Optional.from(editor.getBody())
+    .fold(
+      Fun.constant(args.format === 'tree' ? new AstNode('body', 11) : ''),
+      (body) => getContentFromBody(editor, args, body)
+    );
```
JamesToohey, line 26: Can we move this up

```suggestion
const defaultContentFormatter = (editor: Editor, args: GetContentArgs): Content =>
  Optional.from(editor.getBody()).fold(
    Fun.constant(args.format === 'tree' ? new AstNode('body', 11) : ''),
    (body) => getContentFromBody(editor, args, body)
  );
```

modules/tinymce/src/core/main/ts/content/GetContentImpl.ts
```diff
@@ -15,52 +15,91 @@ import Tools from '../api/util/Tools';
...
+
+const addDefaultGetFormats = (formatCell: Cell<Record<string, GetContentFormatter>>) => {
```
JamesToohey, line 28: This inline type is getting quite verbose and is repeated a few times, could we move it out?

modules/tinymce/src/core/main/ts/content/SetContentImpl.ts
```diff
@@ -98,30 +111,48 @@ const setContentTree = (editor: Editor, body: HTMLElement, content: AstNode, arg
...
   return content;
 };
 
+const getFormatter = (format: string, formatCell: Cell<Record<string, SetContentFormatter>>) => {
+  return Optional.from(formatCell.get()[format]).fold(
+    () => {
+      throw new Error(`Content formatter ${format} not recognized.`);
+    },
+    Fun.identity
+  );
```
JamesToohey, line 123: Can this be a `getOrDie`?

modules/tinymce/src/core/main/ts/content/GetContentImpl.ts
```diff
@@ -15,52 +15,91 @@ import Tools from '../api/util/Tools';
...
+};
+
+const getFormatter = (format: string, getCell: Cell<Record<string, GetContentFormatter>>) => {
+  return Optional.from(getCell.get()[format]).fold(
+    () => {
+      // eslint-disable-next-line no-console
+      throw new Error(`Content formatter ${format} not recognized.`);
+    },
+    Fun.identity
+  );
```
JamesToohey, line 78: Can this be a `getOrDie`?

JamesToohey (changes requested):
Could we get a more detailed description explaining exactly what is happening in this PR? It's quite difficult to follow and I haven't touched a lot of this code, so it took me quite a while to review.

Also please:

- [ ] Add a test for when a content formatter is not found on `setContent`
- [ ] Add a test for when a content formatter is not found on `getContent`

ashfordneil (approved):

modules/tinymce/src/core/main/ts/content/EditorContent.ts
```diff
@@ -5,14 +5,54 @@
...
+import Editor from '../api/Editor';
+import * as AddFormat from './AddFormat';
 import { Content, GetContentArgs, SetContentArgs } from './ContentTypes';
 import { getContent } from './GetContent';
 import { setContent } from './SetContent';
 
+interface EditorContent {
+  addFormat: (format: string, formatGetter: GetContentFormatter, formatSetter: SetContentFormatter) => void;
+  get: (args: Partial<GetContentArgs>) => Content;
+  set: (content: Content, args?: Partial<SetContentArgs>) => Content;
```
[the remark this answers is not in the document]
lnewson: Actually ignore this sorry, as I forgot these are on the public API. Since this logic is now public, it should also be moved to the `api/content/` directory instead. 

- [ ] Move file to `api` directory

modules/tinymce/src/core/main/ts/content/GetContentImpl.ts
```diff
@@ -15,52 +15,91 @@ import Tools from '../api/util/Tools';
...
+    .fold(
+      Fun.constant(args.format === 'tree' ? new AstNode('body', 11) : ''),
+      (body) => getContentFromBody(editor, args, body)
+    );
+
+const addDefaultGetFormats = (formatCell: Cell<Record<string, GetContentFormatter>>) => {
+  addGetContentFormatter('raw', defaultContentFormatter, formatCell);
+  addGetContentFormatter('text', defaultContentFormatter, formatCell);
+  addGetContentFormatter('html', defaultContentFormatter, formatCell);
+  addGetContentFormatter('tree', defaultContentFormatter, formatCell);
```
lnewson, line 32: This is unnecessarily duplicated... we also don't need to use the same formatter for `tree` now so can we please make it a separate formatter instead?

For the duplication, this could be changed to something like this:

```
Arr.each([ 'raw', 'text', 'html' ], (format) {
  addGetContentFormatter(format, defaultContentFormatter, formatCell);
});
addGetContentFormatter('tree'', treeContentFormatter, formatCell);
```

modules/tinymce/src/core/main/ts/content/SetContentImpl.ts
```diff
@@ -19,9 +19,28 @@ import * as NodeType from '../dom/NodeType';
...
+    Fun.constant(content),
+    (body) => isTreeNode(content) ? setContentTree(editor, body, content, updatedArgs) : setContentString(editor, body, content, updatedArgs)
+  );
+};
+
+const addDefaultSetFormats = (formatCell: Cell<Record<string, SetContentFormatter>>) => {
+  addSetContentFormatter('raw', defaultContentFormatter, formatCell);
+  addSetContentFormatter('text', defaultContentFormatter, formatCell);
+  addSetContentFormatter('html', defaultContentFormatter, formatCell);
+  addSetContentFormatter('tree', defaultContentFormatter, formatCell);
```
lnewson, line 41: See my previous comments on the getter about duplication and making `tree` a separate formatter since it deals with `AstNode` instead of `string`.

modules/tinymce/src/core/main/ts/content/SetContentImpl.ts
```diff
@@ -98,30 +111,48 @@ const setContentTree = (editor: Editor, body: HTMLElement, content: AstNode, arg
...
 
-  if (!args.no_events) {
-    editor.fire('SetContent', args);
-  }
-
   return content;
 };
 
+const getFormatter = (format: string, formatCell: Cell<Record<string, SetContentFormatter>>) => {
+  return Optional.from(formatCell.get()[format]).fold(
```
lnewson, line 118:
```suggestion
  return Obj.get(formatCell.get(), format).fold(
```

modules/tinymce/src/core/main/ts/content/SetContentImpl.ts
```diff
@@ -98,30 +111,48 @@ const setContentTree = (editor: Editor, body: HTMLElement, content: AstNode, arg
...
 const setupArgs = (args: Partial<SetContentArgs>, content: Content): SetContentArgs => ({
   format: defaultFormat,
   ...args,
   set: true,
   content: isTreeNode(content) ? '' : content
 });
 
-export const setContentInternal = (editor: Editor, content: Content, args: Partial<SetContentArgs>): Content => {
+const setContentInternal = (editor: Editor, content: Content, args: Partial<SetContentArgs>, formatCell: Cell<Record<string, SetContentFormatter>>): Content => {
+  const formatter = getFormatter(args.format || defaultFormat, formatCell);
```
lnewson, line 134:
```suggestion
  const formatter = getFormatter(args.format ?? defaultFormat, formatCell);
```
This way it only falls back if it's null or undefined and not just falsy.

modules/tinymce/src/core/main/ts/api/Editor.ts
```diff
@@ -342,6 +343,9 @@ class Editor implements EditorObservable {
     this.mode = modeInstance;
     this.setMode = modeInstance.set;
 
+    const content = EditorContent.EditorContent(self);
+    this.content = content;
```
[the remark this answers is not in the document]
HAFRMO: This was done to allow adding `this.getContent = content.get` without the typing going haywire, but it never manifested. I'll change it over to the direct assignment.

modules/mcagar/src/main/ts/ephox/mcagar/alien/EditorTypes.ts
```diff
@@ -51,6 +56,7 @@ export interface Editor {
 
   getContent: (args?: Partial<GetContentArgs>) => string;
   setContent: (content: string, args?: Partial<SetContentArgs>) => void;
+  content: Content;
```
lnewson, line 59: This will need to be optional, as these types need to work for all TinyMCE versions, not just the latest version.

On that note, does this actually need to be added to the common editor type in McAgar? I don't believe it does since McAgar doesn't make use of it.

modules/tinymce/src/core/main/ts/Rtc.ts
```diff
@@ -54,6 +55,7 @@ interface RtcRuntimeApi {
   editor: {
     getContent: (args: Partial<GetContentArgs>) => Content;
     setContent: (content: Content, args: Partial<SetContentArgs>) => Content;
+    addContentFormat: (format: string, formatGetter: GetContentFormatter, getCell: Cell<Record<string, GetContentFormatter>>, formatSetter: SetContentFormatter, setCell: Cell<Record<string, SetContentFormatter>>) => void;
```
lnewson, line 58: @TheSpyder @spocke are you okay with this API for the RTC integration?

modules/tinymce/src/core/main/ts/Rtc.ts
```diff
@@ -396,16 +405,27 @@ export const toggleFormat = (editor: Editor, name: string, vars: Record<string,
...
+  getRtcInstanceWithFallback(editor).editor.setContent(content, args, setCell);
+
+export const addContentFormat = (
+  editor: Editor,
+  format: string,
+  formatGetter: GetContentFormatter,
+  getCell: Cell<Record<string, GetContentFormatter>>,
+  formatSetter: SetContentFormatter,
+  setCell: Cell<Record<string,
+  SetContentFormatter>>
```
lnewson, line 421:
```suggestion
  setCell: Cell<Record<string, SetContentFormatter>>
```

modules/tinymce/src/core/main/ts/content/AddFormat.ts
```diff
@@ -0,0 +1,14 @@
+import { Cell } from '@ephox/katamari';
```
lnewson, line 1: This import is in the wrong location.

modules/tinymce/src/core/main/ts/content/EditorContent.ts
```diff
@@ -5,14 +5,54 @@
...
+import { addDefaultSetFormats } from 'tinymce/core/content/SetContentImpl';
+
+import Editor from '../api/Editor';
+import * as AddFormat from './AddFormat';
 import { Content, GetContentArgs, SetContentArgs } from './ContentTypes';
 import { getContent } from './GetContent';
 import { setContent } from './SetContent';
 
+interface EditorContent {
+  addFormat: (format: string, formatGetter: GetContentFormatter, formatSetter: SetContentFormatter) => void;
```
lnewson, line 20: This name is not per the spec, it should be called `register` format. From the meeting notes:
```
ACTION: change to editor.content.registerFormat
```

- [ ] Fix incorrect API name

modules/tinymce/src/core/main/ts/content/GetContentImpl.ts
lnewson: Something like `GetContentRegistry` and `SetContentRegistry` might be a good name for this :shrug:

modules/tinymce/src/core/main/ts/content/EditorContent.ts
```diff
@@ -5,14 +5,54 @@
...
+  get: (args: Partial<GetContentArgs>) => Content;
+  set: (content: Content, args?: Partial<SetContentArgs>) => Content;
+}
+
+export type GetContentFormatter = (editor: Editor, args: GetContentArgs) => Content;
+export type SetContentFormatter = (editor: Editor, content: Content, args: SetContentArgs) => Content;
```
lnewson, line 26: `Formatter` is misleading, as it sounds like it's related to the formatter API. Can we please rename this to something else? even just changing it to `GetContentCallback` and `SetContentCallback` might be a good start?

modules/tinymce/src/core/main/ts/content/EditorContent.ts
```diff
@@ -5,14 +5,54 @@
...
+
+const EditorContent = (editor: Editor): EditorContent => {
+  const setCell: Cell<Record<string, SetContentFormatter>> = Cell<Record<string, SetContentFormatter>>({});
+  const getCell: Cell<Record<string, GetContentFormatter>> = Cell<Record<string, GetContentFormatter>>({});
```
lnewson, line 30: I'm not really a fan of these names that are used everywhere, as it's not entirely clear. Can we rename them to something like `getContentRegistry` and `setContentRegistry`?

Also we probably shouldn't be passing the cells around, as the addition of new formats should only happen in this file (see how `Mode.ts` does this). As an example, instead of passing `getCell` directly below for `getContent`, we'd update it to do `getCell()` and pass the record around instead

- [ ] Don't pass the cells every, only the "registry"

modules/tinymce/src/core/main/ts/Rtc.ts
TheSpyder: This is news to me, we haven't been made aware of this API or scheduled anything to manage it in 5.10.

lnewson: @TheSpyder this is for 6.0.0 at this stage (see the milestone), but I am surprised you haven't heard about it given this is a HV managed project...

TheSpyder: Ah maybe that's why. Can we get away without involving RTC in this content formatter? Looking at the `GetContentFormatter` API, custom formatters seem to be DOM specific?

lnewson: I don't know the details here, so tagging @lostkeys and @HAFRMO. If RTC don't need this though then I'm all for not doing this (I had assumed it was a requirement, probably incorrectly though).

TheSpyder: RTC has explicit implementations for tree, text, and HTML. Anything not tree or text produces HTML.

I don't see how we can make a content format API that works with RTC when the type is something like `(editor, args) => Content`. We'd need a public API that can abstract over the alt-model.

In theory `setContent` could run in front of RTC, since it's `content => content` but all a get content formatter could do is work from the ContentEditable DOM view (which isn't exactly what we render to a HTML string for get content).

If we want this to be RTC-compatible my suggestion is have both get and set formatters work from `AstNode` explicitly - RTC can deal with those.

modules/tinymce/src/core/test/ts/browser/EditorTest.ts
```diff
@@ -198,6 +198,64 @@ describe('browser.tinymce.core.EditorTest', () => {
...
 
+  it('TINY-7875: addContentFormatter', () => {
+    const editor = hook.editor();
+    let setCount = 0;
+    let getCount = 0;
+
+    const setFormatter = () => {
+      editor.setContent('set');
+      setCount++;
+      return 'set';
```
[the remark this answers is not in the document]
HAFRMO: The specification of the format say these should return a string, and it's used in events.

modules/tinymce/src/core/test/ts/browser/EditorTest.ts
```diff
@@ -198,6 +198,64 @@ describe('browser.tinymce.core.EditorTest', () => {
...
+    assert.equal(setCount, 1, 'test setContent, html getcontent, check setCount');
+    assert.equal(getCount, 0, 'test setContent, html getcontent, check getCount');
+
+    editor.setContent('<p>test3</p>');
+    assert.equal(editor.getContent({ format: 'test' }), 'get', 'html setContent, test getcontent, getContent');
+    assert.equal(setCount, 1, 'html setContent, test getcontent, check setCount');
+    assert.equal(getCount, 1, 'html setContent, test getcontent, check getCount');
+  });
+
+  it('TINY-7875: getContent events updates properly', () => {
```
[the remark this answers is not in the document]
HAFRMO: This is not specific for the custom formats, the events are fired in the same place, so this already covers it.

modules/tinymce/src/core/test/ts/browser/EditorTest.ts
```diff
@@ -198,6 +198,64 @@ describe('browser.tinymce.core.EditorTest', () => {
     assert.equal(editor.getContent(), '<!-- x --><p>\u00a0</p>', 'setContent with comment bug #4409');
   });
 
+  it('TINY-7875: addContentFormatter', () => {
+    const editor = hook.editor();
+    let setCount = 0;
+    let getCount = 0;
+
+    const setFormatter = () => {
+      editor.setContent('set');
```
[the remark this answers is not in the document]
HAFRMO: It's a test function meant to put a distinctive value into the editor for further in the test, that's all.

And we probably don't want to block the further events if it's used like this, since it will be events with different formats, which may be handled differently by those who listen.

And no we can't. If we don't set the content we can't confirm it's been set as expected, and if we don't return a value we don't fit the format type.

modules/tinymce/src/core/main/ts/content/GetContentImpl.ts
HAFRMO: Made the changes, though I think they feel a lot less safe than the old version, at least in the setContentImpl.

modules/tinymce/CHANGELOG.md
```diff
@@ -21,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Fixed an exception getting thrown when disabling events and setting content #TINY-7956
 - Delete operations could behave incorrectly if the selection crossed a table boundary #TINY-7596
+- Added types to `setContent` optional argument #TINY-7875
```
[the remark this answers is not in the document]
HAFRMO: I think it'd be better not to merge in the next-branch all the time, and handle that when we merge in the epic-branch.

modules/tinymce/src/core/main/ts/content/GetContentImpl.ts
```diff
@@ -15,52 +15,91 @@ import Tools from '../api/util/Tools';
...
-  if (updatedArgs.no_events) {
-    return updatedArgs.content;
-  } else {
+  return args.content;
+};
+
+const getFormatter = (format: string, getCell: Cell<Record<string, GetContentFormatter>>) => {
+  return Optional.from(getCell.get()[format]).fold(
+    () => {
+      // eslint-disable-next-line no-console
```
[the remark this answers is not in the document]
HAFRMO: It was, yes, but since we moved over to the getordie it's not needed.

modules/tinymce/src/core/main/ts/content/EditorContent.ts
HAFRMO: We strictly don't need the cell anywhere, as the value it contains ( the object ) is always the same. I was just under the impression that we want to toss around cells instead. If that isn't what we want to do, I can change it, though I'd like to make sure I've understood it correctly: You want to remove the `Cell<>`from around the record where possible?

lnewson: Yes that's correct. We should only be passing the cell where the "registry" can be modified. That should not be happening in the get/set content case as it should only be reading from the "registry", so we should only be passing the cell in the add format case.

modules/tinymce/CHANGELOG.md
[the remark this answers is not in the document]
lnewson: Neil wasn't talking about all the time, only this once as the Editor team setup the next branch to be ready for 6.0 work. This branch however is still based off 5.10, so I agree we should do this to make sure it's version, changelogs, etc... are correct.

modules/mcagar/src/main/ts/ephox/mcagar/alien/EditorTypes.ts
```diff
@@ -1,5 +1,8 @@
 type EventCallback = (event: any) => void;
 
+export type GetContentFormatter = (editor: Editor, args: GetContentArgs) => any;
+export type SetContentFormatter = (editor: Editor, content: any, args: SetContentArgs) => any;
+
```
lnewson, line 5:
```suggestion
export type GetContentFormatter = (editor: Editor, args: GetContentArgs) => any;
export type SetContentFormatter = (editor: Editor, content: any, args: SetContentArgs) => any;

```
This should be reverted as it's not being used anymore.

modules/tinymce/src/core/main/ts/content/GetContentImpl.ts
lnewson: Yeah, you've broken a few things here now, as the relevant events aren't firing for the tree implementation. You've also not split it properly either, as take `defaultContentFormatter` for example it still has a `args.format === 'tree'` check... so it probably needs to be something like this instead:

```
const defaultContentFormatter = (editor: Editor, args: GetContentArgs): string =>
  Optional.from(editor.getBody()).fold(
    Fun.constant(''),
    (body) => getContentFromBody(editor, args, body)
  );

const treeContentFormatter = (editor: Editor, args: GetContentArgs): AstNode =>
  Optional.from(editor.getBody()).fold(
    Fun.constant(new AstNode('body', 11)),
    (body) => getContentFromBody(editor, args, body)
  );
```

- [ ] Fix events not firing for the tree implementation

modules/tinymce/src/core/test/ts/browser/EditorTest.ts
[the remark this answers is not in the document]
JamesToohey: I've unresolved this.

I stand by this test needing to be rethought. We have an assertion in the `afterGetContent` function, which may not be called if there is a problem. If we want to explicitly check that this function was called, we should be doing something more robust.

We should have a test for these content events when there is a custom formatter, to ensure they're still fired correctly.

Should we also have a test to catch this bug that Lee pointed out?

https://github.com/tinymce/tinymce/pull/7170#discussion_r718942654

lostkeys:
Hey, everyone, It seems there are a numerous issues still that has to be sorted out, including the problems with alt-model support. This project has grown much bigger than anticipated and to avoid throwing good money after bad, the project are from here-on-hold until the Product team can reassess the project.

PR #7176, opened 2021-09-06
TINY-7596: Fix all contents of cell being deleted when only part of the contents were selected

JamesToohey:
**Related Ticket:** TINY-7596

**Description of Changes:**

When performing delete operations on selections that contain tables, we now perform a custom delete operation on the cells in the selection that _might_ have only part of their contents selected:

1. If the start of the selection is outside a table and the end of the selection is in a table, then the last cell in the selection
2. If the start of the selection is inside a table and the end is outside a table, then the first cell in that selection
3. If the start of the selection is in a table and the end of the selection is in another table, then the first cell of the selection in the first table and the last cell of the selection in the last table

This means we create a new `Range` inside the cell, delete that, and if the cell still has content inside we do not clean it up and add the bogus `<br>` element.

Any content outside the tables was already being deleted separately. This also appears to simplify the selection/cursor positioning logic.

I found a bug where if there is a nested table inside the cell then the outer paragraph is moved inside the cell, I'm not sure if we want to fix that now or log a separate bug, it appears to not be following any of the code paths inside `TableDelete.ts`.

Screen Shot 2021-09-06 at 1 27 37 pm

**Pre-checks:**
* [x] Changelog entry added

**Review:**
* [x] Milestone set
* [x] Document bug I found that already exists

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Fixed an exception getting thrown when disabling events and setting content #TINY-7956
-- Deleting content within selections that contained content outside a table and content within a cell could result in the entire contents of the cell being incorrectly deleted #TINY-7596
+- Delete operations could behave incorrectly if the selection crossed a table boundary #TINY-7596
 
 ## 5.9.1 - 2021-08-27
 
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+- Deleting content within selections that contained content outside a table and content within a cell could result in the entire contents of the cell being incorrectly deleted #TINY-7596
+
 ## 5.9.1 - 2021-08-27
 
 ### Fixed
--- a/modules/tinymce/src/core/main/ts/delete/TableDelete.ts
+++ b/modules/tinymce/src/core/main/ts/delete/TableDelete.ts
@@ -64,6 +64,17 @@ const deleteContentInsideCell = (cell: SugarElement<HTMLTableCellElement>, rng: 
   insideTableRng.deleteContents();
 };
 
+const collapseAndRestoreCellSelection = (editor: Editor) => {
+  const selectedCells = TableCellSelection.getCellsFromEditor(editor);
+  editor.selection.collapse(true);
+  // Restore the data-mce-selected attribute if multiple cells were selected, as if it was a cef element
+  // then selection overrides would remove it as it was using an offscreen selection clone.
+  if (selectedCells.length > 1) {
+    const firstCell = Arr.find(selectedCells, (cell) => Attribute.has(cell, 'data-mce-first-selected')).getOr(selectedCells[0]);
+    Attribute.set(firstCell, 'data-mce-selected', '1');
+  }
+};
+
 /*
  * Runs when
  * - the start and end of the selection is contained within the same table (called directly from deleteRange)
@@ -102,7 +113,7 @@ const emptySingleTableCells = (editor: Editor, cells: SugarElement<HTMLTableCell
   cleanCells(cellsToClean);
 
   // Collapse the original selection after deleting everything
-  editor.selection.collapse(true);
+  collapseAndRestoreCellSelection(editor);
   return true;
 };
 
@@ -132,7 +143,7 @@ const emptyMultiTableCells = (
   betweenRng.deleteContents();
 
   // This will collapse the selection into the cell of the start table
-  editor.selection.collapse(true);
+  collapseAndRestoreCellSelection(editor);
   return true;
 };
 
--- a/modules/tinymce/src/core/main/ts/delete/TableDelete.ts
+++ b/modules/tinymce/src/core/main/ts/delete/TableDelete.ts
@@ -6,7 +6,7 @@
  */
 
 import { Arr, Fun, Optional, Optionals } from '@ephox/katamari';
-import { Attribute, Compare, Remove, SugarElement, SugarNode, Traverse } from '@ephox/sugar';
+import { Attribute, Compare, Remove, SelectorExists, SugarElement, SugarNode, Traverse } from '@ephox/sugar';
 
 import Editor from '../api/Editor';
 import * as CaretFinder from '../caret/CaretFinder';
@@ -50,14 +50,18 @@ const handleEmptyBlock = (editor: Editor, startInTable: boolean, emptyBlock: Opt
   });
 };
 
-const setCursorInCell = (editor: Editor, cell: SugarElement<HTMLTableCellElement>) => {
-  const selectedCells = TableCellSelection.getCellsFromEditor(editor);
-  editor.selection.setCursorLocation(cell.dom, 0);
-  // Restore the data-mce-selected attribute if multiple cells were selected, as if it was a cef element
-  // then selection overrides would remove it as it was using an offscreen selection clone.
-  if (selectedCells.length > 1) {
-    Attribute.set(cell, 'data-mce-selected', '1');
+const deleteContentInsideCell = (cell: SugarElement<HTMLTableCellElement>, rng: Range, isFirstCellInSelection: boolean) => {
+  const insideTableRng = rng.cloneRange();
+
+  if (isFirstCellInSelection) {
+    insideTableRng.setStart(rng.startContainer, rng.startOffset);
+    insideTableRng.setEndAfter(cell.dom.lastChild);
+  } else {
+    insideTableRng.setStart(cell.dom.firstChild, 0);
+    insideTableRng.setEnd(rng.endContainer, rng.endOffset);
   }
+
+  insideTableRng.deleteContents();
 };
 
 /*
@@ -66,8 +70,24 @@ const setCursorInCell = (editor: Editor, cell: SugarElement<HTMLTableCellElement
  * - part of a table and content outside is selected
  */
 const emptySingleTableCells = (editor: Editor, cells: SugarElement<HTMLTableCellElement>[], outsideDetails: Optional<OutsideTableDetails>): boolean => {
-  // Remove content from selected cells
-  cleanCells(cells);
+  const rng = editor.selection.getRng();
+  const startInTable = SelectorExists.ancestor(SugarElement.fromDom(rng.startContainer), 'table');
+
+  let cellsToClean = cells;
+  // The only time we can have only part of the cell contents selected is when part of the selection
+  // is outside the table (otherwise we use the Darwin fake selection, which always selects entire cells),
+  // in which case we need to delete the contents inside and check if the entire contents of the cell have been deleted.
+  outsideDetails.each(() => {
+    // The endPointCell is the only cell which may have only part of its contents selected.
+    const endPointCell = startInTable ? cells[0] : cells[cells.length - 1];
+    deleteContentInsideCell(endPointCell, rng, startInTable);
+    if (!editor.dom.isEmpty(endPointCell.dom)) {
+      cellsToClean = startInTable ? cellsToClean.slice(1) : cellsToClean.slice(0, cellsToClean.length - 1);
+    }
+  });
+
+  // Remove content from cells we need to clean
+  cleanCells(cellsToClean);
 
   // Delete all content outside of the table that is in the selection
   outsideDetails.map(({ rng, isStartInTable }) => {
@@ -78,15 +98,8 @@ const emptySingleTableCells = (editor: Editor, cells: SugarElement<HTMLTableCell
     handleEmptyBlock(editor, isStartInTable, outsideBlock.filter(Empty.isEmpty));
   });
 
-  // Set the selection:
-  // - to the first emptied cell if the start of the previous selection was inside a table
-  // - otherwise just collapse the previous selection that started in the outside block
-  if (outsideDetails.forall((details) => details.isStartInTable)) {
-    setCursorInCell(editor, cells[0]);
-  } else {
-    editor.selection.collapse(true);
-  }
-
+  // Collapse the original selection after deleting everything
+  editor.selection.collapse(true);
   return true;
 };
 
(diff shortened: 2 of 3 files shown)


modules/tinymce/CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+- Deleting content within selections that contained content outside a table and content within a cell could result in the entire contents of the cell being incorrectly deleted #TINY-7596
```
JamesToohey, line 10: I need help with explaining this bug in shortform changelog format

modules/tinymce/src/core/main/ts/delete/TableDelete.ts
```diff
@@ -50,14 +50,18 @@ const handleEmptyBlock = (editor: Editor, startInTable: boolean, emptyBlock: Opt
   });
 };
 
-const setCursorInCell = (editor: Editor, cell: SugarElement<HTMLTableCellElement>) => {
-  const selectedCells = TableCellSelection.getCellsFromEditor(editor);
-  editor.selection.setCursorLocation(cell.dom, 0);
-  // Restore the data-mce-selected attribute if multiple cells were selected, as if it was a cef element
-  // then selection overrides would remove it as it was using an offscreen selection clone.
-  if (selectedCells.length > 1) {
-    Attribute.set(cell, 'data-mce-selected', '1');
```
lnewson, line 59: This has just been lost and that's absolutely needed, so this needs reverting to some extent.

modules/tinymce/CHANGELOG.md
```suggestion
- Deleting content within selections that contained content outside a table and content within a cell could result in the entire contents of the cell getting deleted #TINY-7596
```
Maybe?

lnewson (changes requested):

modules/tinymce/src/core/main/ts/delete/TableDelete.ts
```diff
@@ -66,8 +70,24 @@ const setCursorInCell = (editor: Editor, cell: SugarElement<HTMLTableCellElement
  * - part of a table and content outside is selected
  */
 const emptySingleTableCells = (editor: Editor, cells: SugarElement<HTMLTableCellElement>[], outsideDetails: Optional<OutsideTableDetails>): boolean => {
-  // Remove content from selected cells
-  cleanCells(cells);
+  const rng = editor.selection.getRng();
+  const startInTable = SelectorExists.ancestor(SugarElement.fromDom(rng.startContainer), 'table');
```
lnewson, line 74: This shouldn't be done... we should be using the `isStartInTable` property from `outsideDetails`.

modules/tinymce/src/core/main/ts/delete/TableDelete.ts
JamesToohey: I can't find where this is needed, nor where `data-mce-selected` is ever removed from the selected cells

modules/tinymce/src/core/main/ts/delete/TableDelete.ts
JamesToohey: Nice I missed that being available

modules/tinymce/src/core/main/ts/delete/TableDelete.ts
```diff
@@ -66,8 +70,24 @@ const setCursorInCell = (editor: Editor, cell: SugarElement<HTMLTableCellElement
...
+
+  let cellsToClean = cells;
```
lnewson, line 76: Instead of the mutation here, this could just be returned via the below function instead. You can simplify the 
"end" slice a little. e.g:
```ts
const cellsToClean = outsideDetails.bind(({ rng, isStartInTable }) => {
    // The endPointCell is the only cell which may have only part of its contents selected.
    const endPointCell = startInTable ? cells[0] : cells[cells.length - 1];
    deleteContentInsideCell(endPointCell, rng, startInTable);
    if (!editor.dom.isEmpty(endPointCell.dom)) {
      return Optional.some(startInTable ? cells.slice(1) : cells.slice(0, -1));
    } else {
      return Optional.none();
    }
}).getOr(cells);
```

modules/tinymce/src/core/main/ts/delete/TableDelete.ts
lnewson: This is needed because when you select a cef element `SelectionOverrides` makes a clone and adds this attribute, however the fake table selection also adds it. As such when we collapse the selection, the `SelectionOverrides` code can remove this attribute here: https://github.com/tinymce/tinymce/blob/8ee937c891eabef49cb86cce605680b913915346/modules/tinymce/src/core/main/ts/SelectionOverrides.ts#L458. As such, this code needs to ensure that it's retained after placing the selection when more than one cell has been selected.

modules/tinymce/src/core/main/ts/delete/TableDelete.ts
```diff
@@ -99,11 +112,30 @@ const emptyMultiTableCells = (
...
+
+  // The cursor is always collapsed back into the start cell, so we never need to clean it
+  const startTableCellsToClean = startTableCells.slice(1);
+
+  let endTableCellsToClean = endTableCells;
+  const lastCell = endTableCells[endTableCells.length - 1];
+  // Only clean empty cells, and the last cell has the potential to still have content
+  if (!editor.dom.isEmpty(lastCell.dom)) {
+    endTableCellsToClean = endTableCells.slice(0, endTableCells.length - 1);
+  }
```
lnewson, line 131:
```suggestion
  const lastCell = endTableCells[endTableCells.length - 1];
  // Only clean empty cells, and the last cell has the potential to still have content
  const endTableCellsToClean = editor.dom.isEmpty(lastCell.dom) ? endTableCells : endTableCells.slice(0, -1);
```
This removes the need for the mutation.

modules/tinymce/src/core/main/ts/delete/TableDelete.ts
```diff
@@ -99,11 +112,30 @@ const emptyMultiTableCells = (
...
+
+  cleanCells(startTableCellsToClean.concat(endTableCellsToClean));
   // Delete all content in between the start table and end table
   betweenRng.deleteContents();
-  // Set the cursor back to the start of the original selection
-  setCursorInCell(editor, startTableCells[0]);
+
+  // This will collapse the selection into the cell of the start table
+  editor.selection.collapse(true);
```
lnewson, line 138: How do we know for certain the selection is in the start of the table? That is why we always manually set it previously, as it's relying on the native selection being moved/retained as you delete content, which can be very brittle. I'd much prefer us to be manually setting this, both here and in `emptySingleTableCells`.

JamesToohey: This function is only called when the selection start and end are in different tables. We can manually store the start of the selection before we do any processing and then set it again?

Part of the problem with `emptySingleTableCells` was that it was snapping the selection back into the cell instead of allowing it to leave the table when the start of the selection wasn't in the table previously

modules/tinymce/src/core/main/ts/delete/TableDelete.ts
JamesToohey: Whoops

modules/tinymce/src/core/main/ts/delete/TableDelete.ts
JamesToohey: The `rng` that we get from `outsideDetails` is only the range outside the table, so we can't use it when deleting contents inside a cell, but I can still just grab the Editor range and do it all inside a `bind`

JamesToohey:
Converting to draft while I work on a fix

modules/tinymce/src/core/main/ts/delete/TableDelete.ts
JamesToohey: We need to continue this discussion because even though I've gone back to similar logic as before, it is still collapsing

modules/tinymce/src/core/main/ts/delete/TableDelete.ts
```diff
@@ -50,13 +50,28 @@ const handleEmptyBlock = (editor: Editor, startInTable: boolean, emptyBlock: Opt
...
 
-const setCursorInCell = (editor: Editor, cell: SugarElement<HTMLTableCellElement>) => {
+const deleteContentInsideCell = (cell: SugarElement<HTMLTableCellElement>, rng: Range, isFirstCellInSelection: boolean) => {
+  const insideTableRng = rng.cloneRange();
+
+  if (isFirstCellInSelection) {
+    insideTableRng.setStart(rng.startContainer, rng.startOffset);
+    insideTableRng.setEndAfter(cell.dom.lastChild);
+  } else {
+    insideTableRng.setStart(cell.dom.firstChild, 0);
```
ashfordneil, line 60:
```suggestion
    insideTableRng.setStartBefore(cell.dom.firstChild);
```

I feel like these two branches should mirror each other. However, it _might_ not make a difference in terms of functionality once you start deleting things?

modules/tinymce/src/core/test/ts/browser/delete/TableDeleteTest.ts
```diff
@@ -444,6 +463,55 @@ describe('browser.tinymce.core.delete.TableDeleteTest', () => {
...
+      TinySelections.setSelection(editor, [ 0, 0, 1, 0, 0 ], 1, [ 2, 0, 0, 1, 0 ], 3);
+      doDelete(editor);
+      TinyAssertions.assertCursor(editor, [ 0, 0, 1, 0, 0 ], 1);
+      TinyAssertions.assertContent(
+        editor,
+        '<table><tbody><tr><td>a</td><td>b</td></tr><tr><td>c</td><td>&nbsp;</td></tr></tbody></table>' +
+        '<table><tbody><tr><td>&nbsp;</td><td>f</td></tr><tr><td>g</td><td>h</td></tr></tbody></table>'
+      );
+    });
   });
```
ashfordneil, line 515: Do you need a test like this?

|a|b123|
|-|-|
|c|d|

|e|f|
|-|-|
|456g|h|

to assert whether the selected rows are completely deleted or just have all their content removed?

ashfordneil (commented):
LGTM but as (1) I was helping pair with you a bit and (2) I'm not sure what else might be going on here, I'll leave this for someone else to approve.

modules/tinymce/CHANGELOG.md
JamesToohey: I like your second one, although it's more 'incorrect' content because it would also not delete any of the content outside the table. Anyone else have any thoughts?

modules/tinymce/src/core/test/ts/browser/delete/TableDeleteTest.ts
JamesToohey: No harm in adding that (if it doesn't already exist), will do

modules/tinymce/src/core/main/ts/delete/TableDelete.ts
```diff
@@ -50,13 +50,28 @@ const handleEmptyBlock = (editor: Editor, startInTable: boolean, emptyBlock: Opt
...
+
+const collapseAndRestoreCellSelection = (editor: Editor) => {
   const selectedCells = TableCellSelection.getCellsFromEditor(editor);
-  editor.selection.setCursorLocation(cell.dom, 0);
+  editor.selection.collapse(true);
   // Restore the data-mce-selected attribute if multiple cells were selected, as if it was a cef element
   // then selection overrides would remove it as it was using an offscreen selection clone.
   if (selectedCells.length > 1) {
-    Attribute.set(cell, 'data-mce-selected', '1');
+    const firstCell = Arr.find(selectedCells, (cell) => Attribute.has(cell, 'data-mce-first-selected')).getOr(selectedCells[0]);
```
lnewson, line 73: I mentioned this yesterday in passing, but we can't use `data-mce-first-selected` in core, as that's a table plugin specific attribute.

- [ ] Refactor this so that we don't use `data-mce-first-selected` or `data-mce-last-selected`.

lnewson: We're probably going to have to use where the selection is after collapsing and then restore based on which cell the selection is in/on.

modules/tinymce/src/core/main/ts/delete/TableDelete.ts
```diff
@@ -66,27 +81,39 @@ const setCursorInCell = (editor: Editor, cell: SugarElement<HTMLTableCellElement
  * - part of a table and content outside is selected
  */
 const emptySingleTableCells = (editor: Editor, cells: SugarElement<HTMLTableCellElement>[], outsideDetails: Optional<OutsideTableDetails>): boolean => {
-  // Remove content from selected cells
-  cleanCells(cells);
+  const editorRng = editor.selection.getRng();
+  const cellsToClean = outsideDetails.bind(({ rng, isStartInTable }) => {
+    /**
+     * Delete all content outside of the table that is in the selection
+     */
```
lnewson, line 88:
```suggestion
    // Delete all content outside of the table that is in the selection
```
This doesn't need to be a multiline comment. If it was meant to be one for visibility, then it shouldn't be an API docs comment (i.e shouldn't have `/**`)

lnewson: Actually, maybe this should be outside this/above the `const cellsToClean ...` part?

modules/tinymce/src/core/main/ts/delete/TableDelete.ts
```diff
@@ -66,27 +81,39 @@ const setCursorInCell = (editor: Editor, cell: SugarElement<HTMLTableCellElement
...
 
-  // Set the selection:
-  // - to the first emptied cell if the start of the previous selection was inside a table
-  // - otherwise just collapse the previous selection that started in the outside block
-  if (outsideDetails.forall((details) => details.isStartInTable)) {
-    setCursorInCell(editor, cells[0]);
-  } else {
-    editor.selection.collapse(true);
-  }
+    /**
```
lnewson, line 96:
```suggestion
    /*
```
As above, 2 asterisks are for API docs comments, a single one is just for a general multiline comment.

modules/tinymce/src/core/main/ts/delete/TableDelete.ts
```diff
@@ -99,11 +126,24 @@ const emptyMultiTableCells = (
...
+
+  const startCell = startTableCells[0];
+  const endCell = endTableCells[endTableCells.length - 1];
+
+  deleteContentInsideCell(startCell, rng, true);
+  deleteContentInsideCell(endCell, rng, false);
+
+  // Only clean empty cells, the first and last cells have the potential to still have content
+  const startTableCellsToClean = editor.dom.isEmpty(startCell.dom) ? startTableCells : startTableCells.slice(1);
+  const endTableCellsToClean = editor.dom.isEmpty(endCell.dom) ? endTableCells : endTableCells.slice(0, -1);
```
lnewson, line 139: Should this be using `Empty.isEmpty` instead? That's what we use above in `emptySingleTableCells`.
```suggestion
  const startTableCellsToClean = Empty.isEmpty(startCell.dom) ? startTableCells : startTableCells.slice(1);
  const endTableCellsToClean = Empty.isEmpty(endCell) ? endTableCells : endTableCells.slice(0, -1);
```

modules/tinymce/src/core/main/ts/delete/TableDelete.ts
```diff
@@ -66,27 +81,39 @@ const setCursorInCell = (editor: Editor, cell: SugarElement<HTMLTableCellElement
...
+    /**
+     * The only time we can have only part of the cell contents selected is when part of the selection
+     * is outside the table (otherwise we use the Darwin fake selection, which always selects entire cells),
+     * in which case we need to delete the contents inside and check if the entire contents of the cell have been deleted.
+     */
+
+    // The endPointCell is the only cell which may have only part of its contents selected.
+    const endPointCell = isStartInTable ? cells[0] : cells[cells.length - 1];
+    deleteContentInsideCell(endPointCell, editorRng, isStartInTable);
+    if (!editor.dom.isEmpty(endPointCell.dom)) {
```
lnewson, line 105:
```suggestion
    if (!Empty.isEmpty(endPointCell)) {
```

modules/tinymce/src/core/main/ts/delete/TableDelete.ts
JamesToohey: I did this because there are two separate chunks of logic in here, happy to change if anyone objects

lnewson: I don't like that this seems to conflict with the comments below then? Can we merge them?

lnewson (approved):

modules/tinymce/CHANGELOG.md
JamesToohey:
```suggestion
- Delete operations could behave incorrectly if the selection crossed a table boundary #TINY-7596
```

ltrouton (approved):

PR #7187, opened 2021-09-10
TINY-7062: Fix editor sometimes scrolling when clicking beside non-editable element

### LINKED ISSUE #6456: Cursor position jump to the start of the editor when dealing with content non-editable elements
Cursor always jumps to the top of the editor when clicked left or right of the editor body (when padding applied)

**Please provide the steps to reproduce and if possible a minimal demo of the problem via [fiddle.tiny.cloud](https://fiddle.tiny.cloud/) or similar.**
1. Have left & right padding set for the editor
2. Place a container which is made up with non-editable elements
image
3. Place cursor inside the editable region of container
4. Click left or right of container
5. Notice that the scroll position is lost and editor is scrolled all the way to top.

Scroll position retained, cursor is placed next to container

**Which versions of TinyMCE, and which browser / OS are affected by this issue? Did this work in previous versions of TinyMCE?
Observed in 4 & 5 major versions, even in latest using Chrome or Edge
scroll issue

ltrouton:
Related Ticket: TINY-7062

Description of Changes:
* Focus body after setting range. Chrome would scroll the editor when calling focus before setting the range when clicking from inside an editable element to beside a non-editable element

Pre-checks:
* [X] Changelog entry added

Review:
* [X] Milestone set

Fixes #6456

### DIFF
--- a/modules/tinymce/src/core/test/ts/browser/SelectionOverridesTest.ts
+++ b/modules/tinymce/src/core/test/ts/browser/SelectionOverridesTest.ts
@@ -29,18 +29,20 @@ describe('browser.tinymce.core.SelectionOverridesTest', () => {
     assert.equal(selectedNode.getAttribute('data-mce-caret'), caretValue);
   };
 
-  const selectBesideContentEditable = (editor: Editor, contentEditableElm: HTMLElement, clickBefore: boolean) => {
+  const selectBesideContentEditable = (editor: Editor, contentEditableElm: HTMLElement, clickPoint: 'before' | 'after') => {
     editor.selection.scrollIntoView(contentEditableElm);
 
     const scrollTop = getScrollTop(editor);
-    const bodyPaddingOffset = 8; // Default is 16 so divide by 2
+    const bodyMarginOffset = 8; // Default is 16 so divide by 2
     const rect = contentEditableElm.getBoundingClientRect();
-    const clientX = clickBefore ? rect.left - bodyPaddingOffset : rect.right + bodyPaddingOffset;
+    const clientX = clickPoint === 'before' ? rect.left - bodyMarginOffset : rect.right + bodyMarginOffset;
     const clientY = rect.top + (rect.height / 2);
 
     Mouse.point('mousedown', 0, TinyDom.documentElement(editor), clientX, clientY);
     // Check the scoll position has not changed
     assert.equal(getScrollTop(editor), scrollTop);
+    // Check fake caret has been added
+    assertSelectionIsCaretBlock(editor, clickPoint);
   };
 
   it('click on link in cE=false', () => {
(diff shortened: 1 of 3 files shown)


modules/tinymce/src/core/test/ts/browser/SelectionOverridesTest.ts
```diff
@@ -15,9 +17,32 @@ describe('browser.tinymce.core.SelectionOverridesTest', () => {
...
 
+  const getScrollTop = (editor: Editor) => Scroll.get(TinyDom.document(editor)).top;
+
+  const assertSelectionIsCaretBlock = (editor: Editor, caretValue: 'before' | 'after') => {
+    const selectedNode = editor.selection.getNode();
+    assert.isTrue(isCaretContainerBlock(selectedNode));
+    assert.equal(selectedNode.getAttribute('data-mce-caret'), caretValue);
+  };
+
+  const selectBesideContentEditable = (editor: Editor, contentEditableElm: HTMLElement, clickBefore: boolean) => {
```
ashfordneil, line 32: Maybe instead of a boolean this could be `'left' | 'right'`?

modules/tinymce/src/core/test/ts/browser/SelectionOverridesTest.ts
```diff
@@ -15,9 +17,32 @@ describe('browser.tinymce.core.SelectionOverridesTest', () => {
...
+    editor.selection.scrollIntoView(contentEditableElm);
+
+    const scrollTop = getScrollTop(editor);
+    const bodyPaddingOffset = 8; // Default is 16 so divide by 2
```
ashfordneil, line 36: Should we call this `bodyMarginOffset` just so that it makes a bit more sense with the CSS?

ashfordneil (approved):

modules/tinymce/src/core/test/ts/browser/SelectionOverridesTest.ts
ltrouton: Yep, forgot to change the variable name. Thanks.

modules/tinymce/src/core/test/ts/browser/SelectionOverridesTest.ts
ltrouton: Have updated the function to use`'before' | 'after'` and do the cart block check as well

modules/tinymce/src/core/test/ts/browser/SelectionOverridesTest.ts
```diff
@@ -15,9 +17,34 @@ describe('browser.tinymce.core.SelectionOverridesTest', () => {
...
+    editor.selection.scrollIntoView(contentEditableElm);
+
+    const scrollTop = getScrollTop(editor);
+    const bodyMarginOffset = 8; // Default is 16 so divide by 2
+    const rect = contentEditableElm.getBoundingClientRect();
+    const clientX = clickPoint === 'before' ? rect.left - bodyMarginOffset : rect.right + bodyMarginOffset;
+    const clientY = rect.top + (rect.height / 2);
+
+    Mouse.point('mousedown', 0, TinyDom.documentElement(editor), clientX, clientY);
+    // Check the scoll position has not changed
```
lnewson, line 42:
```suggestion
    // Check the scroll position has not changed
```

lnewson (approved):

spocke (approved):

JamesToohey (approved):

PR #7191, opened 2021-09-13
TINY-7867: Improve Optional and Result in Katamari

ashfordneil:
Related Ticket: TINY-7867

Description of Changes (all in Katamari):
* Rewrite `Optional.ts` to use a class, so constructing Optionals is faster
* Rewrite `Result.ts` to be exactly the same functionally, but cleaner and way more documented
* Add an `isPlainObject` utility to `Type.ts` as opposed to `isObject`
* Update the deep merge function to check `isPlainObject` instead of `isObject`

* [x] Change target branch to TinyMCE 6.0

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/katamari/CHANGELOG.md
+++ b/modules/katamari/CHANGELOG.md
@@ -7,12 +7,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## Unreleased
 
 ### Changed
-- Swapped `Optional` to be a class, instead of an object with function fields
+- Swapped `Optional` to be a class, instead of an object with function fields.
 
 ### Improved
-- `Optional` objects should now be much faster to construct and work with
-- The `Result` and `Optional` APIs are now thoroughly documented
-- Added private `.tag` and `.value` fields to `Result`, to make them more friendly to view in the debugger or console
+- `Optional` objects should now be faster to construct and work with.
+- The `Result` and `Optional` APIs are now thoroughly documented.
+- Added private `.tag` and `.value` properties to `Result`, to make them friendlier to view in the debugger or console.
 
 ## 8.0.0 - 2021-08-26
 
--- a/modules/katamari/CHANGELOG.md
+++ b/modules/katamari/CHANGELOG.md
@@ -14,6 +14,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - The `Result` and `Optional` APIs are now thoroughly documented.
 - Added private `.tag` and `.value` properties to `Result`, to make them friendlier to view in the debugger or console.
 
+## 8.0.1 - TBD
+
 ## 8.0.0 - 2021-08-26
 
 ### Added
--- a/modules/katamari/CHANGELOG.md
+++ b/modules/katamari/CHANGELOG.md
@@ -6,6 +6,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
 
+### Changed
+- Swapped `Optional` to be a class, instead of an object with function fields
+
+### Improved
+- `Optional` objects should now be much faster to construct and work with
+- The `Result` and `Optional` APIs are now thoroughly documented
+- Added private `.tag` and `.value` fields to `Result`, to make them more friendly to view in the debugger or console
+
 ## 8.0.0 - 2021-08-26
 
 ### Added
--- a/modules/katamari/src/main/ts/ephox/katamari/api/Optional.ts
+++ b/modules/katamari/src/main/ts/ephox/katamari/api/Optional.ts
@@ -5,9 +5,7 @@ import * as Type from './Type';
  * not exist. Any `Optional<T>` can either be a `Some<T>` (in which case the
  * value does exist) or a `None` (in which case the value does not exist). This
  * file defines a whole lot of FP-inspired utility functions for dealing with
- * `Optional` objects. Note: Functions are grouped based on, primarily, their
- * Haskell roots. That said, I'm not a Haskell programmer so it might be
- * imperfect.
+ * `Optional` objects.
  *
  * Comparison with null or undefined:
  * - We don't get fancy null coalescing operators with `Optional`
@@ -18,7 +16,6 @@ import * as Type from './Type';
  * strict-null-checks
  * - Try to use `Optional` instead of null or undefined where you can
  */
-
 export class Optional<T> {
   private readonly tag: boolean;
   private readonly value?: T;
@@ -84,7 +81,7 @@ export class Optional<T> {
     return !this.tag;
   }
 
-  // --- Functor ---
+  // --- Functor (name stolen from Haskell / maths) ---
 
   /**
    * Perform a transform on an `Optional` object, **if** there is a value. If
@@ -102,7 +99,7 @@ export class Optional<T> {
     }
   }
 
-  // --- Monad ---
+  // --- Monad (name stolen from Haskell / maths) ---
 
   /**
    * Perform a transform on an `Optional` object, **if** there is a value.
@@ -117,7 +114,7 @@ export class Optional<T> {
     }
   }
 
-  // --- Traversable ---
+  // --- Traversable (name stolen from Haskell / maths) ---
 
   /**
    * For a given predicate, this function finds out if there **exists** a value
@@ -195,8 +192,8 @@ export class Optional<T> {
    * value.
    *
    * Unlike `or`, in this method the `replacement` value is "thunked" - that is
-   * to say that you don't pass a value to `getOrThunk`, you pass a function
-   * which (if called) will **return** the `value` you want to use.
+   * to say that you don't pass a value to `orThunk`, you pass a function which
+   * (if called) will **return** the `value` you want to use.
    *
    * Unlike `getOrThunk`, in this method the `replacement` value is also
    * `Optional`, meaning that this method will always return an `Optional`.
@@ -238,17 +235,19 @@ export class Optional<T> {
   }
 
   /**
-   * Converts an `Optional` to a nullable type.
+   * Converts an `Optional` to a nullable type, by getting the value if it
+   * exists, or returning `null` if it does not.
    */
   public getOrNull(): T | null {
     return this.tag ? this.value as T : null;
   }
 
   /**
-   * Converts an `Optional` to an undefined-able type.
+   * Converts an `Optional` to an undefined-able type, by getting the value if
+   * it exists, or returning `undefined` if it does not.
    */
   public getOrUndefined(): T | undefined {
-    return this.tag ? this.value as T : undefined;
+    return this.value;
   }
 
   // --- Utilities ---
@@ -262,7 +261,7 @@ export class Optional<T> {
    * doing anything in the case where the `Optional` doesn't have a value inside
    * it. If you're not sure whether your use-case fits into transforming
    * **into** something or **doing** something, check whether it has a return
-   * value. If it does, you're looking at a transform.
+   * value. If it does, you should be performing a transform.
    */
   public each(worker: (value: T) => void): void {
     if (this.tag) {
--- a/modules/katamari/src/main/ts/ephox/katamari/api/Optional.ts
+++ b/modules/katamari/src/main/ts/ephox/katamari/api/Optional.ts
@@ -14,7 +14,6 @@ import * as Type from './Type';
  * another `Optional`)
  * - There is no option to turn off strict-optional-checks like there is for
  * strict-null-checks
- * - Try to use `Optional` instead of null or undefined where you can
  */
 export class Optional<T> {
   private readonly tag: boolean;
--- a/modules/katamari/src/main/ts/ephox/katamari/api/Optional.ts
+++ b/modules/katamari/src/main/ts/ephox/katamari/api/Optional.ts
@@ -4,7 +4,7 @@ import * as Type from './Type';
  * The `Optional` type represents a value (of any type) that potentially does
  * not exist. Any `Optional<T>` can either be a `Some<T>` (in which case the
  * value does exist) or a `None` (in which case the value does not exist). This
- * file defines a whole lot of FP-inspired utility functions for dealing with
+ * module defines a whole lot of FP-inspired utility functions for dealing with
  * `Optional` objects.
  *
  * Comparison with null or undefined:
@@ -24,7 +24,7 @@ export class Optional<T> {
   // reuse the same object
   private static singletonNone = new Optional<any>(false);
 
-  // The internal representation has an `tag` and a `value`, but both are
+  // The internal representation has a `tag` and a `value`, but both are
   // private: able to be console.logged, but not able to be accessed by code
   private constructor(tag: boolean, value?: T) {
     this.tag = tag;
@@ -103,8 +103,7 @@ export class Optional<T> {
 
   /**
    * Perform a transform on an `Optional` object, **if** there is a value.
-   * Unlike the map function earlier in the piece, here the transform itself
-   * also returns an `Optional`.
+   * Unlike `map`, here the transform itself also returns an `Optional`.
    */
   public bind<U>(binder: (value: T) => Optional<U>): Optional<U> {
     if (this.tag) {
@@ -119,8 +118,8 @@ export class Optional<T> {
   /**
    * For a given predicate, this function finds out if there **exists** a value
    * inside this `Optional` object that meets the predicate. In practice, this
-   * means that for empty `Optional`s it returns false (as no predicate-
-   * meeting value exists).
+   * means that for `Optional`s that do not contain a value it returns false (as
+   * no predicate-meeting value exists).
    */
   public exists(predicate: (value: T) => boolean): boolean {
     return this.tag && predicate(this.value as T);
@@ -129,8 +128,8 @@ export class Optional<T> {
   /**
    * For a given predicate, this function finds out if **all** the values inside
    * this `Optional` object meet the predicate. In practice, this means that
-   * for empty `Optional`s it returns true (as all 0 objects do meet the
-   * predicate).
+   * for `Optional`s that do not contain a value it returns true (as all 0
+   * objects do meet the predicate).
    */
   public forall(predicate: (value: T) => boolean): boolean {
     return !this.tag || predicate(this.value as T);
--- a/modules/katamari/src/main/ts/ephox/katamari/api/Optional.ts
+++ b/modules/katamari/src/main/ts/ephox/katamari/api/Optional.ts
@@ -1,141 +1,291 @@
-import * as Fun from './Fun';
+import * as Type from './Type';
 
-/** A value which may be Optional.some(a) or Optional.none().
- *  Useful for representing partiality or pass/fail conditions that produce a value when successful.
- *  Sometimes called a "type-safe alternative to null".
- *  Can be thought of as a list of exactly zero or 1 elements.
+/**
+ * The `Optional` type represents a value (of any type) that potentially does
+ * not exist. Any `Optional<T>` can either be a `Some<T>` (in which case the
+ * value does exist) or a `None` (in which case the value does not exist). This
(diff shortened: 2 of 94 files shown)


modules/katamari/src/main/ts/ephox/katamari/api/Optional.ts
```diff
@@ -1,141 +1,291 @@
...
+  /**
+   * If the `Optional` contains a value, perform an action on that value.
+   * Unlike the rest of the methods on this type, `.each` has side-effects. If
+   * you want to transform an `Optional<T>` **into** something, then this is not
+   * the method for you. If you want to use an `Optional<T>` to **do**
+   * something, then this is the method for you - provided you're okay with not
+   * doing anything in the case where the `Optional` doesn't have a value inside
+   * it. If you're not sure whether your use-case fits into transforming
+   * **into** something or **doing** something, check whether it has a return
+   * value. If it does, you're looking at a transform.
```
JamesToohey, line 265:
```suggestion
   * value. If it does, you should be performing a transform.
```

modules/katamari/src/main/ts/ephox/katamari/api/Optional.ts
```diff
@@ -1,141 +1,291 @@
...
+   */
+  public getOrNull(): T | null {
+    return this.tag ? this.value as T : null;
+  }
+
+  /**
+   * Converts an `Optional` to an undefined-able type.
+   */
+  public getOrUndefined(): T | undefined {
+    return this.tag ? this.value as T : undefined;
```
tiny-james, line 251: Could this be just `return this.value;`?

tiny-james (approved):

modules/katamari/src/main/ts/ephox/katamari/api/Optional.ts
```diff
@@ -1,141 +1,291 @@
...
- *  Sometimes called a "type-safe alternative to null".
- *  Can be thought of as a list of exactly zero or 1 elements.
+/**
+ * The `Optional` type represents a value (of any type) that potentially does
+ * not exist. Any `Optional<T>` can either be a `Some<T>` (in which case the
+ * value does exist) or a `None` (in which case the value does not exist). This
+ * file defines a whole lot of FP-inspired utility functions for dealing with
+ * `Optional` objects. Note: Functions are grouped based on, primarily, their
+ * Haskell roots. That said, I'm not a Haskell programmer so it might be
+ * imperfect.
```
lnewson, line 10:
```suggestion
 * `Optional` objects.
 *
 * Note: Functions are grouped based on, primarily, their Haskell roots.
```
Please no personal `I'm` in this type of thing, as it's not clear who that refers to and really doesn't make sense to be in a doc string.

lnewson: I also wonder if the note should even be in this docstring at all, as it feels like it's a comment specific to the file, not to the `Optional` API.

modules/katamari/CHANGELOG.md
```diff
@@ -6,6 +6,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
 
+### Changed
+- Swapped `Optional` to be a class, instead of an object with function fields
+
+### Improved
+- `Optional` objects should now be much faster to construct and work with
+- The `Result` and `Optional` APIs are now thoroughly documented
+- Added private `.tag` and `.value` fields to `Result`, to make them more friendly to view in the debugger or console
```
lnewson, line 15:
```suggestion
- Swapped `Optional` to be a class, instead of an object with function fields.

### Improved
- `Optional` objects should now be faster to construct and work with.
- The `Result` and `Optional` APIs are now thoroughly documented.
- Added private `.tag` and `.value` properties to `Result`, to make them friendlier to view in the debugger or console.
```
Missing full stops and `much` is very subjective so we shouldn't be including that. In this case we're talking about ~1ms IIRC so I wouldn't call that much faster in the overall scheme of things.

modules/katamari/src/main/ts/ephox/katamari/api/Type.ts
```diff
@@ -20,12 +20,34 @@ const isSimpleType = <Yolo>(type: string) => (value: any): value is Yolo =>
 const eq = <T> (t: T) => (a: any): a is T =>
   t === a;
 
+const getPrototypeOf = Object.getPrototypeOf;
```
lnewson, line 23: This should be at the top of the file, not in the middle like this.

modules/katamari/src/main/ts/ephox/katamari/api/Type.ts
```diff
@@ -20,12 +20,34 @@ const isSimpleType = <Yolo>(type: string) => (value: any): value is Yolo =>
...
 
 export const isObject: (value: any) => value is Object =
   isType('object');
 
+export const isPlainObject = (value: unknown): value is Object => {
+  if (!isObject(value)) {
+    return false;
+  }
+
+  const constructor = value.constructor?.name;
```
lnewson, line 36: Should this type of thing be pulled out into a reusable function, as we're already doing this type of check with `String` and `Array` in `typeOf`? This does cover more cases as well, which are likely not working on IE 11.

lnewson: Actually, I was just thinking about this more... maybe IE keeps the same references across windows, as otherwise the `String` and `Array` tests should be failing. If that's the case, then we can likely skip the to string checks below :shrug: 

Either way, given this is being used multiple times and across multiple functions I definitely think we should be looking to pull this out into a separate function.

ashfordneil: I've done an experiment related to this, and nowI want to see if the tests still pass in all browsers / OSes with this new thing. Either I'll pull the checks into a separate function or I'll remove them, depending on how Jenkins feels

modules/katamari/src/main/ts/ephox/katamari/api/Optional.ts
```diff
@@ -1,141 +1,291 @@
...
+  }
+
+  /**
+   * Get the value out of the inside of the `Optional` object, using a default
+   * `replacement` value if the provided Optional object does not contain a
+   * value.
+   *
+   * Unlike `or`, in this method the `replacement` value is "thunked" - that is
+   * to say that you don't pass a value to `getOrThunk`, you pass a function
+   * which (if called) will **return** the `value` you want to use.
```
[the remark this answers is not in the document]
ashfordneil: It wasn't meant to say `getOrThunk` but it was meant to be very similar to the one above. Have fixed it

modules/katamari/src/main/ts/ephox/katamari/api/Optional.ts
```diff
@@ -1,141 +1,291 @@
...
+  /**
+   * Converts an `Optional` to a nullable type.
+   */
+  public getOrNull(): T | null {
+    return this.tag ? this.value as T : null;
+  }
+
+  /**
+   * Converts an `Optional` to an undefined-able type.
+   */
```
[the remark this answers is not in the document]
ashfordneil: I've updated these to be slightly clearer in what they do, let me know what you think.

They could go in either place, but I think they are a good enough fit here that it's not worth moving them

modules/katamari/src/main/ts/ephox/katamari/api/Optional.ts
ashfordneil: Done (and also ported to Result.ts where I had the same string)

modules/katamari/src/main/ts/ephox/katamari/api/Optional.ts
ashfordneil: It could, good catch thanks

modules/katamari/src/main/ts/ephox/katamari/api/Optional.ts
ashfordneil: No longer in the doc comments. Instead, I've replaced strings like 
```
// --- Functor ---
```

with 

```
// --- Functor (name stolen from Haskell / maths) ---
```

for the names where that was appropriate.

modules/katamari/src/main/ts/ephox/katamari/api/Type.ts
ashfordneil: Done

modules/katamari/src/main/ts/ephox/katamari/api/Optional.ts
```diff
@@ -1,141 +1,290 @@
...
 
-/** A value which may be Optional.some(a) or Optional.none().
- *  Useful for representing partiality or pass/fail conditions that produce a value when successful.
- *  Sometimes called a "type-safe alternative to null".
- *  Can be thought of as a list of exactly zero or 1 elements.
+/**
+ * The `Optional` type represents a value (of any type) that potentially does
+ * not exist. Any `Optional<T>` can either be a `Some<T>` (in which case the
+ * value does exist) or a `None` (in which case the value does not exist). This
+ * file defines a whole lot of FP-inspired utility functions for dealing with
```
lnewson, line 7:
```suggestion
 * module defines a whole lot of FP-inspired utility functions for dealing with
```
Remember this won't be a file when showing as JS doc strings.

modules/katamari/src/main/ts/ephox/katamari/api/Optional.ts
```diff
@@ -1,141 +1,290 @@
...
 
-  readonly isSome: () => boolean;
-  readonly isNone: () => boolean;
+  // Sneaky optimisation: every instance of Optional.none is identical, so just
+  // reuse the same object
+  private static singletonNone = new Optional<any>(false);
 
-  /** If some(x) return x, otherwise return the specified default value */
-  readonly getOr: <T2 = T>(value: T2) => T | T2;
+  // The internal representation has an `tag` and a `value`, but both are
```
lnewson, line 27:
```suggestion
  // The internal representation has a `tag` and a `value`, but both are
```

modules/katamari/src/main/ts/ephox/katamari/api/Optional.ts
```diff
@@ -1,141 +1,290 @@
...
+ * `Optional` objects.
+ *
+ * Comparison with null or undefined:
+ * - We don't get fancy null coalescing operators with `Optional`
+ * - We do get fancy helper functions with `Optional`
+ * - `Optional` support nesting, and allow for the type to still be nullable (or
+ * another `Optional`)
+ * - There is no option to turn off strict-optional-checks like there is for
+ * strict-null-checks
+ * - Try to use `Optional` instead of null or undefined where you can
```
lnewson, line 17: Should this specific item be in the doc string? It seems like a process comment, not something that should be in a doc string.
```suggestion
```

modules/katamari/src/main/ts/ephox/katamari/api/Optional.ts
```diff
@@ -1,141 +1,290 @@
...
-  readonly each: (worker: (x: T) => void) => void;
+  // --- Monad (name stolen from Haskell / maths) ---
 
-  /** "bind"/"flatMap" operation on the Optional Bind/Monad.
-   *  Equivalent to >>= in Haskell/PureScript; flatMap in Scala.
+  /**
+   * Perform a transform on an `Optional` object, **if** there is a value.
+   * Unlike the map function earlier in the piece, here the transform itself
+   * also returns an `Optional`.
    */
```
lnewson, line 108: This needs rewording, as you can't say "earlier in the piece" as when viewing the doc string helper for the function it won't have that context and so each comment here needs to be largely standalone.

- [ ] Reword

modules/katamari/src/main/ts/ephox/katamari/api/Optional.ts
```diff
@@ -1,141 +1,290 @@
...
+  // --- Traversable (name stolen from Haskell / maths) ---
 
-  /** Do all values contained in this option match this predicate? */
-  readonly forall: (f: (x: T) => boolean) => boolean;
+  /**
+   * For a given predicate, this function finds out if there **exists** a value
+   * inside this `Optional` object that meets the predicate. In practice, this
+   * means that for empty `Optional`s it returns false (as no predicate-
+   * meeting value exists).
+   */
```
lnewson, line 124: Maybe instead of saying `empty` Optionals, as that's not very clear (e.g. you could have an empty array), say something like this?
```suggestion
  /**
   * For a given predicate, this function finds out if there **exists** a value
   * inside this `Optional` object that meets the predicate. In practice, this
   * means that for a `None` it returns false (as no predicate-meeting value
   * exists).
   */
```

modules/katamari/src/main/ts/ephox/katamari/api/Optional.ts
```diff
@@ -1,141 +1,290 @@
...
+    return this.tag && predicate(this.value as T);
+  }
 
-  /** Return all values in this Optional that match the predicate.
-   *  The predicate may refine the constituent type using TypeScript type predicates.
+  /**
+   * For a given predicate, this function finds out if **all** the values inside
+   * this `Optional` object meet the predicate. In practice, this means that
+   * for empty `Optional`s it returns true (as all 0 objects do meet the
+   * predicate).
```
lnewson, line 133: Similar to my previous comment, using empty is not clear.
```suggestion
  /**
   * For a given predicate, this function finds out if **all** the values inside
   * this `Optional` object meet the predicate. In practice, this means that
   * for a `None` it returns true (as all 0 objects do meet the predicate).
```

modules/katamari/src/main/ts/ephox/katamari/api/Result.ts
```diff
@@ -1,156 +1,288 @@
 import * as Fun from './Fun';
 import { Optional } from './Optional';
 
+/**
+ * The `Result` type represents a value (of any type) that may instead be an
+ * error (of any type). Any `Result<T, E>` can either be a `Value<T>` (which
+ * contains a value of type `T`) or an `Error<E>` (which contains an error of
+ * type `E`). This file defines a whole lot of FP-inspired utility functions for
+ * dealing with `Result` objects.
```
lnewson, line 9:
```suggestion
 * type `E`). This module defines a whole lot of FP-inspired utility functions
 * for dealing with `Result` objects.
```

modules/katamari/src/main/ts/ephox/katamari/api/Result.ts
```diff
@@ -1,156 +1,288 @@
...
+ *
+ * Comparison with exceptions:
+ * - Each function's type signature says whether it returns a `Result` or not
+ * (they're like checked exceptions).
+ * - The type of the error in a result can be checked by TypeScript.
+ * - You can't forget to catch errors from inside a `Result`, you need to handle
+ * them to access the value.
+ * - Try to use `Result` instead of exceptions where you can (outside of tests)
```
lnewson, line 17: Again, this is a process thing so probably not suitable for docstrings.
```suggestion
```

modules/katamari/src/main/ts/ephox/katamari/api/Result.ts
```diff
@@ -1,156 +1,288 @@
...
-    return onValue(o);
-  };
+  // --- Monad (name stolen from Haskell / maths) ---
 
-  const exists = (f: (value: T) => boolean) => {
-    return f(o);
-  };
+  /**
+   * Perform a transform on a `Result` object, **if** there is a value. If you
+   * have a function that attempts to transform a `T` into a `U` (ie a function
```
lnewson, line 88:
```suggestion
   * have a function that attempts to transform a `T` into a `U` (i.e. a function
```

modules/katamari/src/main/ts/ephox/katamari/api/Result.ts
```diff
@@ -1,156 +1,288 @@
...
+
+  /**
+   * Convert this `Result<T, E>` to an `Optional<T>`. If this object contains a
+   * value, so will the output. If this object contains an error, the error will
+   * be discarded and the output will not contain a value.
+   */
+  readonly toOptional: () => Optional<T>;
+}
+
+/**
```
lnewson, line 192:
```suggestion
/*
```
This doesn't appear to be a doc string and is just a general comment, so shouldn't have 2 asterisks.

modules/katamari/src/main/ts/ephox/katamari/api/Result.ts
```diff
@@ -1,156 +1,288 @@
...
-    getOrThunk,
-    getOrDie,
-    or,
-    orThunk,
-    fold,
-    map,
-    mapError,
-    each: Fun.noop,
-    bind,
+    map: <U>(_mapper) => output as unknown as Result<U, E>,
```
lnewson, line 261: This and `bind` looks like it could also use the `outputHelper` function from `value`?

modules/katamari/src/main/ts/ephox/katamari/api/Result.ts
```diff
@@ -1,156 +1,288 @@
...
-  value
-);
+/**
+ * Creates a new `Result<T, E>` from an `Optional<T>` and an `E`. If the
+ * `Optional` contains a value, so will the outputted `Result`. If it does not,
+ * the outputted `Result` will contain an error (and that error will be the
+ * error passed in).
+ */
+const fromOption = <T, E>(optional: Optional<T>, error: E): Result<T, E> =>
+  optional.fold(() => Result.error(error), Result.value);
```
lnewson, line 285: This references variables before they are defined, which we generally try to avoid... that's why the old version didn't do this, so are we sure we want to make this change?

modules/katamari/src/main/ts/ephox/katamari/api/Type.ts
```diff
@@ -1,10 +1,26 @@
+const getPrototypeOf = Object.getPrototypeOf;
+
+const hasProto = (v: Object, predicate: (v: Object) => boolean, name: string): boolean => {
+  if (predicate(v)) {
+    return true;
+  } else {
+    // String-based fallback time
```
lnewson, line 7: Can we please add a comment that the string based fallback is only needed for IE 11? That way we can remove it once we drop IE support.

lnewson (approved):

modules/katamari/src/main/ts/ephox/katamari/api/Optional.ts
ashfordneil: Reworded

modules/katamari/src/main/ts/ephox/katamari/api/Optional.ts
ashfordneil: I've clarified this to "Optionals that don't contain a value" to match up with the other docs

modules/katamari/src/main/ts/ephox/katamari/api/Optional.ts
ashfordneil: As above

ashfordneil:
Just messed with all of the CHANGELOG files as this is likely getting pushed to a later release.

- [ ] Make a new feature branch to merge this into, and sort out changelogs there

modules/katamari/src/main/ts/ephox/katamari/api/Type.ts
```diff
@@ -1,10 +1,26 @@
+const getPrototypeOf = Object.getPrototypeOf;
+
+const hasProto = (v: Object, predicate: (v: Object) => boolean, name: string): boolean => {
+  if (predicate(v)) {
+    return true;
+  } else {
+    // String-based fallback time (only needed for IE)
```
lnewson, line 7: This comment still isn't very clear, as now it sounds like the `constructor` check is also only needed for IE 11, but it's needed for modern browsers. It's only the `toString` check that's needed for IE 11.

ashfordneil: Good catch, I've moved the "IE only" comment onto the code that's actually IE only

ashfordneil:
Now targeting feature/initialise-next, which will change to just targeting `next` once #7212 has merged. There was a small force push (sorry) to make the changelog git merges not completely horrible.

JamesToohey (approved):

PR #7193, opened 2021-09-14
TINY-7596: Fix entire nodes within cells being cleaned when they still contain content

JamesToohey:
Related Ticket: TINY-7596 (Part 2)

Description of Changes:
Ensures we don't clean out nodes that aren't empty

Pre-checks:
* [x] ~Changelog entry added~ Already exists

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/delete/TableDelete.ts
+++ b/modules/tinymce/src/core/main/ts/delete/TableDelete.ts
@@ -6,7 +6,7 @@
  */
 
 import { Arr, Fun, Optional, Optionals } from '@ephox/katamari';
-import { Attribute, Compare, Remove, SugarElement, SugarNode, Traverse } from '@ephox/sugar';
+import { Attribute, Compare, PredicateExists, Remove, SugarElement, SugarNode, Traverse } from '@ephox/sugar';
 
 import Editor from '../api/Editor';
 import * as CaretFinder from '../caret/CaretFinder';
@@ -154,7 +154,7 @@ const emptyMultiTableCells = (
   return true;
 };
 
-// Runs on a single cell table that has all of its content selected
+// Delete the contents of a range inside a cell. Runs on tables that are a single cell as well as partial selections that need to be cleaned up.
 const deleteCellContents = (editor: Editor, rng: Range, cell: SugarElement<HTMLTableCellElement>, moveSelection: boolean = true): boolean => {
   rng.deleteContents();
   // Pad the last block node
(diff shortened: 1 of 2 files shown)


modules/tinymce/src/core/main/ts/delete/TableDelete.ts
```diff
@@ -154,7 +154,7 @@ const emptyMultiTableCells = (
   return true;
 };
 
-// Runs on a single cell table that has all of its content selected
+// Delete the contents of a range inside a cell. Runs on tables that are a single cell as well as partial selections that need to be cleaned up.
```
lnewson, line 157:
```suggestion
// Delete the contents of a range inside a cell. Runs on tables that are a single cell and partial selections that need to be cleaned up.
```
The current working just seems to mix the 2 cases together which I found confusing.

modules/tinymce/src/core/main/ts/delete/TableDelete.ts
```diff
@@ -171,7 +171,11 @@ const deleteCellContents = (editor: Editor, rng: Range, cell: SugarElement<HTMLT
     const additionalCleanupNodes = Optionals.is(Traverse.parent(lastBlock), cell) ? [] : Traverse.siblings(lastBlock);
     Arr.each(additionalCleanupNodes.concat(Traverse.children(cell)), (node) => {
       if (!Compare.eq(node, lastBlock) && !Compare.contains(node, lastBlock)) {
-        Remove.remove(node);
+        // Don't clear any nodes that still have content in their descendants
+        const hasNonEmptyDescendant = PredicateExists.descendant(node, Fun.not(Empty.isEmpty));
+        if (!hasNonEmptyDescendant) {
+          Remove.remove(node);
+        }
```
lnewson, line 178:
```suggestion
        if (Empty.isEmpty(node)) {
          Remove.remove(node);
        }
```
This isn't necessary and can be replaced with this I believe, as `Empty.isEmpty` already walks down the descendants, so this is just exponentially checking if child nodes are empty when it only needs to be done once.

- [ ] Don't use `PredicateExists.descendant` as it's going to cause child nodes to be checked multiple times.

lnewson (approved):

modules/tinymce/src/core/main/ts/delete/TableDelete.ts
lnewson: Actually this check could be moved up to line 173, instead of using the above suggestion.

JamesToohey: Yep nice, will do it all in there

ltrouton (approved):

PR #7195, opened 2021-09-14
TINY-7875: Added more correct typing to get/set content.

HAFRMO:
Related Ticket:

Description of Changes:
Updated types

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/tinymce/src/core/main/ts/content/GetContentImpl.ts
```diff
@@ -27,10 +27,11 @@ const setupArgs = (args: Partial<GetContentArgs>, format: ContentFormat): GetCon
   ...args,
   format,
   get: true,
-  getInner: true
+  getInner: true,
+  content: ''
```
lnewson, line 31: This is definitely incorrect as it'll overwrite the content passed in via `args`. As such this should be before the `...args` part so it's a default.

modules/tinymce/src/core/main/ts/api/Editor.ts
```diff
@@ -891,13 +891,13 @@ class Editor implements EditorObservable {
    * @method resetContent
    * @param {String} initialContent An optional string to use as the initial content of the editor.
    */
-  public resetContent(initialContent?: string) {
+  public resetContent(initialContent?: string, args?: Partial<EditorContent.SetContentArgs>) {
```
lnewson, line 894: If you're making this change, you need to update the docs string above and include a change log entry for it. However, this doesn't quite seem right so I'd be interested to know why this was added?

modules/tinymce/src/core/main/ts/content/ContentTypes.ts
```diff
@@ -10,23 +10,42 @@ import AstNode from '../api/html/Node';
...
+}
+
+export interface SelectionSetContentArgs extends SetContentArgsBase {
+  content: string;
+  selection?: boolean;
+  contextual?: boolean;
+}
+
+export interface GetSelectionContentArgs extends GetContentArgsBase {
+  selection?: boolean;
```
lnewson, line 44:
```suggestion
  selection: boolean;
```
This one is always defined as per `GetSelectionContentImpl`.

modules/tinymce/src/core/main/ts/content/ContentTypes.ts
```diff
@@ -10,23 +10,42 @@ import AstNode from '../api/html/Node';
...
+  content: Content;
+}
+
+export interface SetContentArgs extends SetContentArgsBase {
+  content: Content;
+}
+
+export interface SelectionSetContentArgs extends SetContentArgsBase {
+  content: string;
+  selection?: boolean;
```
lnewson, line 39:
```suggestion
  selection: boolean;
```
Likewise this is always defined as well.

modules/tinymce/src/core/main/ts/content/ContentTypes.ts
```diff
@@ -10,23 +10,42 @@ import AstNode from '../api/html/Node';
...
+interface SetContentArgsBase {
+  format: string;
+  set: boolean;
+  content: Content;
   no_events?: boolean;
   no_selection?: boolean;
 }
 
+export interface GetContentArgs extends GetContentArgsBase {
+  content: Content;
```
lnewson, line 30: Are we sure this is correct? Looking at the `GetContentImpl` logic, it'll only ever use a string in the event. Maybe we do need to break this up into 2 types, one for the event and one for the args? Additionally, the more I look this it looks like `content` should be optional here as it's not added into **after** `BeforeGetContent` fires, so we may even need 3 types (2 for the different events and 1 for the args).

- [ ] Ensure this is correct, as per the above comments

lnewson: Actually the same is true of `SetContentImpl` as well, in that case `content` will always be a `string` since it serializes the `AstNode`. So I definitely think we're going to need 2 or more different types here (which will likely extend a base type).

lnewson:
Talked to Hannes more, it turns there very large and confusing inconsistencies between `setContent()` and `selection.setContent()` whereby `setContent()` is always a string and `selection.setContent()` seems rather broken as `content` is empty for `tree`.

As such, we're just going to make this target the long running feature branch to unblock the HV team and the editor team will need to fix these oddities for 6.0.

modules/tinymce/src/core/main/ts/content/GetContentImpl.ts
HAFRMO: After discussion, set `content` is made optional and thus not needed here.

modules/tinymce/src/core/main/ts/api/Editor.ts
HAFRMO: Might just have added it because I was adding it everywhere else, to be honest. I'll take it away again.

modules/tinymce/src/core/main/ts/content/ContentTypes.ts
HAFRMO: Woups, good catch.

modules/tinymce/src/core/main/ts/content/ContentTypes.ts
HAFRMO: I'll continue to use the single type for this PR, but I'll add a note on the separate issue to properly fix this that this should be done.

modules/tinymce/src/plugins/table/main/ts/actions/Clipboard.ts
```diff
@@ -50,12 +50,13 @@ const registerEvents = (editor: Editor, selections: Selections, actions: TableAc
...
   editor.on('BeforeSetContent', (e) => {
-    if (e.selection === true && e.paste === true) {
+    const content = e.content;
+    if (e.selection === true && e.paste === true && Type.isString(content)) {
       const selectedCells = TableSelection.getCellsFromSelection(selections);
       Arr.head(selectedCells).each((cell) => {
         TableLookup.table(cell).each((table) => {
 
-          const elements = Arr.filter(SugarElements.fromHtml(e.content), (content) => {
+          const elements = Arr.filter(SugarElements.fromHtml(content), (content) => {
```
ashfordneil, line 59: We should try to avoid shadowing this `content` variable name — either change the name in the outer scope (the variable you added in this PR) or in the inner scope (not code you've touched in this PR directly, but might be easier to find a new name here)

modules/tinymce/src/core/main/ts/content/ContentTypes.ts
```diff
@@ -10,23 +10,42 @@ import AstNode from '../api/html/Node';
...
   no_events?: boolean;
   [key: string]: any;
 }
 
-export interface SetContentArgs {
-  format?: string;
-  set?: boolean;
-  content?: string;
+interface SetContentArgsBase {
+  format: string;
```
ashfordneil, line 22: I know that by changing everything to `Partial` you've managed to avoid any breakages to actual editor APIs — but I'm wondering what the semver implications are of changing the type itself even if the APIs stay the same. Should we find a way to mention that all the fields on these types are now required (while making it clear that you can pass `Partial` types into all of our editor APIs) in the changelog?

ashfordneil (commented):
I'm a little worried about all the no-ops we have when we're handling get/set content events for `format: 'tree'`, but I don't really know what to do about that or what it was doing before then.

modules/tinymce/src/core/main/ts/api/EditorUpload.ts
```diff
@@ -302,7 +302,7 @@ const EditorUpload = (editor: Editor): EditorUpload => {
 
   editor.on('GetContent', (e) => {
     // if the content is not a string, we can't process it
-    if (e.source_view || e.format === 'raw' || e.format === 'tree') {
+    if (e.source_view || e.format === 'raw' || e.format === 'tree' || !Type.isString(e.content)) {
```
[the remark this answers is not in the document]
ashfordneil: I've now realised that this is very common throughout the rest of the PR. I am curious what we are (and were) doing in the not-a-string case, but less worried about this in particular

modules/tinymce/src/core/main/ts/content/ContentTypes.ts
lnewson: @ashfordneil also remember this is for 6.0 and we are looking to split the tree content out into a separate API, so a large amount of this is temporary and why it's not getting merged to a main branch.

HAFRMO:
> I'm a little worried about all the no-ops we have when we're handling get/set content events for `format: 'tree'`, but I don't really know what to do about that or what it was doing before then.

This is more meant as a 'stopgap' fix to allow the HV team to keep developing this epic. A more proper fix will be developed separately.

modules/tinymce/src/core/main/ts/api/EditorUpload.ts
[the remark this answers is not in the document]
HAFRMO: Became very unhappy, is what we did. But yeah, it's a bug that was pointed out by the typing now when it was more properly typed, and this is meant to just let us keep developing this issue separately.

modules/mcagar/src/main/ts/ephox/mcagar/alien/EditorTypes.ts
```diff
@@ -13,14 +13,22 @@ export interface Selection {
 export type ContentFormat = 'raw' | 'text' | 'html' | 'tree';
 
 export interface GetContentArgs {
-  format?: ContentFormat;
-  get?: boolean;
-  content?: string;
+  format: ContentFormat;
+  get: boolean;
+  content: any;
```
lnewson, line 18:
```suggestion
  content?: any;
```
Remember we found that this wasn't always set in the "before" case, so it should still be optional. Either that or we need to clone the various different types you've added to here as well.

modules/tinymce/src/core/main/ts/selection/GetSelectionContentImpl.ts
```diff
@@ -90,10 +85,12 @@ const setupArgs = (args: Partial<GetSelectionContentArgs>, format: ContentFormat
   ...args,
   format,
   get: true,
-  selection: true
+  selection: true,
+  content: '',
```
lnewson, line 89: This should have been reverted since we found out the existing type is meant to be empty/optional in this case.
```suggestion
```

modules/tinymce/src/core/main/ts/content/ContentTypes.ts
```diff
@@ -10,23 +10,42 @@ import AstNode from '../api/html/Node';
...
+export interface SelectionSetContentArgs extends SetContentArgsBase {
+  content: string;
+  selection: boolean;
+  contextual?: boolean;
+}
+
+export interface GetSelectionContentArgs extends GetContentArgsBase {
+  selection: boolean;
+  contextual?: boolean;
+  content: string;
```
lnewson, line 46:
```suggestion
  content?: string;
```
As mentioned elsewhere, at least currently this is meant to be optional as it's not included in the "before" case.

lnewson (approved):
LGTM on the condition this is more of a temporary state until we can fix things properly for 6.0 via TINY-7996 & TINY-8018.

ashfordneil (approved):

modules/mcagar/src/main/ts/ephox/mcagar/alien/EditorTypes.ts
```diff
@@ -13,14 +13,22 @@ export interface Selection {
 export type ContentFormat = 'raw' | 'text' | 'html' | 'tree';
 
 export interface GetContentArgs {
-  format?: ContentFormat;
-  get?: boolean;
-  content?: string;
+  format: ContentFormat;
```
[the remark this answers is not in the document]
HAFRMO: The tests seems to all pass, and since this is mainly just to make the epic be able to move forward while a proper fix is implemented, it should probably be fine?

PR #7197, opened 2021-09-16
TINY-7041: Fix exception thrown when more `col` elements than columns

ltrouton:
Related Ticket: TINY-7041

Description of Changes:
* Rework `TableGrid.subgrid` to better handle colgroup rows in the grid
* Refactored `Warehouse.generate` as I was trying to clamp the excess `col` elements but reverted that (happy to revert the other changes as well) 

Pre-checks:
* [X] Changelog entry added

Review:
* [X] Milestone set


modules/snooker/src/main/ts/ephox/snooker/api/Warehouse.ts
```diff
@@ -80,42 +80,47 @@ const generate = <T extends Structs.Detail>(list: Structs.RowDetail<T>[]): Wareh
...
+    cells.push(Structs.rowdetail(rowData.element, currentRow, rowData.section));
+    rowCount++;
+  });
 
-      maxRows++;
-      cells.push(Structs.rowdetail(rowData.element, currentRow, rowData.section));
-      rowCount++;
-    }
+  // Handle colgroups
+  // Note: Currently only a single colgroup is supported so just use the last one
```
lnewson, line 119: Your change here breaks this specific implementation, so please make sure to revert this part of the change so we are guaranteed to only use the last `colgroup` element.

- [ ] Revert doing colgroups.push

ashfordneil (approved):

modules/tinymce/CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Image resize backdrop element did not have `data-mce-bogus="all"` set #TINY-7854
 - Table cells that were both row and column headers would not retain the correct state when converting back to a regular row or column #TINY-7709
 - Clicking beside a non-editable element could cause the editor to incorrectly scroll to the top of the document #TINY-7062
+- Fixed an exception getting thrown when there were more `col` elements than cell columns in a table #TINY-7041
```
lnewson, line 17: I'm not too keen on using `cell columns` here as that's not overly clear. I think if anything we should just use `cells` here instead, but would be keen to see what others think or if they have any better suggestions?

modules/snooker/src/main/ts/ephox/snooker/model/TableGrid.ts
```diff
@@ -35,15 +35,17 @@ const findDiff = (xs: ElementNew[], comp: (a: SugarElement, b: SugarElement) =>
...
-  const endColIndex = findDiff(restOfRow, comparator);
+  const gridRow = getRow(grid, row);
+  const isColRow = gridRow.section === 'colgroup';
+  const { rows, cols } = GridRow.extractGridDetails(grid);
+  const numColRows = cols.length;
 
-  const restOfColumn = getColumn(grid, column).slice(row);
-  const endRowIndex = findDiff(restOfColumn, comparator);
+  const colspan = findDiff(gridRow.cells.slice(column), comparator);
+  const rowspan = isColRow ? 1 : findDiff(getColumn(rows, column).slice(row - numColRows), comparator);
```
lnewson, line 44: Why was this changed from using `grid` to `rows`? As that doesn't look like it was needed and in fact complicates this since we need to then also adjust the row number passed. If there was a valid reason, can we please move the row number adjustment to a separate variable?

 Also is this even going to work if the `row` passed in is `0`, since the slice would then be `-1` potentially meaning it'll get everything except the last item?

lnewson: Re that last part, I guess it'd be doing `[].slice(-1)` which would still be an empty array. It also shouldn't really be able to get into that state since if you pass in `0` then `isColRow` would be true. However, I'd prefer if we weren't relying on side-effects like that for this to just happen to be working, as it can cause confusion/be hard to reason about.

modules/tinymce/src/plugins/table/test/ts/browser/ExcessColsTest.ts
```diff
@@ -0,0 +1,93 @@
...
+      td: tdCount,
+    });
+  };
+
+  // Note: THe goal of these tests is to make sure an exception is not thrown and each table operation works
+  // when there are more col elements than actual columns
+  Arr.each([
+    {
+      label: 'more cols',
+      tableHTML: '<table><colgroup><col /><col /><col /><col /></colgroup><tbody><tr><td>1</td><td>2</td></tr><tr><td>3</td><td>4</td></tr></tbody></table>',
```
lnewson, line 26: Can we please rename this to `tableHtml` as that's more consistent with our regular variable naming conventions?

lnewson (changes requested):
Requesting changes largely because of that colgroup change in warehouse, as that could cause some serious issues, but also partly because of the subgrid changes as well.

Overall looks good though :+1:

modules/snooker/src/main/ts/ephox/snooker/model/TableGrid.ts
lnewson: So after playing with this some more, I believe this should achieve the desired result here and also saves some cycles, since `getColumn` doesn't need to walk over every row.
```
const rowspan = isColRow ? 1 : findDiff(getColumn(grid.slice(row), column), comparator);
```

modules/tinymce/CHANGELOG.md
ashfordneil: I don't really like `cell columns` either, but I'd prefer we shorten it to just `columns` instead (as the issue wasn't about more `col` elements than cells in the table). Maybe we could do something like this to specify what sort of "columns" we mean?

```suggestion
- Fixed an exception getting thrown when there were more `col` elements than columns of content in a table #TINY-7041
```

modules/tinymce/src/plugins/table/test/ts/browser/ExcessColsTest.ts
ltrouton: Done

modules/snooker/src/main/ts/ephox/snooker/model/TableGrid.ts
ltrouton: I think I just overcomplicated it. I agree that the suggestion you have given should work properly.

modules/snooker/src/main/ts/ephox/snooker/api/Warehouse.ts
ltrouton: Have reworked it based on DMs in Slack

lnewson (approved):

modules/tinymce/src/plugins/table/test/ts/browser/ExcessColsTest.ts
```diff
@@ -0,0 +1,93 @@
...
+
+      it('TINY-7041: insert new column in the middle of the table', () => {
+        const editor = hook.editor();
+        editor.setContent(tableHtml);
+        TinySelections.setCursor(editor, [ 0, 1, 0, 0, 0 ], 0);
+        editor.execCommand('mceTableInsertColAfter');
+        assertBasicTablePresence(editor, 6);
+      });
+
+      it('TINY-7041: insert new column at the emd of the table', () => {
```
JamesToohey, line 60:
```suggestion
      it('TINY-7041: insert new column at the end of the table', () => {
```

JamesToohey (approved):

modules/tinymce/CHANGELOG.md
ltrouton: Have just removed 'cell' from the sentence which I think should still be clear enough

PR #7222, opened 2021-09-23
TINY-8036: Fixed a regression in the format expand range logic when using block formats with a wrapper

lnewson:
Related Ticket: TINY-8036

Description of Changes:

Fixes a regression introduced in https://github.com/tinymce/tinymce/pull/7002 whereby block wrapper formats didn't expand out to the parent block correctly. This in turn caused blockquotes to fail to apply/remove correctly when there were multiple words.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/fmt/ExpandRange.ts
+++ b/modules/tinymce/src/core/main/ts/fmt/ExpandRange.ts
@@ -121,14 +121,14 @@ const findSelectorEndPoint = (dom: DOMUtils, formatList: Format[], rng: Range, c
 };
 
 const findBlockEndPoint = (editor: Editor, formatList: Format[], container: Node, siblingName: Sibling) => {
-  let node = container;
+  let node: Node | null = container;
   const dom = editor.dom;
   const root = dom.getRoot();
   const format = formatList[0];
 
   // Expand to block of similar type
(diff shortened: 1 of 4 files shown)


modules/tinymce/src/core/main/ts/fmt/ExpandRange.ts
```diff
@@ -121,14 +121,14 @@ const findSelectorEndPoint = (dom: DOMUtils, formatList: Format[], rng: Range, c
...
+  let node: Node | null = container;
   const dom = editor.dom;
   const root = dom.getRoot();
   const format = formatList[0];
 
   // Expand to block of similar type
-  if (FormatUtils.isBlockFormat(format) && !format.wrapper) {
-    node = dom.getParent(container, format.block, root);
+  if (FormatUtils.isBlockFormat(format)) {
+    node = format.wrapper ? null : dom.getParent(container, format.block, root);
```
lnewson, line 131: Turns out this needs to be falsy when using a block wrapper so that it flows into the logic on line 134. Without that it fails to expand to the block and only ends up doing a word grab for the current word.

ltrouton (approved):

ashfordneil (approved):

PR #7226, opened 2021-09-24
TINY-7736: Fixed issues when clicking into a table cell when another cell had a cef element

### LINKED ISSUE #6350: Inline Tables and NonEditable Content - Table Cell Focus
Hey Legends,

Loving the editor. Keep up the great work. Just a small bug I've come across when mixing some features together.

**Behaviour:** When clicking certain empty cells in inline tables that both contain NonEditable content, it doesn't bring focus to the cell you've clicked on.

**Example:** https://codesandbox.io/s/tinymce-and-react-table-cell-focus-bug-ng02y?file=/index.js

**Context:** I'm making a document template which populates the preset template fields from an API call.
A single table with 4 columns can't be used as the data can contain multiple lines and each side should not effect eachother's row height.
 
**Expected Behaviour:** Cell you're clicking on should gain focus. (Working in Firefox 84.0)

**Versions:**
MacOS
Chrome - 87.0.4280.88
Safari - 14.0

**Working perfect: Firefox - 84.0**

TinyMCE version: 5.6.2

 Cheers!

### LINKED ISSUE #6547: Mobile device: Caret is positioned to adjacent table cell, when adjacent table cell contains node with attribute contenteditable="false"
Using a mobile device, create a table with multiple columns first. 
tableConfig

Input some text, and followed by inputting a node: `<span contenteditable="false">anything</span>` by a customized plugin for tinymce. In addition, make the span to be the last part of this table cell's first line. Then continue to input some text, they will appear at the second line of this cell. 
cellContent
tableHtml

Clicking on the upper half of the next table cell, the caret goes to the previous cell, and it is located after the not contenteditable span. However, caret locates correctly when the first table cell contains only editable content.

https://user-images.githubusercontent.com/79293925/108457322-62f40500-72ad-11eb-9124-3fadd46e28a6.mp4

Caret is supposed to be located at the cell where the user clicks.

Version of tinymce: 5.3.2
Browser: Chrome
Replicable to android table/phone and ipad.

lnewson:
Related Ticket: TINY-7736

Description of Changes:
Fixes an issue where the "better mouse target" logic didn't handle clicking into an empty block element, which resulted in the selection being placed next to a cef element on the same "line" as the target.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

Fixes #6350
Fixes #6547

### DIFF
--- a/modules/tinymce/src/core/main/ts/SelectionOverrides.ts
+++ b/modules/tinymce/src/core/main/ts/SelectionOverrides.ts
@@ -5,7 +5,7 @@
  * For commercial licenses see https://www.tiny.cloud/
  */
 
-import { Arr, Obj, Unicode } from '@ephox/katamari';
+import { Arr, Obj, Type, Unicode } from '@ephox/katamari';
 import { Attribute, Compare, Remove, SelectorFilter, SelectorFind, SugarElement } from '@ephox/sugar';
 
 import Editor from './api/Editor';
@@ -63,8 +63,7 @@ const SelectionOverrides = (editor: Editor): SelectionOverrides => {
     return container ? container.getElementsByTagName('*')[0] as HTMLElement : container;
   };
 
-  const setRange = (range: Range) => {
-    // console.log('setRange', range);
+  const setRange = (range: Range | null) => {
     if (range) {
       selection.setRng(range);
     }
@@ -132,22 +131,26 @@ const SelectionOverrides = (editor: Editor): SelectionOverrides => {
     editor.on('ResizeWindow FullscreenStateChanged', fakeCaret.reposition);
 
     const hasNormalCaretPosition = (elm: Element) => {
-      const caretWalker = CaretWalker(elm);
-
-      if (!elm.firstChild) {
+      const start = elm.firstChild;
+      if (Type.isNullable(start)) {
         return false;
       }
 
-      const startPos = CaretPosition.before(elm.firstChild);
-      const newPos = caretWalker.next(startPos);
-
-      return newPos && !isNearFakeSelectionElement(newPos);
+      const startPos = CaretPosition.before(start);
+      // If the element has a single br as a child (e.g. is empty), then the start position is a valid cursor position
+      if (NodeType.isBr(startPos.getNode()) && elm.childNodes.length === 1) {
+        return !isNearFakeSelectionElement(startPos);
+      } else {
+        const caretWalker = CaretWalker(elm);
+        const newPos = caretWalker.next(startPos);
+        return newPos && !isNearFakeSelectionElement(newPos);
+      }
     };
 
     const isInSameBlock = (node1: Node, node2: Node) => {
       const block1 = dom.getParent(node1, isBlock);
       const block2 = dom.getParent(node2, isBlock);
-      return block1 === block2;
+      return block1 === block2 || dom.isChildOf(block2, block1);
     };
 
     // Checks if the target node is in a block and if that block has a caret position better than the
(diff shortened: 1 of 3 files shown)


lnewson:
Converting back to a draft as the second test (selecting next to a cef doesn't work on Firefox/IE) for some reason :thinking:

ltrouton (approved):

modules/tinymce/src/core/main/ts/SelectionOverrides.ts
```diff
@@ -132,22 +131,26 @@ const SelectionOverrides = (editor: Editor): SelectionOverrides => {
...
+        const newPos = caretWalker.next(startPos);
+        return newPos && !isNearFakeSelectionElement(newPos);
+      }
     };
 
     const isInSameBlock = (node1: Node, node2: Node) => {
       const block1 = dom.getParent(node1, isBlock);
       const block2 = dom.getParent(node2, isBlock);
-      return block1 === block2;
+      return block1 === block2 || dom.isChildOf(block2, block1);
```
ashfordneil, line 153: Is there a reason we don't also have `|| dom.isChildOf(block1, block2)`?

modules/tinymce/src/core/test/ts/browser/SelectionOverridesTest.ts
```diff
@@ -25,20 +25,20 @@ describe('browser.tinymce.core.SelectionOverridesTest', () => {
 
   const assertSelectionIsCaretBlock = (editor: Editor, caretValue: 'before' | 'after') => {
     const selectedNode = editor.selection.getNode();
-    assert.isTrue(isCaretContainerBlock(selectedNode));
+    assert.isTrue(isCaretContainerBlock(selectedNode), 'Selected node should be a fake caret node');
     assert.equal(selectedNode.getAttribute('data-mce-caret'), caretValue);
   };
 
-  const selectBesideContentEditable = (editor: Editor, contentEditableElm: HTMLElement, clickPoint: 'before' | 'after') => {
+  const selectBesideContentEditable = (editor: Editor, contentEditableElm: HTMLElement, clickPoint: 'before' | 'after', offset: number = 8) => {
```
ashfordneil, line 32: Can we keep the comment explaining where the 8 came from? It appears to have been lost in the change

ashfordneil (approved):

modules/tinymce/src/core/test/ts/browser/SelectionOverridesTest.ts
lnewson: There wasn't really anywhere to put it tbh... maybe I should just not make this optional and then I could put the comment where it's used.

modules/tinymce/src/core/main/ts/SelectionOverrides.ts
lnewson: We don't want that in this case, as `node1` is the `target` element and `node2` is the `caret` element. So this should not return true if `target` is within the `caret` block element. To be honest, as this function is only used for the one purpose we probably should just rename `node1`/`block1` and `node2`/`block2.

lnewson: Actually, I'll just move the `isChildOf` check over to `hasBetterMouseTarget`. That should avoid confusion better than it being here.

lnewson: Done

modules/tinymce/src/core/test/ts/browser/SelectionOverridesTest.ts
lnewson: Have done what I said above. How does that look?

ashfordneil: LGTM

modules/tinymce/src/core/main/ts/SelectionOverrides.ts
```diff
@@ -132,16 +131,20 @@ const SelectionOverrides = (editor: Editor): SelectionOverrides => {
...
+      if (Type.isNullable(start)) {
         return false;
       }
 
-      const startPos = CaretPosition.before(elm.firstChild);
-      const newPos = caretWalker.next(startPos);
-
-      return newPos && !isNearFakeSelectionElement(newPos);
+      const startPos = CaretPosition.before(start);
+      // If the element has a single br as a child (e.g. is empty), then the start position is a valid cursor position
```
JamesToohey, line 140:
```suggestion
      // If the element has a single br as a child (i.e. is empty), then the start position is a valid cursor position
```

or even
```suggestion
      // If the element has a single br as a child (the element is empty), then the start position is a valid cursor position
```

JamesToohey (approved):

PR #7227, opened 2021-09-26
TINY-3463: Fixed the `image` and `media` toolbar buttons showing the incorrect active state in some cases

lnewson:
Related Ticket: TINY-3463

Description of Changes:

This ensures the initial active state is set for the `image` and `media` toolbar buttons, as since this wasn't done the button didn't show as active when in a toolbar drawer or context toolbar.

Note: The associated Jira is about auditing and fixing all cases, but for now this just fixes the immediate case identified in TINY-3463.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/plugins/image/test/ts/browser/ImagePluginTest.ts
+++ b/modules/tinymce/src/plugins/image/test/ts/browser/ImagePluginTest.ts
@@ -11,9 +11,16 @@ import { advancedTabSelectors, assertInputValue, fillActiveDialog, ImageDialogDa
 describe('browser.tinymce.plugins.image.ImagePluginTest', () => {
   const hook = TinyHooks.bddSetupLight<Editor>({
     plugins: 'image',
+    toolbar: 'image',
     indent: false,
-    base_url: '/project/tinymce/js/tinymce'
-  }, [ Plugin, Theme ]);
+    base_url: '/project/tinymce/js/tinymce',
+    setup: (editor) => {
+      editor.ui.registry.addContextToolbar('test-image', {
+        predicate: (node) => node.nodeName.toLowerCase() === 'img',
+        items: 'image'
+      });
+    }
+  }, [ Plugin, Theme ], true);
 
   const pInitAndOpenDialog = async (editor: Editor, content: string, cursorPos: Cursors.CursorSpec | Cursors.RangeSpec) => {
     editor.settings.image_advtab = true;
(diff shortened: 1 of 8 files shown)


ltrouton (approved):

modules/tinymce/src/plugins/image/test/ts/browser/ImagePluginTest.ts
```diff
@@ -218,4 +225,16 @@ describe('browser.tinymce.plugins.image.ImagePluginTest', () => {
       }
     )
   );
+
+  it('TINY-3463: Ensure initial toolbar button state shows correctly', async () => {
+    const editor = hook.editor();
+    editor.setContent('<p>Content</p><p><img src="image.png"></p>');
+
+    TinySelections.setCursor(editor, [ 0, 0 ], 4);
+    await TinyUiActions.pWaitForUi(editor, '.tox-tbtn:not(.tox-tbtn--enabled)');
```
ashfordneil, line 234: Should we add `[title="..."]` or `:contains("...")` to these selectors? I'm a little worried about another toolbar button being added to the hook above without seeing this, and all of these tests no longer checking the right thing

modules/tinymce/src/plugins/media/test/ts/browser/MediaPluginSanityTest.ts
```diff
@@ -62,4 +68,16 @@ describe('browser.tinymce.plugins.media.MediaPluginSanityTest', () => {
     TinyUiActions.submitDialog(editor);
     await Utils.pAssertEditorContent(editor, '<p><video controls="controls" width="300" height="150"><source src="a" /></video></p>');
   });
+
+  it('TINY-3463: Ensure initial toolbar button state shows correctly', async () => {
+    const editor = hook.editor();
+    editor.setContent('<p>Content</p><p><iframe src="https://www.youtube.com/embed/b3XFjWInBog" width="560" height="314" allowFullscreen="1"></iframe></p>');
+
+    TinySelections.setCursor(editor, [ 0, 0 ], 4);
+    await TinyUiActions.pWaitForUi(editor, '.tox-tbtn:not(.tox-tbtn--enabled)');
```
ashfordneil, line 77: Same for these selectors

ashfordneil (approved):

modules/tinymce/src/plugins/image/test/ts/browser/ImagePluginTest.ts
lnewson: We don't anywhere else here or in other tests where we are guaranteed to only have the one toolbar button, so why do you want it in this single place?

lnewson: Sounds like this is maybe something we should discuss and change how we handle this overall? I'm happy to change it either way, but at this stage there's quite a number of places where we don't do this.

lnewson: Have added the tooltips to the CSS selectors.

PR #7230, opened 2021-09-27
TINY-7291: Fixed autoresize incorrectly scrolling to the top of the editor when resizing

### LINKED ISSUE #6250: when use autoresize plugin , click button in toolbar , the editor jump to top
Thank you very much for providing a very useful editor, but I have a headache when using it. When tinymce starts the autoresize plugin, the following picture will appear, and it will always return to the top of the editor. This experience Really bad
屏幕录制2020-11-19 上午10 50 09

**Please provide the steps to reproduce and if possible a minimal demo of the problem via [fiddle.tiny.cloud]https://fiddle.tiny.cloud/Zyhaab/6

I hope that when I click the button in the toolbar, the editor is still in the original position

tinymce: v5

### LINKED ISSUE #6714: Tinymce 5 autoresize plugin jumps scroll position after "Enter" and "Backspace"
Hello,

with the (new) Tinymce Version 5 the autoresize plugin has a Bug.
It change the scroll position after "Enter" und "Backspace" keypress.
See the Screencast below.

1. Scroll to Pos "8."
2. Put cursor position at the end
3. Press "Enter"
4. Then press "Backspace"

Problem -> It scrolls automatically up to another position but the cursor is still at the correct position.

Browser: Chrome

autoresize_failure

You can test this also in this Codepen:
https://codepen.io/Borsti26/pen/OJWxVeX

Any idea how to fix this?

BR
René

### LINKED ISSUE #7131: Autoresize plugin causes editor to jump to top
The demos I set up used CRA and Next starters. But feel free to reproduce it the same way in their official [codepen example](https://codepen.io/pen/?&prefill_data_id=6927c5f6-a2b3-4720-96aa-702c5dc7ae9b)

### Steps to reproduce
-  `git clone https://github.com/laneme/tainy`
-  If you want to use cra instead of next, `git checkout with-cra`
-  `yarn`
-  `yarn dev` or `yarn start` (depending on which branch you're using)
-  Replace the api key and/or text if you need to, and visit `localhost:3000`
-  Go around the bottom and edit/del a block or two. Then press `Ctrl+Z`. The editor should jump to top.

At first I thought it is a react/next specific issue. Maybe due to some weird re-renders. I spent quite some time to figure out it was a core autoresize issue. 

If you inspect the index.tsx file, you'll see there's no other plugins used, nothing that could cause re-renders, and no handlers attached. In fact this happens in the official codepen example too. 

I have a readme file attached to both `main` and `with-cra` branch. Check them out [here](https://github.com/laneme/tainy).

### Browsers (ubuntu 21.04)

-  Firefox 92
-  Chromium 91

They're current latest versions of ubuntu.

lnewson:
Related Ticket: TINY-7291

Description of Changes:
Fixes an issue where doing various actions in the editor would cause the `autoresize` plugin to incorrectly scroll towards the top of the editor. This was primarily caused by 2 things:

1. The common ancestor was either the body, or a large block element and the selection would therefore end up out of view since the top of the common ancestor was out of view, but the selection was still in view. To resolve this, the `editor.selection.scrollIntoView()` API has been updated to be able to scroll the current selection into view.
2. The scroll into view logic was running a bit too often. The case it was added for was inserting/pasting content into the editor, so I've constrained it to just those cases as when you're editing normally (e.g `keyup` or `NodeChange` events) we shouldn't be trying to scroll then since to be able to edit the selection should already be in view (or at least we want it to behave the same as without `autoresize` being enabled).

Note: I've also fixed the `element` type for `scrollIntoView` being inconsistent as part of this.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

Fixes #6250
Fixes #6714
Fixes #7131


modules/tinymce/CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Improved
+- The `editor.selection.scrollIntoView()` API now accepts no element to instead scroll the current selection into view #TINY-7291
```
ashfordneil, line 10:
```suggestion
- The `editor.selection.scrollIntoView()` API now scrolls the current selection into view if no element is provided #TINY-7291
```
Not sure exactly how to word it, but we should try to make it clear that this is an alternative to the current behaviour, not a replacement.

modules/tinymce/src/core/main/ts/api/dom/Selection.ts
```diff
@@ -541,7 +541,14 @@ const EditorSelection = (dom: DOMUtils, win: Window, serializer: DomSerializer,
     return scrollContainer;
   };
 
-  const scrollIntoView = (elm: HTMLElement, alignToTop?: boolean) => ScrollIntoView.scrollElementIntoView(editor, elm, alignToTop);
+  const scrollIntoView = (elm?: HTMLElement, alignToTop?: boolean) => {
+    if (Type.isNonNullable(elm)) {
+      ScrollIntoView.scrollElementIntoView(editor, elm, alignToTop);
+    } else {
+      ScrollIntoView.scrollRangeIntoView(editor, getRng(), alignToTop);
```
ashfordneil, line 548:
```suggestion
      ScrollIntoView.scrollRangeIntoView(editor, getRng());
```
`alignToTop` can't be defined here if we've taken this code path. As such, we should either omit it from the arguments, explicitly pass in false, or write a proper overload where you could call something like `scrollIntoView(alignToTop)` on its own

modules/tinymce/src/core/main/ts/api/dom/Selection.ts
```diff
@@ -100,7 +100,7 @@ interface EditorSelection {
     parents: Element[];
   }) => void) => { unbind: () => void };
   getScrollContainer: () => HTMLElement;
-  scrollIntoView: (elm: Element, alignToTop?: boolean) => void;
+  scrollIntoView: (elm?: HTMLElement, alignToTop?: boolean) => void;
```
ashfordneil, line 103: Should this wait until 6.0, for semver reasons? Technically code that used to compile will no longer compile with this change

ashfordneil (approved):

modules/tinymce/src/core/main/ts/api/dom/Selection.ts
lnewson: Code that didn't pass a `HTMLElement` previously also would have thrown an exception (since it wouldn't have been able to read the dimensions), so no this is simply a bugfix.

modules/tinymce/src/core/main/ts/api/dom/Selection.ts
lnewson: You're incorrect here sorry Neil, you can quite happily pass `scrollIntoView(undefined, false)`. So this is not something we should be doing...

modules/tinymce/src/core/main/ts/api/dom/Selection.ts
ashfordneil: Can we add a changelog line saying "fixed an incorrect type signature where `Element` could be passed into `editor.selection.scrollIntoView`" or something?

modules/tinymce/CHANGELOG.md
lnewson: This is the opposite of what we want to be saying, so I don't like this suggestion sorry.

ashfordneil:
```suggestion
- The `element` argument of the `editor.selection.scrollIntoView()` API is now optional, and if it is not provided the current selection will be scrolled into view instead #TINY-7291
```

lnewson: That seems fine if we remove the `instead` at the end?

ashfordneil: updated

modules/tinymce/src/core/main/ts/api/dom/Selection.ts
lnewson: Point... we should have that. Anyways, this change isn't really needed so if we'd prefer I'm happy to revert it (I just figured you would likely pick it up and raise it as an issue to be honest).

ashfordneil: I'm fine with either reverting or adding a changelog entry. It's always nice to move our type signatures closer to how the editor actually behaves, so I'm quite happy to keep it in.

modules/tinymce/CHANGELOG.md
```diff
@@ -15,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Resizing table columns in some scenarios would resize the column to an incorrect position #TINY-7731
 - Image resize backdrop element did not have `data-mce-bogus="all"` set #TINY-7854
 - Resize handles appeared on top of dialogs and menus when using an inline editor #TINY-3263
+- Fixed the `autoresize` plugin incorrectly scrolling to the top of the editor in some cases when changing content #TINY-7291
```
lnewson, line 21:
```suggestion
- Fixed the `autoresize` plugin incorrectly scrolling to the top of the editor in some cases when changing content #TINY-7291
- Fixed the `editor.selection.scrollIntoView()`  type signature, as it incorrectly required an `Element` instead of `HTMLElement` #TINY-7291
```

@ashfordneil how does that look for a changelog entry, re the types fix?

ashfordneil: LGTM

ltrouton (approved):

PR #7241, opened 2021-09-30
TINY-3463: Fixed the `selection.selectorChanged` API not firing correctly if the selector matched the current selection when registered

lnewson:
Related Ticket: TINY-3463

Description of Changes:

This is a followup for an issue @ltrouton found when QAing that turned out to be an existing core bug whereby the `selectorChanged` API doesn't setup the state correctly if the selector already matches the current selection.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/api/dom/SelectorChanged.ts
+++ b/modules/tinymce/src/core/main/ts/api/dom/SelectorChanged.ts
@@ -29,6 +29,12 @@ export default (dom: DOMUtils, editor: Editor) => {
   let selectorChangedData: Record<string, SelectorChangedCallback[]>;
   let currentSelectors: Record<string, SelectorChangedCallback[]>;
 
+  const matches = (selector: string, nodes: Node[]): boolean =>
+    Arr.exists(nodes, (node) => dom.is(node, selector));
+
+  const getParents = (elem: Element): Element[] =>
+    dom.getParents(elem, null, dom.getRoot());
+
   return {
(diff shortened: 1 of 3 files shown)


modules/tinymce/src/core/main/ts/api/dom/SelectorChanged.ts
```diff
@@ -29,32 +29,37 @@ export default (dom: DOMUtils, editor: Editor) => {
...
-                  currentSelectors[selector] = callbacks;
-                }
-
-                matchedSelectors[selector] = callbacks;
-                return false;
+            if (matches(selector, parents)) {
+              if (!currentSelectors[selector]) {
+                // Execute callbacks
+                Arr.each(callbacks, (callback) => {
+                  callback(true, { node, selector, parents });
```
ltrouton, line 55: In the old code, it looks like `node` passed into the callback could be a parent but in the new code, it will always be the node from the event. Just wanted to check my understanding is correct?

lnewson: You're correct, it shouldn't be doing that... I'll get that fixed.

lnewson: Fixed... this is also why shadowed variables are bad :cry:

ltrouton (approved):

ashfordneil (approved):

PR #7242, opened 2021-09-30
TINY-7998: Ensure urls are cleaned when updating

lnewson:
Related Ticket: TINY-7998

Description of Changes:
* This fixes an issue where in some cases when updating the URL input into the dialog/command was not properly cleaned.
* Adds a new `URI.isDomSafe` helper function as discussed via slack.
* Removed old `quickbars` code that had similar issues, but was not being used.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/tinymce/src/core/main/ts/api/dom/DOMUtils.ts
```diff
@@ -67,8 +67,8 @@ interface AttrHooks {
 const setupAttrHooks = (styles: Styles, settings: Partial<DOMUtilsSettings>, getContext): AttrHooks => {
   const keepValues: boolean = settings.keep_values;
   const keepUrlHook = {
-    set: ($elm, value: string, name: string) => {
-      if (settings.url_converter) {
+    set: ($elm, value: string | null, name: string) => {
+      if (settings.url_converter && value !== null) {
```
ashfordneil, line 71:
```suggestion
      if (settings.url_converter && Type.isNonNullable(value)) {
```

modules/tinymce/src/core/main/ts/api/util/URI.ts
```diff
@@ -36,6 +37,34 @@ export interface URIConstructor {
...
+}
+
+const safeSvgDataUrlElements = [ 'img', 'video' ];
+
+const blockSvgDataUris = (allowSvgDataUrls: boolean | undefined, tagName?: string) => {
+  // Only allow SVGs by default on images/videos since the browser won't execute scripts on those elements
+  if (Type.isNullable(tagName)) {
+    return true;
+  } else {
+    const allowed = Type.isNullable(allowSvgDataUrls) ? Arr.contains(safeSvgDataUrlElements, tagName) : allowSvgDataUrls;
```
ashfordneil, line 53: Could you flip this so that the top-level `if/else` is `Type.isNullable(allowSvgDataUrls)` and then the ternary is `Type.isNullable(tagName) ? !Arr.contains(...`? I just feel like it'll be a little easier to follow if you keep the tagName-related logic together

ashfordneil (approved):

modules/tinymce/src/core/main/ts/api/dom/DOMUtils.ts
lnewson: I wanted to keep this the same as below for now. If we're changing this then we probably need to change a whole lot more.

modules/tinymce/src/core/main/ts/api/util/URI.ts
lnewson: This is literally copied verbatim from the old code in `SaxParser` :shrug: but yeah we should be able to.

lnewson: Fixed

modules/tinymce/src/core/main/ts/api/dom/DOMUtils.ts
lnewson: As discussed we're going to leave this as is for now at least.

ltrouton (approved):

PR #7243, opened 2021-09-30
TINY-7917: Fire dragend event after dropping

JamesToohey:
Related Ticket: TINY-7917

Description of Changes:
* Fire `dragend` after `drop`
* Fire `dragend` event when the escape key is pressed

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/tinymce/CHANGELOG.md
```diff
@@ -30,6 +30,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed the `image` and `media` toolbar buttons showing the incorrect active state in some cases #TINY-3463
 - Inserting content into a `contenteditable="true"` element that was contained within a `contenteditable="false"` element would move the selection to an incorrect location #TINY-7842
 - The `wordcount` plugin was incorrectly treating soft hyphens (`&shy;` entities) as word breaks #TINY-7908
+- Dragging and dropping `contenteditable="false"` elements could result in the element being placed in an unexpected location #TINY-7917
+- Pressing the Escape key would not cancel a drag action that started within the editor #TINY-7917
```
JamesToohey, line 34:
```suggestion
- Pressing the Escape key would not cancel a drag action that started on a `contenteditable="false"` element within the editor #TINY-7917
```

modules/tinymce/src/core/main/ts/DragDropOverrides.ts
```diff
@@ -223,6 +223,8 @@ const drop = (state: Singleton.Value<State>, editor: Editor) => (e: EditorEvent<
             editor._selectionOverrides.hideFakeCaret();
           });
         }
+
+        editor.fire('dragend');
       }
```
lnewson, line 228:
```suggestion
      }
      
      editor.fire('dragend');
```
This is in the incorrect place from what I can tell, as it should fire if dragging regardless of if the drop target is valid.

lnewson (approved):

modules/tinymce/src/core/main/ts/DragDropOverrides.ts
```diff
@@ -267,6 +269,13 @@ const bindFakeDragEvents = (editor: Editor) => {
     pageDom.unbind(rootDocument, 'mousemove', dragHandler);
     pageDom.unbind(rootDocument, 'mouseup', dragEndHandler);
   });
+
+  editor.on('keydown', (e) => {
+    // Fire 'dragend' when the escape key is pressed
+    if (e.keyCode === 27) {
```
ashfordneil, line 275: Can we use a `Keys.ts` or `VK` constant or something?

modules/tinymce/src/core/test/ts/browser/DragDropOverridesTest.ts
```diff
@@ -56,6 +60,42 @@ describe('browser.tinymce.core.DragDropOverridesTest', () => {
...
   });
 
+  it('TINY-7917: Dropping draggable element inside editor fires dragend event', () => {
+    const editor = hook.editor();
+    editor.setContent('<p contenteditable="false">a</p><p>bc123</p>');
+    const target = UiFinder.findIn(TinyDom.body(editor), 'p:contains("a")').getOrDie();
+    const targetRect = target.dom.getBoundingClientRect();
+
+    const dest = UiFinder.findIn(TinyDom.body(editor), 'p:contains("bc123")').getOrDie();
+    const destRect = dest.dom.getBoundingClientRect();
```
ashfordneil, line 70: These could use the `SugarPosition` API

ashfordneil (approved):

modules/tinymce/src/core/main/ts/DragDropOverrides.ts
```diff
@@ -11,6 +11,7 @@ import DOMUtils from './api/dom/DOMUtils';
 import { EventUtilsEvent } from './api/dom/EventUtils';
 import EditorSelection from './api/dom/Selection';
 import Editor from './api/Editor';
+import { VK } from './api/PublicApi';
```
lnewson, line 14: This is the wrong import and as such it's causing a circular dependency. It should be this (not a suggestion as it needs to be moved to the right place):
```
import VK from '../api/util/VK';
```

modules/tinymce/src/core/main/ts/api/util/VK.ts
```diff
@@ -19,6 +19,7 @@ interface VK {
   DELETE: number;
   DOWN: number;
   ENTER: number;
+  ESCAPE: number;
```
lnewson, line 22: We said in the call this should be called `ESC` instead of `ESCAPE`?

- [ ] Rename to `ESC`

lnewson: Also this will need a changelog entry to say we've added it to the `VK` API.

JamesToohey: I didn't hear about calling it `ESC` in the call, but I definitely prefer `ESCAPE` (we also call it `escape` in Agar).

lnewson: I mentioned it in the call and we said yes at the time. I largely mentioned it because this won't be minified since it's on the public API and the actual key on keyboards is `Esc`.

JamesToohey: I made this change, but I do really feel it should have been `ESCAPE` for consistency (e.g. `DELETE`, `PAGE_DOWN`, `SPACEBAR`).

ltrouton (approved):

modules/tinymce/src/core/main/ts/api/util/VK.ts
lnewson: To be clear, this doesn't have to be `ESC`, I was just pointing out it's different to what had been mentioned in the call as what we should do. Maybe lets ask at standup today again, as it's likely others weren't paying much attention either yesterday afternoon.

PR #7301, opened 2021-10-14
Use aria-controls instead of aria-owns in DropdownUtils

### LINKED ISSUE #7300: NVDA & JAWS no longer read sub-menu entries in Chrome 93.0.4556.0+
## What is the current behavior? Describe the bug
When using arrow keys to navigate sub-menus in the TinyMCE menubar, NVDA + JAWS no longer read sub-menu items.

Example:
menu-demo
In the above example, NVDA and JAWS do not read the sub-menu items ("New document", "Preview", etc.) in Chrome  93.0.4556.0 and up

## Steps to reproduce
1. Navigate to the [TinyMCE full-featured demo](https://www.tiny.cloud/docs/demo/full-featured/) (or any TinyMCE instance with a menu bar) in Chrome 93.0.4556.0 or greater
2. Turn on NVDA or JAWS
3. Use the arrow keys to navigate the TinyMCE sub-menus (the "File" sub-menu, for example)
4. Notice that the screen reader fails to read the sub-menu items

## Suggested Fix
Currently, TinyMCE adds an `aria-owns` attribute to the `button` element that opens the sub-menu ("File" button in the above example). This `aria-owns` attribute points to the div containing the sub-menu.

Removing the `aria-owns` attribute from the button elements solves the problem in my testing. I'm assuming `aria-owns` was added to the button because the sub-menu is not nested hierarchically in markup with the button.

W3C ARIA examples and best practices ([here](https://www.w3.org/TR/2016/WD-wai-aria-practices-1.1-20161214/examples/menu-button/menu-button-1/menu-button-1.html), for example) do not recommend using `aria-owns` on a button.

I believe doing so is communicating to assistive technologies that the sub-menu div is nested _inside_ the button, which seems to be causing the issue.

## What is the expected behavior?
Screen readers should read sub-menu items in Chrome

## Which versions of TinyMCE, and which browser / OS are affected by this issue? Did this work in previous versions of TinyMCE?
I've tested this issue in TinyMCE 5.9.2 and 5.8.1, it was reproducible in both.  Chrome must be 93.0.4556.0 or greater to reproduce the issue. Tested with latest NVDA and JAWS

Thanks!

westonkd:
**Description of Changes:**
Don't set `aria-owns` in DropdownUtils.

This fixes the issues with assistive technologies mentioned in #7300.

Test Plan:
* Verify #7300 is resolved
* Verify each of the following continues to work properly with keyboard navigation and screen readers:
   - Dropdowns
   - SplitDropdowns

**Pre-checks:**
* [x] Changelog entry added

**Review:**
* [x] Milestone set

**GitHub issues**: fixes #7300


TheSpyder:
Sorry we didn't look at this earlier. The change seems reasonable, but we'll need to do some accessibility testing before we accept it.

TheSpyder:
This triggered a lot of discussion and research for those of us still working on Christmas eve :)

We now agree with you, `aria-owns` is the wrong way to link these UI components. However since they are disconnected but related components, we don't think removing the link is correct either. Would `aria-controls` be more appropriate?

westonkd:
@TheSpyder Thanks for looking into this! And yes, I agree adding `aria-controls` would be a good path.

> The change to swap to aria-controls is likely to involve a fair bit of code, we are happy to work on that ourselves, we're just trying to figure out the best path forward.

Awesome, thanks! I was interested in taking a stab at it (even if just as an exercise) and pushed up a revision. Feel free to use it or close this PR and continue work yourselves.

I'm going to also do a bit of manual testing with NVDA/JAWS to verify `aria-controls` gives us a good set of behaviors.

westonkd:
Whoops, looks like there is a merge conflict. I'll take care of that in a bit.

lnewson (approved):
Hi @westonkd, sorry we've largely been away over the holiday period so just getting back to this now. The new approach looks good thanks! I've just left some style related comments to fix up. Also could you please add a Fixed changelog entry in the Unreleased section for the Alloy `CHANGELOG.md` file please?

westonkd:
@lnewson Thanks for the feedback! I've made the requested changes.

metricjs (approved):
Thanks for this contribution @westonkd !

An update from our side: We are planning to merge this PR into a feature branch so that we can add a couple of things, such as proper attribution to yourself, and perform our usual testing before merging it into develop. That will hopefully happen in the next couple of weeks. We'll let you know if anything changes.

westonkd:
Sounds great @metricjs! Thanks.

TheSpyder:
This was merged into TinyMCE 6 via #7591

PR #7350, opened 2021-11-01
TINY-7484: Remove zero width space before counting words

JamesToohey:
Related Ticket: TINY-7484

Description of Changes:
Removes the `&#8203;`/`\u200B` character before counting words

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/plugins/wordcount/main/ts/core/Count.ts
+++ b/modules/tinymce/src/plugins/wordcount/main/ts/core/Count.ts
@@ -14,11 +14,14 @@ import { getText } from './GetText';
 
 export type Counter = (node: Node, schema: Schema) => number;
 
+const removeZwsp = (text: string) =>
+  text.replace('\u200B', '');
+
 const strLen = (str: string): number =>
   str.replace(/[\uD800-\uDBFF][\uDC00-\uDFFF]/g, '_').length;
 
 const countWords: Counter = (node: Node, schema: Schema): number => {
-  const text = getText(node, schema).join('\n');
+  const text = removeZwsp(getText(node, schema).join('\n'));
   return Words.getWords(text.split(''), Fun.identity).length;
 };
 
--- a/modules/tinymce/src/plugins/wordcount/main/ts/core/Count.ts
+++ b/modules/tinymce/src/plugins/wordcount/main/ts/core/Count.ts
@@ -15,12 +15,14 @@ import { getText } from './GetText';
 export type Counter = (node: Node, schema: Schema) => number;
 
 const removeZwsp = (text: string) =>
-  text.replace('\u200B', '');
+  text.replace(/\u200B/g, '');
 
 const strLen = (str: string): number =>
   str.replace(/[\uD800-\uDBFF][\uDC00-\uDFFF]/g, '_').length;
 
 const countWords: Counter = (node: Node, schema: Schema): number => {
+  // TINY-7484: The grapheme word boundary logic used by Polaris states a ZWSP (\u200B) should be treated as
+  // a word boundary, however word counting normally does not consider it as anything so we strip it out. 
   const text = removeZwsp(getText(node, schema).join('\n'));
   return Words.getWords(text.split(''), Fun.identity).length;
 };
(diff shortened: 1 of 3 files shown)


modules/tinymce/src/plugins/wordcount/main/ts/core/Count.ts
```diff
@@ -14,11 +14,14 @@ import { getText } from './GetText';
...
 
+const removeZwsp = (text: string) =>
+  text.replace('\u200B', '');
+
 const strLen = (str: string): number =>
   str.replace(/[\uD800-\uDBFF][\uDC00-\uDFFF]/g, '_').length;
 
 const countWords: Counter = (node: Node, schema: Schema): number => {
-  const text = getText(node, schema).join('\n');
+  const text = removeZwsp(getText(node, schema).join('\n'));
```
lnewson, line 24: Can we add a comment here as to why we're removing this here? Something like this maybe?
```suggestion
  // TINY-7484: The grapheme word boundary logic states a ZWSP (\u200B) should be treated as a word
  // boundary however word counting normally does not consider it as anything so we strip it out. 
  const text = removeZwsp(getText(node, schema).join('\n'));
```

modules/tinymce/src/plugins/wordcount/test/ts/browser/PluginTest.ts
```diff
@@ -63,4 +63,19 @@ describe('browser.tinymce.plugins.wordcount.PluginTest', () => {
     editor.setContent('<p>Soft hy&shy;phen</p>');
     await pWaitForWordcount(2);
   });
+
+  it('TINY-7484: Does not treat zwsp character as splitting a word', async () => {
```
lnewson, line 67: Can we please add a test for character counting here, as it should still be included there?

modules/tinymce/src/plugins/wordcount/main/ts/core/Count.ts
```diff
@@ -14,11 +14,14 @@ import { getText } from './GetText';
 
 export type Counter = (node: Node, schema: Schema) => number;
 
+const removeZwsp = (text: string) =>
+  text.replace('\u200B', '');
```
lnewson, line 18: This isn't going to work properly... this will only remove the first instance of a ZWSP character. It needs to be something like this instead:
```suggestion
  text.replace(/\u200B/g, '');
```
Can we please make sure to include multiple characters in one of the tests?

lnewson (changes requested):

modules/tinymce/src/plugins/wordcount/test/ts/browser/PluginTest.ts
lnewson: and I just noticed nothing in this tests characters :cry: we can probably ignore that now and log a tech debt task to add tests for that.

modules/tinymce/src/plugins/wordcount/main/ts/core/Count.ts
JamesToohey: Whoops

modules/tinymce/src/plugins/wordcount/main/ts/core/Count.ts
```diff
@@ -14,11 +14,16 @@ import { getText } from './GetText';
...
+const removeZwsp = (text: string) =>
+  text.replace(/\u200B/g, '');
+
 const strLen = (str: string): number =>
   str.replace(/[\uD800-\uDBFF][\uDC00-\uDFFF]/g, '_').length;
 
 const countWords: Counter = (node: Node, schema: Schema): number => {
-  const text = getText(node, schema).join('\n');
+  // TINY-7484: The grapheme word boundary logic used by Polaris states a ZWSP (\u200B) should be treated as
+  // a word boundary, however word counting normally does not consider it as anything so we strip it out.
```
lnewson:
```suggestion
  // a word boundary, however word counting normally does not consider it as anything so we strip it out.
```
Accidentally left a space here sorry which is causing linter errors :cry:

lnewson (approved):

ltrouton (approved):

PR #7367, opened 2021-11-08
TINY-8207: Added theme registration our mcagar wrapper file. Removed all manual theme registration in tests.

TheSpyder:
Added theme registration our `mcagar` wrapper file.
Removed all manual theme registration in tests.

Related Ticket: TINY-8207

Description of Changes:
* Deleted a lot of code. Feels good.
* search/replace + `eslint --fix` = winning
* Only two significant files to review:
  * `modules/tinymce/src/core/test/ts/module/McAgar.ts`
  * `package.json` where I added an `eslint-fix` script, it's quite handy

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [x] ~Milestone set~

### DIFF
--- a/modules/tinymce/src/core/test/ts/module/McAgar.ts
+++ b/modules/tinymce/src/core/test/ts/module/McAgar.ts
@@ -1,7 +1,9 @@
+/* eslint-disable @tinymce/no-main-module-imports */
 /*
   This is a helper module that allows us to ensure that TinyMCE core is included/compiled
   before mcagar loads. This ensures that the core of the editor isn't loaded from pre-built
   files in node_modules and is always the latest version.
  */
 import 'tinymce';
+import 'tinymce/themes/silver/Main';
 export * from '@ephox/mcagar';
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -69,7 +69,7 @@ runs `tsc -b -w` for those times when you don't need to iterate in the browser.
 an alias to `tsc -b` just in case you forget
 
 `yarn eslint`
-runs `eslint` in _all_ projects, with a rule set that is far more strict than most projects were previously subject to. This is a good source of things to improve when bored.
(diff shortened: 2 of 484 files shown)


package.json
```diff
@@ -16,6 +16,7 @@
     "tsc": "tsc -b",
     "watch": "tsc -b -w",
     "eslint": "lerna exec --stream --no-bail -- eslint src/**/*.ts",
+    "eslint-fix": "yarn -s eslint --fix",
```
lnewson, line 19: Why is this needed? You can just do `yarn eslint --fix` and it's only 1 or 2 extra characters. This is how all our repos are setup so I'm not keen on adding a custom command for this :shrug:

TheSpyder: maybe not everyone knows eslint can fix things for them 🤷‍♂️

modules/tinymce/src/core/test/ts/module/McAgar.ts
```diff
@@ -1,7 +1,10 @@
+/* eslint-disable @tinymce/no-main-module-imports */
 /*
   This is a helper module that allows us to ensure that TinyMCE core is included/compiled
   before mcagar loads. This ensures that the core of the editor isn't loaded from pre-built
   files in node_modules and is always the latest version.
  */
 import 'tinymce';
 export * from '@ephox/mcagar';
+
+import 'tinymce/themes/silver/Main';
```
lnewson, line 10:
```suggestion
import 'tinymce/themes/silver/Main';
export * from '@ephox/mcagar';
```
This should probably be registered before loading the mcagar components. Overall I don't think it'll matter, but this makes more logical sense as it's setting up TinyMCE and then exporting the components to interact with it.

lnewson (approved):

modules/tinymce/src/core/test/ts/module/McAgar.ts
TheSpyder: Oh interesting. The tests pass (both here and on the model branch), but that makes sense.

package.json
lnewson: Maybe we should add something to the Contributing docs section instead then?

JamesToohey (approved):

package.json
TheSpyder: Done, see what you think

lnewson: LGTM :+1:

ashfordneil (approved):

PR #7373, opened 2021-11-09
TINY-8206: Added new Options API and converted core to make use of it

lnewson:
Related Ticket: TINY-8206

Description of Changes:

This implements a replacement API for controlling Settings (now called Options to match the docs). The new Options API requires options to be registered before they can be used. When registering an option TinyMCE will now require a spec to be passed containing a mandatory `processor` which is used to validate and transform/normalise the option and then other optional metadata, such as if the option is immutable or deprecated (some metadata isn't used yet).

As part of this to validate the new API I've also switched the core to use this new API.

Note: This is targeting a feature branch instead of `develop` since we're going to have to do a few more changes before this is fully ready.

Pre-checks:
* [x] Changelog entry added
* [x] Branch prefixed with `feature/` for new features (if applicable)x

Review:
* [x] Milestone set


modules/tinymce/src/core/main/ts/EditorSettings.ts
```diff
@@ -73,73 +63,43 @@ const getSectionConfig = (sectionResult: SectionResult, name: string) => {
...
   return {
     ...baseDefaults,
-    ...isTouch ? defaultTouchSettings : { }
+    ...isTouch ? defaultTouchOptions : { }
   };
 };
 
-const getDefaultMobileSettings = (mobileSettings: RawEditorSettings, isPhone: boolean): RawEditorSettings => {
-  const defaultMobileSettings: RawEditorSettings = {
+const getDefaultMobileOptions = (mobileOptions: RawEditorOptions, isPhone: boolean): RawEditorOptions => {
```
lnewson, line 81: This logic should be able to be removed once we migrate the theme/plugins as it will be responsible for setting the defaults when registering.

modules/tinymce/src/core/main/ts/api/Editor.ts
```diff
@@ -72,7 +73,7 @@ import WindowManager from './WindowManager';
 export interface EditorConstructor {
   readonly prototype: Editor;
 
-  new (id: string, settings: RawEditorSettings, editorManager: EditorManager): Editor;
+  new (id: string, settings: RawEditorOptions, editorManager: EditorManager): Editor;
```
JamesToohey, line 76: Should this be `options`?

```suggestion
  new (id: string, options: RawEditorOptions, editorManager: EditorManager): Editor;
```

modules/tinymce/src/core/main/ts/api/EditorOptions.ts
```diff
@@ -0,0 +1,277 @@
...
+ * @class tinymce.editor.Options
+ */
+
+export interface Options {
+  /**
+   * Register a new option that can be used within TinyMCE.
+   *
+   * @method register
+   * @param {String} name Name of an option.
+   * @param {OptionSpec} spec A option spec containing how to valid the option and other optional metadata.
```
JamesToohey, line 71:
```suggestion
   * @param {OptionSpec} spec An option spec describing how to validate the option and other optional metadata.
```

modules/tinymce/src/core/main/ts/api/EditorOptions.ts
```diff
@@ -0,0 +1,277 @@
...
+ *
+ * @class tinymce.editor.Options
+ */
+
+export interface Options {
+  /**
+   * Register a new option that can be used within TinyMCE.
+   *
+   * @method register
+   * @param {String} name Name of an option.
```
JamesToohey, line 70:
```suggestion
   * @param {String} name Name of the option.
```

modules/tinymce/src/core/main/ts/api/EditorOptions.ts
```diff
@@ -0,0 +1,277 @@
...
+  };
+
+  /**
+   * Clear any set value for the specified option so that it'll revert back to using the default value.
+   *
+   * @method unset
+   * @param {String} name Name of a registered option.
+   * @return {Boolean} True if the option value was successfully reset, otherwise false.
+   */
+  unset: (name: string) => boolean;
```
JamesToohey, line 120: I'm wondering if this should return something different if it's reverted to a default vs. cleared entirely?

modules/tinymce/src/core/main/ts/api/EditorOptions.ts
```diff
@@ -0,0 +1,277 @@
...
+      case 'object':
+        return Type.isObject;
+      case 'string':
+        return Type.isString;
+      case 'string[]':
+        return (val) => Type.isArrayOf(val, Type.isString);
+    }
+  })();
+
+  return (value) => ({ value, valid: validator(value), message: `The value must be a ${type}.` });
```
JamesToohey, line 143: Just because there are some 'an's and 'a's
```suggestion
  return (value) => ({ value, valid: validator(value), message: `The value must be: ${type}.` });
```

modules/tinymce/src/core/main/ts/api/EditorOptions.ts
```diff
@@ -0,0 +1,277 @@
...
+};
+
+const isBuiltInSpec = <K extends BuiltInOptionType>(spec: unknown): spec is BuiltInOptionSpec<K> =>
+  Type.isString((spec as BuiltInOptionSpec<K>).processor);
+
+const getErrorMessage = (message: string, result: ProcessorError): string => {
+  const additionalText = !Strings.isEmpty(result.message) ? '. ' + result.message : '';
```
JamesToohey, line 150:
```suggestion
  const additionalText = Strings.isEmpty(result.message) ? '' : '. ' + result.message;
```

modules/tinymce/src/core/main/ts/api/EditorOptions.ts
```diff
@@ -0,0 +1,277 @@
...
+};
+
+const processDefaultValue = <T, U>(name: string, defaultValue: T, processor: Processor<T, U>): U | undefined => {
+  if (!Type.isUndefined(defaultValue)) {
+    const result = processValue(defaultValue, processor);
+    if (isValidResult(result)) {
+      return result.value;
+    } else {
+      // eslint-disable-next-line no-console
+      console.error(getErrorMessage(`Invalid default value passed for the ${name} option`, result));
```
JamesToohey, line 174: Just in case the option is named something weird like "option"
```suggestion
      console.error(getErrorMessage(`Invalid default value passed for the "${name}" option`, result));
```

modules/tinymce/src/core/main/ts/api/EditorOptions.ts
```diff
@@ -0,0 +1,277 @@
...
+  };
+
+  const isRegistered = (name: string) =>
+    Obj.has(registry, name);
+
+  const get = (name: string) =>
+    Obj.get(values, name).getOrUndefined();
+
+  const set = <T>(name: string, value: T) => {
+    if (!isRegistered(name)) {
```
JamesToohey, line 236: I personally prefer non-negated if statements if possible, but not terribly fussed

modules/tinymce/src/core/main/ts/api/EditorOptions.ts
```diff
@@ -0,0 +1,277 @@
...
+      // eslint-disable-next-line no-console
+      console.warn(`${name} is not a registered option. Ensure the option has been registered before setting a value.`);
```
JamesToohey, line 238:
```suggestion
      console.warn(`"${name}" is not a registered option. Ensure the option has been registered before setting a value.`);
```

modules/tinymce/src/core/main/ts/api/EditorOptions.ts
```diff
@@ -0,0 +1,277 @@
...
+      return false;
+    } else {
+      const spec = registry[name];
+      if (spec.immutable) {
+        // eslint-disable-next-line no-console
+        console.error(`${name} is an immutable option and cannot be updated`);
```
JamesToohey, line 244:
```suggestion
        console.error(`"${name}" is an immutable option and cannot be updated`);
```

modules/tinymce/src/core/main/ts/api/EditorOptions.ts
lnewson: That changes the meaning of this and will make people think they need to put exactly the type as the value.

modules/tinymce/src/core/main/ts/api/EditorOptions.ts
lnewson: I'm not a fan of this as it makes it harder to read.

modules/tinymce/src/core/main/ts/api/EditorOptions.ts
JamesToohey:
```suggestion
  return (value) => ({ value, valid: validator(value), message: `The value must be of type: ${type}.` });
```

modules/tinymce/src/core/main/ts/api/EditorOptions.ts
lnewson: I did it this way because it's the way we conceptualise/walk through the process of what should happen.

modules/tinymce/src/core/main/ts/api/EditorOptions.ts
lnewson: I don't like `the` here as it makes it sound like it's a specific option name you need to pass, not any option name.

modules/tinymce/src/core/main/ts/api/Editor.ts
lnewson: It should be thanks :+1:

modules/tinymce/src/core/main/ts/api/EditorOptions.ts
JamesToohey: I think we should probably have @tylerkelly13 run through all this at some point tbh

lnewson: Yeah we need him to review this at some stage. I was going to add him to the reviewers list when I convert it from a draft.

modules/tinymce/src/core/main/ts/api/EditorOptions.ts
JamesToohey: I always find negated ternaries rough, and also like when the `?` and `:` are closer together

what about

```suggestion
  const additionalText = Strings.isEmpty(result.message) ? '' : `. ${result.message}`;
```

lnewson: That I can live with :+1:

modules/tinymce/src/core/main/ts/api/EditorOptions.ts
lnewson: I mean the default is just undefined for everything if not defined, so I think this is fine personally as from an end users perspective it won't make a difference.

modules/tinymce/src/core/main/ts/api/EditorOptions.ts
```diff
@@ -0,0 +1,277 @@
...
+        return Type.isString;
+      case 'string[]':
+        return (val) => Type.isArrayOf(val, Type.isString);
+    }
+  })();
+
+  return (value) => ({ value, valid: validator(value), message: `The value must be of type: ${type}.` });
+};
+
+const isBuiltInSpec = <K extends BuiltInOptionType>(spec: unknown): spec is BuiltInOptionSpec<K> =>
```
TheSpyder, line 146: Isn't this more `isBuiltInProcessor`? I think predefined processors that can be referenced by name is a great idea.

modules/tinymce/src/core/main/ts/init/InitContentBody.ts
```diff
@@ -66,61 +66,67 @@ const getRootName = (editor: Editor): string => editor.inline ? editor.getElemen
...
-    convert_fonts_to_spans: settings.convert_fonts_to_spans,
-    fix_list_elements: settings.fix_list_elements,
-    font_size_legacy_values: settings.font_size_legacy_values,
-    forced_root_block: settings.forced_root_block,
-    forced_root_block_attrs: settings.forced_root_block_attrs,
-    padd_empty_with_br: settings.padd_empty_with_br,
-    preserve_cdata: settings.preserve_cdata,
-    remove_trailing_brs: settings.remove_trailing_brs,
-    inline_styles: settings.inline_styles,
+    allow_conditional_comments: getOption('allow_conditional_comments'),
```
TheSpyder, line 73: Random question unrelated to this code, how does the switch to strings impact autocomplete of available settings? It just occurred to me this might regress some of the integrator usability improvements that came with publishing types.

TheSpyder (commented):
Looks good so far!

modules/tinymce/src/core/main/ts/init/InitContentBody.ts
lnewson: It didn't impact it at all, as you still get autocomplete suggestions for any settings we've defined in the `OptionTypes` file.

modules/tinymce/src/core/main/ts/api/EditorOptions.ts
lnewson: No, it is actually the spec and the spec defines the processor. The reason it's coupled is because of we need to ensure the relevant types for other properties as well (e.g. the default value matches the processor type)

lnewson: As a side note to this... I wonder if we need a way to define/register built-in processors :thinking: Probably something for later, but I could also see that being quite useful :shrug:

modules/tinymce/src/core/main/ts/api/EditorOptions.ts
```diff
@@ -0,0 +1,277 @@
...
+
+  return undefined;
+};
+
+const create = (editor: Editor, initialOptions: Record<string, unknown>): Options => {
+  const registry: Record<string, OptionSpec<any, any>> = {};
+  const values: Record<string, any> = {};
+
+  editor.on('init', () => {
+    const unregisteredOptions = Arr.partition(Obj.keys(initialOptions), isRegistered).fail;
```
spocke, line 186: Why use a partition if you only care about the fail case. Use `Arr.filter(Obj.keys(initialOptions), not(isRegistered))` instead. But I guess it would have been nice to have an Arr.reject or something that is the inverse of Array.filter.

modules/tinymce/src/core/main/ts/api/EditorOptions.ts
```diff
@@ -0,0 +1,277 @@
...
+    }
+  };
+
+  const register = (name: string, spec: BuiltInOptionSpec<any> | OptionSpec<any, any>) => {
+    const processor = isBuiltInSpec(spec) ? getBuiltInProcessor(spec.processor) : spec.processor;
+
+    // Process and validate the default value
+    const defaultValue = processDefaultValue(name, spec.default, processor);
+
+    // Register the spec with the normalised validator
```
spocke, line 213: `normalized` if we are sticking to US English.

modules/tinymce/src/core/main/ts/api/Options.ts
```diff
@@ -0,0 +1,810 @@
...
+    }
+  } else {
+    return { value, valid, message: 'Must be a string.' };
+  }
+};
+
+const register = (editor: Editor) => {
+  const registerOption = editor.options.register;
+
+  registerOption('id', {
```
spocke, line 55: I wonder if we should have a bulk register function instead or if register should be able to take just a object. So you could do.
```
editor.options.register({
  selector: { processor: 'string' },
  target: { processor: 'object' },
  referrer_policy: { processor: 'string', default: '' }
})
```
It would be far less verbose if you have many settings like this. If we don't want it in the public api then we should at least consider it for this file since it will be come pretty large. It could also be reduced by having predefined register functions for the common types. For this large file might be worth having things like `registerString('id', editor.id)` to avoid having a object structure for every single option.

spocke (approved):
Looks really cool there is a lot of nice things we can do with this processor concept. More predefined ones like `int`, `id-string`, `url`, or `toolbar-items` would have been cool if we supported multiple fixed values as well like `{ processor: { oneof: ['a', 'b', 'c'] } }` or `{ processor: { anyoff: ['a', 'b', 'c'] } }` so many things I guess if we start seeing common processors as we go we could make them predefined ones.

modules/tinymce/src/core/main/ts/api/OptionTypes.ts
```diff
@@ -211,15 +212,23 @@ interface BaseEditorSettings {
...
-  mobile?: RawEditorSettings;
+  mobile?: RawEditorOptions;
   plugins?: string | string[];
 }
 
-// EditorSettings.ts processes the plugins setting to turn it into a string
+// NormaliseOptions.ts processes the plugins setting to turn it into a string
 // and merges in the mobile settings
-export interface EditorSettings extends BaseEditorSettings {
+export interface NormalisedEditorOptions extends BaseEditorOptions {
```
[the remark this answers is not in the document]
lnewson: Thanks I often forget that 😞

modules/tinymce/src/core/main/ts/api/Options.ts
lnewson: We had discussed a `registerAll` function like this but opted to leave it out until actually needed. Happy to re-evaluate and include that now if needed (though the types for that could be nasty).

As for `registerString` we hadn't mentioned this specific case, but in general we specifically don't want to be using function arguments like that as it severely locks in what properties you can use. An object we can continuously add mandatory or optional properties, but that can't be done when its function arguments without making a breaking change.

modules/tinymce/src/core/main/ts/api/EditorOptions.ts
lnewson: Yeah I did consider that, but I preferred this way of handling it, though I don't have any strong reasons. I'll swap to using `Arr.filter` with `Fun.not` then.

modules/tinymce/src/core/main/ts/api/Options.ts
spocke: Yes both a registerAll or a registerString can be local functions to the editor options module. The registerString was more of an alternative to the full register with all properties. Most options will be simple things boolean, strings etc so it seems very verbose to have a config for these simple things as you can editor default options file shows it becomes a lot of duplicated similar objects. Like { processor: 'boolean' } 30 times. But I guess most can be worked around by local functions like the option function I see here and there in this PR.

modules/tinymce/src/core/main/ts/api/EditorOptions.ts
spocke: Just feels wrong to use a branching function and not care about one of the sides making an array that not used at all. But maybe we should consider having an array filter that is the inverse out of scope here but might be useful.

modules/tinymce/src/core/main/ts/api/Options.ts
lnewson: I wonder if instead we could have a public API such as `registerString(name: string, spec: Omit<BuiltInOptionSpec, 'processor'>)` (or similar). That way you wouldn't have to define the processor every time, but would still alleviate some of the syntax duplication...

Anyways, you're right we can use an internal function for big files such as this. We may end up with some duplication elsewhere though amongst the individual plugins though so I'll discuss the above more before going down that path.

lnewson: Discussed this at stand up and the team we're keen on adding public APIs just yet. We also don't really gain anything from the internal `registerAll` function and instead are potentially going to lose type information. So for now we're going to keep this as is and re-evaluate once we've converted more before we do the final merge into `develop`.

modules/tinymce/src/core/main/ts/api/EditorOptions.ts
```diff
@@ -0,0 +1,277 @@
...
+  value: T;
+}
+
+interface ProcessorError {
+  valid: false;
+  message: string;
+}
+
+type SimpleProcessor<T> = (value: T) => boolean;
+type Processor<T, U> = (value: T) => ProcessorSuccess<U> | ProcessorError;
```
ashfordneil, line 24:
```suggestion
type SimpleProcessor<T> = (value: unknown) => value is T;
type Processor<T> = (value: unknown) => ProcessorSuccess<T> | ProcessorError;
```

It just occurred to me that if these processors are trying to replace `Type.isString` or `Type.isNumber` then they would need to do that initial "is this the right type" check before they get to any other logic they might be looking at.

ashfordneil (approved):

modules/tinymce/src/core/main/ts/api/EditorOptions.ts
```diff
@@ -0,0 +1,277 @@
...
+ * For commercial licenses see https://www.tiny.cloud/
+ */
+
+import { Arr, Fun, Obj, Optional, Strings, Type } from '@ephox/katamari';
+
+import Editor from './Editor';
+import { EditorOptions, NormalizedEditorOptions } from './OptionTypes';
+
+interface ProcessorSuccess<T> {
+  valid: true;
```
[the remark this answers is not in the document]
lnewson: Our current rule is that we should never add `readonly` on a TinyMCE public API.

modules/tinymce/src/core/main/ts/api/EditorOptions.ts
lnewson: This is incorrect and will cause issues with public usage, as the `value is T` won't work if you need to do this:
```
const isValid = (value: boolean) => value === false
```
So effectively this is **not** a type guard, it's just a regular predicate so the original is correct here.

ashfordneil: My issue is someone trying to write

```
const isMySecretKeyValid = (value: string) => value.length >= 32
```

and then passing in a number for the setting. We might not be able to make it fully a type guard, but it definitely should take `unknown` as an argument to hint to the users that they'll need to do some form of check

lnewson: Disagree very much there and passing `unknown` is going to cause a number of issues with integrators trying to use this API. In 99% of cases the type will be correct and this is the type of thing where it's external input so we need to assume the type is correct.

lnewson: I should add I do agree that `unknown` is technically the correct type here, don't forget we need to keep this easy to use for others. It's effectively the same as getting a server response whereby the you're just asserting the expected types as you're not going to type guard every little detail from the server response.

modules/tinymce/src/core/main/ts/api/EditorOptions.ts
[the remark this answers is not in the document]
TheSpyder: Might be an interesting rule to revisit, but maybe I'll keep that in the back of my mind for 7.

TheSpyder (approved):

modules/tinymce/src/core/main/ts/api/EditorOptions.ts
ashfordneil: I think the type of users who are enabling typescript and using advanced features of the TinyMCE API (like adding your own plugin with its own setting) are the types of users who would rather trade correctness in types for a little ease of use. Remember that they can just use plain JS, or `ts-ignore`, if its too much of a hassle. It would also make it easier on us developing our own plugins.

lnewson: I still disagree but I don't care enough to argue anymore so have made the change anyways.

modules/tinymce/src/core/main/ts/api/EditorOptions.ts
tylerkelly13: I prefer "Name of the option", we refer to it as the option in the spec doc string. it's the option you're registering, not a specific, ambiguous option

modules/tinymce/src/core/main/ts/api/EditorOptions.ts
```diff
@@ -0,0 +1,284 @@
...
+ * @class tinymce.editor.Options
+ */
+
+export interface Options {
+  /**
+   * Register a new option that can be used within TinyMCE.
+   *
+   * @method register
+   * @param {String} name Name of an option.
+   * @param {OptionSpec} spec An option spec describing how to validate the option and other optional metadata.
```
tylerkelly13, line 71:
```suggestion
   * @param {OptionSpec} spec An option spec describing how to validate the option with other optional metadata.
```
little vague, could be read as "describing how to validate the option and how to validate the other optional metadata" although it that's what it does, you should keep "and" and add a "both" after describing?

modules/tinymce/src/core/main/ts/api/EditorOptions.ts
```diff
@@ -0,0 +1,284 @@
...
+    <K extends keyof NormalizedEditorOptions>(name: K, spec: OptionSpec<NormalizedEditorOptions[K], EditorOptions[K]> | SimpleOptionSpec<NormalizedEditorOptions[K]>): void;
+    <T, U>(name: string, spec: OptionSpec<T, U>): void;
+    <T>(name: string, spec: SimpleOptionSpec<T>): void;
+  };
+
+  /**
+   * Check if an option has been registered.
+   *
+   * @method isRegistered
+   * @param {String} name Name of an option.
```
tylerkelly13, line 84:
```suggestion
   * @param {String} name Name of the option.
```
If you change the one above, then lets change this one too

modules/tinymce/src/core/main/ts/api/EditorOptions.ts
```diff
@@ -0,0 +1,284 @@
...
+   * @return {Boolean} True if the option has already been registered, otherwise false.
+   */
+  isRegistered: (name: string) => boolean;
+
+  /**
+   * Get the value of a registered option.
+   *
+   * @method get
+   * @param {String} name Name of a registered option.
+   * @return {Object} An option value or the registered default if not set.
```
tylerkelly13, line 94: what happens if you try to `get` an option that hasn't been registered?

modules/tinymce/src/core/main/ts/api/EditorOptions.ts
```diff
@@ -0,0 +1,284 @@
...
+    <K extends keyof EditorOptions>(name: K): EditorOptions[K] | undefined;
+    <T>(name: string): T | undefined;
+  };
+
+  /**
+   * Set the value for a registered option.
+   *
+   * @method set
+   * @param {String} name Name of a registered option.
+   * @return {Boolean} True if the option value was successfully set, otherwise false.
```
tylerkelly13, line 106: I'll assume `false` is returned if you forgot to register the option first?

modules/tinymce/src/core/main/ts/api/EditorOptions.ts
```diff
@@ -0,0 +1,284 @@
...
+   */
+  set: {
+    <K extends keyof EditorOptions>(name: K, value: EditorOptions[K]): boolean;
+    <T>(name: string, value: T): boolean;
+  };
+
+  /**
+   * Clear any set value for the specified option so that it'll revert back to using the default value.
```
tylerkelly13, line 114:
```suggestion
   * Clear the set value for the specified option and revert back to the default value.
```
`the` to be specific?

modules/tinymce/src/core/main/ts/api/EditorOptions.ts
lnewson: That's correct

modules/tinymce/src/core/main/ts/api/EditorOptions.ts
lnewson: It'll return undefined

modules/tinymce/src/core/main/ts/api/EditorOptions.ts
lnewson:
```suggestion
   * @param {String} name Name of the option.
```

modules/tinymce/src/core/main/ts/api/EditorOptions.ts
lnewson:
```suggestion
   * @return {Object} An option value, the registered default if not set or undefined if not registered.
```

modules/tinymce/src/core/main/ts/api/EditorOptions.ts
```diff
@@ -0,0 +1,284 @@
...
+   * @return {Boolean} True if the option has already been registered, otherwise false.
+   */
+  isRegistered: (name: string) => boolean;
+
+  /**
+   * Get the value of a registered option.
+   *
+   * @method get
+   * @param {String} name Name of a registered option.
+   * @return {Object} An option value, the registered default if not set or undefined if not registered.
```
lnewson, line 94:
```suggestion
   * @return {Object} An option value, the registered default if not set, or undefined if not registered.
```

modules/tinymce/src/core/main/ts/api/EditorOptions.ts
TheSpyder: I didn't get a chance to say but I'm on Lee's side here. Although we could perhaps keep `unknown` and also alleviate the ease-of-use concerns by providing pre-built extensible processors that do a type guard and then run a custom function to validate.

PR #7379, opened 2021-11-15
TINY-8238: Exclude more internal selectors from the imported CSS list

TheSpyder:
Related Ticket:TINY-8238

Description of Changes:
* Switch from a simple `.mce-` check to a regex that covers all the ways we phrase our content CSS selectors

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/plugins/importcss/main/ts/core/ImportCss.ts
+++ b/modules/tinymce/src/plugins/importcss/main/ts/core/ImportCss.ts
@@ -33,7 +33,7 @@ interface Group extends UserDefinedGroup {
   readonly filter: Filter | undefined;
 }
 
-const internalEditorStyle = /^\.(?:ephox|tiny|mce)(?:-+\w+)+$/;
+const internalEditorStyle = /^\.(?:ephox|tiny-pageembed|mce)(?:-+\w+)+$/;
 
 const removeCacheSuffix = (url: string): string => {
   const cacheSuffix = Env.cacheSuffix;
--- a/modules/tinymce/src/plugins/importcss/main/ts/core/ImportCss.ts
+++ b/modules/tinymce/src/plugins/importcss/main/ts/core/ImportCss.ts
@@ -257,7 +257,7 @@ const setup = (editor: Editor): void => {
     };
 
     Tools.each(getSelectors(editor, editor.getDoc(), compileFilter(Settings.getFileFilter(editor))), (selector) => {
-      if (selector.match(internalEditorStyle) === null) {
+      if (!internalEditorStyle.test(selector)) {
         if (!selectorFilter || selectorFilter(selector)) {
           const selectorGroups = getGroupsBySelector(groups, selector);
 
--- a/modules/tinymce/src/plugins/importcss/main/ts/core/ImportCss.ts
+++ b/modules/tinymce/src/plugins/importcss/main/ts/core/ImportCss.ts
@@ -33,6 +33,8 @@ interface Group extends UserDefinedGroup {
   readonly filter: Filter | undefined;
 }
 
+const INTERNAL_EDITOR_STYLE = /^\.(?:ephox|tiny|mce)(?:-+\w+)+$/;
+
 const removeCacheSuffix = (url: string): string => {
   const cacheSuffix = Env.cacheSuffix;
 
@@ -255,7 +257,7 @@ const setup = (editor: Editor): void => {
     };
 
     Tools.each(getSelectors(editor, editor.getDoc(), compileFilter(Settings.getFileFilter(editor))), (selector) => {
-      if (selector.indexOf('.mce-') === -1) {
+      if (selector.match(INTERNAL_EDITOR_STYLE) === null) {
         if (!selectorFilter || selectorFilter(selector)) {
           const selectorGroups = getGroupsBySelector(groups, selector);
 
--- a/modules/tinymce/src/plugins/importcss/main/ts/core/ImportCss.ts
+++ b/modules/tinymce/src/plugins/importcss/main/ts/core/ImportCss.ts
@@ -33,7 +33,7 @@ interface Group extends UserDefinedGroup {
   readonly filter: Filter | undefined;
 }
 
-const INTERNAL_EDITOR_STYLE = /^\.(?:ephox|tiny|mce)(?:-+\w+)+$/;
+const internalEditorStyle = /^\.(?:ephox|tiny|mce)(?:-+\w+)+$/;
 
 const removeCacheSuffix = (url: string): string => {
   const cacheSuffix = Env.cacheSuffix;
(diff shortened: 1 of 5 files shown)


TheSpyder:
Test failures are IE failing to load, and IE timing out.

modules/tinymce/src/plugins/importcss/main/ts/core/ImportCss.ts
```diff
@@ -33,6 +33,8 @@ interface Group extends UserDefinedGroup {
   readonly filter: Filter | undefined;
 }
 
+const INTERNAL_EDITOR_STYLE = /^\.(?:ephox|tiny|mce)(?:-+\w+)+$/;
```
JamesToohey, line 36: I don't think I've seen many all caps snake case constants, is this common?

modules/tinymce/src/plugins/importcss/main/ts/core/ImportCss.ts
```diff
@@ -255,7 +257,7 @@ const setup = (editor: Editor): void => {
     };
 
     Tools.each(getSelectors(editor, editor.getDoc(), compileFilter(Settings.getFileFilter(editor))), (selector) => {
-      if (selector.indexOf('.mce-') === -1) {
+      if (selector.match(INTERNAL_EDITOR_STYLE) === null) {
```
JamesToohey, line 260: Should we use `Type.isNull` or `Type.isNullable` from Katamari here?

modules/tinymce/src/plugins/importcss/test/ts/browser/ImportCssPluginTest.ts
```diff
@@ -316,4 +316,22 @@ describe('browser.tinymce.plugins.importcss.ImportCssTest', () => {
       ]
     }
   ));
+
+  it('TBA: content_css with two files, basic and internal CSS classes', () => pTestEditorWithSettings(
```
JamesToohey, line 320:
```suggestion
  it('TINY-8238: content_css with two files, basic and internal CSS classes', () => pTestEditorWithSettings(
```

modules/tinymce/src/plugins/importcss/main/ts/core/ImportCss.ts
TheSpyder: I swear that used to be the style... I guess I should've checked. Two years away from the code and I forget some things.

modules/tinymce/src/plugins/importcss/main/ts/core/ImportCss.ts
TheSpyder: I thought that was only really useful when composing nullable checks 🤔 the standard says it will literally be `null` if there are no matches, so this seemed right.

modules/tinymce/src/plugins/importcss/test/ts/browser/ImportCssPluginTest.ts
TheSpyder: Ah of course. I am guilty of copy/pasting.

modules/tinymce/src/plugins/importcss/main/ts/core/ImportCss.ts
TheSpyder: I don't see very many module-level constants but those I have found are camel case not upper-snake case. I'll change it.

modules/tinymce/src/plugins/importcss/main/ts/core/ImportCss.ts
```diff
@@ -255,7 +257,7 @@ const setup = (editor: Editor): void => {
     };
 
     Tools.each(getSelectors(editor, editor.getDoc(), compileFilter(Settings.getFileFilter(editor))), (selector) => {
-      if (selector.indexOf('.mce-') === -1) {
+      if (selector.match(internalEditorStyle) === null) {
```
lnewson, line 260:
```suggestion
      if (!internalEditorStyle.test(selector)) {
```
`test` is faster in most cases than `match` and we don't care about using the match group, so it avoids the `null` check and the various points around it.

modules/tinymce/src/plugins/importcss/main/ts/core/ImportCss.ts
```diff
@@ -33,6 +33,8 @@ interface Group extends UserDefinedGroup {
   readonly filter: Filter | undefined;
 }
 
+const internalEditorStyle = /^\.(?:ephox|tiny|mce)(?:-+\w+)+$/;
```
lnewson, line 36: I sort of hinted at this on the Jira, but should have been more explicit sorry but `tiny` is a common term so I don't think we can do this. For example, this will match `tiny-image` or `tiny-text`, so unfortunately I think we're going to have to use `tiny-pageembed` explicitly here even though it's not very generic or future proof (really we should never use `tiny` for internal class names).

- [ ] Reconsider matching on `tiny-`.

lnewson (approved):
Thanks for getting this done!

modules/tinymce/src/plugins/importcss/main/ts/core/ImportCss.ts
lnewson: Normally we'd using the `Type` check as it'll compile to smaller code and keeps things consistent.

modules/tinymce/src/plugins/importcss/main/ts/core/ImportCss.ts
lnewson: We've never used upper snake case since I've been here fwiw. I remember doing this when I first started and got pinged because it wasn't our normal style 🤷

TheSpyder: yeah I definitely forget these things after so much time away, and having used upper-snake case on previous codebases 😂

modules/tinymce/src/plugins/importcss/main/ts/core/ImportCss.ts
TheSpyder: we're talking about 4 bytes difference which will probably be zero after gzip... I don't even need to change it, looking at your later request :)

modules/tinymce/src/plugins/importcss/main/ts/core/ImportCss.ts
TheSpyder: right, I had a vague memory there was a better way but forgot with everything else going on. Thanks!

modules/tinymce/src/plugins/importcss/main/ts/core/ImportCss.ts
TheSpyder: hmm. I don't think it'll be _that_ much of a problem, it would need to be a customer using styles like that _and also_ the import css plugin (which I don't believe is very popular). But if we're going to avoid using `tiny-*` class names in future this makes sense.

lnewson: The problem is we just really have no actual data on what is being used, so this change could completely break someones integration. We definitely should avoid using `tiny-` prefixed class names for internal plugins in future though and stick to either `ephox` or `mce` prefixes.

ltrouton (approved):

ashfordneil (approved):

PR #7390, opened 2021-11-18
TINY-8249: Separate parsing logic from DOM fixing logic in DomParser

ashfordneil:
Related Ticket: TINY-8249, I think. It's a bit hand-wavey at the moment.

Description of Changes:
* I found the giant chunk of code in `DomParser.parse` that deals with creating, configuring, and calling the `SaxParser` — and I moved it out into its own function (`builder`)
* I took the 3 data structures it was populating as it built the tree (`invalidChildren`, `matchedNodes`, and `matchedAttributes`) — and created a separate function to walk the tree and populate these data structures (`walker`)
* Because it didn't make sense to have them defined on the class itself instead of just inside the `parse` function, I packed all 3 of those variables into an object, and passed that object as an argument into functions that needed it 
* I found the `filterNode` function we had which operated on that state object — and I moved it to a standalone method instead of a class method, because it doesn't really make sense to call it on the class itself. This is a breaking change, though, and I'm open to discussions about reverting this if someone can find a way to reconcile fact that the method is defined at the `DomParser` scope, and the state it operates on is only relevant inside a single `parse` call

Pre-checks:
* [x] Changelog entry added — **not yet**: waiting on confirmation for the `filterNode` removal)
* [x] ~Tests have been added (if applicable)~ all the existing tests (should) pass, and nothing has been added or changed

Review:
* [x] Milestone set


spocke (commented):
If I recall the filterNode was made to be able to filter within a filter. So if a filter adds a new element then you could re-execute the filter on that node. That should probably be input to the filter it self since it has context to the current parse cycle. However there seems to be no code using this anymore so it can likely be removed.
But it's a bit of a performance vs correctness. In theory we should re-run all filters on all changed nodes by a filter but that would be pretty expensive to do.

metricjs (approved):
Approving this as a "these changes make sense in isolation and I can't see anything wrong, but I'll leave whether these make sense in the bigger picture to someone else" way.

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
```diff
@@ -79,21 +79,333 @@ interface DomParser {
...
+      // Check if we are inside a whitespace preserved element
+      if (!isInWhiteSpacePreservedElement && whiteSpaceElements[name]) {
+        isInWhiteSpacePreservedElement = true;
+      }
+    },
+
+    end: (name) => {
+      let textNode, text, sibling, tempNode;
+
+      const elementRule = settings.validate && schema.getElementRule(node.name) || {} as SchemaElement;
```
lnewson, line 227: You're still doing schema validation here by the looks?

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
```diff
@@ -79,21 +79,333 @@ interface DomParser {
...
+
+    return output;
+  };
+
+  const parser = SaxParser({
+    validate: settings.validate,
+    document: settings.document,
+    allow_html_data_urls: settings.allow_html_data_urls,
+    allow_svg_data_urls: settings.allow_svg_data_urls,
+    allow_script_urls: settings.allow_script_urls,
```
lnewson, line 155: These also do validation so this would need to be moved.

lnewson: There's also parts in `SaxParser` that handle sanitizing the attributes. I'm not sure if we can move them in this PR but it's something to keep in mind.

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
ashfordneil: This used to be this:
```typescript
const elementRule: Partial<SchemaElement> = validate ? schema.getElementRule(name) : {};
```
which would validate using the schema. What we have now, with the `schema.getElementRule(node.name) || {}` doesn't involve any validation — it just looks at the schema (with fallback) to inform the whitespace / empty-element removal logic

lnewson:
```
const elementRule = settings.validate && schema.getElementRule(node.name) || {} as SchemaElement;
```
is the same as what you said the original was?
```
const elementRule: Partial<SchemaElement> = validate ? schema.getElementRule(name) : {};
```
as in the `validate` true case you're still going to end up looking up the schema and getting the element rule and in the false case you're going to use the empty object.

For this to be the same as a native parser it cannot use either `validate` or `schema.getElementRule`.

ashfordneil: They both look up the element rule, the difference is in what they do when that element rule can't be found / doesn't exist. Previously it would end up with `const elementRule = null` and then the node gets completely skipped. Now you'd end up with `const elementRule = null || {}`, which means the node still gets processed we just don't apply any specific logic to it. We do need to move towards it not referencing anything at all, which is what I'm working on now on a branch off of this, but it hasn't happened yet.

lnewson: Right I'd missed that because it's a bit hacky and not something we'd normally do. Anyways, I've looked at this more and the `elementRule` is only used for adding padding or removing empty elements. As we said on Friday that can't be in the builder section and needs to be moved out to a later section. So I can't approve this until we've done the separation properly sorry.

lnewson: Alternatively we need to change the scope of this so it's only meant to be a step 1 of separating the logic into the various components/steps that we're going to require moving forward.

lnewson: Discussed this more at stand up and we're moving to the iterative approach whereby this single change can't be released without other PRs.

lnewson (approved):

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
ashfordneil: This has been pushed to the next task

PR #7391, opened 2021-11-18
TINY-8206: Added the isSet function to the editor options API

lnewson:
Related Ticket: TINY-8206/TINY-8235

Description of Changes:
* Adds an `isSet` API to see if a custom value has been set for an option

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/api/EditorOptions.ts
+++ b/modules/tinymce/src/core/main/ts/api/EditorOptions.ts
@@ -5,7 +5,7 @@
  * For commercial licenses see https://www.tiny.cloud/
  */
 
-import { Arr, Fun, Obj, Optional, Strings, Type } from '@ephox/katamari';
+import { Arr, Fun, Obj, Optional, Optionals, Strings, Type } from '@ephox/katamari';
 
 import Editor from './Editor';
 import { EditorOptions, NormalizedEditorOptions } from './OptionTypes';
@@ -118,6 +118,15 @@ export interface Options {
    * @return {Boolean} True if the option value was successfully reset, otherwise false.
    */
   unset: (name: string) => boolean;
+
+  /**
+   * Checks to see if a value has been set for the specified option.
+   *
+   * @method isSet
+   * @param {String} name Name of the option.
+   * @return {Boolean} True if the option has a value set, otherwise false.
+   */
+  isSet: (name: string) => boolean;
 }
 
 const getBuiltInProcessor = <K extends BuiltInOptionType>(type: K): Processor<BuiltInOptionTypeMap[K]> => {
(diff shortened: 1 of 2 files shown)


ashfordneil (approved):

tylerkelly13 (commented):
API docs strings - LGTM

modules/tinymce/src/core/main/ts/api/EditorOptions.ts
```diff
@@ -187,7 +196,7 @@ const processDefaultValue = <T, U>(name: string, defaultValue: T, processor: Pro
 
 const create = (editor: Editor, initialOptions: Record<string, unknown>): Options => {
   const registry: Record<string, OptionSpec<any, any>> = {};
-  const values: Record<string, any> = {};
+  const values: Record<string, { value: any; isDefault: boolean }> = {};
```
spocke, line 199: Is this extra `isDefault` flag really needed you could just have it get the default value from the registry in the get function as a fallback. Sure it would make two lookups but that is usually very fast. So if you unset a value the key is deleted from the `values` and then get will return the fallback value from the `registry`. Is set would then just need to check if the value is defined in values or not.
But maybe I'm not reading the code right or missing something.

spocke (approved):

modules/tinymce/src/core/main/ts/api/EditorOptions.ts
lnewson: The main reason was performance related so that we don't have to keep looking up 2 objects all the time and creating multiple `Optional` instances. I'm probably over optimizing here though so I'll swap it over.

lnewson: Okay have made that change, so does that look better/more like what you were thinking?

spocke: Right yeah hopefully we don't pull the settings object directly that often that this will be a performance issue. If we do we can always optimize that at the call site by using a variable. This change is so much cleaner and makes the data structures simpler.

spocke: Another option would be to not use Obj.get and options at all in the `get` function and just do key lookup directly on the objects and use isUndefined calls etc to check if they exist or not. But that could be done if it becomes a hot spot.

lnewson: Yeah I did consider doing that atm but figured we were better to just see if we do get any perf issues and optimise then instead as this really shouldn't since settings aren't looked up too often.

metricjs (approved):

PR #7392, opened 2021-11-18
TINY-8235: Convert plugins A-C to use the new options API

lnewson:
Related Ticket: TINY-8235

Description of Changes:

This is the first in a set of 5 or 6 PRs to convert all the plugins/themes to use the new Options API. Note that in some cases the old `settings` API may still be used for things that aren't yet migrated (e.g. settings belonging to the theme)

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [x] Milestone set


modules/tinymce/src/plugins/autosave/main/ts/core/Time.ts
```diff
@@ -5,16 +5,15 @@
...
-const parse = (timeString: string | undefined, defaultTime: string): number => {
+const parse = (timeString: string): number => {
   const multiples: Record<string, number> = {
     s: 1000,
     m: 60000
   };
-  const toParse = (timeString || defaultTime);
 
-  const parsedTime = /^(\d+)([ms]?)$/.exec('' + toParse);
+  const parsedTime = /^(\d+)([ms]?)$/.exec(timeString);
```
ashfordneil, line 14: Should we test whether this regex fails, and pass it as an error message back up to the option validator?

ashfordneil (approved):

modules/tinymce/src/plugins/autosave/main/ts/core/Time.ts
lnewson: Out of scope for these changes.

modules/tinymce/src/plugins/autolink/main/ts/api/Options.ts
```diff
@@ -0,0 +1,48 @@
...
+  <K extends keyof EditorOptions>(name: K): (editor: Editor) => EditorOptions[K] | undefined;
+  <T>(name: string): (editor: Editor) => T | undefined;
+} = (name: string) => (editor: Editor) =>
+  editor.options.get(name);
+
+const register = (editor: Editor): void => {
+  const registerOption = editor.options.register;
+
+  registerOption('autolink_pattern', {
+    processor: Type.isNonNullable,
```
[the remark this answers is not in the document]
lnewson: That won't work if the Regexp is created in another frame so I don't want to be touching that here... if we do want this we should look at adding a building `regexp` processor or similar.

metricjs (approved):

modules/tinymce/src/plugins/advlist/main/ts/api/Options.ts
```diff
@@ -0,0 +1,49 @@
...
+const option: {
+  <K extends keyof EditorOptions>(name: K): (editor: Editor) => EditorOptions[K] | undefined;
+  <T>(name: string): (editor: Editor) => T | undefined;
+} = (name: string) => (editor: Editor) =>
+  editor.options.get(name);
+
+const register = (editor: Editor): void => {
+  const registerOption = editor.options.register;
+
+  const stringListProcessor = (value: unknown) => {
```
TheSpyder, line 22: I can already tell "composable built-in processors" are going to be high on my wishlist for this feature. Maybe not for 6.0, but at some point. Everything here except `value.split(/[ ,]/)` could come from the core.

modules/tinymce/src/plugins/autolink/main/ts/core/Keys.ts
```diff
@@ -176,7 +176,7 @@ const parseCurrentLine = (editor: Editor, endOffset: number): void => {
     editor.selection.setRng(rng);
     editor.execCommand('createlink', false, url);
 
-    if (defaultLinkTarget !== false) {
+    if (Type.isString(defaultLinkTarget)) {
```
TheSpyder, line 179: oh wow, a sneaky type mismatch. Good catch!

TheSpyder (approved):

spocke (approved):

modules/tinymce/src/plugins/advlist/main/ts/api/Options.ts
lnewson: Yeah it would definitely be quite useful. There's quite a few places we just need to check if it's a `string` or `boolean` (or something similar to that).

PR #7400, opened 2021-11-19
TINY-8235: Convert the silver theme to use the new options API

lnewson:
Related Ticket: TINY-8235

Description of Changes:

List PR in the set of 6 PRs to convert all the plugins/themes to use the new Options API. This also updates McAgar to work with the new Options API as without this tests fail.

Note: I also realised we need to keep the mobile specific overrides in `NormalizeOptions` because we want them to always override the desktop options. So if someone sets say `toolbar_mode: 'sliding'` in the main options it should remain `scrolling` on mobile unless overridden in the `mobile` config section.

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [x] Milestone set


modules/mcagar/src/main/ts/ephox/mcagar/alien/Settings.ts
```diff
@@ -0,0 +1,31 @@
+import { Editor } from './EditorTypes';
+
+const getSetting = <R>(editor: Editor, name: string): R => {
```
spocke, line 3: Maybe it will be just confusing to any future developer that these functions are called `getSetting` since the options name is the way forward.

modules/mcagar/src/main/ts/ephox/mcagar/api/pipeline/ApiChains.ts
```diff
@@ -57,13 +58,13 @@ const cSetSelection = <T extends Editor>(startPath: number[], soffset: number, f
 
 const cSetSetting = <T extends Editor>(key: string, value: any): Chain<T, T> => {
```
spocke, line 59: Should we deprecate these and add new cSetOption and setOption to mcagar and over time switch to that?

modules/tinymce/src/core/main/ts/api/Options.ts
```diff
@@ -646,6 +655,11 @@ const register = (editor: Editor) => {
     default: false
   });
 
+  registerOption('content_editable_state', {
```
spocke, line 658: Maybe add a todo about this going away or just remove it now?

modules/tinymce/src/plugins/autoresize/main/ts/Plugin.ts
```diff
@@ -23,10 +23,6 @@ import * as Resize from './core/Resize';
 export default (): void => {
   PluginManager.add('autoresize', (editor) => {
     Options.register(editor);
-    // If autoresize is enabled, disable resize if the user hasn't explicitly enabled it
```
spocke, line 26: Curious why was this removed? Will this render a resize handle when the autoresize plugin is enabled or is all that now part of the theme and this is no longer needed?

modules/tinymce/src/themes/silver/main/ts/ui/core/color/Options.ts
```diff
@@ -0,0 +1,117 @@
+/**
```
spocke, line 1: I think if you rename files do that before you do other changes in the same commit and it should be easier to diff. If more than 70% of the file is changed it not picked up as a rename. But I think if you do it in two steps it is.

modules/tinymce/src/themes/silver/main/ts/ui/statusbar/ResizeHandle.ts
```diff
@@ -11,14 +11,14 @@ import { SugarPosition } from '@ephox/sugar';
...
 import Editor from 'tinymce/core/api/Editor';
 
+import * as Options from '../../api/Options';
 import { UiFactoryBackstageProviders } from '../../backstage/Backstage';
 import * as Icons from '../icons/Icons';
 import { resize, ResizeTypes } from '../sizing/Resize';
 
 const getResizeType = (editor: Editor): ResizeTypes => {
   // If autoresize is enabled, disable resize
-  const fallback = !editor.hasPlugin('autoresize');
```
spocke, line 20: What effects does this have on resizing and the autoresize plugin?

spocke (approved):

modules/tinymce/src/core/main/ts/api/Options.ts
lnewson: We have a separate Jira to clean this up so don't want to touch it just yet.

modules/tinymce/src/plugins/autoresize/main/ts/Plugin.ts
lnewson: Logic already existed in Silver for this so I just removed this as it doesn't make sense to have it in 2 places.

modules/tinymce/src/themes/silver/main/ts/ui/statusbar/ResizeHandle.ts
lnewson: Oops the above comment was meant to be removed as well... it has no difference on functionality as this just moved to the default in when we register the option.

modules/tinymce/src/themes/silver/main/ts/ui/statusbar/ResizeHandle.ts
```diff
@@ -11,14 +11,14 @@ import { SugarPosition } from '@ephox/sugar';
 
 import Editor from 'tinymce/core/api/Editor';
 
+import * as Options from '../../api/Options';
 import { UiFactoryBackstageProviders } from '../../backstage/Backstage';
 import * as Icons from '../icons/Icons';
 import { resize, ResizeTypes } from '../sizing/Resize';
 
 const getResizeType = (editor: Editor): ResizeTypes => {
   // If autoresize is enabled, disable resize
```
lnewson, line 20:
```suggestion
```

metricjs (approved):

modules/mcagar/src/main/ts/ephox/mcagar/api/pipeline/ApiChains.ts
lnewson: It's already listed in the `cleanup` settings Jira that's meant to be done after everything is converted. We probably should just mention to rename it there as I really didn't want to touch McAgar here but had to unfortunately... At the moment it's still very much geared more at 4.x/5.x than at 6.x.

modules/tinymce/src/core/main/ts/api/Options.ts
```diff
@@ -369,6 +369,15 @@ const register = (editor: Editor) => {
...
   });
 
+  registerOption('skin', {
+    processor: (value) => Type.isString(value) || value === false,
+    default: 'oxide'
+  });
+
+  registerOption('skin_url', {
+    processor: 'string'
+  });
```
lnewson, line 379: Forgot to mention this when I logged the PR but I was tossing up between whether this should be in Core or in the Silver Theme. Would be keen to get others thoughts?

TheSpyder: Ooh, tough choice. It depends on whether the skin setting is ever read in core (other than to download the skin, although perhaps that counts).

The skin is not useful without the theme, so I suggest putting the registration in the theme, unless that would cause problems with downloading the skin.

lnewson: The skin option I don't believe is ever read in core, as it's loaded by the theme. The reason I placed it in core is it's used by other plugins (e.g. template, preview and exports)

lnewson: Anyways yeah I really just flipped a coin here... I do agree thinking more on this that it should be in the theme, as each theme may have a different default skin.

TheSpyder (approved):

modules/mcagar/src/main/ts/ephox/mcagar/alien/Settings.ts
lnewson: Have renamed this and deprecated the old functions

modules/mcagar/src/main/ts/ephox/mcagar/api/pipeline/ApiChains.ts
lnewson: I did this anyways since it was quick...

modules/tinymce/src/plugins/autoresize/main/ts/Plugin.ts
lnewson: I've reinstated this but as we found out today it's never done anything since 5.0. The reason being the theme is looking up this and rendering the resize handle before the plugins initialise. I've logged TINY-8288 to fix that at some stage.

modules/tinymce/src/themes/silver/main/ts/ui/core/color/Options.ts
lnewson: Good to know, I wasn't aware of that so I'll try and remember to give it a go in future.

PR #7431, opened 2021-12-01
TINY-8168: Now fire input events when overriding events

HAFRMO:
Related Ticket:

Description of Changes:
Fire input and beforeinput events

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/tinymce/src/core/test/ts/browser/keyboard/SpaceKeyTest.ts
```diff
@@ -16,12 +16,19 @@ describe('browser.tinymce.core.keyboard.SpaceKeyTest', () => {
 
   context('Space key around inline boundary elements', () => {
     it('Press space at beginning of inline boundary inserting nbsp', () => {
+      let inputEvents = 0;
       const editor = hook.editor();
       editor.setContent('<p>a <a href="#">b</a> c</p>');
+      editor.on('beforeinput input', (e) => {
+        // eslint-disable-next-line no-console
+        console.log(e.isComposing);
+        inputEvents++;
```
spocke, line 25: Better to use an array to capture e.type and assert that since this could be firing `input` twice and still work.

modules/tinymce/src/core/main/ts/keyboard/EnterKey.ts
```diff
@@ -10,13 +10,14 @@ import { EditorEvent } from '../api/util/EventDispatcher';
...
 import { endTypingLevelIgnoreLocks } from '../undo/TypingState';
+import { mockInputEvents } from './MockInputEvents';
 
 const handleEnterKeyEvent = (editor: Editor, event: EditorEvent<KeyboardEvent>) => {
   if (event.isDefaultPrevented()) {
     return;
   }
 
-  event.preventDefault();
+  mockInputEvents(editor, { inputType: 'insertParagraph', cancelable: false }, event)();
```
spocke, line 20: This needs to fire insertLineBreak on shift+enter so better to move this to the InsertNewLine module if it has a event being passed in then you can fire different events depending on action. Also should this be default preventable this one is easier to do that on. You could fire before input then check isDefaultPrevented on the event then not do the insert actions and not fire the input events. So if you split up the mockInputEvents into two functions mockBeforeInput and mockInput that sets all properties and then fires the event then you could do this in the backspace/delete case maybe rename them to fireFakeBeforeInput and fireFakeInput since they are mock could mean that they just make a mocked event object not actually triggering the event.
```
...
.each(() => {
  fireFakeInputEvent(editor, { inputType: 'deleteContentBackward' }, evt);
  evt.preventDefault();
})
```
and for the newline/paragraph case something like this it also needs to check if evt is something so not exactly like this more of a proof of concept in the InsertNewLine module:
```
const insert = (editor: Editor, evt?: EditorEvent<KeyboardEvent>) => {
  NewLineAction.getAction(editor, evt).fold(
    () => {
      const evt = fireFakeBeforeInputEvent(editor, { inputType: 'insertLineBreak' }, evt);
      if (!evt.isDefaultPrevented()) {
         InsertBr.insert(editor, evt);
         fireFakeInputEvent(editor, { inputType: 'insertLineBreak' }, evt);
      }
    },
    () => {
      const evt = fireFakeBeforeInputEvent(editor, { inputType: 'insertParagraph' }, evt);
      if (!evt.isDefaultPrevented()) {
        InsertBlock.insert(editor, evt);
        fireFakeInputEvent(editor, { inputType: 'insertParagraph' }, evt);
      }
    },
    Fun.noop
  );
};

```

modules/tinymce/src/core/test/ts/browser/keyboard/SpaceKeyTest.ts
```diff
@@ -32,7 +39,7 @@ describe('browser.tinymce.core.keyboard.SpaceKeyTest', () => {
       editor.nodeChanged();
       TinyContentActions.keystroke(editor, Keys.space());
       TinyAssertions.assertSelection(editor, [ 0, 1, 0 ], 2, [ 0, 1, 0 ], 2);
-      TinyAssertions.assertContent(editor, '<p>a <a href="#">b&nbsp;</a> c</p>');
+      TinyAssertions.assertContent(editor, '<p>a <a href="#">b </a> c</p>');
```
lnewson, line 42: This change is not correct and something you've done has introduced a regression/bug as this should not have changed.

modules/tinymce/src/core/test/ts/browser/keyboard/SpaceKeyTest.ts
```diff
@@ -16,12 +16,19 @@ describe('browser.tinymce.core.keyboard.SpaceKeyTest', () => {
 
   context('Space key around inline boundary elements', () => {
     it('Press space at beginning of inline boundary inserting nbsp', () => {
+      let inputEvents = 0;
       const editor = hook.editor();
       editor.setContent('<p>a <a href="#">b</a> c</p>');
+      editor.on('beforeinput input', (e) => {
+        // eslint-disable-next-line no-console
+        console.log(e.isComposing);
```
lnewson, line 24: No console logs please

modules/tinymce/src/core/test/ts/browser/keyboard/SpaceKeyTest.ts
```diff
@@ -16,12 +16,19 @@ describe('browser.tinymce.core.keyboard.SpaceKeyTest', () => {
...
+        inputEvents++;
+      });
       TinySelections.setCursor(editor, [ 0, 1, 0 ], 0);
       editor.nodeChanged();
       TinyContentActions.keystroke(editor, Keys.space());
       TinyAssertions.assertSelection(editor, [ 0, 1, 0 ], 1, [ 0, 1, 0 ], 1);
+      assert.eq(2, inputEvents, 'Events not fired as expected');
```
lnewson, line 31: This doesn't test anything meaningful... as it doesn't test the order nor that you actually got a beforeinput and an  input. You could have ended up getting 2 beforeinput events instead, so please improve this to make sure events are firing in the correct order.

modules/tinymce/src/core/main/ts/keyboard/EnterKey.ts
lnewson, line 20: This is incorrect in some cases, as this can insert either a new paragraph or a line break. So this will need to be `insertLineBreak` or `insertParagraph` depending on what `InsertNewLine.insert` does.

modules/tinymce/src/core/test/ts/browser/keyboard/SpaceKeyTest.ts
spocke, line 24: Is it important to have this console log here?

spocke (changes requested):
This also needs more tests fore the backspace / delete case and the insert newline and insert paragraph case.

lnewson (changes requested):
This was a good attempt but I'm afraid it's not something we can merge in it's current form. The problem is that we need the events to follow the same event order as the browser would. That means that it needs to be something like this:

1. keydown
2. keypress (may or may not fire depending on the key pressed and is also deprecated)
3. beforeinput 
4. (content changed)
5. input
6. keyup

You'll see in your changes `beforeinput` fires after the content has been changed in most cases, but in EnterKeys.ts it actually fires both before the change instead. Another issue is that you have pretty much no tests for this change so please make sure to add some to make sure it works as expected for deleting and enter/space overrides.

modules/tinymce/src/core/main/ts/keyboard/MockInputEvents.ts
```diff
@@ -0,0 +1,52 @@
...
+      srcElement: event.srcElement,
+      metaKey: event.metaKey
+    };
+
+    const beforeSpecific = {
+      ...overrides,
+      cancelable: true,
+    };
+
+    const beforeInput = clone(new window.InputEvent('beforeinput'));
```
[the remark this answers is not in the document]
lnewson: It's because it's a dom-global and the ESlint rules prevent it from being used directly. We can add it to the allowed list which might be a better option here: https://github.com/tinymce/tinymce/blob/7fe5eb819417935aaa468848b03446200aa18a00/modules/tinymce/.eslintrc.json#L10

modules/tinymce/src/core/main/ts/keyboard/DeleteBackspaceKeys.ts
```diff
@@ -43,9 +44,7 @@ const executeKeydownOverride = (editor: Editor, caret: Cell<Text>, evt: Keyboard
     { keyCode: VK.DELETE, action: MatchKeys.action(BlockBoundaryDelete.backspaceDelete, editor, true) },
     { keyCode: VK.BACKSPACE, action: MatchKeys.action(InlineFormatDelete.backspaceDelete, editor, false) },
     { keyCode: VK.DELETE, action: MatchKeys.action(InlineFormatDelete.backspaceDelete, editor, true) }
-  ], evt).each((_) => {
-    evt.preventDefault();
-  });
+  ], evt).each(mockInputEvents(editor, { inputType: 'deleteContentBackward' }, evt));
```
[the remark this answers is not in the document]
lnewson: Yeah `beforeinput` is definitely trickier since it involves a fair bit of refactoring to make it happen before the content changes and to make it preventable. Looking at the Zendesk ticket it does look like `input` was the primary issue so if needed we can most likely look at splitting this into 2 tasks: one for `beforeinput` and one for `input`.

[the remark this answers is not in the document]
HAFRMO: I'll drop the `beforeinput` from this one then, and just make an issue for it to be implemented as well at some point in the future.

modules/tinymce/src/core/test/ts/browser/keyboard/SpaceKeyTest.ts
HAFRMO: Are you sure this is not correct? What's happening with this is that we are triggering inputkeys listener which runs `normalizeNbspsInEditor`. If it isn't correct, do we have a better way than ignoring events with `data: ' '`?

modules/tinymce/src/core/test/ts/browser/keyboard/SpaceKeyTest.ts
HAFRMO: No, that one was a mistake, my bad.

modules/tinymce/src/core/test/ts/browser/keyboard/SpaceKeyTest.ts
lnewson: 100% sure this is incorrect. Take what you've said is the new expected and dump it into a browser. You'll see there's only 1 space after `b` because the browser will collapse adjacent spaces so the original was correct here.

In this case we either have a bug in the `InputKeys` override logic or we don't want the override to run since we've already normalised it. I'm not sure which is the issue so you'll need to investigate that more. If it is the latter, then one way to workaround it would be to add a custom property to the fake events and update the override to ignore the event in that case. We should only do that if we're sure there's not a bug in the overrides though.

modules/tinymce/src/core/main/ts/text/CharType.ts
```diff
@@ -13,7 +13,7 @@ const isNbsp = is(Unicode.nbsp);
 
 const isWhiteSpace = (chr: string) => chr !== '' && ' \f\n\r\t\v'.indexOf(chr) !== -1;
 
-const isContent = (chr: string) => !isWhiteSpace(chr) && !isNbsp(chr);
+const isContent = (chr: string) => !isWhiteSpace(chr) && !isNbsp(chr) && !Unicode.isZwsp(chr);
```
HAFRMO, line 16: I'm not entirely sure this is the correct solution, but it is a solution, and the tests work.
Short of it, there's a ZWSP added to the node which was counted as 'content'. It's removed some time after. The ZWSP may be added by cursor placement.

modules/tinymce/src/core/main/ts/keyboard/EnterKey.ts
```diff
@@ -21,7 +21,11 @@ const handleEnterKeyEvent = (editor: Editor, event: EditorEvent<KeyboardEvent>)
   endTypingLevelIgnoreLocks(editor.undoManager);
   editor.undoManager.transact(() => {
     if (editor.selection.isCollapsed() === false) {
+      // Hacky way to avoid input events to be fired by execCommand idealy we would have our own delete implementation
```
spocke, line 24: Not sure about this change. Delete and range.deleteContents() is not doing the same thing for empty blocks so they are not 1:1 compatible and we don't have our own custom delete logic that works in all cases. So in order to prevent our override to also fire a input event for execCommand it put it into readonly mode that blocks the input events from firing by the event delegation but it feels a bit hacky. But the alternative would be to add some form of flag separate from readonly to only deal with this specific situation to cancel events. This was the best I could come up with.

modules/tinymce/src/core/main/ts/text/CharType.ts
spocke: Yes, that seems correct since zwnbsp are getting trimmed away when getting the content out we should never treat that as content.

modules/tinymce/src/core/main/ts/keyboard/SpaceKey.ts
```diff
@@ -8,14 +8,17 @@
...
 import * as MatchKeys from './MatchKeys';
 
 const executeKeydownOverride = (editor: Editor, evt: KeyboardEvent) => {
   MatchKeys.execute([
     { keyCode: VK.SPACEBAR, action: MatchKeys.action(InsertSpace.insertSpaceOrNbspAtSelection, editor) }
-  ], evt).each((_) => {
+  ], evt).each(() => {
     evt.preventDefault();
+    // Browsers sends space in data even if the dom ends up with a nbsp so we should always be sending a space
+    fireFakeInputEvent(editor, 'insertText', { data: ' ' });
```
metricjs, line 21: Just sanity checking - in DeleteBackspaceKeys we fire the fake event then preventDefault, and here we do the opposite order. That's intended/won't matter?

metricjs (approved):

modules/tinymce/src/core/main/ts/keyboard/SpaceKey.ts
spocke: It doesn't really matter but I tend to want the preventDefault early since if anything goes wrong like it throws an exception or something then the event would still be prevented. I will change that.

modules/tinymce/CHANGELOG.md
```diff
@@ -41,6 +41,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The `editor.selection.getContent()` API did not respect the `no_events` argument #TINY-8018
 - The `GetContent` event was not fired when getting `tree` or `text` formats using the `editor.selection.getContent()` API #TINY-8018
 - The `table` plugin would sometimes not correctly handle headers in the `tfoot` section #TINY-8104
+- Deleting content would sometimes not fire `beforeinput` and `input` events as expected #TINY-8168
```
lnewson, line 44:
```suggestion
- Deleting content would sometimes not fire `input` events as expected #TINY-8168
```
We moved the `beforeinput` logic to a follow up Jira

modules/tinymce/src/core/main/ts/delete/DeleteUtils.ts
```diff
@@ -9,11 +9,20 @@ import { Optional, Optionals } from '@ephox/katamari';
 import { Compare, PredicateFind, SugarElement } from '@ephox/sugar';
 
 import Editor from '../api/Editor';
+import { EditorEvent } from '../api/PublicApi';
```
lnewson, line 12:
```suggestion
import { EditorEvent } from '../api/util/EventDispatcher';
```
This would have just been something done by an IDE we should be avoiding importing from `PublicApi`. We probably need to setup an ESLint rule for that.

modules/tinymce/src/core/main/ts/keyboard/FakeInputEvents.ts
```diff
@@ -0,0 +1,42 @@
...
+    data: null,
+    isComposing: false,
+    detail: 0,
+    view: null,
+    target,
+    currentTarget: target,
+    eventPhase: Event.AT_TARGET,
+    originalTarget: target,
+    explicitOriginalTarget: target,
+    isTrusted: true,
```
lnewson, line 29: Is this correct? As this won't be a trusted event since it's mocked so if someone tries to do something with it that requires that it'll fail.

modules/tinymce/src/core/test/ts/browser/keyboard/SpaceKeyTest.ts
```diff
@@ -16,12 +16,21 @@ describe('browser.tinymce.core.keyboard.SpaceKeyTest', () => {
 
   context('Space key around inline boundary elements', () => {
     it('Press space at beginning of inline boundary inserting nbsp', () => {
+      const inputEvents: { inputType: string; data: string }[] = [];
```
lnewson, line 19:
```suggestion
      const inputEvents: Array<{ inputType: string; data: string }> = [];
```
Unfortunately there isn't a similar rule to our old TSLint rule here, but complex types should use `Array<>`.

modules/tinymce/src/core/test/ts/browser/keyboard/SpaceKeyTest.ts
```diff
@@ -1,5 +1,5 @@
 import { Keys } from '@ephox/agar';
-import { beforeEach, context, describe, it } from '@ephox/bedrock-client';
+import { assert, beforeEach, context, describe, it } from '@ephox/bedrock-client';
```
lnewson, line 2: This should be using `chai` instead of the legacy/deprecated Bedrock assert function (we should delete this next bedrock major). As a side note, we'll also need to use `deepEqual` below.

- [ ] Use chai assertions

modules/tinymce/src/core/test/ts/browser/keyboard/EnterKeyTest.ts
```diff
@@ -718,31 +727,43 @@ describe('browser.tinymce.core.keyboard.EnterKey', () => {
...
     editor.getBody().innerHTML = '<details><summary>ab</summary></details>';
     LegacyUnit.setSelection(editor, 'summary', 1);
-    pressEnter(editor);
+    pressEnter(editor, false);
     assert.equal(editor.getContent(), '<details><summary>a<br />b</summary></details>');
   });
+
+  it('Enter on expanded range', () => {
+    const editor = hook.editor();
+    const apis = TinyApis(editor);
```
lnewson, line 760: This is deprecated/part of the old pipelines API so we should avoid it.
```suggestion
```

modules/tinymce/src/core/test/ts/browser/keyboard/EnterKeyTest.ts
```diff
@@ -718,31 +727,43 @@ describe('browser.tinymce.core.keyboard.EnterKey', () => {
...
+
+    editor.setContent('<p>abc</p>');
+    apis.setSelection([ 0, 0 ], 1, [ 0, 0 ], 2);
```
lnewson, line 763:
```suggestion
    TinySelections.setSelection(editor, [ 0, 0 ], 1, [ 0, 0 ], 2);
```
This is how the newer BDD APIs are used.

modules/tinymce/src/core/test/ts/browser/keyboard/EnterKeyTest.ts
```diff
@@ -718,31 +727,43 @@ describe('browser.tinymce.core.keyboard.EnterKey', () => {
...
+
+    pressEnter(editor, true);
+    apis.assertContent( '<p>a</p><p>c</p>');
+    apis.assertSelection([ 1, 0 ], 0, [ 1, 0 ], 0);
```
lnewson, line 767:
```suggestion
    TinyAssertions.assertContent(editor,  '<p>a</p><p>c</p>');
    TinyAssertions.assertSelection(editor, [ 1, 0 ], 0, [ 1, 0 ], 0);
```

modules/tinymce/src/core/test/ts/browser/keyboard/SpaceKeyTest.ts
```diff
@@ -16,12 +16,21 @@ describe('browser.tinymce.core.keyboard.SpaceKeyTest', () => {
 
   context('Space key around inline boundary elements', () => {
     it('Press space at beginning of inline boundary inserting nbsp', () => {
+      const inputEvents: { inputType: string; data: string }[] = [];
       const editor = hook.editor();
+      const collect = ({ inputType, data }) => {
```
lnewson, line 21: Types?
```suggestion
      const collect = ({ inputType, data }: InputEvent) => {
```

lnewson (approved):

modules/tinymce/src/core/test/ts/browser/keyboard/EnterKeyTest.ts
```diff
@@ -1,5 +1,5 @@
 import { describe, it } from '@ephox/bedrock-client';
-import { LegacyUnit, TinyHooks } from '@ephox/wrap-mcagar';
+import { LegacyUnit, TinyHooks, TinyApis } from '@ephox/wrap-mcagar';
```
lnewson, line 2:
```suggestion
import { LegacyUnit, TinyAssertions, TinyHooks, TinySelections } from '@ephox/wrap-mcagar';
```
This is to match my other suggestions.

modules/tinymce/src/core/main/ts/keyboard/FakeInputEvents.ts
spocke: I guess it depends on how we view that property but it would make it possible to check for simulated events vs real events. But all events are somewhat simulated since they are cloned and dispatched though a event delegator. So then we should switch this to false for all events.

lnewson: Point... I guess leave it as is then :+1:

modules/tinymce/CHANGELOG.md
spocke: Right I thought that had been updated as well I should have double checked that.

modules/tinymce/src/core/test/ts/browser/keyboard/SpaceKeyTest.ts
spocke: Right, I knew about this one. Been out of the TS code so long that I forgot about that rule. :)

modules/tinymce/src/core/main/ts/delete/DeleteUtils.ts
spocke: I fixed this one and I also found one in page up/down that should also be the internal module.

modules/tinymce/src/core/test/ts/browser/keyboard/EnterKeyTest.ts
spocke: Nice, didn't know about these ones good to know for future PRs.

lnewson: For what it's worth or if it's helpful, https://github.com/tinymce/tinymce/blob/develop/modules/mcagar/docs/bdd.md has the details on the new BDD APIs :shrug:

lnewson:
Heads up we've merged this into a feature branch as we're concerned about this being released without the `beforeinput` changes as well. @metricjs is going to bring it up for further discussion at the next sync-up.

PR #7440, opened 2021-12-04
TINY-8253: Remove old IE and legacy Edge logic

lnewson:
Related Ticket: TINY-8253

Description of Changes:

This attempts to remove most of the old IE and Edge logic through out the code. It doesn't do all of the things that comments mention the change is for IE, instead I've just gone with the safer approach and only removed things that had specific plaform checks or that I was reasonably certain only applied to IE/Edge.

Note: I've also cleaned up the old `focus`/`blur` capturing in Firefox since it's supported `focusin` and `focusout` since Firefox 52 (released March 2017).

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/sugar/src/main/ts/ephox/sugar/api/node/SugarComments.ts
```diff
@@ -1,51 +1,25 @@
...
-
-  // the standard wants an object, IE wants a function. But everything is an object, so let's be sneaky
-  // http://www.bennadel.com/blog/2607-finding-html-comment-nodes-in-the-dom-using-treewalker.htm
-  vmlFilter.acceptNode = vmlFilter;
-
-  // Need to use a cast here, as tslib doesn't include the fourth argument anymore since it's deprecated
-  // however we still need it for IE unfortunately.
-  const texas = (document as VendorDocument).createTreeWalker(node.dom, NodeFilter.SHOW_COMMENT, vmlFilter, false);
-
+  const texas = document.createTreeWalker(node.dom, NodeFilter.SHOW_COMMENT, { acceptNode: vmlFilter });
```
lnewson, line 22: This is an actual functionality change here... apparently the browser will work with a boolean but the types were incorrect as the return type for `acceptNode` is a `number`. As such, I've reworked this to use the relevant `NodeFilter` flags instead based on the filter return value.

modules/alloy/src/main/ts/ephox/alloy/dragging/dragndrop/ImageClone.ts
```diff
@@ -40,81 +38,8 @@ const setDragImageFromClone = (transfer: DataTransfer, parent: SugarElement, ima
...
-};
-
-const setImageClone = (transfer: DataTransfer, image: DragnDropImageClone, parent: SugarElement, target: SugarElement): void => {
-  if (DataTransfers.hasDragImageSupport(transfer)) {
-    setDragImageFromClone(transfer, parent, image);
-  } else if (PlatformDetection.detect().browser.isEdge()) {
-    setDragImageFromCloneEdgeFallback(image, parent, target);
-  } else {
-    // We can't provide a fallback on IE 11 since the drag event doesn't update the mouse position
-    blockDefaultGhost(target);
```
lnewson, line 116: I wondered if I should keep this, but according to MDN and the relevant google searches all browsers now support drag images.

modules/oxide/gulpfile.js
```diff
@@ -13,7 +13,7 @@ const fs = require('fs');
 const path = require('path');
 
 const autoprefix = new lessAutoprefix({
-  browsers: ['IE 11', 'last 2 Safari versions', 'iOS 9.0', 'last 2 Chrome versions', 'Firefox ESR'],
+  browsers: ['last 2 Safari versions', 'iOS 13.0', 'last 2 Chrome versions', 'Firefox ESR'],
```
lnewson, line 16: Should this be `iOS 14.0` for 6.0? I just went with the current documented values but given we only support the last 2 majors we should be changing that for 6.0 IIUC.

modules/tinymce/src/themes/silver/main/ts/ui/icons/Icons.ts
```diff
@@ -74,7 +74,7 @@ const needsRtlTransform = (iconName: string) =>
 const addFocusableBehaviour = () =>
   AddEventsBehaviour.config('add-focusable', [
     AlloyEvents.runOnAttached((comp) => {
-      // set focusable=false on SVGs to prevent IE 11 from focusing the toolbar when tabbing into the editor
+      // set focusable=false on SVGs to prevent focusing the toolbar when tabbing into the editor
       SelectorFind.child(comp.element, 'svg').each((svg) => Attribute.set(svg, 'focusable', 'false'));
```
lnewson, line 78: Did we want to remove this as well? It shouldn't be needed on modern browsers but I left it here to be safe :shrug:

modules/oxide/gulpfile.js
TheSpyder: yes, 15 is current

modules/tinymce/src/themes/silver/main/ts/ui/dialog/IFrame.ts
```diff
@@ -8,62 +8,44 @@
...
-  const isSandbox = platformNeedsSandboxing && spec.sandboxed;
+  const isSandbox = spec.sandboxed;
 
   const attributes = {
     ...spec.label.map<{ title?: string }>((title) => ({ title })).getOr({}),
     ...isSandbox ? { sandbox: 'allow-scripts allow-same-origin' } : { }
   };
 
-  const sourcing = getDynamicSource(isSandbox);
+  const sourcing = getDynamicSource();
```
ashfordneil, line 48: Don't we still need to pass the sandbox in through here, just in case the spec asked for sandboxing?

ashfordneil (approved):

modules/tinymce/src/themes/silver/main/ts/ui/dialog/IFrame.ts
lnewson: No, if you look at the reason it was done it was for IE 11/Legacy Edge. It looks like someone originally used that because sandboxing was always disabled for those 2 browsers.

lnewson: I should also add we're fine using `srcdoc` regardless of if we're using a sandbox or not. It'll also avoid those verbose console warnings modern browsers show about `document.write` being deprecated. e.g:
```
[Violation] Avoid using document.write(). https://developers.google.com/web/updates/2016/08/removing-document-write
```

modules/oxide/gulpfile.js
lnewson: Cheers, have changed this to iOS 14 at minimum now then

modules/alloy/src/main/ts/ephox/alloy/dragging/dragndrop/ImageClone.ts
```diff
@@ -29,7 +27,7 @@ const createGhostClone = (image: DragnDropImageClone): SugarElement => {
 // Fire dragDrop on that image
 // remove the image in a setTimeout( 0 )
 // http://jsfiddle.net/stevendwood/akScu/21/
-const setDragImageFromClone = (transfer: DataTransfer, parent: SugarElement, image: DragnDropImageClone): void => {
+const setDragImageFromClone = (transfer: DataTransfer, parent: SugarElement<Node>, image: DragnDropImageClone): void => {
```
spocke, line 30: Shouldn't this be parent: SugarElement<Element> or does that break to much things?

modules/tinymce/src/themes/silver/main/ts/ui/dialog/IFrame.ts
spocke: I guess we should change the init of tinymce to not use document.write as well and use srcdoc.

spocke (approved):
That was a lot of old IE crap to go though. Good work couldn't spot anything that looked weird.

modules/alloy/src/main/ts/ephox/alloy/dragging/dragndrop/ImageClone.ts
spocke: I think it was IE and Edge that was problematic. So not something we need.

modules/alloy/src/main/ts/ephox/alloy/dragging/dragndrop/ImageClone.ts
lnewson: Not sure what you mean here sorry... it's already type as `SugarElement`?

modules/tinymce/src/themes/silver/main/ts/ui/dialog/IFrame.ts
lnewson: Yeah that's a good point... I actually did this like a year ago so I might have an old branch lying around somewhere 🤔 either way I'll log a Jira

modules/alloy/src/main/ts/ephox/alloy/dragging/dragndrop/ImageClone.ts
lnewson: Looking at where this is used we could make the `Node & ParentNode`, but it definitely can't be `Element` because the fallback for `getImageParent` is to get the parent node which isn't guaranteed to be an element.

spocke: What I'm saying it should be ParentNode or even Element since it's highly unlikely that you are drag dropping the documentElement or document fragment nodes since those are not attached to anything.

spocke: So the fallback is wrong it should be Traverse.parentElement it should all be elements.

modules/tinymce/src/themes/silver/main/ts/ui/dialog/IFrame.ts
lnewson: I ended up updating the init sequence to use `srcdoc` as part of this instead since it's also related to IE 11 cleanup/modernisation.

modules/alloy/src/main/ts/ephox/alloy/dragging/dragndrop/ImageClone.ts
lnewson: Have changed the function and types.

PR #7447, opened 2021-12-06
Moved from settings.json to a VSCode workspace file

TheSpyder:
Related Ticket: None

Description of Changes:
* This is a development proposal to get team feedback.

I went with separate folders for code and tests, and grouped them together - we can easily reorder the folders to keep tests next to the code, or to just have one folder per "module" instead of two as I've done here.

I also snuck in a 'fix lint errors on save' command, it's so helpful I thought it was worth trying to enforce.

Pre-checks:
* [ ] ~Changelog entry added~
* [ ] ~Tests have been added (if applicable)~
* [ ] ~Branch prefixed with `feature/` for new features (if applicable)~

Review:
* [ ] Milestone set


spocke:
You could also include a task to run bedrock on a specific folder specific file is no longer required since you have mocha only etc for that. We have this in rockfest but our tests are pretty fast to run so less of an issue there but for tinymce it might be useful to be able to just run the tests for say the theme or some subdir of the core etc without typing paths all day in the shell.

```
{
        "label": "bedrock-manual current dir",
        "command": "yarn",
        "type": "process",
        "args": [
          "bedrock",
          "-d",
          "${fileDirname}"
        ],
        "isBackground": true,
        "options": {
          "cwd": "${workspaceFolder:entire project}"
        },
        "presentation": {
          "echo": true,
          "reveal": "always",
          "focus": false,
          "panel": "shared",
          "showReuseMessage": false,
          "clear": true
        },
        "problemMatcher": []
      }
```

TheSpyder:
I still think it has value, but no. I don't even use vscode anymore.

PR #7438, opened 2021-12-03
TINY-8293: Replace contact wrapper libraries

lnewson:
Related Ticket: TINY-8293

Description of Changes:
* Replace the sizzle wrapper with sizzle directly
* Replace jsverify with fast-check
* Added types to Agar arbitraries

Pre-checks:
* [x] Changelog entry added
* [x] Tests have been added (if applicable)
* [x] Branch prefixed with `feature/` for new features (if applicable)

Review:
* [x] Milestone set
* [x] Review comments resolved

GitHub issues (if applicable):

PR #7448, opened 2021-12-06
TINY-8205: Fixes to add some forward compatibility with the new parser

ashfordneil:
Related Ticket: TINY-8205

Description of Changes:
* This one you could probably review commit by commit if you wanted to (and I'm thinking of rebasing it, instead of squashing, so it would be a good idea to look at the commits separately)
* Fixed a weird test thunking case we'd missed before
* Removed .shortEnded from AstNode
* Removed .fixed from AstNode
* Changed the way we handle inserting invalid content — previously we just parsed and serialised the whole tree it as a way of tidying up the structure, and now we perform tree operations manually (and refactored to take some of that logic out of DomParser)

Pre-checks:
* [x] Changelog entry added
* [x] ~Tests have been added (if applicable)~ nothing new added, just gotta make sure the old tests still pass

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -52,6 +52,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Removed the `toc` plugin, which is now classified as a Premium plugin #TINY-8250
 - Removed the `tinymce.utils.Promise` API #TINY-8241
 - Removed the `tabfocus` plugin #TINY-8315
+- Removed the `shortEnded` property on `tinymce.html.Node` class #TINY-8205
 
 ## 5.10.2 - 2021-11-17
 
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -52,7 +52,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Removed the `toc` plugin, which is now classified as a Premium plugin #TINY-8250
 - Removed the `tinymce.utils.Promise` API #TINY-8241
 - Removed the `tabfocus` plugin #TINY-8315
-- Removed the `shortEnded` property on `tinymce.html.Node` class #TINY-8205
+- Removed the `shortEnded` and `fixed` properties on `tinymce.html.Node` class #TINY-8205
 
 ## 5.10.2 - 2021-11-17
 
--- a/modules/tinymce/src/core/main/ts/api/html/DomParser.ts
+++ b/modules/tinymce/src/core/main/ts/api/html/DomParser.ts
@@ -5,11 +5,12 @@
  * For commercial licenses see https://www.tiny.cloud/
  */
 
-import { Obj } from '@ephox/katamari';
+import { Arr, Obj } from '@ephox/katamari';
 
+import { cleanInvalidNodes } from '../../html/InvalidNodes';
 import * as LegacyFilter from '../../html/LegacyFilter';
 import * as ParserFilters from '../../html/ParserFilters';
-import { hasOnlyChild, isEmpty, isLineBreakNode, isPaddedWithNbsp, paddEmptyNode } from '../../html/ParserUtils';
+import { isEmpty, isLineBreakNode, isPaddedWithNbsp, paddEmptyNode } from '../../html/ParserUtils';
 import { BlobCache } from '../file/BlobCache';
 import Tools from '../util/Tools';
 import AstNode from './Node';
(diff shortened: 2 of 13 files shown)


modules/agar/src/main/ts/ephox/agar/api/Generators.ts
```diff
@@ -8,18 +8,18 @@ import * as TagDecorator from '../arbitrary/TagDecorator';
 
 const selection = (container, exclusions) => GenSelection.selection(container, exclusions);
 
-const describeSelection = (root, generated) =>
-  Hierarchy.path(root, generated.start()).bind((startPath) =>
-    Hierarchy.path(root, generated.finish()).map((finishPath) => ({
+const describeSelection = (root: SugarElement<HTMLElement>, generated: SimRange) =>
```
lnewson, line 11: These changes don't look related to anything here? This is also been fixed already and with other larger changes in https://github.com/tinymce/tinymce/pull/7438 so I'd suggest reverting this.

modules/tinymce/CHANGELOG.md
```diff
@@ -52,6 +52,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Removed the `toc` plugin, which is now classified as a Premium plugin #TINY-8250
 - Removed the `tinymce.utils.Promise` API #TINY-8241
 - Removed the `tabfocus` plugin #TINY-8315
+- Removed the `shortEnded` field on `tinymce.html.Node` class #TINY-8205
+- Removed the `fixed` field on `tinymce.html.Node` class #TINY-8205
```
lnewson, line 56:
```suggestion
- Removed the `shortEnded` and `fixed` field on `tinymce.html.Node` class #TINY-8205
```
This doesn't need 2 separate entries

modules/tinymce/src/plugins/paste/main/ts/core/Utils.ts
```diff
@@ -79,7 +79,7 @@ const innerText = (html: string): string => {
     }
 
     // Walk all children
-    if (!node.shortEnded) {
+    if (!(node.name in schema.getShortEndedElements())) {
```
lnewson, line 82:
```suggestion
    if (!node.name in schema.getShortEndedElements()) {
```
The parentheses aren't needed here

lnewson:
Glanced over things quickly and it looks okay, but will need to review the larger/critical changes when I'm more awake.

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
```diff
@@ -695,10 +568,12 @@ const DomParser = (settings?: DomParserSettings, schema = Schema()): DomParser =
 
     // Fix invalid children or report invalid children in a contextual parsing
     if (validate && state.invalidChildren.length) {
-      if (!args.context) {
-        fixInvalidChildren(state);
+      if (args.context) {
```
lnewson, line 571: You mention `nothing new added, just gotta make sure the old tests still pass` but this is definitely something new as previously it never tried to fix an invalid context so we'll need some tests for this change.

modules/tinymce/src/core/main/ts/api/html/Serializer.ts
```diff
@@ -89,7 +89,7 @@ const HtmlSerializer = (settings?: HtmlSerializerSettings, schema = Schema()): H
 
       if (!handler) {
         const name = node.name;
-        const isEmpty = node.shortEnded;
+        const isEmpty = name in schema.getShortEndedElements();
```
metricjs, line 92: I had completely forgotten you could use `in` like this. nice.

metricjs (approved):

modules/tinymce/src/core/main/ts/api/html/Serializer.ts
lnewson: It's normally a bad idea though and we should be using `Obj.has` generally though, but as this is parser related it's fine here.

modules/tinymce/src/core/main/ts/html/ExtractInvalidNode.ts
```diff
@@ -0,0 +1,152 @@
...
+    for (const childNode of children) {
+      if (!schema.isValidChild(originalNodeParent.name, childNode.name)) {
+        removeOrUnwrapInvalidNode(childNode, schema, originalNodeParent);
+      }
+    }
+    node.unwrap();
+  }
+};
+
+const extractInvalidNode = (node: AstNode, schema: Schema, fixed: Set<AstNode> = new Set(), onCreate: (newNode: AstNode) => void = Fun.noop) => {
```
lnewson, line 31: Return types here please
```suggestion
const extractInvalidNode = (node: AstNode, schema: Schema, fixed: Set<AstNode> = new Set(), onCreate: (newNode: AstNode) => void = Fun.noop): void => {
```

modules/tinymce/src/core/main/ts/html/ExtractInvalidNode.ts
```diff
@@ -0,0 +1,152 @@
...
+
+  // If the invalid element is a text block, and the text block is within a parent LI element
+  // Then unwrap the first text block and convert other sibling text blocks to LI elements similar to Word/Open Office
+  if (textBlockElements[node.name] && node.parent.name === 'li') {
+    // Move sibling text blocks after LI element
+    let sibling = node.next;
+    while (sibling) {
+      if (textBlockElements[sibling.name]) {
+        sibling.name = 'li';
+        fixed.add(sibling);
```
lnewson, line 52: This is mutating the `fixed` data which isn't ideal... can we get rid of the mutation here and return the result instead? That said we're mutating `node` like crazy here so it's probably not an issue...

modules/tinymce/src/core/main/ts/html/ExtractInvalidNode.ts
```diff
@@ -0,0 +1,152 @@
...
+  }
+};
+
+const extractInvalidNode = (node: AstNode, schema: Schema, fixed: Set<AstNode> = new Set(), onCreate: (newNode: AstNode) => void = Fun.noop) => {
+  const textBlockElements = schema.getTextBlockElements();
+  const nonEmptyElements = schema.getNonEmptyElements();
+  const whitespaceElements = schema.getWhiteSpaceElements();
+  const nonSplitableElements = Tools.makeMap('tr,td,th,tbody,thead,tfoot,table');
+
+  let parent: AstNode | undefined, newParent: AstNode | undefined, tempNode: AstNode | undefined;
```
lnewson, line 37:
```suggestion
  let parent: AstNode | undefined;
  let newParent: AstNode | undefined;
  let tempNode: AstNode | undefined;
```
I know this was like this previously, but since we're moving it seemed like a good time to fix this :shrug:

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
```diff
@@ -406,137 +406,10 @@ const DomParser = (settings?: DomParserSettings, schema = Schema()): DomParser =
...
-
-        // Try wrapping the element in a DIV
-        if (schema.isValidChild(node.parent.name, 'div') && schema.isValidChild('div', node.name)) {
-          node.wrap(filterNode(new AstNode('div', 1), nodeFilters, attributeFilters, state));
-        } else {
-          // We failed wrapping it, remove or unwrap it
-          removeOrUnwrapInvalidNode(node);
-        }
-      }
+      extractInvalidNode(nodes[ni], schema, fixedNodes, (newNode) => filterNode(newNode, nodeFilters, attributeFilters, state));
```
lnewson, line 412: This adds potentially a lot of context/stack switches which could be a performance issue here... can we change this to pass in an array instead so we avoid some of this?

modules/agar/src/main/ts/ephox/agar/api/Generators.ts
ashfordneil: It was causing issues when I changed the parser over. Seeing as it's already in another branch though, I've reverted it.

modules/tinymce/src/plugins/paste/main/ts/core/Utils.ts
ashfordneil: Screen Shot 2021-12-08 at 3 42 38 pm
yes they are, otherwise it just evaluates as something like this:

```typescript
if (false in schema.getShortEndedElements()) {
```

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
ashfordneil: New test added to assert that this works, let me know if you want more tests just to be extra safe

modules/tinymce/src/core/main/ts/html/ExtractInvalidNode.ts
ashfordneil: `fixed` is now entirely internal to the function, so there's _less_ external mutation (nodes still get altered quite a bit, but that's kind of the point)

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
ashfordneil: Done — also helped me simplify the arguments to the function a little bit, thankfully

modules/tinymce/src/plugins/paste/main/ts/core/Utils.ts
lnewson: Right you are sorry... must have been getting confused with something else sorry.

modules/tinymce/src/core/main/ts/html/ExtractInvalidNodes.ts
```diff
@@ -0,0 +1,159 @@
...
+    for (const childNode of children) {
+      if (!schema.isValidChild(originalNodeParent.name, childNode.name)) {
+        removeOrUnwrapInvalidNode(childNode, schema, originalNodeParent);
+      }
+    }
+    node.unwrap();
+  }
+};
+
+const extractInvalidNodes = (nodes: AstNode[], schema: Schema, onCreate: (newNode: AstNode) => void = Fun.noop): void => {
```
lnewson, line 31: Should this be called `extract` as that keeps confusing me here since it doesn't return anything? Maybe call it `cleanInvalidNodes` instead?

lnewson: Also maybe just call the module `InvalidNodes` as that's what it's purpose is even if it only exposes the one function atm.

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
lnewson: I'm okay with that, as the rest should largely be tested elsewhere.

lnewson (approved):

PR #7455, opened 2021-12-08
TINY-8326: Removed the old/undocumeted Class API

lnewson:
Related Ticket: TINY-8326

Description of Changes:

Removed the old `Class` API as it's not been used since TinyMCE 4 and doesn't make much sense given we've removed the `Tools.create`/`Tools.createNS` APIs and the `new` operator logic for plugins/themes.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/tinymce/src/core/main/ts/api/Tinymce.ts
```diff
@@ -43,7 +43,6 @@ import Resource from './Resource';
 import Shortcuts, { ShortcutsConstructor } from './Shortcuts';
 import ThemeManager from './ThemeManager';
 import UndoManager from './UndoManager';
-import Class from './util/Class';
```
ltrouton, line 46: This might have come up before in another PR but should we remove the reference in `globals.json` file as well? Looking at that file, some of the other removed APIs are still in the list so maybe that can be done separately.

ltrouton (approved):

ashfordneil (approved):

PR #7471, opened 2021-12-16
TINY-8353: Add FakeClipboard to tinymce global

ltrouton:
Related Ticket: TINY-8353

Description of Changes:
* Add a fake clipboard to the tinymce global that can be used by the table copy row/column logic as well by the paste plugin

Note: I didn't have an exact spec to work with so have just done a basic read, write, clear API as a starting point.

Pre-checks:
* [X] Changelog entry added

Review:
* [X] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/api/FakeClipboard.ts
+++ b/modules/tinymce/src/core/main/ts/api/FakeClipboard.ts
@@ -1,0 +1,63 @@
+
+/**
+ * Copyright (c) Tiny Technologies, Inc. All rights reserved.
+ * Licensed under the LGPL or a commercial license.
+ * For LGPL see License.txt in the project root for license information.
+ * For commercial licenses see https://www.tiny.cloud/
+ */
+
+import { Singleton } from '@ephox/katamari';
+
+/**
+  * TinyMCE FakeClipboard API.
+  *
+  * @class tinymce.FakeClipboard
+  */
+
+interface FakeClipboard {
+
+  /**
+   * Writes arbitrary data to the fake clipboard.
+   *
+   * @method write
+   * @param {any} data data to be written to the fake clipboard
+   */
+  readonly write: <T>(data: T) => void;
+
+  /**
+   * Requests arbitrary data from the fake clipboard.
+   *
+   * @method read
+   */
+  readonly read: <T>() => T | undefined;
+
+  /**
+   * Clear arbitrary data on the fake clipboard.
+   *
+   * @method clear
+   */
+  readonly clear: () => void;
+}
+
+const setup = (): FakeClipboard => {
+  const dataValue = Singleton.value<any>();
+
+  const write = <T>(data: T): void => {
+    dataValue.set(data);
+  };
+
+  const read = <T>(): T | undefined =>
+    dataValue.get().getOrUndefined();
+
+  const clear = dataValue.clear;
+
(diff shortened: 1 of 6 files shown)


modules/tinymce/src/core/main/ts/api/FakeClipboard.ts
```diff
@@ -0,0 +1,89 @@
...
+  * @class tinymce.FakeClipboard
+  */
+
+interface FakeClipboard {
+
+  /**
+   * Create a FakeClipboardItem instance that is used when reading or writing data via the FakeClipboard API.
+   *
+   * @method FakeClipboardItem
+   * @param {Record<string, any>} items an object with the type as the key and any data as the value
```
lnewson, line 32:
```suggestion
   * @param {Object} items An object with the type as the key and any data as the value.
```
This has to be native/primitive types unfortunately.

modules/tinymce/src/core/main/ts/api/FakeClipboard.ts
```diff
@@ -0,0 +1,89 @@
...
+   * @returns {FakeClipboardItem} FakeClipboardItem
```
lnewson, line 33:
```suggestion
   * @returns {tinymce.FakeClipboardItem} A new fake clipboard item to represent the specified items.
```
This has to be the fully qualified name of the class.

modules/tinymce/src/core/main/ts/api/FakeClipboard.ts
```diff
@@ -0,0 +1,89 @@
...
+   */
+  readonly FakeClipboardItem: (items: Record<string, any>) => FakeClipboardItem;
+
+  /**
+   * Writes arbitrary data to the fake clipboard.
+   *
+   * @method write
+   * @param {any} data data to be written to the fake clipboard
```
lnewson, line 41: Hmm... we can't use `any` here I believe (remember this isn't TS types) so this should probably just be `Object`.
```suggestion
   * @param {Object} data Arbitrary data to be written to the fake clipboard.
```

modules/tinymce/src/core/main/ts/api/FakeClipboard.ts
```diff
@@ -0,0 +1,89 @@
...
+   */
+  readonly write: (data: FakeClipboardItem[]) => void;
+
+  /**
+   * Requests arbitrary data from the fake clipboard.
+   *
+   * @method read
+   * @returns {FakeClipboardItem} FakeClipboardItem or undefined
```
lnewson, line 49:
```suggestion
   * @returns {Array} An array of FakeClipboardItems if items exist on the fake clipboard, otherwise undefined.
```
This return type and message is incorrect.

lnewson (approved):

TheSpyder (approved):

modules/tinymce/src/core/main/ts/api/FakeClipboard.ts
ltrouton: Thanks for pointing this out. I rewrote it differently to the suggestion. Let me know if it is seems reasonable.

lnewson: LGTM 👍

ashfordneil (approved):

PR #7478, opened 2021-12-20
TINY-8214: Swap from SaxParser to DomPurify in editor.parser

ashfordneil:
Related Ticket: TINY-8214

Description of Changes:
* Look at the commits for this, it will probably help
* Lots of the changes are just skipping failing tests, and moving them out to future Jira tasks
* I removed the `mceInsertRawHTML` command after a team discussion because it wasn't really any different to `mceInsertContent` in practice, and we aren't using it at all
* I removed the `elements` property from `Schema` as we weren't using it, and it often was invalid
* DomParser you're almost better looking at without the diff, and just reading the (top half of) the new version of the file on its own, as it's largely green code.

Pre-checks:
* [x] Changelog entry added
* [x] ~Tests have been added (if applicable)~ Almost all existing tests pass, and nothing new was added (somehow)

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -84,6 +84,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Removed the `tabfocus` plugin #TINY-8315
 - Removed the `textpattern` plugin's API as part of moving it to core #TINY-8312
 - Removed the `shortEnded` and `fixed` properties on `tinymce.html.Node` class #TINY-8205
+- Removed the `mceInsertRawHTML` command, in favor of `mceInsertContent` #TINY-8214
 
 ### Deprecated
 - The dialog button component `primary` property has been deprecated in favour of the new `buttonType` property #TINY-8304
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -85,6 +85,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Removed the `textpattern` plugin's API as part of moving it to core #TINY-8312
 - Removed the `shortEnded` and `fixed` properties on `tinymce.html.Node` class #TINY-8205
 - Removed the `mceInsertRawHTML` command, in favor of `mceInsertContent` #TINY-8214
+- Removed the `elements` property from `Schema` #TINY-8214
 
 ### Deprecated
 - The dialog button component `primary` property has been deprecated in favour of the new `buttonType` property #TINY-8304
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -84,7 +84,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Removed the `tabfocus` plugin #TINY-8315
 - Removed the `textpattern` plugin's API as part of moving it to core #TINY-8312
 - Removed the `shortEnded` and `fixed` properties on `tinymce.html.Node` class #TINY-8205
-- Removed the `mceInsertRawHTML` command, in favor of `mceInsertContent` #TINY-8214
+- Removed the `mceInsertRawHTML` command #TINY-8214
 - Removed the `elements` property from `Schema` #TINY-8214
 
 ### Deprecated
--- a/modules/tinymce/src/core/main/ts/api/html/DomParser.ts
+++ b/modules/tinymce/src/core/main/ts/api/html/DomParser.ts
@@ -6,7 +6,7 @@
  */
 
 import { Arr, Obj, Type } from '@ephox/katamari';
-import { Remove, SugarElement } from '@ephox/sugar';
+import { Attribute, Insert, InsertAll, Remove, SugarElement, SugarNode, Traverse } from '@ephox/sugar';
 import createDompurify, { Config, DOMPurifyI } from 'dompurify';
 
 import * as NodeType from '../../dom/NodeType';
@@ -17,7 +17,7 @@ import { isEmpty, isLineBreakNode, isPaddedWithNbsp, paddEmptyNode } from '../..
 import { BlobCache } from '../file/BlobCache';
 import Tools from '../util/Tools';
 import AstNode from './Node';
-import Schema, { SchemaElement } from './Schema';
+import Schema from './Schema';
 
 /**
  * This class parses HTML code into a DOM like structure of nodes it will remove redundant whitespace and make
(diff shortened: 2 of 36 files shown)


modules/tinymce/src/core/main/ts/selection/SetSelectionContent.ts
```diff
@@ -102,7 +102,7 @@ const cleanContent = (editor: Editor, args: SetSelectionContentArgs) => {
     const contextBlock = editor.dom.getParent(rng.commonAncestorContainer, editor.dom.isBlock);
     const contextArgs = contextBlock ? { context: contextBlock.nodeName.toLowerCase() } : { };
 
-    const node = editor.parser.parse(args.content, { isRootContent: true, forced_root_block: false, ...contextArgs, ...args });
```
ashfordneil, line 105: I believe this `isRootContent: true` was incorrect, as we're parsing content that will be inserted not into the root, but at a specific point of selection. See the `isAtEdgeOfBlock` function in `DomParser.ts` for where we use this setting.

modules/tinymce/src/core/test/ts/browser/FormatterApplyTest.ts
```diff
@@ -497,15 +497,15 @@ describe('browser.tinymce.core.FormatterApplyTest', () => {
         fontWeight: 'bold'
       }
     });
-    editor.getBody().innerHTML = '<p>a<span id="id" style="font-weight:bold">1234</span>b</p>';
+    editor.getBody().innerHTML = '<p>a<span id="test-id" style="font-weight:bold">1234</span>b</p>';
```
ashfordneil, line 500: Changed the IDs (here and elsewhere) to avoid tripping "prevent DOM Clobbering" checks

modules/tinymce/src/core/test/ts/browser/html/DomParserTest.ts
```diff
@@ -115,7 +121,7 @@ describe('browser.tinymce.core.html.DomParserTest', () => {
   it('Whitespace preserved in PRE', () => {
     parser = DomParser({}, schema);
     root = parser.parse('  \t\r\n  <PRE>  \t\r\n   test  \t\r\n   </PRE>   \t\r\n  ');
-    assert.equal(serializer.serialize(root), '<pre>  \t\r\n   test  \t\r\n   </pre>', 'Whitespace around and inside PRE');
+    assert.equal(serializer.serialize(root), '<pre>  \t\n   test  \t\n   </pre>', 'Whitespace around and inside PRE');
```
ashfordneil, line 124: Even in whitespace sensitive elements, it turns out `\r` isn't meant to be preserved, so our old parser was incorrect here.

modules/tinymce/tsconfig.json
```diff
@@ -7,6 +7,7 @@
     "rootDir": "src",
     "baseUrl": ".",
     "tsBuildInfoFile": "lib/tsconfig.tsbuildinfo",
+    "allowSyntheticDefaultImports": true,
```
ashfordneil, line 10: I needed this due to how the dompurify types had been written — let me know if this is going to cause issues, because there are a lot of build tools we use that this might interfere with.

modules/tinymce/CHANGELOG.md
```diff
@@ -67,6 +67,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Removed the `tabfocus` plugin #TINY-8315
 - Removed the `textpattern` plugin's API as part of moving it to core #TINY-8312
 - Removed the `shortEnded` and `fixed` properties on `tinymce.html.Node` class #TINY-8205
+- Removed the `mceInsertRawHTML` command, in favor of `mceInsertContent` #TINY-8214
```
lnewson, line 70: This isn't a 1:1 replacement so we shouldn't say "in favor of".

```suggestion
- Removed the `mceInsertRawHTML` command #TINY-8214
```

modules/tinymce/src/core/main/ts/selection/SetSelectionContent.ts
lnewson: This shouldn't have made any difference as the `forced_root_block: false` means `isRootContent` should be ignored. That likely means there's a bug somewhere else in your new implementation. I do agree `isRootContent: true` shouldn't have been here though.

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
```diff
@@ -90,253 +92,107 @@ interface WalkResult {
...
+  let uid = 0;
+  const config: Config = {
+    RETURN_DOM: true,
+    ALLOW_DATA_ATTR: true,
+    ALLOWED_URI_REGEXP: /^(?:(?:(?:f|ht)tps?|about|blob|file|mailto|tel|callto|cid|xmpp):|[^a-z]|[a-z+.\-]+(?:[^a-z+.\-:]|$))/i,
+  };
 
-      const textNodeNext = textNode.next;
+  if (settings.allow_html_data_urls) {
+    config.ADD_DATA_URI_TAGS = [ 'a', 'img' ];
```
lnewson, line 105: This doesn't look right... this open is supposed to allow html data URLs on any element, not just `a` and `img`.

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
```diff
@@ -90,253 +92,107 @@ interface WalkResult {
...
+      });
+    } else if (NodeType.isComment(nativeChild) || NodeType.isText(nativeChild) || NodeType.isCData(nativeChild) || NodeType.isPi(nativeChild)) {
+      child.value = nativeChild.data;
+      if (isSpecial) {
+        child.raw = true;
       }
-
-      node = node.parent;
+    } else {
+      // TODO: figure out if anything can arrive in this branch, before merging
```
lnewson, line 178: Do we have a Jira for this or does the comment mean before merging this PR?

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
```diff
@@ -362,6 +218,131 @@ const walker = (root: AstNode, settings: DomParserSettings, schema: Schema, node
...
+  const blockElements: Record<string, string> = extend(makeMap('script,style,head,html,body,title,meta,param'), schema.getBlockElements());
+  const allWhiteSpaceRegExp = /[ \t\r\n]+/g;
+  const startWhiteSpaceRegExp = /^[ \t\r\n]+/;
+  const endWhiteSpaceRegExp = /[ \t\r\n]+$/;
+
+  const hasWhitespaceParent = (node: AstNode) => {
+    if (Obj.has(whitespaceElements, node.name)) {
+      return true;
+    } else if (node.parent) {
+      return hasWhitespaceParent(node.parent);
```
lnewson, line 237: This is potentially a lot of recursion which isn't great for a performance critical bit of code... it would probably be better to just use a for or while loop instead to walk up the parent.

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
```diff
@@ -554,7 +535,10 @@ const DomParser = (settings?: DomParserSettings, schema = Schema()): DomParser =
     };
 
     const rootNode = new AstNode(args.context || settings.root_name, 11);
-    builder(rootNode, html, settings, schema, args);
+    // The settings object we pass to purify is completely ignored, because we called setConfig earlier, but it makes the type signatures work
+    const element = purify.sanitize(html, { RETURN_DOM: true });
+    transferChildren(rootNode, element);
+    simplifyDom(rootNode, schema, settings, args);
     const state = walker(rootNode, settings, schema, nodeFilters, attributeFilters);
```
lnewson, line 542: I'm not fond that this is walking the entire dom 3 times... from the looks of things this could be done in 2 (or maybe less) walks instead so we'd be better doing that unless I'm missing a reason as to why we can't?

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
ashfordneil: - [ ] It meant before merging the PR. Need to figure this out

modules/tinymce/src/core/test/ts/browser/html/DomParserTest.ts
```diff
@@ -37,13 +37,17 @@ describe('browser.tinymce.core.html.DomParserTest', () => {
     assert.equal(serializer.serialize(root), '<b class="class" title="title">test</b>', 'Inline element');
     assert.equal(root.firstChild.type, 1, 'Element type');
     assert.equal(root.firstChild.name, 'b', 'Element name');
-    assert.deepEqual(
-      root.firstChild.attributes, [{ name: 'title', value: 'title' },
-        { name: 'class', value: 'class' }],
-      'Element attributes'
-    );
+    // TODO: TINY-4627/TINY-8202: dompurify shuffles attributes around for some reason
```
ashfordneil, line 40: Just change the assertion, we don't need to preserve attribute orders

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
```diff
@@ -90,276 +92,256 @@ interface WalkResult {
...
-        }
-      }
+      const children = Traverse.children(element);
+      InsertAll.append(replacement, children);
 
-      const sibling = textNode.prev;
-      textNode.remove();
-      textNode = sibling;
+      Insert.after(element, replacement);
+      Remove.remove(element);
```
lnewson, line 167: I mentioned this via DM, but copying here as well... I believe this can be changed to this if you change mutate to insert after as well:
```
      const replacement = Replication.mutate(SugarElement.fromDom(element), rule.outputName);
```

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
```diff
@@ -90,276 +92,263 @@ interface WalkResult {
...
+    if (!parent) {
+      continue;
+    }
 
-      // Change current node if the element wasn't empty i.e not <br /> or <img />
-      if (!empty) {
-        node = newNode;
-      }
+    // Check if node is valid child of the parent node is the child is
+    // unknown we don't collect it since it's probably a custom element
```
ltrouton, line 233: I assume this is more correct?
```suggestion
    // Check if the node is a valid child of the parent node. If the child is
    // unknown we don't collect it since it's probably a custom element
```

ltrouton (approved):

lnewson (approved):

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
lnewson: This will be fixed via TINY-8202

PR #7480, opened 2021-12-20
TINY-7631: Updated the emoticon plugin UI to use "Emojis" instead of "Emoticons"

lnewson:
Related Ticket: TINY-7631

Description of Changes:
* Updated the UI text and internal code to use the correct `Emojis` term instead of `Emoticons`

Note: As per the Jira the plugin name and options are not changing.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/plugins/emoticons/main/ts/ui/Dialog.ts
+++ b/modules/tinymce/src/plugins/emoticons/main/ts/ui/Dialog.ts
@@ -68,7 +68,7 @@ const open = (editor: Editor, database: EmojiDatabase): void => {
       }))
     };
     return {
-      title: 'Emoticons',
+      title: 'Emojis',
       size: 'normal',
       body,
       initialData: initialState,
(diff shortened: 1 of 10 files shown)


modules/tinymce/src/plugins/emoticons/main/ts/ui/Dialog.ts
```diff
@@ -98,23 +98,23 @@ const open = (editor: Editor, database: EmojiDatabase): void => {
...
+        title: 'Emojis',
         body: {
           type: 'panel',
           items: [
             {
               type: 'alertbanner',
               level: 'error',
               icon: 'warning',
-              text: '<p>Could not load emoticons</p>'
+              text: 'Could not load emojis'
```
lnewson, line 117: This doesn't appear to ever actually be used? Anyways I removed the `<p>` as per the name this is supposed to be `text` not `html`.

lnewson: For reference this is what it looks like without the `<p>`, so I don't think it should have ever been wrapped anyways:
image

ashfordneil (approved):

ltrouton (approved):

TheSpyder (approved):

PR #7485, opened 2021-12-21
TINY-8195: Change the annotation api to avoid dom mutation in plugins using it

spocke:
Related Ticket:

Description of Changes:
* This does some changes to the annotation part of the core to avoid direct dom mutation by the remark plugin. Check the JIRA ticket for details.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/annotate/AnnotationChanges.ts
+++ b/modules/tinymce/src/core/main/ts/annotate/AnnotationChanges.ts
@@ -6,10 +6,12 @@
  */
 
 import { Arr, Cell, Obj, Optional, Optionals, Singleton, Throttler } from '@ephox/katamari';
+import { Attribute } from '@ephox/sugar';
 
 import Editor from '../api/Editor';
 import { AnnotationsRegistry } from './AnnotationsRegistry';
-import { identify } from './Identification';
+import { findMarkers, identify } from './Identification';
+import { dataAnnotationActive } from './Markings';
 
 export interface AnnotationChanges {
   readonly addListener: (name: string, f: AnnotationListener) => void;
@@ -24,7 +26,7 @@ export interface AnnotationListenerData {
 
 export type AnnotationListenerMap = Record<string, AnnotationListenerData>;
 
-const setup = (editor: Editor, _registry: AnnotationsRegistry): AnnotationChanges => {
+const setup = (editor: Editor, registry: AnnotationsRegistry): AnnotationChanges => {
   const changeCallbacks = Cell<AnnotationListenerMap>({ });
 
   const initData = (): AnnotationListenerData => ({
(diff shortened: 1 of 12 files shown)


modules/tinymce/src/core/main/ts/annotate/AnnotationChanges.ts
```diff
@@ -6,10 +6,12 @@
  */
 
 import { Arr, Cell, Obj, Optional, Optionals, Singleton, Throttler } from '@ephox/katamari';
+import { Attribute } from '@ephox/sugar';
 
 import Editor from '../api/Editor';
 import { AnnotationsRegistry } from './AnnotationsRegistry';
-import { identify } from './Identification';
+import { findMarkers, identify } from './Identification';
+import { dataAnnotationActive } from './Markings';
```
lnewson, line 14:
```suggestion
import * as Markings from './Markings';
```
We generally prefer the `* as Name` format as it helps make things more readable. Up to you if we do this here. We also probably should do it for `Identification` if we do I guess, but I can't leave a suggestion for that sorry.

modules/tinymce/src/core/main/ts/annotate/AnnotationChanges.ts
```diff
@@ -62,26 +64,38 @@ const setup = (editor: Editor, _registry: AnnotationsRegistry): AnnotationChange
     });
   };
 
+  const toggleActiveAttr = (uid: string, state: boolean) => {
+    Arr.each(findMarkers(editor, uid), (span) => {
+      if (state) {
+        Attribute.set(span, dataAnnotationActive(), 'true');
+      } else {
+        Attribute.remove(span, dataAnnotationActive());
```
lnewson, line 72:
```suggestion
        Attribute.set(span, Markings.dataAnnotationActive(), 'true');
      } else {
        Attribute.remove(span, Markings.dataAnnotationActive());
```

modules/tinymce/src/core/main/ts/annotate/AnnotationFilter.ts
```diff
@@ -15,14 +15,13 @@ import * as Markings from './Markings';
 const setup = (editor: Editor, registry: AnnotationsRegistry): void => {
   const identifyParserNode = (span: AstNode): Optional<AnnotatorSettings> => Optional.from(span.attr(Markings.dataAnnotation())).bind(registry.lookup);
 
-  editor.on('init', () => {
```
lnewson, line 18: For others reading this, I'm assuming this was removed since by the time the annotator is setup the serializer is already initialised so this wasn't needed.

lnewson (approved):

ashfordneil (approved):

modules/tinymce/src/core/main/ts/annotate/AnnotationChanges.ts
spocke: It also makes it way more verbose. But I guess it makes it easier to see if the thing is local or not.

modules/tinymce/src/core/main/ts/annotate/AnnotationFilter.ts
spocke: Yes, and also it should not be added that late in the process it should at least be preinit. Init is more when the editor content is done in theory some one could call editor.getContent at that stage to get the initial snapshot out but I guess the core inits will fire before plugin or custom inits but anyway.

modules/tinymce/src/core/main/ts/annotate/AnnotationChanges.ts
spocke: I guess it make sense some times but not always since if there was a bunch of types that needed to be imported you don't want to have Markings.MyType etc all over. I guess for alloy we have that but that is because it's so many types and that is the only way you can export things in submodules in js though a single entry point.

TheSpyder (approved):
I was confused about the PR title - this appears to _add_ dom mutation, not avoid it - but looking at JIRA the aim is to remove dom mutation from the plugin. That makes more sense.

I'm not sure what to suggest instead but I think the title should change (since it will be the commit message).

spocke:
> I was confused about the PR title - this appears to _add_ dom mutation, not avoid it - but looking at JIRA the aim is to remove dom mutation from the plugin. That makes more sense.
> 
> I'm not sure what to suggest instead but I think the title should change (since it will be the commit message).

Changed the title now.

PR #7488, opened 2021-12-23
TINY-8163: test for paste not changing RGB color

thedrjim:
Related Ticket: TINY-8163

Description of Changes:
* fix paste regression due to incorrect variable being used.

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/plugins/paste/main/ts/core/Quirks.ts
+++ b/modules/tinymce/src/plugins/paste/main/ts/core/Quirks.ts
@@ -73,7 +73,8 @@ const removeWebKitStyles = (editor: Editor, content: string, internal: boolean):
       }
 
       for (let i = 0; i < webKitStyles.length; i++) {
-        let compareInput = inputStyles[webKitStyles[i]];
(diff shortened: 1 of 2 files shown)


lnewson (approved):

modules/tinymce/src/plugins/paste/main/ts/core/Quirks.ts
```diff
@@ -83,7 +82,7 @@ const removeWebKitStyles = (editor: Editor, content: string, internal: boolean):
         }
 
         if (currentValue !== compareInput) {
-          outputStyles[webKitStyles[i]] = inputValue;
+          outputStyles[webKitStyles[i]] = compareInput;
```
TheSpyder, line 85: This was a deliberate change to avoid changing styles to hex; the aim is to just check whether the colours match, not force them to be hex if they do.

It's the same thing we did in the DOM code where we leave rgb alone now instead of converting it to hex.

TheSpyder (approved):

PR #7504, opened 2022-01-10
TINY-8204: Ensure elements are parsed with the appropriate context

lnewson:
Related Ticket: TINY-8204

Description of Changes:
* Ensure that special elements are parsed with the appropriate context

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/api/html/DomParser.ts
+++ b/modules/tinymce/src/core/main/ts/api/html/DomParser.ts
@@ -94,8 +94,8 @@ interface FilterMatches {
 type WalkerCallback = (node: AstNode) => void;
 
 // See https://html.spec.whatwg.org/multipage/parsing.html#parsing-html-fragments
-// for anything special excluding elements that use the RCDATA state
-const specialElements = new Set('style,xmp,iframe,noembed,noframes,script,noscript,plaintext'.split(','));
+const nonRcDataSpecialElements = new Set('style,xmp,iframe,noembed,noframes,script,noscript,plaintext'.split(','));
+const specialElements = new Set([ 'title', 'textarea', ...nonRcDataSpecialElements ]);
 
 const getPurifyConfig = (settings: DomParserSettings): Config => {
   const config: Config = {
@@ -196,7 +196,8 @@ const setupPurify = (settings: DomParserSettings, schema: Schema): DOMPurifyI =>
 };
 
 const transferChildren = (parent: AstNode, nativeParent: Node) => {
-  const isSpecial = specialElements.has(parent.name);
+  // Exclude the special elements where the content is RCDATA as their content needs to be parsed instead of being left as plain text
+  const isSpecial = nonRcDataSpecialElements.has(parent.name);
   Arr.each(nativeParent.childNodes, (nativeChild) => {
     const child = new AstNode(nativeChild.nodeName.toLowerCase(), nativeChild.nodeType);
 
@@ -594,9 +595,16 @@ const DomParser = (settings?: DomParserSettings, schema = Schema()): DomParser =
     const validate = settings.validate;
     const rootBlockName = getRootBlockName(settings, args);
 
-    const rootNode = new AstNode(args.context || settings.root_name, 11);
+    // Determine the root element to wrap the HTML in when parsing. If we're dealing with a
+    // special element then we need to wrap it so the internal content is handled appropriately.
+    const rootName = args.context || settings.root_name;
+    // If the root requires special parsing rules then ensure it's wrapped it in that element
+    const isSpecialRoot = specialElements.has(rootName.toLowerCase());
+    const content = isSpecialRoot ? `<${rootName}>${html}</${rootName}>` : html;
     // The settings object we pass to purify is completely ignored, because we called setConfig earlier, but it makes the type signatures work
-    const element = purify.sanitize(html, { RETURN_DOM: true });
+    const body = purify.sanitize(`<body>${content}</body>`, { RETURN_DOM: true });
+    const element = isSpecialRoot ? body.firstChild : body;
+    const rootNode = new AstNode(rootName, 11);
     transferChildren(rootNode, element);
 
     // Set up whitespace fixes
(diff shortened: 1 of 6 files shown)


HAFRMO (approved):

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
```diff
@@ -93,12 +93,12 @@ interface WalkResult {
 }
 
 // See https://html.spec.whatwg.org/multipage/parsing.html#parsing-html-fragments
-// for anything special excluding elements that use the RCDATA state
-const specialElements = new Set('style,xmp,iframe,noembed,noframes,script,noscript,plaintext'.split(','));
+const specialElements = new Set('title,textarea,style,xmp,iframe,noembed,noframes,script,noscript,plaintext'.split(','));
```
spocke, line 96: Should this be moved to a central schema at some point? I really want to avoid us having hard coded lists of elements all over the place if we can.

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
```diff
@@ -195,7 +195,8 @@ const setupPurify = (settings: DomParserSettings, schema: Schema): DOMPurifyI =>
 };
 
 const transferChildren = (parent: AstNode, nativeParent: Node) => {
-  const isSpecial = specialElements.has(parent.name);
+  // Exclude textarea and title as their content needs to be parsed as RCDATA
+  const isSpecial = specialElements.has(parent.name) && parent.name !== 'textarea' && parent.name !== 'title';
```
spocke, line 199: If this is a special subset of elements that need special treatment should they be their own set instead like. `specialRcDataElements` or something. Are textarea and title the only ones. Could we put a spec link in the comment?

spocke (approved):

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
lnewson: No, as it's not related to the schema in any form... it's related to the HTML parsing spec and how browsers parse, so it's only applicable in this module.

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
lnewson: Have created a variable for it as requested. The spec link was already next to the variable

ltrouton (approved):

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
spocke: I guess we need to discuss how we are to handle the html schema in general in the future. Since in the new alt model it's a dynamic instance being passed around and that is used for editing behavior rules and model normalization:
https://github.com/tinymce/rockfest/blob/master/modules/bs-slate/src/main/re/schema/Slate__Core__Schema.resi

For this case it's probably fine to have a hard coded list of things since they are parser specific and likely not going to change that often. But I want to avoid having lists of elements all over since browsers add things to the spec all the time. However I am guilty of adding some hard coded lists my self mostly used by the content editable support:
https://github.com/tinymce/tinymce/blob/develop/modules/tinymce/src/core/main/ts/dom/ElementType.ts
However since we are switching to a model that has a dynamic schema again and the contenteditable support is going to be rewritten from scratch then that thing will not be used anymore.

The purpose of the Schema when it was designed was to have one spot for all parser rules so it used to be the central place of the HTML spec rules configurable and static ones but also for editing behavior one rule set for everything. So to me it is very much a part of the HTML parser that is why it is passed into the DomParser. This special element could have been a predicate that takes a string on that Schema instance. However not sure this needs to be public api it's unlikely that some plugin needs this information. So I guess it makes sense to hard code the list here then.

However it might be useful to have this for the normalization of the model not just for parsing. Since these elements should never be allowed to have a child elements they should all be converted into string or dropped if they are appended and since the valid children is a dynamic thing so someone could configure a iframe to allow child elements. Having a fixed set there to block that from happening might be useful. But I getting ahead of my self.

lnewson: You're getting confused here sorry Spocke, as we still have the schema and it's used for validation as it previously was. This however relates to how the native parser parses and as such **cannot** be something that can be configured by a custom schema.

lnewson: To clarify further `Slate__Core__Schema.resi` is not related to this piece of code and it instead relates to `Schema.ts`. The custom schema is then used by both DOMPurify to remove elements that aren't allowed and then further once it's been transformed into an AST to clean up invalid children as defined by the custom schema.

spocke: No I'm not confused at all. But I may not be good at explaining what I mean. Parts of the schema is dynamic parts of it is static now it's mostly dynamic it doesn't have to be. I say all HTML rules dynamic or static should be in one central place or perhaps two central places not N places. Since the rules here for the parser might need to be applied to normalization or other things. It's information that might be useful outside the parser it's information about how HTML elements behave in general. I guess void elements should also be a fixed set that the user can't change but that is also useful outside parsing like making atomic selections of them etc or normalizing them if they have elements inside.

lnewson: Sorry for the late reply here, but okay I get where you're coming from now so thanks for persevering. I still don't agree however for this specific case (though I do for most other things), as while the two are similar they are different as the custom schema is not a 1:1 mapping of how the native parser works and the rest of TinyMCE operates or uses the schema. That said, I don't think it's a huge issue either way so for now I've made it reuse the `schema.getSpecialElements()` API, added a TODO to freeze it when we can and made it so it's not deprecated since we're going to need to maintain it if you want this.

spocke: Thanks yes it should be frozen and so should the default set of void elements be for example img should always be a void but you should be able to extend it with custom elements that become atomic units. If we have one schema passed around for everything including parsing it one spot to change, one rule set not multiple ones. It's more how we want the schema to work in the future than how it works today where nearly all of it is dynamic and reconfigurable. I regret adding options for configuring some of these things that should be hard coded that is why they where never documented.

PR #7510, opened 2022-01-13
TINY-8202: More changes to get the parser working and passing tests

lnewson:
Related Ticket: TINY-8202

Description of Changes:
* Got the `xhtml` parser side of things working again.
* Removed the parser `xml` format and replaced references in tests with `xhtml` instead as we've never actually parsed XML and the native DOM parser won't parse XML only XHTML.
* Fixed a few issues with the dom purify configuration I found while fixing tests
* Updated the various invalid tests to reflect how the native parser actually parser per the spec.

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/api/html/DomParser.ts
+++ b/modules/tinymce/src/core/main/ts/api/html/DomParser.ts
@@ -93,21 +93,26 @@ interface FilterMatches {
 
 type WalkerCallback = (node: AstNode) => void;
 
-const getPurifyConfig = (settings: DomParserSettings): Config => {
-  const config: Config = {
-    RETURN_DOM: true,
-    ALLOW_DATA_ATTR: true,
-    ALLOWED_URI_REGEXP: /^(?:(?:(?:f|ht)tps?|about|blob|file|mailto|tel|callto|cid|xmpp):|[^a-z]|[a-z+.\-]+(?:[^a-z+.\-:]|$))/i,
-  };
+const basePurifyConfig: Config & { RETURN_DOM: true } = {
+  RETURN_DOM: true,
+  ALLOW_DATA_ATTR: true,
+  ALLOWED_URI_REGEXP: /^(?:(?:(?:f|ht)tps?|about|blob|file|mailto|tel|callto|cid|xmpp):|[^a-z]|[a-z+.\-]+(?:[^a-z+.\-:]|$))/i,
+  // Deliberately ban all tags and attributes by default, and then un-ban them on demand in hooks
+  // #comment and #cdata-section are always allowed as they aren't controlled via the schema
+  ALLOWED_TAGS: [ '#comment', '#cdata-section' ],
+  ALLOWED_ATTR: []
+};
+
+const getPurifyConfig = (settings: DomParserSettings, args: ParserArgs): Config & { RETURN_DOM: true } => {
+  const config = { ...basePurifyConfig };
+
+  if (args.format === 'xhtml') {
+    config.PARSER_MEDIA_TYPE = 'application/xhtml+xml';
+  }
 
   if (settings.allow_html_data_urls) {
     config.ADD_DATA_URI_TAGS = [ 'a', 'img' ];
   }
-
-  // Deliberately ban all tags and attributes by default, and then un-ban them on demand in hooks
-  // #comment and #cdata-section are always allowed as they aren't controlled via the schema
-  config.ALLOWED_TAGS = [ '#comment', '#cdata-section' ];
-  config.ALLOWED_ATTR = [];
 
   return config;
 };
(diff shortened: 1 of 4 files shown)


modules/tinymce/src/core/main/ts/api/html/DomParser.ts
```diff
@@ -97,21 +98,32 @@ type WalkerCallback = (node: AstNode) => void;
 const nonRcDataSpecialElements = new Set('style,xmp,iframe,noembed,noframes,script,noscript,plaintext'.split(','));
 const specialElements = new Set([ 'title', 'textarea', ...nonRcDataSpecialElements ]);
 
-const getPurifyConfig = (settings: DomParserSettings): Config => {
-  const config: Config = {
-    RETURN_DOM: true,
-    ALLOW_DATA_ATTR: true,
-    ALLOWED_URI_REGEXP: /^(?:(?:(?:f|ht)tps?|about|blob|file|mailto|tel|callto|cid|xmpp):|[^a-z]|[a-z+.\-]+(?:[^a-z+.\-:]|$))/i,
```
lnewson, line 104: This looks to have been added to try and support additional unknown protocols, however there's a separate setting for that so swapped to using it.

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
```diff
@@ -97,21 +98,32 @@ type WalkerCallback = (node: AstNode) => void;
...
+  RETURN_DOM: true,
+  ALLOW_UNKNOWN_PROTOCOLS: true,
+  // Deliberately ban all tags and attributes by default, and then un-ban them on demand in hooks
+  // #comment and #cdata-section are always allowed as they aren't controlled via the schema
+  ALLOWED_TAGS: [ '#comment', '#cdata-section' ],
+  ALLOWED_ATTR: []
+};
 
-  if (settings.allow_html_data_urls) {
-    config.ADD_DATA_URI_TAGS = [ 'a', 'img' ];
```
lnewson, line 108: I'd mentioned this on Neil's previous PR but this was definitely wrong, so I've fixed this and other issues with the allow url settings.

modules/tinymce/src/core/test/ts/browser/html/DomParserTest.ts
```diff
@@ -773,17 +854,16 @@ describe('browser.tinymce.core.html.DomParserTest', () => {
...
   });
 
-  // TODO: TINY-4627/TINY-8202: figure out why dompurify returns a different dom structure to us
-  it.skip('TINY-7756: Parsing invalid nested children', () => {
-    const parser = DomParser();
-    const html = '<table><button><a><meta></meta></a></button></table>';
+  it('TINY-7756: Parsing invalid nested children', () => {
+    const schema = Schema({ valid_children: '-td[button|a|div]' });
+    const parser = DomParser({}, schema);
+    const html = '<table><tr><td><button><a><meta /></a></button></td></tr></table>';
```
lnewson, line 860: Had to change this a little as the native parser will also move the element outside if it's not valid. As such, this now uses valid HTML but changes the schema to get the invalid nested children that we needed to test.

ltrouton (approved):

modules/tinymce/src/core/test/ts/browser/html/DomParserTest.ts
```diff
@@ -151,71 +153,71 @@ describe('browser.tinymce.core.html.DomParserTest', () => {
...
+    assert.equal(serializer.serialize(root), '<p class="a"></p><p class="b">123</p><p></p>', 'P in P, splits outer P');
+    assert.deepEqual(countNodes(root), { 'body': 1, 'p': 3, '#text': 1 }, 'P in P, splits outer P (count)');
 
     parser = DomParser({}, schema);
     root = parser.parse('<p class="a">a<p class="b">b</p><p class="c">c</p>d</p>');
     assert.equal(
       serializer.serialize(root),
-      '<p class="a">a</p><p class="b">b</p><p class="c">c</p><p class="a">d</p>',
-      'Two P in P, no nodes before/after'
+      '<p class="a">a</p><p class="b">b</p><p class="c">c</p>d<p></p>',
```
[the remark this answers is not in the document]
lnewson: Yep, that's correct as that's how the HTML spec says to parse it... basically when it sees a nested `<p>` it closes the current paragraph. Then when it sees `</p>` it says, oh there should have been a paragraph there so it creates an empty one. So since `d` isn't inside a paragraph since the one that was around `a` was closed it ends up being outside.

lnewson:
Admin merging as we're still low on people and @metricjs has skimmed this without finding any red flags, but doesn't have time to do a full review.

PR #7511, opened 2022-01-13
TINY-8261: Update the schema option to default to `html5`

lnewson:
Related Ticket: TINY-8261

Description of Changes:
* Updated the `schema` option to default to `html5` instead of `undefined` (they were fairly similar but it did have slight differences)
* Cleaned up the schema code (removed code duplication, added types and cleaned up variables)

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/api/html/Schema.ts
+++ b/modules/tinymce/src/core/main/ts/api/html/Schema.ts
@@ -5,7 +5,7 @@
  * For commercial licenses see https://www.tiny.cloud/
  */
 
-import { Fun, Obj } from '@ephox/katamari';
+import { Arr, Fun, Obj } from '@ephox/katamari';
 
 import Tools from '../util/Tools';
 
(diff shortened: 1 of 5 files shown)


modules/tinymce/src/core/main/ts/api/html/Schema.ts
```diff
@@ -762,13 +746,11 @@ const Schema = (settings?: SchemaSettings): Schema => {
       children[name] = element.children;
     });
 
-    // Switch these on HTML4
-    if (settings.schema !== 'html5') {
```
lnewson, line 766: This we weren't too confident on what should happen, but as this is what was happening previously it's the closest to what most people would be expecting. In addition to that, if you go by the comment then this check wouldn't have worked for `html5-strict` and in addition to that in HTML 4 `b` and `strong` are very different (see below), so it likely wasn't ever correct anyways 🤷 . So for now we're leaving it as is and we'll discuss later when more people are back as to what exactly this should do.

> In HTML 4, the `strong` tag is used for indicating stronger emphasis than the `em` tag. The strong tag surrounds the emphasised word/phrase.

ltrouton (approved):

PR #7517, opened 2022-01-14
TINY-8373: Update default skin for v6, add tinymce5 skin with old design

noxuhax:
Related Ticket: TINY-8373

Description of Changes:
* Updated default theme styling for TinyMCE 6 facelift
* Updated default content theme for TinyMCE 6 facelift
* Added `tinymce5` UI and content skins with legacy styling
* Renamed `dark` UI and content skins to `tinymce5-dark`

Pre-checks:
* [x] Changelog entry added
* [x] Tests have been updated

Review:
* [x] Milestone set


lostkeys (changes requested):
Found some minor issues

When using multiple toolbars a gray horizontal divider appears. Test using:
```
toolbar1: 'undo redo | formatselect | bold italic strikethrough',
toolbar2: 'styleselect removeformat fullscreen',
```

When using toolbar bottom, a gray border appears above the menubar. Test using:
```
toolbar_location: 'bottom',
```

On safari there are still a faint square outline on input fields. Screenshot is taken from the search and replace dialog.

- [x] Fix issue with gray divider above menubar when using toolbar bottom
- [x] Fix issue with divider between multiple toolbars
- [ ] Fix issue with faint square outline on input fields

Screenshot 2022-01-14 at 16 05 31

Screenshot 2022-01-14 at 16 07 16

Screenshot 2022-01-14 at 15 58 13

lostkeys (approved):

lnewson (changes requested):
The changes themselves are mostly okay, but the new UI has a lot of visual issues so I can't approve this sorry. Looking at this image then I can see the following issues:

image

- The fact the lines removed from the statusbar and toolbar makes it look like it's just part of the content, which is rather confusing and a bad user experience.
- The resize handle is barely visible and looks quite out of place (both horizontal alignment and the color)
- There's no visible separator for the toolbar groups which just makes it look like there's this odd incorrect whitespace there
- The statusbar is quite large and takes up far too much visible space than it should be (remember this is meant to be small and largely out of the way, so it's in the way now). This is made to look even worse when using the sidebar.
- I can't quite put my finger on it but the font, fontsize or color of the statusbar text is different to the menubar and as such looks out of place
- The shadow around the editor should definitely not there by default... I can't think of any application that has a shadow around the editor by default. It looks quite bad to me and I cannot imagine this going down well with customers... I'd certainly be curious to see the UAT or customer feedback on this one and if there isn't any then why are we even doing this? (remember we're not a website so we can't just change and tweak things later/we'll be stuck with this until 7.0)

Then I also saw these issues which aren't in the screenshot:
- The menu/menu items heights are different to the rest (toolbar/menubar) which makes it look out of place
- The menu item separator fails color contrast ratios, so it's too light and needs to be made darker to meet our a11y requirements
- The default margin for the content CSS when not on a small screen is **massive** and far too large, as this is not meant to be Word but instead a HTML editor... we have the `document` content CSS if people want that Word document style. Additionally, this is almost certain to cause other issues we've seen with doing this in `document` so we shouldn't/can't be doing this by default (or at least at this stage).

Then inline mode also has some massive issues, as you can barely even see the header/UI now:

image

lnewson:
There also appear to be some regressions in the tinymce-5 skins, as in the following you can see the element path has quite a large left-padding or similar compared to what it used to be:

New:
image

Original:
image

modules/oxide/src/less/theme/components/menubar/menubar.less
```diff
@@ -11,9 +11,9 @@
 
 @menubar-background-color: @background-color;
 @menubar-background: url("data:image/svg+xml;charset=utf8,%3Csvg height='@{_menubar-height}' viewBox='0 0 40 @{_menubar-height}' width='40' xmlns='http://www.w3.org/2000/svg'%3E%3Crect x='0' y='@{_menubar-divider-height}' width='100' height='1' fill='@{_menubar-row-separator-color}'/%3E%3C/svg%3E") left 0 top 0 @menubar-background-color;
```
lnewson, line 13: Why do we have this if we're just making it transparent/not using it now? This seems like it would be better set to a specific color and then overridden to use the background image in the TinyMCE 5 skins.

lostkeys:
@lnewson, your [items in this comment](https://github.com/tinymce/tinymce/pull/7517#pullrequestreview-863126080) above are mostly about visual looks, and that has all be approved by all stakeholders for relatively long while now. I agree many of the changes are quite controversial but we are going to try something new. Please let's focus on the code/implementation side of things for this PR.

Two comments of my own. The old separators did not pass the contrast check either, and as you will see yourself when checking most UI's out there for example Github's, most gray divider lines and boxes will fail contrast checking. That type of UI is best managed using high contrast modes, which TinyMCE respects. So we will not make any changes there.

Finally, you are completely right that inline mode is broken and has to be addressed. I should have catched that myself. 

- [ ] @noxuhax needs to fix inline mode.

modules/oxide/src/less/theme/components/menubar/menubar.less
lnewson: On that note... I don't see a Jira/task to fix the premium skins to account for these changes either? Has that been done, as when I tried to test it I get an error:
```
'src/less/theme/globals/global-variables.less' wasn't found. Tried - node_modules/@tinymce/oxide/src/less/skins/content/default/src/less/theme/globals/global-variables.less,npm://src/less/theme/globals/global-variables.less,src/less/theme/globals/global-variables.less in file node_modules/@tinymce/oxide/src/less/skins/content/default/content.less line no. 7
```

lnewson:
@lostkeys we need **customer** or UAT feedback here, not stakeholder feedback. As I said in a comment, you all (design) seem to treat this like a website whereby we can change it at the drop of a hat... this is a library and as such not something we can change easily. I even passed this feedback on to @EkimChau about this because the new skins/UI is quite bad:
> For the next major release, I think we need to have all UI redesign work done by scope freeze… the new UI has a lot of issues and I doubt some parts are going to go down well with customers.

modules/oxide/src/less/skins/content/dark/content.less
```diff
@@ -4,17 +4,22 @@
  * For LGPL see License.txt in the project root for license information.
  * For commercial licenses see https://www.tiny.cloud/
  */
+@import 'src/less/theme/globals/global-variables.less';
```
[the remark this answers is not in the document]
lostkeys: I agree with @lnewson  here. It seems through that @noxuhax only uses the break point variable. Instead of pulling in the global variables file here, just hard-code the breakpoint in the media query.

lnewson:
To clarify my previous comment further, you've said this is an experiment... we cannot and should not be doing that in an actual release. If we want to experiment for TinyMCE then it should be done via conventional ways such as UAT to get feedback and make sure it's what customers actually want before releasing it.

lnewson:
> and that has all be approved by all stakeholders for relatively long while now.
Actually on this note, as stakeholders even aware of the actual issues I've raised? They may not have even been aware of them... this also very much feels like you're just going to ignore any feedback we provide which is a problem.

lostkeys:
@lnewson you make it sound like I'm just actively ignoring feedback and running my own show. I'm not. Our designers and developers aren't tools in my hand. I don't want to be (and wouldn't be allowed to become) an auteur of TinyMCE, but instead I want to give people the power to both explore their own ideas and also make their own mistakes and learn from them. So even if I would have done things differenly, I respect that other people make different choices. 

We don't do UAT for many other features and changes, so I don't see the need to hold our designers to a different standard. None of these changes are irreversible, we provide both a fallback skin and an alternative without a drop shadow so I don't really see the need to block or thwart some new fresh ideas from experienced talent.

(and yes this design has been showcased several times, at least two times company wide.)

Again, please focus on the implementation side of things. The design itself has already been approved *and* praised by stakeholders.

lnewson:
> We don't do UAT for many other features and changes, so I don't see the need to hold our designers to a different standard. None of these changes are irreversible, we provide both a fallback skin and an alternative without a drop shadow so I don't really see the need to block or thwart some new fresh ideas from experienced talent.

This is quite different, as with all other functionality it's customer requested... this is not a customer request. Additionally, for a number of changes in 6.0 we did have to discuss it with customers before we could make the change so design is **not** being held to the. same standard.

In addition to this, if you look at any controversial feature Engineering has implemented it is **never** the default and is **always** opt-in if we don't have any customer feedback. So you are not. following normal standards here by experimenting with customers in something that we **cannot** change until 7.0 (you seem to have overlooked my previous comment about this)

> (and yes this design has been showcased several times, at least two times company wide.)

This simply not true... the only time it was properly shown company wide was in the all hands which was done very briefly in a way you couldn't get a good look or interact with it and only done less than a week ago. While yes, some of it is nice there's numerous issues with small/finer details which is what I've raised.

Anyways since you're refusing to acknowledge this and bring the feedback back to stakeholders nor are providing a reason (which is what we in Engineering have to do), then I'm going to have to escalate this I'm afraid.

lnewson:
> but instead I want to give people the power to both explore their own ideas and also make their own mistakes and learn from them. So even if I would have done things differenly, I respect that other people make different choices.

I should also reply to this, I'm all for doing this and wouldn't not have an issue if it was done as per our normal processes as I alluded to in my previous message. That means either being a customer request, having some form of research/customer feedback or having the experimental/controversial change be opt-in before making it final.

lostkeys:
@lnewson Please feel free to escalate, but do not block or obstruct the PR merger because of this. If you intend to do just that, please let me know so I can plan accordingly.

lnewson:
> @lnewson Please feel free to escalate, but do not block or obstruct the PR merger because of this. If you intend to do just that, please let me know so I can plan accordingly.

Such issues mean this cannot be merged, similar to all the other examples I gave where we needed feedback and they were also blocked. So yes, this cannot be merged into `develop` (or any actual release branch) until it's been resolved. Feel free to merge into a feature branch if needed to continue other work with this task.

spocke:
I suggest we discuss this in the `tinymce-dev` slack channel instead of a public PR.

TheSpyder:
I think the gap between buttons is now too big compared to the gap between groups. I'd like to see that adjusted. For everything else, in a separate discussion, @lnewson has been overruled. Sorry.

noxuhax:
> I think the gap between buttons is now too big compared to the gap between groups. I'd like to see that adjusted.

After discussion with Paul this have been addressed, thanks!

Additionally, editor in the Inline mode should work fine now.

TheSpyder:
The toolbar separation is much clearer now, thank you. I would've preferred a smaller icon gap, the toolbar icons now expand significantly when upgrading from 5 to 6, but at least the groups are clear.

Unfortunately there is still one big point of contention, I missed the task @lnewson added in last week's review::
> The default margin for the content CSS when not on a small screen is **massive** and far too large, as this is not meant to be Word but instead a HTML editor... we have the `document` content CSS if people want that Word document style. Additionally, this is almost certain to cause other issues we've seen with doing this in `document` so we shouldn't/can't be doing this by default (or at least at this stage).

Putting aside the design for a moment this actually breaks a couple of features. We don't have time to fix these, or track down (and fix) other features that may be impacted by a change to the content margin. I realise this should have been identified in a design review, sorry about that, for now we will need to revert the content changes.

TheSpyder:
There are still test failures in CI - `ListBoxTest.ts` and `ContextToolbarIframePositionTest.ts`. Once those are fixed I'll merge this.

lnewson:
> There are still test failures in CI - `ListBoxTest.ts` and `ContextToolbarIframePositionTest.ts`. Once those are fixed I'll merge this.

Just checking, but does the new statusbar logo need tweaking as well for the `tinymce-5` skin or will that be done in a separate PR? Just basing that off this comment: https://github.com/tinymce/tinymce/pull/7578#pullrequestreview-869333100

TheSpyder:
> does the new statusbar logo need tweaking as well for the `tinymce-5` skin

I think it looks pretty good after the changes I made, I'm not sure what @noxuhax meant in that comment as it's already smaller in the v5 skin. That's how I found the issue and adjusted it in my PR.

noxuhax:
Since there are going to be changes to design in order to address team’s concerns, I’m gonna sunset this branch and subsequent PR. There is going to be a new Jira to reflect those changes and work will continue in a new branch/PR.

> it's already smaller in the v5 skin. That's how I found the issue and adjusted it in my PR.

It is smaller in v5 skin, but still a touch too big. :) I’ve made adjustments for logo size and it’ll be in the new branch as well.

PR #7520, opened 2022-01-17
TINY-8383: Added back `validate: false` support for the parser

lnewson:
Related Ticket: TINY-8383

Description of Changes:
* Make `validate: false` work again by ignoring the configured schema rules and added some basic tests since we largely had none
* Split out the parsing and sanitizing logic so DOMPurify only handles sanitization
* Made some minor performance improvements
* Did some minor code clean up (largely for legacy code)

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/api/html/DomParser.ts
+++ b/modules/tinymce/src/core/main/ts/api/html/DomParser.ts
@@ -227,13 +227,17 @@ const transferChildren = (parent: AstNode, nativeParent: Node, specialElements: 
   // See: https://html.spec.whatwg.org/multipage/parsing.html#parsing-html-fragments
   const isSpecial = Obj.has(specialElements, parentName) && parentName !== 'title' && parentName !== 'textarea';
 
-  Arr.each(nativeParent.childNodes, (nativeChild) => {
+  const childNodes = nativeParent.childNodes;
+  for (let ni = 0, nl = childNodes.length; ni < nl; ni++) {
+    const nativeChild = childNodes[ni];
     const child = new AstNode(nativeChild.nodeName.toLowerCase(), nativeChild.nodeType);
 
     if (NodeType.isElement(nativeChild)) {
-      Arr.each(nativeChild.attributes, (attr) => {
+      const attributes = nativeChild.attributes;
+      for (let ai = 0, al = attributes.length; ai < al; ai++) {
+        const attr = attributes[ai];
         child.attr(attr.name, attr.value);
-      });
+      }
     } else if (NodeType.isText(nativeChild)) {
       child.value = nativeChild.data;
       if (isSpecial) {
@@ -245,7 +249,7 @@ const transferChildren = (parent: AstNode, nativeParent: Node, specialElements: 
 
     transferChildren(child, nativeChild, specialElements);
     parent.append(child);
-  });
+  }
 };
 
 const walkTree = (root: AstNode, preprocessors: WalkerCallback[], postprocessors: WalkerCallback[]) => {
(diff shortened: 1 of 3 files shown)


ltrouton (approved):

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
```diff
@@ -376,15 +376,35 @@ const getRootBlockName = (settings: DomParserSettings, args: ParserArgs) => {
...
+    validate: true,
+    root_name: 'body',
+    ...settings
+  };
+
+  const parser = new DOMParser();
+  const purify = setupPurify(defaultedSettings, schema);
+
+  const parseAndSanitizeWithContext = (html: string, rootName: string, format: string = 'html') => {
+    const mimeType = format === 'xhtml' ? 'application/xhtml+xml' : 'text/html';
```
spocke, line 394: Will we ever use this xhtml parser part or should we just hard code it to be `text/html`?

modules/tinymce/src/core/test/ts/browser/html/SerializerTest.ts
```diff
@@ -28,14 +28,13 @@ describe('browser.tinymce.core.html.SerializerTest', () => {
...
-  // TODO: TINY-4627/TINY-8383
-  it.skip('Serialize with validate: true, when parsing with validate:false bug', () => {
+  it('Serialize with validate: true, when parsing with validate:false bug', () => {
     const schema = Schema({ valid_elements: 'b' });
     const serializer = HtmlSerializer({}, schema);
 
     assert.equal(
       serializer.serialize(DomParser({ validate: false }, schema).parse('<b a="1" b="2">a</b><i a="1" b="2">b</i>')),
-      '<b a="1" b="2">a</b><i a="1" b="2">b</i>'
+      '<b b="2" a="1">a</b><i b="2" a="1">b</i>'
```
spocke, line 37: Why is this in reverse order now? Should we change the HtmlWriter to sort them even when validate is false to combat this. Could other tests be failing now since it no longer producing the same output as before?

spocke (approved):

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
lnewson: We currently support both XHTML and HTML in TinyMCE 5 so we need to support both unless someone higher up says we can remove both. Given our deadlines we likely aren't going to be able to attempt that for 6.x.

modules/tinymce/src/core/test/ts/browser/html/SerializerTest.ts
lnewson: This was something discussed in stand ups sorry, but we said we weren't concerned about attribute order for things like this as it's not part of the HTML spec. It's also not related to `validate: false` and that's not set for the serializer. So if the order is defined by the schema then it will order them appropriately, but if not it'll just return them however the native parser orders them (or however DOMPurify reorganizes them).

lnewson: I should add I'm not exactly sure why it's in the reverse order, but this had come up in Neil's previous PRs and he had attributed it to the above so I assumed the same. If you're concerned I can look into it more, but it wasn't something I thought we needed to worry about.

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
spocke: Yes, but a html5 parser can parse xhtml will this not force the document to be required to be a XHTML valid document don't think we ever supported that strict ruleset.

The older parser would just parse things like CDATA and PI elements however it would still parse non XML as `<ul><li>foo</ul>` a strict XHTML parser will not and produces a document with a parsererror element.

So it's more about do we need to retain CDATA and PI elements in the parsing step. I don't think we need to and if we do we could preprocess them so they get parsed properly as comments or something that we later can convert back into CDATA and PI elements int he ast tree.

But this is only for people using the parser directly this will not be used if they configure the output format to be xhtml right. So the risk of this new strict parsing behavior causing issues is low?

modules/tinymce/src/core/test/ts/browser/html/SerializerTest.ts
spocke: Ok. I just get worried when I see tests being changed and not knowing the context of that change.

HAFRMO (approved):

modules/tinymce/src/core/test/ts/browser/html/SerializerTest.ts
lnewson: So I checked why this has reversed and it's because of how DOMPurify does it's sanitization of attributes here: https://github.com/cure53/DOMPurify/blob/84f1d45/src/purify.js#L1146. Basically it'll walk backwards and remove than re-add valid attributes, which in turn causes the attribute order to be reversed.

I can't see any good reason why they forcibly remove it and add it back, but the reverse order makes sense to account for the fact that removing an attribute would cause iteration issues if going forwards. We might be able to submit a change to only remove if required which would prevent it reordering if we're really concerned.

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
lnewson: > But this is only for people using the parser directly this will not be used if they configure the output format to be xhtml right. So the risk of this new strict parsing behavior causing issues is low?

Yes, this would only be used when using the API directly... tbh I believe this was only ever used in the `fullpage` plugin so it's something we should look at removing in 7.0.

PR #7521, opened 2022-01-17
TINY-7524: Ensure dialog labels and other text based fields only render as text

lnewson:
Related Ticket: TINY-7524/TINY-8340

Description of Changes:
* Fixed a number of dialog labels and other text-based UI fields to ensure they are always rendered as text content only.

Note: I haven't touched `AlertBanner.ts` or `Notification.ts` as I know those do actually need to allow HTML at this stage.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/acid/src/main/ts/ephox/acid/gui/components/RgbForm.ts
+++ b/modules/acid/src/main/ts/ephox/acid/gui/components/RgbForm.ts
@@ -1,5 +1,5 @@
 import {
-  AddEventsBehaviour, AlloyComponent, AlloyEvents, AlloyTriggers, Behaviour, EventFormat, Focusing, Form, FormField, FormTypes, Input, Invalidating,
+  AddEventsBehaviour, AlloyComponent, AlloyEvents, AlloyTriggers, Behaviour, EventFormat, Focusing, Form, FormField, FormTypes, GuiFactory, Input, Invalidating,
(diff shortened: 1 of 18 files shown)


ltrouton (approved):

spocke (approved):
Should this change have some tests in case we refactor at some point and regress I guess we are pressed for time but maybe just a simple open a dialog with some html stuck in the labels kind of thing?

Also the branch and PR title has different ticket numbers probably because it's a duplicate. I would suggest to have the same at least since doesn't that information end up in the merge commit?

lnewson:
> Should this change have some tests in case we refactor at some point and regress I guess we are pressed for time but maybe just a simple open a dialog with some html stuck in the labels kind of thing?

Yeah, you're right we are pretty under the pump atm as we have about 3x as much work as we have time which is why I had skipped it. That said, I'll add some tests for the more common cases (e.g FieldLabeller) as this is something we really don't want to regress.

> Also the branch and PR title has different ticket numbers probably because it's a duplicate. I would suggest to have the same at least since doesn't that information end up in the merge commit?

Yeah it was a bit of a mess as I realised after I logged the PR that TINY-8340 was actually a duplicate and the original had a lot more information. Anyways, the PR title is the only thing that is used for the commit message since we use squash commits, so the branch name won't matter here (if we used merge commits that'd be another story).

PR #7523, opened 2022-01-17
TINY-8329: Now fire beforeinput events when overriding events

HAFRMO:
Related Ticket:

Description of Changes:
Added beforeInput events, and refactored code to work around it.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/tinymce/src/core/main/ts/delete/TableDelete.ts
```diff
@@ -242,68 +240,70 @@ const isDeleteOfLastCharPos = (fromCaption: SugarElement<HTMLTableCaptionElement
...
+): Optional<() => void> => {
   const from = CaretPosition.fromRangeStart(editor.selection.getRng());
   return Empty.isEmpty(fromCaption) ?
     emptyElement(editor, fromCaption) :
     deleteCaretInsideCaption(editor, rootElm, forward, fromCaption, from);
 };
 
-const isNearTable = (forward: boolean, pos: CaretPosition): boolean =>
-  forward ? isBeforeTable(pos) : isAfterTable(pos);
+const isNearTable = (forward: boolean, pos: CaretPosition): Optional<() => void> => {
```
spocke, line 284: Not sure everything needs to be optional thunks since read operations shouldn't be a problem to have happening before blocking things.

modules/tinymce/src/core/main/ts/delete/TableDelete.ts
```diff
@@ -242,68 +240,70 @@ const isDeleteOfLastCharPos = (fromCaption: SugarElement<HTMLTableCaptionElement
...
 };
 
-const deleteCaret = (editor: Editor, forward: boolean, startElm: SugarElement<Node>): boolean => {
+const deleteCaret = (editor: Editor, forward: boolean, startElm: SugarElement<Node>): Optional<() => void> => {
   const rootElm = SugarElement.fromDom(editor.getBody());
 
   return getParentCaption(rootElm, startElm).fold(
-    () => deleteCaretCells(editor, forward, rootElm, startElm) || isBeforeOrAfterTable(editor, forward),
-    (fromCaption) => deleteCaretCaption(editor, forward, rootElm, fromCaption).getOr(false)
+    () => deleteCaretCells(editor, forward, rootElm, startElm).orThunk(() => isBeforeOrAfterTable(editor, forward)),
```
spocke, line 301: In this file I think this is where you need to thunk not on all predicates.

spocke (commented):
Looks pretty good. I think just some cleanup and maybe reduce the thunking where it's not really needed. Also I wonder if it should have some helper functions to make it more readable in like a OptionalThunk.ts module then have runThunk, makeThunk, makeNoopThunk etc then have all Optional<() => void> be a OptionalThunk type.

Just a suggestion not sure if makes it that much more readable.

modules/tinymce/src/core/main/ts/delete/TableDelete.ts
```diff
@@ -242,68 +240,70 @@ const isDeleteOfLastCharPos = (fromCaption: SugarElement<HTMLTableCaptionElement
         from.isEqual(last) && to.isEqual(first))
   ).getOr(true);
 
-const emptyCaretCaption = (editor: Editor, elm: SugarElement<Node>): Optional<boolean> =>
-  emptyElement(editor, elm);
+const emptyCaretCaption = (editor: Editor, elm: SugarElement<Node>): Optional<Optional<() => void>> =>
```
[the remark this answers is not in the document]
HAFRMO: Because it's a tri-state: "Could not produce result", "Result is none", "Result is some", and in the event of the first we run a thunk.

modules/tinymce/src/core/main/ts/delete/ImageBlockDelete.ts
```diff
@@ -13,19 +13,20 @@ import CaretPosition from '../caret/CaretPosition';
...
     .filter((pos) => forward ? isBeforeImageBlock(pos) : isAfterImageBlock(pos))
     .bind((pos) => Optional.from(getChildNodeAtRelativeOffset(forward ? 0 : -1, pos)))
-    .exists((elm) => {
-      editor.selection.select(elm);
-      return true;
-    });
+    .map((elm) =>
+      () => {
+        editor.selection.select(elm);
+        return true;
```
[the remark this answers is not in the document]
HAFRMO: My guess is that you can return things, and it's still valid, it's just that calling functions know not to expect any value? Same way you can call a function with more parameters than it can accept.

modules/tinymce/src/core/main/ts/delete/TableDelete.ts
```diff
@@ -242,68 +240,70 @@ const isDeleteOfLastCharPos = (fromCaption: SugarElement<HTMLTableCaptionElement
...
+  Optional.some(emptyElement(editor, elm));
 
-const validateCaretCaption = (rootElm: SugarElement<Node>, fromCaption: SugarElement<HTMLTableCaptionElement>, to: CaretPosition): Optional<boolean> =>
+const validateCaretCaption = (rootElm: SugarElement<Node>, fromCaption: SugarElement<HTMLTableCaptionElement>, to: CaretPosition): Optional<Optional<() => void>> =>
   getParentCaption(rootElm, SugarElement.fromDom(to.getNode()))
-    .map((toCaption) => Compare.eq(toCaption, fromCaption) === false);
+    .map((toCaption) => Compare.eq(toCaption, fromCaption) ? Optional.none() : Optional.some(Fun.noop));
```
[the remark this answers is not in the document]
HAFRMO: As stated above, this is a tri-state situation, so this one time it's actually intended.

modules/tinymce/src/core/main/ts/delete/TableDelete.ts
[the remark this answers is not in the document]
HAFRMO: To clarify, this was previously `Optional<boolean>`. Since this refactoring has changed this to `Optional<() => void>` there's no longer any difference between what was before `Optional.none` and `Optional<false>` at this particular point in the code. Since the function here runs a thunk on `Optional.none` but not on false, I had to work around it, and this seemed like the one that worked the best with Optionals.

modules/tinymce/src/core/main/ts/delete/TableDelete.ts
HAFRMO: I tried to keep the generation of `Optional.some` in the same functions that previously had `return true`, as it serves a similar purpose now.

modules/tinymce/src/core/main/ts/delete/TableDelete.ts
HAFRMO: I'm not sure what you mean?

modules/tinymce/src/core/main/ts/delete/TableDelete.ts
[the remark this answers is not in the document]
spocke: I think that could be refactored away. I think the Optional booleans should be able to be reworked into folds seems that the `deleteBetweenCells` and the `deleteCaretInsideCaption` functions are the ones causing this all to be optional booleans. If they are reworked a bit maybe using folds instead of binds then I think it should work. Then `emptyElement` becomes a return of a thunked option of some and the `deleteCaretInsideCaption` would either return a thunked action that some something or a noop thunk to block the default action. The `deleteCaretCells` default action is not to block So that should just return a thunked option if it's some from the `getParentCell` call.

modules/tinymce/src/core/main/ts/delete/TableDelete.ts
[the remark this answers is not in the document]
spocke: So the three different states can be described as Optional.none(), Optional.some(someThunkThatDoesSomething) and Optional.some(Fun.noop).

modules/tinymce/src/core/main/ts/delete/TableDelete.ts
spocke: `isBeforeOrAfterTable` is a predicate and should remain a predicate this is not using options as boolean but rather producing a thunk when something needs to be done or not. That thunk can either be just block the browser native action a Fun.noop or some action that actually alters the dom.

modules/tinymce/src/core/main/ts/delete/TableDelete.ts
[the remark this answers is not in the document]
HAFRMO: I tried to add folds as you wanted, judge for yourself if it improved readability.

modules/tinymce/src/core/main/ts/delete/TableDelete.ts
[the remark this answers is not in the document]
HAFRMO: No, that was not at all what the states were. But I've rewritten it a bit on your request anyway.

modules/tinymce/src/core/main/ts/delete/TableDelete.ts
HAFRMO: See if you like that better.

modules/tinymce/src/core/main/ts/delete/TableDelete.ts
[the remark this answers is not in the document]
spocke: I think the folds are better here since you should never really have an option in an option. It should probably not have been a optional boolean either it's probably my fault being lazy and just wanting the types to align or something.

modules/tinymce/src/core/main/ts/delete/DeleteCommands.ts
```diff
@@ -25,54 +25,47 @@ const nativeCommand = (editor: Editor, command: string): void => {
...
+    BlockBoundaryDelete.backspaceDelete,
+    TableDelete.backspaceDelete,
+    ImageBlockDelete.backspaceDelete,
+    MediaDelete.backspaceDelete,
+    BlockRangeDelete.backspaceDelete,
+    InlineFormatDelete.backspaceDelete
+  ], (item) => item(editor, true));
+
+  result.fold(
+    () => {
```
spocke, line 64: You could get rid of the block { } here just have it be a one line lambda.

spocke (commented):
I guess it also needs tests that the beforeinput is fired and that we can prevent the default action.

modules/tinymce/src/core/main/ts/keyboard/MatchKeys.ts
```diff
@@ -26,7 +32,17 @@ const defaultPatterns = (patterns: KeyPattern[]): KeyPattern[] => Arr.map(patter
...
 }));
 
-const matchesEvent = (pattern: KeyPattern, evt: KeyboardEvent) => (
+const defaultDelayedPatterns = (patterns: KeyPatternDelayed[]): KeyPatternDelayed[] => Arr.map(patterns, (pattern) => ({
+  shiftKey: false,
+  altKey: false,
+  ctrlKey: false,
+  metaKey: false,
+  keyCode: 0,
+  action: Fun.noop,
```
lnewson, line 41: This doesn't match the type, so this shouldn't be compiling... looks like this should be this?
```suggestion
  action: () => Optional.some(Fun.noop),
```

Also there's a lot of duplication here with `defaultPatterns` and `defaultDelayedPatterns`, so can we please move that out into a global variable? e.g.
```
const baseKeyPattern = {
  shiftKey: false,
  altKey: false,
  ctrlKey: false,
  metaKey: false,
  keyCode: 0
};
```

- [ ] Remove duplication

modules/tinymce/src/core/main/ts/keyboard/SpaceKey.ts
```diff
@@ -8,17 +8,22 @@
...
+import { fireFakeBeforeInputEvent, fireFakeInputEvent } from './FakeInputEvents';
 import * as InsertSpace from './InsertSpace';
 import * as MatchKeys from './MatchKeys';
 
 const executeKeydownOverride = (editor: Editor, evt: KeyboardEvent) => {
-  MatchKeys.execute([
+  MatchKeys.executeWithDelayedAction([
     { keyCode: VK.SPACEBAR, action: MatchKeys.action(InsertSpace.insertSpaceOrNbspAtSelection, editor) }
-  ], evt).each(() => {
+  ], evt).each((apply) => {
```
lnewson, line 18: For consistency can you please call this `applyAction` or change your other version to just be `apply`?

modules/tinymce/src/core/main/ts/keyboard/SpaceKey.ts
```diff
@@ -8,17 +8,22 @@
...
     evt.preventDefault();
-    // Browsers sends space in data even if the dom ends up with a nbsp so we should always be sending a space
-    fireFakeInputEvent(editor, 'insertText', { data: ' ' });
+    const event = fireFakeBeforeInputEvent(editor, 'insertText', { data: ' ' });
+
+    if (!event.defaultPrevented) {
```
lnewson, line 22: Similar to the previous comment please use `isDefaultPrevented()` instead.

modules/tinymce/src/core/main/ts/newline/InsertNewLine.ts
```diff
@@ -9,23 +9,31 @@ import { Fun, Type } from '@ephox/katamari';
...
 
 const insert = (editor: Editor, evt?: EditorEvent<KeyboardEvent>) => {
   NewLineAction.getAction(editor, evt).fold(
     () => {
-      InsertBr.insert(editor, evt);
-      if (Type.isNonNullable(evt)) {
-        fireFakeInputEvent(editor, 'insertLineBreak');
+      const event = fireFakeBeforeInputEvent(editor, 'insertLineBreak');
+
+      if (!event.defaultPrevented) {
```
lnewson, line 22: Likewise for the 2 places here, please use `isDefaultPrevented()`.

modules/tinymce/src/core/main/ts/newline/InsertNewLine.ts
```diff
@@ -9,23 +9,31 @@ import { Fun, Type } from '@ephox/katamari';
...
 import * as InsertBr from './InsertBr';
 import * as NewLineAction from './NewLineAction';
 
 const insert = (editor: Editor, evt?: EditorEvent<KeyboardEvent>) => {
   NewLineAction.getAction(editor, evt).fold(
     () => {
-      InsertBr.insert(editor, evt);
-      if (Type.isNonNullable(evt)) {
-        fireFakeInputEvent(editor, 'insertLineBreak');
+      const event = fireFakeBeforeInputEvent(editor, 'insertLineBreak');
```
lnewson, line 20: This logic doesn't look correct... it should not be firing the `beforeinput` event here if `evt` is nullable, as `input` also won't fire if `evt` is nullable.

- [ ] Fix this and below to only fire `beforeInput` if `evt` is nonnullable

modules/tinymce/src/core/test/ts/browser/delete/BeforeInputTest.ts
```diff
@@ -0,0 +1,56 @@
...
+    editor.nodeChanged();
+    TinyContentActions.keydown(editor, Keys.backspace());
+    editor.off('beforeinput', beforeInputCollect);
+    editor.off('input', collect);
+
+    TinyAssertions.assertContent(editor, expectedHtml);
+
+    assert.deepEqual(inputEvents, cancelBeforeInput ? [ 'deleteContentBackward' ] : [ 'deleteContentBackward', 'deleteContentBackward' ]);
+
+    editor.getBody().normalize();
```
lnewson, line 46: Why is this here? That seems quite odd so if it's here for a legit reason can you please add a comment?

modules/tinymce/src/core/test/ts/browser/delete/TableDeleteTest.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable mocha/no-exclusive-tests */
```
lnewson, line 1: This shouldn't be here, I'm guessing it was left in by accident?

modules/tinymce/CHANGELOG.md
```diff
@@ -42,6 +42,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The `GetContent` event was not fired when getting `tree` or `text` formats using the `editor.selection.getContent()` API #TINY-8018
 - The `table` plugin would sometimes not correctly handle headers in the `tfoot` section #TINY-8104
 - Deleting content would sometimes not fire `input` events as expected #TINY-8168
+- Deleting content would sometimes not fire `BeforeInput` events as expected #TINY-8329
```
lnewson, line 45:
```suggestion
- Deleting content would sometimes not fire `beforeinput` and `input` events as expected #TINY-8168 #TINY-8329
```
This can be collapsed to a single line

modules/tinymce/src/core/main/ts/keyboard/SpaceKey.ts
```diff
@@ -8,17 +8,22 @@
 import Editor from '../api/Editor';
 import { EditorEvent } from '../api/util/EventDispatcher';
 import VK from '../api/util/VK';
-import { fireFakeInputEvent } from './FakeInputEvents';
+import { fireFakeBeforeInputEvent, fireFakeInputEvent } from './FakeInputEvents';
 import * as InsertSpace from './InsertSpace';
 import * as MatchKeys from './MatchKeys';
 
 const executeKeydownOverride = (editor: Editor, evt: KeyboardEvent) => {
```
lnewson, line 15: This before input logic needs some tests as I didn't see any added for it.

- [ ] Add `beforeinput` tests for the space key override

lnewson (commented):
Looking quite good overall, nice work @HAFRMO!

modules/tinymce/src/core/main/ts/delete/InlineBoundaryDelete.ts
```diff
@@ -89,42 +87,49 @@ const backspaceDeleteCollapsed = (editor: Editor, caret: Cell<Text>, forward: bo
...
-    .getOrThunk(() => {
-      const toPosition = CaretFinder.navigate(forward, rootNode, from);
-      const toLocation = toPosition.bind((pos) => BoundaryLocation.readLocation(isInlineTarget, rootNode, pos));
-
-      return Optionals.lift2(fromLocation, toLocation, () =>
-        InlineUtils.findRootInline(isInlineTarget, rootNode, from).exists((elm) => {
-          if (hasOnlyTwoOrLessPositionsLeft(elm)) {
+  });
+
+  const newLocation = location.map(setCaretLocation(editor, caret));
```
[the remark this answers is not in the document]
HAFRMO: This was mostly for trying to figure out what was going on, since the chaining made it confusing to read in some spots, and didn't show types, but it's not needed now so I'll chain it back up.

modules/tinymce/src/core/main/ts/delete/InlineBoundaryDelete.ts
```diff
@@ -89,42 +87,49 @@ const backspaceDeleteCollapsed = (editor: Editor, caret: Cell<Text>, forward: bo
...
-      ).orThunk(() => toLocation.bind((_) =>
-        toPosition.map((to) => {
+          });
+        } else {
+          return Optional.none();
+        }
+      })
+    );
+
+    const thunk = () => toLocation.bind(() =>
```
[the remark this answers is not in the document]
HAFRMO: I disagree that it makes it harder to read, I find it quite the opposite, but I'll change it back.

modules/tinymce/src/core/main/ts/keyboard/DeleteBackspaceKeys.ts
```diff
@@ -44,9 +46,14 @@ const executeKeydownOverride = (editor: Editor, caret: Cell<Text>, evt: Keyboard
     { keyCode: VK.DELETE, action: MatchKeys.action(BlockBoundaryDelete.backspaceDelete, editor, true) },
     { keyCode: VK.BACKSPACE, action: MatchKeys.action(InlineFormatDelete.backspaceDelete, editor, false) },
     { keyCode: VK.DELETE, action: MatchKeys.action(InlineFormatDelete.backspaceDelete, editor, true) }
-  ], evt).each((value) => {
+  ], evt).each((applyAction) => {
     evt.preventDefault();
-    fireFakeInputEvent(editor, value.keyCode === VK.BACKSPACE ? 'deleteContentBackward' : 'deleteContentForward');
+    const beforeInput = fireFakeBeforeInputEvent(editor, inputType);
+
+    if (!beforeInput.defaultPrevented) {
```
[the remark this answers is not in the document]
HAFRMO: According to the types `isDefaultPrevented` does not exist on `InputEvent`s, so I don't think I can change it to that? Have I picked the wrong type somewhere?

modules/tinymce/src/core/main/ts/newline/InsertNewLine.ts
HAFRMO: However, we change the DOM on the `InsertBr.Insert`, don't we? Which we don't want to do before the beforeinput, or if it's cancelled.

modules/tinymce/src/core/test/ts/browser/delete/TableDeleteTest.ts
HAFRMO: Correct. I was debugging, and forgot to remove this one. Sorry about that.

modules/tinymce/src/core/test/ts/browser/delete/BeforeInputTest.ts
HAFRMO: I believe the reason is just that I poached the test from a different file, compressed it, and then forgot to clean out all the things that weren't actually needed. I've removed it.

modules/tinymce/src/core/main/ts/delete/DeleteCommands.ts
```diff
@@ -25,54 +25,45 @@ const nativeCommand = (editor: Editor, command: string): void => {
...
-  } else if (MediaDelete.backspaceDelete(editor, true)) {
-    return;
-  } else if (BlockRangeDelete.backspaceDelete(editor, true)) {
-    return;
-  } else if (InlineFormatDelete.backspaceDelete(editor, true)) {
-    return;
-  } else {
-    nativeCommand(editor, 'ForwardDelete');
-  }
+  const result = Arr.findMap([
```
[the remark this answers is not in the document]
HAFRMO: Deduped.

modules/tinymce/src/core/main/ts/delete/MediaDelete.ts
```diff
@@ -32,16 +33,17 @@ const deleteCaret = (editor: Editor, forward: boolean): boolean => {
   } else {
     return Optional.from(InlineUtils.normalizePosition(forward, fromPos))
       .filter((pos) => isNearMedia(pos) && CaretUtils.isMoveInsideSameBlock(fromPos, pos))
-      .exists((pos) => deleteElement(editor, forward, pos.getNode(!forward)));
+      .map((pos) =>
+        () => deleteElement(editor, forward, pos.getNode(!forward)));
```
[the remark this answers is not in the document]
HAFRMO: Done.

modules/tinymce/src/core/main/ts/delete/TableDelete.ts
```diff
@@ -242,41 +240,45 @@ const isDeleteOfLastCharPos = (fromCaption: SugarElement<HTMLTableCaptionElement
...
+const emptyCaretCaption = (editor: Editor, elm: SugarElement<Node>): Optional<() => void> =>
   emptyElement(editor, elm);
 
-const validateCaretCaption = (rootElm: SugarElement<Node>, fromCaption: SugarElement<HTMLTableCaptionElement>, to: CaretPosition): Optional<boolean> =>
+const validateCaretCaption = (rootElm: SugarElement<Node>, fromCaption: SugarElement<HTMLTableCaptionElement>, to: CaretPosition): Optional<() => void> =>
   getParentCaption(rootElm, SugarElement.fromDom(to.getNode()))
-    .map((toCaption) => Compare.eq(toCaption, fromCaption) === false);
+    .fold(
+      () => Optional.some(Fun.noop),
+      (toCaption) => Compare.eq(toCaption, fromCaption) ? Optional.none() : Optional.some(Fun.noop)
```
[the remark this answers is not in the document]
HAFRMO: I didn't realize we had that one, fixed it.

modules/tinymce/src/core/main/ts/keyboard/FakeInputEvents.ts
```diff
@@ -37,6 +39,33 @@ const fireFakeInputEvent = (editor: Editor, inputType: string, specifics: Specif
   editor.fire('input', { ...input, ...overrides, ...specifics });
 };
 
+const fireFakeBeforeInputEvent = (editor: Editor, inputType: string, specifics: SpecificsInput = {}): InputEvent => {
+  const target = editor.getBody();
+  const overrides = {
```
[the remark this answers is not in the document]
HAFRMO: Deduped.

modules/tinymce/src/core/main/ts/keyboard/MatchKeys.ts
HAFRMO: It seems like it doesn't typecheck at all in this one. Good catch with this mistake.
Also deduped as requested.

modules/tinymce/src/core/main/ts/keyboard/SpaceKey.ts
HAFRMO: Added a few more tests. I kept them in separate files to try and avoid making one big filed that tested a bunch of unrelated functions, but if we prefer it in a different way do tell.

modules/tinymce/src/core/main/ts/keyboard/DeleteBackspaceKeys.ts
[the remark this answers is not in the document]
lnewson: Yeah your return types are incorrect, it should be `EditorEvent<InputEvent>` instead as that's the return type for `fire`.

modules/tinymce/src/core/main/ts/newline/InsertNewLine.ts
lnewson: I'm not sure what you mean here... it's not related as to whether or not we change the DOM. It's related to if this was triggered by an event, so if this is triggered programattically then we shouldn't be firing `beforeinput` or `input` events.

modules/tinymce/CHANGELOG.md
```diff
@@ -41,8 +41,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The `editor.selection.getContent()` API did not respect the `no_events` argument #TINY-8018
 - The `GetContent` event was not fired when getting `tree` or `text` formats using the `editor.selection.getContent()` API #TINY-8018
 - The `table` plugin would sometimes not correctly handle headers in the `tfoot` section #TINY-8104
-- Deleting content would sometimes not fire `input` events as expected #TINY-8168
-- Deleting content would sometimes not fire `BeforeInput` events as expected #TINY-8329
+- Deleting content would sometimes not fire `BeforeInput` and `input` events as expected #TINY-8168  #TINY-8329
```
lnewson, line 44:
```suggestion
- Deleting content would sometimes not fire `beforeinput` and `input` events as expected #TINY-8168  #TINY-8329
```
I had corrected this in my previous suggestion but you seem to have missed this.

modules/tinymce/src/core/main/ts/delete/InlineBoundaryDelete.ts
```diff
@@ -56,7 +56,8 @@ const deleteFromTo = (editor: Editor, caret: Cell<Text>, from: CaretPosition, to
 
     BoundaryLocation.readLocation(isInlineTarget, rootNode, CaretPosition.fromRangeStart(editor.selection.getRng()))
       .map(BoundaryLocation.inside)
-      .bind(setCaretLocation(editor, caret)).each((action) => action());
+      .bind(setCaretLocation(editor, caret))
+      .each((action) => action());
```
lnewson, line 60:
```suggestion
      .each(Fun.call);
```

modules/tinymce/src/core/main/ts/newline/InsertNewLine.ts
HAFRMO: No, I mean, I assumed we wanted to be able to cancel the input and prevent changes to the DOM, whereas this one would be un-blockable even when you get the event.
I'm not saying that the event shouldn't be moved not to fire when not appropriate, just that we shouldn't change the DOM even if the input is cancelled? Or did I get something wrong on that?

modules/tinymce/src/core/main/ts/keyboard/DeleteBackspaceKeys.ts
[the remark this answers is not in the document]
HAFRMO: That would explain it.

modules/tinymce/src/core/main/ts/newline/InsertNewLine.ts
lnewson: You can't cancel it using the event in that case, because it's not been triggered by an event. As I said, you should only be able to do that when this logic is triggered via an event.

lnewson: As an example, this logic can be triggered via the `mceInsertNewLine` so firing an event there makes no sense and instead if we wanted that cancelled you'd cancel the command instead.

modules/tinymce/src/core/main/ts/newline/InsertNewLine.ts
```diff
@@ -19,13 +19,21 @@ const insert = (editor: Editor, evt?: EditorEvent<KeyboardEvent>) => {
     () => {
       InsertBr.insert(editor, evt);
       if (Type.isNonNullable(evt)) {
-        fireFakeInputEvent(editor, 'insertLineBreak');
+        const event = fireFakeBeforeInputEvent(editor, 'insertLineBreak');
```
spocke, line 22: Maybe there is some confusion here the branching should be done if there is a event or not. If there is a event passed in then it should fire the before input event to allow the user to prevent the default action, then execute the default action and after that fire a input event. If there is no event passed in it should not fire the beforeInput event or the input event since then it's called from the execCommand. I guess a proper cleanup of this extra parameter should be done after this PR at some point to have two separate functions one that deals with the event and one that deals with direct api calls or something.

- [ ] Branch on the existence of the event and add tests that verify that the default insert action can be prevented if beforeinput is calling preventDefault.

modules/tinymce/src/core/main/ts/newline/InsertNewLine.ts
```diff
@@ -19,13 +19,21 @@ const insert = (editor: Editor, evt?: EditorEvent<KeyboardEvent>) => {
...
+        if (!event.isDefaultPrevented()) {
+          fireFakeInputEvent(editor, 'insertLineBreak');
+        }
       }
     },
     () => {
       InsertBlock.insert(editor, evt);
       if (Type.isNonNullable(evt)) {
-        fireFakeInputEvent(editor, 'insertParagraph');
+        const event = fireFakeBeforeInputEvent(editor, 'insertParagraph');
```
spocke, line 32: Same here branch on the event insert or fire events and insert.

modules/tinymce/src/core/main/ts/newline/InsertNewLine.ts
```diff
@@ -9,21 +9,39 @@ import { Fun, Type } from '@ephox/katamari';
...
 import * as InsertBr from './InsertBr';
 import * as NewLineAction from './NewLineAction';
 
 const insert = (editor: Editor, evt?: EditorEvent<KeyboardEvent>) => {
   NewLineAction.getAction(editor, evt).fold(
     () => {
+      if (Type.isNonNullable(evt)) {
+        const event = fireFakeBeforeInputEvent(editor, 'insertLineBreak');
+
+        if (event.isDefaultPrevented()) {
```
lnewson, line 23:
```suggestion
        const event = fireFakeBeforeInputEvent(editor, 'insertLineBreak');
        if (event.isDefaultPrevented()) {
```
This is a little too verbose/spaced out and as such makes it harder to read, especially given this is all one logical unit.

modules/tinymce/src/core/main/ts/newline/InsertNewLine.ts
```diff
@@ -9,21 +9,39 @@ import { Fun, Type } from '@ephox/katamari';
...
+
       if (Type.isNonNullable(evt)) {
         fireFakeInputEvent(editor, 'insertLineBreak');
       }
     },
     () => {
+      if (Type.isNonNullable(evt)) {
+        const event = fireFakeBeforeInputEvent(editor, 'insertParagraph');
+
+        if (event.isDefaultPrevented()) {
```
lnewson, line 38:
```suggestion
        const event = fireFakeBeforeInputEvent(editor, 'insertParagraph');
        if (event.isDefaultPrevented()) {
```

modules/tinymce/src/core/main/ts/newline/InsertNewLine.ts
lnewson: Thanks for fixing this up 👍  I also agree we should split this logic at some point, but this is fine for now.

lnewson (approved):

modules/tinymce/src/core/test/ts/browser/util/BeforeInputEventUtils.ts
```diff
@@ -0,0 +1,49 @@
...
+    editor.setContent(setupHtml);
+    TinySelections.setCursor(editor, setupPath, setupOffset);
+    performEditAction(editor);
+    editor.off('beforeinput', beforeInputCollect);
+    editor.off('input', collect);
+
+    TinyAssertions.assertContent(editor, expectedHtml);
+
+    const moreEvents = expectMoreEventsIfSafari && Env.browser.isSafari();
+    const expectedNoCancelEvents = moreEvents ? [ eventType, eventType, eventType ] : [ eventType, eventType ];
```
spocke, line 42: This is not the right approach to this patching over the test so that it passes even when it's sending multiple events is not the right way to solve this. We need to figure out why these extra events are being sent. The patch I sent you should work around that I will double check why you couldn't get it to work since it passed on my local Safari.

spocke (changes requested):

modules/tinymce/src/core/test/ts/browser/util/BeforeInputEventUtils.ts
spocke: Ok worked some on this. I made an error in the patch I sent you and only tested the beforeinput tests sorry about that. I corrected that and removed the Safari workarounds and all tests now pass on CI. I put that in a separate branch you can cherry pick the HEAD of that. 

- [ ] Cherry pick from https://github.com/tinymce/tinymce/tree/feature/TINY-8329_2

spocke (changes requested):

modules/tinymce/src/core/test/ts/browser/util/BeforeInputEventUtils.ts
HAFRMO: Are we sure this is the solution we want to go for? This issue about double-firing the input event feels like it's a bug in some unrelated code, and should be fixed separately rather than trying to patch the signs of it like this?

spocke: Yes, this is all about normalizing browser behavior and not firing multiple events. Our custom delete logic should just fire one `beforeInput` and one `input` event. Safari seems to be firing both `beforeInput` and `input` events when doing for example `document.execCommand('Delete')` other browsers seems to only fire the `input` event. So we could just use the same method that we used before for preventing double `input` events from being emitted by the native `execCommand` call by also include the `beforeinput` events to the stop the propagation to other handlers.

So we know why they are emitted and how to prevent that form getting handled by the tinymce event system. So that is what the patch I provided fixes rather than fixing the tests. The tests should only be branched based on browsers if it doesn't really matter if one browser does one thing and another browser does another thing or as a last resort if there really is no workaround for a particular browser. In this case it matters we want to avoid emitting multiple events and we can also work around it.

I also separated the `beforeinput` and `input` events into two buckets as mentioned before we can't put them together into one bucket for assertion since then we don't know what event added that particular input event type. It could be that `beforeinput` fired twice or `input` fired twice or `beforeinput` and `input` fired once all would result in a single array with two items in it. We want to check for the latter that beforeinput fired and input fired once.

spocke (approved):

PR #7527, opened 2022-01-19
TINY-8260: Fix formatter test that was failing in isolation

ltrouton:
Related Ticket: TINY-8260

Description of Changes:
* Make sure editor is focused when running test
* Change `block: 'del'` tests as `del` is an inline element

Pre-checks:

Review:
* [X] Milestone set

### DIFF
--- a/modules/tinymce/src/core/test/ts/browser/FormatterApplyTest.ts
+++ b/modules/tinymce/src/core/test/ts/browser/FormatterApplyTest.ts
@@ -20,7 +20,7 @@ describe('browser.tinymce.core.FormatterApplyTest', () => {
         'margin,margin-top,margin-right,margin-bottom,margin-left,display,text-align'
     },
     base_url: '/project/tinymce/js/tinymce'
-  }, []);
+  }, [], true);
 
   const getContent = (editor: Editor) => {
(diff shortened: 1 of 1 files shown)


lnewson (approved):

modules/tinymce/src/core/test/ts/browser/FormatterApplyTest.ts
```diff
@@ -1771,27 +1771,16 @@ describe('browser.tinymce.core.FormatterApplyTest', () => {
...
     const editor = hook.editor();
     editor.setContent('<p>a</p>');
     LegacyUnit.setSelection(editor, 'p', 0, 'p', 1);
     editor.formatter.register('format', {
-      block: 'del',
+      block: 'div',
       wrapper: true
     });
     editor.formatter.apply('format');
-    assert.equal(getContent(editor), '<p><del><p>a</p></del></p>');
```
spocke, line 1783: Del is one of those weird elements where it can be a inline or a block depending on context. So in this particular case it should be a inline and not allowed to contain that P. At some point we should start looking into supporting these things it's similar to block/inline `a` elements.

spocke (approved):

PR #7534, opened 2022-01-20
TINY-8164: Remove XHR API and update plugins to use fetch

ltrouton:
Related Ticket: TINY-8164

Description of Changes:
* Remove deprecated public tinymce XHR API
* Change code using the tinymce XHR API to use native `fetch` instead

I am not exactly sure if this is how we want to use `fetch` but I tried to make it behave as close to the old XHR API as possible

Pre-checks:
* [X] Changelog entry added

Review:
* [X] Milestone set


modules/tinymce/src/plugins/image/main/ts/core/Utils.ts
```diff
@@ -114,12 +113,12 @@ const createImageList = (editor: Editor, callback: (imageList: false | UserListI
   const imageList = Options.getImageList(editor);
 
   if (Type.isString(imageList)) {
-    XHR.send({
-      url: imageList,
-      success: (text) => {
-        callback(JSON.parse(text));
-      }
-    });
+    window.fetch(imageList)
```
lnewson, line 116: Instead of doing this can we please add `fetch` to the list of globals we can use in the `.eslintrc` configuration? See https://github.com/tinymce/tinymce/blob/develop/modules/tinymce/.eslintrc.json#L10

modules/tinymce/src/plugins/link/main/ts/ui/sections/LinkListOptions.ts
```diff
@@ -27,18 +26,22 @@ const getLinks = (editor: Editor): Promise<Optional<ListItem[]>> => {
...
-  return new Promise<Optional<UserListItem[]>>((callback) => {
+  return new Promise<Optional<UserListItem[]>>((resolve) => {
     // TODO - better handling of failure
     if (Type.isString(linkList)) {
-      XHR.send({
-        url: linkList,
-        success: (text) => callback(parseJson(text)),
-        error: (_) => callback(Optional.none())
-      });
+      window.fetch(linkList)
```
lnewson, line 32: Similar to before, lets add `fetch` to the allowed dom globals list instead of having to use `window.fetch` everywhere.

modules/tinymce/src/plugins/link/main/ts/ui/sections/LinkListOptions.ts
```diff
@@ -27,18 +26,22 @@ const getLinks = (editor: Editor): Promise<Optional<ListItem[]>> => {
...
+        .then((res) => {
+          if (res.ok) {
+            res.text().then((text) => resolve(parseJson(text)));
+          } else {
+            resolve(Optional.none());
+          }
+        })
+        .catch(() => resolve(Optional.none()));
```
lnewson, line 40: This doesn't handle the case where `res.text()` throws an error. I think we'd be better doing something like this maybe?
```suggestion
        .then((res) => res.ok ? res.text().then(parseJson) : Promise.reject())
        .then(resolve, () => resolve(Optional.none()));
```

modules/tinymce/src/plugins/template/main/ts/ui/Dialog.ts
```diff
@@ -110,15 +109,16 @@ const open = (editor: Editor, templateList: ExternalTemplate[]): void => {
...
+      () => resolve(t.value.content.getOr('')),
+      (url) => window.fetch(url).then((res) => {
+        if (res.ok) {
+          return resolve(res.text());
+        } else {
+          return reject();
+        }
+      }).catch(reject)
+    );
   });
```
lnewson, line 122: This can be re-written to remove the `Promise` wrapper now. For example:
```suggestion
  const getTemplateContent = (t: InternalTemplate) =>
    t.value.url.fold(
      () => Promise.resolve(t.value.content.getOr('')),
      (url) => window.fetch(url).then((res) => res.ok ? res.text() : Promise.reject())
    );
```

lnewson (approved):

modules/tinymce/src/plugins/image/main/ts/core/Utils.ts
ltrouton: Done and changed the uses of `window.fetch` to `fetch`

modules/tinymce/src/plugins/link/main/ts/ui/sections/LinkListOptions.ts
ltrouton: This is better. Thanks

modules/tinymce/src/plugins/template/main/ts/ui/Dialog.ts
ltrouton: Fixed

spocke (approved):

TheSpyder (approved):

PR #7535, opened 2022-01-20
TINY-8382: Updated the media plugin to work with the new DomParser

lnewson:
Related Ticket: TINY-8382

Description of Changes:
* Remove the `special` element overrides in the media plugin
* Update the media plugin to serialise the child nodes and use that for the `data-mce-html` to replace the old special override
* Replace uses of `SaxParser` with `DomParser` throughout the media plugin
* Moved the `allowfullscreen` bool attribute config to the core schema
* Fixed a schema issue where it incorrectly allowed nested `audio` and `video` elements
* Sealed the `special` elements in the schema since they aren't something that should be changed

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/tinymce/src/plugins/media/main/ts/core/HtmlToData.ts
```diff
@@ -5,89 +5,90 @@
...
+          data.source = node.attr('movie');
         }
 
         if (name === 'iframe' || name === 'object' || name === 'embed' || name === 'video' || name === 'audio') {
           if (!data.type) {
             data.type = name;
           }
 
-          data = Tools.extend(attrs.map, data);
+          data = Tools.extend(node.attributes.map, data);
```
spocke, line 61: I wonder if we should add a function to return the map instead of using that map object obviously not part of this PR. I have abused that in a few places either for performance or just being lazy. It's also very strange that the map property is stuck into the array so essentially replacing the map function at the time arrays didn't have a map. It should have been it's own _map or something instead on the actual node.

modules/tinymce/src/plugins/media/main/ts/core/FilterContent.ts
```diff
@@ -5,64 +5,54 @@
...
-  editor.on('preInit', () => {
-    // Make sure that any messy HTML is retained inside these
-    const specialElements = editor.schema.getSpecialElements();
-    Tools.each('video audio iframe object'.split(' '), (name) => {
-      specialElements[name] = new RegExp('<\/' + name + '[^>]*>', 'gi');
+  editor.on('PreInit', () => {
+    const { schema, serializer, parser } = editor;
+    // Set browser specific allowFullscreen attribs as boolean
+    const boolAttrs = schema.getBoolAttrs();
+    Arr.each('webkitallowfullscreen mozallowfullscreen'.split(' '), (name) => {
```
spocke, line 23: I wonder how long we need to support these. But I guess Flash is dead as well this plugin really needs to be rewritten.

spocke (approved):

modules/tinymce/src/plugins/media/main/ts/core/FilterContent.ts
lnewson: Yeah I seriously considered removing these tbh and you're right this entire plugin is rather messy or likely has obsolete functionality given todays standards.

modules/tinymce/src/plugins/media/main/ts/core/HtmlToData.ts
lnewson: Looking at the notes for 7.0, one idea is to remove the AST node implementation entirely which would negate having to worry about any of that. That might not happen, but it's been listed as an idea 🤷

ltrouton (approved):

modules/tinymce/src/plugins/media/main/ts/core/HtmlToData.ts
spocke: That would require a new form of api since raw dom elements doesn't have a way to get the attributes out as a pure object map. Also we would have to add unwrap, wrap etc some where else. I guess that will have to be a 8.0 task seems like a lot of work to go though all that now. 😞

PR #7536, opened 2022-01-20
TINY-8411: Various fixes to be able to upgrade to the latest bedrock/monorepo libraries

lnewson:
Related Ticket: TINY-8411/TINY-8414

Description of Changes:
* Upgrade Agar to remove uses of the legacy `assert` API that was removed in Bedrock 13
* Upgrade Agar deps to allow using version 11, 12 or 13 of bedrock
* Fix `SandHTMLElement.isPrototypeOf` throwing an `Illegal invocation` error for `HTMLAnchorElement` on Chromium based browsers (it doesn't like calling `toString` on the prototype)

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/agar/src/main/ts/ephox/agar/assertions/ApproxStructures.ts
+++ b/modules/agar/src/main/ts/ephox/agar/assertions/ApproxStructures.ts
@@ -1,4 +1,4 @@
-import { Assert, assert, TestLabel } from '@ephox/bedrock-client';
+import { Assert, TestLabel } from '@ephox/bedrock-client';
 import { Arr, Fun, Obj, Optional } from '@ephox/katamari';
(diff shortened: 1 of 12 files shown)


modules/agar/src/main/ts/ephox/agar/assertions/ApproxStructures.ts
```diff
@@ -1,4 +1,4 @@
-import { Assert, assert, TestLabel } from '@ephox/bedrock-client';
+import { Assert, TestLabel } from '@ephox/bedrock-client';
```
lnewson, line 1: This intentionally uses `Assert` from bedrock instead of `assert` from chai so that we don't have to add chai as a dependency to agar.

spocke (approved):

ltrouton (approved):

PR #7541, opened 2022-01-20
TINY-8169: added new approach for adding noneditable fake carets

spocke:
Related Ticket:

Description of Changes:
* Adds a new approach to figuring out where to render fake carets.
* Also fixed two unrelated tests that where failing while the 100 test chunk pivot point was moving around. They are depending on state from the previous test.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -23,7 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The dialog `redial` API will now only rerender the changed components instead of the whole dialog #TINY-8334
 - The dialog API `setData` method now uses a deep merge algorithm to support partial nested objects #TINY-8333
 - The dialog spec `initialData` type is now `Partial<T>` to match the underlying implementation details #TINY-8334
-- Improved support for placing the caret before/after noneditable elements within the editor #TINY-8169
+- Improved support for placing the caret before or after noneditable elements within the editor #TINY-8169
 
 ### Changed
 - The `editor.getContent()` API can provide custom content by preventing and overriding `content` in the `BeforeGetContent` event. This makes it consistent with the `editor.selection.getContent()` API #TINY-8018
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The dialog `redial` API will now only rerender the changed components instead of the whole dialog #TINY-8334
 - The dialog API `setData` method now uses a deep merge algorithm to support partial nested objects #TINY-8333
 - The dialog spec `initialData` type is now `Partial<T>` to match the underlying implementation details #TINY-8334
+- Improved support for placing the caret before/after noneditable elements within the editor #TINY-8169
 
 ### Changed
 - The `editor.getContent()` API can provide custom content by preventing and overriding `content` in the `BeforeGetContent` event. This makes it consistent with the `editor.selection.getContent()` API #TINY-8018
--- a/modules/tinymce/src/core/main/ts/SelectionOverrides.ts
+++ b/modules/tinymce/src/core/main/ts/SelectionOverrides.ts
@@ -5,22 +5,18 @@
  * For commercial licenses see https://www.tiny.cloud/
  */
 
-import { Arr, Obj, Type, Unicode } from '@ephox/katamari';
+import { Arr, Obj, Unicode } from '@ephox/katamari';
 import { Attribute, Compare, Css, Focus, Insert, InsertAll, Remove, SelectorFilter, SelectorFind, SugarElement } from '@ephox/sugar';
 
 import Editor from './api/Editor';
 import VK from './api/util/VK';
 import * as CaretContainer from './caret/CaretContainer';
-import CaretPosition from './caret/CaretPosition';
-import { isAfterContentEditableFalse, isAfterMedia, isBeforeContentEditableFalse, isBeforeMedia } from './caret/CaretPositionPredicates';
 import * as CaretUtils from './caret/CaretUtils';
-import { CaretWalker } from './caret/CaretWalker';
+import * as ClosestCaretCandidate from './caret/ClosestCaretCandidate';
 import { FakeCaret, isFakeCaretTarget } from './caret/FakeCaret';
 import * as FakeCaretUtils from './caret/FakeCaretUtils';
-import * as LineUtils from './caret/LineUtils';
 import * as CefUtils from './dom/CefUtils';
 import * as NodeType from './dom/NodeType';
-import * as RangePoint from './dom/RangePoint';
 import * as DragDropOverrides from './DragDropOverrides';
 import * as EditorView from './EditorView';
 import * as CefFocus from './focus/CefFocus';
@@ -28,7 +24,6 @@ import * as EditorFocus from './focus/EditorFocus';
 import * as MediaFocus from './focus/MediaFocus';
 import * as Rtc from './Rtc';
 
-const isContentEditableTrue = NodeType.isContentEditableTrue;
 const isContentEditableFalse = NodeType.isContentEditableFalse;
 
 interface SelectionOverrides {
@@ -54,8 +49,6 @@ const SelectionOverrides = (editor: Editor): SelectionOverrides => {
   // Note: isChildOf will return true if node === rootNode, so we need an additional check for that
   const isFakeSelectionTargetElement = (node: Node): node is HTMLElement =>
     node !== rootNode && (isContentEditableFalse(node) || NodeType.isMedia(node)) && dom.isChildOf(node, rootNode);
-  const isNearFakeSelectionElement = (pos: CaretPosition) =>
-    isBeforeContentEditableFalse(pos) || isAfterContentEditableFalse(pos) || isBeforeMedia(pos) || isAfterMedia(pos);
 
   const getRealSelectionElement = () => {
     const container = dom.get(realSelectionId);
@@ -67,8 +60,6 @@ const SelectionOverrides = (editor: Editor): SelectionOverrides => {
       selection.setRng(range);
     }
   };
-
-  const getRange = selection.getRng;
 
   const showCaret = (direction: number, node: HTMLElement, before: boolean, scrollIntoView: boolean = true): Range => {
     const e = editor.fire('ShowCaret', {
@@ -96,16 +87,6 @@ const SelectionOverrides = (editor: Editor): SelectionOverrides => {
   };
 
   const registerEvents = () => {
-    // Some browsers (Chrome) lets you place the caret after a cE=false
-    // Make sure we render the caret container in this case
-    editor.on('mouseup', (e) => {
-      const range = getRange();
-
-      if (range.collapsed && EditorView.isXYInContentArea(editor, e.clientX, e.clientY)) {
-        FakeCaretUtils.renderCaretAtRange(editor, range, false).each(setRange);
-      }
-    });
-
     editor.on('click', (e) => {
       const contentEditableRoot = getContentEditableRoot(editor, e.target);
       if (contentEditableRoot) {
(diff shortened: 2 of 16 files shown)


modules/tinymce/src/core/main/ts/caret/ClosestCaretCandidate.ts
```diff
@@ -0,0 +1,137 @@
...
+        if (CaretCandidate.isCaretCandidate(node)) {
+          return Optional.some(rect);
+        } else if (NodeType.isElement(node)) {
+          const childRect = closestChildCaretCandidateNodeRect(Arr.from(node.childNodes), clientX, clientY);
+          return childRect.isSome() ? childRect : helper(i + 1, rects);
+        }
+      }
+    };
+
+    return helper(0, rects.sort((r1, r2) => distance(r1, clientX, clientY) - distance(r2, clientX, clientY)));
```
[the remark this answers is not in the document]
spocke: Yes, it probably should forgot that the array function is mutating the array and returning it.

modules/tinymce/src/core/main/ts/caret/CaretContainerInput.ts
```diff
@@ -23,6 +23,7 @@ const findBlockCaretContainer = (editor: Editor): HTMLElement | null =>
 const showBlockCaretContainer = (editor: Editor, blockCaretContainer: HTMLElement): void => {
   if (blockCaretContainer.hasAttribute('data-mce-caret')) {
     CaretContainer.showCaretContainerBlock(blockCaretContainer);
+    editor.selection.setRng(editor.selection.getRng()); // Clears the fake caret state
```
spocke, line 26: This was a regression when the IE specific things got removed. It was doing more than just IE control selection removal. I added a test for this as well since we didn't have one that failed.

modules/tinymce/src/plugins/table/test/ts/browser/FakeSelectionTest.ts
```diff
@@ -139,7 +139,7 @@ describe('browser.tinymce.plugins.table.FakeSelectionTest', () => {
 
   it('TINY-7724: does not select CEF cell if contenteditable=true child is selected', () => {
     const editor = hook.editor();
-    editor.setContent('<table><tr><td contenteditable="false"><p contenteditable="true">1</p></td><td>2</td></tr><tr><td>3</td><td>4</td></tr></table>');
+    editor.setContent('<table><tbody><tr><td contenteditable="false"><p contenteditable="true">1</p></td><td>2</td></tr><tr><td>3</td><td>4</td></tr></tbody></table>');
```
spocke, line 142: I added a tbody here since it was confusing to align the selection paths with browser normalization they always normalize the tables to have a tbody. I didn't do it for the rest of this file though seemed out of scope of this task.

TheSpyder (approved):

modules/tinymce/CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The dialog `redial` API will now only rerender the changed components instead of the whole dialog #TINY-8334
 - The dialog API `setData` method now uses a deep merge algorithm to support partial nested objects #TINY-8333
 - The dialog spec `initialData` type is now `Partial<T>` to match the underlying implementation details #TINY-8334
+- Improved support for placing the caret before/after noneditable elements within the editor #TINY-8169
```
lnewson, line 26:
```suggestion
- Improved support for placing the caret before or after noneditable elements within the editor #TINY-8169
```
Docs has said we can't use `/` as it's not always clear, so just replacing it with something else.

modules/tinymce/src/core/main/ts/SelectionOverrides.ts
```diff
@@ -192,38 +124,28 @@ const SelectionOverrides = (editor: Editor): SelectionOverrides => {
...
-          }
-        }
-      } else if (isFakeSelectionTargetElement(targetElm)) {
-        FakeCaretUtils.selectNode(editor, targetElm).each(setElementSelection);
-      } else if (isFakeCaretTarget(targetElm) === false) {
-        // Remove needs to be called here since the mousedown might alter the selection without calling selection.setRng
-        // and therefore not fire the AfterSetSelectionRange event.
-        removeElementSelection();
-        hideFakeCaret();
+          const range = showCaret(1, caretInfo.node as HTMLElement, caretInfo.position === ClosestCaretCandidate.FakeCaretPosition.Before, false);
```
lnewson, line 139: Not something to do now, but we probably should use an enum here instead of `1` to mean forwards one day 🤷

modules/tinymce/src/core/main/ts/caret/ClosestCaretCandidate.ts
```diff
@@ -0,0 +1,132 @@
...
+  Before = 'before',
+  After = 'after'
+}
+
+export interface FakeCaretInfo {
+  node: Node;
+  position: FakeCaretPosition;
+}
+
+type distanceFn = <T extends GeomClientRect>(rect: T, x: number, y: number) => number;
```
lnewson, line 29: Can we please rename this to `DistanceFn`? as we never start types with a lower case character anywhere else in the code base and it looks too much like a variable.

modules/tinymce/src/core/test/ts/browser/ClickContentEditableFalseTest.ts
```diff
@@ -29,20 +29,32 @@ describe('browser.tinymce.core.ClickContentEditableFalseTest', () => {
...
+  });
+
   it('Click on content editable false inside content editable true', () => {
     const editor = hook.editor();
     editor.setContent('<div contenteditable="true"><p contenteditable="false">a</p></div>');
     clickMiddleOf(editor, [ 0, 1 ]);
     TinyAssertions.assertSelection(editor, [ 0 ], 0, [ 0 ], 1);
   });
 
-  it('Click on content editable true inside content editable false', () => {
```
lnewson, line 39: Is there a reason this test was deleted?

modules/tinymce/src/core/test/ts/browser/caret/ClosestCaretCandidateTest.ts
```diff
@@ -0,0 +1,581 @@
...
+  position: FakeCaretPosition;
+}
+
+interface TestArgs<T> {
+  html: string;
+  targetPath: number[];
+  dx: number;
+  dy: number;
+  expected: Optional<T>;
+}
```
lnewson, line 20:
```suggestion
interface TestCaretInfo {
  readonly path: number[];
  readonly position: FakeCaretPosition;
}

interface TestArgs<T> {
  readonly html: string;
  readonly targetPath: number[];
  readonly dx: number;
  readonly dy: number;
  readonly expected: Optional<T>;
}
```

modules/tinymce/src/core/test/ts/browser/caret/FakeCaretTypeTest.ts
```diff
@@ -0,0 +1,30 @@
...
+import { TinyHooks, TinySelections, TinyAssertions, TinyContentActions } from '@ephox/wrap-mcagar';
+
+import Editor from 'tinymce/core/api/Editor';
+import TablePlugin from 'tinymce/plugins/table/Plugin';
+
+describe('browser.tinymce.core.FakeCaretTypeTest', () => {
+  const hook = TinyHooks.bddSetupLight<Editor>({
+    base_url: '/project/tinymce/js/tinymce',
+    indent: false
+  }, [ TablePlugin ]);
```
lnewson, line 11: Is there a reason we include the Table plugin module here but don't have it in the config?

lnewson (approved):
Nice work Spocke, this must have been a fair effort rewriting this logic 🏆

modules/tinymce/src/core/test/ts/browser/caret/FakeCaretTypeTest.ts
spocke: Oops. I copied the setup from `FirefoxFakeCaretBeforeTableTypeTest` and forgot to remove the table plugin.

modules/tinymce/src/core/test/ts/browser/ClickContentEditableFalseTest.ts
spocke: It used to call editor.selection.placeCaretAt to put the caret inside the content editable element this new approach simply lets the browser default execute for this case. So if we want to keep this it would have to be a webdriver click that actually moves the caret. However not sure it's worth the effort.

modules/tinymce/src/core/main/ts/caret/ClosestCaretCandidate.ts
TheSpyder: Is there a way we can lint this? ReScript is the opposite, types are _required_ to start with a lowercase letter. So it might come up a lot as some of us straddle the two languages.

modules/tinymce/src/core/test/ts/browser/ClickContentEditableFalseTest.ts
lnewson: That's fair, I'm just asking as I seem to recall this being a customer reported issue and I was worried about it regressing. Happy to go with whatever you think though 👍

modules/tinymce/src/core/test/ts/browser/caret/FakeCaretTypeTest.ts
lnewson: I figured it was probably something simple like that 😄

modules/tinymce/CHANGELOG.md
spocke: Will try to remember that. But makes sense since it could mean `and` or `or` depending on how you read it.

modules/tinymce/src/core/main/ts/caret/ClosestCaretCandidate.ts
lnewson: There is, however we currently have it disabled in this repo because we have a few cases of the opposite direction to fix or we need to come up with a rule to match the naming used in TinyMCE IIRC: https://github.com/tinymce/tinymce/blob/develop/.eslintrc.json#L12

spocke: Yes, exactly my thought we probably need a eslint rule I will make this mistake again and maybe others that need to context switch between languages. I looked around for some built in option but couldn't find any might exist some plugin or we could write our own. Should be a fairly easy rule to write since it applies to all type declarations.

modules/tinymce/src/core/test/ts/browser/ClickContentEditableFalseTest.ts
spocke: I did some digging and the test was added for this fix. https://ephocks.atlassian.net/browse/TINY-1001 and commit here: https://github.com/tinymce/tinymce/commit/7745fca2c091c86a4dfb28488f854d7bb3be3383 however the placeCaretAt call has been there before that see: https://github.com/tinymce/tinymce/commit/90e42ea4f6009d7f3808a89cf0167ac3b53c42d5

If I recall correctly it had to do with Chrome not rendering and this related PR suggest that is the case.
https://github.com/tinymce/tinymce/commit/77a24c2ecb554d9f4dda5f0ccf3c3ed27898a1ac

However I could no longer reproduce any caret issues with Chrome at least not on Linux. So I will add this to the QA notes to look for that issue if it still exists. It's very hard to automatically test for since the caret gets placed correctly in the selection model but the actual visual rendering was broken the caret was invisible if I remember correctly I should have added better comments for that one. It has probably been fixed in the browser since the fix was added 7 years ago.

lnewson: Sounds good, thanks for digging that up Spocke!

modules/tinymce/src/core/main/ts/caret/ClosestCaretCandidate.ts
spocke: Seems to be a rule that we might be able to use here https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/naming-convention.md#enforce-the-codebase-follows-eslints-camelcase-conventions

spocke: Added a PR here: https://github.com/tinymce/eslint-plugin/pull/27 but maybe we merge that after the Feb 15 release not super important just a nice to have but might break things if there is projects that are using lowercase.

PR #7544, opened 2022-01-21
TINY-4627: Patch dompurify to apply some minor performance improvements

lnewson:
Related Ticket: TINY-4627

Description of Changes:

This patch is only meant to be short-lived as we should contribute this back upstream if all works well, but this just makes some fairly minor performance improvements I was able to find for DOMPurify.

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [x] Milestone set

### DIFF
--- a/patches/dompurify+2.3.4.patch
+++ b/patches/dompurify+2.3.4.patch
@@ -1,0 +1,104 @@
+diff --git a/node_modules/dompurify/dist/purify.es.js b/node_modules/dompurify/dist/purify.es.js
+index 899e471..5844faf 100644
+--- a/node_modules/dompurify/dist/purify.es.js
++++ b/node_modules/dompurify/dist/purify.es.js
+@@ -617,6 +617,12 @@ function createDOMPurify() {
+ 
+   var HTML_INTEGRATION_POINTS = addToSet({}, ['foreignobject', 'desc', 'title', 'annotation-xml']);
+ 
++  // Certain elements are allowed in both SVG and HTML
++  // namespace. We need to specify them explicitly
++  // so that they don't get erronously deleted from
++  // HTML namespace.
++  var COMMON_SVG_AND_HTML_ELEMENTS = addToSet({}, ['title', 'style', 'font', 'a', 'script']);
++
+   /* Keep track of all possible SVG and MathML tags
+    * so that we can perform the namespace checks
+    * correctly. */
+@@ -701,15 +707,9 @@ function createDOMPurify() {
+         return false;
+       }
+ 
+-      // Certain elements are allowed in both SVG and HTML
+-      // namespace. We need to specify them explicitly
+-      // so that they don't get erronously deleted from
+-      // HTML namespace.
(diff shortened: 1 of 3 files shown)


patches/dompurify+2.3.4.patch
```diff
@@ -0,0 +1,52 @@
...
++++ b/node_modules/dompurify/dist/purify.js
+@@ -623,6 +623,12 @@
+ 
+     var HTML_INTEGRATION_POINTS = addToSet({}, ['foreignobject', 'desc', 'title', 'annotation-xml']);
+ 
++    // Certain elements are allowed in both SVG and HTML
++    // namespace. We need to specify them explicitly
++    // so that they don't get erronously deleted from
++    // HTML namespace.
++    var COMMON_SVG_AND_HTML_ELEMENTS = addToSet({}, ['title', 'style', 'font', 'a', 'script']);
```
lnewson, line 13: This was moved to a constant, as it was getting re-declared for every element being sanitized which caused a lot of additional object creation which in turn triggered the GC more often.

patches/dompurify+2.3.4.patch
```diff
@@ -0,0 +1,52 @@
...
++        return !ALL_MATHML_TAGS[tagName] && (COMMON_SVG_AND_HTML_ELEMENTS[tagName] || !ALL_SVG_TAGS[tagName]);
+       }
+ 
+       // The code should never reach this place (this means
+@@ -908,7 +908,7 @@
+       }
+ 
+       /* Check if tagname contains Unicode */
+-      if (stringMatch(currentNode.nodeName, /[\u0080-\uFFFF]/)) {
++      if (regExpTest(/[\u0080-\uFFFF]/, currentNode.nodeName)) {
```
lnewson, line 40: This likely makes very little difference, but `Regex.text` is faster than `String.match`. There are even eslint rules for this normally, see: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-regexp-test.md

patches/dompurify+2.3.4.patch
```diff
@@ -0,0 +1,52 @@
...
+         _forceRemove(currentNode);
+         return true;
+       }
+@@ -923,7 +923,7 @@
+       });
+ 
+       /* Detect mXSS attempts abusing namespace confusion */
+-      if (!_isNode(currentNode.firstElementChild) && (!_isNode(currentNode.content) || !_isNode(currentNode.content.firstElementChild)) && regExpTest(/<[/\w]/g, currentNode.innerHTML) && regExpTest(/<[/\w]/g, currentNode.textContent)) {
++      if (currentNode.hasChildNodes() && !_isNode(currentNode.firstElementChild) && (!_isNode(currentNode.content) || !_isNode(currentNode.content.firstElementChild)) && regExpTest(/<[/\w]/g, currentNode.innerHTML) && regExpTest(/<[/\w]/g, currentNode.textContent)) {
```
lnewson, line 49: Another check that likely makes little difference, but this saves all the checks happening and the innerHTML/textContent having to be computed if the node has no children.

ltrouton (approved):

spocke (approved):

PR #7550, opened 2022-01-24
TINY-4627: More minor parser optimizations

lnewson:
Related Ticket: TINY-4627

Description of Changes:

This is just some more fixes both to our logic and to DOMPurify to help with the parsing speed. This like the previous PR doesn't make a huge difference (largely a number of micro optimizations), but in my testing between this and the previous PR it's shaven off between 10-20% in general (really depends on the content and how much the GC runs).

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/api/html/DomParser.ts
+++ b/modules/tinymce/src/core/main/ts/api/html/DomParser.ts
@@ -6,7 +6,7 @@
  */
 
 import { Arr, Fun, Obj, Strings, Type } from '@ephox/katamari';
-import { Attribute, Remove, Replication, SugarElement, SugarNode } from '@ephox/sugar';
+import { Attribute, NodeTypes, Remove, Replication, SugarElement } from '@ephox/sugar';
 import createDompurify, { Config, DOMPurifyI } from 'dompurify';
 
 import * as NodeType from '../../dom/NodeType';
@@ -16,7 +16,7 @@ import * as ParserFilters from '../../html/ParserFilters';
 import { isEmpty, isLineBreakNode, isPaddedWithNbsp, paddEmptyNode } from '../../html/ParserUtils';
 import { BlobCache } from '../file/BlobCache';
 import Tools from '../util/Tools';
-import URI from '../util/URI';
+import * as URI from '../util/URI';
 import AstNode from './Node';
 import Schema, { SchemaRegExpMap } from './Schema';
 
@@ -131,19 +131,21 @@ const setupPurify = (settings: DomParserSettings, schema: Schema): DOMPurifyI =>
 
   // We use this to add new tags to the allow-list as we parse, if we notice that a tag has been banned but it's still in the schema
   purify.addHook('uponSanitizeElement', (ele, evt) => {
-    const element = SugarElement.fromDom(ele);
-    const tagName = evt.tagName;
     // Just leave non-elements such as text and comments up to dompurify
-    if (!SugarNode.isElement(element) || SugarNode.name(element) === 'body') {
+    const tagName = evt.tagName;
+    if (ele.nodeType !== NodeTypes.ELEMENT || tagName === 'body') {
       return;
     }
+
+    // Construct the sugar element wrapper
+    const element = SugarElement.fromDom(ele);
 
     // Determine if the schema allows the element and either add it or remove it
     const rule = schema.getElementRule(tagName.toLowerCase());
     if (validate && !rule) {
       Remove.unwrap(element);
       return;
-    } else if (!Obj.has(evt.allowedTags, tagName)) {
+    } else {
       evt.allowedTags[tagName] = true;
     }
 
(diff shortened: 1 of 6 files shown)


modules/tinymce/src/core/main/ts/api/html/DomParser.ts
```diff
@@ -131,19 +131,21 @@ const setupPurify = (settings: DomParserSettings, schema: Schema): DOMPurifyI =>
 
   // We use this to add new tags to the allow-list as we parse, if we notice that a tag has been banned but it's still in the schema
   purify.addHook('uponSanitizeElement', (ele, evt) => {
-    const element = SugarElement.fromDom(ele);
-    const tagName = evt.tagName;
     // Just leave non-elements such as text and comments up to dompurify
-    if (!SugarNode.isElement(element) || SugarNode.name(element) === 'body') {
+    const tagName = evt.tagName;
+    if (ele.nodeType !== NodeTypes.ELEMENT || tagName === 'body') {
```
lnewson, line 136: This saves us having to lookup the node name from the DOM and also avoids constructing the sugar wrapper if not needed.

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
```diff
@@ -198,14 +200,12 @@ const setupPurify = (settings: DomParserSettings, schema: Schema): DOMPurifyI =>
     const { attrName, attrValue } = evt;
 
     evt.keepAttr = !validate || Strings.startsWith(attrName, 'data-') || schema.isValid(tagName, attrName);
-    if (attrName in filteredUrlAttrs && !URI.isDomSafe(attrValue, tagName, settings)) {
+    if (!settings.allow_script_urls && attrName in filteredUrlAttrs && URI.isInvalidUri(settings, attrValue, tagName)) {
```
lnewson, line 203: The `isDomSafe` function decodes the value passed which isn't needed here, as since this was parsed and looked up from the DOM it's already been decoded which was adding unnecessary overhead.

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
```diff
@@ -225,7 +225,7 @@ const transferChildren = (parent: AstNode, nativeParent: Node, specialElements:
   const parentName = parent.name;
   // Exclude the special elements where the content is RCDATA as their content needs to be parsed instead of being left as plain text
   // See: https://html.spec.whatwg.org/multipage/parsing.html#parsing-html-fragments
-  const isSpecial = Obj.has(specialElements, parentName) && parentName !== 'title' && parentName !== 'textarea';
+  const isSpecial = parentName in specialElements && parentName !== 'title' && parentName !== 'textarea';
```
lnewson, line 228: Now to start with, I know this isn't quite the same but our `Obj.has` implementation is a reasonable amount slower than doing this (between 2-3x slower). This meant that we were wasting ~15-20ms for the 1000 para test so I figured the pros outweighed the cons here but am happy to revert this if we want the absolute safety.

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
```diff
@@ -402,6 +402,7 @@ const DomParser = (settings: DomParserSettings = {}, schema = Schema()): DomPars
 
     // Sanitize the content
     purify.sanitize(body, getPurifyConfig(defaultedSettings, mimeType));
+    purify.removed = [];
```
lnewson, line 405: This isn't performance related, but just makes sure we don't end up with a memory leak since DOMPurify retains the elements that are removed when sanitizing.

patches/dompurify+2.3.4.patch
```diff
@@ -1,5 +1,5 @@
 diff --git a/node_modules/dompurify/dist/purify.es.js b/node_modules/dompurify/dist/purify.es.js
-index 899e471..5844faf 100644
+index 899e471..414c5d8 100644
```
lnewson, line 2: This change is just about reducing the amount of remove/set attribute operations to only remove/set the attribute if needed. It also saves the attributes getting reordered which is something @spocke had mentioned as a concern in a previous Jira.

TheSpyder (approved):
Moving towards more low-level code makes me wonder - and I know this is _completely_ off topic - if ReScript would be a good choice for this area of code.

Writing safe code and emitting unsafe code that's guaranteed to be correct is something ReScript is really good at.

ltrouton (approved):

lnewson:
> Writing safe code and emitting unsafe code that's guaranteed to be correct is something ReScript is really good at.

In the Obj.has case it wouldn't help because it's just the difference between the JS implementation and not anything related to the types. Reason being `in` will return true for things like `toString` whereas `hasOwnProperty` would return false. That's totally expected and likely where the time saving is coming in here.

TheSpyder:
I meant in general not just for the specific thing you'd done here. Like I said, it was off topic :)

PR #7554, opened 2022-01-25
TINY-8400: Removed a number of deprecated options

lnewson:
Related Ticket: TINY-8400

Description of Changes:
* Removed the deprecated `autoresize_on_init` option from the `autoresize` plugin
* Removed the deprecated `padd_empty_with_br` core option
* Removed the deprecated `content_editable_state` core option

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


TheSpyder (approved):

ltrouton (approved):

modules/tinymce/src/plugins/autoresize/main/ts/api/Options.ts
```diff
@@ -26,24 +26,17 @@ const register = (editor: Editor): void => {
     processor: 'number',
     default: 50
   });
-
-  registerOption('autoresize_on_init', {
```
spocke, line 30: Hmm, didn't know that we are going to remove this. There is no alternative for this at the moment. Other compeditors have a DIV mode that would act like this where the editor is the same size as the content. But I guess if it has been deprecated for a while then no one is using the editor like that?

spocke (approved):

modules/tinymce/src/plugins/autoresize/main/ts/api/Options.ts
lnewson: Not sure what you mean sorry Spocke, as this currently doesn't do that... due to the event handling we have the editor always resizes on init atm so this option in effect does nothing and that's why it's being removed.

spocke: Hmm, it seems to still resize to the content based on the events. However not sure if this will work as it did before since the wait thing was to wait for style sheets to load and at the time you couldn't add a load event to the stylesheets on webkit and get it to fire. But that has been resolved since.

spocke: Yes, I see now that other changes have been made to fire events when the content css files are loaded. So never mind. I tested it with a slow loading css file and seems to work fine.

PR #7555, opened 2022-01-25
TINY-4627: Replace the custom parser with the native dom parser

lnewson:
Related Ticket: TINY-4627

Description of Changes:

This is just a cumulation of all the other work already done. The only change is the very last commit, which was to just add the changelog entry.

Note: This doesn't remove the old `SaxParser` API yet, instead that'll be done in a follow up PR.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/tinymce/src/core/main/ts/api/html/DomParser.ts
```diff
@@ -79,192 +83,328 @@ interface DomParser {
...
+
+  const parseAndSanitizeWithContext = (html: string, rootName: string, format: string = 'html') => {
+    const mimeType = format === 'xhtml' ? 'application/xhtml+xml' : 'text/html';
+    // Determine the root element to wrap the HTML in when parsing. If we're dealing with a
+    // special element then we need to wrap it so the internal content is handled appropriately.
+    const isSpecialRoot = Obj.has(schema.getSpecialElements(), rootName.toLowerCase());
+    const content = isSpecialRoot ? `<${rootName}>${html}</${rootName}>` : html;
+    // If parsing XHTML then the content must contain the xmlns declaration, see https://www.w3.org/TR/xhtml1/normative.html#strict
+    const wrappedHtml = format === 'xhtml' ? `<html xmlns="http://www.w3.org/1999/xhtml"><head></head><body>${content}</body></html>` : `<body>${content}</body>`;
+    const body = parser.parseFromString(wrappedHtml, mimeType).body;
```
TheSpyder, line 401: LGTM has now picked up on this as [interpreting HTML without escaping meta characters](https://lgtm.com/projects/g/tinymce/tinymce/snapshot/c514d897721262a76c4c391c0825b3dd1b758e38/files/modules/tinymce/src/core/main/ts/api/html/DomParser.ts?sort=name&dir=ASC&mode=heatmap#x40549e7e04818b5e:1). Is this likely to pose a security concern?

modules/tinymce/src/core/main/ts/api/html/Schema.ts
```diff
@@ -67,7 +67,6 @@ export interface SchemaRegExpMap { [name: string]: RegExp }
 
 interface Schema {
   children: Record<string, SchemaMap>;
-  elements: Record<string, SchemaElement>;
```
TheSpyder, line 70: Oh dear. I forgot about this in the initial discussions around schema changes, but RTC relies on updating this record (we make sure this happens before it's used). cc @spocke we will need to have a discussion about how to deal with this change properly - maybe it means putting our alt-model schema patches into core inside one of the "if rtc" checks.

TheSpyder (approved):

modules/tinymce/src/core/main/ts/api/html/Schema.ts
lnewson: Tbh I think we can revert this if needed. I'd need to double check why Neil removed it in the first place though (I think it was just since it was more of an internal attribute, but if that's the case `children` should also be removed)

ltrouton (approved):

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
lnewson: No, as its purpose is to literally interpret HTML as is... this is just a false-positive report from LGTM.

modules/tinymce/src/core/main/ts/api/html/Schema.ts
TheSpyder: I think it was because `setValidElements` does `elements = {}`, meaning the exported reference is no longer used. Aside from making `schema.elements` out of date it's also a memory leak.

lnewson: Ahh right, that does ring a bell now that you mention it 👍

spocke: Yes, putting it in a isRtc branch seems like the best option here. Then we don't need to expose it anymore. So who does that can that be added to this PR? I think it's just copy paste of the raw code from the rockfest repo to the schema. Just put it in a function like mutateRtcElements or something and run that in RTC mode.

lnewson: I'd prefer to not do that, as I already really dislike having a heap of branching in core. So instead I'd like to find a way to expose additional elements to be added.

spocke: Hmm, then it will be more complex since it needs to expose reading it and writing to it so that can't easily be done without putting the reference out as it was before. I'm not sure I will have time to do any extensive work on this on the RTC side so rollback the change or mutate internally as Andy suggested is the two options I have in order to finish this within a week.

lnewson: I can think of 2 options that avoid the branching:
1. Revert so `elements` is exposed and then remove all existing properties instead of doing `elements = {}` (simplest so probably the better given time constraints)
2. Add a new `addValidElement(name, attrSpec?)` API. This is risky doing it quickly and could be confusing without a `removeValidElement` API, but it would work for what RTC needs I believe.

spocke: I think this is the only place we mutate the elements record so 2 might work but as you say that is more work and not scheduled work.
https://github.com/tinymce/rockfest/blob/master/modules/rtc-plugin/src/main/re/model/ImportTinymceSchema.res#L21

I think doing 1 now is better and we can figure out how to do this proper later. Children should also really go though a api at some point.

lnewson: Sounds good 👍  I'll get that done first thing on Thursday then.

lnewson: Have logged https://github.com/tinymce/tinymce/pull/7562 to do option 1.

modules/tinymce/src/core/test/ts/browser/annotate/AnnotationChangedTest.ts
```diff
@@ -295,14 +295,14 @@ describe('browser.tinymce.core.annotate.AnnotationChangedTest', () => {
     annotate(editor, 'beta', 'id-three', { something: 'comment-3' });
 
     // Content should not include the view only active state attributes
-    TinyAssertions.assertContent(editor, [
-      '<p>',
-      '<span class="mce-annotation" data-mce-annotation-uid="id-one" data-mce-annotation="alpha" data-test-anything="comment-1">one ',
-      '<span class="mce-annotation" data-mce-annotation-uid="id-two" data-mce-annotation="beta" data-test-something="comment-2">two ',
-      '<span class="mce-annotation" data-mce-annotation-uid="id-three" data-mce-annotation="beta" data-test-something="comment-3">three</span></span></span>',
+    assertHtmlContent(editor, [
```
spocke, line 298: Why did this need to be changed the output should be deterministic when you call editor.getContent()?

TheSpyder: This doesn't look like a content change, to me, it's just swapping from array `join` to string append

TheSpyder:
CI failure is the flaky headless tests. Admin merging.

modules/tinymce/src/core/test/ts/browser/annotate/AnnotationChangedTest.ts
spocke: It's switching assert function as well not just the content.

PR #7565, opened 2022-01-26
TINY-8288: Fixed the theme UI rendering too early

lnewson:
Related Ticket: TINY-8288

Description of Changes:
- Fixed a number of UI components being rendered a little too early in the silver theme (e.g before the `renderUi` call)
- Cleaned up/broke the render logic into smaller parts
- Removed the hidden `getUi` API on the silver theme, as it exposes Alloy directly and as such also exposes things like Sugar so it really should never have been on a public API
- Added `readonly` to the interfaces in `Render.ts`

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/tinymce/CHANGELOG.md
```diff
@@ -66,6 +66,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The aria labels for the color picker dialog were not translated #TINY-8381
 - The `editor.annotator.remove` did not keep selection when removing the annotation #TINY-8195
 - Dialog labels and other text-based UI properties did not escape HTML markup #TINY-7524
+- The `silver` theme UI was incorrectly rendered before plugins had initialised #TINY-8288
```
spocke, line 69: Not sure it matter should this be `initialized`?

modules/tinymce/src/themes/silver/main/ts/Render.ts
```diff
@@ -389,12 +366,26 @@ const setup = (editor: Editor): RenderInfo => {
...
+  const setupShortcutsAndCommands = (outerContainer: AlloyComponent): void => {
+    editor.addShortcut('alt+F9', 'focus menubar', () => {
+      OuterContainer.focusMenubar(outerContainer);
+    });
+
+    editor.addShortcut('alt+F10', 'focus toolbar', () => {
+      OuterContainer.focusToolbar(outerContainer);
+    });
+
+    editor.addCommand('ToggleToolbarDrawer', () => {
```
spocke, line 378: Should we make another ticket to move the commands to `api/Commands.ts` there is also commands in the ColorSwatch.ts file would be nice to have one spot where they are defined like Options since they are kind of public api.

spocke (approved):
This one was a bit hard to review since you moved around code and it's just one big commit. So I'm just going to assume that this is working correctly. Is there some way you could add a test to this or will that be to hard to get the timing right. Some test plugin that looks for the UI or something would that work? I couldn't find a test that checks that the resize handle is disabled when the autoresize plugin is enabled should that be added. Seems to work now but it might regress if the timing is off again if something is non lazy in some refactor or such.

ltrouton (approved):

modules/tinymce/CHANGELOG.md
lnewson: Yeah it should be... I always seem to forget the `s` -> `z` for US english 😢

modules/tinymce/src/themes/silver/main/ts/Render.ts
lnewson: Will log a Jira for the first part and as for the rest, this was a timebox task so I'd largely skipped looking at tests because of that. However, since we have an extension now I'll go ahead and do them.

modules/tinymce/CHANGELOG.md
lnewson: Fixed

modules/tinymce/src/themes/silver/main/ts/Render.ts
lnewson: Have added some simple regression tests that should cover this and have logged the commands thing as TINY-8437. Let me know if you want to see anything more added.

PR #7567, opened 2022-01-26
TINY-8308: Prevent removing formats if matching selected element

HAFRMO:
Related Ticket:

Description of Changes:
Do not remove format on parent elements if current element matches target.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/tinymce/src/core/main/ts/fmt/RemoveFormat.ts
```diff
@@ -484,8 +484,9 @@ const remove = (ed: Editor, name: string, vars?: FormatVars, node?: Node | Range
       const removed = removeNodeFormat(node);
 
       // TINY-6567/TINY-7393: Include the parent if using an expanded selector format and no match was found for the current node
+      const currentNodeMatches = Arr.exists(formatList, (f) => matchName(dom, node, f));
```
lnewson, line 487: I know this is what I sent you via DMs, but this can probably be optimized a little more.
```suggestion
      const currentNodeMatches = removed || Arr.exists(formatList, (f) => matchName(dom, node, f));
```

modules/tinymce/src/core/main/ts/fmt/RemoveFormat.ts
```diff
@@ -484,8 +484,9 @@ const remove = (ed: Editor, name: string, vars?: FormatVars, node?: Node | Range
...
       const parentNode = node.parentNode;
-      if (!removed && Type.isNonNullable(parentNode) && FormatUtils.shouldExpandToSelector(format)) {
+      if (!removed && !currentNodeMatches && Type.isNonNullable(parentNode) && FormatUtils.shouldExpandToSelector(format)) {
```
lnewson, line 489:
```suggestion
      if (!currentNodeMatches && Type.isNonNullable(parentNode) && FormatUtils.shouldExpandToSelector(format)) {
```

modules/tinymce/CHANGELOG.md
```diff
@@ -66,6 +66,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The aria labels for the color picker dialog were not translated #TINY-8381
 - The `editor.annotator.remove` did not keep selection when removing the annotation #TINY-8195
 - Dialog labels and other text-based UI properties did not escape HTML markup #TINY-7524
+- Alignment would sometimes be removed on parent elements when changing alignment on images #TINY-8308
```
lnewson, line 69:
```suggestion
- Alignment would sometimes be removed on parent elements when changing alignment on certain inline nodes, such as images #TINY-8308 
```

lnewson (approved):

spocke (approved):
I think this might have been there on purpose since if you select an image and the paragraph is right aligned and you press the `align right` button to untoggle that alignment nothing happens with this change. I don't think it's good UX to have active buttons that don't do anything. But the current behavior is very confusing as well so this is more correct.
But we might need to have a follow up ticket where the UI doesn't show the parents alignment if the image is selected it should probably only show the closest one not all parent ones. It's a bit confusing where some times the UI will show multiple alignment buttons enabled at the same time if say the image is left align and the paragraph is right aligned. I see some editors have different alignment for blocks than for images so the image has it's own alignment options that makes more sense since we are really trying to show the state of two different things here.

modules/tinymce/src/core/main/ts/fmt/RemoveFormat.ts
HAFRMO: I really should have realized that improvement.

lnewson:
> I think this might have been there on purpose since if you select an image and the paragraph is right aligned and you press the `align right` button to untoggle that alignment nothing happens with this change. I don't think it's good UX to have active buttons that don't do anything. But the current behavior is very confusing as well so this is more correct. But we might need to have a follow up ticket where the UI doesn't show the parents alignment if the image is selected it should probably only show the closest one not all parent ones. It's a bit confusing where some times the UI will show multiple alignment buttons enabled at the same time if say the image is left align and the paragraph is right aligned. I see some editors have different alignment for blocks than for images so the image has it's own alignment options that makes more sense since we are really trying to show the state of two different things here.

Worth noting this was actually a regression introduced in 5.9.0. Before that it had already worked like this and this was how a customer expected it to continue working 🤷 . I agree the UI showing both selected seems odd though and that is likely a pre-existing bug so we should log a follow up for that.

lnewson:
@HAFRMO I've merged this, but please make sure to log the Jira for the issue Spocke noted.

PR #7571, opened 2022-01-27
TINY-8351: RTC doesn't fire all relevant get/set content events

vpyshnenko:
Related Ticket: TINY-8351

Description of Changes:
* The main goal of this PR is to unify the behavior about sending events on `getContent`, `setContent`, and `insertContent` calls by wrapping these methods with `preProcess(Get/Set)Content` and `postProcess(Get/Set)Content` functions.
* Having this logic in the one place should cover internal `tinymce` methods implementation as well as correspondent RTC's one.
* Move out any fire pre/post events logic from RTC to the `tinymce` core.

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [x] Milestone set


TheSpyder (approved):

lnewson (changes requested):
This isn't what was mentioned in the Jira and now adds more duplication which isn't what we want. We need to change where the RTC integration hooks in, not duplicate the logic in here.

So to be clear, the `SetContent.ts` file should contain the event processing logic and then it should do the RTC adapter delegation from there.

TheSpyder (changes requested):
I was wondering about that, but I assumed it been discussed while I was away.

lnewson:
> I was wondering about that, but I assumed it been discussed while I was away.

As far as I'm aware nothing extra was discussed... I suspect the `insertContent` one could be difficult, in which case this is probably going to be okay if we run out of time. Ideally I'd like to see us avoid that for everything though and only do the current approach should we be short on time.

modules/tinymce/src/core/main/ts/Rtc.ts
```diff
@@ -221,9 +222,14 @@ const makeRtcAdaptor = (rtcEditor: RtcRuntimeApi): RtcAdaptor => {
...
-      getContent: (args, _format) => editor.getContent(args),
+      getContent: (args) => editor.getContent(args),
       setContent: (content, args) => editor.setContent(content, args),
-      insertContent: (content, _details) => editor.insertContent(content),
+      insertContent: (content, details) => {
+        return preProcessSetContent(tinymceEditor, { content, format: 'html', set: false, selection: true, paste: details.paste }).each((args) => {
+          editor.insertContent(content);
+          postProcessSetContent(tinymceEditor, content, args);
+        });
+      },
```
TheSpyder, line 232: I like the changes you've done with set and get content, please do the same with insert content

- [ ] move insert content processing out of `Rtc.ts`

TheSpyder (approved):

modules/tinymce/src/core/main/ts/content/ContentTypes.ts
```diff
@@ -27,6 +27,11 @@ export interface SetContentArgs {
   no_selection?: boolean;
 }
 
+export interface SetContentResult {
+  content: Content;
+  html: string;
```
lnewson, line 32:
```suggestion
  readonly content: Content;
  readonly html: string;
```

As this appears to be an internal only type we should make sure the properties are readonly as well.

modules/tinymce/src/core/main/ts/content/SetContent.ts
```diff
@@ -6,8 +6,31 @@
...
+
+export const setContent = (editor: Editor, content: Content, args: Partial<SetContentArgs> = {}): Content => {
+  const defaultedArgs = setupArgs(args, content);
+
+  return preProcessSetContent(editor, defaultedArgs).map((updatedArgs) => {
+    // Don't use the content from the args for tree, as it'll be an empty string
+    const updatedContent = isTreeNode(content) ? content : updatedArgs.content;
+    const { content: _content, html } = Rtc.setContent(editor, updatedContent, updatedArgs);
+    postProcessSetContent(editor, html, updatedArgs);
+    return _content;
```
lnewson, line 34: Why is this called `_content`? That naming scheme is only meant to be used for unused attributes, but it's definitely used here. Maybe instead of destructuring above, do something like this?
```
    const result = Rtc.setContent(editor, updatedContent, updatedArgs);
    postProcessSetContent(editor, result.html, updatedArgs);
    return result.content;
```

modules/tinymce/src/core/main/ts/content/GetContentImpl.ts
```diff
@@ -15,47 +15,36 @@ import Tools from '../api/util/Tools';
...
+  } else if (args.format === 'tree') {
+    content = editor.serializer.serialize(body, args);
+  } else {
+    content = trimEmptyContents(editor, editor.serializer.serialize(body, args));
+  }
+
+  // Trim if not using a whitespace preserve format/element
+  const shouldTrim = args.format !== 'text' && !isWsPreserveElement(SugarElement.fromDom(body));
+  const trimmedContent = shouldTrim && Type.isString(content) ? Tools.trim(content) : content;
+  return trimmedContent;
```
lnewson, line 43:
```suggestion
  return shouldTrim && Type.isString(content) ? Tools.trim(content) : content;
```

modules/tinymce/src/core/main/ts/content/SetContent.ts
vpyshnenko: Initially, I wanted to avoid using the additional variable name and confusing associations with the` Result` type. But destruction to {content, html} ends up to strange error for the previous line in code :
`error TS2448: Block-scoped variable 'content' used before its declaration.`
 even though the `content` argument looks visible in scope.

lnewson: It's possibly related to the fact that `content` is already in scope. Either way, destructuring isn't really getting any benefit here and I personally don't think it's confusing with the `Result` type. However, if you're worried about that then we should just come up with another name instead... maybe `contentResult` or `setContentResult` to match the `SetContentResult` type?

vpyshnenko: I've reverted to the initially suggested variant without destructuring

modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
```diff
@@ -221,125 +220,116 @@ export const insertHtmlAtCaret = (editor: Editor, value: string, details: Insert
   }, editor.schema);
   const bookmarkHtml = '<span id="mce_marker" data-mce-type="bookmark">&#xFEFF;</span>';
 
-  // Run beforeSetContent handlers on the HTML to be inserted
-  preProcessSetContent(editor, { content: value, format: 'html', set: false, selection: true, paste: details.paste }).each((args) => {
```
lnewson, line 225: There's a few cleanup steps that are run before here, should they also be included in `InsertContent.ts`? (mainly just looking at the whitespace cleanup)

modules/tinymce/src/core/main/ts/content/InsertContent.ts
```diff
@@ -41,9 +42,13 @@ const processValue = (value: string | DetailsWithContent): ProcessedValue => {
 };
 
 const insertAtCaret = (editor: Editor, value: string | DetailsWithContent): void => {
-  const result = processValue(value);
+  const { content, details } = processValue(value);
 
-  Rtc.insertContent(editor, result.content, result.details);
+  preProcessSetContent(editor, { content, format: 'html', set: false, selection: true, paste: details.paste }).each((args) => {
+    Rtc.insertContent(editor, content, details);
+    postProcessSetContent(editor, content, args);
```
lnewson, line 49: This is incorrect... the `Rtc.insertContent` function needs to return the processed HTML and pass that to `postProcessSetContent` as that's what was passed previously here, not the input HTML.

lnewson (changes requested):
Looking good and almost there, but the content being fired in the `SetContent` event when inserting HTML is now incorrect/different. So we need that fixed before we can merge this sorry.

modules/tinymce/src/core/main/ts/content/InsertContent.ts
```diff
@@ -41,9 +42,13 @@ const processValue = (value: string | DetailsWithContent): ProcessedValue => {
 };
 
 const insertAtCaret = (editor: Editor, value: string | DetailsWithContent): void => {
-  const result = processValue(value);
+  const { content, details } = processValue(value);
 
-  Rtc.insertContent(editor, result.content, result.details);
+  preProcessSetContent(editor, { content, format: 'html', set: false, selection: true, paste: details.paste }).each((args) => {
+    Rtc.insertContent(editor, content, details);
```
lnewson, line 48: This line is also incorrect, as we should be passing the pre-processed HTML, not the input HTML.
```suggestion
    Rtc.insertContent(editor, args.content, details);
```

modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
```diff
@@ -221,125 +220,116 @@ export const insertHtmlAtCaret = (editor: Editor, value: string, details: Insert
...
 
-    // Parse the fragment within the context of the parent node
-    const parserArgs: ParserArgs = { context: parentNode.nodeName.toLowerCase(), data: details.data, insert: true };
-    const fragment = parser.parse(value, parserArgs);
-
-    // Custom handling of lists
-    if (details.paste === true && InsertList.isListFragment(editor.schema, fragment) && InsertList.isParentBlockLi(dom, parentNode)) {
-      rng = InsertList.insertAtCaret(serializer, dom, selection.getRng(), fragment);
-      selection.setRng(rng);
-      editor.fire('SetContent', args);
```
TheSpyder, line 264: Was this here incorrectly? Or should it be replaced with firing an event somewhere else?

TheSpyder (approved):

modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
lnewson: It was moved to the `InsertContent.ts` file from what I saw, as it's handled by the `postProcessSetContent` call now I believe?

modules/tinymce/src/core/main/ts/content/InsertContent.ts
TheSpyder: Ah oops. The plugin has never done that.

For set content RTC just re-serializes the HTML model returned by the TinyMCE parser, but that's still worth doing the same here rather than just using the input string. The parser will address things like security issues.

modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
TheSpyder: Ah I didn't realise it was fired elsewhere already. It just seemed weird that the `fire` was removed without another added elsewhere in the PR.

modules/tinymce/src/core/main/ts/content/InsertContent.ts
```diff
@@ -40,10 +42,25 @@ const processValue = (value: string | DetailsWithContent): ProcessedValue => {
...
+  }
+};
+
 const insertAtCaret = (editor: Editor, value: string | DetailsWithContent): void => {
-  const result = processValue(value);
+  const { content, details } = processValue(value);
 
-  Rtc.insertContent(editor, result.content, result.details);
+  preProcessSetContent(editor, { content: trimOrPad(editor, content), format: 'html', set: false, selection: true, paste: details.paste }).each((args) => {
+    postProcessSetContent(editor, Rtc.insertContent(editor, args.content, details), args);
```
lnewson, line 61:
```suggestion
    const insertedContent = Rtc.insertContent(editor, args.content, details);
    postProcessSetContent(editor, insertedContent, args);
```
Maybe? It just looked a little overloaded 🤷

lnewson (approved):

modules/tinymce/src/core/main/ts/content/InsertContent.ts
spocke: Yeah, I just passed out the serialized input ast to get a value to send to the events setContent event I guess you should serialize the alt model fragment using the alt model serializer to get the lists and elements flattning like the actual inserted value. However the whole purpose from what I can see of the setContent event is to do things to the dom when contents gets inserted.

Not sure what the real need would be exactly to get the html being inserted after it has been inserted you can't really alter it at that point that is what the `beforeSetContent` event is for but all this was added before there where proper node filters etc. The events here makes less sense in the alt model. That is why I was not firing all of them.

If you look at every instance we use the `setContent` event it's not even caring about the event data it just does dom mutation or dom reading after some insert is made. So maybe we should deprecate that `content` property from the `setContent` event to me it serves little purpose. Also the serialization process is currently a bit expensive in the alt model since we unflatten marks, unflatten lists etc.

spocke (approved):

modules/tinymce/src/core/main/ts/content/InsertContent.ts
lnewson: No objections to deprecating it here. For RTC you could likely just return an empty string in that case.

spocke: I added a ticket for that #TINY-8457 since those changes would hold up this one.

spocke (approved):

PR #7572, opened 2022-01-27
TINY-7829: Change commands behaviour

spocke:
Related Ticket:

Might be easier to review this commit by commit since it ended up being more changes than just removing the native fallback. This part of the code has been a mess for a long long time it was over due to make some cleanup of things that where parts of tinymce 3.x or even earlier.

Description of Changes:
* Changes the command behavior to all be registered commands no fallback to browsers native commands.
* Separates the commands into separate modules.
* Removed old 3.x commands that are no longer documented or used.
* Did various cleanup of the code to make it more type safe and less js ninja hacky
* Added tests for the editorCommands API most of it before I started the refactor

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/tinymce/CHANGELOG.md
```diff
@@ -59,6 +59,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The `media` plugin no longer treats `iframe`, `video`, `audio` or `object` elements as "special" and will validate the contents against the schema #TINY-8382
 - Renamed the `getShortEndedElements` Schema API to `getVoidElements` #TINY-8344
 - Changed the default statusbar element path delimiter from `»` to `›` #TINY-8372
+- The `editor.editorCommands.queryCommandSupported` will no longer fallback to the browsers native command #TINY-7829
```
TheSpyder, line 62: It's probably best to put the word "function" or "API" in these lines, e.g. `the <x> function will...`

modules/tinymce/CHANGELOG.md
```diff
@@ -101,6 +117,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Removed the `editor.settings` property as it's been replaced by the new Options API #TINY-8236
 - Removed the `shortEnded` and `fixed` properties on `tinymce.html.Node` class #TINY-8205
 - Removed the `mceInsertRawHTML` command #TINY-8214
+- Removed the undocumented `editor.editorCommands.hasCustomCommand` API #TINY-7829
```
TheSpyder, line 120: So, funny story, we are about to start using this API in #7569 to automatically disable table features not supported by RTC. It wasn't our preferred option but we ran out of time to do it a nicer way.

- [ ] please keep `hasCustomCommand` for this release - or maybe we should rename it `hasCommand`?

TheSpyder (changes requested):

modules/tinymce/CHANGELOG.md
spocke: Yeah I saw that. It could be replaced with queryCommandSupported since after this change it will all be custom commands. So the hasCustomCommand is now the same function as queryCommandSupported so it made no sense to keep it unless we want to remove the queryCommandSupported and have a hasCommand instead in the editorCommands but then it would mismatch with the editor.queryCommandSupported api that is documented and most likely used by integrators.

TheSpyder: Oh! yes I'm happy to swap the other branch to `queryCommandSupported`.

TheSpyder (approved):

modules/tinymce/CHANGELOG.md
spocke: I have wondered a bit about the word API to me an API is more than just a function it's kind of the whole thing. Like the tinymce editor has a selection API I think some times we use API for single functions what is the rule here exactly. I've used both my self. But it becomes really fuzzy if you have a function that only does one thing like the select element api like editor.selection.select() is that an API or a function?

modules/tinymce/CHANGELOG.md
```diff
@@ -66,6 +66,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- The `editor.queryCommandValue` function will no longer fallback to the browsers native command #TINY-7829
+- The `editor.execCommand` function will no longer fallback to the browsers native command #TINY-7829
+- The `editor.editorCommands.queryCommandSupported` function will now return `false` on removed editors #TINY-7829
+- The `editor.editorCommands.queryCommandState` function will now return `false` on removed editors #TINY-7829
+- The `editor.editorCommands.queryCommandValue` function will now return empty strings on removed editors #TINY-7829
+- The `editor.editorCommands.execCommand` function will now return false on removed editors #TINY-7829
+- The `editor.queryCommandSupported` function will now return `false` on removed editors #TINY-7829
+- The `editor.queryCommandState` function will now return `false` on removed editors #TINY-7829
+- The `editor.queryCommandValue` function will now return empty strings on removed editors #TINY-7829
+- The `editor.execCommand` function will now return false on removed editors #TINY-7829
```
lnewson, line 84: This all seems quite verbose. Could we trim it down to just something like this? We can always expand more in the documentation if we need to, but I just feel this is too verbose.
```
- The Editor commands APIs will no longer fallback to executing the browsers native command functionality #TINY-7829
- The Editor query command APIs will now return `false` or an empty string on removed editors #TINY-7829
```

If you don't do this, then `false` in the last line needs backticks around it.

modules/tinymce/CHANGELOG.md
```diff
@@ -110,6 +126,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Removed the `editor.settings` property as it's been replaced by the new Options API #TINY-8236
 - Removed the `shortEnded` and `fixed` properties on `tinymce.html.Node` class #TINY-8205
 - Removed the `mceInsertRawHTML` command #TINY-8214
+- Removed the undocumented `editor.editorCommands.hasCustomCommand` API #TINY-7829
+- Removed the undocumented `mceResetDesignMode` command #TINY-7829
+- Removed the undocumented `mceRepaint` command #TINY-7829
+- Removed the undocumented `mceBeginUndoLevel` command #TINY-7829
```
lnewson, line 132:
```suggestion
- Removed the undocumented `mceResetDesignMode`, `mceRepaint` and `mceBeginUndoLevel` commands #TINY-7829
```
Again, this is just a little too verbose

modules/tinymce/CHANGELOG.md
```diff
@@ -110,6 +126,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- Removed the `execCommand` handler function from the plugin interface #TINY-7829
+- Removed the `execCommand` handler function from the theme interface #TINY-7829
```
lnewson, line 134:
```suggestion
- Removed the `execCommand` handler function from the plugin and theme interface #TINY-7829
```

modules/tinymce/src/core/main/ts/api/Editor.ts
```diff
@@ -577,7 +578,7 @@ class Editor implements EditorObservable {
    *
    * @method queryCommandValue
    * @param {string} cmd Command to query value from.
-   * @return {Object} Command specific value, for example the current font size.
+   * @return {string} Command value, for example the current font size or an empty string (`""`) if the query command is not found.
```
lnewson, line 581: I wonder if we'd be better returning `null` or `undefined` here instead of an empty string, as a command might actually need to return an empty string here?

modules/tinymce/src/core/main/ts/api/EditorCommands.ts
```diff
@@ -52,81 +39,49 @@ export interface EditorCommandsConstructor {
...
   public constructor(editor: Editor) {
     this.editor = editor;
-
-    this.setupCommands(editor);
   }
 
   /**
-   * Executes the specified command.
+   * Executes a registered command on the current instance. A list of available commands can be found in
+   * the <a href="https://www.tiny.cloud/docs/advanced/editor-command-identifiers/">command identifiers</a> documentation.
```
lnewson, line 50: This docs link is almost certainly going to break for 6.0, as the docs structure is changing. Can we please make sure we have a followup Jira to fix this or change it to something else that's less hardcoded? At the very least we should remove the domain, as otherwise it won't work on staging for example.

- [ ] Replace with something else or at the very least remove the domain in the docs URLs

modules/tinymce/src/core/main/ts/api/EditorCommands.ts
```diff
@@ -52,81 +39,49 @@ export interface EditorCommandsConstructor {
...
    *
    * @method execCommand
-   * @param {String} command Command to execute.
-   * @param {Boolean} ui Optional user interface state.
-   * @param {Object} value Optional value for command.
-   * @param {Object} args Optional extra arguments to the execCommand.
-   * @return {Boolean} true/false if the command was found or not.
+   * @param {String} cmd Command name to execute, for example mceLink or Bold.
+   * @param {Boolean} ui True/false state if a UI (dialog) should be presented or not.
```
lnewson, line 54:
```suggestion
   * @param {Boolean} ui A boolean that specifies if a UI (dialog) should be presented or not.
```

modules/tinymce/src/core/main/ts/api/EditorCommands.ts
```diff
@@ -52,81 +39,49 @@ export interface EditorCommandsConstructor {
...
-    try {
-      state = self.editor.getDoc().execCommand(command, ui, value);
-    } catch (ex) {
-      // Ignore old IE errors
-    }
-
-    if (state) {
-      self.editor.fire('ExecCommand', { command, ui, value });
+    const func = this.commands.exec[lowerCaseCommand];
+    if (func) {
```
lnewson, line 82: This might be out of scope, so feel free to ignore, but maybe add a `Type.isFunction` guard here?

modules/tinymce/src/core/main/ts/api/EditorCommands.ts
```diff
@@ -141,52 +96,38 @@ class EditorCommands {
...
     }
 
-    // Browser commands
-    try {
-      return this.editor.getDoc().queryCommandState(command);
-    } catch (ex) {
-      // Fails sometimes see bug: 1896577
+    const lowerCaseCommand = command.toLowerCase();
+    const func = this.commands.state[lowerCaseCommand];
+    if (func) {
```
lnewson, line 105: Also here if you do end up adding the `Type.isFunction` guard above.

modules/tinymce/src/core/main/ts/api/EditorCommands.ts
```diff
@@ -52,81 +39,49 @@ export interface EditorCommandsConstructor {
...
    *
    * @method execCommand
-   * @param {String} command Command to execute.
-   * @param {Boolean} ui Optional user interface state.
-   * @param {Object} value Optional value for command.
-   * @param {Object} args Optional extra arguments to the execCommand.
-   * @return {Boolean} true/false if the command was found or not.
+   * @param {String} cmd Command name to execute, for example mceLink or Bold.
+   * @param {Boolean} ui True/false state if a UI (dialog) should be presented or not.
+   * @param {mixed} value Optional command value, this can be anything.
```
lnewson, line 55: `mixed` isn't a valid type... I'd suggest using something like what's used in `EditorManager` instead or just saying `any`.
```suggestion
   * @param {Object/Array/String/Number/Boolean} value Optional command value, this can be anything.
```

modules/tinymce/src/core/main/ts/api/EditorCommands.ts
```diff
@@ -221,371 +162,42 @@ class EditorCommands {
...
+    if (this.commands.exec[lowerCaseCommand]) {
+      return true;
     }
 
     return false;
   }
 
+  /**
+   * Adds a custom query state command to the editor, you can also override existing commands with this method.
+   * The command that you add can be executed with queryCommandState function.
```
lnewson, line 179:
```suggestion
   * Adds a custom query state command to the editor. This function can also be used to override existing commands.
   * The added command can be executed with the <code>queryCommandState<code> function.
```
`You` and other pronouns is something that should almost always be avoided in docs: https://www.tiny.cloud/docs/configure/contributing-docs/#pronouns

modules/tinymce/src/core/main/ts/api/EditorCommands.ts
```diff
@@ -221,371 +162,42 @@ class EditorCommands {
...
+   */
   public addQueryStateHandler(command: string, callback: () => boolean, scope?: any) {
-    command = command.toLowerCase();
-    this.commands.state[command] = () => callback.call(scope || this.editor);
+    this.commands.state[command.toLowerCase()] = () => callback.call(scope ?? this.editor);
   }
 
+  /**
+   * Adds a custom query value command to the editor, you can also override existing commands with this method.
+   * The command that you add can be executed with queryCommandValue function.
```
lnewson, line 192:
```suggestion
   * Adds a custom query value command to the editor. This function can also be used to override existing commands.
   * The added command can be executed with the <code>queryCommandValue<code> function.
```

modules/tinymce/src/core/main/ts/api/EditorCommands.ts
```diff
@@ -221,371 +162,42 @@ class EditorCommands {
...
     return false;
   }
 
+  /**
+   * Adds a custom query state command to the editor, you can also override existing commands with this method.
+   * The command that you add can be executed with queryCommandState function.
+   *
+   * @method addQueryStateHandler
+   * @param {String} name Command name to add/override.
+   * @param {addQueryStateHandlerCallback} callback Function to execute when the command state retrieval occurs.
```
lnewson, line 183:
```suggestion
   * @param {Function} callback Function to execute when the command state retrieval occurs.
```

modules/tinymce/src/core/main/ts/api/EditorCommands.ts
```diff
@@ -221,371 +162,42 @@ class EditorCommands {
...
+    this.commands.state[command.toLowerCase()] = () => callback.call(scope ?? this.editor);
   }
 
+  /**
+   * Adds a custom query value command to the editor, you can also override existing commands with this method.
+   * The command that you add can be executed with queryCommandValue function.
+   *
+   * @method addQueryValueHandler
+   * @param {String} name Command name to add/override.
+   * @param {addQueryValueHandlerCallback} callback Function to execute when the command value retrieval occurs.
```
lnewson, line 196:
```suggestion
   * @param {Function} callback Function to execute when the command value retrieval occurs.
```

modules/tinymce/src/core/main/ts/api/commands/AlignCommands.ts
```diff
@@ -0,0 +1,59 @@
...
+ * Copyright (c) Tiny Technologies, Inc. All rights reserved.
+ * Licensed under the LGPL or a commercial license.
+ * For LGPL see License.txt in the project root for license information.
+ * For commercial licenses see https://www.tiny.cloud/
+ */
+
+import { Arr, Type } from '@ephox/katamari';
+
+import { FormatVars } from '../../fmt/FormatTypes';
+import type Editor from '../Editor';
```
lnewson, line 11:
```suggestion
import Editor from '../Editor';
```
We probably shouldn't be using type imports, as the types are published so this would make it incompatible with older versions of TypeScript so for now given it's not needed we're probably safer doing this (granted type only imports been around for a while now so it might be fine). If we did want this, we probably need to have a quick discussion and document what version of TS our types support.

modules/tinymce/src/core/main/ts/api/commands/AlignCommands.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  editor.editorCommands.addCommands({
+    JustifyLeft: alignStates('alignleft'),
+    JustifyCenter: alignStates('aligncenter'),
+    JustifyRight: alignStates('alignright'),
+    JustifyFull: alignStates('alignjustify')
+  }, 'state');
+};
+
+export const registerCommands = (editor: Editor) => {
```
lnewson, line 56:
```suggestion
export const registerCommands = (editor: Editor): void => {
```
We really need to fix all the issues in the TinyMCE code base so we can re-enable the `@typescript-eslint/explicit-module-boundary-types` rule 😞

modules/tinymce/src/core/main/ts/api/commands/ClipboardCommands.ts
```diff
@@ -0,0 +1,45 @@
+/**
+ * Copyright (c) Tiny Technologies, Inc. All rights reserved.
+ * Licensed under the LGPL or a commercial license.
+ * For LGPL see License.txt in the project root for license information.
+ * For commercial licenses see https://www.tiny.cloud/
+ */
+
+import type Editor from '../Editor';
```
lnewson, line 8:
```suggestion
import Editor from '../Editor';
```

modules/tinymce/src/core/main/ts/api/commands/ClipboardCommands.ts
```diff
@@ -0,0 +1,45 @@
...
+import Env from '../Env';
+
+export const registerCommands = (editor: Editor) => {
```
lnewson, line 11:
```suggestion
export const registerCommands = (editor: Editor): void => {
```

modules/tinymce/src/core/main/ts/api/commands/Commands.ts
```diff
@@ -0,0 +1,61 @@
+/**
+ * Copyright (c) Tiny Technologies, Inc. All rights reserved.
+ * Licensed under the LGPL or a commercial license.
+ * For LGPL see License.txt in the project root for license information.
+ * For commercial licenses see https://www.tiny.cloud/
+ */
+
+import * as EditorFocus from '../../focus/EditorFocus';
+import type Editor from '../Editor';
```
lnewson, line 9:
```suggestion
import Editor from '../Editor';
```

modules/tinymce/src/core/main/ts/api/commands/Commands.ts
```diff
@@ -0,0 +1,61 @@
...
+import * as ContentCommands from './ContentCommands';
+import * as FormatCommands from './FormatCommands';
+import * as HistoryCommands from './HistoryCommands';
+import * as IndentCommands from './IndentCommands';
+import * as LinkCommands from './LinkCommands';
+import * as ListCommands from './ListCommands';
+import * as NewlineCommands from './NewlineCommands';
+import * as SelectionCommands from './SelectionCommands';
+
+const registerSimpleCommands = (editor: Editor) => {
```
lnewson, line 21:
```suggestion
const registerSimpleCommands = (editor: Editor): void => {
```

modules/tinymce/src/core/main/ts/api/commands/ContentCommands.ts
```diff
@@ -0,0 +1,43 @@
+/**
+ * Copyright (c) Tiny Technologies, Inc. All rights reserved.
+ * Licensed under the LGPL or a commercial license.
+ * For LGPL see License.txt in the project root for license information.
+ * For commercial licenses see https://www.tiny.cloud/
+ */
+
+import * as InsertContent from '../../content/InsertContent';
+import type Editor from '../Editor';
```
lnewson, line 9:
```suggestion
import Editor from '../Editor';
```

modules/tinymce/src/core/main/ts/api/commands/ContentCommands.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export const registerCommands = (editor: Editor) => {
```
lnewson, line 11:
```suggestion
export const registerCommands = (editor: Editor): void => {
```

modules/tinymce/src/core/main/ts/api/commands/FormatCommands.ts
```diff
@@ -0,0 +1,85 @@
...
+ * For LGPL see License.txt in the project root for license information.
+ * For commercial licenses see https://www.tiny.cloud/
+ */
+
+import { Type } from '@ephox/katamari';
+
+import * as Font from '../../commands/Font';
+import * as LineHeight from '../../commands/LineHeight';
+import { FormatVars } from '../../fmt/FormatTypes';
+import type Editor from '../Editor';
```
lnewson, line 13:
```suggestion
import Editor from '../Editor';
```

modules/tinymce/src/core/main/ts/api/commands/FormatCommands.ts
```diff
@@ -0,0 +1,85 @@
...
+    'Bold,Italic,Underline,Strikethrough,Superscript,Subscript': (command) => isFormatMatch(command),
+    'mceBlockQuote': () => isFormatMatch('blockquote')
+  }, 'state');
+
+  editor.editorCommands.addQueryValueHandler('FontName', () => Font.fontNameQuery(editor));
+  editor.editorCommands.addQueryValueHandler('FontSize', () => Font.fontSizeQuery(editor));
+  editor.editorCommands.addQueryValueHandler('LineHeight', () => LineHeight.lineHeightQuery(editor));
+};
+
+export const registerCommands = (editor: Editor) => {
```
lnewson, line 82:
```suggestion
export const registerCommands = (editor: Editor): void => {
```

modules/tinymce/src/core/main/ts/api/commands/HistoryCommands.ts
```diff
@@ -0,0 +1,28 @@
+/**
+ * Copyright (c) Tiny Technologies, Inc. All rights reserved.
+ * Licensed under the LGPL or a commercial license.
+ * For LGPL see License.txt in the project root for license information.
+ * For commercial licenses see https://www.tiny.cloud/
+ */
+
+import type Editor from '../Editor';
```
lnewson, line 8:
```suggestion
import Editor from '../Editor';
```

modules/tinymce/src/core/main/ts/api/commands/HistoryCommands.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export const registerCommands = (editor: Editor) => {
+  editor.editorCommands.addCommands({
+    mceAddUndoLevel: () => {
+      editor.undoManager.add();
```
lnewson, line 13: Should we extract `editor.undoManager` out to a separate variable here? It'd minify better than, though granted it likely wouldn't make much difference if we include GZIP compression.

modules/tinymce/src/core/main/ts/api/commands/IndentCommands.ts
```diff
@@ -0,0 +1,25 @@
+/**
+ * Copyright (c) Tiny Technologies, Inc. All rights reserved.
+ * Licensed under the LGPL or a commercial license.
+ * For LGPL see License.txt in the project root for license information.
+ * For commercial licenses see https://www.tiny.cloud/
+ */
+
+import * as IndentOutdent from '../../commands/IndentOutdent';
+import type Editor from '../Editor';
```
lnewson, line 9:
```suggestion
import Editor from '../Editor';
```

modules/tinymce/src/core/main/ts/api/commands/IndentCommands.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export const registerCommands = (editor: Editor) => {
```
lnewson, line 11:
```suggestion
export const registerCommands = (editor: Editor): void => {
```

modules/tinymce/src/core/main/ts/api/commands/IndentCommands.ts
```diff
@@ -0,0 +1,25 @@
...
+    Indent: () => {
+      IndentOutdent.indent(editor);
+    },
+
+    Outdent: () => {
+      IndentOutdent.outdent(editor);
+    },
+  });
+
+  editor.editorCommands.addCommands({
```
lnewson, line 22: Should this be broken up into separate functions for exec and query commands? That's been done everywhere else and it wasn't immediately obvious why this was separate (though it only took reading until the end of the function call to figure it out).

modules/tinymce/src/core/main/ts/api/commands/LinkCommands.ts
```diff
@@ -0,0 +1,50 @@
+/**
+ * Copyright (c) Tiny Technologies, Inc. All rights reserved.
+ * Licensed under the LGPL or a commercial license.
+ * For LGPL see License.txt in the project root for license information.
+ * For commercial licenses see https://www.tiny.cloud/
+ */
+
+import { Type } from '@ephox/katamari';
+
+import type Editor from '../Editor';
```
lnewson, line 10:
```suggestion
import Editor from '../Editor';
```

modules/tinymce/src/core/main/ts/api/commands/LinkCommands.ts
```diff
@@ -0,0 +1,50 @@
...
+
+export const registerCommands = (editor: Editor) => {
```
lnewson, line 12:
```suggestion
export const registerCommands = (editor: Editor): void => {
```

modules/tinymce/src/core/main/ts/api/commands/ListCommands.ts
```diff
@@ -0,0 +1,43 @@
+/**
+ * Copyright (c) Tiny Technologies, Inc. All rights reserved.
+ * Licensed under the LGPL or a commercial license.
+ * For LGPL see License.txt in the project root for license information.
+ * For commercial licenses see https://www.tiny.cloud/
+ */
+
+import type Editor from '../Editor';
```
lnewson, line 8:
```suggestion
import Editor from '../Editor';
```

modules/tinymce/src/core/main/ts/api/commands/ListCommands.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export const registerCommands = (editor: Editor) => {
```
lnewson, line 10:
```suggestion
export const registerCommands = (editor: Editor): void => {
```

modules/tinymce/src/core/main/ts/api/commands/ListCommands.ts
```diff
@@ -0,0 +1,43 @@
...
+        if (/^(H[1-6]|P|ADDRESS|PRE)$/.test(listParent.nodeName)) {
+          const bm = editor.selection.getBookmark();
+          editor.dom.split(listParent, listElm);
+          editor.selection.moveToBookmark(bm);
+        }
+      }
+    }
+  });
+
+  editor.editorCommands.addCommands({
```
lnewson, line 32: Similar to my previous comment, should this be 2 separate functions instead?

modules/tinymce/src/core/main/ts/api/commands/NewlineCommands.ts
```diff
@@ -0,0 +1,22 @@
+/**
+ * Copyright (c) Tiny Technologies, Inc. All rights reserved.
+ * Licensed under the LGPL or a commercial license.
+ * For LGPL see License.txt in the project root for license information.
+ * For commercial licenses see https://www.tiny.cloud/
+ */
+
+import * as InsertBr from '../../newline/InsertBr';
+import * as InsertNewLine from '../../newline/InsertNewLine';
+import type Editor from '../Editor';
```
lnewson, line 10:
```suggestion
import Editor from '../Editor';
```

modules/tinymce/src/core/main/ts/api/commands/NewlineCommands.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export const registerCommands = (editor: Editor) => {
```
lnewson, line 12:
```suggestion
export const registerCommands = (editor: Editor): void => {
```

modules/tinymce/src/core/main/ts/api/commands/SelectionCommands.ts
```diff
@@ -0,0 +1,37 @@
+/**
+ * Copyright (c) Tiny Technologies, Inc. All rights reserved.
+ * Licensed under the LGPL or a commercial license.
+ * For LGPL see License.txt in the project root for license information.
+ * For commercial licenses see https://www.tiny.cloud/
+ */
+
+import * as NodeType from '../../dom/NodeType';
+import type Editor from '../Editor';
```
lnewson, line 9:
```suggestion
import Editor from '../Editor';
```

modules/tinymce/src/core/main/ts/api/commands/SelectionCommands.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export const registerCommands = (editor: Editor) => {
```
lnewson, line 11:
```suggestion
export const registerCommands = (editor: Editor): void => {
```

modules/tinymce/src/core/main/ts/api/commands/SelectionCommands.ts
```diff
@@ -0,0 +1,37 @@
...
+  editor.editorCommands.addCommands({
+    mceSelectNodeDepth: (_command, _ui, value) => {
+      let counter = 0;
+
+      editor.dom.getParent(editor.selection.getNode(), (node) => {
+        if (node.nodeType === 1 && counter++ === value) {
```
lnewson, line 17: Not something to do in this PR, but one day we should clean up all these magic numbers and make them use the relevant helper functions or constants we have already.

modules/tinymce/src/core/main/ts/commands/IndentOutdent.ts
```diff
@@ -77,7 +77,11 @@ const handle = (editor: Editor, command: string): void => {
   });
 };
 
+const indent = (editor: Editor) => handle(editor, 'indent');
+const outdent = (editor: Editor) => handle(editor, 'outdent');
```
lnewson, line 81:
```suggestion
const indent = (editor: Editor): void => handle(editor, 'indent');
const outdent = (editor: Editor): void => handle(editor, 'outdent');
```

lnewson (changes requested):
Changes look good overall Spocke and thanks for doing the cleanup! The reason I'm marking this as needs changes, is one of the Jira requirements hasn't been met. Namely:
```
Implement required additional native commands. See notion for a list.
```
Looking at that list in Notion, I can see a number of these that haven't been implemented.

spocke:
> Implement required additional native commands. See notion for a list.

Not sure what you mean, I added three commands `createLink`, `insertHTML` and `BackColor` the rest if not supported by IE so I doubt anyone is using them or things we don't want to expose according to the notion document. We could implement the ones marked as "probably wont need to implement this" in future releases if they are actually needed and but is a workaround for people who depend on these for the dom model at least to just do a direct call to `editor.getDoc().execCommand`. Adding some of these like increasing/decreasing the font size will be a bit of work and needs to be scheduled as it's own ticket if we feel it's important since I'm not sure how it works in browsers today for different values formats like px, em etc how much does it increase/decrease them is the delta browser dependent etc etc.

I compared these two sets to find things that both browsers support and the above was the things I found:
https://docs.microsoft.com/en-us/previous-versions/windows/internet-explorer/ie-developer/platform-apis/ms533049(v=vs.85)
https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand

modules/tinymce/CHANGELOG.md
spocke: I like the suggestion but should this be `interfaces`?

lnewson: It should be thanks. I'll edit the suggestion.

modules/tinymce/src/core/main/ts/api/commands/NewlineCommands.ts
spocke: Why would we need to specify the void type here it's inferred? If we need to there is a lot of places it needs to be added. It will be quite verbose if we added them to every function for every command?

modules/tinymce/src/core/main/ts/api/commands/SelectionCommands.ts
spocke: Curious why we can't use a type import here? I think all type only imports should be using that since we are not really importing code here and in theory this is a circular reference since Editor imports the command module and the commands import the Editor module. Making it a type only import makes that explicit that no side effects or code are being imported only the type.

modules/tinymce/src/core/main/ts/api/commands/SelectionCommands.ts
spocke: Yes, I think predicates for all these would be better since it can then do things like `is Element` etc on types not sure it can do that by property inspection.

modules/tinymce/src/core/main/ts/api/EditorCommands.ts
spocke: I removed the site link we could improve this later when the new docs site is up. I guess we might need to go though it all I think we have links else were like the UI registry API docs for example.

modules/tinymce/src/core/main/ts/api/Editor.ts
spocke: It will only return the empty string on removed editors or for commands that doesn't exist and both of them could be detected by using the queryCommandSupported so checking if it returned null or not is probably not the right way to see if a command was supported or not. I thought it would be nice to remove the `string | undefined` and `boolean | undefined` types that is there now.
The browsers return an empty string if they are not supported but we are moving away from browser like functionality in this PR so we could switch them to null if we want that type I'm not sure what is the best thing for this edge case. I can switch it I don't mind either way as long as it's not `undefined` as it used to be for removed editors null seems a bit more explicit that we have picked a proper return not just what would be default on a function not returning anything.

modules/tinymce/src/core/main/ts/api/commands/SelectionCommands.ts
lnewson: I explained this here: https://github.com/tinymce/tinymce/pull/7572#discussion_r795337288

modules/tinymce/src/core/main/ts/api/commands/NewlineCommands.ts
lnewson: I explained this here: https://github.com/tinymce/tinymce/pull/7572#discussion_r795337514. The intention of that rule is to make sure you cannot accidentally break a module or public API by making internal changes and has helped out before. As such, that rule requires an explicit return type on exported functions only, which is where I've left the comments. I also don't want us adding more things that would add to the work to get that enabled, as I've already been working slowly to get all issues fixed.

modules/tinymce/src/core/main/ts/api/EditorCommands.ts
lnewson: Looks like Mike has logged TINY-8441 to do this.

modules/tinymce/src/core/main/ts/api/Editor.ts
lnewson: Right I forgot that the native implementation used to do that anyways. In that case I think it's fine, as it's what people would already expect. Anyways, happy to go with whatever you want to do here Spocke 👍

lnewson:
> > Implement required additional native commands. See notion for a list.
> 
> Not sure what you mean, I added three commands `createLink`, `insertHTML` and `BackColor` the rest if not supported by IE so I doubt anyone is using them or things we don't want to expose according to the notion document. We could implement the ones marked as "probably wont need to implement this" in future releases if they are actually needed and but is a workaround for people who depend on these for the dom model at least to just do a direct call to `editor.getDoc().execCommand`. Adding some of these like increasing/decreasing the font size will be a bit of work and needs to be scheduled as it's own ticket if we feel it's important since I'm not sure how it works in browsers today for different values formats like px, em etc how much does it increase/decrease them is the delta browser dependent etc etc.
> 
> I compared these two sets to find things that both browsers support and the above was the things I found: [docs.microsoft.com/en-us/previous-versions/windows/internet-explorer/ie-developer/platform-apis/ms533049(v=vs.85)](https://docs.microsoft.com/en-us/previous-versions/windows/internet-explorer/ie-developer/platform-apis/ms533049(v=vs.85)) [developer.mozilla.org/en-US/docs/Web/API/Document/execCommand](https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand)

I think we're better off discussing this via a call rather than going back and forth here. We have the cross team sync up so lets try and discuss it there.

For reference, I was expecting to see these implemented based on the notes:
- `contentReadOnly`
- `insertImage`
- `insertParagraph`
- `insertText`

modules/tinymce/src/core/main/ts/api/EditorCommands.ts
spocke: Not sure about this one since the type is already a boolean. Maybe I go with just. `Specifies if a UI (dialog) should be presented or not.`

modules/tinymce/src/core/main/ts/api/commands/AlignCommands.ts
spocke: It was added in 3.8 two years ago but there might be legacy projects using tinymce so unsure if we can use this "new" feature. We might also consider adding a step that verifies the types with an older version of typescript so that we don't accidentally use features that might be incompatible.

modules/tinymce/src/core/main/ts/api/commands/NewlineCommands.ts
spocke: But this isn't exported. It just called internally to register things. Or does the rule apply to anything inside the api directory exposed or not?

lnewson: It's a TS lint rule, not a Tiny/Ephox rule so it applies to at the module level.

modules/tinymce/src/core/main/ts/api/commands/HistoryCommands.ts
spocke: We can't since the editor.undoManager gets initialized after the commands gets registered. I don't want to change that order now.

modules/tinymce/src/core/main/ts/commands/IndentOutdent.ts
spocke: I added void return types here but there is probably lots and lots of places if we need to add them to functions that gets called within a block scope without a return statement. Is the idea to always have explicit type annotations that would be very verbose but then again this language is very verbose anyway when it comes to types.

modules/tinymce/src/core/main/ts/api/commands/AlignCommands.ts
spocke: Does that rule mean every function everywhere needs explicit types or can we scope it to the `api` directories only since it seems sad not to ever be able to use type inference.

lnewson (approved):

modules/tinymce/src/core/main/ts/api/EditorCommands.ts
```diff
@@ -221,371 +162,41 @@ class EditorCommands {
    * @return {Boolean} true/false if the command is supported or not.
    */
   public queryCommandSupported(command: string): boolean {
-    command = command.toLowerCase();
-
-    if (this.commands.exec[command]) {
-      return true;
+    if (this.editor.quirks.isHidden() || this.editor.removed) {
+      return false;
     }
```
TheSpyder, line 167: This seems wrong? Whether the editor is hidden or removed seems irrelevant to a boolean check of whether a command string exists on an object.

This breaks our Table Model branch, which needs to switch from `hasCustomCommand` to `queryCommandSupported` before `editor.quirks` is created.

spocke: Oops yes that is wrong it should only check if the editor is removed. But it depends also on how we interpret the function of that should it be is the command registered or is the command available for execution if the editor is removed you can't execute it. But calling commands on removed editors seems like an edge case. But calling it on editors that doesn't have a selection is fairly common. I can make a follow up ticket to this but maybe remove the is removed check so that it is a is the command registered instead.

TheSpyder: I deleted the entire if block on the model PR. If editors that don't have a selection are  problem, perhaps we need further checks here (or better documentation to explain that "supported" commands need further checks to see if they're going to function correctly).

spocke: Yeah we should remove the entire block since we want it to the equal to the hasCsutomCommand and checking for commands on a removed editor is sich an edge case. We need to update the changelog and the code should I still do a separate ticket and or for this or will you just bake it into your branch?

TheSpyder: I was trying to say we already have :)
https://github.com/tinymce/tinymce/pull/7583/commits/aaf1231977f0a3054b3f444b30ce6aaabf5b9071

We forgot to do anything with the changelog though.

PR #7573, opened 2022-01-27
TINY-8371: Replace Powered by Tiny in the status bar with logo

noxuhax:
Related Ticket: TINY-8371

Description of Changes:
* Replaced Powered by Tiny in the status bar with SVG logo

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/tinymce/src/themes/silver/main/ts/ui/statusbar/Statusbar.ts
```diff
@@ -20,7 +20,10 @@ const renderStatusbar = (editor: Editor, providersBackstage: UiFactoryBackstageP
 
   const renderBranding = (): SimpleSpec => {
     const label = I18n.translate([ 'Powered by {0}', 'Tiny' ]);
-    const linkHtml = `<a href="https://www.tiny.cloud/?utm_campaign=editor_referral&amp;utm_medium=poweredby&amp;utm_source=tinymce&amp;utm_content=v5" rel="noopener" target="_blank" tabindex="-1" aria-label="${label}">${label}</a>`;
+    const logo = `<svg width="50" height="16" viewBox="0 0 50 16" xmlns="http://www.w3.org/2000/svg" title="${label}">
```
TheSpyder, line 23: We don't put `<svg` elements in code. We have SVG files that are read into scripts by the build process.

However we're only set up to do this for toolbar icons, we weren't informed there would be a requirement to have an SVG on the status bar for 6.0. I'm not sure how to resolve this. It might have to wait for 6.1, unless we want to put our logo into the icon pack (this seems like a bad idea).

TheSpyder (changes requested):

modules/tinymce/src/themes/silver/main/ts/ui/statusbar/Statusbar.ts
lnewson: This also isn't accessible, so can't be merged unless that is resolved as well. This is something I had said to make sure it was done when this was mentioned in #tinymce-dev, but it looks as though that has been missed.

TheSpyder: Ah, I think that discussion happened right before I went on holiday last year. So I take back my statement about us not being informed.

That didn't result in any tickets to support adding SVGs to the core, though. Which was kind of my point.

lnewson: From what I recall it was never mentioned we'd be using an SVG, it was only stated something general about showing the logo instead and there probably wasn't much thought put into it outside of that. I agree we shouldn't be hardcoding the SVG file in like this though.

Afraithe: Wait, this is our logo, why can't this be hardcoded? The "Powered by Tiny" text is hardcoded? This wont be replaced unless we ourselves make updates?

noxuhax: > `aria-label` attribute added to the anchor. There should be something asserting that exists though, as the SVG won't be focused so the title isn't going to be useful there

`aria-label` has been there before logo update, but I agree adding tests checking for it’s presence would be a good idea. Even though svg is not going to be focused by the user under the normal circumstances, it still going to be present in a11y tree, so I’ve added the `title` attribute.

Not for the sake of arguing, but why are you thinking that inlining svg is a bad idea in that case? This logo is not supposed to be customizable by user, so having it as an external file that would be pulled on each build seems a bit unnecessary. Basically, currently it is as static as any copyright or a class name that is stored in the code.

lostkeys: Why not hardcode it? I feel that is the most pragmatic way. We certainly don't want someone to feel encouraged to swap out an external asset and change our logo so in this case I see it as a feature being inline. And have to do a lot of extra work to just manage a stray file sounds like a lot of extra job and actually more maintenance heavy than simply inline the asset.

spocke: I wonder if the svg should have `role="presentation"` since is the logo really that important for screen readers the parent anchor already have a aria-label then we could remove the extra label on the logo.

Also could we stick this into a separate file like a `Logo.ts` or something. Since this inline svg data bloats up the file also it requires eslint max length to be disabled on the whole file since it's a multi line string and we don't really want to disable that for a whole file like that. Having it in a separate file we can scope that rule to that logo file only. Not sure we have a way to import .svg files directly into code though say a import statement either since it probably needs webpack/rollup/terser modules to handle that and I don't thing we want that kind of imports in the code it adds complexity.

Currently nothing in the actual ts/js code suggest how things is going to be visually presented all visuals are controlled by the skins or the icon packs. This some what a diversion from that philosophy. However it's a gray zone since the logo here doesn't have any colors, it's a logo more like an emoji also it's our logo so no one should make their own tinymce logo in the icon packs we will not have different variants of the logo in out different icons packs. The visual representation color and size in this case is controlled by the skin as it should be so the concerns are properly separated here.

Having it inlined like this doesn't make it harder to remove the branding since we already have a config option to disable it and you can also hide it with css and there is nothing we can do about that.

Also I don't think this is a design decision since it's about but where code and resources lives so more of an editor architecture thing. We are not debating the actual svg data it self, where it's going to be presented or what it looks like visually. So it's not like the feedback on the recent skin PR.

spocke (changes requested):

modules/tinymce/src/themes/silver/main/ts/ui/statusbar/Statusbar.ts
lostkeys: Please take some time to think about the expression _perfect i the enemy of good_". If senior engineers believes continue working on this is money well spent then, as you say, it's on engineering to make the call how it's implemented. But please bring the cost/value into the decision.

TheSpyder: It’s not about perfection. It’s about holding our codebase to a certain standard, and the standard is currently that svg are files not inline strings. There are no other inline svg across the entire codebase (other than in tests).

The fact that our svg icons are designed to be customisable is not why we spent the extra effort in TinyMCE 5 to make them files. It’s just good engineering practice.

lnewson: There's also other benefits to that as well, such as find-ability and the ability to see what it looks like without having to run the application. Think later if someone new has to update this how much easier it would be to find the actual image file rather than some hard coded/inlined string.

As for how to do this cheaply, one option might be to do the same as our major/minor version string substitution whereby in the source we use something like `@@majorVersion@@` and then have it be replaced by webpack/grunt when building.

TheSpyder: The import loaders should be easy enough to set up. We only did something custom for the icons because we do more than just import the file data. I'll see if I can make time to do it today.

TheSpyder: I have a branch that implements this successfully, I'll open a PR (targeting this PR branch) soon.

I'm currently yak shaving because this branch breaks the status bar tests.

TheSpyder:
The requested changes should be solved in #7578

lnewson:
Now that the inline SVG has been fixed, I just wanted to raise the other points that had been noted and also raise a potential issue that might need to be fixed...

For the `role="presentation"` part @spocke mentioned, if that was an a11y issue then it'd also be an issue for all our toolbar icons, etc... as well so we should probably investigate that and follow up in a separate Jira instead. Does anyone have any objections to following that up via a separate Jira?

One thing I noted when testing this is there is no focus style for the logo so you can't see what's focused. That does seem like an a11y issue so we probably need to chase that up with the design team and see if some styles should be added for that.

noxuhax:
After discussion with Paul I’ve updated logo styles to include both focused and hovered states.

I agree on a separate jira regarding SVG `role` attribute since it does not seem to be an immediate issue and more research is needed in order to fully understand whether or not it will ever become one.

lnewson (approved):

lnewson:
> After discussion with Paul I’ve updated logo styles to include both focused and hovered states.
> 
> I agree on a separate jira regarding SVG `role` attribute since it does not seem to be an immediate issue and more research is needed in order to fully understand whether or not it will ever become one.

Thanks Alex! As for the Jira I've gone ahead and logged TINY-8445 to look into that. Feel free to add anything I may have missed.

TheSpyder (approved):

spocke (approved):

PR #7574, opened 2022-01-28
TINY-8391: Fix test relying on previous state

ltrouton:
Related Ticket: TINY-8391

Description of Changes:
* Fix test error that was exposed from chunking shift

Pre-checks:

Review:
* [X] Milestone set

### DIFF
--- a/modules/tinymce/src/themes/silver/test/ts/browser/editor/color/ColorPickerSanityTest.ts
+++ b/modules/tinymce/src/themes/silver/test/ts/browser/editor/color/ColorPickerSanityTest.ts
@@ -97,6 +97,12 @@ describe('browser.tinymce.themes.silver.editor.color.ColorPickerSanityTest', () 
 
       it('TBA: Open dialog, pick a different color, click Cancel and assert color does not change', async () => {
         const editor = hook.editor();
+        // Change color to black
+        await pOpenDialog(editor);
+        await pSetHexBlack(editor);
+        TinyUiActions.submitDialog(editor);
+        await pWaitForDialogClose(editor);
+        // Change color in the dialog but cancel
         await pOpenDialog(editor);
         await pSetHexWhite(editor);
         await pCancelDialog(editor);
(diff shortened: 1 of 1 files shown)


modules/tinymce/src/themes/silver/test/ts/browser/editor/color/ColorPickerSanityTest.ts
```diff
@@ -97,6 +97,12 @@ describe('browser.tinymce.themes.silver.editor.color.ColorPickerSanityTest', ()
 
       it('TBA: Open dialog, pick a different color, click Cancel and assert color does not change', async () => {
```
lnewson, line 98: This needs updating to reflect the changes in the test
```suggestion
      it('TBA: Open dialog, pick black, reopen the dialog, pick a different color, click Cancel and assert color does not change', async () => {
```

lnewson (approved):

modules/tinymce/src/themes/silver/test/ts/browser/editor/color/ColorPickerSanityTest.ts
```diff
@@ -97,6 +97,12 @@ describe('browser.tinymce.themes.silver.editor.color.ColorPickerSanityTest', ()
...
         const editor = hook.editor();
+        // Change color to black
+        await pOpenDialog(editor);
+        await pSetHexBlack(editor);
+        TinyUiActions.submitDialog(editor);
+        await pWaitForDialogClose(editor);
```
lnewson, line 104: Actually... thinking about this more, should just do a `setContent` call instead?

ltrouton: The way the test file is set up, it looks like it doesn't use what is present in the editor at all. I'll confirm that though

TheSpyder (approved):

modules/tinymce/src/themes/silver/test/ts/browser/editor/color/ColorPickerSanityTest.ts
ltrouton: Looks like that is the case. I think I might just change the test message to more accurately point out what is happening. Thanks for pointing that out

PR #7578, opened 2022-01-31
TINY-8371: Swap SVG logo to a file

TheSpyder:
Related Ticket: TINY-8371 (#7573)

Description of Changes:
* Enable file loader for both webpack and rollup
* Extract the SVG to a file
* Swap the now-complex HTML string to an Alloy spec
* Adjusted the branding CSS to fix an issue where the top of the logo was cut off because the image was too close to the top of the status bar

Before:
before
After:
after

Pre-checks:

* [x] ~Changelog entry added~

Review:
* [ ] Milestone set


modules/tinymce/src/themes/silver/main/ts/ui/statusbar/Statusbar.ts
```diff
@@ -9,6 +9,7 @@ import { SimpleSpec } from '@ephox/alloy';
 
 import Editor from 'tinymce/core/api/Editor';
 import I18n from 'tinymce/core/api/util/I18n';
+import { Logo } from 'tinymce/themes/silver/resources/StatusbarLogo';
```
lnewson, line 12: Should this be a relative path since it's something within the silver theme? Normally anything within the `main` dir would need to be relative, but our regular codestyle doesn't really cover this case, so just thought it was worth asking.

modules/tinymce/tools/modules/grunt-webpack.js
```diff
@@ -29,6 +29,11 @@ let create = (entries, tsConfig, outDir, filename) => {
     },
     module: {
       rules: [
+        {
+          test: /\.(jpe?g|png|gif|svg)$/i,
```
lnewson, line 33:
```suggestion
          test: /\.(svg)$/i,
```
We definitely don't want regular images using the raw-loader, so just dropping it back to what we need for now.

patches/@ephox+bedrock-server+11.5.0.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/node_modules/@ephox/bedrock-server/lib/main/ts/bedrock/compiler/Webpack.js b/node_modules/@ephox/bedrock-server/lib/main/ts/bedrock/compiler/Webpack.js
+index 0a33f63..c3c6926 100644
+--- a/node_modules/@ephox/bedrock-server/lib/main/ts/bedrock/compiler/Webpack.js
++++ b/node_modules/@ephox/bedrock-server/lib/main/ts/bedrock/compiler/Webpack.js
+@@ -38,7 +38,7 @@ const webpackSharedRules = webpackRemap.concat([
+         enforce: 'pre'
+     },
+     {
+-        test: /\.(html|htm|css|bower|hex|rtf|xml|yml)$/,
++        test: /\.(html|htm|css|bower|hex|rtf|xml|yml|svg)$/i,
```
lnewson, line 10: For others reading, this was needed as Bedrock can't currently load images and for TinyMCE we're stuck on an older version of Bedrock atm so we can't just add this to bedrock and upgrade. So this was a solution Andy and I discussed that we thought was best until we can upgrade, as this shouldn't be needed elsewhere for older bedrock versions.

lnewson (approved):

modules/tinymce/src/themes/silver/main/ts/ui/statusbar/Statusbar.ts
TheSpyder: The problem is that when looking outside of `main/ts` the relative path might change when this source is compiled into the `lib` folder. I figured a named path like this was more consistently reliable.

lnewson: Yeah that's a good point, lets leave it as is then.

modules/tinymce/tools/modules/grunt-webpack.js
TheSpyder: ah yeah this was added before the swap to `raw`, thanks

spocke (approved):

patches/@ephox+bedrock-server+11.5.0.patch
spocke: Didn't know that we had all those types of imports already in bedrock.

modules/oxide/src/less/theme/components/statusbar/statusbar.less
```diff
@@ -81,8 +81,8 @@
   .tox-statusbar__branding {
     svg {
       fill: currentColor;
-      height: 1.3em;
-      vertical-align: -.3em;
+      height: 1.4em;
```
lnewson, line 84: One thing I just thought of... since this will be shared between 5.x and 6.x skins, this might need to be a variable and adjusted accordingly. I don't know enough about how that's all setup though, but just flagging it as a possible issue.

lnewson: Actually, that's only an issue if we're showing the text for 5.x. I don't believe we are so this probably isn't an issue.

TheSpyder: I think the 6.x skin PR can worry about that after the logo has merged :)

TheSpyder:
@noxuhax @lostkeys  I will wait for design review before merging this

ltrouton (approved):

modules/tinymce/src/themes/silver/main/ts/ui/statusbar/Statusbar.ts
```diff
@@ -19,17 +20,26 @@ import { renderWordCount } from './WordCount';
...
+      },
+      components: [
+        {
+          dom: {
+            tag: 'a',
+            attributes: {
+              'href': 'https://www.tiny.cloud/powered-by-tiny?utm_campaign=editor_referral&utm_medium=poweredby&utm_source=tinymce&utm_content=v6',
+              'rel': 'noopener',
+              'target': '_blank',
+              'tabindex': '-1',
```
noxuhax, line 36: Is there a particular reason for making this link unfocusable via keyboard? I might be wrong but it seems that combining both `tabindex=-1` and `aria-label` attributes is meaningless since screen reader won’t be able to read the lable without the link getting focused in the first place

noxuhax (approved):
From the visual point of view all seems good — great job on catching a pixel being cut off from the top of the logo. Logo size will be adjusted in the legacy `tinymce-5` skin to better fit into the smaller status bar.

modules/tinymce/src/themes/silver/main/ts/ui/statusbar/Statusbar.ts
noxuhax: Oh, and if we decide to remove `tabindex` attribute it would be a good idea to add `role="presentation"` to the svg as @spocke suggested previously.

lnewson: The tabindex=-1 is required because it's handled via our UI framework (Alloy) not via native keyboard navigation. That said though, now this is an Alloy spec this isn't quite the correct way to do this. Instead, we should be adding the `Focusing` behaviour instead.

TheSpyder: I didn't add or remove any attributes, I just converted the existing attributes to Alloy spec:
https://github.com/tinymce/tinymce/blob/8c13482153500dba4a1b16c033c190cc72b91417/modules/tinymce/src/themes/silver/main/ts/ui/statusbar/Statusbar.ts#L23
This discussion makes me happy, as does the new commit to change the URL. These problems existed before but were not noticed.

lnewson: Sorry I had noticed that as well and should have noted they were there beforehand in my reply as well 😞

lnewson: Was talking to @TheSpyder about this via slack. I've gone ahead and swapped it to use the `Focusing` behaviour as I mentioned.

As for the `role="presentation"` part, if that was an a11y issue then it'd also be an issue for all our toolbar icons, etc... as well so we should probably investigate that and follow up in a separate Jira instead. Does anyone have any objections to following that up via a separate Jira?

Now lastly, and this isn't related to these changes but more the original change, but the is no focus style for the logo so you can't see what's focused. That does seem like an a11y issue so we probably need to chase that up with the design team and see if some styles should be added for that. We should handle that after this PR is merged into the existing PR though.

PR #7579, opened 2022-01-31
TINY-6726: Change `:` control character to `~` for schema

ltrouton:
Related Ticket: TINY-6726

Description of Changes:
*  Change `:` control character to `~` that `:` can be used without needing escaping

Pre-checks:
* [X] Changelog entry added

Review:
* [X] Milestone set


modules/tinymce/src/core/main/ts/api/html/Schema.ts
```diff
@@ -486,7 +486,7 @@ const Schema = (settings?: SchemaSettings): Schema => {
     let ei, el, ai, al, matches, element, attr, attrData, elementName, attrName, attrType, attributes, attributesOrder,
       prefix, outputName, globalAttributes, globalAttributesOrder, value;
     const elementRuleRegExp = /^([#+\-])?([^\[!\/]+)(?:\/([^\[!]+))?(?:(!?)\[([^\]]+)])?$/,
-      attrRuleRegExp = /^([!\-])?(\w+[\\:]:\w+|[^=:<]+)?(?:([=:<])(.*))?$/,
+      attrRuleRegExp = /^([!\-])?(\w+[\\:]:\w+|[^=~<]+)?(?:([=~<])(.*))?$/,
```
ltrouton, line 489: I wasn't sure if we still wanted/needed to keep the `[\\:]` portion of the regex?

spocke: I think we should keep it since it will otherwise remove these attributes silently when loading content and the integrator might not notice until it's to late. This new format is more of a convenience thing and the removal of : is less dramatic since it will not throw away information it will just not enforce adding it so less of an issue.

lnewson: I'd agree with Spocke here that for now at least we should keep it. We may want to consider announcing it as deprecated though and remove it in a later version.

modules/tinymce/CHANGELOG.md
```diff
@@ -46,6 +46,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The `target_list` option has been renamed to `link_target_list` for the `link` plugin #TINY-4603
 - The `element_format` option has been set to `html` by default #TINY-8263
 - The `schema` option has been set to `html5` by default #TINY-8261
+- The `:` control chracter has been changed to `~` for the schema `valid_elements` and `extended_valid_elements` options #TINY-6726
```
lnewson, line 49: There's a typo here
```suggestion
- The `:` control character has been changed to `~` for the schema `valid_elements` and `extended_valid_elements` options #TINY-6726
```

metricjs (approved):

lnewson (approved):

modules/tinymce/src/core/main/ts/api/html/Schema.ts
spocke: If we are to keep it then the line that has `attrName = matches[2].replace(/[\\:]:/g, ':');` should be changed to `attrName = matches[2].replace(/[\\:]:/g, '~');` and a test for the escaped old version would be neeeded.

- [ ] Add support for legacy `\\:` and a test for that

spocke (approved):

modules/tinymce/src/core/main/ts/api/html/Schema.ts
ltrouton: I could be misunderstanding but by changing the attrName part, wouldn't that make it so that attributes names could no longer have  `:` in them? For example, `xml:lang` would become `xml~lang`

ltrouton: I did add another test to make sure `:` and `~` can work together. Although, I am still unsure of your suggestion @spocke so some clarification would be helpful

ltrouton: I think there may have been some confusion so have added a comment to one of the test cases clarifying that the literal and escaped versions are being tested

spocke: Thanks. Yes just add support back to the old format as well as the new and of the current code already supported thet then it's fine. I didn't checkout the branch this time to do a review of the whole file so if it had some other workaround or branch for the escaped code then it's all good.

PR #7583, opened 2022-02-01
TINY-8155/TINY-8391: Add DOM model and move non-UI table logic to model and core

ltrouton:
Related Ticket: TINY-8155

This is a checkout of the `spike/TINY-8155-alt-table` branch that has been rebased on the latest develop and the commits cleaned up.

Description of Changes:
* Add model concept to the editor
* Move main table action and commands to the dom model (modules/tinymce/src/models/dom/main/ts/table/api/Commands.ts)
Move fake cell selection handling to core selection (modules/tinymce/src/core/main/ts/api/dom/Selection.ts)
* At the moment, the fake selected cells can be retrieved using editor.selection.getSelectedCells which is what the table plugin and dom model are now using
* Move table resize handling to core selection (modules/tinymce/src/core/main/ts/api/dom/Selection.ts)
* Move tab navigation to core keyboard overrides (modules/tinymce/src/core/main/ts/keyboard/TabKey.ts)
* Move TableFormats to core (modules/tinymce/src/core/main/ts/fmt/FormatRegistry.ts)
* Move `mceInsertTable` command to core (modules/tinymce/src/core/main/ts/api/EditorCommands.ts)
* Remove the need for resize wire (wire) to be passed into Snooker
* Move table tests to where they are most appropriate

Pre-checks:
* [X] Changelog entry added

Review:
* [X] Milestone set

Before merge:
* [ ] Squash commits into reasonable changes


modules/tinymce/src/core/main/ts/init/Render.ts
```diff
@@ -45,9 +50,29 @@ const loadLanguage = (scriptLoader: ScriptLoader, editor: Editor) => {
...
+    const modelUrl = Options.getModelUrl(editor);
+    const url = Type.isString(modelUrl) ? editor.documentBaseURI.toAbsolute(modelUrl) : `models/${model}/model${suffix}.js`;
+    ModelManager.load(model, url).catch(() => {
+      ErrorReporter.modelLoadError(editor, url, model);
+    });
+
+    waitForModel = () => ModelManager.waitFor(model);
+  } else {
+    waitForModel = () => Promise.resolve();
+  }
```
lnewson, line 74: This really seems like the wrong approach here and in addition that, I doubt this is actually working because `Rtc.isRtc` will always return `false` at this point because `editor.plugins` is empty at this point.

modules/tinymce/src/core/main/ts/init/Render.ts
```diff
@@ -45,9 +50,29 @@ const loadLanguage = (scriptLoader: ScriptLoader, editor: Editor) => {
...
+ */
+const loadThemeAndModel = (scriptLoader: ScriptLoader, editor: Editor, suffix: string): Promise<void> => {
   const theme = Options.getTheme(editor);
 
+  // Special case the 'wait for model' code if RTC is loading, as it will provide a model instead
+  let waitForModel: () => Promise<void>;
+  if (!Rtc.isRtc(editor)) {
+    const model = Options.getModel(editor);
+    const modelUrl = Options.getModelUrl(editor);
+    const url = Type.isString(modelUrl) ? editor.documentBaseURI.toAbsolute(modelUrl) : `models/${model}/model${suffix}.js`;
```
lnewson, line 66: I know this is what the theme does, but this seems incorrect... it should be relative to the TinyMCE base url, not the document base url. That's probably a bug in the theme loading code as well.

modules/tinymce/src/core/main/ts/init/Render.ts
TheSpyder: Maybe it's been moved since I last tested it. The aim is to not download the `dom` model when RTC is going to replace it anyway.

modules/tinymce/src/core/main/ts/init/Render.ts
TheSpyder: Yeah, if there's a bug they both need to change 🤔 I can see arguments either way, for a single-page application it might make more sense to the developer for the URL they provide to be relative to the page they are editing. Not the TinyMCE script.

modules/tinymce/src/core/main/ts/init/Render.ts
lnewson: We probably need to discuss this more as it's not clear how RTC is going to register it. It feels like the better option here would be to override or change the default model to load when RTC is enabled (and potentially the url).

The loading order also seems to be wrong here if we want plugins to control the model, as the plugins should load first and then the model in that case. We can check to see if the model is already loaded and skip loading it in that case, as it'd be loaded and registered when the RTC plugin loads.

TheSpyder: I don't want plugins to register a model, RTC is just a special case until the alt-model can become a core feature.

The model needs to load before plugins, so we perhaps just need a better check than `isRtc` here.

lnewson: > The model needs to load before plugins, so we perhaps just need a better check than isRtc here.

I disagree here, as the model needs to init before plugins, so the plugin can use it but it should have no need to load before them.

TheSpyder: You're right, I conflated them, but I don't want to restrict what plugins can do before `init` too much. I guess they're already fairly restricted so that's an ok change to make.

TheSpyder: oh and I forgot to clarify
> it's not clear how RTC is going to register it

That happens automatically on the TinyMCE side
https://github.com/tinymce/tinymce/blob/1f7d16e4d680742369528c9f4033fde08e7fe76f/modules/tinymce/src/core/main/ts/Rtc.ts#L316

lnewson: Okay, that's leading me even more to believe the model manager is something we shouldn't be doing then if we're going with that approach. As if we're using the ModelManager API, then RTC should be calling `ModelManager.add('rtc', () => {...})` instead of monkey patching it like that.

TheSpyder: Ah github didn't update with your second comment before I made mine, sorry. We probably should be discussing this in Slack.

> This seems like a bad design/idea to me as why do we even have the model manager API then if it's not going to be customisable? We might as well just chunk the files and then fork the loading based on if RTC is enabled or not.

I thought it was required if we later have two models in the core? If not then yes we can drop the manager.

TheSpyder: The primary goal of this code is two-fold

* Only download the DOM model if RTC isn't configured
* Download the DOM model in parallel with the theme, because both are quite large and adding an extra round-trip to the init process is highly undesirable

modules/tinymce/src/core/main/ts/table/InsertTable.ts
```diff
@@ -39,11 +37,11 @@ const isPercentage = (width: string): boolean =>
   Type.isString(width) && width.indexOf('%') !== -1;
 
 const insert = (editor: Editor, columns: number, rows: number, colHeaders: number, rowHeaders: number): HTMLTableElement => {
```
spocke, line 39: Same here can't see why the insertTable command couldn't live in the dom core. I don't think any snooker things should be in the core.

modules/tinymce/src/core/main/ts/init/Render.ts
spocke: I think it's this way since you normally using script, css etc import relative to the page and your custom things are likely to live outside the tinymce directory so at the time it made sense to make them relative to the page not the tinymce script. And when loading from cloud it make no sense to have it relative to our cloud url. But the models here is a bit special we are not likely wanting third party to provide models so maybe they should just absolute urls since the only use case I can think of for us using a custom one is when you do cross project development or something. Wanting to load say the dom core from webpack into a project like powerpaste or something.

modules/tinymce/src/core/main/ts/init/Render.ts
TheSpyder: We had a big discussion this morning and we settled on an approach; the `model` option will default to `plugin` when RTC is in the plugins list. This code then changes to check whether the option is set to plugin rather than a specific RTC check. I'm about to add that to the PR and then open a matching PR on the RTC side.


> But I wonder if that change needs to be it's own PR and loading the dom model could be just an empty model in the first PR then we re-open this PR with the table changes moving into that model. This PR Is massive.

yes, sorry about the size, I nearly did a model-only PR months ago when the model API seemed solid. But we didn't have anything for the model to do, and I thought the table changes might impact how the model API was structured.

For now, what you're looking for has been squashed into the first commit of the PR.

TheSpyder: > I wonder why the theme is loaded before the other resources. I tried tweaking it so that they load in parallel instead and from the waterfall it seems to do just that and the editor works just fine. I think it might be some legacy thing where the theme was providing apis that plugins needed on load not on init. So we might be able to change that and then also load the dom model in parallel as well.

We've logged this as TINY-8453 and will see if we can squeeze it in for 6.0

modules/tinymce/src/core/main/ts/table/InsertTable.ts
TheSpyder: At the time we wanted something like insert table to be independent of whether RTC was enabled. I'm not sure if that's still true given we abandoned the tables feature for 6.0

modules/tinymce/src/core/main/ts/keyboard/TableNavigation.ts
```diff
@@ -6,7 +6,8 @@
  */
 
 import { Arr, Fun, Optional } from '@ephox/katamari';
-import { Attribute, Insert, SugarElement } from '@ephox/sugar';
+import { CellLocation, CellNavigation, TableLookup } from '@ephox/snooker';
```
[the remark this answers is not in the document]
TheSpyder: Anything we do in the DOM model needs to be replicated in RTC eventually. I don't know that having a second table navigation implementation is a good idea.

modules/tinymce/src/core/main/ts/api/Options.ts
```diff
@@ -799,6 +867,8 @@ const getEventRoot = option('event_root');
 const getServiceMessage = option('service_message');
 const getTheme = option('theme');
 const getThemeUrl = option('theme_url');
+const getModel = option('model');
```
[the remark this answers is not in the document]
TheSpyder: I dunno, `Options.getModel()` seems just as fine as `Options.getTheme()` to me, it gets an option not the actual concept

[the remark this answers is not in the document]
spocke: Fair point.

modules/tinymce/src/core/main/ts/keyboard/TableNavigation.ts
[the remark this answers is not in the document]
spocke: But we don't yet know what that navigation might look like on the rtc side we might even be able to do it though the model instead of the view. I though snooker would be replaced in the alt model with something new. But I guess we can move thing arou d later the breaking change here is to introduce a new dom model file once we have that and it's required we can choose to move things in and out of it and for the integrator it would work the same.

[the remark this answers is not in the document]
TheSpyder: I guess I hadn't thought of that, it just seemed like something that snooker was already good at. But maybe it's not very good at it and RTC can do better.

At a minimum, we'll evaluate the bundle size after this has merged and look at whether things like this bloat the core too much. We can, as you say, move things back and forth freely without flagging it as a breaking change.

modules/tinymce/src/core/main/ts/init/Render.ts
spocke: > We had a big discussion this morning and we settled on an approach; the model option will default to plugin when RTC is in the plugins list. This code then changes to check whether the option is set to plugin rather than a specific RTC check. I'm about to add that to the PR and then open a matching PR on the RTC side.

That sounds like a viable approach until we move the model. :+1:

modules/tinymce/src/core/main/ts/api/commands/TableCommands.ts
```diff
@@ -0,0 +1,17 @@
+/**
```
ltrouton, line 1: I added this file when rebasing and registered it in [Commands.ts](https://github.com/tinymce/tinymce/pull/7583/files#diff-9e36a5f6df6c7111295466824a34b6a2686bffa1983e57c43aa28bacd9b40361R61)

modules/tinymce/Gruntfile.js
```diff
@@ -175,6 +179,37 @@ module.exports = function (grunt) {
...
           ]
         };
+      }),
+      gruntUtils.generate(models, 'model', (name) => {
+        return {
+          options: {
+            treeshake: true,
+            format: 'iife',
+            onwarn: swag.onwarn,
+            plugins: [
```
lnewson, line 189:
```suggestion
            plugins: [
              FilesAsStrings,
```
I assume we're going to need this here as well given it was added elsewhere?

modules/tinymce/Gruntfile.js
```diff
@@ -175,6 +179,37 @@ module.exports = function (grunt) {
...
+              swag.nodeResolve({
+                basedir: __dirname,
+                prefixes: gruntUtils.prefixes({
+                  'tinymce/core': 'lib/globals/tinymce/core',
+                  'tinymce/ui': 'lib/ui/main/ts'
```
lnewson, line 194: This doesn't look right? That API/module was remove in 5.x and definitely isn't something a model should be using.
```suggestion
```

modules/tinymce/src/models/dom/demo/ts/demo/Demos.ts
```diff
@@ -0,0 +1,7 @@
+import BasicDemo from './BasicDemo';
+
+declare let window: any;
+
+window.demos = {
+  BasicDemo
+};
```
lnewson, line 7: Do we actually need something as "complex" as this for the `dom` model demo? This pattern is normally only used where we're going to need multiple demos (e.g. core uses this but plugins do not) and I don't think that applies here so we could just have the single `Demo.ts` file like in plugins.

modules/tinymce/src/core/main/ts/init/Init.ts
```diff
@@ -94,6 +95,18 @@ const initTheme = (editor: Editor) => {
...
+const initModel = (editor: Editor) => {
+  const model = Options.getModel(editor);
+  const modelUrl = ModelManager.urls[model];
+  const Model = ModelManager.get(model);
+
+  try {
+    editor.model = Model(editor, modelUrl) || {};
+  } catch (e) {
+    ErrorReporter.modelInitError(editor, model, e);
+  }
```
lnewson, line 107: This doesn't seem like the correct behaviour, as if this fails to init the editor cannot continue unlike plugins. This seems like it would be better to follow the same pattern as the theme init logic instead where it just kills the load because an uncaught error is thrown.

modules/tinymce/src/core/main/ts/init/Render.ts
```diff
@@ -45,9 +49,30 @@ const loadLanguage = (scriptLoader: ScriptLoader, editor: Editor) => {
   }
 };
 
-const loadTheme = (scriptLoader: ScriptLoader, editor: Editor, suffix: string): Promise<void> => {
+/**
+ * Initiate the download of both theme and model in parallel, then wait for both.
+ *
+ * Each has their own special case where nothing needs to be waited for.
+ */
+const loadThemeAndModel = (scriptLoader: ScriptLoader, editor: Editor, suffix: string): Promise<void> => {
```
lnewson, line 57: This logic probably needs reworking now sorry given the other change to load the theme in parallel. It could actually probably be a separate function now.

lnewson:
Have reviewed the first and then the last 4 commits which were the parts that hadn't been reviewed yet and overall looks good 👍  Not approving though because of the merge conflicts as I'd like to see how that's resolved before approving.

modules/tinymce/Gruntfile.js
TheSpyder: I doubt it's needed, really only the theme makes use of it, but yes it should go here.

modules/tinymce/Gruntfile.js
lnewson: Looks like the theme spec still has this as well, so it likely never was cleaned up when the `ui` directory was removed in 5.0 😞

TheSpyder: ouch. I guess good thing these changes highlighted that!

TheSpyder:
@lnewson we paired on some changes your comments have been resolved in the rebase + new commits

modules/tinymce/src/core/main/ts/ErrorReporter.ts
```diff
@@ -67,14 +67,6 @@ const pluginInitError = (editor: Editor, name: string, err) => {
   displayError(editor, message);
 };
 
-const modelInitError = (editor: Editor, name: string, err) => {
-  // TODO: TINY-8427 Add translation for this
```
lnewson: I guess we should close this Jira now since we're removing this?

lnewson (approved):

modules/tinymce/src/core/main/ts/ErrorReporter.ts
ltrouton: Yeah, I made a note to do that. I will do that now before I forget

ltrouton: Have marked the Jira as `Won't Do` now

modules/tinymce/src/core/main/ts/table/InsertTable.ts
spocke: Having a table in the content in RTC causes dom mutation due to resize handles and table cell selection and that causes a full reflow. We don't support editing tables yet in RTC but we should avoid all dom mutation. I think this would have been avoided if you some how move the selection model and table resizing out of the core and into the dom model. Even if snooker doesn't really bloat up the core here seems to just use a small part of it it's still causing issues.

- [ ] Branch out table selection and table resize handles if rtc is enabled or move them to the dom model

spocke (approved):

modules/tinymce/src/core/main/ts/table/InsertTable.ts
ltrouton: The latest 4 commits are my initial efforts to move the table resize handling, table cell selection and the `mceInsertTable` command to the model. I haven't done the "get selected cells" method aspect as that introduces more complexity so will have to be discussed further tomorrow on the best way to handle that aspect

modules/tinymce/src/models/dom/main/ts/table/api/TableCellSelection.ts
```diff
@@ -187,7 +185,15 @@ export const TableCellSelection = (editor: Editor): TableCellSelection => {
   const clear = (container: Node) =>
     annotations.clear(SugarElement.fromDom(container));
 
-  return {
-    clear
-  };
+  editor.addCommand('TableCellSelectionClear', (_ui, container: Node) => {
```
spocke, line 188: I wonder if this should instead be a model api function and also a selection api function. Since you have a function to get the selected cells. You could have a editor.selection.clearSelectedCells(). In the alt model it could be just a noop at this point.

modules/tinymce/src/models/dom/main/ts/table/api/TableCellSelection.ts
```diff
@@ -95,6 +92,7 @@ export const TableCellSelection = (editor: Editor): TableCellSelection => {
 
     const keydown = (event: KeyboardEvent) => {
       const wrappedEvent = DomEvent.fromRawEvent(event);
+      // TODO: Using command causes issue with undo state
       editor.execCommand('TableResizeHandlerHide', false, undefined, { skip_focus: true });
```
spocke, line 96: From what I can see this is only used internally inside the model do we need this to be a command? I think none of the commands are really required since it's about modules internally taking to each other within the model. I don't think we need to expose these at an api level.

modules/tinymce/src/core/main/ts/table/InsertTable.ts
spocke: Looking at the code I don't think we need to use commands or events for this. Unless I missed something it seems to be that different modules within the table model needs to talk to each other. That wouldn't need to be public api since commands and events become pubic api. It could be some internal pub/sub thing or shared api object between them. Or are these events needed for the advanced tables plugin?

ltrouton: Thanks for the suggestions @spocke. I have removed the command concept since you are right that they aren't strictly needed. Instead I have added a `getSelectedCells` and `clearSelectedCells` to the model API which core, table plugin and the advatble plugin can then use

TheSpyder (approved):

modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
```diff
@@ -21,7 +21,7 @@ describe('browser.tinymce.core.UndoManagerTest', () => {
     indent: false,
     entities: 'raw',
     base_url: '/project/tinymce/js/tinymce'
-  }, []);
+  }, [], true);
```
lnewson, line 24: This is worrying... why was this added and did the test also fail on `develop` without this? If it didn't then we likely have a bug in these changes.

lnewson:
We're also going to need to squash some of these commits before we merge, so I'm adding a checkbox for that.

ltrouton:
This PR has gotten and bit messy and is quite difficult to review properly so have split the model work and the table work into separate PRs. This should hopefully allow us to get the model changes in even if the table work is delayed from being reviewed
- #7624 (Model PR)
- #7625 (Table PR)

ltrouton:
Closing this as separate PRs have been created. See the previous comment for more details

PR #7589, opened 2022-02-03
TINY-8414: Added a guard for an edge case issue with inline textpattern detection

lnewson:
Related Ticket: TINY-8414

Description of Changes:

This is something that came up with the spellchecker tests when upgrading. The inline textpattern detection code expects that when it's triggered via space the offset will be 1 at minimum, but when using APIs it's possible for that to not be the case. While it's a pretty unlikely case to occur and can't happen naturally, as discussed via slack it was thought it was still worthwhile adding a guard for this case, so that's that guard.

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/textpatterns/core/InlinePattern.ts
+++ b/modules/tinymce/src/core/main/ts/textpatterns/core/InlinePattern.ts
@@ -260,7 +260,7 @@ const findPatterns = (editor: Editor, patterns: InlinePattern[], space: boolean)
   }
 
   return Utils.getParentBlock(editor, rng).bind((block) => {
-    const offset = rng.startOffset > 0 ? rng.startOffset - (space ? 1 : 0) : rng.startOffset;
+    const offset = Math.max(0, rng.startOffset - (space ? 1 : 0));
     return findPatternsRec(editor, patterns, rng.startContainer, offset, block);
   }).fold(() => [], (result) => result.matches);
 };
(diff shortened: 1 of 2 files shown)


modules/tinymce/src/core/main/ts/textpatterns/core/InlinePattern.ts
```diff
@@ -260,7 +260,7 @@ const findPatterns = (editor: Editor, patterns: InlinePattern[], space: boolean)
   }
 
   return Utils.getParentBlock(editor, rng).bind((block) => {
-    const offset = rng.startOffset - (space ? 1 : 0);
+    const offset = rng.startOffset > 0 ? rng.startOffset - (space ? 1 : 0) : rng.startOffset;
```
TheSpyder, line 263: how about
```suggestion
    const offset = Math.max(0, rng.startOffset - (space ? 1 : 0));
```

TheSpyder (approved):

modules/tinymce/src/core/main/ts/textpatterns/core/InlinePattern.ts
lnewson: That'd work 👍  I had originally used `Num.clamp` but it felt like overkill because we don't need to check the upper bounds.

ltrouton (approved):

PR #7590, opened 2022-02-03
TINY-8262: Change default for strikethrough format to an `s` tag for html5 schema types

ltrouton:
Related Ticket: TINY-8262

Description of Changes:
* Change the strikethrough format so that it applies an `s` tag for html5 content and applies the `span` tag with text-decoration for html4 content

Pre-checks:
* [X] Changelog entry added

Review:
* [X] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -60,6 +60,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The `media` plugin no longer treats `iframe`, `video`, `audio` or `object` elements as "special" and will validate the contents against the schema #TINY-8382
 - Renamed the `getShortEndedElements` Schema API to `getVoidElements` #TINY-8344
 - Changed the default statusbar element path delimiter from `»` to `›` #TINY-8372
+- Changed the default tag for the strikethrough format to the `s` tag for html5 schema types #TINY-8262
 - Renamed the `font_formats` option to `font_family_formats` #TINY-8328
 - Renamed the `fontselect` toolbar button and `fontformats` menu item to `fontfamily` #TINY-8328
 - Renamed the `fontsize_formats` option to `font_size_formats` #TINY-8328
--- a/modules/tinymce/src/core/main/ts/fmt/DefaultFormats.ts
+++ b/modules/tinymce/src/core/main/ts/fmt/DefaultFormats.ts
@@ -5,12 +5,15 @@
  * For commercial licenses see https://www.tiny.cloud/
  */
 
-import DOMUtils from '../api/dom/DOMUtils';
+import Editor from '../api/Editor';
 import Tools from '../api/util/Tools';
 import * as NodeType from '../dom/NodeType';
-import { Formats, FormatVars } from './FormatTypes';
+import { Format, Formats, FormatVars } from './FormatTypes';
 
-const get = (dom: DOMUtils) => {
+const get = (editor: Editor) => {
+  const dom = editor.dom;
+  const schemaType = editor.schema.type;
+
   const formats: Formats = {
     valigntop: [
       { selector: 'td,th', styles: { verticalAlign: 'top' }}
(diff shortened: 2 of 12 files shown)


modules/tinymce/src/core/test/ts/browser/fmt/DefaultFormatsWithSchemaTest.ts
```diff
@@ -0,0 +1,79 @@
...
+      const editor = hook.editor();
+      editor.setContent('<p>Test</p>');
+      TinySelections.setCursor(editor, [ 0, 0 ], 2);
+      toggleInlineStyle(editor, 'Strikethrough');
+      TinyAssertions.assertContent(editor, '<p><span style="text-decoration: line-through;">Test</span></p>');
+      toggleInlineStyle(editor, 'Strikethrough');
+      editor.setContent('<p>Test</p>');
+    });
+
+    it('TINY-8262: should convert "strike" tag to "span" tag', () => {
```
lnewson, line 73:
```suggestion
    it('TINY-8262: should convert "strike" tag to a "span" tag', () => {
```

modules/tinymce/CHANGELOG.md
```diff
@@ -60,6 +60,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The `media` plugin no longer treats `iframe`, `video`, `audio` or `object` elements as "special" and will validate the contents against the schema #TINY-8382
 - Renamed the `getShortEndedElements` Schema API to `getVoidElements` #TINY-8344
 - Changed the default statusbar element path delimiter from `»` to `›` #TINY-8372
+- Changed the default tag for the strikethrough format to the `s` tag for html5 schema types #TINY-8262
```
lnewson, line 63:
```suggestion
- Changed the default tag for the strikethrough format to the `s` tag when using a html 5 schema #TINY-8262
```

lnewson (approved):

modules/tinymce/src/core/main/ts/fmt/DefaultFormats.ts
```diff
@@ -167,11 +170,12 @@ const get = (dom: DOMUtils) => {
       { inline: 'u', remove: 'all', preserve_attributes: [ 'class', 'style' ] }
     ],
 
-    strikethrough: [
-      { inline: 'span', styles: { textDecoration: 'line-through' }, exact: true },
-      { inline: 'strike', remove: 'all', preserve_attributes: [ 'class', 'style' ] },
-      { inline: 's', remove: 'all', preserve_attributes: [ 'class', 'style' ] }
-    ],
+    strikethrough: (() => {
```
spocke, line 173: Is this a style we prefer having a anonymous function immediately invoked like this. At lest it's not something I've seen used a lot. It could instead be just a function above the get call.

```
const strikethrough = (schemaType: string): Format => {
  const span: Format = { inline: 'span', styles: { textDecoration: 'line-through' }, exact: true };
  const strike: Format = { inline: 'strike', remove: 'all', preserve_attributes: [ 'class', 'style' ] };
  const s: Format = { inline: 's', remove: 'all', preserve_attributes: [ 'class', 'style' ] };
  return schemaType !== 'html4' ? [ s, span, strike ] : [ span, s, strike ];
});
```

Then at the call site be `strikethough: strikethough(editor.schema.type)`. To me at least that is easier to read. But I guess it's a matter of style.

modules/tinymce/CHANGELOG.md
```diff
@@ -60,6 +60,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The `media` plugin no longer treats `iframe`, `video`, `audio` or `object` elements as "special" and will validate the contents against the schema #TINY-8382
 - Renamed the `getShortEndedElements` Schema API to `getVoidElements` #TINY-8344
 - Changed the default statusbar element path delimiter from `»` to `›` #TINY-8372
+- Changed the default tag for the strikethrough format to the `s` tag when using a html 5 schema #TINY-8262
```
spocke, line 63: It also converts `strike` to `s` elements should be in the changelog.

- [ ] Add details about auto converting strike to s when using a html5 schema.

spocke (approved):

modules/tinymce/src/core/main/ts/fmt/DefaultFormats.ts
ltrouton: I did it this way so that you could still see all of the format configs together without having to look at a separate function in a different location. I don't have any strong opinions so if others think the other way is preferable I am happy to go with that.

modules/tinymce/CHANGELOG.md
ltrouton: A changelog entry has been added for this

modules/tinymce/src/core/main/ts/fmt/DefaultFormats.ts
TheSpyder: It's unusual, but it's the only way to collect a bunch of variables without adding them to the scope. Or, as you said, make it a function nearby.

I don't mind too much which way we decide to go.

TheSpyder (approved):

modules/tinymce/src/core/main/ts/fmt/DefaultFormats.ts
ltrouton: We discussed it post-standup and are happy to go with the current inline function setup

PR #7591, opened 2022-02-03
TINY-8417: Use aria-controls instead of aria-owns in DropdownUtils

lnewson:
Related Ticket: TINY-8417

Description of Changes:

This just cleans up the changes done in https://github.com/tinymce/tinymce/pull/7301/ and adds a couple minor things I found when testing locally.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set
* [x] Rebase merge to keep commit history

Fixes https://github.com/tinymce/tinymce/issues/7300


ltrouton (approved):

modules/alloy/src/main/ts/ephox/alloy/alien/ComponentStructure.ts
```diff
@@ -2,11 +2,14 @@ import { Fun } from '@ephox/katamari';
 import { Compare, PredicateExists, SugarElement } from '@ephox/sugar';
 
 import { AlloyComponent } from '../api/component/ComponentApi';
+import * as AriaControls from '../aria/AriaControls';
 import * as AriaOwner from '../aria/AriaOwner';
 import { AnchorSpec } from '../positioning/mode/Anchoring';
 
 const isAriaPartOf = (component: AlloyComponent, queryElem: SugarElement<Node>): boolean =>
-  AriaOwner.find(queryElem).exists((owner) => isPartOf(component, owner));
+  AriaOwner.find(queryElem)
```
ephox-mogran, line 10: What drove the decision to use either one here? Are we using aria-owner somewhere else? Or just a precaution?

modules/alloy/src/main/ts/ephox/alloy/aria/AriaOwner.ts
```diff
@@ -1,46 +1,7 @@
...
-    Attribute.remove(elem, 'aria-owns');
-  };
-
-  return {
-    id: ariaId,
-    link,
-    unlink
-  };
-};
+const find = AriaManager.find(AriaManager.LinkableAttribute.AriaOwns);
```
ephox-mogran, line 3: Minor. I wonder if we should just combine AriaOwner and AriaControls now. I guess it's unclear what to call the combination.

ephox-mogran (approved):
Ignore my comments. I don't care much about any of them. Nothing to change.

modules/alloy/src/main/ts/ephox/alloy/alien/ComponentStructure.ts
lnewson: Good question... I didn't look into that. I'll see where else it's used, as it's possible we don't need the `AriaOwner` module at all anymore.

lnewson: So looks like it's used by the `FloatingToolbarButton` schema, for which `aria-owner` would be incorrect there again so we should swap that to use `aria-controls` as well. I'll swap it out and just drop all this back to a single `AriaControls` module.

modules/alloy/src/main/ts/ephox/alloy/aria/AriaOwner.ts
lnewson: It's all the one thing now anyways, as we didn't need the `AriaOwner` module in the end (see my other comment).

PR #7593, opened 2022-02-03
TINY-3799: Added a workaround to improve anchor rendering when using allow_html_in_named_anchor: true

lnewson:
Related Ticket: TINY-3799

Description of Changes:

This just makes some improvements and fixes the visual rendering issues with using `allow_html_in_named_anchor: true`. The screenshot below shows the two different versions should render the same now, with the top being an anchor with HTML content (`<a id="dfdfd">dfdfd</a>`) and the second being empty (`<a id="dfdfd"></a>dfdfd`).
image

**Note:** This in no way attempts to solve all the issues with enabling this option and only attempts to fix the poor rendering.

Pre-checks:
* [x] Changelog entry added
* [x] ~Tests have been added (if applicable)~ (Visual changes so no automated tests)

Review:
* [x] Milestone set

### DIFF
--- a/modules/oxide/src/less/theme/content/anchor/anchor.less
+++ b/modules/oxide/src/less/theme/content/anchor/anchor.less
@@ -20,16 +20,26 @@
       background: transparent data-uri('image/svg+xml;charset=UTF-8', '../../../../svg/anchor.svg') no-repeat center;
     }
 
-    cursor: default;
-    display: inline-block;
-    height: 12px !important;
-    padding: 0 2px;
-    -webkit-user-modify: read-only;
-    -moz-user-modify: read-only;
-    -webkit-user-select: all;
-    -moz-user-select: all;
-    user-select: all;
-    width: 8px !important;
+    // empty named anchor (allow_html_in_named_anchor: false)
+    &:empty {
+      cursor: default;
+      display: inline-block;
+      height: 12px !important;
+      padding: 0 2px;
+      -webkit-user-modify: read-only;
+      -moz-user-modify: read-only;
+      -webkit-user-select: all;
+      -moz-user-select: all;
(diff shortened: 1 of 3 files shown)


TheSpyder (approved):
I can't see any difference in the two lines of your screenshot. But maybe I'm just getting old.

lnewson:
> I can't see any difference in the two lines of your screenshot. But maybe I'm just getting old.

That was the intention, as I was trying to make them look the same given they use different styles now. For reference, this is what the content used to look like:
image

spocke (approved):
I guess at some point we should fix the real problem here might be easier to do in the alt model then we could have different element types for these things anchors, bookmarks and block anchors could then be different things and we just import/export/render them all as `a` elements.

TheSpyder:
> That was the intention, as I was trying to make them look the same

so then why does the description say

> The screenshot below shows the differences

maybe `s/differences/similarities/` ?😂

lnewson:
> I guess at some point we should fix the real problem here might be easier to do in the alt model then we could have different element types for these things anchors, bookmarks and block anchors could then be different things and we just import/export/render them all as `a` elements.

Yeah, honestly I don't know that this can be fixed without a massive amount of effort for the dom model, but I'd be pretty confident we could solve it with the alt model (since the model and view are detached). We've had I think 3 timeboxed attempts at it, but never had any luck finding something small to fix the other issues.

lnewson:
> > That was the intention, as I was trying to make them look the same
> 
> so then why does the description say
> 
> > The screenshot below shows the differences
> 
> maybe `s/differences/similarities/` ?😂

It was meant to be about the differences in HTML. I reworded it so it should make more sense.

ltrouton (approved):

PR #7598, opened 2022-02-07
TINY-8455: Make the plugins option return type be a string array

lnewson:
Related Ticket: TINY-8455

Description of Changes:
* Changes the return type for the `plugins` option to be a `string[]` since that's how it's always used
* Fixed an issue with the `Options.set` types, that meant type checking wasn't really working in some cases since it'd fall-through to the more generic override.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/api/EditorOptions.ts
+++ b/modules/tinymce/src/core/main/ts/api/EditorOptions.ts
@@ -107,10 +107,7 @@ export interface Options {
    * @param {String} name Name of a registered option.
    * @return {Boolean} True if the option value was successfully set, otherwise false.
    */
-  set: {
-    <K extends keyof EditorOptions>(name: K, value: EditorOptions[K]): boolean;
-    <T>(name: string, value: T): boolean;
-  };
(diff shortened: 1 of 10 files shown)


modules/tinymce/src/core/main/ts/api/EditorOptions.ts
```diff
@@ -107,10 +107,7 @@ export interface Options {
    * @param {String} name Name of a registered option.
    * @return {Boolean} True if the option value was successfully set, otherwise false.
    */
-  set: {
-    <K extends keyof EditorOptions>(name: K, value: EditorOptions[K]): boolean;
-    <T>(name: string, value: T): boolean;
-  };
+  set: <K extends string, T>(name: K, value: K extends keyof NormalizedEditorOptions ? NormalizedEditorOptions[K] : T) => boolean;
```
lnewson, line 110: This was an intentional type change from `EditorOptions` to `NormalizedEditorOptions` as well. The reason being when setting a value the processor can convert it so the normalized type is what we should be allowing for the setter, not the return type which is what `EditorOptions` is.

TheSpyder (approved):

ltrouton (approved):

spocke (approved):

HAFRMO (approved):

PR #7601, opened 2022-02-07
TINY-8399: Removed support for plugins allowing a mixture of a string array and of space separated strings

lnewson:
Related Ticket: TINY-8399

Description of Changes:

This changes the `plugins` option so it now only accepts a string array or a space separated string, instead of a mixture of the two. For example these are valid:

```js
plugins: 'advlist list image help wordcount'
```
```js
plugins: 'advlist,list,image,help,wordcount'
```
```js
plugins: [ 'advlist', 'list', 'image', 'help', 'wordcount' ]
```
however, this is no longer valid:
```js
plugins: [
  'advlist list image',
  'help wordcount'
]
```

Note: The [documentation](https://www.tiny.cloud/docs/configure/integration-and-setup/#plugins) actually only states we allow the first option, however removing the rest could be too risky and would end up being inconsistent with the built-in `string[]` option processor.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/options/NormalizeOptions.ts
+++ b/modules/tinymce/src/core/main/ts/options/NormalizeOptions.ts
@@ -25,12 +25,14 @@ const deviceDetection = PlatformDetection.detect().deviceType;
 const isPhone = deviceDetection.isPhone();
 const isTablet = deviceDetection.isTablet();
 
-const normalizePlugins = (plugins: string | string[]) => {
-  const pluginNames = Type.isArray(plugins) ? plugins.join(' ') : plugins;
-  const trimmedPlugins = Arr.map(Type.isString(pluginNames) ? pluginNames.split(' ') : [ ], Strings.trim);
-  return Arr.filter(trimmedPlugins, (item) => {
-    return item.length > 0;
-  });
+const normalizePlugins = (plugins: string | string[] | undefined) => {
+  if (Type.isNullable(plugins)) {
+    return [];
+  } else {
+    const pluginNames = Type.isArray(plugins) ? plugins : plugins.split(/[ ,]/);
+    const trimmedPlugins = Arr.map(pluginNames, Strings.trim);
+    return Arr.filter(trimmedPlugins, Strings.isNotEmpty);
+  }
 };
 
 const extractSections = (keys: string[], options: RawEditorOptions) => {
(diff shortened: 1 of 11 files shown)


modules/tinymce/src/core/main/ts/options/NormalizeOptions.ts
```diff
@@ -25,12 +25,14 @@ const deviceDetection = PlatformDetection.detect().deviceType;
...
-  });
+const normalizePlugins = (plugins: string | string[] | undefined) => {
+  if (Type.isNullable(plugins)) {
+    return [];
+  } else {
+    const pluginNames = Type.isArray(plugins) ? plugins : plugins.split(/[ ,]/);
+    const trimmedPlugins = Arr.map(pluginNames, Strings.trim);
+    return Arr.filter(trimmedPlugins, Strings.isNotEmpty);
+  }
 };
```
lnewson, line 36: This logic here never did the comma separation and instead that was done elsewhere after getting the option, for example here: https://github.com/tinymce/tinymce/blob/7a24521f672d7ba3de9ad8fb194681ed98185859/modules/tinymce/src/core/main/ts/init/Render.ts#L109 So this just moves it all into the one place, which will be needed when this is coupled with https://github.com/tinymce/tinymce/pull/7598. It also keeps it in sync with the `string[]` option built-in processor.

ltrouton (approved):

modules/tinymce/src/core/test/ts/browser/options/NormalizeOptionsTest.ts
```diff
@@ -96,13 +96,13 @@ describe('browser.tinymce.core.options.NormalizeOptionsTest', () => {
       assert.equal(settings.plugins, 'a b c d', 'Should be both forced and user plugins');
     });
 
-    it('Override defaults with forced_plugins using mixed types and spaces', () => {
+    it('Override defaults with forced_plugins using strings with commas', () => {
       const defaultSettings = {
-        forced_plugins: '  a   b'
+        forced_plugins: 'a, b'
```
spocke, line 101: Why was the whitespace removed here is it covered elsewhere since it still trims each separated string?

spocke (approved):

HAFRMO (approved):

modules/tinymce/src/core/test/ts/browser/options/NormalizeOptionsTest.ts
lnewson: Honestly, I deleted the old mixed one it's picked up that I updated and copied the one above and adjusted it so that's why it disappeared. I'll adjust one of the two to have more spaces to better test this.

PR #7603, opened 2022-02-07
TINY-8457: Deprecate the content property for setContent event

vpyshnenko:
Related Ticket: TINY-8457

Description of Changes:
* The `content` property is problematic on the SetContent event for rtc and it also doesn't make much sense to have it.
* add @deprecated tag for `SetContentEvent.content` property
* update changelog

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -135,6 +135,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Deprecated
 - The dialog button component `primary` property has been deprecated in favour of the new `buttonType` property #TINY-8304
+- The `SetContent` event `content` property has been deprecated and will be removed in the next major release #TINY-8457
 
 ## 5.10.2 - 2021-11-17
 
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -136,7 +136,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Deprecated
 - The dialog button component `primary` property has been deprecated in favour of the new `buttonType` property #TINY-8304
 - The `content` property on the `SetContent` event has been deprecated and will be removed in the next major release #TINY-8457
-- The return value of the `editor.setContent` API has been deprecated and will be removed in the next major release #TINY-8457
 
 ## 5.10.2 - 2021-11-17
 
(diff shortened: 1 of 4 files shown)


modules/tinymce/CHANGELOG.md
```diff
@@ -135,6 +135,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Deprecated
 - The dialog button component `primary` property has been deprecated in favour of the new `buttonType` property #TINY-8304
+- The `SetContent` event `content` property has been deprecated and will be removed in the next major release #TINY-8457
```
lnewson, line 138:
```suggestion
- The `content` property on the `SetContent` event has been deprecated and will be removed in the next major release #TINY-8457
```

lnewson (approved):
@spocke if we're doing this then do we also need to deprecate the return value from the `setContent` API? As that has a similar meaning to the `content` property in the `SetContent` event.

- [ ] Consider deprecating the return value for `setContent` as well

modules/tinymce/src/core/main/ts/api/Editor.ts
```diff
@@ -782,7 +782,7 @@ class Editor implements EditorObservable {
    * @method setContent
    * @param {String} content Content to set to editor, normally HTML contents but can be other formats as well.
    * @param {Object} args Optional content object, this gets passed around through the whole set process.
-   * @return {String} HTML string that got set into the editor.
+   * @deprecated @return {String} HTML string that got set into the editor.
```
lnewson, line 785: This is going to break the docs, so we either need to mention this in the description or only mention the it's been deprecated via the changelog. The description is probably going to be the better answer, so maybe something like this?

```
  /**
   * Sets the specified content to the editor instance, this will cleanup the content before it gets set using
   * the different cleanup rules options.
   * <br>
   * <em>The content return value was deprecated in TinyMCE 6.0 and has been marked for removal in TinyMCE 7.0.</em>
   * 
   * @method setContent
```

- [ ] Fix this per the above

modules/tinymce/CHANGELOG.md
```diff
@@ -135,6 +135,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Deprecated
 - The dialog button component `primary` property has been deprecated in favour of the new `buttonType` property #TINY-8304
+- The `content` property on the `SetContent` event has been deprecated and will be removed in the next major release #TINY-8457
```
lnewson, line 138: Sorry @vpyshnenko if my previous note was confusing, but we still need the changelog entry here for the `setContent` API.

```suggestion
- The `content` property on the `SetContent` event has been deprecated and will be removed in the next major release #TINY-8457
- The return value of the `editor.setContent` API has been deprecated and will be removed in the next major release #TINY-8457
```

spocke (approved):

PR #7604, opened 2022-02-08
TINY-8444: Fixed node filters not running when inserting certain content

lnewson:
Related Ticket: TINY-8444

Description of Changes:

This is something that regressed in https://github.com/tinymce/tinymce/pull/7448, as since `parser.parse()` no longer runs and instead we manually patch in the AST nodes then that in turn meant the node filters didn't run in this specific case. As such, I found we actually had 2 sets of filtering code paths, so I've merged them into 1 and updated the insert content logic to ensure the filters are run after we've adjusted the tree.

Note: I've also manually built this and linked it to the premium plugins that were failing and have verified the tests now pass there.

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/plugins/media/test/ts/browser/SubmitTest.ts
+++ b/modules/tinymce/src/plugins/media/test/ts/browser/SubmitTest.ts
@@ -24,7 +24,7 @@ describe('browser.tinymce.plugins.media.core.SubmitTest', () => {
   const customEmbed =
   '<div style="left: 0px; width: 100%; height: 0px; position: relative; padding-bottom: 56.338%; max-width: 650px;">' +
   '<iframe src="https://www.youtube.com/embed/IcgmSRJHu_8"' +
-  ' width="560" height="314" allowfullscreen="allowfullscreen">' +
+  ' width="560" height="314" allowfullscreen="allowfullscreen" data-mce-fragment="1">' +
   '</iframe></div>';
 
   const pTestResolvedEmbedContentSubmit = async (editor: Editor, url: string, expected: string) => {
--- a/modules/tinymce/src/plugins/media/test/ts/browser/SubmitTest.ts
+++ b/modules/tinymce/src/plugins/media/test/ts/browser/SubmitTest.ts
@@ -24,7 +24,7 @@ describe('browser.tinymce.plugins.media.core.SubmitTest', () => {
   const customEmbed =
   '<div style="left: 0px; width: 100%; height: 0px; position: relative; padding-bottom: 56.338%; max-width: 650px;">' +
   '<iframe src="https://www.youtube.com/embed/IcgmSRJHu_8"' +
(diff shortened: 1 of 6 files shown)


modules/tinymce/src/plugins/media/test/ts/browser/SubmitTest.ts
```diff
@@ -24,7 +24,7 @@ describe('browser.tinymce.plugins.media.core.SubmitTest', () => {
   const customEmbed =
   '<div style="left: 0px; width: 100%; height: 0px; position: relative; padding-bottom: 56.338%; max-width: 650px;">' +
   '<iframe src="https://www.youtube.com/embed/IcgmSRJHu_8"' +
-  ' width="560" height="314" allowfullscreen="allowfullscreen">' +
+  ' width="560" height="314" allowfullscreen="allowfullscreen" data-mce-fragment="1">' +
```
lnewson, line 27: This was removed in the previous PR mentioned in the description and it turns out it wasn't added because this logic wasn't running. I'm still not convinced it should be here, but it was like this in 5.x so it's a separate issue if it shouldn't be here.

spocke: I ran out of time digging deeper on this but I think it has to do with the attributes being prefixed by a filter in the media plugin. The markFragmentElements will add this attribute to all ast nodes then the unmarkFragmentElements will remove that attribute from all dom elements. However the media plugin will prefix the attributes like this `data-mce-p-data-mce-fragment="1"` then remove that `data-mce-p-` prefix and if I dump the html in the editor body element it still has this prefixed attributes so it's not removing them correctly. I think there is a async media resolver involved here as well so it might be that the element gets converted into an iframe in the dom after the insert is made. So just leaving you with some clues. Might have time to look at this tomorrow unless you solved it by then.

spocke: I think the filters in the media plugin can be changed to not prefix data attributes. Why it does this is that when a for example placeholder for a media element is used an img retains all the data that an object element would have but since the parser would throw away any invalid attributes they are prefixed this way. So if we either black list the data-mce-fragment or any data-mce- prefixed attribute from being prefixed it should theoretically work.

ltrouton (approved):

TheSpyder (approved):

modules/tinymce/src/plugins/media/test/ts/browser/SubmitTest.ts
lnewson: I figured it was probably something like that, so thanks for taking a look 👍 . While out of scope, since it's a quick fix I've added a check so it doesn't copy `data-mce-` prefixed attributes like you suggested. As for cleaning up the rest, I've logged TINY-8479 to look into that as it's a much larger change than I want to make here.

PR #7615, opened 2022-02-10
TINY-3422: Removed style field from image dialog

TheSpyder:
Related Ticket: TINY-3422

Description of Changes:
* Removed style field from the advanced tab of image dialog
* Adjusted dialog logic to calculate the style value on submit instead of on change

This was much harder than it first appeared, because every other field on the advanced tab did not store information independently. When updated they simply patched the style field, and the style field is what was serialised when the dialog was submitted.

Dialogs cannot store information in the API data other than fields that exist in the dialog, so I had to refactor the style generation logic into the submit handler. That itself required some additional work because it was very specifically structured to not have access to the other helper functions.

Technically this also means invalid values in the other fields are not dealt with until submit, but looking at the existing behaviour this validation was already an implicit behaviour. The only sign that invalid values would be ignored was that the style field did not update when they were entered.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/plugins/image/main/ts/ui/Dialog.ts
+++ b/modules/tinymce/src/plugins/image/main/ts/ui/Dialog.ts
@@ -225,40 +225,6 @@ const changeImages = (helpers: Helpers, info: ImageDialogInfo, state: ImageDialo
   changeSrc(helpers, info, state, api);
 };
 
-const calcVSpace = (css: StyleMap): string => {
-  const matchingTopBottom = css['margin-top'] && css['margin-bottom'] && css['margin-top'] === css['margin-bottom'];
-  return matchingTopBottom ? Utils.removePixelSuffix(String(css['margin-top'])) : '';
-};
-
-const calcHSpace = (css: StyleMap): string => {
-  const matchingLeftRight = css['margin-right'] && css['margin-left'] && css['margin-right'] === css['margin-left'];
-  return matchingLeftRight ? Utils.removePixelSuffix(String(css['margin-right'])) : '';
-};
-
-const calcBorderWidth = (css: StyleMap): string =>
-  css['border-width'] ? Utils.removePixelSuffix(String(css['border-width'])) : '';
-
-const calcBorderStyle = (css: StyleMap): string =>
-  css['border-style'] ? String(css['border-style']) : '';
-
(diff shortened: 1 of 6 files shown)


modules/tinymce/src/plugins/image/main/ts/ui/Dialog.ts
```diff
@@ -353,36 +335,17 @@ const makeDialogBody = (info: ImageDialogInfo): DialogType.TabPanelSpec | Dialog
...
-};
-
-const submitHandler = (editor: Editor) => (info: ImageDialogInfo) => (api: API): void => {
+const submitHandler = (editor: Editor, info: ImageDialogInfo, helpers: Helpers) => (api: API): void => {
   const data: ImageDialogData = Merger.deepMerge(fromImageData(info.image), api.getData());
 
-  editor.execCommand('mceUpdateImage', false, toImageData(data, info.hasAccessibilityOptions));
+  // The data architecture relies on passing everything through the style field for validation.
+  // Since the style field was removed that process must be simulated on submit.
+  const finalData = changeStyle2(helpers.parseStyle, helpers.serializeStyle, {
```
lnewson, line 343: Could we rename this now that we have deleted the original `changeStyle` function?

lnewson (approved):

modules/tinymce/src/plugins/image/main/ts/ui/Dialog.ts
TheSpyder: I thought about it, but at the time I was planning for this to be straight up reverted in the future. I now realise that isn't a viable solution to the long term problem so yeah I'll rename it.

spocke: I hope we never put that input field back it has been a pain to handle updating that and the other fields since they need to be in sync.

spocke (approved):

HAFRMO (approved):

modules/tinymce/src/plugins/image/main/ts/ui/Dialog.ts
TheSpyder: I think we will, but it will have validation so we can make it clear when things are wrong instead of simply blocking it.

TheSpyder: Turns out `changeStyle2` wasn't the important part of this code, it was the `style` attribute I was passing in to it. I deleted the method and all tests are still good.

PR #7623, opened 2022-02-13
TINY-8480: Fixed language packs loading incorrectly

lnewson:
Related Ticket: TINY-8480

Description of Changes:
* Fixed an incorrect variable being used in the language url generation which caused it to generate an empty url unless `language_url` was configured.
  * **Note:** The test I've added for this doesn't cover this exact case as we can't do that with our current test setup unfortunately, but we had nothing testing a custom language being loaded so this does at least assert languages can be loaded.
* Fixed an issue where the init sequence wasn't waiting for plugin language packs to be loaded before proceeding.
  * **Note:** This was something I'd noticed the old `ScriptLoader` implementation did but found it poor behaviour to be loading scripts not passed in. However, since it's needed for the this case I've added it back, but we may want to look into better solutions long term. It should have been the only thing left out from the old version, so I'm pretty hopefully the ScriptLoader should be identical now (except it now deals with promises instead of callbacks).

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/api/dom/ScriptLoader.ts
+++ b/modules/tinymce/src/core/main/ts/api/dom/ScriptLoader.ts
@@ -5,7 +5,7 @@
  * For commercial licenses see https://www.tiny.cloud/
  */
 
-import { Arr, Fun, Obj, Optional, Unique } from '@ephox/katamari';
+import { Arr, Fun, Obj, Optional, Type, Unique } from '@ephox/katamari';
 
 import Tools from '../util/Tools';
 import DOMUtils from './DOMUtils';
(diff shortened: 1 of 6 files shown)


TheSpyder (approved):

ltrouton (approved):

modules/tinymce/src/core/main/ts/api/dom/ScriptLoader.ts
```diff
@@ -222,52 +222,66 @@ class ScriptLoader {
...
+          self.states[url] = LOADED;
           execCallbacks('resolve', url);
-          return Promise.resolve();
-        } else if (self.states[url] === FAILED) {
+
+          // Immediately load additional scripts if any were added to the queue while loading this script
+          const queue = self.queue;
+          if (queue.length > 0) {
+            self.queue = [];
+            return load(queue).then(processResults);
```
TheSpyder, line 254: Does this mean that if plugins load additional resources other than translations it will hold up the editor load?

We'll need to audit any plugin that loads non-essential resources to make sure it happens on init rather than blocking editor load.

lnewson: Yes, this is how it worked in v5 as well so it shouldn't make any difference there. Also most plugins don't use the ScriptLoader queue, they just use the `loadScript` API so that wouldn't cause it to block.

PR #7626, opened 2022-02-14
TINY-7996: Ensure only strings are passed to get/set content events

lnewson:
Related Ticket: TINY-7996

Description of Changes:
* Update the `GetContent` event to ensure the `content` property is always a string.
* Update the `BeforeSetContent` event to actually serialize the tree passed to `setContent()` instead of always using an empty string. This also means `BeforeSetContent` is actually usable for tree content now.
* Cleaned up a few `isString` checks for the relevant events since they aren't needed anymore and didn't work with `tree` content (which is one reason we're changing this).
* Fixed/improved a few types for the get/set content operations

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/content/PrePostProcess.ts
+++ b/modules/tinymce/src/core/main/ts/content/PrePostProcess.ts
@@ -9,8 +9,32 @@ import { Result } from '@ephox/katamari';
 
 import Editor from '../api/Editor';
 import * as Events from '../api/Events';
-import AstNode from '../api/html/Node';
-import { Content, GetContentArgs, SetContentArgs } from './ContentTypes';
+import DomParser from '../api/html/DomParser';
+import HtmlSerializer from '../api/html/Serializer';
+import { EditorEvent } from '../api/util/EventDispatcher';
+import { Content, GetContentArgs, isTreeNode, SetContentArgs } from './ContentTypes';
+
+const serializeContent = (content: Content): string =>
+  isTreeNode(content) ? HtmlSerializer({ validate: false }).serialize(content) : content;
+
+const withSerializedContent = <R extends EditorEvent<{ content: string }>>(content: Content, fireEvent: (content: string) => R): R & { content: Content } => {
+  const serializedContent = serializeContent(content);
+  const eventArgs = fireEvent(serializedContent);
+  if (eventArgs.isDefaultPrevented()) {
+    return eventArgs;
+  } else if (isTreeNode(content)) {
+    // Restore the content type back to being an AstNode. If the content has changed we need to
+    // re-parse the new content, otherwise we can return the input.
+    if (eventArgs.content !== serializedContent) {
+      const rootNode = DomParser({ validate: false, forced_root_block: false }).parse(eventArgs.content, { context: content.name });
+      return { ...eventArgs, content: rootNode };
+    } else {
+      return { ...eventArgs, content };
+    }
+  } else {
+    return eventArgs;
+  }
+};
 
 const preProcessGetContent = <T extends GetContentArgs>(editor: Editor, args: T): Result<T, Content> => {
   if (args.no_events) {
@@ -18,8 +42,7 @@ const preProcessGetContent = <T extends GetContentArgs>(editor: Editor, args: T)
   } else {
     const eventArgs = Events.fireBeforeGetContent(editor, args);
     if (eventArgs.isDefaultPrevented()) {
-      const defaultContent = eventArgs.format === 'tree' ? new AstNode('body', 11) : '';
-      return Result.error(Events.fireGetContent(editor, { content: defaultContent, ...eventArgs }).content);
+      return Result.error(Events.fireGetContent(editor, { content: '', ...eventArgs }).content);
     } else {
       return Result.value(eventArgs);
     }
@@ -30,8 +53,8 @@ const postProcessGetContent = <T extends GetContentArgs>(editor: Editor, content
   if (args.no_events) {
     return content;
   } else {
-    const eventArgs = Events.fireGetContent(editor, { ...args, content });
-    return eventArgs.content;
+    const outcome = withSerializedContent(content, (c) => Events.fireGetContent(editor, { ...args, content: c }));
+    return outcome.content;
   }
 };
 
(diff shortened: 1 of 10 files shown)


modules/tinymce/src/core/main/ts/content/SetContentImpl.ts
```diff
@@ -99,5 +96,5 @@ export const setContentInternal = (editor: Editor, content: Content, args: SetCo
     } else {
       return setContentString(editor, body, content, args);
     }
-  }).getOr({ content, html: args.content });
+  }).getOr({ content, html: isTreeNode(args.content) ? '' : args.content });
```
lnewson, line 99: This here is just a case for where the editor has been killed, so it can be anything so I just made it be the same as it was previously. I also didn't see the point of serializing the tree here but I'm happy to if needed.

modules/tinymce/src/core/main/ts/content/PrePostProcess.ts
```diff
@@ -30,21 +53,21 @@ const postProcessGetContent = <T extends GetContentArgs>(editor: Editor, content
...
 };
 
 const preProcessSetContent = <T extends SetContentArgs>(editor: Editor, args: T): Result<T, undefined> => {
   if (args.no_events) {
     return Result.value(args);
   } else {
-    const eventArgs = Events.fireBeforeSetContent(editor, args);
-    if (eventArgs.isDefaultPrevented()) {
-      Events.fireSetContent(editor, eventArgs);
+    const outcome = withSerializedContent(args.content, (content) => Events.fireBeforeSetContent(editor, { ...args, content }));
```
spocke, line 65: Is outcome really a better describing name for this it's as abstract as `value` is it to describe that it got changed maybe `processedEventArgs` would be more suitable? But maybe I'm overly picky. So feel free to ignore this if you want to. :smile:

modules/tinymce/src/core/main/ts/content/PrePostProcess.ts
```diff
@@ -9,17 +9,40 @@ import { Result } from '@ephox/katamari';
...
+const withSerializedContent = <R extends EditorEvent<{ content: string }>>(content: Content, fireEvent: (content: string) => R): R & { content: Content } => {
+  const serializedContent = serializeContent(content);
+  const eventArgs = fireEvent(serializedContent);
+  if (eventArgs.isDefaultPrevented()) {
+    return eventArgs;
+  } else if (isTreeNode(content)) {
+    // Restore the content type back to being an AstNode. If the content has changed we need to
+    // re-parse the new content, otherwise we can return the input.
+    if (eventArgs.content !== serializedContent) {
+      const rootNode = DomParser({ validate: false, forced_root_block: false }).parse(eventArgs.content, { context: content.name });
```
spocke, line 29: So filters like `forced_root_block` is applied later on the tree fragment by running the filters on the ast node. So this is why this doesn't use the editor dom parser instance. So there is no risk here of the content being root level text nodes or other things that the editor can't support. Just want to double check.

spocke (approved):

modules/tinymce/src/core/main/ts/content/PrePostProcess.ts
lnewson: Fair point, I'll rename it. I was mainly just trying to show it wasn't the original event args, so `processedEventArgs` would do that.

modules/tinymce/src/core/main/ts/content/PrePostProcess.ts
lnewson: I should add it can also be a partial bit of content if using `selection.getContent({ format: 'tree' })`, or the node could also have come from `getContent({ format: 'tree' })`.

spocke: > that will be done later when it's actually being inserted

This was the answer I was looking for. Just wanted to double check that it will be filtered so that no one can write a BeforeSetContent filter that takes all paragraphs and turns them into strings and we get strings at root level of the editor or something silly like that.

spocke: Hmm. I just tested this and it seems that you can inject non schema validated elements since the validate is false and that doesn't seems to be enforced later.

This will add the `my-crap` element to the dom inside the editor even though it's not valid.
```
tinymce.activeEditor.on('beforesetcontent', (e) => e.content = '<my-crap></my-crap>');
tinymce.activeEditor.setContent(tinymce.html.DomParser({forced_root_block: false}).parse('hello'));
```

- [ ] Figure out if we should schema validate the fragment

lnewson: @spocke that can be reproduced with TinyMCE 5 as well, so it's not related to this. As an example, try this on TinyMCE 5:
```
tinymce.activeEditor.setContent(tinymce.html.DomParser({validate: false, forced_root_block: false}).parse('<my-crap></my-crap>'));
```

I'll log a separate Jira, as you're right the existing code doesn't run the passed AstNodes via the schema: https://github.com/tinymce/tinymce/blob/8f9b87a830cce3de3ed9810ba7f3999fed64e954/modules/tinymce/src/core/main/ts/content/SetContentImpl.ts#L81-L90

lnewson: This actually also goes all the way back to 4.x apparently, as I can replicate the issue there so it's likely something that was missed when the `tree` concept was first added.

spocke: Yeah the original purpose of this was for wordpress to be able to move content from one editor to another quickly so there was no need to filter anything since that was already done. However since then it's started to being used by integrations etc. I was hoping that we would be able to remove this concept at some point.

spocke: So it was purposely not firing events etc since it was all about getting the state back into an editor as quickly as possible to be used in Gutenberg. It has since then slowly started to become something else an alternative input new content into the editor method so if we add this new event and we are basically continuing on that path so then it should filter etc however then it will no longer do what it was originally intended for.

ltrouton (approved):

TheSpyder (approved):

PR #7628, opened 2022-02-14
TINY-8397: Opening a dialog before the UI renders in inline mode fails

vpyshnenko:
Related Ticket: TINY-8397

Description of Changes:
* Placeholder text

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


lnewson (changes requested):
This is not the right solution here, as this is needed to ensure the initial data is set for every component. While we do pass in the initial data for a number of components as of recent changes, we don't do it for all so this must remain (you'll see a number of tests are also failing because this is needed).

Additionally, this wouldn't solve the issue for the for inline dialogs either. As the Jira notes, this needs to be solved by making sure the window doesn't open at all, it triggers focus to render the full UI, or renders the UI without focusing.

lnewson:
I should add that even if this did prevent the error throwing on open, some of the dialog APIs would still fail because the sink hasn't been rendered/attached yet. So we need something that makes them work across the board in this case.

Another option might be to look at updating the open or instance API logic to do nothing if the dialog isn't in the body. However, you could run into issues then when it does attach that it might not have the initial data so if you do go down that route make sure that's not an issue.

TheSpyder:
Viktor noted in JIRA that the dialog still initialises without this call. Let's move the discussion there.

lnewson:
> Viktor noted in JIRA that the dialog still initialises without this call. Let's move the discussion there.

Except that it doesn't, as noted above and by the numerous test fails 😉  Anyways, for now I think we should close this and discuss in the Jira as mentioned.

TheSpyder (approved):
That looks good to me! Now all it needs is a changelog

- [ ] add changelog

modules/tinymce/src/themes/silver/main/ts/ui/dialog/WindowManager.ts
```diff
@@ -122,6 +122,7 @@ const setup = (extras: WindowManagerSetup): WindowManagerImpl => {
         backstage
       );
 
+      editor.ui.show();
```
lnewson, line 125: This isn't the right place for this, if you read the Jira comments I had mentioned it should go in the **core** WindowManager, this is in the theme and only solve the issue for modal dialogs not inline dialogs, which is why it should be in core.

TheSpyder: oh I missed that it was in the wrong place. good pickup.

lnewson: Looking, I think somewhere in the `storeSelectionAndOpenDialog` function is probably going to be best.

modules/tinymce/CHANGELOG.md
```diff
@@ -107,6 +107,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Anchor elements would render incorrectly when using the `allow_html_in_named_anchor` option #TINY-3799
 - Fixed sub-menu items not read by screen readers. Patch contributed by westonkd #TINY-8417
 - The Home or End keys would move out of a editable element contained within a noneditable element #TINY-8201
+- The dialog could not be opened before the editor has been rendered #TINY-8397
```
lnewson, line 110:
```suggestion
- Dialogs could not be opened in inline mode before the editor had been rendered #TINY-8397
```

modules/tinymce/src/themes/silver/test/ts/browser/window/SilverDialogBeforeEditorRenderedTest.ts
```diff
@@ -0,0 +1,30 @@
...
+        ed.windowManager.open({
+          title: 'Test Dialog',
+          body: {
+            type: 'panel',
+            items: []
+          },
+          buttons: []
+        });
+      });
+    }
```
lnewson, line 21:
```suggestion
```
This would be better suited for inside the actual test.

modules/tinymce/src/themes/silver/test/ts/browser/window/SilverDialogBeforeEditorRenderedTest.ts
```diff
@@ -0,0 +1,30 @@
...
+  }, []);
+
+  it('TINY-8397: Open dialog while the inline editor is hidden should not throw an exception', async () => {
+    const editor = hook.editor();
+    await TinyUiActions.pWaitForDialog(editor, 'div[role="dialog"].tox-dialog');
```
lnewson, line 26:
```suggestion
    const editor = hook.editor();
    editor.windowManager.open({
        title: 'Test Dialog',
        body: {
          type: 'panel',
          items: []
        },
        buttons: []
      });
    });
    await TinyUiActions.pWaitForDialog(editor, 'div[role="dialog"].tox-dialog');
```

lnewson: Also I can't leave a suggestion for it properly, but we should assert the UI hasn't been rendered yet before attempting to open the dialog. For example:
```
UiFinder.notExists(SugarBody.body(), '.tox-tinymce-aux');
```

lnewson (approved):

lnewson (approved):

ltrouton (approved):

PR #7630, opened 2022-02-14
TINY-8201: Pressing the "End" key within a CEF element causes the selection to incorrectly leave the element

spocke:
Related Ticket:

Description of Changes:
* Fixes an issue where home and end keys would move outside the current editable host element
* Sneaked in a fix for a test that was flaking on/off due to the 100 test pivot point moving around
* Cleaned up the tests a bit so that is uses `assertCursor` instead of `assertSelection`

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/caret/CaretUtils.ts
+++ b/modules/tinymce/src/core/main/ts/caret/CaretUtils.ts
@@ -6,7 +6,7 @@
  */
 
 import { Fun, Optional } from '@ephox/katamari';
-import { SugarElement } from '@ephox/sugar';
+import { PredicateFind, SugarElement } from '@ephox/sugar';
 
 import DomTreeWalker from '../api/dom/TreeWalker';
 import * as NodeType from '../dom/NodeType';
@@ -76,14 +76,10 @@ const findNode = (node: Node, direction: number, predicateFn: (node: Node) => bo
   return null;
 };
 
-const getEditingHost = (node: Node, rootNode?: Node): Node | undefined => {
-  for (node = node.parentNode; node && node !== rootNode; node = node.parentNode) {
-    if (isContentEditableTrue(node)) {
-      return node;
-    }
-  }
-
-  return rootNode;
+const getEditingHost = (node: Node, rootNode: HTMLElement): HTMLElement => {
+  return PredicateFind.ancestor(SugarElement.fromDom(node), (node) => isContentEditableTrue(node.dom), (node) => node.dom === rootNode)
+    .map((elm) => elm.dom)
+    .getOr(rootNode) as HTMLElement;
 };
 
 const getParentBlock = (node: Node, rootNode?: Node): Node | null => {
@@ -101,7 +97,7 @@ const getParentBlock = (node: Node, rootNode?: Node): Node | null => {
 const isInSameBlock = (caretPosition1: CaretPosition, caretPosition2: CaretPosition, rootNode?: Node): boolean =>
   getParentBlock(caretPosition1.container(), rootNode) === getParentBlock(caretPosition2.container(), rootNode);
 
-const isInSameEditingHost = (caretPosition1: CaretPosition, caretPosition2: CaretPosition, rootNode?: Node): boolean =>
+const isInSameEditingHost = (caretPosition1: CaretPosition, caretPosition2: CaretPosition, rootNode: HTMLElement): boolean =>
   getEditingHost(caretPosition1.container(), rootNode) === getEditingHost(caretPosition2.container(), rootNode);
 
 const getChildNodeAtRelativeOffset = (relativeOffset: number, caretPosition: CaretPosition): Node | null => {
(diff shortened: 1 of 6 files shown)


modules/tinymce/src/core/test/ts/browser/init/InitEditorLanguageTest.ts
```diff
@@ -22,6 +22,6 @@ describe('browser.tinymce.core.init.InitEditorLanguageTest', () => {
   it('TBA: Should have been able to load a custom language pack', () => {
     UiFinder.notExists(SugarBody.body(), '.tox-notification');
     assert.equal(I18n.getCode(), 'custom', 'I18n language code should be the custom language code');
-    assert.deepEqual(I18n.getData(), { custom: customLanguagePack }, 'I18n data should have custom language pack');
+    assert.deepEqual(I18n.getData().custom, customLanguagePack, 'I18n data should have custom language pack');
```
spocke, line 25: This was flaking since there is a test that adds an Arabic language pack to verify that rtl mode is detected. So I changes this to be more explicit and only check for the custom language pack.

modules/tinymce/src/core/main/ts/caret/CaretUtils.ts
```diff
@@ -76,14 +76,10 @@ const findNode = (node: Node, direction: number, predicateFn: (node: Node) => bo
...
-      return node;
-    }
-  }
-
-  return rootNode;
+const getEditingHost = (node: Node, rootNode: HTMLElement): HTMLElement => {
+  return PredicateFind.ancestor(SugarElement.fromDom(node), (node) => isContentEditableTrue(node.dom), (node) => node.dom === rootNode)
+    .map((elm) => elm.dom)
+    .getOr(rootNode) as HTMLElement;
 };
```
lnewson, line 83:
```suggestion
const getEditingHost = (node: Node, rootNode: HTMLElement): HTMLElement =>
  PredicateFind.ancestor(SugarElement.fromDom(node), (node): node is SugarElement<HTMLElement> => isContentEditableTrue(node.dom), (node) => node.dom === rootNode)
    .map((elm) => elm.dom)
    .getOr(rootNode);
```
This avoids the cast which we should be avoiding if possible. While this still does do some casting, it only applies for the predicate within the `ancestor` lookup. It could also be worthwhile splitting the predicate out into a separate function 🤷  For example:
```ts
const getEditingHost = (node: Node, rootNode: HTMLElement): HTMLElement => {
  const isCETrue = (node: SugarElement<Node>): node is SugarElement<HTMLElement> => isContentEditableTrue(node.dom);
  const isRoot = (node: SugarElement<Node>) => node.dom === rootNode;
  return PredicateFind.ancestor(SugarElement.fromDom(node), isCETrue, isRoot)
    .map((elm) => elm.dom)
    .getOr(rootNode);
};
```

modules/tinymce/src/core/test/ts/browser/init/InitEditorLanguageTest.ts
lnewson: Thanks for this! I wonder if we need to add a way to clean up the I18n data between tests 🤔

lnewson (approved):

modules/tinymce/src/core/main/ts/caret/CaretUtils.ts
spocke: Right yeah it would cast anyway not sure that can be avoided here since we know that it is a HTMLElement because of the predicate so I was unsure where to stick that casting. I updated the function to your suggestions. I wonder how much effort it would be to remove the SugarElement wrapping object it serves little to no purpose in a typed language and it would make things like this where you try to marry sugar with regular dom more easy. In rockfest we use a sugar like library but with regular nodes, elements etc. Anyway, thanks for the suggestion.

modules/tinymce/src/core/test/ts/browser/init/InitEditorLanguageTest.ts
spocke: Maybe the resource thing required it since it was using the same key. This uses different keys so less of an issue to have a unload language pack here.

modules/tinymce/src/core/main/ts/caret/CaretUtils.ts
TheSpyder: We use unwrapped references in RTC because it's an abstract type. Arguably sugar wrappers serve a lot less purpose now that they're not thunked, but they do at least (mostly) enforce a 'not null' requirement.

TheSpyder (approved):

modules/tinymce/src/core/main/ts/caret/CaretUtils.ts
spocke: But wouldn't strict mode eventually give us that for ts too or is that only solving implicit anys not enforcing foo | null on everything.

TheSpyder: The not null benefit happens through our `fromDom` method. Otherwise we're relying on the TS libdom to declare when DOM methods might return null. Maybe strict mode will help with this? We're a very long way from enabling strict in TinyMCE though.

spocke: Right yeah that makes sense. I guess my brain has been destroyed by not having to deal with null values for a couple of years. :smile:

PR #7632, opened 2022-02-14
TINY-6946: Added regression tests for base64 image parsing

HAFRMO:
Related Ticket:

Description of Changes:
Adding tests to ensure that corrupted base64 file data does not cause the editor to crash.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/test/ts/atomic/html/Base64UrisTest.ts
+++ b/modules/tinymce/src/core/test/ts/atomic/html/Base64UrisTest.ts
@@ -134,7 +134,12 @@ describe('atomic.tinymce.core.html.Base64UrisTest', () => {
     KAssert.eqOptional('Data uri without base64', Optional.none(), parseDataUri('data:image/svg+xml,R4/yw=='));
     KAssert.eqOptional('Data with spaces, tabs and line breaks', Optional.some({ type: 'image/png', data: 'SGVsbG8sI\r\n  Hdv		cmxkIQ==' }), parseDataUri('data:image/png;base64,SGVsbG8sI\r\n  Hdv		cmxkIQ=='));
     KAssert.eqOptional('Corrupted data', Optional.some({ type: 'image/gif', data: 'R0' }), parseDataUri('data:image/gif;base64,R0ÖlGODdhIAAgAIABAP8AAP///ywAAAAAIAAgAAACHoSPqcvtD6OctNqLs968+w+G4kiW5omm6sq27gubBQA7AA==%A0'));
-    // eslint-disable-next-line max-len
-    KAssert.eqOptional('Corrupted data due to added letters', Optional.some({ type: 'image/gif', data: 'R0lGODdhIAAgAIABAP8AAP///ywAAAAAIAAgAAACHoSPqcvtD6OctNqLs968+w+G4kiW5omm6sq27gubBQA7AA==' }), parseDataUri('data:image/gif;base64,R0lGODdhIAAgAIABAP8AAP///ywAAAAAIAAgAAACHoSPqcvtD6OctNqLs968+w+G4kiW5omm6sq27gubBQA7AA==%A0'));
+    KAssert.eqOptional(
+      'Corrupted data due to added letters',
+      Optional.some({
+        type: 'image/gif',
+        data: 'R0lGODdhIAAgAIABAP8AAP///ywAAAAAIAAgAAACHoSPqcvtD6OctNqLs968+w+G4kiW5omm6sq27gubBQA7AA=='
+      }),
+      parseDataUri('data:image/gif;base64,R0lGODdhIAAgAIABAP8AAP///ywAAAAAIAAgAAACHoSPqcvtD6OctNqLs968+w+G4kiW5omm6sq27gubBQA7AA==%A0'));
   });
 });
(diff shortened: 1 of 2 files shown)


modules/tinymce/src/core/test/ts/browser/EditorTest.ts
```diff
@@ -469,6 +469,18 @@ describe('browser.tinymce.core.EditorTest', () => {
     input.parentNode.removeChild(input);
   });
 
+  it('TINY-6946: Images should be properly cleaned up and not cause a crash', () => {
+    const editor = hook.editor();
+    editor.setContent('<img src="data:image/gif;base64,R0lGODdhIAAgAIABAP8AAP///ywAAAAAIAAgAAACHoSPqcvtD6OctNqLs968+w+G4kiW5omm6sq27gubBQA7AA==%A0">');
+    TinyAssertions.assertContent(editor, '<p><img src="data:image/gif;base64,R0lGODdhIAAgAIABAP8AAP///ywAAAAAIAAgAAACHoSPqcvtD6OctNqLs968+w+G4kiW5omm6sq27gubBQA7AA=="></p>');
+  });
+
+  it('TINY-6946: Images should cut off invalid data, even if the image remains invalid', () => {
```
lnewson, line 478: This doesn't look right? Generally our rule is to try to not lose data... so are you sure this is expected behaviour?

lnewson: To add to this, browsers don't cut off invalid data when parsing so this looks like a bug to me.

lnewson: Ahh, it's probably because we convert it to a blob and then convert back to base64 content. In which case it'd be expected since the blob can't contain invalid data. If that is the reason, then can you please add a comment about why it's expected we're losing content?

modules/tinymce/src/core/test/ts/atomic/html/Base64UrisTest.ts
```diff
@@ -133,5 +133,8 @@ describe('atomic.tinymce.core.html.Base64UrisTest', () => {
     KAssert.eqOptional('Data uri without mime', Optional.none(), parseDataUri('data:base64,R3/yw=='));
     KAssert.eqOptional('Data uri without base64', Optional.none(), parseDataUri('data:image/svg+xml,R4/yw=='));
     KAssert.eqOptional('Data with spaces, tabs and line breaks', Optional.some({ type: 'image/png', data: 'SGVsbG8sI\r\n  Hdv		cmxkIQ==' }), parseDataUri('data:image/png;base64,SGVsbG8sI\r\n  Hdv		cmxkIQ=='));
+    KAssert.eqOptional('Corrupted data', Optional.some({ type: 'image/gif', data: 'R0' }), parseDataUri('data:image/gif;base64,R0ÖlGODdhIAAgAIABAP8AAP///ywAAAAAIAAgAAACHoSPqcvtD6OctNqLs968+w+G4kiW5omm6sq27gubBQA7AA==%A0'));
+    // eslint-disable-next-line max-len
+    KAssert.eqOptional('Corrupted data due to added letters', Optional.some({ type: 'image/gif', data: 'R0lGODdhIAAgAIABAP8AAP///ywAAAAAIAAgAAACHoSPqcvtD6OctNqLs968+w+G4kiW5omm6sq27gubBQA7AA==' }), parseDataUri('data:image/gif;base64,R0lGODdhIAAgAIABAP8AAP///ywAAAAAIAAgAAACHoSPqcvtD6OctNqLs968+w+G4kiW5omm6sq27gubBQA7AA==%A0'));
```
spocke, line 138: You could separate this up into multiple lines so that it easier to read and to avoid that lint disable.

```suggestion
    KAssert.eqOptional(
      'Corrupted data due to added letters',
      Optional.some({
        type: 'image/gif',
        data: 'R0lGODdhIAAgAIABAP8AAP///ywAAAAAIAAgAAACHoSPqcvtD6OctNqLs968+w+G4kiW5omm6sq27gubBQA7AA=='
      }),
      parseDataUri('data:image/gif;base64,R0lGODdhIAAgAIABAP8AAP///ywAAAAAIAAgAAACHoSPqcvtD6OctNqLs968+w+G4kiW5omm6sq27gubBQA7AA==%A0')
    );
```

modules/tinymce/src/core/test/ts/browser/EditorTest.ts
```diff
@@ -469,6 +469,18 @@ describe('browser.tinymce.core.EditorTest', () => {
     input.parentNode.removeChild(input);
   });
 
+  it('TINY-6946: Images should be properly cleaned up and not cause a crash', () => {
```
spocke, line 472: Maybe change this from crash to `Images should be properly cleaned up if they contain invalid trailing data` or something?

modules/tinymce/src/core/test/ts/browser/EditorTest.ts
spocke: This is because the base64 parser will parse valid things as long as it can then convert that to a blob url and replace the whole attribute contents with that new blob url. So anything that isn't valid base64 after the valid one will be thrown away. In theory we could try to retain the invalid data here but not sure what the point of that would be exactly since the browser will not render it anyway since it's corrupt.
So it's kind of cleaning up invalid base64 urls by accident but that is what the customer requested here so we just add a test so that we don't break this in the future.

spocke (approved):

TheSpyder (approved):

modules/tinymce/src/core/test/ts/atomic/html/Base64UrisTest.ts
```diff
@@ -133,5 +133,13 @@ describe('atomic.tinymce.core.html.Base64UrisTest', () => {
     KAssert.eqOptional('Data uri without mime', Optional.none(), parseDataUri('data:base64,R3/yw=='));
     KAssert.eqOptional('Data uri without base64', Optional.none(), parseDataUri('data:image/svg+xml,R4/yw=='));
     KAssert.eqOptional('Data with spaces, tabs and line breaks', Optional.some({ type: 'image/png', data: 'SGVsbG8sI\r\n  Hdv		cmxkIQ==' }), parseDataUri('data:image/png;base64,SGVsbG8sI\r\n  Hdv		cmxkIQ=='));
+    KAssert.eqOptional('Corrupted data', Optional.some({ type: 'image/gif', data: 'R0' }), parseDataUri('data:image/gif;base64,R0ÖlGODdhIAAgAIABAP8AAP///ywAAAAAIAAgAAACHoSPqcvtD6OctNqLs968+w+G4kiW5omm6sq27gubBQA7AA==%A0'));
+    KAssert.eqOptional(
+      'Corrupted data due to added letters',
```
lnewson, line 138:
```suggestion
      'Corrupted data due to added URL encoded nbsp',
```

lnewson (approved):

PR #7633, opened 2022-02-15
TINY-8463: Update default skin for v6, add tinymce5 skin with old design

noxuhax:
Related Ticket: TINY-8463

Description of Changes:
* Updated default theme styling for TinyMCE 6 facelift
* Updated default content theme for TinyMCE 6 facelift
* Added tinymce5 UI and content skins with legacy styling
* Renamed dark UI and content skins to tinymce5-dark

Pre-checks:
* [x] Changelog entry added
* [x] Tests have been updated

Review:
* [x] Milestone set


lostkeys (approved):

lnewson:
Looking good! Looks like we have some more test failures I believe just due to positional assertions that still need updating though. Let us know if you need a hand to resolve them.

One question I did have though, is it expected to have the toolbar separator for a single toolbar row as I don't recall seeing that in the figma designs? (in this image there's an extra line between the toolbar and the edge of the "header" container)
image

lnewson:
Another potential bug I found with this, the color swatch in the format menu has some extra padding that I'm guessing shouldn't be there? This could also be caused by the nesting we need to cleanup at some stage (see TINY-3604), as I believe we had to add some specific CSS previously to workaround this until that could be fixed.
image

noxuhax:
> One question I did have though, is it expected to have the toolbar separator for a single toolbar row as I don't recall seeing that in the figma designs? (in this screenshot of our full featured demo there's an extra line between the toolbar and the edge of the "header" container) image

I believe you might have some scaling enabled in your browser? I was able to reproduce that with 110% zoom in Chrome and it seems like an error in rounding of background scale. I’ve modified a formula for bg size as a failsafe so it should be fine now, would be great if you could re-check.

> the color swatch in the format menu has some extra padding that I'm guessing shouldn't be there

Fixed that one as well, thanks for letting me know about this case.

lnewson:
> I believe you might have some scaling enabled in your browser? I was able to reproduce that with 110% zoom in Chrome and it seems like an error in rounding of background scale. I’ve modified a formula for bg size as a failsafe so it should be fine now, would be great if you could re-check.

I checked again as I'd took that from my personal laptop so hadn't made any changes and it wasn't zoomed when I took that screenshot (or at least I can still reproduce it without the changes and it's on the default 100% zoom). Unfortunately, I can still reproduce this with the latest changes as well 😞. The machine I'm using is a Macbook Pro running macOS 12.2.1 and Google Chrome 98.0.4758.102. I'll see if I can replicate it on some other set ups as well.

TheSpyder (approved):

TheSpyder:
Weird pixel changes such as you're describing can come up when using very thin lines - it depends whether the screen is retina or not. And it can definitely change depending on both operating system and browser.

TheSpyder:
If that’s the case the solution is to make the line thicker. It’s probably become a subpixel line which retina screens are able to display but not regular screens. I don’t know if there are media queries to tell the difference or if it just needs to be thicker for everyone.

noxuhax:
After almost losing my mind over this subpixel line (I was able to reproduce it myself as well), I figured out what’s the issue: apparently, the line only appears on demo pages with sans-serif headers (`font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen, Ubuntu, Cantarell, 'Open Sans', 'Helvetica Neue', sans-serif;` from content.css). The reason is page title’s size and margins getting calculated into a particularly unfortunate values (44.797 pixels tall plus two vertical margins of 21.44px) thus making editor’s top position 87.671875 pixels. Apparently, Chrome does not like that and draws a top half-pixel line of a pixel-tall toolbar divider.

To fix that, I’ve added a `transform` rule that visually does nothing, but forces toolbar to be drawn on pixel grid removing this half-line. Could you please verify that this issue is now fixed on your machines as well?

lnewson (approved):
Oh wow, so it was sub pixel rendering 😥  Nice work tracking that down and getting a fix for that! I can confirm I can no longer replicate the issue now either.

TheSpyder:
Ugh fractional pixels are the worst. I hate that there's no way to avoid them without weird tricks.

PR #7636, opened 2022-02-15
TINY-6559: Now exit the blockquote when pressing enter twice

HAFRMO:
Related Ticket:

Description of Changes:
Blockquotes should now be possible to exit by pressing shift+enter.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


lnewson:
@HAFRMO which version is this meant to target as you haven't set the milestone and the Jira fix version is missing? As this is a change/improvement by the looks, so we've missed the 6.0 freeze now so is this meant to be for 6.1 or do we need an exception for including this in 6.0?

Afraithe:
It's for 6.1, Jira updated.

modules/tinymce/src/core/main/ts/newline/InsertBlock.ts
```diff
@@ -242,7 +242,21 @@ const addBrToBlockIfNeeded = (dom, block) => {
   }
 };
 
-const insert = (editor: Editor, evt?: EditorEvent<KeyboardEvent>) => {
+const findParent = (editor: Editor, container: Node, stepUntilFinding?: string) => {
```
spocke, line 245: Maybe this could take a predicate instead that way this parameter wouldn't need to be optional. Then it would just be  `insertBlock(Fun.always)` and for the blockquote `insertBlock((node) => node.tagName === 'BLOCKQUOTE'` and then you could use PredicateFind.closest(SugarElement.fromDom(container), (node) => editor.dom.isBlock(node.dom) && predicate(node.dom)) and not need to have this `findDirectParent` fallback.

So the entire function becomes something like:
```
const findParent = (editor: Editor, container: Node, predicate: (node: DOMNode) => boolean) => {
  const isRoot = (el: SugarElement<Node>) => el.dom === editor.getBody();
  return PredicateFind.closest(SugarElement.fromDom(container), (node) => editor.dom.isBlock(node.dom) && predicate(node.dom), isRoot).map((element) => element.dom);
};
```

Also we have defined this isRoot so many times it starting to become ridiculous. Maybe we should have a `EditorBody.ts` (I could have sworn that we did have that at some point). That module could have a curried isRoot function so you could do `PredicateFind.closest(.. EditorBody.isRoot(editor))`.

Also if we want to marry these two worlds the sugar world and the dom world at some point I could have sworn I saw someone adding a `SugarElement.toDom` function that just unwraps the node like that. I think it was Josh maybe that was in some PR that never made it. Obviously we shouldn't overuse that but it would be useful here and there especially in mixed code like this.

modules/tinymce/src/core/main/ts/newline/InsertNewLine.ts
```diff
@@ -15,35 +15,39 @@ import * as InsertBr from './InsertBr';
...
       if (Type.isNonNullable(evt)) {
-        const event = fireFakeBeforeInputEvent(editor, 'insertLineBreak');
+        const event = fireFakeBeforeInputEvent(editor, 'insertParagraph');
         if (event.isDefaultPrevented()) {
           return;
         }
       }
 
-      InsertBr.insert(editor, evt);
+      InsertBlock.insert(editor, evt, stepUntilFinding);
```
spocke, line 27: Weird why are these flipped from InsertBr to InsertBlock on the adt the new case `newBlock` is added after the block so that shouldn't affect the ordering. This fold should still be (br, block, newBlock, none) maybe I'm not reading the code right here?

modules/tinymce/src/core/main/ts/newline/NewLineUtils.ts
```diff
@@ -104,6 +104,9 @@ const getEditableRoot = (dom, node) => {
   return parent !== root ? editableRoot : root;
 };
 
+const isBlockQuote = (editor: Editor) =>
+  SelectorFind.ancestor(SugarElement.fromDom(editor.selection.getStart(true)), 'BLOCKQUOTE', (element) => element.dom === editor.dom.getRoot()).isSome();
```
spocke, line 108: Since this is a css selector you could use a lowercase string here. Also if we move the root predicate here to that EditorBody.ts module I mentioned that I think we should have this would be less code. Also there is a `SelectorExists.ancestor` so you don't need the isSome check here.

modules/tinymce/src/core/test/ts/browser/newline/InsertNewLineTest.ts
```diff
@@ -162,4 +162,24 @@ describe('browser.tinymce.core.newline.InsertNewLineTest', () => {
     TinyAssertions.assertContent(editor, '<p><a href="#">a</a></p><p><a href="#"><img src="about:blank"></a></p>');
     TinyAssertions.assertSelection(editor, [ 1, 0 ], 0, [ 1, 0 ], 0);
   });
+
+  context('Blockquote manipulation', () => {
+    it('Press Enter in blockquote', () => {
```
spocke, line 167: Should these be prefixed with the ticket ID I never really understood if that should be on all changes so that we can track it back to the ticket or just bug fixes.

modules/tinymce/src/core/test/ts/browser/newline/InsertNewLineTest.ts
```diff
@@ -162,4 +162,24 @@ describe('browser.tinymce.core.newline.InsertNewLineTest', () => {
     TinyAssertions.assertContent(editor, '<p><a href="#">a</a></p><p><a href="#"><img src="about:blank"></a></p>');
     TinyAssertions.assertSelection(editor, [ 1, 0 ], 0, [ 1, 0 ], 0);
   });
+
+  context('Blockquote manipulation', () => {
```
spocke, line 166: Maybe drop the `manipulation` here since the whole test file is about inserting newlines and the special case here is `Blockquote`.

modules/tinymce/src/core/main/ts/newline/InsertNewLine.ts
HAFRMO: This is just git being confused when showing the differences. Since the two versions of InsertBlock were almost identical I moved the function out to a variable, but git then confuses it with the InsertBr-function that was first. The InsertBr-function is unchanged.

modules/tinymce/src/core/test/ts/browser/newline/InsertNewLineTest.ts
HAFRMO: The It-blocks should, you're right. That was my oversight.

modules/tinymce/src/core/main/ts/newline/InsertNewLine.ts
spocke: Ahh right yes when I view the whole file it looks correct. Thanks make sense now.

modules/tinymce/src/core/main/ts/newline/NewLineUtils.ts
HAFRMO: I'll leave creating the editorbody.ts to a separate issue, since I feel that when we do that we should change all the other ones in the code base.

modules/tinymce/CHANGELOG.md
```diff
@@ -92,6 +92,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Moved non-UI table functionality to core #TINY-8273
 - The `table_style_by_css` option has been set to `true` by default #TINY-8259
 - The `table_use_colgroups` option has been set to `true` by default #TINY-8259
+- Pressing Shift + Enter will now step out of `blockquote` elements to match `pre` element behavior #TINY-6559
```
lnewson, line 95: This needs rebasing on top of `next` since it's for 6.1 and this entry will need moving to the 6.1 unreleased section (as this section is for 6.0).

spocke (approved):

lnewson:
Is this really the behaviour we want? I've been testing this and you now no longer can create `<br>` elements inside a `<blockquote>`. I think a better behaviour would be to have this only run if it's at the end of a blockquote. Take for example the following HTML, if you place the caret at the end of the first line and press shift+enter it just splits the blockquote.

```
<blockquote>
<p>Line 1</p>
<p>Line 2</p>
</blockquote>
```

modules/tinymce/CHANGELOG.md
HAFRMO: I'll do this after the discussion that you requested.

lnewson:
I've converted this back to a draft since we're going to change the approach to solving this issue as discussed on Slack.

modules/tinymce/src/core/main/ts/newline/InsertBlock.ts
```diff
@@ -417,7 +434,7 @@ const insert = (editor: Editor, evt?: EditorEvent<KeyboardEvent>) => {
   }
 
   // Find parent block and setup empty block paddings
-  parentBlock = dom.getParent(container, dom.isBlock);
+  parentBlock = editor.dom.getParent(container, editor.dom.isBlock);
```
spocke, line 437: Any particular reason why this was changed this will no longer name mangle producing a slightly bigger minified js file?

modules/tinymce/src/core/test/ts/browser/newline/InsertNewLineTest.ts
```diff
@@ -162,4 +162,44 @@ describe('browser.tinymce.core.newline.InsertNewLineTest', () => {
     TinyAssertions.assertContent(editor, '<p><a href="#">a</a></p><p><a href="#"><img src="about:blank"></a></p>');
     TinyAssertions.assertSelection(editor, [ 1, 0 ], 0, [ 1, 0 ], 0);
   });
+
+  context('Blockquote', () => {
```
spocke, line 166: We need additional tests for the new option types. If it set to say `div, blockquote` you should be able to split divs and blockquotes but not other container elements. If it's set to false it should not split anything. If it's set to true it should split any container block. This should probably be in it's own context maybe name it `end_container_on _empty_block` or something. There are some tests for this in the `EnterKeyTest.ts` but these are in a older legacy format so I think it's better that we write new tests for this here instead.

modules/tinymce/CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Changed
+- Pressing Shift + Enter will now step out of `blockquote` elements to match `pre` element behavior #TINY-6559gp
```
spocke, line 10: This is no longer the case it needs to be updated to include the changes to the option and that the default is blockquote. Also this like has `gp` at the end of the text not sure what that is? Maybe it needs to be two log items since we are altering the option and also changing the default behavior to split blockquotes not sure what make sense.

spocke (changes requested):

modules/tinymce/src/core/main/ts/newline/InsertBlock.ts
HAFRMO: I was changing it manually, and missed this detail, that's all.

modules/tinymce/CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## 6.0.1 - TBD
 
+### Changed
+- The option`end_container_on_empty_block` can now take a string of blocks to split when pressing Enter twice #TINY-6559
+- The default value for option`end_container_on_empty_block` changed to `blockquote` #TINY-6559
```
lnewson, line 16: This should be in the `Unreleased` section of the changelog since it's for 6.1.0 instead of 6.0.1.

- [ ] Move to correct place

Also here's a suggestion for some minor edits:
```suggestion
### Changed
- The `end_container_on_empty_block` option can now take a string of blocks to split when pressing Enter twice #TINY-6559
- The default value for `end_container_on_empty_block` option has been changed to ` 'blockquote' ` #TINY-6559
```

modules/tinymce/src/core/main/ts/newline/InsertBlock.ts
```diff
@@ -1,4 +1,11 @@
-import { Arr, Obj, Optional, Optionals } from '@ephox/katamari';
+/**
+ * Copyright (c) Tiny Technologies, Inc. All rights reserved.
+ * Licensed under the LGPL or a commercial license.
+ * For LGPL see License.txt in the project root for license information.
+ * For commercial licenses see https://www.tiny.cloud/
+ */
+
```
lnewson, line 7:
```suggestion
```
This looks like a merge commit that's gone bad, as this header is no longer needed.

modules/tinymce/src/core/main/ts/newline/InsertBlock.ts
```diff
@@ -235,6 +242,15 @@ const addBrToBlockIfNeeded = (dom, block) => {
   }
 };
 
+const shouldEndContainer = (editor: Editor, container: Node) => {
```
lnewson, line 245:
```suggestion
const shouldEndContainer = (editor: Editor, container: Node | undefined) => {
```
This should be this if we have the nullable check and looking at how it's defined, it looks like the case we need to protect against is `undefined`.

modules/tinymce/src/core/main/ts/newline/InsertBlock.ts
```diff
@@ -235,6 +242,15 @@ const addBrToBlockIfNeeded = (dom, block) => {
...
+  const optionValue = Options.shouldEndContainerOnEmptyBlock(editor);
+  if (Type.isString(optionValue)) {
+    return Type.isNonNullable(container) && Arr.contains(optionValue.split(/, ?/), container.nodeName.toLowerCase());
+  } else {
+    return Type.isNonNullable(container) && optionValue;
+  }
+};
```
lnewson, line 252:
```suggestion
const shouldEndContainer = (editor: Editor, container: Node | undefined) => {
  const optionValue = Options.shouldEndContainerOnEmptyBlock(editor);
  if (Type.isNullable(container)) {
    return false;
  } else if (Type.isString(optionValue)) {
    return Arr.contains(Tools.explode(optionValue), container.nodeName.toLowerCase());
  } else {
    return optionValue;
  }
};
```
I believe this is a little easier to read and avoids the duplication of the `Type.isNonNullable` checks. I also swapped to use `Tools.explode` as thats the more consistent API rather than manually splitting the string.

modules/tinymce/src/core/main/ts/newline/NewLineUtils.ts
```diff
@@ -97,6 +97,9 @@ const getEditableRoot = (dom, node) => {
   return parent !== root ? editableRoot : root;
 };
 
+const isBlockQuote = (editor: Editor) =>
```
lnewson, line 100: This doesn't look to be used at anymore? Should this still be here?

- [ ] Consider removing this unused code

modules/tinymce/src/core/test/ts/browser/newline/InsertNewLineTest.ts
```diff
@@ -162,4 +162,335 @@ describe('browser.tinymce.core.newline.InsertNewLineTest', () => {
...
+    it('TINY-6559: Press Enter in blockquote without wrapping p', () => {
+      const editor = hook.editor();
+      editor.setContent('<blockquote>Line 1</blockquote>');
+      TinySelections.setCursor(editor, [ 0 ], 1);
+      insertNewline(editor, { });
+      TinyAssertions.assertContent(editor, '<blockquote>Line 1</blockquote><p>&nbsp;</p>');
+      TinyAssertions.assertSelection(editor, [ 1 ], 0, [ 1 ], 0);
+    });
+
+    context('Has default value', () => {
```
lnewson, line 176:
```suggestion
    context('With the default value', () => {
```

lnewson (approved):

modules/tinymce/src/core/main/ts/newline/InsertBlock.ts
```diff
@@ -1,10 +1,11 @@
-import { Arr, Obj, Optional, Optionals } from '@ephox/katamari';
+import { Arr, Obj, Optional, Optionals, Type } from '@ephox/katamari';
 import { Css, PredicateFilter, SugarElement, SugarNode } from '@ephox/sugar';
 
 import DOMUtils from '../api/dom/DOMUtils';
 import DomTreeWalker from '../api/dom/TreeWalker';
 import Editor from '../api/Editor';
 import * as Options from '../api/Options';
+import { Tools } from '../api/PublicApi';
 import { EditorEvent } from '../api/util/EventDispatcher';
```
lnewson, line 9:
```suggestion
import { EditorEvent } from '../api/util/EventDispatcher';
import Tools from '../api/util/Tools';
```
This was importing from the wrong place. We really should look at adding a lint rule for this.

modules/tinymce/src/core/main/ts/newline/NewLineUtils.ts
```diff
@@ -1,5 +1,5 @@
 import { Fun, Optional, Unicode } from '@ephox/katamari';
-import { SugarElement } from '@ephox/sugar';
+import { SelectorExists, SugarElement } from '@ephox/sugar';
```
lnewson, line 2:
```suggestion
import { SugarElement } from '@ephox/sugar';
```
This is why the LGTM warning is occurring.

modules/tinymce/CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Changed
+- The `end_container_on_empty_block` option can now take a string of blocks to split when pressing Enter twice #TINY-6559
+- The default value for `end_container_on_empty_block` option has been changed to ` 'blockquote' ` #TINY-6559
```
```suggestion
- The default value for `end_container_on_empty_block` option has been changed to `'blockquote'` #TINY-6559
```
Just a minor formatting issue here.

ltrouton (approved):

spocke (approved):

PR #7638, opened 2022-02-17
Reorganising the changelog for clarity plus minor edits

metricjs:
Related Ticket: N/A - general 6.0 release prep

Description of Changes:
* Reorganised the changelog entries, trying to group by what kind of change (e.g. renaming, removing, deprecating) or what they affect (e.g. APIs, commands, events). Made a few exceptions, e.g. grouped moved plugins + their related setting renames. 
* Edited some entries for consistency e.g. "Renamed x to y" instead of a mix of that and "x was renamed to y", etc.
** Unsure about the wording I ended up with for deprecated items. Opinions welcome.

Sorry the diff might be a bit confusing.

Pre-checks:
* [x] Changelog entry added
* [x] Tests have been added (if applicable) - N/A

Review:
* [x] Milestone set


modules/tinymce/CHANGELOG.md
```diff
@@ -8,113 +8,134 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- Removed the `tinymce.utils.Promise` API #TINY-8241
+- Removed the `toHex` function for the `DOMUtils` and `Styles` APIs #TINY-8163
+- Removed the `execCommand` handler function from the plugin and theme interfaces #TINY-7829
+- Removed the `editor.settings` property as it has been replaced by the new Options API #TINY-8236
+- Removed the `shortEnded` and `fixed` properties on `tinymce.html.Node` class #TINY-8205
+- Removed the `mceInsertRawHTML` command #TINY-8214
+- Removed the style field from the `image` plugin dialog advanced tab #TINY-3422
+- Removed the legacy `mobile` theme #TINY-7832
+- Removed the legacy browser detection properties from `Env` #TINY-8162
+- Removed support for Microsoft Internet Explorer 11 #TINY-8194 #TINY-8241
```
lnewson, line 132: Should this be up the top since it's one of the key things we need to highlight?

metricjs: I'd be fine with that. Can do

metricjs: Done

modules/tinymce/CHANGELOG.md
```diff
@@ -8,113 +8,134 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
-- New `AutocompleterStart`, `AutocompleterUpdate` and `AutocompleterEnd` events #TINY-8279
-- New `mceAutocompleterClose`, `mceAutocompleterReload` commands #TINY-8279
-- New `FakeClipboard` API on the `tinymce` global #TINY-8353
-- New `slider` dialog component #TINY-8304
-- New `buttonType` property on dialog button components, supporting `toolbar` style in addition to `primary` and `secondary` #TINY-8304
-- New `imagepreview` dialog component, allowing preview and zoom of any image URL #TINY-8333
 - New `editor.annotator.removeAll` API to remove all annotations by name #TINY-8195
 - New `Resource.unload` API to make it possible to unload resources #TINY-8431
+- New `FakeClipboard` API on the `tinymce` global #TINY-8353
 - New `dispatch()` function to replace the now deprecated `fire()` function in various APIs #TINY-8102
```
EkimChau, line 14: Should this be "change" instead of New? dispatch() replaces fire()

lnewson: We didn't remove the `fire` function so the `dispatch` function is actually new.

metricjs: Hmm, true. I think it was basically a rename? Not entirely sure

lnewson: It sort of was but largely wasn't, as the change added a new function and then replaced all internal uses to use the new name, instead of the deprecated one. Then as a last step it deprecated the old function name, which is why there's an entry in Deprecated for that as well.

metricjs: Right, makes sense. We should probably keep it as it is then

modules/tinymce/CHANGELOG.md
```diff
@@ -8,152 +8,153 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 - Dialog labels and other text-based UI properties did not escape HTML markup #TINY-7524
-- Deleting content would sometimes not fire `beforeinput` and `input` events as expected #TINY-8168 #TINY-8329
 - Anchor elements would render incorrectly when using the `allow_html_in_named_anchor` option #TINY-3799
-- Fixed sub-menu items not read by screen readers. Patch contributed by westonkd #TINY-8417
 - The Home or End keys would move out of a editable element contained within a noneditable element #TINY-8201
 
 ### Removed
-- Removed the jQuery integration #TINY-4518
+- Removed support for Microsoft Internet Explorer 11 #TINY-8194 #TINY-8241
+- Removed support for Word from the `paste` plugin #TINY-7493
```
lnewson, line 114: Does this need rewording now that we no longer have a paste plugin?

modules/tinymce/CHANGELOG.md
```diff
@@ -8,152 +8,153 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
-- Renamed the `getShortEndedElements` Schema API to `getVoidElements` #TINY-8344
-- Changed the default statusbar element path delimiter from `»` to `›` #TINY-8372
+- The `GetContent` event will now always pass a `string` for the `content` property #TINY-7996
 - Changed the default tag for the strikethrough format to the `s` tag when using a html 5 schema #TINY-8262
 - The `strike` tag is automatically converted to the `s` tag when using a html 5 schema #TINY-8262
+- Aligning a table to the left or right will now use margin styling instead of float styling #TINY-6558
+- The `:` control character has been changed to `~` for the schema `valid_elements` and `extended_valid_elements` options #TINY-6726
+- The `primary` property on dialog buttons has been deprecated. Use the new `buttonType` property instead #TINY-8304
+- Changed the default statusbar element path delimiter from `»` to `›` #TINY-8372
+- Replaced 'Powered by Tiny' link text with logo #TINY-8371
```
lnewson, line 70:
```suggestion
- Replaced the `Powered by Tiny` link text with the Tiny logo #TINY-8371
```
Backticks instead of quotes for this maybe? Also I don't think this fully made sense, so made a minor change there.

modules/tinymce/CHANGELOG.md
```diff
@@ -8,152 +8,153 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
 ### Improved
+- The default height of editor has been increased from `200px` to `400px` to improve the usability of the editor #TINY-6860
+- The upload results returned from the `editor.uploadImages()` API now includes a `removed` flag, reflecting if the image was removed after a failed upload #TINY-7735
 - The `ScriptLoader`, `StyleSheetLoader`, `AddOnManager`, `PluginManager` and `ThemeManager` APIs will now return a `Promise` when loading resources instead of using callbacks #TINY-8325
 - A `ThemeLoadError` event is now fired if the theme fails to load #TINY-8325
-- The upload results returned from the `editor.uploadImages()` API now includes a `removed` flag, reflecting if the image was removed after a failed upload #TINY-7735
+- The `BeforeSetContent` event will now include the actual serialized content when passing in an `AstNode` to the `editor.setContent` API #TINY-7996
+- Improved support for placing the caret before or after noneditable elements within the editor #TINY-8169
+- Calls to `editor.selection.setRng` now update the cursor position bookmark used when focus is returned to the editor #TINY-8450
```
lnewson, line 30:
```suggestion
- Calls to `editor.selection.setRng` now update the caret position bookmark used when focus is returned to the editor #TINY-8450
```
Just noticed this uses the wrong term, as cursor is the mouse pointer.

lnewson (approved):

modules/tinymce/CHANGELOG.md
metricjs: I honestly can not think of a good way to reword this to clarify that

lnewson: Yeah nor can I to be honest. The best I could come up with was something like this... Let's leave it as is then if you don't like the suggestion or would prefer to leave it as is.
```suggestion
- Removed support for Word from the opensource paste functionality #TINY-7493
```

TheSpyder: I suggest `Microsoft Word` instead of just `Word`

TheSpyder (approved):

modules/tinymce/CHANGELOG.md
lnewson: That's a good point, yeah it should definitely be the full product name. I'll update my suggestion to include that.

metricjs: I'll go with this, thanks!

PR #7643, opened 2022-02-21
TINY-6536: The Shadowdom appears to distort menu element sizes

spocke:
Related Ticket:

Description of Changes:
* Did a lot of experiments trying to get a proper workaround for this issue. But the basic reality is that we are trying to fight the browsers rendering model. We are trying to bust out of the constrains of the viewport that a shadowdom is designed to provide. The only way I can think of that lets you go from a constrained in element position/size system to a viewport position/size system is to use position fixed. So this PR adds position fixed to the sink so that all elements gets positioned/size constrained relative the the viewport not relative to any parent container it might have. It does so only when rendered inside shadow dom since using position fixed forces scrolling UI elements to be required to be positioned by code. If we ever change the sink to be appended relative to where the editor is being rendered and not the body we will get a similar problem as we have with shadwo dom today.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/tinymce/src/themes/silver/main/ts/Render.ts
```diff
@@ -229,12 +230,13 @@ const setup = (editor: Editor): RenderInfo => {
         tag: 'div',
         classes: [ 'tox', 'tox-silver-sink', 'tox-tinymce-aux' ].concat(deviceClasses),
         attributes: {
-          ...I18n.isRtl() ? { dir: 'rtl' } : {}
+          ...I18n.isRtl() ? { dir: 'rtl' } : {},
+          style: isInShadowRoot ? 'top: 0; left: 0; right: 0; position: fixed' : ''
```
TheSpyder, line 234: Alloy should have a way to do this already, it shouldn't need custom styles?

TheSpyder (commented):
`position: fixed` may work around these specific problems but it's not a great user experience. Menus no longer move as the page scrolls (and attempts to update positions in response to scroll lead to jittering).

We also found on TBIO that page styles can reset the origin inside where the menu is trying to measure resulting in wildly incorrect position calculations.

lnewson:
> `position: fixed` may work around these specific problems but it's not a great user experience. Menus no longer move as the page scrolls (and attempts to update positions in response to scroll lead to jittering).
> 
> We also found on TBIO that page styles can reset the origin inside where the menu is trying to measure resulting in wildly incorrect position calculations.

I've been meaning to try this because I expected that to happen... generally the sink should only ever be fixed if the rest of the editor is also fixed. As such, I definitely don't think this is something we can enable by default and would have to be an opt-in option if we want to go ahead with it.

As a side note, we're going to have to figure something out for this, as this same issue is going to happen once we look at scrollable divs, etc... so we'll need to keep that in mind as well. I think to fully solve this we need to look at instead adjusting where the menu's are attached and what their bounds are. As an example, one of the issues linked is the dialog listbox menu is distorted because it doesn't attach inside the dialog container which is fixed position, so if it was attached there it wouldn't have the overflow clipping issues. As for the menus, to solve that we probably need to have a separate sink beside the editor and potentially adjust the bounds to be the shadow root/container instead so it doesn't try to overflow outside them and cause overflow issues.

TheSpyder:
> As a side note, we're going to have to figure something out for this, as this same issue is going to happen once we look at scrollable divs, etc...

yes I half expect we will need a mode that either abandons the sink or puts the sink inside the editor container. We can't do this by default, long menus will get cut off by `overflow: scroll` boundaries, but it definitely needs to be an option.

lnewson:
> yes I half expect we will need a mode that either abandons the sink or puts the sink inside the editor container. We can't do this by default, long menus will get cut off by `overflow: scroll` boundaries, but it definitely needs to be an option.

Yeah, I've long suspected we're going to effectively need 2 different sinks or something similar... one for modal UI whereby it must be attached to the body and another for things that attach directly to the editor.

modules/tinymce/src/themes/silver/main/ts/Render.ts
spocke: I couldn't find anything that sets styles to position fixed and 100% viewport width. The isFixed either sets position relative or no styles at all. Maybe I missed something.

spocke:
I know about the jitter on scroll and this needs to be an option eventually. Since different people will want different things depending on where the editor is rendered. If it's rendered in a scrolling container you might want to constrain and scroll it within that container if it's in a container only used for positioning your editor you might want it to work like it does now with a fixed sink with a viewport constrain instead.
After trying many hacky things to try to avoid the browser from constraining the menus to the parent container this is the best thing I could come up with any other ideas are welcome. But these hacks felt more like trying to fight what the browser is designed to do in this case and that is to constrain everything inside the element.
For shadow dom we shouldn't put anything outside of the shadow root (since that is the whole point) so having a separate body based sink is not really an option but I guess we could have a fixed sink and a relative sink inside it instead but that would still mean menu jitter.

modules/tinymce/src/themes/silver/main/ts/Render.ts
ephox-mogran: I think what Andy means is that alloy doesn't tend to use `attribute` to set `style`.

ephox-mogran: Actually, no .. that's not what he meant ... but it's what I mean :)

ephox-mogran: Hmm. I thought it would set position: fixed, but maybe it just assumed that the integrator would do that. I'll check.

Yeah, looks like it's been that way for ages.

PR #7646, opened 2022-02-21
TINY-8486: Fix DataTransfer image check in DragDrop.ts

ltrouton:
Related Ticket: TINY-8486

Description of Changes:
* Fix regex to not include `data:` as that is never part of an image `file.type`
* Add a test as I couldn't find any existing ones

Pre-checks:
* [X] Changelog entry added

Review:
* [X] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -110,7 +110,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Anchor elements would render incorrectly when using the `allow_html_in_named_anchor` option #TINY-3799
 - The `AstNode` HTML serializer did not serialize `pre` or `textarea` elements correctly when they contained newlines #TINY-8446
 - Fixed sub-menu items not read by screen readers. Patch contributed by westonkd #TINY-8417
-- Drag and drop would not properly detect images and respect the `paste_data_images` option in some cases #TINY-8486
 - The Home or End keys would move out of a editable element contained within a noneditable element #TINY-8201
 - Dialogs could not be opened in inline mode before the editor had been rendered #TINY-8397
 
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -110,6 +110,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Anchor elements would render incorrectly when using the `allow_html_in_named_anchor` option #TINY-3799
 - The `AstNode` HTML serializer did not serialize `pre` or `textarea` elements correctly when they contained newlines #TINY-8446
 - Fixed sub-menu items not read by screen readers. Patch contributed by westonkd #TINY-8417
+- Drag and drop would not properly detect images and respect the `paste_data_images` option in some cases #TINY-8486
 - The Home or End keys would move out of a editable element contained within a noneditable element #TINY-8201
 - Dialogs could not be opened in inline mode before the editor had been rendered #TINY-8397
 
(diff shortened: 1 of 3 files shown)


modules/tinymce/src/core/test/ts/browser/paste/DragDropImageTest.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    return new window.File([ bytes ], filename, { type });
+  };
+
+  const fireDropEvent = (editor: Editor, files: File[]): DragEvent => {
+    const data = {
+      dataTransfer: { files }
+    } as any;
+    return editor.fire('drop', data);
+  };
```
lnewson, line 35: This is a very old way of doing this and not something we should be doing anymore. Please swap this to use the Agar `DragnDrop` module instead.

- [ ] Change this so we aren't manually firing events via the editor

modules/tinymce/src/core/test/ts/browser/paste/DragDropImageTest.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  const pWaitForSelector = (editor: Editor, selector: string) =>
+    Waiter.pTryUntilPredicate(`Wait for ${selector} to exist`, () => editor.dom.select(selector).length > 0);
```
lnewson, line 38: This should probably be using something like this instead? It could also be inlined as we only use this once.
```ts
  const pWaitForSelector = (editor: Editor, selector: string) =>
    UiFinder.pWaitFor(`Wait for ${selector} to exist`, TinyDom.body(), selector);
```

tiny-ben-tran (approved):

modules/tinymce/src/core/test/ts/browser/paste/DragDropImageTest.ts
```diff
@@ -0,0 +1,84 @@
...
+      base_url: '/project/tinymce/js/tinymce'
+    });
+
+    editor.setContent('<p>a</p>');
+    TinySelections.setCursor(editor, [ 0, 0 ], 0);
+
+    const event = fireDropEvent(editor, [
+      base64ToBlob(base64ImgSrc, 'image/gif', 'image.gif')
+    ]);
+    assert.isTrue(event.defaultPrevented);
```
lnewson, line 53: This is duplicated between the 2 tests, can we remove the duplication here please?

lnewson (approved):

modules/tinymce/src/core/test/ts/browser/paste/DragDropImageTest.ts
ltrouton: Fixed

modules/tinymce/src/core/test/ts/browser/paste/DragDropImageTest.ts
ltrouton: Ah, I forgot about that. Thanks. Hopefully this is good now.

modules/tinymce/CHANGELOG.md
```diff
@@ -110,6 +110,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Anchor elements would render incorrectly when using the `allow_html_in_named_anchor` option #TINY-3799
 - The `AstNode` HTML serializer did not serialize `pre` or `textarea` elements correctly when they contained newlines #TINY-8446
 - Fixed sub-menu items not read by screen readers. Patch contributed by westonkd #TINY-8417
+- Drag and drop would not properly detect images and respect the `paste_data_images` option in some cases #TINY-8486
```
lnewson, line 113: Was this broken in 5.x? I believe this is something that got broken in the 6.0 work, so in that case we shouldn't have a changelog.

lnewson: For reference here's the old logic: https://github.com/tinymce/tinymce/blob/master/modules/tinymce/src/plugins/paste/main/ts/core/DragDrop.ts#L47. As you can see it didn't filter on images at all, which was something we tried to fix IIRC. So I'm pretty sure this wasn't a bug in 5.x.

TheSpyder (approved):

PR #7647, opened 2022-02-22
TINY-2316: Replace LGPL license with MIT

TheSpyder:
Related Ticket: TINY-2316

Description of Changes:
* TinyMCE is now MIT licensed!
* A lot of search/replace. File headers have been removed as a part of this - apologies for the mammoth PR as a result. If we find this too difficult I can try to rewrite the commits so the file headers change in one and everything else in another. From what I can see the source code is grouped together in one section of the PR list, so hopefully it's not too bad.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/tinymce/README.md
```diff
@@ -60,12 +60,12 @@ TinyMCE is an open source software project and we encourage developers to contri
 
 __Basic Rules__
 
-* Contributed code will be licensed under the LGPL license but not limited to LGPL
+* Contributed code will be licensed under the MIT license
```
tiny-james, line 63: Do we need the " but not limited to MIT" bit?

TheSpyder: I'm not sure. There are a few parts of this change that weren't fully specced out.

tiny-james (approved):

modules/tinymce/README.md
tiny-james: I assume it was added for our commercial users in the case of LGPL so we could re-license it to suite them. So the question would be if they'd be happy with MIT being the only option.

EkimChau (approved):

spocke (approved):

modules/tinymce/README.md
TheSpyder: I've thought about this and it seems more about making clear to contributors that code we accept may be relicensed and not released as LGPL (i.e. our commercial release).

The MIT license allows commercial use so I think "not limited to MIT" is unnecessary.

PR #7651, opened 2022-02-23
TINY-2316: Corrected company name

TheSpyder:
Related Ticket: TINY-2316

Description of Changes:
* This was missed in #7647. The company name in our license txt is different to what we've been using in the package.json files:
https://github.com/tinymce/tinymce/blob/236891ffcdf2fd1b698d020e67c31838fb110019/LICENSE.TXT#L3

This PR updates everywhere we use the company name to the proper full string.

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [x] Milestone set

### DIFF
--- a/modules/acid/package.json
+++ b/modules/acid/package.json
@@ -14,7 +14,7 @@
     "@ephox/sugar": "^9.0.0-alpha.5",
     "tslib": "^2.0.0"
   },
-  "author": "Tiny Technologies, Inc",
+  "author": "Ephox Corporation DBA Tiny Technologies, Inc",
   "license": "MIT",
   "files": [
     "lib/main",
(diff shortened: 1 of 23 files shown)


spocke (approved):

lnewson (approved):
Does this actually need changing? as this isn't a legal field IIUC, in which case the old name was probably fine 🤷  Then again I don't know for sure so it also can't hurt.

TheSpyder:
> Does this actually need changing? as this isn't a legal field IIUC, in which case the old name was probably fine 🤷 Then again I don't know for sure so it also can't hurt.

I tried to make the case and was told to change it anyway. There are probably a bunch of places we use the old name that will be lurking around for ages.

PR #7451, opened 2021-12-08
TINY-8163: Moved toHex APIs from tinymce to acid

thedrjim:
Related Ticket:

Description of Changes:
* lower case changes and updates to paste tests 

Pre-checks:
* [x] Changelog entry added
* [x] Tests have been added (if applicable)
* [x] Branch prefixed with `feature/` for new features (if applicable)

Review:
* [x] Milestone set
* [x] Review comments resolved

GitHub issues (if applicable):

PR #7655, opened 2022-02-24
TINY-8525: Fixed a regression with webkit style handling while pasting content

lnewson:
Related Ticket: TINY-8525

Description of Changes:

Fixes a regression introduced in https://github.com/tinymce/tinymce/pull/7451/ as the string value passed to `rgbToHex` can be undefined if the styles attribute doesn't have the configured webkit styles to retain. I've largely just made it do what it did previously to solve this, so that it both cleans the value and handles the undefined value since `Tools.trim` will convert a nullable value to an empty string. I also fixed the incorrect types for `Tools.trim` to reflect it takes nullable values.

Pre-checks:
* [x] ~Changelog entry added~ 6.0 only issue, so no changelog

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/test/ts/browser/paste/PasteStylesTest.ts
+++ b/modules/tinymce/src/core/test/ts/browser/paste/PasteStylesTest.ts
@@ -67,4 +67,13 @@ describe('browser.tinymce.core.paste.PasteStylesTest', () => {
     Clipboard.pasteItems(TinyDom.body(editor), { 'text/html': '<span style="color: rgb(224, 62, 45);">b</span>' });
     TinyAssertions.assertContent(editor, `<p><span style="color: rgb(224, 62, 45);">b</span></p>`);
   });
+
+  it('TINY-8525: paste span without any color styles, paste_webkit_styles: color,font-family', () => {
+    const editor = hook.editor();
+    editor.options.set('paste_webkit_styles', 'color,font-family');
+    editor.setContent('<p>test</p>');
+    TinySelections.setSelection(editor, [ 0, 0 ], 0, [ 0, 0 ], 4);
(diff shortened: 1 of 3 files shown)


TheSpyder (approved):

modules/tinymce/src/core/test/ts/browser/paste/PasteStylesTest.ts
```diff
@@ -67,4 +67,13 @@ describe('browser.tinymce.core.paste.PasteStylesTest', () => {
     Clipboard.pasteItems(TinyDom.body(editor), { 'text/html': '<span style="color: rgb(224, 62, 45);">b</span>' });
     TinyAssertions.assertContent(editor, `<p><span style="color: rgb(224, 62, 45);">b</span></p>`);
   });
+
+  it('TINY-8525: paste span without any color styles, paste_webkit_styles: color,font-family', () => {
+    const editor = hook.editor();
+    editor.options.set('paste_webkit_styles', 'color,font-family');
```
ltrouton, line 73: Should we unset this at the end of the test?

ltrouton (approved):

modules/tinymce/src/core/test/ts/browser/paste/PasteStylesTest.ts
lnewson: There's a hook that does that: https://github.com/tinymce/tinymce/blob/feature/TINY-8525/modules/tinymce/src/core/test/ts/browser/paste/PasteStylesTest.ts#L23. Though looking at that it's got the wrong option name 🤦

lnewson: Have fixed the incorrect option name

ltrouton: Ah right 👍

PR #7670, opened 2022-03-04
TINY-8533: Adjust naked button and notifications contrast

noxuhax:
Related Ticket: TINY-8533

Description of Changes:
* Adjusted contrast for Naked Button component on darker backgrounds
* Adjusted contrast in Notifications component

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -9,8 +9,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Fixed the dev ZIP missing the required `bin` scripts to build from the source #TINY-8542
 - Fixed a regression whereby text patterns couldn't be updated at runtime #TINY-8540
-
-### Improved
 - Naked buttons better adapt to various background colors, improved text contrast in notifications #TINY-8533
 
 ## 6.0.0 - 2022-03-03
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -10,6 +10,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed the dev ZIP missing the required `bin` scripts to build from the source #TINY-8542
 - Fixed a regression whereby text patterns couldn't be updated at runtime #TINY-8540
 
+### Improved
+- Naked buttons better adapt to various background colors, improved text contrast in notifications #TINY-8533
+
 ## 6.0.0 - 2022-03-03
 
 ### Added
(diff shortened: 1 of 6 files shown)


tiny-ben-tran (approved):

TheSpyder (approved):

lnewson:
This appears to still have some contrast issues in with these changes, is that expected (both in dark and default skins)? According to WCAG standards (e.g [G195](https://www.w3.org/WAI/WCAG21/Techniques/general/G195)) it should have a 3:1 contrast ratio and this is well below that.
image
also for reference what it looked like before these changes, so this does improve things:
image

As a side note, I thought all these buttons were supposed to be changed to use the same blueish background color for the hover/active styles as the toolbars. What ever happened to that as that would potentially help with the contrast ratios?

noxuhax:
@lnewson though not ideal, new Naked buttons appearance works better with various background colors of all notification types, both from contrast and visual perspectives. After discussion with Paul we decided that there is no simple way to improve contrast for this buttons that would be consistent with other UI elements. 

There is a better solution that would also affect Primary and Secondary buttons as well as Textfields making focus state more obvious since it is currently not so clear for other types of buttons. For this, I’ll made a separate Jira since it branches too far away from the initial issue of the X button in notifications.

lnewson:
Sounds good, thanks @noxuhax 👍

lnewson (approved):

lnewson:
@noxuhax looking at the Jira this is meant to be included in 6.0.1, that means it should instead be based off/targeting `master` instead of `develop`. Could you please rebase that, update the PR target branch and also add a changelog entry so it can be merged? Let us know if you want us to do that instead.

noxuhax:
> Could you please rebase that, update the PR target branch and also add a changelog entry so it can be merged?

Done and done

modules/tinymce/CHANGELOG.md
```diff
@@ -10,6 +10,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed the dev ZIP missing the required `bin` scripts to build from the source #TINY-8542
 - Fixed a regression whereby text patterns couldn't be updated at runtime #TINY-8540
 
+### Improved
+- Naked buttons better adapt to various background colors, improved text contrast in notifications #TINY-8533
```
```suggestion
- Fixed a regression whereby text patterns couldn't be updated at runtime #TINY-8540
- Naked buttons better adapt to various background colors, improved text contrast in notifications #TINY-8533
```
Improvements can't go in a patch release and the Jira is a bug, so this should be under the Fixed section

lnewson: Merging this suggestion so that we can get this merged today.

PR #7678, opened 2022-03-09
TINY-8540: Fixed text patterns not able to be updated at runtime

lnewson:
Related Ticket: TINY-8540

Description of Changes:

Fixed a regression in functionality from 5.x whereby text patterns used to be able to be updated at runtime via the `plugins.textpattern.setPatterns()` API. While we don't want to bring that API back, as this is needed for connections our concession here was to make it so the option value can be updated at runtime and the text pattern logic will instead use the updated value. As such this moves the logic around a little so the options processor does the normalization now instead of doing it separately and we then use the normalized option value when matching the patterns.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/test/ts/browser/textpatterns/TextPatternsTest.ts
+++ b/modules/tinymce/src/core/test/ts/browser/textpatterns/TextPatternsTest.ts
@@ -213,4 +213,17 @@ describe('browser.tinymce.core.textpatterns.TextPatternsTest', () => {
     TinyContentActions.keyup(editor, Keys.space());
     TinyAssertions.assertContent(editor, '<p><a href="about:blank"><span class="test">www</span>.google.com</a></p>');
   });
+
+  it('TINY-8540: should be able to add additional patterns (connections requirement)', () => {
+    const editor = hook.editor();
+    Utils.setContentAndPressSpace(editor, 'brb');
+    TinyAssertions.assertContent(editor, '<p>brb&nbsp;</p>');
+
+    editor.options.set('text_patterns', [
+      ...editor.options.get('text_patterns'),
+      { start: 'brb', replacement: 'be right back' }
+    ]);
+    Utils.setContentAndPressSpace(editor, 'brb');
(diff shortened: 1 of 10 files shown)


TheSpyder (approved):

ltrouton (approved):

modules/tinymce/src/core/test/ts/browser/textpatterns/TextPatternsTest.ts
```diff
@@ -213,4 +213,17 @@ describe('browser.tinymce.core.textpatterns.TextPatternsTest', () => {
     TinyContentActions.keyup(editor, Keys.space());
     TinyAssertions.assertContent(editor, '<p><a href="about:blank"><span class="test">www</span>.google.com</a></p>');
   });
+
+  it('TINY-8540: should be able to add additional patterns (connections requirement)', () => {
```
spocke, line 217: Could you unset the option as well so that the default gets set back if someone adds a test after this.

- [ ] Unset option so that the default sets restored for other tests

spocke: Also not sure we need to mention that this is for a particular product that is an internal concern and can be tracked in JIRA? We don't generally do this for other things if it's customer specific fixes etc.

spocke (approved):
Very nice cleanup as well. Good work.

modules/tinymce/src/core/test/ts/browser/textpatterns/TextPatternsTest.ts
lnewson: Yeah fair point, I was mainly trying to make it very clear why this exists in case someone looks at it in a couple years as they're less likely to track it back then.

Also good catch on resetting the option value.

lnewson: Have removed the reference and added the `unset` call at the end.

PR #7681, opened 2022-03-11
TINY-8552: Fixed the autocompleter failing to start and close properly in some cases

lnewson:
Related Ticket: TINY-8552

Description of Changes:

This fixes an issue whereby the autocompleter wouldn't start correctly which in turn meant the UI state believed the autocompleter wasn't active and therefore wouldn't close the menu. This was largely only an issue when using a throttled fetch function such that the first `fetch` call never ends up resolving.

I've also fixed a very minor regression I noticed in the `esc` key handling as in 5.x it would cancel even when a fetch was running but that's no longer the case and with the 6.x changes it'd only cancel if the UI had been triggered (e.g. the fetch had completed). I'm happy to revert this if we don't think this should happen, but it made sense to me that if you press escape you don't want the menu to pop up after.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/test/ts/browser/keyboard/AutocompleterTest.ts
+++ b/modules/tinymce/src/core/test/ts/browser/keyboard/AutocompleterTest.ts
@@ -1,6 +1,7 @@
 import { Keys, Waiter } from '@ephox/agar';
 import { describe, it } from '@ephox/bedrock-client';
-import { Arr } from '@ephox/katamari';
+import { InlineContent } from '@ephox/bridge';
+import { Arr, Throttler } from '@ephox/katamari';
 import { TinyContentActions, TinyHooks, TinySelections } from '@ephox/wrap-mcagar';
 import { assert } from 'chai';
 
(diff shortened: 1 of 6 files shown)


modules/tinymce/src/core/test/ts/browser/keyboard/AutocompleterTest.ts
```diff
@@ -9,64 +10,92 @@ import { EditorEvent } from 'tinymce/core/api/PublicApi';
...
     }
   }, []);
 
-  const pOpenAutocompleter = async (editor: Editor) => {
+  const pOpenAutocompleter = async (editor: Editor, triggerChar: string) => {
     editor.focus();
     editor.setContent(`<p>${triggerChar}</p>`);
-    TinySelections.setCursor(editor, [ 0, 0 ], triggerChar.length);
-    TinyContentActions.keypress(editor, triggerChar.charCodeAt(0));
+    TinySelections.setCursor(editor, [ 0, 0 ], plusTriggerChar.length);
```
ltrouton, line 57: Should this be `tiggerChar` instead and for below as well?
```suggestion
    TinySelections.setCursor(editor, [ 0, 0 ], triggerChar.length);
```

ltrouton (approved):

modules/tinymce/src/core/test/ts/browser/keyboard/AutocompleterTest.ts
lnewson: Yes it should be thanks!

modules/tinymce/src/core/test/ts/browser/keyboard/AutocompleterTest.ts
```diff
@@ -9,64 +10,92 @@ import { EditorEvent } from 'tinymce/core/api/PublicApi';
...
+    TinyContentActions.keypress(editor, plusTriggerChar.charCodeAt(0));
```
lnewson, line 58:
```suggestion
    TinySelections.setCursor(editor, [ 0, 0 ], triggerChar.length);
    TinyContentActions.keypress(editor, triggerChar.charCodeAt(0));
```

TheSpyder (approved):

spocke:
Thanks for fixing this one. I guess my attempt to avoid having two mutating states didn't really work for slow or broken promise calls. I guess in theory this could be modeled as some form of single three state thing if we really wanted to avoid having this extra cell.

PR #7683, opened 2022-03-11
TINY-8534: Add transparent prop to dialog iframe component

noxuhax:
Related Ticket: TINY-8354

Description of Changes:
* Add `transparent` prop for dialog iframes
* Update Preview and Insert Template dialogs

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


lnewson:
Just going to convert this to a draft, as I'd prefer to keep the discussion about this in our slack thread for now.

HAFRMO (approved):

modules/bridge/src/main/ts/ephox/bridge/components/dialog/Iframe.ts
```diff
@@ -6,15 +6,18 @@ import { FormComponentWithLabel, formComponentWithLabelFields, FormComponentWith
 export interface IframeSpec extends FormComponentWithLabelSpec {
   type: 'iframe';
   sandboxed?: boolean;
+  opaque?: boolean;
```
lnewson, line 9: This name is a little inconsistent, so I'd suggest we call this `transparent` instead. That's the CSS name and similar to the old IE `allowtransparency` iframe attribute name as well. That also means we'd need to set the default to `true` below.

modules/tinymce/CHANGELOG.md
```diff
@@ -14,6 +14,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Dialogs will not exceed the window height on smaller screens #TINY-8146
 - Text alignment could not be applied to `pre` elements #TINY-7715
 
+### Improved
```
lnewson, line 17: The `Improved` section should be above the `Changed` section and below the `Added` section.

modules/tinymce/CHANGELOG.md
```diff
@@ -14,6 +14,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- New `opaque` prop for `iframe` dialog component fixes an issue in Preview and Insert Template dialogs with dark skins #TINY-8534
```
lnewson, line 18: This is really 2 changelog entries. We should add one for the new property and one for the preview/template dialog fixes.

Also as a side note, we shouldn't be using short hand in changelogs (e.g. `prop` should be `property`).

modules/tinymce/src/themes/silver/main/ts/ui/dialog/IFrame.ts
```diff
@@ -34,11 +34,13 @@ const getDynamicSource = (initialData: Optional<string>): IFrameSourcing => {
...
 const renderIFrame = (spec: IframeSpec, providersBackstage: UiFactoryBackstageProviders, initialData: Optional<string>) => {
   const isSandbox = spec.sandboxed;
+  const isOpaque = spec.opaque;
 
   const attributes = {
     ...spec.label.map<{ title?: string }>((title) => ({ title })).getOr({}),
     ...initialData.map((html) => ({ srcdoc: html })).getOr({}),
-    ...isSandbox ? { sandbox: 'allow-scripts allow-same-origin' } : { }
+    ...isSandbox ? { sandbox: 'allow-scripts allow-same-origin' } : { },
+    ...isOpaque ? { class: 'tox-dialog__iframe tox-dialog__iframe--opaque' } : { class: 'tox-dialog__iframe' }
```
lnewson, line 43: As Andy mentioned on slack this should be using the `classes` property in the `dom` part of the alloy spec instead of setting it as an attribute. e.g:
```
dom: {
  tag: 'iframe',
  attributes,
  classes: isOpaque ? [ 'tox-dialog__iframe', 'tox-dialog__iframe--opaque' ] : [ 'tox-dialog__iframe' ]
},
```

Note: The above has some duplication so if possible we should probably create a base array and then extend it in the above instead.

lnewson (changes requested):
Looking good overall, just a few little things to fix up 👍

modules/tinymce/CHANGELOG.md
noxuhax: Ok, I was having issues fitting it into one entry but did not know if it’s fine to have to entries for one task :) Should new property be under `Improved` and dialog fix is under `Fixed` then?

modules/tinymce/src/themes/silver/main/ts/ui/dialog/IFrame.ts
noxuhax: Thanks for clarifying that, now it makes sense

lnewson:
> @lnewson I think a little help with tests would be quite useful here. As I understand it, they already should be failing due to class being present on the `iframe` element and [IframeTest](https://github.com/tinymce/tinymce/blob/develop/modules/tinymce/src/themes/silver/test/ts/headless/components/iframe/IframeTest.ts?rgh-link-date=2022-04-04T16%3A13%3A07Z#L36) seems to check for empty classes but it runs fine for me. I’m a bit confused here :)

Haven't forgotten about this yet, just been tied up with the TinyMCE 6 enterprise release sorry. Should have time tomorrow or later this afternoon to help with this.

modules/tinymce/CHANGELOG.md
```diff
@@ -6,13 +6,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- New `transparent` property for `iframe` dialog component #TINY-8534
+
 ### Changed
 - The `end_container_on_empty_block` option can now take a string of blocks to split when pressing Enter twice #TINY-6559
 - The default value for `end_container_on_empty_block` option has been changed to `'blockquote'` #TINY-6559
 
 ### Fixed
 - Dialogs will not exceed the window height on smaller screens #TINY-8146
 - Text alignment could not be applied to `pre` elements #TINY-7715
+- Preview and Insert Template dialogs display correct content backgorund in dark skins #TINY-8534
```
lnewson, line 19:
```suggestion
- Preview and Insert Template dialogs display the correct content background color when using dark skins #TINY-8534
```

lnewson (approved):

modules/tinymce/src/themes/silver/main/ts/ui/dialog/IFrame.ts
```diff
@@ -51,7 +53,8 @@ const renderIFrame = (spec: IframeSpec, providersBackstage: UiFactoryBackstagePr
       uid: newSpec.uid,
       dom: {
         tag: 'iframe',
-        attributes
+        attributes,
+        classes: (isTransparent ? [ baseClass ] : [ baseClass, baseClass + '--opaque' ])
```
spocke, line 57: Maybe better not sure:
```suggestion
        classes: (isTransparent ? [ baseClass ] : [ baseClass, `${baseClass}--opaque` ])
```

spocke (approved):
Should this have a test?
I guess it's pretty simple thing so maybe it not likely to break.

noxuhax:
@lnewson pinging you to ask if your help offering with tests is still valid :)

lnewson:
I'll check what's happening as the task for helping out was passed over to others when I was on leave. If that wasn't done then I'll take a look today or tomorrow sorry @noxuhax.

lnewson:
> As I understand it, they already should be failing due to class being present on the iframe element and [IframeTest](https://github.com/tinymce/tinymce/blob/develop/modules/tinymce/src/themes/silver/test/ts/headless/components/iframe/IframeTest.ts?rgh-link-date=2022-04-04T16%3A13%3A07Z#L36) seems to check for empty classes but it runs fine for me. I’m a bit confused here :)
Regarding this, I should also add the structure assertions only check for the presence of things defined. So having `classes: [ ]` doesn't mean we expect it to be empty and instead just means we aren't asserting anything regarding the classes. So by adding some assertions we can make sure it has the classes we want.

noxuhax:
@lnewson got it! Thanks for the help and clarification, I really did misunderstood the meaning of `classes: [ ]`.

PR #7689, opened 2022-03-15
TINY-8575: Fixed the notification manager tests failing in isolation

lnewson:
Related Ticket: TINY-8575

Description of Changes:

This test includes a service message that means it opens the notification during init. That in turn causes issues if the first test in the suite doesn't run first, so this adjusts the tests to make sure it always works when the tests are run in isolation.

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [x] ~Milestone set~

### DIFF
--- a/modules/tinymce/src/core/test/ts/browser/NotificationManagerTest.ts
+++ b/modules/tinymce/src/core/test/ts/browser/NotificationManagerTest.ts
@@ -28,13 +28,16 @@ describe('browser.tinymce.core.NotificationManagerTest', () => {
         }
       }, []);
 
-      afterEach(() => {
+      const resetNotifications = () => {
         const editor = hook.editor();
         const notifications = [].concat(editor.notificationManager.getNotifications());
         Arr.each(notifications, (notification) => notification.close());
-
         beforeOpenEvents = [];
         openEvents = [];
+      };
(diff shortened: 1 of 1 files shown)


modules/tinymce/src/core/test/ts/browser/NotificationManagerTest.ts
```diff
@@ -57,6 +60,8 @@ describe('browser.tinymce.core.NotificationManagerTest', () => {
 
       it('TBA: Should not add duplicate text message', () => {
         const editor = hook.editor();
+        resetNotifications();
```
spocke, line 63: Could this be a beforeEach or would that be harder to understand.

spocke (approved):

modules/tinymce/src/core/test/ts/browser/NotificationManagerTest.ts
lnewson: Not every test needs this to be reset (the first test for example), so using a `beforeEach` would break things. I did wonder if this maybe should be 2 separate tests though, but ended up deciding on this (though happy to do something else if others prefer).

tiny-ben-tran (approved):

PR #7707, opened 2022-03-19
TINY-8582: Fixed the buttonType property not working for dialog footer buttons

tiny-ben-tran:
Related Ticket: TINY-8582

Description of Changes:
* Deprecated primary property of dialog footer buttons
* Fixed the buttonType property not working for dialog footer buttons
* Removed `buttonType = Optional.none()` in Button.ts

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/bridge/src/main/ts/ephox/bridge/components/dialog/DialogFooterButton.ts
+++ b/modules/bridge/src/main/ts/ephox/bridge/components/dialog/DialogFooterButton.ts
@@ -11,6 +11,7 @@ export type DialogFooterToggleMenuItem = DialogToggleMenuItem;
 interface BaseDialogFooterButtonSpec {
   name?: string;
   align?: 'start' | 'end';
+  /** @deprecated use `buttonType: "primary"` instead */
   primary?: boolean;
   enabled?: boolean;
   icon?: string;
--- a/modules/bridge/src/main/ts/ephox/bridge/components/dialog/DialogFooterButton.ts
+++ b/modules/bridge/src/main/ts/ephox/bridge/components/dialog/DialogFooterButton.ts
@@ -14,6 +14,7 @@ interface BaseDialogFooterButtonSpec {
   primary?: boolean;
   enabled?: boolean;
   icon?: string;
+  buttonType?: 'primary' | 'secondary';
 }
 
 export interface DialogFooterNormalButtonSpec extends BaseDialogFooterButtonSpec {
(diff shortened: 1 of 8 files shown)


modules/bridge/src/main/ts/ephox/bridge/components/dialog/DialogFooterButton.ts
```diff
@@ -14,6 +14,7 @@ interface BaseDialogFooterButtonSpec {
   primary?: boolean;
```
lnewson, line 14: This needs the `@deprecated` comment as well.
```suggestion
  /** @deprecated use `buttonType: "primary"` instead */
  primary?: boolean;
```

lnewson (changes requested):
Looking good thanks Ben! However, we need a test for this so it doesn't regress and a TinyMCE changelog entry (in addition to the bridge one).

modules/bridge/src/main/ts/ephox/bridge/components/dialog/DialogFooterButton.ts
tiny-ben-tran: I thought to write a test for this but I saw [this test](https://github.com/tinymce/tinymce/blob/496a2b9662f9988e5cde4900c9ad67961cee7bc0/modules/tinymce/src/themes/silver/test/ts/headless/components/dialogbutton/DialogButtonTest.ts#L11) which seems to cover `buttonType` property. Should there be another one specific for dialog footer button?

lnewson: If that test covered this then it would have been failing... if you look closer at it then it only handles dialog buttons, not dialog footer buttons.

modules/tinymce/CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Dialogs will not exceed the window height on smaller screens #TINY-8146
+- The `buttonType` property did not work for dialog footer buttons #TINY-8582
```
lnewson, line 11: This is in the wrong place... if you look closer you'll see the 6.0.1 entries are below.

modules/bridge/src/main/ts/ephox/bridge/components/dialog/DialogFooterButton.ts
tiny-ben-tran: Figured out the difference. I am writing tests as we speak..

TheSpyder (approved):

TheSpyder:
Pending Lee's feedback

lnewson:
I only just DM'd @tiny-ben-tran about this, but I noticed this is also targeting/based off the wrong branch. This is a hotfix for 6.0.1, so it should be based off of and targeting `master`. Ben is going to fix that up.

tiny-ben-tran:
Closed in favour of #7710 which is based of a right branch

Issue #7719, opened 2022-03-25
Canceling indent/outdent in BeforeExecCommand does not affect lists


rozpuszczalny:
**What is the current behavior? Describe the bug**
When `indent` or `outdent` action is defaultPrevent'ed from `BeforeExecCommand` event handler, lists do indent/outdent regardless of being canceled.

**Please provide the steps to reproduce and if possible a minimal demo of the problem via [fiddle.tiny.cloud](https://fiddle.tiny.cloud/) or similar.**
https://fiddle.tiny.cloud/1Zhaab

**What is the expected behavior?**
Lists ignore the indent/outdent commands if they were canceled.

**Which versions of TinyMCE, and which browser / OS are affected by this issue? Did this work in previous versions of TinyMCE?**
Stable 5 Lists plugin for sure. Probably on all of the browsers

**Footnotes**
Looks like lists uses `.on('BeforeExecCommand')`, instead of `.on('ExecCommand')`. See this handler in register:
https://github.com/tinymce/tinymce/blob/70ab2cf51d411750e0c285d8b1b0026d742265f7/modules/tinymce/src/plugins/lists/main/ts/api/Commands.ts#L22-L31

I ~~can~~ created a simple PR, which changes `BeforeExecCommand` to `ExecCommand`, but I'm not sure why `BeforeExecCommand` was made in the first place. #7720

PR #7720, opened 2022-03-25
Lists listen to ExecCommand events, so it's possible to cancel indent/outdent via BeforeExecCommand

rozpuszczalny:
Related Ticket: N/A

Description of Changes:
* Lists listen to ExecCommand events, so it's possible to cancel indent/outdent via BeforeExecCommand

Pre-checks:
* [x] Changelog entry added
* [ ] ~~Tests have been added (if applicable)~~
* [ ] ~~Branch prefixed with `feature/` for new features (if applicable)~~

Review:
* [ ] Milestone set

GitHub issues (if applicable): #7719

### DIFF
--- a/modules/tinymce/src/plugins/lists/main/ts/api/Commands.ts
+++ b/modules/tinymce/src/plugins/lists/main/ts/api/Commands.ts
@@ -20,7 +20,7 @@ const registerDialog = (editor: Editor): void => {
 };
 
 const register = (editor: Editor): void => {
-  editor.on('BeforeExecCommand', (e) => {
+  editor.on('ExecCommand', (e) => {
     const cmd = e.command.toLowerCase();
 
     if (cmd === 'indent') {
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Dialogs will not exceed the window height on smaller screens #TINY-8146
+- Lists are listening to indent/outdent events though ExecCommand, so it's possible to cancel it via BeforeExecCommand
 
 ## 6.0.1 - 2022-03-23
 


HAFRMO (approved):

modules/tinymce/src/plugins/lists/main/ts/api/Commands.ts
```diff
@@ -20,7 +20,7 @@ const registerDialog = (editor: Editor): void => {
 };
 
 const register = (editor: Editor): void => {
-  editor.on('BeforeExecCommand', (e) => {
+  editor.on('ExecCommand', (e) => {
```
lnewson, line 23: This isn't the correct way to handle this case, as we want to override the default behavior and this would cause it to run after the default behavior instead. So what we should be doing to solve this is to add a check to see if it's already been prevented instead. As such this probably needs to be something like this:
```
editor.on('BeforeExecCommand', (e) => {
  if (!e.isDefaultPrevented()) {
    const cmd = e.command.toLowerCase();

    if (cmd === 'indent') {
      indentListSelection(editor);
    } else if (cmd === 'outdent') {
      outdentListSelection(editor);
    }
  }
});
```

lnewson (changes requested):
Thanks @rozpuszczalny for this PR!

The approach was a good attempt, but isn't quite the right solution here so I've left a different suggestion. Also are you able to add some tests for this so it doesn't regress in the future? If not then let us know and we'll look at merging this into a temporary feature branch to be able to add tests.

modules/tinymce/src/plugins/lists/main/ts/api/Commands.ts
TheSpyder: Don't plugins register their handlers before the core? So swapping to `ExecCommand` will still override the core indent/outdent behaviour? It seems strange to me that lists had to use `Before` to achieve the override.

lnewson: No, that's not how events, nor the commands work. We have the `BeforeExecCommand` event which fires before the core or native commands run and then the `ExecCommand` event which fires after that. So in this case the flow of events is:
- `BeforeExecCommand` event is fired
  - List indent/outdent behaviour override
- Core indent/outdent command runs
- `ExecCommand` event is fired

I'm really not sure how you expect the `ExecCommand` event to run before depending on the plugin registration order to be honest. I'm wondering if you've got mixed up thinking Core does the processing in an event, rather than in the command? Worth noting that we also can't just override the indent/outdent commands, as it only needs to override in specific cases.

lnewson:
Closing as there has been no update. If you can make the changes requested we'll be happy to re-evaluate this.

PR #7726, opened 2022-03-29
TINY-8215: Inconsistent triple clicking when around nested contenteditable

vpyshnenko:
Related Ticket:

Description of Changes:
* Placeholder text

Pre-checks:
* [ ] Changelog entry added

Review:
* [ ] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/selection/MultiClickSelection.ts
+++ b/modules/tinymce/src/core/main/ts/selection/MultiClickSelection.ts
@@ -1,44 +1,82 @@
 import { SugarElement } from '@ephox/sugar';
 
 import Editor from '../api/Editor';
-import * as CaretFinder from '../caret/CaretFinder';
 import { CaretPosition } from '../caret/CaretPosition';
-import { isTextBlock } from '../dom/ElementType';
+import { isBlock } from '../dom/ElementType';
+import { isCaretCandidate } from '../caret/CaretCandidate';
 import * as NodeType from '../dom/NodeType';
 import * as RangeNormalizer from './RangeNormalizer';
+import DomTreeWalker from '../api/dom/TreeWalker';
 
-const isTextBlockNode = (node: Node): node is Element => NodeType.isElement(node) && isTextBlock(SugarElement.fromDom(node));
+const isBr = NodeType.isBr;
+const isText = NodeType.isText;
 
-const normalizeSelection = (editor: Editor) => {
-  const rng = editor.selection.getRng();
-  const startPos = CaretPosition.fromRangeStart(rng);
-  const endPos = CaretPosition.fromRangeEnd(rng);
+const isBoundary = (node: Node): boolean => {
+  return isBr(node) || isBlock(SugarElement.fromDom(node));
+}
 
+const walkBackwardWhile = (startNode: Node, rootNode: Node): Node => {
+  const walker = new DomTreeWalker(startNode, rootNode);
+  let result: Node = startNode;
+  for (let next = walker.prev(); next && !isBoundary(next); next = walker.prev()){
+    if(isCaretCandidate(next)){
+      result = next
+    }
+  }
+  return result
+}
+
+const walkForwardWhile = (startNode: Node, rootNode: Node): Node => {
+  const walker = new DomTreeWalker(startNode, rootNode);
+  let result: Node = startNode;
+  for (let next = startNode; next && !isBoundary(next); next = walker.next()){
+    if(isCaretCandidate(next)){
+      result = next
+    }
+  }
+  return result
+}
+
+const findClosestBlockRange = (startRng: Range, rootNode: Node) => {
+  const startPos = CaretPosition.fromRangeStart(startRng);
+  let clickNode = startPos.container()
   if (CaretPosition.isElementPosition(startPos)) {
-    const container = startPos.container();
-    if (isTextBlockNode(container)) {
-      CaretFinder.firstPositionIn(container).each((pos) => rng.setStart(pos.container(), pos.offset()));
-    }
+    clickNode = clickNode.childNodes[startPos.offset()]
   }
 
-  if (CaretPosition.isElementPosition(endPos)) {
-    const container = startPos.container();
-    if (isTextBlockNode(container)) {
-      CaretFinder.lastPositionIn(container).each((pos) => rng.setEnd(pos.container(), pos.offset()));
-    }
+  const startNode = walkBackwardWhile(clickNode, rootNode)
+  const endNode = walkForwardWhile(clickNode, rootNode)
+
+  const rng = document.createRange();
+  if (isText(startNode)) {
+    rng.setStart(startNode, 0);
+  } else {
+    rng.setStartBefore(startNode);
   }
 
+  if (isText(endNode)) {
+    rng.setEnd(endNode, endNode.data.length);
+  } else {
+    rng.setEndAfter(endNode);
+  }
+  return rng
+}
+
+const onTripleClickSelect = (editor) => {
+  const rng = findClosestBlockRange(editor.selection.getRng(), editor.getBody())
   editor.selection.setRng(RangeNormalizer.normalize(rng));
-};
+}
 
 const setup = (editor: Editor) => {
-  editor.on('click', (e) => {
+  editor.on('mousedown', (e) => {
     if (e.detail >= 3) {
-      normalizeSelection(editor);
+      e.preventDefault();
+      onTripleClickSelect(editor)
     }
   });
 };
 
 export {
+  findClosestBlockRange,
   setup
 };
--- a/modules/tinymce/src/core/test/ts/browser/selection/MultiClickSelectionTest.ts
+++ b/modules/tinymce/src/core/test/ts/browser/selection/MultiClickSelectionTest.ts
@@ -1,32 +1,128 @@
 import { describe, it } from '@ephox/bedrock-client';
+import { assert } from 'chai';
 import { TinyAssertions, TinyHooks, TinySelections } from '@ephox/wrap-mcagar';
+
+import DOMUtils from 'tinymce/core/api/dom/DOMUtils';
+import * as ViewBlock from '../../module/test/ViewBlock';
+import { findClosestBlockRange } from 'tinymce/core/selection/MultiClickSelection';
 
 import Editor from 'tinymce/core/api/Editor';
 
 describe('browser.tinymce.core.selection.MultiClickSelectionTest', () => {
-  const hook = TinyHooks.bddSetupLight<Editor>({
-    base_url: '/project/tinymce/js/tinymce'
-  }, [], true);
+  describe('findClosestBlockRange test', () => {
+    const DOM = DOMUtils.DOM;
+    const viewBlock = ViewBlock.bddSetup();
+    const getRoot = viewBlock.get;
+    const setupHtml = viewBlock.update;
 
-  const fakeMultiClick = (editor: Editor, clickCount) => {
-    editor.dispatch('click', { detail: clickCount } as MouseEvent);
-  };
+    const createRange = (sc: Node, so: number, ec: Node, eo: number) => {
+      const rng = DOM.createRng();
+      rng.setStart(sc, so);
+      rng.setEnd(ec, eo);
+      return rng;
+    };
 
-  for (let clickCount = 3; clickCount <= 10; clickCount++) {
-    it(`Normalize selection from index text node to text node offsets with ${clickCount} clicks`, () => {
-      const editor = hook.editor();
-      editor.setContent('<p>abc</p>');
-      TinySelections.setSelection(editor, [ 0 ], 0, [ 0 ], 1);
-      fakeMultiClick(editor, clickCount);
-      TinyAssertions.assertSelection(editor, [ 0, 0 ], 0, [ 0, 0 ], 3);
+    const assertRange = (expected: Range, actual: Range) => {
+      assert.strictEqual(actual.startContainer, expected.startContainer, 'startContainers should be equal');
+      assert.strictEqual(actual.startOffset, expected.startOffset, 'startOffset should be equal');
+      assert.strictEqual(actual.endContainer, expected.endContainer, 'endContainer should be equal');
+      assert.strictEqual(actual.endOffset, expected.endOffset, 'endOffset should be equal');
+    };
+
+    const testFindClosestBlockRange = (startRng: Range, expectedRng: Range) =>
+      assertRange(expectedRng, findClosestBlockRange(startRng, getRoot()))
+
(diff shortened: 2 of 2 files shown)


modules/tinymce/src/core/test/ts/browser/selection/MultiClickSelectionTest.ts
```diff
@@ -1,32 +1,128 @@
...
+      assert.strictEqual(actual.endOffset, expected.endOffset, 'endOffset should be equal');
+    };
+
+    const testFindClosestBlockRange = (startRng: Range, expectedRng: Range) =>
+      assertRange(expectedRng, findClosestBlockRange(startRng, getRoot()));
+
+    it(`should return the range with the whole text content of the block`, () => {
+      setupHtml('<p>aaa bIb ccc</p>');
+      const textNode = getRoot().firstChild.firstChild;
+      const startRng = createRange(textNode, 4, textNode, 7);
```
spocke, line 38: I guess you could use Agar Cursors here or sugar Path directly to resolve paths to ranges. It would avoid having these long node.firstNode.firstNode.childNodes[1] chains since then you can have a path like [0, 0, 1] to get to the same place. So if you had a way to just take paths and offsets and produce dom ranges it would fit into the existing tests. Something like this `const expectedRng = resolveRange([ 0, 0 ], 0, [ 0, 0 ], 11)` to get a range of the whole range node but maybe my brain has just gotten more used to reading that syntax than having traversing node paths. I kind of like that this style is also explicitly named nodes as variables.

- [ ] Consider using path based assertion like other tests

modules/tinymce/src/core/test/ts/browser/selection/MultiClickSelectionTest.ts
```diff
@@ -1,32 +1,128 @@
...
+    });
+  });
+
+  describe('fake multi click browser test', () => {
+    const hook = TinyHooks.bddSetupLight<Editor>({
+      base_url: '/project/tinymce/js/tinymce'
+    }, [], true);
+
+    const fakeMultiClick = (editor: Editor, clickCount: number, target: HTMLElement) => {
+      editor.dispatch('mousedown', { detail: clickCount, target } as unknown as MouseEvent);
```
spocke, line 87: I wonder if we should add details as to mouse down settings in Agar Mouse so that it's more of a simulated real even than a simulated tinymce event.

modules/tinymce/src/core/main/ts/selection/MultiClickSelection.ts
```diff
@@ -1,44 +1,82 @@
...
-  const endPos = CaretPosition.fromRangeEnd(rng);
+const isBoundary = (node: Node): boolean => {
+  return isBr(node) || isBlock(SugarElement.fromDom(node));
+};
 
-  if (CaretPosition.isElementPosition(startPos)) {
-    const container = startPos.container();
-    if (isTextBlockNode(container)) {
-      CaretFinder.firstPositionIn(container).each((pos) => rng.setStart(pos.container(), pos.offset()));
+const walkBackwardWhile = (startNode: Node, rootNode: Node): Node => {
```
spocke, line 18: I wonder if these two could be the same function but maybe that will just make the code more confusing. Another thing is that it should probably also not select the last caret candidate if it finds one within a noneditable element. So for this case `foo<span contenteditable="false">bar</span>` the selection should be a element position after the noneditable span not the text node within it. I think this could just be a separate step where you check if the found nodes closest contentEditable state is false then place the point after that element. I'm guessing some browsers will normalize the selection for us in this case anyway moving it outside.

spocke (changes requested):

modules/tinymce/src/core/main/ts/selection/MultiClickSelection.ts
```diff
@@ -1,44 +1,82 @@
...
+    if (isCaretCandidate(next)) {
+      result = next;
     }
   }
+  return result;
+};
+
+const findClosestBlockRange = (startRng: Range, rootNode: Node) => {
+  const startPos = CaretPosition.fromRangeStart(startRng);
+  let clickNode = startPos.container();
```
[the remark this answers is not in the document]
vpyshnenko: I see that startPos.getNode() function uses deprecated RangeNodes.getNodeUnsafe function. Is the current moment the appropriate time to remove it and replace with RangeNodes.getNode() function?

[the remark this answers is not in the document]
lnewson: Yeah it does rely on it returning undefined. When I first added the newer/safer `getNode` function I swapped that but then a number of tests failed as it's relying on the unsafe behaviour still. As such given time constraints, I renamed the original function to include unsafe so we hopefully wouldn't use it in future and left this using the older logic.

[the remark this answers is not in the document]
lnewson: If I recall correctly from looking at it sometime later it was https://github.com/tinymce/tinymce/blob/236891ffcdf2fd1b698d020e67c31838fb110019/modules/tinymce/src/core/main/ts/caret/CaretPosition.ts#L203 that relied on it returning `undefined` as the index ended up being `-1`. As such, the safe method constrained that back to `0` and returned a `node`, whereas the unsafe version returned `undefined`.

I could be remembering incorrectly though, but yeah if we're concerned about it I'd suggest logging a separate Jira and handling the changes there.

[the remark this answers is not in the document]
vpyshnenko: Actually, I'm not sure `getNodeUnsafe` should be deprecated, as we can see it has its own place. It might have certain usage, which depends on desired logic.

modules/tinymce/src/core/main/ts/selection/MultiClickSelection.ts
```diff
@@ -1,44 +1,78 @@
...
+const isBr = NodeType.isBr;
+const isText = NodeType.isText;
+const isContentEditableFalse = (elm: SugarElement<Node>): boolean => NodeType.isContentEditableFalse(elm.dom);
+const isRoot = (rootNode: Node) => (elm: SugarElement<Node>): boolean => Compare.eq(SugarElement.fromDom(rootNode), elm);
 
-const normalizeSelection = (editor: Editor) => {
-  const rng = editor.selection.getRng();
-  const startPos = CaretPosition.fromRangeStart(rng);
-  const endPos = CaretPosition.fromRangeEnd(rng);
+const getClosestBlock = (node: Node, rootNode: Node): Node =>
```
spocke, line 16: Maybe rename this to closest scope and also make sure that it covers the case of triple clicking in this content:
```
<p contenteditable="false">foo<span contenteditable="true">hello</span>bar</p>
```

Since triple click on hello should only select `hello` not `foo` and `bar`. At least that is how the Chrome does it today I guess it's debatable if it should select the parent noneditable block in that case.

modules/agar/src/main/ts/ephox/agar/mouse/Clicks.ts
```diff
@@ -26,6 +26,7 @@ interface Settings {
   metaKey?: boolean;
   bubbles?: boolean;
   cancelable?: boolean;
+  detail?: number;
```
spocke, line 29: This needs a changelog item under unreleased in the agar package.

modules/tinymce/src/core/main/ts/selection/MultiClickSelection.ts
```diff
@@ -1,44 +1,78 @@
...
-      CaretFinder.lastPositionIn(container).each((pos) => rng.setEnd(pos.container(), pos.offset()));
+const findClosestBlockRange = (startRng: Range, rootNode: Node) => {
+  const startPos = CaretPosition.fromRangeStart(startRng);
+  const clickNode = startPos.getNode();
+  const scope = getClosestBlock(clickNode, rootNode);
+  const startNode = findEdgeCaretCandidate(clickNode, scope, false);
+  const endNode = findEdgeCaretCandidate(clickNode, scope, true);
+
+  const rng = document.createRange();
+  getClosestCef(startNode, scope).fold(() => {
```
spocke, line 43: Maybe extremely picky but I tend to format each case of the fold like this:
```
getClosestCef(startNode, scope).fold(
  () => {
    if (isText(startNode)) {
      rng.setStart(startNode, 0);
    } else {
      rng.setStartBefore(startNode);
    }
  },
  (cef) => rng.setStartBefore(cef.dom)
);
```
But maybe that is a old style I used when we had adts faked using folds like this where each case would be it's own line.

spocke (approved):

modules/tinymce/src/core/main/ts/selection/MultiClickSelection.ts
```diff
@@ -1,44 +1,83 @@
...
+import { isBlock } from '../dom/ElementType';
 import * as NodeType from '../dom/NodeType';
 import * as RangeNormalizer from './RangeNormalizer';
 
-const isTextBlockNode = (node: Node): node is Element => NodeType.isElement(node) && isTextBlock(SugarElement.fromDom(node));
+const isBr = NodeType.isBr;
+const isText = NodeType.isText;
+const isContentEditableFalse = (elm: SugarElement<Node>): boolean => NodeType.isContentEditableFalse(elm.dom);
+const isContentEditableTrue = (elm: SugarElement<Node>): boolean => NodeType.isContentEditableTrue(elm.dom);
+const isRoot = (rootNode: Node) => (elm: SugarElement<Node>): boolean => Compare.eq(SugarElement.fromDom(rootNode), elm);
```
spocke, line 15: I wonder how many times we have invented this function maybe we should add some tech dept ticket to just include this partial function in sugar Compare. This is more of a general comment. I think Hannes also invented this function in the other PR. :smile:

spocke (approved):

modules/tinymce/src/core/main/ts/selection/MultiClickSelection.ts
lnewson: It's probably a lot... the problem with this one is that we need that shared library for TinyMCE and other monorepo libs as we have nowhere to place this sort of thing atm and we can't create a circular reference.

modules/tinymce/src/core/main/ts/selection/MultiClickSelection.ts
```diff
@@ -1,44 +1,83 @@
...
+        rng.setEnd(endNode, endNode.data.length);
+      } else {
+        rng.setEndAfter(endNode);
+      }
+    }, (cef) => rng.setEndAfter(cef.dom)
+  );
+  return rng;
+};
 
+const onTripleClickSelect = (editor) => {
```
lnewson, line 66: Types?
```suggestion
const onTripleClickSelect = (editor: Editor) => {
```

modules/tinymce/src/core/main/ts/selection/MultiClickSelection.ts
```diff
@@ -1,44 +1,83 @@
...
+
+  const rng = document.createRange();
+  getClosestCef(startNode, scope).fold(
+    () => {
+      if (isText(startNode)) {
+        rng.setStart(startNode, 0);
+      } else {
+        rng.setStartBefore(startNode);
+      }
+    }, (cef) => rng.setStartBefore(cef.dom)
```
lnewson, line 51:
```suggestion
    },
    (cef) => rng.setStartBefore(cef.dom)
```
This is how this is normally formatted if the none case is on a new line.

modules/tinymce/src/core/main/ts/selection/MultiClickSelection.ts
```diff
@@ -1,44 +1,83 @@
...
+  );
+
+  getClosestCef(endNode, scope).fold(
+    () => {
+      if (isText(endNode)) {
+        rng.setEnd(endNode, endNode.data.length);
+      } else {
+        rng.setEndAfter(endNode);
+      }
+    }, (cef) => rng.setEndAfter(cef.dom)
```
lnewson, line 61:
```suggestion
    },
    (cef) => rng.setEndAfter(cef.dom)
```

modules/tinymce/src/core/test/ts/browser/selection/MultiClickSelectionTest.ts
```diff
@@ -1,32 +1,162 @@
...
 import { TinyAssertions, TinyHooks, TinySelections } from '@ephox/wrap-mcagar';
+import { assert } from 'chai';
 
+import DOMUtils from 'tinymce/core/api/dom/DOMUtils';
 import Editor from 'tinymce/core/api/Editor';
+import { findClosestBlockRange } from 'tinymce/core/selection/MultiClickSelection';
+
+import * as ViewBlock from '../../module/test/ViewBlock';
+
+type RangePath = [startContainerPath: number[], startOffset: number, endContainerPath: number[], endOffset: number];
```
lnewson, line 13:
```suggestion
type RangePath = [ startContainerPath: number[], startOffset: number, endContainerPath: number[], endOffset: number ];
```

modules/tinymce/src/core/test/ts/browser/selection/MultiClickSelectionTest.ts
```diff
@@ -1,32 +1,162 @@
...
+  describe('findClosestBlockRange test', () => {
+    const DOM = DOMUtils.DOM;
+    const viewBlock = ViewBlock.bddSetup();
+    const getRoot = viewBlock.get;
+    const setupHtml = viewBlock.update;
+
+    const resolveRange = (args: RangePath): Range => {
+      const root = SugarElement.fromDom(getRoot());
+      const startContainer = Cursors.calculateOne(root, args[0]);
+      const endContainer = Cursors.calculateOne(root, args[2]);
```
lnewson, line 25: Is there a reason we're not using `Cursors.range()` here, as then you wouldn't need the custom `RangePath` type and could just use `Cursors.CursorPath`. That would also make things a lot more readable below as you'd have named parts instead of trying to remember what's the start path, offset, etc...

- [ ] Consider changing this to use `CursorPath` and `Cursors.range`

lnewson (approved):

modules/tinymce/src/core/test/ts/browser/selection/MultiClickSelectionTest.ts
vpyshnenko: Such a way to specify ranges was suggested [here](https://github.com/tinymce/tinymce/pull/7726#discussion_r838417439) by @spocke . And it turns out this is a matter of taste.

lnewson: What I've said doesn't contradict what Spocke said and instead expands on it further. In general we prefer named structures instead of unnamed as you can't tell what each array item represents easily.

lnewson (approved):

PR #7728, opened 2022-03-29
TINY-7005: Changed escape key handling from keydown to keyup

HAFRMO:
Related Ticket:

Description of Changes:
Changed the alloy window closing to react to key up rather than key down, to prevent escape's key up to be sent to the main window instead of the dialog. No tests added, check will be performed by QA.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/tinymce/CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Dialogs will not exceed the window height on smaller screens #TINY-8146
 - Text alignment could not be applied to `pre` elements #TINY-7715
+- Dialogs would in some cases not prevent key presses from escalating to listeners #TINY-7005
```
spocke, line 16: Do we need to be more specific here that it's the keyup events that would be triggered for Esc and Enter?

spocke (approved):

modules/alloy/src/main/ts/ephox/alloy/keying/TabbingTypes.ts
```diff
@@ -112,11 +112,12 @@ const create = (cyclicField: FieldProcessor): KeyingType.KeyingType<TabbingConfi
   const getKeydownRules = Fun.constant([
     KeyRules.rule(KeyMatch.and([ KeyMatch.isShift, KeyMatch.inSet(Keys.TAB) ]), goBackwards),
     KeyRules.rule(KeyMatch.inSet(Keys.TAB), goForwards),
-    KeyRules.rule(KeyMatch.inSet(Keys.ESCAPE), exit),
     KeyRules.rule(KeyMatch.and([ KeyMatch.isNotShift, KeyMatch.inSet(Keys.ENTER) ]), execute)
   ]);
 
-  const getKeyupRules = Fun.constant([ ]);
+  const getKeyupRules = Fun.constant([
```
lnewson, line 118: This needs an Alloy changelog entry and you also need to add the minor version bump in `versisons.txt`.

- [ ] Alloy changelog
- [ ] Add version bump to `versions.txt`

modules/tinymce/CHANGELOG.md
lnewson: Hmm... technically this applies to many things. I don't think we should be saying `Dialogs` specifically here, as this could also apply to other UI components. I do agree we should say this applies to `Esc` only though. Maybe something like this?
```suggestion
- UI components, such as dialogs, would in some cases cause the `esc` keyup event to incorrectly trigger inside the editor #TINY-7005
```

modules/alloy/src/main/ts/ephox/alloy/keying/TabbingTypes.ts
```diff
@@ -112,11 +112,12 @@ const create = (cyclicField: FieldProcessor): KeyingType.KeyingType<TabbingConfi
...
+    KeyRules.rule(KeyMatch.inSet(Keys.ESCAPE), exit)
```
lnewson, line 119: If you're changing this then you really should be changing other keying escape handlers. For example `EscapingType.ts`

lnewson (changes requested):
Looks good for this specific change, but if you're changing one escape type then we should be changing the rest of them (see the various keying types in `modules/alloy/src/main/ts/ephox/alloy/keying/`).

modules/alloy/src/main/ts/ephox/alloy/keying/SpecialType.ts
```diff
@@ -45,12 +45,13 @@ const getKeydownRules = (component: AlloyComponent, simulatedEvent: NativeSimula
...
+  KeyRules.rule(KeyMatch.inSet(Keys.SPACE), specialInfo.onSpace)
 ];
 
-const getKeyupRules = (component: AlloyComponent, simulatedEvent: NativeSimulatedEvent, specialInfo: SpecialConfig): Array<KeyRules.KeyRule<SpecialConfig, Stateless>> => specialInfo.stopSpaceKeyup ? [
+const getKeyupRules = (component: AlloyComponent, simulatedEvent: NativeSimulatedEvent, specialInfo: SpecialConfig): Array<KeyRules.KeyRule<SpecialConfig, Stateless>> => (specialInfo.stopSpaceKeyup ? [
   KeyRules.rule(KeyMatch.inSet(Keys.SPACE), stopEventForFirefox)
-] : [ ];
+] : [ ]).concat([
+  KeyRules.rule(KeyMatch.inSet(Keys.ESCAPE), specialInfo.onEscape)
+]);
```
lnewson, line 55: This might be a little more readable?
```suggestion
const getKeyupRules = (component: AlloyComponent, simulatedEvent: NativeSimulatedEvent, specialInfo: SpecialConfig): Array<KeyRules.KeyRule<SpecialConfig, Stateless>> => [
  ...(specialInfo.stopSpaceKeyup ? [ KeyRules.rule(KeyMatch.inSet(Keys.SPACE), stopEventForFirefox) ] : [ ]),
  KeyRules.rule(KeyMatch.inSet(Keys.ESCAPE), specialInfo.onEscape)
];
```

lnewson (approved):

PR #7733, opened 2022-03-31
TINY-8592: List backspace function no longer steps into tables

HAFRMO:
Related Ticket:

Description of Changes:
List plugin should no longer step into tables if the last cell of the table contains a list at the end. Also added typing to reflect TreeWalker's ability return undefined in addition to a Node ( such as when reaching the end of a walk ).

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -8,12 +8,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - The `end_container_on_empty_block` option can now take a string of blocks to split when pressing Enter twice #TINY-6559
-- Added optional function parameter to `TreeWalker.prev2`, allowing aborting tree walking efforts #TINY-8592
 - The default value for `end_container_on_empty_block` option has been changed to `'blockquote'` #TINY-6559
 
 ### Fixed
 - Dialogs will not exceed the window height on smaller screens #TINY-8146
-- Some types on functions in `editor.dom.TreeWalker` class missed that it could return undefined #TINY-8592
+- Some types on functions in `tinymce.dom.TreeWalker` class missed that it could return undefined #TINY-8592
 - In some cases backspace would step into tables rather than remain outside #TINY-8592
 - Text alignment could not be applied to `pre` elements #TINY-7715
 
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -12,8 +12,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Dialogs will not exceed the window height on smaller screens #TINY-8146
-- Some types on functions in `tinymce.dom.TreeWalker` class missed that it could return undefined #TINY-8592
-- In some cases backspace would step into tables rather than remain outside #TINY-8592
+- Some types on functions in the `tinymce.dom.TreeWalker` class missed that it could return undefined #TINY-8592
+- In some cases pressing the backspace key would incorrectly step into tables rather than remain outside #TINY-8592
 - Text alignment could not be applied to `pre` elements #TINY-7715
 
 ## 6.0.1 - 2022-03-23
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -8,10 +8,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - The `end_container_on_empty_block` option can now take a string of blocks to split when pressing Enter twice #TINY-6559
+- Added optional function parameter to `TreeWalker.prev2`, allowing aborting tree walking efforts #TINY-8592
 - The default value for `end_container_on_empty_block` option has been changed to `'blockquote'` #TINY-6559
 
 ### Fixed
 - Dialogs will not exceed the window height on smaller screens #TINY-8146
+- Some types on functions in `editor.dom.TreeWalker` class missed that it could return undefined #TINY-8592
+- In some cases backspace would step into tables rather than remain outside #TINY-8592
 - Text alignment could not be applied to `pre` elements #TINY-7715
 
 ## 6.0.1 - 2022-03-23
--- a/modules/tinymce/src/core/main/ts/api/dom/TreeWalker.ts
+++ b/modules/tinymce/src/core/main/ts/api/dom/TreeWalker.ts
@@ -1,5 +1,3 @@
-import { Fun, Type } from '@ephox/katamari';
-
 export interface DomTreeWalkerConstructor {
   readonly prototype: DomTreeWalker;
 
@@ -66,8 +64,8 @@ class DomTreeWalker {
     return this.node;
   }
 
-  public prev2(shallow?: boolean, abortIf?: (node: Node | undefined) => boolean): Node | undefined {
-    this.node = this.findPreviousNode(this.node, Type.isFunction(abortIf) ? abortIf : Fun.never, shallow);
+  public prev2(shallow?: boolean): Node | undefined {
+    this.node = this.findPreviousNode(this.node, shallow);
     return this.node;
   }
 
@@ -98,12 +96,12 @@ class DomTreeWalker {
     }
   }
 
-  private findPreviousNode(node: Node | undefined, abortIf: (node: Node | undefined) => boolean, shallow?: boolean): undefined | Node {
+  private findPreviousNode(node: Node | undefined, shallow?: boolean): undefined | Node {
     let sibling: Node, parent: Node, child: Node;
 
     if (node) {
       sibling = node.previousSibling;
-      if (this.rootNode && sibling === this.rootNode || abortIf(sibling)) {
+      if (this.rootNode && sibling === this.rootNode) {
         return;
       }
 
@@ -111,10 +109,6 @@ class DomTreeWalker {
         if (!shallow) {
           // Walk down to the most distant child
           for (child = sibling.lastChild; child; child = child.lastChild) {
-            if (abortIf(child)) {
-              return;
-            }
-
             if (!child.lastChild) {
               return child;
             }
@@ -125,7 +119,7 @@ class DomTreeWalker {
       }
 
       parent = node.parentNode;
-      if (parent && parent !== this.rootNode && !abortIf(parent)) {
+      if (parent && parent !== this.rootNode) {
         return parent;
       }
     }
--- a/modules/tinymce/src/core/main/ts/api/dom/TreeWalker.ts
+++ b/modules/tinymce/src/core/main/ts/api/dom/TreeWalker.ts
@@ -36,7 +36,7 @@ class DomTreeWalker {
    * Returns the current node.
    *
    * @method current
-   * @return {Node | undefined} Current node where the walker is, or undefined if the walker has reached the end.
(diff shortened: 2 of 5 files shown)


spocke:
I think you can fix this issue without altering the tree walker at all. Just check for the closest table cell parent of the LI that it finds when traversing using the tree walker. Then find the closest table cell of the range startContainer if they are the same then you are deleting within a table cell and that should be fine. If it's not the same table cell or one or the other is missing then you are deleting between table cell boundaries and that should not be allowed. If you find no parent table cells then you are deleting in normal content and that should be allowed. You should be able to use PredicateFind.closest and Compare.contains for this from sugar. Cells are `td`, `th` or `caption` and the changes should be able to be scoped to the list plugin only.

HAFRMO:
> I think you can fix this issue without altering the tree walker at all. Just check for the closest table cell parent of the LI that it finds when traversing using the tree walker. Then find the closest table cell of the range startContainer if they are the same then you are deleting within a table cell and that should be fine. If it's not the same table cell or one or the other is missing then you are deleting between table cell boundaries and that should not be allowed. If you find no parent table cells then you are deleting in normal content and that should be allowed. You should be able to use PredicateFind.closest and Compare.contains for this from sugar. Cells are `td`, `th` or `caption` and the changes should be able to be scoped to the list plugin only.

That was the other method I could do, and I did try it. But it also seemed less reliable at finding the problems, and less efficient, so I went with the current solution.
If we really don't want to use this one I can switch.

spocke:
> That was the other method I could do, and I did try it. But it also seemed less reliable at finding the problems, and less efficient, so I went with the current solution.
> If we really don't want to use this one I can switch.

I just want to avoid making the public api more complex if we can and the tree walker is such an api. Since the need here seems to be scoped to the list plugin. Also I don't think performance here is a concern since editor contents trees tends to be wide not deep, it would be surprised to see a 1000 nested elements document where this would be a performance impact. It would have to check the depth of the document twice and compare but that should be a cheep operation for not super deep documents and also it wouldn't need to do one or the other if just one of them fails to meet the predicate like if there is no TD for the found LI node that would be the common case then there is no need to check the parents of the current caret location.

modules/tinymce/src/plugins/lists/main/ts/core/Delete.ts
```diff
@@ -209,6 +209,15 @@ const backspaceDeleteIntoListCaret = (editor: Editor, isForward: boolean): boole
     const rng = ListRangeUtils.normalizeRange(editor.selection.getRng());
     const otherLi = dom.getParent(findNextCaretContainer(editor, rng, isForward, root), 'LI', root);
 
+    const findValidElement = (element: SugarElement<Node>) => Arr.contains([ 'td', 'th', 'caption' ], SugarNode.name(element));
+    const findRoot = (node: SugarElement<Node>) => node.dom === root;
+    const otherLiCell = PredicateFind.closest(SugarElement.fromDom(otherLi), findValidElement, findRoot);
+    const caretCell = PredicateFind.closest(SugarElement.fromDom(rng.startContainer), findValidElement, findRoot);
+
+    if (otherLiCell !== caretCell) {
```
spocke, line 217: This can't be right since `PredicateFind.closest` returns a Optional. Use something like `Optionals.equals`. Also since this is not breaking any test that suggest that more tests are needed.

- [ ] Switch to proper equals

modules/tinymce/src/plugins/lists/test/ts/browser/BackspaceDeleteFromBlockIntoLiTest.ts
```diff
@@ -13,6 +13,27 @@ describe('Browser Test: .RemoveTrailingBlockquoteTest', () => {
     base_url: '/project/tinymce/js/tinymce'
   }, [ Plugin ]);
 
+  it('TINY-8592: backspace from p outside of table with no change', () => {
```
spocke, line 16: Since there is no test failing for the incorrect compare more tests are needed here.

- [ ] Add tests that covers delete from within a the same cell, from cell to text from text to cell from cell to other cell and from text to text.

modules/tinymce/src/core/main/ts/api/dom/TreeWalker.ts
```diff
@@ -96,20 +96,20 @@ class DomTreeWalker {
     }
   }
 
-  private findPreviousNode(node: Node, startName: 'lastChild', siblingName: 'previousSibling', shallow?: boolean) {
+  private findPreviousNode(node: Node | undefined, shallow?: boolean): undefined | Node {
```
spocke, line 99: Bit of a nice cleanup I think I just copied the function so that is why it had params. They are clearly not needed.

modules/tinymce/CHANGELOG.md
```diff
@@ -8,10 +8,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - The `end_container_on_empty_block` option can now take a string of blocks to split when pressing Enter twice #TINY-6559
+- Added optional function parameter to `TreeWalker.prev2`, allowing aborting tree walking efforts #TINY-8592
```
spocke, line 11: This is no longer the case right so update this.

- [ ] Remove changelog item

modules/tinymce/CHANGELOG.md
```diff
@@ -8,10 +8,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 - The default value for `end_container_on_empty_block` option has been changed to `'blockquote'` #TINY-6559
 
 ### Fixed
 - Dialogs will not exceed the window height on smaller screens #TINY-8146
+- Some types on functions in `editor.dom.TreeWalker` class missed that it could return undefined #TINY-8592
```
spocke, line 16: This is not correct it should be `tinymce.dom.TreeWalker`.

- [ ] Rename to proper namespace

spocke (changes requested):

modules/tinymce/src/core/main/ts/api/dom/TreeWalker.ts
```diff
@@ -36,19 +36,19 @@ class DomTreeWalker {
    * Returns the current node.
    *
    * @method current
-   * @return {Node} Current node where the walker is.
+   * @return {Node | undefined} Current node where the walker is, or undefined if the walker has reached the end.
```
lnewson, line 39: This isn't the correct syntax for different doc types. It should be something like this:
```suggestion
   * @return {Node/undefined} Current node where the walker is, or undefined if the walker has reached the end.
```

modules/tinymce/src/core/main/ts/api/dom/TreeWalker.ts
```diff
@@ -36,19 +36,19 @@ class DomTreeWalker {
...
+  public current(): Node | undefined {
     return this.node;
   }
 
   /**
    * Walks to the next node in tree.
    *
    * @method next
-   * @return {Node} Current node where the walker is after moving to the next node.
+   * @return {Node | undefined} Current node where the walker is after moving to the next node, or undefined if the walker has reached the end.
```
lnewson, line 49:
```suggestion
   * @return {Node/undefined} Current node where the walker is after moving to the next node, or undefined if the walker has reached the end.
```

modules/tinymce/src/core/main/ts/api/dom/TreeWalker.ts
```diff
@@ -57,19 +57,19 @@ class DomTreeWalker {
    * Walks to the previous node in tree.
    *
    * @method prev
-   * @return {Node} Current node where the walker is after moving to the previous node.
+   * @return {Node | undefined} Current node where the walker is after moving to the previous node, or undefined if the walker has reached the end.
```
lnewson, line 60:
```suggestion
   * @return {Node/undefined} Current node where the walker is after moving to the previous node, or undefined if the walker has reached the end.
```

modules/tinymce/src/plugins/lists/main/ts/core/Delete.ts
```diff
@@ -210,6 +210,15 @@ const backspaceDeleteIntoListCaret = (editor: Editor, isForward: boolean): boole
     const otherLi = dom.getParent(findNextCaretContainer(editor, rng, isForward, root), 'LI', root);
 
     if (otherLi) {
+      const findValidElement = (element: SugarElement<Node>) => Arr.contains([ 'td', 'th', 'caption' ], SugarNode.name(element));
+      const findRoot = (node: SugarElement<Node>) => node.dom === root;
+      const otherLiCell = PredicateFind.closest(SugarElement.fromDom(otherLi), findValidElement, findRoot);
+      const caretCell = PredicateFind.closest(SugarElement.fromDom(rng.startContainer), findValidElement, findRoot);
+
+      if (!Optionals.equals(otherLiCell, caretCell, (a, b) => a.dom === b.dom)) {
```
lnewson, line 218:
```suggestion
      if (!Optionals.equals(otherLiCell, caretCell, Compare.eq) {
```

spocke (approved):

ltrouton (approved):

modules/tinymce/src/plugins/lists/main/ts/core/Delete.ts
```diff
@@ -37,7 +37,7 @@ const findNextCaretContainer = (editor: Editor, rng: Range, isForward: boolean,
     }
   }
 
-  while ((node = walker[isForward ? 'next' : 'prev2']())) {
+  while ((node = (isForward ? walker.next() : walker.prev2()))) {
```
lnewson, line 40:
```suggestion
  const walkFn = isForward ? walker.next : walker.prev2;
  while ((node = walkFn())) {
```
This is a little easier to read? It also saves having to do the ternary evaluation on every loop.

modules/tinymce/src/core/main/ts/api/dom/TreeWalker.ts
lnewson, line 99:
```suggestion
  private findPreviousNode(node: Node | undefined, shallow?: boolean): Node | undefined {
```
For consistency.

modules/tinymce/CHANGELOG.md
```diff
@@ -12,6 +12,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Dialogs will not exceed the window height on smaller screens #TINY-8146
+- Some types on functions in `tinymce.dom.TreeWalker` class missed that it could return undefined #TINY-8592
```
```suggestion
- Some types on functions in the `tinymce.dom.TreeWalker` class missed that it could return undefined #TINY-8592
```

modules/tinymce/CHANGELOG.md
```diff
@@ -12,6 +12,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- In some cases backspace would step into tables rather than remain outside #TINY-8592
```
lnewson, line 16:
```suggestion
- In some cases pressing the backspace key would incorrectly step into tables rather than remain outside #TINY-8592
```

lnewson (approved):

modules/tinymce/src/plugins/lists/main/ts/core/Delete.ts
lnewson: Forgot we'd need to bind `this`, so it should actually be this sorry:
```suggestion
  const walkFn = isForward ? walker.next.bind(walker) : walker.prev2.bind(walker);
  while ((node = walkFn())) {
```

PR #7744, opened 2022-04-11
TINY-8578: Newlines for more than one empty line.

HAFRMO:
Related Ticket:

Description of Changes:
Only return empty string if there is only one paragraph ( or at least, if the `body.innerText` returns zero or one newlines... ).

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Changed
+- Calling `tinymce.getContent({format: 'text'})` now returns newlines instead of empty string if more than one paragraph exists #TINY-8578
 - The `end_container_on_empty_block` option can now take a string of blocks to split when pressing Enter twice #TINY-6559
 - The default value for `end_container_on_empty_block` option has been changed to `'blockquote'` #TINY-6559
 
--- a/modules/tinymce/src/core/main/ts/content/GetContentImpl.ts
+++ b/modules/tinymce/src/core/main/ts/content/GetContentImpl.ts
@@ -23,7 +23,8 @@ const getContentFromBody = (editor: Editor, args: GetContentArgs, body: HTMLElem
     content = Tools.trim(TrimHtml.trimExternal(editor.serializer, body.innerHTML));
   } else if (args.format === 'text') {
     // return empty string for text format when editor is empty to avoid bogus elements being returned in content
-    content = editor.dom.isEmpty(body) ? '' : Zwsp.trim(body.innerText || body.textContent);
+    content = Zwsp.trim(body.innerText || body.textContent);
+    content = '\n' === content ? '' : content;
   } else if (args.format === 'tree') {
     content = editor.serializer.serialize(body, args);
   } else {
--- a/modules/tinymce/src/core/test/ts/browser/content/EditorContentTest.ts
+++ b/modules/tinymce/src/core/test/ts/browser/content/EditorContentTest.ts
@@ -94,6 +94,46 @@ describe('browser.tinymce.core.content.EditorContentTest', () => {
     assertEventsContentType();
   });
 
+  it('TINY-8578: getContent text, empty line in div', () => {
+    const editor = hook.editor();
+    editor.setContent('<div><p></p></div>');
+    const text = editor.getContent({ format: 'text' });
+    const expected = '';
+    Assertions.assertEq('Should be expected text', expected, text);
+    assertEventsFiredInOrder();
+    assertEventsContentType();
+  });
+
+  it('TINY-8578: getContent text, empty line', () => {
+    const editor = hook.editor();
+    editor.setContent('<p></p>');
+    const text = editor.getContent({ format: 'text' });
+    const expected = '';
+    Assertions.assertEq('Should be expected text', expected, text);
+    assertEventsFiredInOrder();
+    assertEventsContentType();
+  });
+
+  it('TINY-8578: getContent text, two empty lines in div', () => {
+    const editor = hook.editor();
+    editor.setContent('<div><p></p><p></p></div>');
+    const text = editor.getContent({ format: 'text' });
+    const isSafari = PlatformDetection.detect().browser.isSafari();
+    Assertions.assertEq('Should be expected text', isSafari ? '\n\n' : '\n\n\n\n', text);
+    assertEventsFiredInOrder();
+    assertEventsContentType();
+  });
+
+  it('TINY-8578: getContent text, two empty lines', () => {
+    const editor = hook.editor();
+    editor.setContent('<p></p><p></p>');
+    const text = editor.getContent({ format: 'text' });
+    const isSafari = PlatformDetection.detect().browser.isSafari();
+    Assertions.assertEq('Should be expected text', isSafari ? '\n\n' : '\n\n\n\n', text);
+    assertEventsFiredInOrder();
+    assertEventsContentType();
+  });
+
   it('TINY-6281: getContent text with empty editor', () => {
     const editor = hook.editor();
     editor.setContent('');


modules/tinymce/src/core/main/ts/content/GetContentImpl.ts
```diff
@@ -23,7 +23,8 @@ const getContentFromBody = (editor: Editor, args: GetContentArgs, body: HTMLElem
     content = Tools.trim(TrimHtml.trimExternal(editor.serializer, body.innerHTML));
   } else if (args.format === 'text') {
     // return empty string for text format when editor is empty to avoid bogus elements being returned in content
-    content = editor.dom.isEmpty(body) ? '' : Zwsp.trim(body.innerText || body.textContent);
+    content = Zwsp.trim(body.innerText || body.textContent);
+    content = '\n' === content ? '' : content;
```
spocke, line 27: I think we tend to have these flipped so `content = content === '\n' ? '' : content;`

modules/tinymce/src/core/test/ts/browser/content/EditorContentTest.ts
```diff
@@ -94,6 +94,46 @@ describe('browser.tinymce.core.content.EditorContentTest', () => {
     assertEventsContentType();
   });
 
+  it('TINY-8578: getContent text, empty line in div', () => {
+    const editor = hook.editor();
```
spocke, line 98: Maybe this could be a function like `testGetTextContent('input html', 'expected text')` since all of the tests have the same structure. isSafari could be a scoped variable then used.

This would then become:
```
it('TINY-8578: getContent text, empty line', () => testGetTextContent('<div><p></p></div>', ''));
```

modules/tinymce/src/core/main/ts/content/GetContentImpl.ts
```diff
@@ -23,7 +23,8 @@ const getContentFromBody = (editor: Editor, args: GetContentArgs, body: HTMLElem
     content = Tools.trim(TrimHtml.trimExternal(editor.serializer, body.innerHTML));
   } else if (args.format === 'text') {
     // return empty string for text format when editor is empty to avoid bogus elements being returned in content
-    content = editor.dom.isEmpty(body) ? '' : Zwsp.trim(body.innerText || body.textContent);
+    content = Zwsp.trim(body.innerText || body.textContent);
```
spocke, line 26: You can remove `textContent` here while you are at it. Firefox has implemented innerText a long time ago see: https://caniuse.com/innertext

spocke (approved):

modules/tinymce/src/core/test/ts/browser/content/EditorContentTest.ts
```diff
@@ -94,6 +94,46 @@ describe('browser.tinymce.core.content.EditorContentTest', () => {
...
+    assertEventsFiredInOrder();
+    assertEventsContentType();
+  });
+
+  it('TINY-8578: getContent text, two empty lines in div', () => {
+    const editor = hook.editor();
+    editor.setContent('<div><p></p><p></p></div>');
+    const text = editor.getContent({ format: 'text' });
+    const isSafari = PlatformDetection.detect().browser.isSafari();
+    Assertions.assertEq('Should be expected text', isSafari ? '\n\n' : '\n\n\n\n', text);
```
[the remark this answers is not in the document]
HAFRMO: Yeah, we'd need to make something quite big and complex to work around it. I'd say it's something we should look into with time. But for now we probably don't have any option than to let this issue be browser dependent, I don't think we want to ad-hoc a parser of that complexity. At least not without deciding what output is correct.

tiny-ben-tran (approved):

modules/tinymce/src/core/test/ts/browser/content/EditorContentTest.ts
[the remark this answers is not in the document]
TheSpyder: We did attempt generic "get text" logic in TBIO. It only extracts the text from text nodes, everything else it decides how many newlines to add. The logic was included in the port of libraries to TinyMCE:
https://github.com/tinymce/tinymce/blob/2d8b95517a43d1f1967a2f3898371865a2fa18e4/modules/phoenix/src/main/ts/ephox/phoenix/api/dom/DomExtract.ts#L31-L33
Which ends up in this code:
https://github.com/tinymce/tinymce/blob/2d8b95517a43d1f1967a2f3898371865a2fa18e4/modules/phoenix/src/main/ts/ephox/phoenix/extract/ExtractText.ts#L13-L17

I'm not saying we need to use this _now_, but it might be worth logging as a future JIRA improvement.

TheSpyder (approved):

modules/tinymce/src/core/test/ts/browser/content/EditorContentTest.ts
```diff
@@ -80,28 +91,20 @@ describe('browser.tinymce.core.content.EditorContentTest', () => {
...
-    const isSafari = PlatformDetection.detect().browser.isSafari();
-    const expected = 'Text to be retrieved' + (isSafari ? '\n\n' : '');
-    Assertions.assertHtml('Should be expected text', expected, text);
-    assertEventsFiredInOrder();
-    assertEventsContentType();
-  });
+  // TODO: TINY-8367 The table plugin code has been moved to core so there is now always an extra DIV in the dom for the table resize bars
+  // Safari differs in behaviour compared to the other browsers when getting text content for inline mode.
+  // When the resize bar div is included in the DOM, editor.getBody().innerText includes two extra \n at the end
+  it('TINY-6281: getContent text', () => testGetTextContent('<p>Text to be retrieved</p>', 'Text to be retrieved' + (isSafari() ? '\n\n' : '')));
```
lnewson, line 97: I looked into this more and the cause appears to be that Safari adds on the newlines at the end because it's an inline editor that uses a `<div>` element as a wrapper. If you put this same content in an iframe editor (i.e as a child of the body) then you don't end up with the newlines. I guess on the bright side, that should mean these oddities shouldn't happen as often when using a "classic" editor.

I wonder if instead of trying to use `innerText` here we should make it share the same logic as the `selection.getContent()` API and use a range over all the content to get the text instead. So this is another option we could use to improve this further, so yeah I'd suggest logging a Jira and detail both this option and the option Andy mentioned as potential solutions.

- [ ] Log a follow up Jira to improve/fix these Safari oddities

modules/tinymce/src/core/test/ts/browser/content/EditorContentTest.ts
```diff
@@ -80,28 +91,20 @@ describe('browser.tinymce.core.content.EditorContentTest', () => {
...
-    const text = editor.getContent({ format: 'text' });
-    Assertions.assertHtml('Should be expected text', '', text);
-    assertEventsFiredInOrder();
-    assertEventsContentType();
-  });
+  it('TINY-8578: getContent text, empty line in div', () => testGetTextContent('<div><p></p></div>', ''));
+
+  it('TINY-8578: getContent text, empty line', () => testGetTextContent('<p></p>', ''));
+
+  it('TINY-8578: getContent text, two empty lines in div', () => testGetTextContent('<div><p></p><p></p></div>', isSafari() ? '\n\n' : '\n\n\n\n'));
```
lnewson, line 103: With regards to this difference, it appears that Safari only counts `<p><br></p>` as a single newline, instead of 2 newlines unlike other browsers.

modules/tinymce/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Changed
+- Calling `tinymce.getContent({format: 'text'})` now returns newlines instead of empty string if more than one paragraph exists #TINY-8578
```
```suggestion
- Getting the text content of the editor now returns newlines instead of an empty string if more than one empty paragraph exists #TINY-8578
```

modules/tinymce/src/core/test/ts/browser/content/EditorContentTest.ts
```diff
@@ -58,6 +58,17 @@ describe('browser.tinymce.core.content.EditorContentTest', () => {
     assert.isTrue(isExpectedTypes);
   };
 
+  const isSafari = () => PlatformDetection.detect().browser.isSafari();
+
+  const testGetTextContent = (content: string, expected: string) => {
+    const editor = hook.editor();
+    editor.setContent(content);
+    const text = editor.getContent({ format: 'text' });
+    Assertions.assertEq('Should be expected text', expected, text);
```
lnewson, line 67: This is a legacy API that shouldn't be used anymore (only the HTML functions should be used in `Assertions`). Please use the chai assertion library instead.

- [ ] Use chai assertions here instead of the legacy custom assertion API

lnewson (approved):

modules/tinymce/src/core/test/ts/browser/content/EditorContentTest.ts
```diff
@@ -135,7 +138,7 @@ describe('browser.tinymce.core.content.EditorContentTest', () => {
     const editor = hook.editor();
     editor.setContent('<p>html</p>');
     editor.setContent('<p>new html</p>');
-    TinyAssertions.assertContent(editor, '<p>new html</p>');
+    assert.deepEqual(editor.getContent(), '<p>new html</p>');
```
lnewson, line 141: The wrong thing appears to have changed here... `TinyAssertions` is the correct thing to use here. It was the `Assertions` module above that was the issue. Also this doesn't need to be `deepEqual` it can just be `assert.equal`.

In saying that though, that part above could actually be this instead:
```
TinyAssertions.assertContent(editor, expected, { format: 'text' });
```

HAFRMO: Derp. Changed the ( hopefully ) correct one this time. And moved this one back to the original one.
No idea what I was thinking.

modules/tinymce/src/core/test/ts/browser/content/EditorContentTest.ts
lnewson: For tracking purposes, this was logged as TINY-8690.

PR #7747, opened 2022-04-14
TINY-8594: Fixed inline toolbars flickering when switching between editors

tiny-ben-tran:
Related Ticket: TINY-8594, also fixed TINY-8503

Description of Changes:
* Reverted changes that were made in #6916
* Bumped package.json version to 6.0.2

The issue occurring in a single frame makes it difficult to write a reliable test

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


lnewson (changes requested):
As I said on slack, this needs tests.

modules/tinymce/src/themes/silver/test/ts/browser/editor/toolbar/MultipleToolbarVisibilityTest.ts
```diff
@@ -0,0 +1,43 @@
...
+  };
+
+  it('TINY-8503: Does not leave two toolbars showing', async () => {
+    const editorOne = await McEditor.pFromSettings<Editor>(settings);
+    const editorTwo = await McEditor.pFromSettings<Editor>(settings);
+    editorOne.setContent('<p id="number1"><strong>blarg</strong></p>');
+    editorTwo.setContent('<p id="number2">blarg</p>');
+    editorOne.focus();
+    editorTwo.focus();
+    await Waiter.pWait(500);
```
lnewson, line 24: This is a very long wait and something we normally discourage... why exactly do we need this?

lnewson: I should add the normal preference is to wait for some state to change. A timeout should only be used where nothing changes and in that case it should be as small as possible.

tiny-ben-tran: The timeout was 100 initially but the tests were flaky. I can try another approach to avoid using timeout

TheSpyder: The problem is we're waiting for the bug to _not_ happen. The timeout that triggered the bug is removed by this PR.

tiny-ben-tran: I guess the other approach is to wait for the editor's`focus` event (if there is) and then timeout a bit before calling assertions. Might be less flaky even with a small timeout like 100.

lnewson: Yeah, from what I can see we can wait for the editor UI to be rendered first. We will likely still need a small delay to ensure nothing renders after but it would be less arbitrary and shouldn't need to be as large.

modules/tinymce/CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+- Inline toolbars flicker when switching between editors #TINY-8594
+- Multiple inline toolbars are shown after scrolling down and up #TINY-8503
```
```suggestion
- Multiple inline toolbars were shown if focused too quickly #TINY-8503
```
This wasn't caused by scrolling up and down. That was the spellchecker part which was causing the editor to be focused. Also just as a heads up, changelog entries should be past tense.

lnewson (approved):
Looks good, just please try to reduce or remove both of those timeouts.

modules/tinymce/src/themes/silver/test/ts/browser/editor/toolbar/MultipleToolbarVisibilityTest.ts
tiny-ben-tran: @TheSpyder, `timeout`s are replaced with waiting for `focus` event promises. They are no longer flaky and fail if the fix is removed.

TheSpyder: As long as it fails with the fix removed (and you can see two toolbars in the failure) I'm happy 👍

modules/tinymce/src/themes/silver/test/ts/browser/editor/toolbar/MultipleToolbarVisibilityTest.ts
```diff
@@ -0,0 +1,53 @@
+import { Assert, describe, it } from '@ephox/bedrock-client';
+import { Css } from '@ephox/sugar';
+import { McEditor, TinyDom } from '@ephox/wrap-mcagar';
+
+import Editor from 'tinymce/core/api/Editor';
+
+describe('browser.tinymce.themes.silver.editor.toolbar.MultipleToolbarVisibilityTest', () => {
```
lnewson, line 7: Should this include `Inline` in the test/file name? As this is very specifically an inline mode only issue. Maybe `MultipleInlineToolbarVisibilityTest`?

lnewson (approved):

ltrouton (approved):

PR #7749, opened 2022-04-14
TINY-8599: added selection direction to bookmarks

spocke:
Related Ticket:

Description of Changes:
* Added a extra `forward` parameter to all bookmark types where it makes sense
* Did some cleanup while I was at it to avoid local mutation or `any` types.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/tinymce/src/core/main/ts/bookmark/ResolveBookmark.ts
```diff
@@ -25,16 +33,14 @@ const addBogus = (dom: DOMUtils, node: Node): Node => {
 };
 
 const resolveCaretPositionBookmark = (dom: DOMUtils, bookmark: StringPathBookmark) => {
-  let pos;
-
   const rng = dom.createRng();
-  pos = CaretBookmark.resolve(dom.getRoot(), bookmark.start);
-  rng.setStart(pos.container(), pos.offset());
+  const spos = CaretBookmark.resolve(dom.getRoot(), bookmark.start);
```
ltrouton, line 37: For better readability, could use `sPos` or `startPos` as the variable name instead? Same for `epos` name as well

ltrouton (approved):

modules/tinymce/src/core/main/ts/bookmark/ResolveBookmark.ts
```diff
@@ -15,6 +15,14 @@ import {
 } from './BookmarkTypes';
 import * as CaretBookmark from './CaretBookmark';
 
+export interface BookmarkResolveResult {
+  range: Range;
+  forward: boolean;
```
lnewson, line 20:
```suggestion
  readonly range: Range;
  readonly forward: boolean;
```

modules/tinymce/src/core/main/ts/bookmark/ResolveBookmark.ts
lnewson: Yeah I'd agree with Luke that these should be using camel case at the very least. That said, I do like `startPos` as it's clear 🤷 

- [ ] Fix naming please

modules/tinymce/src/core/main/ts/bookmark/ResolveBookmark.ts
```diff
@@ -224,18 +230,18 @@ const resolveId = (dom: DOMUtils, bookmark: IdBookmark): Optional<Range> => {
       const rng = dom.createRng();
```
lnewson, line 230: Should we call this and other variables throughout this file `range` instead to match the resolve result property name (or change the result type to use `rng`)? That way we can shorten it to just `{ range, forward: isForwardBookmark(bookmark) }`.

lnewson (approved):
Should this have a test for the reproduction case as well? i.e. moving the caret with visual characters enabled.

- [x] Consider adding a test for the original reported issue

modules/tinymce/src/core/main/ts/bookmark/ResolveBookmark.ts
spocke: I guess we could try to setup some form of waiter that waits for the visual chars wrapping spans to appear. I avoided adding a test since it's throttled and needs to wait a while until we can check if the selection is still backwards. But I see now that it can be a wait for a specific condition. I usually hate the other kind where it needs to just wait a while for something to not happen since it's likely to flake or pass when it shouldn't pass so if I can avoid that I will add a test if I can't then I will not add a test for the original case. .

spocke:
> Should this have a test for the reproduction case as well? i.e. moving the caret with visual characters enabled.
> 
> * [x]  Consider adding a test for the original reported issue

Since moving the caret is a native operation I didn't want to involve webdriver tests. So instead I added a test that verifies that the selection direction is maintained when toggling the visualchars on/off. Internally that is what is happening on selection change anyway it seems.

PR #7750, opened 2022-04-14
TINY-8557: Improved keyboard focus states for buttons

noxuhax:
Related Ticket: TINY-8557

Description of Changes:
* Additional visual highlight for buttons with `:focus-visible` pseudoselector targeted

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


lnewson:
Should this also apply to other types of buttons? For example dialog menu buttons, such as on the find and replace dialog as it feels like that's sort of missing or inconsistent when testing locally?
image
vs
image

Then there's also other buttons such as toolbar buttons?

lnewson:
Should this also apply to other types of buttons? For example dialog menu buttons, such as on the find and replace dialog as it feels like that's sort of missing or inconsistent when testing locally?
image
vs
image

Then there's also other buttons such as toolbar buttons?

noxuhax:
@lnewson for the focus was only to improve regular, non-toolbar buttons since their focus state is the least noticeable. In the future we would want to update focus states for all elements that might need a bump in clarity, but it is hard to make one-size-fits-all solution for all possible controls we currently have in one go. This is just a first quick step that we considered brings some noticeable improvement to the user experience. Applying this solution to the toolbar buttons wouldn’t work because of their smaller size and tighter environment they’re in.

As for the Find & Replace dialog, I think it’s a bit of an oddball considering that the settings button uses toolbar button component outside of the toolbar. Maybe a better idea would be find those oddities and to make them a regular button in a separate task?

lnewson:
> @lnewson for this task focus was only to improve regular, non-toolbar buttons since their focus state is the least noticeable. In the future we would want to update focus states for all elements that might need a bump in clarity as part of the a11y improvements, but it is hard to make one-size-fits-all solution for all possible controls we currently have in one go. Applying this solution to the toolbar buttons wouldn’t work because of their smaller size and tighter environment they’re in. This is just a first quick step that we considered brings some noticeable improvement to the user experience.

Hmm, I'm certainly not a fan of inconsistent UI but ultimately it is the design teams call. I should add though I don't mind iterative approaches, but having things be insistent in the same components (e.g. a single dialog) is not something I think we should be doing.

> As for the Find & Replace dialog, I think it’s a bit of an oddball considering that the settings button uses toolbar button component outside of the toolbar. Maybe a better idea would be find those oddities and to make them a regular button in a separate task?

These are actually shared more than you might think (take a look at the editimage dialog for example). Anyways, yes I agree they ideally should have different classes, etc... to the toolbar buttons but it should not be a regular button as this isn't a regular button, it's much closer to a dropdown. Doing that in a separate task seems fine, but I am still concerned that this doesn't look great in the interim.

lnewson:
Paul DM'd me about this and to paraphrase, he basically agreed that we shouldn't have an inconsistent UI but he was also concerned about how much it would take to fix. So his preference was to proceed with this solution for now even though we won't know when the other part would be fixed.

So that's design's stance here by the looks, so in that case can I please just ask that we ensure a Jira gets logged to fix the dialog menu footer button?

- [ ] Log Jira for the dialog menu footer style issues

TheSpyder (approved):

tiny-ben-tran (approved):

PR #7751, opened 2022-04-14
fix: base64 encoded tiny editor as iframe src bug

Don-Cryptus:
Related Ticket:

Description of Changes:
if tinymce is shown in an iframe with src being data url it breaks, because it doesnt have a real url so the path spliting is broken

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


metricjs:
HI @Don-Cryptus,

Thanks for this PR. Could you please give us some more information about what the issue is and what behaviour it causes? A replication case would be great - would it be possible for you to create a [Fiddle](http://fiddle.tinymce.com/)?

Don-Cryptus:
I cant exactly replicate it on fiddle, but it would look something like [this](https://fiddle.tiny.cloud/1biaab/2).
I made a new github repo where I am using webpack to put everything into one html file convert the file to base64 and put it as iframe src. [ github repo](https://github.com/Don-Cryptus/tinyMCE-iframe)

my example repo works like this open the ```tinymce iframe.html``` select either the fixed or original html and it will set the source of the iframe with the encoded string my forked repo has the fix already implemented so the fixed index.html shows the working editor but the orignal will through the error.

PR #7758, opened 2022-04-18
TINY-8572: Fix incorrect word wrapping in menu dropdowns

noxuhax:
Related Ticket: TINY-8572

Description of Changes:
* Changed word-breaking behaviour for collection items inside menu component so it would only break on spaces

Pre-checks:
* [X] Changelog entry added
* [X] ~Tests have been added (if applicable)~

Review:
* [x] Milestone set


ltrouton (approved):

HAFRMO (approved):

lnewson (changes requested):
As noted on slack, this was done because without this in some languages content ends up going out of view on smaller devices (such as mobiles). Here's an example of what this looks like on mobile with German:
image

So whatever solution we come up with we need to make sure to account for/cater for that use case as well.

lnewson (approved):

modules/oxide/src/less/theme/globals/global-variables.less
```diff
@@ -62,5 +62,7 @@
 @z-index-menu: 1150;
 
 // Responsive breakpoints
+@breakpoint-sm-value: 767px;
 @breakpoint-md: ~"only screen and (min-width:1000px)";
-@breakpoint-sm: ~"only screen and (max-width:767px)";
+@breakpoint-sm: ~"only screen and (max-width:" @breakpoint-sm-value ~")";
+@breakpoint-lt-sm: ~"only screen and (min-width:" @breakpoint-sm-value + 1px ~")";
```
TheSpyder, line 68: What is `lt-sm` supposed to indicate? Because my first impression was it meant "less than small", but this is greater than 😉

TheSpyder (approved):

modules/oxide/src/less/theme/globals/global-variables.less
noxuhax: Well that’s embarrassing 😅

modules/oxide/src/less/theme/components/menu/menu.less
```diff
@@ -35,6 +35,15 @@
     padding: @menu-item-collection-grid-padding;
   }
 
+  @media @breakpoint-lt-sm {
```
lnewson, line 38: This needs updating to match the new name done in 517e80ca76eb817cae000d016d64a15544b74711

PR #7770, opened 2022-04-21
TINY-5947: Fixed images not showing as selected when selecting all content

### LINKED ISSUE #5450: set image selection
behavior: 
can't set img selection style

steps: 
Ctrl+A or Cmd+A can't set img selection style

image

expected behavior:
want to be able to set the image selection style

versions:
V5.1.3

lnewson:
Related Ticket: TINY-5947

Description of Changes:

This fixes an issue where the browser selection would never show for images when more than just the image was selected. The main cause is the CSS we used, however to make that work without flickering it required a change to how we do the selection as we need to ensure the `data-mce-selected` attribute is set immediately, not in the next event loop. This does mean a little bit of the logic runs more often on node change or similar events, but I've tried to keep it minimised so as to not cause performance problems.

Note: I also did some very minor cleanup while I was messing with the code.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

Fixes #5450

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The default value for `end_container_on_empty_block` option has been changed to `'blockquote'` #TINY-6559
 
 ### Fixed
-- Images were not showing as selected when selecting images with alongside other content #TINY-5947
+- Images were not showing as selected when selecting images alongside other content #TINY-5947
 - Dialogs will not exceed the window height on smaller screens #TINY-8146
 - Text alignment could not be applied to `pre` elements #TINY-7715
 
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The default value for `end_container_on_empty_block` option has been changed to `'blockquote'` #TINY-6559
 
 ### Fixed
+- Images were not showing as selected when selecting all content #TINY-5947
 - Dialogs will not exceed the window height on smaller screens #TINY-8146
 - Text alignment could not be applied to `pre` elements #TINY-7715
 
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The default value for `end_container_on_empty_block` option has been changed to `'blockquote'` #TINY-6559
 
 ### Fixed
-- Images were not showing as selected when selecting all content #TINY-5947
+- Images were not showing as selected when selecting images with alongside other content #TINY-5947
 - Dialogs will not exceed the window height on smaller screens #TINY-8146
 - Text alignment could not be applied to `pre` elements #TINY-7715
 
--- a/modules/tinymce/src/core/main/ts/api/dom/ControlSelection.ts
+++ b/modules/tinymce/src/core/main/ts/api/dom/ControlSelection.ts
@@ -39,6 +39,23 @@ interface SelectedResizeHandle extends ResizeHandle {
   };
 }
 
+const elementSelectionAttr = 'data-mce-selected';
+const controlElmSelector = 'table,img,figure.image,hr,video,span.mce-preview-object';
+const abs = Math.abs;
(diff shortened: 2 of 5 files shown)


ltrouton (approved):

modules/tinymce/src/core/test/ts/browser/dom/ControlSelectionTest.ts
```diff
@@ -187,4 +188,14 @@ describe('browser.tinymce.core.dom.ControlSelectionTest', () => {
     const wrapper = UiFinder.findIn(TinyDom.body(editor), 'span.mce-preview-object').getOrDie();
     getAndAssertDimensions(wrapper, 402 + 100, 202 + 50);
   });
+
+  it('TINY-5947: data-mce-selected should be set synchronously when selecting control elements', async () => {
```
spocke, line 192: This also checks the async nature of the resize handles so maybe either spit this up to two tests or change the description. Should it maybe also check that they are not present first sync then check if they are present after a while with the waiter.

modules/tinymce/src/core/main/ts/api/dom/ControlSelection.ts
```diff
@@ -408,38 +403,36 @@ const ControlSelection = (selection: EditorSelection, editor: Editor): ControlSe
...
-    // Ignore all events while resizing or if the editor instance was removed
-    if (resizeStarted || editor.removed) {
+  const updateResizeRect = (e: EditorEvent<unknown>) => {
+    // Ignore all events while resizing, if the editor instance is composing or the editor was removed
+    if (resizeStarted || editor.removed || editor.composing) {
       return;
     }
 
+    const targetElm = e.type === 'mousedown' ? e.target : selection.getNode();
+    const controlElm = SelectorFind.closest<HTMLElement>(SugarElement.fromDom(targetElm), 'table,img,figure.image,hr,video,span.mce-preview-object')
```
spocke, line 416: While you are at it maybe pull this hard coded selector out to a constant since it a pretty long line. There are a few other constants in the constructor scope unsure if it would improve the file if we moved then out to module scope.

modules/tinymce/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The default value for `end_container_on_empty_block` option has been changed to `'blockquote'` #TINY-6559
 
 ### Fixed
+- Images were not showing as selected when selecting all content #TINY-5947
```
spocke, line 14: This also changes how images look while selecting around contents. Like `a[ab]b` so there should be a changelog item about that change. It now shows a proper background for those as it should so it feels like a bug fix to me.

- [ ] Add changelog item about background selection change for images

spocke (approved):

modules/tinymce/CHANGELOG.md
lnewson: I did consider this originally, however I ended up not including it. I don't think this should be a separate changelog though and instead should just be a single changelog entry so I'll look at updating the existing one instead.

modules/tinymce/src/core/test/ts/browser/dom/ControlSelectionTest.ts
lnewson: I probably should just have left the last line out, but I threw it in just as a double check they still do render at some stage. The main thing this is testing is the sync part and we don't really care if the resize handles are rendered immediately or in say the following event loop. So I don't really think anything should change here, but if you really want something done I'll just remove the wait for the handles at the end.

modules/tinymce/src/core/main/ts/api/dom/ControlSelection.ts
lnewson: Done everything that I thought made sense (I left the `rootDocument` as it was clearer beside `rootElement`).

modules/tinymce/CHANGELOG.md
lnewson: Have reworked the changelog to try and address this. How does that look?

modules/tinymce/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The default value for `end_container_on_empty_block` option has been changed to `'blockquote'` #TINY-6559
 
 ### Fixed
+- Images were not showing as selected when selecting images with alongside other content #TINY-5947
```
```suggestion
- Images were not showing as selected when selecting images alongside other content #TINY-5947
```

TheSpyder (approved):

PR #7775, opened 2022-04-21
TINY-8647: Fix custom toolbar select buttons width on touch devices

noxuhax:
Related Ticket: TINY-8647

Description of Changes:
* Replace fixed width for toolbar buttons in selects with decreased padding on touch devices

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/oxide/src/less/theme/components/toolbar-button/toolbar-select-button.less
```diff
@@ -37,4 +38,13 @@
...
   }
+
+  // Correct specific styles for split button on touch devices
+  // See toolbar-split-button.less
+  &.tox-platform-touch {
+    .tox-split-button .tox-tbtn.tox-tbtn--select.tox-tbtn--select {
+      padding: 0 (@toolbar-button-select-padding - @toolbar-split-button-chevron-touch-padding);
+      width: auto;
+    }
+  }
```
lnewson, line 49: This really seems like the wrong place for this. It should be inside `toolbar-split-button.less` since the button is still a split button, not a regular one.

lnewson: Additionally, a better solution I believe would be to make the existing rule not match instead. For example adding `:not(.tox-tbtn--select)`:
```
  // Increase the chevron size and decrease the button size on touch devices
  &.tox-platform-touch {
    .tox-split-button .tox-tbtn:not(.tox-tbtn--select):first-child {
      width: @toolbar-button-width - @toolbar-split-button-chevron-touch-padding;
    }

    .tox-split-button__chevron {
      width: @toolbar-button-chevron-width + @toolbar-split-button-chevron-touch-padding;
    }
  }
```

lnewson (changes requested):
This fix also needs a changelog entry.

modules/oxide/src/less/theme/components/toolbar-button/toolbar-select-button.less
noxuhax: Agree on both cases. I did not initially took into account existence of toolbar select buttons on their own, outside of split buttons.

modules/tinymce/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Inline toolbars flickered when switching between editors #TINY-8594
 - Multiple inline toolbars were shown if focused too quickly #TINY-8503
 - Added background and additional spacing for the text labeled buttons in the toolbar to improve visual clarity #TINY-8617
+- Fix custom toolbar select buttons width on touch devices #TINY-8647
```
metricjs, line 14:
```suggestion
- Toolbar split buttons with text used an incorrect width on touch devices #TINY-8647
```

modules/oxide/src/less/theme/components/toolbar-button/toolbar-split-button.less
```diff
@@ -40,23 +40,27 @@
...
   }
 
+  .tox-split-button.tox-tbtn--disabled:hover,
+  .tox-split-button.tox-tbtn--disabled:focus,
+  .tox-split-button.tox-tbtn--disabled .tox-tbtn:hover,
+  .tox-split-button.tox-tbtn--disabled .tox-tbtn:focus {
+    background: @toolbar-button-disabled-background-color;
+    box-shadow: @toolbar-button-disabled-box-shadow;
+    color: @toolbar-button-disabled-text-color;
+  }
```
lnewson, line 50: This doesn't seem related to this Jira so should it have been included? If so then it's not working as expected with keyboard navigation because the inner button never gets focus and its background overrides the parent one.

- [ ] Fix issues fully or split out into a separate Jira

lnewson (approved):

modules/oxide/src/less/theme/components/toolbar-button/toolbar-split-button.less
lnewson: This maybe should also have a changelog entry if you keep it, since it looks to be trying to fix focus style issues when the button is disabled.

lnewson: Ignore this sorry, as @ltrouton pointed out on slack this was just moved from below instead.

ltrouton (approved):

modules/tinymce/CHANGELOG.md
lnewson: Accepting this sorry Alex, as we'd like to get ready for the release. If there's any issues with it let us know and we'll make sure to correct it before we actually release.

PR #7780, opened 2022-04-24
TINY-8601: Fixed inline boundaries keyboard navigation with ranged selections

vpyshnenko:
Related Ticket: TINY-8601

Description of Changes:
* Fold current editor selection to one of its edges before passing it to inline boundaries handling

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/tinymce/CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - In some cases pressing the backspace key would incorrectly step into tables rather than remain outside #TINY-8592
 - Text alignment could not be applied to `pre` elements #TINY-7715
 - The `InsertParagraph` or `mceInsertNewLine` commands did not delete the current selection like the native command used to #TINY-8606
+- The caret moved before the selected inline boundary element when the right arrow key pressing #TINY-8601
```
spocke, line 19:
```suggestion
- The caret moved before the selected inline boundary element when the right arrow key was pressed #TINY-8601
```

or
```suggestion
- The caret moved before the selected inline boundary element when pressing the right arrow key #TINY-8601
```
or
```suggestion
- The caret moved before the selected inline boundary element when using the right arrow key #TINY-8601
```

Writing changelog items is hard. Almost as hard as naming things. :smile: not sure if my suggestions are good.

modules/tinymce/src/core/test/ts/browser/keyboard/ArrowKeysInlineBoundariesTest.ts
```diff
@@ -336,4 +336,27 @@ describe('browser.tinymce.core.keyboard.ArrowKeysInlineBoundariesTest', () => {
       TinyAssertions.assertCursor(editor, [ 0, 0 ], 0);
     });
   });
+
+  context('Arrow keys on inline boundary node when selection is not collapsed', () => {
+    it('TINY-8601: It should move caret before selected inline boundary node when arrow left', () => {
+      const editor = hook.editor();
+      editor.setContent('<p>test <span class="mce-annotation">span</span> selection</p>', { format: 'raw' });
+      TinySelections.select(editor, 'p', [ 1 ]);
```
spocke, line 344: If you use the `setSelection` api then it has a extra parameter to fire a node change. So maybe change this to:
`TinySelections.setSelection(editor, [0], 1, [0], 2, true);`
But then again the `TinySelections.select` call here is probably easier to understand than the paths above not sure. I just tend to like using the `setSelection` api my self since it's more exact of what it's doing you kind of see both end points where they are going.

spocke (approved):

modules/tinymce/CHANGELOG.md
lnewson:
```suggestion
- When pressing the right arrow key, the caret incorrectly moved before any selected inline boundary element #TINY-8601
```
Is another option that flips it a little?

modules/tinymce/src/core/main/ts/keyboard/BoundarySelection.ts
```diff
@@ -35,9 +36,19 @@ const renderCaretLocation = (editor: Editor, caret: Cell<Text>, location: Bounda
     return location;
   });
 
+const getPositionFromRange = (range: Range, root: HTMLElement, forward: boolean): CaretPosition => {
+  const start = CaretPosition.fromRangeStart(range);
+  if (range.collapsed) {
+    return start;
+  } else {
+    const end = CaretPosition.fromRangeEnd(range);
+    return forward ? CaretFinder.prevPosition(root, end).getOr(end) : CaretFinder.nextPosition(root, start).getOr(start);
```
lnewson, line 45: I haven't looked into this much yes, but is this correct? I would have expected the start position to be the same regardless of if it was collapsed. This instead will make it move back inside the inline boundary element.

modules/tinymce/src/core/test/ts/browser/keyboard/ArrowKeysInlineBoundariesTest.ts
```diff
@@ -336,4 +336,27 @@ describe('browser.tinymce.core.keyboard.ArrowKeysInlineBoundariesTest', () => {
       TinyAssertions.assertCursor(editor, [ 0, 0 ], 0);
     });
   });
+
+  context('Arrow keys on inline boundary node when selection is not collapsed', () => {
```
lnewson, line 340: This is quite verbose and just duplicates the information (e.g. we already know this test is about arrow keys on inline boundaries). As such, maybe it should be something like this (while trying to keep it similar to what's already there)?
```suggestion
  context('Arrow keys when selection is not collapsed', () => {
```

modules/tinymce/src/core/test/ts/browser/keyboard/ArrowKeysInlineBoundariesTest.ts
```diff
@@ -336,4 +336,27 @@ describe('browser.tinymce.core.keyboard.ArrowKeysInlineBoundariesTest', () => {
...
+    it('TINY-8601: It should move caret before selected inline boundary node when arrow left', () => {
```
lnewson, line 341: We shouldn't start this with `it`, so maybe something like this?
```suggestion
    it('TINY-8601: should move caret before selected inline boundary node when arrow left', () => {
```

modules/tinymce/src/core/test/ts/browser/keyboard/ArrowKeysInlineBoundariesTest.ts
```diff
@@ -336,4 +336,27 @@ describe('browser.tinymce.core.keyboard.ArrowKeysInlineBoundariesTest', () => {
...
+      const editor = hook.editor();
+      editor.setContent('<p>test <span class="mce-annotation">span</span> selection</p>', { format: 'raw' });
+      TinySelections.select(editor, 'p', [ 1 ]);
+      editor.nodeChanged();
+      TinyContentActions.keystroke(editor, Keys.left());
+      TinyAssertions.assertCursor(editor, [ 0, 0 ], 5);
+      assertCaretAtZwsp(editor);
+    });
+
+    it('TINY-8601: It should move caret after selected inline boundary node when arrow right', () => {
```
lnewson, line 351:
```suggestion
    it('TINY-8601: should move caret after selected inline boundary node when arrow right', () => {
```

modules/tinymce/src/core/main/ts/keyboard/BoundarySelection.ts
vpyshnenko: Without rebasing position to previous one  caret can not leave inline boundary element

https://user-images.githubusercontent.com/44773772/165530437-afba2b99-4d8f-43ca-ba3d-0ae7f7107e1a.mov

lnewson: I tested this locally and the current behaviour with these changes isn't correct. With these changes it causes it to effectively just collapse the selection, but it shouldn't be doing that it should be moving to the next viable caret location. So lets say I have:
```html
<p>Some <a href="#">content</a> here.</p>
```
So in the above, lets say we select the inner content of the anchor. For example (where `|` is the selection):
```html
<p>Some <a href="#">|content|</a> here.</p>
```
Now, when I press the left key the selection should be in the before the anchor element for example:
```html
<p>Some |<a href="#">content</a> here.</p>
```
Instead all it's doing is collapsing it so it's still inside the anchor at the very first caret position.

Likewise in reverse, if we pressed the right arrow key instead then I'd expect it to end up here:
```html
<p>Some <a href="#">content</a>| here.</p>
```

I haven't had a chance to look into this more, but something needs to change here to fix the above. We also should add tests for cases like the above, as the current ones only handle if the outside edges of the anchor are selected but not the inside edges. So it'd need to be something like this:
```
editor.setContent('<p>test <span class="mce-annotation">span</span> selection</p>', { format: 'raw' });
TinySelections.setSelection(editor, [ 0, 1, 0 ], 0, [ 0, 1, 0 ], 4, true);
TinyContentActions.keystroke(editor, Keys.left());
TinyAssertions.assertCursor(editor, [ 0, 0 ], 5);
```

- [ ] Add tests for the cases above and fix issues

modules/tinymce/src/core/test/ts/browser/keyboard/ArrowKeysInlineBoundariesTest.ts
```diff
@@ -336,4 +336,25 @@ describe('browser.tinymce.core.keyboard.ArrowKeysInlineBoundariesTest', () => {
       TinyAssertions.assertCursor(editor, [ 0, 0 ], 0);
     });
   });
+
+  context('Arrow keys when selection is not collapsed', () => {
+    it('TINY-8601: should move caret before selected inline boundary node when arrow left', () => {
+      const editor = hook.editor();
+      editor.setContent('<p>test <span class="mce-annotation">span</span> selection</p>', { format: 'raw' });
```
lnewson, line 343: I don't believe you need `format: 'raw'` here. Was there a reason that's being done?

modules/tinymce/src/core/main/ts/keyboard/BoundarySelection.ts
lnewson: Actually, really sorry @vpyshnenko. In my haste I realised I've got this backwards and your current behaviour is correct for the case mentioned above (i.e. it should collapse, which is what the browser does natively as well).

Anyways, it'd probably still be good to add some tests for where the selection is inside and then we press left/right if you don't mind?

modules/tinymce/src/core/test/ts/browser/keyboard/ArrowKeysInlineBoundariesTest.ts
vpyshnenko: The only reason is that I tried to follow previous tests style.

HAFRMO (approved):

modules/tinymce/src/core/test/ts/browser/keyboard/ArrowKeysInlineBoundariesTest.ts
```diff
@@ -336,4 +336,43 @@ describe('browser.tinymce.core.keyboard.ArrowKeysInlineBoundariesTest', () => {
...
+      TinyAssertions.assertCursor(editor, [ 0, 2 ], 1);
+      assertCaretAfterZwsp(editor);
+    });
+
+    it('TINY-8601: should collapse inner text selection to the start when arrow left', () => {
+      const editor = hook.editor();
+      editor.setContent('<p>test <span class="mce-annotation">span</span> selection</p>');
+      TinySelections.setSelection(editor, [ 0, 1, 0 ], 0, [ 0, 1, 0 ], 4, true);
+      TinyContentActions.keystroke(editor, Keys.left());
+      TinyAssertions.assertCursor(editor, [ 0, 1, 0 ], 1);
```
lnewson, line 364: This assertion isn't right... it should be collapsing the selection or moving it outside the span, not moving it one position to the right (especially when we press the left arrow key).

lnewson: Oh wait... is this the ZWSP text node that it's on? If so that'd make some sense, in which case can you please add a comment to clarify that.

vpyshnenko: I've added the comment

lnewson: Thankyou!

PR #7781, opened 2022-04-25
TINY-8588: NBSP characters weren't preserved near cef block elements

HAFRMO:
Related Ticket:

Description of Changes:
NBSP would not be preserved in some situations with CEF-blocks that were styled as blocks.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/tinymce/src/core/test/ts/browser/content/EditorNbspTest.ts
```diff
@@ -0,0 +1,35 @@
+import { describe, it } from '@ephox/bedrock-client';
+import { TinyAssertions, TinyHooks, TinySelections } from '@ephox/wrap-mcagar';
+
+import Editor from 'tinymce/core/api/Editor';
+
+describe('tinymce.src.core.test.ts.browser.content.EditorNbspTest', () => {
```
lnewson, line 6:
```suggestion
describe('browser.tinymce.core.content.EditorNbspTest', () => {
```
This is named incorrectly.

lnewson: Also these tests should be in the `SpaceKeyTest` instead, as that's where all the existing tests are.

- [ ] Move the tests to the `SpaceKeyTest.ts` file

modules/tinymce/src/core/main/ts/keyboard/Nbsps.ts
```diff
@@ -54,6 +55,25 @@ const isAtLineBoundary = (root: SugarElement, pos: CaretPosition) => (
...
+const isCefBlock = (node?: Node) =>
+  Type.isNonNullable(node) && NodeType.isContentEditableFalse(node) && isBlockLike(node);
+
+const isBeforeCefBlock = (root: SugarElement, pos: CaretPosition) => {
+  if (pos.isAtEnd()) {
+    return isCefBlock(new DomTreeWalker(pos.container(), root.dom).next());
+  } else {
+    return false;
+  }
+};
```
spocke, line 67: I guess you could just make this a one liner the same goes with the `isAfterCefBlock`.

```suggestion
const isBeforeCefBlock = (root: SugarElement, pos: CaretPosition) =>
  pos.isAtEnd() && isCefBlock(new DomTreeWalker(pos.container(), root.dom).next());
```

modules/tinymce/src/core/main/ts/keyboard/Nbsps.ts
```diff
@@ -54,6 +55,25 @@ const isAtLineBoundary = (root: SugarElement, pos: CaretPosition) => (
     isBeforeBr(root, pos)
 );
 
+const isCefBlock = (node?: Node) =>
+  Type.isNonNullable(node) && NodeType.isContentEditableFalse(node) && isBlockLike(node);
```
spocke, line 59: I think this affects more than just cef elements I think the space will not be rendered for any block like element. If that is the case then we should make a more general fix and just call isBlockLike without the cef check and rename the function here to `isBlockLikeNode` or something.

HAFRMO: No, the issue is that the cursor does not go into cef blocks while treewalking, resulting in this issue. There's already things that will do this for other ones, it just didn't do it for cef.

modules/tinymce/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The default value for `end_container_on_empty_block` option has been changed to `'blockquote'` #TINY-6559
 
 ### Fixed
+- Spaces would not be added or shown on Chrome in certain situations #TINY-8588
```
```suggestion
- Spaces would not be added correctly on some browsers when before or after a contenteditable block element #TINY-8588
```
This would also happen on Edge and I suspect Safari, so this might be better?

lnewson (approved):

ltrouton (approved):

PR #7784, opened 2022-04-26
TINY-8663: Fixed some tests that failed to run in isolation

lnewson:
Related Ticket: TINY-8663

Description of Changes:
* Fixed various tests that failed to run in isolation. This for example then caused flakes when used with bedrock's chunking.

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/themes/silver/test/ts/browser/editor/core/UndoRedoTest.ts
+++ b/modules/tinymce/src/themes/silver/test/ts/browser/editor/core/UndoRedoTest.ts
@@ -18,7 +18,7 @@ describe('browser.tinymce.themes.silver.editor.core.UndoRedoTest', () => {
   const assertMenuItemState = (editor: Editor, item: string, disabled: boolean) =>
     TinyUiActions.pWaitForUi(editor, `div[title="${item}"][role="menuitem"][aria-disabled="${disabled}"]`);
 
-  const waitForToolbarState = async (editor: Editor, disabled: boolean) => {
+  const pWaitForToolbarState = async (editor: Editor, disabled: boolean) => {
     const overlord = UiFinder.findIn(SugarBody.body(), '.tox-toolbar-overlord').getOrDie();
     await Waiter.pTryUntil(
       'Waiting for toolbar state',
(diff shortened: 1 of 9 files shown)


tiny-ben-tran (approved):

ltrouton (approved):

modules/tinymce/src/themes/silver/test/ts/browser/editor/core/UndoRedoTest.ts
```diff
@@ -47,7 +47,8 @@ describe('browser.tinymce.themes.silver.editor.core.UndoRedoTest', () => {
   it('TINY-8101: Undo/redo should be disabled/enabled if dirty', async () => {
     const editor = hook.editor();
     // 1. Insert content, undo should be enabled and redo should be still disabled
-    editor.insertContent('<p>slsl</p>');
+    editor.resetContent('');
+    editor.insertContent('<p>test</p>');
```
TheSpyder, line 51: Are we reusing tinymce instances across an entire `describe` block? I can see how that would be faster, but it will only encourage cross-test state accidents 🤔

TheSpyder (approved):

modules/tinymce/src/themes/silver/test/ts/browser/editor/core/UndoRedoTest.ts
lnewson: Yes we are in general, as it's far too slow to setup a new editor for each of the 4k+ tests. I've often wondered if we should include the reset logic for between each test in McAgar but then some tests wouldn't work with that either.

spocke (approved):

modules/tinymce/src/themes/silver/test/ts/browser/editor/core/UndoRedoTest.ts
TheSpyder: Interesting. I'm pretty sure we did re-init in every test for TBIO - I wonder whether it was just faster to load, or whether it had less overall tests. The lack of plugins and customisability perhaps made it quite easy to load.

lnewson: One thing to remember is that TBIO used pipeline tests, so each "test" wasn't actually a single unit and was instead multiple units so it may have had the same issue where it just reused the editor for multiple test units. I can't say for sure as I never looked at them in depth, but many tests used to do that type of thing.

PR #7792, opened 2022-04-28
TINY-8349: Convert xhr to fetch where possible

ltrouton:
Related Ticket: TINY-8349

Description of Changes:
* Convert native uses of xhr to fetch

There is still one use of xhr in the repo but since it requires progress state and fetch does not provide a simple way to do this, it is better to leave that as is for now

Pre-checks:
* [X] ~~Changelog entry added~~ From an integrator perspective nothing should have changed so I am not sure if a changelog is needed
* [X] Tests have been added (if applicable) Looked like there was sufficient tests already

Review:
* [X] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/file/Conversions.ts
+++ b/modules/tinymce/src/core/main/ts/file/Conversions.ts
@@ -12,38 +12,10 @@ interface DataUriResult {
   readonly data: string;
 }
 
-const blobUriToBlob = (url: string): Promise<Blob> => {
-  return new Promise((resolve, reject) => {
-
-    const rejectWithError = () => {
-      reject('Cannot convert ' + url + ' to Blob. Resource might not exist or is inaccessible.');
-    };
-
-    try {
-      const xhr = new XMLHttpRequest();
-
-      xhr.open('GET', url, true);
-      xhr.responseType = 'blob';
-
-      xhr.onload = () => {
-        if (xhr.status === 200) {
-          resolve(xhr.response);
-        } else {
-          // IE11 makes it into onload but responds with status 500
-          rejectWithError();
-        }
-      };
-
-      // Chrome fires an error event instead of the exception
-      // Also there seems to be no way to intercept the message that is logged to the console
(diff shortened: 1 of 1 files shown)


modules/tinymce/src/core/main/ts/file/Conversions.ts
```diff
@@ -12,38 +12,10 @@ interface DataUriResult {
...
-
-      xhr.send();
-    } catch (ex) {
-      rejectWithError();
-    }
-  });
-};
+const blobUriToBlob = (url: string): Promise<Blob> =>
+  fetch(url)
+    .then((res) => res.ok ? res.blob() : Promise.reject())
```
lnewson, line 17: This isn't quite a 1:1 replacement as this will allow any 2xx status code, whereas previously it only allowed 200. Do we care?

lnewson (approved):

modules/tinymce/src/core/main/ts/file/Conversions.ts
ltrouton: When I removed the XHR API in #7534 I used `res.ok` as replacement for `xhr.status === 200` where needed. I am inclined to say using `res.ok` is fine given that is the recommended way for `fetch` but I am by no means an expert on status codes so could be missing something where this could lead to issues.

metricjs (approved):

modules/tinymce/src/core/main/ts/file/Conversions.ts
spocke: Any ok status code is file here since this is for converting blob urls to blobs its very unlikely that it would produce other 200 codes there since it's all data urls not real http urls.

spocke (approved):

modules/tinymce/src/core/main/ts/file/Conversions.ts
lnewson: Yeah I didn't expect this to be an issue tbh, was more so just highlighting it 😄

PR #7798, opened 2022-04-29
TINY-8625: Can now reset multiple cells' height, or other values.

HAFRMO:
Related Ticket:

Description of Changes:
Removing values from the dialogs would not remove them when multiple cells/rows were selected.

Pre-checks:
* [x] Changelog entry added

Review:
* [ ] Milestone set

### DIFF
--- a/modules/tinymce/src/plugins/table/main/ts/ui/CellDialog.ts
+++ b/modules/tinymce/src/plugins/table/main/ts/ui/CellDialog.ts
@@ -38,18 +38,34 @@ const getSelectedCells = (table: SugarElement<HTMLTableElement>, cells: SugarEle
   }));
 };
 
-const updateSimpleProps = (modifier: DomModifier, colModifier: DomModifier, data: CellData): void => {
-  modifier.setAttrib('scope', data.scope);
-  modifier.setAttrib('class', data.class);
-  modifier.setStyle('height', Utils.addPxSuffix(data.height));
-  colModifier.setStyle('width', Utils.addPxSuffix(data.width));
+const updateSimpleProps = (modifier: DomModifier, colModifier: DomModifier, data: CellData, shouldUse: (key: string) => boolean): void => {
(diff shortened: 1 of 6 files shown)


modules/tinymce/src/plugins/table/main/ts/ui/CellDialog.ts
```diff
@@ -38,18 +38,34 @@ const getSelectedCells = (table: SugarElement<HTMLTableElement>, cells: SugarEle
...
 
-const updateSimpleProps = (modifier: DomModifier, colModifier: DomModifier, data: CellData): void => {
-  modifier.setAttrib('scope', data.scope);
-  modifier.setAttrib('class', data.class);
-  modifier.setStyle('height', Utils.addPxSuffix(data.height));
-  colModifier.setStyle('width', Utils.addPxSuffix(data.width));
+const updateSimpleProps = (modifier: DomModifier, colModifier: DomModifier, data: CellData, shouldUse: (key: string) => boolean): void => {
+  if (shouldUse('scope')) {
+    modifier.setAttrib('scope', data.scope);
+  }
```
TheSpyder, line 44: I wish we could swap cell data to optional values instead of doing this, but that seems a little much for this fix.

TheSpyder (approved):

lnewson (changes requested):
According to the Jira this is meant to be targeted at 6.0.3, so this needs to be recreated as a `hotfix`, rebased and target `master` instead. This is why setting the milestone when creating a PR can be useful 😉

modules/tinymce/src/plugins/table/main/ts/ui/CellDialog.ts
HAFRMO: We'd still need to do this though, since "No data" is valid input here. We're checking if the data was changed, nothing else.

PR #7804, opened 2022-05-04
TINY-8646: Fix base64 encoded image not displayed following `data:` text

### LINKED ISSUE #7737: base64 encoded images are not displayed following "data:" typed as text -- INCLUDES FIX
From a blank document, type "My data:" then either on the same line or the next paragraph insert an image.
It appears to work fine, but the html that will be generated will not open in tinymce correctly the next time.

Now, on page load set the Tinymce content to that html.
```
<p>My data:</p>
<p><img src="data:image/png;base64,abcdefg..." ...></p>
```

When Tinymce receives that valid html, it winds up html escaping everything between the first "data:" and the base64 part, so the editor appears to show raw html.

**Please provide the steps to reproduce and if possible a minimal demo of the problem via [fiddle.tiny.cloud](https://fiddle.tiny.cloud/) or similar.**
https://fiddle.tiny.cloud/taiaab

If the text typed is slightly different like `data :` or `data:;` the bug does not happen, the expected behavior is to behave like that.

Tested in version 5.  It happens in all browsers.  I'm looking for the fix in version 5

**The Fix**
This PR 7098 touches the same line and a PR for this fix can be modeled after it.  But I'm not familiar with the testing requirements.
https://github.com/tinymce/tinymce/pull/7098

The first line of the `extractBase64DataUris` function has a regex that starts with `/data:[^;]+;`.
If you insert one more character so it starts with `/data:[^;<]+;` that fixes this bug.
The logic there is wherever "data:" is used correctly, the greater than and less than characters will be required to be escaped.  So to prevent crossing a `<` prevents it from selecting into a separate html tag.

To be consistent with that PR 7098, one line in the `parseDataUri` function should get the same edit as well.

tiny-ben-tran:
Related Ticket: TINY-8646

Description:
- Updated the regex in `Base64Uris()` function as suggested in the ticket

**Note:** The bug seems to be fixed without updating `parseDataUri()` as suggested in the fix. I just leave it as is

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

Fixes #7737

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -6,10 +6,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
-## 5.10.5 - 2022-05-05
-
 ### Fixed
-- Base64 data URI was not extracted correctly following `data:` text #TINY-8646
+- Base64 data URIs were not extracted correctly during parsing when proceded by `data:` text #TINY-8646
 
 ## 5.10.4 - 2022-04-27
 
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Fixed
-- Base64 data URIs were not extracted correctly during parsing when proceded by `data:` text #TINY-8646
+- Base64 data URIs were not extracted correctly during parsing when proceeded by `data:` text #TINY-8646
 
 ## 5.10.4 - 2022-04-27
 
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+- Base64 encoded image was not displayed following `data:` text #TINY-8646
+
 ## 5.10.4 - 2022-04-27
 
 ### Fixed
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -6,8 +6,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## 5.10.5 - 2022-05-05
+
 ### Fixed
-- Base64 encoded image was not displayed following `data:` text #TINY-8646
+- Base64 data URI was not extracted correctly following `data:` text #TINY-8646
 
 ## 5.10.4 - 2022-04-27
 
--- a/modules/tinymce/src/core/main/ts/html/Base64Uris.ts
+++ b/modules/tinymce/src/core/main/ts/html/Base64Uris.ts
@@ -22,7 +22,7 @@ export interface Base64UriParts {
 }
 
 export const extractBase64DataUris = (html: string): Base64Extract => {
-  const dataImageUri = /data:[^;<" ]+;base64,([a-z0-9\+\/=\s]+)/gi;
+  const dataImageUri = /data:[^;<"'\s]+;base64,([a-z0-9\+\/=\s]+)/gi;
   const chunks: string[] = [];
   const uris: UriMap = {};
   const prefix = Id.generate('img');
--- a/modules/tinymce/src/core/main/ts/html/Base64Uris.ts
+++ b/modules/tinymce/src/core/main/ts/html/Base64Uris.ts
@@ -22,7 +22,7 @@ export interface Base64UriParts {
 }
 
 export const extractBase64DataUris = (html: string): Base64Extract => {
-  const dataImageUri = /data:[^;]+;base64,([a-z0-9\+\/=\s]+)/gi;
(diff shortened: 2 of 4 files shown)


modules/tinymce/src/core/main/ts/html/Base64Uris.ts
```diff
@@ -22,7 +22,7 @@ export interface Base64UriParts {
 }
 
 export const extractBase64DataUris = (html: string): Base64Extract => {
-  const dataImageUri = /data:[^;]+;base64,([a-z0-9\+\/=\s]+)/gi;
+  const dataImageUri = /data:[^;<]+;base64,([a-z0-9\+\/=\s]+)/gi;
```
lnewson, line 25: This still is far too open... take for example this HTML:
```
<img data-custom="data:" src="data:image/png;base64,R0/yw==">
```
The above regex will incorrectly walk across the attribute boundaries, e.g: `data:" src="data:image/png;base64,R0/yw==`. So at the very least we need to make sure that spaces aren't allowed in the mime-type part of this regex either. Ideally, we should do some quick investigation if we can get something that will only match a valid mime-type (though this could be difficult, so doing the space might be all we can do).

lnewson (changes requested):

modules/tinymce/CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+- Base64 encoded image was not displayed following `data:` text #TINY-8646
```
lnewson, line 10: This also wasn't restricted to images, it was base64 URIs in general and it was related to parsing not really how it was "rendered". The rendering was a side-effect of it not parsing correctly.

modules/tinymce/CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
```
lnewson, line 9: Another thing you'll need to do here is bump the version in `package.json` to `5.10.5` since this is the first commit for that.

modules/tinymce/src/core/main/ts/html/Base64Uris.ts
tiny-ben-tran: this regex `/data:([a-z0-9]+\/[a-z0-9-.+]+);base64,([a-z0-9\+\/=\s]+)/gi` works if we only concern about matching mime-type formats. I prefer this since it is more precise

lnewson: Except that doesn't match everything that's allowed in a mime type... take a look at the RFC: https://datatracker.ietf.org/doc/html/rfc6838#section-4.2. You can see it allows other things such as `!`, `&`, etc...  which wouldn't be matched by the above.

tiny-ben-tran: It would open a can of worm if we are going to support what the spec says. I am just going to update the current regex to handle more cases for now

modules/tinymce/CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## 5.10.5 - 2022-05-05
+
+### Fixed
+- Base64 data URI was not extracted correctly following `data:` text #TINY-8646
```
lnewson, line 12:
```suggestion
- Base64 data URIs were not extracted correctly during parsing when proceeded by `data:` text #TINY-8646
```

modules/tinymce/CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## 5.10.5 - 2022-05-05
+
```
lnewson, line 10: This shouldn't be here yet.
```suggestion
```

modules/tinymce/src/core/test/ts/atomic/html/Base64UrisTest.ts
```diff
@@ -98,6 +98,72 @@ describe('atomic.tinymce.core.html.Base64UrisTest', () => {
...
+        html: '<img name="data:"src="$prefix_0"/>',
+        uris: {
+          $prefix_0: 'data:image/gif;base64,R0/yw=='
+        }
+      }
+    );
+
+    testExtract(
+      'TINY-8646: Should extract based64 encoded image when single quotes are used',
+      '<img name=\'data:\'src=\'data:image/gif;base64,R0/yw==\'/>',
```
lnewson, line 148: It might be more readable if you use template literals here instead?
```suggestion
      `<img name='data:'src='data:image/gif;base64,R0/yw=='/>`,
```

modules/tinymce/src/core/test/ts/atomic/html/Base64UrisTest.ts
```diff
@@ -98,6 +98,72 @@ describe('atomic.tinymce.core.html.Base64UrisTest', () => {
...
+      {
+        html: '<img name=\'data:\'src=\'$prefix_0\'/>',
```
lnewson, line 150:
```suggestion
        html: `<img name='data:'src='$prefix_0'/>`,
```

lnewson (approved):

ltrouton (approved):

PR #7807, opened 2022-05-05
TINY-8661: Stop links from opening when preventDefault is called

hamza0867:
Related Ticket: TINY-8661

Description of Changes:
Stop links from opening using alt+enter (option+enter on Mac) when preventDefault is called

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -12,7 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The default value for `end_container_on_empty_block` option has been changed to `'blockquote'` #TINY-6559
 
 ### Fixed
-- Links would open when using alt+enter (option+enter on Mac) even when preventDefault is called on the keydown event #TINY-8661
+- Links would open when using alt+enter (option+enter on Mac) even when `preventDefault()` is called on the keydown event #TINY-8661
 - Spaces would not be added correctly on some browsers when before or after a contenteditable block element #TINY-8588
 - Images were not showing as selected when selecting images alongside other content #TINY-5947
 - Dialogs will not exceed the window height on smaller screens #TINY-8146
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -12,7 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The default value for `end_container_on_empty_block` option has been changed to `'blockquote'` #TINY-6559
 
 ### Fixed
-- Links would open when using alt+enter (option+enter on Mac) even when preventDefault is called on the event #TINY-8661
+- Links would open when using alt+enter (option+enter on Mac) even when preventDefault is called on the keydown event #TINY-8661
 - Spaces would not be added correctly on some browsers when before or after a contenteditable block element #TINY-8588
 - Images were not showing as selected when selecting images alongside other content #TINY-5947
 - Dialogs will not exceed the window height on smaller screens #TINY-8146
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The default value for `end_container_on_empty_block` option has been changed to `'blockquote'` #TINY-6559
 
 ### Fixed
+- Links would open when using alt+enter (option+enter on Mac) even when preventDefault is called on the event #TINY-8661
 - Spaces would not be added correctly on some browsers when before or after a contenteditable block element #TINY-8588
 - Images were not showing as selected when selecting images alongside other content #TINY-5947
 - Dialogs will not exceed the window height on smaller screens #TINY-8146
--- a/modules/tinymce/src/plugins/link/main/ts/core/Actions.ts
+++ b/modules/tinymce/src/plugins/link/main/ts/core/Actions.ts
@@ -49,10 +49,12 @@ const setupGotoLinks = (editor: Editor): void => {
   });
 
   editor.on('keydown', (e) => {
-    const link = getSelectedLink(editor);
-    if (!e.isDefaultPrevented() && link && e.keyCode === 13 && hasOnlyAltModifier(e)) {
-      e.preventDefault();
-      gotoLink(editor, link);
+    if (!e.isDefaultPrevented() && e.keyCode === 13 && hasOnlyAltModifier(e)) {
+      const link = getSelectedLink(editor);
+      if (link) {
+        e.preventDefault();
+        gotoLink(editor, link);
+      }
     }
   });
 };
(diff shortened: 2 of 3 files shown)


modules/tinymce/CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The default value for `end_container_on_empty_block` option has been changed to `'blockquote'` #TINY-6559
 
 ### Fixed
+- Links would open when using alt+enter (option+enter on Mac) even when preventDefault is called on the event #TINY-8661
```
spocke, line 15: Maybe:
```suggestion
- Links would open when using alt+enter (option+enter on Mac) even when preventDefault is called on the keydown event #TINY-8661
```

spocke (approved):

tiny-ben-tran (commented):
@hamza0867 , can you add tests for this fix?

modules/tinymce/src/plugins/link/main/ts/core/Actions.ts
```diff
@@ -50,7 +50,7 @@ const setupGotoLinks = (editor: Editor): void => {
 
   editor.on('keydown', (e) => {
     const link = getSelectedLink(editor);
-    if (link && e.keyCode === 13 && hasOnlyAltModifier(e)) {
+    if (!e.defaultPrevented && link && e.keyCode === 13 && hasOnlyAltModifier(e)) {
```
lnewson, line 53: This should be using the `e.isDefaultPrevented()` function instead, as that's the TinyMCE provided API that is designed to handle cross browser cases and custom events.

Additionally, this isn't the best place to put this. Currently, you're running `getSelectedLink()` when it's never going to be used, so this check should instead be around the entire block instead of just this if statement.

modules/tinymce/CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The default value for `end_container_on_empty_block` option has been changed to `'blockquote'` #TINY-6559
 
 ### Fixed
+- Links would open when using alt+enter (option+enter on Mac) even when preventDefault is called on the keydown event #TINY-8661
```
```suggestion
- Links would open when using alt+enter (option+enter on Mac) even when `preventDefault()` is called on the keydown event #TINY-8661
```

lnewson (changes requested):

lnewson:
> @hamza0867 , can you add tests for this fix?

I'm guessing this can't be done, since we can't assert that a new tab wasn't opened. That said, if that is the case it would be good to have confirmation and should really be noted in the description as to why there are no tests.

lnewson:
> > @hamza0867 , can you add tests for this fix?
> 
> I'm guessing this can't be done, since we can't assert that a new tab wasn't opened. That said, if that is the case it would be good to have confirmation and should really be noted in the description as to why there are no tests.

Actually, as the `gotoLink` function works for in page links, then this should be able to be tested by using the key combination and asserting the page hasn't scrolled. So yeah we should try and write some tests for this.

modules/tinymce/src/plugins/link/main/ts/core/Actions.ts
spocke: I noticed this as well that the more modern property was used however since we are dropping IE support we could start using this property. This wasn't a standard property at the time so the naming here of the function of the workaround is more based on how it works in other popular frameworks at the time. But I guess for custom events it would be wrong since it will not set the `defaultPrevented` flag if you can `preventDefault` on those but maybe it should then we could slowly migrate toward the more standard property. So for now change it but should we consider a modernization here?

spocke: Is afk so couldn't test this it seems that the normalization code for events already handle the property state change. https://github.com/tinymce/tinymce/blob/develop/modules/tinymce/src/core/main/ts/events/EventUtils.ts#L86

hamza0867: > Additionally, this isn't the best place to put this. Currently, you're running `getSelectedLink()` when it's never going to be used, so this check should instead be around the entire block instead of just this if statement.

@lnewson   It seems the result of `getSelectedLink()` is used in the if condition. Should I remove it from the if condition and put it inside the body of if statement ?

hamza0867:
> > > @hamza0867 , can you add tests for this fix?
> > 
> > 
> > I'm guessing this can't be done, since we can't assert that a new tab wasn't opened. That said, if that is the case it would be good to have confirmation and should really be noted in the description as to why there are no tests.
> 
> Actually, as the `gotoLink` function works for in page links, then this should be able to be tested by using the key combination and asserting the page hasn't scrolled. So yeah we should try and write some tests for this.

From my testing, pressing alt+enter opens a new window (not tab) (it ignores if target=_blank or not) so the page won't scroll ?

> > > @hamza0867 , can you add tests for this fix?
> > 
> > 
> > I'm guessing this can't be done, since we can't assert that a new tab wasn't opened. That said, if that is the case it would be good to have confirmation and should really be noted in the description as to why there are no tests.
> 
> Actually, as the `gotoLink` function works for in page links, then this should be able to be tested by using the key combination and asserting the page hasn't scrolled. So yeah we should try and write some tests for this.

You are correct, a test case has been added

modules/tinymce/src/plugins/link/main/ts/core/Actions.ts
lnewson: I'm not sure how that relates to what I mentioned sorry? To be clear, this is what I expected to see here:
```
  editor.on('keydown', (e) => {
    if (!e.isDefaultPrevented()) {
      const link = getSelectedLink(editor);
      if (link && e.keyCode === 13 && hasOnlyAltModifier(e)) {
        e.preventDefault();
        gotoLink(editor, link);
      }
    }
  });
```
In saying that though, now that we have a separate if condition we can optimize the above a little more, such as:
```
  editor.on('keydown', (e) => {
    if (!e.isDefaultPrevented() && e.keyCode === 13 && hasOnlyAltModifier(e)) {
      const link = getSelectedLink(editor);
      if (link) {
        e.preventDefault();
        gotoLink(editor, link);
      }
    }
  });
```
This way we're only doing the link lookup if alt+enter is pressed and the event hasn't been prevented. Ideally we'd also use a non falsy check for `link` as well.

modules/tinymce/src/plugins/link/test/ts/browser/PreventDefaultTest.ts
```diff
@@ -0,0 +1,51 @@
...
+import { TestLinkUi } from '../module/TestLinkUi';
+
+describe('browser.tinymce.plugins.link.PreventDefaultTest', () => {
+
+  const hook = TinyHooks.bddSetupLight<Editor>({
+    plugins: 'link',
+    toolbar: 'link openlink unlink',
+    base_url: '/project/tinymce/js/tinymce',
+    setup: (ed) => {
+      const hasOnlyAltModifier = (e) => {
```
lnewson, line 18: Types for `e` here?
```suggestion
      const hasOnlyAltModifier = (e: KeyboardEvent) => {
```

lnewson (approved):

modules/tinymce/src/plugins/link/main/ts/core/Actions.ts
hamza0867: Hey @lnewson should this conversation be resolved ?

lnewson: Yep, it's all good now. The point Spocke mentioned I've been talking to him via DM's about that and we're going to raise it via #tinymce-dev.

PR #7816, opened 2022-05-11
TINY-8579: Changed how blank lines are generated in the text output

HAFRMO:
Related Ticket:

Description of Changes:
Changed how newlines are generated to normalize the behavior, and to correct the output of newlines.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


tiny-ben-tran (approved):

modules/tinymce/src/core/main/ts/content/GetContentImpl.ts
```diff
@@ -22,9 +22,15 @@ const getContentFromBody = (editor: Editor, args: GetContentArgs, body: HTMLElem
...
-    content = content === '\n' ? '' : content;
+    content = getContentFromBody(editor, { ...args, format: 'html' }, body);
+
+    if (content === '' || !Type.isString(content)) {
+      return content;
+    }
+
+    const parseBody = document.createElement('body');
+    parseBody.innerHTML = trimEmptyContents(editor, content);
+    content = Zwsp.trim(parseBody.innerText.replace(/\xa0/g, ''));
```
lnewson, line 33: This isn't going to work because it will remove all `&nbsp;` characters from the output. For example try the following HTML:
```
<p>this is test&nbsp; content</p>
```

Instead of having 2 spaces between `test` and `content` it will only have 1 space. Also as a side note, we normally use `\u00a0` instead of `\xa0` for nbsp characters.

lnewson (changes requested):
As discussed previously this approach won't work for the reason I mentioned in the review. We need to look into other options here, but at this stage I suspect we're going to need the custom text serialization that we've mentioned previously to solve this issue (see TINY-8690 and https://github.com/tinymce/tinymce/pull/7744).

modules/tinymce/src/core/main/ts/html/ParserUtils.ts
```diff
@@ -6,7 +6,9 @@ import Schema, { SchemaMap } from '../api/html/Schema';
 
 const paddEmptyNode = (settings: DomParserSettings, args: ParserArgs, blockElements: SchemaMap, node: AstNode): void => {
   if (args.insert && blockElements[node.name]) {
-    node.empty().append(new AstNode('br', 1));
+    const astNode = new AstNode('br', 1);
+    astNode.attr('data-mce-bogus', '1');
```
spocke, line 10: Why was this needed, this could potentially break things?

modules/tinymce/src/core/main/ts/content/GetContentImpl.ts
```diff
@@ -22,9 +22,15 @@ const getContentFromBody = (editor: Editor, args: GetContentArgs, body: HTMLElem
...
-    // return empty string for text format when editor is empty to avoid bogus elements being returned in content
-    content = Zwsp.trim(body.innerText);
-    content = content === '\n' ? '' : content;
+    content = getContentFromBody(editor, { ...args, format: 'html' }, body);
+
+    if (content === '' || !Type.isString(content)) {
+      return content;
+    }
+
+    const parseBody = document.createElement('body');
```
spocke, line 31: Interesting so sticking it into a separate body element it will render the innerText correctly on Safari? Is it that we are getting innerText on a contenteditable and it has that special rendering mode where br:s are treated differently that causes the bug?

modules/tinymce/src/core/main/ts/html/ParserUtils.ts
HAFRMO: Because the initial data did not have the bogus-attribute added as they apparently should have. Tests seems to be passing as expected.

spocke: Right so it's only when it's being inserted and that empty block will still be treated as empty and get padded on export.

modules/tinymce/src/core/main/ts/content/GetContentImpl.ts
HAFRMO: The issue there was that safari treated "Content in body" and "Content in a div" differently, which could cause issues with inline/not inline editors.

modules/tinymce/src/core/test/ts/browser/content/EditorContentTest.ts
```diff
@@ -92,17 +89,19 @@ describe('browser.tinymce.core.content.EditorContentTest', () => {
...
+  it('TINY-8578: getContent text, two empty lines in div', () => testGetTextContent('<div><p></p><p></p></div>', '\n\n\n\n\n\n\n\n\n\n'));
 
-  it('TINY-8578: getContent text, two empty lines in div', () => testGetTextContent('<div><p></p><p></p></div>', isSafari() ? '\n\n' : '\n\n\n\n'));
+  it('TINY-8578: getContent text, repeating two empty lines in divs', () => testGetTextContent('<div><p></p><p></p></div><div><p></p><p></p></div>', '\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n\n'));
 
-  it('TINY-8578: getContent text, two empty lines', () => testGetTextContent('<p></p><p></p>', isSafari() ? '\n\n' : '\n\n\n\n'));
+  it('TINY-8578: getContent text, two empty lines', () => testGetTextContent('<p></p><p></p>', '\n\n\n\n\n\n\n\n'));
 
-  it('TINY-6281: getContent text with empty editor', () => testGetTextContent('', ''));
+  it('TINY-6281: getContent text with empty editor', () => testGetTextContent('', '\n\n\n\n\n'));
```
HAFRMO, line 104: The new method have added a lot of newlines everywhere. While this is now the same between the browsers, do we want to continue in this direction or try a different one?

lnewson: Yeah I agree with Spocke here... the TBIO solution is not going to work for TinyMCE as it doesn't cover a lot of cases so it'll be a regression in functionality. It also doesn't have context of the schema nor things we generally need to ignore like bogus elements. If we need a custom walker solution then we are going to have to expand on that quite a bit, implement our own (as we maybe can't change the TBIO one depending on where and how it's being used) or look at a third-party library for this. Both are going to be a very large task.

modules/tinymce/src/core/main/ts/content/GetContentImpl.ts
lnewson: @HAFRMO have you considered or tried replacing the bogus elements with a nonbreaking zero-width space to solve this instead? As the problem if I understand correctly was that with `<br>` that added a newline and the problem with `&nbsp;` was that it added additional spaces. However if we use `\uFEFF` that's already an internal character and we can then strip it safely from the text after getting the `innerText`.

I tested this theory via the HTML in https://fiddle.tiny.cloud/5eiaab and it appeared to work fine on Chrome.

lnewson (changes requested):
Just marking this again because of the previous comments. I suspect my idea about replacing the bogus br elements with a nonbreaking zero width space should help solve this though without having to write a custom text serializer. You've looked into this more though, so let me know if I've missed something there.

modules/tinymce/src/core/test/ts/browser/content/EditorContentTest.ts
TheSpyder: It was never going to be a silver bullet. I'd be surprised if the TBIO logic is used anywhere that would have a problem with us adjusting the newline behaviour, if it's even used at all.

I do think a custom serializer is the correct solution because at least then it's the same for all browsers. But I guess if we can find a workaround that will solve the customer issue we can schedule something larger down the track.

spocke: I wonder if we could do something like this at least until we can make a proper get text mode:
```
body = editor.getBody();
clonedBody = body.cloneNode(false);
clonedBody.innerHTML = editor.getContent();
body.parentNode.insertBefore(clonedBody, body);
text = clonedBody.innerText;
clonedBody.remove();
console.log(text);
```
Would not be efficient but it would remove any bogus nodes and also render that in context since I think a detached body might do very strange things when it doesn't know how to render things. Does it use the default user agent css. Seems safer to stick next to the body element of the editor.

TheSpyder (dismissed):

modules/tinymce/src/core/main/ts/content/GetContentImpl.ts
```diff
@@ -22,9 +22,14 @@ const getContentFromBody = (editor: Editor, args: GetContentArgs, body: HTMLElem
   if (args.format === 'raw') {
     content = Tools.trim(TrimHtml.trimExternal(editor.serializer, body.innerHTML));
   } else if (args.format === 'text') {
-    // return empty string for text format when editor is empty to avoid bogus elements being returned in content
-    content = Zwsp.trim(body.innerText);
-    content = content === '\n' ? '' : content;
+    content = body.innerHTML.replace(/<br data-mce-bogus="1">/g, '\uFEFF');
+    const parseBody = document.createElement('body');
+    parseBody.setAttribute('style', 'position:fixed;left:400%;top:0px');
```
lnewson, line 27: As discussed on slack, this won't work with stricter CSP's and may still show in some edge cases... please use the normal method of a negative position instead.

modules/tinymce/src/core/main/ts/content/GetContentImpl.ts
```diff
@@ -22,9 +22,14 @@ const getContentFromBody = (editor: Editor, args: GetContentArgs, body: HTMLElem
   if (args.format === 'raw') {
     content = Tools.trim(TrimHtml.trimExternal(editor.serializer, body.innerHTML));
   } else if (args.format === 'text') {
-    // return empty string for text format when editor is empty to avoid bogus elements being returned in content
-    content = Zwsp.trim(body.innerText);
-    content = content === '\n' ? '' : content;
+    content = body.innerHTML.replace(/<br data-mce-bogus="1">/g, '\uFEFF');
+    const parseBody = document.createElement('body');
```
lnewson, line 26: We should not be creating elements from, nor appending to the root document. We should be using the editor document instead so please swap the uses here to use `editor.getDoc()` instead (ideally that should be stored as a variable). This is also needed to ensure any content css/styling is applied to the elements when we're getting the text content.

modules/tinymce/src/core/main/ts/content/GetContentImpl.ts
```diff
@@ -22,9 +22,14 @@ const getContentFromBody = (editor: Editor, args: GetContentArgs, body: HTMLElem
...
+    parseBody.setAttribute('style', 'position:fixed;left:400%;top:0px');
+    const root = document.getElementsByTagName('html').item(0);
```
lnewson, line 28: You don't need to do this... you can just use `document.documentElement` instead.

lnewson (changes requested):

lnewson:
Also don't forget this will need rebasing onto the 6.2 branch and the changelog will need moving.

modules/tinymce/src/core/main/ts/content/GetContentImpl.ts
```diff
@@ -22,9 +22,17 @@ const getContentFromBody = (editor: Editor, args: GetContentArgs, body: HTMLElem
   if (args.format === 'raw') {
     content = Tools.trim(TrimHtml.trimExternal(editor.serializer, body.innerHTML));
   } else if (args.format === 'text') {
-    // return empty string for text format when editor is empty to avoid bogus elements being returned in content
-    content = Zwsp.trim(body.innerText);
-    content = content === '\n' ? '' : content;
+    const doc = editor.getDoc();
+    content = body.innerHTML.replace(/<br data-mce-bogus="1">/g, Unicode.zeroWidth);
+    const parseBody = doc.createElement('body');
```
lnewson, line 27: I'm still pretty concerned about us having 2 body elements, as it's not valid HTML. So while it might work now, it might break easily in the future. I really think we need to try and find a different solution here.

HAFRMO: I agree, but if we use a div we will get different output on different browsers. If you have a workaround or better solution I'd be delighted to hear it.

modules/tinymce/src/core/main/ts/content/GetContentImpl.ts
```diff
@@ -22,9 +22,17 @@ const getContentFromBody = (editor: Editor, args: GetContentArgs, body: HTMLElem
   if (args.format === 'raw') {
     content = Tools.trim(TrimHtml.trimExternal(editor.serializer, body.innerHTML));
   } else if (args.format === 'text') {
-    // return empty string for text format when editor is empty to avoid bogus elements being returned in content
-    content = Zwsp.trim(body.innerText);
-    content = content === '\n' ? '' : content;
+    const doc = editor.getDoc();
+    content = body.innerHTML.replace(/<br data-mce-bogus="1">/g, Unicode.zeroWidth);
```
lnewson, line 26: This replacement string won't work if the content is XHTML, for example `<br data-mce-bogus="1"/>`.

lnewson: It also wouldn't catch cases where we have 2 attributes on a bogus element (granted I can't think of any off the top of my head). Either way, we probably want to consider doing something more like this: https://github.com/tinymce/tinymce/blob/236891ffcdf2fd1b698d020e67c31838fb110019/modules/tinymce/src/core/main/ts/dom/TrimHtml.ts#L90.

modules/tinymce/src/core/main/ts/content/GetContentImpl.ts
```diff
@@ -22,9 +22,17 @@ const getContentFromBody = (editor: Editor, args: GetContentArgs, body: HTMLElem
   if (args.format === 'raw') {
     content = Tools.trim(TrimHtml.trimExternal(editor.serializer, body.innerHTML));
   } else if (args.format === 'text') {
-    // return empty string for text format when editor is empty to avoid bogus elements being returned in content
-    content = Zwsp.trim(body.innerText);
-    content = content === '\n' ? '' : content;
+    const doc = editor.getDoc();
```
lnewson, line 25: I wonder if we should move this out to a separate function to make it a little more readable?

modules/tinymce/src/core/main/ts/content/GetContentImpl.ts
```diff
@@ -16,15 +16,36 @@ const trimEmptyContents = (editor: Editor, html: string): string => {
   return html.replace(emptyRegExp, '');
 };
 
+const bogusAllRegExp = /<(\w+) [^>]*data-mce-bogus="(all|1)"[^>]*>/g;
```
lnewson, line 19: @HAFRMO this doesn't look correct... if we have a `all` element then the entire element needs to be stripped not converted into a zero-width space.

I wonder if instead of doing this string based replacement, if instead we should do something like this:
```ts
const parseBody = doc.createElement('body');
parseBody.style.position = 'fixed';
parseBody.style.left = '-9999999px';
parseBody.style.top = '0px';
parseBody.innerHTML = body.innerHTML;

const bogusElements = SelectorFind.decendants(SugarElement.fromDom(parseBody), '[data-mce-bogus]');
Arr.each(bogusElements, (elem) => {
  if (Attribute.get(elem, 'data-mce-bogus') !== 'all) {
    Insert.before(elem, SugarElement.fromText(Unicode.zeroWidth));
  }
  Remove.remove(elem);
});
```

lnewson: Actually the more I think about this, the more I realise that this needs to handle bogus elements in general when getting the text content.

lnewson: Something like this I think should work?
```ts
// Cleanup bogus elements
const bogusElements = SelectorFilter.descendants(SugarElement.fromDom(parseBody), '[data-mce-bogus]');
Arr.each(bogusElements, (elem) => {
  const bogusValue = Attribute.get(elem, 'data-mce-bogus');
  if (bogusValue === 'all') {
    Remove.remove(elem);
  } else if (isBr(elem)) {
    // Need to keep bogus represented as a zero-width space so that they aren't collapsed by the browser
    Insert.before(elem, SugarElement.fromText(Unicode.zeroWidth));
    Remove.remove(elem);
  } else {
    Remove.unwrap(elem);
  }
});
```

modules/tinymce/src/core/main/ts/content/GetContentImpl.ts
lnewson: I've spiked out a solution which I believe works, that doesn't need to append 2 body elements (or at least all the tests pass): https://github.com/tinymce/tinymce/compare/feature/TINY-8579...spike/TINY-8579. Take a look at that and see if you can see any issues.

modules/tinymce/src/core/test/ts/browser/content/EditorContentTest.ts
```diff
@@ -92,15 +89,17 @@ describe('browser.tinymce.core.content.EditorContentTest', () => {
   // TODO: TINY-8367 The table plugin code has been moved to core so there is now always an extra DIV in the dom for the table resize bars
```
lnewson, line 89: Is this comment still accurate? As we've changed how things are retrieved I don't believe it's valid anymore. TINY-8367 could also be closed/linked to this fix if it's not longer valid as well.

modules/tinymce/src/core/main/ts/content/GetContentImpl.ts
```diff
@@ -16,15 +16,36 @@ const trimEmptyContents = (editor: Editor, html: string): string => {
...
+  const parseBody = doc.createElement('body');
+  parseBody.style.position = 'fixed';
+  parseBody.style.left = '-9999999px';
+  parseBody.style.top = '0px';
+
+  const root = doc.documentElement;
+  root.appendChild(parseBody);
+  parseBody.innerHTML = content;
+
+  content = Zwsp.trim(Unicode.removeZwsp(parseBody.innerText));
```
lnewson, line 36: This is stripping the Zwsp characters twice, as `Zwsp.trim` and `Unicode.removeZwsp` are the same thing.

modules/tinymce/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - The Autolink plugin did not work when the text nodes in the content were fragmented #TINY-3723
 - Fixed various incorrect types on public APIs found while enabling TypeScript strict mode #TINY-8806
+- Text returned from `editor.getContent({format: 'text'})` would differ between some browsers #TINY-8579
```
lnewson, line 12:
```suggestion
- Text returned from `editor.getContent({format: 'text'})` would differ with blank lines between some browsers #TINY-8579
```
Maybe to just make it a little clearer as to how it would differ?

modules/tinymce/src/core/test/ts/browser/content/EditorContentTest.ts
```diff
@@ -92,15 +89,19 @@ describe('browser.tinymce.core.content.EditorContentTest', () => {
...
 
-  it('TINY-8578: getContent text, two empty lines in div', () => testGetTextContent('<div><p></p><p></p></div>', isSafari() ? '\n\n' : '\n\n\n\n'));
+  it('TINY-8578: getContent text, two empty lines in div', () => testGetTextContent('<div><p></p><p></p></div>', '\n\n'));
+
+  it('TINY-8578: getContent text, repeating two empty lines in divs', () => testGetTextContent('<div><p></p><p></p></div><div><p></p><p></p></div>', '\n\n\n\n\n\n'));
+
+  it('TINY-8578: getContent text, two empty lines', () => testGetTextContent('<p></p><p></p>', '\n\n'));
 
-  it('TINY-8578: getContent text, two empty lines', () => testGetTextContent('<p></p><p></p>', isSafari() ? '\n\n' : '\n\n\n\n'));
+  it('TINY-8578: getContent text with bogus element', () => testGetTextContent(`<p data-mce-bogus="all">content</p><p></p>`, '', 'raw'));
```
lnewson, line 104: I wonder if these tests should be in the `EditorGetContentTextFormatTest` test file instead?

Also do we have tests that check the iframe output as well as inline? It looks like this only tests inline.

- [ ] Ensure tests exist for both iframe and inline modes

lnewson (approved):

modules/tinymce/src/core/test/ts/browser/content/EditorContentTest.ts
HAFRMO: No, I think this is now properly irrellevant.

ltrouton (approved):

vpyshnenko (approved):

PR #7819, opened 2022-05-12
TINY-8701: Ensure fullscreen fires a ResizeEditor event

HAFRMO:
Related Ticket:

Description of Changes:
This issue was detected as one where notifications are not appropriately moved when toggling fullscreen. I thought it might be better for the fullscreen plugin to fire the `resizeeditor`-event instead of making the notifications listen to the `fullscreenstatechanged` as it seemed more generic, since fullscreen does change the size of the editor.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Changed
+- Toggling fullscreen mode with `fullscreen` plugin now also fires the `ResizeEditor` event #TINY-8701
 - Getting the text content of the editor now returns newlines instead of an empty string if more than one empty paragraph exists #TINY-8578
 - The `end_container_on_empty_block` option can now take a string of blocks to split when pressing Enter twice #TINY-6559
 - The default value for `end_container_on_empty_block` option has been changed to `'blockquote'` #TINY-6559
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Spaces would not be added correctly on some browsers when before or after a contenteditable block element #TINY-8588
 - Images were not showing as selected when selecting images alongside other content #TINY-5947
-- Notifications would not properly reposition when toggling to or from fullscreen #TINY-8701
+- Notifications would not properly reposition when toggling fullscreen mode #TINY-8701
 - Dialogs will not exceed the window height on smaller screens #TINY-8146
 - Some types on functions in the `tinymce.dom.TreeWalker` class missed that it could return undefined #TINY-8592
 - In some cases pressing the backspace key would incorrectly step into tables rather than remain outside #TINY-8592
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Changed
-- Toggling fullscreen mode with `fullscreen` plugin now also fires the `ResizeEditor` event #TINY-8701
+- Toggling fullscreen mode with the `fullscreen` plugin now also fires the `ResizeEditor` event #TINY-8701
 - Getting the text content of the editor now returns newlines instead of an empty string if more than one empty paragraph exists #TINY-8578
 - The `end_container_on_empty_block` option can now take a string of blocks to split when pressing Enter twice #TINY-6559
 - The default value for `end_container_on_empty_block` option has been changed to `'blockquote'` #TINY-6559
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Spaces would not be added correctly on some browsers when before or after a contenteditable block element #TINY-8588
 - Images were not showing as selected when selecting images alongside other content #TINY-5947
+- Notifications would not properly reposition when toggling to or from fullscreen #TINY-8701
 - Dialogs will not exceed the window height on smaller screens #TINY-8146
 - Some types on functions in the `tinymce.dom.TreeWalker` class missed that it could return undefined #TINY-8592
 - In some cases pressing the backspace key would incorrectly step into tables rather than remain outside #TINY-8592
--- a/modules/tinymce/src/plugins/fullscreen/main/ts/api/Events.ts
+++ b/modules/tinymce/src/plugins/fullscreen/main/ts/api/Events.ts
@@ -2,6 +2,7 @@ import Editor from 'tinymce/core/api/Editor';
 
 const fireFullscreenStateChanged = (editor: Editor, state: boolean): void => {
   editor.dispatch('FullscreenStateChanged', { state });
+  editor.dispatch('ResizeEditor', { state });
 };
 
 export {
--- a/modules/tinymce/src/plugins/fullscreen/main/ts/api/Events.ts
+++ b/modules/tinymce/src/plugins/fullscreen/main/ts/api/Events.ts
@@ -2,7 +2,7 @@ import Editor from 'tinymce/core/api/Editor';
 
 const fireFullscreenStateChanged = (editor: Editor, state: boolean): void => {
   editor.dispatch('FullscreenStateChanged', { state });
-  editor.dispatch('ResizeEditor', { state });
+  editor.dispatch('ResizeEditor');
 };
 
 export {
--- a/modules/tinymce/src/plugins/fullscreen/test/ts/browser/FullScreenPluginTest.ts
+++ b/modules/tinymce/src/plugins/fullscreen/test/ts/browser/FullScreenPluginTest.ts
@@ -1,12 +1,13 @@
 import { UiFinder } from '@ephox/agar';
 import { afterEach, context, describe, it } from '@ephox/bedrock-client';
-import { Arr } from '@ephox/katamari';
+import { Arr, Type } from '@ephox/katamari';
 import { PlatformDetection } from '@ephox/sand';
-import { Attribute, Classes, Css, Html, SelectorFind, SugarBody, SugarDocument, SugarShadowDom, Traverse } from '@ephox/sugar';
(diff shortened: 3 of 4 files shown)


modules/tinymce/src/plugins/fullscreen/main/ts/api/Events.ts
```diff
@@ -2,6 +2,7 @@ import Editor from 'tinymce/core/api/Editor';
 
 const fireFullscreenStateChanged = (editor: Editor, state: boolean): void => {
   editor.dispatch('FullscreenStateChanged', { state });
+  editor.dispatch('ResizeEditor', { state });
```
lnewson, line 5: This isn't correct, as `ResizeEditor` shouldn't have a state.

modules/tinymce/src/plugins/fullscreen/test/ts/browser/FullScreenPluginTest.ts
```diff
@@ -89,13 +90,16 @@ describe('browser.tinymce.plugins.fullscreen.FullScreenPluginTest', () => {
...
-            lastEventArgs.set(e);
+          firedEvents = [];
+          editor.on('FullscreenStateChanged ResizeEditor', (e: any) => {
+            firedEvents.push(e.type);
+            firedEvents.push(e.state);
           });
         }
       }, [ FullscreenPlugin, LinkPlugin ]);
 
+      afterEach(() => firedEvents = []);
```
lnewson, line 101:
```suggestion
      afterEach(() => {
        firedEvents = [];
      });
```
This should never be inlined for before/after hooks as the return value is passed to the hook.

modules/tinymce/src/plugins/fullscreen/test/ts/browser/FullScreenPluginTest.ts
```diff
@@ -37,7 +37,8 @@ describe('browser.tinymce.plugins.fullscreen.FullScreenPluginTest', () => {
 
   const assertApiAndLastEvent = (editor: Editor, state: boolean) => {
     assert.equal(editor.plugins.fullscreen.isFullscreen(), state, 'Editor isFullscreen state');
-    assert.equal(lastEventArgs.get().state, state, 'FullscreenStateChanged event state');
+    assert.deepEqual(firedEvents, [ 'fullscreenstatechanged', state, 'resizeeditor', state ], 'FullscreenStateChanged event state');
```
lnewson, line 40: This label doesn't make sense anymore... maybe something like this?
```suggestion
    assert.deepEqual(firedEvents, [ 'fullscreenstatechanged', state, 'resizeeditor', state ], 'fired event state');
```

modules/tinymce/src/plugins/fullscreen/test/ts/browser/FullScreenPluginTest.ts
```diff
@@ -37,7 +37,8 @@ describe('browser.tinymce.plugins.fullscreen.FullScreenPluginTest', () => {
...
+    firedEvents = [];
```
lnewson, line 41: This is pretty confusing to have here... as this is only meant to be an assertion but it's changing test state. Please change this so that this function doesn't alter state.

modules/tinymce/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Changed
+- Toggling fullscreen mode with `fullscreen` plugin now also fires the `ResizeEditor` event #TINY-8701
```
```suggestion
- Toggling fullscreen mode with the `fullscreen` plugin now also fires the `ResizeEditor` event #TINY-8701
```

modules/tinymce/src/plugins/fullscreen/main/ts/api/Events.ts
lnewson: Also if you're making this change then we should look at removing some of the duplicate event hooks we have (for example in `ControlSelection.ts`).

lnewson (changes requested):
Looks good overall, however you don't have a changelog entry nor a test for the actual issue in the Jira. Please add that as well .

modules/tinymce/src/plugins/fullscreen/main/ts/api/Events.ts
HAFRMO: There doesn't seem to be any other overlap, except for the one in the file you provided. However there's a few places where we could change events, though I'm not sure if that's what we want to do.

lnewson: Yeah we don't want to change the fullscreen event to resize editor, only remove it where it's duplicated.

modules/tinymce/src/plugins/fullscreen/test/ts/browser/FullScreenPluginTest.ts
lnewson: Hmm... that's been missed in review then I guess sorry. They shouldn't be like that.

modules/tinymce/src/plugins/fullscreen/test/ts/browser/FullScreenPluginTest.ts
```diff
@@ -76,6 +76,25 @@ describe('browser.tinymce.plugins.fullscreen.FullScreenPluginTest', () => {
...
+
+  const getNotificationPosition = (notification: NotificationApi) => {
+    const elem = Traverse.parent(SugarElement.fromDom(notification.getEl())).getOrDie() as SugarElement<HTMLElement>;
+    return {
+      top: elem.dom.offsetTop,
+      left: elem.dom.offsetLeft
+    };
+  };
+
+  const asserPositionChanged = (notification: NotificationApi, oldPos: { left: number; top: number }) => {
```
tiny-ben-tran, line 93:
```suggestion
  const assertPositionChanged = (notification: NotificationApi, oldPos: { left: number; top: number }) => {
```

modules/tinymce/CHANGELOG.md
```diff
@@ -7,13 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 ### Changed
+- Toggling fullscreen mode with the `fullscreen` plugin now also fires the `ResizeEditor` event #TINY-8701
 - Getting the text content of the editor now returns newlines instead of an empty string if more than one empty paragraph exists #TINY-8578
 - The `end_container_on_empty_block` option can now take a string of blocks to split when pressing Enter twice #TINY-6559
 - The default value for `end_container_on_empty_block` option has been changed to `'blockquote'` #TINY-6559
 
 ### Fixed
 - Spaces would not be added correctly on some browsers when before or after a contenteditable block element #TINY-8588
 - Images were not showing as selected when selecting images alongside other content #TINY-5947
+- Notifications would not properly reposition when toggling to or from fullscreen #TINY-8701
```
```suggestion
- Notifications would not properly reposition when toggling fullscreen mode #TINY-8701
```
Maybe? As toggling already means turning it on and off.

lnewson (approved):

modules/tinymce/src/plugins/fullscreen/test/ts/browser/FullScreenPluginTest.ts
```diff
@@ -1,19 +1,24 @@
...
 import { assert } from 'chai';
 
 import Editor from 'tinymce/core/api/Editor';
+import { NotificationApi } from 'tinymce/core/api/PublicApi';
 import FullscreenPlugin from 'tinymce/plugins/fullscreen/Plugin';
 import LinkPlugin from 'tinymce/plugins/link/Plugin';
 
 describe('browser.tinymce.plugins.fullscreen.FullScreenPluginTest', () => {
-  const lastEventArgs = Cell(null);
+  let firedEvents = [];
```
spocke, line 15: This becomes `any[]` so maybe have it be `boolean | string` that seem to be the type looking at the test.
```suggestion
  let firedEvents: boolean | string = [];
```

modules/tinymce/src/plugins/fullscreen/test/ts/browser/FullScreenPluginTest.ts
```diff
@@ -37,7 +42,7 @@ describe('browser.tinymce.plugins.fullscreen.FullScreenPluginTest', () => {
 
   const assertApiAndLastEvent = (editor: Editor, state: boolean) => {
```
spocke, line 43: This should be renames to maybe `assertApiEvents` since it's now checking more than just that last event state change.

modules/tinymce/src/plugins/fullscreen/test/ts/browser/FullScreenPluginTest.ts
```diff
@@ -37,7 +42,7 @@ describe('browser.tinymce.plugins.fullscreen.FullScreenPluginTest', () => {
...
     assert.equal(editor.plugins.fullscreen.isFullscreen(), state, 'Editor isFullscreen state');
-    assert.equal(lastEventArgs.get().state, state, 'FullscreenStateChanged event state');
+    assert.deepEqual(firedEvents, [ 'fullscreenstatechanged', state, 'resizeeditor' ], 'Fired event state');
```
spocke, line 45: The message here should be changed to reflect the code change maybe.
```suggestion
    assert.deepEqual(firedEvents, [ 'fullscreenstatechanged', state, 'resizeeditor' ], 'Should be expected events and state');
```

modules/tinymce/src/plugins/fullscreen/test/ts/browser/FullScreenPluginTest.ts
```diff
@@ -89,9 +113,12 @@ describe('browser.tinymce.plugins.fullscreen.FullScreenPluginTest', () => {
...
         setup: (editor: Editor) => {
-          lastEventArgs.set(null);
-          editor.on('FullscreenStateChanged', (e: Editor) => {
-            lastEventArgs.set(e);
+          firedEvents = [];
+          editor.on('FullscreenStateChanged ResizeEditor', (e: any) => {
+            firedEvents.push(e.type);
+            if (Type.isBoolean(e.state)) {
+              firedEvents.push(e.state);
+            }
```
spocke, line 121: I think this would be a lot better if we refactor it to all be strings. So make them be [`'fullscreenstatechanged:true', 'resizeeditor'`] or something that would also make the assert if failing be a lot more clear of what was wrong. I really dislike having multiple different types on things. Even if TS allows that other languages that are typed generally don't it also simplifies reasoning and avoids type guards etc on other parts of the code, in this example it not a big problem but it's the general idea of mixing types that should be avoided if possible.

```suggestion
          editor.on('FullscreenStateChanged ResizeEditor', (e: any) => {
            if (Type.isBoolean(e.state)) {
              firedEvents.push(`${e.type}:${e.state}`);
            } else {
              firedEvents.push(e.type);
            |
```

spocke (approved):

ltrouton (approved):

PR #7824, opened 2022-05-12
TINY-8069: Fixed incorrect URL detection for path segments

lnewson:
Related Ticket: TINY-8069

Description of Changes:
Updated the URL detection logic used by autolink and paste to ensure that path segments parse correctly when some uncommon characters are used, such as `!` and `:`. These are valid as per the URL parsing spec here: https://url.spec.whatwg.org/#url-path-segment-string

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/polaris/src/main/ts/ephox/polaris/api/Regexes.ts
+++ b/modules/polaris/src/main/ts/ephox/polaris/api/Regexes.ts
@@ -28,7 +28,7 @@
   - allow up to 15 character schemes including all valid characters from the spec https://url.spec.whatwg.org/#url-scheme-string (TINY-5074)
   - changed instances of 0-9 to be \d (TINY-5074)
   - reduced duplication (TINY-5074)
-  - include additional valid characters in the path segment (TINY-8069)
+  - allow [*!;:'@$] in the path segment as they are valid characters per the spec: https://url.spec.whatwg.org/#url-path-segment-string (TINY-8069)
 
 (?:
   (?:
@@ -43,7 +43,7 @@
 (?:
   \/
   (?:
-     [-.~*+=!;:'%@^$(),\/\w]*[-~*+=%@^$()\/\w]
+     [-.~*+=!;:'%@$(),\/\w]*[-~*+=%@$()\/\w]
    )?
 )?
 (?:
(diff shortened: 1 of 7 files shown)


tiny-ben-tran (approved):

ltrouton (approved):

modules/polaris/src/main/ts/ephox/polaris/api/Regexes.ts
```diff
@@ -28,6 +28,7 @@
   - allow up to 15 character schemes including all valid characters from the spec https://url.spec.whatwg.org/#url-scheme-string (TINY-5074)
   - changed instances of 0-9 to be \d (TINY-5074)
   - reduced duplication (TINY-5074)
+  - include additional valid characters in the path segment (TINY-8069)
```
spocke, line 31:
```suggestion
  - included additional valid characters `!` and `:` in the path segment (TINY-8069)
```

spocke (approved):

modules/polaris/src/main/ts/ephox/polaris/api/Regexes.ts
lnewson: We included more than just `!` though, so that wouldn't be accurate. You're right it should be `included` though.

lnewson: Looking at the diff, the full list is: `*`, `!`, `;`, `:`, `', `@`, `^` and `$`.

lnewson: Reading the spec again though `^` shouldn't be there though (I copied it largely from what was in the query string so looks like I missed removing that).

lnewson: Fixed that invalid character and reworded this message.

HAFRMO (approved):

modules/polaris/src/main/ts/ephox/polaris/api/Regexes.ts
spocke: Right the changelog item for polaris confused me to think that it was only these characters and also the ticket information. But good to have a full list and spec reference here. Thanks.

PR #7826, opened 2022-05-13
TINY-7795: Ctrl+Shift+Home/End doesn't work when content starts/ends with CEF block

vpyshnenko:
Related Ticket: TINY-7795

Description of Changes:
Support following operations when cef element at the start/end of the content 
* expand selection on Cmd+Shift+Up/Down
* collapse selection with arrow keys
* delete the selected content on Delete

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


spocke (changes requested):

modules/tinymce/src/core/main/ts/keyboard/NavigationUtils.ts
```diff
@@ -23,13 +25,26 @@ const moveToRange = (editor: Editor, rng: Range) => {
 const renderRangeCaretOpt = (editor: Editor, range: Range, scrollIntoView: boolean): Optional<Range> =>
   Optional.some(FakeCaretUtils.renderRangeCaret(editor, range, scrollIntoView));
 
+const isCefAtEdgeSelected = (editor: Editor): boolean => {
+  const body = SugarElement.fromDom(editor.getBody());
+  const rng = editor.selection.getRng();
+  return !rng.collapsed
+    && (Traverse.firstChild(body).exists((el) => isContentEditableFalse(el.dom) && el.dom === editor.selection.getStart())
```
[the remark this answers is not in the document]
vpyshnenko: done

modules/tinymce/src/core/main/ts/keyboard/CefNavigation.ts
```diff
@@ -91,8 +92,28 @@ const moveToLineEndPoint = (editor: Editor, forward: boolean): boolean => {
   return NavigationUtils.moveToLineEndPoint(editor, forward, isCefPosition);
 };
 
+const selectToEndPoint = (editor: Editor, forward: boolean): boolean => {
+  const root = editor.getBody();
+
+  const getEdgePosition = (forward: boolean): Optional<CaretPosition> =>
```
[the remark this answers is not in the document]
vpyshnenko: done

modules/tinymce/src/core/test/ts/browser/delete/CefDeleteTest.ts
```diff
@@ -140,4 +140,33 @@ describe('browser.tinymce.core.delete.CefDeleteTest', () => {
       })
     );
   });
+
+  context('cef is at the start/end of the content and covered with selection', () => {
+    it('TINY-7795: shoud delete selected content when cef block at the start', () => {
```
spocke, line 145:
```suggestion
    it('TINY-7795: should delete selected content when cef block at the start', () => {
```

modules/tinymce/src/core/test/ts/browser/keyboard/ArrowKeysEdgeCefTest.ts
```diff
@@ -0,0 +1,282 @@
...
+
+import Editor from 'tinymce/core/api/Editor';
+import * as CaretContainer from 'tinymce/core/caret/CaretContainer';
+
+describe('browser.tinymce.core.keyboard.ArrowKeysEdgeCefTest', () => {
+  const hook = TinyHooks.bddSetupLight<Editor>({
+    base_url: '/project/tinymce/js/tinymce',
+  }, [], true);
+
+  const assertStartContainer = (editor: Editor, f: (node: Node) => boolean) => {
```
spocke, line 14: Is the predicate option needed here from what I can tell it's always getting the same predicate passed in in this file.

modules/tinymce/src/core/main/ts/keyboard/HomeEndKeys.ts
```diff
@@ -11,6 +11,8 @@ const executeKeydownOverride = (editor: Editor, caret: Cell<Text>, evt: Keyboard
   MatchKeys.execute([
     { keyCode: VK.END, action: MatchKeys.action(CefNavigation.moveToLineEndPoint, editor, true) },
     { keyCode: VK.HOME, action: MatchKeys.action(CefNavigation.moveToLineEndPoint, editor, false) },
+    { keyCode: VK.END, action: MatchKeys.action(CefNavigation.selectToEndPoint, editor, true), ctrlKey: true, shiftKey: true },
```
spocke, line 14: It seems that this code path is not tested in the tests. At lest two basic tests one for each key would be useful. Also I wonder are we sure that these are not conflicting. That mac uses say Ctrl-Shift-Home for something else. Or does home even exist on a mac keyboard. Not a mac user so I have no idea. Just curious?

modules/tinymce/src/core/test/ts/browser/keyboard/ArrowKeysEdgeCefTest.ts
```diff
@@ -0,0 +1,282 @@
...
+    const startContainer = editor.selection.getRng().startContainer;
+    assert.isTrue(f(startContainer), 'Check selection is in caret container');
+  };
+
+  context('Cmd+Shift+Up/Down should expand selection to the edge of the content', () => {
+    it('TINY-7795: Cmd+Shift+Up when cef block at the start', () => {
```
spocke, line 20:
```suggestion
    it('TINY-7795: Cmd+Shift+Up when a cef block is at the start', () => {
```

Maybe add the `a` and `is` on all of them. Not sure. Maybe the native English people has better grammar ideas. :)

spocke (approved):
Think it looks very good Viktor, good work.

modules/tinymce/src/core/main/ts/keyboard/HomeEndKeys.ts
vpyshnenko: I've decided to handle Cmd+Shift+Up/Down only for the mac platform and Ctrl+Shift+Home/Down for the rest.

modules/tinymce/CHANGELOG.md
```diff
@@ -32,6 +32,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Switching between unordered and ordered lists would incorrectly alter any parent element that contained that list #TINY-8068
 - The URL detection used for `autolink` and smart paste didn't work if a path segment contained valid characters such as `!` and `:` #TINY-8069
 - Copy events were not dispatched in readonly mode #TINY-6800
+- Using Ctrl+Shift+Home/End (Cmd+Shift+Up/Down on Mac) could not expand the selection when `contenteditable="false"` element was at the edge of content #TINY-7795
```
lnewson, line 35:
```suggestion
- Using Ctrl+Shift+Home/End (Cmd+Shift+Up/Down on Mac) would not expand the selection when a `contenteditable="false"` element was at the edge of content #TINY-7795
```

modules/tinymce/src/core/main/ts/delete/CefDelete.ts
```diff
@@ -77,6 +78,13 @@ const backspaceDeleteRange = (editor: Editor, forward: boolean): Optional<() =>
       () => Optional.some(Fun.noop)
     );
   }
+
+  if (isCefAtEdgeSelected(editor)) {
+    return Optional.some(() => {
+      editor.selection.getRng().deleteContents();
+    });
+  }
```
lnewson, line 86: Why is this here? This Jira is only supposed to be able Home/End key navigation and this impacts deletion behaviour. I'm also concerned about this logic as `deleteContents()` doesn't clean up very well. This looks like it should be in a separate Jira and definitely needs a changelog entry.

modules/tinymce/src/core/main/ts/keyboard/ArrowKeys.ts
```diff
@@ -32,7 +32,13 @@ const executeKeydownOverride = (editor: Editor, caret: Cell<Text>, evt: Keyboard
     { keyCode: VK.LEFT, ctrlKey: !os.isMacOS(), altKey: os.isMacOS(), action: MatchKeys.action(BoundarySelection.movePrevWord, editor, caret) },
     { keyCode: VK.UP, action: MatchKeys.action(ContentEndpointNavigation.moveV, editor, false) },
     { keyCode: VK.DOWN, action: MatchKeys.action(ContentEndpointNavigation.moveV, editor, true) }
-  ], evt).each((_) => {
+  ];
+  if (os.isMacOS()) {
+    keyPatterns.push({ keyCode: VK.UP, action: MatchKeys.action(CefNavigation.selectToEndPoint, editor, false), metaKey: true, shiftKey: true });
+    keyPatterns.push({ keyCode: VK.DOWN, action: MatchKeys.action(CefNavigation.selectToEndPoint, editor, true), metaKey: true, shiftKey: true });
+  }
```
lnewson, line 39: This is doing mutation which we generally try to avoid... this looks like it could be avoided by using a spread inside the original and would also make sure we can order it appropriately.

For example:
```ts
  ...(os.isMacOS() ? [
    { keyCode: VK.UP, action: MatchKeys.action(CefNavigation.selectToEndPoint, editor, false), metaKey: true, shiftKey: true },
    { keyCode: VK.DOWN, action: MatchKeys.action(CefNavigation.selectToEndPoint, editor, true), metaKey: true, shiftKey: true }
  ] : [])
```

modules/tinymce/src/core/main/ts/keyboard/CefNavigation.ts
```diff
@@ -91,8 +92,25 @@ const moveToLineEndPoint = (editor: Editor, forward: boolean): boolean => {
...
+      forward ? rng.setStart(curRng.startContainer, curRng.startOffset) : rng.setEnd(curRng.endContainer, curRng.endOffset);
+      return rng;
+    })
+    .fold(
+      Fun.never,
+      (rng) => {
+        NavigationUtils.moveToRange(editor, rng);
+        return true;
+      }
+    );
```
lnewson, line 109: A `fold` where the none case is `Fun.never` is the same as using `exists` instead.
```suggestion
    .exists((rng) => {
      NavigationUtils.moveToRange(editor, rng);
      return true;
    });
```

modules/tinymce/src/core/main/ts/keyboard/CefNavigation.ts
```diff
@@ -91,8 +92,25 @@ const moveToLineEndPoint = (editor: Editor, forward: boolean): boolean => {
   return NavigationUtils.moveToLineEndPoint(editor, forward, isCefPosition);
 };
 
+const selectToEndPoint = (editor: Editor, forward: boolean): boolean =>
+  getEdgeCefPosition(editor, !forward)
+    .map((pos) => {
+      const rng = pos.toRange();
+      const curRng = editor.selection.getRng();
+      forward ? rng.setStart(curRng.startContainer, curRng.startOffset) : rng.setEnd(curRng.endContainer, curRng.endOffset);
```
lnewson, line 100: Ternaries should not be used for setting values. Please convert this to an if/else.
```suggestion
      if (forward) {
        rng.setStart(curRng.startContainer, curRng.startOffset);
      } else {
        rng.setEnd(curRng.endContainer, curRng.endOffset);
      }
```

modules/tinymce/src/core/main/ts/keyboard/HomeEndKeys.ts
```diff
@@ -8,14 +9,23 @@ import * as MatchKeys from './MatchKeys';
 import * as MediaNavigation from './MediaNavigation';
 
 const executeKeydownOverride = (editor: Editor, caret: Cell<Text>, evt: KeyboardEvent) => {
-  MatchKeys.execute([
+  const os = PlatformDetection.detect().os;
```
lnewson, line 12: This should be using the `Env` API instead of doing this. Basically for TinyMCE we should always use `Env` except for in tests.

modules/tinymce/src/core/main/ts/keyboard/HomeEndKeys.ts
```diff
@@ -8,14 +9,23 @@ import * as MatchKeys from './MatchKeys';
...
     { keyCode: VK.HOME, action: MatchKeys.action(CefNavigation.moveToLineEndPoint, editor, false) },
+    { keyCode: VK.END, action: MatchKeys.action(CefNavigation.selectToEndPoint, editor, true), ctrlKey: true, shiftKey: true },
+    { keyCode: VK.HOME, action: MatchKeys.action(CefNavigation.selectToEndPoint, editor, false), ctrlKey: true, shiftKey: true },
     { keyCode: VK.END, action: MatchKeys.action(MediaNavigation.moveToLineEndPoint, editor, true) },
     { keyCode: VK.HOME, action: MatchKeys.action(MediaNavigation.moveToLineEndPoint, editor, false) },
     { keyCode: VK.END, action: MatchKeys.action(BoundarySelection.moveToLineEndPoint, editor, true, caret) },
     { keyCode: VK.HOME, action: MatchKeys.action(BoundarySelection.moveToLineEndPoint, editor, false, caret) }
-  ], evt).each((_) => {
+  ];
+  if (!os.isMacOS()) {
```
lnewson, line 24: Likewise this would be better using a spread operator, especially in this case where it should be beside the other `CefNavigation` logic.

modules/tinymce/src/core/main/ts/keyboard/NavigationUtils.ts
```diff
@@ -30,6 +31,11 @@ const moveHorizontally = (editor: Editor, direction: HDirection, range: Range, i
   const getNextPosFn = Fun.curry(CaretUtils.getVisualCaretPosition, forwards ? caretWalker.next : caretWalker.prev);
   const isBeforeFn = forwards ? isBefore : isAfter;
 
+  if (isCefAtEdgeSelected(editor)) {
+    const newRange = range.cloneRange();
+    newRange.collapse(direction === HDirection.Backwards);
+    return Optional.from(newRange);
+  }
   if (!range.collapsed) {
```
lnewson, line 39:
```suggestion
  } else if (!range.collapsed) {
```
We should never have 2 if's together like this, especially when they are mutually exclusive.

modules/tinymce/src/core/test/ts/browser/delete/CefDeleteTest.ts
```diff
@@ -140,4 +140,33 @@ describe('browser.tinymce.core.delete.CefDeleteTest', () => {
...
+    });
+
+    it('TINY-7795: should delete selected content when cef block at the start and at the end', () => {
+      const editor = hook.editor();
+      editor.setContent('<p contenteditable="false">CEF</p><p>abc</p><p contenteditable="false">CEF</p>');
+      editor.execCommand('SelectAll');
+      TinyContentActions.keystroke(editor, Keys.backspace());
+      TinyAssertions.assertContent(editor, '');
+    });
+  });
```
lnewson, line 171: As mentioned elsewhere this definitely isn't part of the scope for TINY-7795. I'm guessing this is something you found? If so it should be logged as a separate Jira and done in that instead of grouping it in with this change.

modules/tinymce/src/core/main/ts/keyboard/NavigationUtils.ts
```diff
@@ -69,6 +75,9 @@ const moveHorizontally = (editor: Editor, direction: HDirection, range: Range, i
 
 const moveVertically = (editor: Editor, direction: LineWalker.VDirection, range: Range, isBefore: (caretPosition: CaretPosition) => boolean,
                         isAfter: (caretPosition: CaretPosition) => boolean, isElement: (node: Node) => node is Element): Optional<Range> => {
+  if (!range.collapsed) {
```
lnewson, line 78: Is there a reason we're doing this in the vertical case and not the horizontal case? This is just a code smell sorry as it seems like an odd change.

modules/tinymce/src/core/test/ts/browser/keyboard/ArrowKeysEdgeCefTest.ts
```diff
@@ -0,0 +1,299 @@
...
+      TinyAssertions.assertCursor(editor, [ 0 ], 0);
+
+      editor.execCommand('SelectAll');
+      TinyContentActions.keystroke(editor, Keys.down());
+      assertRangeInCaretContainerBlock(editor);
+      // actual content:<p contenteditable="false">CEF</p><p>abc</p><p contenteditable="false">CEF</p><p data-mce-caret="after"><br data-mce-bogus="1"></p>
+      TinyAssertions.assertCursor(editor, [ 3 ], 0);
+    });
+  });
+
```
lnewson, line 297:
```suggestion
```

lnewson (changes requested):
Looks good overall Viktor and nice work on the test coverage! Marking as needs changes though largely because of the delete thing being adding in when it's not in scope.

modules/tinymce/src/core/test/ts/browser/delete/CefDeleteTest.ts
vpyshnenko: TINY-7795

vpyshnenko: I've created follow up TINY-8729

lnewson: @vpyshnenko sorry for the confusion here, I didn't mean it had to be deleted from this PR, only that it wasn't about navigation and as such should have been a separate Jira with an associated changelog entry. I have no issues with the 2 being done in this single PR.

As for the text, given I was in the call that was created for... the intention/meaning was to make sure the selection was in such a state that deletion would still work, not to create a custom deletion override. I haven't looked into it enough if we actually need to the override though but I'm assuming it was given your investigation.

lnewson: As discussed on slack, we're going to leave it as 2 separate PRs since Viktor has already reverted this change.

modules/tinymce/src/core/main/ts/keyboard/NavigationUtils.ts
vpyshnenko: Fixed

modules/tinymce/src/core/main/ts/keyboard/NavigationUtils.ts
```diff
@@ -72,13 +78,21 @@ const moveVertically = (editor: Editor, direction: LineWalker.VDirection, range:
...
   const forwards = direction === LineWalker.VDirection.Down;
+  const root = editor.getBody();
 
   if (!caretClientRect) {
     return Optional.none();
   }
 
+  if (isCefAtEdgeSelected(editor)) {
+    const caretPosition = forwards ? CaretPosition.fromRangeEnd(range) : CaretPosition.fromRangeStart(range);
+    return (!forwards ? LineReader.getClosestPositionAbove(root, caretPosition) : LineReader.getClosestPositionBelow(root, caretPosition))
```
lnewson, line 89: This line has a fair bit of complexity with the ternary as well. I'd suggest we split this up over 2 lines which will also help with readability. For example:
```suggestion
    const getClosestFn = !forwards ? LineReader.getClosestPositionAbove : LineReader.getClosestPositionBelow;
    return getClosestFn(root, caretPosition)
```

modules/tinymce/src/core/main/ts/keyboard/NavigationUtils.ts
```diff
@@ -72,13 +78,21 @@ const moveVertically = (editor: Editor, direction: LineWalker.VDirection, range:
...
+      .or(Optional.from(caretPosition))
```
lnewson, line 90: This is going to eagerly evaluate and create the Optional when not needed, so this should be using `orThunk`.
```suggestion
      .orThunk(() => Optional.from(caretPosition))
```

lnewson (approved):

PR #7830, opened 2022-05-16
TINY-8078: fixed list normalization on cut/delete/forwarddelete

spocke:
Related Ticket:

Description of Changes:
* Added list normalization on delete/forward delete these are used for the cut operation

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/tinymce/CHANGELOG.md
```diff
@@ -31,6 +31,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Indenting or outdenting list items inside a block element that was inside another list item would not work #TINY-7209
 - Switching between unordered and ordered lists would incorrectly alter any parent element that contained that list #TINY-8068
 - The URL detection used for `autolink` and smart paste didn't work if a path segment contained valid characters such as `!` and `:` #TINY-8069
+- Cutting contents to clipboard while selecting between the parent list and a nested list would not always set the list style to `none` on the parent list #TINY-8078
```
spocke, line 34: Not sure about how to describe this to make any sense for anyone reading it in a short sentence.

modules/tinymce/src/plugins/lists/main/ts/core/Delete.ts
```diff
@@ -237,13 +237,16 @@ const backspaceDeleteIntoListCaret = (editor: Editor, isForward: boolean): boole
 const backspaceDeleteCaret = (editor: Editor, isForward: boolean): boolean => {
   return backspaceDeleteFromListToListCaret(editor, isForward) || backspaceDeleteIntoListCaret(editor, isForward);
 };
-
-const backspaceDeleteRange = (editor: Editor): boolean => {
+const hasListSelection = (editor: Editor) => {
```
lnewson, line 240:
```suggestion

const hasListSelection = (editor: Editor) => {
```

modules/tinymce/src/plugins/lists/main/ts/core/Delete.ts
```diff
@@ -260,6 +263,13 @@ const backspaceDelete = (editor: Editor, isForward: boolean): boolean => {
 };
 
 const setup = (editor: Editor): void => {
+  editor.on('ExecCommand', (e) => {
+    const cmd = e.command.toLocaleLowerCase();
```
lnewson, line 267:
```suggestion
    const cmd = e.command.toLowerCase();
```
We probably don't want this differing based on the users locale.

modules/tinymce/src/plugins/lists/main/ts/core/Delete.ts
```diff
@@ -260,6 +263,13 @@ const backspaceDelete = (editor: Editor, isForward: boolean): boolean => {
...
+    if ((cmd === 'delete' || cmd === 'forwarddelete') && hasListSelection(editor)) {
```
lnewson, line 268: Should this run if the command has been prevented? I guess it shouldn't hurt.

modules/tinymce/CHANGELOG.md
lnewson:
```suggestion
- Cutting content to the clipboard while selecting between the parent list and a nested list would not always set the list style to `none` on the parent list #TINY-8078
```
I can't think of anything better to describe this either, so my suggestion is only for grammar.

lnewson (approved):

modules/tinymce/CHANGELOG.md
spocke: Yeah, sometimes it's hard to describe these edge case scenarios.

modules/tinymce/src/plugins/lists/main/ts/core/Delete.ts
spocke: Ooops. Good catch I should have seen that.

modules/tinymce/src/plugins/lists/main/ts/core/Delete.ts
spocke: I don't think we should prevent it here since the command has already been executed at that point it's not like it's preventing the `ExecCommand` from executing that is the `BeforeExecCommand` event. I guess the event name here is a bit missleading maybe they should have been `ExecCommand` and `AfterExecCommand` instead not sure.

lnewson: Yeah I realised this was what happens after, it was more just in case someone wanted to prevent this for whatever reason. It's an extreme edge case though so more than happy to leave it as is, I just thought it was worth asking.

ltrouton (approved):

PR #7834, opened 2022-05-17
TINY-4550: Selecting all content with a single image in the content was inconsistent for the keyboard shortcut and menu item

hamza0867:
Related Ticket: TINY-4550

Description of Changes:
When there is only an image in the content, then running ctrl+a (cmd+a) should not select only the image but the content wrapping the image

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/tinymce/src/core/main/ts/util/Quirks.ts
```diff
@@ -445,10 +445,16 @@ const Quirks = (editor: Editor): Quirks => {
   const normalizeSelection = () => {
     // Normalize selection for example <b>a</b><i>|a</i> becomes <b>a|</b><i>a</i>
     editor.on('keyup focusin mouseup', (e) => {
+
+      // The next three lines address TINY-4550: when there is only an image in the editor then normalization should not run
+      const selectedNode = selection.getNode();
```
spocke, line 450: Not sure but I think I would prefer something like this to see if all of the content is selected:
```
const body = editor.getBody();
const selectionIsAllContent = rng.startContainer === rng.endContainer && rng.startContainer === body && rng.startOffset === 0 && rng.endOffset === body.childNodes.length;
```
Since the 'SelectAll' command sets the explicit range to the whole contents of the editor.

modules/tinymce/src/core/test/ts/browser/dom/SelectionQuirksTest.ts
```diff
@@ -68,4 +69,16 @@ describe('browser.tinymce.core.dom.SelectionQuirksTest', () => {
...
+  it('TINY-4550: Normalization should not run after selecting all with keyboard shortcut when there is only an image in the content', () => {
+    const editor = hook.editor();
+    resetNormalizeCounter();
+    editor.setContent('<p><img src="https://www.tiny.cloud/images/glyph-tinymce@2x.png" alt="" width="354" height="116"></p>');
+    TinySelections.setCursor(editor, [ 0 ], 1);
+    editor.shortcuts.add('meta+a', null, 'SelectAll');
+    const isMac = Env.os.isMacOS() || Env.os.isiOS();
+    TinyContentActions.keydown(editor, 65, { metaKey: isMac, ctrlKey: !isMac });
+    TinyContentActions.keyup(editor, isMac ? 224 : 17, { });
+    Assertions.assertEq('Selection node should be the p node', '<p><img src="https://www.tiny.cloud/images/glyph-tinymce@2x.png" alt="" width="354" height="116"></p>', editor.selection.getContent());
```
spocke, line 82: You could use the TinyAssertions.assertSelection(editor, [0], 0, [0], 1) here to verify that the image is selected. When checking the selection state we usually use that or assertCursor.

modules/tinymce/src/core/main/ts/util/Quirks.ts
lnewson: There's also the `SelectionUtils.hasAllContentsSelected` internal API that might be able to be used here instead?

spocke:
The branch should be `feature/TINY-4550` also could you add a test for the editor.selection.normalize() method as well. Just set the selection to enclose the whole content using `TinySelection.setSelection(editor, [], 0, [], 1);` on a paragraph with a image then run `editor.selection.normalize();` then `TInyAssertions.assertSelection(editor, [], 0, [], 1);`.

modules/tinymce/src/core/main/ts/util/Quirks.ts
spocke: That seems to travel down so it would be true for the <p>[]</p> case and in theory we would want to normalize say when the start point is before a text node.

modules/tinymce/src/core/main/ts/api/util/VK.ts
```diff
@@ -22,9 +22,11 @@ interface VK {
   PAGE_DOWN: number;
   END: number;
   HOME: number;
+  CTRL: number;
+  META: number;
 
-  modifierPressed: (e: KeyboardLikeEvent) => boolean;
-  metaKeyPressed: (e: KeyboardLikeEvent) => boolean;
+  modifierPressed: (e: (KeyboardLikeEvent & { keyCode?: number })) => boolean;
+  metaKeyPressed: (e: (KeyboardLikeEvent & { keyCode?: number })) => boolean;
```
lnewson, line 29: This is should be part of the `KeyboardLikeEvent` instead of patched here.

modules/tinymce/src/core/main/ts/api/util/VK.ts
```diff
@@ -46,14 +48,16 @@ const VK: VK = {
...
+  CTRL: 17,
+  META: 224,
 
-  modifierPressed: (e: KeyboardLikeEvent): boolean => {
+  modifierPressed: (e: (KeyboardLikeEvent & { keyCode?: number })): boolean => {
     return e.shiftKey || e.ctrlKey || e.altKey || VK.metaKeyPressed(e);
   },
 
-  metaKeyPressed: (e: KeyboardLikeEvent): boolean => {
+  metaKeyPressed: (e: (KeyboardLikeEvent & { keyCode?: number })): boolean => {
```
lnewson, line 58: This is a breaking change so not something we should be doing. This is used within the `modifierPressed` logic and a modifier isn't pressed if pressing the actual meta key.

To clarify, these aren't meant to be used to see if the key itself is pressed, it's meant to be used to see if the key is pressed in addition to another key.

lnewson (changes requested):

lnewson:
Also @hamza0867 please make sure to fix the PR title as it's used for the commit message when being squashed. It's also hard to know what this change is about when looking at the PR list.

modules/tinymce/CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The default value for `end_container_on_empty_block` option has been changed to `'blockquote'` #TINY-6559
 
 ### Fixed
+- Select all with a single image is inconsistent #TINY-4550
```
lnewson, line 16: Changelog entries should be in past tense. So maybe something like this?
```suggestion
- Selecting all content with a single image in the content was inconsistent for the keyboard shortcut and menu item #TINY-4550
```

modules/tinymce/src/core/main/ts/api/util/VK.ts
hamza0867: I thought putting it inside `KeyboardLikeEvent` breaks some other code. I was mistaken. Will fix now!

modules/tinymce/src/core/main/ts/api/util/VK.ts
spocke: Any file that is in the api directory is a public api. So changing things there is more risky since third party might use that. Also agree with Lee here that it's supposed to check for modifiers states of key events so it behavior is being changed from that to check for modifiers and the modifier key it self. Just move the keyCode check to the normalize if statement where this is being called.

modules/tinymce/src/core/main/ts/api/util/VK.ts
```diff
@@ -22,9 +23,11 @@ interface VK {
   PAGE_DOWN: number;
   END: number;
   HOME: number;
+  CTRL: number;
+  META: number;
 
-  modifierPressed: (e: KeyboardLikeEvent) => boolean;
-  metaKeyPressed: (e: KeyboardLikeEvent) => boolean;
+  modifierPressed: (e: KeyboardLikeEvent ) => boolean;
```
spocke, line 29: Why is there a extra space on these?

modules/tinymce/src/core/test/ts/browser/dom/SelectionQuirksTest.ts
```diff
@@ -64,8 +65,21 @@ describe('browser.tinymce.core.dom.SelectionQuirksTest', () => {
     TinySelections.setSelection(editor, [ 0, 0, 0 ], 0, [ 0, 0 ], 0);
     TinyContentActions.keyup(editor, Keys.left(), { shift: true });
     assertNormalizeCounter(0);
-    TinyContentActions.keyup(editor, 17, { }); // single ctrl
-    assertNormalizeCounter(1);
+    const isMac = Env.os.isMacOS() || Env.os.isiOS();
+    TinyContentActions.keyup(editor, isMac ? 224 : 17, { }); // single ctrl
```
spocke, line 69: I guess you could use the VK key constants here as well. If we decide to keep them that is.

modules/tinymce/src/core/main/ts/api/util/VK.ts
```diff
@@ -46,14 +49,16 @@ const VK: VK = {
   PAGE_DOWN: 34,
   END: 35,
   HOME: 36,
+  CTRL: 17,
```
spocke, line 52: I wonder if we need to document these in the changelog or is it such a fringe api thing that we don't care. What do you guys think? We could also have a set of internal constants but I do see this as being useful in say plugins. At least until we migrate everything to the modern `core` property where everything is strings.

modules/tinymce/src/core/test/ts/browser/dom/SelectionQuirksTest.ts
```diff
@@ -64,8 +65,21 @@ describe('browser.tinymce.core.dom.SelectionQuirksTest', () => {
...
+    assertNormalizeCounter(0);
     TinyAssertions.assertSelection(editor, [ 0, 0 ], 0, [ 0, 0 ], 0);
   });
+
+  it('TINY-4550: Normalization should not run after selecting all with keyboard shortcut when there is only an image in the content', () => {
+    const editor = hook.editor();
+    resetNormalizeCounter();
+    editor.setContent('<p><img src="https://www.tiny.cloud/images/glyph-tinymce@2x.png" alt="" width="354" height="116"></p>');
```
spocke, line 77: Maybe use a data url or just `about:blank` since having tests that does a http request out would be a think that would be nice to avoid if the internet is flaky or slow. The `about:blank` is an approach I've used else ware sure the image gets broken since that is not a real url but it doesn't really matter for the tests. Also the image size is irrelevant so it could just be `<p></p>` short and clean.

spocke (changes requested):

modules/tinymce/src/core/main/ts/api/util/VK.ts
hamza0867: Just forgotten, I thought eslint would automatically remove them.
Done!

modules/tinymce/src/core/main/ts/api/util/VK.ts
hamza0867: Done!

modules/tinymce/src/core/main/ts/api/util/VK.ts
hamza0867: Done!

modules/tinymce/src/core/test/ts/browser/dom/SelectionQuirksTest.ts
hamza0867: Done!

modules/tinymce/src/core/test/ts/browser/dom/SelectionQuirksTest.ts
hamza0867: Nice, I like it!
Done!

modules/tinymce/src/core/main/ts/api/util/VK.ts
lnewson: I really don't think we should even be including these here... at the very least the information here isn't correct as there's a left/right key for each of these that needs to be accounted for both here and in the actual fix. In addition to that, the meta keycode can actually be different values (e.g. on Chrome its 91/93 and on Firefox it's always 224).

lnewson (changes requested):
Looks better thanks @hamza0867, however as noted there's 2 different keys that can be used to trigger Select All and that needs to be accounted for as well.

modules/tinymce/src/core/main/ts/api/util/VK.ts
lnewson: I should add that, yes if we are leaving these here then we'll need a changelog entry for this change.

modules/tinymce/src/core/test/ts/browser/dom/SelectionQuirksTest.ts
```diff
@@ -4,6 +4,8 @@ import { TinyAssertions, TinyContentActions, TinyDom, TinyHooks, TinySelections
 import { assert } from 'chai';
 
 import Editor from 'tinymce/core/api/Editor';
+import Env from 'tinymce/core/api/Env';
+import { VK } from 'tinymce/core/api/PublicApi';
```
lnewson, line 8: This should be importing from VK directly instead of from the `PublicApi` that just exports everything.
```suggestion
import VK from 'tinymce/core/api/util/VK';
```

modules/tinymce/src/core/test/ts/browser/dom/SelectionQuirksTest.ts
```diff
@@ -64,8 +65,20 @@ describe('browser.tinymce.core.dom.SelectionQuirksTest', () => {
     TinySelections.setSelection(editor, [ 0, 0, 0 ], 0, [ 0, 0 ], 0);
     TinyContentActions.keyup(editor, Keys.left(), { shift: true });
     assertNormalizeCounter(0);
-    TinyContentActions.keyup(editor, 17, { }); // single ctrl
-    assertNormalizeCounter(1);
+    const isMac = Env.os.isMacOS() || Env.os.isiOS();
+    editor.dispatch('keyup', new KeyboardEvent('keyup', { key: 'Control' }));
+    assertNormalizeCounter(isMac ? 1 : 0);
```
lnewson, line 70: Why is this different on Mac? The goal should be to have it be consistent across browsers.

modules/tinymce/src/core/main/ts/util/Quirks.ts
hamza0867: > Not sure but I think I would prefer something like this to see if all of the content is selected:
> 
> ```
> const body = editor.getBody();
> const rng = editor.selection.getRng();
> const selectionIsAllContent = rng.startContainer === rng.endContainer && rng.startContainer === body && rng.startOffset === 0 && rng.endOffset === body.childNodes.length;
> ```
> 
> Since the 'SelectAll' command sets the explicit range to the whole contents of the editor.

Done!

modules/tinymce/src/core/test/ts/browser/dom/SelectionQuirksTest.ts
hamza0867: Done!

modules/tinymce/src/core/main/ts/api/util/VK.ts
hamza0867: This change was removed. Should this discussion be resolved then ?

modules/tinymce/src/core/main/ts/util/Quirks.ts
```diff
@@ -448,7 +448,7 @@ const Quirks = (editor: Editor): Quirks => {
       // no point to exclude Ctrl+A, since normalization will still run after Ctrl will be unpressed
       // better exclude any key combinations with the modifiers to avoid double normalization
       // (also addresses TINY-1130)
-      if (!VK.modifierPressed(e)) {
+      if (!VK.modifierPressed(e) && (isMac ? (e.key !== 'Meta') : (e.key !== 'Control'))) {
```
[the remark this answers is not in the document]
hamza0867: This change was removed. Should this discussion be resolved ?

modules/tinymce/src/core/test/ts/browser/dom/SelectionQuirksTest.ts
hamza0867: This change was removed. Should this discussion be resolved ?

modules/tinymce/src/core/test/ts/browser/dom/SelectionQuirksTest.ts
```diff
@@ -64,8 +65,20 @@ describe('browser.tinymce.core.dom.SelectionQuirksTest', () => {
     TinySelections.setSelection(editor, [ 0, 0, 0 ], 0, [ 0, 0 ], 0);
     TinyContentActions.keyup(editor, Keys.left(), { shift: true });
     assertNormalizeCounter(0);
-    TinyContentActions.keyup(editor, 17, { }); // single ctrl
+    TinyContentActions.keyup(editor, 17, { }); // Single Ctrl
     assertNormalizeCounter(1);
     TinyAssertions.assertSelection(editor, [ 0, 0 ], 0, [ 0, 0 ], 0);
   });
+
+  it('TINY-4550: Normalization should not run after selecting all with keyboard shortcut when there is only an image in the content', () => {
```
lnewson, line 73:
```suggestion
  it('TINY-4550: Normalization should not run after selecting all when there is only an image in the content', () => {
```
This should happen regardless of how the select all is triggered, so this just helps clarify that I think.

modules/tinymce/src/core/main/ts/util/Quirks.ts
```diff
@@ -448,7 +448,12 @@ const Quirks = (editor: Editor): Quirks => {
       // no point to exclude Ctrl+A, since normalization will still run after Ctrl will be unpressed
       // better exclude any key combinations with the modifiers to avoid double normalization
       // (also addresses TINY-1130)
-      if (!VK.modifierPressed(e)) {
+
+      // The next three lines address TINY-4550
+      const body = editor.getBody();
+      const rng = editor.selection.getRng();
+      const selectionIsAllContent = rng.startContainer === rng.endContainer && rng.startContainer === body && rng.startOffset === 0 && rng.endOffset === body.childNodes.length;
```
lnewson, line 455: Can we move this out to a separate helper function? That'll have 2 benefits: 1) it's more readable and clear what's going on and 2) it won't evaluate it when it's not needed (e.g. if a modifier key is pressed).

modules/tinymce/src/core/test/ts/browser/dom/SelectionQuirksTest.ts
lnewson: Also should we have a test for `mouseup` as well (or similar)?

modules/tinymce/src/core/main/ts/util/Quirks.ts
```diff
@@ -438,6 +438,16 @@ const Quirks = (editor: Editor): Quirks => {
     });
   };
 
+  /**
```
spocke, line 441: Is this stating the obvious? The function name is pretty clear. So I think the comment is not needed.

modules/tinymce/src/core/main/ts/util/Quirks.ts
```diff
@@ -438,6 +438,16 @@ const Quirks = (editor: Editor): Quirks => {
...
+   * This is a helper function to test if the selection is all the content available
+   */
+  const isAllContentSelected = (editor: Editor): boolean => {
+    // The next three lines address TINY-4550
```
spocke, line 445: Maybe move this to the comment below instead at the call site and maybe explain why it's there that we check if all the contents is selected to avoid it from normalizing inwards to an image not sure about the wording.

spocke (approved):

lnewson (approved):

TheSpyder (approved):

PR #7835, opened 2022-05-18
TINY-8639: Retain formatted blank lines when `format_empty_lines` is true

ltrouton:
Related Ticket: TINY-8639

Description of Changes:
* Update the schema and DomParser so that text inline elements can be padded if they are a child of an empty block. As to whether text inline elements are padded by default is controlled by the `format_empty_lines` option

Coming up with setting and property names for the schema was tricky so am open to suggestions if we are not happy with what I have got currently

Pre-checks:
* [X] Changelog entry added

Review:
* [X] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/api/html/DomParser.ts
+++ b/modules/tinymce/src/core/main/ts/api/html/DomParser.ts
@@ -12,7 +12,7 @@ import { BlobCache } from '../file/BlobCache';
 import Tools from '../util/Tools';
 import * as URI from '../util/URI';
 import AstNode from './Node';
-import Schema, { SchemaRegExpMap } from './Schema';
+import Schema, { getTextRootBlockElements, SchemaRegExpMap } from './Schema';
 
 /**
  * @summary
@@ -279,8 +279,7 @@ const whitespaceCleaner = (root: AstNode, schema: Schema, settings: DomParserSet
   const nonEmptyElements = schema.getNonEmptyElements();
   const whitespaceElements = schema.getWhitespaceElements();
   const blockElements: Record<string, string> = extend(makeMap('script,style,head,html,body,title,meta,param'), schema.getBlockElements());
-  // NOTE: This is duplicated in ApplyFormat.ts
-  const textRootBlockElements: Record<string, string> = extend(makeMap('td,th,li,dt,dd,figcaption,caption,details,summary'), schema.getTextBlockElements());
+  const textRootBlockElements = getTextRootBlockElements(schema);
   const allWhiteSpaceRegExp = /[ \t\r\n]+/g;
   const startWhiteSpaceRegExp = /^[ \t\r\n]+/;
   const endWhiteSpaceRegExp = /[ \t\r\n]+$/;
--- a/modules/tinymce/src/core/main/ts/api/html/DomParser.ts
+++ b/modules/tinymce/src/core/main/ts/api/html/DomParser.ts
@@ -279,6 +279,8 @@ const whitespaceCleaner = (root: AstNode, schema: Schema, settings: DomParserSet
   const nonEmptyElements = schema.getNonEmptyElements();
   const whitespaceElements = schema.getWhitespaceElements();
   const blockElements: Record<string, string> = extend(makeMap('script,style,head,html,body,title,meta,param'), schema.getBlockElements());
+  // NOTE: This is duplicated in ApplyFormat.ts
+  const textRootBlockElements: Record<string, string> = extend(makeMap('td,th,li,dt,dd,figcaption,caption,details,summary'), schema.getTextBlockElements());
   const allWhiteSpaceRegExp = /[ \t\r\n]+/g;
   const startWhiteSpaceRegExp = /^[ \t\r\n]+/;
   const endWhiteSpaceRegExp = /[ \t\r\n]+$/;
@@ -290,6 +292,18 @@ const whitespaceCleaner = (root: AstNode, schema: Schema, settings: DomParserSet
         return true;
       } else {
         node = node.parent;
+      }
+    }
+    return false;
+  };
+
+  const isTextRootBlockEmpty = (node: AstNode) => {
+    let tempNode = node;
+    while (Type.isNonNullable(tempNode)) {
+      if (tempNode.name in textRootBlockElements) {
+        return isEmpty(schema, nonEmptyElements, whitespaceElements, tempNode);
+      } else {
+        tempNode = tempNode.parent;
       }
     }
     return false;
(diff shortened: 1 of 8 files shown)


modules/tinymce/src/core/main/ts/api/html/DomParser.ts
```diff
@@ -279,6 +279,8 @@ const whitespaceCleaner = (root: AstNode, schema: Schema, settings: DomParserSet
   const nonEmptyElements = schema.getNonEmptyElements();
   const whitespaceElements = schema.getWhitespaceElements();
   const blockElements: Record<string, string> = extend(makeMap('script,style,head,html,body,title,meta,param'), schema.getBlockElements());
+  // NOTE: This is duplicated in ApplyFormat.ts
+  const textRootBlockElements: Record<string, string> = extend(makeMap('td,th,li,dt,dd,figcaption,caption,details,summary'), schema.getTextBlockElements());
```
lnewson, line 283: Instead of duplicating this can we make it a shared function or similar please?

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
```diff
@@ -332,14 +346,27 @@ const whitespaceCleaner = (root: AstNode, schema: Schema, settings: DomParserSet
...
+            paddEmptyNode(settings, args, blockElements, node);
           } else {
-            node.unwrap();
+            removeOrPadd();
           }
-        } else if (elementRule.paddEmpty && (isNodeEmpty || isPaddedWithNbsp(node))) {
-          paddEmptyNode(settings, args, blockElements, node);
+        } else {
+          removeOrPadd();
         }
```
lnewson, line 370: This is over complicated and can be simplified to something like this I believe:
```ts
        if (elementRule.paddInEmptyBlock && isNodeEmpty && isTextRootBlockEmpty(node)) {
          paddEmptyNode(settings, args, blockElements, node);
        } else if (elementRule.removeEmpty && isNodeEmpty) {
          if (blockElements[node.name]) {
            node.remove();
          } else {
            node.unwrap();
          }
        } else if (elementRule.paddEmpty && (isNodeEmpty || isPaddedWithNbsp(node))) {
          paddEmptyNode(settings, args, blockElements, node);
        }
```
This also helps with readability to make it show the order the properties are evaluated.

modules/tinymce/src/core/main/ts/api/html/Schema.ts
```diff
@@ -15,6 +15,7 @@ export interface SchemaSettings {
   valid_elements?: string;
   valid_styles?: string | Record<string, string>;
   verify_html?: boolean;
+  retain_empty_block_inline_children?: boolean;
```
lnewson, line 18:
```suggestion
  keep_empty_block_inline_children?: boolean;
```
Maybe? As looking elsewhere it seems we've used `keep` previously. That said given this enables the padding functionality maybe we should be using `padd` or similar instead?

modules/tinymce/src/core/main/ts/api/html/Schema.ts
```diff
@@ -464,7 +466,7 @@ const Schema = (settings?: SchemaSettings): Schema => {
   const blockElementsMap = createLookupTable('block_elements', 'hr table tbody thead tfoot ' +
     'th tr td li ol ul caption dl dt dd noscript menu isindex option ' +
     'datalist select optgroup figcaption details summary', textBlockElementsMap);
-  const textInlineElementsMap = createLookupTable('text_inline_elements', 'span strong b em i font strike u var cite ' +
+  const textInlineElementsMap = createLookupTable('text_inline_elements', 'span strong b em i font s strike u var cite ' +
```
lnewson, line 469: Does this need a changelog entry as it looks like an additional bug fix?

modules/tinymce/src/core/test/ts/browser/fmt/FormatEmptyLineTest.ts
```diff
@@ -45,163 +53,377 @@ describe('browser.tinymce.core.fmt.FormatEmptyLineTest', () => {
...
+    TinyAssertions.assertContentPresence(editor, expectedPresence);
+    if (Type.isNonNullable(expectedHtml)) {
+      TinyAssertions.assertContent(editor, expectedHtml);
+    }
+    if (Type.isNonNullable(expectedRawHtml)) {
+      TinyAssertions.assertRawContent(editor, expectedRawHtml);
+    }
+    config.remove(editor);
+    TinyAssertions.assertContentPresence(editor, expectedPresenceOnRemove);
+    // TinyAssertions.assertContent(editor, html);
```
lnewson, line 76: Is this needed here? If not we probably should remove this.

lnewson (commented):
Just leaving a few comments for now, I still need to come back and review properly sorry Luke.

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
ltrouton: I had something similar stored in a text file but initially thought an early return would be required. Alhtough, what you have is much better and is equivalent to what I had previously so I will make the change. Thanks

modules/tinymce/src/core/main/ts/api/html/Schema.ts
ltrouton: I think `padd_empty_block_inline_children` could work well and is the clearest representation of what is happening

ltrouton: Switched over to `padd_empty_block_inline_children`

modules/tinymce/src/core/main/ts/api/html/Schema.ts
ltrouton: Good point. I added a couple more changelog entries for the other schema fixes I made

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
ltrouton: Done

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
ltrouton: I can't quite think of a good place for it so not sure if you have any thoughts? Maybe a new file is needed. It would be good to integrate into the schema somehow but I pretty sure that would require a minor.

spocke (approved):
- [ ] Update the title to the correct ticket id should I think it should be TINY-8639 everywhere

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
lnewson: Is there any reason it can't just be in `Schema.ts`? That seems to make the most sense to me. It could either be an API on schema or just a floating function outside the schema that's not exposed on the public API.

modules/tinymce/src/core/test/ts/browser/fmt/FormatEmptyLineTest.ts
```diff
@@ -1,12 +1,17 @@
...
+import { Type, Unicode } from '@ephox/katamari';
+import { TinyAssertions, TinyContentActions, TinyHooks, TinySelections, TinyUiActions } from '@ephox/wrap-mcagar';
 
 import Editor from 'tinymce/core/api/Editor';
 
 interface TestConfig {
   readonly selector: string;
   readonly selectorCount: number;
   readonly html: string;
+  readonly contentFormatRaw?: boolean;
```
lnewson, line 12: Instead of making this a boolean maybe just make this a string/`Content` type? That way it's clearer what it's used for in the test. e.g: `contentFormat: Content` and then you'd do `contentFormat: 'raw'` 🤷

lnewson: I think the other thing that's confusing me is that the content is passed via `html` here and then this uses `content` so they don't appear to be related. So another option would be to call this `rawHtml` instead which would probably be clearer than my previous suggestion.

modules/tinymce/src/core/test/ts/browser/html/DomParserTest.ts
```diff
@@ -1065,6 +1065,127 @@ describe('browser.tinymce.core.html.DomParserTest', () => {
     );
   });
 
+  it('TINY-8639: handling empty text inline elements when root block is empty', () => {
```
lnewson, line 1068: Do we need any tests for what the default `padd_empty_block_inline_children` value is here (i.e. false)? or is that covered by existing tests?

lnewson (approved):

modules/tinymce/src/core/test/ts/browser/html/DomParserTest.ts
ltrouton: I have added cases for when the option isn't specified. Thanks

modules/tinymce/src/core/test/ts/browser/fmt/FormatEmptyLineTest.ts
ltrouton: Renamed property to `rawHtml`

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
ltrouton: I have made the change. Let me know if this is along the lines of what you were thinking

HAFRMO (approved):

PR #7836, opened 2022-05-18
TINY-6800: copy event should be dispatched also in readonly mode

lorenzo-pomili:
Related Ticket: TINY-6800

Description of Changes:
  * [x] At the moment is possible to copy content but the event is not dispatched, to solve it I added a function named `isReadOnlyAllowedEvent` that checks if an event should be allowed in `readonly` mode.
  * [x] I change `fireEvent` function in observable splitting the 2 cases:
         * !hidde && (!readonly || eventAllowedInReadOnly) -> dispatch event
         * readonly -> processReadonlyEvents

I split it so that the events allowed in `readonly` mode have the same behavior as the others

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -11,7 +11,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Getting the text content of the editor now returns newlines instead of an empty string if more than one empty paragraph exists #TINY-8578
 - The `end_container_on_empty_block` option can now take a string of blocks to split when pressing Enter twice #TINY-6559
 - The default value for `end_container_on_empty_block` option has been changed to `'blockquote'` #TINY-6559
-- Copy events were not dispatched in readonly mode #TINY-6800
 
 ### Fixed
 - Links would open when using alt+enter (option+enter on Mac) even when `preventDefault()` is called on the keydown event #TINY-8661
@@ -32,6 +31,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Indenting or outdenting list items inside a block element that was inside another list item would not work #TINY-7209
 - Switching between unordered and ordered lists would incorrectly alter any parent element that contained that list #TINY-8068
 - The URL detection used for `autolink` and smart paste didn't work if a path segment contained valid characters such as `!` and `:` #TINY-8069
+- Copy events were not dispatched in readonly mode #TINY-6800
 
 ## 6.0.3 - TBD
 
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Getting the text content of the editor now returns newlines instead of an empty string if more than one empty paragraph exists #TINY-8578
 - The `end_container_on_empty_block` option can now take a string of blocks to split when pressing Enter twice #TINY-6559
 - The default value for `end_container_on_empty_block` option has been changed to `'blockquote'` #TINY-6559
+- Copy events were not dispatched in readonly mode #TINY-6800
 
 ### Fixed
 - Links would open when using alt+enter (option+enter on Mac) even when `preventDefault()` is called on the keydown event #TINY-8661
@@ -37,7 +38,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Could not remove values when multiple cells were selected with the cell properties dialog #TINY-8625
 - Could not remove values when multiple rows were selected with the row properties dialog #TINY-8625
-- Copy events are not dispatched in readonly mode #TINY-6800
 
 ## 6.0.2 - 2022-04-27
 
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -37,6 +37,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Could not remove values when multiple cells were selected with the cell properties dialog #TINY-8625
 - Could not remove values when multiple rows were selected with the row properties dialog #TINY-8625
+- Copy events are not dispatched in readonly mode #TINY-6800
 
 ## 6.0.2 - 2022-04-27
 
--- a/modules/tinymce/src/core/main/ts/api/EditorObservable.ts
+++ b/modules/tinymce/src/core/main/ts/api/EditorObservable.ts
@@ -54,16 +54,14 @@ const getEventTarget = (editor: Editor, eventName: string): Node => {
   return editor.getBody();
 };
 
-const isListening = (editor: Editor) => !editor.hidden && !isReadOnly(editor);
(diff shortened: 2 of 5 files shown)


modules/tinymce/src/core/main/ts/api/EditorObservable.ts
```diff
@@ -54,12 +54,13 @@ const getEventTarget = (editor: Editor, eventName: string): Node => {
   return editor.getBody();
 };
 
-const isListening = (editor: Editor) => !editor.hidden && !isReadOnly(editor);
+const isListening = (editor: Editor, e: Event) => !editor.hidden && (!isReadOnly(editor) || isReadOnlyAllowedEvent(e));
```
lnewson, line 57: This is confusing/mixing the logic here, as. it's meant to be divided into listening to any event and then processing when in readonly mode. As such, this would be better handled by emitting the event within the `processReadonlyEvents` function instead if it's in the whitelist, as its very purpose is to decide what to do with events while in readonly mode.

- [ ] Revert these changes and move this logic into `processReadonlyEvents` please

modules/tinymce/src/core/main/ts/api/EditorObservable.ts
```diff
@@ -54,12 +54,13 @@ const getEventTarget = (editor: Editor, eventName: string): Node => {
...
 
 const fireEvent = (editor: Editor, eventName: string, e: Event) => {
-  if (isListening(editor)) {
+  if (isListening(editor, e)) {
     editor.dispatch(eventName, e);
-  } else if (isReadOnly(editor)) {
+  }
+  if (isReadOnly(editor)) {
```
lnewson, line 63: This shouldn't have been changed, as it this meant to be one or the other and never both.

modules/tinymce/src/core/main/ts/mode/Readonly.ts
```diff
@@ -116,6 +116,9 @@ const registerReadOnlyContentFilters = (editor: Editor) => {
 
 const isClickEvent = (e: Event): e is MouseEvent => e.type === 'click';
 
+const allowedEvents: ReadonlyArray<string> = [ 'copy' ];
+
+const isReadOnlyAllowedEvent = (e: Event) => allowedEvents.includes(e.type);
```
lnewson, line 121: This should be using our Katamari APIs instead, so in this case `Arr.contains`.

Additionally there should be a blank line between functions.

modules/tinymce/src/core/test/ts/browser/ModeTest.ts
```diff
@@ -108,4 +109,14 @@ describe('browser.tinymce.core.ModeTest', () => {
     assertMode(editor, 'readonly');
     assertBodyClass(editor, 'mce-content-readonly', true);
   });
+
+  it('TINY-6800: even in readonly mode copy event should be dispatched', () => {
```
lnewson, line 113: This would be better suited in `ReadOnlyModeTest` instead of here.

modules/tinymce/src/core/test/ts/browser/ModeTest.ts
```diff
@@ -108,4 +109,14 @@ describe('browser.tinymce.core.ModeTest', () => {
...
+    const editor = hook.editor();
+
+    let copyEventCount = 0;
+    editor.on('copy', () => copyEventCount++);
```
lnewson, line 117: This isn't unbound at the end of the test and we should be cleaning up after ourselves.

modules/tinymce/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Could not remove values when multiple cells were selected with the cell properties dialog #TINY-8625
 - Could not remove values when multiple rows were selected with the row properties dialog #TINY-8625
+- Copy events are not dispatched in readonly mode #TINY-6800
```
lnewson, line 12: This is the 6.0.3 changelog... I believe you've branched this from the wrong branch. Please make sure to rebase this on top of `develop` instead of `master` and move this to the appropriate place.

Also here's a minor suggestion for the wording (entries should use past tense):
```suggestion
- Copy events were not dispatched in readonly mode #TINY-6800
```

lnewson (changes requested):

modules/tinymce/CHANGELOG.md
lorenzo-pomili: rebased and fixed

modules/tinymce/src/core/test/ts/browser/ModeTest.ts
lorenzo-pomili: done

modules/tinymce/src/core/test/ts/browser/ModeTest.ts
lorenzo-pomili: moved

modules/tinymce/src/core/main/ts/mode/Readonly.ts
lorenzo-pomili: done

modules/tinymce/src/core/main/ts/api/EditorObservable.ts
lorenzo-pomili: reverted sorry, I had interpreted `processReadonlyEvents` as something that changes the read-only events more than something that manages them (and now I figured out that in that case, I should have inverted the 2 ifs :sweat_smile: )

modules/tinymce/src/core/main/ts/mode/Readonly.ts
```diff
@@ -116,6 +116,9 @@ const registerReadOnlyContentFilters = (editor: Editor) => {
 
 const isClickEvent = (e: Event): e is MouseEvent => e.type === 'click';
 
+const allowedEvents: ReadonlyArray<string> = [ 'copy' ];
+
+const isReadOnlyAllowedEvent = (e: Event) => Arr.contains(allowedEvents, e.type);
```
lnewson, line 121:
```suggestion
const isReadOnlyAllowedEvent = (e: Event) => Arr.contains(allowedEvents, e.type);

```
This still needs an empty line to break up the functions 😄

modules/tinymce/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Getting the text content of the editor now returns newlines instead of an empty string if more than one empty paragraph exists #TINY-8578
 - The `end_container_on_empty_block` option can now take a string of blocks to split when pressing Enter twice #TINY-6559
 - The default value for `end_container_on_empty_block` option has been changed to `'blockquote'` #TINY-6559
+- Copy events were not dispatched in readonly mode #TINY-6800
```
lnewson, line 14: This looks to be in the wrong place... this should be in the `Fixed` section given how it's worded.

- [ ] Move to correct section or reword

modules/tinymce/src/core/main/ts/api/EditorObservable.ts
lorenzo-pomili: done

lnewson (approved):
That looks much cleaner, thanks Lorenzo! 👍

ltrouton (approved):

PR #7840, opened 2022-05-20
TINY-8641: Add the `dispatchChange` function to the `undoManager`

lorenzo-pomili:
Related Ticket: TINY-8641

Description of Changes:
Add the `dispatchChange` function that triggers a change event with current editor status as level and current `undoManager` layer as `lastLevel` in the `undoManager` and the current content of the editor

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -58,7 +58,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed incorrect word breaks in menu dropdowns with scrollbar #TINY-8572
 - The `InsertLineBreak` command did not replace selected content #TINY-8458
 - Delete operations could behave incorrectly if the selection contains a `contenteditable="false"` element located at the edge of content #TINY-8729
-- `uploadImages` no longer triggers two `change` events if there is an error on upload #TINY-8641
+- `uploadImages` no longer triggers two `change` events if there is a removal of images on upload #TINY-8641
 
 ## 6.0.3 - 2022-05-25
 
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The default value for `end_container_on_empty_block` option has been changed to `'blockquote'` #TINY-6559
 - Custom elements are now treated as non-empty elements via the schema #TINY-4784
 - The autocompleter menu element is now positioned instead of the wrapper #TINY-6476
+- Now on `uploadImages` does not trigger `change` event 2 times if there is an error on upload #TINY-8641
 
 ### Fixed
 - Selecting all content with a single image in the content was inconsistent for the keyboard shortcut and menu item #TINY-4550
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - New `removeAttributeFilter` and `removeNodeFilter` functions to the `DomParser` and DOM `Serializer` APIs #TINY-7847
 - New `iframe_template_callback` option in the `media` plugin. Patch provided by Namstel #TINY-8684
 - New `editor.undoManager.fireIfChanged` API to fire change event only if the last undoManager content layer is diffrent from the current editor
-- New `dispatchIfChanged` UndoManager API to fire the `change` event if the last UndoManager content layer is different from the current editor content #TINY-8641
+- New `dispatchChange` UndoManager API to fire the `change` with current editor status as level and current undoManager layer as lastLevel #TINY-8641
 
 ### Improved
 - Clearer focus states for buttons while navigating with keyboard #TINY-8557
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - New `transparent` property for `iframe` dialog component #TINY-8534
 - New `removeAttributeFilter` and `removeNodeFilter` functions to the `DomParser` and DOM `Serializer` APIs #TINY-7847
 - New `iframe_template_callback` option in the `media` plugin. Patch provided by Namstel #TINY-8684
+- New `editor.undoManager.fireIfChanged` API to fire change event only if the last undoManager content layer is diffrent from the current editor
 
 ### Improved
 - Clearer focus states for buttons while navigating with keyboard #TINY-8557
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -24,7 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The default value for `end_container_on_empty_block` option has been changed to `'blockquote'` #TINY-6559
 - Custom elements are now treated as non-empty elements via the schema #TINY-4784
 - The autocompleter menu element is now positioned instead of the wrapper #TINY-6476
-- Now on `uploadImages` does not trigger `change` event 2 times if there is an error on upload #TINY-8641
+- `uploadImages` no longer triggers two `change` events if there is an error on upload #TINY-8641
 
 ### Fixed
 - Selecting all content with a single image in the content was inconsistent for the keyboard shortcut and menu item #TINY-4550
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -24,7 +24,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The default value for `end_container_on_empty_block` option has been changed to `'blockquote'` #TINY-6559
 - Custom elements are now treated as non-empty elements via the schema #TINY-4784
 - The autocompleter menu element is now positioned instead of the wrapper #TINY-6476
-- `uploadImages` no longer triggers two `change` events if there is an error on upload #TINY-8641
 
 ### Fixed
 - Selecting all content with a single image in the content was inconsistent for the keyboard shortcut and menu item #TINY-4550
@@ -59,6 +58,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed incorrect word breaks in menu dropdowns with scrollbar #TINY-8572
 - The `InsertLineBreak` command did not replace selected content #TINY-8458
 - Delete operations could behave incorrectly if the selection contains a `contenteditable="false"` element located at the edge of content #TINY-8729
+- `uploadImages` no longer triggers two `change` events if there is an error on upload #TINY-8641
 
 ## 6.0.3 - 2022-05-25
 
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -12,7 +12,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - New `transparent` property for `iframe` dialog component #TINY-8534
 - New `removeAttributeFilter` and `removeNodeFilter` functions to the `DomParser` and DOM `Serializer` APIs #TINY-7847
 - New `iframe_template_callback` option in the `media` plugin. Patch provided by Namstel #TINY-8684
-- New `editor.undoManager.fireIfChanged` API to fire change event only if the last undoManager content layer is diffrent from the current editor
 - New `dispatchChange` UndoManager API to fire the `change` with current editor status as level and current undoManager layer as lastLevel #TINY-8641
 
 ### Improved
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - New `removeAttributeFilter` and `removeNodeFilter` functions to the `DomParser` and DOM `Serializer` APIs #TINY-7847
 - New `iframe_template_callback` option in the `media` plugin. Patch provided by Namstel #TINY-8684
 - New `editor.undoManager.fireIfChanged` API to fire change event only if the last undoManager content layer is diffrent from the current editor
+- New `dispatchIfChanged` UndoManager API to fire the `change` event if the last UndoManager content layer is different from the current editor content #TINY-8641
 
 ### Improved
 - Clearer focus states for buttons while navigating with keyboard #TINY-8557
--- a/modules/tinymce/src/core/main/ts/api/EditorUpload.ts
+++ b/modules/tinymce/src/core/main/ts/api/EditorUpload.ts
@@ -1,4 +1,4 @@
-import { Arr } from '@ephox/katamari';
+import { Arr, Strings } from '@ephox/katamari';
 import { Attribute, SugarElement } from '@ephox/sugar';
 
 import * as ErrorReporter from '../ErrorReporter';
@@ -121,7 +121,7 @@ const EditorUpload = (editor: Editor): EditorUpload => {
           let removed = false;
 
           if (uploadInfo.status && Options.shouldReplaceBlobUris(editor)) {
-            if (uploadInfo.url && !image.src.includes(uploadInfo.url)) {
+            if (uploadInfo.url && !Strings.contains(image.src, uploadInfo.url)) {
               shouldDispatchChange = true;
             }
             blobCache.removeByUri(image.src);
--- a/modules/tinymce/src/core/main/ts/api/EditorUpload.ts
+++ b/modules/tinymce/src/core/main/ts/api/EditorUpload.ts
@@ -149,8 +149,7 @@ const EditorUpload = (editor: Editor): EditorUpload => {
           };
         });
 
-        const isRemovingAnImage = Arr.exists(imagesToRemove, (element) => !!blobCache.getByUri(element.src));
-        if (shouldDispatchChange || isRemovingAnImage) {
+        if (shouldDispatchChange) {
           editor.undoManager.dispatchChange();
         }
 
--- a/modules/tinymce/src/core/main/ts/api/EditorUpload.ts
+++ b/modules/tinymce/src/core/main/ts/api/EditorUpload.ts
@@ -149,7 +149,8 @@ const EditorUpload = (editor: Editor): EditorUpload => {
           };
         });
 
-        if (shouldDispatchChange || Arr.exists(imagesToRemove, (element) => !!blobCache.getByUri(element.src))) {
+        const isRemovingAnImage = Arr.exists(imagesToRemove, (element) => !!blobCache.getByUri(element.src));
+        if (shouldDispatchChange || isRemovingAnImage) {
           editor.undoManager.dispatchChange();
         }
 
--- a/modules/tinymce/src/core/main/ts/api/EditorUpload.ts
+++ b/modules/tinymce/src/core/main/ts/api/EditorUpload.ts
@@ -149,7 +149,7 @@ const EditorUpload = (editor: Editor): EditorUpload => {
           };
         });
 
-        if (shouldDispatchChange) {
+        if (shouldDispatchChange || Arr.exists(imagesToRemove, (element) => !!blobCache.getByUri(element.src))) {
           editor.undoManager.dispatchChange();
         }
 
--- a/modules/tinymce/src/core/main/ts/api/EditorUpload.ts
+++ b/modules/tinymce/src/core/main/ts/api/EditorUpload.ts
@@ -112,8 +112,8 @@ const EditorUpload = (editor: Editor): EditorUpload => {
       const blobInfos = Arr.map(imageInfos, (imageInfo) => imageInfo.blobInfo);
 
       return uploader.upload(blobInfos, openNotification(editor)).then(aliveGuard((result) => {
-        let shouldDispatchChange = false;
         const imagesToRemove: HTMLImageElement[] = [];
+        let shouldDispatchChange = false;
 
         const filteredResult: UploadResult[] = Arr.map(result, (uploadInfo, index) => {
           const blobInfo = imageInfos[index].blobInfo;
@@ -153,9 +153,6 @@ const EditorUpload = (editor: Editor): EditorUpload => {
           editor.undoManager.transact(() => {
             Arr.each(imagesToRemove, (element) => {
               editor.dom.remove(element);
-              if (blobCache.getByUri(element.src)) {
-                shouldDispatchChange = true;
-              }
               blobCache.removeByUri(element.src);
             });
           });
--- a/modules/tinymce/src/core/main/ts/api/EditorUpload.ts
+++ b/modules/tinymce/src/core/main/ts/api/EditorUpload.ts
@@ -149,6 +149,10 @@ const EditorUpload = (editor: Editor): EditorUpload => {
           };
         });
 
+        if (shouldDispatchChange) {
+          editor.undoManager.dispatchChange();
+        }
+
         if (imagesToRemove.length > 0 && !Rtc.isRtc(editor)) {
           editor.undoManager.transact(() => {
             Arr.each(imagesToRemove, (element) => {
@@ -156,10 +160,6 @@ const EditorUpload = (editor: Editor): EditorUpload => {
               blobCache.removeByUri(element.src);
             });
           });
-        }
-
-        if (shouldDispatchChange) {
-          editor.undoManager.dispatchChange();
         }
 
         return filteredResult;
--- a/modules/tinymce/src/core/main/ts/api/EditorUpload.ts
+++ b/modules/tinymce/src/core/main/ts/api/EditorUpload.ts
@@ -149,11 +149,12 @@ const EditorUpload = (editor: Editor): EditorUpload => {
           };
         });
 
-        if (shouldDispatchChange && imagesToRemove.length === 0 && !Rtc.isRtc(editor)) {
+        const shouldRemoveImages = imagesToRemove.length > 0 && !Rtc.isRtc(editor);
+        if (shouldDispatchChange && !shouldRemoveImages) {
           editor.undoManager.dispatchChange();
         }
 
-        if (imagesToRemove.length > 0 && !Rtc.isRtc(editor)) {
+        if (shouldRemoveImages) {
           editor.undoManager.transact(() => {
             Arr.each(imagesToRemove, (element) => {
               editor.dom.remove(element);
--- a/modules/tinymce/src/core/main/ts/api/EditorUpload.ts
+++ b/modules/tinymce/src/core/main/ts/api/EditorUpload.ts
@@ -1,4 +1,4 @@
-import { Arr } from '@ephox/katamari';
+import { Arr, Cell } from '@ephox/katamari';
 import { Attribute, SugarElement } from '@ephox/sugar';
 
 import * as ErrorReporter from '../ErrorReporter';
@@ -6,6 +6,8 @@ import { BlobInfoImagePair, ImageScanner } from '../file/ImageScanner';
 import { Uploader } from '../file/Uploader';
 import { UploadStatus } from '../file/UploadStatus';
 import * as Rtc from '../Rtc';
+import * as Levels from '../undo/Levels';
+import { UndoLevel } from '../undo/UndoManagerTypes';
 import Editor from './Editor';
 import Env from './Env';
 import { BlobCache, BlobInfo } from './file/BlobCache';
@@ -36,11 +38,37 @@ interface EditorUpload {
   destroy: () => void;
 }
 
+const UploadChangeHandler = (editor: Editor) => {
+  const lastChangedLevel = Cell<UndoLevel>(null);
+
+  editor.on('change AddUndo', (e) => {
+    lastChangedLevel.set({ ...e.level });
+  });
+
+  const fireIfChanged = () => {
+    const data = editor.undoManager.data;
+    Arr.last(data).filter((level) => {
+      return !Levels.isEq(lastChangedLevel.get(), level);
+    }).each((level) => {
+      editor.setDirty(true);
+      editor.dispatch('change', {
+        level,
+        lastLevel: Arr.get(data, data.length - 2).getOrNull()
+      });
+    });
+  };
+
+  return {
+    fireIfChanged
+  };
+};
+
 const EditorUpload = (editor: Editor): EditorUpload => {
   const blobCache = BlobCache();
   let uploader: Uploader, imageScanner: ImageScanner;
   const uploadStatus = UploadStatus();
   const urlFilters: Array<(img: HTMLImageElement) => boolean> = [];
+  const changeHandler = UploadChangeHandler(editor);
 
   const aliveGuard = <T, R> (callback?: (result: T) => R) => {
     return (result: T) => {
@@ -146,7 +174,7 @@ const EditorUpload = (editor: Editor): EditorUpload => {
         });
 
         if (filteredResult.length > 0) {
-          editor.undoManager.fireIfChanged();
+          changeHandler.fireIfChanged();
         }
 
         if (imagesToRemove.length > 0 && !Rtc.isRtc(editor)) {
--- a/modules/tinymce/src/core/main/ts/api/EditorUpload.ts
+++ b/modules/tinymce/src/core/main/ts/api/EditorUpload.ts
@@ -1,4 +1,4 @@
-import { Arr, Cell } from '@ephox/katamari';
+import { Arr } from '@ephox/katamari';
 import { Attribute, SugarElement } from '@ephox/sugar';
 
 import * as ErrorReporter from '../ErrorReporter';
@@ -6,8 +6,6 @@ import { BlobInfoImagePair, ImageScanner } from '../file/ImageScanner';
 import { Uploader } from '../file/Uploader';
 import { UploadStatus } from '../file/UploadStatus';
 import * as Rtc from '../Rtc';
-import * as Levels from '../undo/Levels';
-import { UndoLevel } from '../undo/UndoManagerTypes';
 import Editor from './Editor';
 import Env from './Env';
 import { BlobCache, BlobInfo } from './file/BlobCache';
@@ -38,37 +36,11 @@ interface EditorUpload {
   destroy: () => void;
 }
 
-const UploadChangeHandler = (editor: Editor) => {
-  const lastChangedLevel = Cell<UndoLevel>(null);
-
-  editor.on('change AddUndo', (e) => {
-    lastChangedLevel.set({ ...e.level });
-  });
-
-  const fireIfChanged = () => {
-    const data = editor.undoManager.data;
-    Arr.last(data).filter((level) => {
-      return !Levels.isEq(lastChangedLevel.get(), level);
-    }).each((level) => {
-      editor.setDirty(true);
-      editor.dispatch('change', {
-        level,
-        lastLevel: Arr.get(data, data.length - 2).getOrNull()
-      });
-    });
-  };
-
-  return {
-    fireIfChanged
-  };
-};
-
 const EditorUpload = (editor: Editor): EditorUpload => {
   const blobCache = BlobCache();
   let uploader: Uploader, imageScanner: ImageScanner;
   const uploadStatus = UploadStatus();
   const urlFilters: Array<(img: HTMLImageElement) => boolean> = [];
-  const changeHandler = UploadChangeHandler(editor);
 
   const aliveGuard = <T, R> (callback?: (result: T) => R) => {
     return (result: T) => {
@@ -140,6 +112,7 @@ const EditorUpload = (editor: Editor): EditorUpload => {
       const blobInfos = Arr.map(imageInfos, (imageInfo) => imageInfo.blobInfo);
 
       return uploader.upload(blobInfos, openNotification(editor)).then(aliveGuard((result) => {
+        let shouldDispatchChange = false;
         const imagesToRemove: HTMLImageElement[] = [];
 
         const filteredResult: UploadResult[] = Arr.map(result, (uploadInfo, index) => {
@@ -148,6 +121,9 @@ const EditorUpload = (editor: Editor): EditorUpload => {
           let removed = false;
 
           if (uploadInfo.status && Options.shouldReplaceBlobUris(editor)) {
+            if (uploadInfo.url && !image.src.includes(uploadInfo.url)) {
+              shouldDispatchChange = true;
+            }
             blobCache.removeByUri(image.src);
             if (Rtc.isRtc(editor)) {
               // RTC handles replacing the image URL through callback events
@@ -173,17 +149,20 @@ const EditorUpload = (editor: Editor): EditorUpload => {
           };
         });
 
-        if (filteredResult.length > 0) {
-          changeHandler.fireIfChanged();
-        }
-
         if (imagesToRemove.length > 0 && !Rtc.isRtc(editor)) {
           editor.undoManager.transact(() => {
             Arr.each(imagesToRemove, (element) => {
               editor.dom.remove(element);
+              if (blobCache.getByUri(element.src)) {
+                shouldDispatchChange = true;
+              }
               blobCache.removeByUri(element.src);
             });
           });
+        }
+
+        if (shouldDispatchChange) {
+          editor.undoManager.dispatchChange();
         }
 
         return filteredResult;
--- a/modules/tinymce/src/core/main/ts/api/EditorUpload.ts
+++ b/modules/tinymce/src/core/main/ts/api/EditorUpload.ts
@@ -149,7 +149,7 @@ const EditorUpload = (editor: Editor): EditorUpload => {
           };
         });
 
-        if (shouldDispatchChange) {
+        if (shouldDispatchChange && imagesToRemove.length === 0 && !Rtc.isRtc(editor)) {
           editor.undoManager.dispatchChange();
         }
 
--- a/modules/tinymce/src/core/main/ts/api/EditorUpload.ts
+++ b/modules/tinymce/src/core/main/ts/api/EditorUpload.ts
@@ -1,4 +1,4 @@
-import { Arr, Cell } from '@ephox/katamari';
+import { Arr } from '@ephox/katamari';
 import { Attribute, SugarElement } from '@ephox/sugar';
 
 import * as ErrorReporter from '../ErrorReporter';
@@ -6,8 +6,6 @@ import { BlobInfoImagePair, ImageScanner } from '../file/ImageScanner';
 import { Uploader } from '../file/Uploader';
 import { UploadStatus } from '../file/UploadStatus';
 import * as Rtc from '../Rtc';
-import * as Levels from '../undo/Levels';
-import { UndoLevel } from '../undo/UndoManagerTypes';
 import Editor from './Editor';
 import Env from './Env';
 import { BlobCache, BlobInfo } from './file/BlobCache';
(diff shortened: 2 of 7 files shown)


modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
```diff
@@ -611,4 +611,67 @@ describe('browser.tinymce.core.UndoManagerTest', () => {
...
+    let changeEventCounter = 0;
+    let currentChangeEvent;
+
+    const initialContent = '<p>some inital content</p>';
+    const manualModifiedLevel = 'a modified last level';
+
+    const fireAndCheckEvent = (event, level, lastLevel) => assert.isTrue(
+      itTheExpectedChangeEvent(event, level, lastLevel),
+      'change event should have last level as the current content of the editor and lastLevel as the last level in undoManager'
+    );
```
spocke, line 629: The function doesn't fire the event it asserts properties on it. Also you could instead have separate assertments for the different properties. That way a better error message with a diff would be displayed. Also it seems that all the assertions send in the same content so why not have it hard coded in the function. Suggested code below should obviously also have types.

```suggestion
    const assertChangeEvent = (event) => {
      assert.equal(e.level, manualModifiedLevel, 'Level has not the expected content');
      assert.equal(e.lastLevel, initialLevel, 'Last level has not the expected content');
    };
```

modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
```diff
@@ -611,4 +611,67 @@ describe('browser.tinymce.core.UndoManagerTest', () => {
...
+
+    assert.equal(changeEventCounter, 0, '0 event should be detected at start');
+    editor.undoManager.fireIfChanged();
+    assert.equal(changeEventCounter, 0, '0 event should be detected if content is coherent with the undoManager history');
+
+    Arr.last(editor.undoManager.data).each((lastLevel) => {
+      lastLevel.content = manualModifiedLevel;
+    });
+
+    assert.equal(editor.isDirty(), false, 'Editor should be not dirty before fireIfChanged');
```
spocke, line 647:
```suggestion
    assert.equal(editor.isDirty(), false, 'Editor should not be dirty before fireIfChanged');
```

modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
```diff
@@ -611,4 +611,67 @@ describe('browser.tinymce.core.UndoManagerTest', () => {
...
+
+    editor.undoManager.fireIfChanged();
+    fireAndCheckEvent(currentChangeEvent, initialContent, manualModifiedLevel);
+
+    editor.undoManager.fireIfChanged();
+    fireAndCheckEvent(currentChangeEvent, initialContent, manualModifiedLevel);
+
+    editor.undoManager.fireIfChanged();
+    fireAndCheckEvent(currentChangeEvent, initialContent, manualModifiedLevel);
+    assert.equal(changeEventCounter, 4, 'it should countinue to call change till the editor and last level are different');
```
spocke, line 661:
```suggestion
    assert.equal(changeEventCounter, 4, 'it should continue to call change till the editor and last level are different');
```

spocke (changes requested):

modules/tinymce/src/core/main/ts/api/UndoManager.ts
```diff
@@ -50,6 +51,30 @@ const UndoManager = (editor: Editor): UndoManager => {
...
+     *
+     * @method fireIfChanged
+     */
+    fireIfChanged: (): void => {
+      const data = undoManager.data;
+      const currentLevel = Levels.createFromEditor(editor);
+
+      Arr.last(data).filter((lastLevel) => {
+        return !Levels.isEq(lastLevel, currentLevel);
+      }).fold(
```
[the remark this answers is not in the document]
lorenzo-pomili: changed

modules/tinymce/src/core/main/ts/api/UndoManager.ts
```diff
@@ -50,6 +51,30 @@ const UndoManager = (editor: Editor): UndoManager => {
       return Rtc.addUndoLevel(editor, undoManager, index, locks, beforeBookmark, level, event);
     },
 
+    /**
+     * Trigger change event if the content in last layer of the undoManager and the current editor content are different
```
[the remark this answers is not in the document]
lorenzo-pomili: changed

modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
```diff
@@ -611,4 +611,67 @@ describe('browser.tinymce.core.UndoManagerTest', () => {
 
     assert.equal(changeEventCounter, 0, 'No events should be detected');
   });
+
+  it('TINY-8641: fire change and set editor dirty if the current content and the undoManager history are incoherent', () => {
+    const itTheExpectedChangeEvent = (e, level, lastLevel): boolean => e && e.level === level && e.lastLevel === lastLevel;
```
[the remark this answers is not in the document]
lorenzo-pomili: I changed as you suggested here: https://github.com/tinymce/tinymce/pull/7840#discussion_r878179286

modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
```diff
@@ -611,4 +611,67 @@ describe('browser.tinymce.core.UndoManagerTest', () => {
...
+
+    const editor = hook.editor();
+
+    let changeEventCounter = 0;
+    let currentChangeEvent;
```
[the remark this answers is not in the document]
lorenzo-pomili: done

modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
```diff
@@ -611,4 +611,67 @@ describe('browser.tinymce.core.UndoManagerTest', () => {
...
+
+    const initialContent = '<p>some inital content</p>';
+    const manualModifiedLevel = 'a modified last level';
+
+    const fireAndCheckEvent = (event, level, lastLevel) => assert.isTrue(
```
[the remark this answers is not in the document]
lorenzo-pomili: I changed as you suggested here: https://github.com/tinymce/tinymce/pull/7840#discussion_r878179286

modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
```diff
@@ -611,4 +611,67 @@ describe('browser.tinymce.core.UndoManagerTest', () => {
...
+      itTheExpectedChangeEvent(event, level, lastLevel),
+      'change event should have last level as the current content of the editor and lastLevel as the last level in undoManager'
+    );
+
+    editor.resetContent(initialContent);
+    const onChange = (e) => {
```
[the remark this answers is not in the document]
lorenzo-pomili: I did it, but it may be a better type that fit this case

modules/tinymce/src/core/main/ts/api/UndoManager.ts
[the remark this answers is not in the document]
lnewson: Looks like the API is still `fireIfChanged` here? It should also be changed to `dispatchIfChanged` to match that `fire` was renamed to `dispatch`.

modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
lorenzo-pomili: done

modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
lorenzo-pomili: done

modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
lorenzo-pomili: done

modules/tinymce/src/core/main/ts/api/UndoManager.ts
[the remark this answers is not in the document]
lorenzo-pomili: ok, sorry, now I changed it

vpyshnenko (approved):

modules/tinymce/CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+- New `editor.undoManager.fireIfChanged` API to fire change event only if the last undoManager content layer is diffrent from the current editor content #TINY-8641
```
```suggestion
- New `dispatchIfChanged` UndoManager API to fire the `change` event if the last UndoManager content layer is different from the current editor content #TINY-8641
```
This still mentioned `fireIfChanged`. I've also reworded it a little.

modules/tinymce/src/core/main/ts/api/UndoManager.ts
```diff
@@ -50,6 +51,27 @@ const UndoManager = (editor: Editor): UndoManager => {
       return Rtc.addUndoLevel(editor, undoManager, index, locks, beforeBookmark, level, event);
     },
 
+    /**
+     * Dispatch a change event if the content in last layer of the undoManager and the current editor content are different
+     *
+     * @method dispatchIfChanged
+     */
+    dispatchIfChanged: (): void => {
```
lnewson, line 59: We need to remove the `EditorUpload` logic and make it use this new API as well.

- [ ] Replace old editor upload logic

modules/tinymce/src/core/main/ts/api/UndoManager.ts
```diff
@@ -50,6 +51,27 @@ const UndoManager = (editor: Editor): UndoManager => {
...
+      Arr.last(data).filter((lastLevel) => {
+        return !Levels.isEq(lastLevel, currentLevel);
+      }).each((lastLevel) => {
+        editor.setDirty(true);
+        editor.dispatch('change', {
+          level: currentLevel,
+          lastLevel
+        });
+      });
+
```
lnewson, line 72:
```suggestion
```

modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
```diff
@@ -611,4 +612,68 @@ describe('browser.tinymce.core.UndoManagerTest', () => {
...
+    assert.equal(editor.isDirty(), true, 'Editor should be dirty after dispatchIfChanged');
+
+    editor.undoManager.dispatchIfChanged();
+    assertChangeEvent(currentChangeEvent);
+
+    editor.undoManager.dispatchIfChanged();
+    assertChangeEvent(currentChangeEvent);
+
+    editor.undoManager.dispatchIfChanged();
+    assertChangeEvent(currentChangeEvent);
```
lnewson, line 662:
```suggestion
    Arr.range(3, () => {
      editor.undoManager.dispatchIfChanged();
      assertChangeEvent(currentChangeEvent);
    });
```
No need to manually repeat this 3 times as we can use a loop to do that. This also helps to make it clearer about how many times it has run.

modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
```diff
@@ -611,4 +612,68 @@ describe('browser.tinymce.core.UndoManagerTest', () => {
...
+    editor.undoManager.add();
+    assert.equal(changeEventCounter, 5, 'add should trigger a change as well');
+
+    Arr.last(editor.undoManager.data).each((lastLevel) => {
+      assert.equal(editor.getContent(), lastLevel.content, 'add should add a last layer with the editor content');
+    });
+
+    editor.undoManager.dispatchIfChanged();
+    editor.undoManager.dispatchIfChanged();
+    editor.undoManager.dispatchIfChanged();
```
lnewson, line 674:
```suggestion
    Arr.range(3, () => {
      editor.undoManager.dispatchIfChanged();
    });
```

modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
```diff
@@ -611,4 +612,68 @@ describe('browser.tinymce.core.UndoManagerTest', () => {
...
+    assertChangeEvent(currentChangeEvent);
+    assert.equal(changeEventCounter, 4, 'it should continue to call change till the editor and last level are different');
+
+    editor.undoManager.add();
+    assert.equal(changeEventCounter, 5, 'add should trigger a change as well');
+
+    Arr.last(editor.undoManager.data).each((lastLevel) => {
+      assert.equal(editor.getContent(), lastLevel.content, 'add should add a last layer with the editor content');
+    });
```
lnewson, line 670: This assertion isn't going to be working as you might expect... If there is no last level then it's not going to fail. Generally in tests you want to assert the length of the array as well before accessing the item.

- [ ] Improve this assertion to fail if there is no `data`

modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
```diff
@@ -611,4 +612,68 @@ describe('browser.tinymce.core.UndoManagerTest', () => {
...
+    let changeEventCounter = 0;
+    let currentChangeEvent: { level: string; lastLevel: string };
+
+    const initialContent = '<p>some inital content</p>';
+    const manualModifiedLevel = 'a modified last level';
+
+    const assertChangeEvent = (event: { level: string; lastLevel: string }) => {
+      assert.equal(event.level, initialContent, 'Level has not the expected content');
+      assert.equal(event.lastLevel, manualModifiedLevel, 'Last level has not the expected content');
+    };
```
lnewson, line 628: Can we please put this outside of the test as we don't normally do inline assert functions like this for tests? For example at the top of this before any tests as thats where we normally put helper functions. It does mean you'll need to pass in the initial and modified content as variables though.

modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
```diff
@@ -611,4 +612,68 @@ describe('browser.tinymce.core.UndoManagerTest', () => {
...
+      level: UndoLevel;
+      lastLevel: UndoLevel;
+    }>) => {
+      changeEventCounter++;
+      currentChangeEvent = { level: e.level.content, lastLevel: e.lastLevel.content };
+    };
+
+    editor.on('change', onChange);
+
+    assert.equal(changeEventCounter, 0, '0 event should be detected at start');
```
lnewson, line 641:
```suggestion
    assert.equal(changeEventCounter, 0, 'no events should be detected at start');
```

modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
```diff
@@ -611,4 +612,68 @@ describe('browser.tinymce.core.UndoManagerTest', () => {
...
+    editor.undoManager.dispatchIfChanged();
+    assert.equal(changeEventCounter, 0, '0 event should be detected if content is coherent with the undoManager history');
```
lnewson, line 643:
```suggestion
    assert.equal(changeEventCounter, 0, 'no events should be detected if content is consistent with the undoManager history');
```
I think this is what you want here?

modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
```diff
@@ -611,4 +612,68 @@ describe('browser.tinymce.core.UndoManagerTest', () => {
...
+
+    Arr.last(editor.undoManager.data).each((lastLevel) => {
+      lastLevel.content = manualModifiedLevel;
+    });
+
+    assert.equal(editor.isDirty(), false, 'Editor should not be dirty before dispatchIfChanged');
```
lnewson, line 649: Please use the `assert.isFalse` function instead for these type of assertions. There's also `assert.isTrue` for the cases below where you do a truthy check instead.

- [ ] Use the `isTrue` or `isFalse` assertions instead of manually doing it via `equal`

modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
```diff
@@ -611,4 +612,68 @@ describe('browser.tinymce.core.UndoManagerTest', () => {
...
+    editor.undoManager.dispatchIfChanged();
+    assertChangeEvent(currentChangeEvent);
+    assert.equal(changeEventCounter, 1, '1 event should be detected if the content is not coherent with the undoManager history');
```
lnewson, line 652:
```suggestion
    assert.equal(changeEventCounter, 1, '1 event should be detected if the content is not consistent with the undoManager history');
```

modules/tinymce/src/core/main/ts/api/UndoManager.ts
```diff
@@ -50,6 +51,27 @@ const UndoManager = (editor: Editor): UndoManager => {
...
+    /**
+     * Dispatch a change event if the content in last layer of the undoManager and the current editor content are different
+     *
+     * @method dispatchIfChanged
+     */
+    dispatchIfChanged: (): void => {
+      const data = undoManager.data;
+      const currentLevel = Levels.createFromEditor(editor);
+
+      Arr.last(data).filter((lastLevel) => {
```
lnewson, line 63: This doesn't account for the `index` property which says which undo level we should be looking at currently (which can move backwards or forwards as you undo/redo changes). This probably needs to use similar logic to `addUndoLevel` from `Operations.ts`, e.g:
```
const lastLevel = undoManager.data[index.get()];
```

Another issue I see here now looking at this, is that it won't work if the `lastLevel` is `undefined`. It probably still should be firing in that case, so we should add a test for that as well.

lnewson (changes requested):
Looking pretty good Lorenzo! Just a couple of issues with the implementation and a couple things that have been missed.

Also don't forget to update the PR title/description to match the actual API name.

modules/tinymce/src/core/main/ts/api/UndoManager.ts
spocke: That was well spotted.

- [ ] Add a test that does a change then undo then run this then it should not fire a change event
- [ ] Add a test that clears all undo levels then it should always fire I guess since we have nothing to compare against

modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
spocke: Maybe it should be wrapped in a context and the function gets moved there. Since there will likely be more than one test here. I'm also thinking this might need to be split up it's testing a few different things if they where separate tests you would be able to describe what it's testing in each test.

Like `it('dispatchIfChanged should always fire on empty stack', ...)` or `it('dispatchIfChanged should compare with the level at the current undo index', ...)` for example.

spocke (changes requested):

modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
lorenzo-pomili: done, I also replaced the terms `coherent` and `incoherent`, in those 2 places: [1](https://github.com/tinymce/tinymce/pull/7840/files#diff-a02245522e6da74182ef7dede5d329ddcce06854c05146e4d4abc9194c4d4ea8R645), [2](https://github.com/tinymce/tinymce/pull/7840/files#diff-a02245522e6da74182ef7dede5d329ddcce06854c05146e4d4abc9194c4d4ea8R654)

modules/tinymce/src/core/main/ts/api/UndoManager.ts
lorenzo-pomili: done, instead of `data[index.get()]`, I used `Arr.get` I hope is ok

modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
lorenzo-pomili: done

modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
lorenzo-pomili: done

modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
lorenzo-pomili: done

modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
lorenzo-pomili: ok, I put a context and extracted the helper functions and variables

modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
lorenzo-pomili: done

modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
lorenzo-pomili: done

modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
lorenzo-pomili: done

modules/tinymce/src/core/main/ts/api/UndoManager.ts
lorenzo-pomili: done

modules/tinymce/CHANGELOG.md

modules/tinymce/src/core/main/ts/api/UndoManager.ts
lorenzo-pomili: ok, I opened a thread

modules/tinymce/src/core/main/ts/api/UndoManager.ts
```diff
@@ -50,6 +51,32 @@ const UndoManager = (editor: Editor): UndoManager => {
...
+      const dispatchEvent = (lastLevel: UndoLevel) => {
+        editor.setDirty(true);
+        editor.dispatch('change', {
+          level: currentLevel,
+          lastLevel
+        });
+      };
+
+      Arr.get(data, index.get()).fold(
+        () => dispatchEvent(currentLevel),
```
lnewson, line 72: This implementation is incorrect. When there is no previous undo level the `lastLevel` should be `undefined` instead. You'll see that's what the old logic did (technically `null` though that was incorrect) and also what `Operations.ts` does as well.

lorenzo-pomili: fixed, sorry

modules/tinymce/src/core/main/ts/api/UndoManager.ts
lorenzo-pomili: ok, I simplified the API and moved the check in `EditorUpload`

modules/tinymce/src/core/main/ts/api/EditorUpload.ts
```diff
@@ -173,19 +149,22 @@ const EditorUpload = (editor: Editor): EditorUpload => {
...
 
-        if (filteredResult.length > 0) {
-          changeHandler.fireIfChanged();
-        }
-
         if (imagesToRemove.length > 0 && !Rtc.isRtc(editor)) {
           editor.undoManager.transact(() => {
             Arr.each(imagesToRemove, (element) => {
               editor.dom.remove(element);
+              if (blobCache.getByUri(element.src)) {
```
vpyshnenko, line 156: `undoManager.transact` should care about dispatching `change` events on its own

spocke: Yeah, I think we should probably just fix the test and not fire the change twice if we remove the image. The whole point of tweaking this code is to avoid change events that are not needed. So lets tweak it some more. I doubt anyone is dependent on this being emitted twice.

lorenzo-pomili: ok, I removed it and fixed the test, now on error we have no more change events triggered

modules/tinymce/src/core/main/ts/api/EditorUpload.ts
```diff
@@ -186,6 +158,10 @@ const EditorUpload = (editor: Editor): EditorUpload => {
           });
         }
 
+        if (shouldDispatchChange) {
```
lnewson, line 161: Why was this moved? This means the `undoManager.transact()` logic runs first which could alter the behaviour for integrators.

modules/tinymce/src/core/main/ts/api/EditorUpload.ts
```diff
@@ -141,13 +113,17 @@ const EditorUpload = (editor: Editor): EditorUpload => {
...
         const imagesToRemove: HTMLImageElement[] = [];
+        let shouldDispatchChange = false;
 
         const filteredResult: UploadResult[] = Arr.map(result, (uploadInfo, index) => {
           const blobInfo = imageInfos[index].blobInfo;
           const image = imageInfos[index].image;
           let removed = false;
 
           if (uploadInfo.status && Options.shouldReplaceBlobUris(editor)) {
+            if (uploadInfo.url && !image.src.includes(uploadInfo.url)) {
```
lnewson, line 124: Please use the Katamari methods instead here, so this should be `!Strings.contains(image.src, uploadInfo.url)`.

modules/tinymce/src/core/main/ts/api/UndoManager.ts
```diff
@@ -50,6 +51,29 @@ const UndoManager = (editor: Editor): UndoManager => {
...
+    /**
+     * Dispatch a change event with current editor status as level and current undoManager layer as lastLevel
+     *
+     * @method dispatchChange
+     */
+    dispatchChange: (): void => {
+      const data = undoManager.data;
+      const currentLevel = Levels.createFromEditor(editor);
+
+      const dispatchEvent = (lastLevel: UndoLevel) => {
```
lnewson, line 63: This type is incorrect, as you're literally passing `undefined` below.

modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
```diff
@@ -611,4 +612,80 @@ describe('browser.tinymce.core.UndoManagerTest', () => {
...
+      assert.equal(event.level, expectedLevel, 'Level has not the expected content');
+      assert.equal(event.lastLevel, expectedLastlevel, 'Last level has not the expected content');
+    };
+
+    let changeEventCounter: number;
+    let currentChangeEvent: { level: string; lastLevel: string };
+
+    const onChange = (e: EditorEvent<{
+      level: UndoLevel;
+      lastLevel: UndoLevel;
```
lnewson, line 634: This type is also missing that `lastLevel` could be `undefined`.

modules/tinymce/src/core/test/ts/browser/EditorUploadTest.ts
```diff
@@ -358,8 +358,8 @@ describe('browser.tinymce.core.EditorUploadTest', () => {
       callCount++;
 
       if (callCount === 2) {
-        // Note: This is 2 as the removal of the image also triggers the addition of an undo level and a change event
-        assertEventsLength(2);
+        // Note: This is 1 as only the removal of the image triggers the addition of an undo level and a change event
```
lnewson, line 361: This is probably a side-effect of you changing the placement of the `change` call and proves that you've changed behaviour which isn't what we want here.

modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
```diff
@@ -611,4 +612,80 @@ describe('browser.tinymce.core.UndoManagerTest', () => {
...
+    const initialContent = '<p>some inital content</p>';
+    const manualModifiedLevel = 'a modified last level';
+
+    const assertChangeEvent = (
+      event: { level: string; lastLevel: string },
+      expectedLevel: string | undefined,
+      expectedLastlevel: string | undefined
+    ) => {
+      assert.equal(event.level, expectedLevel, 'Level has not the expected content');
+      assert.equal(event.lastLevel, expectedLastlevel, 'Last level has not the expected content');
```
lnewson, line 626: These really should be using `deepEqual` as we don't need them to be the same reference, only the same data.

lnewson: I should add if we don't use the level then can we please change the variable names as we aren't passing the level here at the moment, only the content.

lorenzo-pomili: done, I also changed the type of the function since the change event could pass undefined as a `lastLevel`

modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
lorenzo-pomili: done

modules/tinymce/src/core/main/ts/api/UndoManager.ts
lorenzo-pomili: fixed

modules/tinymce/src/core/main/ts/api/EditorUpload.ts
lorenzo-pomili: changed

modules/tinymce/src/core/main/ts/api/EditorUpload.ts
lorenzo-pomili: I changed it back, I moved it in order to try to preserve the old behavior of change events when there is an error, I'll write more details here: https://github.com/tinymce/tinymce/pull/7840#discussion_r895247193

modules/tinymce/src/core/test/ts/browser/EditorUploadTest.ts
lorenzo-pomili: I did a fix that seems to preserve the old behavior but I change it back with this [commit](https://github.com/tinymce/tinymce/pull/7840/commits/df5c497c1985ac4f54aaefc5c2d4bb3c5ff333e9) because of this [comment](https://github.com/tinymce/tinymce/pull/7840#discussion_r894208749).

At the moment the only way to preserve the behavior that I found was my [previous fix](https://github.com/tinymce/tinymce/pull/7840/commits/df5c497c1985ac4f54aaefc5c2d4bb3c5ff333e9#diff-b611e1623b7290a18e6eac5c3465926dae2ac0f148142325b5e2f08ca8636e39L156-L158), I made this [commit](https://github.com/tinymce/tinymce/pull/7840/commits/8e65e8e9725d6edb4284711937ecd20a079b3996) to fix it better than how was in the previous code, but I'd also like the opinion of @spocke since he wrote this https://github.com/tinymce/tinymce/pull/7840#discussion_r894208749

spocke: @lnewson I really think the behavior we have here is wrong that if you upload the same image in parallel and remove the image you get two change events. I would expect to only get one. I doubt anyone if depending on us producing two change events in a row for that particular case.
Switching this code from the current compare with undo levels approach would make that 3 times even more incorrect. So the code needs to be tweaked if we want it to use the new dispatchChange event.
The transact call could be enough to detect the remove image change and fire the change event when it's needed when a new undo level is added.
For uploading images that doesn't get removed then there should be more no new undo level but a change event should be fired so that people can save after an image gets uploaded. At least that is my understanding of this code and problem with changing it.

lnewson: @spocke per the comment, the first change is to notify that the existing undomanager state has changed and the second is related to the additional remove steps which is really a separate process and hence the new undomanager state. So there definitely should be the 2 change events here, as people might need to handle that state change.

lnewson: Also the original logic that was here previously would still have resulted in 2 change events firing because it was this:
```
        if (imagesToRemove.length > 0 && !Rtc.isRtc(editor)) {
            Arr.each(imagesToRemove, (element) => {
              editor.dom.remove(element);
              blobCache.removeByUri(element.src);
            });
          });
        }

        if (shouldDispatchChange) {
          editor.undoManager.dispatchChange();
        }
```
That meant that you'd then get 2 change events for the exact same undo level which definitely wasn't what we wanted.

modules/tinymce/src/core/main/ts/api/UndoManager.ts
```diff
@@ -50,6 +51,29 @@ const UndoManager = (editor: Editor): UndoManager => {
...
+
+      const dispatchEvent = (lastLevel: UndoLevel | undefined) => {
+        editor.setDirty(true);
+        editor.dispatch('change', {
+          level: currentLevel,
+          lastLevel
+        });
+      };
+
+      Arr.get(data, index.get()).fold(
```
lnewson, line 71: This logic could be simplified as we don't need the fold or separate function. eg:
```ts
const lastLevel = Arr.get(data, index.get()).getOrUndefined();
editor.setDirty(true);
editor.dispatch('change', {
  level: currentLevel,
  lastLevel
});
```

lnewson: You could even do `data[index.get()]` since we don't need to do a safe index based lookup here.

lorenzo-pomili: ok, I removed all the variables since these don't give any additional information, the result is:
```JavaScript
editor.setDirty(true);
editor.dispatch('change', {
  level: Levels.createFromEditor(editor),
  lastLevel: Arr.get(undoManager.data, index.get()).getOrUndefined()
});
```

modules/tinymce/src/core/test/ts/browser/EditorUploadTest.ts
spocke: Yes, and that is the thing I think is wrong. If your upload handler removes an image then removes the same image again only one change should be fired. The scanner will find one image, then do an async upload, then remove the image when the first one resolves then tries to remove it again when the second one gets resolved. So shouldDispatchChange could check both if there are no url mutations or node remove mutations then don't fire a change event. It's more that we need to tweak the code to not produce 3 change events then we might as well tweak it to not produce 2. It can't be uploaded and removed at the same time since the upload either passes or fails if it passes then it's an url mutation to change the url to the new one if it fails and has the remove flag then it should remove the image and add an extra undo level since it's a visual change. It would still fire 2 events if you for example upload one image that passed and one that fails and got removed. But I guess we can go all though hoops to get this to work exactly as it did before dispatching additional unnecessary change events if we need to but then I think we might not change this code at all to use the new dispatchChangeEvent since we are afraid of potential behavioral changes and since it's not comparing the undo stack as before it would be a behavioral change in theory if someone has some weird dependency of it having to do that.

lnewson: That's a fair point about `uploadImages` being a single operation that either fails or succeeds. I'm okay with making it only fire the single change here (though as I said the previous logic didn't do that) and the very small risk of breaking that one edge case as long as we put a changelog entry in about it as well.

modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
lnewson: @lorenzo-pomili I can't see these changes, as it's still got the misleading names and isn't passing the full level only the content. Did you forget to push them?

lorenzo-pomili: Sorry, I was sure that I also changed the names of expectation parameters and the name of the keys of the input object. (fixed with the last commits)

But I'm probably still missing the point about the use of `deepEqual` + use directly the `UndoLevel`, in this test the expectations are strings so making `deepEqual` with `UndoLevel` would mean to recreate an `UndoLevel` from the expectation string.

If we want to pass the entire event to the function it could be refactored in this way:
```TypeScript
const assertChangeEvent = (
      event: { level: UndoLevel; lastLevel: UndoLevel | undefined } | undefined,
      expectedLevelContent: string | undefined,
      expectedLastlevelContent: string | undefined
    ) => {
      assert.equal(event?.level?.content, expectedLevelContent, 'Level has not the expected content');
      assert.equal(event?.lastLevel?.content, expectedLastlevelContent, 'Last level has not the expected content');
    };

    let changeEventCounter: number;
    let currentChangeEvent: { level: UndoLevel; lastLevel: UndoLevel | undefined } | undefined;

    const onChange = (e: EditorEvent<{
      level: UndoLevel;
      lastLevel: UndoLevel | undefined;
    }>) => {
      changeEventCounter++;
      currentChangeEvent = e;
    };
```

modules/tinymce/src/core/test/ts/browser/EditorUploadTest.ts
lorenzo-pomili: At the moment the new code behave as the old one, should I left this behavior or should I remove the check that trigger the change and add a change to the changelog that says:
```
- Now on `uploadImages` does not trigger `change` event if there is an error on upload #TINY-8641
```
?

lorenzo-pomili: Sorry I checked again the behaviour, and the 2 events are triggered for the same reason so the correct entry probably would be:
```
- Now on `uploadImages` does not trigger `change` event 2 times if there is an error on upload #TINY-8641
```

modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
```diff
@@ -611,4 +612,82 @@ describe('browser.tinymce.core.UndoManagerTest', () => {
...
+      changeEventCounter = 0;
+      currentChangeEvent = undefined;
+    });
+
+    it('TINY-8641: Dispatch change with current editor status as level and current undoManager layer as lastLevel', () => {
+      const editor = hook.editor();
+
+      editor.resetContent(initialContent);
+
+      editor.on('change', onChange);
```
spocke, line 652: I guess this binding/unbinding could in theory be a beforeEach/afterEach thing since the onChange is a function outside these tests.

spocke (approved):

modules/tinymce/src/core/test/ts/browser/EditorUploadTest.ts
TheSpyder: I believe adding a changelog fix entry is the right way forward. It's become an intentional change, and we need some indication to integrators that they have something watch out for. What you wrote is very close, but I suggest:

```
<under the 'fixed' heading>
- `uploadImages` no longer triggers two `change` events if there is an error on upload #TINY-8641
- ```

**Review by TheSpyder — APPROVED** [review]

File: modules/tinymce/src/core/test/ts/browser/EditorUploadTest.ts (hunk 28 above)
Comment (lorenzo-pomili) [reply modules/tinymce/src/core/test/ts/browser/EditorUploadTest.ts:361]: I added the new entry too

File: modules/tinymce/src/core/main/ts/api/EditorUpload.ts [hunk 32]
```diff
@@ -173,8 +149,8 @@ const EditorUpload = (editor: Editor): EditorUpload => {
           };
         });
 
-        if (filteredResult.length > 0) {
-          changeHandler.fireIfChanged();
+        if (shouldDispatchChange) {
+          editor.undoManager.dispatchChange();
         }
 
         if (imagesToRemove.length > 0 && !Rtc.isRtc(editor)) {
```
lnewson, line 156: This doesn't solve the issue I've repeatedly said here if we're going with the single event logic... this is going to end up with 2 change events when you have both a URL changed and an image that's going to be removed.

modules/tinymce/CHANGELOG.md
```diff
@@ -57,6 +58,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed incorrect word breaks in menu dropdowns with scrollbar #TINY-8572
 - The `InsertLineBreak` command did not replace selected content #TINY-8458
 - Delete operations could behave incorrectly if the selection contains a `contenteditable="false"` element located at the edge of content #TINY-8729
+- `uploadImages` no longer triggers two `change` events if there is an error on upload #TINY-8641
```
lnewson, line 61: This isn't correct... it only triggers if removing images, not on error.

modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
lnewson: Yes, that's the point I was referring to. As for the `deepEqual` part, you're right I've got that a bit mixed up sorry. I instead was thinking of doing a partial deep equal which isn't possible out of the box. Anyways, what you've got now is fine minus a little typo. If you preferred, you really could just go back to using `assert.equal` for each level content sorry.

modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
```diff
@@ -611,4 +612,77 @@ describe('browser.tinymce.core.UndoManagerTest', () => {
...
+      assertChangeEvent(currentChangeEvent, initialContent, manualModifiedLevelContent);
+      assert.equal(changeEventCounter, 1, '1 event should be detected');
+      assert.isTrue(editor.isDirty(), 'Editor should be dirty after dispatchChange');
+
+      Arr.range(3, () => {
+        editor.undoManager.dispatchChange();
+        assertChangeEvent(currentChangeEvent, initialContent, manualModifiedLevelContent);
+      });
+
+      assert.equal(changeEventCounter, 4, 'it should continue to call change');
```
lnewson, line 670: Does this really have any benefit now that we want it to fire at all times anyways? It was useful for when we wanted it to skip, but this doesn't really seem to be testing anything meaningful now.

It looks to me at least that all of this could be condensed down to a single `it` test.

lnewson (changes requested):
Marking as needs changes because of the issue I mentioned about it still firing multiple change events when @spocke had mentioned that's a bug and should only fire once.

modules/tinymce/src/core/main/ts/api/EditorUpload.ts
lnewson: We also should make sure we have a test whereby 2 images are uploaded and one image succeeds in uploading and the other fails with the `remove` flag set. In that case, given what's been said here then we should only have the single `change` event fire.

modules/tinymce/src/core/main/ts/api/EditorUpload.ts
```diff
@@ -173,8 +149,8 @@ const EditorUpload = (editor: Editor): EditorUpload => {
           };
         });
 
-        if (filteredResult.length > 0) {
-          changeHandler.fireIfChanged();
+        if (shouldDispatchChange && imagesToRemove.length === 0 && !Rtc.isRtc(editor)) {
```
lnewson: This logic doesn't quite look right based on the previous logic... this should fire if `shouldDispatchChange` is true and not take into account RTC. So what we probably want is to merge it into the if block below. For example:
```
if (imagesToRemove.length > 0 && !Rtc.isRtc(editor)) {
  ...
} else if (shouldDispatchChange) {
  editor.undoManager.dispatchChange();
}

Comment (lorenzo-pomili) [reply modules/tinymce/src/core/main/ts/api/EditorUpload.ts]: yes, I saw that the logic was not correct and I already changed it [here](https://github.com/tinymce/tinymce/pull/7840/commits/98a7038c85e19badf34c44fa4cd79d16cadf956c), but in this way have more sense I'll refactor it

File: modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts (hunk 34 above)
Comment (lorenzo-pomili) [reply modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts]: ok, I removed the loops but I think that have 2 tests make sense since the second one cover an edge case

File: modules/tinymce/CHANGELOG.md (hunk 33 above)
Comment (lorenzo-pomili) [reply modules/tinymce/CHANGELOG.md]: ok, I changed it in:
```
`uploadImages` no longer triggers two `change` events if there is a removal of images on upload #TINY-8641
```

modules/tinymce/src/core/main/ts/api/EditorUpload.ts
lorenzo-pomili: changed, we can continue this discussion [here](https://github.com/tinymce/tinymce/pull/7840#discussion_r900071060)

modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
lorenzo-pomili: refactored, in the way that I wrote in the previous message

modules/tinymce/src/core/main/ts/api/EditorUpload.ts
lorenzo-pomili: done

lnewson (approved):
Looks good now and good work on the additional tests, thanks @lorenzo-pomili! 👍

PR #7847, opened 2022-05-24
TINY-7433: Fixed values not escaped in generated regular expressions for the template plugin

lnewson:
Related Ticket: TINY-7433

Description of Changes:

Fixes some cases where values provided via options were used (that weren't escaped) when constructing a regular expression leading to them being interpreted as regex syntax. In the end I removed using regex entirely for one case as it wasn't a good use case for regex.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/tinymce/src/plugins/template/test/ts/browser/DatesTest.ts
```diff
@@ -91,4 +91,50 @@ describe('browser.tinymce.plugins.template.DatesTest', () => {
       '</div>'
     ].join('\n'));
   });
+
+  it('TBA: Multiple replacement classes provided via options', async () => {
```
lnewson, line 95: This is intentionally using `TBA` as I just added a test for this existing functionality to make sure my change to `FilterContent` and `Templates` didn't break it as there were no tests.

ltrouton (approved):

HAFRMO (approved):

PR #7850, opened 2022-05-25
TINY-8710: Add sidebar_show option

tiny-ben-tran:
Related Ticket: TINY-8710

Description of Changes:
* Added new option `sidebar_show` to show the specified bar on initialization
* Added  `immediateGrow()` to support showing the specified sidebar without animation
* Removed `Css.reflow` in `doImmediateShrink` as suggested

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/alloy/src/main/ts/ephox/alloy/behaviour/sliding/SlidingApis.ts
+++ b/modules/alloy/src/main/ts/ephox/alloy/behaviour/sliding/SlidingApis.ts
@@ -156,7 +156,6 @@ const toggleGrow = (component: AlloyComponent, slideConfig: SlidingConfig, slide
 const immediateGrow = (component: AlloyComponent, slideConfig: SlidingConfig, slideState: SlidingState): void => {
   if (!slideState.isExpanded()) {
     setGrown(component, slideConfig);
-    // Force current dimension to begin transition
     Css.set(component.element, getDimensionProperty(slideConfig), getDimension(slideConfig, component.element));
     Css.reflow(component.element);
 
--- a/modules/alloy/src/main/ts/ephox/alloy/behaviour/sliding/SlidingApis.ts
+++ b/modules/alloy/src/main/ts/ephox/alloy/behaviour/sliding/SlidingApis.ts
@@ -34,7 +34,7 @@ const doImmediateShrink = (component: AlloyComponent, slideConfig: SlidingConfig
 
   // Force current dimension to begin transition
   Css.set(component.element, getDimensionProperty(slideConfig), getDimension(slideConfig, component.element));
-  Css.reflow(component.element);
+  // TINY-8710: we don't think reflow is required (as has been done elsewhere) as the animation is not needed
 
   disableTransitions(component, slideConfig);
 
--- a/modules/alloy/src/main/ts/ephox/alloy/behaviour/sliding/SlidingApis.ts
+++ b/modules/alloy/src/main/ts/ephox/alloy/behaviour/sliding/SlidingApis.ts
@@ -153,6 +153,15 @@ const toggleGrow = (component: AlloyComponent, slideConfig: SlidingConfig, slide
   f(component, slideConfig, slideState);
 };
 
+const immediateGrow = (component: AlloyComponent, slideConfig: SlidingConfig, slideState: SlidingState): void => {
+  if (!slideState.isExpanded()) {
+    slideState.setExpanded();
+    setGrown(component, slideConfig);
+    slideConfig.onStartGrow(component);
+    slideConfig.onGrown(component);
+  }
+};
+
 export {
   refresh,
   grow,
@@ -164,5 +173,6 @@ export {
   isShrinking,
   isTransitioning,
   toggleGrow,
-  disableTransitions
+  disableTransitions,
+  immediateGrow
 };
--- a/modules/alloy/src/main/ts/ephox/alloy/behaviour/sliding/SlidingApis.ts
+++ b/modules/alloy/src/main/ts/ephox/alloy/behaviour/sliding/SlidingApis.ts
@@ -155,15 +155,12 @@ const toggleGrow = (component: AlloyComponent, slideConfig: SlidingConfig, slide
 
 const immediateGrow = (component: AlloyComponent, slideConfig: SlidingConfig, slideState: SlidingState): void => {
   if (!slideState.isExpanded()) {
+    setGrown(component, slideConfig);
     // Force current dimension to begin transition
     Css.set(component.element, getDimensionProperty(slideConfig), getDimension(slideConfig, component.element));
     Css.reflow(component.element);
 
     disableTransitions(component, slideConfig);
-
-    setGrown(component, slideConfig);
-    // Set the dimension after grow
-    Css.set(component.element, getDimensionProperty(slideConfig), getDimension(slideConfig, component.element));
     slideState.setExpanded();
     slideConfig.onStartGrow(component);
     slideConfig.onGrown(component);
--- a/modules/alloy/src/main/ts/ephox/alloy/behaviour/sliding/SlidingApis.ts
+++ b/modules/alloy/src/main/ts/ephox/alloy/behaviour/sliding/SlidingApis.ts
@@ -157,9 +157,7 @@ const immediateGrow = (component: AlloyComponent, slideConfig: SlidingConfig, sl
   if (!slideState.isExpanded()) {
     setGrown(component, slideConfig);
     Css.set(component.element, getDimensionProperty(slideConfig), getDimension(slideConfig, component.element));
-    Css.reflow(component.element);
-
-    disableTransitions(component, slideConfig);
+    // TINY-8710: we don't think reflow is required (as has been done elsewhere) as the animation is not needed
     slideState.setExpanded();
     slideConfig.onStartGrow(component);
     slideConfig.onGrown(component);
--- a/modules/alloy/src/main/ts/ephox/alloy/behaviour/sliding/SlidingApis.ts
+++ b/modules/alloy/src/main/ts/ephox/alloy/behaviour/sliding/SlidingApis.ts
@@ -155,13 +155,9 @@ const toggleGrow = (component: AlloyComponent, slideConfig: SlidingConfig, slide
 
 const immediateGrow = (component: AlloyComponent, slideConfig: SlidingConfig, slideState: SlidingState): void => {
   if (!slideState.isExpanded()) {
-    // Force current dimension to begin transition
-    Css.set(component.element, getDimensionProperty(slideConfig), getDimension(slideConfig, component.element));
-    Css.reflow(component.element);
-
-    disableTransitions(component, slideConfig);
-
     setGrown(component, slideConfig);
+
+    Css.set(component.element, getDimensionProperty(slideConfig), getDimension(slideConfig, component.element));
     slideState.setExpanded();
     slideConfig.onStartGrow(component);
     slideConfig.onGrown(component);
--- a/modules/alloy/src/main/ts/ephox/alloy/behaviour/sliding/SlidingApis.ts
+++ b/modules/alloy/src/main/ts/ephox/alloy/behaviour/sliding/SlidingApis.ts
@@ -158,6 +158,8 @@ const immediateGrow = (component: AlloyComponent, slideConfig: SlidingConfig, sl
     setGrown(component, slideConfig);
     Css.set(component.element, getDimensionProperty(slideConfig), getDimension(slideConfig, component.element));
     // TINY-8710: we don't think reflow is required (as has been done elsewhere) as the animation is not needed
+    // Keep disableTransition to handle the case where it's part way through transitioning
+    disableTransitions(component, slideConfig);
     slideState.setExpanded();
     slideConfig.onStartGrow(component);
     slideConfig.onGrown(component);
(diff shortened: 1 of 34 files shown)


modules/tinymce/src/themes/silver/main/ts/ui/sidebar/Sidebar.ts
```diff
@@ -94,9 +94,20 @@ const makeSidebar = (panelConfigs: SidebarConfig) => SlotContainer.sketch((parts
...
+
+  optSlider.each((slider) => {
+    Replacing.set(slider, [ makeSidebar(panelConfigs) ]);
+
+    // Show the default sidebar
+    if (Type.isString(showSidebar)) {
+      Composing.getCurrent(slider).each((slotContainer) => {
+        SlotContainer.showSlot(slotContainer, showSidebar);
+      });
+      Sliding.grow(slider);
```
lnewson, line 108: This will still cause the sidebar to animate which we don't want per the Jira requirements. We probably need to implement a `growImmediately` API in the alloy sliding behaviour to solve this.

The other issue is that we probably don't want to grow if there is no slot container found. So we should put this beside the `SlotContainer.showSlot` call instead.

modules/tinymce/src/core/test/ts/browser/options/SidebarShowOptionTest.ts
```diff
@@ -0,0 +1,87 @@
...
+    await Waiter.pTryUntil('Checking initial events', () => store.assertEq('Asserting initial show of sidebars', [
+      {
+        name: 'sidebartwo:show',
+        index: 1
+      },
+    ]));
+    McEditor.remove(editor);
+  });
+
+  it('TINY-8710: Show no side bar if name does not exist', async () => {
```
lnewson, line 78:
```suggestion
  it('TINY-8710: Show no sidebar if the name does not exist', async () => {
```

modules/tinymce/src/core/test/ts/browser/options/SidebarShowOptionTest.ts
```diff
@@ -0,0 +1,87 @@
...
+          tooltip: 'side bar one',
+          icon: 'comment',
+          onShow: logEvent('sidebartwo:show'),
+        });
+      }
+    };
+  };
+
+  it('TINY-8710: Show no sidebar on init if not set', async () => {
+    const store = TestHelpers.TestStore();
```
lnewson, line 40: This probably doesn't need to be initialised for each test. We could init this outside and then just do a `beforeEach` that calls `clear()` to reset the store. That'd just remove some duplication here is all.

modules/tinymce/src/core/test/ts/browser/options/SidebarShowOptionTest.ts
```diff
@@ -0,0 +1,87 @@
...
+    await Waiter.pTryUntil('Checking initial events', () => store.assertEq('Asserting initial show of sidebars', [
+      {
+        name: 'sidebarone:show',
+        index: 0
+      },
+    ]));
+    McEditor.remove(editor);
+  });
+
+  it('TINY-8710: Show just one sidebar on init', async () => {
```
lnewson, line 63: How is this different to the test above? All I can see is that we're initialising with a different value. If we do want to keep this we probably need to improve this description to clarify the differences.

modules/tinymce/src/core/test/ts/browser/options/SidebarShowOptionTest.ts
```diff
@@ -0,0 +1,87 @@
...
+    McEditor.remove(editor);
+  });
+
+  it('TINY-8710: Show a sidebar on init', async () => {
+    const store = TestHelpers.TestStore();
+    const editor = await McEditor.pFromSettings({
+      ...settingsFactory(store),
+      sidebar_show: 'sidebarone'
+    });
+    await Waiter.pTryUntil('Checking initial events', () => store.assertEq('Asserting initial show of sidebars', [
```
lnewson, line 54: We shouldn't need a `pTryUntil` here because all of the logic should happen before the editor initialises/this logic runs.

lnewson (changes requested):
Looks good for a first attempt with dealing with this type of thing 👍  just a couple of things were missed and some minor cleanup is needed 😄

modules/tinymce/src/themes/silver/main/ts/modes/Iframe.ts
```diff
@@ -96,10 +96,10 @@ const render = (editor: Editor, uiComponents: RenderUiComponents, rawUiConfig: R
       outerContainer,
       identifyMenus(editor, rawUiConfig)
     );
-
```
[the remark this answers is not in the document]
tiny-ben-tran: Reverted as it was accidentally removed

modules/tinymce/src/themes/silver/test/ts/browser/sidebar/SidebarShowOptionTest.ts
```diff
@@ -0,0 +1,104 @@
+
```
lnewson, line 1:
```suggestion
```

modules/tinymce/src/themes/silver/test/ts/browser/sidebar/SidebarShowOptionTest.ts
```diff
@@ -0,0 +1,104 @@
...
+import { UiFinder, Waiter } from '@ephox/agar';
+import { TestHelpers } from '@ephox/alloy';
+import { beforeEach, describe, it } from '@ephox/bedrock-client';
+import { Sidebar } from '@ephox/bridge';
+import { SugarBody, SugarElement, Traverse } from '@ephox/sugar';
+import { McEditor } from '@ephox/wrap-mcagar';
+interface EventLog {
```
lnewson, line 8:
```suggestion
import { McEditor } from '@ephox/wrap-mcagar';

interface EventLog {
```

modules/alloy/src/test/ts/browser/behaviour/sliding/SlidingTest.ts
```diff
@@ -236,6 +236,33 @@ UnitTest.asynctest('SlidingTest', (success, failure) => {
...
+        GeneralSteps.sequence([
+          store.sAssertEq('Finished growing', [ 'onStartGrow', 'onGrown' ]),
+          Assertions.sAssertStructure(
+            'Checking structure',
+            ApproxStructure.build((s, str, arr) => s.element('div', {
+              classes: [
+                arr.not('test-sliding-width-shrinking'),
+                arr.not('test-sliding-width-growing'),
+                arr.not('test-sliding-closed'),
+                arr.has('test-sliding-open')
```
lnewson, line 254: We need to assert the width here as well by the looks.

modules/tinymce/src/themes/silver/test/ts/browser/sidebar/SidebarShowOptionTest.ts
```diff
@@ -0,0 +1,92 @@
...
+          onShow: logEvent('sidebarone:show'),
+        });
+        ed.ui.registry.addSidebar('sidebartwo', {
+          tooltip: 'side bar one',
+          icon: 'comment',
+          onShow: logEvent('sidebartwo:show'),
+        });
+      }
+    };
+  };
```
lnewson, line 39: Typo here by the looks? `no` -> `not`

lnewson (approved):
Looks good 👍

modules/alloy/src/main/ts/ephox/alloy/behaviour/sliding/SlidingApis.ts
```diff
@@ -153,6 +153,20 @@ const toggleGrow = (component: AlloyComponent, slideConfig: SlidingConfig, slide
   f(component, slideConfig, slideState);
 };
 
+const immediateGrow = (component: AlloyComponent, slideConfig: SlidingConfig, slideState: SlidingState): void => {
+  if (!slideState.isExpanded()) {
+    setGrown(component, slideConfig);
+    // Force current dimension to begin transition
```
spocke, line 159: This looks a bit odd. We put in in the final state then a few lines below the transition is disabled. Wouldn't that stop the animation directly. The comment here isn't very clear. Is the reflow really needed if the class gets changed in the same animation frame?

modules/tinymce/src/themes/silver/main/ts/ui/sidebar/Sidebar.ts
```diff
@@ -94,9 +94,20 @@ const makeSidebar = (panelConfigs: SidebarConfig) => SlotContainer.sketch((parts
   ])
 }));
 
-const setSidebar = (sidebar: AlloyComponent, panelConfigs: SidebarConfig) => {
+const setSidebar = (sidebar: AlloyComponent, panelConfigs: SidebarConfig, showSidebar?: string) => {
```
spocke, line 97: This showSidebar should either be an option or a `string | undefined` type since having a ? on it makes it an optional parameter and I think that is not the intent here.

spocke (approved):

modules/alloy/src/main/ts/ephox/alloy/behaviour/sliding/SlidingApis.ts
lnewson: This is how it's done elsewhere here so I'm assuming @ephox-mogran had a good reason for that. We definitely need to remove the transition classes for the case where it's already in a partial growing/shrinking state. I'm not so convinced about the need for the reflow though.

As for the comment, good point it's not really accurate here. What we're really doing is setting the final size value and then making sure it doesn't transition to get to that state.

TheSpyder: I too am very curious why the transitions can't be disabled before setting the width. The reflow wouldn't be required at that point.

TheSpyder (approved):
I was hoping this could be done while the sidebar is created, instead of creating it and then setting it (thus needing to deal with the transition stuff)... but this works.

tiny-ben-tran:
> I was hoping this could be done while the sidebar is created, instead of creating it and then setting it (thus needing to deal with the transition stuff)... but this works.

I am not familiar with Alloy yet and still rely on Lee to explain how sidebar, slot container, etc work. As explained from Lee:

The sidebar components aren't rendered until `setSidebar()` is called so it is the earliest possible time we can show it. It is also related to the slot container being used, as that is why the sidebar components aren't rendered until later.  It would require a large rewrite of the sidebar logic to render it initially when the sidebar itself is rendered. 

Also noted that [the spec](https://github.com/tinymce/tinymce/blob/236891ffcdf2fd1b698d020e67c31838fb110019/modules/tinymce/src/themes/silver/main/ts/ui/sidebar/Sidebar.ts#L92-L94) that renders the slot container actually hides all of them when attached to the DOM. So we could have done something to keep the slot container opened, but that still wouldn't have opened that slide. 

We could pass a flag through  [here](https://github.com/tinymce/tinymce/blob/236891ffcdf2fd1b698d020e67c31838fb110019/modules/tinymce/src/themes/silver/main/ts/ui/sidebar/Sidebar.ts#L159-L162) to control whether the sidebar should be rendered expanded or collapsed. However, the sidebar is empty for a period of time when opened.

TheSpyder:
Ah yeah the way plugins render into the sidebar does suggest not showing the area immediately. OK.

modules/alloy/src/main/ts/ephox/alloy/behaviour/sliding/SlidingApis.ts
ephox-mogran: > This is how it's done elsewhere here so I'm assuming @ephox-mogran had a good reason for that.

Trial and error from several years ago, usually.  I'd have to look back through the history. It's possible this was done during apollo as well. Sorry, I can't remember off-hand.

ephox-mogran: Looking at the code with Ben, generally the reason you needed reflow was to trigger an animation, so if this isn't being animated, it may not be necessary. However, the edge cases were always when it was half-way through something.

ephox-mogran: The reflow was usually to force it to register its current size so that it could animate to the next one.

ephox-mogran: IIRC, the reflow was required for situations where the css or classes that had just changed impacted the size of the element and its transition states. So we did a reflow after removing classes or adding classes that would have changed CSS width family values. In this case, because we just want to jump to the immediate size, I don't think any of it should be necessary.

ephox-mogran: Summary. Not sure. Remove it. See what happens.

Probably also the one in immediateShrink. If something is relying on that size being set, it could always call Css.reflow. From looking through history, immediateShrink was added as part of expandable forms from pre-apollo ... so it's been there for ages.

tiny-ben-tran: So I removed the `reflow` as we don't need to trigger an animation and `disableTransition` is also removed as the result. Just noted that it seems to work fine in the manual test, including in the edge case where it is grown from a halfway shrink

lnewson: @tiny-ben-tran that's not what we said to do... per my above comment `disableTransition` is needed and must be kept. In the above we only discussed removing the reflow, so that's all that should have been removed here.

lnewson (changes requested):
Marking as needs changes again because `disableTransition` was removed and that's definitely required to handle the case where it's part way through transitioning.

modules/alloy/src/main/ts/ephox/alloy/behaviour/sliding/SlidingApis.ts
lnewson: Morgan also mentioned above that if we're removing it here then we should also look at removing the one in `immediateShrink`:
> Summary. Not sure. Remove it. See what happens.
> 
> Probably also the one in immediateShrink

lnewson (approved):

PR #7851, opened 2022-05-25
TINY-8040: Copying and pasting columns will no longer generate illegal tables when interacting with colspan

HAFRMO:
Related Ticket:

Description of Changes:
This issue is only concerned with preventing the editor from creating illegal tables.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/snooker/src/main/ts/ephox/snooker/api/CopyCols.ts
+++ b/modules/snooker/src/main/ts/ephox/snooker/api/CopyCols.ts
@@ -4,6 +4,7 @@ import { Attribute, InsertAll, Replication, SugarElement } from '@ephox/sugar';
 import { onUnlockedCells, TargetSelection } from '../model/RunOperation';
 import * as CellUtils from '../util/CellUtils';
 import { CellElement } from '../util/TableTypes';
+import { ColumnExt, DetailExt } from './Structs';
 import { Warehouse } from './Warehouse';
 
 const constrainSpan = (element: SugarElement<CellElement>, property: 'colspan' | 'rowspan' | 'span', value: number) => {
@@ -15,9 +16,16 @@ const constrainSpan = (element: SugarElement<CellElement>, property: 'colspan' |
   }
 };
 
+const isColInRange = (minColRange: number, maxColRange: number) =>
+  (cell: ColumnExt | DetailExt) => {
+    const endCol = cell.column + cell.colspan - 1;
+    const startCol = cell.column;
+    return endCol >= minColRange && startCol < maxColRange;
+  };
+
 const generateColGroup = (house: Warehouse, minColRange: number, maxColRange: number): SugarElement<HTMLTableColElement>[] => {
   if (Warehouse.hasColumns(house)) {
-    const colsToCopy = Arr.filter(Warehouse.justColumns(house), (col) => col.column >= minColRange && col.column < maxColRange);
+    const colsToCopy = Arr.filter(Warehouse.justColumns(house), isColInRange(minColRange, maxColRange));
     const copiedCols = Arr.map(colsToCopy, (c) => {
       const clonedCol = Replication.deep(c.element);
       constrainSpan(clonedCol, 'span', maxColRange - minColRange);
@@ -33,7 +41,7 @@ const generateColGroup = (house: Warehouse, minColRange: number, maxColRange: nu
 
 const generateRows = (house: Warehouse, minColRange: number, maxColRange: number): SugarElement<HTMLTableRowElement>[] =>
   Arr.map(house.all, (row) => {
-    const cellsToCopy = Arr.filter(row.cells, (cell) => (cell.column + cell.colspan > minColRange) && cell.column < maxColRange);
+    const cellsToCopy = Arr.filter(row.cells, isColInRange(minColRange, maxColRange));
     const copiedCells = Arr.map(cellsToCopy, (cell) => {
       const clonedCell = Replication.deep(cell.element);
       constrainSpan(clonedCell, 'colspan', maxColRange - minColRange);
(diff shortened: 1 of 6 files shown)


lnewson (changes requested):
I admittedly haven't looked at this fix properly sorry Hannes but at a glance the changes seem to be reasonable. However, just glancing at it I can see if doesn't meet the DoD in the Jira:
> - Fix the copy column logic to include a cell for the colspan row so that the copied column in the internal clipboard is correct
> 
> - Fix the paste logic to make sure the correct number of cells are added to the colspan row in the table

So of the above, the first requirement has not been met with these changes. Also you're missing changelog entries for Snooker.

HAFRMO:
> I admittedly haven't looked at this fix properly sorry Hannes but at a glance the changes seem to be reasonable. However, just glancing at it I can see if doesn't meet the DoD in the Jira:
> 
> > * Fix the copy column logic to include a cell for the colspan row so that the copied column in the internal clipboard is correct
> > * Fix the paste logic to make sure the correct number of cells are added to the colspan row in the table
> 
> So of the above, the first requirement has not been met with these changes. Also you're missing changelog entries for Snooker.

I thought about that, but I'm not sure that DoD is correct. The column doesn't have a cell, adding one would honestly make the internal clipboard incorrect. The code already compensates for that case, so it does not cause any problems.

lnewson:
> I thought about that, but I'm not sure that DoD is correct. The column doesn't have a cell, adding one would honestly make the internal clipboard incorrect. The code already compensates for that case, so it does not cause any problems.

The content on the clipboard still must be correct, so it's still a bug and something that needs to be fixed. So yes, it's still valid.

spocke:
I just have to little knowledge of snooker. Last time I did anything on snooker or darwin was years ago I could try to get up to speed with things but that would take to long so I'm removing my self and adding in Andy instead.

modules/tinymce/CHANGELOG.md
```diff
@@ -32,6 +32,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Indenting or outdenting list items inside a block element that was inside another list item would not work #TINY-7209
 - Switching between unordered and ordered lists would incorrectly alter any parent element that contained that list #TINY-8068
 - Custom elements on blank lines would be removed during serialization #TINY-4784
+- Pasting columns in tables could sometimes result in an invalid table #TINY-8040
```
TheSpyder, line 35: Please add the copying log as well

- [ ] add extra log entry

TheSpyder (approved):
Much better, thanks Hannes!

modules/snooker/src/test/ts/browser/CopyColumnsTest.ts
```diff
@@ -17,7 +17,7 @@ UnitTest.test('CopyColumnsTest', () => {
     Insert.append(SugarBody.body(), table);
 
     const rowsOpt = copyCols(table, {
-      selection: [ Hierarchy.follow(table, [ section, row, column, 0 ]).getOrDie(label + ': could not follow path') ]
+      selection: [Hierarchy.follow(table, [section, row, column, 0]).getOrDie(label + ': could not follow path')]
```
lnewson, line 20: Something seems to have gone wrong here, as this should have the extra spaces.
```suggestion
      selection: [ Hierarchy.follow(table, [ section, row, column, 0 ]).getOrDie(label + ': could not follow path') ]
```

modules/snooker/src/main/ts/ephox/snooker/api/CopyCols.ts
```diff
@@ -33,7 +33,7 @@ const generateColGroup = (house: Warehouse, minColRange: number, maxColRange: nu
 
 const generateRows = (house: Warehouse, minColRange: number, maxColRange: number): SugarElement<HTMLTableRowElement>[] =>
   Arr.map(house.all, (row) => {
-    const cellsToCopy = Arr.filter(row.cells, (cell) => cell.column >= minColRange && cell.column < maxColRange);
+    const cellsToCopy = Arr.filter(row.cells, (cell) => (cell.column + cell.colspan > minColRange) && cell.column < maxColRange);
```
lnewson, line 36: This doesn't fix the issue... I've pulled down the changes and the content on the clipboard is still an empty row when selecting a column that has a row with a colspan.

lnewson (changes requested):

modules/snooker/src/main/ts/ephox/snooker/api/CopyCols.ts
lnewson: Additionally you've changed `>=` -> `>` which is going to definitely cause issues as it's meant to include the minimum value.

lnewson:
Also please make sure to fix the PR title, as it should describe the change since it'll be used for the final commit message so it shouldn't be something vague like `Making the tables legal`

modules/snooker/src/main/ts/ephox/snooker/api/CopyCols.ts
lnewson: Did some more debugging as it looked like the tests should have caught this and it seems the reason this isn't working is because `tsc` failed to recompile snooker sorry Hannes 😞

That said though, the `>` change is a little confusing and the parenthesis is in quite an odd position. So maybe something like this instead?
```suggestion
    const cellsToCopy = Arr.filter(row.cells, (cell) => {
      const endCol = cell.column + cell.colspan - 1;
      const startCol = cell.column;
      return endCol >= minColRange && startCol < maxColRange;
    });
```
This then keeps it more consistent to how it's done in `CopySelected.ts` as well. We could also abstract this out into a separate function to share between `generateRows` and `generateColGroup`.

lnewson (approved):

PR #7857, opened 2022-05-30
TINY-8750: Fixed a failing test in CssTest on Chrome 102

tiny-ben-tran:
Related Ticket: TINY-8750

Chrome 102 added further MathML support so that those elements can be styled via the browser APIs. And that fails a test in `modules/sugar/src/test/ts/browser/CssTest.ts`. 

Description of Changes:
* Update the check to clobber to the expected CSS for Chrome 102

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [x] Milestone set

### DIFF
--- a/modules/sugar/src/test/ts/browser/CssTest.ts
+++ b/modules/sugar/src/test/ts/browser/CssTest.ts
@@ -39,8 +39,9 @@ UnitTest.test('CssTest', () => {
     Css.copy(c, c2);
     Css.copy(m, c2);
 
-    // NOTE: Safari, Firefox 71+ and Chrome 102+ seem to support styles for math ml tags, so the Css.copy(m, c2) clobbers the previous style
+    // NOTE: Safari, Firefox 71+ and Chromium 102+ seem to support styles for math ml tags, so the Css.copy(m, c2) clobbers the previous style
     const browser = PlatformDetection.detect().browser;
+    // Would cover other Chromium based browsers such as Chrome and Edge
     if (browser.isSafari() || browser.isFirefox() && browser.version.major >= 71 || browser.isChromium() && browser.version.major >= 102) {
       Css.copy(c, c2);
     }
--- a/modules/sugar/src/test/ts/browser/CssTest.ts
+++ b/modules/sugar/src/test/ts/browser/CssTest.ts
@@ -39,7 +39,7 @@ UnitTest.test('CssTest', () => {
     Css.copy(c, c2);
     Css.copy(m, c2);
 
-    // NOTE: Safari, Firefox 71+ and Chrome 102 seem to support styles for math ml tags, so the Css.copy(m, c2) clobbers the previous style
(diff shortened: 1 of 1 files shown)


modules/sugar/src/test/ts/browser/CssTest.ts
```diff
@@ -41,7 +41,7 @@ UnitTest.test('CssTest', () => {
 
     // NOTE: Safari and Firefox 71+ seems to support styles for math ml tags, so the Css.copy(m, c2) clobbers the previous style
```
lnewson, line 42: This description also needs updating, as it has the reason "why" this is done.

lnewson (changes requested):

modules/sugar/src/test/ts/browser/CssTest.ts
```diff
@@ -39,7 +39,7 @@ UnitTest.test('CssTest', () => {
     Css.copy(c, c2);
     Css.copy(m, c2);
 
-    // NOTE: Safari and Firefox 71+ seems to support styles for math ml tags, so the Css.copy(m, c2) clobbers the previous style
+    // NOTE: Safari, Firefox 71+ and Chrome 102 seem to support styles for math ml tags, so the Css.copy(m, c2) clobbers the previous style
```
lnewson, line 42:
```suggestion
    // NOTE: Safari, Firefox 71+ and Chrome 102+ seem to support styles for math ml tags, so the Css.copy(m, c2) clobbers the previous style
```
This needs to be future proof as well.

lnewson: @tiny-ben-tran looks like only half of the suggestion was made here? Admittedly I did edit that immediately after submitting the comment so you may have missed it because of that.

modules/sugar/src/test/ts/browser/CssTest.ts
```diff
@@ -39,8 +39,9 @@ UnitTest.test('CssTest', () => {
     Css.copy(c, c2);
     Css.copy(m, c2);
 
-    // NOTE: Safari, Firefox 71+ and Chrome 102+ seem to support styles for math ml tags, so the Css.copy(m, c2) clobbers the previous style
+    // NOTE: Safari, Firefox 71+ and Chromium 102+ seem to support styles for math ml tags, so the Css.copy(m, c2) clobbers the previous style
     const browser = PlatformDetection.detect().browser;
+    // Would cover other Chromium based browsers such as Chrome and Edge
```
lnewson, line 44:
```suggestion
```
This is unnecessary and implied by `Chromium`. If you really want this it should be in the above comment, not here.

lnewson (approved):

metricjs (approved):

TheSpyder:
I feel like this might need to be rethought completely now that all supported browsers include mathml. The copy inside the browser check is a workaround for styles getting clobbered when mathml is supported.

Not now, though, perhaps log a new task for it.

tiny-ben-tran:
@TheSpyder , I created a task for it so we can have a look at this in the future https://ephocks.atlassian.net/browse/TINY-8763

PR #7859, opened 2022-05-31
TINY-7719: preserve `pre` format on copy and paste

lorenzo-pomili:
Related Ticket: TINY-7719

Description of Changes:

This PR aims to preserve `pre` tags on copy and paste.
To do so now `pre` tags are preserved on copy, and on paste are managed similarly to how the lists work so the element will be unwrapped if it is pasting in the same tag to be flattened when a `pre` element is copied in another `pre` element

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -43,7 +43,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Some option values for the `template` plugin weren't escaped properly when doing replacement lookups via a `RegExp` #TINY-7433
 - Copy events were not dispatched in readonly mode #TINY-6800
 - Using Ctrl+Shift+Home/End (Cmd+Shift+Up/Down on Mac) would not expand the selection when a `contenteditable="false"` element was at the edge of content #TINY-7795
-- Pre tags were not preserved in copy #TINY-7719
+- `<pre>` tags were not preserved when copying and pasting content #TINY-7719
 
 ## 6.0.3 - 2022-05-25
 
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -8,6 +8,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Added
 - New `sidebar_show` option to show the specified sidebar on initialization #TINY-8710
+- New `newline_behavior` option to control what happens when the enter key is pressed or using commands such as `mceInsertNewLine` #TINY-8458
+- New `transparent` property for `iframe` dialog component #TINY-8534
+- New `removeAttributeFilter` and `removeNodeFilter` functions to the `DomParser` and DOM `Serializer` APIs #TINY-7847
+- New `iframe_template_callback` option in the `media` plugin. Patch provided by Namstel #TINY-8684
 
 ### Changed
 - Toggling fullscreen mode with the `fullscreen` plugin now also fires the `ResizeEditor` event #TINY-8701
@@ -15,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The `end_container_on_empty_block` option can now take a string of blocks to split when pressing Enter twice #TINY-6559
 - The default value for `end_container_on_empty_block` option has been changed to `'blockquote'` #TINY-6559
 - Custom elements are now treated as non-empty elements via the schema #TINY-4784
+- The autocompleter menu element is now positioned instead of the wrapper #TINY-6476
 
 ### Fixed
 - Selecting all content with a single image in the content was inconsistent for the keyboard shortcut and menu item #TINY-4550
@@ -40,10 +45,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Copying columns in tables could sometimes result in an invalid copy #TINY-8040
 - The URL detection used for `autolink` and smart paste didn't work if a path segment contained valid characters such as `!` and `:` #TINY-8069
 - Cutting content to the clipboard while selecting between the parent list and a nested list would not always set the list style to `none` on the parent list #TINY-8078
+- The autocompleter would not trigger at the start of nested list items #TINY-8759
 - Some option values for the `template` plugin weren't escaped properly when doing replacement lookups via a `RegExp` #TINY-7433
 - Copy events were not dispatched in readonly mode #TINY-6800
 - Using Ctrl+Shift+Home/End (Cmd+Shift+Up/Down on Mac) would not expand the selection when a `contenteditable="false"` element was at the edge of content #TINY-7795
 - `<pre>` tags were not preserved when copying and pasting content #TINY-7719
+- Preview and Insert Template dialogs display the correct content background color when using dark skins #TINY-8534
 
 ## 6.0.3 - 2022-05-25
 
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -43,6 +43,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Some option values for the `template` plugin weren't escaped properly when doing replacement lookups via a `RegExp` #TINY-7433
 - Copy events were not dispatched in readonly mode #TINY-6800
 - Using Ctrl+Shift+Home/End (Cmd+Shift+Up/Down on Mac) would not expand the selection when a `contenteditable="false"` element was at the edge of content #TINY-7795
+- Pre tags were not preserved in copy #TINY-7719
 
 ## 6.0.3 - 2022-05-25
 
--- a/modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
+++ b/modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
@@ -29,7 +29,7 @@ const shouldFlatten = (fragment: AstNode, parentNode: Element) => {
 
   const isAFlattenableTag = Arr.contains(wrappedElements, currentNode.name);
   const isPastingInTheSameTag = currentNode.name === parentNode.tagName.toLowerCase();
-  // TINY-7719: last children is the bookmarkHtml attached by the function insertHtmlAtCaret
+  // The last children is the bookmarkHtml attached by the function insertHtmlAtCaret
   const isCopingOnlyOneTag = currentNode === fragment.lastChild.prev;
 
   return isCopingOnlyOneTag && isAFlattenableTag && isPastingInTheSameTag;
--- a/modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
+++ b/modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
@@ -24,11 +24,11 @@ import * as InsertList from './InsertList';
 
 const wrappedElements = [ 'pre' ];
 
-const shouldFlatten = (fragment: AstNode, parentNodeTagName: Uppercase<string>) => {
+const shouldFlatten = (fragment: AstNode, parentNode: Element) => {
   const currentNode = fragment.firstChild;
 
   const isAFlattenableTag = Arr.contains(wrappedElements, currentNode.name);
-  const isPastingInTheSameTag = currentNode.name.toUpperCase() === parentNodeTagName;
+  const isPastingInTheSameTag = currentNode.name === parentNode.tagName;
 
   return isAFlattenableTag && isPastingInTheSameTag;
 };
@@ -258,7 +258,7 @@ export const insertHtmlAtCaret = (editor: Editor, value: string, details: Insert
     return value;
   }
 
-  if (details.paste === true && shouldFlatten(fragment, parentNode.tagName)) {
+  if (details.paste === true && shouldFlatten(fragment, parentNode)) {
     fragment.firstChild.unwrap();
   }
 
--- a/modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
+++ b/modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
@@ -24,7 +24,7 @@ import * as InsertList from './InsertList';
 
 const wrappedElements = [ 'pre' ];
 
-const shouldFlatten = (fragment: AstNode, parentNode: Element) => {
+const shouldPasteContentOnly = (fragment: AstNode, parentNode: Element) => {
   const currentNode = fragment.firstChild;
 
   const isAFlattenableTag = Arr.contains(wrappedElements, currentNode.name);
@@ -260,7 +260,7 @@ export const insertHtmlAtCaret = (editor: Editor, value: string, details: Insert
     return value;
   }
 
-  if (details.paste === true && shouldFlatten(fragment, parentNode)) {
+  if (details.paste === true && shouldPasteContentOnly(fragment, parentNode)) {
     fragment.firstChild.unwrap();
   }
 
--- a/modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
+++ b/modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
@@ -22,10 +22,11 @@ import * as SelectionUtils from '../selection/SelectionUtils';
 import { InsertContentDetails } from './ContentTypes';
 import * as InsertList from './InsertList';
 
+const wrappedElements = [ 'pre' ];
+
 const isFlattenable = (fragment: AstNode) => {
   const currentNode = fragment.firstChild;
 
-  const wrappedElements = [ 'pre' ];
   const isAFlattenableTag = Arr.contains(wrappedElements, currentNode.name);
   const isPastingInTheSameTag = currentNode.name === currentNode.parent.name;
 
--- a/modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
+++ b/modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
@@ -25,12 +25,12 @@ import * as InsertList from './InsertList';
 const wrappedElements = [ 'pre' ];
 
 const shouldPasteContentOnly = (fragment: AstNode, parentNode: Element) => {
-  const currentNode = fragment.firstChild;
+  const firstNode = fragment.firstChild;
 
-  const isAFlattenableTag = Arr.contains(wrappedElements, currentNode.name);
-  const isPastingInTheSameTag = currentNode.name === parentNode.tagName.toLowerCase();
+  const isAFlattenableTag = Arr.contains(wrappedElements, firstNode.name);
+  const isPastingInTheSameTag = firstNode.name === parentNode.tagName.toLowerCase();
   const lastNode = fragment.lastChild.attr('data-mce-type') === 'bookmark' ? fragment.lastChild.prev : fragment.lastChild;
-  const isCopingOnlyOneTag = currentNode === lastNode;
+  const isCopingOnlyOneTag = firstNode === lastNode;
 
   return isCopingOnlyOneTag && isAFlattenableTag && isPastingInTheSameTag;
 };
--- a/modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
+++ b/modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
@@ -30,7 +30,7 @@ const shouldFlatten = (fragment: AstNode, parentNode: Element) => {
   const isAFlattenableTag = Arr.contains(wrappedElements, currentNode.name);
   const isPastingInTheSameTag = currentNode.name === parentNode.tagName.toLowerCase();
   // TINY-7719: last children is the bookmarkHtml attached by the function insertHtmlAtCaret
-  const isCopingOnlyOneTag = fragment.firstChild === fragment.lastChild.prev;
+  const isCopingOnlyOneTag = currentNode === fragment.lastChild.prev;
 
   return isCopingOnlyOneTag && isAFlattenableTag && isPastingInTheSameTag;
 };
--- a/modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
+++ b/modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
@@ -24,11 +24,11 @@ import * as InsertList from './InsertList';
 
 const wrappedElements = [ 'pre' ];
 
-const shouldFlatten = (fragment: AstNode, parentNode: Element) => {
+const shouldFlatten = (fragment: AstNode, parentNodeTagName: Uppercase<string>) => {
   const currentNode = fragment.firstChild;
 
   const isAFlattenableTag = Arr.contains(wrappedElements, currentNode.name);
-  const isPastingInTheSameTag = currentNode.name.toUpperCase() === parentNode.tagName;
+  const isPastingInTheSameTag = currentNode.name.toUpperCase() === parentNodeTagName;
 
   return isAFlattenableTag && isPastingInTheSameTag;
 };
@@ -258,7 +258,7 @@ export const insertHtmlAtCaret = (editor: Editor, value: string, details: Insert
     return value;
   }
 
-  if (details.paste === true && shouldFlatten(fragment, parentNode)) {
+  if (details.paste === true && shouldFlatten(fragment, parentNode.tagName)) {
     fragment.firstChild.unwrap();
   }
 
--- a/modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
+++ b/modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
@@ -28,7 +28,7 @@ const shouldFlatten = (fragment: AstNode, parentNode: Element) => {
   const currentNode = fragment.firstChild;
 
   const isAFlattenableTag = Arr.contains(wrappedElements, currentNode.name);
-  const isPastingInTheSameTag = currentNode.name === parentNode.tagName;
+  const isPastingInTheSameTag = currentNode.name.toUpperCase() === parentNode.tagName;
 
   return isAFlattenableTag && isPastingInTheSameTag;
 };
--- a/modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
+++ b/modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
@@ -28,7 +28,7 @@ const shouldFlatten = (fragment: AstNode, parentNode: Element) => {
   const currentNode = fragment.firstChild;
 
   const isAFlattenableTag = Arr.contains(wrappedElements, currentNode.name);
-  const isPastingInTheSameTag = currentNode.name.toUpperCase() === parentNode.tagName;
+  const isPastingInTheSameTag = currentNode.name === parentNode.tagName.toLowerCase();
 
   return isAFlattenableTag && isPastingInTheSameTag;
 };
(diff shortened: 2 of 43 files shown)


modules/tinymce/src/core/test/ts/webdriver/paste/CopyAndPaste.ts
```diff
@@ -0,0 +1,143 @@
...
+      'a',
+      'bdo',
+      'map',
+      'q',
+      'span',
+      'sub',
+      'sup',
+      'button',
+      'label'
+    ];
```
lnewson, line 136: This is quite verbose and takes up a lot of room... it could be squashed down to something like this:
```suggestion
    const inlineElements = [
      'big', 'small', 'tt', 'abbr', 'acronym', 'cite', 'code',
      'dfn', 'em', 'kbd', 'strong', 'samp', 'var', 'a', 'bdo',
      'map', 'q', 'span', 'sub', 'sup', 'button', 'label'
    ];
```

modules/tinymce/src/core/test/ts/webdriver/paste/CopyAndPaste.ts
```diff
@@ -0,0 +1,143 @@
...
+
+    for (const tagName of inlineElements) {
+      await testInlineTags(tagName);
+    }
+  });
+
```
lnewson, line 142:
```suggestion
```

modules/tinymce/src/core/test/ts/webdriver/paste/CopyAndPaste.ts
```diff
@@ -0,0 +1,143 @@
...
+type Selection = [ number[], number, number[], number ];
+
+const copyAndPast = async (editor: Editor, source: Selection, target: Selection): Promise<void> => {
+  const ui = TinyUi(editor);
+  const api = TinyApis(editor);
+
+  api.setSelection(...source);
+  ui.clickOnMenu('button:contains("Edit")');
+  await ui.pWaitForUi('*[role="menu"]');
+  await RealMouse.pClickOn('div[title="Copy"]');
```
lnewson, line 22: Why is this needed? Shouldn't this just be able to be `RealClipboard.pCopy()`?

modules/tinymce/src/core/test/ts/webdriver/paste/CopyAndPaste.ts
```diff
@@ -0,0 +1,143 @@
...
+  api.setSelection(...target);
+  await RealClipboard.pPaste('iframe => body => h1');
```
lnewson, line 24:
```suggestion
  await RealClipboard.pPaste('iframe => body');
```
This doesn't look right, as `=>` means select across frames and there's not an iframe in the editor. Additionally, we don't have a `h1` element in all these tests.

modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
```diff
@@ -22,6 +22,16 @@ import * as SelectionUtils from '../selection/SelectionUtils';
 import { InsertContentDetails } from './ContentTypes';
 import * as InsertList from './InsertList';
 
+const isFlattenable = (fragment: AstNode) => {
```
lnewson, line 25: A better name for this might be `shouldFlatten`? As that's more what we're checking is if the fragment content should be flattened when merged into the existing content.

modules/tinymce/src/core/test/ts/webdriver/paste/CopyAndPaste.ts
```diff
@@ -0,0 +1,143 @@
+import { RealClipboard, RealMouse } from '@ephox/agar';
```
lnewson, line 1: This test won't be running on CI because it's missing the `Test` suffix.

- [ ] Rename the file to `CopyAndPasteTest.ts`

modules/tinymce/src/core/test/ts/webdriver/paste/CopyAndPaste.ts
```diff
@@ -0,0 +1,143 @@
...
+
+  api.setSelection(...source);
+  ui.clickOnMenu('button:contains("Edit")');
+  await ui.pWaitForUi('*[role="menu"]');
+  await RealMouse.pClickOn('div[title="Copy"]');
+  api.setSelection(...target);
+  await RealClipboard.pPaste('iframe => body => h1');
+};
+
+describe('webdriver.tinymce.core.paste.CopyAndPaste', () => {
```
lnewson, line 27:
```suggestion
describe('webdriver.tinymce.core.paste.CopyAndPasteTest', () => {
```

modules/tinymce/src/core/test/ts/webdriver/paste/CopyAndPaste.ts
```diff
@@ -0,0 +1,143 @@
+import { RealClipboard, RealMouse } from '@ephox/agar';
+import { describe, it } from '@ephox/bedrock-client';
+import { TinyApis, TinyAssertions, TinyHooks, TinyUi } from '@ephox/mcagar';
```
lnewson, line 3:
```suggestion
import { TinyApis, TinyAssertions, TinyHooks, TinyUi } from '@ephox/wrap-mcagar';
```

TinyMCE tests should always use the `wrap-mcagar` library instead of `mcagar` directly otherwise it won't use local changes when testing. We have a Jira in the backlog to add a linter rule for this IIRC, but it's not been scheduled yet.

modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
```diff
@@ -22,6 +22,16 @@ import * as SelectionUtils from '../selection/SelectionUtils';
...
+  const currentNode = fragment.firstChild;
+
+  const wrappedElements = [ 'pre' ];
+  const isAFlattenableTag = Arr.contains(wrappedElements, currentNode.name);
+  const isPastingInTheSameTag = currentNode.name === currentNode.parent.name;
+
+  return isAFlattenableTag && isPastingInTheSameTag;
```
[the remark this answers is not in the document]
lorenzo-pomili: Sorry, I'm not sure that I understood what should be the behaviour.

  is the case:
    Option 1:
    ```
    <pre>content</pre><p>paragraph</p>
    <pre>target<pre>
    ```

    ```
    <pre>conten|t</pre><p>p|aragraph</p>
    <pre>target<pre>
    ```

    `copy`

    ```
    <pre>content</pre><p>paragraph</p>
    <pre>tar|get<pre>
    ```

    `paste`

    ```
    <pre>content</pre><p>paragraph</p>
    <pre>tartpget<pre>
    ```

    Option 2:
    ```
    <pre>content</pre><p>paragraph</p>
    <pre>source<pre>
    ```

    ```
    <pre>content</pre><p>paragraph</p>
    <pre>so|ur|ce<pre>
    ```

    `copy`

    ```
    <pre>conten|t</pre><p>p|aragraph</p>
    <pre>source<pre>
    ```

    `paste`

    ```
    <p><pre>contenur</pre>aragraph</p>
    <pre>source<pre>
    ```

modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
```diff
@@ -22,6 +22,16 @@ import * as SelectionUtils from '../selection/SelectionUtils';
 import { InsertContentDetails } from './ContentTypes';
 import * as InsertList from './InsertList';
 
+const isFlattenable = (fragment: AstNode) => {
+  const currentNode = fragment.firstChild;
+
+  const wrappedElements = [ 'pre' ];
```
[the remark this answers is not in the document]
lorenzo-pomili: done

modules/tinymce/CHANGELOG.md
```diff
@@ -43,6 +43,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Some option values for the `template` plugin weren't escaped properly when doing replacement lookups via a `RegExp` #TINY-7433
 - Copy events were not dispatched in readonly mode #TINY-6800
 - Using Ctrl+Shift+Home/End (Cmd+Shift+Up/Down on Mac) would not expand the selection when a `contenteditable="false"` element was at the edge of content #TINY-7795
+- Pre tags were not preserved in copy #TINY-7719
```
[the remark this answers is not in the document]

modules/tinymce/src/core/main/ts/dom/ElementType.ts
```diff
@@ -49,6 +51,8 @@ const isVoid = lazyLookup(voids);
 const isTableSection = lazyLookup(tableSections);
 const isTableCell = lazyLookup<HTMLTableCellElement>(tableCells);
 const isWsPreserveElement = lazyLookup(wsElements);
+const isWrapElementsWithoutInLineElements = lazyLookup(wrapElementsWithoutInLineElements);
+const isWrapElement = (node: SugarElement<Node>) => isWrapElementsWithoutInLineElements(node) || isInline(node);
```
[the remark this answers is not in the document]
lorenzo-pomili: done

modules/tinymce/src/core/test/ts/webdriver/paste/CopyAndPaste.ts
```diff
@@ -0,0 +1,143 @@
...
+        `<${tagName}>abc</${tagName}>\n` +
+        `<${tagName}>e tag</${tagName}>`
+      );
+    };
+
+    for (const tagName of [ 'h1', 'h2', 'h3', 'h4', 'h5', 'h6' ]) {
+      await testBlockTags(tagName);
+    }
+
+    const testPreTag = async (tagName) => {
```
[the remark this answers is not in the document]
lorenzo-pomili: done

modules/tinymce/src/core/test/ts/webdriver/paste/CopyAndPaste.ts
lorenzo-pomili: done

modules/tinymce/src/core/test/ts/webdriver/paste/CopyAndPaste.ts
lorenzo-pomili: done

modules/tinymce/src/core/test/ts/webdriver/paste/CopyAndPaste.ts
lorenzo-pomili: replacing with it seems not to trigger the copy, is that possible that I'm using the wrong `selector`, I tried with `RealClipboard.pCopy('iframe => body')`, but it does not work, so I may have misunderstood how selectors work in those functions

modules/tinymce/src/core/test/ts/webdriver/paste/CopyAndPaste.ts
lorenzo-pomili: done

modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
lorenzo-pomili: done

modules/tinymce/src/core/test/ts/webdriver/paste/CopyAndPaste.ts
lorenzo-pomili: done

modules/tinymce/src/core/test/ts/webdriver/paste/CopyAndPaste.ts
lorenzo-pomili: done

modules/tinymce/src/core/test/ts/webdriver/paste/CopyAndPaste.ts
lorenzo-pomili: sorry I forgot a detail about this one, in reality copy event is triggered (I'm checking with `editor.on('copy', (e) => console.log('copy: ', e));`), but the content seems not copied in the clipboard, using:
```
yarn bedrock-auto -b chrome -f modules/tinymce/src/core/test/ts/webdriver/paste/CopyAndPasteTest.ts
```
I get as a result of the `paste` the last thing that I copy in my clipboard and using `chrome-headless` I get nothing

modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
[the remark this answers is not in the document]
spocke: Check if the fragment.firstChild and fragment.lastChild is the same. Maybe just unpack them as:
```
const { firstChild, lastChild } = fragment;
....
return firstChild === lastChild && isAFlattenableTag;
```
It probably needs a undefined check if we pass in an empty fragment.

modules/tinymce/src/core/test/ts/webdriver/paste/CopyAndPasteTest.ts
```diff
@@ -0,0 +1,124 @@
...
+import Editor from 'tinymce/core/api/Editor';
+
+describe('webdriver.tinymce.core.paste.CopyAndPasteTest', () => {
+  const hook = TinyHooks.bddSetup<Editor>({
+    base_url: '/project/tinymce/js/tinymce',
+    toolbar: false,
+    statusbar: false
+  }, []);
+
+  type Selection = [ number[], number, number[], number ];
```
spocke, line 14: Maybe use the CursorPath type from agar Cursors here. It is an object structure though so `{ startPath: [0], soffset: 0, finishPath: [0], foffset: 0 }` having an object is a bit easier to read I think.

modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
```diff
@@ -22,6 +22,17 @@ import * as SelectionUtils from '../selection/SelectionUtils';
 import { InsertContentDetails } from './ContentTypes';
 import * as InsertList from './InsertList';
 
+const wrappedElements = [ 'pre' ];
+
+const shouldFlatten = (fragment: AstNode, parentNode: Element) => {
+  const currentNode = fragment.firstChild;
+
+  const isAFlattenableTag = Arr.contains(wrappedElements, currentNode.name);
+  const isPastingInTheSameTag = currentNode.name.toUpperCase() === parentNode.tagName;
```
lnewson, line 31: Just for consistency as we normally lowercase the tag name instead in these comparisons.
```suggestion
  const isPastingInTheSameTag = currentNode.name === parentNode.tagName.toLowerCase();
```

In all honesty, we probably should look a making a helper function in say `Strings.ts` for case insensitive comparison. It might be worth logging that as a separate tech debt task?

modules/tinymce/src/core/test/ts/webdriver/paste/CopyAndPaste.ts
lnewson: `RealClipboard.pCopy('iframe => body')` should work in this case... oh well, if you've tried it and it's not working then I'm happy with this workaround.

lnewson: Ahh I just looked up the test cases for this API and it has this which might explain why that didn't work: https://github.com/tinymce/tinymce/blob/674a17a92c2f91e8ca9af4e6abd188e8993f5fdd/modules/agar/src/test/ts/webdriver/RealEffectsTest.ts#L21

modules/tinymce/src/core/test/ts/webdriver/paste/CopyAndPasteTest.ts
lorenzo-pomili: ok, actually, I used this signature to imitate functions like `setSelection`, changing it to an object would result in this:

```
await pCopyAndPaste(
  editor,
  { startPath: [ 0, 0 ], soffset: 0, finishPath: [ 0, 0 ], foffset: 3 },
  { startPath: [ 2, 0 ], soffset: 1, finishPath: [ 2, 0 ], foffset: 3 }
);
```

could make sense to make a helper function like this?

```
await pCopyAndPaste(
  editor,
  toSelection([ 0, 0 ], 0, [ 0, 0 ], 3),
  toSelection([ 2, 0 ], 1, [ 2, 0 ], 3)
);
```

or

```
await pCopyAndPaste(
  editor, {
    source: toSelection([ 0, 0 ], 0, [ 0, 0 ], 3),
    target: toSelection([ 2, 0 ], 1, [ 2, 0 ], 3)
  }
);
```

in this way, the type `Selection` would be more readable and the signature of `pCopyAndPaste` would stay similar to the other functions like `setSelection`

modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
lorenzo-pomili: done

modules/tinymce/src/core/test/ts/webdriver/paste/CopyAndPaste.ts
lorenzo-pomili: ok, I added a comment, if it is ok, I'll resolve the conversation

modules/tinymce/src/core/test/ts/webdriver/paste/CopyAndPasteTest.ts
lnewson: The problem with both of those aren't as readable... the named approach lets you see what each component signifies with no background required for what the type/order is.

lorenzo-pomili: ok, so I'll go with the object solution, thanks

lorenzo-pomili: done

lnewson:
@lorenzo-pomili tests are failing here so this isn't in a state to review yet sorry.

modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
```diff
@@ -22,6 +22,19 @@ import * as SelectionUtils from '../selection/SelectionUtils';
...
 
+const wrappedElements = [ 'pre' ];
+
+const shouldFlatten = (fragment: AstNode, parentNode: Element) => {
+  const currentNode = fragment.firstChild;
+
+  const isAFlattenableTag = Arr.contains(wrappedElements, currentNode.name);
+  const isPastingInTheSameTag = currentNode.name === parentNode.tagName.toLowerCase();
+  // TINY-7719: last children is the bookmarkHtml attached by the function insertHtmlAtCaret
+  const isCopingOnlyOneTag = fragment.firstChild === fragment.lastChild.prev;
```
lnewson, line 33:
```suggestion
  const isCopingOnlyOneTag = currentNode === fragment.lastChild.prev;
```
We should be reusing the variable here. Also this might make more sense to be called `firstChild` or something instead of `currentNode`.

modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
```diff
@@ -22,6 +22,19 @@ import * as SelectionUtils from '../selection/SelectionUtils';
...
 import * as InsertList from './InsertList';
 
+const wrappedElements = [ 'pre' ];
+
+const shouldFlatten = (fragment: AstNode, parentNode: Element) => {
+  const currentNode = fragment.firstChild;
+
+  const isAFlattenableTag = Arr.contains(wrappedElements, currentNode.name);
+  const isPastingInTheSameTag = currentNode.name === parentNode.tagName.toLowerCase();
+  // TINY-7719: last children is the bookmarkHtml attached by the function insertHtmlAtCaret
```
lnewson, line 32:
```suggestion
  // The last children is the bookmarkHtml attached by the function insertHtmlAtCaret
```
This logic/comment has no relation to the Jira so we shouldn't be mentioning it here.

modules/tinymce/src/core/test/ts/webdriver/paste/CopyAndPasteTest.ts
```diff
@@ -0,0 +1,172 @@
...
+import Editor from 'tinymce/core/api/Editor';
+
+describe('webdriver.tinymce.core.paste.CopyAndPasteTest', () => {
+  const hook = TinyHooks.bddSetup<Editor>({
+    base_url: '/project/tinymce/js/tinymce',
+    toolbar: false,
+    statusbar: false
+  }, []);
+
+  interface Selection { startPath: number[]; soffset: number; finishPath: number[]; foffset: number }
```
lnewson, line 14: This type should be at the top of the file... Also this should not be needed as you should be able to use the `CursorPath` type instead.

- [ ] Use the `CursorPath` type instead of a duplicate/custom type

modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
lorenzo-pomili: done

modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
lorenzo-pomili: done

modules/tinymce/src/core/test/ts/webdriver/paste/CopyAndPasteTest.ts
lorenzo-pomili: done

lorenzo-pomili:
@lnewson I solved the problem with the tests, it seems that the paste on Safari was not working, I add an if, for Safari that uses the `Paste` from the menu (similar to what is used for the copy), but I'm not sure if this is the best solution, let me know what do you think about that

modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
```diff
@@ -22,6 +22,19 @@ import * as SelectionUtils from '../selection/SelectionUtils';
 import { InsertContentDetails } from './ContentTypes';
 import * as InsertList from './InsertList';
 
+const wrappedElements = [ 'pre' ];
+
+const shouldFlatten = (fragment: AstNode, parentNode: Element) => {
+  const currentNode = fragment.firstChild;
```
lnewson, line 28: This name is still confusing to me at least, can we please change it?

modules/tinymce/src/core/test/ts/webdriver/paste/CopyAndPasteTest.ts
```diff
@@ -0,0 +1,177 @@
...
+  }, []);
+
+  const pCopyAndPaste = async (editor: Editor, source: Cursors.CursorPath, target: Cursors.CursorPath): Promise<void> => {
+    TinySelections.setSelection(editor, source.startPath, source.soffset, source.finishPath, source.foffset);
+    // at the moment: RealClipboard.pCopy('iframe => body'), doesn't work in with all browser (see https://github.com/webdriverio/webdriverio/issues/622)
+    // chrome, chrome-headless, firefox-headless -> it doesn't work
+    // firefox -> it works
+    TinyUiActions.clickOnMenu(editor, 'button:contains("Edit")');
+    await TinyUiActions.pWaitForUi(editor, '*[role="menu"]');
+    await RealMouse.pClickOn('div[title="Copy"]');
```
lnewson, line 22: This is now duplicated. Can we please abstract it? For example:
```
const pClickEditMenu = async (item: string) => {
  TinyUiActions.clickOnMenu(editor, 'button:contains("Edit")');
  await TinyUiActions.pWaitForUi(editor, '*[role="menu"]');
  await RealMouse.pClickOn(`div[title="${item}"]`);
};
```

modules/tinymce/src/core/test/ts/webdriver/paste/CopyAndPasteTest.ts
```diff
@@ -0,0 +1,177 @@
...
+    TinySelections.setSelection(editor, target.startPath, target.soffset, target.finishPath, target.foffset);
+    if (PlatformDetection.detect().browser.isSafari()) {
```
lnewson, line 24: I wonder if we should just do this logic for all cases, as it doesn't really matter which means is used 🤷

lnewson (approved):

modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
```diff
@@ -22,6 +22,19 @@ import * as SelectionUtils from '../selection/SelectionUtils';
...
+
+  const isAFlattenableTag = Arr.contains(wrappedElements, currentNode.name);
+  const isPastingInTheSameTag = currentNode.name === parentNode.tagName.toLowerCase();
+  // The last children is the bookmarkHtml attached by the function insertHtmlAtCaret
+  const isCopingOnlyOneTag = currentNode === fragment.lastChild.prev;
```
lnewson, line 33: Should this double check that the lastChild is the bookmark span? As this could still lead to unexpected cases e.g:
```suggestion
const lastNode = fragment.lastChild.attr('data-mce-type') === 'bookmark' ? fragment.lastChild.prev : fragment.lastChild;
const isCopingOnlyOneTag =  currentNode === lastNode;
```

modules/tinymce/src/core/test/ts/webdriver/paste/CopyAndPasteTest.ts
lorenzo-pomili: Actually, it does not work in all cases some browsers forbid using it by saying that direct access to the clipboard is not supported

modules/tinymce/src/core/test/ts/webdriver/paste/CopyAndPasteTest.ts
lorenzo-pomili: done

modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
lorenzo-pomili: done

modules/tinymce/src/core/test/ts/webdriver/paste/CopyAndPasteTest.ts
lnewson: Ahh right forgot that sorry. Lets leave it forking then 👍

modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
lorenzo-pomili: I'm thinking about what name could be more clear but at the moment the only idea that I have is `shouldBeFlattened`.

something about that is just one tag pasted in the same tag like: `isPastingInTheSameTag` would lose the information that only certain tags should be flattened

something like: `shouldBeUnwapped` seems less correct to me since the unwrapping is used the way we use to just copy the content and it is not what we want to achieve.

Maybe `shouldPastContentOnly`?

lnewson: I wasn't talking about that above... this is about the `currentNode` name.

lnewson: That being said `shouldPasteContentOnly` is probably a better name for this function 👍

lorenzo-pomili: ok sorry I misunderstood :sweat_smile: , so I'll also rename the function, about the name of the variable, I'm thinking about: `pastingNode`, `firstNode` or removing the variable and using `fragment.firstChild` for the 3 occurrences

lnewson: Using a variable means it will minify better so I think we should keep that. My preference would just be `firstNode` to match `lastNode`.

lorenzo-pomili: ok, done

spocke (approved):

PR #7863, opened 2022-06-01
TINY-8710: Fix bugs in sidebar_show option

tiny-ben-tran:
Related Ticket: TINY-8710

This is to handle the case where the comments sidebar takes up the whole editor when the skin isn't loaded yet by the time calling `setSidebar()`

Description of Changes:
* Remove the wrong width that is set to the sidebar when the skin isn't loaded yet.
* Fix a case where the slot container is open even if the specified sidebar doesn't exist 

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/themes/silver/main/ts/api/Options.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/api/Options.ts
@@ -301,7 +301,7 @@ const useElementPath = option('elementpath');
 const useBranding = option('branding');
 const getResize = option('resize');
 const getPasteAsText = option('paste_as_text');
-const getSidebarShow = option('sidebar_show');
+const getSidebarShow = wrapOptional(option('sidebar_show'));
 
 const isSkinDisabled = (editor: Editor): boolean =>
   editor.options.get('skin') === false;
--- a/modules/tinymce/src/themes/silver/main/ts/api/Options.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/api/Options.ts
@@ -301,7 +301,7 @@ const useElementPath = option('elementpath');
 const useBranding = option('branding');
 const getResize = option('resize');
 const getPasteAsText = option('paste_as_text');
-const getSidebarShow = wrapOptional(option('sidebar_show'));
+const getSidebarShow = option('sidebar_show');
 
 const isSkinDisabled = (editor: Editor): boolean =>
   editor.options.get('skin') === false;
--- a/modules/tinymce/src/themes/silver/main/ts/modes/Iframe.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/modes/Iframe.ts
@@ -100,7 +100,7 @@ const render = (editor: Editor, uiComponents: RenderUiComponents, rawUiConfig: R
     OuterContainer.setSidebar(
       outerContainer,
       rawUiConfig.sidebar,
-      Options.getSidebarShow(editor).fold(() => undefined, (value) => value.toLowerCase())
+      Options.getSidebarShow(editor)
     );
 
     setupEvents(editor, uiComponents);
--- a/modules/tinymce/src/themes/silver/main/ts/modes/Iframe.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/modes/Iframe.ts
@@ -100,7 +100,7 @@ const render = (editor: Editor, uiComponents: RenderUiComponents, rawUiConfig: R
     OuterContainer.setSidebar(
       outerContainer,
       rawUiConfig.sidebar,
-      Options.getSidebarShow(editor)
+      Options.getSidebarShow(editor).fold(() => undefined, (value) => value.toLowerCase())
     );
 
     setupEvents(editor, uiComponents);
--- a/modules/tinymce/src/themes/silver/main/ts/ui/sidebar/Sidebar.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/ui/sidebar/Sidebar.ts
@@ -105,6 +105,8 @@ const setSidebar = (sidebar: AlloyComponent, panelConfigs: SidebarConfig, showSi
       Composing.getCurrent(slider).each((slotContainer) => {
         SlotContainer.showSlot(slotContainer, showSidebar);
         Sliding.immediateGrow(slider);
+        // TINY-8710: Remove the wrong width from being set to the sidebar when the skin isn't loaded yet
+        Css.remove(slider.element, 'width');
       });
     }
   });
--- a/modules/tinymce/src/themes/silver/main/ts/ui/sidebar/Sidebar.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/ui/sidebar/Sidebar.ts
@@ -103,10 +103,12 @@ const setSidebar = (sidebar: AlloyComponent, panelConfigs: SidebarConfig, showSi
     // Show the default sidebar
     if (Type.isString(showSidebar)) {
       Composing.getCurrent(slider).each((slotContainer) => {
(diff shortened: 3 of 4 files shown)


modules/tinymce/src/themes/silver/main/ts/ui/sidebar/Sidebar.ts
```diff
@@ -103,8 +103,12 @@ const setSidebar = (sidebar: AlloyComponent, panelConfigs: SidebarConfig, showSi
     // Show the default sidebar
     if (Type.isString(showSidebar)) {
       Composing.getCurrent(slider).each((slotContainer) => {
-        SlotContainer.showSlot(slotContainer, showSidebar);
-        Sliding.immediateGrow(slider);
+        if (!SlotContainer.getSlot(slotContainer, showSidebar).isNone()) {
```
lnewson, line 106: This could potentially be simplified by just checking the name is in the panel configs instead. For example:
```
if (Type.isString(showSidebar) && Obj.has(panelConfigs, showSidebar)) {
```

Another thing I realised we missed here, is that the sidebar names will have been lowercased here. I mentioned this a while back but I forgot this fact as well.

- [ ] Ensure to lowercase the sidebar name so it matches the registry panelConfigs name

modules/tinymce/src/themes/silver/test/ts/browser/sidebar/SidebarShowOptionTest.ts
```diff
@@ -39,6 +39,8 @@ describe('browser.tinymce.themes.silver.sidebar.SidebarShowOptionTest', () => {
   };
 
   const assertNotAnimating = () => UiFinder.notExists(SugarBody.body(), '.tox-sidebar--sliding-growing');
+  const assertSlotContainerOpen = () => UiFinder.exists(SugarBody.body(), '.tox-sidebar--sliding-open');
+  const assertSlotContainerNotOpen = () => UiFinder.notExists(SugarBody.body(), '.tox-sidebar--sliding-open');
```
lnewson, line 43: This isn't asserting the slot contain is open, it's checking the sidebar so the should probably be called `assertSidebarOpen` and `assertSidebarClosed` (or similar).

modules/tinymce/src/themes/silver/test/ts/browser/sidebar/SidebarShowOptionTest.ts
```diff
@@ -73,6 +77,7 @@ describe('browser.tinymce.themes.silver.sidebar.SidebarShowOptionTest', () => {
       sidebar_show: 'sidebartwo'
     });
     assertNotAnimating();
+    assertSlotContainerNotOpen();
```
lnewson, line 80: This doesn't look right? Based on the test `sidebartwo` should be open here.
```suggestion
    assertSlotContainerOpen();
```

lnewson (approved):

spocke (approved):

modules/tinymce/src/themes/silver/main/ts/modes/Iframe.ts
```diff
@@ -100,7 +100,7 @@ const render = (editor: Editor, uiComponents: RenderUiComponents, rawUiConfig: R
     OuterContainer.setSidebar(
       outerContainer,
       rawUiConfig.sidebar,
-      Options.getSidebarShow(editor)
+      Options.getSidebarShow(editor).fold(() => undefined, (value) => value.toLowerCase())
```
ltrouton, line 103: Wondering if this would be clearer?
```suggestion
      Options.getSidebarShow(editor).map((value) => value.toLowerCase()).getOrUndefined()
```

ltrouton (approved):

modules/tinymce/src/themes/silver/main/ts/api/Options.ts
```diff
@@ -301,7 +301,7 @@ const useElementPath = option('elementpath');
 const useBranding = option('branding');
 const getResize = option('resize');
 const getPasteAsText = option('paste_as_text');
-const getSidebarShow = option('sidebar_show');
+const getSidebarShow = wrapOptional(option('sidebar_show'));
```
lnewson, line 304: Why is this an `Optional` value now when we're immediately unwrapping it anyways?

modules/tinymce/src/themes/silver/main/ts/ui/sidebar/Sidebar.ts
```diff
@@ -101,10 +101,12 @@ const setSidebar = (sidebar: AlloyComponent, panelConfigs: SidebarConfig, showSi
     Replacing.set(slider, [ makeSidebar(panelConfigs) ]);
 
     // Show the default sidebar
-    if (Type.isString(showSidebar)) {
+    if (Type.isString(showSidebar) && Obj.has(panelConfigs, showSidebar)) {
```
lnewson, line 104: Doing the `lowerCase` check here would make more sense and would be more foolproof as this could be called from different places.

modules/tinymce/src/themes/silver/main/ts/api/Options.ts
tiny-ben-tran: So I could use `Optional` helper functions to lowercase the input value. Realized that it is a not so good idea

modules/tinymce/src/themes/silver/main/ts/ui/sidebar/Sidebar.ts
```diff
@@ -101,10 +101,13 @@ const setSidebar = (sidebar: AlloyComponent, panelConfigs: SidebarConfig, showSi
     Replacing.set(slider, [ makeSidebar(panelConfigs) ]);
 
     // Show the default sidebar
-    if (Type.isString(showSidebar)) {
+    const configKey = Optional.from(showSidebar).map((value) => value.toLowerCase()).getOrUndefined();
```
lnewson, line 104:
```suggestion
    const configKey = value?.toLowerCase();
```
This can be simplified to this I believe?

modules/tinymce/src/themes/silver/main/ts/ui/sidebar/Sidebar.ts
```diff
@@ -101,10 +101,13 @@ const setSidebar = (sidebar: AlloyComponent, panelConfigs: SidebarConfig, showSi
...
+    if (Type.isString(showSidebar) && Obj.has(panelConfigs, configKey)) {
       Composing.getCurrent(slider).each((slotContainer) => {
         SlotContainer.showSlot(slotContainer, showSidebar);
         Sliding.immediateGrow(slider);
+        // TINY-8710: Remove the width as since the skins/styles won't have loaded yet it's going to be incorrect
```
lnewson, line 109:
```suggestion
        // TINY-8710: Remove the width as since the skins/styles won't have loaded yet, so it's going to be incorrect
```

modules/tinymce/src/themes/silver/test/ts/browser/sidebar/SidebarShowOptionTest.ts
```diff
@@ -67,12 +71,13 @@ describe('browser.tinymce.themes.silver.sidebar.SidebarShowOptionTest', () => {
     McEditor.remove(editor);
   });
 
-  it('TINY-8710: Show a different sidebar on init', async () => {
+  it('TINY-8710: Sidebar name should be lowercased', async () => {
     const editor = await McEditor.pFromSettings<Editor>({
       ...settingsFactory(store),
-      sidebar_show: 'sidebartwo'
+      sidebar_show: 'SideBarTwo'
```
lnewson, line 77: If you're changing this, then shouldn't the name in the `addSidebar` call be changed as well? As the intention here is that they should match.

PR #7704, opened 2022-03-18
TINY-8573: Fix autcomplete border

noxuhax:
Related Ticket: TINY-8573

Description of Changes:
* Fix autcomplete border

Pre-checks:
* [x] Changelog entry added
* [x] ~Tests have been added (if applicable)~
* [x] Branch prefixed with `feature/` for new features (if applicable)

Review:
* [x] Milestone set
* [x] Review comments resolved

PR #7864, opened 2022-06-02
TINY-6476: Fixed the autocompleter positioning the menu incorrectly

lnewson:
Related Ticket:

Description of Changes:
* Fixes the `tox-autocompleter` sandbox being positioned instead of the menu (which in turn caused some issues with styling). The core issue is that we were using the wrong `InlineView` API as the one being used is meant to position the sandbox so that content can be swapped in and out (such as in context forms). 
* Reverted the CSS workaround done in https://github.com/tinymce/tinymce/pull/7704 due to this bug.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/oxide/src/less/theme/components/form/autocomplete.less
+++ b/modules/oxide/src/less/theme/components/form/autocomplete.less
@@ -9,11 +9,7 @@
   }
 
   .tox-autocompleter .tox-menu {
-    @menu-border-color: @border-color;
-    @menu-box-shadow: none;
-
-    border-color: @menu-border-color;
-    box-shadow: @menu-box-shadow;
+    box-sizing: border-box;
     max-width: 25em;
   }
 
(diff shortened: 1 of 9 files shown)


modules/oxide/src/less/theme/components/form/autocomplete.less
```diff
@@ -9,11 +9,7 @@
   }
 
   .tox-autocompleter .tox-menu {
-    @menu-border-color: @border-color;
-    @menu-box-shadow: none;
-
-    border-color: @menu-border-color;
-    box-shadow: @menu-box-shadow;
+    box-sizing: border-box;
```
lnewson, line 12: This was added because without this when the menu is positioned the size changes because it was previously constrained within the `.tox-autocompleter` wrapper. We also want to ensure the `max-width` is actually `25em` including the padding to keep it consistent with how this worked previously.

noxuhax (approved):

PR #7878, opened 2022-06-07
TINY-7847: Added new parser/serializer remove filter APIs

lnewson:
Related Ticket: TINY-7847

Description of Changes:
* Added the new `removeNodeFilter` and `removeAttributeFilter` functions to the parser/serializer APIs.
* Abstracted the filter registry logic out into a separate module to avoid duplication between nodes/attributes.
* Fixed a couple of minor issues with the existing parser/serializer doc strings/examples.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/api/dom/Serializer.ts
+++ b/modules/tinymce/src/core/main/ts/api/dom/Serializer.ts
@@ -111,7 +111,7 @@ const DomSerializer = (settings: DomSerializerSettings, editor?: Editor): DomSer
      * Removes a node filter function or removes all filter functions from the parser used by the serializer for the node names provided.
      *
      * @method removeNodeFilter
-     * @param {String} name Comma separated list of nodes names to remove filters for.
+     * @param {String} name Comma separated list of node names to remove filters for.
      * @param {Function} callback Optional callback function to only remove a specific callback.
      * @example
      * // Remove a single filter
@@ -130,7 +130,7 @@ const DomSerializer = (settings: DomSerializerSettings, editor?: Editor): DomSer
      * Removes an attribute filter function or removes all filter functions from the parser used by the serializer for the attribute names provided.
      *
      * @method removeAttributeFilter
-     * @param {String} name Comma separated list of attributes names to remove filters for.
+     * @param {String} name Comma separated list of attribute names to remove filters for.
      * @param {Function} callback Optional callback function to only remove a specific callback.
      * @example
      * // Remove a single filter
--- a/modules/tinymce/src/core/main/ts/api/dom/Serializer.ts
+++ b/modules/tinymce/src/core/main/ts/api/dom/Serializer.ts
@@ -29,7 +29,7 @@ const DomSerializer = (settings: DomSerializerSettings, editor?: Editor): DomSer
      * and then execute the callback once it has finished parsing the document.
      *
      * @method addNodeFilter
-     * @method {String} name Comma separated list of nodes to collect.
+     * @param {String} name Comma separated list of nodes to collect.
      * @param {Function} callback Callback function to execute once it has collected nodes.
      * @example
      * serializer.addNodeFilter('p,h1', (nodes, name) => {
@@ -46,7 +46,7 @@ const DomSerializer = (settings: DomSerializerSettings, editor?: Editor): DomSer
      * and then execute the callback once it has finished parsing the document.
      *
      * @method addAttributeFilter
-     * @method {String} name Comma separated list of attributes to collect.
+     * @param {String} name Comma separated list of attributes to collect.
      * @param {Function} callback Callback function to execute once it has collected nodes.
      * @example
      * serializer.addAttributeFilter('src,href', (nodes, name) => {
--- a/modules/tinymce/src/core/main/ts/api/dom/Serializer.ts
+++ b/modules/tinymce/src/core/main/ts/api/dom/Serializer.ts
@@ -32,7 +32,7 @@ const DomSerializer = (settings: DomSerializerSettings, editor?: Editor): DomSer
      * @method {String} name Comma separated list of nodes to collect.
      * @param {Function} callback Callback function to execute once it has collected nodes.
      * @example
-     * parser.addNodeFilter('p,h1', (nodes, name) => {
+     * serializer.addNodeFilter('p,h1', (nodes, name) => {
      *   for (let i = 0; i < nodes.length; i++) {
      *     console.log(nodes[i].name);
      *   }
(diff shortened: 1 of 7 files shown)


modules/tinymce/src/core/main/ts/api/dom/Serializer.ts
```diff
@@ -105,7 +105,45 @@ const DomSerializer = (settings: DomSerializerSettings, editor?: Editor): DomSer
...
     getNodeFilters: domSerializer.getNodeFilters,
 
-    getAttributeFilters: domSerializer.getAttributeFilters
+    getAttributeFilters: domSerializer.getAttributeFilters,
+
+    /**
+     * Removes a node filter function or removes all filter functions from the parser used by the serializer for the node names provided.
+     *
+     * @method removeNodeFilter
+     * @param {String} name Comma separated list of nodes names to remove filters for.
```
tiny-ben-tran, line 114:
```suggestion
     * @param {String} name Comma separated list of node names to remove filters for.
```

modules/tinymce/src/core/main/ts/api/dom/Serializer.ts
```diff
@@ -105,7 +105,45 @@ const DomSerializer = (settings: DomSerializerSettings, editor?: Editor): DomSer
...
+     * // Remove all filters
+     * serializer.removeNodeFilter('p,h1');
+     */
+    removeNodeFilter: domSerializer.removeNodeFilter,
+
+    /**
+     * Removes an attribute filter function or removes all filter functions from the parser used by the serializer for the attribute names provided.
+     *
+     * @method removeAttributeFilter
+     * @param {String} name Comma separated list of attributes names to remove filters for.
```
tiny-ben-tran, line 133:
```suggestion
     * @param {String} name Comma separated list of attribute names to remove filters for.
```

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
```diff
@@ -434,36 +434,35 @@ const DomParser = (settings: DomParserSettings = {}, schema = Schema()): DomPars
...
-    }
+  const getNodeFilters = nodeFilterRegistry.getFilters;
 
-    return out;
-  };
+  /**
+   * Removes a node filter function or removes all filter functions from the parser for the node names provided.
+   *
+   * @method removeNodeFilter
+   * @param {String} name Comma separated list of nodes names to remove filters for.
```
tiny-ben-tran, line 445:
```suggestion
   * @param {String} name Comma separated list of node names to remove filters for.
```

tiny-ben-tran (approved):

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
```diff
@@ -472,22 +471,28 @@ const DomParser = (settings: DomParserSettings = {}, schema = Schema()): DomPars
...
-    });
-  };
+  const getAttributeFilters = attributeFilterRegistry.getFilters;
 
-  const getAttributeFilters = (): ParserFilter[] => [].concat(attributeFilters);
+  /**
+   * Removes an attribute filter function or removes all filter functions from the parser for the attribute names provided.
+   *
+   * @method removeAttributeFilter
+   * @param {String} name Comma separated list of attributes names to remove filters for.
```
lnewson, line 482:
```suggestion
   * @param {String} name Comma separated list of attribute names to remove filters for.
```

modules/tinymce/src/core/main/ts/api/dom/Serializer.ts
```diff
@@ -105,7 +105,45 @@ const DomSerializer = (settings: DomSerializerSettings, editor?: Editor): DomSer
...
+
+    /**
+     * Removes a node filter function or removes all filter functions from the parser used by the serializer for the node names provided.
+     *
+     * @method removeNodeFilter
+     * @param {String} name Comma separated list of node names to remove filters for.
+     * @param {Function} callback Optional callback function to only remove a specific callback.
+     * @example
+     * // Remove a single filter
+     * serializer.removeNodeFilter('p,h1', (nodes, name) => {
```
spocke, line 118: This example makes little sense. Maybe just have it be `serializer.removeNodeFilter('p,h1', someFunction)` or maybe just remove the example here since this is not very commonly used. I tend to add examples to things that are used a lot.

modules/tinymce/src/core/main/ts/api/dom/Serializer.ts
```diff
@@ -105,7 +105,45 @@ const DomSerializer = (settings: DomSerializerSettings, editor?: Editor): DomSer
...
+     */
+    removeNodeFilter: domSerializer.removeNodeFilter,
+
+    /**
+     * Removes an attribute filter function or removes all filter functions from the parser used by the serializer for the attribute names provided.
+     *
+     * @method removeAttributeFilter
+     * @param {String} name Comma separated list of attribute names to remove filters for.
+     * @param {Function} callback Optional callback function to only remove a specific callback.
+     * @example
```
spocke, line 135: Same here remove or simplify the example since passing in a function like that needs to be a reference to an existing function.

modules/tinymce/src/core/main/ts/html/FilterRegistry.ts
```diff
@@ -0,0 +1,58 @@
...
+  callbacks: C[];
+}
+
+export interface FilterRegistry<C extends Function> {
+  readonly addFilter: (name: string, callback: C) => void;
+  readonly getFilters: () => Filter<C>[];
+  readonly removeFilter: (name: string, callback?: C) => void;
+}
+
+export const FilterRegistry = <C extends Function>(): FilterRegistry<C> => {
```
spocke, line 18: I wonder what style we should use here. The constructor style here is more of a legacy in my mind from the `new Foo` style. So maybe this could be renamed `makeFilterRegistry` instead?

spocke (approved):

modules/tinymce/src/core/main/ts/html/FilterRegistry.ts
lnewson: I hadn't seen any indication that we'd switched to be honest, but I don't mind changing it. Though I think we can just get away with just `create` if we do that as we already have the context based on the module name.

lnewson: That would also be consistent with other more recent APIs I guess, such as Mode and Options: https://github.com/tinymce/tinymce/blob/91b891eb142d14040a12a29842338bc985284c80/modules/tinymce/src/core/main/ts/api/Mode.ts#L71

modules/tinymce/src/core/main/ts/api/dom/Serializer.ts
lnewson: Yeah I did wonder that myself when creating these. I'll swap it to basically what you suggested, though I might use `someCallback` just to match the parameter name.

modules/tinymce/src/core/main/ts/html/FilterRegistry.ts
spocke: Yeah, internally at least we should try to use more normal functions I guess then externally it might be hard since there we already have a bunch of legacy named things when it was a constructor function and some things still are classes for example.

lnewson: Done

PR #7884, opened 2022-06-09
TINY-8458: Fix newline commands 

TheSpyder:
Related Ticket: TINY-8458

Description of Changes:
- Corrected behaviour of InsertParagraph with `newline_behavior` set.
- Fix `InsertLineBreak` bug where it did not delete selection.
- This adds a little bit to our bundle size (the `InsertBreak` and `InsertBr` modules now generate an export object in the bundle where previously the functions were used directly) but it seems worth it.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Added
 - New `sidebar_show` option to show the specified sidebar on initialization #TINY-8710
-- New `newline_behavior` option to control what happens when the enter key is pressed or using commands such as `mceInsertNewLine` #TINY-8458
+- New `newline_behavior` option to control what happens when the enter key is pressed or the `mceInsertNewLine` command is used #TINY-8458
 - New `transparent` property for `iframe` dialog component #TINY-8534
 - New `removeAttributeFilter` and `removeNodeFilter` functions to the `DomParser` and DOM `Serializer` APIs #TINY-7847
 - New `iframe_template_callback` option in the `media` plugin. Patch provided by Namstel #TINY-8684
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -51,6 +51,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Using Ctrl+Shift+Home/End (Cmd+Shift+Up/Down on Mac) would not expand the selection when a `contenteditable="false"` element was at the edge of content #TINY-7795
 - `<pre>` tags were not preserved when copying and pasting content #TINY-7719
 - Preview and Insert Template dialogs display the correct content background color when using dark skins #TINY-8534
+- `InsertLineBreak` command did not replace selected content #TINY-8458
 
 ## 6.0.3 - 2022-05-25
 
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -51,7 +51,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Using Ctrl+Shift+Home/End (Cmd+Shift+Up/Down on Mac) would not expand the selection when a `contenteditable="false"` element was at the edge of content #TINY-7795
 - `<pre>` tags were not preserved when copying and pasting content #TINY-7719
 - Preview and Insert Template dialogs display the correct content background color when using dark skins #TINY-8534
-- `InsertLineBreak` command did not replace selected content #TINY-8458
+- The `InsertLineBreak` command did not replace selected content #TINY-8458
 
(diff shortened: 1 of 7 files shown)


modules/tinymce/src/core/main/ts/api/commands/NewlineCommands.ts
```diff
@@ -1,19 +1,20 @@
+import * as InsertBlock from '../../newline/InsertBlock';
 import * as InsertBr from '../../newline/InsertBr';
 import * as InsertNewLine from '../../newline/InsertNewLine';
 import Editor from '../Editor';
 
 export const registerCommands = (editor: Editor): void => {
   editor.editorCommands.addCommands({
     insertParagraph: () => {
-      InsertNewLine.insert(editor);
+      InsertNewLine.insertBreak(InsertBlock, editor);
```
lnewson, line 9: Using a module as an object has caused various issues before, so I really don't like seeing us do that here. For example, Alloy used to do this and then was using `Obj` with it and that worked until Rollup fixed a bug where it wasn't following the standard (basically modules shouldn't have a prototype) and this caused things to die.

Can we come up with another way of passing the insert function and fake event name?

lnewson (changes requested):
Overall looks good, but using the module import/export as an object is quite dangerous so I'd like to see that changed.

modules/tinymce/src/core/main/ts/api/commands/NewlineCommands.ts
```diff
@@ -1,19 +1,20 @@
...
     },
 
     mceInsertNewLine: (_command, _ui, value) => {
       InsertNewLine.insert(editor, value);
     },
 
-    InsertLineBreak: (_command, _ui, value) => {
-      InsertBr.insert(editor, value);
+    InsertLineBreak: (_command, _ui, _value) => {
+      InsertNewLine.insertBreak(InsertBr, editor);
```
lnewson, line 17: Also just a though... do we need a `Fixed` changelog entry for this?

modules/tinymce/src/core/main/ts/api/commands/NewlineCommands.ts
TheSpyder: does `export const {}` work?

modules/tinymce/src/core/main/ts/api/commands/NewlineCommands.ts
TheSpyder: Ah probably, yes

modules/tinymce/src/core/main/ts/api/commands/NewlineCommands.ts
lnewson: yeah that should work fine 👍

modules/tinymce/CHANGELOG.md
```diff
@@ -51,6 +51,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Using Ctrl+Shift+Home/End (Cmd+Shift+Up/Down on Mac) would not expand the selection when a `contenteditable="false"` element was at the edge of content #TINY-7795
 - `<pre>` tags were not preserved when copying and pasting content #TINY-7719
 - Preview and Insert Template dialogs display the correct content background color when using dark skins #TINY-8534
+- `InsertLineBreak` command did not replace selected content #TINY-8458
```
lnewson, line 54:
```suggestion
- The `InsertLineBreak` command did not replace selected content #TINY-8458
```
Maybe?

lnewson (approved):

modules/tinymce/src/core/main/ts/newline/InsertNewLine.ts
```diff
@@ -5,53 +5,56 @@ import { getNewlineBehavior } from '../api/Options';
...
-  const insertBreak = (breakType: (editor: Editor, evt?: EditorEvent<KeyboardEvent>) => void, fakeEvent: string) => () => {
-    if (!editor.selection.isCollapsed()) {
-      execDeleteCommand(editor);
-    }
-    if (Type.isNonNullable(evt)) {
-      const event = fireFakeBeforeInputEvent(editor, fakeEvent);
-      if (event.isDefaultPrevented()) {
-        return;
-      }
+const insertBreak = (breakType: {
```
spocke, line 12: Should this be it's own type a BreakType i guess it would have to be a separate module since three modules uses that type. I just tend to want to use named types for things since it is easier when you refactor but maybe overkill here.

spocke (approved):

MitchCTiny (approved):

modules/tinymce/src/core/main/ts/newline/InsertNewLine.ts
TheSpyder: We wouldn't need to make all 3 modules use the type, it would just be structurally equal as it is today using an anonymous type. I wasn't sure which style we prefer but thought I'd give this a go for now.

PR #7885, opened 2022-06-09
TINY-8588: CEF elements wrapped in elements were not accounted for

HAFRMO:
Related Ticket:

Description of Changes:
Replication case simplied removed the wrapping element, which prevents things from working. This is an attempt to fix this oversight. No changelog added, 8588 not yet released.

Pre-checks:
* Changelog entry added

Review:
* [x] Milestone set


modules/tinymce/src/core/main/ts/keyboard/Nbsps.ts
```diff
@@ -58,11 +58,20 @@ const isAtLineBoundary = (root: SugarElement, pos: CaretPosition) => (
...
+};
 
-const isAfterCefBlock = (root: SugarElement, pos: CaretPosition) =>
-  pos.isAtStart() && isCefBlock(new DomTreeWalker(pos.container(), root.dom).prev2());
+const isAfterCefBlock = (root: SugarElement, pos: CaretPosition) => {
+  const prevPos = CaretFinder.prevPosition(root.dom, pos).getOr(pos);
+  const prevElement = isCefBlock(new DomTreeWalker(pos.container(), root.dom).prev2());
+  const prevExpectedPos = isCefBlock(new DomTreeWalker(prevPos.container(), root.dom).prev2());
+  return pos.isAtStart() && (prevElement || prevExpectedPos);
+};
```
lnewson, line 74: Maybe something like this to remove some of the duplication/eager evaluation?
```suggestion
const isSiblingCefBlock = (root: Element, direction: 'next' | 'prev2') => (container: Node) =>
  isCefBlock(new DomTreeWalker(container, root)[direction]());

// Check the next/previous element in case it is a cef and the next/previous caret position then would skip it, then check
// the next next/previous caret position ( for example in case the next element is a strong, containing a cef ).
const isBeforeCefBlock = (root: SugarElement, pos: CaretPosition) => {
  const nextPos = CaretFinder.nextPosition(root.dom, pos).getOr(pos);
  const isNextCefBlock = isSiblingCefBlock(root.dom, 'next');
  return pos.isAtEnd() && (isNextCefBlock(pos.container()) || isNextCefBlock(nextPos.container()));
};

const isAfterCefBlock = (root: SugarElement, pos: CaretPosition) => {
  const prevPos = CaretFinder.prevPosition(root.dom, pos).getOr(pos);
  const isPrevCefBlock = isSiblingCefBlock(root.dom, 'prev2');
  return pos.isAtStart() && (isPrevCefBlock(pos.container()) || isPrevCefBlock(prevPos.container()));
};
```

lnewson (approved):

modules/tinymce/src/core/main/ts/keyboard/Nbsps.ts
```diff
@@ -58,11 +58,21 @@ const isAtLineBoundary = (root: SugarElement, pos: CaretPosition) => (
...
+  isCefBlock(new DomTreeWalker(container, root)[direction]());
+
+// Check the next/previous element in case it is a cef and the next/previous caret position then would skip it, then check
+// the next next/previous caret position ( for example in case the next element is a strong, containing a cef ).
+const isBeforeCefBlock = (root: SugarElement, pos: CaretPosition) => {
+  const nextPos = CaretFinder.nextPosition(root.dom, pos).getOr(pos);
+  const isNextCefBlock = isSiblingCefBlock(root.dom, 'next');
+  return pos.isAtEnd() && (isNextCefBlock(pos.container()) || isNextCefBlock(nextPos.container()));
+};
+const isAfterCefBlock = (root: SugarElement, pos: CaretPosition) => {
```
lnewson, line 71:
```suggestion

const isAfterCefBlock = (root: SugarElement, pos: CaretPosition) => {
```

modules/tinymce/src/core/test/ts/browser/keyboard/SpaceKeyTest.ts
```diff
@@ -109,5 +109,13 @@ describe('browser.tinymce.core.keyboard.SpaceKeyTest', () => {
       editor.selection.setContent(' ');
       TinyAssertions.assertContent(editor, '<p><span class="filler">s&nbsp;</span><span style="display: block;" contenteditable="false">a</span></p>');
     });
+
+    it('TINY-8588: Add one space before a block inside a strong', () => {
```
spocke, line 113: Since the after is also change should there be a test for that as well as the before test. Seems that there are only tests for the before case on all of these should we have them on after the cef.

- [ ] Consider adding a test pressing space after the cef element

spocke (approved):

modules/tinymce/src/core/test/ts/browser/keyboard/SpaceKeyTest.ts
```diff
@@ -93,6 +93,14 @@ describe('browser.tinymce.core.keyboard.SpaceKeyTest', () => {
       TinyAssertions.assertContent(editor, '<p>s&nbsp;<span style="display: block;" contenteditable="false">a</span></p>');
     });
 
+    it('TINY-8588: Add one space just after a block', () => {
+      const editor = hook.editor();
+      editor.setContent('<p><span style="display: block;" contenteditable="false">a</span>s</p>');
+      TinySelections.setCursor(editor, [ 0, 1 ], 0);
+      editor.selection.setContent(' ');
```
lnewson, line 100: I must have missed this before, but why are we doing `editor.selection.setContent(' ')` here? Don't we want to be pressing the space key here instead since that is the bug and a more direct use case of `Nbsp.ts`?

lnewson (changes requested):
Marking as needs changes, as the last commit just deleted a massive block that shouldn't have been done from what I can tell.

HAFRMO:
> Marking as needs changes, as the last commit just deleted a massive block that shouldn't have been done from what I can tell.

It was deleted on purpose. There's a fun second bug that'll require it's own JIRA, as the solution will need to be different, but which prevents the usage of the now-deleted stuff. If it's needed it should be added there, instead of here.

lnewson:
> It was deleted on purpose. There's a fun second bug that'll require it's own JIRA, as the solution will need to be different, but which prevents the usage of the now-deleted stuff. If it's needed it should be added there, instead of here.

There's absolutely no details about this anywhere... please provide the details on the Jira or somewhere visible.

HAFRMO:
> There's absolutely no details about this anywhere... please provide the details on the Jira or somewhere visible.

An issue has been added to the JIRA with a second issue to solve this second bug.

lnewson:
> An issue has been added to the JIRA with a second issue to solve this second bug.

Thank you for that, however you've get test failures so I can't approve while those are failing.

lnewson (changes requested):
Just marking this as needs changes because of the test failures I mentioned in my previous comment and to get it off my review board while waiting for that to be looked into.

lnewson (approved):

PR #7892, opened 2022-06-10
TINY-8729: Delete is broken when the selection contains CEF at the edge of the content

vpyshnenko:
Related Ticket: TINY-8729

Description of Changes:
* Overrides default deletion when the selection contains cef at the start or the end of the document and explicitly removes selected content

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/delete/CefDelete.ts
+++ b/modules/tinymce/src/core/main/ts/delete/CefDelete.ts
@@ -6,7 +6,6 @@ import CaretPosition from '../caret/CaretPosition';
 import * as CefUtils from '../dom/CefUtils';
 import * as NodeType from '../dom/NodeType';
 import { isCefAtEdgeSelected } from '../keyboard/CefUtils';
-import { deleteRng } from '../textpatterns/utils/Utils';
 import * as CefDeleteAction from './CefDeleteAction';
 import * as DeleteElement from './DeleteElement';
 import * as DeleteUtils from './DeleteUtils';
@@ -82,7 +81,7 @@ const backspaceDeleteRange = (editor: Editor, forward: boolean): Optional<() => 
 
   if (isCefAtEdgeSelected(editor)) {
     return Optional.some(() => {
-      deleteRng(editor.dom, editor.selection.getRng(), (elm) => elm === editor.getBody());
+      DeleteUtils.deleteRangeContents(editor, editor.selection.getRng(), SugarElement.fromDom(editor.getBody()));
     });
   }
   return Optional.none();
--- a/modules/tinymce/src/core/main/ts/delete/CefDelete.ts
+++ b/modules/tinymce/src/core/main/ts/delete/CefDelete.ts
@@ -5,6 +5,8 @@ import Editor from '../api/Editor';
 import CaretPosition from '../caret/CaretPosition';
 import * as CefUtils from '../dom/CefUtils';
 import * as NodeType from '../dom/NodeType';
+import { isCefAtEdgeSelected } from '../keyboard/CefUtils';
+import { deleteRng } from '../textpatterns/utils/Utils';
 import * as CefDeleteAction from './CefDeleteAction';
 import * as DeleteElement from './DeleteElement';
 import * as DeleteUtils from './DeleteUtils';
@@ -78,6 +80,11 @@ const backspaceDeleteRange = (editor: Editor, forward: boolean): Optional<() => 
     );
   }
 
+  if (isCefAtEdgeSelected(editor)) {
+    return Optional.some(() => {
+      deleteRng(editor.dom, editor.selection.getRng(), (elm) => elm === editor.getBody());
+    });
+  }
   return Optional.none();
 };
 
--- a/modules/tinymce/src/core/main/ts/delete/DeleteUtils.ts
+++ b/modules/tinymce/src/core/main/ts/delete/DeleteUtils.ts
@@ -1,11 +1,13 @@
-import { Optional, Optionals } from '@ephox/katamari';
-import { Compare, PredicateFind, SugarElement } from '@ephox/sugar';
+import { Arr, Optional, Optionals } from '@ephox/katamari';
+import { Compare, PredicateFind, Remove, SugarElement, SugarNode, Traverse } from '@ephox/sugar';
 
 import Editor from '../api/Editor';
 import { EditorEvent } from '../api/util/EventDispatcher';
 import * as CaretFinder from '../caret/CaretFinder';
 import CaretPosition from '../caret/CaretPosition';
 import { isListItem, isTextBlock } from '../dom/ElementType';
+import * as Empty from '../dom/Empty';
+import * as PaddingBr from '../dom/PaddingBr';
 import * as InlineUtils from '../keyboard/InlineUtils';
 
 const execCommandIgnoreInputEvents = (editor: Editor, command: string) => {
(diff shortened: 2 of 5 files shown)


lnewson:
@vpyshnenko a reminder you need to add your team to the reviewers list now. I've done it this time.

modules/tinymce/src/core/main/ts/delete/CefDelete.ts
```diff
@@ -78,6 +80,11 @@ const backspaceDeleteRange = (editor: Editor, forward: boolean): Optional<() =>
     );
   }
 
+  if (isCefAtEdgeSelected(editor)) {
+    return Optional.some(() => {
+      deleteRng(editor.dom, editor.selection.getRng(), (elm) => elm === editor.getBody());
```
lnewson, line 85: This is only intended to be used for `textpattern` as it removes all content and doesn't pad, since the text pattern will replace it (which is also why it's in the `textpatterns` directory). This isn't a viable option here and we need to use something like what's used in `TableDelete` (see `deleteCellContents`) which will both clean up and pad the empty element. A good test case here would be something like a CEF element inside a list item.

Anyways, I'd suggest we move and rename that logic from `TableDelete` to a more shared space and that should be able to be used here I believe.

lnewson (changes requested):
As noted the function used isn't appropriate here and if we were to use it, it'd need to be moved. So can you please look at fixing that as noted and can you also can you please make sure to add some tests for inline cef elements.

modules/tinymce/src/core/main/ts/delete/DeleteUtils.ts
```diff
@@ -67,7 +69,35 @@ const willDeleteLastPositionInElement = (forward: boolean, fromPos: CaretPositio
...
 
+const freefallRtl = (root: SugarElement<Node>): Optional<SugarElement<Node>> => {
+  const child = SugarNode.isComment(root) ? Traverse.prevSibling(root) : Traverse.lastChild(root);
+  return child.bind(freefallRtl).orThunk(() => Optional.some(root));
+};
+
+const deleteRangeContents = (editor: Editor, rng: Range, root: SugarElement<HTMLElement>, moveSelection: boolean = true): void => {
+  rng.deleteContents();
+  // Pad the last block node
+  const lastNode = freefallRtl(root).getOr(root);
```
lorenzo-pomili, line 80: Could we put the `getOr` in `freefallRtl` or avoid the `.orThunk`?

spocke (approved):
Looks good Viktor.

modules/tinymce/src/core/main/ts/delete/DeleteUtils.ts
vpyshnenko: `getOr` is needed here to actually extract the node
`orThunk` is for handling the haltering case of recursion in `freefallRtl `

lorenzo-pomili (approved):

modules/tinymce/src/core/test/ts/browser/delete/CefDeleteTest.ts
```diff
@@ -140,4 +141,84 @@ describe('browser.tinymce.core.delete.CefDeleteTest', () => {
       })
     );
   });
+
+  context('cef block is at the start/end of the content and covered with the selection', () => {
+    Arr.each([ Keys.delete, Keys.backspace ], (ks) => {
```
lnewson, line 146: This is going to be problematic when debugging, as you're not going to know where it's a delete or backspace that failed. As such, we probably need to turn this into an array of objects like the following and then replace `delete` in each `it` block with the `label`.
```
[
  { label: 'Delete', key: Keys.delete },
  { label: 'Backspace', key: Keys.backspace }
]
```

- [ ] Ensure the tests have some way to report/log what key is actually pressed

modules/tinymce/src/core/test/ts/browser/delete/CefDeleteTest.ts
```diff
@@ -140,4 +141,84 @@ describe('browser.tinymce.core.delete.CefDeleteTest', () => {
...
+        TinySelections.setSelection(editor, [ 0 ], 0, [ 0 ], 4);
+        TinyContentActions.keystroke(editor, ks());
+        TinyAssertions.assertCursor(editor, [ 0 ], 0);
+        TinyAssertions.assertContent(editor, '');
+      });
+    });
+  });
+
+  context('Cleaning up after rng.deleteContens call', () => {
+    Arr.each([ Keys.backspace, Keys.delete ], (ks) => {
```
lnewson, line 212: Same issue here as above.

modules/tinymce/src/core/test/ts/browser/delete/CefDeleteTest.ts
```diff
@@ -140,4 +141,84 @@ describe('browser.tinymce.core.delete.CefDeleteTest', () => {
...
+        // actual content: <p data-mce-caret="before"></p><p contenteditable="false">CEF</p><p>abc</p><p contenteditable="false">CEF</p>
+        TinySelections.setSelection(editor, [ 0 ], 0, [], 4);
+        TinyContentActions.keystroke(editor, ks());
+        TinyAssertions.assertContent(editor, '');
+      });
+    });
+  });
+
+  context('inline cef is at the start/end of the content and covered with the selection', () => {
+    Arr.each([ Keys.delete, Keys.backspace ], (ks) => {
```
lnewson, line 178: Same issue here as above.

modules/tinymce/src/core/test/ts/browser/delete/CefDeleteTest.ts
```diff
@@ -140,4 +141,84 @@ describe('browser.tinymce.core.delete.CefDeleteTest', () => {
...
+        TinyContentActions.keystroke(editor, ks());
+        TinyAssertions.assertCursor(editor, [ 0 ], 0);
+        TinyAssertions.assertContent(editor, '');
+      });
+    });
+  });
+
+  context('Cleaning up after rng.deleteContens call', () => {
+    Arr.each([ Keys.backspace, Keys.delete ], (ks) => {
+      it('TINY-8729: should clean up empty nodes and padd empty block with bogus br', () => {
```
lnewson, line 213: Can we please add one for a list here with an inline cef as well? As that's a somewhat more complicated case as I mentioned in my previous review, or will that never trigger this logic?

For example, something like:
```
<ul><li><span contenteditable="false">CEF</span></li></ul>
```
which should end up with something like:
```
<ul><li><br data-mce-bogus="1"></li></ul>
```

- [ ] Add list test case unless it's impossible for this logic to run into that.

lnewson (approved):
This looks much better, thanks @vpyshnenko 👍

modules/tinymce/src/core/test/ts/browser/delete/CefDeleteTest.ts
```diff
@@ -140,4 +147,94 @@ describe('browser.tinymce.core.delete.CefDeleteTest', () => {
...
+        const editor = hook.editor();
+        editor.setContent('<p><strong>W</strong>elcom<strong>e</strong></p><p contenteditable="false">CEF</p>');
+        TinySelections.setSelection(editor, [ 0, 0, 0 ], 0, [ ], 2);
+        TinyContentActions.keystroke(editor, key());
+        // it should remove empty  <p><strong></strong></p> nodes
+        TinyAssertions.assertRawContent(editor, '<p><br data-mce-bogus="1"></p>');
+        TinyAssertions.assertCursor(editor, [ 0 ], 0);
+      });
+
+      it(`TINY-8729-4: should delete selected content and padd empty list item with bogus br when ${label} is pressed`, () => {
```
lnewson, line 229:
```suggestion
      it(`TINY-8729: should delete selected content and padd empty list item with bogus br when ${label} is pressed`, () => {
```

PR #7901, opened 2022-06-15
TINY-8802: Set up for strict Typescript check

tiny-ben-tran:
Related Ticket:  TINY-8802

Description of Changes:
- Added a `tsconfig.strictTypeCheck.json` file
- Enabled strict type checking in the `ci` command. Placed it before the regular compiler because it falls through when there is nothing new

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [x] ~Milestone set~

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,8 @@
     "headless-test-firefox": "yarn -s grunt headless-auto --bedrock-browser=firefox-headless",
     "headless-test-manual": "yarn -s grunt headless-manual",
     "bedrock": "bedrock --customRoutes modules/tinymce/src/core/test/json/routes.json",
-    "test-one": "yarn tsc && yarn bedrock-auto -b chrome-headless -f"
+    "test-one": "yarn tsc && yarn bedrock-auto -b chrome-headless -f",
+    "strict-null-checks": "tsc -b tsconfig.strictNullChecks.json"
   },
   "devDependencies": {
     "@ephox/bedrock-client": "^13.0.0",
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "start": "yarn -s tinymce-grunt start",
     "dev": "npm-run-all oxide-icons-build oxide-build -p tsc \"tinymce-grunt dev\"",
     "ci": "npm-run-all -p oxide-ci oxide-icons-ci -p tsc \"tinymce-grunt dev\"",
-    "ci-all": "npm-run-all -p eslint -p strict-type-check ci -s tinymce-rollup",
+    "ci-all": "npm-run-all -p eslint ci -s strict-type-check -s tinymce-rollup",
     "local-ci": "npm-run-all ci-all -s headless-test browser-test",
     "test": "run-s local-ci",
     "build": "npm-run-all -p oxide-icons-ci oxide-ci -s tinymce-grunt",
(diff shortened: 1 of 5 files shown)


spocke (approved):

lnewson (changes requested):
This isn't what's needed here sorry Ben. We need something that runs during normal development so as changes are made it doesn't regress and it needs to only run on a subset for files (for example just core, not plugins). Feel free to DM me if you want to chat more.

TheSpyder:
huh. Maybe I misread the blog post. Ben has actually followed it exactly.

tsconfig.strictTypeCheck.json
```diff
@@ -0,0 +1,10 @@
+{
+  "extends": "./modules/tinymce/tsconfig.json",
+  "compilerOptions": {
+    "strict": true,
+    "noEmit": true
+  },
+  "include": [
+    "modules/tinymce/src/core/main/ts/ui"
```
tiny-ben-tran, line 8: Added this here because Typescript complains if there is no input

TheSpyder: yes it doesn't surprised me we need to have at least one folder passing as strict to start using this config in CI.

package.json
```diff
@@ -19,7 +19,7 @@
     "eslint": "lerna exec --stream --no-bail -- eslint --max-warnings=0 src/**/*.ts",
     "start": "yarn -s tinymce-grunt start",
     "dev": "npm-run-all oxide-icons-build oxide-build -p tsc \"tinymce-grunt dev\"",
-    "ci": "npm-run-all -p oxide-ci oxide-icons-ci -p tsc \"tinymce-grunt dev\"",
+    "ci": "npm-run-all strict-type-check -p oxide-ci oxide-icons-ci -p tsc \"tinymce-grunt dev\"",
```
lnewson, line 22: This will run on actual browser nodes as well. If you might recall, in the call today @TheSpyder mentioned it should really only run on the compile node. That means this needs to be in `ci-all` instead of `ci`. It also means we might need to add it to `local-ci` as well.

tsconfig.strictTypeCheck.json
```diff
@@ -0,0 +1,10 @@
+{
```
lnewson, line 1: This is still in the wrong location... this file should be inside the `modules/tinymce` directory since it only applies to that. I'd also suggest we just shorten the name to `tsconfig.strict.json`.

package.json
lnewson: @tiny-ben-tran with the latest commit that means it's no longer running locally when running `yarn test`. As I mentioned above, you'll likely need to add it to `local-ci` as well to fix that.

tiny-ben-tran: Don't need to add the strict type check to `local-ci` as it already calls `ci-all`

package.json
```diff
@@ -20,7 +20,7 @@
     "start": "yarn -s tinymce-grunt start",
     "dev": "npm-run-all oxide-icons-build oxide-build -p tsc \"tinymce-grunt dev\"",
     "ci": "npm-run-all -p oxide-ci oxide-icons-ci -p tsc \"tinymce-grunt dev\"",
-    "ci-all": "npm-run-all -p eslint ci -s tinymce-rollup",
+    "ci-all": "npm-run-all -p eslint -p strict-type-check ci -s tinymce-rollup",
```
lnewson, line 23: This likely isn't what we want... we want `eslint` to run alongside `ci` not by itself. Additionally, this may fail without the oxide build happening first. I think this is what we'd want instead:
```suggestion
    "ci-all": "npm-run-all -p eslint ci -s strict-type-check tinymce-rollup",
```

package.json
lnewson: Ahh yeah you're right sorry 🤦

package.json
tiny-ben-tran: I knew something wasn't right there

lnewson (approved):

PR #7903, opened 2022-06-15
TINY-8698: Allow blocks to be annotated directly

ltrouton:
Related Ticket: TINY-8698

Description of Changes:
* Update the annotator so that it is able to annotate directly on blocks and make sure all of the annotator APIs now work with this update
* Updated the `comments.less` file so that it adds a yellow outline to blocks that have been commented on

Pre-checks:
* [X] Changelog entry added

Review:
* [X] Milestone set


modules/oxide/src/less/theme/content/comments/comments.less
```diff
@@ -2,13 +2,46 @@
...
+  .tox-comment[contenteditable="false"][data-mce-annotation-active="true"] {
+    &:not([data-mce-selected]) {
+      outline: 3px solid @document-comment-active-outline-color;
+    }
+  }
+
+  span.tox-comment[data-mce-annotation-active="true"] {
+    & img,
+    & > audio,
+    & > video {
```
lnewson, line 32: Does this also need `span.mce-preview-object` to be included in this list?

modules/tinymce/CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - New `transparent` property for `iframe` dialog component #TINY-8534
 - New `removeAttributeFilter` and `removeNodeFilter` functions to the `DomParser` and DOM `Serializer` APIs #TINY-7847
 - New `iframe_template_callback` option in the `media` plugin. Patch provided by Namstel #TINY-8684
+- Support annotating block elements directly when using the editor annotator #TINY-8698
```
lnewson, line 15: Should this be in `Improved`? Also we aren't supporting all block elements, only specific ones so we probably need to be a little clearer about that...

modules/tinymce/src/core/main/ts/annotate/AnnotationFilter.ts
```diff
@@ -6,17 +6,42 @@ import { AnnotationsRegistry, AnnotatorSettings } from './AnnotationsRegistry';
...
+    node.attr(Markings.dataAnnotation(), null);
+    node.attr(Markings.dataAnnotationActive(), null);
+
+    const customAttrNames = Optional.from(node.attr(Markings.dataAnnotationAttributes())).map((names) => names.split(',')).getOr([]);
+    const customClasses = Optional.from(node.attr(Markings.dataAnnotationClasses())).map((names) => names.split(',')).getOr([]);
+    Arr.each(customAttrNames, (name) => node.attr(name, null));
+
+    const classList = new Set(node.attr('class')?.split(' ') ?? []);
+    Arr.each([ Markings.annotation() ].concat(customClasses), (clazz) => classList.delete(clazz));
+    const newClassList = Array.from(classList.values());
```
lnewson, line 24: This can be simplified to this I believe?
```suggestion
    const classList = node.attr('class')?.split(' ') ?? [];
    const newClassList = Arr.difference(classList, [ Markings.annotation() ].concat(customClasses));
```

modules/tinymce/src/core/main/ts/annotate/Identification.ts
```diff
@@ -42,19 +45,31 @@ const identify = (editor: Editor, annotationName: Optional<string>): Optional<{u
 
 const isAnnotation = (elem: any): boolean => SugarNode.isElement(elem) && Class.has(elem, Markings.annotation());
 
+const offscreenSelectionClass = 'mce-offscreen-selection';
```
lnewson, line 48: This might be better at the top, as constants normally are at the top of the file instead of hidden in the middle somewhere?

modules/tinymce/src/core/main/ts/annotate/Identification.ts
```diff
@@ -42,19 +45,31 @@ const identify = (editor: Editor, annotationName: Optional<string>): Optional<{u
...
+
+const isOffscreenSelectionElement = (elem: SugarElement<Node>) =>
+  Traverse.parent(elem).exists((parent) => Class.has(parent, offscreenSelectionClass));
+
 const findMarkers = (editor: Editor, uid: string): Array<SugarElement<Element>> => {
   const body = SugarElement.fromDom(editor.getBody());
-  return SelectorFilter.descendants(body, `[${Markings.dataAnnotationId()}="${uid}"]`);
+  const descendants = SelectorFilter.descendants(body, `[${Markings.dataAnnotationId()}="${uid}"]`);
+  const offscreenSelection = hasOffscreenSelection(body);
+  return offscreenSelection ? Arr.filter(descendants, (descendant) => !isOffscreenSelectionElement(descendant)) : descendants;
```
lnewson, line 59: This seems like a code smell to me... as this is one limited edge case and I think instead we need to be ignoring bogus elements in general not just the offscreen selection element.

- [ ] Swap to excluding `data-mce-bogus` elements instead (be careful of `all` vs `1`).

modules/tinymce/src/themes/silver/test/ts/browser/editor/header/StickyHeaderInitialPlacementTest.ts
```diff
@@ -26,7 +27,9 @@ describe('browser.tinymce.themes.silver.editor.header.StickyHeaderInitialPlaceme
         const isToolbarTop = test.location === ToolbarLocation.top;
         await StickyUtils.pAssertHeaderDocked(isToolbarTop);
       }
-      StickyUtils.assertEditorClasses(test.expectDocked);
+      await Waiter.pTryUntil('should have correct classes', () => {
+        StickyUtils.assertEditorClasses(test.expectDocked);
+      });
```
lnewson, line 32: Don't forget to revert this since it didn't solve the issues.

modules/tinymce/src/core/main/ts/api/Annotator.ts
```diff
@@ -83,7 +85,13 @@ const Annotator = (editor: Editor): Annotator => {
...
       const bookmark = editor.selection.getBookmark();
       identify(editor, Optional.some(name)).each(({ elements }) => {
-        Arr.each(elements, Remove.unwrap);
+        Arr.each(elements, (element) => {
+          if (isSpan(element)) {
+            Remove.unwrap(element);
+          } else {
+            removeDirectAnnotation(element);
+          }
+        });
```
lnewson, line 94: This appears to be duplicated with the below `removeAll` logic. Can we extract it out into a single function?

modules/oxide/src/less/theme/content/comments/comments.less
ltrouton: Yeah, it should be. Thanks for that. I am working on adding a test to cover that

modules/tinymce/CHANGELOG.md
ltrouton: Yeah, I was tossing between the two when creating the entry. I think if it supported all blocks or was more general then maybe it would make more sense to be in `Added`. Although, given it is quite specific and not integrator configurable, then `Improved` might be safer/better.

I am happy to move to `Improved` and fix the message to say certain blocks.

modules/tinymce/src/core/main/ts/annotate/AnnotationFilter.ts
ltrouton: Yep, that should work. Thanks

modules/tinymce/src/core/main/ts/api/Annotator.ts
ltrouton: I have refactored it out. Thanks for the suggestion

modules/oxide/src/less/theme/content/comments/comments.less
ltrouton: Done

modules/tinymce/CHANGELOG.md
ltrouton: I have moved it over

modules/tinymce/src/core/main/ts/annotate/Identification.ts
ltrouton: I could be misunderstanding but were you thinking of something like this?

```
const isBogusElement = (elem: SugarElement<Element>, root: SugarElement<Node>) =>
  Attribute.has(elem, 'data-mce-bogus') || SelectorExists.ancestor(elem, '[data-mce-bogus="all"]', isRoot(root));

const findMarkers = (editor: Editor, uid: string): Array<SugarElement<Element>> => {
  const body = SugarElement.fromDom(editor.getBody());
  const descendants = SelectorFilter.descendants(body, `[${Markings.dataAnnotationId()}="${uid}"]`);
  return Arr.filter(descendants, (descendant) => !isBogusElement(descendant, body)) ;
};
```

lnewson: Yeah that's basically what I was thinking here.

ltrouton: Thanks. I will make the relevant updates

ltrouton: Have pushed up change. Thanks for the suggestion for a more general solution

modules/oxide/src/less/theme/content/comments/comments.less
```diff
@@ -2,13 +2,47 @@
...
-  background-color: @document-comment-background-color;
-}
+.tox-comments-visible {
+  .tox-comment[contenteditable="false"],
+  span.tox-comment img,
+  span.tox-comment > audio,
+  span.tox-comment > video,
+  span.tox-comment span.mce-preview-object {
+    &:not([data-mce-selected]) {
+      outline: 3px solid @document-comment-outline-color;
```
lnewson, line 19: Should this width and the ones below be using the `@object-block-selected-outline-width` LESS variable so it keeps it as the same size as the selection? Either that or maybe we should have this at the top as a new variable? e.g
```
@document-comment-outline-width: @object-block-selected-outline-width;
```

modules/tinymce/CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Improved
 - Clearer focus states for buttons while navigating with keyboard #TINY-8557
+- Support annotating certain block elements directly when using the editor annotator #TINY-8698
```
```suggestion
- Support annotating certain block elements directly when using the editor annotator API #TINY-8698
```
Maybe?

modules/tinymce/src/core/main/ts/annotate/Identification.ts
```diff
@@ -42,19 +45,25 @@ const identify = (editor: Editor, annotationName: Optional<string>): Optional<{u
 
 const isAnnotation = (elem: any): boolean => SugarNode.isElement(elem) && Class.has(elem, Markings.annotation());
 
+const isBogusElement = (elem: SugarElement<Node>, root: SugarElement<Node>) =>
+  Attribute.has(elem, 'data-mce-bogus') || SelectorExists.ancestor(elem, '[data-mce-bogus="all"]', isRoot(root));
```
lnewson, line 49: Should this be using `closest` incase the element itself has `data-mce-bogus="all"`?
```suggestion
  Attribute.has(elem, 'data-mce-bogus') || SelectorExists.closest(elem, '[data-mce-bogus="all"]', isRoot(root));
```

modules/tinymce/src/core/test/ts/browser/annotate/AnnotationPersistenceTest.ts
```diff
@@ -49,59 +50,82 @@ describe('browser.tinymce.core.annotate.AnnotationPersistenceTest', () => {
...
     TinyAssertions.assertContentPresence(editor, {
-      '.mce-annotation': 1,
       'p:contains("This is the only paragraph here")': 1
     });
   };
 
-  const contentContains = (editor: Editor, pattern: string, isContained: boolean) => {
+  const setupSingleBlockAnnotation = (editor: Editor) => {
+    editor.setContent(`<pre class="language-markup" contenteditable="false">test</pre>`);
+    editor.undoManager.add();
```
lnewson, line 81:
```suggestion
    editor.resetContent(`<pre class="language-markup" contenteditable="false">test</pre>`);
```
This might be better here as it'll make sure the undo state is cleared as well?

modules/tinymce/src/core/test/ts/browser/annotate/AnnotationStylingTest.ts
```diff
@@ -0,0 +1,311 @@
...
+
+        Arr.each(parents, (e) => assert.deepEqual(getOutline(e), noOutline, 'parent should not have outline'));
+        Arr.each(children, (e) => assert.deepEqual(getOutline(e), isFigCaption(e) ? emptyFigCaptionOutline : noOutline, 'child should not have outline'));
+      }
+    });
+
+  before(() => {
+    const editor = hook.editor();
+    Class.add(TinyDom.body(editor), 'tox-comments-visible');
+  });
```
lnewson, line 106: This should be up the top near the other hooks to keep everything grouped (e.g. `TinyHook.bddSetup`)

lnewson (approved):

modules/oxide/src/less/theme/content/comments/comments.less
ltrouton: That's a good idea. Will make the change

modules/tinymce/src/core/main/ts/annotate/Identification.ts
ltrouton: Doesn't having `Attribute.has(elem, 'data-mce-bogus')` before check that? I guess it can't hurt though

modules/oxide/src/less/theme/content/comments/comments.less
ltrouton: Done

modules/tinymce/src/core/main/ts/annotate/Identification.ts
lnewson: No you're correct... I'd compartmentalised it incorrectly as the left side checked `data-mce-bogus="1"` only. Feel free to ignore then.

PR #7907, opened 2022-06-16
TINY-8602: Ensure choice menu items use the "menuitemradio" aria role

lnewson:
Related Ticket: TINY-8602

Description of Changes:

This fixes an accessibility issue whereby our choice menu items were using the `menuitemcheckbox` role when the options were exclusive. As part of this, to be able to tell when different menu items are toggled I've also added a `onToggled` callback to the Toggling configuration spec.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/alloy/src/main/ts/ephox/alloy/menu/build/ItemType.ts
+++ b/modules/alloy/src/main/ts/ephox/alloy/menu/build/ItemType.ts
@@ -24,6 +24,20 @@ const getItemRole = (detail: NormalItemDetail): ItemRole =>
     .map((toggling) => toggling.exclusive ? 'menuitemradio' : 'menuitemcheckbox')
     .getOr('menuitem');
 
+const getTogglingSpec = (tConfig: Partial<TogglingConfigSpec> & { exclusive?: boolean }): TogglingConfigSpec => ({
+  aria: {
+    mode: 'checked'
+  },
+  // Filter out the additional properties
+  ...Obj.filter(tConfig as { [K in keyof TogglingConfigSpec]: TogglingConfigSpec[K] }, (_value, name) => name !== 'exclusive'),
+  onToggled: (component, state) => {
+    if (Type.isFunction(tConfig.onToggled)) {
+      tConfig.onToggled(component, state);
+    }
+    ItemEvents.onToggled(component, state);
+  }
+});
+
 const builder = (detail: NormalItemDetail): AlloySpec => ({
   dom: detail.dom,
   domModification: {
@@ -41,18 +55,7 @@ const builder = (detail: NormalItemDetail): AlloySpec => ({
     detail.itemBehaviours,
     [
       // Investigate, is the Toggling.revoke still necessary here?
-      detail.toggling.fold(Toggling.revoke, (tConfig: TogglingConfigSpec) => Toggling.config({
-        aria: {
-          mode: 'checked'
-        },
-        ...Obj.filter(tConfig as { [K in keyof TogglingConfigSpec]: TogglingConfigSpec[K] }, (_value, name) => name !== 'exclusive'),
-        onToggled: (component, state) => {
-          if (Type.isFunction(tConfig.onToggled)) {
-            tConfig.onToggled(component, state);
-          }
-          ItemEvents.onToggled(component, state);
-        }
-      })),
+      detail.toggling.fold(Toggling.revoke, (tConfig) => Toggling.config(getTogglingSpec(tConfig))),
       Focusing.config({
         ignore: detail.ignoreFocus,
         // Rationale: because nothing is focusable, when you click
--- a/modules/alloy/src/main/ts/ephox/alloy/menu/build/ItemType.ts
+++ b/modules/alloy/src/main/ts/ephox/alloy/menu/build/ItemType.ts
@@ -1,4 +1,5 @@
 import { FieldProcessor, FieldSchema } from '@ephox/boulder';
+import { Obj, Type } from '@ephox/katamari';
 
 import * as AddEventsBehaviour from '../../api/behaviour/AddEventsBehaviour';
 import { Focusing } from '../../api/behaviour/Focusing';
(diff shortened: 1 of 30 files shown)


modules/alloy/src/main/ts/ephox/alloy/menu/build/ItemType.ts
```diff
@@ -37,7 +45,13 @@ const builder = (detail: NormalItemDetail): AlloySpec => ({
         aria: {
           mode: 'checked'
         },
-        ...tConfig
+        ...Obj.filter(tConfig as { [K in keyof TogglingConfigSpec]: TogglingConfigSpec[K] }, (_value, name) => name !== 'exclusive'),
```
ephox-mogran, line 48: This is a bit hard to unpack. Can we move it above `builder`?

modules/alloy/src/main/ts/ephox/alloy/ui/single/MenuSpec.ts
```diff
@@ -58,6 +60,23 @@ const make: CompositeSketchFactory<MenuDetail, MenuSpec> = (detail, components,
...
       Highlighting.highlight(menu, item);
+    }),
+
+    // Enforce only a single radio menu item is toggled by finding any other toggled
+    // radio menu items and untoggling them when a certain item is toggled
+    AlloyEvents.run<MenuItemToggledEvent>(ItemEvents.toggled(), (menu, simulatedEvent) => {
+      const { item, state } = simulatedEvent.event;
+      if (state) {
+        // TODO: TINY-8812 - This ideally should be done in a way such that a menu can have multiple radio groups.
+        const checkedRadioItems = SelectorFilter.descendants(menu.element, '[role="menuitemradio"][aria-checked="true"]');
```
ephox-mogran, line 71: Yeah, as you mention, it would be ideal to avoid this kind of DOM inspection. It probably looks like the kind of thing we should use Highlighting (with single select), but I think it already has another purpose (probably selection). I wonder if some sort of `Grouping` behaviour would be applicable here, but probably not. Anyway, I'm not giving you any better idea than you have already.

ephox-mogran: The main improvement I would suggest is just moving it outside of the main spec definition to a separate function.

modules/alloy/src/main/ts/ephox/alloy/ui/types/ItemTypes.ts
```diff
@@ -57,7 +57,7 @@ export interface NormalItemSpec {
   components?: AlloySpec[];
   dom: RawDomSchema;
   // INVESTIGATE: this might not be right
```
ephox-mogran, line 59: Do we still need to investigate this ? :)

modules/alloy/src/test/ts/browser/ui/dropdown/MenuRadioTest.ts
```diff
@@ -0,0 +1,167 @@
...
+
+    triggerFocusItem(beta);
+    Assertions.assertStructure(
+      'After focusing item on beta',
+      ApproxStructure.build((s, str, arr) => s.element('ol', {
+        classes: [
+          arr.has('test-menu')
+        ],
+        children: [
+          itemStructure(false, false)(s, str, arr),
```
ephox-mogran, line 114: Maybe make this one argument with two fields, just do it's a bit more self-documenting. Or can you use named parameters in TS? - that would be another option.

ephox-mogran (approved):
Apart from the JIRA you've already logged, this looks good enough. Professional work as always.

modules/alloy/src/main/ts/ephox/alloy/ui/single/MenuSpec.ts
lnewson: Yeah, I suspect we need to look at a `Grouping` or similar sketcher rather than a behaviour here.

modules/alloy/src/main/ts/ephox/alloy/ui/types/ItemTypes.ts
lnewson: No idea to be honest and isn't really something that's in scope of this task 😉 It's also why a TODO generally should never exist without a Jira for more context.

modules/alloy/src/test/ts/browser/ui/dropdown/MenuRadioTest.ts
lnewson: Yeah good point, have done that. Also no named parameters in TS unfortunately 😞

modules/alloy/src/main/ts/ephox/alloy/ui/single/MenuSpec.ts
lnewson: Have split it out and added a comment... hopefully that's what you meant.

modules/alloy/src/main/ts/ephox/alloy/menu/build/ItemType.ts
lnewson: Have split it out and added a comment... hopefully that's what you meant.

tiny-ben-tran (commented):
The changes look ok to me. I don't understand alloy enough to give this an approval

modules/alloy/src/main/ts/ephox/alloy/ui/types/ItemTypes.ts
ephox-mogran: I imagine I wrote that when looking through the types that had been added, and it didn't look quite right. So if the type is right, it can be removed.

lnewson: Removed these TODO statements as discussed on slack

ltrouton (approved):

modules/alloy/src/main/ts/ephox/alloy/ui/types/ItemTypes.ts
```diff
@@ -56,8 +56,7 @@ export interface NormalItemSpec {
   data: ItemDataTuple;
   components?: AlloySpec[];
   dom: RawDomSchema;
-  // INVESTIGATE: this might not be right
-  toggling?: Partial<TogglingConfigSpec>;
+  toggling?: Partial<TogglingConfigSpec> & { exclusive?: boolean };
```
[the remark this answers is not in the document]
lnewson: Its explicitly part of the item toggling logic which is an extension to the normal toggling spec. So it wouldn't make sense to include it as part of the `TogglingConfigSpec`.

PR #7908, opened 2022-06-16
TINY-8758: Default with value would add value when not desired

HAFRMO:
Related Ticket: 

Description of Changes:
When not using css, the default css value would be used regardless.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/tinymce/src/plugins/table/main/ts/ui/TableDialog.ts
```diff
@@ -81,7 +81,11 @@ const applyDataToElement = (editor: Editor, tableElm: HTMLTableElement, data: Ta
     styles['border-style'] = data.borderstyle;
   }
 
-  attrs.style = dom.serializeStyle({ ...Options.getDefaultStyles(editor), ...styles });
+  const fullStyleAttribute = { ...Options.getDefaultStyles(editor), ...styles };
+  if (dom.getAttrib(tableElm, 'width') && !Options.shouldStyleWithCss(editor)) {
```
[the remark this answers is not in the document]
TheSpyder: ah right, yeah the check for the existing width needs to be here.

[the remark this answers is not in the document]
TheSpyder: I wonder if this could use the same check done above on line 47, refactored into a variable 🤔

modules/tinymce/src/plugins/table/main/ts/ui/TableDialog.ts
```diff
@@ -81,7 +81,11 @@ const applyDataToElement = (editor: Editor, tableElm: HTMLTableElement, data: Ta
     styles['border-style'] = data.borderstyle;
   }
 
-  attrs.style = dom.serializeStyle({ ...Options.getDefaultStyles(editor), ...styles });
+  const fullStyleAttribute = { ...Options.getDefaultStyles(editor), ...styles };
```
lnewson, line 84: This seems like the wrong approach to take... I've looked into it more and the cause is that `table_default_styles` has `width: 100%` on it when `table_style_by_css` is `false`. That seems like the root bug here instead, as we're not supposed to be using CSS we're supposed to be using attributes wherever possible.

As such, I'd suggest changing the defaults instead so that it puts the width in via attributes when set to `false` instead of via CSS>

lnewson (changes requested):
The approach taken here seems okay, but it seems like a workaround and I'd like us to look into the root issue instead. If that's not going to work though, then I'm okay taking this approach.

modules/tinymce/src/plugins/table/main/ts/ui/TableDialog.ts
HAFRMO: Since the default option doesn't also provide an attribute object, if we go with this, do we want to create some kind of default attribute object in turn, or strip attriutes from the css, since we still want some sort of default for them?

lnewson: Not sure what you mean sorry? There is `table_default_attributes`?

HAFRMO: I am blind, that is what I mean. Blind and missing what is staring me right in the face.

modules/tinymce/src/plugins/table/main/ts/api/Options.ts
```diff
@@ -18,21 +18,41 @@ const defaultCellBorderStyles = Arr.map([ 'Solid', 'Dotted', 'Dashed', 'Double',
...
 const determineDefaultStyles = (editor: Editor, defaultStyles: Record<string, string>): Record<string, string> => {
-  if (isPixelsForced(editor)) {
-    // Determine the inner size of the parent block element where the table will be inserted
-    const dom = editor.dom;
-    const parentBlock = dom.getParent<HTMLElement>(editor.selection.getStart(), dom.isBlock) ?? editor.getBody();
-    const contentWidth = Width.getInner(SugarElement.fromDom(parentBlock));
-    return { ...defaultStyles, width: contentWidth + 'px' };
-  } else if (isResponsiveForced(editor)) {
-    return Obj.filter(defaultStyles, (_value, key) => key !== 'width');
+  if (isResponsiveForced(editor) || !shouldStyleWithCss(editor)) {
```
ltrouton, line 32: As hinted in the `Note` above, if we are making changes to this function, they will also have to be made in `modules/tinymce/src/models/dom/main/ts/table/api/Options.ts`

modules/tinymce/src/models/dom/main/ts/table/api/Options.ts
```diff
@@ -14,21 +14,42 @@ const option: {
...
+  if (editor.options.isSet('table_default_styles')) {
+    return defaultAttributes;
+  } else if (shouldStyleWithCss(editor)) {
+    return defaultAttributes;
+  } else if (isTablePixelsForced(editor)) {
+    return { ...defaultAttributes, ...Obj.filter(defaultStyles, (_value, key) => Arr.contains(stylesAvailableAsAttributes, key)), width: getPixelForcedWidth(editor) };
+  } else {
+    return { ...defaultAttributes, ...Obj.filter(defaultStyles, (_value, key) => Arr.contains(stylesAvailableAsAttributes, key)) };
+  }
+};
```
lnewson, line 51: This should definitely not be looking at the `table_default_styles` option... we should just be setting the width in the attribute instead and not setting it in the default styles since it cannot be a fixed value due to the `table_style_by_css` option.

- [ ] Remove `width: '100%'` from the defaults for `table_default_styles`
- [ ] Change this to set the default width as a variable and use it when in the `determineDefault...` functions

lnewson: To be clear, I'd expect something like this:
```ts
const defaultWidth = '100%';

const getPixelForcedWidth = (editor: Editor) => {
  // Determine the inner size of the parent block element where the table will be inserted
  const dom = editor.dom;
  const parentBlock = dom.getParent<HTMLElement>(editor.selection.getStart(), dom.isBlock) ?? editor.getBody();
  return Width.getInner(SugarElement.fromDom(parentBlock)) + 'px';
};

// Note: This is also contained in the table plugin Options.ts file
const determineDefaultTableStyles = (editor: Editor, defaultStyles: Record<string, string>): Record<string, string> => {
  if (isTablePercentagesForced(editor)) {
    return { ...defaultStyles, width: defaultWidth };
  } else if (isTablePixelsForced(editor)) {
    return { ...defaultStyles, width: getPixelForcedWidth(editor) };
  } else {
    return defaultStyles;
  }
};

// Note: This is also contained in the table plugin Options.ts file
const determineDefaultTableAttributes = (editor: Editor, defaultAttributes: Record<string, string>): Record<string, string> => {
  if (isTablePercentagesForced(editor)) {
    return { ...defaultAttributes, width: defaultWidth };
  } else if (isTablePixelsForced(editor)) {
    return { ...defaultAttributes, width: getPixelForcedWidth(editor) };
  } else {
    return defaultAttributes;
  }
};
```

Note: In the above I did intentionally swap to using `isTablePercentagesForced` as it's easy to reason what's going on and makes sure we only add the width when it needs to be added.

lnewson (changes requested):

modules/tinymce/src/models/dom/main/ts/table/api/Options.ts
lnewson: Oops the above doesn't account for the `table_style_by_css` option sorry... let me have a bit more of a think then.

lnewson: I thought it'd just be easier to do locally, so have made the changes mentioned above and pushed them here so you can review: https://github.com/tinymce/tinymce/compare/feature/TINY-8758...feature/TINY-8758_2. If all looks good, feel free to just cherry-pick that across.

The other thing missed was that we needed to move the registration of the `table_style_by_css` option to the model from the plugin.

modules/tinymce/src/plugins/table/test/ts/browser/ui/TableDialogWithoutCssTest.ts
```diff
@@ -0,0 +1,85 @@
+import { describe, it } from '@ephox/bedrock-client';
```
ltrouton, line 1: I think it would be better to add these tests to the existing TableDialogWithCssTest.ts file instead. That way we can properly check the width is correctly applied and kept for both when `table_style_css` is `false` and `true`.

HAFRMO: There are already tests that check this in that file. The reason this file was added was to test it with that option set to false.

modules/tinymce/src/models/dom/main/ts/table/api/Options.ts
HAFRMO: It was an attempt to have it in the css and moved over to the attributes when needed, and to allow more attributes to be moved over when appropriate instead of hard-coding them like this. I suppose it might have been a bit over-ambitious.

lnewson (approved):

vpyshnenko (approved):

modules/tinymce/src/plugins/table/test/ts/browser/ui/TableDialogWithoutCssTest.ts
ltrouton: I could be missing it but I don't see any tests that check the width is correct after submitting the dialog in the existing test file.

modules/tinymce/src/plugins/table/test/ts/browser/ui/TableDialogTest.ts
```diff
@@ -366,4 +366,52 @@ describe('browser.tinymce.plugins.table.TableDialogTest', () => {
     await TableTestUtils.pClickDialogButton(editor, false);
     editor.options.unset('table_appearance_options');
   });
+
+  it('TINY-8758: width should be retained when changing border.', async () => {
```
lnewson, line 370:
```suggestion
  it('TINY-8758: width should be retained when changing border', async () => {
```
There shouldn't be a fullstop here.

modules/tinymce/src/plugins/table/main/ts/api/Options.ts
```diff
@@ -144,6 +155,13 @@ const getDefaultStyles = (editor: Editor): Record<string, string> => {
   return options.isSet('table_default_styles') ? defaultStyles : determineDefaultStyles(editor, defaultStyles);
 };
 
+const getDefaultAttributes = (editor: Editor): Record<string, string> => {
+  // Note: The we don't rely on the default here as we need to dynamically lookup the widths based on the current editor state
+  const options = editor.options;
+  const defaultStyles = options.get('table_default_attributes');
+  return options.isSet('table_default_attributes') ? defaultStyles : determineDefaultAttributes(editor, defaultStyles);
```
ltrouton, line 162: I think this would be clearer?
```suggestion
  const defaultAttributes = options.get('table_default_attributes');
  return options.isSet('table_default_attributes') ? defaultAttributes : determineDefaultAttributes(editor, defaultAttributes);
```

lnewson: Looks like the wrong variables were changed in https://github.com/tinymce/tinymce/pull/7908/commits/37aa3595b8d5e75ad5752704097b83cbc34cd3e6

- [ ] Fix incorrect variable names

brian-forte (approved):

PR #7910, opened 2022-06-17
fix: Multiple font-family  display

YangChao-uniontech:
Related Ticket: 

Description of Changes:
When copying HTML to Tinymce, if the HTML has multiple font styles, Tinymce will display them all.However in fact, only the first font style is applied to the current HTML, which can result in inaccurate font styles or multiple font styles being displayed. Now, the problem of multiple font styles is solved by getting the current number of fonts and only reading the first font style as a display

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/commands/Font.ts
+++ b/modules/tinymce/src/core/main/ts/commands/Font.ts
@@ -46,8 +46,10 @@ export const fontNameAction = (editor: Editor, value: string): void => {
   editor.nodeChanged();
 };
 
-export const fontNameQuery = (editor: Editor): string => mapRange(editor, (elm: SugarElement<Element>) =>
-  FontInfo.getFontFamily(editor.getBody(), elm.dom)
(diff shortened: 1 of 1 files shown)


lnewson:
Hi @YangChao-uniontech,

Firstly, thanks for contributing this PR. Unfortunately though, this is currently working as intended and is therefore not a bug, not to mention doing this would be a breaking API change so I'm afraid we cannot accept it. The reason this is expected, is that the browsers won’t let us know the actual font being used as it can be used for fingerprinting or similar and the font being used could be anyone in the font-family list provided. As such, since the font being used can technically be any of the ones in the computed font list so we now show all of them.

If you'd like to see the UI only show one option, then one option is to ensure the font is configured via the [font_family_formats](https://www.tiny.cloud/docs/tinymce/6/user-formatting-options/#font_family_formats) option. If that isn't applicable or won't work for your use case, then this would have to be a new feature including a way to opt into it. So I'd suggest logging a feature request instead unless you'd like to attempt doing that yourself and logging a new PR for our product team to review.

PR #7911, opened 2022-06-17
TINY-8057: Changed Link menu and toolbar buttons to execute 'mceLink' command

### LINKED ISSUE #7235: Link plugin does not execute commands.
For example, when you open code editor dialog it executes `mceCodeEditor` command. When you click `unlink` button it executes `unlink` command and both of these examples can be overriden by `addCommand`.  But with adding/editing link.. It does not execute any commands, so you cannot track when the dialog is getting opened and etc.

However, `mceLink` is executed when opening dialog via keyboard shortcut (`CTRL +K`, https://www.tiny.cloud/docs/plugins/opensource/link/#link_quicklink)

When link dialog is opened it should execute a command.

MitchCTiny:
Related Ticket: 

Description of Changes:
* Added optional `dialog` boolean value as flag to `mceLink` command where `true` always opens link dialog.
* Changed usage of `Link` menu and toolbar buttons to execute command with `dialog` = `true`
* Added tests for each flag and for `Ctrl+K` shortcut behaviour

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

Fixes #7235

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Improved
 - Clearer focus states for buttons while navigating with keyboard #TINY-8557
-- `mceLink` command can now take the value `{ dialog: true }` to always open the link dialog #TINY-8057
+- The `mceLink` command can now take the value `{ dialog: true }` to always open the link dialog #TINY-8057
 
 ### Changed
 - Toggling fullscreen mode with the `fullscreen` plugin now also fires the `ResizeEditor` event #TINY-8701
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -23,6 +23,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The default value for `end_container_on_empty_block` option has been changed to `'blockquote'` #TINY-6559
 - Custom elements are now treated as non-empty elements via the schema #TINY-4784
 - The autocompleter menu element is now positioned instead of the wrapper #TINY-6476
+- `mceLink` command can now take `true` value to always open link dialog #TINY-8057
+- `Link` menu and toolbar buttons now always execute `mceLink` command #TINY-8057
 
 ### Fixed
 - Selecting all content with a single image in the content was inconsistent for the keyboard shortcut and menu item #TINY-4550
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Improved
 - Clearer focus states for buttons while navigating with keyboard #TINY-8557
+- `mceLink` command can now take the value `{ dialog: true }` to always open the link dialog #TINY-8057
 
 ### Changed
 - Toggling fullscreen mode with the `fullscreen` plugin now also fires the `ResizeEditor` event #TINY-8701
@@ -23,8 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The default value for `end_container_on_empty_block` option has been changed to `'blockquote'` #TINY-6559
 - Custom elements are now treated as non-empty elements via the schema #TINY-4784
 - The autocompleter menu element is now positioned instead of the wrapper #TINY-6476
-- `mceLink` command can now take `true` value to always open link dialog #TINY-8057
-- `Link` menu and toolbar buttons now always execute `mceLink` command #TINY-8057
+- `Link` menu and toolbar buttons now always execute the `mceLink` command #TINY-8057
 
 ### Fixed
 - Selecting all content with a single image in the content was inconsistent for the keyboard shortcut and menu item #TINY-4550
--- a/modules/tinymce/src/plugins/link/main/ts/api/Commands.ts
+++ b/modules/tinymce/src/plugins/link/main/ts/api/Commands.ts
@@ -3,9 +3,13 @@ import Editor from 'tinymce/core/api/Editor';
 import * as Dialog from '../ui/Dialog';
 import * as Options from './Options';
 
+interface Dialog {
+  readonly 'dialog': boolean;
+}
+
 const register = (editor: Editor): void => {
-  editor.addCommand('mceLink', (_ui, dialog?: boolean ) => {
-    if (dialog || !Options.useQuickLink(editor)) {
+  editor.addCommand('mceLink', (_ui, value?: Dialog) => {
+    if (value.dialog || !Options.useQuickLink(editor)) {
       Dialog.open(editor);
     } else {
       editor.dispatch('contexttoolbar-show', {
--- a/modules/tinymce/src/plugins/link/main/ts/api/Commands.ts
+++ b/modules/tinymce/src/plugins/link/main/ts/api/Commands.ts
@@ -4,12 +4,12 @@ import * as Dialog from '../ui/Dialog';
 import * as Options from './Options';
 
 interface Dialog {
-  readonly 'dialog': boolean;
+  readonly dialog?: boolean;
 }
 
 const register = (editor: Editor): void => {
   editor.addCommand('mceLink', (_ui, value?: Dialog) => {
-    if (value.dialog || !Options.useQuickLink(editor)) {
+    if (value?.dialog === true || !Options.useQuickLink(editor)) {
       Dialog.open(editor);
     } else {
       editor.dispatch('contexttoolbar-show', {
--- a/modules/tinymce/src/plugins/link/main/ts/api/Commands.ts
+++ b/modules/tinymce/src/plugins/link/main/ts/api/Commands.ts
@@ -1,10 +1,11 @@
 import Editor from 'tinymce/core/api/Editor';
 
+import * as Options from '../api/Options';
 import * as Dialog from '../ui/Dialog';
 
 const register = (editor: Editor): void => {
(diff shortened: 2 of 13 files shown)


modules/tinymce/src/plugins/link/main/ts/api/Commands.ts
```diff
@@ -1,17 +1,16 @@
...
 
-import * as Actions from '../core/Actions';
+import * as Dialog from '../ui/Dialog';
 import * as Options from './Options';
 
 const register = (editor: Editor): void => {
-  editor.addCommand('mceLink', () => {
-    if (Options.useQuickLink(editor)) {
-      // Taken from ContextEditorEvents in silver. Find a better way.
+  editor.addCommand('mceLink', (ui, dialog?: boolean ) => {
```
tiny-ben-tran, line 7:
```suggestion
  editor.addCommand('mceLink', (_ui, dialog?: boolean ) => {
```

tiny-ben-tran (approved):

modules/tinymce/CHANGELOG.md
```diff
@@ -23,6 +23,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The default value for `end_container_on_empty_block` option has been changed to `'blockquote'` #TINY-6559
 - Custom elements are now treated as non-empty elements via the schema #TINY-4784
 - The autocompleter menu element is now positioned instead of the wrapper #TINY-6476
+- `mceLink` command can now take `true` value to always open link dialog #TINY-8057
```
lnewson, line 26: We possibly don't really want to document this, given it's largely internal. If we do, then this is an improvement so should be in the `Improved` section instead.

modules/tinymce/CHANGELOG.md
```diff
@@ -23,6 +23,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- `Link` menu and toolbar buttons now always execute `mceLink` command #TINY-8057
```
lnewson, line 27:
```suggestion
- `Link` menu and toolbar buttons now always execute the `mceLink` command #TINY-8057
```

modules/tinymce/src/plugins/link/main/ts/api/Commands.ts
```diff
@@ -1,17 +1,16 @@
...
 
-import * as Actions from '../core/Actions';
+import * as Dialog from '../ui/Dialog';
 import * as Options from './Options';
 
 const register = (editor: Editor): void => {
-  editor.addCommand('mceLink', () => {
-    if (Options.useQuickLink(editor)) {
-      // Taken from ContextEditorEvents in silver. Find a better way.
+  editor.addCommand('mceLink', (_ui, dialog?: boolean ) => {
```
lnewson, line 7: There was a reason I mentioned using an object on slack for this, it's because this isn't extensible. This locks us in heavily and if in future we need to pass 2 different things in we can't without doing a breaking change. So generally we try to use objects for these use cases as they can be extended in a non-breaking way.

- [ ] Change the command value to be an object with a dialog or similar property instead

modules/tinymce/src/plugins/link/test/ts/browser/QuickLinkTest.ts
```diff
@@ -156,4 +159,26 @@ describe('browser.tinymce.plugins.link.QuickLinkTest', () => {
...
+
+  it('TINY-8057: Checking that mceLink command can open Quicklink and dialog', async () => {
+    const editor = hook.editor();
+    editor.setContent('');
+    editor.execCommand('mcelink', false, true);
+    await TinyUiActions.pWaitForDialog(editor);
+    TinyUiActions.closeDialog(editor);
+    editor.execCommand('mcelink');
+    // Waiting for Quicklink dialog as in pOpenQuickLink
+    await Waiter.pWait(100);
```
lnewson, line 171: Using `Waiter.pWait` is normally a pretty big red flag, as in many cases it'll lead to flaky tests. It should only be used when there's nothing to wait for generally. So in this case we do have something to wait for and that's for the context toolbar UI to render. For example:
```ts
await TinyUiActions.pWaitForPopup(editor, '.tox-pop .tox-toolbar');
```
and then as that's more descriptive, you probably don't need the comment above 🤷 

- [ ] Don't use `Waiter.pWait` to wait for the quicklink context toolbar to show

modules/tinymce/src/plugins/link/test/ts/browser/QuickLinkTest.ts
```diff
@@ -156,4 +159,26 @@ describe('browser.tinymce.plugins.link.QuickLinkTest', () => {
...
+    await FocusTools.pTryOnSelector('Selector should be in context form input', doc, '.tox-toolbar input');
+  });
+
+  it('TINY-8057: Checking Quicklink opens with keyboard shortcut', async () => {
+    const editor = hook.editor();
+    editor.setContent('<p>blah blah blah</p>');
+    TinyUiActions.keystroke(editor, 0x001E, metaKey); // Meta+K
```
lnewson, line 178: Why are you using `0x001E` here?

modules/tinymce/src/plugins/link/test/ts/browser/QuickLinkTest.ts
```diff
@@ -156,4 +159,26 @@ describe('browser.tinymce.plugins.link.QuickLinkTest', () => {
...
+    TinyUiActions.keystroke(editor, 75, metaKey); // Meta+K
```
lnewson, line 179:
```suggestion
    TinyUiActions.keystroke(editor, 'k'.charCodeAt(0), metaKey);
```
This makes it more readable and so that you don't need the comment.

lnewson (changes requested):
Good job on this @MitchCTiny!

I'm only marking it as needs changes because nothing tests the buttons/menu uses the `mceLink` command so that could still regress. So can you please add a test for that as well?

modules/tinymce/CHANGELOG.md
MitchCTiny: Seeing as the documentation needs to be updated I was considering including how to use the `mceLink` command to open the dialog even with quicklink enabled as that may be important, so I'll ask documentation how best to describe this for the changelog also

modules/tinymce/src/plugins/link/test/ts/browser/QuickLinkTest.ts
MitchCTiny: Left that in by accident

modules/tinymce/src/plugins/link/test/ts/browser/QuickLinkTest.ts
MitchCTiny: Done, thanks Lee

modules/tinymce/CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Improved
 - Clearer focus states for buttons while navigating with keyboard #TINY-8557
+- `mceLink` command can now take the value `{ dialog: true }` to always open the link dialog #TINY-8057
```
```suggestion
- The `mceLink` command can now take the value `{ dialog: true }` to always open the link dialog #TINY-8057
```

modules/tinymce/src/plugins/link/main/ts/api/Commands.ts
```diff
@@ -1,17 +1,20 @@
 import Editor from 'tinymce/core/api/Editor';
 
-import * as Actions from '../core/Actions';
+import * as Dialog from '../ui/Dialog';
 import * as Options from './Options';
 
+interface Dialog {
```
lnewson, line 6: This name conflicts with the import above and it's not related to the dialog anyways, it's the command options. So maybe something like `CommandOptions` instead?

modules/tinymce/src/plugins/link/main/ts/api/Commands.ts
```diff
@@ -1,17 +1,20 @@
...
+  readonly 'dialog': boolean;
```
lnewson, line 7:
```suggestion
  readonly dialog?: boolean;
```
There shouldn't be quotes here and I'm not convinced this should be mandatory? It would lock us in again in future.

modules/tinymce/src/plugins/link/main/ts/api/Commands.ts
```diff
@@ -1,17 +1,20 @@
...
+}
+
 const register = (editor: Editor): void => {
-  editor.addCommand('mceLink', () => {
-    if (Options.useQuickLink(editor)) {
-      // Taken from ContextEditorEvents in silver. Find a better way.
+  editor.addCommand('mceLink', (_ui, value?: Dialog) => {
+    if (value.dialog || !Options.useQuickLink(editor)) {
```
lnewson, line 12: `value` can be `undefined` here so this might throw an error. I'd suggest we do something like this instead:
```suggestion
    if (value?.dialog === true || !Options.useQuickLink(editor)) {
```

modules/tinymce/src/plugins/link/main/ts/api/Commands.ts
MitchCTiny: Ah that explains the failures haha, thanks @lnewson

MitchCTiny:
> Good job on this @MitchCTiny!
> 
> I'm only marking it as needs changes because nothing tests the buttons/menu uses the `mceLink` command so that could still regress. So can you please add a test for that as well?

Will add tests for this as well ASAP

MitchCTiny:
@lnewson link toolbar and menu button tests have been added

lnewson (changes requested):
Getting there Mitch, but we're still not quite testing what we need to though. I'd suggest testing by removing the `value?.dialog === true ||` check and making sure your tests actually fail locally.

lnewson (approved):

PR #7918, opened 2022-06-21
TINY-8337: Fixed incorrect handling of image data URIs that aren't base64 encoded

lnewson:
Related Ticket: TINY-8337

Description of Changes:
* Fixes the editor upload/conversions logic to properly handle data URIs where the data isn't base64 encoded as previously in this case it corrupted the image by specifying the blob data as being nothing.
* Cleaned up the logic a little, fixed some incorrect types and removed some duplicated code between the parser and image scanner.
* Merged the `Base64Uris` logic into the `Conversions` module as they were both handling similar things. I did however remove the `restoreDataUris` function logic as its no longer used in TinyMCE 6.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


TheSpyder (approved):

modules/tinymce/src/core/main/ts/api/EditorUpload.ts
```diff
@@ -1,4 +1,4 @@
-import { Arr, Strings } from '@ephox/katamari';
+import { Arr, Cell, Strings, Type } from '@ephox/katamari';
```
lnewson, line 1:
```suggestion
import { Arr, Strings, Type } from '@ephox/katamari';
```

ltrouton (approved):

PR #7925, opened 2022-06-22
TINY-8806: First pass at fixing strict mode issues

lnewson:
Related Ticket: TINY-8806

Description of Changes:

This is the first PR for the work done today to convert core to strict mode. There's still **a lot** more to do, but in an effort to try and keep the reviews manageable I'm going to log once per day. I should also add that while I've fixed many, I still can't add them to the `tsconfig.strict.json` configuration because they rely on Editor or DOMUtils.

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [x] Milestone set


modules/tinymce/src/core/main/ts/DragDropOverrides.ts
```diff
@@ -193,10 +195,15 @@ const move = (state: Singleton.Value<State>, editor: Editor) => {
 };
 
 // Returns the raw element instead of the fake cE=false element
-const getRawTarget = (selection: EditorSelection) => {
-  const rng = selection.getSel().getRangeAt(0);
-  const startContainer = rng.startContainer;
-  return startContainer.nodeType === 3 ? startContainer.parentNode : startContainer;
+const getRawTarget = (selection: EditorSelection): Node | null => {
+  const sel = selection.getSel();
+  if (Type.isNonNullable(sel)) {
```
lnewson, line 200: Minor code change here given that `getSel()` can be nullable.

modules/tinymce/src/core/main/ts/ForceBlocks.ts
```diff
@@ -29,19 +29,19 @@ const isValidTarget = (blockElements, node) => {
...
   });
 };
 
 // const is
 
-const shouldRemoveTextNode = (blockElements, node) => {
+const shouldRemoveTextNode = (blockElements: SchemaMap, node: Node) => {
   if (NodeType.isText(node)) {
-    if (node.nodeValue.length === 0) {
+    if (node.data.length === 0) {
```
lnewson, line 42: Minor code change here and in a couple of places... `nodeValue` is nullable whereas `data` is not. They are the same thing for Text nodes though.

modules/tinymce/src/core/main/ts/ForceBlocks.ts
```diff
@@ -88,7 +87,7 @@ const addRootBlocks = (editor: Editor) => {
 
       if (!rootBlockNode) {
         rootBlockNode = dom.create(forcedRootBlock, Options.getForcedRootBlockAttrs(editor));
-        node.parentNode.insertBefore(rootBlockNode, node);
+        rootNode.insertBefore(rootBlockNode, node);
```
lnewson, line 90: Another minor code change here as `node.parentNode` is nullable per the TS dom types. Since we know that parent is `rootNode` here and it's not nullable we can use it instead.

modules/tinymce/src/core/main/ts/NodeChange.ts
```diff
@@ -104,20 +105,21 @@ class NodeChange {
...
-      args = args || {};
-      args.element = node;
-      args.parents = parents;
-
-      this.editor.dispatch('NodeChange', args);
+      this.editor.dispatch('NodeChange', {
+        ...args,
+        element: node,
+        parents
+      });
```
lnewson, line 122: More prominent code change here, as the old code here mutated the `args` passed in, whereas this clones it now.

modules/tinymce/src/core/main/ts/api/dom/Selection.ts
```diff
@@ -310,11 +310,12 @@ const EditorSelection = (dom: DOMUtils, win: Window, serializer: DomSerializer,
     }
 
     try {
-      if ((selection = getSel()) && !NodeType.isRestrictedNode(selection.anchorNode)) {
+      const selection = getSel();
+      if (selection && !NodeType.isRestrictedNode(selection.anchorNode)) {
         if (selection.rangeCount > 0) {
           rng = selection.getRangeAt(0);
         } else {
-          rng = selection.createRange ? selection.createRange() : doc.createRange();
```
lnewson, line 317: Minor code change here, as `selection.createRange()` is an older IE API that no longer exists.

modules/tinymce/src/core/main/ts/api/util/Tools.ts
```diff
@@ -82,24 +83,17 @@ const is = (obj: any, type: string) => {
...
-  i = items.length;
+  let i = resolvedItems.length;
   while (i--) {
-    map[items[i]] = {};
+    newMap[resolvedItems[i]] = {};
   }
 
-  return map;
+  return newMap;
 };
```
lnewson, line 97: This is a rework to make it work nicer with types. It also prevents mutating in the passed `map` variable.

modules/tinymce/src/core/main/ts/api/util/Tools.ts
```diff
@@ -194,15 +185,17 @@ const resolve = (n, o?) => {
...
+  if (Type.isNullable(s) || Type.isArray(s)) {
     return s;
+  } else if (s === '') {
+    return [];
+  } else {
+    return ArrUtils.map(s.split(d || ','), trim);
   }
-
-  return ArrUtils.map(s.split(d || ','), trim);
 };
```
lnewson, line 196: Another case where I reworked the logic to make it play nicer with types. Also makes sure that an empty string returns an empty array rather than an empty string.

modules/tinymce/src/core/main/ts/fmt/Preview.ts
```diff
@@ -192,39 +191,40 @@ const getCssText = (editor: Editor, format: any) => {
   }
 
   // Removes any variables since these can't be previewed
-  const removeVars = (val): string => {
-    return val.replace(/%(\w+)/g, '');
+  const removeVars = (val: FormatAttrOrStyleValue): string => {
+    return Type.isString(val) ? val.replace(/%(\w+)/g, '') : '';
```
lnewson, line 195: Minor bug fix here technically, as this didn't account for the `val` being a function. As an example, run `editor.formatter.getCssText('lang')` and it'll throw an exception.

modules/tinymce/src/core/main/ts/selection/ElementSelection.ts
```diff
@@ -89,58 +87,60 @@ const getNode = (root: Element, rng: Range): Element => {
...
 
   return selectedBlocks;
 };
 
-const select = (dom, node: Node, content?: boolean) => Optional.from(node).map((node) => {
-  const idx = dom.nodeIndex(node);
-  const rng = dom.createRng();
+const select = (dom: DOMUtils, node: Node | null, content?: boolean) =>
+  Optional.from(node).bind((node) =>
+    Optional.from(node.parentNode).map((parent) => {
```
lnewson, line 128: Minor code change here to account for `node.parentNode` being nullable.

modules/tinymce/src/core/main/ts/textpatterns/core/Pattern.ts
```diff
@@ -97,41 +97,6 @@ const normalizePattern = (pattern: RawPattern): Result<Pattern, PatternError> =>
   }
 };
 
-const denormalizePattern = (pattern: Pattern): RawPattern => {
```
lnewson, line 100: This logic wasn't used so I deleted it.

modules/tinymce/src/core/main/ts/util/Private.ts
```diff
@@ -1,44 +0,0 @@
-import * as Uuid from './Uuid';
```
lnewson, line 1: This module isn't used anymore.

ltrouton (approved):
Have looked through about 2/3 of the files so far and can't see anything incorrect. Happy to give tick overall

modules/tinymce/src/core/main/ts/api/util/Tools.ts
```diff
@@ -194,15 +185,17 @@ const resolve = (n, o?) => {
  * // Split a string into an array with a,b,c
  * const arr = tinymce.explode('a, b,   c');
  */
-const explode = (s, d?) => {
-  if (!s || is(s, 'array')) {
+const explode = (s: string | string[], d?: string | RegExp): string[] => {
+  if (Type.isNullable(s) || Type.isArray(s)) {
```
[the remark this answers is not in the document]
lnewson: Yeah I should have commented on that sorry. I wasn't game to remove that check, as while nothing in our code is nullable something else externally potentially could have been. So I figured it was worth keeping for anyone not using the types and I didn't add it to the types as a) that'd complicate things a lot and b) I'd prefer if we didn't allow something nullable to be passed.

modules/tinymce/src/core/main/ts/fmt/Preview.ts
TheSpyder: should we have a test for this? 😉

modules/tinymce/src/core/main/ts/selection/ElementSelection.ts
TheSpyder: Could this be collapsed to `Optional.from(node?.parentNode)` and avoid the extra level of nesting?

modules/tinymce/src/core/main/ts/textpatterns/core/InlinePattern.ts
```diff
@@ -123,7 +123,7 @@ const findPattern = (editor: Editor, block: Node, details: PatternDetails): Opti
     } else {
       // Find any nested patterns, making sure not to process the current pattern again
       const resultsOpt = findPatternsRec(editor, details.remainingPatterns, spot.container, spot.offset, block);
-      const results = resultsOpt.getOr({ matches: [], position: spot });
+      const results = resultsOpt.getOrThunk((): SearchResults => ({ matches: [], position: spot }));
```
TheSpyder, line 126: it seems weird that we have to introduce a constant function just for the types 🤔

TheSpyder (approved):

modules/tinymce/src/core/main/ts/api/util/Tools.ts
lnewson: Hmm I understood that a different way, to me that meant that it'll extend it. I can see your point though so yeah I'll revert that part so it keeps mutating it.

lnewson: Worth noting there's very few places this is used in TinyMCE... the only place I can recall is in the schema where it chains multiple `mapMap` functions like so: 
```
const value = makeMap(defaultValue, ' ', makeMap(defaultValue.toUpperCase(), ' '));
```

modules/tinymce/src/core/main/ts/Rtc.ts
```diff
@@ -82,16 +82,16 @@ interface RtcAdaptor {
       index: Index,
       locks: Locks,
       beforeBookmark: UndoBookmark,
-      level?: UndoLevel,
+      level?: Partial<UndoLevel>,
       event?: Event
-    ) => UndoLevel;
-    undo: (undoManager: UndoManager, locks: Locks, index: Index) => UndoLevel;
-    redo: (index: Index, data: UndoLevel[]) => UndoLevel;
+    ) => UndoLevel | null;
```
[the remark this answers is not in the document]
lnewson: This isn't caused by the adaptor... the actual implementation returns a `null` here in some cases. For example: https://github.com/tinymce/tinymce/blob/bd10b2474e1be0f51cd804811635fa9e556d681c/modules/tinymce/src/core/main/ts/undo/Operations.ts#L29-L36

modules/tinymce/src/core/main/ts/api/util/Tools.ts
lnewson: Have restored the mutation.

modules/tinymce/src/core/main/ts/api/dom/Selection.ts
```diff
@@ -442,9 +441,10 @@ const EditorSelection = (dom: DOMUtils, win: Window, serializer: DomSerializer,
    * // Alerts the currently selected elements node name
    * alert(tinymce.activeEditor.selection.getNode().nodeName);
    */
-  const getNode = (): Element => ElementSelection.getNode(editor.getBody(), getRng());
+  // TODO: Investigate if this type is correct
```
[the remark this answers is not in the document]
lnewson: Ahh yeah, I added this as something to come back to when I got to strict typing this part of the code. The problem is that it appears as though `ElementSelection.getNode` should be `Node | null`. In my quick manual testing I'm not sure that was actually possible though.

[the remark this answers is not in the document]
lnewson: I wonder if we should change `ElementSelection.getNode()` to something like this to be 100% sure it's always going to be an element?
```ts
  const elm = NodeType.isText(node) ? node.parentNode : node;
  return NodeType.isElement(elm) ? elm : root;
```

modules/tinymce/src/core/main/ts/selection/ElementSelection.ts
TheSpyder: ugh of course it does 😩

modules/tinymce/src/core/main/ts/Rtc.ts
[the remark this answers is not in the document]
TheSpyder: ah fair enough then

modules/tinymce/src/core/main/ts/textpatterns/core/InlinePattern.ts
TheSpyder: Depends whether you consider an eager object or eager function to be more expensive 🤔

modules/tinymce/src/core/main/ts/api/dom/Selection.ts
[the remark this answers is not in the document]
TheSpyder: yes, looking at this now I think the change of `getNode` to return `Node | null` instead of `Element` is incorrect. Lots of things become easier if it just returns an element.

modules/tinymce/src/core/main/ts/textpatterns/core/InlinePattern.ts
lnewson: Yeah that's also a good point... it's probably negligible and the performance changes we're talking about isn't going to matter here, so I'll swap it back to the above instead.
```suggestion
      const results: SearchResults = resultsOpt.getOr({ matches: [], position: spot });
```

modules/tinymce/src/core/main/ts/api/util/Tools.ts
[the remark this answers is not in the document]
lnewson: Removed the `isNullable` check as discussed in our call today.

modules/tinymce/src/core/main/ts/api/dom/Selection.ts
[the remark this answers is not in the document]
lnewson: Changed this to use the above as discussed and have removed the TODO/`Element` cast.

PR #7927, opened 2022-06-23
TINY-8806: Second pass at fixing strict mode issues

lnewson:
Related Ticket: TINY-8806

Description of Changes:

This is the second PR for the work done today to convert core to strict mode. There's still **a lot** more to do, but in an effort to try and keep the reviews manageable I'm going to log once per day. I should also add that while I've fixed many, I still can't add them to the `tsconfig.strict.json` configuration because they rely on Editor or DOMUtils.

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [x] Milestone set


modules/tinymce/src/core/main/ts/api/Editor.ts
```diff
@@ -679,30 +680,34 @@ class Editor implements EditorObservable {
...
+    const elm = self.getElement();
 
     if (self.removed) {
       return '';
     }
 
     if (elm) {
-      args = args || {};
-      args.load = true;
+      const loadArgs: Partial<EditorContent.SetContentArgs> & { load: boolean } = {
```
lnewson, line 692: Various changes here to make sure we're not mutating the passed arguments.

modules/tinymce/src/core/main/ts/api/Editor.ts
```diff
@@ -715,52 +720,58 @@ class Editor implements EditorObservable {
...
+    let elm: HTMLElement | null = self.getElement();
 
     if (!elm || !self.initialized || self.removed) {
-      return;
+      return '';
     }
 
-    args = args || {};
-    args.save = true;
+    const getArgs: Partial<EditorContent.GetContentArgs> = {
```
lnewson, line 731: More changes here to prevent mutation of the passed arguements.

modules/tinymce/src/core/main/ts/api/Editor.ts
```diff
@@ -1015,8 +1026,14 @@ class Editor implements EditorObservable {
       return urlConverterCallback.call(self, url, elm, true, name);
     }
 
-    // Don't convert link href since thats the CSS files that gets loaded into the editor also skip local file URLs
-    if (!getOption('convert_urls') || (elm && elm.nodeName === 'LINK') || url.indexOf('file:') === 0 || url.length === 0) {
+    // Don't convert link href since that's the CSS files that gets loaded into the editor also skip local file URLs
+    if (
+      !getOption('convert_urls') ||
+      elm === 'link' ||
```
lnewson, line 1032: Code change here as `elm` can be a string or an Element and this only accounted for the element path.

modules/tinymce/src/core/main/ts/api/dom/NativeTypes.ts
```diff
@@ -1,2 +0,0 @@
-// tslint:disable-next-line:no-empty-interface
-export interface NativeSelection extends Selection {}
```
lnewson, line 2: Deleted as this wasn't used anymore.

modules/tinymce/src/core/main/ts/api/util/Observable.ts
```diff
@@ -78,7 +89,7 @@ const Observable: Observable<any> = {
       let parent = self.parent();
       while (parent && !dispatcherArgs.isPropagationStopped()) {
         parent.dispatch(name, dispatcherArgs, false);
-        parent = parent.parent();
+        parent = parent.parent ? parent.parent() : undefined;
```
lnewson, line 92: Code change here, as while everything that uses this doesn't implement `parent()`, if it did this may have thrown an exception.

modules/tinymce/src/core/main/ts/fmt/FormatContainer.ts
```diff
@@ -1,15 +1,17 @@
...
+  NodeType.isElement(node) && node.id === CARET_ID;
 
-const getParentCaretContainer = (body: Node, node: Node) => {
-  while (node && node !== body) {
-    if ((node as Element).id === CARET_ID) {
-      return node;
+const getParentCaretContainer = (body: Node, node: Node): Element | null => {
+  let currentNode: Node | null = node;
+  while (currentNode && currentNode !== body) {
+    if (isCaretNode(currentNode)) {
```
lnewson, line 10: Code change here, as this was checking the same thing as the function above so I just made it reuse the above function instead.

modules/tinymce/src/core/main/ts/init/Init.ts
```diff
@@ -175,18 +176,17 @@ const init = (editor: Editor) => {
   initPlugins(editor);
   const renderInfo = renderThemeUi(editor);
   augmentEditorUiApi(editor, Optional.from(renderInfo.api).getOr({}));
-  const boxInfo = {
```
lnewson, line 178: Moved this as it's only needed in the iframe case, so initializing it here wasn't necessary and saved having to define a type.

modules/tinymce/src/core/main/ts/init/InitIframe.ts
```diff
@@ -72,11 +77,11 @@ const createIframe = (editor: Editor, boxInfo) => {
   DOM.add(boxInfo.iframeContainer, ifr);
 };
 
-const init = (editor: Editor, boxInfo) => {
+const init = (editor: Editor, boxInfo: BoxInfo): void => {
   createIframe(editor, boxInfo);
 
   if (boxInfo.editorContainer) {
-    DOM.get(boxInfo.editorContainer).style.display = editor.orgDisplay;
```
lnewson, line 79: Minor code change here as `boxInfo.editorContainer` has to be a `Element`.

modules/tinymce/src/core/main/ts/keyboard/MatchKeys.ts
```diff
@@ -23,17 +23,15 @@ const baseKeyPattern = {
   keyCode: 0
 };
 
-const defaultPatterns = (patterns: KeyPattern[]): KeyPattern[] =>
+const defaultPatterns = (patterns: KeyPattern[]): Required<KeyPattern>[] =>
   Arr.map(patterns, (pattern) => ({
     ...baseKeyPattern,
-    action: Fun.noop,
```
lnewson, line 29: Removed this as `action` is mandatory for `KeyPattern` so this was never used and strict mode raises that as an issue.

modules/tinymce/src/core/main/ts/newline/InsertBlock.ts
```diff
@@ -20,71 +21,66 @@ import * as NewLineUtils from './NewLineUtils';
...
+    rawNode.nodeValue = Zwsp.trim(rawNode.data);
   });
 };
 
-const isEmptyAnchor = (dom: DOMUtils, elm: Element) => {
+const isEmptyAnchor = (dom: DOMUtils, elm: Node): boolean => {
   return elm && elm.nodeName === 'A' && dom.isEmpty(elm);
 };
 
-const isTableCell = (node) => {
```
lnewson, line 31: Made this use the `NodeType.isTableCellOrCaption` function instead to remove some duplication.

modules/tinymce/src/core/main/ts/newline/InsertBlock.ts
```diff
@@ -93,39 +89,23 @@ const normalizeZwspOffset = (start, container, offset) => {
...
-      editableRoot = parent;
-    }
-
-    parent = parent.parentNode;
-  }
-
-  return parent !== root ? editableRoot : root;
+    currentNode = currentNode.firstChild;
+  } while (currentNode);
 };
```
lnewson, line 109: Deleted this as it was duplicate code and already defined in `NewLineUtils.ts`.

modules/tinymce/src/core/main/ts/newline/InsertLi.ts
```diff
@@ -48,11 +50,12 @@ const isFirstOrLastLi = (containerBlock, parentBlock, first) => {
...
 
 // Inserts a block or br before/after or in the middle of a split list of the LI is empty
-const insert = (editor: Editor, createNewBlock, containerBlock, parentBlock, newBlockName: string) => {
+const insert = (editor: Editor, createNewBlock: (name: string) => Element, containerBlock: Element, parentBlock: Element, newBlockName: string): void => {
   const dom = editor.dom;
   const rng = editor.selection.getRng();
+  const containerParent = containerBlock.parentNode;
 
-  if (containerBlock === editor.getBody()) {
+  if (containerBlock === editor.getBody() || !containerParent) {
```
lnewson, line 58: Code change here to early exit if we don't have a parent node for the container (i.e. the container is detached).

modules/tinymce/src/core/main/ts/selection/NormalizeRange.ts
```diff
@@ -8,49 +8,54 @@ import * as NodeType from '../dom/NodeType';
...
+const isCeFalseCaretContainer = (node: Node, rootNode: Node) => CaretContainer.isCaretContainer(node) && !hasParent(node, rootNode, isCaretNode);
 
 const hasBrBeforeAfter = (dom: DOMUtils, node: Node, left: boolean) => {
-  const walker = new DomTreeWalker(node, dom.getParent(node.parentNode, dom.isBlock) || dom.getRoot());
-
-  while ((node = walker[left ? 'prev' : 'next']())) {
-    if (NodeType.isBr(node)) {
-      return true;
+  const parentNode = node.parentNode;
+  if (parentNode) {
```
lnewson, line 34: Code change here to ensure that node is not detached and has a parent node

modules/tinymce/src/core/main/ts/selection/NormalizeRange.ts
```diff
@@ -59,17 +64,22 @@ const hasContentEditableFalseParent = (body: HTMLElement, node: Node) => {
...
 const findTextNodeRelative = (dom: DOMUtils, isAfterNode: boolean, collapsed: boolean, left: boolean, startNode: Node): Optional<CaretPosition> => {
-  let lastInlineElement;
   const body = dom.getRoot();
-  let node;
   const nonEmptyElementsMap = dom.schema.getNonEmptyElements();
+  const parentNode = startNode.parentNode;
+  let lastInlineElement: Node | undefined;
+  let node: Node | undefined;
+
+  if (!parentNode) {
```
lnewson, line 73: Another check to handle the node potentially being detached/where it's parent node is null.

modules/tinymce/src/core/main/ts/selection/NormalizeRange.ts
```diff
@@ -8,49 +8,54 @@ import * as NodeType from '../dom/NodeType';
...
 
-const hasParentWithName = (node: Node, rootNode: Node, name: string) => hasParent(node, rootNode, (node) => {
-  return node.nodeName === name;
-});
+const hasParentWithName = (node: Node, rootNode: Node, name: string): boolean =>
+  hasParent(node, rootNode, (node) => node.nodeName === name);
 
-const isTable = (node: Node) => node && node.nodeName === 'TABLE';
-
-const isTableCell = (node: Node) => node && /^(TD|TH|CAPTION)$/.test(node.nodeName);
```
lnewson, line 31: Replaced these with the `NodeType` functions to remove code duplication.

modules/tinymce/src/core/main/ts/selection/SplitRange.ts
```diff
@@ -13,30 +13,32 @@ const split = (rng: RangeLikeObject): RangeLikeObject => {
...
+    if (startOffset > 0 && startOffset < startContainer.data.length) {
       endContainer = splitText(startContainer, startOffset);
-      startContainer = endContainer.previousSibling;
+      startContainer = endContainer.previousSibling as Text;
 
       if (endOffset > startOffset) {
         endOffset = endOffset - startOffset;
-        startContainer = endContainer = splitText(endContainer as Text, endOffset).previousSibling;
-        endOffset = endContainer.nodeValue.length;
+        const newContainer = splitText(endContainer as Text, endOffset).previousSibling as Text;
```
lnewson, line 22: Extracted this out to a separate variable so we can assert the type, since TS can't track that the split nodes previousSibling will definitely be a Text node here.

modules/tinymce/src/core/main/ts/caret/LineWalker.ts
```diff
@@ -96,16 +101,18 @@ const positionsUntil = (direction: VDirection, root: Node, predicateFn: RectPred
...
-    if (direction === 1) {
-      return ArrUtils.last(caretPosition.getClientRects());
+  const getClientRect = (caretPosition: CaretPosition): ClientRect => {
+    // TODO: Are we really guaranteed that we're get at least 1 client rect back?
+    // Also should this really be returning the last item for both directions?
+    if (direction === VDirection.Down) {
+      return ArrUtils.last(caretPosition.getClientRects()) as ClientRect;
+    } else {
+      return ArrUtils.last(caretPosition.getClientRects()) as ClientRect;
     }
```
lnewson, line 115: This is almost definitely a bug... however the git history shows it's been like that since at least 2015. Should we just remove the branching here?

modules/tinymce/src/core/main/ts/api/Editor.ts
TheSpyder: And once again certain ways to read the API definition thwart your attempts to improve the code 😩 

`* @param {Object} args Optional content object, this gets passed around through the whole save process.`

Although in this case I'm tempted to leave it, easy to swap back if someone has a really good reason to need the actual object instance to be passed around.

modules/tinymce/src/core/main/ts/api/util/FakeStorage.ts
```diff
@@ -1,20 +1,20 @@
 // Simple stub of localstorage for IE11 with strict security settings #TINY-1782
```
TheSpyder, line 1: maybe we just delete this? 😉

modules/tinymce/src/core/main/ts/caret/LineWalker.ts
TheSpyder: I would think the intent is to return the last for one and first for the other (not sure which is which). Maybe there's been a weird lurking bug since 2015?

I believe at least one client rect is guaranteed - for a range `getClientRects` aggregates all client rects of elements in the range. I guess if it's not connected to the body it might be zero.

TheSpyder (approved):

modules/tinymce/src/core/main/ts/api/Editor.ts
lnewson: I disagree here... this is no way prevents the args being passed. They are still being passed through the whole process.

modules/tinymce/src/core/main/ts/api/util/FakeStorage.ts
lnewson:
```suggestion
// Simple stub of localstorage with strict security settings #TINY-1782
```

modules/tinymce/src/core/main/ts/caret/LineWalker.ts
lnewson: As discussed today, this logic is only used when for the `isAboveFn` so using `last` appears to make sense there. As such we're just going to remove the `direction` check and rename the function.

ltrouton (approved):

PR #7934, opened 2022-06-28
TINY-8785:  editor focused via the auto_focus option isn't scrolled into to view

vpyshnenko:
Related Ticket: TINY-8785

Description of Changes:
* add `editor.getBody().scrollIntoView()` to `autoFocus` function

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


hamza0867 (approved):
LGTM. Nice work Viktor 👍

modules/tinymce/src/core/main/ts/init/InitContentBody.ts
```diff
@@ -219,6 +219,7 @@ const autoFocus = (editor: Editor) => {
 
       if (!focusEditor.destroyed) {
         focusEditor.focus();
+        focusEditor.getBody().scrollIntoView();
```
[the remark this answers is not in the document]
lnewson: Another issue is that `scrollIntoView` does not work very well when the editor is already in view... it can often cause it to scroll to the top. For example, try https://fiddle.tiny.cloud/ffiaab/7 and click the focus second button. Even though the second editor is in view it causes it to scroll to the top.

As such, we probably want to consider using our `ScrollIntoView` logic instead which accounts for this case.

- [ ] Fix this issue and add a test to cover it

modules/tinymce/src/core/test/ts/browser/EditorAutoFocusTest.ts
```diff
@@ -0,0 +1,63 @@
...
+import { EditorManager } from 'tinymce/core/api/PublicApi';
+
+const isInViewport = (editor: Editor) => {
+  const containerRect = editor.getContentAreaContainer().getBoundingClientRect();
+  const innerRect = editor.selection.getNode().getBoundingClientRect();
+  const top = containerRect.top + innerRect.top + 1;
+  return top > 0 && top + innerRect.height < window.innerHeight;
+};
+
+const setupEditorAutoFocus = (id: string) => {
```
lnewson, line 16: Per the naming conversions, anything that's async in tests should be prefixed with `p`. E.g. `pSetupEditorAutoFocus`.

modules/tinymce/src/core/test/ts/browser/EditorAutoFocusTest.ts
```diff
@@ -0,0 +1,63 @@
...
+  return new Promise((resolve) => {
+    EditorManager.init({
+      selector: 'textarea',
+      base_url: '/project/tinymce/js/tinymce/',
+      toolbar: '',
+      menubar: false,
+      statusbar: false,
```
lnewson, line 23: `toolbar: ''` is not the correct way to disable the toolbar, you should be using `toolbar: false`. In saying that though, as mentioned elsewhere we need these enabled to test they are still in view.

modules/tinymce/src/core/test/ts/browser/EditorAutoFocusTest.ts
```diff
@@ -0,0 +1,63 @@
...
+const testEditorAutoFocus = async (id: string) => {
+  await setupEditorAutoFocus(id);
+  const editor = EditorManager.get(id);
+  assert.isTrue(editor.hasFocus());
+  assert.isTrue(isInViewport(editor));
+};
+
+describe('browser.tinymce.core.EditorAutoFocusTest', () => {
+  before(() => {
+    Insert.append(SugarBody.body(), SugarElement.fromHtml(`<div>
```
lnewson, line 42: These elements are never being cleaned up... please make sure to clean up any setup logic in an `after` hook or similar.

modules/tinymce/src/core/test/ts/browser/EditorAutoFocusTest.ts
```diff
@@ -0,0 +1,63 @@
...
+      height: 1000,
+      auto_focus: id,
+      init_instance_callback: (editor: Editor) => {
+        editor.on('focus', resolve);
+      }
+    });
+  });
+};
+
+const testEditorAutoFocus = async (id: string) => {
```
lnewson, line 33: Likewise here this doesn't follow naming conventions.

modules/tinymce/src/core/test/ts/browser/EditorAutoFocusTest.ts
```diff
@@ -0,0 +1,63 @@
...
+  await setupEditorAutoFocus(id);
+  const editor = EditorManager.get(id);
+  assert.isTrue(editor.hasFocus());
+  assert.isTrue(isInViewport(editor));
+};
+
+describe('browser.tinymce.core.EditorAutoFocusTest', () => {
```
lnewson, line 40: We also need to make sure this works in inline mode correctly, so can you please add tests for that as well? I'd suggest using the `Arr.each` with `context` method I mentioned last night to do that.

- [ ] Add tests to make sure this works with inline mode as well

modules/tinymce/src/core/test/ts/browser/EditorAutoFocusTest.ts
```diff
@@ -0,0 +1,63 @@
...
+    <textarea id="mce_0">Editor_0</textarea>
+    <textarea id="mce_1">Editor_1</textarea>
+    <textarea id="mce_2">Editor_2</textarea>
+  </div>`));
+    TinyHooks.bddSetupLight({}); // it puts tinymce to global scope as a side effect
+  });
+
+  afterEach(() => EditorManager.remove());
+
+  it('TINY-8785: it should autofocus the first editor', async () => {
```
lnewson, line 52:
```suggestion
  it('TINY-8785: should autofocus the first editor', async () => {
```
`It` shouldn't be in the message like this, as it's already implied by the `it` function.

lnewson (dismissed):

modules/tinymce/src/core/test/ts/browser/EditorAutoFocusTest.ts
```diff
@@ -0,0 +1,63 @@
+import { describe, it, before, afterEach } from '@ephox/bedrock-client';
+import { Insert, SugarBody, SugarElement } from '@ephox/sugar';
+import { TinyHooks } from '@ephox/wrap-mcagar';
+import { assert } from 'chai';
+
+import Editor from 'tinymce/core/api/Editor';
+import { EditorManager } from 'tinymce/core/api/PublicApi';
```
lnewson, line 7: This should not be importing from the `PublicApi`.

modules/tinymce/src/core/main/ts/init/InitContentBody.ts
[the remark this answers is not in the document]
TheSpyder: I'm fairly sure we wrote some custom scroll code for TBIO to provide an approximation of the 'if needed' code from Safari on all browsers. I wonder if we could dig that up. Sadly I don't really remember where to look.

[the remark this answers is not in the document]
lnewson: The `ScrollIntoView` module I mentioned above was rewritten back around 5.1 based on the original TBIO logic.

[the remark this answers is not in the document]
TheSpyder: ah I think I misread what you said. Yes that's exactly what we need.

- [ ] use ScrollIntoView logic

TheSpyder (approved):
Excellent, thank you 👍

TheSpyder:
Lee is away so I'm dismissing his review so it can be merged.

modules/tinymce/src/core/main/ts/init/InitContentBody.ts
```diff
@@ -220,6 +221,7 @@ const autoFocus = (editor: Editor) => {
 
       if (focusEditor && !focusEditor.destroyed) {
         focusEditor.focus();
+        scrollElementIntoView(focusEditor, focusEditor.selection.getNode());
```
lnewson, line 224: This should just be using the `editor.selection.scrollIntoView(`) API instead.

Issue #7941, opened 2022-06-29
[Tiny 6.1.0] "only in private projects"


Georg-Git:
Security warning while building:  **"warning Workspaces can only be enabled in private projects"**

It is the first time I am seeing this warning while building tinymce.

Tiny Master => ~6.1.0
Windows Yarn 1.22.19

The package.json file of Tiny begins with:
```
{
  "name": "root",
  "private": true,
  "workspaces": [
    "modules/*"
  ],
...
```

So what?
The required declaration of private: true is done. 
What is the reason for this security warning?

TheSpyder:
Huh. It doesn't block the build, and we have had so many warnings about out of date packages over time, that I don't think any of us had noticed. I'll look into it, thanks for logging this!

TheSpyder:
It's a yarn bug:
https://github.com/yarnpkg/yarn/issues/8580

I used `grep` to search the codebase and the problem is we upgraded `webdriverio` in this release, which depends on a version of `keyv` that published a `package.json` including workspaces.

They have fixed it (jaredwray/keyv#347) so we'll see whether an upgrade of wdio is available or if we need to add a resolution override.

TheSpyder:
A simple upgrade of _our_ transitive dependencies upgraded `keyv` and fixed the problem!

TheSpyder:
Merged

Georg-Git:
Yes - workspace warning is gone.

By the way: 
I've been noticing for a while 3 remarkable warnings.
And the first one makes me nervous:

```
warning Pattern ["grunt-cli@^1.4.3"] is trying to unpack in the same destination 
..Local\\Yarn\\Cache\\v6\\npm-grunt-cli-1.4.3-22c9f1a3d2780bf9b0d206e832e40f8f499175ff-integrity\\node_modules\\grunt-cli
as pattern ["grunt-cli@~1.4.3"]. 
This could result in non-deterministic behavior, skipping.
```


```
warning "lerna > @lerna/version > @lerna/github-client > @octokit/rest > @octokit/plugin-request-log@1.0.4"
has unmet peer dependency "@octokit/core@>=3"
```

```
warning " > stylelint-order@5.0.0" 
has incorrect peer dependency "stylelint@^14.0.0"
```

TheSpyder:
yes, we need to figure out what to do with the out of date lerna and stylelint. Neither are causing us issues, as far as we can tell, it's just a matter of checking for breaking changes in the new major version.

[...]

PR #7942, opened 2022-06-30
TINY-8864: Upgrade transitive dependencies to avoid yarn workspace warning

TheSpyder:
Related Ticket: TINY-8864

Description of Changes:
* Doesn't seem like major version changes so I think it can be a hotfix.
* This doesn't need to involve publishing a new version since it only impacts building from source.
* I removed the experimental flag in `.yarnrc` while trying to track this down, and it doesn't seem to make a difference anymore, so I decided to leave it out.

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [x] Milestone set

GitHub issues (if applicable): #7941

### DIFF
--- a/.yarnrc
+++ b/.yarnrc
@@ -1,5 +1,4 @@
 --install.frozen-lockfile true
-workspaces-experimental true
 "registry" "https://registry.yarnpkg.com"
 "@ephox:registry" "https://registry.yarnpkg.com"
 "@tinymce:registry" "https://registry.yarnpkg.com"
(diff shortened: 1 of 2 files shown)


.yarnrc
```diff
@@ -1,5 +1,4 @@
 --install.frozen-lockfile true
-workspaces-experimental true
```
lnewson, line 2: This was needed for yarn 0.x, so given we only use 1.x now and I'd be surprised if anyone still used 0.x this seems fine.

lnewson (approved):

ltrouton (approved):

tiny-ben-tran (approved):

PR #7947, opened 2022-07-01
TINY-8806: Fourth pass at fixing strict mode issues

lnewson:
Related Ticket: 

Description of Changes:
* Placeholder text

Pre-checks:
* [ ] Changelog entry added

Review:
* [ ] Milestone set

…ere the node couldn't be nullable


modules/tinymce/src/core/main/ts/annotate/Identification.ts
```diff
@@ -23,24 +23,17 @@ const identify = (editor: Editor, annotationName: Optional<string>): Optional<{
...
+    Attribute.getOpt(c, `${Markings.dataAnnotationId()}`).bind((uid) =>
+      Attribute.getOpt(c, `${Markings.dataAnnotation()}`).map((name) => {
+        const elements = findMarkers(editor, uid);
+        return {
+          uid,
+          name,
+          elements
+        };
+      })
+    ));
```
lnewson, line 36: This is just a simplified version to avoid a nullable return value from `Attribute.get`. It should be equivalent though, as doing `has` and `get` should be the same as using an `Optional`.

modules/tinymce/src/core/main/ts/api/Editor.ts
```diff
@@ -298,6 +301,7 @@ class Editor implements EditorObservable {
     });
     this.baseURI = this.baseUri;
     this.inline = Options.isInline(self);
+    this.hasVisual = Options.isVisualAidsEnabled(self);
```
lnewson, line 304: Moved this here just to make sure it's initialized to a valid value initially.

modules/tinymce/src/core/main/ts/api/EditorOptions.ts
```diff
@@ -181,6 +181,8 @@ const getBuiltInProcessor = <K extends BuiltInOptionType>(type: K): Processor<Bu
         return (val) => Type.isArrayOf(val, Type.isObject);
       case 'regexp':
         return (val) => Type.is(val, RegExp);
+      default:
+        return Fun.always;
```
lnewson, line 185: Not sure why this is needed, as every case was covered by the `case` statements above and I double checked a couple times. Either way, this seems like a reasonable default.

modules/tinymce/src/core/main/ts/api/dom/DOMUtils.ts
```diff
@@ -33,119 +33,18 @@ import DomTreeWalker from './TreeWalker';
...
 
-const legacySetAttribute = (elm: SugarElement<Element>, name: string, value: string | number | boolean | null) => {
-  if (Type.isNullable(value) || value === '') {
-    Attribute.remove(elm, name);
-  } else {
-    Attribute.set(elm, name, value);
-  }
-};
-
-const setupAttrHooks = (styles: Styles, settings: Partial<DOMUtilsSettings>, getContext: () => any): AttrHooks => {
```
lnewson, line 65: Moved these below so all interfaces, etc... are grouped together near the top.

modules/tinymce/src/core/main/ts/api/dom/DOMUtils.ts
```diff
@@ -161,17 +60,17 @@ export interface DOMUtilsSettings {
   keep_values: boolean;
   update_styles: boolean;
   root_element: HTMLElement | null;
-  collect: Function;
+  collect: boolean;
```
lnewson, line 63: This was apparently always a boolean, but in TinyMCE 5 it was changed from `collect: editor.inline` to `collect: () => inline` in `InitContentBody.ts`, see https://github.com/tinymce/tinymce/commit/da184ded52e9cb99226b151708bb67bbc0091fde. This meant it was always truthy when used below. As such, I put this back to the original behaviour but if this should always be truthy I'm happy to just add `collect: true` instead in `InitContentBody.ts`.

modules/tinymce/src/core/main/ts/api/dom/DOMUtils.ts
```diff
@@ -33,119 +33,18 @@ import DomTreeWalker from './TreeWalker';
...
-  };
-
-  const attrHooks: AttrHooks = {
-    style: {
-      set: (elm, value: string | {} | null) => {
-        const sugarElm = SugarElement.fromDom(elm);
-        if (Type.isObject(value)) {
-          Css.setAll(sugarElm, value as Record<string, string>);
-          return;
-        }
```
lnewson, line 92: Removed this in the logic below because according to the types and my own debugging this will never be an object.

modules/tinymce/src/core/main/ts/api/dom/DOMUtils.ts
```diff
@@ -460,10 +447,19 @@ const DOMUtils = (doc: Document, settings: Partial<DOMUtilsSettings> = {}): DOMU
   };
 
   const setStyles = (elm: string | Element | Element[], stylesArg: StyleMap) => {
-    setStyle(elm, stylesArg);
```
lnewson, line 463: Changed this to make it work better with the types, so I moved the internals of `setStyle` out to separate functions that are called by both `setStyle` and `setStyles`.

modules/tinymce/src/core/main/ts/bookmark/ResolveBookmark.ts
```diff
@@ -32,27 +32,31 @@ const addBogus = (dom: DOMUtils, node: Node): Node => {
...
 
-const resolveCaretPositionBookmark = (dom: DOMUtils, bookmark: StringPathBookmark) => {
-  const range = dom.createRng();
-  const startPos = CaretBookmark.resolve(dom.getRoot(), bookmark.start);
-  range.setStart(startPos.container(), startPos.offset());
+const resolveCaretPositionBookmark = (dom: DOMUtils, bookmark: StringPathBookmark): Optional<BookmarkResolveResult> => {
+  const startPos = Optional.from(CaretBookmark.resolve(dom.getRoot(), bookmark.start));
+  const endPos = Optional.from(CaretBookmark.resolve(dom.getRoot(), bookmark.end));
+
+  return Optionals.lift2(startPos, endPos, (start, end) => {
```
lnewson, line 39: Change here as `CaretBookmark.resolve` can return a nullable value in specific cases and this logic didn't handle that case at all.

modules/tinymce/src/core/main/ts/delete/MergeBlocks.ts
```diff
@@ -78,9 +78,9 @@ const findInsertionPoint = (toBlock: SugarElement<Element>, block: SugarElement<
 const getInsertionPoint = (fromBlock: SugarElement<Element>, toBlock: SugarElement<Element>): Optional<SugarElement> =>
   Compare.contains(toBlock, fromBlock) ? findInsertionPoint(toBlock, fromBlock) : Optional.none();
 
-const trimBr = (first: boolean, block: SugarElement) => {
+const trimBr = (first: boolean, block: SugarElement<Element>) => {
   CaretFinder.positionIn(first, block.dom)
-    .map((position) => position.getNode())
+    .bind((position) => Optional.from(position.getNode()))
```
lnewson, line 83: Here's an example of one of the cases I did change to make sure to account for the fact that `CaretPosition.getNode()` can returned `undefined` in specific cases.

ltrouton (approved):

PR #7963, opened 2022-07-08
TINY-8755: Whitelist internal attributes to narrow down what is considered an internal attribute.

HAFRMO:
Related Ticket: 

Description of Changes:
Previously all attributes starting with `data-` were considered internal attributes, this assumes that it has be prefixed with `data-mce-` or similar to denote it is internal. A whitelist of already-existing attributes that does not follow this rule is included for backwards compatibility.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/dom/ElementUtils.ts
+++ b/modules/tinymce/src/core/main/ts/dom/ElementUtils.ts
@@ -24,7 +24,7 @@ const each = Tools.each;
 
 export interface ElementUtils {
   readonly compare: (node1: Node, node2: Node) => boolean;
-  readonly attributeIsInternal: (attribute: string) => boolean;
+  readonly isAttributeInternal: (attribute: string) => boolean;
 }
 
 const ElementUtils = (editor: Editor): ElementUtils => {
@@ -59,7 +59,7 @@ const ElementUtils = (editor: Editor): ElementUtils => {
         const name = attr.nodeName.toLowerCase();
 
         // Don't compare internal attributes or style
-        if (name !== 'style' && !attributeIsInternal(name)) {
+        if (name !== 'style' && !isAttributeInternal(name)) {
           attribs[name] = dom.getAttrib(node, name);
         }
       });
@@ -122,12 +122,12 @@ const ElementUtils = (editor: Editor): ElementUtils => {
     return !Bookmarks.isBookmarkNode(node1) && !Bookmarks.isBookmarkNode(node2);
   };
 
-  const attributeIsInternal = (attributeName: string): boolean =>
+  const isAttributeInternal = (attributeName: string): boolean =>
     Arr.find(internalAttributesPrefixes, (value) => Strings.startsWith(attributeName, value)).isSome() || internalAttributes.has(attributeName);
 
   return {
     compare,
-    attributeIsInternal
+    isAttributeInternal
   };
 };
 
--- a/modules/tinymce/src/core/main/ts/dom/ElementUtils.ts
+++ b/modules/tinymce/src/core/main/ts/dom/ElementUtils.ts
@@ -1,10 +1,17 @@
-import { Obj, Type } from '@ephox/katamari';
+import { Arr, Obj, Strings, Type } from '@ephox/katamari';
 
-import * as Bookmarks from '../../bookmark/Bookmarks';
-import * as NodeType from '../../dom/NodeType';
-import { attributeIsInternal } from '../../fmt/FormatUtils';
-import Tools from '../util/Tools';
-import DOMUtils from './DOMUtils';
+import Editor from '../api/Editor';
+import Tools from '../api/util/Tools';
+import * as Bookmarks from '../bookmark/Bookmarks';
+import * as NodeType from './NodeType';
+
+const internalAttributesPrefixes = [
+  'data-ephox-',
+  'data-mce-',
+  'data-alloy-',
+  'data-snooker-',
+  '_'
+];
 
 /**
  * Utility class for various element specific functions.
@@ -17,9 +24,12 @@ const each = Tools.each;
 
 export interface ElementUtils {
   readonly compare: (node1: Node, node2: Node) => boolean;
+  readonly attributeIsInternal: (attribute: string) => boolean;
 }
 
-const ElementUtils = (dom: DOMUtils): ElementUtils => {
+const ElementUtils = (editor: Editor): ElementUtils => {
+  const dom = editor.dom;
+  const internalAttributes = new Set<string>(editor.serializer.getTempAttrs());
   /**
    * Compares two nodes and checks if it's attributes and styles matches.
    * This doesn't compare classes as items since their order is significant.
@@ -112,8 +122,12 @@ const ElementUtils = (dom: DOMUtils): ElementUtils => {
     return !Bookmarks.isBookmarkNode(node1) && !Bookmarks.isBookmarkNode(node2);
   };
 
+  const attributeIsInternal = (attributeName: string): boolean =>
+    Arr.find(internalAttributesPrefixes, (value) => Strings.startsWith(attributeName, value)).isSome() || internalAttributes.has(attributeName);
+
   return {
-    compare
+    compare,
+    attributeIsInternal
   };
 };
 
--- a/modules/tinymce/src/core/main/ts/dom/ElementUtils.ts
+++ b/modules/tinymce/src/core/main/ts/dom/ElementUtils.ts
@@ -123,7 +123,7 @@ const ElementUtils = (editor: Editor): ElementUtils => {
   };
 
   const isAttributeInternal = (attributeName: string): boolean =>
-    Arr.find(internalAttributesPrefixes, (value) => Strings.startsWith(attributeName, value)).isSome() || internalAttributes.has(attributeName);
+    Arr.exists(internalAttributesPrefixes, (value) => Strings.startsWith(attributeName, value)) || internalAttributes.has(attributeName);
 
   return {
     compare,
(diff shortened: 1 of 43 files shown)


modules/tinymce/src/core/main/ts/fmt/FormatUtils.ts
```diff
@@ -9,6 +9,49 @@ import * as NodeType from '../dom/NodeType';
...
+const internalAttributesPrefixes = [
+  'data-ephox-',
+  'data-mce-',
+  'data-alloy-',
+  'data-snooker-',
+  '_',
+  'aria-'
+];
+
+const internalDataValues = [
```
HAFRMO, line 21: Not all of these values are likely to be needed, for example `data-bedrockid` is probably not something we need to worry about ever showing up in production. But I thought it'd be better to add all of them at the start and then remove them when others point out they're probably not needed than to miss some. These values include the ones I could find from premium plugins. Many are likely attributes used very locally ( such as in snooker ) which never leaves it.

lorenzo-pomili (approved):

modules/tinymce/src/core/main/ts/fmt/FormatUtils.ts
```diff
@@ -9,6 +9,49 @@ import * as NodeType from '../dom/NodeType';
 import * as Whitespace from '../text/Whitespace';
 import { BlockFormat, Format, FormatAttrOrStyleValue, FormatVars, InlineFormat, MixedFormat, SelectorFormat } from './FormatTypes';
 
+const internalAttributesPrefixes = [
+  'data-ephox-',
+  'data-mce-',
+  'data-alloy-',
+  'data-snooker-',
+  '_',
+  'aria-'
```
tiny-ben-tran, line 18: A lot of Aria attributes have `aria-` prefixes. I don't think they are TinyMCE specific

modules/tinymce/src/core/main/ts/fmt/FormatUtils.ts
TheSpyder: I think I'd prefer to change them all to `data-mce` or similar. Some of these probably aren't used very often, and it's a lot of strings to bundle and add to the lookup function. Every one of these we can (and in my opinion should) update.

It may be a little annoying, but the long term effects seem worth it. Less code to maintain, and less ability for new data attributes to sneak in that aren't prefixed correctly.

modules/tinymce/src/core/main/ts/fmt/FormatUtils.ts
TheSpyder: But does it matter if they're removed?

I think I agree, this is new and it probably isn't a good idea.

- [ ] remove `aria-`

TheSpyder (approved):

modules/tinymce/src/core/main/ts/fmt/FormatUtils.ts
HAFRMO: I agree with this, however, this would require us to also change values in the premium plugins ( some of these values are from them, and they use some of the others ) and they would not be fully backwards compatible against earlier versions, which I have been informed is not really acceptable. This is meant to allow us to be backwards compatible until we can fully fix this, presumably around 7.0...
Or at least this have been my understanding of the situation.

modules/tinymce/src/core/main/ts/fmt/FormatUtils.ts
HAFRMO: Ah, this probably hung around from an earlier bit, so I'll remove it.

modules/tinymce/src/core/main/ts/fmt/FormatUtils.ts
TheSpyder: That's... debatable. If we do what I suggested then new plugins will still work with old tinymce, but old plugins will start leaving data attributes around when used with new tinymce.

I'm not sure how much we really try to make compatibility work in that direction. Compatibility is more about making sure customer integrations continue to work after upgrading tinymce, not so much old versions of our own plugins.

lnewson (changes requested):

modules/tinymce/src/core/test/ts/browser/FormatterRemoveTest.ts
```diff
@@ -3,6 +3,7 @@ import { LegacyUnit, TinyAssertions, TinyHooks, TinySelections } from '@ephox/wr
 import { assert } from 'chai';
 
 import Editor from 'tinymce/core/api/Editor';
+import { ZWSP } from 'tinymce/src/core/main/ts/text/Zwsp';
```
lnewson, line 6:
```suggestion
import { ZWSP } from 'tinymce/core/text/Zwsp';
```
Just noticed this import path isn't quite correct either sorry... You could also just use `Unicode.zwsp` directly if you preferred 🤷

modules/tinymce/src/core/main/ts/dom/ElementUtils.ts
```diff
@@ -112,8 +122,12 @@ const ElementUtils = (dom: DOMUtils): ElementUtils => {
     return !Bookmarks.isBookmarkNode(node1) && !Bookmarks.isBookmarkNode(node2);
   };
 
+  const attributeIsInternal = (attributeName: string): boolean =>
```
lnewson, line 125: Just to follow normal naming standards/conventions, maybe call this `isAttributeInternal` as it's common to start functions like this with `is`.

modules/tinymce/src/core/main/ts/dom/ElementUtils.ts
```diff
@@ -111,8 +122,12 @@ const ElementUtils = (dom: DOMUtils): ElementUtils => {
     return !Bookmarks.isBookmarkNode(node1) && !Bookmarks.isBookmarkNode(node2);
   };
 
+  const isAttributeInternal = (attributeName: string): boolean =>
+    Arr.find(internalAttributesPrefixes, (value) => Strings.startsWith(attributeName, value)).isSome() || internalAttributes.has(attributeName);
```
lnewson, line 126:
```suggestion
    Arr.exists(internalAttributesPrefixes, (value) => Strings.startsWith(attributeName, value)) || internalAttributes.has(attributeName);
```
Sorry had missed this previously, but `exists` is the same as using `find` with `isSome()`

modules/tinymce/src/core/main/ts/dom/ElementUtils.ts
```diff
@@ -16,9 +24,12 @@ const each = Tools.each;
 
 export interface ElementUtils {
   readonly compare: (node1: Node, node2: Node) => boolean;
+  readonly isAttributeInternal: (attribute: string) => boolean;
 }
 
-const ElementUtils = (dom: DOMUtils): ElementUtils => {
+const ElementUtils = (editor: Editor): ElementUtils => {
+  const dom = editor.dom;
+  const internalAttributes = new Set<string>(editor.serializer.getTempAttrs());
```
lnewson, line 32:
```suggestion
  const internalAttributes = new Set<string>(editor.serializer.getTempAttrs());

```

modules/tinymce/src/core/test/ts/browser/FormatterRemoveTest.ts
```diff
@@ -603,4 +604,14 @@ describe('browser.tinymce.core.FormatterRemoveTest', () => {
...
   });
+
+  it('TINY-8755: Non-internal attributes are no longer removed', () => {
+    const editor = hook.editor();
+    // eslint-disable-next-line max-len
+    editor.setContent('<p><strong>bold<span data-field-type="TEXT"><span class="my-class-1"></span><span class="my-class-2"><span style="display: flex; align-items: flex-start;" data-mce-style="display: flex; align-items: flex-start;"><span class="my-class-3">' + ZWSP + '</span></span></span></span>text</strong></p>', { format: 'raw' });
+    TinySelections.setSelection(editor, [ 0, 0, 0 ], 2, [ 0, 0, 2 ], 2);
+    editor.formatter.remove('bold');
+    // eslint-disable-next-line max-len
+    TinyAssertions.assertContent(editor, '<p><strong>bo</strong>ld<span data-field-type="TEXT"><span class="my-class-1"></span><span class="my-class-2"><span style="display: flex; align-items: flex-start;" data-mce-style="display: flex; align-items: flex-start;"><span class="my-class-3"></span></span></span></span>te<strong>xt</strong></p>', { format: 'raw' });
```
lnewson, line 615:
```suggestion
    TinyAssertions.assertRawContent(editor, '<p><strong>bo</strong>ld<span data-field-type="TEXT"><span class="my-class-1"></span><span class="my-class-2"><span style="display: flex; align-items: flex-start;" data-mce-style="display: flex; align-items: flex-start;"><span class="my-class-3"></span></span></span></span>te<strong>xt</strong></p>');
```

modules/tinymce/src/core/test/ts/browser/FormatterRemoveTest.ts
```diff
@@ -603,4 +604,14 @@ describe('browser.tinymce.core.FormatterRemoveTest', () => {
     editor.formatter.remove('blockquote');
     TinyAssertions.assertContent(editor, '<p>test test</p>');
   });
+
+  it('TINY-8755: Non-internal attributes are no longer removed', () => {
```
lnewson, line 608:
```suggestion
  it('TINY-8755: Non-internal attributes are not removed', () => {
```
We generally want these to be using past tense since that's how they'll be read in future.

lnewson (approved):

modules/tinymce/src/core/main/ts/dom/ElementUtils.ts
HAFRMO: I'm not sure what the change here is?

PR #7968, opened 2022-07-13
TINY-8905: Allow formatter to be able to wrap noneditable elements

ltrouton:
Related Ticket: TINY-8905

Description of Changes:
* Allow a non-editable element to be wrapped if it is specifically marked as wrappable and also does not have any editable descendants
	* Currently the marker is a `data-mce-cef-wrappable` data attribute. I am open to suggestions if there is a better way to indicate wrappable elements whether that be an option or something else
* Refactor some of the `if` blocks in ApplyFormat.ts to hopefully make them easier to reason about
* Make sure styling on a non-editable element or on a child of a non-editable element is not eagerly removed so that the non-editable behaviour is respected	 

Pre-checks:
* [x] Changelog entry added

Review:
* [X] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/api/OptionTypes.ts
+++ b/modules/tinymce/src/core/main/ts/api/OptionTypes.ts
@@ -109,7 +109,7 @@ interface BaseEditorOptions {
   forced_root_block?: string;
   forced_root_block_attrs?: Record<string, string>;
   formats?: Formats;
-  format_wrap_noneditable_selectors?: string[];
+  format_noneditable_selector?: string;
   height?: number | string;
   hidden_input?: boolean;
   icons?: string;
@@ -259,6 +259,7 @@ export interface EditorOptions extends NormalizedEditorOptions {
   font_size_style_values: string;
   forced_root_block: string;
   forced_root_block_attrs: Record<string, string>;
+  format_noneditable_selector: string;
   iframe_attrs: Record<string, string>;
   images_file_types: string;
   images_upload_base_path: string;
--- a/modules/tinymce/src/core/main/ts/api/OptionTypes.ts
+++ b/modules/tinymce/src/core/main/ts/api/OptionTypes.ts
@@ -109,6 +109,7 @@ interface BaseEditorOptions {
   forced_root_block?: string;
   forced_root_block_attrs?: Record<string, string>;
   formats?: Formats;
+  format_wrap_noneditable_selectors?: string[];
   height?: number | string;
   hidden_input?: boolean;
   icons?: string;
--- a/modules/tinymce/src/core/main/ts/api/Options.ts
+++ b/modules/tinymce/src/core/main/ts/api/Options.ts
@@ -380,6 +380,11 @@ const register = (editor: Editor): void => {
     default: false
   });
 
+  registerOption('format_wrap_noneditable_selectors', {
+    processor: 'string[]',
+    default: []
+  });
+
   registerOption('preview_styles', {
     processor: (value) => {
       const valid = value === false || Type.isString(value);
@@ -822,6 +827,7 @@ const isInlineBoundariesEnabled = option('inline_boundaries');
 const getFormats = option('formats');
 const getPreviewStyles = option('preview_styles');
 const canFormatEmptyLines = option('format_empty_lines');
+const getWrapNoneditableSelectors = option('format_wrap_noneditable_selectors');
 const getCustomUiSelector = option('custom_ui_selector');
 const isInline = option('inline');
 const hasHiddenInput = option('hidden_input');
@@ -924,6 +930,7 @@ export {
   getFormats,
   getPreviewStyles,
   canFormatEmptyLines,
+  getWrapNoneditableSelectors,
   getCustomUiSelector,
   getThemeUrl,
   getModelUrl,
--- a/modules/tinymce/src/core/main/ts/api/Options.ts
+++ b/modules/tinymce/src/core/main/ts/api/Options.ts
@@ -381,9 +381,9 @@ const register = (editor: Editor): void => {
     default: false
   });
 
-  registerOption('format_wrap_noneditable_selectors', {
-    processor: 'string[]',
-    default: []
+  registerOption('format_noneditable_selector', {
+    processor: 'string',
+    default: ''
   });
 
   registerOption('preview_styles', {
@@ -842,7 +842,7 @@ const isInlineBoundariesEnabled = option('inline_boundaries');
 const getFormats = option('formats');
 const getPreviewStyles = option('preview_styles');
 const canFormatEmptyLines = option('format_empty_lines');
-const getWrapNoneditableSelectors = option('format_wrap_noneditable_selectors');
+const getFormatNoneditableSelector = option('format_noneditable_selector');
 const getCustomUiSelector = option('custom_ui_selector');
 const isInline = option('inline');
 const hasHiddenInput = option('hidden_input');
@@ -949,7 +949,7 @@ export {
   getFormats,
   getPreviewStyles,
   canFormatEmptyLines,
-  getWrapNoneditableSelectors,
+  getFormatNoneditableSelector,
   getCustomUiSelector,
   getThemeUrl,
   getModelUrl,
--- a/modules/tinymce/src/core/main/ts/fmt/ApplyFormat.ts
+++ b/modules/tinymce/src/core/main/ts/fmt/ApplyFormat.ts
@@ -87,6 +87,11 @@ const applyFormat = (ed: Editor, name: string, vars?: FormatVars, node?: Node | 
         return false;
       }
 
+      // Check if the node is nonediatble and if the format can override noneditable node
+      if (dom.getContentEditable(node) === 'false' && !format.ceFalseOverride) {
+        return true;
+      }
+
       // Check collapsed state if it exists
       if (Type.isNonNullable(format.collapsed) && format.collapsed !== isCollapsed) {
         return true;
@@ -139,7 +144,7 @@ const applyFormat = (ed: Editor, name: string, vars?: FormatVars, node?: Node | 
             FormatUtils.isValid(ed, wrapName, nodeName) &&
             FormatUtils.isValid(ed, parentName, wrapName) &&
             Type.isNonNullable(wrapElm);
-      // If it is not node specific, it means that it was not passed into 'formatter.apply` and is withiin the editor selection
+      // If it is not node specific, it means that it was not passed into 'formatter.apply` and is within the editor selection
       const isZwsp = !nodeSpecific && NodeType.isText(node) && Zwsp.isZwsp(node.data);
       const isCaret = isCaretNode(node);
       const isCorrectFormatForNode = !FormatUtils.isInlineFormat(format) || !dom.isBlock(node);
@@ -309,23 +314,6 @@ const applyFormat = (ed: Editor, name: string, vars?: FormatVars, node?: Node | 
     });
   };
 
-  const selectedNode = selection.getNode();
-  if (dom.getContentEditable(selectedNode) === 'false') {
-    node = selectedNode;
-    for (let i = 0, l = formatList.length; i < l; i++) {
-      const formatItem = formatList[i];
-      if (formatItem.ceFalseOverride && FormatUtils.isSelectorFormat(formatItem) && dom.is(node, formatItem.selector)) {
-        setElementFormat(node as Element, formatItem);
-        break;
-      }
-    }
-
-    if (!FormatUtils.isWrappableNoneditable(ed, node)) {
-      Events.fireFormatApply(ed, name, selectedNode, vars);
-      return;
-    }
-  }
-
   if (format) {
     if (node) {
       if (FormatUtils.isNode(node)) {
@@ -349,7 +337,9 @@ const applyFormat = (ed: Editor, name: string, vars?: FormatVars, node?: Node | 
           });
         });
 
-        FormatUtils.moveStart(dom, selection, selection.getRng());
+        if (dom.getContentEditable(selection.getNode()) !== 'false') {
+          FormatUtils.moveStart(dom, selection, selection.getRng());
+        }
         ed.nodeChanged();
       } else {
         CaretFormat.applyCaretFormat(ed, name, vars);
--- a/modules/tinymce/src/core/main/ts/fmt/ApplyFormat.ts
+++ b/modules/tinymce/src/core/main/ts/fmt/ApplyFormat.ts
@@ -221,6 +221,7 @@ const applyFormat = (ed: Editor, name: string, vars?: FormatVars, node?: Node | 
             parentNode.insertBefore(currentWrapElm, node);
             newWrappers.push(currentWrapElm);
           }
+
           // Wrappable noneditable element has been handled so go back to previous state
           if (isWrappableNoneditableElm && hasContentEditableState) {
             contentEditable = lastContentEditable;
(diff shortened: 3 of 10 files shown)


modules/tinymce/src/core/main/ts/fmt/ApplyFormat.ts
```diff
@@ -282,6 +291,7 @@ const applyFormat = (ed: Editor, name: string, vars?: FormatVars, node?: Node |
           node = mergeStyles(node);
         }
 
+        // TODO: Make sure all of the below handle noneditable elements correctly
```
ltrouton, line 294: I still need to sanity check noneditable nodes are handled correctly by these merge functions

lnewson (changes requested):
According to the Jira this is only meant to be a spike, so not sure why we're trying to merge this. I'm also not keen on this attribute approach, as we need customers to be able to configure this as well and `data-mce-*` attributes are internal only. When this was discussed previously, I seem to recall mentioning this would best be done via a format configuration or via an option (e.g. a CSS selector) to specify what can be formatted.

So because of the inconsistencies in the Jira and this PR I'm marking as Needs changes until it can be discussed further.

modules/tinymce/src/core/main/ts/api/Options.ts
```diff
@@ -380,6 +380,11 @@ const register = (editor: Editor): void => {
     default: false
   });
 
+  registerOption('format_wrap_noneditable_selectors', {
```
ltrouton, line 383: I have added a new option that can be used by integrators. Would we be happy with `format_wrap_noneditable_selectors` as the option name?

modules/tinymce/src/core/main/ts/fmt/FormatUtils.ts
```diff
@@ -99,6 +106,25 @@ const isEmptyTextNode = (node: Node | null): boolean => {
   return Type.isNonNullable(node) && NodeType.isText(node) && node.length === 0;
 };
 
+const isWrapNoneditableTarget = (editor: Editor, node: Node): boolean => {
+  const selectors = [ '[data-mce-cef-wrappable]' ].concat(Options.getWrapNoneditableSelectors(editor)).join(',');
+  return Selectors.is(SugarElement.fromDom(node), selectors);
+};
+
```
ltrouton, line 113: This is where a non-editable element is determined to be wrappable.

At the moment, either a selector provided by the new option or `data-mce-cef-wrappable` attribute can be used to indicate it is wrappable. It was discussed during the knowledge share that having both could be useful so that we can use the data attribute in our plugins/code without worrying about what the integrator sets the option to.
Are we happy with `data-mce-cef-wrappable` as the attribute name?

In addition, as noted in the comment, I have made it so that a non-editable element can only be wrapped if it has no editable descendants. If there are concerns about that, a follow-up task can be created.

modules/tinymce/src/core/main/ts/fmt/ApplyFormat.ts
ltrouton: To expand further on why this needs to be checked. I have made the assumption that we do not want to remove any styling from the `contenteditable="false"` element itself and/or its descendants as by definition of it being `contenteditable="false"`, we shouldn't be changing its structure

TheSpyder (approved):
Seems reasonable now 👍

modules/tinymce/src/core/main/ts/api/OptionTypes.ts
```diff
@@ -109,6 +109,7 @@ interface BaseEditorOptions {
   forced_root_block?: string;
   forced_root_block_attrs?: Record<string, string>;
   formats?: Formats;
+  format_wrap_noneditable_selectors?: string[];
```
lnewson, line 112: Since this is largely aimed at noneditable, might I suggest: `noneditable_format_wrap_selector`.

Also this option type is inconsistent and having an array doesn't make sense so this should just be a single string.

- [ ] Change option to be a single selector for consistency

lnewson: Also as this has a default, we should declare this as non-optional in the `EditorOptions` type. For example, `format_wrap_noneditable_selectors: string`

modules/tinymce/src/core/main/ts/fmt/ApplyFormat.ts
```diff
@@ -122,6 +118,10 @@ const applyFormat = (ed: Editor, name: string, vars?: FormatVars, node?: Node |
     const newWrappers: Element[] = [];
     let contentEditable = true;
 
+    const isWrappingBlockFormat = FormatUtils.isBlockFormat(format) && format.wrapper;
+    const isNonWrappingBlockFormat = FormatUtils.isBlockFormat(format) && !format.wrapper;
```
lnewson, line 122: Should these be in `FormatUtils` instead? They definitely seem out of place here is all.

modules/tinymce/src/core/main/ts/fmt/ApplyFormat.ts
```diff
@@ -122,6 +118,10 @@ const applyFormat = (ed: Editor, name: string, vars?: FormatVars, node?: Node |
...
+    const isWrapNameFormat = FormatUtils.isInlineFormat(format) || FormatUtils.isBlockFormat(format);
```
lnewson, line 123: This doesn't look correct? I block format is not something that wraps content. Should it be using is `isWrappingBlockFormat` above instead?

modules/tinymce/src/core/main/ts/fmt/ApplyFormat.ts
```diff
@@ -134,17 +134,34 @@ const applyFormat = (ed: Editor, name: string, vars?: FormatVars, node?: Node |
...
+        const isEditableDescendant = contentEditable && !hasContentEditableState;
+        const isValidBlockFormatForNode = isNonWrappingBlockFormat && FormatUtils.isTextBlock(ed, nodeName) && FormatUtils.isValid(ed, parentName, wrapName);
+        const canRenameBlock = isEditableDescendant && isValidBlockFormatForNode;
+
+        const isValidWrapNode = isWrapNameFormat && FormatUtils.isValid(ed, wrapName, nodeName) && FormatUtils.isValid(ed, parentName, wrapName);
+        // If it is not node specific, it means that it was not passed into 'formatter.apply` and is withiin the editor selection
+        const isZwsp = !nodeSpecific && NodeType.isText(node) && Zwsp.isZwsp(node.data);
+        const isCaret = isCaretNode(node);
+        const isCorrectFormatForNode = !FormatUtils.isInlineFormat(format) || !dom.isBlock(node);
+        const canWrapNode = (isEditableDescendant || isWrappableNoneditableElm) && isValidWrapNode && !isZwsp && !isCaret && isCorrectFormatForNode;
```
lnewson, line 163: This is doing a lot more eager evaluation now... can we break this out into functions that are called when needed to help with that? It's also help to make this more readable I believe.

modules/tinymce/src/core/main/ts/fmt/ApplyFormat.ts
lnewson, line 294: Just adding a task for this

- [ ] Make sure this TODO is done and cleaned up

modules/tinymce/src/core/main/ts/fmt/ApplyFormat.ts
```diff
@@ -292,18 +302,23 @@ const applyFormat = (ed: Editor, name: string, vars?: FormatVars, node?: Node |
...
         break;
       }
     }
 
-    Events.fireFormatApply(ed, name, node, vars);
-    return;
+    if (formatApplied || !FormatUtils.isWrappableNoneditable(ed, node)) {
+      Events.fireFormatApply(ed, name, selectedNode, vars);
+      return;
+    }
```
lnewson, line 321: This looks to be a breaking change... as this will make the event change from any `editor.formatter.apply()` call to only firing when a format was applied.

Are we sure we want to do this? If so then it definitely needs a changelog entry.

modules/tinymce/src/core/main/ts/fmt/MergeFormats.ts
```diff
@@ -12,7 +12,7 @@ const each = Tools.each;
 
 const mergeTextDecorationsAndColor = (dom: DOMUtils, format: ApplyFormat, vars: FormatVars | undefined, node: Node): void => {
   const processTextDecorationsAndColor = (n: Node) => {
-    if (NodeType.isElement(n) && NodeType.isElement(n.parentNode)) {
+    if (NodeType.isElement(n) && NodeType.isElement(n.parentNode) && dom.getContentEditable(n.parentNode) !== 'false' && dom.getContentEditable(n) !== 'false') {
```
lnewson, line 15: Hmm... this logic doesn't look right. Wouldn't this prevent merging formats if you were only applying it inside a `contenteditable="true"` container (see below) either that or wouldn't it fail to walk up if nested a few levels deep? I think instead we need to check they are within the same contenteditable block.

```html
<p contentediable="false">noneditable <span contenteditable="true"><span style="color: red">editable</span> text</span></p>
```

modules/tinymce/src/core/main/ts/fmt/RemoveFormat.ts
```diff
@@ -623,18 +628,23 @@ const remove = (ed: Editor, name: string, vars?: FormatVars, node?: Node | Range
...
         }
       }
     }
 
-    Events.fireFormatRemove(ed, name, node, vars);
-    return;
+    if (formatRemoved) {
+      Events.fireFormatRemove(ed, name, node, vars);
+      return;
+    }
```
lnewson, line 647: Likewise here, this is breaking behaviour...

lnewson (changes requested):
Still need to look at the tests sorry Luke, but just submitting what I have atm. Also don't forget to add a changelog entry for this improvement/feature.

modules/tinymce/src/core/main/ts/api/OptionTypes.ts
lnewson: Another name that might work? `format_noneditable_selector` I know that drops the wrap, but it's not very clear what that applies to just from the name so I think we'd be better clarifying that via docs if needed instead 🤷

modules/tinymce/src/core/main/ts/fmt/ApplyFormat.ts
ltrouton: Good idea. Will do

modules/tinymce/src/core/main/ts/api/OptionTypes.ts
ltrouton: I think I like `format_noneditable_selector` the best so am happy to go with that.

ltrouton: The option name and return type has been updated

modules/tinymce/src/core/main/ts/fmt/ApplyFormat.ts
ltrouton: Thanks for the suggestion. I have set up some functions to avoid eager evaluation. Let me know if this still isn't quite what were you were thinking

modules/tinymce/src/core/main/ts/fmt/ApplyFormat.ts
ltrouton: Looking at the existing code again, I realised I don't actually need this variable so have just removed it

modules/tinymce/src/core/main/ts/fmt/ApplyFormat.ts
lnewson: This still isn't quite what I was thinking but it's close... these are all still inline functions and I was hoping we could move them out so it's easier to read.

modules/tinymce/src/core/main/ts/fmt/ApplyFormat.ts
ltrouton: The event should still be fired by the `Events.fireFormatApply` call near the end of the file even if a format wasn't applied. I could be misunderstanding something though.

lnewson: It won't look at your logic... it'll only fire if a format applied now or it's not a non-wrappable noneditable node. So if it's a wrappable non-editable element and a format doesn't match then it no longer fires. Basically we don't want these if conditions.

lnewson: Ahh wait, I see what you mean now. There is a second/separate call at the bottom of this. Hmm... I'll need to look more but this does worry me.

modules/tinymce/src/core/main/ts/fmt/ApplyFormat.ts
ltrouton: TODO has been looked into and handled now

modules/tinymce/src/core/main/ts/fmt/ApplyFormat.ts
ltrouton: I have moved them up. Thanks for pointing that out

modules/tinymce/src/core/main/ts/fmt/MergeFormats.ts
ltrouton: I have simplified the noneditable check a bit for this function so I think it is all good now

modules/tinymce/src/core/main/ts/fmt/ApplyFormat.ts
ltrouton: I have refactored and removed a fair bit of code in both ApplyFormat.ts and RemoveFormat.ts to remove the need to have the specific logic for when a non-editable element is selected. There is a risk by doing this but all of the tests pass which is a pretty good sign given there are a lot of formatted tests.

lnewson (commented):
Have a meeting to attend, so just leaving the comments I have so far.

modules/tinymce/src/core/main/ts/fmt/FormatUtils.ts
```diff
@@ -1,20 +1,30 @@
...
 const isNode = (node: any): node is Node => !!(node).nodeType;
 
+const isElementNode = (node: Node): node is Element =>
+  NodeType.isElement(node) && !Bookmarks.isBookmarkNode(node) && !isCaretNode(node) && !NodeType.isBogus(node);
+
 const isInlineBlock = (node: Node): boolean => {
   return node && /^(IMG)$/.test(node.nodeName);
 };
 
+const isContentEditable = (elm: HTMLElement): boolean =>
```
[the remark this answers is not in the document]
ltrouton: Good point. Will make the change

modules/tinymce/src/core/main/ts/fmt/MergeFormats.ts
```diff
@@ -33,26 +35,28 @@ const mergeBackgroundColorAndFontSize = (dom: DOMUtils, format: ApplyFormat, var
   // nodes with font-size should have their own background color as well to fit the line-height (see TINY-882)
   if (format.styles && format.styles.backgroundColor) {
     processChildElements(node,
-      hasStyle(dom, 'fontSize'),
+      (elm) => hasStyle(dom, 'fontSize')(elm) && FormatUtils.isContentEditable(elm),
```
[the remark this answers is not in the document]
ltrouton: I think I might lean more toward the latter mainly because the related `applyStyle` function still uses the currying.
I can do `const hasFontSize = hasStyle(dom, 'fontSize');` and use it that like you mentioned

modules/tinymce/src/core/main/ts/fmt/ApplyFormat.ts
```diff
@@ -188,21 +213,18 @@ const applyFormat = (ed: Editor, name: string, vars?: FormatVars, node?: Node |
...
-        ) {
+        if (canWrapNode(node, parentName, isEditableDescendant, isWrappableNoneditableElm)) {
           // Start wrapping
           if (!currentWrapElm) {
             // Wrap the node
             currentWrapElm = dom.clone(wrapElm, false) as Element;
             parentNode.insertBefore(currentWrapElm, node);
             newWrappers.push(currentWrapElm);
           }
+          // Wrappable noneditable element has been handled so go back to previous state
```
lnewson, line 224:
```suggestion

          // Wrappable noneditable element has been handled so go back to previous state
```

lnewson (approved):

PR #7969, opened 2022-07-13
TINY-8778: New `text_patterns_lookup` option

tiny-ben-tran:
Related Ticket: TINY-8778: 

Description of Changes:
* Added new option `text_patterns_lookup` which allows prepending additional patterns on the fly. The patterns take precedence over `text_patterns` during the matching process

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/api/OptionTypes.ts
+++ b/modules/tinymce/src/core/main/ts/api/OptionTypes.ts
@@ -1,5 +1,5 @@
 import { UploadHandler } from '../file/Uploader';
-import { DynamicPatternsLookup, Pattern, RawPattern } from '../textpatterns/core/PatternTypes';
+import { Pattern, RawDynamicPatternsLookup, RawPattern } from '../textpatterns/core/PatternTypes';
 import Editor from './Editor';
 import { PastePostProcessEvent, PastePreProcessEvent } from './EventTypes';
 import { Formats } from './fmt/Format';
@@ -192,7 +192,7 @@ interface BaseEditorOptions {
   table_tab_navigation?: boolean;
   target?: HTMLElement;
   text_patterns?: RawPattern[] | false;
-  text_patterns_lookup?: DynamicPatternsLookup;
+  text_patterns_lookup?: RawDynamicPatternsLookup;
   theme?: string | ThemeInitFunc | false;
   theme_url?: string;
   toolbar?: boolean | string | string[] | Array<ToolbarGroup>;
@@ -278,7 +278,7 @@ export interface EditorOptions extends NormalizedEditorOptions {
   preview_styles: string;
   readonly: boolean;
   text_patterns: Pattern[];
-  text_patterns_lookup: DynamicPatternsLookup;
+  text_patterns_lookup: RawDynamicPatternsLookup;
   visual: boolean;
   visual_anchor_class: string;
   visual_table_class: string;
--- a/modules/tinymce/src/core/main/ts/api/OptionTypes.ts
+++ b/modules/tinymce/src/core/main/ts/api/OptionTypes.ts
@@ -1,5 +1,5 @@
 import { UploadHandler } from '../file/Uploader';
-import { Pattern, RawPattern } from '../textpatterns/core/PatternTypes';
+import { DynamicPatternsLookup, Pattern, RawPattern } from '../textpatterns/core/PatternTypes';
 import Editor from './Editor';
 import { PastePostProcessEvent, PastePreProcessEvent } from './EventTypes';
 import { Formats } from './fmt/Format';
@@ -277,6 +277,7 @@ export interface EditorOptions extends NormalizedEditorOptions {
   preview_styles: string;
   readonly: boolean;
   text_patterns: Pattern[];
+  text_patterns_lookup?: DynamicPatternsLookup;
   visual: boolean;
   visual_anchor_class: string;
   visual_table_class: string;
--- a/modules/tinymce/src/core/main/ts/api/OptionTypes.ts
+++ b/modules/tinymce/src/core/main/ts/api/OptionTypes.ts
@@ -1,5 +1,5 @@
 import { UploadHandler } from '../file/Uploader';
-import { Pattern, RawDynamicPatternsLookup, RawPattern } from '../textpatterns/core/PatternTypes';
+import { DynamicPatternsLookup, Pattern, RawDynamicPatternsLookup, RawPattern } from '../textpatterns/core/PatternTypes';
 import Editor from './Editor';
 import { PastePostProcessEvent, PastePreProcessEvent } from './EventTypes';
 import { Formats } from './fmt/Format';
@@ -278,7 +278,7 @@ export interface EditorOptions extends NormalizedEditorOptions {
   preview_styles: string;
   readonly: boolean;
   text_patterns: Pattern[];
-  text_patterns_lookup: RawDynamicPatternsLookup;
+  text_patterns_lookup: DynamicPatternsLookup;
   visual: boolean;
   visual_anchor_class: string;
   visual_table_class: string;
--- a/modules/tinymce/src/core/main/ts/api/OptionTypes.ts
+++ b/modules/tinymce/src/core/main/ts/api/OptionTypes.ts
@@ -192,6 +192,7 @@ interface BaseEditorOptions {
   table_tab_navigation?: boolean;
   target?: HTMLElement;
   text_patterns?: RawPattern[] | false;
+  text_patterns_lookup?: DynamicPatternsLookup;
   theme?: string | ThemeInitFunc | false;
   theme_url?: string;
   toolbar?: boolean | string | string[] | Array<ToolbarGroup>;
@@ -277,7 +278,7 @@ export interface EditorOptions extends NormalizedEditorOptions {
   preview_styles: string;
   readonly: boolean;
   text_patterns: Pattern[];
-  text_patterns_lookup?: DynamicPatternsLookup;
+  text_patterns_lookup: DynamicPatternsLookup;
   visual: boolean;
   visual_anchor_class: string;
   visual_table_class: string;
--- a/modules/tinymce/src/core/main/ts/api/Options.ts
+++ b/modules/tinymce/src/core/main/ts/api/Options.ts
@@ -737,8 +737,7 @@ const register = (editor: Editor): void => {
     processor: (value) => {
       if (Type.isFunction(value)) {
         return {
-          // This (ctx => value(ctx)) isn't ETA reduced so that it compiles
-          value: Pattern.fromRawPatternsLookup((ctx) => value(ctx)),
+          value: Pattern.fromRawPatternsLookup(value as PatternTypes.RawDynamicPatternsLookup),
           valid: true,
         };
       } else {
--- a/modules/tinymce/src/core/main/ts/api/Options.ts
+++ b/modules/tinymce/src/core/main/ts/api/Options.ts
@@ -2,6 +2,7 @@ import { Arr, Obj, Strings, Type } from '@ephox/katamari';
 import { PlatformDetection } from '@ephox/sand';
 
 import * as Pattern from '../textpatterns/core/Pattern';
+import * as PatternTypes from '../textpatterns/core/PatternTypes';
 import DOMUtils from './dom/DOMUtils';
 import Editor from './Editor';
 import { EditorOptions } from './OptionTypes';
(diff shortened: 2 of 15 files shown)


modules/tinymce/src/core/test/ts/browser/textpatterns/FindBlockPatternTest.ts
```diff
@@ -0,0 +1,207 @@
...
+import { TinyHooks, TinySelections } from '@ephox/wrap-mcagar';
+import { assert } from 'chai';
+
+import Editor from 'tinymce/core/api/Editor';
+import * as Options from 'tinymce/core/api/Options';
+import * as BlockPattern from 'tinymce/core/textpatterns/core/BlockPattern';
+import * as Pattern from 'tinymce/core/textpatterns/core/Pattern';
+import { PatternSet } from 'tinymce/core/textpatterns/core/PatternTypes';
+
+const getPatternSetFor = (hook: TinyHooks.Hook<Editor>) => Thunk.cached((): PatternSet => {
```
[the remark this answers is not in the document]
lnewson: Also looking further, this seems to be largely duplicated with the `FindInlinePatternTest.ts` logic, so this should probably be declared in `TextPatternsUtils.ts` instead.

modules/tinymce/src/core/test/ts/browser/textpatterns/FindBlockPatternTest.ts
```diff
@@ -0,0 +1,207 @@
...
+            start: [ 0, 0, 0 ],
+            end: [ 0, 0, 0 ]
+          }
+        }
+      ],
+      'Checking block pattern matches'
+      );
+    });
+
+    it('TINY-8778: Does not trigger a match if the block tag does not match forced_root_block', () => {
```
lnewson, line 178: Likewise here, this isn't something that's expected so please either remove it or add a comment referencing TINY-4303.

modules/tinymce/src/core/test/ts/browser/textpatterns/FindBlockPatternTest.ts
```diff
@@ -0,0 +1,207 @@
...
+        { start: '#', format: 'h1' },
+        { start: '##', format: 'h2' },
+        { start: '###', format: 'h3' }
+      ],
+      base_url: '/project/tinymce/js/tinymce'
+    }, []);
+
+    const getPatternSet = getPatternSetFor(hook);
+
+    it('TINY-8778: # triggers heading match when block tag matches forced_root_block', () => {
```
lnewson, line 38: This does not relate to the new lookup feature, so it should not be using `TINY-8778` as the Jira reference. It should just be using `TBA` since we don't know what it applies to.

- [ ] Fix incorrect Jira number references in tests

modules/tinymce/src/core/test/ts/browser/textpatterns/FindInlinePatternTest.ts
```diff
@@ -17,285 +17,423 @@ interface ExpectedPatternMatch {
...
-
-  it('TBA: Check match when input pattern has an empty end value', () => {
-    const editor = hook.editor();
-    setContentAndCursor(editor, 'irl', [ 0 ], 3);
-    const matches = getInlinePattern(editor, inlinePatterns());
-    assertPatterns(matches, [
-      { pattern: { start: '', end: 'irl', value: 'in real life' }, startRng: { start: [ 0, 0 ], end: [ 0, 3 ] }, endRng: { start: [ 0, 0 ], end: [ 0, 3 ] }}
-    ]);
-  });
+const getInlinePatternSetFor = (hook: TinyHooks.Hook<Editor>) => Thunk.cached((): InlinePatternSet => {
```
lnewson, line 20: Likewise here, functions should not be outside the root `describe` block

- [ ] Fix test code style so that all test functions are within the describe/context block they apply to

modules/tinymce/src/core/main/ts/textpatterns/core/InlinePattern.ts
```diff
@@ -159,8 +163,27 @@ const findPatternsRec = (editor: Editor, patterns: InlinePattern[], node: Node,
...
+      // inline patterns are triggered via a space keyup, so spaces might be present.
+      // When triggering inline patterns via enter, we don't allow trailing spaces due to TINY-8779
+      allowTrailingSpaces: space
+    });
+    const dynamicPatterns = getInlinePatterns(extraPatterns);
+    // Dynamic patterns take precedence over static patterns
+    const patterns = [
+      ...dynamicPatterns,
+      ...patternSet.inlinePatterns,
+    ];
```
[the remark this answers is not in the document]
lnewson: Ignore this sorry, I got myself confused and this is correct.

modules/tinymce/src/core/main/ts/textpatterns/keyboard/Keyboard.ts
```diff
@@ -9,6 +9,8 @@ const setup = (editor: Editor): void => {
   const charCodes = [ ',', '.', ';', ':', '!', '?' ];
   const keyCodes = [ 32 ];
 
+  const isSelectionCollapsed = editor.selection.isCollapsed();
```
ephox-mogran, line 12: I don't think this is the right spot. This will just calculate if the selection is collapsed the first time it's setup. You want to do it every keydown event. So closer to line 28, or maybe line 24 to avoid the text pattern lookup.

ephox-mogran: Or alternatively, this should be a function, and then it will be invoked elsewhere. At the moment, it's a value.

ephox-mogran: But ideally, a test should fail because of this. So instead of just changing it, see if you can make a test that fails because it's in the wrong spot.

tiny-ben-tran: I have fixed it. I didn't move and convert it correctly.

modules/tinymce/src/core/main/ts/textpatterns/keyboard/Keyboard.ts
```diff
@@ -9,28 +9,50 @@ const setup = (editor: Editor): void => {
...
     if (e.keyCode === 13 && !VK.modifierPressed(e)) {
-      if (KeyHandler.handleEnter(editor, getPatternSet())) {
+      const patternSet = getPatternSet();
+      // Do not process anything if we don't have any inline patterns, block patterns,
+      // or dynamic lookup defined
+      const hasPatterns = patternSet.inlinePatterns.length > 0 ||
+        patternSet.blockPatterns.length > 0 ||
+        hasDynamicPatterns();
+
+      if (editor.selection.isCollapsed() && hasPatterns && KeyHandler.handleEnter(editor, patternSet)) {
```
lnewson, line 30: This still eagerly evaluates the patterns which we don't want if the selection is collapsed.

modules/tinymce/src/core/main/ts/textpatterns/keyboard/Keyboard.ts
tiny-ben-tran: FYI, `ReplacementTest` already covers that case. Its tests failed before when I had the isSelectionCollapsed checking in the wrong spot

modules/tinymce/src/core/main/ts/textpatterns/core/Pattern.ts
```diff
@@ -115,10 +116,23 @@ const fromRawPatterns = (patterns: RawPattern[]): Pattern[] => {
   return normalized.values;
 };
 
+const fromRawPatternsLookup = (lookupFn: (ctx: DynamicPatternContext) => RawPattern[]): DynamicPatternsLookup => {
+  return (ctx: DynamicPatternContext) => {
+    const rawPatterns = lookupFn({
+      text: ctx.text,
+      allowTrailingSpaces: ctx.allowTrailingSpaces,
+      block: ctx.block,
+    });
```
lnewson, line 125:
```suggestion
    const rawPatterns = lookupFn(ctx);
```
I believe this can just be simplified to this? I think the original intention was to make sure we didn't expose something public we didn't want to. If that was the case though then we should have a different type in `lookupFn`.

modules/tinymce/src/core/main/ts/api/OptionTypes.ts
```diff
@@ -192,6 +192,7 @@ interface BaseEditorOptions {
   table_tab_navigation?: boolean;
   target?: HTMLElement;
   text_patterns?: RawPattern[] | false;
+  text_patterns_lookup?: DynamicPatternsLookup;
```
lnewson, line 195: This type is incorrect... the pre-processed type needs to return a `RawPattern[]` and this will require it to be a `Pattern[]` instead. This should be the same type as what's used for `lookupFn` in `Pattern.ts`.

lnewson: Maybe we need to create a `RawDynamicPatternsLookup` type? For example:
```ts
export type RawDynamicPatternsLookup = (ctx: DynamicPatternContext) => RawPattern[];
```

modules/tinymce/src/core/main/ts/textpatterns/core/Pattern.ts
```diff
@@ -115,10 +116,19 @@ const fromRawPatterns = (patterns: RawPattern[]): Pattern[] => {
   return normalized.values;
 };
 
+const fromRawPatternsLookup = (lookupFn: (ctx: DynamicPatternContext) => RawPattern[]): DynamicPatternsLookup => {
+  return (ctx: DynamicPatternContext) => {
+    const rawPatterns = lookupFn(ctx);
+
+    return fromRawPatterns(rawPatterns);
```
lnewson, line 123:
```suggestion
    const rawPatterns = lookupFn(ctx);
    return fromRawPatterns(rawPatterns);
```

modules/tinymce/src/core/test/ts/browser/textpatterns/FindBlockPatternsTest.ts
```diff
@@ -0,0 +1,198 @@
...
+        { start: '*', end: '*', format: 'italic' },
+        { start: '**', end: '**', format: 'bold' },
+        { start: '#', format: 'h1' },
+        { start: '##', format: 'h2' },
+        { start: '###', format: 'h3' }
+      ],
+      base_url: '/project/tinymce/js/tinymce'
+    }, []);
+
+    const getPatternSet = getPatternSetFor(hook);
```
lnewson, line 26:
```suggestion
    const getPatternSet = getPatternSetFor(hook);

```

modules/tinymce/src/core/test/ts/browser/textpatterns/FindInlinePatternTest.ts
lnewson: @tiny-ben-tran looks like this hasn't been fixed yet.

modules/tinymce/src/core/main/ts/textpatterns/core/BlockPattern.ts
```diff
@@ -73,8 +74,21 @@ const findPatterns = (editor: Editor, patterns: BlockPattern[]): BlockPatternMat
...
+      // Block patterns do not allow trailing spaces currently. This is related to TINY-8779
+      allowTrailingSpaces: false
+    });
+    // search in the dynamic patterns first
+    const patterns = getBlockPatterns(extraPatterns);
+    const matchedPattern = findPattern(patterns, blockText).orThunk(() => {
+      // Search in the static patterns
+      const patterns = getBlockPatterns(patternSet.blockPatterns);
+      return findPattern(patterns, blockText);
+    });
```
TheSpyder, line 90: I've been a little concerned about the performance implications of textpattern for a while, but this is looking more and more expensive.

Can we please log a technical debt task under either this epic or the small changes area to improve the way this operates?

Dynamic patterns will always be more expensive, that's expected, but I see a lot of logic that could be run in response to option changes instead of re-loaded and re-analysed every time a trigger key is pressed.

TheSpyder (approved):

modules/tinymce/src/core/main/ts/api/Options.ts
```diff
@@ -732,6 +733,21 @@ const register = (editor: Editor): void => {
     ]
   });
 
+  registerOption('text_patterns_lookup', {
+    processor: (value) => {
+      if (Type.isFunction(value)) {
+        return {
+          // This (ctx => value(ctx)) isn't ETA reduced so that it compiles
```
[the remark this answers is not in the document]
tiny-ben-tran: @TheSpyder, I have used a new type to make this look a bit better. Thought?

modules/tinymce/src/core/main/ts/textpatterns/core/BlockPattern.ts
tiny-ben-tran: @TheSpyder , I created TINY-8904 as per your request. Can you please have a look as I think additional info would be helpful.

lnewson (approved):

modules/tinymce/src/core/main/ts/api/Options.ts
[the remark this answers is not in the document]
TheSpyder: Yeah, that'll do 🙂

PR #7972, opened 2022-07-13
TINY-8828: Prevent selection of cef content with the mouse on Firefox

vpyshnenko:
Related Ticket: TINY-8828

Description of Changes:
* add css rule to forbid the user to select the text inside of cef element
*  add css rule to retain the ability to select the content of cet element wrapped with cef element on Safari

I'm struggling to simulate dragging the mouse from an adjacent element to cef content, so probably this issue can be a candidate for manual testing by QA

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/oxide/src/less/theme/content/contenteditable/contenteditable.less
+++ b/modules/oxide/src/less/theme/content/contenteditable/contenteditable.less
@@ -40,6 +40,7 @@
     cursor: text;
   }
 
+  // needed to select the content of cet wrapped in cef on Safari, should be fixed in Safari 16
   *[contentEditable=false] *[contentEditable=true] {
     user-select: text;
   }
--- a/modules/oxide/src/less/theme/content/contenteditable/contenteditable.less
+++ b/modules/oxide/src/less/theme/content/contenteditable/contenteditable.less
@@ -33,9 +33,14 @@
 
   *[contentEditable=false] {
     cursor: default;
+    user-select: none;
   }
 
   *[contentEditable=true] {
     cursor: text;
+  }
+
+  *[contentEditable=false] *[contentEditable=true] {
+    user-select: text;
   }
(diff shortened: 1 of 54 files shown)


lnewson:
I've converted this to a draft as it doesn't look ready to review yet.

modules/oxide/src/less/theme/content/contenteditable/contenteditable.less
```diff
@@ -33,9 +33,14 @@
...
   *[contentEditable=false] {
     cursor: default;
+    user-select: none;
   }
 
   *[contentEditable=true] {
     cursor: text;
   }
+
+  *[contentEditable=false] *[contentEditable=true] {
```
TheSpyder, line 43: Couldn't this just go in the `*[contentEditable=true]` block? `user-select` is not inherited so it should still work the same way `cursor` does.

- [ ] consider moving `user-select` into the `true` block

TheSpyder (approved):

modules/oxide/src/less/theme/content/contenteditable/contenteditable.less
vpyshnenko: It seems there is an exception for Safari: https://fiddle.tiny.cloud/yiiaab/7
Without  `*[contentEditable=false] *[contentEditable=true] `rule  I can not select content of nested `cet` element

TheSpyder: I can select text in both strong tags, but I'm running Safari 16 beta. I guess it'll be another year or so before we can rely on that. Maybe leave a note about it in a comment? Doesn't seem worth making a task for.

lorenzo-pomili (approved):

PR #7976, opened 2022-07-14
TINY-8877: Fixed a regression that caused an exception to be thrown when deleting all content

lnewson:
Related Ticket: TINY-8873

Description of Changes:
* Fixed a regression caused by https://github.com/tinymce/tinymce/pull/7892. The problem is that in `DeleteUtils.ts` the `getParent` call can return `null` and will if the node passed in is the root/body node. So this corrects that logic and also makes the delete logic properly handle padding an empty editor by inserting the root block.
  * Note: This exception actually isn't thrown in `develop` because the strict types made me do a similar fix (though it didn't handle padding the root/body properly). So +1 for strict types I guess?
* I also cleaned up some duplication for other cases where I found we were inserting the root block in the delete/navigation logic, so we now hopefully only have the one bit of code for generating that.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/delete/DeleteUtils.ts
+++ b/modules/tinymce/src/core/main/ts/delete/DeleteUtils.ts
@@ -36,16 +36,9 @@ const getParentBlock = (rootNode: SugarElement<Node>, elm: SugarElement<Node>): 
   }
 };
 
-const placeCaretInEmptyBody = (editor: Editor): void => {
-  const body = editor.getBody();
-  const node = body.firstChild && editor.dom.isBlock(body.firstChild) ? body.firstChild : body;
-  editor.selection.setCursorLocation(node, 0);
-};
-
-const paddEmptyBody = (editor: Editor): void => {
+const paddEmptyBody = (editor: Editor, moveSelection: boolean = true): void => {
   if (editor.dom.isEmpty(editor.getBody())) {
-    editor.setContent('');
-    placeCaretInEmptyBody(editor);
+    editor.setContent('', { no_selection: !moveSelection });
   }
 };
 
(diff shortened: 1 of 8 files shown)


modules/tinymce/src/core/main/ts/delete/DeleteUtils.ts
```diff
@@ -78,8 +71,11 @@ const deleteRangeContents = (editor: Editor, rng: Range, root: SugarElement<HTML
   rng.deleteContents();
   // Pad the last block node
   const lastNode = freefallRtl(root).getOr(root);
-  const lastBlock = SugarElement.fromDom(editor.dom.getParent(lastNode.dom, editor.dom.isBlock));
-  if (Empty.isEmpty(lastBlock)) {
+  const lastBlock = SugarElement.fromDom(editor.dom.getParent(lastNode.dom, editor.dom.isBlock) ?? root.dom);
+  // If the block is the editor body then we need to insert the root block as well
+  if (lastBlock.dom === editor.getBody()) {
+    paddEmptyBody(editor, moveSelection);
```
lnewson, line 77: @ltrouton pointed out today that this function already existed, so the `ForceBlocks.insertNewLine` code is no longer required for this fix. However, as it does remove duplication I think it's worth keeping. If anyone would prefer it to be reverted though let me know and I'll do it.

modules/tinymce/src/core/main/ts/delete/DeleteUtils.ts
```diff
@@ -36,16 +36,9 @@ const getParentBlock = (rootNode: SugarElement<Node>, elm: SugarElement<Node>):
...
-  const body = editor.getBody();
-  const node = body.firstChild && editor.dom.isBlock(body.firstChild) ? body.firstChild : body;
-  editor.selection.setCursorLocation(node, 0);
-};
-
-const paddEmptyBody = (editor: Editor): void => {
+const paddEmptyBody = (editor: Editor, moveSelection: boolean = true): void => {
   if (editor.dom.isEmpty(editor.getBody())) {
-    editor.setContent('');
-    placeCaretInEmptyBody(editor);
```
lnewson, line 48: Removed this, as it was unnecessary since `setContent` does set the cursor position nowadays, plus I believe this wouldn't have been the correct behaviour in some edge cases anyways.

tiny-ben-tran (approved):

ltrouton (approved):

TheSpyder (approved):

PR #7981, opened 2022-07-18
TINY-8779: Fixed text pattern throwing a range error when there were fragmented text nodes in a paragraph

tiny-ben-tran:
Related Ticket: TINY-8779

Description of Changes:
* Fixed a bug where the text pattern logic threw an error when there were two or more text nodes in a paragraph

Pre-checks:
* [x] Changelog entry added
* [x] Tests have been added (if applicable)
* [x] Branch prefixed with `feature/`, `hotfix/` or `spike/`

Review:
* [x] Milestone set

GitHub issues (if applicable):

PR #7985, opened 2022-07-20
TINY-8779: Fixed text pattern throwing a range error when there were fragmented text nodes in a paragraph

lnewson:
Related Ticket: TINY-8779

Description of Changes:

This is an alternative take on https://github.com/tinymce/tinymce/pull/7981 that I've discussed with @tiny-ben-tran as I was concerned about us normalizing the DOM on every enter keypress. So this instead makes the path range logic return a normalized range when requested so that the enter lookups will work after it's been normalized by the undomanager actions (similar to how the undo bookmark manager works and yes there is some duplication for which I've logged TINY-8909).

Additionally, this also cleans up all the old TODOs attached to this Jira which was something that had been missed.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


tiny-ben-tran (approved):

ephox-mogran (approved):

TheSpyder (approved):
Much better - I would've blocked the original PR for the performance cost.

PR #7991, opened 2022-07-21
TINY-8778: Fixed `text_patterns_lookup` being called more than expected when matching patterns

tiny-ben-tran:
Related Ticket: TINY-8778

Description of Changes:
* This PR address the issue,  which was found in the QA of TINY-8778,  that `text_patterns_lookup` was called more than expected (> 1 time) when matching a pattern on pressing enter or space key.
* Also reduced the number of sorting patterns

@TheSpyder, this might also address TINY-8904

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/textpatterns/core/InlinePattern.ts
+++ b/modules/tinymce/src/core/main/ts/textpatterns/core/InlinePattern.ts
@@ -8,7 +8,7 @@ import { createMarker, Marker, rangeFromMarker, removeMarker } from '../utils/Ma
 import { generatePathRange, generatePathRangeFromRange } from '../utils/PathRange';
 import * as Utils from '../utils/Utils';
 import { getInlinePatterns } from './Pattern';
-import { InlinePattern, InlinePatternMatch, InlinePatternSet } from './PatternTypes';
+import { InlinePattern, InlinePatternMatch, InlinePatternSet, Pattern } from './PatternTypes';
 
 interface PatternDetails {
   readonly pattern: InlinePattern;
@@ -261,7 +261,7 @@ const addMarkers = (dom: DOMUtils, matches: InlinePatternMatch[]): InlinePattern
   }, [] as InlinePatternMatchWithMarkers[]);
 };
 
-const findPatterns = (editor: Editor, patternSet: InlinePatternSet, normalizedMatches: boolean, space: boolean): InlinePatternMatch[] => {
+const findPatterns = (editor: Editor, patternSet: InlinePatternSet, dynamicPattern: Pattern[], normalizedMatches: boolean, space: boolean): InlinePatternMatch[] => {
   const rng = editor.selection.getRng();
   if (!rng.collapsed) {
     return [];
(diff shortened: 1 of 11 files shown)


modules/tinymce/src/core/main/ts/textpatterns/core/InlinePattern.ts
```diff
@@ -167,18 +166,6 @@ const findPatternsRec = (
     rng.setEnd(node, offset);
     const text = rng.toString();
 
-    // TINY-8781: TODO: text_patterns should announce their changes for accessibility
-    const extraPatterns = patternSet.dynamicPatternsLookup({
-      text,
```
lnewson, line 172: Hmm... I hadn't noticed this when discussing last week sorry Ben, but the new implementation returns the whole block not from the start of the trigger character back to the first element. Is that going to be an issue?

I think it is and the inline key logic needs to do something like this instead of the whole block, as if we trigger something in the middle of a paragraph then we really only want to deal with where the text pattern is starting to match, not the whole block.

tiny-ben-tran: It is an issue. So I changed the code to only pass the text from the beginning to the cursor location into `dynamicPatternsLookup` when matching inline patterns. Also added a test where matching is triggered in the middle of the paragraph. 

Not sure whether passing extra info such as matching mode (inline or block) to the `dynamicPatternsLookup` would be helpful. Maybe leave it for now until there is a clear use case for it

modules/tinymce/src/core/test/ts/browser/textpatterns/FindInlinePatternTest.ts
```diff
@@ -217,7 +224,7 @@ describe('browser.tinymce.textpatterns.FindInlinePatternTest', () => {
       it('TINY-8778: force only ** pattern and test return on not existing *** pattern', () => {
         const editor = hook.editor();
         setContentAndCursor(editor, '***x***', [ 0 ], 7);
-        const matches = getInlinePattern(editor, inlinePatternSet);
+        const matches = getInlinePattern(editor, inlinePatternSet as PatternSet);
```
lnewson, line 227: Why is the case needed here?

modules/tinymce/src/core/test/ts/browser/textpatterns/FindInlinePatternTest.ts
```diff
@@ -326,7 +333,7 @@ describe('browser.tinymce.textpatterns.FindInlinePatternTest', () => {
               start: '*', end: '*', format: 'bold'
             }
           ];
-        } else if (parentTag === 'div' && ctx.text === 'replace-me') {
+        } else if (parentTag === 'div' && ctx.text.includes('replace-me')) {
```
lnewson, line 336: This won't test what you want properly for the `Inline pattern matches text in the middle` test... because this would still match if you passed the entire block text.

- [ ] Add or fix the test to that ensures `replace-me` (or similar) isn't returned if it's after the trigger point

modules/tinymce/src/core/main/ts/textpatterns/utils/Utils.ts
```diff
@@ -36,8 +38,31 @@ const deleteRng = (dom: DOMUtils, rng: Range, isRoot: (e: Node) => boolean, clea
...
+
+  // dynamic patterns take precedence here
+  return {
+    ...patternSet,
+    blockPatterns: getBlockPatterns(dynamicPatterns).concat(patternSet.blockPatterns),
+    inlinePatterns: getInlinePatterns(dynamicPatterns).concat(patternSet.inlinePatterns)
+  };
+};
+
+const getBeforeText = (dom: DOMUtils, block: Node, node: Node, offset: number): string => {
```
lnewson, line 55:
```suggestion
const getBeforeText = (dom: DOMUtils, block: Element, node: Node, offset: number): string => {
```

lnewson (approved):

TheSpyder (approved):
That does look like it will perform better. We can always do more detailed performance measuring if it becomes a problem.

modules/tinymce/src/core/test/ts/browser/textpatterns/FindInlinePatternTest.ts
tiny-ben-tran: Didn't clean up after the test was changed

modules/tinymce/src/core/test/ts/browser/textpatterns/FindInlinePatternTest.ts
tiny-ben-tran: Have changed the logic and added another test

PR #7996, opened 2022-07-22
TINY-8861: fix cursor position after pasting a `codesample` snippet

lorenzo-pomili:
Related Ticket: TINY-8861

Description of Changes:
The problem here is that copying and pasting snippet generated from `codesample` plugin the cursor is placed in the wrong place so pressing enter the row is added into the code and not outside.

the solution that I tried is to replicate logic similar to the one used when it is inserted a new [snippet](https://github.com/tinymce/tinymce/blob/develop/modules/tinymce/src/plugins/codesample/main/ts/core/CodeSample.ts#L25), on `SetContent` event of `codesimple` plugin

Pre-checks:
* [x] Changelog entry added

Review:
* [ ] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - When a sidebar was opened using the `sidebar_show` option, its associated toggle button was not highlighted #TINY-8873
 - The `autolink` plugin when converting a URL to a link did not fire an `ExecCommand` event, nor did it create an undo level #TINY-8896
 - Worked around a Firefox bug whereby cookies weren't available inside the editor content #TINY-88916
+- The cursor position after pasting a `codesimple` snippet is now in the correct position #TINY-8861
 
 ## 6.1.0 - 2022-06-29
 
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -12,7 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - When a sidebar was opened using the `sidebar_show` option, its associated toggle button was not highlighted #TINY-8873
 - The `autolink` plugin when converting a URL to a link did not fire an `ExecCommand` event, nor did it create an undo level #TINY-8896
 - Worked around a Firefox bug whereby cookies weren't available inside the editor content #TINY-88916
-- The cursor position after pasting a `codesimple` snippet is now in the correct position #TINY-8861
+- The cursor position after pasting or setting content with a `codesample` snippet no longer places the selection inside the non-editable codesample #TINY-8861
 
 ## 6.1.0 - 2022-06-29
 
--- a/modules/tinymce/src/plugins/codesample/main/ts/core/FilterContent.ts
+++ b/modules/tinymce/src/plugins/codesample/main/ts/core/FilterContent.ts
@@ -6,6 +6,9 @@ import Tools from 'tinymce/core/api/util/Tools';
 import * as Prism from '../prism/Prism';
 import * as Utils from '../util/Utils';
 
+const isPre = (node: Node): node is HTMLPreElement =>
+  node.nodeName.toLowerCase() === 'pre';
+
 const setup = (editor: Editor): void => {
   editor.on('PreProcess', (e) => {
     const dom = editor.dom;
@@ -46,6 +49,11 @@ const setup = (editor: Editor): void => {
           Prism.get(editor).highlightElement(elm);
           elm.className = Strings.trim(elm.className);
         });
+
+        const endContainer = editor.selection.getRng().endContainer;
+        if (isPre(endContainer)) {
+          editor.selection.select(endContainer);
+        }
       });
     }
   });
--- a/modules/tinymce/src/plugins/codesample/main/ts/core/FilterContent.ts
+++ b/modules/tinymce/src/plugins/codesample/main/ts/core/FilterContent.ts
@@ -6,9 +6,6 @@ import Tools from 'tinymce/core/api/util/Tools';
 import * as Prism from '../prism/Prism';
 import * as Utils from '../util/Utils';
 
-const isPre = (node: Node): node is HTMLPreElement =>
-  node.nodeName.toLowerCase() === 'pre';
-
 const setup = (editor: Editor): void => {
   editor.on('PreProcess', (e) => {
     const dom = editor.dom;
@@ -18,6 +15,7 @@ const setup = (editor: Editor): void => {
 
       dom.setAttrib(elm, 'class', Strings.trim(dom.getAttrib(elm, 'class')));
       dom.setAttrib(elm, 'contentEditable', null);
+      dom.setAttrib(elm, 'data-mce-highlighted', null);
 
       // Empty the pre element
       let child: Node;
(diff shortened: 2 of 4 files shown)


modules/tinymce/src/plugins/codesample/main/ts/core/FilterContent.ts
```diff
@@ -46,6 +49,11 @@ const setup = (editor: Editor): void => {
           Prism.get(editor).highlightElement(elm);
           elm.className = Strings.trim(elm.className);
         });
+
+        const endContainer = editor.selection.getRng().endContainer;
+        if (isPre(endContainer)) {
+          editor.selection.select(endContainer);
+        }
```
lnewson, line 56: This isn't the correct action for this as it shouldn't highlight the node for this usecase, it should put the selection beside it and it's also going to fail in some cases whereby the selection could be inside another element inside the non-editable pre element.

So the root cause of this bug appears instead to be that it's not setting the contenteditable state when inserting and instead is trying to modify it after. That's never going to end well, so instead what we need to do to fix this is ensure we use a node filter to add the `contenteditable="false"` attribute instead.

modules/tinymce/src/plugins/codesample/test/ts/webdriver/CodeSampleCopyAndPasteTest.ts
```diff
@@ -0,0 +1,82 @@
...
+    editor.dispatch('beforeinput', getInputEventArgs('beforeinput'));
+    editor.dispatch('input', getInputEventArgs('input'));
+    editor.dispatch('keyup', getKeyboardEventArgs('keyup'));
+  };
+
+  beforeEach(function () {
+    const browser = PlatformDetection.detect().browser;
+    if (browser.isSafari()) {
+      this.skip();
+    }
```
lnewson, line 34: Skipping a browser just because it fails tests isn't the right approach... we need to find out why it's failing and fix it instead of doing this.

modules/tinymce/src/plugins/codesample/main/ts/core/FilterContent.ts
lnewson:
```typescript
  editor.on('SetContent', () => {
    const dom = editor.dom;
    const unprocessedCodeSamples = Tools.grep(dom.select('pre'), (elm) => {
      return Utils.isCodeSample(elm) && dom.getAttrib(elm, 'data-mce-highlighted') !== 'true';
    });

    if (unprocessedCodeSamples.length) {
        Tools.each(unprocessedCodeSamples, (elm) => {
          Tools.each(dom.select('br', elm), (elm) => {
            dom.replace(editor.getDoc().createTextNode('\n'), elm);
          });

          elm.innerHTML = dom.encode(elm.textContent ?? '');
          Prism.get(editor).highlightElement(elm);
          dom.setAttrib(elm, 'data-mce-highlighted', true);
          elm.className = Strings.trim(elm.className);
        });
      });
    }
  });

  editor.on('PreInit', () => {
    editor.parser.addNodeFilter('pre', (nodes) => {
      for (let i = 0, l = nodes.length; i < l; i++) {
        const node = nodes[i];
        const isCodeSample = (node.attr('class') ?? '').indexOf('language-') !== -1 ;
        if (isCodeSample) {
          node.attr('contenteditable', 'false');
          node.attr('data-mce-highlighted', 'false');
        }
      }
    });
  });
```
Here is something that I spun up quickly that I believe should resolve this issue. As you can see, in the above we add a node filter to the parser to find codesample pre elements and add the contenteditable state there. This ensures that it'll always be treated as a CEF element and other actions that occur when inserting won't try and modify it either.

Now we still need some way to detect that the codesample needs processing, which is why I've added the `data-mce-highlighted` attribute to track what needs highlighting. Doing this also need `dom.setAttrib(elm, 'data-mce-highlighted', null);` or similar added to the `PreProcess` step above, but I've excluded that just to keep this shorter..

modules/tinymce/src/plugins/codesample/test/ts/webdriver/CodeSampleCopyAndPasteTest.ts
```diff
@@ -0,0 +1,82 @@
...
+    TinyAssertions.assertContent(editor,
+      '<p><br></p>' +
+      '<p><br></p>' +
+      '<pre class="language-markup" contenteditable="false">test content</pre>' +
+      '<pre class="language-markup" contenteditable="false">test content</pre>',
+      { format: 'raw' }
+    );
+  });
+
+  it('TINY-8861: coping and pating a piece of code and a text should leave the cursor on the text after paste', async () => {
```
lnewson, line 58:
```suggestion
  it('TINY-8861: copying and pasting a piece of code and a text should leave the cursor on the text after paste', async () => {
```

modules/tinymce/src/plugins/codesample/test/ts/webdriver/CodeSampleCopyAndPasteTest.ts
```diff
@@ -0,0 +1,82 @@
...
+  const pressEnter = (editor: Editor) => {
+    const overrides = { keyCode: Keys.enter() };
+
+    const getInputEventArgs = (eventType: string) => ({ ...new window.InputEvent(eventType), ...overrides });
+    const getKeyboardEventArgs = (eventType: string) => ({ ...new window.KeyboardEvent(eventType), ...overrides });
+
+    editor.dispatch('keydown', getKeyboardEventArgs('keydown'));
+    editor.dispatch('beforeinput', getInputEventArgs('beforeinput'));
+    editor.dispatch('input', getInputEventArgs('input'));
+    editor.dispatch('keyup', getKeyboardEventArgs('keyup'));
```
lnewson, line 27: You pretty much should never be doing this, especially in a webdriver test. In a webdriver test we should be using `RealKeys` and elsewhere we should be using `Keyboard` to do that.

The only time we don't use them is if we need something else to manually happen between the keyboard events (in Permanent Pen for example Agar doesn't fire the input event, though that is something we need to add to Agar).

modules/tinymce/src/plugins/codesample/test/ts/webdriver/CodeSampleCopyAndPasteTest.ts
```diff
@@ -0,0 +1,82 @@
...
+import { beforeEach, describe, it } from '@ephox/bedrock-client';
+import { PlatformDetection } from '@ephox/sand';
+import { TinyAssertions, TinyHooks, TinySelections } from '@ephox/wrap-mcagar';
+
+import Editor from 'tinymce/core/api/Editor';
+import Plugin from 'tinymce/plugins/codesample/Plugin';
+
+import * as TestUtils from '../module/CodeSampleTestUtils';
+
+describe('webdriver.tinymce.plugins.codesample.CodeSampleCopyAndPasteTest', () => {
```
lnewson, line 11: We also need a test here that does just a plain `setContent()` call to make sure it works correctly as well. As such I'd also suggest renaming this to `CodeSampleInsertTest` or similar...

modules/tinymce/CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Selecting a figure with `class="image"` would incorrectly highlight the link toolbar button #TINY-8832
 - Fixed various issues that occurred when formatting `contenteditable` elements. #TINY-8905
 - The text pattern logic threw an error when there were fragmented text nodes in a paragraph #TINY-8779
+- The cursor position after pasting a `codesimple` snippet is now in the correct position #TINY-8861
```
lnewson, line 25:
```suggestion
- The cursor position after pasting or setting content with a `codesample` snippet no longer places the selection inside the non-editable codesample #TINY-8861
```

lnewson (changes requested):

lnewson:
Another thing... you've used the wrong base branch and target branch here. If you look at the Jira this was supposed to be for TinyMCE 6.1.1 which means you should have used `master` not `develop`. However, as this is unlikely to make code freeze for 6.1.1 at this stage, we maybe should leave this and push this fix out to 6.2. I believe this discussion will be happing between PO's today sometime.

modules/tinymce/src/plugins/codesample/test/ts/webdriver/CodeSampleCopyAndPasteTest.ts
lorenzo-pomili: done

modules/tinymce/src/plugins/codesample/test/ts/webdriver/CodeSampleCopyAndPasteTest.ts
lorenzo-pomili: done

modules/tinymce/CHANGELOG.md

modules/tinymce/src/plugins/codesample/test/ts/webdriver/CodeSampleCopyAndPasteTest.ts
lnewson: @lorenzo-pomili sorry I missed something in the above... the other thing we should use is the McAgar helper functions. So we could use `TinyContentActions.keystroke(editor, Keys.enter());` here.

Also as a side note, if it's not for TinyMCE, there's the `Keyboard.activeKeystroke` API as well instead of manually getting the focused element and passing it.

lorenzo-pomili: ok, I changed it, thanks

TheSpyder:
Branch name is wrong, needs a new PR, but it has to be done before @lorenzo-pomili wakes up. I'll take care of it.

PR #7997, opened 2022-07-22
TINY-8887: Add autocompleter multi-character trigger support

vpyshnenko:
Related Ticket: TINY-8887

Description of Changes:
* `ch` option is replaced with new `trigger` option
* replaced the `findChar` function with the` findTrigger` function that searches for trigger substring instead of one character trigger.
* added ticket for docs update TINY-1774

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/tinymce/src/core/main/ts/autocomplete/AutocompleteContext.ts
```diff
@@ -14,17 +14,18 @@ export interface AutocompleteContext {
...
 
   for (i = index - 1; i >= 0; i--) {
-    const char = text.charAt(i);
+    const char = str.charAt(i);
     if (isWhitespace(char)) {
       return Optional.none();
     }
 
-    if (char === ch) {
+    if (str.includes(ch, i)) {
```
lnewson, line 28: This should be using `Strings.contains` from Katamari instead.

modules/tinymce/src/core/test/ts/atomic/autocomplete/ParserTest.ts
```diff
@@ -1,25 +1,45 @@
 import { describe, it } from '@ephox/bedrock-client';
 import { assert } from 'chai';
 
-import * as AutocompleteContext from 'tinymce/core/autocomplete/AutocompleteContext';
+import { findTrigger } from 'tinymce/core/autocomplete/AutocompleteContext';
```
lnewson, line 4: This should be using a named import like it was previously.

lnewson (changes requested):
This was a good attempt, but this needs to use a new property in the autocompleter as it's `char` which is very specifically not a plural... if this is accepting multiple characters then it needs a new property instead. Maybe we need to deprecate the `char` property and only use it as a fallback?

Another issue I see with this, is this logic is going to fail to detect characters split over multiple text nodes. For example, say you had 2 fragmented text nodes such as `@` and `@abc` then this will fail to detect `@@` as a trigger since it's only looking across a single text node at any one time.

- [ ] Add a test for fragmented text node matching
- [ ] Add a new property to replace the singular `char` autocompleter property

modules/tinymce/src/core/main/ts/autocomplete/AutocompleteContext.ts
vpyshnenko: `Strings.contains` doesn't support necessary here `position` parameter

lnewson: Then maybe we should change it 😉

modules/katamari/src/main/ts/ephox/katamari/api/Strings.ts
```diff
@@ -38,8 +39,12 @@ export const ensureTrailing = (str: string, suffix: string): string => {
   return endsWith(str, suffix) ? str : StrAppend.addToEnd(str, suffix);
 };
 
-export const contains = (str: string, substr: string): boolean => {
-  return str.indexOf(substr) !== -1;
+export const contains = (str: string, substr: string, start: number = 0, end?: number): boolean => {
+  if (Type.isUndefined(end) && start === 0) {
+    return str.indexOf(substr) !== -1;
+  } else {
+    return str.substring(start, end).indexOf(substr) !== -1;
```
HAFRMO, line 46: Couldn't this use the `position`-parameter of the `indexOf`-function, instead of generating new strings? Maybe javascript substring generation is more efficient than I give it credit for or the `position`-parameter introduces more complexity than it solves, but it feels like that might be faster.

modules/bridge/src/main/ts/ephox/bridge/components/content/Autocompleter.ts
```diff
@@ -87,4 +88,4 @@ export const createAutocompleterItem = (spec: AutocompleterItemSpec): Result<Aut
   StructureSchema.asRaw<AutocompleterItem>('Autocompleter.Item', autocompleterItemSchema, spec);
 
 export const createAutocompleter = (spec: AutocompleterSpec): Result<Autocompleter, StructureSchema.SchemaError<any>> =>
-  StructureSchema.asRaw<Autocompleter>('Autocompleter', autocompleterSchema, spec);
+  StructureSchema.asRaw<Autocompleter>('Autocompleter', autocompleterSchema, { ...spec, trigger: Obj.get(spec, 'trigger').getOr(spec.ch) });
```
HAFRMO, line 91: Is there any plans or issues to remove this stopgap fix in the future? It's for backwards compatibility, sure, but we'd probably prefer to remove it at some point, and not forgetting would be nice.

modules/katamari/src/main/ts/ephox/katamari/api/Strings.ts
vpyshnenko: Good point! So I will resort to `substring` only when the `end` parameter is applied.

modules/bridge/src/main/ts/ephox/bridge/components/content/Autocompleter.ts
vpyshnenko: I've created TINY-8929 for that

modules/katamari/src/main/ts/ephox/katamari/api/Strings.ts
lnewson: This could actually be optimised further I believe. For example:
```ts
const idx = str.indexOf(substr, start);
if (idx !== -1) {
  return Type.isUndefined(end) ? true : idx + substr.length < end;
} else {
  return false;
}
```
This would safe the overhead of having to do the `substring` call all the time when `end` is passed since we can just compare if the matched index is within the bounds of the end.

modules/bridge/src/main/ts/ephox/bridge/components/content/Autocompleter.ts
lnewson: Could this be simplified to this?
```suggestion
  StructureSchema.asRaw<Autocompleter>('Autocompleter', autocompleterSchema, { trigger: spec.ch, ...spec });
```

modules/katamari/src/main/ts/ephox/katamari/api/Strings.ts
vpyshnenko: there is no limit to perfection...

TheSpyder: Indeed. I would have written that third line as `return Type.isUndefined(end) || idx + substr.length <= end;` 😂

modules/katamari/src/test/ts/atomic/api/str/ContainsTest.ts
```diff
@@ -52,4 +52,21 @@ describe('atomic.katamari.api.str.ContainsTest', () => {
...
+    };
+
+    check(true, 'abc', 'b', 0);
+    check(true, 'abc', 'b', 1);
+    check(false, 'abc', 'b', 2);
+    check(false, 'abc', 'b', 0, 0);
+    check(true, 'abc', 'b', 0, 2);
+    check(true, 'abc', 'b', 1, 3);
+    check(false, 'abc', 'b', 2, 3);
+    check(false, 'abc', 'b', 0, 1);
```
TheSpyder, line 69: please add a test for cases where the `substr` parameter is longer than the remaining string

for example `check(false, 'abc', 'bc', 1, 2);` should do it?

modules/bridge/src/main/ts/ephox/bridge/components/content/Autocompleter.ts
```diff
@@ -30,7 +30,9 @@ export interface AutocompleterItem {
 
 export interface AutocompleterSpec {
   type?: 'autocompleter';
-  ch: string;
+  // TODO: TINY-8929: Remove 'trigger' fallback to 'ch'
+  ch?: string;
+  trigger: string;
```
TheSpyder, line 35: If I'm reading this right it will be a breaking change, at least for TypeScript customers, requiring third party implementations to add `trigger`?

I don't know if there's a nice way to make this migration. Setting them both to optional will require extra validation to make sure at least one of them is supplied.

TheSpyder (changes requested):

modules/bridge/src/main/ts/ephox/bridge/components/content/Autocompleter.ts
vpyshnenko: I made it optional in autocompleter spec, but mandatory in autocompleter schema, so if none of this is applied there will be a runtime error: `Could not find valid *required* value for "trigger"`

TheSpyder: That's the way to do it 👍 

I'm not totally happy with the error message but that's just what we get out of boulder.

TheSpyder (approved):

modules/bridge/src/main/ts/ephox/bridge/components/content/Autocompleter.ts
```diff
@@ -30,7 +30,9 @@ export interface AutocompleterItem {
 
 export interface AutocompleterSpec {
   type?: 'autocompleter';
-  ch: string;
+  // TODO: TINY-8929: Remove 'trigger' fallback to 'ch'
+  ch?: string;
+  trigger?: string; // TODO: TINY-8929: upgrade to non-optional
```
lnewson, line 35: Why is this after when all the other comments are above? This also feels quite verbose, I think we could just merge it in with the first TODO.

modules/katamari/src/test/ts/atomic/api/str/ContainsTest.ts
```diff
@@ -52,4 +52,22 @@ describe('atomic.katamari.api.str.ContainsTest', () => {
...
+    check(false, 'abc', 'bc', 1, 2);
+
```
lnewson, line 71:
```suggestion
```

modules/tinymce/CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - New `text_patterns_lookup` option to provide additional text patterns dynamically #TINY-8778
 
+### Improved
+- The autocompleter now supports a multiple character trigger #TINY-8887
```
lnewson, line 13:
```suggestion
- The autocompleter now supports a multiple character trigger using the new `trigger` configuration #TINY-8887
```
Maybe? As we have no mention about about the new configuration option. We also need to add a `Deprecated` heading below `Fixed` and say that `ch` is now deprecated, use `trigger` instead.

- [ ] Add deprecated changelog entry

modules/tinymce/src/core/main/ts/autocomplete/AutocompleteContext.ts
```diff
@@ -37,10 +37,15 @@ const findStart = (dom: DOMUtils, initRange: Range, ch: string, minChars: number
     return Optional.none();
   }
 
-  const findTriggerChIndex = (element: Text, offset: number, text: string) =>
+  const buf: Array<{text: string; offset: number}> = [];
```
lnewson, line 40:
```suggestion
  const buf: Array<{ text: string; offset: number }> = [];
```

lnewson: Also can you please add a comment about what's happening here, as it's not very clear without looking at it in more detail.

lnewson: Actually having looked into this more... if you're doing mutation anyways then having to do the `Arr.foldl` below is rather inefficient and we might as well just update as we go. For example:
```ts
const buffer = { text: '', offset: 0 };

const findTriggerChIndex = (element: Text, offset: number, text: string) => {
  buffer.text = text + buffer.text;
  buffer.offset += offset;
  // Stop searching by just returning the current offset if whitespace was found (eg Optional.none())
  // and we'll handle the final checks below instead
  return findTrigger(buffer.text, buffer.offset, ch).getOr(offset);
};
```

- [ ] Don't loop over the buffer to create the text/offsets on every text node when it's not needed

modules/tinymce/src/core/main/ts/autocomplete/AutocompleteContext.ts
```diff
@@ -37,10 +37,15 @@ const findStart = (dom: DOMUtils, initRange: Range, ch: string, minChars: number
...
+
+  const findTriggerChIndex = (element: Text, offset: number, text: string) => {
     // Stop searching by just returning the current offset if whitespace was found (eg Optional.none())
     // and we'll handle the final checks below instead
```
lnewson, line 44: This comment should be below the buffer logic, as it applies to the `findChar(text, offset, ch).getOr(offset);` logic.

modules/tinymce/src/core/main/ts/autocomplete/AutocompleteContext.ts
```diff
@@ -14,7 +14,7 @@ export interface AutocompleteContext {
 
 const stripTriggerChar = (text: string, triggerCh: string) => text.substring(triggerCh.length);
 
-const findChar = (text: string, index: number, ch: string): Optional<number> => {
+const findTrigger = (text: string, index: number, ch: string): Optional<number> => {
```
lnewson, line 17: Please rename `ch` here as this is no longer a single character.

- [ ] Rename to match that this isn't a single character anymore

lnewson: This function also isn't very performant anymore since it has to do a `indexOf` call on every character which could be slower on long strings (granted it should stop at the first whitespace char). This code also runs on every keypress so we need to try and make sure it's fairly quick. I think we'd be better refactoring this a little so that it's something like this instead:
```ts
const findTrigger = (text: string, index: number, trigger: string): Optional<number> => {
  // Identify the `char` in, and start the text from that point forward. If there is ever any whitespace, fail
  let i: number;
  const firstChar = trigger.charAt(0)

  for (i = index - 1; i >= 0; i--) {
    const char = text.charAt(i);
    if (isWhitespace(char)) {
      return Optional.none();
    }

    if (firstChar === char && Strings.contains(text, trigger, i, index)) {
      break;
    }
  }

  return Optional.some(i);
};
```
This way we only do the `indexOf` check if the first character in the trigger is found first.

- [ ] Consider improving the performance of this logic a little, like in the above

lnewson (approved):

modules/tinymce/src/core/main/ts/autocomplete/AutocompleteContext.ts
lnewson: Actually looking more the `Autocompleters` logic also needs `ch` renamed to reflect it's no longer a single character and the `triggerChars` property in `AutocompleterDatabase` also needs renaming... maybe to just `triggers`?

- [ ] Rename anything related to `ch` in the autocompleter logic

tiny-james (approved):

modules/katamari/src/main/ts/ephox/katamari/api/Strings.ts
```diff
@@ -38,8 +39,13 @@ export const ensureTrailing = (str: string, suffix: string): string => {
   return endsWith(str, suffix) ? str : StrAppend.addToEnd(str, suffix);
 };
 
-export const contains = (str: string, substr: string): boolean => {
-  return str.indexOf(substr) !== -1;
+export const contains = (str: string, substr: string, start: number = 0, end?: number): boolean => {
```
lnewson, line 42: Sorry I missed this in my previous review, but this needs an `Improved` section added to the Katamari changelog. We also need to add `katamari@9.1.0` to `versions.txt` so we know it needs a minor bump.

modules/tinymce/CHANGELOG.md
```diff
@@ -24,6 +25,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The text pattern logic threw an error when there were fragmented text nodes in a paragraph #TINY-8779
 - The content of the `contenteditable="false"` element could be selected with the mouse on Firefox  #TINY-8828
 
+### Deprecated
+- The autocompleter configuration option `ch` has been deprecated and will be removed in the next major release. Use the `trigger` option instead. #TINY-8887
```
lnewson, line 29:
```suggestion
- The autocompleter `ch` configuration property has been deprecated and will be removed in the next major release. Use the `trigger` property instead. #TINY-8887
```

lnewson (approved):

lnewson (approved):

PR #7998, opened 2022-07-22
TINY-8874: Fix dragging CEF element toward the edges doesn't cause sc…

hamza0867:
…rolling

Related Ticket: TINY-8874

Description of Changes:
Detect when we are in a dragging state and the mouse in on the edge of the editor then cause scrolling accordingly 

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/tinymce/src/core/main/ts/DragDropOverrides.ts
```diff
@@ -121,6 +131,54 @@ const moveGhost = (
...
+      // (within MouseRange pixels of the bottom edge)
+      if (mouseY + MouseRange >= clientHeight) {
+        const scrollDown = (currentTop: number) => {
+          win.scroll({
+            top: currentTop + ScrollingValue,
+            behavior: 'smooth'
+          });
+        };
+        scrollDown(currentTop);
+        state.intervalId = Optional.some(setInterval(() => {
```
tiny-ben-tran, line 160: Wonder whether this can be optimized further by clearing the previous timer before setting a new one

hamza0867: I do clear previous timers before calling `moveGhost` [here](https://github.com/tinymce/tinymce/blob/1cb5ddce8eac335b535725a19355f81ad4a8a78f/modules/tinymce/src/core/main/ts/DragDropOverrides.ts#L249)

modules/tinymce/src/core/test/ts/browser/DragDropOverridesTest.ts
```diff
@@ -117,4 +117,59 @@ describe('browser.tinymce.core.DragDropOverridesTest', () => {
     ]);
     await pAssertNotification('Dropped file type is not supported');
   });
+
+});
+describe('browser.tinymce.core.DragDropOverridesTest.TINY-8874', () => {
```
HAFRMO, line 122: This seems to be two different sets of tests that doesn't share any code, so I'd say you should either merge the two sets together due to both being for the dragdropoverride or create a separate file for this set. Putting them like this, being entirely separate, just seems like it'd end up being confusing.

modules/tinymce/src/core/main/ts/DragDropOverrides.ts
```diff
@@ -21,6 +21,11 @@ import * as Predicate from './util/Predicate';
  * @class tinymce.DragDropOverrides
  */
 
+// Arbitrary values needed when scrolling CEF elements
+const ScrollingValue = 32;
+const IntervalValue = 100;
+const MouseRange = 4;
```
HAFRMO, line 27: You refer to these as arbitrary, yet i assume they actually do affect the behavior. Do we want to include a proper description of what they do, or even make them options? And if not, how well does the behavior match CET elements being dragged, and how do we refine it? Should these depend on the browser used, to better match behavior there?

modules/tinymce/src/core/main/ts/DragDropOverrides.ts
```diff
@@ -121,6 +131,54 @@ const moveGhost = (
...
   ghostElm.style.width = (width - overflowX) + 'px';
   ghostElm.style.height = (height - overflowY) + 'px';
+
+  // Code needed for dragging CEF elements (specifically fixing TINY-8874)
+  // The idea behind the algorithm is that the user will start dragging the
+  // CEF element to the edge of the editor and that would cause scrolling.
+  // The way that happens is that the user will trigger a mousedown event,
+  // then a mousemove event until they reach the edge of the editor. Then
+  // no event triggers. That's when I set an interval to keep scrolling the editor.
+  // Once a new event triggers I clear the existing interval and set it back to none.
```
HAFRMO, line 141: How well does this handle 'strange' behavior, like losing focus in the middle of things? It's not out of the realm for someone to suddenly tab away.

modules/tinymce/src/core/main/ts/DragDropOverrides.ts
```diff
@@ -121,6 +131,54 @@ const moveGhost = (
...
+  const clientHeight = contentAreaContainer.clientHeight;
+
+  // current scroll level
+  const currentTop = win.scrollY;
+
+  state.on((state) => {
+    if (state.dragging) {
+      // This basically means that the mouse is close to the bottom edge
+      // (within MouseRange pixels of the bottom edge)
+      if (mouseY + MouseRange >= clientHeight) {
```
HAFRMO, line 152: This implementation doesn't support scrolling to the side, is this intentional? Is that solved elsewhere?

hamza0867: Because I believe the editor is not scrollable to the sides so it supporting scrolling to the side wouldn't have benefits and would be hard (maybe impossible ?) to test.

modules/tinymce/src/core/main/ts/DragDropOverrides.ts
hamza0867: If they suddenly tab away then a new 'mouseup' event would be submitted and that is handled elsewhere

modules/tinymce/src/core/main/ts/DragDropOverrides.ts
HAFRMO: It's very easy to scroll sideways in the editor, I did check to make sure. Unless this is referring to something else? You can make a `div` with very wide width, and that needs to scroll?

modules/tinymce/src/core/main/ts/DragDropOverrides.ts
HAFRMO: If that's fired, then that's good.

modules/tinymce/src/core/main/ts/DragDropOverrides.ts
hamza0867: You are correct, I forgot about that. The way I tested was using a very large sentence and it was broken down multiple line instead of being one large line that can be scrolled sideways. I will add support for scrolling to the side 👍

modules/tinymce/src/core/test/ts/browser/DragDropOverridesTest.ts
hamza0867: Even though the tests don't share code, but the tests I made, test stuff that's in the `DragDropOverrides.ts`. So it made sense to me to put the tests in the `DragDropOverridesTest.ts`. Also the reason why I made a new describe block is because I wanted to control the `height` of the editor and it seems I can only set this property when initializing the editor but not after. So I had to create a new describe block to make a new editor hook with the `height` property already set up. What do you think ?

lnewson: Just chiming in quickly here and I haven't reviewed this... anyways Hannes is correct, this isn't what we normally do. Normally we'd use `context` functions inside the `describe` to break it up if possible, otherwise we use a separate test file.

Also as a side note, your "height" won't be working as `innerHeight` is not a valid TinyMCE option.

hamza0867: Yeah I figured about the `innerHeight` and I changed it to `height` I just still haven't pushed yet. So if I understand correctly I should use `context` instead of the `describe` I made, and put that inside the existing `describe` block ?

lnewson: Yes, but only if you can move the existing content into a describe since you need a different editor instance. Is there a reason the existing tests can't use a fixed height as well though? I can't see anything about them where that would be an issue.

hamza0867: I used height and with and put everything in the same describe block and it worked fine. Thanks 👍

modules/tinymce/src/core/main/ts/DragDropOverrides.ts
hamza0867: Done!

modules/tinymce/src/core/test/ts/browser/DragDropOverridesTest.ts
hamza0867: Done!

modules/tinymce/src/core/main/ts/DragDropOverrides.ts
hamza0867: Arbitrary doesn't mean random, it just means they could have been different but I believe these values to be sane defaults.
Yes we can include a proper description of what they affect. As for making them options configurable by the client, I don't think that would be necessary but that's more up to the product team to decide.

> how well does the behavior match CET elements being dragged, and how do we refine it? 

Well, Safari doesn't scroll when dragging CET elements so the behavior doesn't match there. For other editors it seems to me to match well but that can be debated. We refine the behavior through the use/change of these variables.

> Should these depend on the browser used, to better match behavior there?

Yes we can make them depend in the browser, but the behavior doesn't change much between chrome and firefox. So having a sane default can suffice, I think.

modules/tinymce/src/core/main/ts/DragDropOverrides.ts
```diff
@@ -33,8 +38,16 @@ interface State {
...
   ghost: HTMLElement;
+  intervalId: Optional<number>;
 }
 
+const clearStateInterval = (state: State) => {
+  if (state.intervalId.isSome()) {
+    clearInterval(state.intervalId.getOrUndefined());
+    state.intervalId = Optional.none();
+  }
+};
```
TheSpyder, line 49: We have a concept called `Singleton` that seems like it might make this easier - it ensures a "revoke" function is run whenever a new value is set, which could be used to clear the interval. No need to manually manage set and clear.
https://github.com/tinymce/tinymce/blob/2d8b95517a43d1f1967a2f3898371865a2fa18e4/modules/katamari/src/main/ts/ephox/katamari/api/Singleton.ts#L22

I'm not sure it _quite_ applies here, because currently it is designed for objects with `destroy` or `unbind` functions, but I think the concept might be useful. It would certainly cut down on some of the boilerplate code you've had to add here.

Perhaps we need a new singleton that supports intervals like this; it might be a good candidate to go in the [Throttler](https://github.com/tinymce/tinymce/blob/2d8b95517a43d1f1967a2f3898371865a2fa18e4/modules/katamari/src/main/ts/ephox/katamari/api/Throttler.ts) module. Throttler uses timeout, not interval, but it seems like the most appropriate place.

modules/tinymce/src/core/main/ts/DragDropOverrides.ts
```diff
@@ -121,6 +139,83 @@ const moveGhost = (
...
+          win.scroll({
+            top: currentTop - ScrollingValue,
+            behavior: 'smooth'
+          });
+        };
+        scrollUp(currentTop);
+        state.intervalId = Optional.some(setInterval(() => {
+          const currentTop = win.scrollY;
+          scrollUp(currentTop);
+        }, IntervalValue));
```
TheSpyder, line 187: For all of these, do you need to call the function before setting the interval? Does it really matter if nothing happens for the first 100ms?

If you remove that, the function can be inlined into the `setInterval` code.

TheSpyder (approved):

modules/tinymce/src/core/main/ts/DragDropOverrides.ts
```diff
@@ -21,6 +21,11 @@ import * as Predicate from './util/Predicate';
  * @class tinymce.DragDropOverrides
  */
 
+// Arbitrary values needed when scrolling CEF elements
+const ScrollingValue = 32;
```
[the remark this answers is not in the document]
hamza0867: Done! Can you please confirm if the new names are more obvious ?

modules/tinymce/src/core/main/ts/DragDropOverrides.ts
hamza0867: Done!

modules/tinymce/src/core/main/ts/DragDropOverrides.ts
hamza0867: Done!

modules/katamari/src/main/ts/ephox/katamari/api/Singleton.ts
```diff
@@ -50,6 +50,8 @@ export const destroyable = <T extends { destroy: () => void }> (): Revocable<T>
 
 export const unbindable = <T extends { unbind: () => void }> (): Revocable<T> => singleton<T>((s) => s.unbind());
 
+export const repeatable = (): Revocable<number> => singleton((intervalId) => clearInterval(intervalId));
```
TheSpyder, line 53: I wonder if the fact that it's an interval can be moved inside the `repeatable`.

I'm thinking create it as `Singleton.repeatable(ScrollIntervalValue)` and then `interval.set(functionToRepeat)`. That would be a pretty cool data type, and avoid all the manual interval management.

modules/tinymce/src/core/main/ts/DragDropOverrides.ts
```diff
@@ -208,6 +288,9 @@ const getRawTarget = (selection: EditorSelection): Node | null => {
 
 const drop = (state: Singleton.Value<State>, editor: Editor) => (e: EditorEvent<MouseEvent>) => {
   state.on((state) => {
+    if (state.intervalId.isSet()) {
+      state.intervalId.clear();
+    }
```
TheSpyder, line 293: If you look at the way `clear` works:
https://github.com/tinymce/tinymce/blob/c7e1029ff3fabd36d1a31a6d45c19509aa811fed/modules/katamari/src/main/ts/ephox/katamari/api/Singleton.ts#L25-L30

You don't need to check if it's set before calling it. By design it can be called no matter what the state is.
```suggestion
    state.intervalId.clear();
```

modules/tinymce/src/core/main/ts/DragDropOverrides.ts
```diff
@@ -235,6 +318,9 @@ const drop = (state: Singleton.Value<State>, editor: Editor) => (e: EditorEvent<
 
 const stop = (state: Singleton.Value<State>, editor: Editor) => () => {
   state.on((state) => {
+    if (state.intervalId.isSet()) {
+      state.intervalId.clear();
+    }
```
TheSpyder, line 323:
```suggestion
    state.intervalId.clear();
```

TheSpyder (commented):
Very good! Just a few small suggestions.

modules/tinymce/src/core/main/ts/DragDropOverrides.ts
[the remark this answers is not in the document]
TheSpyder: Yes

modules/katamari/src/main/ts/ephox/katamari/api/Singleton.ts
```diff
@@ -9,6 +9,13 @@ interface Singleton<T> {
   readonly set: (value: T) => void;
 }
 
+export interface Repeatable {
```
lnewson, line 12: @hamza0867 this PR is missing quite a number of things for this change... you've not added any Katamri tests nor a changelog entry. Please log a follow up PR to correct that.

lnewson: Also if making large changes after a review such as this it's normally a good idea to re-request reviews.

modules/tinymce/src/core/main/ts/DragDropOverrides.ts
```diff
@@ -21,6 +21,11 @@ import * as Predicate from './util/Predicate';
  * @class tinymce.DragDropOverrides
  */
 
+// Arbitrary values needed when scrolling CEF elements
+const ScrollPixelsPerInterval = 32;
```
lnewson, line 25: These variables are also PascalCase when they should be camelCase. So please fix that when you add the other missing parts.

modules/tinymce/src/core/main/ts/DragDropOverrides.ts
```diff
@@ -121,6 +132,75 @@ const moveGhost = (
...
+  const clientHeight = contentAreaContainer.clientHeight;
+  const clientWidth = contentAreaContainer.clientWidth;
+
+  state.on((state) => {
+    if (state.dragging) {
+      // This basically means that the mouse is close to the bottom edge
+      // (within MouseRange pixels of the bottom edge)
+      if (mouseY + MouseRangeToTriggerScroll >= clientHeight) {
+        const scrollDown = (currentTop: number) => {
+          win.scroll({
```
lnewson, line 153: Have you tested this works with say the autoresize plugin enabled? As the window is always the same height so it's rather unlikely using window.scroll() will work in that case.

modules/tinymce/CHANGELOG.md
```diff
@@ -24,12 +24,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 - Fixed various issues that occurred when formatting `contenteditable` elements. #TINY-8905
 - The text pattern logic threw an error when there were fragmented text nodes in a paragraph #TINY-8779
+- Dragging a `contentEditable=false` element towards the edges would not cause scrolling #TINY-8874
 - The content of the `contenteditable="false"` element could be selected with the mouse on Firefox  #TINY-8828
 - Parsing large documents no longer throws a `Maximum call stack size exceeded` exception #TINY-6945
 
 ### Deprecated
 - The autocompleter `ch` configuration property has been deprecated and will be removed in the next major release. Use the `trigger` property instead. #TINY-8887
 
+
```
lnewson, line 34: This blank line should not have been added here either, so another thing to fix.

modules/tinymce/src/core/test/ts/browser/DragDropOverridesTest.ts
```diff
@@ -117,4 +119,87 @@ describe('browser.tinymce.core.DragDropOverridesTest', () => {
...
+      <p>CEF can get dragged after this one</p>
+    `);
+    const target = UiFinder.findIn(TinyDom.body(editor), 'p:contains("Draggable CEF")').getOrDie();
+    const dest = UiFinder.findIn(TinyDom.body(editor), '#separator').getOrDie();
+    const initialScrollY = editor.getBody().ownerDocument.defaultView.scrollY;
+    Mouse.mouseDown(target);
+    Mouse.mouseMoveTo(dest, 0, 98); // Move the mouse close to the bottom edge of the editor to trigger scrolling
+    await Waiter.pWait(1000); // Wait a small amount of time to ensure the scrolling happens
+    Mouse.mouseUp(target);
+    assert.isTrue(editor.getBody().ownerDocument.defaultView.scrollY > initialScrollY); // Make sure scrolling happened
```
lnewson, line 139: This should have been using `isAbove` like you did in the test below. Reason being you can't see the values when it fails so the error message information is very unhelpful if you do this type of thing.

modules/tinymce/src/core/test/ts/browser/DragDropOverridesTest.ts
```diff
@@ -117,4 +119,87 @@ describe('browser.tinymce.core.DragDropOverridesTest', () => {
...
+      top: 450
+    });
+    const target = UiFinder.findIn(TinyDom.body(editor), 'p:contains("Draggable CEF")').getOrDie();
+    const dest = UiFinder.findIn(TinyDom.body(editor), '#separator').getOrDie();
+    const initialScrollY = editor.getBody().ownerDocument.defaultView.scrollY;
+    Mouse.mouseDown(target);
+    Mouse.mouseMoveTo(dest, 0, 2); // Move the mouse close to the top edge of the editor to trigger scrolling
+    await Waiter.pWait(1000); // Wait a small amount of time to ensure the scrolling happens
+    Mouse.mouseUp(target);
+    assert.isTrue(editor.getBody().ownerDocument.defaultView.scrollY < initialScrollY); // Make sure scrolling happened
```
lnewson, line 161: This should have been using `isBelow` as well, like you did below.

PR #7999, opened 2022-07-25
TINY-8860: Correctly merge pre content from paste

MitchCTiny:
Related Ticket: TINY-8860

Description of Changes:
* Fixed pasting codesample onto any `pre` content removing class attribute
* Fixed pasting any `pre` content into styled `pre` content not merging correctly

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -11,7 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - An exception was thrown when deleting all content if the start or end of the document had a `contenteditable="false"` element #TINY-8877
 - When a sidebar was opened using the `sidebar_show` option, its associated toggle button was not highlighted #TINY-8873
 - The `autolink` plugin when converting a URL to a link did not fire an `ExecCommand` event, nor did it create an undo level #TINY-8896
-- Worked around a Firefox bug whereby cookies weren't available inside the editor content #TINY-88916
+- Worked around a Firefox bug whereby cookies weren't available inside the editor content #TINY-8916
+- Pasting styled `pre` content into a `pre` block now correctly merges content #TINY-8860
 
 ## 6.1.0 - 2022-06-29
 
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -12,7 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - When a sidebar was opened using the `sidebar_show` option, its associated toggle button was not highlighted #TINY-8873
 - The `autolink` plugin when converting a URL to a link did not fire an `ExecCommand` event, nor did it create an undo level #TINY-8896
 - Worked around a Firefox bug whereby cookies weren't available inside the editor content #TINY-8916
-- Pasting `pre` content into a `pre` block that had inline styles or was noneditable now correctly merges content #TINY-8860
+- `<pre>` content pasted into a `<pre>` block that had inline styles or was `noneditable` now merges correctly with the surrounding content. #TINY-8860
 
 ## 6.1.0 - 2022-06-29
 
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -12,7 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - When a sidebar was opened using the `sidebar_show` option, its associated toggle button was not highlighted #TINY-8873
 - The `autolink` plugin when converting a URL to a link did not fire an `ExecCommand` event, nor did it create an undo level #TINY-8896
 - Worked around a Firefox bug whereby cookies weren't available inside the editor content #TINY-8916
-- Pasting styled `pre` content into a `pre` block now correctly merges content #TINY-8860
+- Pasting `pre` content into a `pre` block that had inline styles or was noneditable now correctly merges content #TINY-8860
 
 ## 6.1.0 - 2022-06-29
 
--- a/modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
+++ b/modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
@@ -22,19 +22,20 @@ import * as SelectionUtils from '../selection/SelectionUtils';
 import { InsertContentDetails } from './ContentTypes';
 import * as InsertList from './InsertList';
 
-const wrappedElements = [ 'pre', 'h1', 'h2', 'h3', 'h4', 'h5', 'h6' ];
+const wrappedElements = [ 'pre' ];
 
-const shouldPasteContentOnly = (dom: DOMUtils, fragment: AstNode, parentNode: Element) => {
+const shouldPasteContentOnly = (dom: DOMUtils, fragment: AstNode, parentNode: Element, editor: Editor): boolean => {
   const firstNode = fragment.firstChild as AstNode;
   const lastNode = fragment.lastChild as AstNode;
   const last = lastNode.attr('data-mce-type') === 'bookmark' ? lastNode.prev : lastNode;
 
   const isCopyingOnlyOneTag = firstNode === last;
   const isWrappedElement = Arr.contains(wrappedElements, firstNode.name);
-  const isPastingInTheSameTag = dom.getParent(parentNode, (node) => firstNode.name === node.nodeName.toLowerCase());
-  const isContentEditableTrue = Optional.from(CefUtils.getContentEditableRoot(dom.getRoot(), parentNode)).map(NodeType.isContentEditableTrue).getOr(true);
+  const isPastingInTheSameTag = firstNode.name === parentNode.nodeName.toLowerCase() ? true : dom.getParent(parentNode, (node) => firstNode.name === node.nodeName.toLowerCase()) !== null;
+  const isSameClass = firstNode.attr('class') !== undefined ? firstNode.attr('class') === parentNode.className : parentNode.className === '';
+  const isPastingInContentEditableTrue = Optional.from(CefUtils.getContentEditableRoot(editor.getBody(), parentNode)).map(NodeType.isContentEditableTrue).getOr(true);
 
-  return isCopyingOnlyOneTag && isWrappedElement && isPastingInTheSameTag && isContentEditableTrue;
+  return isCopyingOnlyOneTag && isWrappedElement && isPastingInTheSameTag && isSameClass && isPastingInContentEditableTrue;
 };
 
 const isTableCell = NodeType.isTableCell;
@@ -262,7 +263,7 @@ export const insertHtmlAtCaret = (editor: Editor, value: string, details: Insert
     return value;
   }
 
-  if (details.paste === true && shouldPasteContentOnly(dom, fragment, parentNode)) {
+  if (details.paste === true && shouldPasteContentOnly(dom, fragment, parentNode, editor)) {
     fragment.firstChild?.unwrap();
   }
 
--- a/modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
+++ b/modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
@@ -22,7 +22,7 @@ import * as SelectionUtils from '../selection/SelectionUtils';
 import { InsertContentDetails } from './ContentTypes';
 import * as InsertList from './InsertList';
 
-const wrappedElements = [ 'pre' ];
+const wrappedElements = [ 'pre', 'h1', 'h2', 'h3', 'h4', 'h5', 'h6' ];
 
 const shouldPasteContentOnly = (dom: DOMUtils, fragment: AstNode, parentNode: Element) => {
   const firstNode = fragment.firstChild as AstNode;
@@ -31,8 +31,8 @@ const shouldPasteContentOnly = (dom: DOMUtils, fragment: AstNode, parentNode: El
 
   const isCopyingOnlyOneTag = firstNode === last;
   const isWrappedElement = Arr.contains(wrappedElements, firstNode.name);
-  const isPastingInTheSameTag = firstNode.name === parentNode.tagName.toLowerCase();
-  const isContentEditableTrue = dom.getContentEditable(parentNode) !== 'false';
+  const isPastingInTheSameTag = dom.getParent(parentNode, (node) => firstNode.name === node.nodeName.toLowerCase());
+  const isContentEditableTrue = Optional.from(CefUtils.getContentEditableRoot(dom.getRoot(), parentNode)).map(NodeType.isContentEditableTrue).getOr(true);
 
   return isCopyingOnlyOneTag && isWrappedElement && isPastingInTheSameTag && isContentEditableTrue;
 };
--- a/modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
+++ b/modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
@@ -22,20 +22,25 @@ import * as SelectionUtils from '../selection/SelectionUtils';
 import { InsertContentDetails } from './ContentTypes';
 import * as InsertList from './InsertList';
 
-const wrappedElements = [ 'pre' ];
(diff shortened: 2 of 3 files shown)


modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
```diff
@@ -24,15 +24,18 @@ import * as InsertList from './InsertList';
...
-
-  return isCopingOnlyOneTag && isAFlattenableTag && isPastingInTheSameTag;
+const shouldPasteContentOnly = (dom: DOMUtils, fragment: AstNode, parentNode: Element, editor: Editor): boolean => {
+  const firstNode = fragment.firstChild as AstNode;
+  const lastNode = fragment.lastChild as AstNode;
+  const last = lastNode.attr('data-mce-type') === 'bookmark' ? lastNode.prev : lastNode;
+
+  const isCopyingOnlyOneTag = firstNode === last;
+  const isWrappedElement = Arr.contains(wrappedElements, firstNode.name);
+  const isPastingInTheSameTag = firstNode.name === parentNode.nodeName.toLowerCase() ? true : dom.getParent(parentNode, (node) => firstNode.name === node.nodeName.toLowerCase()) !== null;
```
lnewson, line 34: This can be simplified like so, because `getParent` runs on the node passed into it (it's weird but it's how it works):
```suggestion
  const isPastingInTheSameTag = dom.getParent(parentNode, (node) => firstNode.name === node.nodeName.toLowerCase()) !== null;
```

In saying that though, I have to wonder if this is really the correct logic we want. While this is okay for this case, since we only have one `wrappedElements` item, it wouldn't work well if we had multiple. For example, say the wrapped elements list had `pre` and `blockquote` in it. If we then pasted in a `blockquote` this would unwrap that and break things...

As such, I think a better solution would be to find the nearest block element and see if it compares as we really should only be unwrapping block elements. For example:
```ts
  const isPastingInTheSameBlockTag = dom.getParent(parentNode, dom.isBlock)?.nodeName == firstNode.name;
```

modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
```diff
@@ -24,15 +24,18 @@ import * as InsertList from './InsertList';
...
-const shouldPasteContentOnly = (fragment: AstNode, parentNode: Element) => {
-  const firstNode = fragment.firstChild;
-
-  const isAFlattenableTag = Arr.contains(wrappedElements, firstNode.name);
-  const isPastingInTheSameTag = firstNode.name === parentNode.tagName.toLowerCase();
-  const lastNode = fragment.lastChild.attr('data-mce-type') === 'bookmark' ? fragment.lastChild.prev : fragment.lastChild;
-  const isCopingOnlyOneTag = firstNode === lastNode;
-
-  return isCopingOnlyOneTag && isAFlattenableTag && isPastingInTheSameTag;
+const shouldPasteContentOnly = (dom: DOMUtils, fragment: AstNode, parentNode: Element, editor: Editor): boolean => {
```
lnewson, line 27: Instead of passing the `editor` here, it'd be cleaner and easier to test in isolation if we pass the root element through to this function instead since that's all we're using `editor` for. For example:
```ts
const shouldPasteContentOnly = (dom: DOMUtils, fragment: AstNode, parentNode: Element, root: Node): boolean => {
```

modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
```diff
@@ -24,15 +24,18 @@ import * as InsertList from './InsertList';
...
+  const firstNode = fragment.firstChild as AstNode;
```
lnewson, line 28: We also need to consider if the fragment itself is marked as a CEF element here I believe. So we probably need to add a check such as `firstNode.attr('contenteditable') !== 'false' so that it won't unwrap a CEF element, though I'd double check that's not handled elsewhere already. A good way to test that would be to add a test for it.

- [ ] Add a test for pasting in a CEF pre block

modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
```diff
@@ -24,15 +24,18 @@ import * as InsertList from './InsertList';
...
+  const lastNode = fragment.lastChild as AstNode;
+  const last = lastNode.attr('data-mce-type') === 'bookmark' ? lastNode.prev : lastNode;
+
+  const isCopyingOnlyOneTag = firstNode === last;
+  const isWrappedElement = Arr.contains(wrappedElements, firstNode.name);
+  const isPastingInTheSameTag = firstNode.name === parentNode.nodeName.toLowerCase() ? true : dom.getParent(parentNode, (node) => firstNode.name === node.nodeName.toLowerCase()) !== null;
+  const isSameClass = firstNode.attr('class') !== undefined ? firstNode.attr('class') === parentNode.className : parentNode.className === '';
```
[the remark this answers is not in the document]
lnewson: Oh actually... was this added to exclude merging a codesample block? If so then yeah we might need to leave this as is for now, but either way we should have a test for it.

Also the change that I suggested in Lorenzo's [PR](https://github.com/tinymce/tinymce/pull/7996) would likely fix this without needing this check since it would then be marked as a CEF element while processing here. So maybe we need to make sure they are both merged for 6.1.1 🤔

modules/tinymce/src/core/test/ts/browser/content/InsertContentTest.ts
```diff
@@ -352,6 +352,42 @@ describe('browser.tinymce.core.content.InsertContentTest', () => {
     TinyAssertions.assertContent(editor, '<pre> <strong> a </strong> </pre>');
   });
 
+  it('TINY-8860: insertAtCaret - insert pre block on empty p', () => {
```
lnewson, line 355: I don't see any test here for inserting into/over a CEF block. Can you please add one?

- [ ] Add a test for pasting over a CEF block

modules/tinymce/CHANGELOG.md
```diff
@@ -11,7 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - An exception was thrown when deleting all content if the start or end of the document had a `contenteditable="false"` element #TINY-8877
 - When a sidebar was opened using the `sidebar_show` option, its associated toggle button was not highlighted #TINY-8873
 - The `autolink` plugin when converting a URL to a link did not fire an `ExecCommand` event, nor did it create an undo level #TINY-8896
-- Worked around a Firefox bug whereby cookies weren't available inside the editor content #TINY-88916
+- Worked around a Firefox bug whereby cookies weren't available inside the editor content #TINY-8916
+- Pasting styled `pre` content into a `pre` block now correctly merges content #TINY-8860
```
```suggestion
- Pasting `pre` content into a `pre` block that had inline styles or was noneditable now correctly merges content #TINY-8860
```

lnewson (changes requested):
Good work on this Mitch, it's mostly just little things left to clean up by the looks 👍

modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
```diff
@@ -24,15 +24,18 @@ import * as InsertList from './InsertList';
...
-  const lastNode = fragment.lastChild.attr('data-mce-type') === 'bookmark' ? fragment.lastChild.prev : fragment.lastChild;
-  const isCopingOnlyOneTag = firstNode === lastNode;
-
-  return isCopingOnlyOneTag && isAFlattenableTag && isPastingInTheSameTag;
+const shouldPasteContentOnly = (dom: DOMUtils, fragment: AstNode, parentNode: Element, editor: Editor): boolean => {
+  const firstNode = fragment.firstChild as AstNode;
+  const lastNode = fragment.lastChild as AstNode;
+  const last = lastNode.attr('data-mce-type') === 'bookmark' ? lastNode.prev : lastNode;
+
+  const isCopyingOnlyOneTag = firstNode === last;
```
lnewson, line 32: We're also getting a lot of checks in this function now that eagerly evaluate. Most are quick but some are slower... I'd suggest breaking this up to something like:
```
const isCopyingOnlyOneTag = firstNode === last;
const isWrappedElement = Arr.contains(wrappedElements, firstNode.name);
if (isCopyingOnlyOneTag && isWrappedElement) {
  ...other checks
} else {
  return false;
}
```

modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
MitchCTiny: Looks good, pretty much what I was trying to do with finding the parent block in the case `parentNode` is a `<strong>` or similar. Is the issue you mention with unwrapping when pasting `blockquote` content into another `blockquote` block? Or pasting into a different block also in `wrappedElements`?

modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
MitchCTiny: Did consider that, will do.

modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
lnewson: I'll give some demos shortly to explain it better as to be honest I don't think I've thought it through fully either, it was just more a potential code smell... what I've mentioned may not end up being an actual issue.

lnewson: Okay, so here's a slightly different case for if `li` was a wrapped element. So say we were pasting into this content:
```html
<ul>
<li>
  <p>
    insert here
  </p>
</li>
</ul>
```
and this was what was on the clipboard:
```html
<li>pasted content</li>
```
In that case, we wouldn't want this unwrapped and instead it should split the paragraph and start a new list. In saying that though, you shouldn't ever have just an `li` on the clipboard so it's still very much theoretical... either way, I'm happy to leave it with the version you have and we can adjust it later should we ever need to.

modules/tinymce/src/core/test/ts/browser/content/InsertContentTest.ts
```diff
@@ -346,12 +313,59 @@ describe('browser.tinymce.core.content.InsertContentTest', () => {
...
+    editor.setContent('<p></p>');
+    TinySelections.setSelection(editor, [ 0, 0 ], 0, [ 0, 0 ], 0);
+    InsertContent.insertAtCaret(editor, { content: '<pre>abc</pre>', paste: true });
+    TinyAssertions.assertContent(editor, '<pre>abc</pre>');
+  });
+
+  it('TINY-8860: insertAtCaret - insert styled pre block', () => {
+    const editor = hook.editor();
+    editor.setContent('<pre>abc</pre>');
+    TinySelections.setSelection(editor, [ 0, 0 ], 1, [ 0, 0 ], 1);
```
lnewson, line 332:
```suggestion
    TinySelections.setCursor(editor, [ 0, 0 ], 1);
```

modules/tinymce/src/core/test/ts/browser/content/InsertContentTest.ts
```diff
@@ -346,12 +313,59 @@ describe('browser.tinymce.core.content.InsertContentTest', () => {
...
-    editor.focus();
     TinySelections.setSelection(editor, [ 0, 0 ], 0, [ 0, 0 ], 0);
     InsertContent.insertAtCaret(editor, ' <strong> a </strong> ');
     TinyAssertions.assertContent(editor, '<pre> <strong> a </strong> </pre>');
   });
 
+  it('TINY-8860: insertAtCaret - insert pre block on empty p', () => {
+    const editor = hook.editor();
+    editor.setContent('<p></p>');
+    TinySelections.setSelection(editor, [ 0, 0 ], 0, [ 0, 0 ], 0);
```
lnewson, line 324:
```suggestion
    TinySelections.setCursor(editor, [ 0, 0 ], 0);
```

modules/tinymce/src/core/test/ts/browser/content/InsertContentTest.ts
```diff
@@ -346,12 +313,59 @@ describe('browser.tinymce.core.content.InsertContentTest', () => {
...
+    editor.setContent('<pre>abc</pre>');
+    TinySelections.setSelection(editor, [ 0, 0 ], 1, [ 0, 0 ], 1);
+    InsertContent.insertAtCaret(editor, { content: '<pre><strong>123</strong></pre>', paste: true });
+    TinyAssertions.assertContent(editor, '<pre>a<strong>123</strong>bc</pre>');
+  });
+
+  it('TINY-8860: insertAtCaret - insert plain pre in styled pre block', () => {
+    const editor = hook.editor();
+    editor.setContent('<pre><strong>123</strong>abc</pre>');
+    TinySelections.setSelection(editor, [ 0, 0, 0 ], 1, [ 0, 0, 0 ], 1);
```
lnewson, line 340:
```suggestion
    TinySelections.setCursor(editor, [ 0, 0, 0 ], 1);
```

modules/tinymce/src/core/test/ts/browser/content/InsertContentTest.ts
```diff
@@ -346,12 +313,59 @@ describe('browser.tinymce.core.content.InsertContentTest', () => {
...
+    InsertContent.insertAtCaret(editor, { content: '<pre>abc</pre>', paste: true });
+    TinyAssertions.assertContent(editor, '<pre><strong>1abc23</strong>abc</pre>');
+  });
+
+  it('TINY-8860: insertAtCaret - insert styled pre in differently styled pre block', () => {
+    const editor = hook.editor();
+    editor.setContent('<pre><strong>1abc23</strong>abc</pre>');
+    TinySelections.setSelection(editor, [ 0, 0, 0 ], 5, [ 0, 0, 0 ], 5);
```
lnewson, line 348:
```suggestion
    TinySelections.setCursor(editor, [ 0, 0, 0 ], 5);
```

modules/tinymce/src/core/test/ts/browser/content/InsertContentTest.ts
```diff
@@ -346,12 +313,59 @@ describe('browser.tinymce.core.content.InsertContentTest', () => {
...
+
+  it('TINY-8860: insertAtCaret - insert cef pre in pre block', () => {
+    const editor = hook.editor();
+    editor.setContent('<pre>abc</pre>');
+    TinySelections.setCursor(editor, [ 0, 0 ], 1);
+    InsertContent.insertAtCaret(editor, { content: '<pre contenteditable="false">123</pre>', paste: true });
+    TinyAssertions.assertContent(editor, '<pre>a</pre><pre contenteditable="false">123</pre><pre>bc</pre>');
+  });
+
+  it('TINY-8860: insertAtCaret - insert pre in cef pre block', () => {
```
lnewson, line 361:
```suggestion
  it('TINY-8860: insertAtCaret - insert pre over cef pre block', () => {
```

modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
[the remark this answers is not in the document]
MitchCTiny: isSameClass to be removed once change from TINY-8861 is implemented, attaching `contenteditable="false"` attribute to codesample elements. This will remove the requirement to match class for codesample elements, and instead prevent unwrap since `isContentEditable` will be false.

ltrouton (approved):

lnewson (approved):
Looks better, thanks Mitch 👍

modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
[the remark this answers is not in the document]
lnewson: Just adding an action item here so we can't forget this.

- [ ] Either revert this once the other PR is merged or add test cases for this

modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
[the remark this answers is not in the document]
MitchCTiny: That does sound better, I'll add it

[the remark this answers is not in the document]
MitchCTiny: Added as `isPastingSingleElement`

PR #8001, opened 2022-07-26
TINY-8921: Fixed more strict type errors in the core demos and tests

lnewson:
Related Ticket: TINY-8921

Description of Changes:

Here we go again with more strict type PRs... this one aims at getting all the tests and demos in core compiling under strict mode. Most of these changes are just casting the `firstChild` type in the tests since we can be sure it's of a specific type and won't be nullable. I also found while fixing these some core types needed further improvement so this also contains those changes. Note that I've not included the `tests` directory in the `tsconfig.strict.json` file because they include silver, the dom model and some plugins which won't compile yet.

P.S. Hopefully this is still small enough to review and I've broken it up into 3 commits in an attempt to make it easier to see the changes specifically done for core, tests and demos.

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/api/AddOnManager.ts
+++ b/modules/tinymce/src/core/main/ts/api/AddOnManager.ts
@@ -101,7 +101,7 @@ interface AddOnManager<T> {
   urls: Record<string, string>;
   lookup: Record<string, { instance: AddOnConstructor<T> }>;
   get: (name: string) => AddOnConstructor<T> | undefined;
-  requireLangPack: (name: string, languages: string) => void;
+  requireLangPack: (name: string, languages?: string) => void;
   add: (id: string, addOn: AddOnConstructor<T>) => AddOnConstructor<T>;
   remove: (name: string) => void;
   createUrl: (baseUrl: UrlObject, dep: string | UrlObject) => UrlObject;
@@ -131,7 +131,7 @@ const AddOnManager = <T>(): AddOnManager<T> => {
     return undefined;
   };
 
-  const loadLanguagePack = (name: string, languages: string): void => {
+  const loadLanguagePack = (name: string, languages?: string): void => {
     const language = I18n.getCode();
     const wrappedLanguages = ',' + (languages || '') + ',';
 
@@ -142,7 +142,7 @@ const AddOnManager = <T>(): AddOnManager<T> => {
     ScriptLoader.ScriptLoader.add(urls[ name ] + '/langs/' + language + '.js');
   };
 
-  const requireLangPack = (name: string, languages: string) => {
+  const requireLangPack = (name: string, languages?: string) => {
     if (AddOnManager.languageLoad !== false) {
       if (isLoaded(name)) {
         loadLanguagePack(name, languages);
--- a/modules/tinymce/src/core/main/ts/api/UndoManager.ts
+++ b/modules/tinymce/src/core/main/ts/api/UndoManager.ts
@@ -1,6 +1,7 @@
 import { Arr, Cell, Singleton } from '@ephox/katamari';
 
 import { Bookmark } from '../bookmark/BookmarkTypes';
+import * as GetBookmark from '../bookmark/GetBookmark';
 import * as Rtc from '../Rtc';
 import * as Levels from '../undo/Levels';
 import { addKeyboardShortcuts, registerEvents } from '../undo/Setup';
@@ -45,7 +46,7 @@ const UndoManager = (editor: Editor): UndoManager => {
      * @method add
      * @param {Object} level Optional undo level object to add.
      * @param {DOMEvent} event Optional event responsible for the creation of the undo level.
-     * @return {Object} Undo level that got added or null it a level wasn't needed.
+     * @return {Object} Undo level that got added or null if a level wasn't needed.
      */
     add: (level?: Partial<UndoLevel>, event?: Event): UndoLevel | null => {
       return Rtc.addUndoLevel(editor, undoManager, index, locks, beforeBookmark, level, event);
@@ -58,8 +59,10 @@ const UndoManager = (editor: Editor): UndoManager => {
      */
     dispatchChange: (): void => {
       editor.setDirty(true);
+      const level = Levels.createFromEditor(editor) as UndoLevel;
+      level.bookmark = GetBookmark.getUndoBookmark(editor.selection);
       editor.dispatch('change', {
-        level: Levels.createFromEditor(editor),
+        level,
         lastLevel: Arr.get(undoManager.data, index.get()).getOrUndefined()
       });
     },
--- a/modules/tinymce/src/core/test/ts/browser/EditorManagerTest.ts
+++ b/modules/tinymce/src/core/test/ts/browser/EditorManagerTest.ts
@@ -38,9 +38,9 @@ describe('browser.tinymce.core.EditorManagerTest', () => {
         assert.equal(EditorManager.get(0), EditorManager.activeEditor);
         assert.isNull(EditorManager.get(1));
         assert.isNull(EditorManager.get('noid'));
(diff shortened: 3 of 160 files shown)


lnewson:
Note: The DeepScan errors all seem relate to having to cast the types in the tests and as such are not actual errors.

modules/tinymce/src/core/main/ts/api/UndoManager.ts
```diff
@@ -58,8 +59,10 @@ const UndoManager = (editor: Editor): UndoManager => {
      */
     dispatchChange: (): void => {
       editor.setDirty(true);
+      const level = Levels.createFromEditor(editor) as UndoLevel;
+      level.bookmark = GetBookmark.getUndoBookmark(editor.selection);
```
lnewson, line 63: Minor functional change here... the the `bookmark` needs to be populated for a change event so this adds that.

modules/tinymce/src/core/test/ts/browser/EditorManagerTest.ts
```diff
@@ -38,9 +38,9 @@ describe('browser.tinymce.core.EditorManagerTest', () => {
         assert.equal(EditorManager.get(0), EditorManager.activeEditor);
         assert.isNull(EditorManager.get(1));
         assert.isNull(EditorManager.get('noid'));
-        assert.isNull(EditorManager.get(undefined));
+        assert.isNull(EditorManager.get(undefined as any));
         assert.equal(EditorManager.get()[0], EditorManager.activeEditor);
-        assert.equal(EditorManager.get(EditorManager.activeEditor.id), EditorManager.activeEditor);
+        assert.equal(EditorManager.get((EditorManager.activeEditor as Editor).id), EditorManager.activeEditor);
```
tiny-ben-tran, line 43: Not sure why we need to assert the type of `EditorManager.activeEditor` as Editor when it already is.

lnewson: Take another look 😉 it's `Editor | null` since if there's nothing activated it can be null...

modules/tinymce/src/core/test/ts/browser/EditorManagerTest.ts
```diff
@@ -55,7 +55,7 @@ describe('browser.tinymce.core.EditorManagerTest', () => {
 
         // Re-init on same id
         EditorManager.init({
-          selector: '#' + EditorManager.activeEditor.id,
+          selector: '#' + (EditorManager.activeEditor as Editor).id,
```
tiny-ben-tran, line 58: Here was well. At least there is no error shown when doing a strict type check

lnewson: See previous comment...

modules/tinymce/src/core/test/ts/browser/dom/DomUtilsTest.ts
```diff
@@ -10,6 +11,9 @@ import * as HtmlUtils from '../../module/test/HtmlUtils';
 describe('browser.tinymce.core.dom.DOMUtilsTest', () => {
   const DOM = DOMUtils(document, { keep_values: true, schema: Schema() });
 
+  const getTestElement = () =>
```
tiny-ben-tran, line 14:
```suggestion
  const getTestElement = (name: string) =>
      DOM.get(name) as HTMLElement;
```
And replace `DOM.get()` below
I don't have a strong opinion about this suggestion. Thought it would  reduce some code

modules/tinymce/src/core/test/ts/browser/dom/DomUtilsTest.ts
```diff
@@ -59,61 +63,61 @@ describe('browser.tinymce.core.dom.DOMUtilsTest', () => {
...
+    assert.equal(getTestElement().className, 'abc 123', 'incorrect classname');
 
-    DOM.get('test').innerHTML = '<span id="test2"></span><span id="test3"></span><span id="test4"></span>';
+    getTestElement().innerHTML = '<span id="test2"></span><span id="test3"></span><span id="test4"></span>';
     DOM.addClass(DOM.select('span', 'test'), 'abc');
-    assert.equal(DOM.get('test2').className, 'abc', 'incorrect classname');
-    assert.equal(DOM.get('test3').className, 'abc', 'incorrect classname');
-    assert.equal(DOM.get('test4').className, 'abc', 'incorrect classname');
-    DOM.get('test').innerHTML = '';
+    assert.equal((DOM.get('test2') as HTMLElement).className, 'abc', 'incorrect classname');
```
tiny-ben-tran, line 75:
```suggestion
    assert.equal((getTestElement('test2')).className, 'abc', 'incorrect classname');
```

And so on

modules/tinymce/src/core/test/ts/browser/dom/DomUtilsTest.ts
lnewson: That wouldn't reduce code, it'd add a lot more as there's only a couple cases where `test2` or similar is used, but many where `test` is used, so you've have to pass the `test` id through instead. That's why I didn't do it that way... I guess we could have it default to `test` though (e.g. `name: string = 'test'`).

tiny-ben-tran (approved):

ltrouton (approved):

modules/tinymce/src/core/test/ts/browser/dom/DomUtilsTest.ts
```diff
@@ -50,7 +54,7 @@ describe('browser.tinymce.core.dom.DOMUtilsTest', () => {
     // equal(dom.getAttrib('test2', 'style'), Env.ue && !window.getSelection ?
     // 'border: #00ff00 1px solid;' : 'border: 1px solid #00ff00;'); // IE has a separate output
 
-    dom.get('test').innerHTML = '<span id="test2" style="background-image: url(http://www.site.com/test.gif);"></span>';
+    (dom.get('test') as HTMLElement).innerHTML = '<span id="test2" style="background-image: url(http://www.site.com/test.gif);"></span>';
```
TheSpyder, line 57: Missed one?
```suggestion
    getTestElement().innerHTML = '<span id="test2" style="background-image: url(http://www.site.com/test.gif);"></span>';
```

TheSpyder (approved):

modules/tinymce/src/core/test/ts/browser/dom/DomUtilsTest.ts
lnewson: Oops, not sure how I missed that 😅 thanks for catching that!

lnewson: Oh wait, no this one is different as it's using `dom` not `DOM` so this wouldn't be the same conversion. While it should still work, I didn't want to convert this one in case the looks do end up diverging somehow.

modules/tinymce/src/core/main/ts/api/AddOnManager.ts
```diff
@@ -131,9 +131,9 @@ const AddOnManager = <T>(): AddOnManager<T> => {
     return undefined;
   };
 
-  const loadLanguagePack = (name: string, languages: string): void => {
+  const loadLanguagePack = (name: string, languages: string = ''): void => {
     const language = I18n.getCode();
-    const wrappedLanguages = ',' + (languages || '') + ',';
+    const wrappedLanguages = ',' + languages + ',';
 
     if (!language || languages && wrappedLanguages.indexOf(',' + language + ',') === -1) {
```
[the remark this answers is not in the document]
lnewson: This ended up breaking the tests since if languages is falsy it needed to still attempt to load the language script. Looking into this more I believe the intention was to have `languages` be optional, so I've reverted this and updated it to reflect that.

PR #8005, opened 2022-07-26
TINY-8921: Fixed strict type errors in the silver theme demos and tests

lnewson:
Related Ticket: TINY-8921

Description of Changes:

This PR is a somewhat smaller than that last thankfully and aims at fixing all the strict type errors in the silver theme demo and tests. Note that I've not included the `tests` directory in the `tsconfig.strict.json` file because they include silver and some plugins which won't compile yet.

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/themes/silver/test/ts/browser/editor/SilverRenderTest.ts
+++ b/modules/tinymce/src/themes/silver/test/ts/browser/editor/SilverRenderTest.ts
@@ -25,7 +25,7 @@ describe('browser.tinymce.themes.silver.editor.SilverRenderTest', () => {
   const hook = TinyHooks.bddSetupLight<Editor>({
     plugins: 'test',
     base_url: '/project/tinymce/js/tinymce',
-    setup: (editor) => {
+    setup: (editor: Editor) => {
       editor.on('PostRender init', (e) => {
         initStates[e.type] = getRenderState(editor);
       });
--- a/modules/tinymce/.eslintrc.json
+++ b/modules/tinymce/.eslintrc.json
@@ -17,7 +17,7 @@
     },
     // Re-enable things that are passing for explicit module boundary types
     {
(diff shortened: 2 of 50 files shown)


modules/tinymce/src/themes/silver/test/ts/browser/editor/SilverRenderTest.ts
```diff
@@ -25,7 +25,7 @@ describe('browser.tinymce.themes.silver.editor.SilverRenderTest', () => {
   const hook = TinyHooks.bddSetupLight<Editor>({
     plugins: 'test',
     base_url: '/project/tinymce/js/tinymce',
-    setup: (editor) => {
+    setup: (editor: Editor) => {
```
MitchCTiny, line 28: Just curious, why is the type removed for some setups, and added for others?

MitchCTiny (approved):

modules/tinymce/src/themes/silver/test/ts/browser/editor/SilverRenderTest.ts
lnewson: It's removed in demos because we have the `TinyMCE` type that it can infer the editor from. For McAgar, it needs to work with TinyMCE 4, 5, 6, etc... so it unfortunately doesn't have the type information available for the options. Now we could change that, if we passed through another generic like we do for `Editor` (which is the return type), however that's a bigger change I'd prefer to do separately if we wanted to go down that path.

MitchCTiny: Interesting, thanks @lnewson

tiny-ben-tran (approved):

TheSpyder (approved):

PR #8011, opened 2022-07-27
TINY-8773: Prevent stripping of "name" and "id" attributes on iframe and img elements

MitchCTiny:
Related Ticket: TINY-8773

Description of Changes:
* Regression introduced in clobbering check #TINY-7756
* Filter through new document for each attribute, rather than same document
* SaxParser method now works same as in 6.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+- The `name` and `id` attributes of any `img` or `iframe` elements were incorrectly removed #TINY-8773
+
 ## 5.10.5 - 2022-05-25
 
 ### Fixed
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Fixed
-- The `name` and `id` attributes of any `img` or `iframe` elements were incorrectly removed #TINY-8773
+- The `name` and `id` attributes of some elements were incorrectly removed during serialization #TINY-8773
 
 ## 5.10.5 - 2022-05-25
 
--- a/modules/tinymce/src/core/main/ts/api/html/SaxParser.ts
+++ b/modules/tinymce/src/core/main/ts/api/html/SaxParser.ts
@@ -211,7 +211,7 @@ const checkBogusAttribute = (regExp: RegExp, attrString: string): string | null 
  */
 const SaxParser = (settings?: SaxParserSettings, schema = Schema()): SaxParser => {
   settings = settings || {};
-  const doc = settings.document ?? document;
+  const doc = (settings.document ?? document).implementation.createHTMLDocument('parser');
   const form = doc.createElement('form');
 
   if (settings.fix_self_closing !== false) {
--- a/modules/tinymce/src/core/main/ts/api/html/SaxParser.ts
+++ b/modules/tinymce/src/core/main/ts/api/html/SaxParser.ts
@@ -107,6 +107,8 @@ const filteredClobberElements = Tools.makeMap('button,fieldset,form,iframe,img,i
 
 const isValidPrefixAttrName = (name: string): boolean => name.indexOf('data-') === 0 || name.indexOf('aria-') === 0;
 
+const lazyTempDocument = Thunk.cached(() => document.implementation.createHTMLDocument('parser'));
+
 /**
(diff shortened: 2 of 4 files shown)


MitchCTiny:
Wondering if it would be worth adding any tests, or is it unnecessary after CI and QA

modules/tinymce/src/core/main/ts/api/html/SaxParser.ts
```diff
@@ -211,7 +211,7 @@ const checkBogusAttribute = (regExp: RegExp, attrString: string): string | null
  */
 const SaxParser = (settings?: SaxParserSettings, schema = Schema()): SaxParser => {
   settings = settings || {};
-  const doc = settings.document ?? document;
+  const doc = (settings.document ?? document).implementation.createHTMLDocument('parser');
```
lnewson, line 214: You probably don't want to do this... you can just use `document` for this use case as it isn't going to matter which version is used. The other thing is this will create a new document every time the content parses... that's not very good for performance. I'd suggest either using the same global or doing something similar as to what's done in the undoManager: https://github.com/tinymce/tinymce/blob/304fecd538e49fc23b5f85284f7dd056887550d9/modules/tinymce/src/core/main/ts/undo/Levels.ts#L12

lnewson:
> Wondering if it would be worth adding any tests, or is it unnecessary after CI and QA

Yes, this should have a regression test... otherwise someone might break this in future. Remember 5.10.x needs to be supported for quite some time yet.

modules/tinymce/src/core/main/ts/api/html/SaxParser.ts
TheSpyder: how often do we parse content though 🤔 

I told him to stick with `settings.document` in case it made some sort of difference we needed to retain.

lnewson: Surprisingly often in some cases... either way this is definitely a hot code path so we shouldn't be doing this.

lnewson: @MitchCTiny the change you've made here is no different to the old version, in fact it's a little worse. You're still constructing a new Document every time the SaxParser is initialised. We need to ensure it's only setup once, not every time SaxParser is initialised.

modules/tinymce/CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+- The `name` and `id` attributes of any `img` or `iframe` elements were incorrectly removed #TINY-8773
```
lnewson, line 10: Don't forget to bump the `modules/tinymce/package.json` version to 5.10.6 as well.

modules/tinymce/src/core/test/ts/browser/html/SaxParserTest.ts
```diff
@@ -284,6 +284,20 @@ describe('browser.tinymce.core.html.SaxParserTest', () => {
     parser.parse('<p title="gre>gererg"/>');
     assert.equal(writer.getContent(), '<p title="gre&gt;gererg"></p>', 'Parse elements with quoted > characters.');
     assert.deepEqual(counter.counts, { start: 1, end: 1 }, 'Parse elements with quoted > characters (counts).');
+
+    counter = createCounter(writer);
+    parser = SaxParser(counter, schema);
+    writer.reset();
+    parser.parse('<iframe id="someid" src="test" name="my_iframe"></iframe>');
```
tiny-ben-tran, line 291:
```suggestion
    const content = '<iframe id="someid" src="test" name="my_iframe"></iframe>';
    parser.parse(content);
    assert.equal(writer.getContent(), content, 'Parse iframe elements with name and id attributes.');
```
Just less code

tiny-ben-tran: I would also add one or more attributes

MitchCTiny: Theres dozens of cases in each test in this file, so I don't want to declare new const variables for each case.

modules/tinymce/src/core/main/ts/api/html/SaxParser.ts
MitchCTiny: Sorry, I misunderstood the method, should be better now

modules/tinymce/CHANGELOG.md
```suggestion
- The `name` and `id` attributes of some elements were incorrectly removed during serialization #TINY-8773
```
This can happen with other elements, so this just makes it more generic.

lnewson (approved):

ltrouton (approved):

PR #8015, opened 2022-07-28
TINY-8936: Reverted the undo level part of the fixes done for f5c2ce9

lnewson:
Related Ticket: TINY-8936

Description of Changes:

Reverts the undo level fixes so this works the same as it did in 5.x and we'll re-evaluate and provide a better fix for 6.2.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/plugins/autolink/main/ts/core/Keys.ts
+++ b/modules/tinymce/src/plugins/autolink/main/ts/core/Keys.ts
@@ -1,7 +1,6 @@
 import { Strings, Type } from '@ephox/katamari';
 
 import Editor from 'tinymce/core/api/Editor';
-import { EditorEvent } from 'tinymce/core/api/util/EventDispatcher';
 
 import * as Options from '../api/Options';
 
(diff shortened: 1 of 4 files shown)


HAFRMO (approved):

modules/tinymce/src/plugins/autolink/main/ts/core/Keys.ts
```diff
@@ -203,21 +200,17 @@ const handleSpacebar = (editor: Editor): void => {
 
 const handleBracket = handleSpacebar;
 
-const handleEnter = (editor: Editor, e: EditorEvent<KeyboardEvent>): void => {
+const handleEnter = (editor: Editor): void => {
   const result = parseCurrentLine(editor, -1);
   if (Type.isNonNullable(result)) {
```
ephox-mogran, line 205: Can you briefly explain why we had the mceInsertNewline before, and why we are removing it?

ephox-mogran (approved):

modules/tinymce/src/plugins/autolink/main/ts/core/Keys.ts
lnewson: Because you need to ensure the newline is inserted before doing the replacement if you want the undo to only undo the link change and not the newline instead.

ltrouton (approved):

modules/tinymce/src/plugins/autolink/main/ts/core/Keys.ts
lnewson: To word that differently, it was only added originally to fix the undo stack, but that's part of what caused the issues here so for now we're just abandoning it and undo will first undo the new line then the url conversion.

ephox-mogran: K. That makes sense.

PR #8017, opened 2022-07-29
TINY-8721: Make sure single value is respected for advlist options

ltrouton:
Related Ticket: TINY-8721

Description of Changes:
* Make sure  a single value for the `advlist_bullet_styles` and `advlist_number_styles` is handled correctly when using the toolbar buttons

Pre-checks:
* [X] Changelog entry added

Review:
* [X] Milestone set


modules/tinymce/src/plugins/advlist/main/ts/ui/Buttons.ts
```diff
@@ -80,21 +83,23 @@ const addSplitButton = (editor: Editor, id: string, tooltip: string, cmd: string
...
+    onAction: () => editor.queryCommandState(cmd) || value === '' ? editor.execCommand(cmd) : Actions.applyListFormat(editor, nodeName, value)
   });
 };
 
 const addControl = (editor: Editor, id: string, tooltip: string, cmd: string, nodeName: ListType, styles: string[]): void => {
   if (styles.length > 1) {
     addSplitButton(editor, id, tooltip, cmd, nodeName, styles);
   } else {
-    addButton(editor, id, tooltip, cmd, nodeName, styles);
+    addButton(editor, id, tooltip, cmd, nodeName, styles[0] || '');
```
MitchCTiny, line 102: Would it be better to call normalizeStyleValue on the first style[] element here? And replace the empty string with 'default' so that it is known an empty string will be returned?

MitchCTiny (approved):

modules/tinymce/src/plugins/advlist/main/ts/ui/Buttons.ts
ltrouton: I am not quite sure if I understand what you are suggesting so I would probably need to see a code suggestion.
In saying that, maybe just doing `addButton(editor, id, tooltip, cmd, nodeName, styles[0] || 'default');` would be slightly better/easier to reason about?

MitchCTiny:
```suggestion
const addButton = (editor: Editor, id: string, tooltip: string, cmd: string, nodeName: ListType, value: string): void => {
```
```suggestion
    addButton(editor, id, tooltip, cmd, nodeName, normalizeStyleValue(styles[0]));
```
I suppose I was meaning something like this, removing the `const value = normalizeStyleValue(styleValue);` from `addButton`, but I also like your suggestion, so up to you.

MitchCTiny: BTW I messed up the position of the suggestions, still getting the hang of it, so don't commit these ones haha

ltrouton: I thought that might be what you were referring to for a suggestion. I would have to make a small change to `normalizeStyleValue` but it should work 👍

ltrouton: I have pushed up a change more in line with what you were thinking. Thanks for the suggestion

modules/tinymce/src/plugins/advlist/main/ts/ui/Buttons.ts
```diff
@@ -80,21 +85,22 @@ const addSplitButton = (editor: Editor, id: string, tooltip: string, cmd: string
...
+    onAction: () => editor.queryCommandState(cmd) || styleValue === '' ? editor.execCommand(cmd) : Actions.applyListFormat(editor, nodeName, styleValue)
   });
 };
 
 const addControl = (editor: Editor, id: string, tooltip: string, cmd: string, nodeName: ListType, styles: string[]): void => {
   if (styles.length > 1) {
     addSplitButton(editor, id, tooltip, cmd, nodeName, styles);
   } else {
-    addButton(editor, id, tooltip, cmd, nodeName, styles);
+    addButton(editor, id, tooltip, cmd, nodeName, normalizeStyleValue(styles[0]));
```
lnewson, line 103: I wonder if we should just be normalizing these originally? That'd also save normalizing them every time the split toolbar opens. It also makes sure the logic is consistent, as I noticed this doesn't hand the `disc` or `decimal` styles being transformed to `default` like they do in the split buttons. So something like this as an example:
```ts
const normalizeStyleValue = (styleValue: string): string => {
  const style = styleValue === 'disc' || styleValue === 'decimal' ? 'default' : styleValue;
  return style === 'default' ? '' : style;
}

const addControl = (editor: Editor, id: string, tooltip: string, cmd: string, nodeName: ListType, styles: string[]): void => {
  const normalizedValues = Arr.map(styles, normalizeStyleValue);
  if (styles.length > 1) {
    addSplitButton(editor, id, tooltip, cmd, nodeName, normalizedValues);
  } else {
    addButton(editor, id, tooltip, cmd, nodeName, normalizedValues[0] ?? '');
  }
};
```

lnewson (approved):

TheSpyder (approved):

PR #8003, opened 2022-07-26
TINY-8861: fix cursor position after pasting a `codesample` snippet

TheSpyder:
Related Ticket: TINY-8861

Description of Changes:
* Ensure contenteditable state is set before inserting pasted `codesample` snippets
* Make better use of the `data-mce-highlighted` attribute flag rather than simply `contenteditable=false`
* See further discussion in #7996

Pre-checks:
* [x] Changelog entry added
* [x] Tests have been added (if applicable)
* [x] Branch prefixed with `feature/`, `hotfix/` or `spike/`

Review:
* [x] Milestone set

PR #8018, opened 2022-07-29
TINY-8861: Add selection assertions to codesample test

ltrouton:
Related Ticket: TINY-8861

Description of Changes:
* Adds some selection assertions that were missed in the tests added in #8003 

Pre-checks:

Review:
* [X] ~~Milestone set~~

### DIFF
--- a/modules/tinymce/package.json
+++ b/modules/tinymce/package.json
@@ -1,6 +1,6 @@
 {
   "name": "tinymce",
-  "version": "6.1.2",
+  "version": "6.1.3",
   "private": true,
   "repository": {
     "type": "git",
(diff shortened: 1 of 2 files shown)


lnewson (changes requested):
Thanks for this Luke, but as mentioned on slack (or was it a call?) we don't want this on `master` nor for it to be a `hotfix`. This should just target `develop` instead so it doesn't regress in future.

ltrouton:
> Thanks for this Luke, but as mentioned on slack (or was it a call?) we don't want this on `master` nor for it to be a `hotfix`, especially since we have no current plans for a 6.1.3. As such, this should just target `develop` instead so it doesn't regress in future.

Ah, that must have escaped my memory. Sorry about it. Will fix this up on Monday 👍

PR #8019, opened 2022-07-31
TINY-8940: Investigate and fix getAnchorElement changes

hamza0867:
Related Ticket: TINY-8940

Description of Changes:
* Fix a regression introduced in TINY-8832 where clicking in the editor would always show the link context toolbar when the option `link_context_toolbar` is set to true

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -14,7 +14,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The formatter can now apply a format to a `contenteditable="false"` element by wrapping it. Configurable using the `format_noneditable_selector` option. #TINY-8905
 
 ### Fixed
-- Fixed a regression where clicking in the editor would always show the link context toolbar when the option `link_context_toolbar` is set to true #TINY-8940
 - The Autolink plugin did not work when the text nodes in the content were fragmented #TINY-3723
 - Fixed various incorrect types on public APIs found while enabling TypeScript strict mode #TINY-8806
 - Text returned from `editor.getContent({format: 'text'})` would differ with blank lines between some browsers #TINY-8579
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The formatter can now apply a format to a `contenteditable="false"` element by wrapping it. Configurable using the `format_noneditable_selector` option. #TINY-8905
 
 ### Fixed
+- A regression where clicking in the editor would always show the link context toolbar when the option `link_context_toolbar` is set to true #TINY-8940
 - The Autolink plugin did not work when the text nodes in the content were fragmented #TINY-3723
 - Fixed various incorrect types on public APIs found while enabling TypeScript strict mode #TINY-8806
 - Text returned from `editor.getContent({format: 'text'})` would differ with blank lines between some browsers #TINY-8579
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -14,7 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The formatter can now apply a format to a `contenteditable="false"` element by wrapping it. Configurable using the `format_noneditable_selector` option. #TINY-8905
 
 ### Fixed
-- A regression where clicking in the editor would always show the link context toolbar when the option `link_context_toolbar` is set to true #TINY-8940
+- Fixed a regression where clicking in the editor would always show the link context toolbar when the option `link_context_toolbar` is set to true #TINY-8940
 - The Autolink plugin did not work when the text nodes in the content were fragmented #TINY-3723
 - Fixed various incorrect types on public APIs found while enabling TypeScript strict mode #TINY-8806
 - Text returned from `editor.getContent({format: 'text'})` would differ with blank lines between some browsers #TINY-8579
--- a/modules/tinymce/src/plugins/link/main/ts/core/Utils.ts
+++ b/modules/tinymce/src/plugins/link/main/ts/core/Utils.ts
@@ -78,6 +78,9 @@ const getAnchorElement = (editor: Editor, selectedElm?: Element): Optional<HTMLA
   }
 };
 
+const isInAnchor = (editor: Editor, selectedElm?: Element): boolean =>
+  getAnchorElement(editor, selectedElm).isSome();
+
 const getAnchorText = (selection: EditorSelection, anchorElm: Optional<HTMLAnchorElement>): string => {
   const text = anchorElm.fold(
     () => selection.getContent({ format: 'text' }),
@@ -297,6 +300,7 @@ export {
   getHref,
   isOnlyTextSelected,
   getAnchorElement,
+  isInAnchor,
   getAnchorText,
   applyRelTargetRules,
   hasProtocol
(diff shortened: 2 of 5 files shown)


modules/tinymce/CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The formatter can now apply a format to a `contenteditable="false"` element by wrapping it. Configurable using the `format_noneditable_selector` option. #TINY-8905
 
 ### Fixed
+- Fixed a regression where clicking in the editor would always show the link context toolbar when the option `link_context_toolbar` is set to true #TINY-8940
```
lnewson, line 17: This was never released to customers, so there was no regression to them and as such there should be no changelog entry.

```suggestion
```

modules/tinymce/src/plugins/link/test/ts/browser/ContextToolbarTest.ts
```diff
@@ -28,6 +29,9 @@ describe('browser.tinymce.plugins.link.ContextToolbarTest', () => {
     const editorBody = TinyDom.body(editor);
     editor.setContent('<a href="http://www.google.com">google</a>');
     Mouse.trueClickOn(editorBody, 'a');
+    // I believe this does not do much, since the toolbar will never be in
+    // in the editor body. The toolbar exists in the .tox-tinymce-aux portal
+    // which exists at the top level.
     UiFinder.notExists(editorBody, '.tox-toolbar button[aria-label="Link"]');
```
lnewson, line 35: Leaving a comment like this doesn't make sense/doesn't solve the issue. I agree though with your findings though.
```suggestion
    UiFinder.notExists(SugarBody.body(), '.tox-toolbar button[aria-label="Link"]');
```

modules/tinymce/src/plugins/link/test/ts/browser/ContextToolbarTest.ts
```diff
@@ -43,6 +47,22 @@ describe('browser.tinymce.plugins.link.ContextToolbarTest', () => {
...
+    const editor = hook.editor();
+    editor.options.set('link_context_toolbar', true);
+    editor.setContent('<p>google</p>');
+    Mouse.trueClickOn(TinyDom.body(editor), 'p');
+    // You might be wondering why I did not use UiFinder.notExists.
+    // And the reason being simply it doesn't work in this case.
+    // I am not sure why, but you can put a  link in the editor content
+    // and click on it. The tollbar will show as it should, while the call
+    // UiFinder.notExists(SugarBody.body(), '.tox-toolbar') will still pass
+    // when it should fail since the toolbar does exist
```
lnewson, line 60: I'm not sure what you did here, but I tested this locally and `notExists` does work if I revert your changes to `Controls.ts`. Additionally this is the exact same implementation of `UiFinder.notExists`, so it won't be helping anyways/will have the same issues. Here's what I used if it helps:
```ts
    await Waiter.pWait(50); // Need to wait a little for the toolbar to trigger
    UiFinder.notExists(SugarBody.body(), '.tox-pop__dialog .tox-toolbar');
```

Now the couple of things you'll note are different are that I added the `.tox-pop__dialog` part to the selector because we want to be sure it's the context toolbar that doesn't show and the toolbar class alone can be used elsewhere. The second thing I added was the wait, as the context toolbar does **not** trigger synchronously... it has a `17ms` timeout/throttle as can be seen here, which is potentially why you thought `notExists` wasn't working: https://github.com/tinymce/tinymce/blob/236891ffcdf2fd1b698d020e67c31838fb110019/modules/tinymce/src/themes/silver/main/ts/ui/context/ContextToolbar.ts#L244

lnewson (changes requested):
The fix itself looks good 👍 just some cleanup work to do elsewhere.

modules/tinymce/src/plugins/link/main/ts/ui/Controls.ts
```diff
@@ -95,7 +95,7 @@ const setupContextToolbars = (editor: Editor): void => {
       onSetup: Actions.toggleActiveState(editor)
     },
     label: 'Link',
-    predicate: (node) => !!Utils.getAnchorElement(editor, node) && Options.hasContextToolbar(editor),
+    predicate: (node) => Utils.getAnchorElement(editor, node).isSome() && Options.hasContextToolbar(editor),
```
lnewson, line 98: We seem to be using `Utils.getAnchorElement(editor, node).isSome()` quite a bit/have duplicated this in a number of places. Instead of doing all this duplication, can we move this to a helper function inside `Utils` as well? For example:
```ts
const isInAnchor = (editor: Editor, selectedElm?: Element): boolean =>
  getAnchorElement(editor, selectedElm).isSome()
```

modules/tinymce/src/plugins/link/test/ts/browser/ContextToolbarTest.ts
hamza0867: That seemed to do the trick 👍

modules/tinymce/src/plugins/link/main/ts/ui/Controls.ts
hamza0867: Done!

modules/tinymce/src/plugins/link/main/ts/core/Utils.ts
```diff
@@ -78,6 +78,9 @@ const getAnchorElement = (editor: Editor, selectedElm?: Element): Optional<HTMLA
   }
 };
 
+const isInAnchor = (editor: Editor, selectedElm?: Element): boolean =>
```
lnewson, line 81: Can you also update the uses in `Actions.ts` as well to remove this duplication please?

- [ ] Replace all uses of `getAnchorElement(editor, selectedElm).isSome()` with `isInAnchor`

modules/tinymce/src/plugins/link/test/ts/browser/ContextToolbarTest.ts
```diff
@@ -43,6 +43,16 @@ describe('browser.tinymce.plugins.link.ContextToolbarTest', () => {
     await UiFinder.pWaitForState<HTMLInputElement>('check link content', SugarBody.body(), '.tox-toolbar input', (ele) => ele.dom.value === 'http://www.google.com');
   });
 
+  it('TINY-8940: toolbar doesn\'t show on non link elements', async () => {
+    const editor = hook.editor();
+    editor.options.set('link_context_toolbar', true);
+    editor.setContent('<p>google</p>');
+    Mouse.trueClickOn(TinyDom.body(editor), 'p');
+    await Waiter.pWait(100);
```
lnewson, line 51:
```suggestion
    await Waiter.pWait(50);
```
I gave 50 for a specific reason, as that's as high as we like to use normally as it slows done tests if we have too many long waits like this. Additionally, 50 is at least 2 times higher than the 17ms throttle we have in place so it should have given it plenty of time to render.

lnewson (approved):

modules/tinymce/src/plugins/link/main/ts/ui/Controls.ts
```diff
@@ -95,7 +95,7 @@ const setupContextToolbars = (editor: Editor): void => {
       onSetup: Actions.toggleActiveState(editor)
     },
     label: 'Link',
-    predicate: (node) => !!Utils.getAnchorElement(editor, node) && Options.hasContextToolbar(editor),
+    predicate: (node) => Utils.isInAnchor(editor, node) && Options.hasContextToolbar(editor),
```
lnewson, line 98: I know this was like this before, but for performance this would be better if the order was swapped since getting the anchor by walking the DOM is much more over than checking if an option is enabled. Feel free to ignore this suggestion if you'd prefer since it's not related to this change.

```suggestion
    predicate: (node) => Options.hasContextToolbar(editor) && Utils.isInAnchor(editor, node),
```

modules/tinymce/src/plugins/link/main/ts/core/Utils.ts
hamza0867: Done!

HAFRMO (approved):

TheSpyder (approved):

PR #8024, opened 2022-08-01
TINY-8778: Fixed the lookup patterns taking priority over default patterns even though the latter has a better match

tiny-ben-tran:
Related Ticket: TINY-8778

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/textpatterns/core/BlockPattern.ts
+++ b/modules/tinymce/src/core/main/ts/textpatterns/core/BlockPattern.ts
@@ -55,9 +55,8 @@ const applyPattern = (editor: Editor, match: BlockPatternMatch): boolean => {
   return true;
 };
 
-const sortPatterns = <P extends Pattern>(patterns: P[]): P[] => Arr.sort(patterns, (a, b) => {
-return b.start.length - a.start.length
-});
+const sortPatterns = <P extends Pattern>(patterns: P[]): P[] =>
+  Arr.sort(patterns, (a, b) => b.start.length - a.start.length);
 
 // Finds a matching pattern to the specified text
 const findPattern = <P extends Pattern>(patterns: P[], text: string): Optional<P> => {
--- a/modules/tinymce/src/core/main/ts/textpatterns/core/BlockPattern.ts
+++ b/modules/tinymce/src/core/main/ts/textpatterns/core/BlockPattern.ts
@@ -56,10 +56,7 @@ const applyPattern = (editor: Editor, match: BlockPatternMatch): boolean => {
 };
 
 const sortPatterns = <P extends Pattern>(patterns: P[]): P[] => Arr.sort(patterns, (a, b) => {
-  if (a.start.length === b.start.length) {
-    return 0;
-  }
-  return a.start.length > b.start.length ? -1 : 1;
+return b.start.length - a.start.length
 });
 
 // Finds a matching pattern to the specified text
--- a/modules/tinymce/src/core/main/ts/textpatterns/core/BlockPattern.ts
+++ b/modules/tinymce/src/core/main/ts/textpatterns/core/BlockPattern.ts
@@ -57,9 +57,17 @@ const applyPattern = (editor: Editor, match: BlockPatternMatch): boolean => {
 
 // Finds a matching pattern to the specified text
 const findPattern = <P extends Pattern>(patterns: P[], text: string): Optional<P> => {
+  const sortedPatterns = sortPatterns(patterns);
   const nuText = text.replace(Unicode.nbsp, ' ');
-  return Arr.find(patterns, (pattern) => text.indexOf(pattern.start) === 0 || nuText.indexOf(pattern.start) === 0);
+  return Arr.find(sortedPatterns, (pattern) => text.indexOf(pattern.start) === 0 || nuText.indexOf(pattern.start) === 0);
 };
+
+const sortPatterns = <T extends Pattern>(patterns: T[]): T[] => Arr.sort(patterns, (a, b) => {
+  if (a.start.length === b.start.length) {
+    return 0;
+  }
+  return a.start.length > b.start.length ? -1 : 1;
+});
 
 const findPatterns = (editor: Editor, block: Element, patternSet: PatternSet, normalizedMatches: boolean): BlockPatternMatch[] => {
   const dom = editor.dom;
--- a/modules/tinymce/src/core/main/ts/textpatterns/core/BlockPattern.ts
+++ b/modules/tinymce/src/core/main/ts/textpatterns/core/BlockPattern.ts
@@ -55,19 +55,19 @@ const applyPattern = (editor: Editor, match: BlockPatternMatch): boolean => {
   return true;
 };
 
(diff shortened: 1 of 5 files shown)


modules/tinymce/src/core/test/ts/atomic/textpatterns/FindBlockPatternsTest.ts
```diff
@@ -27,7 +28,7 @@ describe('atomic.tinymce.textpatterns.FindBlockPatternsTest', () => {
     const defaultPatterns = patternSet.blockPatterns;
 
     const testFindStartPattern = (text: string, expectedPattern: string) => {
-      const actual = BlockPattern.findPattern(defaultPatterns, text).getOrNull();
+      const actual = BlockPattern.findPattern(sortPatterns(defaultPatterns), text).getOrNull();
```
lnewson, line 31: Instead of having to do this in every location, this would be better within `findPattern` I believe? It's also not going to add any performance overhead, in fact it'd actually reduce it as then the sorting wouldn't happen when just doing the inline key handling.

modules/tinymce/src/core/main/ts/textpatterns/utils/Utils.ts
```diff
@@ -38,17 +38,23 @@ const deleteRng = (dom: DOMUtils, rng: Range, isRoot: (e: Node) => boolean, clea
 const getParentBlock = (editor: Editor, rng: Range): Optional<Element> =>
   Optional.from(editor.dom.getParent(rng.startContainer, editor.dom.isBlock));
 
+const sortPatterns = <T extends Pattern>(patterns: T[]): T[] => Arr.sort(patterns, (a, b) => {
```
lnewson, line 41: This logic is specific to block patterns so this name/location may not make sense.

modules/tinymce/src/core/main/ts/textpatterns/core/BlockPattern.ts
```diff
@@ -57,10 +57,18 @@ const applyPattern = (editor: Editor, match: BlockPatternMatch): boolean => {
 
 // Finds a matching pattern to the specified text
 const findPattern = <P extends Pattern>(patterns: P[], text: string): Optional<P> => {
+  const sortedPatterns = sortPatterns(patterns);
   const nuText = text.replace(Unicode.nbsp, ' ');
-  return Arr.find(patterns, (pattern) => text.indexOf(pattern.start) === 0 || nuText.indexOf(pattern.start) === 0);
+  return Arr.find(sortedPatterns, (pattern) => text.indexOf(pattern.start) === 0 || nuText.indexOf(pattern.start) === 0);
 };
 
+const sortPatterns = <T extends Pattern>(patterns: T[]): T[] => Arr.sort(patterns, (a, b) => {
```
lnewson, line 65: Functions shouldn't be used before the are defined if we can help it... so can you please move this somewhere that makes sense above `findPattern`?

lnewson: Also for consistency, the generic should be called `P` here as it is above.

ephox-mogran, line 65: Is the basic idea here that we feel that if a pattern is _longer_, it's a higher priority?

tiny-ben-tran: We want to return the best match for the patterns like markdown headings syntax (`#`, `##`, etc)

MitchCTiny (approved):

ltrouton (approved):

modules/tinymce/src/core/main/ts/textpatterns/core/BlockPattern.ts
```diff
@@ -55,10 +55,18 @@ const applyPattern = (editor: Editor, match: BlockPatternMatch): boolean => {
   return true;
 };
 
+const sortPatterns = <P extends Pattern>(patterns: P[]): P[] => Arr.sort(patterns, (a, b) => {
+  if (a.start.length === b.start.length) {
+    return 0;
+  }
+  return a.start.length > b.start.length ? -1 : 1;
```
TheSpyder, line 62: It doesn't need to be `1` and `-1`, only >0 and <0, so this whole thing can become
```suggestion
return b.start.length - a.start.length
```

modules/tinymce/src/core/main/ts/textpatterns/core/BlockPattern.ts
```diff
@@ -55,10 +55,18 @@ const applyPattern = (editor: Editor, match: BlockPatternMatch): boolean => {
...
+});
+
 // Finds a matching pattern to the specified text
 const findPattern = <P extends Pattern>(patterns: P[], text: string): Optional<P> => {
+  const sortedPatterns = sortPatterns(patterns);
```
TheSpyder, line 67: Instead of sorting them every time they're used, can't we sort them when the pattern set is created?

TheSpyder (approved):

modules/tinymce/src/core/main/ts/textpatterns/core/BlockPattern.ts
tiny-ben-tran: Sorting when the pattern was set, was the original logic and it caused the bug that this PR is meant to fix

lnewson: The other thing to remember is that the pattern set is generated every time we attempt to find the patterns anyways due to the dynamic lookups, so moving it there would make no difference anyways.

modules/tinymce/src/core/main/ts/textpatterns/core/BlockPattern.ts
```diff
@@ -55,10 +55,15 @@ const applyPattern = (editor: Editor, match: BlockPatternMatch): boolean => {
   return true;
 };
 
+const sortPatterns = <P extends Pattern>(patterns: P[]): P[] => Arr.sort(patterns, (a, b) => {
+return b.start.length - a.start.length
+});
```
lnewson, line 60:
```suggestion
const sortPatterns = <P extends Pattern>(patterns: P[]): P[] =>
  Arr.sort(patterns, (a, b) => b.start.length - a.start.length);
```
I think this is more readable?

modules/tinymce/src/core/main/ts/textpatterns/utils/Utils.ts
```diff
@@ -43,7 +43,6 @@ const resolveFromDynamicPatterns = (patternSet: PatternSet, block: Element, befo
     text: beforeText,
     block
   });
-
```
lnewson, line 46: This line should not have been deleted... please add it back as it makes it harder to read without this separator.

lnewson (approved):

modules/tinymce/src/core/main/ts/textpatterns/core/BlockPattern.ts
lnewson: Also as I noted in a previous comment, doing it when we create the pattern set is actually worse for performance in the  "inline key" case since it never uses block patterns.

PR #8027, opened 2022-08-02
 TINY-8840:  Added new SPAP promotion to the editor

HAFRMO:
Related Ticket:

Description of Changes:
Added SPAP to the editor as dictated by earlier spike

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/tinymce/src/themes/silver/main/ts/Render.ts
```diff
@@ -68,6 +68,9 @@ export interface RenderArgs {
   readonly height: string;
 }
 
+const SpapMessage = 'Try Premium!';
```
lnewson, line 71: Can we please move all this "SPAP" logic out into a separate module? It just makes this messy. Ideally it'd also not be called SPAP as that is not clear what it means... I'd suggest renaming all this to just be `promotion` instead.

To clarify, I mean have a separate module that has a `render` function or similar that returns the promotion `AlloySpec` which is then used in the `OuterContainer.parts.spap` call below.

modules/tinymce/src/themes/silver/main/ts/Render.ts
```diff
@@ -166,14 +171,16 @@ const setup = (editor: Editor): RenderInfo => {
       }
     };
 
+    const menuBarCollection = shouldHavePromotion ? [ partSpap, partMenubar ] : [ partMenubar ];
```
lnewson, line 174: "menubar" is one word, so it shouldn't be `menuBarCollection` and instead should be `menubarCollection`

modules/tinymce/src/themes/silver/main/ts/api/Options.ts
```diff
@@ -302,6 +307,12 @@ const useBranding = option('branding');
 const getResize = option('resize');
 const getPasteAsText = option('paste_as_text');
 const getSidebarShow = option('sidebar_show');
+const promotionEnabled = option('promotion');
+
+const hasApiKey = (editor: Editor) => {
+  const key = option('api_key')(editor);
```
lnewson, line 313:
```suggestion
  const key = editor.options.get('api_key');
```
This should just be doing this, either that or the option should be added as a function above.

modules/tinymce/CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - New `text_patterns_lookup` option to provide additional text patterns dynamically #TINY-8778
+- New SPAP element to editor to promote premium #TINY-8840
```
lnewson, line 11: We need to document the `promotion` option to be able to disable this... additionally no customer is going to know what SPAP means. So maybe something like this:
```suggestion
- New promotion element has been added to the UI and can be disabled using the new `promotion` option #TINY-8840
```

modules/tinymce/src/themes/silver/main/ts/Render.ts
```diff
@@ -155,6 +158,8 @@ const setup = (editor: Editor): RenderInfo => {
     const hasMultipleToolbar = Options.isMultipleToolbars(editor);
     const hasToolbar = Options.isToolbarEnabled(editor);
     const hasMenubar = Options.isMenubarEnabled(editor);
+    const shouldHavePromotion = Options.promotionEnabled(editor) && !Options.hasApiKey(editor);
```
lnewson, line 161: This logic would likely be better off in the `Options` module so it's consistent across the board (e.g. `promotionEnabled` should return false if an API key is enabled).

modules/tinymce/src/themes/silver/main/ts/Render.ts
lnewson: Also these names aren't camel case when they should be.

modules/tinymce/src/themes/silver/main/ts/Render.ts
```diff
@@ -68,6 +68,9 @@ export interface RenderArgs {
   readonly height: string;
 }
 
+const SpapMessage = 'Try Premium!';
+const SpapLink = 'http://tiny.cloud';
```
lnewson, line 72: This URL does not look correct. Are we sure this is what it should be pointed to? At the very least it should be `https://www.tiny.cloud/`. That said though, I'd expect this to point to something highlighting the premium features.

modules/tinymce/src/themes/silver/main/ts/Render.ts
```diff
@@ -184,6 +191,30 @@ const setup = (editor: Editor): RenderInfo => {
...
+        tag: 'div',
+        classes: [ 'tox-spap' ],
+      },
+      components: [
+        {
+          dom: {
+            tag: 'a',
+            attributes: {
+              href: SpapLink,
+              target: '_blank'
```
lnewson, line 206: This needs `rel="noopener"` to prevent security concerns.

modules/tinymce/src/themes/silver/test/ts/browser/editor/header/SpapTest.ts
```diff
@@ -0,0 +1,35 @@
...
+describe('browser.tinymce.themes.silver.editor.header.SpapTest', () => {
+  const promotionClass = 'tox-spap';
+
+  context('spap turned off', () => {
+    const hook = TinyHooks.bddSetup<Editor>({
+      base_url: '/project/tinymce/js/tinymce',
+      promotion: false
+    }, []);
+
+    it('promotion should not be displayed', async () => {
```
lnewson, line 17: Tests should be prefixed with Jira numbers.

modules/tinymce/src/themes/silver/test/ts/browser/editor/header/SpapTest.ts
```diff
@@ -0,0 +1,35 @@
...
+      const editor = hook.editor();
+
+      await Waiter.pTryUntil('Should not have spap', () => UiFinder.notExists(SugarElement.fromDom(editor.getContainer()), '.' + promotionClass));
```
lnewson, line 20: This waiter is doing nothing meaningful... we should just remove it.

modules/tinymce/src/themes/silver/test/ts/browser/editor/header/StickyHeaderInitialPlacementTest.ts
```diff
@@ -17,6 +17,7 @@ describe('browser.tinymce.themes.silver.editor.header.StickyHeaderInitialPlaceme
     it(`Test toolbar initial placement with toolbar_location: ${test.location} and height: ${test.height}`, async () => {
       const editor = await McEditor.pFromSettings<Editor>({
         base_url: '/project/tinymce/js/tinymce',
+        promotion: false,
```
lnewson, line 20: We probably should put this in the mcagar overrides instead maybe, as we wouldn't want this showing in most cases. See https://github.com/tinymce/tinymce/blob/6afd94ff80c9ffdebae118136f069a2ce1ace615/modules/mcagar/src/main/ts/ephox/mcagar/api/McEditor.ts#L13 and https://github.com/tinymce/tinymce/blob/6afd94ff80c9ffdebae118136f069a2ce1ace615/modules/mcagar/src/main/ts/ephox/mcagar/loader/Loader.ts#L91

modules/tinymce/src/themes/silver/test/ts/browser/editor/header/SpapTest.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  context('spap turned on', () => {
+    const hook = TinyHooks.bddSetup<Editor>({
+      base_url: '/project/tinymce/js/tinymce',
+    }, []);
+
+    it('promotion should not be displayed', async () => {
+      const editor = hook.editor();
+
+      await Waiter.pTryUntil('Should have spap', () => UiFinder.exists(SugarElement.fromDom(editor.getContainer()), '.' + promotionClass));
```
lnewson, line 32: Likewise the waiter here is not doing anything meaningful.

modules/tinymce/src/themes/silver/main/ts/api/Options.ts
```diff
@@ -257,6 +257,11 @@ const register = (editor: Editor): void => {
     default: true
   });
 
+  registerOption('promotion', {
```
lnewson, line 260: This new option needs adding in the `OptionTypes.ts` module so it's in the bundled types.

modules/oxide/src/less/theme/components/menubar/spap.less
```diff
@@ -0,0 +1,33 @@
...
+
+// private variables
+@_spap-row-separator-color: escape('@{menubar-row-separator-color}');
+@_spap-height: @spap-select-height + (@spap-select-spacing-y * 2) + 1px;
+@_spap-divider-height: @_spap-height - 1px;
+@spap-select-height: @toolbar-button-height;
+@spap-select-spacing-y: @toolbar-button-spacing-y;
+
+.tox {
+  .tox-spap {
```
lnewson, line 16: The cursor for this is incorrect, as atm it shows the text icon instead of a pointer to show it's a button.

modules/tinymce/src/themes/silver/main/ts/ui/general/OuterContainer.ts
```diff
@@ -252,6 +252,13 @@ const partHeader = Composite.partType.optional<OuterContainerSketchDetail, Heade
   ]
 });
 
+const partSPAP = Composite.partType.optional<OuterContainerSketchDetail, HeaderSpec>({
+  name: 'spap',
```
lnewson, line 256: This should have a render function and be called like `renderHeader` above. This would also work well with what I suggested previously about moving the SPAP logic to a separate module.

modules/tinymce/CHANGELOG.md
HAFRMO: I wasn't sure if we were to document the option or not, due to some suggestions that it shouldn't be. But I might have misunderstood something.

modules/tinymce/src/themes/silver/main/ts/Render.ts
HAFRMO: No link was provided, so I added the first one available. I wasn't sure if this one should point to anything specific.

modules/tinymce/src/themes/silver/test/ts/browser/editor/header/SpapTest.ts
HAFRMO: Ah, completely forgot, that's my bad.

modules/tinymce/src/themes/silver/main/ts/Render.ts
lnewson: Actually, @ephox-mogran reminded me that this doesn't even really make sense to do the API key check within TinyMCE. We'd be better adding the `promotion: false` to Tiny Cloud.

modules/tinymce/src/themes/silver/main/ts/ui/promotion/Promotion.ts
```diff
@@ -0,0 +1,45 @@
...
+
+import { UiFactoryBackstageShared } from '../../backstage/Backstage';
+
+const promotionMessage = 'Try Premium!';
+const promotionLink = 'http://tiny.cloud';
+
+export interface PromotionSpec extends SimpleSpec {
+  editor: Editor;
+  sharedBackstage: UiFactoryBackstageShared;
+}
```
lnewson, line 13: Why do we need this? It's never used below is all 🤷

modules/tinymce/src/themes/silver/main/ts/ui/promotion/Promotion.ts
```diff
@@ -0,0 +1,45 @@
...
+const renderPromotion = (spec: PromotionSpec): AlloySpec => {
+  return {
+    uid: spec.uid,
+    dom: spec.dom,
+
+    components: [
+      {
+        dom: {
+          tag: 'div',
+          classes: [ 'tox-promotion' ],
```
lnewson, line 24: This doesn't look right... won't this result in 2 `tox-promotion` divs being created? We should either be merging this with the `spec.dom` logic instead I believe or just not include it.

modules/tinymce/src/themes/silver/test/ts/browser/editor/header/PromotionTest.ts
```diff
@@ -0,0 +1,35 @@
+import { UiFinder } from '@ephox/agar';
+import { context, describe, it } from '@ephox/bedrock-client';
+import { SugarElement } from '@ephox/sugar';
+import { TinyHooks } from '@ephox/wrap-mcagar';
+
+import Editor from 'tinymce/core/api/Editor';
+
+describe('browser.tinymce.themes.silver.editor.header.PromotionTest', () => {
```
lnewson, line 8: Can we please make sure to add a `Assertions.assertStructure` check to these to ensure the DOM is correct for the promotion given my concern in the previous comment?

modules/tinymce/src/themes/silver/test/ts/browser/editor/header/PromotionTest.ts
```diff
@@ -0,0 +1,35 @@
...
+  const promotionClass = 'tox-promotion';
+
+  context('TINY-8840: promotion turned off', () => {
```
lnewson, line 11: The Jira number should be on the `it` instead of `context`, as that's what it's specific to.

modules/mcagar/src/main/ts/ephox/mcagar/api/McEditor.ts
```diff
@@ -12,6 +12,7 @@ const pFromElement = <T extends EditorType = EditorType>(element: SugarElement<E
   return new Promise((resolve, reject) => {
     const nuSettings: Record<string, any> = {
       toolbar_mode: 'wrap',
+      promotion: false,
```
lnewson, line 15: These changes will need a McAgar changelog added and if it's not a patch release (this might be a minor? not really sure what to call it), then we'd also need to add the version to `versions.txt`.

modules/oxide/src/less/theme/components/menubar/promotion.less
```diff
@@ -0,0 +1,33 @@
...
+
+@promotion-background-color: @background-color;
+@promotion-background: url("data:image/svg+xml;charset=utf8,%3Csvg height='@{_promotion-height}' viewBox='0 0 40 @{_promotion-height}' width='40' xmlns='http://www.w3.org/2000/svg'%3E%3Crect x='0' y='@{_promotion-divider-height}' width='100' height='1' fill='@{_promotion-row-separator-color}'/%3E%3C/svg%3E") left 0 top 0 @promotion-background-color;
+
+// private variables
+@_promotion-row-separator-color: escape('@{menubar-row-separator-color}');
+@_promotion-height: @promotion-select-height + (@promotion-select-spacing-y * 2) + 1px;
+@_promotion-divider-height: @_promotion-height - 1px;
+@promotion-select-height: @toolbar-button-height;
+@promotion-select-spacing-y: @toolbar-button-spacing-y;
```
lnewson, line 13: These don't appear to be used?

modules/tinymce/CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - New `text_patterns_lookup` option to provide additional text patterns dynamically #TINY-8778
+- New Promotion element to editor to promote premium #TINY-8840
```
TheSpyder, line 11:
```suggestion
- New Promotion menu element to promote premium features #TINY-8840
```

modules/tinymce/src/themes/silver/main/ts/ui/promotion/Promotion.ts
HAFRMO: Because I got distracted trying to get the tests to pass and forgot to delete it. Fixed.

modules/tinymce/src/themes/silver/main/ts/ui/promotion/Promotion.ts
HAFRMO: I was going to check that, but forgot. My bad.

modules/oxide/src/less/theme/components/menubar/promotion.less
HAFRMO: They are used in the above `promotion-background`-definition.

lnewson: Ahh I had missed that sorry... in that case though then they are using the wrong fallbacks, because this is in the menubar so this should be using `@menubar-select-` not `@toolbar-button-`.

lnewson: Also the last 2 are not private (no underscore prefix) so they should be above the background anyways. Either that or they should be named properly as a private variable.

modules/agar/CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
 
+### Added
```
lnewson, line 9: Ahh... no changes were made to `agar`. I think you've put this in the wrong place, as this should be in `mcagar`.

- [ ] Fix changelogs and versions.txt to update `mcagar` instead of `agar`

modules/tinymce/src/themes/silver/test/ts/browser/editor/header/PromotionTest.ts
```diff
@@ -0,0 +1,73 @@
...
+  context('promotion turned on', () => {
+    const hook = TinyHooks.bddSetup<Editor>({
+      base_url: '/project/tinymce/js/tinymce',
+      promotion: true
+    }, []);
+
+    it('TINY-8840: promotion should not be displayed', () => {
+      const editor = hook.editor();
+
+      const editorContainer = SugarElement.fromDom(editor.getContainer());
```
lnewson, line 32: This should be using `TinyDom.container(editor)` instead to get access to that.

modules/tinymce/src/themes/silver/test/ts/browser/editor/header/PromotionTest.ts
```diff
@@ -0,0 +1,73 @@
...
+                          classes: [ arr.has('tox-promotion-link') ],
+                          attrs: {
+                            rel: str.is('noopener'),
+                            target: str.is('_blank'),
+                            href: str.startsWith('http')
+                          }
+                        }),
+                      ]
+                    }),
+                    s.element('div', {}),
```
lnewson, line 58: This really should be checking the next element is the menubar, otherwise it could be something else unexpected.

Also we could simplify this/remove some of the unclear `s.element('div', {})` calls if instead we only do the assertion on the `tox-editor-header`. For example:
```ts
const editorContainer = TinyDom.container(editor);
const header = SelectorFind.descendant(editorContainer, '.tox-editor-header');
Assertions.assertStructure(
  'Editor header structure',
  ApproxStructure.build((s, str, arr) => s.element('div', {
    ...
  })),
  header
);
```
We also could consider using `s.theRest()` or `s.oneOrMore()` if we don't care about the contents of the elements after.

modules/tinymce/src/themes/silver/test/ts/browser/editor/header/PromotionTest.ts
```diff
@@ -0,0 +1,73 @@
...
+    });
+  });
+
+  context('promotion turned on', () => {
+    const hook = TinyHooks.bddSetup<Editor>({
+      base_url: '/project/tinymce/js/tinymce',
+      promotion: true
+    }, []);
+
+    it('TINY-8840: promotion should not be displayed', () => {
```
lnewson, line 29:
```suggestion
    it('TINY-8840: promotion should be displayed', () => {
```

modules/tinymce/src/themes/silver/test/ts/browser/editor/header/PromotionTest.ts
HAFRMO: I swear I have done this correctly before...

modules/oxide/src/less/theme/components/menubar/promotion.less
HAFRMO: Alright, I'll use the menubar ones.

lnewson (approved):

modules/tinymce/src/themes/silver/main/ts/Render.ts
lnewson: According to what was said vaguely on the ticket yesterday, it seems this is still TBD... so can you please ensure a follow up JIRA is logged and then add a TODO here?

- [ ] Log follow up Jira to correct this and add a TODO comment referencing the Jira

modules/tinymce/src/themes/silver/main/ts/ui/promotion/Promotion.ts
```diff
@@ -0,0 +1,30 @@
+import { AlloySpec, SimpleSpec } from '@ephox/alloy';
+
+// TODO final text and link to be decided in https://ephocks.atlassian.net/browse/TINY-8954
+const promotionMessage = 'Try Premium!';
```
lnewson, line 4: According to the edits made to the Jira overnight, this should say `Upgrade` for now 🤷

modules/mcagar/CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
 
+### Added
+- Added `promotion` set to `false` as a default setting.
```
lnewson, line 10:
```suggestion
### Changed
- The new `promotion` option added in TinyMCE 6.2 is set to `false` by default.
```
Added is meant to be new features, since this isn't new then it should probably be either an enhancement or a simple change... I figured the latter made more sense in this usecase 🤷

modules/oxide/src/less/theme/components/menubar/promotion.less
```diff
@@ -0,0 +1,34 @@
...
+    background-color: @promotion-background-color;
+    cursor: pointer;
+    grid-column: 2;
+    grid-row: 1;
+    padding-inline-end: 8px;
+    padding-inline-start: 4px;
+    padding-top: 5px;
+  }
+
+  .tox-promotion-link {
```
lnewson, line 27: I noticed while reviewing the design spec today this is missing the hover states, so that needs adding/fixing as well. According to the figma details it's supposed to have the following background colors:
- Normal - #E8F1F8
- Hover - #B4D7FF

The design doc also had an icon next to the text... not sure if that's required but it'd be worth double checking or maybe putting in the follow up Jira.

- [ ] Fix hover/focus styles to make the design document

modules/oxide/src/less/theme/components/menubar/promotion.less
```diff
@@ -0,0 +1,34 @@
...
+
+// private variables
+@_promotion-row-separator-color: escape('@{menubar-row-separator-color}');
+@_promotion-height: @_promotion-select-height + (@_promotion-select-spacing-y * 2) + 1px;
+@_promotion-divider-height: @_promotion-height - 1px;
+@_promotion-select-height: @menubar-select-height;
+@_promotion-select-spacing-y: @menubar-select-spacing-y;
+
+.tox {
+  .tox-promotion {
```
lnewson, line 16: I noticed while reviewing the design spec today this is missing the hover states, so that needs adding/fixing as well. According to the figma details it's supposed to have the following background colors:
- Normal - #E8F1F8
- Hover - #B4D7FF

The design doc also had an icon next to the text... not sure if that's required but it'd be worth double checking or maybe putting in the follow up Jira.

- [ ] Fix hover/focus styles to match the design document

lnewson: Actually the link text also doesn't match the design doc either, as it's supposed to be `#086BE6` for the oxide skin.

- [ ] Fix link text color to be `#086BE6`

modules/oxide/src/less/theme/components/menubar/promotion.less
```diff
@@ -0,0 +1,34 @@
...
+    padding-inline-end: 8px;
+    padding-inline-start: 4px;
+    padding-top: 5px;
+  }
+
+  .tox-promotion-link {
+    background: lightblue;
+    border-radius: 5px;
+    font-size: 12px;
+    padding: 5px;
```
lnewson, line 31:
```suggestion
    padding: 4px 8px;
```
This is what the padding is supposed to be according to the design document.

modules/oxide/src/less/theme/components/menubar/promotion.less
lnewson: This also looks like it's meant to be aligned with the other menu button text, which doesn't happen currently.

- [ ] Review and make sure this matches the design spec

HAFRMO: Yeah, I followed the spiked design, I thought it'd been made with the finished design in mind. Should have double-checked just to be sure.

modules/oxide/src/less/theme/components/menubar/promotion.less
```diff
@@ -0,0 +1,46 @@
...
+@promotion-background-hover-color: #B4D7FF;
+@promotion-background-active-color: #D9EDF7;
+@promotion-color: #086BE6;
+
+// private variables
+@_promotion-row-separator-color: escape('@{menubar-row-separator-color}');
+@_promotion-height: @_promotion-select-height + (@_promotion-select-spacing-y * 2) + 1px;
+@_promotion-divider-height: @_promotion-height - 1px;
+@_promotion-select-height: @menubar-select-height;
+@_promotion-select-spacing-y: @menubar-select-spacing-y;
```
lnewson, line 15: I know I commented about this previously, but these definitely aren't used anymore so we can remove them.

```suggestion
```

modules/oxide/src/less/theme/components/menubar/promotion.less
```diff
@@ -0,0 +1,46 @@
+//
+// Promotion
+//
+
+@promotion-background-color: #E8F1F8;
+@promotion-background-hover-color: #B4D7FF;
+@promotion-background-active-color: #D9EDF7;
+@promotion-color: #086BE6;
```
lnewson, line 8: This should probably be called `promotion-text-color` for consistency with other variables and it also makes it clearer what it applies to.

modules/oxide/src/less/theme/components/menubar/promotion.less
HAFRMO: Right, I forgot to remove them when I removed the background. That's entirely on me.

modules/tinymce/src/themes/silver/test/ts/browser/editor/header/PromotionTest.ts
```diff
@@ -0,0 +1,64 @@
+import { ApproxStructure, Assertions, UiFinder } from '@ephox/agar';
+import { context, describe, it } from '@ephox/bedrock-client';
+import { SelectorFind } from '@ephox/sugar';
+import { TinyDom, TinyHooks } from '@ephox/wrap-mcagar';
+
+import Editor from 'tinymce/core/api/Editor';
+
+describe('browser.tinymce.themes.silver.editor.header.PromotionTest', () => {
+  const promotionClass = 'tox-promotion';
```
spocke, line 9: I guess this could be a separate `promotionSelector` instead then it could be.
```const promotionSelector = '.tox-promotion';```

So then you wouldn't need the ```'.' + promotionClass``` two times. But since the `tox-promotion` is also in the approx structure you might want to keep the promotionClass and use that there but maybe that just makes the whole thing harder to read. DRY is fine for the most part but when it comes to test code I tend to like having a bit of duplication just so that you don't need to jump around just to understand a test case.

File: modules/tinymce/src/themes/silver/test/ts/browser/editor/header/PromotionTest.ts [hunk 34]
```diff
@@ -0,0 +1,64 @@
...
+
+  context('promotion turned off', () => {
+    const hook = TinyHooks.bddSetup<Editor>({
+      base_url: '/project/tinymce/js/tinymce',
```
spocke, line 13: Maybe it's better to be specific here and pass in `false` even if you don't need to. Since it would be hard for someone just reading the test and the code in the theme to know that the default is changed inside the `bddSetup` call to `false`.

spocke (approved):

codedmonkey:
Please stop pushing your marketing on my clients.

PR #8035, opened 2022-08-03
TINY-8888: Check parser filter matching in between calling each filter

### LINKED ISSUE #7959: Unsetting attribute serializer nodeFilter causes exception if attribute filter matches attribute
If serializer node filter unsets attribute, attribute filter gets old list, which doesn't apply anymore and error can occure.

https://fiddle.tiny.cloud/ohiaab/1

Open fiddle, open source and insert iframe (e.g. YouTube share).

Expected behaviour:
iframe in editor should have src attribute, but in view source and on save should be transformed in data-src

Actual behaviour:
opening code and saving throws error:
`Uncaught TypeError: uri is undefined`

TheSpyder:
Related Ticket: TINY-8888

Description of Changes:
* The logic that calls filters did not expect that filters could remove things that a later filter was matched on.
* The logged issue happened when a node filter removed the `src` attribute from an `iframe`, our internal `src` attribute filter [received an unexpected `undefined` and triggered an exception](https://github.com/tinymce/tinymce/blob/ece3bf774b9ae20910d073af5351f162c4fb52d5/modules/tinymce/src/core/main/ts/dom/DomSerializerFilters.ts#L40).
* The fix is to check the filter still matches the node array before _every_ filter callback is executed.
* To keep the performance impact to a minimum, this uses careful mutation of the node array rather than copying the contents. Our expectation is that the case where a filter no longer matches is rare, so the majority case should not need to pay the copying penalty of a `filter` that allows all nodes.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

GitHub issues (if applicable): Fixes #7959

### DIFF
--- a/modules/tinymce/src/core/main/ts/html/FilterNode.ts
+++ b/modules/tinymce/src/core/main/ts/html/FilterNode.ts
@@ -82,7 +82,7 @@ const runFilters = (matches: FilterMatches, args: ParserArgs): void => {
 
           // Remove already removed children, and nodes that no longer match the filter
           if (Type.isNullable(node.parent) || filteringAttributes ? node.attr(match.filter.name) === undefined : node.name !== match.filter.name) {
-            nodes.splice(i);
+            nodes.splice(i, 1);
           }
         }
 
--- a/modules/tinymce/src/core/main/ts/html/FilterNode.ts
+++ b/modules/tinymce/src/core/main/ts/html/FilterNode.ts
@@ -81,7 +81,8 @@ const runFilters = (matches: FilterMatches, args: ParserArgs): void => {
           const node = nodes[i];
 
           // Remove already removed children, and nodes that no longer match the filter
-          if (Type.isNullable(node.parent) || filteringAttributes ? node.attr(match.filter.name) === undefined : node.name !== match.filter.name) {
+          const valueMatches = filteringAttributes ? node.attr(match.filter.name) !== undefined : node.name === match.filter.name;
+          if (!valueMatches || Type.isNullable(node.parent)) {
             nodes.splice(i, 1);
           }
         }
--- a/modules/tinymce/src/core/main/ts/html/FilterNode.ts
+++ b/modules/tinymce/src/core/main/ts/html/FilterNode.ts
@@ -72,14 +72,25 @@ const findMatchingNodes = (nodeFilters: ParserFilter[], attributeFilters: Parser
 const runFilters = (matches: FilterMatches, args: ParserArgs): void => {
   const run = (matchRecord: Record<string, FilterMatch>, filteringAttributes: boolean) => {
     Obj.each(matchRecord, (match) => {
-      // Remove already removed children, and nodes that no longer contain the attribute to be matched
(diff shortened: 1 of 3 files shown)


modules/tinymce/src/core/main/ts/html/FilterNode.ts
```diff
@@ -75,19 +75,31 @@ const findMatchingNodes = (nodeFilters: ParserFilter[], attributeFilters: Parser
...
+      const nodes = Arr.from(match.nodes);
 
       Arr.each(match.filter.callbacks, (callback) => {
-        callback(nodes, match.filter.name, args);
+        // very very carefully mutate the nodes array based on whether the filter still matches them
+        for (let i = nodes.length - 1; i >= 0; i--) {
+          const node = nodes[i];
+
+          // Remove already removed children, and nodes that no longer match the filter
+          if (Type.isNullable(node.parent) || filteringAttributes ? node.attr(match.filter.name) === undefined : node.name !== match.filter.name) {
```
lnewson, line 89:
```suggestion
          const valueMatches = filteringAttributes ? node.attr(match.filter.name) !== undefined : node.name === match.filter.name;
          if (!valueMatches || Type.isNullable(node.parent)) {
```
Maybe? Just feels a bit complex having a ternary inside an if

lnewson (approved):

modules/tinymce/src/core/main/ts/html/FilterNode.ts
TheSpyder: Yeah I guess that'll collapse in minification anyway

modules/tinymce/src/core/test/ts/browser/html/DomParserTest.ts
```diff
@@ -386,6 +386,52 @@ describe('browser.tinymce.core.html.DomParserTest', () => {
...
+  it('TINY-8888: mutating addAttributeFilter only removes matching nodes', () => {
+    const parser = DomParser({});
+    parser.addAttributeFilter('src', (nodes) => {
+      nodes[0].attr('src', null);
+    });
+    let ranIdFilter = false;
+    parser.addAttributeFilter('src', () => {
+      ranIdFilter = true;
+    });
+    parser.parse('<b>a<img src="1.gif" />b<img src="1.gif" />c</b>');
```
ltrouton, line 431: Wondering if it would be good to verify the correct node still exists. 
```suggestion
    parser.addAttributeFilter('src', (nodes) => {
      ranIdFilter = true;
      assert.lengthOf(nodes, 1);
      assert.equal(node[0].attr('src'), '2.gif');
    });
    parser.parse('<b>a<img src="1.gif" />b<img src="2.gif" />c</b>');
```

ltrouton (approved):

modules/tinymce/src/core/test/ts/browser/html/DomParserTest.ts
TheSpyder: yeah I guess that makes sense. I'll merge the suggestion then double check it fails correctly.

PR #8036, opened 2022-08-04
TINY-8920: Prevent list actions when in non-editable lists

MitchCTiny:
Related Ticket: TINY-8920

Description of Changes:
* When in non-editable content:
- Content cannot be split into new block by pressing enter (not prevented on shift-enter)?
- Editable list content cannot be merged into non-editable content by delete (but other block content can from start of element)
- List types cannot be toggled or changed, and toolbar buttons are disabled
- Lists cannot be indented
- List properties dialog isw disabled and unopenable

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set
* [x] Docs ticket created (if applicable): DOC-1781 (missing list menu item/commands)

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -29,7 +29,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The content of the `contenteditable="false"` element could be selected with the mouse on Firefox  #TINY-8828
 - Parsing large documents no longer throws a `Maximum call stack size exceeded` exception #TINY-6945
 - DomParser filter matching was not checked between filters, which could lead to an exception in the parser #TINY-8888
-- Lists with `contenteditable="false"` can no longer be toggled, and `contenteditable="true"` list elements within them no longer deleted, added, or indented. #TINY-8920
+- Lists with `contenteditable="false"` can no longer be toggled, and `contenteditable="true"` list elements within them can no longer deleted, added, or indented. #TINY-8920
 
 ### Deprecated
 - The autocompleter `ch` configuration property has been deprecated and will be removed in the next major release. Use the `trigger` property instead. #TINY-8887
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -28,6 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The content of the `contenteditable="false"` element could be selected with the mouse on Firefox  #TINY-8828
 - Parsing large documents no longer throws a `Maximum call stack size exceeded` exception #TINY-6945
 - DomParser filter matching was not checked between filters, which could lead to an exception in the parser #TINY-8888
+- Lists with `contenteditable="false"` can no longer be toggled, and `contenteditable="true"` list elements within them no longer deleted, added, or indented. #TINY-8920
 
 ### Deprecated
 - The autocompleter `ch` configuration property has been deprecated and will be removed in the next major release. Use the `trigger` property instead. #TINY-8887
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -9,7 +9,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - New `text_patterns_lookup` option to provide additional text patterns dynamically #TINY-8778
 - New `format_noneditable_selector` option to specify the `contenteditable="false"` elements that can be wrapped in a format. #TINY-8905
-- New `getContentEditableRoot` function added to the `DOMUtils` API which retrieves the state of `contenteditable` parent until reaching a `false` value or the root #TINY-8920
 
 ### Improved
 - The formatter can now apply a format to a `contenteditable="false"` element by wrapping it. Configurable using the `format_noneditable_selector` option. #TINY-8905
@@ -29,7 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The content of the `contenteditable="false"` element could be selected with the mouse on Firefox  #TINY-8828
 - Parsing large documents no longer throws a `Maximum call stack size exceeded` exception #TINY-6945
 - DomParser filter matching was not checked between filters, which could lead to an exception in the parser #TINY-8888
-- Lists with `contenteditable="false"` can no longer be toggled, and `contenteditable="true"` list elements within them can no longer deleted, added, or indented. #TINY-8920
+- Lists with `contenteditable="false"` can no longer be toggled, and `contenteditable="true"` list elements within them can no longer be deleted or indented. #TINY-8920
 
 ### Deprecated
 - The autocompleter `ch` configuration property has been deprecated and will be removed in the next major release. Use the `trigger` property instead. #TINY-8887
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -28,7 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The content of the `contenteditable="false"` element could be selected with the mouse on Firefox  #TINY-8828
 - Parsing large documents no longer throws a `Maximum call stack size exceeded` exception #TINY-6945
 - DomParser filter matching was not checked between filters, which could lead to an exception in the parser #TINY-8888
-- Lists with `contenteditable="false"` can no longer be toggled, and `contenteditable="true"` list elements within them can no longer be indented, split into another list element, or appended to the previous list element by deletion. #TINY-8920
+- Lists with `contenteditable="false"` can no longer be toggled, and `contenteditable="true"` list elements within them can no longer be indented, split into another list element, or appended to the previous list element by deletion #TINY-8920
 
 ### Deprecated
 - The autocompleter `ch` configuration property has been deprecated and will be removed in the next major release. Use the `trigger` property instead. #TINY-8887
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - New `text_patterns_lookup` option to provide additional text patterns dynamically #TINY-8778
 - New `format_noneditable_selector` option to specify the `contenteditable="false"` elements that can be wrapped in a format. #TINY-8905
+- New `getContentEditableRoot` function added to the `DOMUtils` API which retrieves the state of `contenteditable` parent until reaching a `false` value or the root #TINY-8920
 
 ### Improved
 - The formatter can now apply a format to a `contenteditable="false"` element by wrapping it. Configurable using the `format_noneditable_selector` option. #TINY-8905
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -28,7 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The content of the `contenteditable="false"` element could be selected with the mouse on Firefox  #TINY-8828
 - Parsing large documents no longer throws a `Maximum call stack size exceeded` exception #TINY-6945
 - DomParser filter matching was not checked between filters, which could lead to an exception in the parser #TINY-8888
-- Lists with `contenteditable="false"` can no longer be toggled, and `contenteditable="true"` list elements within them can no longer be deleted or indented. #TINY-8920
+- Lists with `contenteditable="false"` can no longer be toggled, and `contenteditable="true"` list elements within them can no longer be indented, split into another list element, or appended to the previous list element by deletion. #TINY-8920
 
 ### Deprecated
 - The autocompleter `ch` configuration property has been deprecated and will be removed in the next major release. Use the `trigger` property instead. #TINY-8887
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -28,7 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The content of the `contenteditable="false"` element could be selected with the mouse on Firefox  #TINY-8828
 - Parsing large documents no longer throws a `Maximum call stack size exceeded` exception #TINY-6945
 - DomParser filter matching was not checked between filters, which could lead to an exception in the parser #TINY-8888
-- Lists with `contenteditable="false"` can no longer be toggled, and `contenteditable="true"` list elements within them can no longer be indented, split into another list element, or appended to the previous list element by deletion #TINY-8920
+- Lists with `contenteditable="false"` can no longer be toggled, and `contenteditable="true"` list elements within them can no longer be indented, split into another list element, or appended to the previous list element by deletion. #TINY-8920
 
 ### Deprecated
 - The autocompleter `ch` configuration property has been deprecated and will be removed in the next major release. Use the `trigger` property instead. #TINY-8887
--- a/modules/tinymce/src/core/main/ts/api/dom/DOMUtils.ts
+++ b/modules/tinymce/src/core/main/ts/api/dom/DOMUtils.ts
@@ -1124,7 +1124,7 @@ const DOMUtils = (doc: Document, settings: Partial<DOMUtilsSettings> = {}): DOMU
 
   const getContentEditableRoot = (node: Node) => {
     const root = getRoot();
-    let editableRoot: string | null;
+    let editableRoot: string | null = null;
     // Get all parents until we hit a non editable parent or the root
     let parent: Node | null = node;
     while (parent !== root && parent) {
--- a/modules/tinymce/src/core/main/ts/api/dom/DOMUtils.ts
+++ b/modules/tinymce/src/core/main/ts/api/dom/DOMUtils.ts
@@ -1842,9 +1842,34 @@ const DOMUtils = (doc: Document, settings: Partial<DOMUtilsSettings> = {}): DOMU
      */
     dispatch,
 
-    // Returns the content editable state of a node
+    /**
+     * Returns the content editable state of a node.
+     *
+     * @method getContentEditable
+     * @param {Node} node Node to find editable state of.
+     * @return {String} Value of content editable attribute.
+     */
     getContentEditable,
+
+    /**
+     * Returns the content editable state of the first parent of a node with the `contenteditable` attribute.
+     *
+     * @method getContentEditableParent
+     * @param {Node} node Node to find state of editable parent.
+     * @return {String} Value of content editable attribute.
+     */
     getContentEditableParent,
+
+    /**
+     * Returns the state of the `contenteditable` root parent of a node;
+     * `'false'` if any parent has `contenteditable="false"`,
+     * `'true'` if at least one parent `contenteditable="true"` and no `contenteditable="false"`,
+     * or `null` if no parent has the `contenteditable` attribute.
+     *
+     * @method getContentEditableRoot
+     * @param {Node} node Node to find state of editable root.
+     * @return {String} Value of content editable attribute.
+     */
     getContentEditableRoot,
 
     /**
--- a/modules/tinymce/src/core/main/ts/api/dom/DOMUtils.ts
+++ b/modules/tinymce/src/core/main/ts/api/dom/DOMUtils.ts
@@ -1125,18 +1125,17 @@ const DOMUtils = (doc: Document, settings: Partial<DOMUtilsSettings> = {}): DOMU
   const getContentEditableRoot = (node: Node) => {
     const root = getRoot();
     let editableRoot: string | null = null;
-    // Get all parents until we hit a non editable parent or the root
-    let parent: Node | null = node;
-    while (parent !== root && parent) {
-      editableRoot = getContentEditable(parent) ?? editableRoot;
+    for (let tempNode: Node | null = node; tempNode && tempNode !== root; tempNode = tempNode.parentNode) {
+      // editableRoot only updates if tempNode has contenteditable attribute
+      editableRoot = getContentEditable(tempNode) ?? editableRoot;
+
       if (editableRoot === 'false') {
         // node is within contenteditable="false" element, return false
-        return editableRoot;
+        break;
       }
-      parent = parent.parentNode;
     }
-    // return editableRoot unless parent is root
-    return parent !== root ? editableRoot : null;
+
+    return editableRoot;
   };
 
   const destroy = () => {
--- a/modules/tinymce/src/core/main/ts/api/dom/DOMUtils.ts
+++ b/modules/tinymce/src/core/main/ts/api/dom/DOMUtils.ts
@@ -199,7 +199,7 @@ interface DOMUtils {
   dispatch: (target: Node | Window, name: string, evt?: {}) => EventUtils;
   getContentEditable: (node: Node) => string | null;
   getContentEditableParent: (node: Node) => string | null;
-  getContentEditableRoot: (node: Node) => HTMLElement | null;
+  getContentEditableRoot: (node: Node) => string | null;
   destroy: () => void;
   isChildOf: (node: Node, parent: Node) => boolean;
   dumpRng: (r: Range) => string;
@@ -1124,20 +1124,19 @@ const DOMUtils = (doc: Document, settings: Partial<DOMUtilsSettings> = {}): DOMU
 
   const getContentEditableRoot = (node: Node) => {
     const root = getRoot();
-    let editableRoot: HTMLElement | undefined;
+    let editableRoot: string | null;
     // Get all parents until we hit a non editable parent or the root
     let parent: Node | null = node;
     while (parent !== root && parent) {
-      if (getContentEditable(parent) === 'false') {
-        // node is within contenteditable="false" element, return null
-        return null;
-      } else if (getContentEditable(parent) === 'true') {
-        editableRoot = parent as HTMLElement;
+      editableRoot = getContentEditable(parent) ?? editableRoot;
+      if (editableRoot === 'false') {
+        // node is within contenteditable="false" element, return false
+        return editableRoot;
       }
       parent = parent.parentNode;
     }
-    // return the ancestor contenteditable="true" or root
-    return parent !== root && editableRoot ? editableRoot : root;
+    // return editableRoot unless parent is root
+    return parent !== root ? editableRoot : null;
   };
 
   const destroy = () => {
--- a/modules/tinymce/src/core/main/ts/api/dom/DOMUtils.ts
+++ b/modules/tinymce/src/core/main/ts/api/dom/DOMUtils.ts
@@ -199,7 +199,6 @@ interface DOMUtils {
   dispatch: (target: Node | Window, name: string, evt?: {}) => EventUtils;
   getContentEditable: (node: Node) => string | null;
   getContentEditableParent: (node: Node) => string | null;
-  getContentEditableRoot: (node: Node) => string | null;
   destroy: () => void;
   isChildOf: (node: Node, parent: Node) => boolean;
   dumpRng: (r: Range) => string;
@@ -1122,22 +1121,6 @@ const DOMUtils = (doc: Document, settings: Partial<DOMUtilsSettings> = {}): DOMU
     return state;
   };
 
-  const getContentEditableRoot = (node: Node) => {
-    const root = getRoot();
-    let editableRoot: string | null = null;
-    for (let tempNode: Node | null = node; tempNode && tempNode !== root; tempNode = tempNode.parentNode) {
-      // editableRoot only updates if tempNode has contenteditable attribute
-      editableRoot = getContentEditable(tempNode) ?? editableRoot;
-
-      if (editableRoot === 'false') {
-        // node is within contenteditable="false" element, return false
-        break;
-      }
-    }
-
-    return editableRoot;
-  };
-
   const destroy = () => {
     // Unbind all events bound to window/document by editor instance
     if (boundEvents.length > 0) {
@@ -1841,35 +1824,9 @@ const DOMUtils = (doc: Document, settings: Partial<DOMUtilsSettings> = {}): DOMU
      */
     dispatch,
 
-    /**
-     * Returns the content editable state of a node.
-     *
-     * @method getContentEditable
-     * @param {Node} node Node to find editable state of.
-     * @return {String} Value of content editable attribute.
-     */
+    // Returns the content editable state of a node
     getContentEditable,
-
-    /**
-     * Returns the content editable state of the first parent of a node with the `contenteditable` attribute.
-     *
-     * @method getContentEditableParent
-     * @param {Node} node Node to find state of editable parent.
-     * @return {String} Value of content editable attribute.
-     */
     getContentEditableParent,
-
-    /**
-     * Returns the state of the `contenteditable` root parent of a node;
-     * `'false'` if any parent has `contenteditable="false"`,
-     * `'true'` if at least one parent has `contenteditable="true"` and no `contenteditable="false"`,
-     * or `null` if no parent has the `contenteditable` attribute.
-     *
-     * @method getContentEditableRoot
-     * @param {Node} node Node to find state of editable root.
-     * @return {String} Value of content editable attribute.
-     */
-    getContentEditableRoot,
 
     /**
      * Destroys all internal references to the DOM to solve memory leak issues.
--- a/modules/tinymce/src/core/main/ts/api/dom/DOMUtils.ts
+++ b/modules/tinymce/src/core/main/ts/api/dom/DOMUtils.ts
@@ -199,6 +199,7 @@ interface DOMUtils {
   dispatch: (target: Node | Window, name: string, evt?: {}) => EventUtils;
   getContentEditable: (node: Node) => string | null;
   getContentEditableParent: (node: Node) => string | null;
+  getContentEditableRoot: (node: Node) => HTMLElement | null;
   destroy: () => void;
   isChildOf: (node: Node, parent: Node) => boolean;
   dumpRng: (r: Range) => string;
@@ -1121,6 +1122,24 @@ const DOMUtils = (doc: Document, settings: Partial<DOMUtilsSettings> = {}): DOMU
     return state;
   };
 
+  const getContentEditableRoot = (node: Node) => {
+    const root = getRoot();
+    let editableRoot: HTMLElement | undefined;
+    // Get all parents until we hit a non editable parent or the root
+    let parent: Node | null = node;
+    while (parent !== root && parent) {
+      if (getContentEditable(parent) === 'false') {
+        // node is within contenteditable="false" element, return null
+        return null;
+      } else if (getContentEditable(parent) === 'true') {
+        editableRoot = parent as HTMLElement;
+      }
+      parent = parent.parentNode;
+    }
+    // return the ancestor contenteditable="true" or root
+    return parent !== root && editableRoot ? editableRoot : root;
+  };
+
   const destroy = () => {
     // Unbind all events bound to window/document by editor instance
     if (boundEvents.length > 0) {
@@ -1827,6 +1846,7 @@ const DOMUtils = (doc: Document, settings: Partial<DOMUtilsSettings> = {}): DOMU
     // Returns the content editable state of a node
     getContentEditable,
     getContentEditableParent,
+    getContentEditableRoot,
 
     /**
      * Destroys all internal references to the DOM to solve memory leak issues.
--- a/modules/tinymce/src/core/main/ts/api/dom/DOMUtils.ts
+++ b/modules/tinymce/src/core/main/ts/api/dom/DOMUtils.ts
@@ -1862,7 +1862,7 @@ const DOMUtils = (doc: Document, settings: Partial<DOMUtilsSettings> = {}): DOMU
     /**
      * Returns the state of the `contenteditable` root parent of a node;
      * `'false'` if any parent has `contenteditable="false"`,
-     * `'true'` if at least one parent `contenteditable="true"` and no `contenteditable="false"`,
+     * `'true'` if at least one parent has `contenteditable="true"` and no `contenteditable="false"`,
      * or `null` if no parent has the `contenteditable` attribute.
      *
      * @method getContentEditableRoot
--- a/modules/tinymce/src/core/main/ts/newline/InsertBlock.ts
+++ b/modules/tinymce/src/core/main/ts/newline/InsertBlock.ts
@@ -250,7 +250,7 @@ const insert = (editor: Editor, evt?: EditorEvent<KeyboardEvent>): void => {
   const createNewBlock = (name?: string): Element => {
     let node: Node | null = container;
     const textInlineElements = schema.getTextInlineElements();
-
+    const editableRoot = CefUtils.getContentEditableRoot(dom.getRoot(), container);
     let block: Element;
     if (name || parentBlockName === 'TABLE' || parentBlockName === 'HR') {
       block = dom.create(name || newBlockName);
@@ -283,7 +283,7 @@ const insert = (editor: Editor, evt?: EditorEvent<KeyboardEvent>): void => {
             block.appendChild(clonedNode);
           }
         }
-      } while ((node = node.parentNode) && editableRoot !== node);
+      } while ((node = node.parentNode) && node !== editableRoot);
     }
 
     setForcedBlockAttrs(editor, block);
@@ -402,8 +402,6 @@ const insert = (editor: Editor, evt?: EditorEvent<KeyboardEvent>): void => {
   if (dom.getContentEditableRoot(container) === 'false') {
     return;
   }
-  // Get editable root node, normally the body element but sometimes a div or span
-  const editableRoot = CefUtils.getContentEditableRoot(dom.getRoot(), container);
 
   // Wrap the current node and it's sibling in a default block if it's needed.
   // for example this <td>text|<b>text2</b></td> will become this <td><p>text|<b>text2</p></b></td>
--- a/modules/tinymce/src/core/main/ts/newline/InsertBlock.ts
+++ b/modules/tinymce/src/core/main/ts/newline/InsertBlock.ts
@@ -282,7 +282,7 @@ const insert = (editor: Editor, evt?: EditorEvent<KeyboardEvent>): void => {
             block.appendChild(clonedNode);
           }
         }
-      } while ((node = node.parentNode) && editableRoot.forall((o) => o !== node));
+      } while ((node = node.parentNode) && !editableRoot.exists((o) => o === node));
     }
 
     setForcedBlockAttrs(editor, block);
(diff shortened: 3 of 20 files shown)


lnewson:
Moving back to a draft, as talked to Mitch last night and he only logged this because he thought it was needed to get CI to build which I've clarified isn't needed. Anyways, since tests aren't done yet this isn't ready for review.

modules/tinymce/src/core/main/ts/newline/NewLineAction.ts
```diff
@@ -66,8 +66,7 @@ const hasShiftKey = (_editor: Editor, shiftKey: boolean) => {
 const canInsertIntoEditableRoot = (editor: Editor) => {
   const forcedRootBlock = Options.getForcedRootBlock(editor);
   const rootEditable = NewLineUtils.getEditableRoot(editor.dom, editor.selection.getStart());
-
-  return rootEditable && editor.schema.isValidChild(rootEditable.nodeName, forcedRootBlock);
+  return rootEditable.map((rE) => editor.schema.isValidChild(rE.nodeName, forcedRootBlock)).getOr(false);
```
TheSpyder, line 69: Rather than mapping to a boolean, I prefer filter. `x.map(y).getOr(false)` should be equivalent to `x.filter(y).isSome()`

Does that make sense?

modules/tinymce/src/core/main/ts/newline/InsertBlock.ts
```diff
@@ -282,7 +282,7 @@ const insert = (editor: Editor, evt?: EditorEvent<KeyboardEvent>): void => {
             block.appendChild(clonedNode);
           }
         }
-      } while ((node = node.parentNode) && node !== editableRoot);
+      } while ((node = node.parentNode) && !editableRoot.exists((o) => o === node));
```
TheSpyder, line 285: Option comparison used to be on `Optional`, but it was moved to `Optionals`

I believe this can be changed to `!Optionals.is(editableRoot, node)`

modules/tinymce/src/plugins/advlist/main/ts/ui/Buttons.ts
```diff
@@ -43,9 +43,22 @@ const isWithinList = (editor: Editor, e: EditorEvent<NodeChangeEvent>, nodeName:
...
+const isEditableSelection = (editor: Editor, e: EditorEvent<NodeChangeEvent>): boolean => {
+  const root = editor.getBody();
+  let parent: HTMLElement | null = e.element as HTMLElement;
+  while (parent !== root && parent) {
+    if (editor.dom.getContentEditable(parent) === 'false') {
+      return false;
+    }
+    parent = parent.parentElement;
+  }
+  return true;
```
TheSpyder, line 55: This smells like `PredicateExists.ancestor()` but I'm not sure swapping it to use sugar nodes is worth the slight reduction in code. It is a _very_ common pattern, though.

There's sugar `ContentEditable.isEditable`, `editor.dom.getContentEditableParent`, probably others I'm forgetting, surely one of them is usable for this 🤔

modules/tinymce/src/plugins/lists/main/ts/core/Util.ts
```diff
@@ -1,25 +1,55 @@
...
+export const isEditableSelection = (editor: Editor, node: Element): boolean => {
+  const root = editor.getBody();
+  let parent: HTMLElement | null = node as HTMLElement;
+  while (parent !== root && parent) {
+    if (editor.dom.getContentEditable(parent) === 'false') {
+      return false;
+    }
+    parent = parent.parentElement;
+  }
+  return true;
```
TheSpyder, line 54: Duplicate with `advlist` `Buttons.ts`, hopefully we can find another method to use in both places

modules/tinymce/src/plugins/lists/main/ts/core/Util.ts
```diff
@@ -1,25 +1,55 @@
...
-  editor.on('NodeChange', nodeChangeHandler);
+const getInitial = (editor: Editor): NodeChangeEvent => {
+  const initialNode = editor.selection.getNode();
+  return {
+    parents: editor.dom.getParents(initialNode),
+    element: initialNode
+  };
+};
 
+const setNodeChangeHandler = (editor: Editor, nodeChangeHandler: (e: NodeChangeEvent) => void, initial: NodeChangeEvent): () => void => {
```
TheSpyder, line 33: Since the way to get `initial` is the same for both, why not just call it inside this function?

Our style tends to be defining functions before they are used, so I'd appreciate if you could reorder your changes in this module

modules/tinymce/src/plugins/lists/test/ts/webdriver/ContentEditableFalseTest.ts
```diff
@@ -0,0 +1,96 @@
+import { RealKeys } from '@ephox/agar';
+import { describe, it } from '@ephox/bedrock-client';
+import { TinyAssertions, TinyHooks, TinySelections } from '@ephox/wrap-mcagar';
+
+import Editor from 'tinymce/core/api/Editor';
+import Plugin from 'tinymce/plugins/lists/Plugin';
+
+describe('webdriver.tinymce.plugins.lists.BackspaceDeleteTest', () => {
```
TheSpyder, line 8: This is doing more than just backspace and delete (I guess the file moved too).

I don't see any test for the toolbar button list actions?

TheSpyder (changes requested):

modules/tinymce/src/plugins/advlist/main/ts/ui/Buttons.ts
MitchCTiny: Yea that was the takeaway from @lnewson where this is reused within the lists plugin its only used once in Advlist. This function check all contenteditable parents and returns false if there is one false element, not only finding the first parent. This is because getContentEditableParent will be true when editing a contenteditable true list element which is within a contenteditable false list.

modules/tinymce/src/plugins/lists/main/ts/core/Util.ts
MitchCTiny: I'm wondering if importing the getContentEditableRoot method and using the optional.isNone() as a false flag in these cases would be a better option? Though that would require moving that method elsewhere more suitable as well.

modules/tinymce/src/plugins/lists/main/ts/core/Util.ts
MitchCTiny: Will do.

modules/tinymce/src/plugins/lists/test/ts/webdriver/ContentEditableFalseTest.ts
MitchCTiny: Missed when writing the test, I copied the content.

modules/tinymce/src/core/main/ts/newline/NewLineAction.ts
MitchCTiny: Yep, fixing now.

modules/tinymce/src/core/main/ts/newline/NewLineUtils.ts
```diff
@@ -84,21 +84,21 @@ const moveToCaretPosition = (editor: Editor, root: Node): void => {
...
+    } else if (dom.getContentEditable(parent) === 'true') {
       editableRoot = parent as HTMLElement;
     }
 
     parent = parent.parentNode;
   }
-
-  return parent !== root && editableRoot ? editableRoot : root;
+  return Optional.some(parent !== root && editableRoot ? editableRoot : root);
 };
```
MitchCTiny, line 102: @TheSpyder Where would be a good place to move this in the core api so that it can be referenced in Lists and Advlist?

MitchCTiny: Moved to DOMUtils

modules/tinymce/src/core/main/ts/api/dom/DOMUtils.ts
```diff
@@ -199,6 +199,7 @@ interface DOMUtils {
   dispatch: (target: Node | Window, name: string, evt?: {}) => EventUtils;
   getContentEditable: (node: Node) => string | null;
   getContentEditableParent: (node: Node) => string | null;
+  getContentEditableRoot: (node: Node) => HTMLElement | null;
```
lnewson, line 202: If you're adding a new public API then you **must** add JSDoc (moxiedoc) content so it shows in the API docs and you need a changelog entry as well.

Additionally the implementation below really does not make sense given this function name... the contenteditable root should be the one that is either `false` or `true` but your logic below only returns an element for `true`. It's also not likely going to be useful to as many if it only returns an element for a true context, so I'd suggest making it return an element in both cases and then change the code where it's being used to check if it's `true` or `false`.

- [ ] Rethink this API and make sure to add the appropriate docs/changelog entries if you keep it

modules/tinymce/src/core/main/ts/api/dom/DOMUtils.ts
```diff
@@ -1121,6 +1122,24 @@ const DOMUtils = (doc: Document, settings: Partial<DOMUtilsSettings> = {}): DOMU
...
+      if (getContentEditable(parent) === 'false') {
+        // node is within contenteditable="false" element, return null
+        return null;
+      } else if (getContentEditable(parent) === 'true') {
+        editableRoot = parent as HTMLElement;
+      }
+      parent = parent.parentNode;
+    }
+    // return the ancestor contenteditable="true" or root
+    return parent !== root && editableRoot ? editableRoot : root;
```
lnewson, line 1140: This here is another issue if you're making this a public API... this doesn't make sense to be returning the root here, because it then would be inconsistent with the existing `getContentEditableParent` which returns `null` if the root is what's matched.

modules/tinymce/src/core/main/ts/api/dom/DOMUtils.ts
MitchCTiny: Will rework this to return the false element as well as you suggest, the original use can be changed so that the method is more suitable for public use. I plan to still return the first non-editable parent element, or furthest editable parent element, as this fits the original purpose the most.

Should I add the JSDoc content for `getContentEditable` and `getContentEditableParent` as well, since they are missing?

modules/tinymce/src/plugins/lists/test/ts/browser/ContentEditableFalseActionsTest.ts
```diff
@@ -0,0 +1,96 @@
...
+    TinyAssertions.assertContent(editor, content);
+    editor.execCommand('InsertOrderedList');
+    TinyAssertions.assertCursor(editor, [ 0, 2, 0 ], 0);
+    TinyAssertions.assertContent(editor, content);
+    editor.execCommand('InsertDefinitionList');
+    TinyAssertions.assertCursor(editor, [ 0, 2, 0 ], 0);
+    TinyAssertions.assertContent(editor, content);
+    editor.execCommand('mceListProps');
+    TinyAssertions.assertCursor(editor, [ 0, 2, 0 ], 0);
+    TinyAssertions.assertContent(editor, content);
```
tiny-ben-tran, line 71: I think we should also test command `mceListUpdate`

modules/tinymce/src/plugins/lists/test/ts/browser/ContentEditableFalseActionsTest.ts
```diff
@@ -0,0 +1,96 @@
...
+    editor.execCommand('InsertUnorderedList');
+    TinyAssertions.assertCursor(editor, [ 0, 2, 0 ], 0);
+    TinyAssertions.assertContent(editor, content);
+    editor.execCommand('InsertOrderedList');
+    TinyAssertions.assertCursor(editor, [ 0, 2, 0 ], 0);
+    TinyAssertions.assertContent(editor, content);
+    editor.execCommand('InsertDefinitionList');
+    TinyAssertions.assertCursor(editor, [ 0, 2, 0 ], 0);
+    TinyAssertions.assertContent(editor, content);
```
tiny-ben-tran, line 94: is `mceListProps` command applicable to UL list?

modules/tinymce/src/plugins/lists/test/ts/webdriver/ContentEditableFalseTest.ts
```diff
@@ -0,0 +1,96 @@
...
+      '<li contenteditable="true">editable</li>\n' +
+      '<li>noneditable</li>\n' +
+      '<li contenteditable="true">editable</li>\n' +
+    '</ol>';
+    editor.setContent(content);
+    TinySelections.setCursor(editor, [ 1, 2, 0 ], 0); // HTML is fake-caret
+    await RealKeys.pSendKeysOn('iframe => body li:last-child', [ RealKeys.combo({ shift: true }, 'tab') ]);
+    TinyAssertions.assertCursor(editor, [ 0, 0, 0 ], 0);
+    TinyAssertions.assertContent(editor, content);
+  });
```
tiny-ben-tran, line 95: Why are tests for UL not included? It might be tedious but i think it is worth it

tiny-ben-tran (commented):
We should add a test with a nested list as `getContentEditableRoot` function is better covered in that case

modules/tinymce/src/plugins/lists/test/ts/webdriver/ContentEditableFalseTest.ts
MitchCTiny: No reason, I will update some of these to use UL

modules/tinymce/src/plugins/lists/test/ts/browser/ContentEditableFalseActionsTest.ts
MitchCTiny: No, that command opens a dialog which allows the choice of a starting number for an OL list.

modules/tinymce/src/plugins/lists/test/ts/browser/ContentEditableFalseActionsTest.ts
MitchCTiny: I'm not sure what the expected behaviour of this would be, since this command is able to change attributes, including `contenteditable`. I am leaning towards it still working on cef content, so I can add a test which confirms that this command still works?

lnewson: The list dialog definitely shouldn't work if the list itself is noneditable. As that would be the same as changing the list type via the advlist buttons for example.

MitchCTiny: Updated to prevent `mceListUpdate` on non-editable lists.

modules/tinymce/src/core/main/ts/api/dom/DOMUtils.ts
```diff
@@ -1121,6 +1122,22 @@ const DOMUtils = (doc: Document, settings: Partial<DOMUtilsSettings> = {}): DOMU
     return state;
   };
 
+  const getContentEditableRoot = (node: Node) => {
```
lnewson, line 1125: Where's the tests for this new public API? Also how is this any different to the above now as we are **not** returning the root anymore? Instead we're returning a true/false string based editable state if anything in the tree is a CEF element which is drastically different to what this used to do and cannot be used as a replacement for the old `NewLineAction` logic at the very least since it'll fail if you do something like this and pass in the `editable` text node (the editable root is the CET span in this case):

```html
<p contenteditable="false">Some <span contenteditable="true">editable</span> content</p>
```

Please have a discussion with either Morgan or Andy, or the wider team on how this public API should work, as it looks like it's just being thrown together without out any real consideration for API design sorry.

modules/tinymce/src/core/main/ts/newline/InsertBlock.ts
```diff
@@ -147,12 +148,12 @@ const setForcedBlockAttrs = (editor: Editor, node: Element) => {
 // Wraps any text nodes or inline elements in the specified forced root block name
 const wrapSelfAndSiblingsInDefaultBlock = (editor: Editor, newBlockName: string, rng: Range, container: Node, offset: number) => {
   const dom = editor.dom;
-  const editableRoot = NewLineUtils.getEditableRoot(dom, container);
+  const editableRoot = CefUtils.getContentEditableRoot(dom.getRoot(), container) as Element;
```
lnewson, line 151: This cast should definitely not be done... this could potentially be nullable as the types mention so please don't pretend it's not.

modules/tinymce/src/core/main/ts/newline/InsertBlock.ts
```diff
@@ -249,7 +250,7 @@ const insert = (editor: Editor, evt?: EditorEvent<KeyboardEvent>): void => {
   const createNewBlock = (name?: string): Element => {
     let node: Node | null = container;
     const textInlineElements = schema.getTextInlineElements();
-
```
lnewson, line 252: This blank separator line should not have been removed.

modules/tinymce/src/core/main/ts/newline/InsertBlock.ts
lnewson: In this case it looks like we shouldn't even be getting this again here and we instead should be passing it down from where it's retrieved and checked in `insert`.

modules/tinymce/src/plugins/lists/main/ts/actions/Indendation.ts
```diff
@@ -9,9 +9,12 @@ import * as Selection from '../core/Selection';
 import { Indentation } from '../listmodel/Indentation';
 import { listIndentation } from '../listmodel/ListsIndendation';
 
+const getContentEditableElements = (elements: HTMLElement[]): SugarElement<HTMLElement>[] =>
```
lnewson, line 12: This name is rather confusing as it sounds like it's just looking for anything with `contenteditable` as an attribute... I'd suggest naming this `getEditableElements` instead since that's what you're looking for here.

modules/tinymce/CHANGELOG.md
```diff
@@ -28,6 +29,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The content of the `contenteditable="false"` element could be selected with the mouse on Firefox  #TINY-8828
 - Parsing large documents no longer throws a `Maximum call stack size exceeded` exception #TINY-6945
 - DomParser filter matching was not checked between filters, which could lead to an exception in the parser #TINY-8888
+- Lists with `contenteditable="false"` can no longer be toggled, and `contenteditable="true"` list elements within them can no longer deleted, added, or indented. #TINY-8920
```
lnewson, line 32:
```suggestion
- Lists with `contenteditable="false"` can no longer be toggled, and `contenteditable="true"` list elements within them can no longer be deleted, added, or indented. #TINY-8920
```

modules/tinymce/src/plugins/lists/test/ts/browser/ContentEditableFalseActionsTest.ts
```diff
@@ -0,0 +1,102 @@
...
+  it('TINY-8920: List toolbar buttons are disabled when in noneditable OL list', () => {
+    const editor = hook.editor();
+    const content = '<ol contenteditable="false">\n' +
+      '<li contenteditable="true">editable</li>\n' +
+      '<li>noneditable</li>\n' +
+      '<li contenteditable="true">editable</li>\n' +
+    '</ol>';
+    editor.setContent(content);
+    TinySelections.setCursor(editor, [ 1, 2, 0 ], 0);
+    TinyUiActions.clickOnToolbar(editor, 'button[aria-label="Numbered list"]');
```
lnewson, line 23: Why are you trying to click on something that's supposed to be disabled? We shouldn't be doing that as it's not actually checking it's disabled, all you're doing here is checking that the button does nothing which isn't the same thing.

modules/tinymce/src/plugins/lists/test/ts/browser/ContentEditableFalseActionsTest.ts
```diff
@@ -0,0 +1,102 @@
...
+    TinyAssertions.assertCursor(editor, [ 0, 2, 0 ], 0);
+    TinyAssertions.assertContent(editor, content);
+    TinyUiActions.clickOnToolbar(editor, 'button[aria-label="Bullet list"]');
```
lnewson, line 26: Checking numbered list vs bullet list should be 2 separate tests, as one could fail and then it's going to be hard to diagnose which one failed.

modules/tinymce/src/plugins/advlist/main/ts/ui/Buttons.ts
```diff
@@ -46,6 +46,7 @@ const isWithinList = (editor: Editor, e: EditorEvent<NodeChangeEvent>, nodeName:
 const makeSetupHandler = (editor: Editor, nodeName: ListType) => (api: Toolbar.ToolbarSplitButtonInstanceApi | Toolbar.ToolbarToggleButtonInstanceApi) => {
   const nodeChangeHandler = (e: EditorEvent<NodeChangeEvent>) => {
     api.setActive(isWithinList(editor, e, nodeName));
+    api.setEnabled(editor.dom.getContentEditableRoot(e.element) !== 'false');
```
lnewson, line 49: These changes also need tests.

modules/tinymce/src/plugins/lists/test/ts/browser/ContentEditableFalseActionsTest.ts
MitchCTiny: Since they both use the same `NodeChangeEvent` to disable them in `contenteditable="false"` contexts, if one fails (by changing the content) they would both fail.

modules/tinymce/src/plugins/lists/test/ts/browser/ContentEditableFalseActionsTest.ts
MitchCTiny: How would I check that the button is disabled?

modules/tinymce/src/plugins/lists/test/ts/browser/ContentEditableFalseActionsTest.ts
lnewson: That's not always going to be the case though, someone in the next PR could change that and then the tests wouldn't cover that...

modules/tinymce/src/plugins/lists/main/ts/actions/Indendation.ts
MitchCTiny: Sure thing

modules/tinymce/src/core/main/ts/newline/InsertBlock.ts
MitchCTiny: My mistake

modules/tinymce/src/core/main/ts/newline/InsertBlock.ts
MitchCTiny: Would a better solution be adding the original `NewLineUtils.getEditableRoot` function to CefUtils?

modules/tinymce/src/core/main/ts/api/dom/DOMUtils.ts
MitchCTiny: The other solution was to have three different instances of the same function, in `NewLineUtils`, lists `Utils`, and Advlist `Utils` respectively. It was discussed that this would be better as a function in the `DOMUtils` API, however if you have a better solution I would appreciate it.

MitchCTiny: Function has been removed from DOMUtils API and changes have been reverted.

modules/tinymce/src/core/main/ts/newline/InsertBlock.ts
MitchCTiny: Change has been reverted

modules/tinymce/src/plugins/advlist/test/ts/browser/ContentEditableFalseTest.ts
```diff
@@ -0,0 +1,99 @@
...
+    nonEditableList,
+    divNestedNonEditableList
+  ]);
+
+  const listActions: ListAction[] = [
+    { title: 'Numbered list toolbar button', action: (editor: Editor) => TinyUiActions.clickOnToolbar(editor, '[aria-label="Numbered list"] > .tox-tbtn') },
+    { title: 'Bullet list toolbar button', action: (editor: Editor) => TinyUiActions.clickOnToolbar(editor, '[aria-label="Bullet list"] > .tox-tbtn') }
+  ];
+
+  Arr.each(listActions, (listAction: ListAction) =>
```
lnewson, line 85:
```suggestion
  Arr.each(listActions, (listAction) =>
```
This shouldn't be needed as it can be inferred.

modules/tinymce/src/plugins/advlist/test/ts/browser/ContentEditableFalseTest.ts
```diff
@@ -0,0 +1,99 @@
...
+  readonly content: string;
+  readonly startPath: number[];
+}
+
+interface ListAction {
+  readonly title: string;
+  readonly action: (editor: Editor) => any;
+}
+
+describe('browser.tinymce.plugins.lists.ContentEditableFalseActionsTest', () => {
```
lnewson, line 25:
```suggestion
describe('browser.tinymce.plugins.advlist.ContentEditableFalseActionsTest', () => {
```
This isn't a `list` test so the name should be this.

modules/tinymce/src/plugins/advlist/test/ts/browser/ContentEditableFalseTest.ts
```diff
@@ -0,0 +1,99 @@
...
+  ]);
+
+  const listActions: ListAction[] = [
+    { title: 'Numbered list toolbar button', action: (editor: Editor) => TinyUiActions.clickOnToolbar(editor, '[aria-label="Numbered list"] > .tox-tbtn') },
+    { title: 'Bullet list toolbar button', action: (editor: Editor) => TinyUiActions.clickOnToolbar(editor, '[aria-label="Bullet list"] > .tox-tbtn') }
+  ];
+
+  Arr.each(listActions, (listAction: ListAction) =>
+    context(listAction.title, () =>
+      Arr.each(contentCombinations, (listContent: ListContents) =>
```
lnewson, line 87:
```suggestion
      Arr.each(contentCombinations, (listContent) =>
```

modules/tinymce/src/plugins/lists/test/ts/webdriver/ContentEditableFalseTest.ts
```diff
@@ -0,0 +1,110 @@
...
+  );
+
+  const contentCombinations: ListParameters[] = Arr.flatten([
+    nonEditableList,
+    divNestedNonEditableList,
+    nestedNonEditableList
+  ]);
+
+  const listActions: ListAction[] = [
+    // TODO include 'enter' key here
```
lnewson, line 91: Is this still valid? If so it should either be done as part of this PR or a Jira should be logged and the comment should be updated to reference the Jira.

modules/tinymce/src/plugins/lists/test/ts/webdriver/ContentEditableFalseTest.ts
```diff
@@ -0,0 +1,110 @@
...
+    { title: 'backspace key', keyPress: [ RealKeys.backspace() ] },
+    { title: 'tab key', keyPress: [ RealKeys.combo({}, 'tab') ] },
+    { title: 'shift and tab keys', keyPress: [ RealKeys.combo({ shift: true }, 'tab') ] },
+  ];
+
+  Arr.each(listActions, (listAction) =>
+    context(listAction.title, () =>
+      Arr.each(contentCombinations, (list) =>
+        it('TINY-8920: Pressing ' + listAction.title + ' is disabled when in ' + list.title, async () => {
```
lnewson, line 100:
```suggestion
        it(`TINY-8920: Pressing ${listAction.title} is disabled when in ${list.title}`, async () => {
```
Template strings are normally preferred instead of doing string concatenation.

modules/tinymce/src/plugins/advlist/main/ts/core/ListUtils.ts
```diff
@@ -20,8 +20,16 @@ const getSelectedStyleType = (editor: Editor): Optional<string> => {
   return Optional.from(style);
 };
 
+// Lists/core/Util.ts - Duplicated in Lists plugin
+const isEditableList = (editor: Editor, element: Element): boolean => {
+  const parentList = editor.dom.getParent(element, 'ol,ul,dl');
+  const editableList = editor.dom.getContentEditable(parentList ?? element) === 'true';
+  return editableList || editor.dom.getContentEditableParent(parentList ?? element) !== 'false';
```
lnewson, line 27: You've got code duplicate here... Also why are we checking if it is both true and not false here? That doesn't make a lot of sense here... I think this could be simplified down to just this:
```suggestion
  return editor.dom.getContentEditableParent(parentList ?? element) !== 'false';
```

modules/tinymce/src/plugins/lists/test/ts/webdriver/ContentEditableFalseTest.ts
MitchCTiny: I was about to start a conversation about that, should I use the `isEditableList` function in `InsertBlock` to prevent list item splitting as well? I was thinking of asking other team members since the discussion last week.

modules/tinymce/src/plugins/advlist/main/ts/core/ListUtils.ts
lnewson: Ahh nvm sorry Mitch, I missed one does `getContentEditableParent` and the other just `getContentEditable`. In that case we should just clean up the duplication... So maybe something like this:
```
const isEditableList = (dom: DOMUtils, element: Element): boolean => {
  const parentList =dom.getParent(element, 'ol,ul,dl') ?? element;
  const editableList = dom.getContentEditable(parentList) === 'true';
  return editableList || dom.getContentEditableParent(parentList) !== 'false';
}
```

MitchCTiny: Yep, that's how it was originally. Forgot to revert back when I was trying to fix the Advlist `api.setEnabled` implementation.

modules/tinymce/src/plugins/lists/main/ts/core/Util.ts
```diff
@@ -1,25 +1,47 @@
...
-  const nodeChangeHandler = (e: { parents: Node[] }) => {
-    const inList = Arr.findUntil(e.parents, NodeType.isListNode, NodeType.isTableCellNode)
-      .filter((list: HTMLElement) => list.nodeName === listName && !isCustomList(list))
-      .isSome();
-    activate(inList);
-  };
+const inList = (parents: Node[], listName: string): boolean =>
+  Arr.findUntil(parents, NodeType.isListNode, NodeType.isTableCellNode)
+    .filter((list: HTMLElement) => list.nodeName === listName && !isCustomList(list))
+    .isSome();
```
lnewson, line 15:
```suggestion
    .exists((list) => list.nodeName === listName && !isCustomList(list));
```
I know this was like this previously, but doing a `filter` and `isSome` is the same as just doing `exists`. Feel free to ignore this though, but I just thought while we're changing this it might be worth it to fix this.

modules/tinymce/src/plugins/lists/main/ts/core/Util.ts
```diff
@@ -1,25 +1,47 @@
+import { Toolbar, Menu } from '@ephox/bridge';
```
lnewson, line 1: Bridge should never be used directly outside of `core` and `silver`. You should use the `Ui` module from TinyMCE core instead, for example:
```ts
import { Menu, Toolbar } from 'tinymce/core/api/ui/Ui';
```

modules/tinymce/src/plugins/lists/main/ts/core/Util.ts
MitchCTiny: Will do

modules/tinymce/CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The content of the `contenteditable="false"` element could be selected with the mouse on Firefox  #TINY-8828
 - Parsing large documents no longer throws a `Maximum call stack size exceeded` exception #TINY-6945
 - DomParser filter matching was not checked between filters, which could lead to an exception in the parser #TINY-8888
+- Lists with `contenteditable="false"` can no longer be toggled, and `contenteditable="true"` list elements within them can no longer be indented, split into another list element, or appended to the previous list element by deletion. #TINY-8920
```
spocke, line 31: Should this include the `.` other items doesn't seem to have that except the Deprecated one. Also the log item is rather long not sure if we care anymore. I think the details should go into release notes instead. This would more be an overview that lists now works better with contenteditable=false.

modules/tinymce/src/plugins/advlist/main/ts/ui/Buttons.ts
```diff
@@ -46,6 +46,7 @@ const isWithinList = (editor: Editor, e: EditorEvent<NodeChangeEvent>, nodeName:
 const makeSetupHandler = (editor: Editor, nodeName: ListType) => (api: Toolbar.ToolbarSplitButtonInstanceApi | Toolbar.ToolbarToggleButtonInstanceApi) => {
   const nodeChangeHandler = (e: EditorEvent<NodeChangeEvent>) => {
     api.setActive(isWithinList(editor, e, nodeName));
+    api.setEnabled(ListUtils.isEditableList(editor, e.element));
```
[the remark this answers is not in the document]
MitchCTiny: Each operation has also been disabled, this is just to show that the toggle is disabled as well.

modules/tinymce/src/plugins/advlist/test/ts/browser/ContentEditableFalseTest.ts
```diff
@@ -0,0 +1,100 @@
...
+describe('browser.tinymce.plugins.advlist.ContentEditableFalseTest', () => {
+  const hook = TinyHooks.bddSetupLight<Editor>({
+    plugins: 'advlist lists',
+    advlist_bullet_styles: 'default,lower-alpha,lower-greek,lower-roman,upper-alpha,upper-roman',
+    advlist_number_styles: 'default,circle,square',
+    toolbar: 'numlist bullist',
+    base_url: '/project/tinymce/js/tinymce'
+  }, [ AdvListPlugin, ListsPlugin ], true);
+
+  const styleAttr = ' style="list-style-type: ';
```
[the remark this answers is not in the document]
MitchCTiny: Fixed, same with cefAttr

modules/tinymce/src/plugins/lists/main/ts/core/Util.ts
```diff
@@ -1,25 +1,45 @@
...
   /\btox\-/.test(list.className);
 
-export const listState = (editor: Editor, listName: string, activate: (active: boolean) => void): () => void => {
-  const nodeChangeHandler = (e: { parents: Node[] }) => {
-    const inList = Arr.findUntil(e.parents, NodeType.isListNode, NodeType.isTableCellNode)
-      .filter((list: HTMLElement) => list.nodeName === listName && !isCustomList(list))
-      .isSome();
-    activate(inList);
-  };
+const inList = (parents: Node[], listName: string): boolean =>
```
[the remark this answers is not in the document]
MitchCTiny: Done as described, looks much cleaner thank you

modules/tinymce/src/plugins/lists/test/ts/browser/ContentEditableFalseActionsTest.ts
```diff
@@ -0,0 +1,115 @@
...
+    divNestedNonEditableList,
+    nestedNonEditableList
+  ]);
+
+  const checkToolbarDisabled = (editor: Editor, listType: string) => {
+    UiFinder.exists(SugarBody.body(), `button[aria-label="${listType}"][aria-disabled="true"]`);
+    TinyUiActions.clickOnToolbar(editor, `button[aria-label="${listType}"][aria-disabled="true"]`);
+  };
+
+  const listActions: ListAction[] = [
```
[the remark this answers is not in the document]
MitchCTiny: Done for each new test file, different test `it` for each action, still iterates over the content array however.

modules/tinymce/CHANGELOG.md
MitchCTiny: Fixed, thank you

lnewson: @MitchCTiny what you had was correct actually with the fullstop. Remember @brian-forte had asked us to include it and if you look at all the released versions it does have fullstops. I thought Millie was going to pass that on to the Husky team but I guess that got missed before she left.

Either way, not something we need to enforce but it saves us having to fix it later.

MitchCTiny: Ah right, I remember. Fixed

modules/tinymce/src/plugins/advlist/test/ts/browser/ContentEditableFalseTest.ts
```diff
@@ -0,0 +1,97 @@
...
+    advlist_bullet_styles: 'default,lower-alpha,lower-greek,lower-roman,upper-alpha,upper-roman',
+    advlist_number_styles: 'default,circle,square',
+    toolbar: 'numlist bullist',
+    base_url: '/project/tinymce/js/tinymce'
+  }, [ AdvListPlugin, ListsPlugin ], true);
+
+  const orderedListStyles = [ 'lower-alpha', 'lower-greek', 'lower-roman', 'upper-alpha', 'upper-roman' ];
+  const numberedListStyles = [ 'circle', 'square' ];
+
+  const OlListTypes: ListStyle[] = Arr.bind(orderedListStyles, (style) => [{ type: 'ol', style }]);
```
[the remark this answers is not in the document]
MitchCTiny: Yep, just wasn't sure of the convention with a record, will change now.

modules/tinymce/src/plugins/advlist/test/ts/browser/ContentEditableFalseTest.ts
```diff
@@ -0,0 +1,97 @@
...
+  const UlListTypes: ListStyle[] = Arr.bind(numberedListStyles, (style) => [{ type: 'ul', style }]);
+  const listTypes = Arr.flatten([ OlListTypes, UlListTypes ]);
```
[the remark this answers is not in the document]
MitchCTiny: Ahh, spread is still new to me but it looks suitable here.

modules/tinymce/CHANGELOG.md
spocke: Time to make that changelog linter and add a punctuation rule. :smile:

modules/tinymce/src/core/main/ts/newline/InsertBlock.ts
```diff
@@ -25,6 +25,11 @@ const trimZwsp = (fragment: DocumentFragment) => {
   });
 };
 
+const isEditableList = (editor: Editor, node: Node): boolean => {
+  const parentList = editor.dom.getParent(node, 'ol,ul,dl');
+  return !parentList ? true : editor.dom.getContentEditableParent(parentList) !== 'false';
```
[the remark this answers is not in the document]
spocke: Using `closest` as the name for that would be better. I have always regretted that the `dom.getParent` is not called `dom.getClosest` since that is what it's really doing.

modules/tinymce/src/plugins/advlist/test/ts/browser/ContentEditableFalseTest.ts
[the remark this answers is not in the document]
spocke: I think you can just add parentheses around this.
```
const OlListTypes: ListStyle[] = Arr.bind(orderedListStyles, (style) => ({ type: 'ol', style }));
```

modules/tinymce/src/plugins/lists/test/ts/browser/ContentEditableFalseActionsTest.ts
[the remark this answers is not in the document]
spocke: Yeah, that is fine it's more likely that you want to drill down on a particular command than a particular type of content.

modules/tinymce/src/plugins/advlist/test/ts/browser/ContentEditableFalseTest.ts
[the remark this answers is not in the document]
MitchCTiny: Nice, thanks

modules/tinymce/src/plugins/lists/main/ts/actions/Indendation.ts
```diff
@@ -1,20 +1,25 @@
...
+  const lists = SugarElements.fromDom(Arr.filter(Selection.getSelectedListRoots(editor),
+    (list) => Util.isEditableList(editor, list)));
+  const dlItems = SugarElements.fromDom(Arr.filter(Selection.getSelectedDlItems(editor),
+    (list) => Util.isEditableList(editor, list)));
+
+  const parentList = Selection.getParentList(editor);
   let isHandled = false;
 
-  if (lists.length || dlItems.length) {
+  if (parentList && Util.isEditableList(editor, parentList) && (lists.length || dlItems.length)) {
```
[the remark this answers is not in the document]
MitchCTiny: After our discussion I would appreciate your approval of my/our implementation.

modules/tinymce/src/core/main/ts/newline/InsertBlock.ts
```diff
@@ -25,6 +25,11 @@ const trimZwsp = (fragment: DocumentFragment) => {
   });
 };
 
+const isEditableList = (editor: Editor, node: Node): boolean => {
+  const parentList = editor.dom.getParent(node, 'ol,ul,dl');
```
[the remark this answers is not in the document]
MitchCTiny: I figured the editor might leave them in some cases.

spocke (approved):

modules/tinymce/src/plugins/advlist/test/ts/browser/ContentEditableFalseTest.ts
```diff
@@ -0,0 +1,97 @@
...
+    toolbar: 'numlist bullist',
+    base_url: '/project/tinymce/js/tinymce'
+  }, [ AdvListPlugin, ListsPlugin ], true);
+
+  const orderedListStyles = [ 'lower-alpha', 'lower-greek', 'lower-roman', 'upper-alpha', 'upper-roman' ];
+  const numberedListStyles = [ 'circle', 'square' ];
+
+  const OlListTypes: ListStyle[] = Arr.map(orderedListStyles, (style) => ({ type: 'ol', style }));
+  const UlListTypes: ListStyle[] = Arr.map(numberedListStyles, (style) => ({ type: 'ul', style }));
+  const listTypes = [ ...OlListTypes, ...UlListTypes ];
```
lnewson, line 36:
```suggestion
  const olListTypes: ListStyle[] = Arr.map(orderedListStyles, (style) => ({ type: 'ol', style }));
  const ulListTypes: ListStyle[] = Arr.map(numberedListStyles, (style) => ({ type: 'ul', style }));
  const listTypes = [ ...olListTypes, ...ulListTypes ];
```
These are using the wrong naming convention as we use camel case, not pascal case for variable names.

TheSpyder (approved):

ephox-mogran (approved):

PR #8041, opened 2022-08-08
TINY-8814: Spaces after cef-blocks should now be added and seen as expected

HAFRMO:
Related Ticket: 

Description of Changes:
CEF blocks caused trouble due to cursor movement, similar to spaces before blocks. Additional code needed to handle the problem of there being no NBSP in some of these new situations, thus not triggering the code.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


vpyshnenko (approved):

modules/tinymce/CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed various incorrect types on public APIs found while enabling TypeScript strict mode #TINY-8806
 - Text returned from `editor.getContent({format: 'text'})` would differ with blank lines between some browsers #TINY-8579
 - The editor focused via the `auto_focus` option was not scrolled into the viewport #TINY-8785
+- Adding spaces just after a `contenteditable="false"` block would not work properly in some situations TINY-8814
```
lnewson, line 22:
```suggestion
- Adding spaces just after a `contenteditable="false"` block would not work properly in some situations #TINY-8814
```

modules/tinymce/src/core/main/ts/keyboard/Nbsps.ts
```diff
@@ -149,22 +162,37 @@ const normalizeNbspInMiddleOfTextNode = (node: Text): boolean => {
   }
 };
 
-const normalizeNbspAtEnd = (root: SugarElement<Node>, node: Text): boolean => {
+const normalizeNbspAtEnd = (root: SugarElement<Node>, node: Text, makeNbsb: boolean): boolean => {
```
lnewson, line 165: Typo in `makeNbsp` here (`Nbsp` vs `Nbsb`).

modules/tinymce/src/core/main/ts/keyboard/Nbsps.ts
```diff
@@ -149,22 +162,37 @@ const normalizeNbspInMiddleOfTextNode = (node: Text): boolean => {
...
 
-const normalizeNbsps = (root: SugarElement<Node>, pos: CaretPosition): Optional<CaretPosition> => Optional.some(pos).filter(hasNbsp).bind((pos) => {
-  const container = pos.container() as Text;
-  const normalized = normalizeNbspAtStart(root, container) || normalizeNbspInMiddleOfTextNode(container) || normalizeNbspAtEnd(root, container);
-  return normalized ? Optional.some(pos) : Optional.none();
-});
+const normalizeNbsps = (root: SugarElement<Node>, pos: CaretPosition): Optional<CaretPosition> =>
+  Optional.some(pos).filter(hasNbsp).fold(
+    () => {
+      if (needsToBeNbsp(root, pos) && SugarNode.isText(SugarElement.fromDom(pos.container()))) {
```
lnewson, line 182: This should be using `NodeType.isText()` instead of converting it to a SugarElement just to check this.

HAFRMO: That's much neater.

TheSpyder (approved):

modules/tinymce/src/core/main/ts/keyboard/Nbsps.ts
```diff
@@ -126,13 +136,16 @@ const normalizeNbspMiddle = (text: string): string => {
   }).join('');
 };
 
-const normalizeNbspAtStart = (root: SugarElement<Node>, node: Text): boolean => {
+const normalizeNbspAtStart = (root: SugarElement<Node>, node: Text, makeNbsb: boolean): boolean => {
```
lnewson, line 139: You still have a typo here by the looks... it seems you only fixed some of them instead of all of them.

modules/tinymce/src/core/main/ts/keyboard/Nbsps.ts
```diff
@@ -126,13 +136,16 @@ const normalizeNbspMiddle = (text: string): string => {
...
   const text = node.data;
   const firstPos = CaretPosition(node, 0);
 
-  if (isNbspAt(text, 0) && !needsToBeNbsp(root, firstPos)) {
+  if (isNbspAt(text, 0) && !needsToBeNbsp(root, firstPos) && !makeNbsb) {
```
lnewson, line 143: It's probably worth putting the `makeNbsp` checks at the start of these if statements in this function and in `normalizeNbspAtEnd` below. As remember JavaScript will evaluate from left to right, so in this case it'll evaluate the more costly `needsToBeNbsp` when it's never going to be useful if `makeNbsp` is true anyways.

modules/tinymce/src/core/main/ts/keyboard/Nbsps.ts
```diff
@@ -149,22 +162,37 @@ const normalizeNbspInMiddleOfTextNode = (node: Text): boolean => {
...
-  const normalized = normalizeNbspAtStart(root, container) || normalizeNbspInMiddleOfTextNode(container) || normalizeNbspAtEnd(root, container);
-  return normalized ? Optional.some(pos) : Optional.none();
-});
+const normalizeNbsps = (root: SugarElement<Node>, pos: CaretPosition): Optional<CaretPosition> =>
+  Optional.some(pos).filter(hasNbsp).fold(
+    () => {
+      if (needsToBeNbsp(root, pos) && NodeType.isText(pos.container())) {
+        const container = pos.container() as Text;
+        const normalized = normalizeNbspAtStart(root, container, true) || normalizeNbspAtEnd(root, container, true);
+        return normalized ? Optional.some(pos) : Optional.none();
```
lnewson, line 185: Up to you if you do this, but this could be using the `Optionals.someIf(normalized, pos)` helper here and below to reduce the bundle size (albeit very marginally).

modules/tinymce/src/plugins/nonbreaking/test/ts/webdriver/NonbreakingWrapTypingTest.ts
```diff
@@ -148,8 +148,8 @@ describe('webdriver.tinymce.plugins.nonbreaking.NonbreakingWrapTypingTest', () =
                 s.text(str.is(Unicode.nbsp))
               ]
             }),
-            s.text(str.is(isGecko ? Unicode.zeroWidth + 'test ' : Unicode.zeroWidth + 'test\u00a0'))
-          ].concat(isGecko ? [ s.element('br', {}) ] : [])
+            s.text(str.is(Unicode.zeroWidth + 'test\u00a0'))
```
lnewson, line 151: Is there a reason you've swapped `\u00a0` to `Unicode.nbsp` everywhere else except in this file? It's fine either way, it's just inconsistent 🤷

spocke (approved):

modules/tinymce/src/core/main/ts/keyboard/Nbsps.ts
```diff
@@ -149,22 +162,37 @@ const normalizeNbspInMiddleOfTextNode = (node: Text): boolean => {
...
   }
 };
 
-const normalizeNbsps = (root: SugarElement<Node>, pos: CaretPosition): Optional<CaretPosition> => Optional.some(pos).filter(hasNbsp).bind((pos) => {
-  const container = pos.container() as Text;
-  const normalized = normalizeNbspAtStart(root, container) || normalizeNbspInMiddleOfTextNode(container) || normalizeNbspAtEnd(root, container);
-  return normalized ? Optional.some(pos) : Optional.none();
-});
+const normalizeNbsps = (root: SugarElement<Node>, pos: CaretPosition): Optional<CaretPosition> =>
+  Optional.some(pos).filter(hasNbsp).fold(
```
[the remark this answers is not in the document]
TheSpyder: I agree, there's a place and time for optionals but doing _anything_ with an `Optional.some()` other than returning it or passing it to a function is generally incorrect.

I'm also dubious about this fold implementation, it's using `pos` from the function argument in the `none` branch which is extremely confusing. Once it's unpacked into a basic if/else structure it might be more obvious where to combine branches.

modules/tinymce/src/core/main/ts/keyboard/Nbsps.ts
```diff
@@ -149,22 +162,35 @@ const normalizeNbspInMiddleOfTextNode = (node: Text): boolean => {
...
+    return Optionals.someIf(normalized, pos);
+  } else {
+    if (needsToBeNbsp(root, pos) && NodeType.isText(pos.container())) {
+      const container = pos.container() as Text;
+      const normalized = normalizeNbspAtStart(root, container, true) || normalizeNbspAtEnd(root, container, true);
+      return Optionals.someIf(normalized, pos);
+    } else {
+      return Optional.none();
+    }
+  }
```
lnewson, line 192: You've got nested ifs here when you don't need them, nor does it really make sense to have them nested.
```suggestion
  } else if (needsToBeNbsp(root, pos) && NodeType.isText(pos.container())) {
    const container = pos.container() as Text;
    const normalized = normalizeNbspAtStart(root, container, true) || normalizeNbspAtEnd(root, container, true);
    return Optionals.someIf(normalized, pos);
  } else {
    return Optional.none();
  }
```

modules/tinymce/src/core/main/ts/keyboard/Nbsps.ts
```diff
@@ -149,22 +162,35 @@ const normalizeNbspInMiddleOfTextNode = (node: Text): boolean => {
...
   }
 };
 
-const normalizeNbsps = (root: SugarElement<Node>, pos: CaretPosition): Optional<CaretPosition> => Optional.some(pos).filter(hasNbsp).bind((pos) => {
-  const container = pos.container() as Text;
-  const normalized = normalizeNbspAtStart(root, container) || normalizeNbspInMiddleOfTextNode(container) || normalizeNbspAtEnd(root, container);
-  return normalized ? Optional.some(pos) : Optional.none();
-});
+const normalizeNbsps = (root: SugarElement<Node>, pos: CaretPosition): Optional<CaretPosition> => {
+  if (hasNbsp(pos)) {
```
lnewson, line 180: Should this also have the `NodeType.isText(pos.container())` check? It seems quite odd we'd have it in one place but not both... especially given we cast it as a text node below.

lnewson: Maybe this needs an early exit function before all this? For example:
```
const container = pos.container();
if (!NodeType.isText(container)) {
  return Optional.none();
}
```
That would also mean you can remove the casts if we do something like that.

modules/tinymce/src/core/main/ts/keyboard/Nbsps.ts
HAFRMO: True, I shorted it down without thinking, that's on me.

modules/tinymce/src/core/main/ts/keyboard/Nbsps.ts
HAFRMO: The `hasNbsp`-function already checks for if it is a text or not, that's why it wasn't included. But yeah, I can add that check as you suggest.

lnewson (approved):

PR #8042, opened 2022-08-08
TINY-8811: Fixed strict type errors in katamari

lnewson:
Related Ticket: TINY-8811

Description of Changes:

This enables strict mode for katamari and katamari-assertions and fixes the various compilation errors. Most of the changes are either in tests or in `Obj` since we cannot index an object using `string` and it needs to be `keyof object`. This also includes the `T extends {}` generic fix that will be needed to upgrade to TS 4.8 when it's released later this month.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/katamari/src/main/ts/ephox/katamari/api/Obj.ts
+++ b/modules/katamari/src/main/ts/ephox/katamari/api/Obj.ts
@@ -54,7 +54,7 @@ const internalFilter = <T extends {}>(obj: T, pred: ObjPredicate<T>, onTrue: Obj
 };
 
 export const bifilter = <T extends {}>(obj: T, pred: ObjPredicate<T>): { t: Record<string, T[keyof T]>; f: Record<string, T[keyof T]> } => {
-  const t: Record<string, T[keyof T]> = {} ;
+  const t: Record<string, T[keyof T]> = {};
   const f: Record<string, T[keyof T]> = {};
   internalFilter(obj, pred, objAcc(t), objAcc(f));
   return { t, f };
--- a/modules/katamari/src/main/ts/ephox/katamari/api/Obj.ts
+++ b/modules/katamari/src/main/ts/ephox/katamari/api/Obj.ts
@@ -3,6 +3,11 @@ import { Eq } from '@ephox/dispute';
 import * as Fun from './Fun';
 import { Optional } from './Optional';
 
+type ObjKeys<T extends {}> = Extract<keyof T, string>;
+type ObjCallback<T extends {}> = (value: T[keyof T], key: ObjKeys<T>) => void;
+type ObjMorphism<T extends {}, R> = (value: T[keyof T], key: ObjKeys<T>) => R;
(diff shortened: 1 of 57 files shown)


modules/katamari/src/main/ts/ephox/katamari/api/Obj.ts
```diff
@@ -37,41 +42,39 @@ export const tupleMap = <R, T>(obj: T, f: (value: T[keyof T], key: string) => {k
...
   r[i] = x;
 };
 
-const internalFilter = <V>(obj: Record<string, V>, pred: (value: V, key: string) => boolean, onTrue: (value: V, key: string) => void, onFalse: (value: V, key: string) => void) => {
-  const r: Record<string, V> = {};
+const internalFilter = <T extends {}>(obj: T, pred: ObjPredicate<T>, onTrue: ObjCallback<T>, onFalse: ObjCallback<T>) => {
   each(obj, (x, i) => {
     (pred(x, i) ? onTrue : onFalse)(x, i);
   });
-  return r;
```
lnewson, line 49: Removed this as it's only ever used internally and the return value is never used... not to mention the return value was always `{}` anyways.

modules/katamari/src/test/ts/atomic/api/struct/AdtTest.ts
```diff
@@ -143,7 +143,7 @@ describe('atomic.katamari.api.struct.AdtTest', () => {
         const branches = Arr.mapToObject(original, () => Fun.identity);
         subject.match(branches);
         return false;
-      } catch (err) {
+      } catch (err: any) {
```
spocke, line 146: I guess there is not much point in spending time to add proper types for the the tests for the Adt stuff since it's so legacy by now. Since wouldn't this be `Error` or something?

spocke (approved):

ltrouton (approved):

modules/katamari/src/test/ts/atomic/api/struct/AdtTest.ts
lnewson: Strict mode requires this to be `any` or `unknown` and then you need to add type guards or casts to handle the error (some details are in https://devblogs.microsoft.com/typescript/announcing-typescript-4-4/#use-unknown-catch-variables). So since it's just in tests I didn't see it as worthwhile to add that 🤷.

lnewson: For reference, this is the error you get if you try to use anything other than `any` or `unknown`:
```
 error TS1196: Catch clause variable type annotation must be 'any' or 'unknown' if specified.
```

spocke: Right makes sense in a language without exception tracking like Java. It could be other things than error instances being thrown.

lnewson:
Merging this as the LGTM alert appears to be a false-positive specific to the TinyMCE tests. At the very least isn't related to this change.

PR #8048, opened 2022-08-10
TINY-8874: Dragging CEF element toward the edges doesn't cause scrolling when the edge of the editor is hidden

hamza0867:
Related Ticket: TINY-8874

Description of Changes:
* Support cases when dragging a CEF element towards an edge of the screen when that edge of the editor is hidden

Pre-checks:
* [x] ~Changelog entry added~ (Done in #7998)

Review:
* [x] Milestone set


modules/tinymce/src/core/test/ts/browser/DragDropOverridesTest.ts
```diff
@@ -8,198 +8,288 @@ import { assert } from 'chai';
...
+      <p id="separator" style="height: 100px"></p>
+      <p class="hidden" style="height: 1500px"></p>
+      <p>CEF can get dragged after this one</p>
+    `);
+      const target = UiFinder.findIn(TinyDom.body(editor), 'p:contains("Draggable CEF")').getOrDie();
+      const dest = UiFinder.findIn(TinyDom.body(editor), 'p.hidden').getOrDie();
+      const initialScrollY = window.scrollY;
+      Mouse.mouseDown(target);
+      Mouse.mouseMoveTo(dest, 0, window.innerHeight - 2); // Move the mouse close to the bottom edge of the editor to trigger scrolling
+      await Waiter.pWait(1000); // Wait a small amount of time to ensure the scrolling happens
```
spocke, line 227: The problem with these types of waits is that they slow down the whole testing. This entire test file now just waits for 10+ seconds. I think it's better to use a waiter here. Wait for the scroll to happen at intervals. Then it will hammer say every 10 ms to check did it scroll. That way it's likely to pass the test after 100ms instead of 1000 this goes for all these tests that has wait for a while better wait for a state to happen. If we really want to test that something happens after 100 ms then it could validate that the time before and after is more than 100 ms in a separate state and test. So use agars `Waiter.pTryUntil` or `Waiter.pTryUntilPredicate`. Maybe even create some form of pWaitForScrollX/pWaitForScrollY functions that does some form of setup, grabs the initial value, and a done. Since there seems to be a lot of repetition here but not sure that makes it easier to read the tests.

modules/katamari/src/test/ts/browser/RepeatableTest.ts
```diff
@@ -0,0 +1,38 @@
...
+
+describe('browser.katamari.RepeatableTests', () => {
+  it('Make a repeatable then clean it', async () => {
+    const intervalId = Singleton.repeatable(100);
+    assert.strictEqual(intervalId.get(), Optional.none());
+    let counter = 0;
+    intervalId.set(() => {
+      counter++;
+    });
+    await delay(250);
```
spocke, line 19: I think this could be a waiter that check two things that the counter is greater that 2 and that the now - start time is more than 200ms that way it would be a bit more robust if the browser would lag and fire way the interval call 50ms to late. These time based tests tend to flake, less now I guess when we don't have to deal with IE anymore but still.

modules/katamari/CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## 9.0.0 - 2022-03-03
 
 ### Added
+- Added a `Repeatable` type to the `Singleton` module
```
spocke, line 15: This should be in a new Unreleased section above the 9.0.0 version and also I think the version.txt needs a bump or the library.

modules/tinymce/src/core/test/ts/browser/DragDropOverridesTest.ts
```diff
@@ -8,198 +8,288 @@ import { assert } from 'chai';
...
+      base_url: '/project/tinymce/js/tinymce',
+      height: 300,
+      width: 300,
+    }, [], true);
+
+    before(async () => {
+      hook.editor().on('dragend', () => {
+        fired.set(true);
+      });
+      await Waiter.pWait(100); // Wait a small amount of time to ensure the events have been bound
```
spocke, line 26: Curious why is this needed?

spocke (changes requested):
I noticed that this feature doesn't quite work right if you try to stop the drag process by dragging up again. It should stop scrolling if you drag towards the bottom edge then drag up from that bottom edge again. It should only scroll when the mouse cursor is within that edge window and it should work a bit like a toggle. So maybe it needs a state.intervalId.clear in the branching of the edges an else if it matches none of them.

modules/tinymce/src/core/test/ts/browser/DragDropOverridesTest.ts
hamza0867: I don't know. It was there already 🤷‍♂️

lnewson: From memory it's because some tests here check the drag/drop blocking behaviour and that's bound in the next event loop after initial runs, however the mcagar ready callback runs when init is fired in most cases. See https://github.com/tinymce/tinymce/blob/develop/modules/tinymce/src/core/main/ts/DragDropOverrides.ts#L408-L411

modules/tinymce/src/core/test/ts/browser/DragDropOverridesTest.ts
```diff
@@ -8,198 +8,288 @@ import { assert } from 'chai';
...
+      const initialScrollX = editor.getWin().scrollX;
+      Mouse.mouseDown(target);
+      Mouse.mouseMoveTo(SugarElement.fromDom(editor.getBody()), 2, 9); // Move the mouse close to the right edge of the editor to trigger scrolling
+      await Waiter.pWait(1500); // Wait a small amount of time to ensure the scrolling happens
+      Mouse.mouseUp(target);
+      assert.isBelow(editor.getWin().scrollX, initialScrollX); // Make sure scrolling happened
+    });
+  });
+
+  context('Tests with autoresize', () => {
```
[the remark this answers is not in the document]
hamza0867: Agreed. Changed!

modules/tinymce/src/core/test/ts/browser/DragDropOverridesTest.ts
hamza0867: Done! I used `Waiter.pTryUntil`

modules/katamari/CHANGELOG.md
hamza0867: Since this is a new type I also bumped the version to `9.1.0-rc`. Is that correct ?

modules/katamari/src/test/ts/browser/RepeatableTest.ts
hamza0867: I added assertions around time. Is that what you meant ?

modules/katamari/CHANGELOG.md
lnewson: No that's not correct... TinyMCE is managed by lerna. I'd suggest reading the contributing docs 😉 https://github.com/tinymce/tinymce/blob/develop/CONTRIBUTING.md#version-management

lnewson: So if something needs a minor/major bump we track that by adding the version change to `versions.txt` as Spocke mentioned. In this case though that's already been done because of previous changes so it can be skipped this time.

hamza0867: And I should rollback the change in `package.json` right ?

lnewson: Yes, as noted in the contributing document the versions in the package.json should never be edited manually.

modules/tinymce/src/core/test/ts/browser/DragDropOverridesTest.ts
```diff
@@ -8,194 +8,299 @@ import { assert } from 'chai';
...
-    editor.getWin().scroll({
-      top: 450
+      editor.getWin().scroll({
+        top: 400,
+        behavior: 'auto'
+      });
+      await Waiter.pWait(1500);
+      const target = UiFinder.findIn(TinyDom.body(editor), 'p:contains("Draggable CEF")').getOrDie();
+      const initialScrollY = editor.getWin().scrollY;
+      const dest = SugarElement.fromDom(editor.getBody());
```
lnewson, line 158: This and other places below should be using the `TinyDom.body(editor)` McAgar helper instead.

modules/katamari/CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
 
+### Added
+- Added a `Repeatable` type to the `Singleton` module
```
lnewson, line 10:
```suggestion
- Added a `Repeatable` API to the `Singleton` module to repeat a function at a fixed interval.
```
Maybe? Just that it's not a `type` so that's misleading.

modules/tinymce/src/core/test/ts/browser/DragDropOverridesTest.ts
hamza0867: Done!

modules/tinymce/src/core/test/ts/browser/DragDropOverridesTest.ts
hamza0867: This conversation should be resolved then ?

spocke: Weird that github shows that as a new line. Thought it was part of this PR. Anyway I think this is resolved. Makes sense that it needs some time to finish binding the events.

TheSpyder (approved):

modules/tinymce/src/core/main/ts/DragDropOverrides.ts
```diff
@@ -187,17 +191,73 @@ const moveGhost = (
...
         state.intervalId.set(() => {
           const currentLeft = win.scrollX;
           scrollLeft(currentLeft);
         });
+        // This basically means that the mouse is close to the bottom edge
+        // of the page (within MouseRange pixels) when the bottom of
+        // the editor is offscreen
+      } else if (outerMouseY + mouseRangeToTriggerScrollOutsideEditor >= window.innerHeight) {
+        const scrollDown = (currentTop: number) => {
+          window.scroll({
```
spocke, line 210: Also I first thought this would be a to simplistic view of scrolling things since in theory you could have any number of parent scrollable containers. However did some experimentation and it seems that native drag scrolling only scrolls the closest view at least on Chrome. So if the point is outside of the parent viewport then that is scrolled if it's outside the inner viewport then that is scrolled. But that is kind of what is implemented just find it a bit strange that you can drag scroll and end up with no scroll at all even if there are scrollbars. But since we try to mimic the native behavior here I think this should be fine.

spocke (approved):

modules/tinymce/src/core/main/ts/DragDropOverrides.ts
```diff
@@ -143,15 +144,18 @@ const moveGhost = (
 
   const clientHeight = contentAreaContainer.clientHeight;
   const clientWidth = contentAreaContainer.clientWidth;
+  const outerMouseY = mouseY + contentAreaContainer.getBoundingClientRect().top;
+  const outerMouseX = mouseX + contentAreaContainer.getBoundingClientRect().left;
 
   state.on((state) => {
```
lnewson, line 150: All this logic is over 250 lines now... that's a pretty red flag that this is doing too much. Is there any way we can break this up into smaller functions?

tiny-james (approved):

PR #8053, opened 2022-08-12
TINY-8775: Reduce spaces in urlinput when they should not be there

HAFRMO:
Related Ticket: 

Description of Changes:
UrlInput would not remove spaces on its own, and it would need to be triggered from the outside. This makes the spaces always be removed.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Selecting a figure with `class="image"` would incorrectly highlight the link toolbar button #TINY-8832
 - Specifying a single, non-default list style for the `advlist_bullet_styles` and `advlist_number_styles` options was not respected. #TINY-8721
 - Fixed various issues that occurred when formatting `contenteditable` elements. #TINY-8905
+- Link url field would sometimes allow users to add spaces when a title was set #TINY-8775
 - The text pattern logic threw an error when there were fragmented text nodes in a paragraph #TINY-8779
 - Dragging a `contentEditable=false` element towards the edges would not cause scrolling #TINY-8874
 - The content of the `contenteditable="false"` element could be selected with the mouse on Firefox  #TINY-8828
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -23,7 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Selecting a figure with `class="image"` would incorrectly highlight the link toolbar button #TINY-8832
 - Specifying a single, non-default list style for the `advlist_bullet_styles` and `advlist_number_styles` options was not respected. #TINY-8721
 - Fixed various issues that occurred when formatting `contenteditable` elements. #TINY-8905
-- Link url field would sometimes allow users to add spaces when a title was set #TINY-8775
+- Spaces could be incorrectly added to urlinput dialog components in certain cases #TINY-8775
 - The text pattern logic threw an error when there were fragmented text nodes in a paragraph #TINY-8779
 - Dragging a `contentEditable=false` element towards the edges would not cause scrolling #TINY-8874
 - The content of the `contenteditable="false"` element could be selected with the mouse on Firefox  #TINY-8828
--- a/modules/tinymce/src/themes/silver/main/ts/ui/dialog/UrlInput.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/ui/dialog/UrlInput.ts
@@ -4,7 +4,7 @@ import {
 } from '@ephox/alloy';
 import { Dialog } from '@ephox/bridge';
 import { Arr, Fun, Future, FutureResult, Id, Optional, Result } from '@ephox/katamari';
-import { Attribute, Traverse } from '@ephox/sugar';
+import { Attribute, Traverse, Value } from '@ephox/sugar';
 
 import { UiFactoryBackstage } from '../../backstage/Backstage';
 import { UiFactoryBackstageForUrlInput } from '../../backstage/UrlInputBackstage';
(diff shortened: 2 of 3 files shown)


modules/tinymce/src/themes/silver/test/ts/webdriver/components/urlinput/UrlInputTest.ts
```diff
@@ -0,0 +1,78 @@
...
+    return component.getSystem().getByDom(
+      SelectorFind.descendant(component.element, 'input').getOrDie('Could not find input')
+    ).getOrDie();
+  };
+
+  const assertTrimSpaces = (inputKeys: any[], expectedOutput: string) =>
+    async () => {
+      const input = getInput();
+
+      await RealKeys.pSendKeysOn('body input', inputKeys);
```
TheSpyder, line 50: is `RealKeys` really the only way to trigger a form input event? That doesn't seem right. The more we rely on real keys the less tests we can run on our dev machines.

modules/tinymce/src/themes/silver/test/ts/webdriver/components/urlinput/UrlInputTest.ts
```diff
@@ -0,0 +1,78 @@
...
+  beforeEach(() => {
+    hook.store().clear();
+    UiControls.setValue(getInput().element, '');
+  });
+
+  context('Trim spaces as expected for an URL', () => {
+    it('TINY-8775: Trim a space of it is the only thing in the input', assertTrimSpaces([ RealKeys.text(' ') ], ''));
+    it('TINY-8775: Trim a space of it is at the start', assertTrimSpaces([ RealKeys.text(' A') ], 'A'));
+    it('TINY-8775: Trim a space of it is at the end', assertTrimSpaces([ RealKeys.text('A ') ], 'A'));
+    it('TINY-8775: Trim all appropriate spaces', assertTrimSpaces([ RealKeys.text(' A B '), RealKeys.combo({}, 'arrowLeft'), RealKeys.text(' ') ], 'A B'));
```
TheSpyder, line 76: Does this really result in `A B` when typing one character at a time on a keyboard? I'm not sure the input filter logic would really allow for that if it trims `"A "` before the `B` is typed. I assume `RealKeys.text()` is just inserting all at once as if it was pasted.

TheSpyder (approved):
Nice clean solution, just a couple of minor concerns :)

modules/tinymce/src/themes/silver/test/ts/webdriver/components/urlinput/UrlInputTest.ts
lnewson: Bedrock uses `elementSendKeys` which will handle them one character at a time as per https://w3c.github.io/webdriver/#dfn-dispatch-the-events-for-a-typeable-string. So typing ` A B ` results in `AB` being in the output instead of it being similar to being pasted and that can be verified by removing the other keys here. So after the state is `AB`, pressing left then moves the caret between `A` and `B` and finally pressing space adds the space in the middle so it ends up as `A B`.

modules/tinymce/src/themes/silver/test/ts/webdriver/components/urlinput/UrlInputTest.ts
lnewson: I had told Hannes to do this from our chat last week... we cannot use any simulated APIs (such as those in Agar) here because the issue is that the browser itself trims all whitespace when using the `value` API on the input element. So if we used them then the input element would never end up with leading/trailing spaces anyways.

modules/tinymce/src/themes/silver/test/ts/webdriver/components/urlinput/UrlInputTest.ts
TheSpyder: ok. thanks for the update.

modules/tinymce/src/themes/silver/test/ts/webdriver/components/urlinput/UrlInputTest.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  beforeEach(() => {
+    hook.store().clear();
+    UiControls.setValue(getInput().element, '');
+  });
+
+  context('Trim spaces as expected for an URL', () => {
+    it('TINY-8775: Trim a space of it is the only thing in the input', assertTrimSpaces([ RealKeys.text(' ') ], ''));
+    it('TINY-8775: Trim a space of it is at the start', assertTrimSpaces([ RealKeys.text(' A') ], 'A'));
+    it('TINY-8775: Trim a space of it is at the end', assertTrimSpaces([ RealKeys.text('A ') ], 'A'));
```
ltrouton, line 75:
```suggestion
    it('TINY-8775: Trim a space if it is the only thing in the input', assertTrimSpaces([ RealKeys.text(' ') ], ''));
    it('TINY-8775: Trim a space if it is at the start', assertTrimSpaces([ RealKeys.text(' A') ], 'A'));
    it('TINY-8775: Trim a space if it is at the end', assertTrimSpaces([ RealKeys.text('A ') ], 'A'));
```

modules/tinymce/CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Selecting a figure with `class="image"` would incorrectly highlight the link toolbar button #TINY-8832
 - Specifying a single, non-default list style for the `advlist_bullet_styles` and `advlist_number_styles` options was not respected. #TINY-8721
 - Fixed various issues that occurred when formatting `contenteditable` elements. #TINY-8905
+- Link url field would sometimes allow users to add spaces when a title was set #TINY-8775
```
ltrouton, line 26: I don't think we typically use the word 'users'. Maybe something like this?
```suggestion
- Spaces could be incorrectly added to the Link plugin url field in certain cases #TINY-8775
```

ltrouton (approved):

modules/tinymce/CHANGELOG.md
lnewson:
```suggestion
- Urlinput dialog components would sometimes incorrectly allow spaces to be added before or after the URL #TINY-8775
```
I think this would be better, as this is about the whole `urlinput` component not just the link plugin specifically.

lnewson: Or alternatively trying to use Luke's wording:
```suggestion
- Spaces could be incorrectly added to urlinput dialog components in certain cases #TINY-8775
```

modules/tinymce/src/themes/silver/main/ts/ui/dialog/UrlInput.ts
```diff
@@ -119,14 +119,20 @@ export const renderUrlInput = (
...
         // We want to get fast feedback for the link dialog, but not sure about others
-        spec.filetype === 'file' ? [
-          AlloyEvents.run(NativeEvents.input(), (comp) => {
-            AlloyTriggers.emitWith(comp, formChangeEvent, { name: spec.name });
-          })
-        ] : [ ],
         [
+          AlloyEvents.run(NativeEvents.input(), (comp) => {
+            const currentValue = Value.get(comp.element);
+            const trimmedValue = currentValue.trim();
```
spocke, line 127: Should we be using Strings.trim from katamari on this or do we not care I guess it's so small and not that common that minification is less of a concern?

spocke (approved):

modules/tinymce/src/themes/silver/main/ts/ui/dialog/UrlInput.ts
TheSpyder: I'm not sure `Strings.trim` is really the best solution anymore. It's using regex? And was probably created when we had to support IE8. It might be _slightly_ smaller when minified, but gzip erases the difference.

spocke: Right yeah there are probably weird unicode white space characters the native trim is better of handling than say a regexp.

PR #8059, opened 2022-08-18
TINY-8997: Fixed urlinput components updating the value when highlighting dropdown values

lnewson:
Related Ticket: TINY-8997

Description of Changes:

This fixes a regression I accidentally introduced when doing the strict types due to the alloy types being incorrect for the `TypeaheadSpec`. As part of this, I also had to add a way to specify a custom gui for the alloy `TestHelpers` because the URL input was never working properly because the `input` element and menu were in different GUI motherships. This then meant that all the logic in https://github.com/tinymce/tinymce/blob/16676bc48409eb440ad1e2bb3021d62c24d975fa/modules/alloy/src/main/ts/ephox/alloy/ui/schema/TypeaheadSchema.ts was never running since the `getByUid` would always return a `Result.error`.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/themes/silver/main/ts/ui/dialog/UrlInput.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/ui/dialog/UrlInput.ts
@@ -146,7 +146,7 @@ export const renderUrlInput = (
     model: {
       getDisplayText: (itemData) => itemData.value,
       selectsOver: false,
-      populateFromBrowser: false
+      populateFromBrowse: false
     },
 
     markers: {
--- a/modules/tinymce/src/themes/silver/test/ts/headless/components/urlinput/UrlInputTest.ts
+++ b/modules/tinymce/src/themes/silver/test/ts/headless/components/urlinput/UrlInputTest.ts
@@ -1,5 +1,5 @@
 import { ApproxStructure, Assertions, Keyboard, Keys, Mouse, UiControls, UiFinder, Waiter } from '@ephox/agar';
-import { AlloyTriggers, Disabling, Focusing, GuiFactory, NativeEvents, Representing, TestHelpers } from '@ephox/alloy';
+import { Disabling, Focusing, GuiFactory, Representing, TestHelpers } from '@ephox/alloy';
 import { beforeEach, describe, it } from '@ephox/bedrock-client';
 import { Future, Optional } from '@ephox/katamari';
 import { SelectorFind, SugarDocument, Value } from '@ephox/sugar';
@@ -35,7 +35,7 @@ describe('headless.tinymce.themes.silver.components.urlinput.UrlInputTest', () =
           {
             type: 'header' as LinkTargetType,
             title: 'Header2',
-            url: '#h_2abefd32',
+            url: '#header2',
             level: 0,
             attach: store.adder('header2.attach')
           }
@@ -49,7 +49,7 @@ describe('headless.tinymce.themes.silver.components.urlinput.UrlInputTest', () =
         return Future.pure({ value: 'http://tiny.cloud', meta: { before: entry.value }, fieldname: 'test' });
       })
     }, Optional.none())
-  ));
+  ), helpers.uiMothership);
 
   TestHelpers.GuiSetup.bddAddStyles(SugarDocument.getDocument(), [
     '.tox-menu { background: white; }',
@@ -75,7 +75,7 @@ describe('headless.tinymce.themes.silver.components.urlinput.UrlInputTest', () =
   const closeMenu = () => {
     const doc = hook.root();
     // Close the menu
-    Keyboard.activeKeydown(doc, Keys.escape());
+    Keyboard.activeKeystroke(doc, Keys.escape());
   };
 
   beforeEach(() => {
@@ -140,8 +140,7 @@ describe('headless.tinymce.themes.silver.components.urlinput.UrlInputTest', () =
     const input = getInput();
 
     await pOpenMenu();
-    UiControls.setValue(input.element, 'He');
-    AlloyTriggers.emit(input, NativeEvents.input());
+    UiControls.setValue(input.element, 'He', 'input');
     await Waiter.pTryUntil(
       'Waiting for the menu to update',
       () => {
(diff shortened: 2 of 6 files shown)


modules/tinymce/src/themes/silver/test/ts/headless/components/urlinput/UrlInputTest.ts
```diff
@@ -35,7 +35,7 @@ describe('headless.tinymce.themes.silver.components.urlinput.UrlInputTest', () =
           {
             type: 'header' as LinkTargetType,
             title: 'Header2',
-            url: '#h_2abefd32',
```
lnewson, line 38: This was quite weird and I think was probably a copy/paste accident at some point 🤷

modules/tinymce/src/themes/silver/test/ts/headless/components/urlinput/UrlInputTest.ts
```diff
@@ -75,7 +75,7 @@ describe('headless.tinymce.themes.silver.components.urlinput.UrlInputTest', () =
   const closeMenu = () => {
     const doc = hook.root();
     // Close the menu
-    Keyboard.activeKeydown(doc, Keys.escape());
+    Keyboard.activeKeystroke(doc, Keys.escape());
```
lnewson, line 78: This stopped working when @HAFRMO changed alloy to listen to `keyup` for the escape key. As such the tests were flaky and only worked because we have retry enabled so this corrects that.

modules/tinymce/src/themes/silver/main/ts/ui/dialog/UrlInput.ts
```diff
@@ -146,7 +146,7 @@ export const renderUrlInput = (
     model: {
       getDisplayText: (itemData) => itemData.value,
       selectsOver: false,
-      populateFromBrowser: false
```
spocke, line 149: Is it really this fragile I thought the specs where validated though boulder? But maybe that just ignores things it doesn't recognize if so should we harden boulder so that unknown things produces a validation error?

spocke (approved):

modules/tinymce/src/themes/silver/main/ts/ui/dialog/UrlInput.ts
lnewson: Unfortunately yes... boulder just completely ignores any key it doesn't know about. As for changing it, that might be a good idea but it also has a lot of potential for causing a heap of failures as I suspect at the moment full specs are passed that would have a lot of unregistered properties.

ephox-mogran: Boulder has a process for this already, but it isn't used by that many things (https://github.com/tinymce/tinymce/blob/310bafec6dbd453be7746cd2ecf5f5c40b18ed1b/modules/boulder/src/main/ts/ephox/boulder/api/StructureSchema.ts#L73) due to a potential performance hit.

ephox-mogran: The top-level schemas use it (https://github.com/tinymce/tinymce/blob/310bafec6dbd453be7746cd2ecf5f5c40b18ed1b/modules/alloy/src/main/ts/ephox/alloy/spec/SpecSchema.ts#L34), but it doesn't inherit that setting all the way down, so you have to specify it every time you are composing a new object of fields.

modules/tinymce/src/themes/silver/test/ts/headless/components/urlinput/UrlInputTest.ts
ephox-mogran: Yeah. Good change. I was keeping it because I thought maybe the intent was to clearly distinguish between the values we wanted and the ones we didn't. But copy/paste sounds just as likely.

modules/tinymce/src/themes/silver/test/ts/headless/components/urlinput/UrlInputTest.ts
```diff
@@ -160,7 +159,7 @@ describe('headless.tinymce.themes.silver.components.urlinput.UrlInputTest', () =
             classes: [ arr.has('tox-collection__group') ],
             children: [
               s.element('div', {
-                classes: [ arr.has('tox-collection__item') ],
+                classes: [ arr.has('tox-collection__item'), arr.has('tox-collection__item--active') ],
```
ephox-mogran, line 162: I'll have to update my changes in-line with these changes, but it might be a good idea to assert that the menu does not exist in pOpenMenu to catch the issue with a menu not being closed earlier.

versions.txt
```diff
@@ -3,5 +3,6 @@
 
 mcagar@8.1.0
```
ephox-mogran, line 4: What is this file for?

ephox-mogran (approved):

versions.txt
lnewson: It's how we track what version to release come 6.2 as we can't manually bump the package.json because of how lerna works.

modules/tinymce/src/themes/silver/main/ts/ui/dialog/UrlInput.ts
lnewson: Worth noting this would likely be much less of a problem/less fragile if we had better boulder types... atm there's a lot if `any`'s in there.

modules/tinymce/src/themes/silver/test/ts/headless/components/urlinput/UrlInputTest.ts
lnewson: Hmm... would we want to assert that though? As the expectation should always be that it's not open. If anything I'd say we should branch to skip opening if it's already open. That said I'm hesitant to make that change though because it seems a little overkill 🤷

lnewson: Actually... on second thoughts I'll make the `closeMenu` function assert it was able to successfully close it. That would make more sense and allows us to keep our expectations.

ltrouton (approved):

Issue #8040, opened 2022-08-04
The newline_behavior attribute is not working as expected in inline editors.


muhammadmohie:

muhammadmohie:
You are right but I tried to replace `newline_behavior: "default"` with `newline_behavior: "invert"` to swap Enter and Shift+Enter behaviors, but with no effect applied.

muhammadmohie:
Any updates? @tiny-james?

TheSpyder:
That does seem like a bug, given that shift+enter works with `default` but enter doesn't work with `invert`. Thanks for the report, we will investigate.

muhammadmohie:
Thanks a lot for your time, waiting for the fix.

micaelpreis:
I'm also the same issue, I'm using a span and the newline_behaviour is also not working as expected there.

TheSpyder:
We've tentatively scheduled the fix for 6.3, due next month, but we haven't confirmed whether it will make it into that release yet.

TheSpyder:
It didn't fit into the 6.3 release and it hasn't been rescheduled yet. Sorry for the lack of updates, I labelled our internal report incorrectly and it wasn't properly connected to this issue.

DGovender:
Hi @TheSpyder, has this fix been scheduled for any release as this issue is still open?

DAdq26:
Hi 

for us when using newline_behavior: 'linebreak' to avoid new paragraph after hiting enter, this option breaks key Enter with list
(with version 7.1.1)

Nor enter nor shift enter allows you to add a new list item
It's a real issue for us

![image](https://github.com/tinymce/tinymce/assets/134490838/9000e38c-b214-4d29-8da4-53d6d0750184)

Best

dikowski:
I'm in the same situation as @DAdq26. newline_behavior: 'linebreak' messes up the ordered and unordered list functionality. This is a problem. Hoping a solution is coming soon.

TheSpyder:
This is getting off topic. The issue is regarding inline editors; please log a new issue if you need to discuss other things.

[...]

PR #8060, opened 2022-08-19
fix: newline behavior for heading and paragraphs tags

atarek12:
Related Ticket: https://github.com/tinymce/tinymce/issues/8040

Description of Changes:
- Enabling the `Enter` key to insert `line-break` for inline editors.
- I have forced the behavior to be `line-break` for the inline editors whenever we have `P` and `h*` tags as root blocks  as we cannot insert the `Block` break inside them.

Pre-checks:
* [x] Changelog entry added

Review:
* [ ] Milestone set

GitHub issues (if applicable): https://github.com/tinymce/tinymce/issues/8040

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -29,6 +29,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The content of the `contenteditable="false"` element could be selected with the mouse on Firefox  #TINY-8828
 - Parsing large documents no longer throws a `Maximum call stack size exceeded` exception #TINY-6945
 - DomParser filter matching was not checked between filters, which could lead to an exception in the parser #TINY-8888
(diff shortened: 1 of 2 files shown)


atarek12:
@TheSpyder @spocke any updates for this PR?

TheSpyder:
Apologies for the lack of follow up on this. We're scheduled to look at it soon. I'm not sure whether your PR is the right approach; I was suspecting something in the `invert` logic might be wrong, or perhaps `ContextSelectors.shouldInsertBr()` should also match the inline root block.

If it is the right approach I'd want to adjust the logic to use an existing defined list of blocks rather than hard coding extra ones after `summary`, but that's not a simple task.
https://github.com/tinymce/tinymce/blob/ece3bf774b9ae20910d073af5351f162c4fb52d5/modules/tinymce/src/core/main/ts/dom/ElementType.ts#L4-L11

PR #8072, opened 2022-08-24
TINY-8988: Adding Merge Tags and Footnotes to the Plugin Lists

ephox-mogran:
Related Ticket: TINY-8988 and TINY-9019

Description of Changes:
* added mergetags and footnotes to the available plugin lists
* sorted the available and installed plugin lists that are displayed

Pre-checks:
* [x] Changelog entry added (for sorting)

Review:
* [X] Milestone set
* [N/A] Docs ticket created (if applicable)

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -21,7 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - The `@menubar-row-separator-color` oxide variable no longer affects the divider between the Menubar and Toolbar and only controls the color of the separator lines drawn in multiline Menubars. #TINY-8632
 - The `@toolbar-separator-color` oxide variable now exclusively affects the color of the separator between the Menubar and Toolbar. #TINY-8632
-- The plugin lists in the Help dialog are no longer translated. #TINY-9019
+- The available premium plugins listed in the Help dialog are no longer translated. #TINY-9019
 
 ### Fixed
 - The Autolink plugin did not work when the text nodes in the content were fragmented #TINY-3723
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -16,10 +16,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The formatter can now apply a format to a `contenteditable="false"` element by wrapping it. Configurable using the `format_noneditable_selector` option. #TINY-8905
 - The autocompleter now supports a multiple character trigger using the new `trigger` configuration #TINY-8887
 - The formatter will now apply some inline formats like color and font size to list item elements when the entire item content is selected. #TINY-8961
+- The plugin lists in the Help dialog are now sorted alphabetically. #TINY-9019
 
 ### Changed
 - The `@menubar-row-separator-color` oxide variable no longer affects the divider between the Menubar and Toolbar and only controls the color of the separator lines drawn in multiline Menubars. #TINY-8632
 - The `@toolbar-separator-color` oxide variable now exclusively affects the color of the separator between the Menubar and Toolbar. #TINY-8632
+- The plugin lists in the Help dialog are no longer translated. #TINY-9019
 
 ### Fixed
 - The Autolink plugin did not work when the text nodes in the content were fragmented #TINY-3723
--- a/modules/tinymce/src/plugins/help/main/ts/ui/PluginsTab.ts
+++ b/modules/tinymce/src/plugins/help/main/ts/ui/PluginsTab.ts
@@ -50,7 +50,6 @@ const tab = (editor: Editor): Dialog.TabSpec & { name: string } => {
   const getPluginData = (editor: Editor, key: string): PluginData => Arr.find(PluginUrls.urls, (x) => {
     return x.key === key;
   }).fold(() => {
-    // This is an unknown plugin, so see if there is any metadata, or fallback to its name
     return identifyUnknownPlugin(editor, key);
   }, (x) => {
     // We know this plugin, so use our stored details.
--- a/modules/tinymce/src/plugins/help/main/ts/ui/PluginsTab.ts
+++ b/modules/tinymce/src/plugins/help/main/ts/ui/PluginsTab.ts
@@ -7,13 +7,24 @@ import I18n from 'tinymce/core/api/util/I18n';
 import * as Options from '../api/Options';
 import * as PluginUrls from '../data/PluginUrls';
 
+interface PluginData {
+  // The name is just used for sorting alphabetically.
+  readonly name: string;
+  readonly html: string;
+}
+
 const tab = (editor: Editor): Dialog.TabSpec & { name: string } => {
   const availablePlugins = () => {
     const premiumPlugins = Arr.filter(PluginUrls.urls, ({ type }) => {
       return type === PluginUrls.PluginType.Premium;
     });
-    const premiumPluginList = Arr.map(premiumPlugins, (plugin) => '<li>' + I18n.translate(plugin.name) + '</li>').join('');
 
+    const sortedPremiumPlugins = Arr.sort(
+      Arr.map(premiumPlugins, (p) => p.name),
+      (s1, s2) => s1.localeCompare(s2)
+    );
+
+    const premiumPluginList = Arr.map(sortedPremiumPlugins, (pluginName) => `<li>${pluginName}</li>`).join('');
     return '<div data-mce-tabstop="1" tabindex="-1">' +
       '<p><b>' + I18n.translate('Premium plugins:') + '</b></p>' +
       '<ul>' +
@@ -26,14 +37,26 @@ const tab = (editor: Editor): Dialog.TabSpec & { name: string } => {
   const makeLink = (p: { name: string; url: string }): string =>
     `<a href="${p.url}" target="_blank" rel="noopener">${p.name}</a>`;
 
-  const maybeUrlize = (editor: Editor, key: string) => Arr.find(PluginUrls.urls, (x) => {
+  const identifyUnknownPlugin = (editor: Editor, key: string): PluginData => {
+    const getMetadata = editor.plugins[key].getMetadata;
+    if (Type.isFunction(getMetadata)) {
+      const metadata = getMetadata();
+      return { name: metadata.name, html: makeLink(metadata) };
+    } else {
+      return { name: key, html: key };
+    }
+  };
+
+  const getPluginData = (editor: Editor, key: string): PluginData => Arr.find(PluginUrls.urls, (x) => {
     return x.key === key;
   }).fold(() => {
-    const getMetadata = editor.plugins[key].getMetadata;
-    return typeof getMetadata === 'function' ? makeLink(getMetadata()) : key;
+    // This is an unknown plugin, so see if there is any metadata, or fallback to its name
+    return identifyUnknownPlugin(editor, key);
   }, (x) => {
+    // We know this plugin, so use our stored details.
     const name = x.type === PluginUrls.PluginType.Premium ? `${x.name}*` : x.name;
-    return makeLink({ name, url: `https://www.tiny.cloud/docs/tinymce/6/${x.slug}/` });
+    const html = makeLink({ name, url: `https://www.tiny.cloud/docs/tinymce/6/${x.slug}/` });
+    return { name, html };
   });
 
   const getPluginKeys = (editor: Editor) => {
@@ -47,8 +70,13 @@ const tab = (editor: Editor): Dialog.TabSpec & { name: string } => {
 
   const pluginLister = (editor: Editor) => {
     const pluginKeys = getPluginKeys(editor);
-    const pluginLis = Arr.map(pluginKeys, (key) => {
-      return '<li>' + maybeUrlize(editor, key) + '</li>';
+    const sortedPluginData = Arr.sort(
+      Arr.map(pluginKeys, (k) => getPluginData(editor, k)),
+      (pd1, pd2) => pd1.name.localeCompare(pd2.name)
+    );
+
+    const pluginLis = Arr.map(sortedPluginData, (key) => {
+      return '<li>' + key.html + '</li>';
     });
     const count = pluginLis.length;
     const pluginsString = pluginLis.join('');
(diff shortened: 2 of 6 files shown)


modules/tinymce/src/plugins/help/main/ts/ui/PluginsTab.ts
```diff
@@ -47,8 +80,13 @@ const tab = (editor: Editor): Dialog.TabSpec & { name: string } => {
 
   const pluginLister = (editor: Editor) => {
     const pluginKeys = getPluginKeys(editor);
-    const pluginLis = Arr.map(pluginKeys, (key) => {
-      return '<li>' + maybeUrlize(editor, key) + '</li>';
+    const sortedPluginData = Arr.sort(
```
lnewson, line 83: This will need a changelog entry won't it as it is a visible change for everyone?

modules/tinymce/src/plugins/help/main/ts/ui/PluginsTab.ts
```diff
@@ -7,13 +7,24 @@ import I18n from 'tinymce/core/api/util/I18n';
 import * as Options from '../api/Options';
 import * as PluginUrls from '../data/PluginUrls';
 
+interface PluginData {
+  // The name is just used for sorting alphabetically.
+  name: string;
+  html: string;
```
lnewson, line 13:
```suggestion
  readonly name: string;
  readonly html: string;
```

ephox-mogran: Sorry. Pointing this out every time should eventually change my muscle memory.

modules/tinymce/src/plugins/help/main/ts/ui/PluginsTab.ts
ephox-mogran: I crossed out changelog because the other change (autocorrect) didn't have an entry, so I just assumed that was how these changes work.

ephox-mogran: But should it have had a changelog entry? Or just because things are now sorted?

lnewson: The premium names generally don't have a changelog since it's not something opensource/non-premium customers would be impacted by. However the sorting impacts everyone, so it should have a changelog.

ephox-mogran: I'll add it for https://ephocks.atlassian.net/browse/TINY-9019 then, as that's probably the ticket I ended up doing here.

modules/tinymce/src/plugins/help/main/ts/ui/PluginsTab.ts
ephox-mogran: Didn't update before rebasing... fixing now.

modules/tinymce/src/plugins/help/test/ts/browser/PluginTabsOrderTest.ts
```diff
@@ -0,0 +1,63 @@
...
+      rawEntries,
+      Arr.sort(rawEntries),
+      `${label} were not sorted alphabetically`
+    );
+
+    // Close the dialog after a successful assertion.
+    Keyboard.activeKeystroke(doc, Keys.escape(), { });
+  };
+
+  it('TINY-8988: Installed Plugin Lists are alphabetically ordered', async () => {
```
ephox-mogran, line 54: I'll change these too.

modules/tinymce/src/plugins/help/main/ts/ui/PluginsTab.ts
```diff
@@ -7,13 +7,24 @@ import I18n from 'tinymce/core/api/util/I18n';
...
+
 const tab = (editor: Editor): Dialog.TabSpec & { name: string } => {
   const availablePlugins = () => {
     const premiumPlugins = Arr.filter(PluginUrls.urls, ({ type }) => {
       return type === PluginUrls.PluginType.Premium;
     });
-    const premiumPluginList = Arr.map(premiumPlugins, (plugin) => '<li>' + I18n.translate(plugin.name) + '</li>').join('');
 
+    // ASSUMPTION: We want the entries to be sorted POST-translation.
+    const sortedPremiumPlugins = Arr.sort(
```
TheSpyder, line 23: We need to use something more advanced than `[].sort()` if we are doing this post translations.

https://stackoverflow.com/a/61094581/7546 looks like it is available on our supported browsers

TheSpyder (changes requested):

modules/tinymce/src/plugins/help/main/ts/ui/PluginsTab.ts
lnewson: `Intl.Collator` is the same as `localeCompare` btw: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare. You are right we do need to have a custom comparator function here though.

ephox-mogran: Is using `localCompare` for the sortBy function sufficient?

ephox-mogran: As in: 

```
     const sortedPremiumPlugins = Arr.sort(
-      Arr.map(premiumPlugins, (p) => I18n.translate(p.name))
+      Arr.map(premiumPlugins, (p) => I18n.translate(p.name)),
+      (s1, s2) => s1.localeCompare(s2)
```

ephox-mogran: Or do I need to detect the locale first and pass it through to `localeCompare` ?

ephox-mogran: If so, is `editor.options.getLanguageCode` going to yield the correct answer? 
Or is `tinymce.I18n.getCode` going to be better? That looks like it sets a cell, and it's going to be consistent with what translate is doing. Yeah, that seems safer. I'll try that.

ephox-mogran: From what I'm seeing, we have a singleton for the entire tinymce variable for the current locale for translation. When I try to load editors in different languages, the last one clobbers the first one. From a quick search, I couldn't find a JIRA for this, and talking to QA, they weren't aware it worked like this. This fiddle demonstrates that if you open a German menubar menu, you get Spanish items.

So I guess that I'll use tinymce.I18n.getCode because that is the assumption that everything else seems to be using. `editor.translate` just seems to delegate to `I18n` which is global.

lnewson: You can't use the `I18n.getCode` because the locale you have to pass to `localeCompare` must be a BCP 47 locale code and that's not what's used by TinyMCE's language codes.

ephox-mogran: Do we have conversion code anywhere? 

I'm happy to say that 9019 **hasn't been** done and just go with basic sorting for now, with a known issue that it doesn't work well for all international languages. It's a step forward to sort at least English, given that we've just been trying to do that manually ? If the code isn't a supported language tag, what happens? Should I identify which language codes aren't using a valid format? 

Thoughts? 

I can revert all sorting code if you'd really prefer, because we aren't handling every case. I disagree with that as an approach (because this is not a step backwards), but I haven't worked on this codebase for a long time, so it's not really my call.

ephox-mogran: Maybe I should just wait until https://ephocks.atlassian.net/browse/TINY-9019?focusedCommentId=120080 is resolved. We might just get able to get rid of the translations.

ephox-mogran: Product has decided that the plugin names shouldn't be translated. Removed translation sorting attempt. Re-requesting review.

spocke: Did you forget to push the sorting is still there?

modules/tinymce/src/plugins/help/main/ts/ui/PluginsTab.ts
```diff
@@ -26,14 +37,36 @@ const tab = (editor: Editor): Dialog.TabSpec & { name: string } => {
...
+      };
+    }
+  };
+
+  const getPluginData = (editor: Editor, key: string): PluginData => Arr.find(PluginUrls.urls, (x) => {
     return x.key === key;
   }).fold(() => {
-    const getMetadata = editor.plugins[key].getMetadata;
-    return typeof getMetadata === 'function' ? makeLink(getMetadata()) : key;
+    // This is an unknown plugin, so see if there is any metadata, or fallback to its name
```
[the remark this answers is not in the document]
ephox-mogran: Ok. I'll delete it.

modules/tinymce/CHANGELOG.md
```diff
@@ -16,10 +16,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The formatter can now apply a format to a `contenteditable="false"` element by wrapping it. Configurable using the `format_noneditable_selector` option. #TINY-8905
 - The autocompleter now supports a multiple character trigger using the new `trigger` configuration #TINY-8887
 - The formatter will now apply some inline formats like color and font size to list item elements when the entire item content is selected. #TINY-8961
+- The plugin lists in the Help dialog are now sorted alphabetically. #TINY-9019
 
 ### Changed
 - The `@menubar-row-separator-color` oxide variable no longer affects the divider between the Menubar and Toolbar and only controls the color of the separator lines drawn in multiline Menubars. #TINY-8632
 - The `@toolbar-separator-color` oxide variable now exclusively affects the color of the separator between the Menubar and Toolbar. #TINY-8632
+- The plugin lists in the Help dialog are no longer translated. #TINY-9019
```
lnewson, line 24: Is this accurate? As only the premium plugin list was translated to my knowledge. Maybe something like this instead?
```suggestion
- The available premium plugins listed in the Help dialog are no longer translated. #TINY-9019
```

modules/tinymce/src/plugins/help/test/ts/browser/PluginTabsOrderTest.ts
```diff
@@ -0,0 +1,63 @@
...
+import { assert } from 'chai';
+
+import Editor from 'tinymce/core/api/Editor';
+import HelpPlugin from 'tinymce/plugins/help/Plugin';
+
+import { selectors } from '../module/Selectors';
+import FakePlugin from '../module/test/FakePlugin';
+import NoMetaFakePlugin from '../module/test/NoMetaFakePlugin';
+
+describe('Browser Test: .PluginTabsOrderTest', () => {
```
lnewson, line 15:
```suggestion
describe('browser.tinymce.plugins.help.PluginTabsOrderTest', () => {
```
TinyMCE tests use this format for the name 🤷

modules/tinymce/src/plugins/help/test/ts/browser/PluginTabsOrderTest.ts
```diff
@@ -0,0 +1,63 @@
...
+    return Arr.map(
+      // We want to exclude the "read-more link at the bottom"
+      SelectorFilter.children(list, `li:not(.${selectors.pluginsTabLists.readMoreClass})`),
+      (x) => TextContent.get(x) ?? ''
+    );
+  };
+
+  const pTestPluginItems = async (label: string, editor: Editor, selector: string): Promise<void> => {
+    TinyUiActions.clickOnToolbar(editor, selectors.toolbarHelpButton);
+    const dialog = await UiFinder.pWaitFor('Could not find help dialog', body, selectors.dialog);
```
lnewson, line 41: This probably should be using the `TinyUiActions.pWaitForUi` helper instead.

modules/tinymce/src/plugins/help/test/ts/browser/PluginTabsOrderTest.ts
```diff
@@ -0,0 +1,63 @@
...
+    Mouse.clickOn(dialog, selectors.pluginsTab);
+    const rawEntries = await pExtractItemsFrom(label, selector);
+    assert.deepEqual(
+      rawEntries,
+      Arr.sort(rawEntries, (s1, s2) => s1.localeCompare(s2)),
+      `${label} were not sorted alphabetically`
+    );
+
+    // Close the dialog after a successful assertion.
+    Keyboard.activeKeystroke(doc, Keys.escape(), { });
```
lnewson, line 51: Likewise this should use the `TinyUiActions.keystroke` helper.

Note: While it doesn't make a huge difference here, the reason we have these helpers is because this logic wouldn't work for shadow dom, etc...

lnewson (approved):

modules/tinymce/CHANGELOG.md
ephox-mogran: Yeah, I thought about putting in "available" plugins but the world available wasn't appearing on the dialog so I just left it out. There's a Premium Plugins heading, though, so that sounds good.

modules/tinymce/src/plugins/help/test/ts/browser/PluginTabsOrderTest.ts
ephox-mogran: The one I used as a basis (Metadata) was the *only* test that didn't do this. I'll update it too.

modules/tinymce/src/plugins/help/test/ts/browser/PluginTabsOrderTest.ts
ephox-mogran: Yep. I can see that now looking at the code. Thanks. I was wondering where all the ShadowDom aware code was.

modules/tinymce/src/plugins/help/test/ts/browser/PluginTabsOrderTest.ts
lnewson: ahh damn thanks for that! We did try to fix all of them a while back but must have missed that.

modules/tinymce/src/plugins/help/test/ts/browser/PluginTabsOrderTest.ts
ephox-mogran: Is it a general rule that any UI interactions shouldn't go through agar directly any more?

lnewson: Not so much shouldn't go through Agar, but more if a `TinyUiActions` or `TinyContentActions` helper exists we should prefer it.

TheSpyder (approved):

tiny-ben-tran (approved):

lnewson (approved):

PR #8076, opened 2022-08-25
TINY-8966: Include CSS for Footnotes plugin

MitchCTiny:
Related Ticket: TINY-8966

Description of Changes:
* Added CSS to affect visuals of `mce-footnote` and `mce-footnotes` elements.

Pre-checks:
* [x] ~Changelog entry added~ N/A

Review:
* [x] Milestone set

### DIFF
--- a/modules/oxide/src/less/theme/content/footnotes/footnotes.less
+++ b/modules/oxide/src/less/theme/content/footnotes/footnotes.less
@@ -1,0 +1,32 @@
+//
+// Footnotes
+//
+
+sup.mce-footnote a {
+  color: black;
+  text-decoration: none;
+}
+
+div.mce-footnotes {
+  hr {
+    margin-left: 0;
+    margin-right: auto;
+    width: 25%;
+  }
+
+  li > a:first-child {
+    color: black;
+    text-decoration: none;
+  }
+}
+
+@media print {
+  div.mce-footnotes {
+    break-inside: avoid;
+    width: 100%;
+
+    li > a:first-child {
+      display: none;
+    }
+  }
+}
--- a/modules/oxide/src/less/theme/content/footnotes/footnotes.less
+++ b/modules/oxide/src/less/theme/content/footnotes/footnotes.less
@@ -2,30 +2,29 @@
 // Footnotes
 //
 
-sup.mce-footnote a {
-  color: black;
-  text-decoration: none;
-}
-
 div.mce-footnotes {
   hr {
-    margin-left: 0;
-    margin-right: auto;
+    margin-inline-end: auto;
+    margin-inline-start: 0;
     width: 25%;
   }
 
-  li > a:first-child {
-    color: black;
+  li > a.mce-footnotes-backlink {
     text-decoration: none;
   }
 }
 
 @media print {
+  sup.mce-footnote a {
+    color: black;
+    text-decoration: none;
+  }
+
   div.mce-footnotes {
     break-inside: avoid;
-    width: 100%;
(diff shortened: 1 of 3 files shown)


modules/oxide/src/less/theme/content/footnotes/footnotes.less
```diff
@@ -0,0 +1,32 @@
...
+    text-decoration: none;
+  }
+}
+
+@media print {
+  div.mce-footnotes {
+    break-inside: avoid;
+    width: 100%;
+
+    li > a:first-child {
```
tiny-ben-tran, line 28: Which element does this hide?

MitchCTiny: The '^' symbol used to navigate back to the footnote in the content.

modules/oxide/src/less/theme/content/footnotes/footnotes.less
```diff
@@ -0,0 +1,32 @@
...
+}
+
+div.mce-footnotes {
+  hr {
+    margin-left: 0;
+    margin-right: auto;
+    width: 25%;
+  }
+
+  li > a:first-child {
```
ephox-mogran, line 17: @ltrouton is there anything that stops users putting in an <a> tag into the footnotes section entry? If not, should we have a class on the `a` so that it is more restrictive about which `a` we are styling / removing?

tiny-ben-tran (approved):

modules/oxide/src/less/theme/content/footnotes/footnotes.less
TheSpyder: I think, similar to @ephox-mogran's comment above, it should have a unique class. Please coordinate with @ltrouton on what the class name should be.

TheSpyder (changes requested):

TheSpyder:
The rest looks fine!

modules/oxide/src/less/theme/content/footnotes/footnotes.less
ltrouton: They can add a link but it can only be added to the editable span which is the second child of the `li`. The backlink anchor `a` tag is noneditable and should always be the first child of the `li`.

To be honest, adding a class to the backlink `a` would probably be helpful and the most robust approach

ephox-mogran: Right. I misread `first-child` again. Very out of practice reading CSS selectors. I thought it was saying the first child of the `a` tag, not choose only the `a` tags that are the first children. But never rely on position like this, if you can control and annotate the DOM elements (as a rule)

MitchCTiny:
@ltrouton @ephox-mogran @TheSpyder Would the class name `mce-footnotes-backlink` suit this specific anchor?

modules/oxide/src/less/theme/content/footnotes/footnotes.less
```diff
@@ -0,0 +1,32 @@
...
+
+sup.mce-footnote a {
+  color: black;
+  text-decoration: none;
+}
+
+div.mce-footnotes {
+  hr {
+    margin-left: 0;
+    margin-right: auto;
```
noxuhax, line 13: Should this be a `margin-inline-start` and `margin-inline-end` to support rtl content as well?

MitchCTiny: Figured it out now, thanks for the tip!

ltrouton (approved):

TheSpyder (approved):

PR #8080, opened 2022-08-26
TINY-8687: added support for aligning media plugin and ephox embed elements

spocke:
Related Ticket: TINY-8687

Description of Changes:
* Adds two new alignment formats for media plugin wrapping elements and for ephox embeded elements (enhanced media plugin)
* Tested a few variants of this one was using classes instead like the figure element another one was trying to have the display be `block` in import/exporting but `table` internally both became messy. So this is the simplest solution and the customer can override this alignment if they want to in one spot with the `formats` config option.
* The case where you select an inline editable element seems to work with other changes we made recently to cef so I just added an extra test for that

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The autocompleter now supports a multiple character trigger using the new `trigger` configuration #TINY-8887
 - The formatter will now apply some inline formats like color and font size to list item elements when the entire item content is selected. #TINY-8961
 - The plugin lists in the Help dialog are now sorted alphabetically. #TINY-9019
+- Alignment can now be applied to more types of embedded media elements. #TINY-8687
 
 ### Changed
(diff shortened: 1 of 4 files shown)


modules/tinymce/CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The autocompleter now supports a multiple character trigger using the new `trigger` configuration #TINY-8887
 - The formatter will now apply some inline formats like color and font size to list item elements when the entire item content is selected. #TINY-8961
 - The plugin lists in the Help dialog are now sorted alphabetically. #TINY-9019
+- Alignment can now be applied to more types of embedded media elements. #TINY-8687
```
spocke, line 20: Not sure about this changelog item. I don't want to add details about enhanced media embed and not make it specific to the media plugin since it's really not. Or should it be split up into one for media plugin elements and one for ephox embedded elements. Suggestions are welcome.

vpyshnenko (approved):

lorenzo-pomili (approved):

modules/tinymce/CHANGELOG.md
TheSpyder: This seems fine!

TheSpyder (approved):

PR #8081, opened 2022-08-26
TINY-9023: added oxide content ui css for mergetags

spocke:
Related Ticket: TINY-9023

Description of Changes:
* Added new mergetags ui css to visualize the merge tags while editing

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/oxide/src/less/theme/content/mergetags/mergetag.less
+++ b/modules/oxide/src/less/theme/content/mergetags/mergetag.less
@@ -1,7 +1,12 @@
+@mergetag-hover-background-color: rgba(64, 153, 255, .1);
 @mergetag-affix-text-color: #3788EC;
-@mergetag-affix-background-color: rgba(64, 153, 255, .1);
+@mergetag-affix-background-color: @mergetag-hover-background-color;
 
 .mce-content-body {
+  .mce-mergetag:hover {
+    background-color: @mergetag-hover-background-color;
+  }
+
   .mce-mergetag-affix {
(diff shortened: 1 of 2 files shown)


spocke:
I wonder if this needs a changelog or not. I don't think we added changelog items before when we added things for the UI for preimum plugins.

vpyshnenko (approved):

TheSpyder (approved):

modules/oxide/src/less/theme/content/mergetags/mergetag.less
```diff
@@ -0,0 +1,14 @@
+@mergetag-hover-background-color: rgba(64, 153, 255, .1);
+@mergetag-affix-text-color: #3788EC;
```
noxuhax, line 2: It would be better to use a variables for color and background so it would adjust with the user’s skin. Since we don’t seem to have an exact match for the color in the design, Paul suggested using the same color scheme as do tabs in dialog popups.

```suggestion
@mergetag-hover-background-color: fade(@color-tint, 10%);
@mergetag-affix-text-color: @color-tint;
```

noxuhax (changes requested):

modules/oxide/src/less/theme/content/mergetags/mergetag.less
spocke: That makes sense to use existing variables as much as we can here.

noxuhax (approved):

PR #8082, opened 2022-08-29
TINY-9053: Exposing more types of TextPatterns 

ephox-mogran:
Related Ticket: TINY-9053

Description of Changes:
* exposed more types through TextPatterns in PublicApi
* created a test that uses just PublicApi to check that managing text pattern lookups are more managable

Questions
- does this need documentation changes? 
- is there a precedent for how verbose our changelog entry is for this? Does it need to list every type exposed? 

Pre-checks:
* [x] Changelog entry added

Review:
* [X] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/api/textpatterns/TextPatterns.ts
+++ b/modules/tinymce/src/core/main/ts/api/textpatterns/TextPatterns.ts
@@ -1,7 +1,14 @@
-import { RawPattern, RawDynamicPatternsLookup, DynamicPatternContext } from '../../textpatterns/core/PatternTypes';
+import { RawPattern, RawDynamicPatternsLookup, DynamicPatternContext, DynamicPatternsLookup, Pattern, BlockCmdPattern, BlockFormatPattern, InlineCmdPattern, InlinePattern, InlineFormatPattern } from '../../textpatterns/core/PatternTypes';
 
 export {
+  Pattern,
   RawPattern,
+  DynamicPatternsLookup,
   RawDynamicPatternsLookup,
-  DynamicPatternContext
+  DynamicPatternContext,
+  BlockCmdPattern,
+  BlockFormatPattern,
+  InlineCmdPattern,
+  InlinePattern,
+  InlineFormatPattern
 };
--- a/modules/tinymce/src/core/main/ts/api/textpatterns/TextPatterns.ts
+++ b/modules/tinymce/src/core/main/ts/api/textpatterns/TextPatterns.ts
@@ -1,4 +1,4 @@
-import { RawPattern, RawDynamicPatternsLookup, DynamicPatternContext, DynamicPatternsLookup, Pattern, BlockCmdPattern, BlockFormatPattern, InlineCmdPattern, InlinePattern, InlineFormatPattern } from '../../textpatterns/core/PatternTypes';
+import { RawPattern, RawDynamicPatternsLookup, DynamicPatternContext, DynamicPatternsLookup, Pattern, BlockCmdPattern, BlockFormatPattern, InlineCmdPattern, InlinePattern, InlineFormatPattern, BlockPattern } from '../../textpatterns/core/PatternTypes';
 
 export {
   Pattern,
@@ -7,6 +7,7 @@ export {
   RawDynamicPatternsLookup,
   DynamicPatternContext,
   BlockCmdPattern,
+  BlockPattern,
   BlockFormatPattern,
   InlineCmdPattern,
   InlinePattern,
--- a/modules/tinymce/src/core/main/ts/textpatterns/core/PatternTypes.ts
+++ b/modules/tinymce/src/core/main/ts/textpatterns/core/PatternTypes.ts
@@ -53,7 +53,7 @@ export type Pattern = InlinePattern | BlockPattern;
 
 export interface DynamicPatternContext {
   readonly text: string; // the string from the start of the block to the cursor
-  readonly block: Element; // the parent block node
+  readonly block: Element; // the parent block element
 }
 
 export type DynamicPatternsLookup = (ctx: DynamicPatternContext) => Pattern[];
--- a/modules/tinymce/src/core/main/ts/textpatterns/core/PatternTypes.ts
+++ b/modules/tinymce/src/core/main/ts/textpatterns/core/PatternTypes.ts
@@ -53,7 +53,7 @@ export type Pattern = InlinePattern | BlockPattern;
 
 export interface DynamicPatternContext {
   readonly text: string; // the string from the start of the block to the cursor
-  readonly block: Node; // the parent block node
+  readonly block: Element; // the parent block node
 }
 
 export type DynamicPatternsLookup = (ctx: DynamicPatternContext) => Pattern[];
--- a/modules/tinymce/src/core/test/ts/browser/textpatterns/TextPatternsPublicApiTest.ts
+++ b/modules/tinymce/src/core/test/ts/browser/textpatterns/TextPatternsPublicApiTest.ts
@@ -12,7 +12,13 @@ describe('browser.tinymce.textpatterns.TextPatternsPublicApiTest', () => {
   const makePlugin = () => {
     PluginManager.add('custom-plugin', (editor) => {
 
-      const replacementLookup: TextPatterns.RawDynamicPatternsLookup = (_ctx: TextPatterns.DynamicPatternContext): TextPatterns.RawPattern[] => {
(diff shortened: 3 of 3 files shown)


modules/tinymce/src/core/main/ts/api/textpatterns/TextPatterns.ts
```diff
@@ -1,7 +1,14 @@
-import { RawPattern, RawDynamicPatternsLookup, DynamicPatternContext } from '../../textpatterns/core/PatternTypes';
+import { RawPattern, RawDynamicPatternsLookup, DynamicPatternContext, DynamicPatternsLookup, Pattern, BlockCmdPattern, BlockFormatPattern, InlineCmdPattern, InlinePattern, InlineFormatPattern } from '../../textpatterns/core/PatternTypes';
 
 export {
+  Pattern,
```
ephox-mogran, line 4: I'm not sure if this is too many. Do we have a general reason to ever suppress types that are part of public API functions?

lnewson:
> does this need documentation changes?

Types in general probably need documentation. For now I'd say no, but see what others say.

> is there a precedent for how verbose our changelog entry is for this? Does it need to list every type exposed?

In this case the bug was never released as it was only added in 6.2 so there should be no changelog entry for this.

modules/tinymce/src/core/main/ts/api/textpatterns/TextPatterns.ts
```diff
@@ -1,7 +1,14 @@
...
   RawPattern,
+  DynamicPatternsLookup,
   RawDynamicPatternsLookup,
-  DynamicPatternContext
+  DynamicPatternContext,
```
lnewson, line 8: The incorrect type on `block` hasn't been fixed in this type by the looks of it. Is that expected?

ephox-mogran: Nope. Completely forgot that part. Got carried away with exposing types. Please hold.

ephox-mogran: That should be added now.

lnewson (approved):

modules/tinymce/src/core/main/ts/api/textpatterns/TextPatterns.ts
```diff
@@ -1,7 +1,14 @@
...
+  BlockCmdPattern,
```
lnewson, line 9: Should `BlockPattern` also be exposed here given we're exposing the `InlinePattern` type which is equivalent.

ephox-mogran: Yeah. I think so for consistency. I'll update.

ephox-mogran: Exposed BlockPattern and tested it was exposed.

modules/tinymce/src/core/main/ts/textpatterns/core/PatternTypes.ts
```diff
@@ -53,7 +53,7 @@ export type Pattern = InlinePattern | BlockPattern;
 
 export interface DynamicPatternContext {
   readonly text: string; // the string from the start of the block to the cursor
-  readonly block: Node; // the parent block node
+  readonly block: Element; // the parent block node
```
spocke, line 56: Maybe change the comment as well to `the parent block element`

modules/tinymce/src/core/test/ts/browser/textpatterns/TextPatternsPublicApiTest.ts
```diff
@@ -0,0 +1,124 @@
+import { describe, it } from '@ephox/bedrock-client';
+import { Arr } from '@ephox/katamari';
+import { TinyAssertions, TinyHooks, TinyUiActions } from '@ephox/wrap-mcagar';
+
+// eslint-disable-next-line @tinymce/no-publicapi-module-imports
+import { Editor, PluginManager, TextPatterns } from 'tinymce/core/api/PublicApi';
+
+// The purpose of this test is to ensure that we have all the types required
+// for managing TextPatterns exposed through PublicApi
+describe('browser.tinymce.textpatterns.TextPatternsPublicApiTest', () => {
```
spocke, line 10: I this the first time we have a test that only has the purpose of testing the types. Should these be a new category of tests or should type testing be else where?

spocke (approved):

modules/tinymce/src/core/test/ts/browser/textpatterns/TextPatternsPublicApiTest.ts
lnewson: It's actually been done before fwiw. Here's the first one I can think of: https://github.com/tinymce/tinymce/blob/b69c943324af5e127fe53f296f8ff1d4df0769b2/modules/katamari/src/test/ts/atomic/api/data/CovarianceAssertionTest.ts.

modules/tinymce/src/core/main/ts/api/textpatterns/TextPatterns.ts
```diff
@@ -1,7 +1,15 @@
-import { RawPattern, RawDynamicPatternsLookup, DynamicPatternContext } from '../../textpatterns/core/PatternTypes';
+import { RawPattern, RawDynamicPatternsLookup, DynamicPatternContext, DynamicPatternsLookup, Pattern, BlockCmdPattern, BlockFormatPattern, InlineCmdPattern, InlinePattern, InlineFormatPattern, BlockPattern } from '../../textpatterns/core/PatternTypes';
```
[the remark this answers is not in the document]
lnewson: We really should look at fixing all those max line length linter issues one day so we can re-enable the rule. Atm it's configured but it's really large (260 vs 160).

modules/tinymce/src/core/test/ts/browser/textpatterns/TextPatternsPublicApiTest.ts
ephox-mogran: I think testing the presence of types on the PublicApi in lots of places sounds like a good way to make sure we don't accidentally break something. Especially when some of the exports from PublicApi are "all the exports in this other module"

modules/tinymce/src/core/main/ts/api/textpatterns/TextPatterns.ts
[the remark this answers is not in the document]
ephox-mogran: What is the format for multilines. Is it 

```
import { ..., ...., ..., 
  ..., ..., } from '..'
```

or

```
import {
  ...., ...., ..., ....,
  ...., ...., ...., ....
} from '...'

or something else ?

modules/tinymce/src/core/main/ts/textpatterns/core/PatternTypes.ts
ephox-mogran: Good point. 

- [ ] change comment.

modules/tinymce/src/core/main/ts/api/textpatterns/TextPatterns.ts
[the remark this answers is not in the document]
lnewson: We really just need to either find a linter rule for this or write one... it's the latter anyways. So it's either a single line or if it's multiline then the import/from statements should be on their own lines.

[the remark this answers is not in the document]
TheSpyder: yes I'm trying to put a stop to style questions in PRs. If there's a style issue, please just log a ticket to update (or add) a linter rule to cover it.

modules/tinymce/src/core/test/ts/browser/textpatterns/TextPatternsPublicApiTest.ts
TheSpyder: As long as refactoring doesn't auto-update the tests and make it pass incorrectly 😆

TheSpyder (approved):

modules/tinymce/src/core/test/ts/browser/textpatterns/TextPatternsPublicApiTest.ts
ephox-mogran: Well, then it's the review's job to see that the test changed :P

modules/tinymce/src/core/main/ts/api/textpatterns/TextPatterns.ts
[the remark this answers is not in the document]
lnewson: Found https://www.npmjs.com/package/eslint-plugin-import-newlines which I think should do what's needed here.

PR #8095, opened 2022-09-04
TINY-9098: Fixed a regression that caused enter to work within noneditable elements

lnewson:
Related Ticket: TINY-9098

Description of Changes:

Fixes a regression introduced in 6.2 development when I was doing the strict types as the new line logic relies on `getEditableRoot` returning a nullable value. So this swaps that back and adds other checks/fallbacks where required to handle that.

Pre-checks:
* [x] ~Changelog entry added~ N/A as this bug hasn't been released

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/plugins/table/test/ts/browser/settings/TableGridFalseTest.ts
+++ b/modules/tinymce/src/plugins/table/test/ts/browser/settings/TableGridFalseTest.ts
@@ -1,4 +1,4 @@
-import { Assertions, Waiter } from '@ephox/agar';
+import { Assertions, UiFinder, Waiter } from '@ephox/agar';
 import { describe, it } from '@ephox/bedrock-client';
 import { TinyHooks, TinyUiActions } from '@ephox/wrap-mcagar';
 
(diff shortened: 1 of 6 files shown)


modules/tinymce/src/plugins/table/test/ts/browser/settings/TableGridFalseTest.ts
```diff
@@ -18,7 +18,8 @@ describe('browser.tinymce.plugins.table.TableGridFalse', () => {
     await Waiter.pTryUntil('click table menu', () =>
       TinyUiActions.clickOnUi(editor, 'div.tox-menu div.tox-collection__item .tox-collection__item-label:contains("Table")')
     );
-    const dialog = await TinyUiActions.pWaitForDialog(editor, 'div.tox-dialog:has(div.tox-dialog__title:contains("Table Properties"))');
+    const dialog = await TinyUiActions.pWaitForDialog(editor);
```
lnewson, line 21: This test fails on Chrome v105 as it introduces `:has` selector support and this doesn't work (I'm assuming since `:contains` is a Sizzle only selector). So while this isn't exactly the same, it does still ensure we're getting the correct dialog.

HAFRMO (approved):

spocke (approved):

hamza0867 (approved):

PR #8096, opened 2022-09-05
TINY-8935: Fix formatter placing cursor inside a noneditable node

ltrouton:
Related Ticket: TINY-8935

Description of Changes:
* Fixes a regression that was caused when adding support for wrapping noneditable elements. After either applying or removing a format, the selection is kept on the noneditable element if it was already selected. Additionally, there is some extra checks to prevent placing the cursor inside a noneditable element.
 
Pre-checks:
* [X] Changelog entry added (This was never released so not including a changelog entry)

Review:
* [X] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/fmt/ApplyFormat.ts
+++ b/modules/tinymce/src/core/main/ts/fmt/ApplyFormat.ts
@@ -1,4 +1,4 @@
-import { Arr, Obj, Type } from '@ephox/katamari';
+import { Arr, Fun, Obj, Type } from '@ephox/katamari';
 import { PredicateExists, SugarElement } from '@ephox/sugar';
 
 import DOMUtils from '../api/dom/DOMUtils';
@@ -137,17 +137,17 @@ const applyFormat = (ed: Editor, name: string, vars?: FormatVars, node?: Node | 
 
     const canRenameBlock = (nodeName: string, parentName: string, isEditableDescendant: boolean) => {
       const isValidBlockFormatForNode =
-            FormatUtils.isNonWrappingBlockFormat(format) &&
-            FormatUtils.isTextBlock(ed, nodeName) &&
-            FormatUtils.isValid(ed, parentName, wrapName);
+        FormatUtils.isNonWrappingBlockFormat(format) &&
+        FormatUtils.isTextBlock(ed, nodeName) &&
+        FormatUtils.isValid(ed, parentName, wrapName);
       return isEditableDescendant && isValidBlockFormatForNode;
     };
 
     const canWrapNode = (node: Node, parentName: string, isEditableDescendant: boolean, isWrappableNoneditableElm: boolean) => {
       const nodeName = node.nodeName.toLowerCase();
       const isValidWrapNode =
-            FormatUtils.isValid(ed, wrapName, nodeName) &&
-            FormatUtils.isValid(ed, parentName, wrapName);
+        FormatUtils.isValid(ed, wrapName, nodeName) &&
+        FormatUtils.isValid(ed, parentName, wrapName);
       // If it is not node specific, it means that it was not passed into 'formatter.apply` and is within the editor selection
       const isZwsp = !nodeSpecific && NodeType.isText(node) && Zwsp.isZwsp(node.data);
       const isCaret = isCaretNode(node);
@@ -335,16 +335,16 @@ const applyFormat = (ed: Editor, name: string, vars?: FormatVars, node?: Node | 
       if (!isCollapsed || !FormatUtils.isInlineFormat(format) || TableCellSelection.getCellsFromEditor(ed).length) {
         // Apply formatting to selection
         selection.setRng(RangeNormalizer.normalize(selection.getRng()));
-        SelectionUtils.preserve(selection, true, () => {
-          SelectionUtils.runOnRanges(ed, (selectionRng, fake) => {
-            const expandedRng = fake ? selectionRng : ExpandRange.expandRng(ed, selectionRng, formatList);
-            applyRngStyle(dom, expandedRng, false);
-          });
-        });
 
-        if (dom.getContentEditable(selection.getNode()) !== 'false') {
-          FormatUtils.moveStart(dom, selection, selection.getRng());
-        }
+        FormatUtils.adjustSelectionAfter(ed, () => {
+          SelectionUtils.preserve(selection, true, () => {
+            SelectionUtils.runOnRanges(ed, (selectionRng, fake) => {
+              const expandedRng = fake ? selectionRng : ExpandRange.expandRng(ed, selectionRng, formatList);
+              applyRngStyle(dom, expandedRng, false);
+            });
+          });
+        }, Fun.always);
+
         ed.nodeChanged();
       } else {
         CaretFormat.applyCaretFormat(ed, name, vars);
--- a/modules/tinymce/src/core/main/ts/fmt/ApplyFormat.ts
+++ b/modules/tinymce/src/core/main/ts/fmt/ApplyFormat.ts
@@ -336,14 +336,16 @@ const applyFormat = (ed: Editor, name: string, vars?: FormatVars, node?: Node | 
         // Apply formatting to selection
         selection.setRng(RangeNormalizer.normalize(selection.getRng()));
 
-        FormatUtils.adjustSelectionAfter(ed, () => {
-          SelectionUtils.preserve(selection, true, () => {
+        FormatUtils.preserveSelection(
+          ed,
+          () => {
             SelectionUtils.runOnRanges(ed, (selectionRng, fake) => {
               const expandedRng = fake ? selectionRng : ExpandRange.expandRng(ed, selectionRng, formatList);
               applyRngStyle(dom, expandedRng, false);
             });
-          });
-        }, Fun.always);
+          },
+          Fun.always
+        );
 
         ed.nodeChanged();
       } else {
--- a/modules/tinymce/src/core/main/ts/fmt/FormatUtils.ts
+++ b/modules/tinymce/src/core/main/ts/fmt/FormatUtils.ts
@@ -9,6 +9,7 @@ import Editor from '../api/Editor';
 import * as Options from '../api/Options';
 import * as Bookmarks from '../bookmark/Bookmarks';
 import * as NodeType from '../dom/NodeType';
+import * as SelectionUtils from '../selection/SelectionUtils';
 import * as Whitespace from '../text/Whitespace';
 import { isCaretNode } from './FormatContainer';
 import { BlockFormat, Format, FormatAttrOrStyleValue, FormatVars, InlineFormat, MixedFormat, SelectorFormat } from './FormatTypes';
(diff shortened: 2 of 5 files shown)


modules/tinymce/src/core/main/ts/fmt/FormatUtils.ts
```diff
@@ -18,39 +18,73 @@ const isNode = (node: any): node is Node => !!(node).nodeType;
 const isElementNode = (node: Node): node is Element =>
   NodeType.isElement(node) && !Bookmarks.isBookmarkNode(node) && !isCaretNode(node) && !NodeType.isBogus(node);
 
-const isInlineBlock = (node: Node): boolean => {
-  return node && /^(IMG)$/.test(node.nodeName);
+// Control blocks are any elements e.g. (img, iframe, noneditable element) that can be selected individually
+// In TinyMCE, selected blocks are indicated with the data-mce-selected attribute
+const isControlBlockSelected = (dom: DOMUtils, node: Node): boolean => {
```
spocke, line 23: Maybe name this `isControlElementSelected` since block could mean block elements and not include the inline variants that also have this attribute.

modules/tinymce/src/core/main/ts/fmt/FormatUtils.ts
```diff
@@ -18,39 +18,73 @@ const isNode = (node: any): node is Node => !!(node).nodeType;
...
 const isEditable = (elm: HTMLElement): boolean =>
   elm.isContentEditable === true;
 
-const moveStart = (dom: DOMUtils, selection: EditorSelection, rng: Range): void => {
-  const offset = rng.startOffset;
-  let container = rng.startContainer;
+const adjustSelectionAfter = <T>(editor: Editor, action: () => T, shouldMoveStart: (startNode: Node) => boolean): void => {
+  const { selection, dom } = editor;
+  const selectedNodeBeforeAction = selection.getNode();
+  const isSelectedBeforeNodeNoneditable = dom.getContentEditable(selectedNodeBeforeAction) === 'false';
```
spocke, line 41: Probably a bit shorter when minifed.

```suggestion
  const isSelectedBeforeNodeNoneditable = NodeType.isContentEditableFalse(selectedNodeBeforeAction);
```

modules/tinymce/src/core/main/ts/fmt/FormatUtils.ts
```diff
@@ -18,39 +18,73 @@ const isNode = (node: any): node is Node => !!(node).nodeType;
...
+  if (isBeforeNodeStillNoneditable && dom.isChildOf(selectedNodeBeforeAction, editor.getBody())) {
+    editor.selection.select(selectedNodeBeforeAction);
+  } else if (shouldMoveStart(selection.getStart())) {
+    moveStartToNearestText(dom, selection);
   }
+};
+
+// Note: The reason why we only care about moving the start is because MatchFormat and its function use the start of the selection to determine if a selection has a given format or not
+// The format toolbar buttons use the MatchFormat functions to determine whether the button should be active or not
+// Therefore, if the start is not moved forward in certain cases, the match format detection and by extension,the toolbar button state will be incorrect
```
spocke, line 57: I think the first note here is enough so I would remove this. It's how the formatter api works that the UI is using that api is besides the point.

modules/tinymce/src/core/main/ts/fmt/FormatUtils.ts
```diff
@@ -18,39 +18,73 @@ const isNode = (node: any): node is Node => !!(node).nodeType;
...
+  }
 };
 
 const isEditable = (elm: HTMLElement): boolean =>
   elm.isContentEditable === true;
 
-const moveStart = (dom: DOMUtils, selection: EditorSelection, rng: Range): void => {
-  const offset = rng.startOffset;
-  let container = rng.startContainer;
+const adjustSelectionAfter = <T>(editor: Editor, action: () => T, shouldMoveStart: (startNode: Node) => boolean): void => {
```
spocke, line 38: Why did you need the T here is there action functions that return something. Could this be `() => void`?

modules/tinymce/src/core/main/ts/fmt/FormatUtils.ts
```diff
@@ -18,39 +18,73 @@ const isNode = (node: any): node is Node => !!(node).nodeType;
...
 
-  if (container === rng.endContainer) {
-    if (isInlineBlock(container.childNodes[offset])) {
-      return;
-    }
+  action();
+
+  // Check previous selected node before the action still exists in the DOM
+  // and is still noneditable
+  const isBeforeNodeStillNoneditable = isSelectedBeforeNodeNoneditable && dom.getContentEditable(selectedNodeBeforeAction) === 'false';
```
spocke, line 47: Same here there is a predicate that you could use a.f.a.i.k.

spocke (approved):

modules/tinymce/src/core/main/ts/fmt/FormatUtils.ts
lnewson, line 23: I think this function would be better suited in `ControlSelection` as it's responsible for defining how something is selected like this... Additionally, if someone changes the logic later they are more likely to see it being in the same spot instead of being scattered in a completely separate section of the code.

lnewson: Actually, the more I read this, this is not about Control Elements, as a noneditable element is **not** a control selection (remember control elements are ones that can be resized). It appears to just be an entire element selection. So maybe just rename this by removing `Control` and update the comment?

modules/tinymce/src/core/main/ts/fmt/FormatUtils.ts
ltrouton: Yeah, you are probably right. Happy to remove those lines

modules/tinymce/src/core/main/ts/fmt/FormatUtils.ts
ltrouton: Yeah, having the T probably is unnecessary. Will remove

modules/tinymce/src/core/main/ts/fmt/FormatUtils.ts
ltrouton: Ah yep. My use of terms was incorrect. I'll try to figure out a better name. Thanks

ltrouton: I ended up renaming it to `isElementDirectlySelected` as I think that makes it a bit clearer

MitchCTiny (approved):

modules/tinymce/src/core/main/ts/fmt/FormatUtils.ts
```diff
@@ -18,39 +18,71 @@ const isNode = (node: any): node is Node => !!(node).nodeType;
...
+// Note: The reason why we only care about moving the start is because MatchFormat and its function use the start of the selection to determine if a selection has a given format or not
+const moveStartToNearestText = (dom: DOMUtils, selection: EditorSelection): void => {
+  const rng = selection.getRng();
+  const { startContainer, startOffset } = rng;
+  const selectedNode = selection.getNode();
 
-  // Move startContainer/startOffset in to a suitable node
-  if (NodeType.isElement(container)) {
-    const nodes = container.childNodes;
+  if (isElementDirectlySelected(dom, selectedNode)) {
```
ephox-mogran, line 61: This code path appears to be new, but doesn't look like it's tested anywhere for elements that aren't cef. Is that accurate?

modules/tinymce/src/core/main/ts/fmt/FormatUtils.ts
```diff
@@ -18,39 +18,71 @@ const isNode = (node: any): node is Node => !!(node).nodeType;
...
 
 const isEditable = (elm: HTMLElement): boolean =>
   elm.isContentEditable === true;
 
-const moveStart = (dom: DOMUtils, selection: EditorSelection, rng: Range): void => {
-  const offset = rng.startOffset;
-  let container = rng.startContainer;
+const adjustSelectionAfter = (editor: Editor, action: () => void, shouldMoveStart: (startNode: Node) => boolean): void => {
+  const { selection, dom } = editor;
+  const selectedNodeBeforeAction = selection.getNode();
```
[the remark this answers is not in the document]
ltrouton: Naming these was a struggle but I figured seeing `selection.getNode();` makes it pretty obvious what is happening.

modules/tinymce/src/core/main/ts/fmt/ApplyFormat.ts
```diff
@@ -335,16 +335,16 @@ const applyFormat = (ed: Editor, name: string, vars?: FormatVars, node?: Node |
       if (!isCollapsed || !FormatUtils.isInlineFormat(format) || TableCellSelection.getCellsFromEditor(ed).length) {
         // Apply formatting to selection
         selection.setRng(RangeNormalizer.normalize(selection.getRng()));
-        SelectionUtils.preserve(selection, true, () => {
-          SelectionUtils.runOnRanges(ed, (selectionRng, fake) => {
-            const expandedRng = fake ? selectionRng : ExpandRange.expandRng(ed, selectionRng, formatList);
-            applyRngStyle(dom, expandedRng, false);
+
+        FormatUtils.adjustSelectionAfter(ed, () => {
```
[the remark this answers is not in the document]
ltrouton: I think this is the best we can do for now. Maybe we should create a Jira to look at adding the CEF checking to `SelectionUtils.preserve` instead? Can then add a comment in the code

modules/tinymce/src/core/main/ts/fmt/FormatUtils.ts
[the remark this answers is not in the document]
ephox-mogran: Yeah, fair enough. My problem was I didn't read from the start... There's no need to change anything here.

modules/tinymce/src/core/main/ts/fmt/ApplyFormat.ts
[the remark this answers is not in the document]
ephox-mogran: Yeah that's fine.

lnewson, line 339: This is rather messy and hard to read... is they any reason we couldn't move the `SelectionUtils.preserve` logic inside the `FormatUtils` helper and rename it to something like `preserveSelection` as that's what it's really trying to do here? We could then just treat this as a custom wrapper to `SelectionUtils.preserve` for formatting specifically and it could then easily be moved later should we need to.

ltrouton: True. That would probably be a cleaner abstraction. Happy to make that change

ltrouton: Have made the update. Thanks for the suggestion

modules/tinymce/src/core/main/ts/fmt/FormatUtils.ts
ltrouton: Yep, I am only testing CEF at the moment. I should be able and another one for formatting an image as another test case.

ltrouton: Additional test has been added 👍

PR #8097, opened 2022-09-05
TINY-8881: drag and drop on the left of CEF elements

lorenzo-pomili:
Related Ticket: TINY-8881

Description of Changes:
In order to allow an element to be correctly placed when dragged over a CEF element, this change:

- find the last CEF ancestor
- if it exists, check if the parent node exists
- insert the element in the parent node at the last CEF ancestor index

P.S. The nested `if` is only to avoid searching for the last non-editable element each time

Pre-checks:
* [x] Changelog entry added

Review:
* [ ] Milestone set


modules/tinymce/src/core/test/ts/browser/DragnDropCEFTest.ts
```diff
@@ -0,0 +1,82 @@
+import { Mouse, UiFinder } from '@ephox/agar';
+import { describe, it } from '@ephox/bedrock-client';
+import { Arr } from '@ephox/katamari';
+import { SugarLocation } from '@ephox/sugar';
+import { TinyAssertions, TinyDom, TinyHooks } from '@ephox/wrap-mcagar';
+
+import Editor from 'tinymce/core/api/Editor';
+
+describe('browser.tinymce.core.DragnDropCEFTest', () => {
+  const getBaseCEFElement = (name: string) => [ `<div class="${name}" style="margin: 40px;" contenteditable="false">`,
```
spocke, line 10: Does the HTML need to be this complex to simulate the fix here I think a div with a width and height should be enough.

modules/tinymce/src/core/test/ts/browser/DragnDropCEFTest.ts
```diff
@@ -0,0 +1,82 @@
...
+    '<article style="padding: 10px 50px;" contenteditable="true">',
+    `<p>${name} description</p>`,
+    '</article>',
+    '</div>',
+    '</div>',
+    '</div>',
+    '</div>'
+  ].join('');
+
+  const getContentWithCefElements = (elementsNames: string[]): string => `<div>${Arr.foldl(elementsNames, (acc, elementName) =>
```
spocke, line 28: A fold inside a template string that is a bit hard to read at least for me. Maybe just pull out the fold result into a variable and use that in the template string.

modules/tinymce/src/core/test/ts/browser/DragnDropCEFTest.ts
```diff
@@ -0,0 +1,82 @@
...
+    acc + getBaseCEFElement(elementName)
+  , '')}</div>`;
+
+  const wait = (ms: number) => new Promise((r) => setTimeout(r, ms));
```
spocke, line 32: Use Waiter.pWait in agar.

- [ ] Switch to agar pWait

modules/tinymce/src/core/test/ts/browser/DragnDropCEFTest.ts
```diff
@@ -0,0 +1,82 @@
...
+    Mouse.mouseMoveTo(toDrag, xDelta - 5, yDelta - 5);
+
+    // little trick that give "time" to CaretRange.fromPoint to find the position
+    await wait(0);
+    Mouse.mouseUp(toDrag);
+
+    TinyAssertions.assertContent(editor, getContentWithCefElements([ 'obstacle', 'toDrag', 'destination' ]));
+  });
+
+  it('TINY-8881: Dragging CEF element over the first element should work as expected', async () => {
```
spocke, line 61: The code of these two tests are very similar could they be a reusable function that takes different deltas so that if we add more than these 2 cases then it would be easier? Like `pDragDropCef(dx, dy)` or something?

modules/tinymce/CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Line separator scrolling in floating toolbars. #TINY-8948
 - A double bottom border appeared on inline mode editor for the `tinymce-5` skin. #TINY-9108
 - The editor header showed up even with no menubar and toolbar configured. #TINY-8819
+- Dragging over CEF element now works correctly #TINY-8881
```
spocke, line 19: Not sure if it's all that better but maybe since CEF is kind of internal name:

```suggestion
- Dragging a noneditable element before or after another noneditable element now works correctly. #TINY-8881
```

spocke (approved):

modules/tinymce/src/core/main/ts/caret/ClosestCaretCandidate.ts
```diff
@@ -102,10 +102,14 @@ const closestChildCaretCandidateNodeRect = (children: ChildNode[], clientX: numb
 
 const traverseUp = (rootElm: SugarElement<Node>, scope: SugarElement<Element>, clientX: number, clientY: number): Optional<NodeClientRect> => {
   const helper = (scope: SugarElement<Element>, prevScope: Optional<SugarElement<Element>>): Optional<NodeClientRect> => {
+    const childNodesWithoutGhost = Arr.filter(Arr.from(scope.dom.childNodes), (node) =>
+      !(node as HTMLElement)?.classList?.contains('mce-drag-container')
+    );
```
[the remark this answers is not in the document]
lorenzo-pomili: ok I changed to the second version

modules/tinymce/src/core/test/ts/browser/DragnDropCEFTest.ts
lorenzo-pomili: ok, I simplified it

modules/tinymce/src/core/test/ts/browser/DragnDropCEFTest.ts
lorenzo-pomili: simplifying it I removed the `join` and now is just a template string

modules/tinymce/src/core/test/ts/browser/DragnDropCEFTest.ts
lorenzo-pomili: done

modules/tinymce/src/core/test/ts/browser/DragnDropCEFTest.ts
lorenzo-pomili: I refactored it, let me know if it is ok

modules/tinymce/CHANGELOG.md

modules/tinymce/src/core/test/ts/browser/DragnDropCEFTest.ts
```diff
@@ -0,0 +1,62 @@
+import { Mouse, UiFinder, Waiter } from '@ephox/agar';
```
lnewson, line 1: Is there a reason this wasn't added to the existing `DragDropOverridesTest`? As this is just updating that and we already have a test for that does the same or a very similar thing.

- [ ] Reconsider using a new test for this when a test file already exists with similar tests

modules/tinymce/src/core/test/ts/browser/DragnDropCEFTest.ts
```diff
@@ -0,0 +1,62 @@
...
+    if (!Arr.contains(elementsNames, 'toDrag') || !Arr.contains(elementsNames, 'destination')) {
+      // eslint-disable-next-line no-throw-literal
+      throw new Error('This function require to have an element named toDrag and one destination');
+    }
+    return `<div>${Arr.foldl(elementsNames, (acc, elementName) =>
+      acc + getBaseCEFElement(elementName)
+    , '')}</div>`;
+  };
+
+  const hook = TinyHooks.bddSetup<Editor>({
```
lnewson, line 23: If keeping this, why do you need a full editor and not a lightweight one? This doesn't appear to use the UI at all so this should be using `bddSetupLight` instead.

modules/tinymce/src/core/test/ts/browser/DragnDropCEFTest.ts
lorenzo-pomili: you are right, I moved the tests to a new `context` in `DragDropOverridesTest`

modules/tinymce/src/core/test/ts/browser/DragnDropCEFTest.ts
lorenzo-pomili: changed

lorenzo-pomili:
close since now there is the PR with the correct ticket number: https://github.com/tinymce/tinymce/pull/8195

PR #8079, opened 2022-08-26
TINY-8954: Adding the final link and text

HAFRMO:
Related Ticket: 

Description of Changes:
Add the final text and link ( Link not yet set up on the site ). No changelog needed, no tests needed.

Pre-checks:
* [x] Changelog entry added
* [x] Tests have been added (if applicable)
* [x] Branch prefixed with `feature/`, `hotfix/` or `spike/`

Review:
* [x] Milestone set
* [x] Docs ticket created (if applicable)

GitHub issues (if applicable):

PR #8099, opened 2022-09-05
TINY-9106: Fix promotion button height difference between hidpi and regular screens

noxuhax:
Related Ticket: TINY-9106

Description of Changes:
* Fix promotion button height and center it’s content regardless of content’s height
* Moved `cursor: pointer` property from promotion wrapper to button itself

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [x] Milestone set

### DIFF
--- a/modules/oxide/src/less/theme/components/menubar/promotion.less
+++ b/modules/oxide/src/less/theme/components/menubar/promotion.less
@@ -5,13 +5,13 @@
 @promotion-background-color: #E8F1F8;
 @promotion-background-hover-color: #B4D7FF;
 @promotion-background-active-color: #D9EDF7;
+@promotion-button-height: @menubar-select-font-size * 1.9;
 @promotion-text-color: #086BE6;
 
 .tox {
   .tox-promotion {
     background: @menubar-background;
     background-color: @menubar-background-color;
-    cursor: pointer;
     grid-column: 2;
     grid-row: 1;
     padding-inline-end: 8px;
@@ -20,11 +20,14 @@
   }
 
   .tox-promotion-link {
+    align-items: unsafe center;
     background-color: @promotion-background-color;
     border-radius: 5px;
     color: @promotion-text-color;
-    display: block;
+    cursor: pointer;
+    display: flex;
     font-size: @menubar-select-font-size;
+    height: @promotion-button-height;
     padding: 4px 8px;
     white-space: nowrap;
 


vpyshnenko (approved):

lnewson:
This is how this looks on my screen now:
image
It's still not quite right as the text is not centered and is partly towards the bottom, but it's much better. Unfortunately I think this is just what we're going to have to live with given the Product decision to not use a proper icon here.

lnewson (approved):

lorenzo-pomili (approved):

modules/oxide/src/less/theme/components/menubar/promotion.less
```diff
@@ -20,11 +20,14 @@
   }
 
   .tox-promotion-link {
+    align-items: unsafe center;
     background-color: @promotion-background-color;
     border-radius: 5px;
     color: @promotion-text-color;
-    display: block;
+    cursor: pointer;
```
spocke, line 27: Is a pointer what we really want here the Tiny logo in the statusbar has a default link icon. And this is also kind of a link even if it looks like a button.

modules/oxide/src/less/theme/components/menubar/promotion.less
```diff
@@ -20,11 +20,14 @@
   }
 
   .tox-promotion-link {
+    align-items: unsafe center;
```
spocke, line 23: So it will start clipping to the height of the container if the emoji renders to high. Never used `unsafe` before?

spocke (approved):

spocke:
> This is how this looks on my screen now: image (`tinymce-5`) image (`oxide`) It's still not quite right as the text is not centered and is partly towards the bottom, but it's much better. Unfortunately I think this is just what we're going to have to live with given the Product decision to not use a proper icon here.

I can barely see the icon on Linux. Since it has no borders it just yellow on light blue and that floats into each other.
image

Should we switch this to an icon in 6.3 it seems that it has all kinds of problems.

lnewson:
@spocke I raised that in https://github.com/tinymce/tinymce/pull/8079#issuecomment-1229837240 and was told very clearly Product & Design don't want an icon.

modules/oxide/src/less/theme/components/menubar/promotion.less
noxuhax: `pointer` is a hand-with-extended-finger one, the one that’s used on links, so I see no issue here. You might be confusing it with `default` which is usually an arrow, the one that used pretty much on every screen surface that is not interactive. Or am I reading your comment wrong?

modules/oxide/src/less/theme/components/menubar/promotion.less
noxuhax: It won’t be clipping anything unless we set an `overflow` for the button. `unsafe` just prevents contents from stretching parent container but since the intended content here is a line of text it should be fine.

modules/oxide/src/less/theme/components/menubar/promotion.less
spocke: Ahh good got confused there about the icons. See now that it has the right one. We use the default arrow for all other UI elements the finger for this one and the other promotional link so all is good.

modules/oxide/src/less/theme/components/menubar/promotion.less
spocke: But if it prevents the parent container from stretching what does it do if it doesn't clip scroll, overflow outside of the box?

spocke: I'm just curious if say an emoji some how renders very large on some system since the default emoji rendering is different will what would happen.

noxuhax: Yeah, it’s contents are gonna extend beyond the container. In that case on both top and bottom of the button, keeping text centered. 
I highly doubt that there is a system that has default emoji size almost twice the size of a text that it sits in, since emoji are intended to be used along the text. But yeah, if something goes really south then thunderbolt will be sticking out at the top and the bottom of the button.

spocke: Right thanks for clarifying that.

PR #8105, opened 2022-09-06
TINY-9111: Revert changes for how selected nested noneditable element is formatted

ltrouton:
Related Ticket: TINY-9111

Description of Changes:
* It was found during QA that when a noneditable element was directly selected and had editable children, applying a format would format the editable children. Whilst this may have been fine, there was then no way to toggle the format off again when the nonediatble element was directly selected. Because of this, part of the code deleted from #7968 has been put back so that the 6-stable behaviour is maintained.

It might be better if the editable children of the noneditable element could be formatted even if the noneditable element was selected so that it is consistent with how a ranged selection works. Although, this would require various changes to MatchFormat and RemoveFormat so is best left for another ticket if we wanted to do it.

Pre-checks:
* [X] ~Changelog entry added~ (just reverting some changes that hadn't been released yet)

Review:
* [X] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/fmt/ApplyFormat.ts
+++ b/modules/tinymce/src/core/main/ts/fmt/ApplyFormat.ts
@@ -319,6 +319,16 @@ const applyFormat = (ed: Editor, name: string, vars?: FormatVars, node?: Node | 
     });
   };
 
+  // TODO: TINY-9142: Remove this to make nested noneditable formatting work
+  const targetNode = FormatUtils.isNode(node) ? node : selection.getNode();
+  if (dom.getContentEditable(targetNode) === 'false' && !FormatUtils.isWrappableNoneditable(ed, targetNode)) {
+    // node variable is used by other functions above in the same scope so need to set it here
+    node = targetNode;
+    applyNodeStyle(formatList, node);
+    Events.fireFormatApply(ed, name, node, vars);
+    return;
+  }
+
   if (format) {
     if (node) {
       if (FormatUtils.isNode(node)) {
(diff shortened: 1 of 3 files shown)


MitchCTiny (approved):

ephox-mogran:
Does this need a changelog entry either added or removed?

modules/tinymce/src/core/main/ts/fmt/ApplyFormat.ts
```diff
@@ -319,6 +319,16 @@ const applyFormat = (ed: Editor, name: string, vars?: FormatVars, node?: Node |
     });
   };
 
+  // TODO: TINY-9142: Remove this to make nested noneditable formatting work
+  const targetNode = FormatUtils.isNode(node) ? node : selection.getNode();
+  if (dom.getContentEditable(targetNode) === 'false' && !FormatUtils.isWrappableNoneditable(ed, targetNode)) {
+    // node variable is used by other functions above in the same scope so need to set it here
+    node = targetNode;
+    applyNodeStyle(formatList, node);
```
ephox-mogran, line 327: What is `applyNodeStyle` for here? Aren't we just skipping formatting this node altogether? Or is the behaviour a lot more nuanced than that? What is the desired behaviour? 

* we can format the contenteditable false itself because we are able to toggle off that format
* we can't format the contenteditable false because it doesn't have the wrapper class ?

ephox-mogran: Or is it because the applying of styling will put something _outside_ that we still need to apply the styles? If so, what's the significance of the wrapper class? What difference in behaviour does it create?

ephox-mogran (approved):

modules/tinymce/src/core/main/ts/fmt/ApplyFormat.ts
lnewson: To try and help answer this since Luke is pretty busy atm, selector formats can be configured to format the CEF element regardless by adding `ceFalseOverride: true` to the format. So it's that case whereby we still need to apply formats and `applyNodeStyle` handles that case and returns otherwise.

ephox-mogran: Oh, I see. Each format has its own `ceFalseOverride` setting, so the checking is done in applyNodeStyle

```
 // Check if the node is nonediatble and if the format can override noneditable node
      if (dom.getContentEditable(node) === 'false' && !format.ceFalseOverride) {
        return true;
      }
```

TheSpyder (approved):

PR #8132, opened 2022-09-15
TINY-9029: Added behavior to be consistent in both directions.

HAFRMO:
Related Ticket: 

Description of Changes:
The ExpandRange would behave somewhat differently depending on the direction due to the start range position being looked at as one step further to the right than it really was.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


TheSpyder (approved):

modules/tinymce/src/core/main/ts/fmt/ExpandRange.ts
```diff
@@ -38,9 +38,9 @@ const findParentContentEditable = (dom: DOMUtils, node: Node) => {
 
 const walkText = (start: boolean, node: Text, offset: number, predicate: (chr: string) => boolean) => {
   const str = node.data;
-  for (let i = offset; start ? i >= 0 : i < str.length; start ? i-- : i++) {
-    if (predicate(str.charAt(i))) {
-      return start ? i + 1 : i;
+  for (let i = offset; start ? i > 0 : i < str.length; start ? i-- : i++) {
+    if (predicate(str.charAt(start ? i - 1 : i))) {
+      return i;
     }
```
spocke, line 44: Are we sure that we don't want to return the value that the predicate is checking like so:
```suggestion
    const idx = start ? i - 1 : i;
    if (predicate(str.charAt(idx))) {
      return idx;
    }
```

- [ ] The previous code checked one thing but returned another thing. Now it checks one thing differently from the return value only on start. Is that right?

- [ ] Are there edge cases with end or beginning of strings that is not covered by tests?

modules/tinymce/src/core/main/ts/fmt/ExpandRange.ts
```diff
@@ -38,9 +38,9 @@ const findParentContentEditable = (dom: DOMUtils, node: Node) => {
 
 const walkText = (start: boolean, node: Text, offset: number, predicate: (chr: string) => boolean) => {
   const str = node.data;
-  for (let i = offset; start ? i >= 0 : i < str.length; start ? i-- : i++) {
-    if (predicate(str.charAt(i))) {
-      return start ? i + 1 : i;
+  for (let i = offset; start ? i > 0 : i < str.length; start ? i-- : i++) {
+    if (predicate(str.charAt(start ? i - 1 : i))) {
+      return i;
```
spocke, line 43: I wonder if this would be better as two separate loops this is crazy complex now.

- [ ] Consider splitting up the loops to two loops

```suggestion
  if (start) {
   for (let i = offset; i > 0; i--) {
       if (predicate(str.charAt(i - 1))) {
         return i; // Should this be i - 1?
       }
    }
  } else {
   for (let i = offset; i < str.length; i++) {
       if (predicate(str.charAt(i))) {
         return i;
       }
   }
 }
```

spocke (approved):

modules/tinymce/src/core/main/ts/fmt/ExpandRange.ts
HAFRMO: What happened was that if it was placed in the middle of a string, say `fo|ur` at the |, the end-position would check 'u' and 'r', while the start-position would check 'u', 'o', and 'f'. This meant that the result was unintentionally behaving differently in the two different directions. This change just makes the start-position check the expected 'o' and 'f', instead of starting one character 'behind'. This caused some issues if it was placed 'beside' something of interest. As for the end or beginning, it'll return -1, as per design.

modules/tinymce/src/core/main/ts/fmt/ExpandRange.ts
HAFRMO: I did think about it, but I wasn't sure it was any more readable, or needed.

modules/tinymce/src/core/main/ts/fmt/ExpandRange.ts
spocke: Ok, I assume that we have tests that cover those cases now.

modules/tinymce/src/core/main/ts/fmt/ExpandRange.ts
HAFRMO: It also gets a bit tricky to read when separate, but I get you. I'll change it.

modules/tinymce/src/core/main/ts/fmt/ExpandRange.ts
HAFRMO: We have a couple of tests, yes. Not the most tested part, but at least it's not un-tested.

PR #8138, opened 2022-09-19
TINY-9164: Improve accessibility of charmap dialog

tiny-ben-tran:
Related Ticket: TINY-9164

Description of Changes:
* Placeholder text

Pre-checks:
* [ ] Changelog entry added

Review:
* [ ] Milestone set

### DIFF
--- a/modules/bridge/src/demo/ts/dialogs/CharmapDialog.ts
+++ b/modules/bridge/src/demo/ts/dialogs/CharmapDialog.ts
@@ -11,9 +11,13 @@ export const createCharmapDialog = (): void => {
         items: [
           {
             name: 'chars',
-            type: 'collection'
+            type: 'collection',
+            aria: {
+              role: 'button',
+              description: 'Insert this character and close dialog'
+            }
           }
-        ]
+        ],
       },
       buttons: [
         {
--- a/modules/bridge/src/demo/ts/dialogs/CharmapDialog.ts
+++ b/modules/bridge/src/demo/ts/dialogs/CharmapDialog.ts
@@ -17,7 +17,7 @@ export const createCharmapDialog = (): void => {
               description: 'Insert this character and close dialog'
             }
           }
-        ],
+        ]
       },
       buttons: [
         {
--- a/modules/tinymce/src/plugins/charmap/main/ts/ui/Dialog.ts
+++ b/modules/tinymce/src/plugins/charmap/main/ts/ui/Dialog.ts
@@ -18,7 +18,11 @@ const open = (editor: Editor, charMap: CharMap[]): void => {
     },
     {
       type: 'collection',
-      name: 'results'
+      name: 'results',
+      aria: {
+        role: 'button',
+        description: 'Insert this character and close the dialog'
+      }
       // TODO TINY-3229 implement collection columns properly
       // columns: 'auto'
     }
--- a/modules/tinymce/src/themes/silver/main/ts/ui/dialog/Collection.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/ui/dialog/Collection.ts
@@ -28,33 +28,32 @@ export const renderCollection = (
 ): SketchSpec => {
   // DUPE with TextField.
   const pLabel = spec.label.map((label) => renderLabel(label, providersBackstage));
+  const accessibilityHtml = spec.aria.fold(
+    Fun.constant({ insertAndCloseInfo: '', itemAttributes: '' }),
+    (value) => {
+      let insertAndCloseInfo = '';
+      const attributes = [];
+
+      if (value.role.isSome()) {
+        attributes.push(` role="${value.role.getOr('')}"`);
+      }
+
+      if (value.description.isSome()) {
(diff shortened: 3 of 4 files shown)


modules/bridge/src/demo/ts/dialogs/CharmapDialog.ts
```diff
@@ -11,9 +11,13 @@ export const createCharmapDialog = (): void => {
...
             name: 'chars',
-            type: 'collection'
+            type: 'collection',
+            aria: {
+              role: 'button',
+              description: 'Insert this character and close dialog'
+            }
           }
-        ]
+        ],
```
tiny-ben-tran, line 20:
```suggestion
        ]
```

lnewson (changes requested):
As I said on slack, we’ve specifically avoided putting the aria attribute as we don’t want that (i.e. is inconsistent with all other UI specs). It should be a basic spec that's not specifically tied to HTML generally. It also should be aimed at a specific UI/type of UI so having an open ended role doesn't make sense. The correct solution here is to do TINY-8764 (make it a grid so the titles work, keyboard nav works, etc...) as I said on slack and was ignored.

Additionally having the aria field on a collection makes it appear as though you’re setting the aria on the collection, not each collection item so it doesn’t make sense anyways.

modules/tinymce/src/themes/silver/main/ts/ui/dialog/Collection.ts
```diff
@@ -108,7 +129,10 @@ export const renderCollection = (
     dom: {
       tag: 'div',
       // FIX: Read from columns
-      classes: [ 'tox-collection' ].concat(spec.columns !== 1 ? [ 'tox-collection--grid' ] : [ 'tox-collection--list' ])
+      classes: [ 'tox-collection' ].concat(spec.columns !== 1 ? [ 'tox-collection--grid' ] : [ 'tox-collection--list' ]),
+      attributes: {
+        'aria-live': 'polite'
```
tiny-ben-tran, line 134: Just a couple of notes I found about using this `aria-live`:
- All characters that appear in the dialog content, are announced. This may not be the same as the audit's suggestion, which is to announce `how many items there are in the list from the start or if it has filtered out any results upon typing`
- The announcement is cut off early. We might need to escalate the politeness of `aria-live` to `aggressive`.

ephox-mogran: Yeah, this isn't really what we want to do. We're after something a bit more like `modules/alloy/src/main/ts/ephox/alloy/aria/AriaVoice.ts`, where it creates a span and adds it to the page with a specific message, and then deletes it. Because it has the appropriate `aria-live` settings, it is read by a screen-reader. We don't want to rely on it reading the whole collection.

modules/tinymce/src/themes/silver/main/ts/ui/dialog/Collection.ts
```diff
@@ -57,13 +77,14 @@ export const renderCollection = (
       const ariaLabel = itemText.replace(/\_| \- |\-/g, (match) => mapItemName[match]);
 
       const disabledClass = providersBackstage.isDisabled() ? ' tox-collection__item--state-disabled' : '';
-      return `<div class="tox-collection__item${disabledClass}" tabindex="-1" data-collection-item-value="${Entities.encodeAllRaw(item.value)}" title="${ariaLabel}" aria-label="${ariaLabel}">${iconContent}${textContent}</div>`;
+      return `<div class="tox-collection__item${disabledClass}" tabindex="-1" data-collection-item-value="${Entities.encodeAllRaw(item.value)}"` +
```
ephox-mogran, line 80: This is getting really hard to read. I wonder if we can break it up into an array that joins.

modules/tinymce/src/plugins/charmap/main/ts/ui/Dialog.ts
```diff
@@ -18,7 +18,11 @@ const open = (editor: Editor, charMap: CharMap[]): void => {
     },
     {
       type: 'collection',
-      name: 'results'
+      name: 'results',
+      aria: {
```
ephox-mogran, line 22: @lnewson is pretty insistent on that we don't want to have `aria` as part of any spec. One of the options we considered when we were doing this was making the items themselves have the description on them, though that's less positioned to be able to take advantage of a common description element. I don't know what we should do to have this description tacked on. Probably have to push it to ARCOM.

tiny-ben-tran: Pushing it to ARCOM is a good idea.

lnewson: As this has been said it's not something we want to do (re bridge changes) and needs to go back to our Accessibility vendor anyways for the button role thing, should this be closed for now? We can keep the branch around if necessary but it'd be nice to not have a long running PR sitting here that likely won't be merged (at least in its current state).

PR #8151, opened 2022-09-26
TINY-9025: Dragging CEF element toward the bottom edge no longer causes the page to scroll up

hamza0867:
Related Ticket: TINY-9025

Description of Changes:
*  Dragging CEF element toward the bottom edge no longer causes the page to scroll up

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - New `expand` function added to `tinymce.dom.RangeUtils` to return a new range expanded around the nearest word. #TINY-9001
 
 ### Fixed
+- Dragging CEF element toward the bottom edge would cause the page to scroll up. #TINY-9025
 - Range expanding capabilities would behave inconsistently depending on where the cursor was placed. #TINY-9029
 - Compilation errors were thrown when using TypeScript 4.8. #TINY-9161
 - Line separator scrolling in floating toolbars. #TINY-8948
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - New `expand` function added to `tinymce.dom.RangeUtils` to return a new range expanded around the nearest word. #TINY-9001
 
 ### Fixed
-- Dragging CEF element toward the bottom edge would cause the page to scroll up. #TINY-9025
+- Dragging a noneditable element toward the bottom edge would cause the page to scroll up. #TINY-9025
 - Range expanding capabilities would behave inconsistently depending on where the cursor was placed. #TINY-9029
 - Compilation errors were thrown when using TypeScript 4.8. #TINY-9161
 - Line separator scrolling in floating toolbars. #TINY-8948
--- a/modules/tinymce/src/core/main/ts/DragDropOverrides.ts
+++ b/modules/tinymce/src/core/main/ts/DragDropOverrides.ts
@@ -269,7 +269,7 @@ const move = (state: Singleton.Value<State>, editor: Editor) => {
     }
 
     if (state.dragging) {
-      const mouseEventOriginatedFromWithinTheEditor = e.currentTarget === editor.getDoc().firstElementChild;
+      const mouseEventOriginatedFromWithinTheEditor = e.currentTarget === editor.getDoc().documentElement;
       const targetPos = applyRelPos(state, MousePosition.calc(editor, e));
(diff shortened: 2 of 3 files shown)


modules/tinymce/src/core/main/ts/DragDropOverrides.ts
```diff
@@ -268,9 +269,10 @@ const move = (state: Singleton.Value<State>, editor: Editor) => {
     }
 
     if (state.dragging) {
+      const mouseEventOriginatedFromWithinTheEditor = e.currentTarget === editor.getDoc().firstElementChild;
```
spocke, line 272: Clearer to know what the first element it could in theory have been the head element.

```suggestion
      const mouseEventOriginatedFromWithinTheEditor = e.currentTarget === editor.getDoc().documentElement;
```

spocke: This won't work for inline editors. The events are then bound on the editor.getBody() does this drag/drop overrides work with inline mode should we branch out of it in inline mode.

spocke (changes requested):
Can this be tested, dispatch a fake event from outside?

hamza0867:
> Can this be tested, dispatch a fake event from outside?

I did test this by dispatching a fake event from outside 👍

modules/tinymce/src/core/test/ts/browser/DragDropOverridesTest.ts
```diff
@@ -257,6 +257,31 @@ describe('browser.tinymce.core.DragDropOverridesTest', () => {
...
+      const target = UiFinder.findIn(TinyDom.body(editor), 'p:contains("Draggable CEF")').getOrDie();
+      const initialScrollY = window.scrollY;
+      editor.getWin().scroll({
+        top: editor.getWin().innerHeight
+      });
+      Mouse.mouseDown(target);
+      window.dispatchEvent(new MouseEvent('mousemove', {
+        clientY: window.innerHeight,
+      }));
+      await Waiter.pWait(100);
```
spocke, line 280: Can this be a wait for waiter maybe with an interval of 100 if the setInterval of the scroll thing is 100. I always get a bit worried that this might misbehave if the timing is slightly off since the browser on the VM is slow.

spocke (approved):

modules/tinymce/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - New `expand` function added to `tinymce.dom.RangeUtils` to return a new range expanded around the nearest word. #TINY-9001
 
 ### Fixed
+- Dragging CEF element toward the bottom edge would cause the page to scroll up. #TINY-9025
```
spocke, line 14: Not sure that the audience knows what a CEF is? So maybe:

```suggestion
- Dragging a noneditable element toward the bottom edge would cause the page to scroll up. #TINY-9025
```

modules/tinymce/src/core/test/ts/browser/DragDropOverridesTest.ts
spocke: Right this is a waiter for something to now happen since it compares with the initial. So then I guess it's fine to have just a wait here since we don't know how long we need to vapid for that condition to not be true.

modules/tinymce/src/core/test/ts/browser/DragDropOverridesTest.ts
```diff
@@ -257,6 +257,31 @@ describe('browser.tinymce.core.DragDropOverridesTest', () => {
...
+      window.scroll({
+        top: window.innerHeight + 2000
+      });
+      const target = UiFinder.findIn(TinyDom.body(editor), 'p:contains("Draggable CEF")').getOrDie();
+      const initialScrollY = window.scrollY;
+      editor.getWin().scroll({
+        top: editor.getWin().innerHeight
+      });
+      Mouse.mouseDown(target);
+      window.dispatchEvent(new MouseEvent('mousemove', {
```
lorenzo-pomili, line 277: should be used `Mouse.mouseMoveTo` instead?

lorenzo-pomili (approved):

TheSpyder (approved):

modules/tinymce/src/core/test/ts/browser/DragDropOverridesTest.ts
hamza0867: No, `Mouse.mouseMoveTo` only works within the editor, not outside of it if I'm not mistaken 🤷‍♂️

lnewson: @hamza0867 you are incorrect there... anything Agar related can be used by anything. It's `mcagar` helpers that are editor specific.

PR #8155, opened 2022-09-27
TINY-9200: Fix broken tests on Firefox

spocke:
Related Ticket: TINY-9200

Description of Changes:
Tests started to fail since Firefox was updated to a newer version
* Changed the `ResizeTest.ts` and `TableSizeTest.ts` for snooker to have an inline block with exact pixels to expand the TD rather than just text that could cause issues with kerning, oses etc the intent of the test is still the same.
* Added a fix for Firefox scrollbars. By default Firefox 100+ on Linux has no scrollbars visible. Makes local testing a bit easier.
* Removed `source` from the children elements for the `AnnotationStylingTest.ts` test since it now return an empty string instead of a proper rgb value. It is a hidden element so it kind of makes sense to return nothing.
* Increased the fuzzy asserts from 2 to 4 to compensate for rendering changes in latest Firefox.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/sugar/src/test/ts/browser/LocationTest.ts
+++ b/modules/sugar/src/test/ts/browser/LocationTest.ts
@@ -136,7 +136,7 @@ UnitTest.asynctest('LocationTest', (success, failure) => {
     pos = SugarLocation.viewport(body);
     Assert.eq('', 0, pos.top);
     Assert.eq('', 0, pos.left);
-    const noVisibleScrollbarBrowser = platform.os.isMacOS() || (platform.browser.isFirefox() && platform.os.isLinux());
+    const noVisibleScrollbarBrowser = platform.os.isMacOS() || (platform.browser.isFirefox() && platform.os.isLinux()) && (platform.os.isWindows() && platform.os.version.major >= 11);
     Assert.eq('scroll bar width, got=' + scrollBarWidth, true, scrollBarWidth > 5 && scrollBarWidth < 50 || (noVisibleScrollbarBrowser && scrollBarWidth === 0));
   };
 
--- a/modules/sugar/src/test/ts/browser/LocationTest.ts
+++ b/modules/sugar/src/test/ts/browser/LocationTest.ts
@@ -136,8 +136,8 @@ UnitTest.asynctest('LocationTest', (success, failure) => {
     pos = SugarLocation.viewport(body);
     Assert.eq('', 0, pos.top);
     Assert.eq('', 0, pos.left);
-    const noScrollBrowser = platform.os.isMacOS() || (platform.browser.isFirefox() && platform.os.isLinux());
-    Assert.eq('scroll bar width, got=' + scrollBarWidth, true, scrollBarWidth > 5 && scrollBarWidth < 50 || (noScrollBrowser && scrollBarWidth === 0));
+    const noVisibleScrollbarBrowser = platform.os.isMacOS() || (platform.browser.isFirefox() && platform.os.isLinux());
+    Assert.eq('scroll bar width, got=' + scrollBarWidth, true, scrollBarWidth > 5 && scrollBarWidth < 50 || (noVisibleScrollbarBrowser && scrollBarWidth === 0));
   };
 
   const disconnectedChecks = () => {
--- a/modules/sugar/src/test/ts/browser/LocationTest.ts
+++ b/modules/sugar/src/test/ts/browser/LocationTest.ts
@@ -136,7 +136,7 @@ UnitTest.asynctest('LocationTest', (success, failure) => {
     pos = SugarLocation.viewport(body);
     Assert.eq('', 0, pos.top);
     Assert.eq('', 0, pos.left);
-    const noVisibleScrollbarBrowser = platform.os.isMacOS() || (platform.browser.isFirefox() && platform.os.isLinux()) && (platform.os.isWindows() && platform.os.version.major >= 11);
+    const noVisibleScrollbarBrowser = platform.os.isMacOS() || (platform.browser.isFirefox() && platform.os.isLinux()) || (platform.browser.isFirefox() && platform.os.isWindows() && platform.os.version.major >= 11);
     Assert.eq('scroll bar width, got=' + scrollBarWidth, true, scrollBarWidth > 5 && scrollBarWidth < 50 || (noVisibleScrollbarBrowser && scrollBarWidth === 0));
   };
 
--- a/modules/sugar/src/test/ts/browser/LocationTest.ts
+++ b/modules/sugar/src/test/ts/browser/LocationTest.ts
@@ -136,7 +136,8 @@ UnitTest.asynctest('LocationTest', (success, failure) => {
     pos = SugarLocation.viewport(body);
     Assert.eq('', 0, pos.top);
     Assert.eq('', 0, pos.left);
-    Assert.eq('scroll bar width, got=' + scrollBarWidth, true, scrollBarWidth > 5 && scrollBarWidth < 50 || (platform.os.isMacOS() && scrollBarWidth === 0));
+    const noScrollBrowser = platform.os.isMacOS() || (platform.browser.isFirefox() && platform.os.isLinux());
+    Assert.eq('scroll bar width, got=' + scrollBarWidth, true, scrollBarWidth > 5 && scrollBarWidth < 50 || (noScrollBrowser && scrollBarWidth === 0));
   };
 
   const disconnectedChecks = () => {
--- a/modules/sugar/src/test/ts/browser/ScrollTest.ts
+++ b/modules/sugar/src/test/ts/browser/ScrollTest.ts
@@ -127,7 +127,7 @@ UnitTest.asynctest('ScrollTest', (success, failure) => {
     const cX = Math.round(center.left);
     const cY = Math.round(center.top);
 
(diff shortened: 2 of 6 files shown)


modules/tinymce/src/models/dom/test/ts/module/table/TableTestUtils.ts
```diff
@@ -53,7 +53,7 @@ const assertWidth = (editor: Editor, elm: HTMLElement, expectedWidth: number | n
   if (expectedWidth === null) {
     assert.isNull(widthData.raw, `${nodeName} width should not be set`);
   } else {
-    assert.approximately(widthData.raw ?? -1, expectedWidth, 2, `${nodeName} width is ${expectedWidth} ~= ${widthData.raw}`);
+    assert.approximately(widthData.raw ?? -1, expectedWidth, 4, `${nodeName} width is ${expectedWidth} ~= ${widthData.raw}`);
```
hamza0867, line 56: This looks weird. I guess the width is off by 1 pixel in both right and left, which gives us off by 2 pixels. Right ?

hamza0867 (approved):

modules/sugar/src/test/ts/browser/LocationTest.ts
```diff
@@ -136,7 +136,8 @@ UnitTest.asynctest('LocationTest', (success, failure) => {
     pos = SugarLocation.viewport(body);
     Assert.eq('', 0, pos.top);
     Assert.eq('', 0, pos.left);
-    Assert.eq('scroll bar width, got=' + scrollBarWidth, true, scrollBarWidth > 5 && scrollBarWidth < 50 || (platform.os.isMacOS() && scrollBarWidth === 0));
+    const noVisibleScrollbarBrowser = platform.os.isMacOS() || (platform.browser.isFirefox() && platform.os.isLinux());
```
lnewson, line 139: This doesn't fix it for Firefox on Windows 11 as it also has no scrollbar. Should that be added now as well?

spocke: I don't have any way to test that do we have Windows 11 VMs somewhere?

lnewson: There is a build nodes for windows 11 yes... it needs you to either SSH though or use the `bedrock-windows-11` label instead from what I know.

lnewson: I think adding this should solve it though:
```ts
(platform.os.isWindows() && platform.os.version.major >= 11)
```

spocke: I guess I could just add it now and then if it breaks things on Win 11 we figure it out later.

lnewson: Oops sorry Spocke, I stuffed up my previous example as that would apply to all browsers. We just want firefox. I'll leave some suggestions.

modules/sugar/src/test/ts/browser/LocationTest.ts
```diff
@@ -136,7 +136,8 @@ UnitTest.asynctest('LocationTest', (success, failure) => {
     pos = SugarLocation.viewport(body);
     Assert.eq('', 0, pos.top);
     Assert.eq('', 0, pos.left);
-    Assert.eq('scroll bar width, got=' + scrollBarWidth, true, scrollBarWidth > 5 && scrollBarWidth < 50 || (platform.os.isMacOS() && scrollBarWidth === 0));
+    const noVisibleScrollbarBrowser = platform.os.isMacOS() || (platform.browser.isFirefox() && platform.os.isLinux()) && (platform.os.isWindows() && platform.os.version.major >= 11);
```
lnewson, line 139:
```suggestion
    const noVisibleScrollbarBrowser = platform.os.isMacOS() || (platform.browser.isFirefox() && platform.os.isLinux()) || (platform.browser.isFirefox() && platform.os.isWindows() && platform.os.version.major >= 11);
```

modules/sugar/src/test/ts/browser/ScrollTest.ts
```diff
@@ -127,7 +127,8 @@ UnitTest.asynctest('ScrollTest', (success, failure) => {
     const cX = Math.round(center.left);
     const cY = Math.round(center.top);
 
-    Assert.eq('scroll bar width, got=' + scrollBarWidth, true, scrollBarWidth > 5 && scrollBarWidth < 50 || (platform.os.isMacOS() && scrollBarWidth === 0));
+    const noVisibleScrollbarBrowser = platform.os.isMacOS() || (platform.browser.isFirefox() && platform.os.isLinux()) && (platform.os.isWindows() && platform.os.version.major >= 11);
```
lnewson, line 130:
```suggestion
    const noVisibleScrollbarBrowser = platform.os.isMacOS() || (platform.browser.isFirefox() && platform.os.isLinux()) || (platform.browser.isFirefox() && platform.os.isWindows() && platform.os.version.major >= 11);
```

modules/sugar/src/test/ts/browser/LocationTest.ts
lnewson: I fixed another bug here whereby the check used an `&&` instead of `||` for the different linux/windows checks.

ephox-mogran: Has this condition been determined empirically, or is it documented somewhere in a Firefox changelog?

modules/sugar/src/test/ts/browser/ScrollTest.ts
ephox-mogran: If this is the same as the other place, should we put it in agar?

modules/tinymce/src/core/test/ts/browser/annotate/AnnotationStylingTest.ts
```diff
@@ -97,7 +97,7 @@ describe('browser.tinymce.core.annotate.AnnotationStylingTest', () => {
 
       if (checkOtherNodes) {
         const parents = SelectorFilter.ancestors(elm, '*', (e) => Compare.eq(e, body));
-        const children = SelectorFilter.children(elm, '*');
+        const children = SelectorFilter.children(elm, '*:not(source)');
```
ephox-mogran, line 100: I think a descriptive comment explaining why we are excluding `source` tags here might help the future code reader.

modules/tinymce/src/models/dom/test/ts/module/table/TableTestUtils.ts
ephox-mogran: Maybe just put a comment explaining why `4` has been chosen (even if it's just: that's what it seemed to be based on actual data)

ephox-mogran (approved):

modules/sugar/src/test/ts/browser/LocationTest.ts
spocke: Ooops I should have read that code better as well.

spocke: Doing code changes while on a meeting and listening one one ear is probably not a good this. Will fix this now.

modules/sugar/src/test/ts/browser/LocationTest.ts
spocke: It's documented. However we should probably feature detect this instead since it's depends on settings in Firefox or the OS. But that is a later problem.

modules/sugar/src/test/ts/browser/ScrollTest.ts
spocke: We should have some form of scroll width feature detection in sand/agar/sugar not sure where it fits I think I've seen code in sugar that measures scrollbar widths.

modules/tinymce/src/core/test/ts/browser/annotate/AnnotationStylingTest.ts
spocke: Done

modules/tinymce/src/models/dom/test/ts/module/table/TableTestUtils.ts
spocke: It didn't have a comment for 2 either. Not 100% sure why 4 is needed might need some heavy investigation why the Firefox now is off by 2 more pixels. So not sure at this time what to put into the comment.

ltrouton (approved):

modules/tinymce/src/models/dom/test/ts/module/table/TableTestUtils.ts
ephox-mogran: K. Just putting something like "this 4 seems to be large enough to handle the different values we are seeing, but it's never been thoroughly investigated"

spocke: I added a comment and referencing the ticket added information on what failed and the weirdness around that in the JIRA ticket instead.

modules/sugar/src/test/ts/browser/LocationTest.ts
spocke: Oh bummer it seems that UA in FF on Win 11 is still 10. So we can't sniff out Windows 11 it seems.
I created a follow up ticket here: https://ephocks.atlassian.net/browse/TINY-9203 so make some form of feature detection. I added it as a training ticket since it seems small in scope and only affects tests.

PR #8157, opened 2022-09-28
TINY-9184: Set the separate color swatches

HAFRMO:
Related Ticket: 

Description of Changes:
Split cache and property, while keeping the color_map as the backup color for both.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/tinymce/src/themes/silver/main/ts/ui/core/color/Options.ts
```diff
@@ -86,16 +110,25 @@ const register = (editor: Editor): void => {
 
 const getColorCols = option('color_cols');
 const hasCustomColors = option('custom_colors');
-const getColors = option<Menu.ChoiceMenuItemSpec[]>('color_map');
+const getColors = (editor: Editor, id: string): Menu.ChoiceMenuItemSpec[] => {
```
spocke, line 113: This fallback behavior is not fully covered in tests. Setup an editor with foreground only see if background falls back. Setup with background only see if foreground falls back etc. You could mutate the options directly I guess. Though in theory these are immutable options.

spocke (changes requested):

modules/tinymce/src/themes/silver/main/ts/ui/core/color/ColorCache.ts
```diff
@@ -2,15 +2,14 @@ import { Arr, Type } from '@ephox/katamari';
...
-
 export interface ColorCache {
   readonly add: (key: string) => void;
   readonly state: () => string[];
 }
 
-export const ColorCache = (max: number = 10): ColorCache => {
-  const storageString = LocalStorage.getItem(storageName);
+export const ColorCache = (id: string, max: number = 10): ColorCache => {
+  const storageId = 'tinymce-custom-colors-' + id;
```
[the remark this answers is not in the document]
HAFRMO: I suppose we do.

modules/tinymce/src/themes/silver/main/ts/ui/core/color/Options.ts
```diff
@@ -7,7 +7,21 @@ import { Menu } from 'tinymce/core/api/ui/Ui';
...
+const getCatcheForId = (id: string) => {
+  let storage = cacheStorage[id];
+
+  if (Type.isNullable(storage)) {
+    storage = ColorCache(id, 10);
+    cacheStorage[id] = storage;
+  }
+
+  return storage;
+};
```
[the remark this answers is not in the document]
HAFRMO: I'm not sure this is more readable, and to me this wouldn't be a good way to go about it.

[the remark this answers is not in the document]
lnewson: This is definitely a safer option, as if you do `cacheStorage[id]` you run into all sorts of issues. Spocke's suggestion is how this is done generally so it makes sense to me.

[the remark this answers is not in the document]
lnewson: Another option that can be used now that IE 11 isn't supported anymore, that's still rarely used atm is a [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map).

modules/tinymce/src/themes/silver/main/ts/ui/core/color/Options.ts
HAFRMO: Yeah, I'm not keen on the idea of mutating these. Doubly so since some values are calculated beforehand, using the values that exist then, such as the cols ( even though those wouldn't change anything here )

modules/tinymce/src/themes/silver/main/ts/ui/core/color/Options.ts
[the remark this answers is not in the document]
HAFRMO: I'll just use the Obj.

modules/tinymce/src/themes/silver/main/ts/ui/core/color/Options.ts
```diff
@@ -84,18 +117,35 @@ const register = (editor: Editor): void => {
...
 };
 
-const getColorCols = option('color_cols');
+const getColorCols = (editor: Editor, id: string): number => {
+  if (id === foregroundId) {
+    return option('color_cols_foreground')(editor);
+  } else if (id === backgroundId) {
+    return option('color_cols_background')(editor);
+  } else {
+    return option('color_cols')(editor);
```
spocke, line 126: I think the `calcCols` need to move here getColorCols since it needs to compute different defaults depending on the format color map being used. So `getColorCols` could call getColors and do the calculation in the fallback.

HAFRMO: I'm not sure what you mean that it's needed to move here? It might be good to add a check that `color_map_background` or `color_map_foreground` are set before using those values, if we want them tied to those values, but the calculations should be just fine where they are?

modules/tinymce/src/themes/silver/main/ts/ui/core/color/ColorSwatch.ts
```diff
@@ -86,7 +86,7 @@ const applyColor = (editor: Editor, format: ColorFormat, value: string, onChoice
     const dialog = colorPickerDialog(editor);
     dialog((colorOpt) => {
       colorOpt.each((color) => {
-        Options.addColor(color);
+        Options.addColor(color, format);
```
spocke, line 89: This is flipped around. I think it should be `Options.addColor(format, color);`
I found this by adding a proper type to id making it `'forecolor' | 'hilitecolor' | 'default'` by changing all the IDs to that type.

So I think we should really consider making this a proper type than just a string. That would help out with branching as well forcing one to make sure that all variants are covered if you instead of doing if/else doing if on each of the cases.

modules/tinymce/src/themes/silver/main/ts/ui/core/color/Options.ts
```diff
@@ -84,18 +117,35 @@ const register = (editor: Editor): void => {
...
+const getCurrentColors = (id: string): Menu.ChoiceMenuItemSpec[] =>
+  Arr.map(getCatcheForId(id).state(), (color) => ({
+    type: 'choiceitem',
+    text: color,
+    value: color
+  }));
 
-const addColor = (color: string): void => {
-  colorCache.add(color);
+const addColor = (id: string, color: string): void => {
```
spocke, line 147: I don't think this should live in the options module. Since all it does is to add a color to the cache.

modules/tinymce/src/themes/silver/main/ts/ui/core/color/Options.ts
spocke: You are right. They fallback to the defaults for each option. So never mind. :+1:

spocke (changes requested):

modules/tinymce/src/themes/silver/main/ts/ui/core/color/Options.ts
```diff
@@ -1,13 +1,26 @@
...
 import { EditorOptions } from 'tinymce/core/api/OptionTypes';
 import { Menu } from 'tinymce/core/api/ui/Ui';
 
 import { ColorCache } from './ColorCache';
 
-const colorCache = ColorCache(10);
+const foregroundId = 'forecolor';
+const backgroundId = 'hilitecolor';
+
+interface CacheStorage {
```
spocke, line 13: I think this belongs in a CacheStore.ts module then the state getCacheForId, addColor could all live in there instead of this Options.ts. Separation of concerns.

modules/tinymce/src/themes/silver/main/ts/ui/core/color/ColorSwatch.ts
HAFRMO: It was intentionally set as string to allow us or users to reference these fully custom, since that'd help us add more, separate cache's in other places, something we've had a problem with before. Adding the type would effectively make that impossible.

spocke: Good point the cache could be on custom things I thought we only had bespoke color pickers but if they are to be used on table dialogs I guess then it could be anything. Either way the types found a bug so more tests are needed then to cover everything. In theory the id's could be a wrapped type but seems overkill of the feature is properly covered by tests.

HAFRMO: I'm surprised none of the existing tests caught this. Don't we have tests to check this?

spocke: Apparently not but the parameter order here is new code so should be added to this test.

modules/tinymce/src/themes/silver/main/ts/ui/core/color/ColorSwatch.ts
```diff
@@ -6,6 +6,7 @@ import { Dialog, Menu, Toolbar } from 'tinymce/core/api/ui/Ui';
 import { EditorEvent } from 'tinymce/core/api/util/EventDispatcher';
 
 import * as Events from '../../../api/Events';
+import { addColor, getCurrentColors } from './ColorCache';
```
lnewson, line 9: This should be using `import * as ColorCache from './ColorCache';` instead.

modules/tinymce/src/themes/silver/main/ts/ui/dialog/ColorInput.ts
```diff
@@ -10,8 +10,8 @@ import { UiFactoryBackstageShared } from '../../backstage/Backstage';
 import { UiFactoryBackstageForColorInput } from '../../backstage/ColorInputBackstage';
 import * as ReadOnly from '../../ReadOnly';
 import { renderLabel } from '../alien/FieldLabeller';
+import { addColor } from '../core/color/ColorCache';
```
lnewson, line 13: Likewise here, as we always prefer using star imports generally to make it more readable.

lorenzo-pomili (approved):

modules/bridge/src/main/ts/ephox/bridge/components/dialog/ColorInput.ts
```diff
@@ -1,17 +1,21 @@
-import { StructureSchema, ValueType } from '@ephox/boulder';
+import { FieldSchema, StructureSchema, ValueType } from '@ephox/boulder';
 import { Result } from '@ephox/katamari';
 
 import { FormComponentWithLabel, formComponentWithLabelFields, FormComponentWithLabelSpec } from './FormComponent';
 
 export interface ColorInputSpec extends FormComponentWithLabelSpec {
   type: 'colorinput';
+  swatchKey?: string;
```
lnewson, line 8: Does anyone know what `swatch` is here? It's a rather internal thing so I don't think this makes sense having thought about this more. I'd suggest just calling this either `key` or `storageKey`.

lnewson: I guess we do have `colorswatch` for the menuitem, but that doesn't make sense for `colorinput` though.

modules/tinymce/src/themes/silver/main/ts/ui/core/color/ColorCache.ts
```diff
@@ -1,16 +1,21 @@
...
 
-export const ColorCache = (max: number = 10): ColorCache => {
-  const storageString = LocalStorage.getItem(storageName);
+interface CacheStorage {
+  [index: string]: ColorCache;
+}
+const cacheStorage: CacheStorage = {};
+
+const ColorCache = (id: string, max: number = 10): ColorCache => {
+  const storageId = `tinymce-custom-colors-${id}`;
```
lnewson, line 17: This is a **breaking** change for end users so cannot be done... it needs to continue to use `tinymce-custom-colors` as the default key so that users custom colors continue to work when they upgrade.

lnewson (dismissed):

modules/tinymce/src/themes/silver/main/ts/ui/core/color/ColorCache.ts
TheSpyder: Yes I think prefer the new values, but if they don't exist read the old one and copy it no matter what the ID is. That way the value carries forward to both but can then diverge.

modules/tinymce/src/themes/silver/test/ts/browser/editor/color/ColorCacheTest.ts
```diff
@@ -0,0 +1,60 @@
+import { afterEach, describe, it } from '@ephox/bedrock-client';
+import { assert } from 'chai';
+
+import { Menu } from 'tinymce/core/api/ui/Ui';
+import LocalStorage from 'tinymce/core/api/util/LocalStorage';
+import * as ColorCache from 'tinymce/themes/silver/ui/core/color/ColorCache';
+
+describe('browser.tinymce.themes.silver.editor.color.ColorCatcheTest', () => {
```
lnewson, line 8:
```suggestion
describe('browser.tinymce.themes.silver.editor.color.ColorCacheTest', () => {
```
Minor typo here.

modules/tinymce/src/themes/silver/main/ts/ui/core/color/ColorCache.ts
```diff
@@ -47,3 +51,39 @@ export const ColorCache = (max: number = 10): ColorCache => {
     state
   };
 };
+
+const getCatcheForId = (id: string): ColorCache =>
```
spocke, line 55:
```suggestion
const getCacheForId = (id: string): ColorCache =>
```

modules/tinymce/src/themes/silver/test/ts/browser/editor/color/ColorCacheTest.ts
```diff
@@ -0,0 +1,60 @@
...
+      {
+        type: 'choiceitem',
+        text: '#0F00F0',
+        value: '#0F00F0'
+      }
+    ];
+    assert.deepEqual(colors, expectedResult, 'Color cache should have the correct entry');
+  });
+
+  it('TINY-9184: A cache should use not be tied to the legacy cache', () => {
```
spocke, line 32: Maybe the intent was?

```suggestion
  it('TINY-9184: A cache should not be tied to the legacy cache', () => {
```

modules/tinymce/src/themes/silver/test/ts/browser/editor/color/TextColorSanityTest.ts
```diff
@@ -88,4 +90,27 @@ describe('browser.tinymce.themes.silver.editor.color.TextColorSanityTest', () =>
...
+    TinyUiActions.clickOnUi(editor, 'button[title="Custom color"]');
+    const dialog = await TinyUiActions.pWaitForDialog(editor);
+    const input = UiFinder.findIn<HTMLInputElement>(dialog, 'label:contains("R") + input').getOrDie();
+    UiControls.setValue(input, '255');
+    const evt = new Event('input', {
+      bubbles: true,
+      cancelable: true
+    });
+    input.dom.dispatchEvent(evt);
+    await Waiter.pWait(500);
```
spocke, line 109: Why is the wait needed here is the input some how async? Could you instead await a change of color on the preview or the hex representation to reduce the possibility of a flake here?

modules/tinymce/src/themes/silver/test/ts/browser/editor/color/TextCustomColorMapBackgroundColorSanityTest.ts
```diff
@@ -0,0 +1,46 @@
+import { UiFinder } from '@ephox/agar';
+import { Assert, describe, it } from '@ephox/bedrock-client';
+import { SugarShadowDom } from '@ephox/sugar';
+import { TinyDom, TinyHooks, TinySelections, TinyUiActions } from '@ephox/wrap-mcagar';
+
+import Editor from 'tinymce/core/api/Editor';
+import { getColorCols } from 'tinymce/themes/silver/ui/core/color/Options';
+
+describe('browser.tinymce.themes.silver.editor.color.TextCustomColorMapBackgroundColorSanityTest', () => {
```
spocke, line 9: These three test are so similar I think they should be combined into one test file. You can just have three contexts with the hook in them so that you can share the `setupContent` and `assertUiElementDoesNotExist` between them. So stick them all into `TextCustomColorMapSanityTest.ts`?

modules/tinymce/CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - New `expand` function added to `tinymce.selection` which expands the selection around the nearest word. #TINY-9001
 - New `expand` function added to `tinymce.dom.RangeUtils` to return a new range expanded around the nearest word. #TINY-9001
+- New `color_map_background` option which sets the default colors in the background highlight colorswatch #TINY-9184
+- New `color_map_foreground` option which sets the default colors in the textcolor colorswatch #TINY-9184
```
spocke, line 13:
```suggestion
- New `color_map_background` option which sets the default colors in the background highlight colorswatch. #TINY-9184
- New `color_map_foreground` option which sets the default colors in the textcolor colorswatch. #TINY-9184
```

spocke (approved):

modules/tinymce/src/themes/silver/test/ts/browser/editor/color/TextCustomColorMapBackgroundColorSanityTest.ts
HAFRMO: Can't, they use different color setups, which includes the col-calculations, which seems to be have been done ahead of time. Unless I have completely misunderstood something, making these tests requires them to be separate.

modules/tinymce/src/themes/silver/test/ts/browser/editor/color/TextColorSanityTest.ts
HAFRMO: I don't know exactly why it's needed, but it made things work. I wasn't too concerned about exactly why since it's honestly unrelated to this issue, and I was concerned that trying to wait for the change with some sort of waiter might take some time to implement, since it's in a dialog which requires extra fiddling around, at least for me.

modules/tinymce/src/themes/silver/test/ts/browser/editor/color/ColorCacheTest.ts
HAFRMO: Yes, deleting a word is hard for me, it seems.

modules/tinymce/src/themes/silver/test/ts/browser/editor/color/TextCustomColorMapBackgroundColorSanityTest.ts
spocke: Seems to work fine to combine them. The cols are calculated on registry of the option that is done on setup of a editor instance so as long as the configs and editor setup are within it's own context in the test they should be treated as they where separate files. The file separation is irrelevant even the test framework doesn't even know that there are multiple files all it sees is a long list of describes, context and it calls in a `scratch/compiled/tests-imports.ts` file bundle.

modules/tinymce/src/themes/silver/test/ts/browser/editor/color/TextColorSanityTest.ts
spocke: Please switch to a waiter if we can. Since if we don't know why it's needed then we might need to know why also if we start sprinkling 500ms waits all over the tests then they start going slow when they start adding up. They already take tons of time to finish we shouldn't add more time to it just because we don't know why something needs to be async. If it's to hard to wait for a state then fine we could use a wait call here but then a comment should be added to why it needs to be there or a follow up ticket reference or something if it's really unknown why it needs to be there there are a few such cases in other parts of the codebase.

lnewson:
Similar to https://github.com/tinymce/tinymce/pull/8165/files#r990853904, please make sure to work with @hamza0867 to make sure the option names are consistent.

- [ ]  Ensure the option names are consistent.

modules/tinymce/src/themes/silver/test/ts/browser/editor/color/TextColorSanityTest.ts
lnewson: - [ ] Combine the tests using `context` to remove duplication

modules/tinymce/src/themes/silver/test/ts/browser/editor/color/TextCustomColorMapBackgroundColorSanityTest.ts
lnewson: - [ ] Combine the tests using `context` to remove duplication

HAFRMO: Putting them physically into the same file would be possible, but we have historically frowned upon putting what is, effectively, different tests in the same file. But very well, I'll do it since it was requested.

lnewson: I'd disagree with your statement there... I can't speak for Spocke, but from what he's said and what I've read this appears to be the same tests just with different options. In that case we do try and keep it together as it's more maintainable since it's all running via the same code.

modules/tinymce/src/themes/silver/test/ts/browser/editor/color/TextColorSanityTest.ts
TheSpyder: Agreed

- [ ] Investigate why a 500ms wait works around the problem

TheSpyder (approved):

modules/bridge/src/main/ts/ephox/bridge/components/dialog/ColorInput.ts
```diff
@@ -1,17 +1,21 @@
-import { StructureSchema, ValueType } from '@ephox/boulder';
+import { FieldSchema, StructureSchema, ValueType } from '@ephox/boulder';
 import { Result } from '@ephox/katamari';
 
 import { FormComponentWithLabel, formComponentWithLabelFields, FormComponentWithLabelSpec } from './FormComponent';
 
 export interface ColorInputSpec extends FormComponentWithLabelSpec {
   type: 'colorinput';
+  storageKey?: string;
```
lnewson, line 8: This needs a bridge changelog entry as well.

- [ ] Add a bridge changelog entry

modules/tinymce/CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - New `expand` function added to `tinymce.selection` which expands the selection around the nearest word. #TINY-9001
 - New `expand` function added to `tinymce.dom.RangeUtils` to return a new range expanded around the nearest word. #TINY-9001
+- New `color_map_background` option which sets the default colors in the background highlight colorswatch. #TINY-9184
+- New `color_map_foreground` option which sets the default colors in the textcolor colorswatch. #TINY-9184
```
lnewson, line 13: In @hamza0867 now merged PR he combined these based on something @ltrouton said which makes sense to me, so this matches that (also removes some confusion around `default` related to the other PR).
```suggestion
- New `color_map_background` and `color_map_foreground` option which sets the base colors used in the `backcolor` and `forecolor` toolbar buttons and menu items. #TINY-9184
```

modules/tinymce/CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
```
lnewson, line 9: There doesn't appear to be any changelog entry for the new `storageKey` option either. Is that expected as it's a public API?

lnewson (commented):
Just a couple other things that have been missed here I saw from other comments lately sorry.

modules/tinymce/CHANGELOG.md
HAFRMO: Sure, I'll add an entry for that.

lnewson: @HAFRMO you appear to have missed this...

PR #8161, opened 2022-10-03
TINY-9102: Added exactAttrs, exactClasses and exactStyles to ApproxStructures.element

shanmen-tiny:
Related Ticket: TINY-9102

Description of Changes:
* Added `elementWithSpecifics` to `structApi` (`ApproxStructure`) for strict comparison of styles, attributes and classes

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/agar/src/main/ts/ephox/agar/assertions/ApproxStructures.ts
+++ b/modules/agar/src/main/ts/ephox/agar/assertions/ApproxStructures.ts
@@ -298,7 +298,7 @@ const assertClassesSpecific = (expectedClasses: string[], actual: SugarElement<E
   );
 
   if (!isClassesEqual) {
-    throw new Error('Attribute names were not matching. actual: ' + actualClasses.join(', ') + ', expected: ' + expectedClasses.join(', '));
+    throw new Error('Class names were not matching. actual: ' + actualClasses.join(', ') + ', expected: ' + expectedClasses.join(', '));
   }
 };
 
--- a/modules/agar/src/main/ts/ephox/agar/assertions/ApproxStructures.ts
+++ b/modules/agar/src/main/ts/ephox/agar/assertions/ApproxStructures.ts
@@ -46,6 +46,10 @@ export interface ElementFields {
   children?: StructAssert[];
 }
 
+export interface ElementWithSpecificFields extends Omit<ElementFields, 'classes'> {
+  classes?: string[];
+}
+
 const elementQueue = (items: SugarElement<Node>[], container: Optional<SugarElement<Node>>): ElementQueue => {
   let i = -1;
 
@@ -112,6 +116,33 @@ const element = (tag: string, fields: ElementFields): StructAssert => {
       assertAttrs(attrs, actual);
       assertClasses(classes, actual);
       assertStyles(styles, actual);
+      assertHtml(html, actual);
+      assertValue(value, actual);
+
+      assertChildren(children, actual);
+    } else {
+      Assert.eq('Incorrect node type for: ' + Truncate.getHtml(actual), 1, SugarNode.type(actual));
+    }
+  };
+
+  return {
+    doAssert
+  };
+};
+
+const elementWithSpecifics = (tag: string, fields: ElementWithSpecificFields): StructAssert => {
+  const doAssert = (actual: SugarElement<Node>): void => {
+    if (SugarNode.isHTMLElement(actual)) {
+      Assert.eq(() => 'Incorrect node name for: ' + Truncate.getHtml(actual), tag, SugarNode.name(actual));
+      const attrs = fields.attrs !== undefined ? fields.attrs : {};
+      const classes = fields.classes !== undefined ? fields.classes : [];
+      const styles = fields.styles !== undefined ? fields.styles : {};
+      const html = fields.html !== undefined ? Optional.some(fields.html) : Optional.none<StringAssert>();
+      const value = fields.value !== undefined ? Optional.some(fields.value) : Optional.none<StringAssert>();
+      const children = fields.children !== undefined ? Optional.some(fields.children) : Optional.none<StructAssert[]>();
+      assertAttrs(attrs, actual, true);
+      assertClassesSpecific(classes, actual);
+      assertStyles(styles, actual, true);
       assertHtml(html, actual);
       assertValue(value, actual);
 
@@ -221,7 +252,21 @@ const anythingStruct: StructAssert = {
   doAssert: Fun.noop
 };
 
-const assertAttrs = (expectedAttrs: Record<string, StringAssert>, actual: SugarElement<Element>) => {
+const assertAttrs = (expectedAttrs: Record<string, StringAssert>, actual: SugarElement<Element>, specificMode: boolean = false) => {
+  if (specificMode) {
+    const allDefinedAttrs = Obj.keys(expectedAttrs);
+    const actualDefinedAttrs = Obj.keys(Obj.bifilter(Attribute.clone(actual), (_value, key) => !Arr.contains([ 'style', 'class' ], key)).t);
+
+    const isAttrsEqual = Arr.equal(
+      Arr.sort(allDefinedAttrs),
+      Arr.sort(actualDefinedAttrs)
+    );
+
+    if (!isAttrsEqual) {
+      throw new Error('Attribute names were not matching. actual: ' + actualDefinedAttrs.join(', ') + ', expected: ' + allDefinedAttrs.join(', '));
+    }
+  }
+
   Obj.each(expectedAttrs, (v, k) => {
     if (v.strAssert === undefined) {
       throw new Error(JSON.stringify(v) + ' is not a *string assertion*.\nSpecified in *expected* attributes of ' + Truncate.getHtml(actual));
@@ -244,7 +289,33 @@ const assertClasses = (expectedClasses: ArrayAssert[], actual: SugarElement<Elem
   });
 };
 
-const assertStyles = (expectedStyles: Record<string, StringAssert>, actual: SugarElement<Element>) => {
+const assertClassesSpecific = (expectedClasses: string[], actual: SugarElement<Element>) => {
+  const actualClasses = Classes.get(actual);
+
+  const isClassesEqual = Arr.equal(
+    Arr.sort(actualClasses),
+    Arr.sort(expectedClasses)
+  );
+
+  if (!isClassesEqual) {
+    throw new Error('Attribute names were not matching. actual: ' + actualClasses.join(', ') + ', expected: ' + expectedClasses.join(', '));
+  }
+};
+
+const assertStyles = (expectedStyles: Record<string, StringAssert>, actual: SugarElement<Element>, specificMode: boolean = false) => {
+  if (specificMode) {
+    const allDefinedStyles = Obj.keys(expectedStyles);
+    const actualDefinedStyles = Obj.keys(Css.getAllRaw(actual));
+    const isStylesEqual = Arr.equal(
+      Arr.sort(allDefinedStyles),
+      Arr.sort(actualDefinedStyles)
+    );
+
+    if (!isStylesEqual) {
+      throw new Error('Style names were not matching. actual: ' + actualDefinedStyles.join(', ') + ', expected: ' + allDefinedStyles.join(', '));
+    }
+  }
+
   Obj.each(expectedStyles, (v, k) => {
     const actualValue = Css.getRaw(actual, k).getOrThunk(ApproxComparisons.missing);
     if (v.strAssert === undefined) {
@@ -311,6 +382,7 @@ export {
   elementQueue,
   anything,
   element,
+  elementWithSpecifics,
   text,
   either,
   repeat,
--- a/modules/agar/src/main/ts/ephox/agar/assertions/ApproxStructures.ts
+++ b/modules/agar/src/main/ts/ephox/agar/assertions/ApproxStructures.ts
@@ -254,7 +254,7 @@ const assertAttrs = (expectedAttrs: Record<string, StringAssert>, actual: SugarE
 
 const assertExactMatchAttrs = (expectedAttrs: Record<string, StringAssert>, actual: SugarElement<Element>) => {
   const allDefinedAttrs = Obj.keys(expectedAttrs);
-  const actualDefinedAttrs = Obj.keys(Obj.bifilter(Attribute.clone(actual), (_value, key) => !Arr.contains([ 'style', 'class' ], key)).t);
+  const actualDefinedAttrs = Arr.filter(Obj.keys(Attribute.clone(actual)), (attr) => attr !== 'class' && attr !== 'style');
 
   const isEqual = assertEqualArray(
     allDefinedAttrs,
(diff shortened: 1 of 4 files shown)


modules/agar/src/main/ts/ephox/agar/assertions/ApproxStructures.ts
```diff
@@ -221,7 +252,21 @@ const anythingStruct: StructAssert = {
   doAssert: Fun.noop
 };
 
-const assertAttrs = (expectedAttrs: Record<string, StringAssert>, actual: SugarElement<Element>) => {
+const assertAttrs = (expectedAttrs: Record<string, StringAssert>, actual: SugarElement<Element>, specificMode: boolean = false) => {
```
ephox-mogran, line 255: Generally, try to avoid passing through a boolean that causes branching in the implementation. If you can write the new assertion function by doing a step first, then doing the rest, it would be better to make a separate function, do the additional check first, then call the other one. If it causes some efficiency, you can make a shared function that gets passed the `allDefinedAttrs` or something similar

modules/agar/src/main/ts/ephox/agar/assertions/ApproxStructures.ts
```diff
@@ -244,7 +289,33 @@ const assertClasses = (expectedClasses: ArrayAssert[], actual: SugarElement<Elem
...
+    Arr.sort(actualClasses),
+    Arr.sort(expectedClasses)
+  );
+
+  if (!isClassesEqual) {
+    throw new Error('Class names were not matching. actual: ' + actualClasses.join(', ') + ', expected: ' + expectedClasses.join(', '));
+  }
+};
+
+const assertStyles = (expectedStyles: Record<string, StringAssert>, actual: SugarElement<Element>, specificMode: boolean = false) => {
```
ephox-mogran, line 305: See comment about `assertAttrs`

modules/agar/src/main/ts/ephox/agar/assertions/ApproxStructures.ts
```diff
@@ -126,6 +130,33 @@ const element = (tag: string, fields: ElementFields): StructAssert => {
   };
 };
 
+const elementWithSpecifics = (tag: string, fields: ElementWithSpecificFields): StructAssert => {
```
MitchCTiny, line 133: The name `ElementWithSpecifics` is vague, maybe a name reflecting that it exactly matches classes would be better.

MitchCTiny (approved):

modules/agar/src/main/ts/ephox/agar/assertions/ApproxStructures.ts
```diff
@@ -234,6 +265,22 @@ const assertAttrs = (expectedAttrs: Record<string, StringAssert>, actual: SugarE
...
+  const actualDefinedAttrs = Obj.keys(Obj.bifilter(Attribute.clone(actual), (_value, key) => !Arr.contains([ 'style', 'class' ], key)).t);
+
+  const isEqual = Arr.equal(
+    Arr.sort(allDefinedAttrs),
+    Arr.sort(actualDefinedAttrs)
+  );
+
+  if (!isEqual) {
+    throw new Error('Attribute names were not matching. actual: ' + actualDefinedAttrs.join(', ') + ', expected: ' + allDefinedAttrs.join(', '));
+  }
```
MitchCTiny, line 279: If this sorting then asserting is identical functionality for classes, styles, and attributes, perhaps a helper function which does this would be more suitable.
```suggestion
const assertEqualArrays = (expected: T[], actual: T[]): boolean =>  Arr.equal(
  Arr.sort(expected),
  Arr.sort(actual)
);
```

modules/agar/src/test/ts/browser/ApproxStructureTest.ts
```diff
@@ -1,139 +1,426 @@
...
+            }))
+          ]
+        }));
+
+      check(struct2, '<div><span class="hello"></span><div></div></div>');
+      check(struct2, '<div><span class="hello"></span><div></div><span class="bye"></span></div>');
+      check(struct2, '<div><span class="hello"></span><span class="hello"></span><span class="hello"></span><span class="hello"></span><div></div><span class="bye"></span></div>');
+    });
+
+    it('TINY-9102: ApproxStructure text combineSiblings', () => {
```
tiny-ben-tran, line 249: Not sure what combineSiblings means here

tiny-ben-tran (approved):

modules/agar/src/main/ts/ephox/agar/assertions/ApproxStructures.ts
shanmen-tiny: Moved this to separate functions.

The specific checking function uses the key from `allDefinedAttrs`, but the general function retrieves the key and value from `allDefinedAttrs`. The difference is the value is used for string/array assertions.

modules/agar/src/main/ts/ephox/agar/assertions/ApproxStructures.ts
```diff
@@ -46,6 +46,10 @@ export interface ElementFields {
   children?: StructAssert[];
 }
 
+export interface ElementWithSpecificFields extends Omit<ElementFields, 'classes'> {
+  classes?: string[];
+}
```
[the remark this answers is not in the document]
shanmen-tiny: Sorry correct me if I'm wrong; if we added the `string[]` type to the `ElementFields` interface, we would need to do type assertion, ex: `fields.classes as string[]` or `fields.classes as ArrayAssert[]` and pass it to the class asserting functions is that fine?

modules/agar/src/main/ts/ephox/agar/assertions/ApproxStructures.ts
shanmen-tiny: Not sure if `elementWithExactMatch`, and the rest `assertExactMatchClasses` sounds better?

modules/agar/src/main/ts/ephox/agar/assertions/ApproxStructures.ts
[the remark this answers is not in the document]
ephox-mogran: The problem I have with changing `ElementFields` is it makes it *much* easier to write the wrong thing. The idea of a separate function is to be explicit about the intent: for this structure, I *care* about the classes being only these. If you made `ElementFields` support either, then you get into the situation where if the API is poorly understood, and `str` assertions versus a string is misunderstood, they might just write strings because they assumed that's what they should do, and the compiler didn't complain.

modules/agar/src/main/ts/ephox/agar/assertions/ApproxStructures.ts
ephox-mogran: It also matches styles and attributes. Maybe using `exact` is better, as you say, @shanmen-tiny

modules/agar/src/main/ts/ephox/agar/assertions/ApproxStructures.ts
```diff
@@ -234,6 +265,22 @@ const assertAttrs = (expectedAttrs: Record<string, StringAssert>, actual: SugarE
   });
 };
 
+const assertAttrsWithSpecifics = (expectedAttrs: Record<string, StringAssert>, actual: SugarElement<Element>) => {
+  const allDefinedAttrs = Obj.keys(expectedAttrs);
+  const actualDefinedAttrs = Obj.keys(Obj.bifilter(Attribute.clone(actual), (_value, key) => !Arr.contains([ 'style', 'class' ], key)).t);
+
+  const isEqual = Arr.equal(
```
ephox-mogran, line 272: I thought I saw this comment elsewhere, but it's probably worth making a shared function here for asserting sorted arrays

modules/agar/src/test/ts/browser/ApproxStructureTest.ts
```diff
@@ -1,139 +1,426 @@
...
+    it('TINY-9102: ApproxStructure theRest', () => {
+      check(ApproxStructure.build((s, str, _arr) =>
+        s.element('div', {
+          children: [
+            s.element('div', {
+              attrs: {
+                selected: str.is('true')
+              }
+            }),
+            s.theRest()
```
ephox-mogran, line 78: Interesting. I didn't even know this was a thing.

modules/agar/src/test/ts/browser/ApproxStructureTest.ts
ephox-mogran: Looks like it means that adjacent text nodes can be normalised into one in the expected.

modules/agar/src/test/ts/browser/ApproxStructureTest.ts
```diff
@@ -1,139 +1,426 @@
...
+          classes: [
+            'test1', 'root', 'non-existent-class'
+          ],
+          styles: {
+            display: str.is('block')
+          },
+        })), html);
+    });
+
+    it('TINY-9102: ApproxStructure throws error when there are extra styles', () => {
```
ephox-mogran, line 313: It looks like we don't have any tests to check that if there aren't enough classes/styles/attributes, it still fails. Is that correct?

modules/agar/src/main/ts/ephox/agar/assertions/ApproxStructures.ts
spocke: Could it be separate properties instead for the normal element. Like `exactAttributes`, `exactClasses` then just error out of both `attributes` and `exactClasses` are defined. The type system could have both one with and one without so might not need proper validation?

spocke: Or maybe it's fine to mix them having like `element('div', { exactClasses: [ str.is('foo') ], attributes: { bar: str.is('bar') } } )`

ephox-mogran: That's a good point that you might want to mix them. I'm wary of using runtime errors though to avoid defining both, though, so there isn't really a good solution for that. There's arguments both ways. I like the separate API approach, but I didn't think that you might want to mix and match them in the same element. I think the key thing that I care about is the decision has to be intentional. I don't want something that just detects the type you use, and acts accordingly, because it will be too easy to write the wrong thing.

modules/agar/src/main/ts/ephox/agar/assertions/ApproxStructures.ts
```diff
@@ -126,6 +130,33 @@ const element = (tag: string, fields: ElementFields): StructAssert => {
   };
 };
 
+const elementWithSpecifics = (tag: string, fields: ElementWithSpecificFields): StructAssert => {
+  const doAssert = (actual: SugarElement<Node>): void => {
+    if (SugarNode.isHTMLElement(actual)) {
+      Assert.eq(() => 'Incorrect node name for: ' + Truncate.getHtml(actual), tag, SugarNode.name(actual));
+      const attrs = fields.attrs !== undefined ? fields.attrs : {};
+      const classes = fields.classes !== undefined ? fields.classes : [];
+      const styles = fields.styles !== undefined ? fields.styles : {};
```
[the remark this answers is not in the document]
shanmen-tiny: Thanks, done.

modules/agar/src/main/ts/ephox/agar/assertions/ApproxStructures.ts
shanmen-tiny: Added a shared function here for asserting sorted arrays.

modules/agar/src/test/ts/browser/ApproxStructureTest.ts
shanmen-tiny: Thanks, yeah, I have just added it after you commented. It should also throw errors when there are fewer classes/styles/attributes.

modules/agar/src/main/ts/ephox/agar/assertions/ApproxStructures.ts
lnewson: I'd agree with Spocke here that having it be part of the existing API with specific keys is probably the better way here. While you might accidentally combine `classes` with `exactClasses` we basically already have that problem today anyways with `attributes` and `classes`/`styles` as you can mix and match them.

ephox-mogran: Fair enough. You've both convinced me. Sorry @shanmen-tiny , I'll make some alterations to the ticket. Most of your approach is fine ... it's just that it will be all be part of `s.element` instead of a separate API. It also avoids having to come up with better names as well, as `exactClasses` and `exactStyles` etc. are pretty self-explanatory.

For now (unless anyone strongly objects, we won't have a runtime error if both variants are provided, but the *exact* variants will take priority.

ephox-mogran: I've updated the ticket.

modules/agar/src/main/ts/ephox/agar/assertions/ApproxStructures.ts
```diff
@@ -257,6 +304,22 @@ const assertStyles = (expectedStyles: Record<string, StringAssert>, actual: Suga
...
+  const allDefinedStyles = Obj.keys(expectedStyles);
+  const actualDefinedStyles = Obj.keys(Css.getAllRaw(actual));
+
+  const isEqual = assertEqualArray(
+    allDefinedStyles,
+    actualDefinedStyles
+  );
+
+  if (!isEqual) {
+    throw new Error('Style names were not matching. actual: ' + actualDefinedStyles.join(', ') + ', expected: ' + allDefinedStyles.join(', '));
```
ltrouton, line 317: Would it be clearer to have wrapping `[]` in the string to better distinguish between actual and expected values?
It might actually be better to use a template string
```suggestion
    throw new Error(`Style names were not matching. actual: [${actualDefinedStyles.join(', ')}], expected: [{allDefinedStyles.join(', '))}]`;
```

ltrouton (approved):

modules/agar/src/main/ts/ephox/agar/assertions/ApproxStructures.ts
```diff
@@ -234,6 +252,22 @@ const assertAttrs = (expectedAttrs: Record<string, StringAssert>, actual: SugarE
   });
 };
 
+const assertExactMatchAttrs = (expectedAttrs: Record<string, StringAssert>, actual: SugarElement<Element>) => {
+  const allDefinedAttrs = Obj.keys(expectedAttrs);
+  const actualDefinedAttrs = Obj.keys(Obj.bifilter(Attribute.clone(actual), (_value, key) => !Arr.contains([ 'style', 'class' ], key)).t);
```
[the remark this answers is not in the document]
shanmen-tiny: Thanks, applied the suggestions.

modules/agar/src/main/ts/ephox/agar/assertions/ApproxStructures.ts
shanmen-tiny: Yep, I applied the suggestion, clearer to read the actual and expected values now.

modules/agar/src/main/ts/ephox/agar/assertions/ApproxStructures.ts
shanmen-tiny: Thank you all for the suggestions. I have applied the changes of having the ability to mix and match different variants.

ephox-mogran (approved):

PR #8162, opened 2022-10-03
TINY-8886: Update prismjs

TheSpyder:
Related Ticket: TINY-8886

Description of Changes:
* Small version bump to include security patches
* We don't actually use the vulnerable code, but there's no harm in upgrading (primjs is [in maintenance mode](https://github.com/PrismJS/prism/blob/59e5a3471377057de1f401ba38337aca27b80e03/README.md?plain=1#L18-L20) until v2 is released)

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [x] Milestone set

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "@ephox/oxide-icons-tools": "^2.2.2",
     "@ephox/swag": "^4.4.0",
     "@ephox/wrap-jsverify": "^2.0.1",
-    "@ephox/wrap-prismjs": "^1.25.0",
+    "@ephox/wrap-prismjs": "^1.29.1-patchfix.3558",
     "@ephox/wrap-promise-polyfill": "^2.2.1",
     "@tinymce/eslint-plugin": "^1.9.1",
     "@tinymce/moxiedoc": "^0.1.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -305,10 +305,10 @@
   resolved "https://registry.yarnpkg.com/@ephox/wrap-jsverify/-/wrap-jsverify-2.0.1.tgz#8984b292eeab3bd570eab773541650d257793d4d"
   integrity sha512-RSSsUFBV9uVUPzmCRvYXZBEZhVahYXs/j0JVgCy0CrbS98nV60BeYd97QB37sr05Lv7haLWzKf9XAcVAO7M8Ng==
 
-"@ephox/wrap-prismjs@^1.25.0":
-  version "1.29.0"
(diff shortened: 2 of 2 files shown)


TheSpyder:
Looks like the upgrade doesn't work on IE11. Should we abandon this?

spocke:
This one is concerning. It breaks on IE with an invalid regexp error. So maybe the DOS security fix uses some new regexp feature to avoid catastrophic backtracking haven't looked. Was to much code changes between these two prism versions: https://github.com/PrismJS/prism/compare/v1.25.0...v1.29.0

TheSpyder:
Yeah we're working on patching it in `wrap-prismjs`, the bug isn't hard to find. But the obvious fix didn't work so now I'll have to dig into it.

spocke:
I think it's just that the 5.10 branch doesn't have the `postinstall` script for the patch-package it's in develop but not in the 5.10 branch.

TheSpyder:
I thought that was automatic 🤔 thanks!

MitchCTiny (approved):

ltrouton (approved):

patches/@ephox+wrap-prismjs+1.29.0.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/node_modules/@ephox/wrap-prismjs/src/main/js/ephox/wrap-prismjs/PrismJS.js b/node_modules/@ephox/wrap-prismjs/src/main/js/ephox/wrap-prismjs/PrismJS.js
+index a0eed9a..60075a8 100644
+--- a/node_modules/@ephox/wrap-prismjs/src/main/js/ephox/wrap-prismjs/PrismJS.js
++++ b/node_modules/@ephox/wrap-prismjs/src/main/js/ephox/wrap-prismjs/PrismJS.js
+@@ -220,7 +220,7 @@ var exports = {}, module = { exports: exports }, global = {};
+ 				setLanguage: function (element, language) {
+ 					// remove all `language-xxxx` classes
+ 					// (this might leave behind a leading space)
+-					element.className = element.className.replace(RegExp(lang, 'gi'), '');
```
ephox-mogran, line 9: I don't think I understand what is going on here? Do we own this code? If so, why are we patching it? Can't we just make another version? Or is there an issue with version number syncing with the original prism?

TheSpyder: Ah thanks for the poke. This is indeed incorrect. I need to take this patch and apply it to prism inside `wrap-prismjs` directly, `contact` already has patching built into it. The application here was just to avoid testing on IE11 myself.

tiny-ben-tran (approved):

TheSpyder:
ok, patch removed and this PR is just a simple dependency update. Once it builds I'll merge it.

PR #8171, opened 2022-10-05
TINY-9172: Implement block level links

spocke:
Related Ticket: TINY-9172

Description of Changes:
* Transparent elements are now accepted at editor root
* Transparent elements are now allowed to contain other blocks
* Transparent elements in the root gets annotated with `data-mce-block` if they have blocks
* Transparent elements that contain blocks gets annotated with `data-mce-block`
* Lots of new tests for this new behavior of transparent elements

This is just the first batch known issues have been logged.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -15,10 +15,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - New `color_default_foreground` and `color_default_background` options to set the initial default color for the `forecolor` and `backcolor` toolbar buttons and menu items. #TINY-9183
 
 ### Changed
-- Transparent elements like anchors are now allowed in the root of the editor body. #TINY-9172
+- Transparent elements, like anchors, are now allowed in the root of the editor body if they contain blocks. #TINY-9172
 
 ### Improved
-- Transparent elements like anchors can now contain block elements. #TINY-9172
+- Transparent elements, like anchors, can now contain block elements. #TINY-9172
 
 ### Fixed
 - Dragging a noneditable element toward the bottom edge would cause the page to scroll up. #TINY-9025
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - New `expand` function added to `tinymce.selection` which expands the selection around the nearest word. #TINY-9001
 - New `expand` function added to `tinymce.dom.RangeUtils` to return a new range expanded around the nearest word. #TINY-9001
+- New `getTransparentElements` function added to `tinymce.html.Schema` to return a map object of transparent HTML elements. #TINY-9172
 
 ### Fixed
 - Dragging a noneditable element toward the bottom edge would cause the page to scroll up. #TINY-9025
@@ -17,6 +18,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Line separator scrolling in floating toolbars. #TINY-8948
 - A double bottom border appeared on inline mode editor for the `tinymce-5` skin. #TINY-9108
 - The editor header showed up even with no menubar and toolbar configured. #TINY-8819
+
+### Changed
+- Transparent elements like anchors are now allowed in the root of the editor body. #TINY-9172
+- Transparent elements like anchors can now contain block elements. #TINY-9172
 
 ## 6.2.0 - 2022-09-08
 
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Compilation errors were thrown when using TypeScript 4.8. #TINY-9161
 - Line separator scrolling in floating toolbars. #TINY-8948
 - A double bottom border appeared on inline mode editor for the `tinymce-5` skin. #TINY-9108
+- The editor header showed up even with no menubar and toolbar configured. #TINY-8819
 
 ## 6.2.0 - 2022-09-08
 
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -14,17 +14,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - New `ToggleView` command which makes it possible to hide or show registered custom views. #TINY-9210
 - New `color_default_foreground` and `color_default_background` options to set the initial default color for the `forecolor` and `backcolor` toolbar buttons and menu items. #TINY-9183
 
+### Changed
+- Transparent elements like anchors are now allowed in the root of the editor body. #TINY-9172
+
+### Improved
+- Transparent elements like anchors can now contain block elements. #TINY-9172
+
 ### Fixed
 - Dragging a noneditable element toward the bottom edge would cause the page to scroll up. #TINY-9025
 - Range expanding capabilities would behave inconsistently depending on where the cursor was placed. #TINY-9029
 - Compilation errors were thrown when using TypeScript 4.8. #TINY-9161
 - Line separator scrolling in floating toolbars. #TINY-8948
 - A double bottom border appeared on inline mode editor for the `tinymce-5` skin. #TINY-9108
-- The editor header showed up even with no menubar and toolbar configured. #TINY-8819
-
-### Changed
-- Transparent elements like anchors are now allowed in the root of the editor body. #TINY-9172
-- Transparent elements like anchors can now contain block elements. #TINY-9172
 
 ## 6.2.0 - 2022-09-08
 
--- a/modules/tinymce/src/core/main/ts/api/html/DomParser.ts
+++ b/modules/tinymce/src/core/main/ts/api/html/DomParser.ts
@@ -14,7 +14,7 @@ import { BlobCache } from '../file/BlobCache';
 import Tools from '../util/Tools';
 import * as URI from '../util/URI';
 import AstNode from './Node';
-import Schema, { getTextRootBlockElements, SchemaRegExpMap } from './Schema';
+import Schema, { getTextRootBlockElements, SchemaMap, SchemaRegExpMap } from './Schema';
 
 /**
  * @summary
@@ -503,8 +503,15 @@ const DomParser = (settings: DomParserSettings = {}, schema = Schema()): DomPars
     }
   };
 
+  const isWrappableNode = (blockElements: SchemaMap, node: AstNode) => {
+    const isInternalElement = Type.isString(node.attr(internalElementAttr));
+    const isInlineElement = node.type === 1 && (!Obj.has(blockElements, node.name) && !TransparentElements.isTransparentAstBlock(schema, node));
+
+    return node.type === 3 || (isInlineElement && !isInternalElement);
+  };
+
   const addRootBlocks = (rootNode: AstNode, rootBlockName: string): void => {
-    const blockElements = extend(makeMap('script,style,head,html,body,title,meta,param'), schema.getBlockElements(), schema.getTransparentElements());
+    const blockElements = extend(makeMap('script,style,head,html,body,title,meta,param'), schema.getBlockElements());
     const startWhiteSpaceRegExp = /^[ \t\r\n]+/;
     const endWhiteSpaceRegExp = /[ \t\r\n]+$/;
 
@@ -534,8 +541,7 @@ const DomParser = (settings: DomParserSettings = {}, schema = Schema()): DomPars
     while (node) {
       const next = node.next;
 
-      if (node.type === 3 || (node.type === 1 && node.name !== 'p' &&
-        !blockElements[node.name] && !node.attr(internalElementAttr))) {
+      if (isWrappableNode(blockElements, node)) {
         if (!rootBlockNode) {
           // Create a new root block element
           rootBlockNode = new AstNode(rootBlockName, 1);
@@ -573,7 +579,7 @@ const DomParser = (settings: DomParserSettings = {}, schema = Schema()): DomPars
     // Parse and sanitize the content
     const element = parseAndSanitizeWithContext(html, rootName, args.format);
 
-    TransparentElements.update(schema, element, Type.isUndefined(args.context));
+    TransparentElements.updateChildren(schema, element);
 
     // Create the AST representation
     const rootNode = new AstNode(rootName, 11);
--- a/modules/tinymce/src/core/main/ts/api/html/DomParser.ts
+++ b/modules/tinymce/src/core/main/ts/api/html/DomParser.ts
@@ -2,10 +2,11 @@ import { Arr, Fun, Obj, Strings, Type } from '@ephox/katamari';
 import { Attribute, NodeTypes, Remove, Replication, SugarElement } from '@ephox/sugar';
 import createDompurify, { Config, DOMPurifyI } from 'dompurify';
 
+import * as TransparentElements from '../../content/TransparentElements';
 import * as NodeType from '../../dom/NodeType';
 import * as FilterNode from '../../html/FilterNode';
 import * as FilterRegistry from '../../html/FilterRegistry';
-import { cleanInvalidNodes } from '../../html/InvalidNodes';
+import * as InvalidNodes from '../../html/InvalidNodes';
 import * as LegacyFilter from '../../html/LegacyFilter';
 import * as ParserFilters from '../../html/ParserFilters';
 import { isEmpty, isLineBreakNode, isPaddedWithNbsp, paddEmptyNode } from '../../html/ParserUtils';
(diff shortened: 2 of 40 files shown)


modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
```diff
@@ -344,5 +345,6 @@ export const insertHtmlAtCaret = (editor: Editor, value: string, details: Insert
   moveSelectionToMarker(editor, dom.get('mce_marker'));
   unmarkFragmentElements(editor.getBody());
   trimBrsFromTableCell(dom, selection.getStart());
+  TransparentElements.update(editor.schema, editor.getBody(), true);
```
TheSpyder, line 348: Do we have to do this on the entire body? Couldn't we just do the inserted content, or perhaps the parent of the inserted content?

TheSpyder (approved):

modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
spocke: Did some optimization now it's a bit tricky since it might need to update nodes up to the root.
Say you insert `<p>bar</p>` in `<div><a href="#">a|b</a></div>` then it should update data-mce-block state so that it becomes `<div><a href="#" data-mce-block="true">a<p>bar|</p>b</a></div>` so I changed it a bit so it goes down to the element just above the root and updates that if it find one otherwise it updates all elements.

spocke: Could be more complex structures as well say you insert `<a href="#><p>foo</p></a><i><b>bar</b></i>` in `<div>baz|</div>` then it should become `<div>baz<a href="#" data-mce-block="true"><p>foo</p></a><i><b>bar|</b></i></div>` there is a risk that we miss things unless we cast a wide net. So checking the parent wouldn't be enough in this case.

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
```diff
@@ -573,6 +573,8 @@ const DomParser = (settings: DomParserSettings = {}, schema = Schema()): DomPars
     // Parse and sanitize the content
     const element = parseAndSanitizeWithContext(html, rootName, args.format);
 
+    TransparentElements.update(schema, element, Type.isUndefined(args.context));
```
lnewson, line 576: This doesn't look right. A context does not specify if it's for the root block, it should be using `isRootContent` instead from memory.

modules/tinymce/src/core/main/ts/content/TransparentElements.ts
```diff
@@ -0,0 +1,57 @@
...
+import Schema, { SchemaMap } from '../api/html/Schema';
+import * as NodeType from '../dom/NodeType';
+
+export const transparentBlockAttr = 'data-mce-block';
+
+const makeSelectorFromSchemaMap = (map: SchemaMap) => Arr.filter(Obj.keys(map), (key) => /^[a-z]+$/.test(key)).join(',');
+
+const updateAt = (schema: Schema, root: Element, scope: Element, inEditorRoot: boolean) => {
+  const transparentSelector = makeSelectorFromSchemaMap(schema.getTransparentElements());
+  const blocksSelector = makeSelectorFromSchemaMap(schema.getBlockElements());
```
lnewson, line 14: As this is only used if a transparent element is found I wonder if we should be lazily doing this (e.g `Thunk.cached`)? I'm not sure on the performance impact though, but this doe happen on every parse 🤷

modules/tinymce/src/core/main/ts/content/TransparentElements.ts
```diff
@@ -0,0 +1,57 @@
...
+      if (transparent.getAttribute('data-mce-selected') === 'inline-boundary') {
+        transparent.removeAttribute('data-mce-selected');
+      }
+    } else {
+      transparent.removeAttribute(transparentBlockAttr);
+    }
+  });
+};
+
+export const update = (schema: Schema, root: Element, inEditorRoot: boolean): void => updateAt(schema, root, root, inEditorRoot);
```
lnewson, line 29:
```suggestion
export const update = (schema: Schema, root: Element, inEditorRoot: boolean): void =>
  updateAt(schema, root, root, inEditorRoot);
```
This is just a little easier to read 🤷 We've also mixed doing this below.

modules/tinymce/src/core/main/ts/content/TransparentElements.ts
```diff
@@ -0,0 +1,57 @@
...
+  const parents = Traverse.parents(SugarElement.fromDom(caretParent), isRoot);
+  // Check the element just above below the root so in if caretParent is I in this
+  // case <body><p><b><i>|</i></b></p></body> it would use the P as the scope
+  Arr.get(parents, parents.length - 2).filter(SugarNode.isElement).fold(
+    () => update(schema, root, true),
+    (scope) => updateAt(schema, root, scope.dom, true)
+  );
+};
+
+export const isTransparentElementName = (schema: Schema, name: string): boolean => name in schema.getTransparentElements();
```
lnewson, line 42:
```suggestion
export const isTransparentElementName = (schema: Schema, name: string): boolean =>
  name in schema.getTransparentElements();
```

lnewson: I also wonder if this should be using `Obj.has` as well as this will pick up things like `__proto__`.

modules/tinymce/src/core/main/ts/fmt/FormatUtils.ts
```diff
@@ -121,12 +122,8 @@ const getNonWhiteSpaceSibling = (node: Node | null, next?: boolean, inc?: boolea
...
 
-const isTextBlock = (schema: Schema, name: string | Node): boolean => {
-  if (isNode(name)) {
-    name = name.nodeName;
-  }
-
-  return !!schema.getTextBlockElements()[name.toLowerCase()];
+const isTextBlock = (schema: Schema, node: Node): boolean => {
+  return !!schema.getTextBlockElements()[node.nodeName.toLowerCase()] || TransparentElements.isTransparentBlock(schema, node);
 };
```
lnewson, line 127:
```suggestion
const isTextBlock = (schema: Schema, node: Node): boolean =>
  !!schema.getTextBlockElements()[node.nodeName.toLowerCase()] || TransparentElements.isTransparentBlock(schema, node);
```
This can be trimmed down now.

modules/tinymce/src/core/test/ts/browser/EditorTest.ts
```diff
@@ -79,25 +79,25 @@ describe('browser.tinymce.core.EditorTest', () => {
     editor.options.set('relative_urls', true);
     editor.documentBaseURI = new URI('http://www.site.com/dirA/dirB/dirC/');
 
-    editor.setContent('<a href="test.html">test</a>');
+    editor.setContent('<p><a href="test.html">test</a></p>');
```
lnewson, line 82: Are we sure we want this? I honestly would have expected this to be unnecessary and it to keep it as an inline anchor in this case since it doesn't contain another block element, but I guess that might be difficult to know what the intention was.

I'm also concerned this could be a **breaking** change if someone has this content and loads it into the editor is all. I guess it shouldn't happen when loading in existing content though, only new content.

modules/tinymce/src/core/test/ts/browser/content/InsertContentTest.ts
```diff
@@ -662,4 +663,63 @@ describe('browser.tinymce.core.content.InsertContentTest', () => {
     editor.insertContent('<foo-bar contenteditable="false" data-name="foobar"></foo-bar>');
     TinyAssertions.assertContent(editor, '<p><foo-bar contenteditable="false" data-name="foobar"></foo-bar></p>');
   });
+
+  it('TINY-9172: Insert block anchor in regual block', () => {
```
lnewson, line 667:
```suggestion
  it('TINY-9172: Insert block anchor in regular block', () => {
```
Typo

modules/tinymce/src/core/test/ts/browser/content/TransparentElementsTest.ts
```diff
@@ -0,0 +1,192 @@
...
+
+    it('TINY-9172: Should update anchor closest to root', () => {
+      const root = rootState.get().getOrDie();
+
+      Html.set(root, '<a href="#" data-mce-block="true">link</a>');
+      const scope = Hierarchy.follow(root, [ 0 ]).filter(SugarNode.isElement).getOrDie();
+      TransparentElements.updateCaret(schema, root.dom, scope.dom);
+      assert.equal( Html.get(root), '<a href="#" data-mce-block="true">link</a>');
+    });
+
```
lnewson, line 72:
```suggestion
```

modules/tinymce/src/core/test/ts/browser/html/DomParserTest.ts
```diff
@@ -1416,5 +1416,35 @@ describe('browser.tinymce.core.html.DomParserTest', () => {
 
       assert.equal(serializedHtml, html);
     });
+
+    context('transparent elements', () => {
+      const getTransparentElements = (schema: Schema) => Arr.unique(Arr.map(Obj.keys(schema.getTransparentElements()), (s) => s.toLowerCase()));
+
+      it('TINY-9172: inline transparents should not get data-mce-block attribute', () => {
+        const parser = DomParser();
+        const html = '<p>' + Arr.map(getTransparentElements(parser.schema), (name) => `<${name}>text</${name}>`).join('') + '</p>';
```
lnewson, line 1425: It might be more readable to break this up into 2 variables and use a template string maybe?

modules/tinymce/src/plugins/media/test/ts/browser/ContentFormatsTest.ts
```diff
@@ -37,7 +37,7 @@ describe('browser.tinymce.plugins.media.ContentFormatsTest', () => {
   it('TBA: Embed retained as is', () => {
     const editor = hook.editor();
     editor.setContent(
-      '<embed src="320x240.ogg" width="100" height="200">text<a href="#">link</a></embed>'
+      '<p><embed src="320x240.ogg" width="100" height="200">text<a href="#">link</a></embed></p>'
```
lnewson, line 40: Why was this needed? This definitely doesn't seem right as the root is not a transparent element, it's an inline element.

lnewson (commented):
Good work on taking my initial quick spike and getting it this far! 🏆

modules/tinymce/CHANGELOG.md
```diff
@@ -21,6 +22,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - A double bottom border appeared on inline mode editor for the `tinymce-5` skin. #TINY-9108
 - The editor header showed up even with no menubar and toolbar configured. #TINY-8819
 
+### Changed
```
lnewson, line 25: Forgot to leave a comment... this should be above `Fixed` and I don't think this makes sense as `Changed`. Maybe the first one, but the second one is definitely `Improved` or `Added`.

modules/tinymce/src/core/test/ts/browser/EditorTest.ts
spocke: A lot of these changes is to avoid having tests that rely on forced root blocks when they are testing something else. That is why I changed a lot of them. So it doesn't set one thin then tests for another thing.

modules/tinymce/src/core/main/ts/content/TransparentElements.ts
spocke: I only move them down if they don't fit on a line. I think this is readable. Lambdas is supposed to be one liners if they can be.

lnewson: It makes sense when you don't have a lot of args, but here you do and just makes it many times harder to read.

modules/tinymce/CHANGELOG.md
```diff
@@ -21,6 +22,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- Transparent elements like anchors are now allowed in the root of the editor body. #TINY-9172
+- Transparent elements like anchors can now contain block elements. #TINY-9172
```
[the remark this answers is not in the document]
spocke: Changed it to `Transparent elements, like anchors, are now allowed in the root of the editor body if they contain blocks` since that is now the new truth.

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
spocke: Got rid of this in root thing. Transparents are now block like if they contain blocks. That is the only rule.

modules/tinymce/src/core/main/ts/content/TransparentElements.ts
spocke: Not sure I agree anyway I could ETA reduce that function away it was just an alias now when the root check was removed.

modules/tinymce/src/core/main/ts/content/TransparentElements.ts
spocke: I just followed what we had before and I think we had some reasoning that Obj.has could have a bit of overhead but I guess we never investigated that.

modules/tinymce/src/plugins/media/test/ts/browser/ContentFormatsTest.ts
spocke: Rolled this back. Tried to avoid having one input and another output of things. Not really related to this PR but in general.

modules/tinymce/src/core/main/ts/content/TransparentElements.ts
spocke: Not sure Thunk.cached can be used here since it would create a new thunk each time. It would have to be some global state bag. Maybe we optimize this later if it need to be optimized. The sets here are pretty small.

modules/tinymce/src/core/main/ts/content/TransparentElements.ts
spocke: Switched to Obj.has optimizations can be done later once we figure all the quirks out.

lnewson: Ahh right I forgot we'd done that in some cases sorry. This was very much an edge case on code that shouldn't be public so it wasn't really an issue anyways. Whatever you think is best anyways 👍

modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
spocke: Added some more tests around this.

modules/tinymce/CHANGELOG.md
```diff
@@ -9,17 +9,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+### Improved
+- Transparent elements, like anchors, can now contain block elements. #TINY-9172
+
 ### Fixed
 - Dragging a noneditable element toward the bottom edge would cause the page to scroll up. #TINY-9025
 - Range expanding capabilities would behave inconsistently depending on where the cursor was placed. #TINY-9029
 - Compilation errors were thrown when using TypeScript 4.8. #TINY-9161
 - Line separator scrolling in floating toolbars. #TINY-8948
 - A double bottom border appeared on inline mode editor for the `tinymce-5` skin. #TINY-9108
-- The editor header showed up even with no menubar and toolbar configured. #TINY-8819
```
lnewson, line 22: Did you mean to remove this btw? I'm guessing ti was an accident.

modules/tinymce/src/core/main/ts/content/TransparentElements.ts
lnewson: Doesn't matter then... anyways, to explain myself. When you need to search across a line to try and find the start it involves a lot more scanning. So at all times it's just easier to have the function definition and then the code separately. So as you add more to the definition it just gets harder to find the end of the definition and start of the actual code.

Code highlighting can help, but you don't always have access to that or some systems might not do it as well.

modules/tinymce/CHANGELOG.md
spocke: Wops! How did that happen.

modules/tinymce/src/core/main/ts/content/TransparentElements.ts
spocke: If we only had a language that didn't have all these noisy type annotations.

Like I rather have this (rescript):
```
let filter = (v, ~f) => keep(v, f)
let filterWithIndex = (v, ~f) => keepWithIndex(v, f)
let filterMap = (v, ~f) => keepMap(v, f)
```

Than this it's so much harder to read for me at least:
```
let filter = (v, ~f) =>
  keep(v, f)
let filterWithIndex = (v, ~f) =>
  keepWithIndex(v, f)
let filterMap = (v, ~f) =>
  keepMap(v, f)
```

I think the => in typescript is easy to spot there is where the body starts and the annotations stop. Having less lines of code is also a plus. I would prefer to have just one or two pages of code per module very hard to do sometime when one just adds and adds things and want to hide the details of a module.

We can debate this in slack if you like. I don't think there is a single truth here.

modules/tinymce/CHANGELOG.md
```diff
@@ -9,10 +9,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - New `expand` function added to `tinymce.selection` which expands the selection around the nearest word. #TINY-9001
 - New `expand` function added to `tinymce.dom.RangeUtils` to return a new range expanded around the nearest word. #TINY-9001
+- New `getTransparentElements` function added to `tinymce.html.Schema` to return a map object of transparent HTML elements. #TINY-9172
```
vpyshnenko, line 12: In console `tinymce.html.Schema.getTransparentElements` returns `undefined`

vpyshnenko (approved):

PR #8176, opened 2022-10-10
TINY-9237: Limit overscroll for floating toolbar

noxuhax:
Related Ticket: TINY-9237

Description of Changes:
* Limited overscroll for floating toolbar

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [x] Milestone set

### DIFF
--- a/modules/oxide/src/less/theme/components/toolbar/toolbar.less
+++ b/modules/oxide/src/less/theme/components/toolbar/toolbar.less
@@ -125,6 +125,7 @@
   border: none;
   border-radius: @pop-border-radius;
   box-shadow: @pop-box-shadow;
+  overscroll-behavior: none;
   padding: @pad-xs 0;
 }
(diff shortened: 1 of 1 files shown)


yacodes (approved):

TheSpyder (approved):

spocke (commented):
Wouldn't this change require a changelog item. It would be something that users would notice that we fixed?

noxuhax:
> Wouldn't this change require a changelog item. It would be something that users would notice that we fixed?

@spocke this is a follow up fix for an unreleased TINY-8948 which already has it’s [changelog entry](https://github.com/tinymce/tinymce/blob/develop/modules/tinymce/CHANGELOG.md?plain=1#L18). This PR does not fundamentally add anything new to that task, only fixing a crossbrowser inconsistency.

PR #8178, opened 2022-10-11
TINY-9223: Starting to separate out the dialog and popup sinks/motherships

ephox-mogran:
Related Ticket: TINY-9223

Description of Changes:
* passing through different sinks and motherships for modal dialogs vs other popups. This is in preparation for supporting tinymce within a scrollable element. Note: there is still just one mothership and sink, but the code has been separated to (theoretically) support more than one more easily. This also doesn't include the code to make Positioning work with the popup sink - that's going to be TINY-9226.

Pre-checks:
* [X] Changelog entry added

Review:
* [X] Milestone set
* [N/A] Docs ticket created (if applicable) - it's internal

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -13,9 +13,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - New `ToggleView` command which makes it possible to hide or show registered custom views. #TINY-9210
 - New `color_default_foreground` and `color_default_background` options to set the initial default color for the `forecolor` and `backcolor` toolbar buttons and menu items. #TINY-9183
 
-### Changed
-- Separated the sinks and backstage setups between dialogs and other popups. #TINY-9223
-
 ### Fixed
 - Dragging a noneditable element toward the bottom edge would cause the page to scroll up. #TINY-9025
 - Range expanding capabilities would behave inconsistently depending on where the cursor was placed. #TINY-9029
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -13,6 +13,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - New `ToggleView` command which makes it possible to hide or show registered custom views. #TINY-9210
 - New `color_default_foreground` and `color_default_background` options to set the initial default color for the `forecolor` and `backcolor` toolbar buttons and menu items. #TINY-9183
 
+### Changed
+- Separated the sinks and backstage setups between dialogs and other popups. #TINY-9223
+
 ### Fixed
 - Dragging a noneditable element toward the bottom edge would cause the page to scroll up. #TINY-9025
 - Range expanding capabilities would behave inconsistently depending on where the cursor was placed. #TINY-9029
--- a/modules/tinymce/src/themes/silver/main/ts/Events.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/Events.ts
@@ -6,20 +6,23 @@ import Editor from 'tinymce/core/api/Editor';
 import { AfterProgressStateEvent } from 'tinymce/core/api/EventTypes';
 import { EditorEvent } from 'tinymce/core/api/util/EventDispatcher';
 
-const setup = (editor: Editor, mothership: Gui.GuiSystem, uiMothership: Gui.GuiSystem): void => {
+const setup = (editor: Editor, mothership: Gui.GuiSystem, uiMotherships: Gui.GuiSystem[]): void => {
   const broadcastEvent = (name: string, evt: EventArgs) => {
-    Arr.each([ mothership, uiMothership ], (ship) => {
-      ship.broadcastEvent(name, evt);
+    Arr.each([ mothership, ...uiMotherships ], (m) => {
+      m.broadcastEvent(name, evt);
     });
   };
 
   const broadcastOn = (channel: string, message: Record<string, any>) => {
-    Arr.each([ mothership, uiMothership ], (ship) => {
-      ship.broadcastOn([ channel ], message);
+    Arr.each([ mothership, ...uiMotherships ], (m) => {
+      m.broadcastOn([ channel ], message);
     });
   };
 
-  const fireDismissPopups = (evt: EventArgs) => broadcastOn(Channels.dismissPopups(), { target: evt.target });
+  const fireDismissPopups = (evt: EventArgs) => broadcastOn(
+    Channels.dismissPopups(),
+    { target: evt.target }
+  );
 
   // Document touch events
   const doc = SugarDocument.getDocument();
@@ -101,10 +104,8 @@ const setup = (editor: Editor, mothership: Gui.GuiSystem, uiMothership: Gui.GuiS
   });
 
   editor.on('detach', () => {
-    Attachment.detachSystem(mothership);
-    Attachment.detachSystem(uiMothership);
-    mothership.destroy();
-    uiMothership.destroy();
+    Arr.each([ mothership, ...uiMotherships ], Attachment.detachSystem);
+    Arr.each([ mothership, ...uiMotherships ], (m) => m.destroy());
   });
 };
 
--- a/modules/tinymce/src/themes/silver/main/ts/Render.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/Render.ts
@@ -14,6 +14,7 @@ import * as Backstage from './backstage/Backstage';
 import * as Events from './Events';
 import * as Iframe from './modes/Iframe';
 import * as Inline from './modes/Inline';
+import { LazyUiReferences, ReadyUiReferences } from './modes/UiReferences';
 import * as ReadOnly from './ReadOnly';
 import * as ContextToolbar from './ui/context/ContextToolbar';
 import * as FormatControls from './ui/core/FormatControls';
@@ -49,13 +50,6 @@ export interface RenderInfo {
   readonly renderUI: () => ModeRenderInfo;
 }
 
-export interface RenderUiComponents {
-  readonly mothership: Gui.GuiSystem;
-  readonly uiMothership: Gui.GuiSystem;
-  readonly outerContainer: AlloyComponent;
-  readonly sink: AlloyComponent;
-}
-
 export type ToolbarConfig = Array<string | Options.ToolbarGroupOption> | string | boolean;
 
 export interface RenderUiConfig extends RenderToolbarConfig, MenuRegistry {
@@ -74,10 +68,14 @@ const getLazyMothership = (singleton: Singleton.Value<Gui.GuiSystem>) =>
 const setup = (editor: Editor): RenderInfo => {
   const isInline = editor.inline;
   const mode = isInline ? Inline : Iframe;
+
+  // We use a different component for creating the sticky toolbar behaviour. The
+  // most important difference is it needs "Docking" configured and all of the
+  // ripple effects that creates.
   const header = Options.isStickyToolbar(editor) ? StickyHeader : StaticHeader;
 
-  const lazySink = Singleton.value<AlloyComponent>();
-  const lazyOuterContainer = Singleton.value<AlloyComponent>();
+  const lazyUiRefs = LazyUiReferences();
+
   const lazyMothership = Singleton.value<Gui.GuiSystem>();
   const lazyUiMothership = Singleton.value<Gui.GuiSystem>();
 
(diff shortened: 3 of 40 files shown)


modules/tinymce/CHANGELOG.md
```diff
@@ -10,6 +10,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - New `expand` function added to `tinymce.selection` which expands the selection around the nearest word. #TINY-9001
 - New `expand` function added to `tinymce.dom.RangeUtils` to return a new range expanded around the nearest word. #TINY-9001
 
+### Changed
+- Separated the sinks and backstage setups between dialogs and other popups. #TINY-9223
```
TheSpyder, line 14: This doesn't seem like a user-facing change? We probably don't need a changelog until we have the final scroll container fix 🤔

TheSpyder (approved):

modules/tinymce/CHANGELOG.md
ephox-mogran: Should the changelog have only user-facing changes? Isn't that the difference between the changelog and the release notes? The changelog might impact someone contributing, whereas the release notes is for the user?

lnewson: It's always been our rule for anything customer facing it should only contain things that impact them. 🤷

ephox-mogran: Fair enough. I don't mind either way. Just trying to understand the boundaries and reasons.

lnewson: I should add at least one of the reasons are to avoid confusing users with things they won't understand. As remember this is effectively what community use to know what changed in TinyMCE when upgrading.

ephox-mogran: - [ ] remove non user-facing entry from changelog

modules/tinymce/src/themes/silver/main/ts/Events.ts
```diff
@@ -6,20 +6,23 @@ import Editor from 'tinymce/core/api/Editor';
 import { AfterProgressStateEvent } from 'tinymce/core/api/EventTypes';
 import { EditorEvent } from 'tinymce/core/api/util/EventDispatcher';
 
-const setup = (editor: Editor, mothership: Gui.GuiSystem, uiMothership: Gui.GuiSystem): void => {
+const setup = (editor: Editor, mothership: Gui.GuiSystem, uiMotherships: Gui.GuiSystem[]): void => {
```
spocke, line 9: Is there a reason we need to distinct the motherships here. From what I can see it does the same thing to both mothership and the uiMotherships? So one just concats the arrays before calling this or is that confusing.

spocke (approved):
Think it looks fine. However take my review with a grain of salt not that familiar with the ins and outs of silver.

modules/tinymce/src/themes/silver/main/ts/backstage/Backstage.ts
```diff
@@ -41,32 +41,66 @@ export interface UiFactoryBackstage {
...
+      interpreter: (s) => UiFactory.interpretWithoutForm(s, {}, popupBackstage),
+      getSink: lazySinks.popup
+    }
+  };
+
+  const dialogBackstage: UiFactoryBackstage = {
+    ...commonBackstage,
+    shared: {
+      ...commonBackstage.shared,
+      interpreter: (s) => UiFactory.interpretWithoutForm(s, {}, popupBackstage),
```
lnewson, line 95: Is this right? I believe this should be using `dialogBackstage`. If not it probably needs a TODO to change later.

modules/tinymce/src/themes/silver/main/ts/backstage/Backstage.ts
```diff
@@ -41,32 +41,66 @@ export interface UiFactoryBackstage {
...
+      getSink: lazySinks.popup
```
lnewson, line 96: Likewise here as this is using the popup sink for the dialog backstage.

modules/tinymce/src/themes/silver/main/ts/Render.ts
```diff
@@ -37,19 +38,17 @@ export interface ModeRenderInfo {
 }
 
 export interface RenderInfo {
-  readonly getMothership: () => Gui.GuiSystem;
-  readonly getUiMothership: () => Gui.GuiSystem;
-  readonly backstage: Backstage.UiFactoryBackstage;
+  readonly forPopups: {
```
lnewson, line 41: Is there a reason why we aren't consistent with the naming here? Everywhere else we use `popups` and `dialogs`

modules/tinymce/src/themes/silver/main/ts/Render.ts
```diff
@@ -88,13 +92,45 @@ const setup = (editor: Editor): RenderInfo => {
...
+    .bind(OuterContainer.getHeader);
+
+  const lazyDialogSinkResult = () => Result.fromOption(
+    lazyUiRefs.dialogUi.get().map((ui) => ui.sink),
+    'UI has not been rendered'
+  );
+
+  const lazyPopupSinkResult = () => Result.fromOption(
+    lazyUiRefs.popupUi.get().map((ui) => ui.sink),
+    '(adjacent) UI has not been rendered'
```
lnewson, line 106: Is this still correct as it's using a different name? Should it be this instead:
```suggestion
    '(popup) UI has not been rendered'
```

modules/tinymce/src/themes/silver/main/ts/backstage/Backstage.ts
```diff
@@ -41,32 +41,66 @@ export interface UiFactoryBackstage {
...
+  readonly popup: UiFactoryBackstage;
+  readonly dialog: UiFactoryBackstage;
+}
+
+const init = (lazySinks: { popup: () => Result<AlloyComponent, string>; dialog: () => Result<AlloyComponent, string> }, editor: Editor, lazyAnchorbar: () => AlloyComponent): UiFactoryBackstagePair => {
   const contextMenuState = Cell(false);
   const toolbar = HeaderBackstage(editor);
-  const backstage: UiFactoryBackstage = {
+
+  const providers: UiFactoryBackstage['shared']['providers'] = {
```
lnewson, line 53: This should be using the `UiFactoryBackstageProviders` type instead if we're keeping things consistent.
```suggestion
  const providers: UiFactoryBackstageProviders = {
```

modules/tinymce/src/themes/silver/main/ts/modes/Iframe.ts
```diff
@@ -78,17 +83,25 @@ const setupEvents = (editor: Editor, uiComponents: RenderUiComponents) => {
   });
 };
 
-const render = (editor: Editor, uiComponents: RenderUiComponents, rawUiConfig: RenderUiConfig, backstage: UiFactoryBackstage, args: RenderArgs): ModeRenderInfo => {
+// TINY-9226: When introducing two sinks, the dialog mothership should be attached to the ui
```
lnewson, line 86: Is this still accurate? We are attaching it below is all

modules/tinymce/src/themes/silver/main/ts/modes/Inline.ts
```diff
@@ -71,11 +72,18 @@ const setupEvents = (editor: Editor, targetElm: SugarElement, ui: InlineHeader,
   });
 };
 
-const render = (editor: Editor, uiComponents: RenderUiComponents, rawUiConfig: RenderUiConfig, backstage: UiFactoryBackstage, args: RenderArgs): ModeRenderInfo => {
-  const { mothership, uiMothership, outerContainer } = uiComponents;
+// TINY-9226: When introducing two sinks, the dialog mothership should be attached to the ui
```
lnewson, line 75: Likewise here this seems obsolete.

modules/tinymce/src/themes/silver/main/ts/ui/header/InlineHeader.ts
```diff
@@ -27,11 +27,10 @@ const { ToolbarLocation, ToolbarMode } = Options;
 export const InlineHeader = (
   editor: Editor,
   targetElm: SugarElement<HTMLElement>,
-  uiComponents: RenderUiComponents,
+  uiRefs: ReadyUiReferences,
   backstage: UiFactoryBackstage,
   floatContainer: Singleton.Value<AlloyComponent>
 ): InlineHeader => {
-  const { uiMothership, outerContainer } = uiComponents;
```
lnewson, line 34: Why did we remove this? It helps keep the code smaller, etc... to deconstruct this first. I saw the same problem in WindowManager where every time you want to access anything you end up with a massive string.

- [ ] Reconsider doing this type of thing as it makes things harder to use/write

modules/tinymce/src/themes/silver/main/ts/ui/header/InlineHeader.ts
```diff
@@ -196,20 +197,25 @@ export const InlineHeader = (
...
+    });
     updateMode(false);
     updateChromeUi();
   };
 
   const hide = () => {
     visible.set(false);
-    if (uiComponents.outerContainer) {
-      Css.set(outerContainer.element, 'display', 'none');
+    // In what situation would `outerContainer` not be set?
```
lnewson, line 211: This looks like something that was left in unintentionally? Either that or the check needs to be removed with this comment or a TODO and Jira to investigate. I suspect this was probably left over from refactoring where at some point this might have been nullable.

modules/tinymce/src/themes/silver/main/ts/Events.ts
ephox-mogran: The separation is a bit more conceptual. I don't mind either way, but that's the distinction.

modules/tinymce/src/themes/silver/main/ts/Render.ts
ephox-mogran: Definitely no reason.

- [ ] fix consistency with naming.

modules/tinymce/src/themes/silver/main/ts/Render.ts
ephox-mogran: Yep. I changed the names between the first spike and this task. It should be popup for now

- [ ] change all usages of adjacent to popup

modules/tinymce/src/themes/silver/main/ts/backstage/Backstage.ts
ephox-mogran: Definitely wrong. Wasn't picked up in tests because this would only surface when we tried to create a UI control that cared about the backstage sink ... so I might add a test with a URLInput and check that inline dialogs and modal dialogs are putting the dropdown menu in the right place

- [ ] fix and add test that fails because this was wrong.

modules/tinymce/src/themes/silver/main/ts/backstage/Backstage.ts
ephox-mogran: Hmm. Also wrong. I'll need to investigate why the test wasn't failing for this one. I'm probably not testing what I thought I was. The backstage in tests does separate the two, so the overall placement of dialogs was being tested, but I'll double-check what this was used for

- [ ] fix and write test case that fails because this was wrong.

modules/tinymce/src/themes/silver/main/ts/modes/Iframe.ts
ephox-mogran: It's probably a bit confusing, but TINY-9226 is a future task. It hasn't been done yet. I was trying to mark what it would need to change. Same with the other TINY-9226.

modules/tinymce/src/themes/silver/main/ts/modes/Inline.ts
ephox-mogran: See other comment about TINY-9226

modules/tinymce/src/themes/silver/main/ts/ui/header/InlineHeader.ts
ephox-mogran: My end goal was to have different types at each layer that only had the required parts in, but still mirrored the original structure just to avoid a whole lot of marshalling code, but I didn't do that here just because it would bloat the PR. However, I'm not sure if you'd prefer that anyway.

My main issue with RenderUiComponents is that it was just a bag of things, and wasn't structured in any meaningful way. I preferred the explicitness of defining what exactly I was using (and its context through namespacing), As soon as ReadyuiReferences was given structure, then destructuring it quickly became less clear. Also, not everything in there was a component (motherships etc.).

I can revert the change, but I would disagree with your statement that it makes it hard to read/use/write. However, that doesn't make my view right. Happy to go with those who have had to work with the code the most.

modules/tinymce/src/themes/silver/main/ts/ui/header/InlineHeader.ts
ephox-mogran: Yeah, I'm still not 100% sure what I should do when I see code that doesn't make sense to me. I guess my instinct is to tag it with a comment, but that doesn't really help anyone else.

I think in this case, the way the code has been structured now, this will never not be set, so I'll remove the check.

- [ ] remove the check.

modules/tinymce/src/themes/silver/main/ts/backstage/Backstage.ts
ephox-mogran: Ah, the tests use an alternate test backstage, so of course they aren't testing this!

modules/tinymce/src/themes/silver/main/ts/ui/header/InlineHeader.ts
lnewson: I maybe wasn't clear enough here, or maybe I'm not understanding what you're saying as we seem to be talking about 2 separate things.

I was fine with the type changes. What I'm talking about is you do `uiRefs.mainUi.outerContainer` numerous times below. The editor team has spent time removing that in numerous places in silver previous and instead destructing early such as `const { mainUi, uiMotherships } = uiRefs;` or passing only what's needed to reduce all that duplication and the long `a.b.c.d`, etc... strings as they were not liked and slowed down development and made it harder to find what to use when you're not used to it.

ephox-mogran: Discussed. I'll change it to max one level of reference (e.g. mainUi.outerContainer).

- [ ] change references.

modules/tinymce/src/themes/silver/main/ts/backstage/Backstage.ts
ephox-mogran: Added tests that check these details.

modules/tinymce/src/themes/silver/main/ts/backstage/Backstage.ts
ephox-mogran: Added tests that check these details.

modules/tinymce/src/themes/silver/main/ts/ui/header/InlineHeader.ts
ephox-mogran: I've made it so that in most places, there is at most one level of property access.

modules/tinymce/src/themes/silver/main/ts/ui/header/InlineHeader.ts
ephox-mogran: Removed.

modules/tinymce/src/themes/silver/main/ts/backstage/Backstage.ts
ephox-mogran: Changed.

PR #8182, opened 2022-10-12
TINY-9190: Stopped quickbars context toolbar from being shown in the fake caret

tiny-ben-tran:
Related Ticket:  TINY-9190

Description of changes:
- Updated the predicate to not show `Quickbars` context tool bar when the cursor is in bogus elements
- Added tests for predicate

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -34,6 +34,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Matches of inline text patterns that are similar are now managed correctly. #TINY-8949
 - The context toolbar prevented the user from placing the cursor at the edges of the editor. #TINY-8890
 - The Quick Insert context toolbar provided by the `quickbars` plugin showed when the cursor was in a fake block caret. #TINY-9190
+- The `editor.selection.getRng()` API was not returning a proper range on hidden editors in Firefox. #TINY-9259
+- The `editor.selection.getBookmark()` API was not returning a proper bookmark on hidden editors in Firefox. #TINY-9259
 
 ## 6.2.0 - 2022-09-08
 
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -17,7 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Line separator scrolling in floating toolbars. #TINY-8948
 - A double bottom border appeared on inline mode editor for the `tinymce-5` skin. #TINY-9108
 - The editor header showed up even with no menubar and toolbar configured. #TINY-8819
-- The context toolbar of `quickbars` plugin was showed in the fake block caret. #TINY-9190
+- The Quick Insert context toolbar provided by the `quickbars` plugin showed when the cursor was in a fake block caret. #TINY-9190
 
 ## 6.2.0 - 2022-09-08
 
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -9,14 +9,30 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - New `expand` function added to `tinymce.selection` which expands the selection around the nearest word. #TINY-9001
 - New `expand` function added to `tinymce.dom.RangeUtils` to return a new range expanded around the nearest word. #TINY-9001
+- New `color_map_background` and `color_map_foreground` option which sets the base colors used in the `backcolor` and `forecolor` toolbar buttons and menu items. #TINY-9184
+- Added optional `storageKey` property to colorinput component #TINY-9184
+- New `addView` function added to `editor.ui.registry` which makes it possible to register custom editor views. #TINY-9210
+- New `ToggleView` command which makes it possible to hide or show registered custom views. #TINY-9210
+- New `color_default_foreground` and `color_default_background` options to set the initial default color for the `forecolor` and `backcolor` toolbar buttons and menu items. #TINY-9183
+- New `getTransparentElements` function added to `tinymce.html.Schema` to return a map object of transparent HTML elements. #TINY-9172
+
+### Changed
+- Transparent elements, like anchors, are now allowed in the root of the editor body if they contain blocks. #TINY-9172
+
+### Improved
+- Transparent elements, like anchors, can now contain block elements. #TINY-9172
 
 ### Fixed
+- Parsing media content would cause a memory leak, which for example occurred when using the `getContent` API. #TINY-9186
 - Dragging a noneditable element toward the bottom edge would cause the page to scroll up. #TINY-9025
 - Range expanding capabilities would behave inconsistently depending on where the cursor was placed. #TINY-9029
 - Compilation errors were thrown when using TypeScript 4.8. #TINY-9161
 - Line separator scrolling in floating toolbars. #TINY-8948
 - A double bottom border appeared on inline mode editor for the `tinymce-5` skin. #TINY-9108
 - The editor header showed up even with no menubar and toolbar configured. #TINY-8819
+- Inline text pattern no longer triggers if it matches only the end but not the start. #TINY-8947
+- Matches of inline text patterns that are similar are now managed correctly. #TINY-8949
+- The context toolbar prevented the user from placing the cursor at the edges of the editor. #TINY-8890
 - The Quick Insert context toolbar provided by the `quickbars` plugin showed when the cursor was in a fake block caret. #TINY-9190
 
 ## 6.2.0 - 2022-09-08
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Line separator scrolling in floating toolbars. #TINY-8948
 - A double bottom border appeared on inline mode editor for the `tinymce-5` skin. #TINY-9108
 - The editor header showed up even with no menubar and toolbar configured. #TINY-8819
+- The context toolbar of `quickbars` plugin was showed in the fake block caret. #TINY-9190
 
 ## 6.2.0 - 2022-09-08
 
--- a/modules/tinymce/src/plugins/quickbars/main/ts/insert/Toolbars.ts
+++ b/modules/tinymce/src/plugins/quickbars/main/ts/insert/Toolbars.ts
@@ -1,5 +1,5 @@
 import { Fun } from '@ephox/katamari';
-import { PredicateExists, SelectorFind, SugarElement, SugarNode } from '@ephox/sugar';
+import { Attribute, PredicateExists, SelectorFind, SugarElement, SugarNode } from '@ephox/sugar';
 
 import Editor from 'tinymce/core/api/Editor';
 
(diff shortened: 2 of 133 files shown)


modules/tinymce/src/plugins/quickbars/main/ts/insert/Toolbars.ts
```diff
@@ -12,10 +12,16 @@ const addToEditor = (editor: Editor): void => {
       predicate: (node) => {
         const sugarNode = SugarElement.fromDom(node);
         const textBlockElementsMap = editor.schema.getTextBlockElements();
-        const isRoot = (elem: SugarElement<Node>) => elem.dom === editor.getBody();
+        const isRoot = (elem) => elem.dom === editor.getBody();
+        const isFakeCaret = (elem) =>
+          Attribute.has(elem, 'data-mce-bogus') || SelectorExists.ancestor(elem, '[data-mce-bogus="all"]', (el) => isRoot(el));
```
lnewson, line 17:
```suggestion
        const isRoot = (elem: SugarElement<Node>) => elem.dom === editor.getBody();
        const isFakeCaret = (elem: SugarElement<Node>) =>
          Attribute.has(elem, 'data-mce-bogus') || SelectorExists.ancestor(elem, '[data-mce-bogus="all"]', isRoot);
```
Just copying what I mentioned in DMs. Sorry for the confusion 😞

modules/tinymce/src/plugins/quickbars/test/ts/browser/ContentEditableTest.ts
```diff
@@ -57,4 +57,11 @@ describe('browser.tinymce.plugins.quickbars.ContentEditableTest', () => {
     TinySelections.select(editor, 'p span', []);
     await pAssertToolbarNotVisible();
   });
+
+  it('TINY-9190: Toolbar is not shown in the fake caret', async () => {
+    const editor = hook.editor();
+    editor.setContent('<p contenteditable="false">CEF element</p>');
+    // Selection is already in the fake caret
```
spocke, line 64: Maybe this should also assert that the fake caret is actually showing or that the `editor.selection.getRng` returns the position before the `p` tag though TinyAssertions. Since now the test kind of just assumes that the caret will be where the comment says it is.

spocke (approved):

modules/tinymce/CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Line separator scrolling in floating toolbars. #TINY-8948
 - A double bottom border appeared on inline mode editor for the `tinymce-5` skin. #TINY-9108
 - The editor header showed up even with no menubar and toolbar configured. #TINY-8819
+- The context toolbar of `quickbars` plugin was showed in the fake block caret. #TINY-9190
```
ltrouton, line 20: Just wondering if we should be more specific.
```suggestion
- The Quick Insert context toolbar provided by the `quickbars` plugin showed when the cursor was in the fake block caret. #TINY-9190
```

TheSpyder (approved):
I'm still on the fence about whether this should go into core, but I guess it can move there if we find other context toolbars need this.

ltrouton (approved):

modules/tinymce/src/plugins/quickbars/main/ts/insert/Toolbars.ts
```diff
@@ -13,9 +13,15 @@ const addToEditor = (editor: Editor): void => {
...
         const isRoot = (elem: SugarElement<Node>) => elem.dom === editor.getBody();
+        const isFakeCaret = (elem: SugarElement<Node>) =>
+          Attribute.has(elem, 'data-mce-bogus') || SelectorExists.ancestor(elem, '[data-mce-bogus="all"]', isRoot);
+
         return SelectorFind.closest(sugarNode, 'table', isRoot).fold(
-          () => PredicateFind.closest(sugarNode, (elem) =>
-            SugarNode.name(elem) in textBlockElementsMap && editor.dom.isEmpty(elem.dom), isRoot).isSome(),
+          () => PredicateExists.closest(
+            sugarNode,
+            (elem) => !isFakeCaret(elem) && SugarNode.name(elem) in textBlockElementsMap && editor.dom.isEmpty(elem.dom),
```
ephox-mogran, line 22: I think the order of these calculations could be improved. Let's break it down: 

* isFakeCaret is potentially going to look all the way up the tree to the root.
* SugarNode is an O(1) operation

So I think the first improvement would be to do the SugarNode.name check first, as that will potentially prevent you from needing to do any "up the dom searching" with isFakeCaret. Remember, that for the common case of the current element **doesn't have** `data-mce-bogus`, `isFakeCaret` is going to search all the way up to the root on *every* PredicateExists call. That's a lot of repeated effort, and I don't think the result is ever going to change if it's false.

Then, you could probably combine the two searches so that we aren't looking up every step. Or you could look at doing a Thunk.cached for the SelectorExists.ancestor. It just seems that there is something inefficient here, but I might be misreading the code.

modules/tinymce/src/plugins/quickbars/test/ts/browser/ContentEditableTest.ts
```diff
@@ -57,4 +57,13 @@ describe('browser.tinymce.plugins.quickbars.ContentEditableTest', () => {
     TinySelections.select(editor, 'p span', []);
     await pAssertToolbarNotVisible();
   });
+
+  it('TINY-9190: Toolbar is not shown in the fake caret', async () => {
```
ephox-mogran, line 61: We probably need some test cases that include some checking of the ancestors.

modules/tinymce/src/plugins/quickbars/main/ts/insert/Toolbars.ts
tiny-ben-tran: I moved `isFakeCaret` so it is the last check

modules/tinymce/src/plugins/quickbars/test/ts/browser/ContentEditableTest.ts
tiny-ben-tran: A test for the ancestor with `data-mce-bogus="all"` is added

modules/tinymce/src/plugins/quickbars/main/ts/insert/Toolbars.ts
lnewson: @tiny-ben-tran this only does the first part of what Morgan mentioned. Was there a reason the second part was skipped?

tiny-ben-tran: Thanks Lee, I totally missed it.

I combined the two searches as suggested.

lnewson: One thing to remember here is that the predicate function gets called multiple times for every node up the tree from the selection.

ephox-mogran: Yeah, my read of the original code is what we *want* to happen is this: 

* if anything up the tree is a table, then do not render
* otherwise, if we find an ancestor text block-capable node up the tree, and it's empty, render. Otherwise, keep looking.

The additional things to add are: 

* if anything up the tree is a data-mce-bogus=all, then do not render
* otherwise, if we find an ancestor that has data-mce-bogus, then do not render.

So it seems like we have two different types of checks. My instinct suggests you should be able to consolidate them. So the first `SelectorFind.closest` could look for either table, or [data-mce-bogus=all], as both conditions will stop it from rendering. Otherwise, you run through the PredicateExists.closest and then you'll only need to check for the data-mce-bus attribute and the other things it was already checking for.

Is that preserving the right intent @lnewson ?

lnewson: Sorry I missed this, but yes that sounds correct to me.

modules/tinymce/src/plugins/quickbars/test/ts/browser/PredicateTest.ts
```diff
@@ -0,0 +1,51 @@
...
+  const hook = TinyHooks.bddSetup<Editor>({
+    plugins: 'quickbars link',
+    inline: true,
+    toolbar: false,
+    menubar: false,
+    base_url: '/project/tinymce/js/tinymce'
+  }, [ Plugin ], true);
+
+
+  it('TINY-9190: Toolbar is not shown when the ancestor element has a data-mce-bogus attribute', async () => {
```
ephox-mogran, line 19: You'll probably need to add some tests here for when it should be shown. Otherwise, all of this passes with Fun.constant(false).

TheSpyder (approved):

modules/tinymce/src/plugins/quickbars/main/ts/insert/Toolbars.ts
tiny-ben-tran: Should the toolbar be shown if the cursor is placed in the deepest span element of `<p><span><span data-mce-bogus="1"><span>{cursor here}</span></span></span></p>`?

tiny-ben-tran: @lnewson @ephox-mogran, can you give this PR another review?

modules/tinymce/src/plugins/quickbars/test/ts/browser/PredicateTest.ts
```diff
@@ -0,0 +1,69 @@
+import { describe, it } from '@ephox/bedrock-client';
+import { TinyHooks, TinySelections } from '@ephox/wrap-mcagar';
+
+import Editor from 'tinymce/core/api/Editor';
+import Plugin from 'tinymce/plugins/quickbars/Plugin';
+
+import { pAssertToolbarNotVisible, pAssertToolbarVisible } from '../module/test/Utils';
+
+describe('browser.tinymce.plugins.quickbars.PredicateTest', () => {
```
lnewson, line 9: Should this test/file be called `InsertToolbarPredicateTest` or even `InsertToolbarTest` (to match `SelectionToolbarTest`)? As this sounds like it's generic, but this only tests the insert quickbar and nothing else.

modules/tinymce/src/plugins/quickbars/test/ts/module/test/Utils.ts
```diff
@@ -0,0 +1,14 @@
+import { UiFinder, Waiter } from '@ephox/agar';
+import { SugarBody } from '@ephox/sugar';
+
+const pAssertToolbarNotVisible = async (): Promise<void> => {
+  await Waiter.pWait(50);
+  UiFinder.notExists(SugarBody.body(), '.tox-pop__dialog .tox-toolbar');
+};
+
+const pAssertToolbarVisible = (): Promise<void> => Waiter.pTryUntil('toolbar should exist', () => UiFinder.exists(SugarBody.body(), '.tox-toolbar'));
```
lnewson, line 9:
```suggestion
const pAssertToolbarVisible = (): Promise<void> =>
  Waiter.pTryUntil('toolbar should exist', () => UiFinder.exists(SugarBody.body(), '.tox-toolbar'));
```

modules/tinymce/src/plugins/quickbars/test/ts/module/test/Utils.ts
```diff
@@ -0,0 +1,15 @@
+import { UiFinder, Waiter } from '@ephox/agar';
+import { SugarBody } from '@ephox/sugar';
+
+const pAssertToolbarNotVisible = async (): Promise<void> => {
+  await Waiter.pWait(50);
+  UiFinder.notExists(SugarBody.body(), '.tox-pop__dialog .tox-toolbar');
+};
+
+const pAssertToolbarVisible = (): Promise<void> =>
+  Waiter.pTryUntil('toolbar should exist', () => UiFinder.exists(SugarBody.body(), '.tox-toolbar'));
```
lnewson, line 10: Sorry Ben you beat me to it as I accidentally submitted the comment... I meant to add this as well otherwise you could pickup the regular toolbar:
```suggestion
  Waiter.pTryUntil('toolbar should exist', () => UiFinder.exists(SugarBody.body(), '.tox-pop__dialog .tox-toolbar'));
```

lnewson: This was like this previously though, so not caused by any change you made.

tiny-ben-tran: No worries.

modules/tinymce/src/plugins/quickbars/main/ts/insert/Toolbars.ts
lnewson: LGTM 👍

ephox-mogran (approved):

PR #8183, opened 2022-10-12
TINY-9180: Improve discoverability of the question mark read more link

shanmen-tiny:
Related Ticket: TINY-9180

Description of Changes:
* Setting icon alignment to be middle 
* Decrease font size of 'issue' and 'warning/error'
* Remove borders around descriptions
* Set warning and success title colours to meet the contrast ratio
* Setting the background colour of the 'question mark' button

Pre-checks:
* [x] ~~Changelog entry added~~ (Do we need changelog entry for changes made to premium plugin?)

Review:
* [x] Milestone set

Screenshot of UI changes: 
- The `info` state is to demo

**Light Mode:**
SCR-20221013-b5l
SCR-20221013-ayn
SCR-20221013-ayx
SCR-20221013-az3

**Dark Mode:**
SCR-20221013-b5b
SCR-20221013-b3q
SCR-20221013-b3u
SCR-20221013-b41

### DIFF
--- a/modules/oxide/src/less/theme/components/accessibility-checker/accessibility-checker.less
+++ b/modules/oxide/src/less/theme/components/accessibility-checker/accessibility-checker.less
@@ -5,22 +5,18 @@
 @accessibility-issue-info-heading-text-color: contrast(@dialog-background-color, @color-white, @color-tint);
 @accessibility-issue-info-body-text-color: contrast(@dialog-background-color, @color-white, @text-color);
 @accessibility-issue-info-background-color: contrast(@dialog-background-color, fade(@color-tint, 10%), fade(@color-tint, 50%));
-@accessibility-issue-info-border-color: contrast(@dialog-background-color, fade(@color-tint, 40%), @color-tint);
 
 @accessibility-issue-warn-heading-text-color: contrast(@dialog-background-color, @color-white, #cc8500);
 @accessibility-issue-warn-body-text-color: contrast(@dialog-background-color, @color-white, @text-color);
 @accessibility-issue-warn-background-color: contrast(@dialog-background-color, fade(orange, 10%), fade(orange, 50%));
-@accessibility-issue-warn-border-color: contrast(@dialog-background-color, fade(orange, 50%), fade(orange, 80%));
 
 @accessibility-issue-error-heading-text-color: contrast(@dialog-background-color, @color-white, @color-error);
 @accessibility-issue-error-body-text-color: contrast(@dialog-background-color, @color-white, @text-color);
 @accessibility-issue-error-background-color: contrast(@dialog-background-color, fade(@color-error, 10%), fade(@color-error, 50%));
-@accessibility-issue-error-border-color: contrast(@dialog-background-color, fade(@color-error, 40%), fade(@color-error, 80%));
 
 @accessibility-issue-success-heading-text-color: contrast(@dialog-background-color, @color-white, @color-success);
 @accessibility-issue-success-body-text-color: contrast(@dialog-background-color, @color-white, @text-color);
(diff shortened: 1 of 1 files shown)


lnewson:
For these sort of changes, it's often a good idea to include a screenshot to show how others it looks in the PR description.

modules/oxide/src/less/theme/components/accessibility-checker/accessibility-checker.less
```diff
@@ -108,21 +108,29 @@
         fill: @accessibility-issue-warn-heading-text-color;
       }
 
-      a .tox-icon {
-        color: @accessibility-issue-warn-heading-text-color;
+      /* stylelint-disable-next-line no-descending-specificity */
```
ephox-mogran, line 111: Should we just nest it inside `a.tox-button--naked` so that we don't have a disable a linting rule. Is this really that much easier to parse/read? Is that what the rule is complaining about?

modules/oxide/src/less/theme/components/accessibility-checker/accessibility-checker.less
```diff
@@ -47,12 +48,11 @@
         align-items: center;
         display: flex;
         margin-bottom: @pad-xs;
-      }
-    }
 
-    > *:last-child:not(:only-child) {
-      border-color: @border-color;
-      border-style: solid;
+        .tox-icon svg {
```
ephox-mogran, line 52: I probably didn't expect this much of a change, but I also haven't looked into it. What's the reason for this change?

modules/oxide/src/less/theme/components/accessibility-checker/accessibility-checker.less
```diff
@@ -83,21 +78,26 @@
         fill: @accessibility-issue-info-heading-text-color;
       }
 
-      a .tox-icon {
-        color: @accessibility-issue-info-heading-text-color;
+      a.tox-button--naked.tox-button--icon {
```
ephox-mogran, line 81: Why does this one not need to disable the linting rule?

modules/oxide/src/less/theme/components/accessibility-checker/accessibility-checker.less
shanmen-tiny: This is to set the icons to be middle aligned. The `svg` defaults to `inline`, it has an extra space which makes the icon a bit off.

Current: 
image

modules/oxide/src/less/theme/components/accessibility-checker/accessibility-checker.less
TheSpyder: I think it doesn't complain about `info` because it's first. If you swapped the order and put something else first then `info` would get the warning too.

Likely the only way to work around the warning is specify the 3 button colours and then repeat the 3 button selectors to set hover/focus/active later in the file. Given we know `error`/`warning`/`info` won't be used at the same time I guess suppressing the warning is fine.

TheSpyder (approved):

modules/oxide/src/less/theme/components/accessibility-checker/accessibility-checker.less
ephox-mogran: I wonder if we should split that into a separate ticket. This is a different way from how it is addressed in the notification banners, but maybe that doesn't matter. There, they seem to use grid and align-self etc. I guess it all depends on how many other `tox-icon`s we have around with this problem.

modules/oxide/src/less/theme/components/accessibility-checker/accessibility-checker.less
ephox-mogran: Fair enough. Withdrawn.

ephox-mogran (approved):

PR #8185, opened 2022-10-13
TINY-9186: Fix reported memory leak when parsing content

hamza0867:
Related Ticket: TINY-9186

Description of Changes:
* Fix memory leak caused in getContent method

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -14,7 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - New `color_default_foreground` and `color_default_background` options to set the initial default color for the `forecolor` and `backcolor` toolbar buttons and menu items. #TINY-9183
 
 ### Fixed
-- calling getContent method would cause a memory leak. #TINY-9186
+- Parsing media content would cause a memory leak, which for example occurred when using the `getContent` API. #TINY-9186
 - Dragging a noneditable element toward the bottom edge would cause the page to scroll up. #TINY-9025
 - Range expanding capabilities would behave inconsistently depending on where the cursor was placed. #TINY-9029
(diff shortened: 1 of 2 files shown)


modules/tinymce/CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - New `color_default_foreground` and `color_default_background` options to set the initial default color for the `forecolor` and `backcolor` toolbar buttons and menu items. #TINY-9183
 
 ### Fixed
+- calling getContent method would cause a memory leak. #TINY-9186
```
lnewson, line 17:
```suggestion
- Parsing media content would cause a memory leak, which for example occurred when using the `getContent` API. #TINY-9186
```
Maybe? Just to show it happening during parsing and only with media (e.g. videos). It also relates it to getting content 🤷

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
```diff
@@ -577,6 +577,10 @@ const DomParser = (settings: DomParserSettings = {}, schema = Schema()): DomPars
     const rootNode = new AstNode(rootName, 11);
     transferChildren(rootNode, element, schema.getSpecialElements());
 
+    // This next line is needed to fix memory leak in chrome and firefox.
```
lnewson, line 580:
```suggestion
    // This next line is needed to fix a memory leak in chrome and firefox.
```

lnewson (commented):
Looks good to me 👍 One thing you could do to write a test for this would be to do similar to the fiddle and parse a few thousand times. That however would be pretty slow so we maybe will just have to go without.

yacodes (approved):

TheSpyder (approved):

PR #8191, opened 2022-10-17
TINY-9213: Color picker dialog now defaults to approriate color 

HAFRMO:
Related Ticket: 

Description of Changes:
Color picker dialog uses cursor position for color.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -21,7 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Improved
 - Transparent elements, like anchors, can now contain block elements. #TINY-9172
-- Color picker dialog now starts on the appropriate color for the cursor position #TINY-9213
+- Color picker dialog now starts on the appropriate color for the cursor position. #TINY-9213
 
 ### Fixed
 - Parsing media content would cause a memory leak, which for example occurred when using the `getContent` API. #TINY-9186
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Improved
 - Transparent elements, like anchors, can now contain block elements. #TINY-9172
+- Color picker dialog now starts on the approriate color for the cursor position #TINY-9213
 
 ### Fixed
 - Parsing media content would cause a memory leak, which for example occurred when using the `getContent` API. #TINY-9186
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -21,7 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Improved
 - Transparent elements, like anchors, can now contain block elements. #TINY-9172
-- Color picker dialog now starts on the approriate color for the cursor position #TINY-9213
+- Color picker dialog now starts on the appropriate color for the cursor position #TINY-9213
 
 ### Fixed
 - Parsing media content would cause a memory leak, which for example occurred when using the `getContent` API. #TINY-9186
--- a/modules/tinymce/src/themes/silver/main/ts/ui/core/color/ColorSwatch.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/ui/core/color/ColorSwatch.ts
@@ -1,5 +1,5 @@
 import { HexColour, RgbaColour } from '@ephox/acid';
-import { Cell, Fun, Optional, Strings, Type } from '@ephox/katamari';
+import { Cell, Fun, Optional } from '@ephox/katamari';
 import { Css, SugarElement } from '@ephox/sugar';
 
 import Editor from 'tinymce/core/api/Editor';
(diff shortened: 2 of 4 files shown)


modules/tinymce/src/themes/silver/main/ts/ui/core/color/ColorSwatch.ts
```diff
@@ -83,13 +84,15 @@ const getAdditionalColors = (hasCustom: boolean): Menu.ChoiceMenuItemSpec[] => {
 const applyColor = (editor: Editor, format: ColorFormat, value: string, onChoice: (v: string) => void) => {
   if (value === 'custom') {
     const dialog = colorPickerDialog(editor);
+    const cssRgbValue = Css.get(SugarElement.fromDom(editor.selection.getStart()), format === 'hilitecolor' ? 'background-color' : 'color');
```
lnewson, line 87: Why not use the current `getCurrentColor` function above?

HAFRMO: Because for some reason I'd mentally filed it as checking the memory in options instead of the placement of the cursor. I'll update it, and use it.

hamza0867 (approved):

modules/tinymce/CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Improved
 - Transparent elements, like anchors, can now contain block elements. #TINY-9172
+- Color picker dialog now starts on the approriate color for the cursor position #TINY-9213
```
spocke, line 24:
```suggestion
- Color picker dialog now starts on the appropriate color for the cursor position. #TINY-9213
```

lorenzo-pomili (approved):

modules/tinymce/src/themes/silver/main/ts/ui/core/color/ColorSwatch.ts
```diff
@@ -117,12 +106,10 @@ const registerTextColorButton = (editor: Editor, name: string, format: ColorForm
...
-      const optCurrentRgb = getCurrentColor(editor, format);
-      return optCurrentRgb.bind((currentRgb) => RgbaColour.fromString(currentRgb).map((rgba) => {
-        const currentHex = HexColour.fromRgba(rgba).value;
-        // note: value = '#FFFFFF', currentHex = 'ffffff'
-        return Strings.contains(value.toLowerCase(), currentHex);
-      })).getOr(false);
+      const optCurrenthex = getCurrentColor(editor, format);
+      return optCurrenthex.map((currentHex) => {
+        return value.toUpperCase() === currentHex;
+      }).getOr(false);
```
lnewson, line 112: Typo here... `currenthex` isn't a word, so missing a capital `H`. We also could just slim this down.
```suggestion
      const optCurrentHex = getCurrentColor(editor, format);
      return optCurrentHex.exists((currentHex) => value.toUpperCase() === currentHex);
```
You could also use `Optionals.is` instead to make this even more readable, but I can do a suggestion for that. For example:
```
      const optCurrentHex = getCurrentColor(editor, format);
      return Optionals.is(optCurrentHex, value.toUpperCase());
```

modules/tinymce/src/themes/silver/main/ts/ui/core/color/ColorSwatch.ts
```diff
@@ -17,21 +18,9 @@ export interface ColorSwatchDialogData {
...
-  editor.dom.getParents(editor.selection.getStart(), (elm) => {
-    const value = hasStyleApi(elm) ? elm.style[format === 'forecolor' ? 'color' : 'backgroundColor'] : null;
-
-    if (value) {
-      color = color ? color : value;
-    }
-  });
-
-  return Optional.from(color);
+  const cssRgbValue = Css.get(SugarElement.fromDom(editor.selection.getStart()), format === 'hilitecolor' ? 'background-color' : 'color');
```
lnewson, line 22: One side effect of this is that previously it never used the computed style... it only used the raw styles (e.g. CSS stylesheets wouldn't impact this). Are we sure we want it to use the computed style instead here now?

HAFRMO: This was discussed with husky and decided that we will use the css stylesheets, meaning that the current code is the one we want to go with.

TheSpyder: I agree, I think the colour the user sees is the one that should be in the dialog.

modules/tinymce/src/themes/silver/main/ts/ui/core/color/ColorSwatch.ts
```diff
@@ -117,12 +106,8 @@ const registerTextColorButton = (editor: Editor, name: string, format: ColorForm
...
     icon: name === 'forecolor' ? 'text-color' : 'highlight-bg-color',
     select: (value) => {
-      const optCurrentRgb = getCurrentColor(editor, format);
-      return optCurrentRgb.bind((currentRgb) => RgbaColour.fromString(currentRgb).map((rgba) => {
-        const currentHex = HexColour.fromRgba(rgba).value;
-        // note: value = '#FFFFFF', currentHex = 'ffffff'
-        return Strings.contains(value.toLowerCase(), currentHex);
-      })).getOr(false);
+      const optCurrentHex = getCurrentColor(editor, format);
+      return Optionals.is(optCurrentHex, value.toUpperCase());
```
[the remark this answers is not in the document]
TheSpyder: Don't we have another feature in this release related to selecting the colour? I see "breaking change" and my immediate reaction here is to say not unless it's a major but I need to understand more about the context.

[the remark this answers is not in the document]
lnewson: So one thing to remember here is that `select` just marks the item as "checked"/"enabled" when the color swatch opens... so this is very similar to the dialog. Should the color be marked as enabled based on the raw or computed styles?

At the moment there is no CSS anyways for this so you can't see anything anyways (unless someone adds custom CSS outside the skin). So any change done here really won't make any visual difference anyways unless that's been done.

Another issue is this current code doesn't even work! As this appears to accidentally have been broken back in 2021 by normalizing the Hex color so it returns as uppercase which means atm it's always false: https://github.com/tinymce/tinymce/commit/097520477639312c12d58ceefdc693baf7d00c6b#diff-957d1482d9a8b718e7eae0a09aa8c4aaf65af2eaceb0018ac6cb9e26d20d471bR32 so this shows there are no tests.

modules/tinymce/src/themes/silver/test/ts/browser/editor/color/ColorPickerSanityTest.ts
```diff
@@ -13,9 +13,41 @@ describe('browser.tinymce.themes.silver.editor.color.ColorPickerSanityTest', ()
...
     { label: 'Shadow Dom Editor', setup: TinyHooks.bddSetupInShadowRoot }
   ], (tester) => {
+    context(`${tester.label}, Test of different color types`, () => {
+      const hook = tester.setup<Editor>({
+        base_url: '/project/tinymce/js/tinymce',
+        toolbar: 'forecolor backcolor',
+        content_style: 'p{color: pink; background-color: hsl(120, 100%, 75%);}'
+      }, []);
+
+      it('TINY-9213: Color detected on color, background', async () => {
```
spocke, line 23: Should all the tests also close/cancel the dialog? Does the test end up with multiple dialogs open at the same time?

spocke (approved):

PR #8200, opened 2022-10-19
TINY-9216: Add working example for toolbar button with updateable text

hamza0867:
Related Ticket: TINY-9216

Description of Changes:
*  Add working example for toolbar button with updateable text

To test the example:
1. open the code plugin example page
2. click on the button with the text `Before`
3. Notice how the text updates to become `After`

Pre-checks:
* [ ] Changelog entry added

Review:
* [ ] Milestone set

### DIFF
--- a/modules/tinymce/src/themes/silver/main/ts/ui/toolbar/button/ToolbarButtons.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/ui/toolbar/button/ToolbarButtons.ts
@@ -1,6 +1,7 @@
 import {
   AddEventsBehaviour, AlloyComponent, AlloyEvents, AlloyTriggers, Behaviour, Button as AlloyButton, Disabling, FloatingToolbarButton, Focusing,
-  Keying, NativeEvents, Reflecting, Replacing, SketchSpec, SplitDropdown as AlloySplitDropdown, SystemEvents, TieredData, TieredMenuTypes, Toggling,
+  GuiFactory,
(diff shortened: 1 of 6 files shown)


modules/tinymce/src/themes/silver/main/ts/ui/toolbar/button/ToolbarButtons.ts
```diff
@@ -79,6 +87,57 @@ const getTooltipAttributes = (tooltip: Optional<string>, providersBackstage: UiF
...
+            AlloyTriggers.emit(button, focusButtonEvent);
+          })
+        ])
+      ].concat(
+        receiver.map((r) => Reflecting.config({
+          channel: r,
+          initialData: { icon, text },
+          renderComponents: (data: ButtonState, _state) => componentRenderPipeline([
+            data.icon.map((iconName) => renderIconFromPack(iconName, providersBackstage.icons)),
+            data.text.map((text) => renderLabel(text, ToolbarButtonClasses.Button, providersBackstage))
```
ephox-mogran, line 133: I haven't looked at it properly yet, but at a glance, the thing that worries me is that in the UpdateMenuTextEvent handler, we are relying on the optMemDisplayText Memento to exist, but here on renderComponents, we don't keep that memento id. I'm actually not sure why you have `Reflecting` at all. And I think we probably don't want so much duplication between `renderCommonStructure` and `renderCommonStructureUpdateableText`.

I have a branch where I was cleaning up toolbar buttons because it's very messy and inconsistent (it was one section of the sink work that was shelved), and that might be the underlying problem - that the existing code is hard to follow. I'll have a proper look through this later, and maybe we can have a call about it so I can see what if there's a cleaner way to make this change.

PR #8205, opened 2022-10-21
TINY-9222: Fix scrolling to the cursor after undo or redo actions

yacodes:
Related Ticket: TINY-9222

Description of Changes:
* Fixed scrolling to the new cursor position after undo or redo action. This is fixed in `undo/Levels.ts:applyToEditor` since this function is called both from `undo` and `redo` actions. However it is also called in `undoManager.extra` function which I do not fully comprehend.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -36,7 +36,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The `editor.selection.getRng()` API was not returning a proper range on hidden editors in Firefox. #TINY-9259
 - The `editor.selection.getBookmark()` API was not returning a proper bookmark on hidden editors in Firefox. #TINY-9259
 - Dragging a noneditable element before or after another noneditable element now works correctly. #TINY-9253
-- Scrolling to the cursor after undo/redo. #TINY-9222
+- The restored selection after a redo or undo action was not scrolled into view. #TINY-9222
 
 ## 6.2.0 - 2022-09-08
 
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -36,6 +36,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The `editor.selection.getRng()` API was not returning a proper range on hidden editors in Firefox. #TINY-9259
 - The `editor.selection.getBookmark()` API was not returning a proper bookmark on hidden editors in Firefox. #TINY-9259
 - Dragging a noneditable element before or after another noneditable element now works correctly. #TINY-9253
+- Scrolling to the cursor after undo/redo. #TINY-9222
 
 ## 6.2.0 - 2022-09-08
 
--- a/modules/tinymce/src/core/main/ts/undo/Levels.ts
+++ b/modules/tinymce/src/core/main/ts/undo/Levels.ts
@@ -65,6 +65,8 @@ const applyToEditor = (editor: Editor, level: UndoLevel, before: boolean): void 
   if (bookmark) {
     editor.selection.moveToBookmark(bookmark);
   }
+
+  editor.selection.scrollIntoView();
 };
 
 const getLevelContent = (level: NewUndoLevel): string => {
--- a/modules/tinymce/src/core/main/ts/undo/Levels.ts
+++ b/modules/tinymce/src/core/main/ts/undo/Levels.ts
@@ -64,9 +64,8 @@ const applyToEditor = (editor: Editor, level: UndoLevel, before: boolean): void 
 
   if (bookmark) {
     editor.selection.moveToBookmark(bookmark);
+    editor.selection.scrollIntoView();
   }
-
-  editor.selection.scrollIntoView();
 };
 
 const getLevelContent = (level: NewUndoLevel): string => {
--- a/modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
+++ b/modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
@@ -689,7 +689,7 @@ describe('browser.tinymce.core.UndoManagerTest', () => {
     editor.focus();
 
     const height = 5000;
-    editor.resetContent(`<p class="first">top paragraph</p><p style="height: ${height}px"></p><p class="last">last paragraph</p>`);
+    editor.setContent(`<p class="first">top paragraph</p><p style="height: ${height}px"></p><p class="last">last paragraph</p>`);
     TinySelections.select(editor, 'p.last', [ 0 ]);
     TinyContentActions.type(editor, 'updated ');
 
--- a/modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
+++ b/modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
@@ -1,8 +1,9 @@
-import { Keys } from '@ephox/agar';
+import { Keys, UiFinder } from '@ephox/agar';
 import { beforeEach, context, describe, it } from '@ephox/bedrock-client';
 import { Arr, Fun } from '@ephox/katamari';
 import { PlatformDetection } from '@ephox/sand';
-import { LegacyUnit, TinyAssertions, TinyContentActions, TinyHooks, TinySelections } from '@ephox/wrap-mcagar';
+import { SugarElement, Scroll } from '@ephox/sugar';
+import { TinyDom, LegacyUnit, TinyAssertions, TinyContentActions, TinyHooks, TinySelections } from '@ephox/wrap-mcagar';
 import { assert } from 'chai';
 
 import Editor from 'tinymce/core/api/Editor';
@@ -682,4 +683,25 @@ describe('browser.tinymce.core.UndoManagerTest', () => {
     });
   });
 
+  context('TINY-9222: Scroll to the cursor after undo or redo actions', () => {
+    ([ 'undo', 'redo' ] as const).forEach((action) => {
+      it('TINY-9222: Scroll to the cursor after ' + action, () => {
+        const editor = hook.editor();
+        editor.setContent(`<p class="first">top paragraph</p>${'<br>'.repeat(50)}<p class="last">last paragraph</p>`);
+        TinySelections.select(editor, 'p.last', [ 0 ]);
+        TinyContentActions.type(editor, 'updated ');
+
+        const doc = TinyDom.document(editor);
+        Scroll.to(0, 0, doc);
+        editor.undoManager[action]();
+        const yPos = Scroll.get(doc).top;
+        assert.notEqual(0, yPos, 'should change scroll after undo');
+
+        const el = UiFinder.findIn(doc, 'p.last').getOrDie();
+        Scroll.intoViewIfNeeded(el, SugarElement.fromDom(doc.dom.scrollingElement as Element));
+        const newYPos = Scroll.get(doc).top;
+        assert.equal(newYPos, yPos, 'should scroll to the cursor after undo');
+      });
+    });
+  });
 });
--- a/modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
+++ b/modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
@@ -685,11 +685,9 @@ describe('browser.tinymce.core.UndoManagerTest', () => {
 
   it('TINY-9222: Scroll to the cursor after undo and redo', () => {
     const editor = hook.editor();
-    editor.undoManager.clear();
-    editor.focus();
 
     const height = 5000;
-    editor.setContent(`<p class="first">top paragraph</p><p style="height: ${height}px"></p><p class="last">last paragraph</p>`);
+    editor.resetContent(`<p class="first">top paragraph</p><p style="height: ${height}px"></p><p class="last">last paragraph</p>`);
     TinySelections.select(editor, 'p.last', [ 0 ]);
     TinyContentActions.type(editor, 'updated ');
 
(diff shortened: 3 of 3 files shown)


modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
```diff
@@ -682,4 +683,25 @@ describe('browser.tinymce.core.UndoManagerTest', () => {
...
+        const doc = TinyDom.document(editor);
+        Scroll.to(0, 0, doc);
+        editor.undoManager[action]();
+        const yPos = Scroll.get(doc).top;
+        assert.notEqual(0, yPos, 'should change scroll after undo');
+
+        const el = UiFinder.findIn(doc, 'p.last').getOrDie();
+        Scroll.intoViewIfNeeded(el, SugarElement.fromDom(doc.dom.scrollingElement as Element));
+        const newYPos = Scroll.get(doc).top;
+        assert.equal(newYPos, yPos, 'should scroll to the cursor after undo');
```
vpyshnenko, line 703: replace `undo` in messages with $action placeholder

modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
```diff
@@ -682,4 +683,25 @@ describe('browser.tinymce.core.UndoManagerTest', () => {
...
+    ([ 'undo', 'redo' ] as const).forEach((action) => {
+      it('TINY-9222: Scroll to the cursor after ' + action, () => {
+        const editor = hook.editor();
+        editor.setContent(`<p class="first">top paragraph</p>${'<br>'.repeat(50)}<p class="last">last paragraph</p>`);
+        TinySelections.select(editor, 'p.last', [ 0 ]);
+        TinyContentActions.type(editor, 'updated ');
+
+        const doc = TinyDom.document(editor);
+        Scroll.to(0, 0, doc);
+        editor.undoManager[action]();
```
vpyshnenko, line 696: on `redo` action phase: how do you test `redo` without the previous `undo`? It seems strange here that `redo` after the update scrolls content down since nothing to redo.

vpyshnenko (changes requested):

modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
```diff
@@ -682,4 +683,25 @@ describe('browser.tinymce.core.UndoManagerTest', () => {
...
+        const yPos = Scroll.get(doc).top;
+        assert.notEqual(0, yPos, 'should change scroll after undo');
+
+        const el = UiFinder.findIn(doc, 'p.last').getOrDie();
```
spocke, line 700: Maybe the assert could just be that the yPos is greater than 5000 if you put a 5000px height. Instead of doing another scroll to compare against.

spocke (approved):

modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
spocke: Yeah, maybe do both in one test. Make a change, undo, check scroll pos, redo check scroll pos or something.

modules/tinymce/CHANGELOG.md
```diff
@@ -36,6 +36,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The `editor.selection.getRng()` API was not returning a proper range on hidden editors in Firefox. #TINY-9259
 - The `editor.selection.getBookmark()` API was not returning a proper bookmark on hidden editors in Firefox. #TINY-9259
 - Dragging a noneditable element before or after another noneditable element now works correctly. #TINY-9253
+- Scrolling to the cursor after undo/redo. #TINY-9222
```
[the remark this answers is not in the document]
lnewson: As a side note, `/` should not be used in changelogs or docs either as it's unclear if it means `and` or `or`, plus doesn't translate well. So use one of them specifically instead. Taking Viktor's suggestion and fixing that, maybe something like this then?
```suggestion
- The restored selection after a redo or undo action was not scrolled into view. #TINY-9222
```

modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
```diff
@@ -682,4 +683,25 @@ describe('browser.tinymce.core.UndoManagerTest', () => {
     });
   });
 
+  context('TINY-9222: Scroll to the cursor after undo or redo actions', () => {
```
lnewson, line 686: Jira numbers should not be used in a `context` block because the context is about a type of action not a specific Jira like an actual test (plus they could contain multiple tests for different Jiras). In all honesty I don't see a need for a `context` here, especially when it's the same as the `it` block text so it should probably be removed entirely.

modules/tinymce/CHANGELOG.md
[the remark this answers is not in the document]
yacodes: Thanks for the suggestions! Reworded [here](https://github.com/tinymce/tinymce/pull/8205/commits/d1506b5662ac1ffee3e9146b8c5da7679ab6839b)

modules/tinymce/src/core/main/ts/undo/Levels.ts
```diff
@@ -65,6 +65,8 @@ const applyToEditor = (editor: Editor, level: UndoLevel, before: boolean): void
   if (bookmark) {
     editor.selection.moveToBookmark(bookmark);
   }
+
+  editor.selection.scrollIntoView();
```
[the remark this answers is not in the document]
yacodes: Updated [here](https://github.com/tinymce/tinymce/pull/8205/commits/47f1ba57e090f27f4a41f16ab6e6711f7488ee2a)

modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
yacodes: Updated [here](https://github.com/tinymce/tinymce/pull/8205/commits/d03dcd64f8958af59ee6b7984f043033afb135d8)

modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
yacodes: Yeah, that makes sense. However, I believe that the `undoManager` state was persistent and `redo` actually worked! Weird. Updated to a single test [here](https://github.com/tinymce/tinymce/pull/8205/commits/d03dcd64f8958af59ee6b7984f043033afb135d8)

modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
```diff
@@ -682,4 +683,25 @@ describe('browser.tinymce.core.UndoManagerTest', () => {
...
+    ([ 'undo', 'redo' ] as const).forEach((action) => {
+      it('TINY-9222: Scroll to the cursor after ' + action, () => {
+        const editor = hook.editor();
+        editor.setContent(`<p class="first">top paragraph</p>${'<br>'.repeat(50)}<p class="last">last paragraph</p>`);
```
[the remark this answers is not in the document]
yacodes: Thanks for the suggestion, that's much better! Updated [here](https://github.com/tinymce/tinymce/pull/8205/commits/d03dcd64f8958af59ee6b7984f043033afb135d8)

modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
yacodes: Updated [here](https://github.com/tinymce/tinymce/pull/8205/commits/d03dcd64f8958af59ee6b7984f043033afb135d8)

modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
yacodes: Removed the `context` block whatsoever [here](https://github.com/tinymce/tinymce/pull/8205/commits/d03dcd64f8958af59ee6b7984f043033afb135d8)

lnewson:
@yacodes just a reminder that you should be setting the milestone here as per the PR description checklist based on what the `Fix version` is in Jira (it's only needed for public projects like TinyMCE). You should be able to cross out the docs item as well since this shouldn't need a separate docs ticket.

modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
```diff
@@ -682,4 +683,26 @@ describe('browser.tinymce.core.UndoManagerTest', () => {
     });
   });
 
+  it('TINY-9222: Scroll to the cursor after undo and redo', () => {
+    const editor = hook.editor();
+    editor.undoManager.clear();
+    editor.focus();
+
+    const HEIGHT = 5000;
```
lnewson, line 691: This should be lower case as this uppercase naming format isn't used at Tiny for constants (there was a thread about this yesterday in a premium library PR). As @spocke there though a lint rule really should be added to enforce this so we probably need a tech debt Jira logged to follow up.

modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
```diff
@@ -682,4 +683,26 @@ describe('browser.tinymce.core.UndoManagerTest', () => {
...
+    editor.resetContent(`<p class="first">top paragraph</p><p style="height: ${HEIGHT}px"></p><p class="last">last paragraph</p>`);
```
vpyshnenko, line 692: `editor.resetContent` clears undo history internally

modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
yacodes: Updated [here](https://github.com/tinymce/tinymce/pull/8205/commits/242f06e4430020e4e065f3ea6957594e18f5a9c1)

yacodes:
> @yacodes just a reminder that you should be setting the milestone here as per the PR description checklist based on what the `Fix version` is in Jira (it's only needed for public projects like TinyMCE). You should be able to cross out the docs item as well since this shouldn't need a separate docs ticket.

Thanks for the reminder! Added the milestone.

modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
yacodes: @vpyshnenko Not sure what should I update here, that's kind of the point. Can I ask you to elaborate, please?

lnewson: I wonder if this means you don't need the `editor.undoManager.clear()` above since `resetContent()` already does that?

lnewson: On that note, the `editor.focus()` should also be unnecessary the `TinyHooks` logic is already configured to focus the editor on init.

yacodes: Oh, I see. Replaced `resetContent` with `setContent` to explicitly underline the intention of clearing the `undoManager` [here](https://github.com/tinymce/tinymce/pull/8205/commits/6ce4ec0751aacae05b4a1819b09be4689d7ad628).

lnewson: @yacodes that's not the right thing to do... you should remove the `clear` call instead as now you have a case where there is no initial undo state which definitely isn't what we want here.

modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
```diff
@@ -682,4 +683,26 @@ describe('browser.tinymce.core.UndoManagerTest', () => {
...
+    TinySelections.select(editor, 'p.last', [ 0 ]);
+    TinyContentActions.type(editor, 'updated ');
+
+    const doc = TinyDom.document(editor);
+    const editorHeight = editor.getWin().innerHeight;
+
+    const checkScroll = (action: 'undo' | 'redo') => {
+      Scroll.to(0, 0, doc);
+      editor.undoManager[action]();
+      Scroll.intoViewIfNeeded(UiFinder.findIn(doc, 'p.last').getOrDie(), SugarElement.fromDom(doc.dom.scrollingElement as Element));
```
vpyshnenko, line 702: It looks like  `Scroll.intoViewIfNeeded` call here is redundant since this should be done during undo operation. Try to remove it. The test should pass without it.

modules/tinymce/src/core/test/ts/browser/UndoManagerTest.ts
yacodes: @lnewson I do not understand. That's exactly what I want:
1. Start with an empty undo state;
2. Add content to the editor to add undo state;
3. Scroll up;
4. Undo;
5. Check if the scroll is down.

Without `editor.undoManager.clear()` the `undo` state is not empty I believe. Other tests are clearing the undo state as well.

lnewson: @yacodes that's not what you want... take a look at a clean editor, it will always have one value in the editor state (the initial state). If it has zero then it's in an invalid state and `undo` will not be allowed/work properly until there are at least 2 data objects in the undo manager. For example, in fiddle run clear and then click the bold button and you'll see the undo button is still disabled. So using `resetContent` is what you want here as it will put the editor back into a "clean" state where you cannot undo or redo without a change.

So while the steps are what I'd expect, the first step isn't right/what's happening here and to correct that using `resetContent` is the right API to use. Also remember to remove `editor.focus()` as I said above unless there's a good reason for it.

yacodes: @lnewson I see the difference now, thanks for the clarification. Updated [here](https://github.com/tinymce/tinymce/pull/8205/commits/2e4b85db39f93e3d2a815f21a8bee3cb80e687cc)

vpyshnenko (approved):

spocke (approved):

PR #8215, opened 2022-10-25
TINY-9143: Add tests for insert content and get content events

yacodes:
Related Ticket: TINY-9143

Description of Changes:
* Added tests to check if content manipulation in `BeforeSetContent` callback is possible;
* Added tests to check if custom data object can be passed to all get content and set content events.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Improved
 - Transparent elements, like anchors, can now contain block elements. #TINY-9172
+- Tests for insert content and get content events. #TINY-9143
 
 ### Fixed
 - Parsing media content would cause a memory leak, which for example occurred when using the `getContent` API. #TINY-9186
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -18,10 +18,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - Transparent elements, like anchors, are now allowed in the root of the editor body if they contain blocks. #TINY-9172
+- `setContent` is now allowed to accept any custom keys and values as a second options argument. #TINY-9143
 
 ### Improved
 - Transparent elements, like anchors, can now contain block elements. #TINY-9172
-- Tests for insert content and get content events. #TINY-9143
 
 ### Fixed
 - Parsing media content would cause a memory leak, which for example occurred when using the `getContent` API. #TINY-9186
--- a/modules/tinymce/src/core/main/ts/content/ContentTypes.ts
+++ b/modules/tinymce/src/core/main/ts/content/ContentTypes.ts
@@ -22,6 +22,7 @@ export interface SetContentArgs {
   paste?: boolean;
   load?: boolean;
   initial?: boolean;
+  [key: string]: any;
 }
 
 export interface SetContentResult {
--- a/modules/tinymce/src/core/test/ts/browser/content/EditorContentEventsTest.ts
+++ b/modules/tinymce/src/core/test/ts/browser/content/EditorContentEventsTest.ts
@@ -123,4 +123,28 @@ describe('browser.tinymce.core.content.EditorContentEventsTest', () => {
     assertEvents([]);
     TinyAssertions.assertContent(editor, '<p>Selection content</p>');
   });
+
+  const data = { hello: 'world', test: 132 };
+  Arr.each([ 'BeforeSetContent', 'SetContent' ], (action) => {
+    it(`TINY-9143: Can pass custom data object to "${action}" event`, () => {
+      const editor = hook.editor();
+      editor.setContent('<p>initial</p>');
+      editor.once(action, (e) => {
+        assert.equal(data.hello, e.hello);
+        assert.equal(data.test, e.test);
+      });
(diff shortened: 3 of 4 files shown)


modules/tinymce/CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Improved
 - Transparent elements, like anchors, can now contain block elements. #TINY-9172
+- Tests for insert content and get content events. #TINY-9143
```
spocke, line 24: Remove this since tests are not something that affects the release. So a integrator would not see anything different in the usage of tinymce based on this change.

- [ ] Remove changelog for tests change

modules/tinymce/src/core/test/ts/browser/content/EditorContentEventsTest.ts
```diff
@@ -123,4 +123,28 @@ describe('browser.tinymce.core.content.EditorContentEventsTest', () => {
...
+  const data = { hello: 'world', test: 132 };
+  Arr.each([ 'BeforeSetContent', 'SetContent' ], (action) => {
+    it(`TINY-9143: Can pass custom data object to "${action}" event`, () => {
+      const editor = hook.editor();
+      editor.setContent('<p>initial</p>');
+      editor.once(action, (e) => {
+        assert.equal(data.hello, e.hello);
+        assert.equal(data.test, e.test);
+      });
+      editor.setContent('<p>new</p>', data);
```
spocke, line 136: This will pass even if the event never fires since the assert is inside the event handler. It would be more safe something like this:

```suggestion
      const editor = hook.editor();
      const lastEventState = Singleton.value<SetContentEvent>();
      editor.setContent('<p>initial</p>');
      editor.once(action, (e) => {
        lastEventState.set(e);
      });
      editor.setContent('<p>new</p>', data);
      const lastEvent = lastEvent.get().getOrDie('Should be set');
      assert.equal(lastEvent.hello, data.hello);
      assert.equal(lastEvent.test, data.test);
```

Or maybe use a let and null not since it's just test code we could be a bit more relaxed. I also flipped the assert.equal since I think it's (actual, expected) in chai.

modules/tinymce/src/core/test/ts/browser/content/EditorContentTest.ts
```diff
@@ -269,6 +269,25 @@ describe('browser.tinymce.core.content.EditorContentTest', () => {
           const tree = editor.getContent({ format: 'tree' });
           assert.equal(toHtml(tree), '<p>replaced</p>', 'Should be replaced html');
         });
+
+        const INITIAL_CONTENT = '<p>initial</p>';
+        const NEW_CONTENT = '<p>new content</p>';
+        const MANIPULATED_CONTENT = '<p>manipulated</p>';
```
spocke, line 275: Camelcase these seems that we settled on using camel case.

modules/tinymce/src/core/main/ts/content/ContentTypes.ts
```diff
@@ -22,6 +22,7 @@ export interface SetContentArgs {
   paste?: boolean;
   load?: boolean;
   initial?: boolean;
+  [key: string]: any;
```
spocke, line 25: Does this end up as a public api change. Then maybe that needs to be logged since that does affect what types of events you can pass into the content apis. I wonder if we even want to be relaxed about this. The code allows anything but that is maybe more of a side effect of how this is implemented.

yacodes: What are the implications of being strict here? I do not think that passing custom keys can break something, but it can satisfy integrators' and plugin-developers' needs. Maybe it makes sense to have similar behavior in both `GetContentArgs` and `SetContentArgs`?

modules/tinymce/src/core/test/ts/browser/content/EditorContentEventsTest.ts
yacodes: `Singleton` solution is clear and readable, thanks a lot! Updated [here](https://github.com/tinymce/tinymce/pull/8215/commits/f1d9ca09b1fea6a3c05996853cedcce6b7bd489b)

modules/tinymce/src/core/test/ts/browser/content/EditorContentTest.ts
yacodes: Sorry, forgot to change. Get a bit used to such constants wording. Updated [here](https://github.com/tinymce/tinymce/pull/8215/commits/bd4f3b04de484d0738dc2fc71d59cf2b2f89a8cd).

modules/tinymce/src/core/main/ts/content/ContentTypes.ts
spocke: The problem could be that we then invite people to add custom things here that might clash with things we want to add our self later. But since we have no runtime validation of this and it can be used in JS projects I guess it's already very relaxed here. But if this is a public api change we need a changelog of this and not the fact that we added tests. Something that describes that the types now accept custom properties.

modules/tinymce/CHANGELOG.md
yacodes: Removed this line [here](https://github.com/tinymce/tinymce/pull/8215/commits/014f9e2f3f9121a3f0f12c70c4bfd5183a223999)

TheSpyder (approved):

vpyshnenko (approved):

PR #8218, opened 2022-10-27
TINY-9258: Update to waluigi 7 and run tests on Windows 11

lnewson:
Related Ticket: TINY-9258

Description of Changes:
* Update to Waluigi 7
* Update the builds to run using the `windows` label instead which will allow them to run on Windows 11
* Modernize the Jenkinsfile a little to match changes done in newer Waluigi versions. This means tests will now show as "unstable" (orange) instead of "failed" (red) when a test doesn't pass

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [x] Milestone set

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,7 +1,9 @@
 #!groovy
-@Library('waluigi@v4.5.0') _
+@Library('waluigi@release/7') _
 
-def runTests(name, bedrockCommand, runAll) {
+standardProperties()
+
+def runTests(String name, String bedrockCommand, Boolean runAll) {
   // Clean out the old XML files before running tests, since we junit import *.XML files
   dir('scratch') {
     if (isUnix()) {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -12,18 +14,20 @@ def runTests(name, bedrockCommand, runAll) {
...
   junit allowEmptyResults: true, testResults: 'scratch/TEST-*.xml'
 
-  if (!successfulTests) {
-    echo "Tests failed for " + name + " so passing failure as exit code for node: $NODE_NAME"
-    exec("exit 1")
+  // If the tests failed (exit code 4) then just mark it as unstable
+  if (testStatus == 4) {
+    unstable("Tests failed for ${name}")
+  } else if (testStatus != 0) {
+    error("Unexpected error running tests for ${name} so passing failure as exit code")
```
spocke, line 26: Maybe change it to this since the old code had a exit 1 so then it kind of described what the code below it did. Now it's just an error described by the function call it self.

```suggestion
    error("Unexpected error running tests for ${name}")
```

spocke (approved):

Jenkinsfile
lnewson: It's just a 1:1 copy of waluigi atm. I wanted to keep it the same, but I don't have a strong attachment to that so if you still want to I'll change it here.

Jenkinsfile
```diff
@@ -12,18 +14,20 @@ def runTests(name, bedrockCommand, runAll) {
...
 
-  echo "Writing JUnit results for " + name + " on node: $NODE_NAME"
+  echo "Writing JUnit results for ${name} on node: $NODE_NAME"
   junit allowEmptyResults: true, testResults: 'scratch/TEST-*.xml'
 
-  if (!successfulTests) {
-    echo "Tests failed for " + name + " so passing failure as exit code for node: $NODE_NAME"
-    exec("exit 1")
+  // If the tests failed (exit code 4) then just mark it as unstable
+  if (testStatus == 4) {
```
ephox-mogran, line 23: is `4` something we should have as a constant in waluigi somewhere or is that overkill?

ephox-mogran (approved):

Jenkinsfile
lnewson: Sounds like overkill to me... if we did want something in Waluigi I'd abstract the entire function instead. That was something I considered, but it was likely this previously and isn't really in scope of just upgrading.

lnewson: I'll log a tech debt ticket and throw it in the editor team backlog for this, as it should probably be cleaned up. It needs a function similar to `bedrockTests` where it allows a command to be passed in instead basically, as that's why it can't be used atm.

ltrouton (approved):

Jenkinsfile
lnewson: Logged as TINY-9313.

Jenkinsfile
spocke: Yes but the message is like throw new Error('Error: something') kind of the same information twice an error call is kind of implied that it will produce an error code. But I have no strong opinion here you could just leave it as is just looked a bit funny.

lnewson: I'm just going to leave it then so it's clear it's a copy. There's TINY-9313 for you all to remove the duplication anyways.

PR #8221, opened 2022-10-27
TINY-9308: Focus editor before reversing selection in visualchars PluginTest

mkzhx:
Related Ticket: 
TINY-9308

Description of Changes:
* Add `editor.focus()` before an `editor.setRng` that reverses current selection to address browser quirk in Firefox where editor loses focus when toolbar button is clicked

Pre-checks:
* [x] ~~Changelog entry added~~ - not applicable as involves fixing tests
* [x] ~~Tests have been added (if applicable)~~ - not applicable as involves fixing existing tests

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/plugins/visualchars/test/ts/browser/PluginTest.ts
+++ b/modules/tinymce/src/plugins/visualchars/test/ts/browser/PluginTest.ts
@@ -56,6 +56,7 @@ describe('browser.tinymce.plugins.visualchars.PluginTest', () => {
     editor.setContent('<p>abc&nbsp;&nbsp;</p>');
 
     TinySelections.setSelection(editor, [ 0, 0 ], 0, [ 0, 0 ], 3);
+    editor.focus();
     editor.selection.setRng(editor.selection.getRng(), false);
 
     TinyUiActions.clickOnToolbar(editor, 'button');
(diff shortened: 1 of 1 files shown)


modules/tinymce/src/plugins/visualchars/test/ts/browser/PluginTest.ts
```diff
@@ -56,6 +56,7 @@ describe('browser.tinymce.plugins.visualchars.PluginTest', () => {
     editor.setContent('<p>abc&nbsp;&nbsp;</p>');
 
     TinySelections.setSelection(editor, [ 0, 0 ], 0, [ 0, 0 ], 3);
+    editor.focus();
```
MitchCTiny, line 59: I'm sure this is fine, but it would probably be better to set focus right after instantiating the hook.

mkzhx: Focus is only required for the `setRng` method though, I felt this was more explicit. From my testing the test behaviours between setting focus right after instantiating the hook and placing it right before `setRng` are the same.

spocke: Maybe move the focus call to the start of the test function or use focusOnInit on the hook since the other tests doesn't seem to depend on focus being there since they all block on toolbars buttons first.

spocke (approved):

modules/tinymce/src/plugins/visualchars/test/ts/browser/PluginTest.ts
mkzhx: Yeah fair, thinking about it using focusOnInit would probably be more elegant. Mitch has also mentioned that before.

ltrouton (approved):

MitchCTiny (approved):

PR #8223, opened 2022-10-27
DOC-1757: Improve documentation of overrideDefaults editor method

MitchCTiny:
Related Ticket: DOC-1757

Previous PR: https://github.com/tinymce/tinymce-docs/pull/2554

Description of Changes:
* Update EditorManager API docs to improve description of `overrideDefaults` method, with example
* Document the `defaultOptions` property
* Credit to @kemister85 and @lnewson for the new documentation

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/api/EditorManager.ts
+++ b/modules/tinymce/src/core/main/ts/api/EditorManager.ts
@@ -128,7 +128,7 @@ const EditorManager: EditorManager = {
   baseURL: null as any,
 
   /**
-   * Default options for TinyMCE.
+   * Object which contains required, and user-specific, default values for cloud-related options.
    *
    * @property defaultOptions
    * @type Object
--- a/modules/tinymce/src/core/main/ts/api/EditorManager.ts
+++ b/modules/tinymce/src/core/main/ts/api/EditorManager.ts
@@ -278,11 +278,11 @@ const EditorManager: EditorManager = {
   },
 
   /**
-   * Overrides the default options for editor instances. The <code>overrideDefaults</code> method replaces any of the defaults set by a previous call to the <code>overrideDefaults</code> function.
+   * Overrides the default options for editor instances. The <code>overrideDefaults</code> method replaces the <code>defaultOptions</code>, including any set by a previous call to the <code>overrideDefaults</code> method.
    * <br /><br />
    * When using the Tiny Cloud, some of these defaults are required for the cloud-based editor to function.
    * <br /><br />
-   * Therefore, when using <code>overrideDefaults</code> with the cloud-based editor, combine the previous defaults, <code>tinymce.defaultOptions</code>, with the new options that you want to integrate.
+   * Therefore, when using <code>overrideDefaults</code> with the cloud-based editor, any newly integrated options must be combined with the options in <code>tinymce.defaultOptions</code>.
    *
    * @method overrideDefaults
    * @param {Object} defaultOptions Default options object.
--- a/modules/tinymce/src/core/main/ts/api/EditorManager.ts
+++ b/modules/tinymce/src/core/main/ts/api/EditorManager.ts
@@ -278,7 +278,7 @@ const EditorManager: EditorManager = {
   },
 
   /**
-   * Overrides the default options for editor instances. <code>overrideDefaults</code> replaces any of the defaults set by a previous call to the <code>overrideDefaults</code> function.
+   * Overrides the default options for editor instances. The <code>overrideDefaults</code> method replaces any of the defaults set by a previous call to the <code>overrideDefaults</code> function.
    * <br /><br />
    * When using the Tiny Cloud, some of these defaults are required for the cloud-based editor to function.
    * <br /><br />
@@ -288,7 +288,7 @@ const EditorManager: EditorManager = {
    * @param {Object} defaultOptions Default options object.
    * @example
    * const customOptions = {
-   *   promotion: true
+   *   toolbar_sticky: true
    * };
    *
    * tinymce.overrideDefaults({
--- a/modules/tinymce/src/core/main/ts/api/EditorManager.ts
+++ b/modules/tinymce/src/core/main/ts/api/EditorManager.ts
@@ -126,7 +126,15 @@ const EditorManager: EditorManager = {
 
   baseURI: null as any,
   baseURL: null as any,
+
+  /**
+   * Default options for TinyMCE.
+   *
+   * @property defaultOptions
+   * @type Object
+   */
   defaultOptions: {},
+
   documentBaseURL: null as any,
   suffix: null as any,
 
@@ -272,7 +280,7 @@ const EditorManager: EditorManager = {
   /**
    * Overrides the default options for editor instances. <code>overrideDefaults</code> replaces any of the defaults set by a previous call to the <code>overrideDefaults</code> function.
    * <br /><br />
-   * When using the cloud, some of these defaults are required for the cloud-based editor to function.
+   * When using the Tiny Cloud, some of these defaults are required for the cloud-based editor to function.
    * <br /><br />
    * Therefore, when using <code>overrideDefaults</code> with the cloud-based editor, combine the previous defaults, <code>tinymce.defaultOptions</code>, with the new options that you want to integrate.
    *
--- a/modules/tinymce/src/core/main/ts/api/EditorManager.ts
+++ b/modules/tinymce/src/core/main/ts/api/EditorManager.ts
@@ -128,7 +128,7 @@ const EditorManager: EditorManager = {
   baseURL: null as any,
 
   /**
-   * Object containing the options that will be passed by default to the `init` method upon each initialization of an editor.
+   * Object containing the default options that will be shallow merged with the options object passed to the `init` method upon each initialization of an editor.
    *
    * @property defaultOptions
    * @type Object
--- a/modules/tinymce/src/core/main/ts/api/EditorManager.ts
+++ b/modules/tinymce/src/core/main/ts/api/EditorManager.ts
@@ -270,10 +270,23 @@ const EditorManager: EditorManager = {
   },
 
   /**
-   * Overrides the default options for editor instances.
+   * Overrides the default options for editor instances. <code>overrideDefaults</code> replaces any of the defaults set by a previous call to the <code>overrideDefaults</code> function.
+   * <br /><br />
+   * When using the cloud, some of these defaults are required for the cloud-based editor to function.
+   * <br /><br />
+   * Therefore, when using <code>overrideDefaults</code> with the cloud-based editor, combine the previous defaults, <code>tinymce.defaultOptions</code>, with the new options that you want to integrate.
    *
    * @method overrideDefaults
-   * @param {Object} defaultOptions Defaults options object.
+   * @param {Object} defaultOptions Default options object.
+   * @example
+   * const customOptions = {
+   *   promotion: true
+   * };
+   *
+   * tinymce.overrideDefaults({
+   *   ...tinymce.defaultOptions,
+   *   ...customOptions
+   * });
    */
   overrideDefaults(defaultOptions) {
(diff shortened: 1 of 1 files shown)


shanmen-tiny (approved):

modules/tinymce/src/core/main/ts/api/EditorManager.ts
```diff
@@ -270,10 +270,23 @@ const EditorManager: EditorManager = {
   },
 
   /**
-   * Overrides the default options for editor instances.
+   * Overrides the default options for editor instances. <code>overrideDefaults</code> replaces any of the defaults set by a previous call to the <code>overrideDefaults</code> function.
+   * <br /><br />
+   * When using the cloud, some of these defaults are required for the cloud-based editor to function.
```
lnewson, line 275: Just copying across a fix from what @brian-forte said in the linked PR.
```suggestion
   * When using the Tiny Cloud, some of these defaults are required for the cloud-based editor to function.
```

modules/tinymce/src/core/main/ts/api/EditorManager.ts
```diff
@@ -270,10 +270,23 @@ const EditorManager: EditorManager = {
...
    * @method overrideDefaults
-   * @param {Object} defaultOptions Defaults options object.
+   * @param {Object} defaultOptions Default options object.
+   * @example
+   * const customOptions = {
+   *   promotion: true
+   * };
+   *
+   * tinymce.overrideDefaults({
+   *   ...tinymce.defaultOptions,
```
spocke, line 287: If we are exposing this one then that needs to be documented as well. Since that is now public api and refereed to from the docs at this point.

spocke (approved):

modules/tinymce/src/core/main/ts/api/EditorManager.ts
ephox-mogran: My recollection is that there is also an asciidoc attribute for `Tiny Cloud`.

Yep: `cloudname` (source: https://github.com/tinymce/tinymce-docs/blob/staging/docs-6/antora.yml#L27)

modules/tinymce/src/core/main/ts/api/EditorManager.ts
ephox-mogran: @kemister85 we'll need to create another ticket for documenting `tinymce.defaultOptions` (and `tinymce.defaultSettings` for v5)

modules/tinymce/src/core/main/ts/api/EditorManager.ts
```diff
@@ -270,10 +270,23 @@ const EditorManager: EditorManager = {
   },
 
   /**
-   * Overrides the default options for editor instances.
+   * Overrides the default options for editor instances. <code>overrideDefaults</code> replaces any of the defaults set by a previous call to the <code>overrideDefaults</code> function.
```
ephox-mogran, line 281: Is it a problem that a lowercased `overrideDefaults` starts a sentence here? Or is that acceptable because it's code?

ephox-mogran (approved):

modules/tinymce/src/core/main/ts/api/EditorManager.ts
lnewson: You cannot use that in Moxiedoc though, as it's not AsciiDoc based nor is that variable available to it.

ephox-mogran: Oh. Ignore me. I thought this was asciidoc :)

modules/tinymce/src/core/main/ts/api/EditorManager.ts
brian-forte: > Is it a problem that a lowercased `overrideDefaults` starts a sentence here? Or is that acceptable because it’s code?

Ooh, look, a can asking to be opened. I wonder if there are worms inside?

Fun fact: modern English capitalisation practice largely derives from the [First Folio](https://en.wikipedia.org/wiki/First_Folio), the edition of Shakespeare’s plays published in 1623, seven years after the playwrights’s death.

It is broadly believed the capitalisation practices used therein (always for the first word of a sentence and only for nouns that named specific things or people; what we now call Proper Nouns) was reflective of Shakespeare’s personal preference, and that he acquired that preference because the community of theatre writers and journalists he was a part of had been pushing for this approach during his working life.

During Shakespeare’s working life (and in the centuries before Shakespeare’s life), English capitalisation practices were inconsistent, to say the least.

And, because Shakespeare’s work is on par with the KJV as a source of authority when it comes to formal and prestige English usage, the _always capitalise the first word of a sentence_ practice is regularly argued for with doctrinal zeal.

But there have long been exceptions, even among the true believers.

[Tussenvoegsel’s](https://en.wikipedia.org/wiki/Tussenvoegsel), for example.

Also names that begin with the _ﬀ_ (the digraph f). This glyph exists for two reasons. First, because written English in the Middle Ages didn’t have the _F_ glyph. In this case, strictly speaking, _ﬀ_ and _F_ are equivalent glyphs.

But _ﬀ_ is also used to spell Welsh names that begin with a fricative labiodental (represented by **f** in the IPA). This is because the _F_ glyph represents the **v** sound in written Welsh.

So, given exceptions existed even in the days when the practice was becoming the forcefully enforced norm, and even in the most formal and pompous of Englishes, there is no good argument for insisting that Proper Nouns denoting things — _iPad_ and _iPhone_ and _eBay_, for example — should be forced to take an initial capital when they are the first word of a sentence.

And, by simple extension, there is no good argument for mangling the name of a code object (eg `overrideDefaults`) just because it is the first word of a sentence.

You can make a case for re-writing a sentence to avoid having the argument arise at all. In this particular case, for example:

```
The <code>overrideDefaults</code> replace any of the defaults set by a previous call to the <code>overrideDefaults</code> function.
```
But that’s a style question and, speaking for myself, I don’t have a strong preference for either approach in this case.

MitchCTiny: I changed the sentence to `The <code>overrideDefaults</code> method replaces any of the defaults set by a previous call to the <code>overrideDefaults</code> function.`

ephox-mogran: Fascinating as always Brian. Yeah, I was probably only suggesting adding a word before it, not changing its capitalisation (as the function name is case-sensitive anyway). Your example of iPad and iPhone are interesting. I think I also go out of my way to put a word before them, usually. It's good to know that I belong to the elitist group of enforcing capitalisation in others based on an ignorance of context, history, and justification :P

Now, of course, it's all different in German again. I'm sure that's also a fascinating story as well. But try not to bite ...

brian-forte (approved):

modules/tinymce/src/core/main/ts/api/EditorManager.ts
```diff
@@ -126,7 +126,15 @@ const EditorManager: EditorManager = {
 
   baseURI: null as any,
   baseURL: null as any,
+
+  /**
+   * Default options for TinyMCE.
```
ephox-mogran, line 131: There's not really a lot of explanation here. Is this really sufficient? What do `defaultOptions` do, for example? What is their relationship with options being passed through to `init`? Do we need to document that?

brian-forte (approved):

modules/tinymce/src/core/main/ts/api/EditorManager.ts
```diff
@@ -126,7 +126,15 @@ const EditorManager: EditorManager = {
 
   baseURI: null as any,
   baseURL: null as any,
+
+  /**
+   * Object which contains required, and user-specific, default values for cloud-related options.
```
ephox-mogran, line 131: Unfortunately, that's not what `defaultOptions` is. That's what we use it for on the cloud, but that's not what it is. It is a set of options that will be passed in by default to every `init` call, but the `init` call's settings / options etc. can override them.

ephox-mogran (changes requested):
The `defaultOptions` explanation isn't quite correct.

modules/tinymce/src/core/main/ts/api/EditorManager.ts
MitchCTiny: Thanks for the explanation! I wasn't sure where to find this out, so this was an uneducated placeholder. I've updated the description now, if you could please approve it.

lnewson: That still doesn't really explain how it's merged or that it's even be merged at all with the options passed to init. Maybe add something like this?
```
Object containing the options that will be passed by default to the `init` method upon each initialization of an editor. These defaults will be shallow merged with other options passed to `init`.
```

MitchCTiny: Any better now?

ephox-mogran: It's probably clearer if you keep it as two sentences. The reason is that the "that will be" has a slightly more ambiguous wording in the one-sentence version. 

> Object containing the default options that will be shallow merged with the options object passed to the `init` method upon each initialization of an editor.

It can be interpreted as it will contain *only* those default options that will be shallow merged, but other default options might be stored elsewhere. The two sentence version makes it a bit clearer that this will contain *all* the default options, and that all the default options will be passed through and shallow merged.

> Object containing the options that will be passed by default to the init method upon each initialization of an editor. These defaults will be shallow merged with other options passed to init.

But it's probably pretty subjective, so you might want to get another opinion.

ephox-mogran (approved):

modules/tinymce/src/core/main/ts/api/EditorManager.ts
```diff
@@ -270,10 +278,23 @@ const EditorManager: EditorManager = {
   },
 
   /**
-   * Overrides the default options for editor instances.
+   * Overrides the default options for editor instances. The <code>overrideDefaults</code> method replaces any of the defaults set by a previous call to the <code>overrideDefaults</code> function.
+   * <br /><br />
+   * When using the Tiny Cloud, some of these defaults are required for the cloud-based editor to function.
+   * <br /><br />
+   * Therefore, when using <code>overrideDefaults</code> with the cloud-based editor, combine the previous defaults, <code>tinymce.defaultOptions</code>, with the new options that you want to integrate.
```
brian-forte, line 285: A Socratic question:

Is _combine the previous defaults_ an instruction or a consequence?

To explain, one conceptual rendering of the current structure is as follows:

> _[state] has [requirements]_
> 
> _Therefore, when [state] + [thing], [instruction or consequence?], plus [thing] to create [state + more]._

If the phrase in question is an instruction, I’d argue for editing thus:
```suggestion
   * Overrides the default options for editor instances. The <code>overrideDefaults</code> method replaces any of the defaults set by a previous call to the <code>overrideDefaults</code> function.
   * <br /><br />
   * When using the Tiny Cloud, some of these defaults are required for the cloud-based editor to function.
   * <br /><br />
   * Therefore, when using <code>overrideDefaults</code> with the cloud-based editor, you must combine the previous defaults, <code>tinymce.defaultOptions</code>, with the new options that you want to integrate.
```

If, OTOH, the phrase in question is an expectation, I’d argue for editing thus:
```suggestion
   * Overrides the default options for editor instances. The <code>overrideDefaults</code> method replaces any of the defaults set by a previous call to the <code>overrideDefaults</code> function.
   * <br /><br />
   * When using the Tiny Cloud, some of these defaults are required for the cloud-based editor to function.
   * <br /><br />
   * Therefore, when using <code>overrideDefaults</code> with the cloud-based editor, code>overrideDefaults</code> combines the previous defaults, <code>tinymce.defaultOptions</code>, with the new options that you want to integrate.
```

MitchCTiny: I'd say it's an instruction rather than a consequence, but I'll try to improve the wording without using a pronoun.

TheSpyder (approved):

PR #8224, opened 2022-10-27
TINY-9290: fixed bug where names with numbers was not working correctly

spocke:
Related Ticket: TINY-9290

Description of Changes:
* The node name matches was incorrect since they didn't match H1 so they are changed to look for things that are not uppercase instead.
* Changed all duplicates of the function in tests by exposing the function also makes it testable.
* Improved tests so that all text blocks elements are tested would be more difficult to test all block elements since UL etc require a child element to be valid.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/content/TransparentElements.ts
+++ b/modules/tinymce/src/core/main/ts/content/TransparentElements.ts
@@ -7,7 +7,9 @@ import * as NodeType from '../dom/NodeType';
 
 export const transparentBlockAttr = 'data-mce-block';
 
-const makeSelectorFromSchemaMap = (map: SchemaMap) => Arr.filter(Obj.keys(map), (key) => /^[a-z]+$/.test(key)).join(',');
+export const elementNames = (map: SchemaMap): string[] => Arr.filter(Obj.keys(map), (key) => !/[A-Z]/.test(key));
+
+const makeSelectorFromSchemaMap = (map: SchemaMap) => elementNames(map).join(',');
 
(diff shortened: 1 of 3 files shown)


modules/tinymce/src/core/main/ts/content/TransparentElements.ts
```diff
@@ -7,7 +7,9 @@ import * as NodeType from '../dom/NodeType';
 
 export const transparentBlockAttr = 'data-mce-block';
 
-const makeSelectorFromSchemaMap = (map: SchemaMap) => Arr.filter(Obj.keys(map), (key) => /^[a-z]+$/.test(key)).join(',');
+export const elementNames = (map: SchemaMap): string[] => Arr.filter(Obj.keys(map), (key) => !/[A-Z]/.test(key));
```
ltrouton, line 10: Could just be me but I initially found it a bit hard to reason about why this works instead of the old way. I understand there is a test to ensure this doesn't get changed and regress which is the main thing. I am just wondering if it could be worth having a comment that mentions what the function is doing and how.

ltrouton (approved):

modules/tinymce/src/core/main/ts/content/TransparentElements.ts
spocke: Added a comment hopefully it's makes it easier to understand.

lorenzo-pomili (approved):

PR #8234, opened 2022-11-01
TINY-9193: The caret is moved to the previous line after Enter key pressed

vpyshnenko:
Related Ticket: TINY-9193

Description of Changes:
* Placeholder text

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -37,6 +37,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The `editor.selection.getRng()` API was not returning a proper range on hidden editors in Firefox. #TINY-9259
 - The `editor.selection.getBookmark()` API was not returning a proper bookmark on hidden editors in Firefox. #TINY-9259
 - Dragging a noneditable element before or after another noneditable element now works correctly. #TINY-9253
+- The caret was moved to the previous line when the inline text pattern have been executed on Enter key pressing in Firefox. #TINY-9193
 
 ## 6.2.0 - 2022-09-08
 
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -42,7 +42,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Dragging a noneditable element before or after another noneditable element now works correctly. #TINY-9253
 - The restored selection after a redo or undo action was not scrolled into view. #TINY-9222
 - A newline could not be inserted when the selection was restored from a bookmark after an inline `contenteditable="false"` element. #TINY-9194
-- The caret was moved to the previous line when the inline text pattern have been executed on Enter key pressing in Firefox. #TINY-9193
+- The caret was moved to the previous line when a text pattern executed a `mceInsertContent` command on Enter key when running on Firefox. #TINY-9193
 
 ## 6.2.0 - 2022-09-08
 
--- a/modules/tinymce/src/core/test/ts/browser/textpatterns/TextPatternsTest.ts
+++ b/modules/tinymce/src/core/test/ts/browser/textpatterns/TextPatternsTest.ts
@@ -303,4 +303,21 @@ describe('browser.tinymce.core.textpatterns.TextPatternsTest', () => {
 
     editor.options.unset('text_patterns');
   });
+
+  it('TINY-9193: should move the caret to the next line after the Enter key pressed', () => {
+    const editor = hook.editor();
+    editor.options.set('text_patterns', [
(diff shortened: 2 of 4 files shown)


vpyshnenko:
Something wrong on CI:
```
Unexpected Bedrock Error -> Server Quitting **********

[2022-11-03T15:27:38.752Z] Unexpected runner error: Error

[2022-11-03T15:27:38.752Z] 

[2022-11-03T15:27:38.752Z] Stack:

[2022-11-03T15:27:38.752Z] promiseReactionJob@[native code]

[2022-11-03T15:27:45.495Z] Warning: Task "bedrock-auto:browser" failed. Use --force to continue.
```

modules/tinymce/src/core/test/ts/browser/textpatterns/TextPatternsTest.ts
```diff
@@ -303,4 +304,24 @@ describe('browser.tinymce.core.textpatterns.TextPatternsTest', () => {
...
     editor.options.unset('text_patterns');
   });
+
+  it('TINY-9193: should move the caret to the next line after the Enter key pressed', () => {
+    const editor = hook.editor();
+    editor.options.set('text_patterns', [
+      { start: '~', end: '~', cmd: 'mceInsertContent', value: 'world' }
+    ]);
+    editor.setContent('<p>Hello</p><p>~test~</p>');
+    TinySelections.setCursor(editor, [ 1, 0 ], 6 );
```
lorenzo-pomili, line 314:
```suggestion
    editor.setContent('<p>Hello<br /></p><p>~test</p>');
    TinySelections.setCursor(editor, [ 1, 0 ], 5 );
    TinyContentActions.type(editor, '~');
```
I tried to run the test with the old code but it seems to not make the test fail.
With this close seems to fail without the fix and work with it but I'm not sure that is it the correct way to do it

lorenzo-pomili (changes requested):

modules/tinymce/src/core/test/ts/browser/textpatterns/TextPatternsTest.ts
vpyshnenko: Did you try the test with the old code in Firefox/MacOS?

TheSpyder (dismissed):
Sorry, you can't just remove the "delete" line from the "delete selected content" function. Before even getting to why it's there, now the function only does what it says some of the time.

There's a very good reason why we delete selected content before inserting and I would want to see far more evidence and discussion about why this is the best solution before I can consider accepting such a big change.

modules/tinymce/src/core/test/ts/browser/textpatterns/TextPatternsTest.ts
lorenzo-pomili: actually in Firefox/Linux, but it seems to work as in Mac, now that you made me think, it is possible that in Linux it works differently

vpyshnenko:
I've traced the issue down to the point of the deletion of the matched content through the native `Delete` command. Unfortunately in Firefox, this command implementation is accompanied by moving the caret to the previous block. While searching for a workaround I found that down the road the `insertHtmlAtCaret `passes through the `editor.selection.SetContent()` which internally ends up to `rng.deleteContents()`. Surprisingly removing the line with the `editor.getDoc().execCommand('Delete', false) `call passes all the tests. That makes me assume that it is a redundant call.  It that's not the case, probably we need to figure out a test case where we strictly depend on the `Delete `command for the deletion of ranged selection.

TheSpyder:
> I found that down the road the `insertHtmlAtCaret `passes through the `editor.selection.SetContent()` which internally ends up to `rng.deleteContents()`.

Ah that's not as bad as I thought. Thanks for narrowing the condition, I'll look at it soon.

modules/tinymce/CHANGELOG.md
```diff
@@ -42,6 +42,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Dragging a noneditable element before or after another noneditable element now works correctly. #TINY-9253
 - The restored selection after a redo or undo action was not scrolled into view. #TINY-9222
 - A newline could not be inserted when the selection was restored from a bookmark after an inline `contenteditable="false"` element. #TINY-9194
+- The caret was moved to the previous line when the inline text pattern have been executed on Enter key pressing in Firefox. #TINY-9193
```
spocke, line 45: Not sure if this is better:

```suggestion
- The caret was moved to the previous line when a text pattern executed a `mceInsertContent` command on Enter key when running on Firefox. #TINY-9193
```

spocke (approved):

lorenzo-pomili (approved):

PR #8235, opened 2022-11-02
TINY-9217: Toolbar menu button as inputs

lorenzo-pomili:
Related Ticket: TINY-9217

Description of Changes:
This spike includes two new components:
  [BespokeTypeahead](modules/tinymce/src/themes/silver/main/ts/ui/core/complex/BespokeTypeahead.ts) -> this could be used as a replacement for any toolbar menu select
  [BespokeNumberInput](modules/tinymce/src/themes/silver/main/ts/ui/core/complex/BespokeNumberInput.ts) -> this can be used as a number input with a unit, units also give the possibility to define specific behaviors for different units, by now is defined only in the step but it could also be used for others behavior like minimum or maximum value

Pre-checks:
* [ ] Changelog entry added

Review:
* [ ] Milestone set

### DIFF
--- a/modules/alloy/src/main/ts/ephox/alloy/ui/schema/TypeaheadSchema.ts
+++ b/modules/alloy/src/main/ts/ephox/alloy/ui/schema/TypeaheadSchema.ts
@@ -55,7 +55,7 @@ const schema = Fun.constant([
     Focusing, Representing, Streaming, Keying, Toggling, Coupling
   ]),
 
-  FieldSchema.customField('lazyTypeaheadComp', () => Cell(Optional.none)),
+  FieldSchema.customField('lazyTypeaheadComp', () => Cell(Optional.none())),
 
   FieldSchema.customField('previewing', () => Cell(true))
 ].concat(
--- a/modules/tinymce/src/themes/silver/main/ts/ui/core/complex/BespokeNumberInput.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/ui/core/complex/BespokeNumberInput.ts
@@ -27,15 +27,18 @@ const createBespokeNumberInput = (editor: Editor, backstage: UiFactoryBackstage,
 
   const customEvents = Id.generate('custom-number-input-events');
 
+  const isValidValue = (value: number): boolean => value >= 1;
+
   const changeValue = (f: (v: number) => number): void => {
     const text = currentValue.get();
     const unit = text.match(/\D+$/)?.join('');
     const value = parseInt(text.match(/^\d+/)?.join('') || '0', 10);
-    const newValue = `${f(value)}${unit}`;
+    const newValue = f(value);
+    const newValueWithUnit = `${isValidValue(newValue) ? newValue : value}${unit}`;
 
-    spec.onAction(newValue);
-    currentValue.set(newValue);
-    currentComp.get().each((comp) => Representing.setValue(comp, newValue));
+    spec.onAction(newValueWithUnit);
+    currentValue.set(newValueWithUnit);
+    currentComp.get().each((comp) => Representing.setValue(comp, newValueWithUnit));
   };
 
   const buttonStyles = {
--- a/modules/tinymce/src/themes/silver/main/ts/ui/core/complex/BespokeNumberInput.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/ui/core/complex/BespokeNumberInput.ts
@@ -1,4 +1,4 @@
-import { AddEventsBehaviour, AlloyComponent, AlloyEvents, Behaviour, Button, Input, Representing, SketchSpec } from '@ephox/alloy';
+import { AddEventsBehaviour, AlloyComponent, AlloyEvents, Behaviour, Button, Input, NativeEvents, Representing, SketchSpec } from '@ephox/alloy';
 import { Cell, Fun, Id, Optional } from '@ephox/katamari';
 
 import Editor from 'tinymce/core/api/Editor';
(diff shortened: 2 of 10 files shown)


modules/alloy/src/main/ts/ephox/alloy/ui/schema/TypeaheadSchema.ts
```diff
@@ -55,7 +55,7 @@ const schema = Fun.constant([
     Focusing, Representing, Streaming, Keying, Toggling, Coupling
   ]),
 
-  FieldSchema.customField('lazyTypeaheadComp', () => Cell(Optional.none)),
+  FieldSchema.customField('lazyTypeaheadComp', () => Cell(Optional.none())),
```
ephox-mogran, line 58: NOTE to other reviewers: This is just a bug. It only surfaces when you clobber the EventBehaviour that does onAttached in Typeahead, so it only happens when the code is broken, but it does need to be fixed.

modules/tinymce/src/themes/silver/main/ts/ui/core/complex/BespokeNumberInput.ts
```diff
@@ -0,0 +1,126 @@
...
+import { onSetupEvent } from '../ControlUtils';
+import { NumberInputSpec } from './FontSizeBespoke';
+
+interface BespokeSelectApi {
+  readonly getComponent: () => AlloyComponent;
+}
+
+const createBespokeNumberInput = (editor: Editor, backstage: UiFactoryBackstage, spec: NumberInputSpec): SketchSpec => {
+  const currentValue = Cell('');
+  const currentComp: Cell<Optional<AlloyComponent>> = Cell(Optional.none());
```
ephox-mogran, line 19: Generally, if you find yourself creating a cell with a reference to the current component, then you're probably not doing something in the cleanest way. In this case, it looks like you need it for the `changeValue` function, and it's currently storing the input. This is what you would use Mementos for.

ephox-mogran: Although I guess you don't really care about that level of detail, and you just want an assessment of the overall approach, correct? I'll try to focus more on the big-picture :)

modules/tinymce/src/themes/silver/main/ts/ui/core/complex/BespokeNumberInput.ts
```diff
@@ -0,0 +1,126 @@
...
+            return Optional.some(true);
+          } else {
+            return Optional.none();
+          }
+        },
+        onEscape: (wrapperComp) => {
+          if (Focus.hasFocus(wrapperComp.element)) {
+            return Optional.none();
+          } else {
+            Focusing.focus(wrapperComp);
```
ephox-mogran, line 117: Be wary of switching between Focus an Focusing because they are slightly different.

modules/tinymce/src/themes/silver/main/ts/ui/core/complex/BespokeNumberInput.ts
```diff
@@ -0,0 +1,126 @@
+import { AddEventsBehaviour, AlloyComponent, AlloyEvents, Behaviour, Button, Focusing, Input, Keying, NativeEvents, Representing, SketchSpec } from '@ephox/alloy';
+import { Cell, Fun, Id, Optional } from '@ephox/katamari';
+import { Focus, SugarElement, Traverse } from '@ephox/sugar';
+
```
ephox-mogran, line 4: So as a high-level overview of what you are doing here

* spec.updateText is used to update the input field to the appropriate value
* the DOM structure is a wrapper containing [minus button, input, plus button]
* pressing Escape on the wrapper isn't handled, and should be handled by the calling code
* pressing Escape within the wrapper should focus the wrapper
* pressing Enter on the wrapper should focus the 1st child inside the wrapper
* pressing Enter within the wrapper should do nothing, and isn't handled
* you use cells to keep values so that you can refer to them across components
* when plus or minus is triggered, you identify the unit via regex, and try to add / subtract and appropriate step size for that unit based on a config
* if the value is larger than 0, you keep it ... otherwise you change to the parsed value before adding/subtracting
* on clicking +/-, you execute the action to make the appropriate formatting change in the editor
* on typing in the input, you execute the action to make the appropriate formatting change in the editor. 
* on typing in the input, you do not validate the data

Does that sound right?

modules/tinymce/src/themes/silver/main/ts/ui/core/complex/BespokeSelect.ts
```diff
@@ -68,6 +68,13 @@ export interface SelectSpec {
   readonly dataset: SelectDataset;
 }
 
+export interface SelectTypeaheadSpec {
+  readonly dataset: SelectDataset;
+  readonly onAction: (item: FormatterFormatItem) => (api: Menu.ToggleMenuItemInstanceApi) => void;
```
ephox-mogran, line 73: I'll have a closer look at the actual code, but this `onAction` looks a bit strange.

modules/tinymce/src/themes/silver/main/ts/ui/core/complex/BespokeTypeahead.ts
```diff
@@ -0,0 +1,181 @@
...
+
+        const selectedStyle = Arr.find(styles, (style) => {
+          if (style.type === 'formatter') {
+            return style.title === value.meta.text;
+          }
+          if (style.type === 'separator') {
+            return false;
+          }
+          if (style.type === 'submenu') {
+            return Arr.exists(style.items, (subItem) => subItem.title === value.meta.text);
```
ephox-mogran, line 118: Are these styles going to be nested? Will you have to look inside subItem.items for example?

modules/tinymce/src/themes/silver/main/ts/ui/core/complex/BespokeNumberInput.ts
```diff
@@ -0,0 +1,126 @@
...
+  const editorOffCell = Cell(Fun.noop);
+
+  const customEvents = Id.generate('custom-number-input-events');
+
+  const isValidValue = (value: number): boolean => value >= 0;
+
+  const changeValue = (f: (v: number, step: number) => number): void => {
+    const text = currentValue.get();
+    const unit = text.match(/\D+$/)?.join('') || '';
+    const value = parseFloat(text.match(/^[\d\.]+/)?.join('') || '0');
```
ephox-mogran, line 36: So we do some basic normalising here, in that we restore to `value` not `text`. But we don't do any normalising for `change`. Is that intended?

modules/tinymce/src/themes/silver/main/ts/ui/core/complex/BespokeNumberInput.ts
```diff
@@ -0,0 +1,126 @@
...
+}
+
+const createBespokeNumberInput = (editor: Editor, backstage: UiFactoryBackstage, spec: NumberInputSpec): SketchSpec => {
+  const currentValue = Cell('');
+  const currentComp: Cell<Optional<AlloyComponent>> = Cell(Optional.none());
+
+  const onSetup = onSetupEvent(editor, 'NodeChange', (api: BespokeSelectApi) => {
+    const comp = api.getComponent();
+    currentComp.set(Optional.some(comp));
+    spec.updateText(comp);
```
ephox-mogran, line 24: I realise it's what we've called it elsewhere, but using `updateText` for a function that doesn't take the value to update it to keeps causing me problems as I read this. Is there a name that we can use (assuming it's not public-facing) that suggests that the function itself works out the value, rather than being provided the value as an argument?

PR #8238, opened 2022-11-03
TINY-9271: Add 'ToggleToolbarDrawer' event

yacodes:
Related Ticket: TINY-9271

Description of Changes:
* Added new `ToggleToolbarDrawer` event to subsribe to toolbar's opening and closing actions.
* In order to keep alloy components without any knowledge of the editor instance, added a new `onToggled` callback and propagated in up to the `OuterContainer` component for `SplitSlidingToolbar` and `SplitFloatingToolbar`.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/alloy/CHANGELOG.md
+++ b/modules/alloy/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
 
+### Added
+- Added `ToggleToolbarDrawer` event to subscribe to toolbar's opening and closing. #TINY-9271
+
 ### Removed
 - Moved `TestStore` to Agar. #TINY-9157
 
--- a/modules/alloy/CHANGELOG.md
+++ b/modules/alloy/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## Unreleased
 
 ### Added
-- Added `onToggled` event for `SplitFloatingToolbar` and `SplitSlidingToolbar` components. #TINY-9271
+- Added the `onToggled` callback for `SplitFloatingToolbar` and `SplitSlidingToolbar` components. #TINY-9271
 
 ### Removed
 - Moved `TestStore` to Agar. #TINY-9157
--- a/modules/alloy/CHANGELOG.md
+++ b/modules/alloy/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## Unreleased
 
 ### Added
-- Added `ToggleToolbarDrawer` event to subscribe to toolbar's opening and closing. #TINY-9271
+- Added `onToggled` event for `SplitFloatingToolbar` and `SplitSlidingToolbar` components. #TINY-9271
 
 ### Removed
 - Moved `TestStore` to Agar. #TINY-9157
--- a/modules/alloy/CHANGELOG.md
+++ b/modules/alloy/CHANGELOG.md
@@ -7,7 +7,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## Unreleased
 
 ### Added
-- Added the `onToggled` callback for `SplitFloatingToolbar` and `SplitSlidingToolbar` components. #TINY-9271
+- Added the `onToggled` callback for the `FloatingToolbarButton` component. #TINY-9271
+- Added the `onOpened` and `onClosed` callbacks for the `SplitFloatingToolbar` component. #TINY-9271
 
 ### Removed
 - Moved `TestStore` to Agar. #TINY-9157
--- a/modules/alloy/src/main/ts/ephox/alloy/ui/schema/FloatingToolbarButtonSchema.ts
+++ b/modules/alloy/src/main/ts/ephox/alloy/ui/schema/FloatingToolbarButtonSchema.ts
@@ -48,7 +48,7 @@ const parts: () => PartType.PartTypeAdt[] = Fun.constant([
            * For FloatingToolbars, we can hook up our `onToggled` handler directly to the Toggling
            * because we don't have to worry about any animations.
            *
-           * Unfortunately, for SlidingToolbars, Toggling is more directly hooked into the animation for growing, 
+           * Unfortunately, for SlidingToolbars, Toggling is more directly hooked into the animation for growing,
            * so to have an event `onToggled` that doesn't care about the animation, we can't just hook into the Toggling config.
            */
           onToggled: (_api, state) => detail.onToggled(state)
--- a/modules/alloy/src/main/ts/ephox/alloy/ui/schema/FloatingToolbarButtonSchema.ts
+++ b/modules/alloy/src/main/ts/ephox/alloy/ui/schema/FloatingToolbarButtonSchema.ts
@@ -51,7 +51,7 @@ const parts: () => PartType.PartTypeAdt[] = Fun.constant([
            * Unfortunately, for SlidingToolbars, Toggling is more directly hooked into the animation for growing,
            * so to have an event `onToggled` that doesn't care about the animation, we can't just hook into the Toggling config.
            */
-          onToggled: (_api, state) => detail.onToggled(state)
+          onToggled: detail.onToggled
         })
       ])
     })
--- a/modules/alloy/src/main/ts/ephox/alloy/ui/schema/FloatingToolbarButtonSchema.ts
+++ b/modules/alloy/src/main/ts/ephox/alloy/ui/schema/FloatingToolbarButtonSchema.ts
@@ -24,7 +24,8 @@ const schema = Fun.constant([
   FieldSchema.optionObjOf('fireDismissalEventInstead', [
     FieldSchema.defaulted('event', SystemEvents.dismissRequested())
   ]),
-  AnchorLayouts.schema()
+  AnchorLayouts.schema(),
+  Fields.onHandler('onToggled'),
 ]);
 
 const parts: () => PartType.PartTypeAdt[] = Fun.constant([
@@ -42,7 +43,8 @@ const parts: () => PartType.PartTypeAdt[] = Fun.constant([
           aria: {
             mode: 'expanded'
           },
-          toggleOnExecute: false
+          toggleOnExecute: false,
+          onToggled: (_api, state) => detail.onToggled(state)
         })
       ])
     })
--- a/modules/alloy/src/main/ts/ephox/alloy/ui/schema/FloatingToolbarButtonSchema.ts
+++ b/modules/alloy/src/main/ts/ephox/alloy/ui/schema/FloatingToolbarButtonSchema.ts
@@ -44,6 +44,13 @@ const parts: () => PartType.PartTypeAdt[] = Fun.constant([
             mode: 'expanded'
           },
           toggleOnExecute: false,
+          /**
+           * For FloatingToolbars, we can hook up our `onToggled` handler directly to the Toggling
+           * because we don't have to worry about any animations.
+           *
+           * Unfortunately, for SlidingToolbars, Toggling is more directly hooked into the animation for growing, 
+           * so to have an event `onToggled` that doesn't care about the animation, we can't just hook into the Toggling config.
+           */
           onToggled: (_api, state) => detail.onToggled(state)
         })
       ])
--- a/modules/alloy/src/main/ts/ephox/alloy/ui/schema/SplitSlidingToolbarSchema.ts
+++ b/modules/alloy/src/main/ts/ephox/alloy/ui/schema/SplitSlidingToolbarSchema.ts
@@ -66,7 +66,7 @@ const parts: () => PartType.PartTypeAdt[] = Fun.constant([
             },
             onStartShrink: () => {
               /**
-               * Calling the `onToggled` callback here instead of `Toggling` behaviour,
+               * Calling the `onToggled` callback here instead of `Toggling` behaviour's configuration,
                * because we want the instant state change.
                * Otherwise the callback is called only after the animation is completed.
                */
--- a/modules/alloy/src/main/ts/ephox/alloy/ui/schema/SplitSlidingToolbarSchema.ts
+++ b/modules/alloy/src/main/ts/ephox/alloy/ui/schema/SplitSlidingToolbarSchema.ts
@@ -18,7 +18,8 @@ import * as ToolbarSchema from './ToolbarSchema';
 const schema = Fun.constant([
   Fields.markers([ 'closedClass', 'openClass', 'shrinkingClass', 'growingClass', 'overflowToggledClass' ]),
   Fields.onHandler('onOpened'),
-  Fields.onHandler('onClosed')
+  Fields.onHandler('onClosed'),
+  Fields.onHandler('onToggled'),
 ].concat(
   SplitToolbarBase.schema()
 ));
@@ -58,7 +59,11 @@ const parts: () => PartType.PartTypeAdt[] = Fun.constant([
             },
             onStartGrow: (comp) => {
               AlloyParts.getPart(comp, detail, 'overflow-button').each(Toggling.on);
-            }
+              detail.onToggled(true);
+            },
+            onStartShrink: () => {
+              detail.onToggled(false);
+            },
           }),
           Keying.config({
             mode: 'acyclic',
--- a/modules/alloy/src/main/ts/ephox/alloy/ui/schema/SplitSlidingToolbarSchema.ts
+++ b/modules/alloy/src/main/ts/ephox/alloy/ui/schema/SplitSlidingToolbarSchema.ts
@@ -59,9 +59,17 @@ const parts: () => PartType.PartTypeAdt[] = Fun.constant([
             },
             onStartGrow: (comp) => {
               AlloyParts.getPart(comp, detail, 'overflow-button').each(Toggling.on);
+              /**
+               * Calling the `onToggled` for the consistency with `onStartShrink` callback.
+               */
               detail.onToggled(true);
             },
             onStartShrink: () => {
+              /**
+               * Calling the `onToggled` callback here instead of `Toggling` behaviour,
+               * because I want the instant state change.
+               * Otherwise the callback is called only after the animation is completed.
+               */
               detail.onToggled(false);
             },
           }),
--- a/modules/alloy/src/main/ts/ephox/alloy/ui/schema/SplitSlidingToolbarSchema.ts
+++ b/modules/alloy/src/main/ts/ephox/alloy/ui/schema/SplitSlidingToolbarSchema.ts
@@ -59,19 +59,7 @@ const parts: () => PartType.PartTypeAdt[] = Fun.constant([
             },
             onStartGrow: (comp) => {
               AlloyParts.getPart(comp, detail, 'overflow-button').each(Toggling.on);
-              /**
-               * Calling the `onToggled` for the consistency with the `onStartShrink` callback.
-               */
-              detail.onToggled(true);
-            },
-            onStartShrink: () => {
-              /**
-               * Calling the `onToggled` callback here instead of `Toggling` behaviour's configuration,
-               * because we want the instant state change.
-               * Otherwise the callback is called only after the animation is completed.
-               */
-              detail.onToggled(false);
-            },
+            }
           }),
           Keying.config({
             mode: 'acyclic',
@@ -95,13 +83,6 @@ const parts: () => PartType.PartTypeAdt[] = Fun.constant([
             mode: 'pressed'
           },
           toggleOnExecute: false
-          /**
-           * For FloatingToolbars, we can hook up our `onToggled` handler directly to the Toggling
-           * because we don't have to worry about any animations.
-           *
-           * Unfortunately, for SlidingToolbars, Toggling is more directly hooked into the animation for growing, 
-           * so to have an event `onToggled` that doesn't care about the animation, we can't just hook into the Toggling config.
-           */
         })
       ])
     })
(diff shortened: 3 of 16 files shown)


modules/tinymce/src/themes/silver/main/ts/Render.ts
```diff
@@ -423,7 +426,6 @@ const setup = (editor: Editor): RenderInfo => {
 
     editor.addCommand('ToggleToolbarDrawer', () => {
       OuterContainer.toggleToolbarDrawer(outerContainer);
-      // TODO: Consider firing event - TINY-6371
```
TheSpyder, line 426: When removing old TODOs please close the old ticket and link it to the current ticket, don't leave things dangling

TheSpyder (approved):

TheSpyder:
Also please remember to add the rest of the team when opening a PR - our code owners are set up such that only the required reviewers are added by default.

modules/tinymce/src/themes/silver/main/ts/Render.ts
yacodes: Closed the `TINY-6371` ticket as a duplicate.

modules/alloy/CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
 
+### Added
```
lnewson, line 9: This doesn't look like an `alloy` change? I suspect this was meant for the TinyMCE changelog as we're missing a changelog there. We'd then still need one here for the `onToggled` callback added.

- [ ] Make sure an appropriate changelog exists for `alloy` and `tinymce`

modules/tinymce/src/themes/silver/main/ts/Render.ts
```diff
@@ -165,6 +165,9 @@ const setup = (editor: Editor): RenderInfo => {
       onEscape: () => {
         editor.focus();
       },
+      onToolbarToggled: (state: boolean) => {
+        editor.dispatch('ToggleToolbarDrawer', { isToolbarDrawerToggled: state });
```
lnewson, line 169: This should be in the `api/Events.ts` file instead per the TinyMCE codestyle guidelines. The idea is to keep all the events fired by a plugin, theme, etc... to know what is used/exposed. See https://github.com/tinymce/tinymce/blob/ea5387c05c4a47a0aecb7289836a9c885ca6adf6/modules/tinymce/src/themes/silver/main/ts/api/Events.ts

Note: This is normally a lint rule but atm is disabled in TinyMCE due to some legacy things needing to be fixed.

- [ ] Move dispatch call to `api/Events.ts`

modules/tinymce/src/themes/silver/test/ts/browser/editor/toolbar/ToolbarDrawerToggleTest.ts
```diff
@@ -68,4 +69,43 @@ describe('browser.tinymce.themes.silver.editor.toolbar.ToolbarDrawerToggleTest',
...
+    assert.isFalse(lastState.get().getOrDie('Should be set'));
+    McEditor.remove(editor);
+  };
+
+  context(`Should emit 'ToogleToolbarDrawer' event when toggled`, () => {
+    Arr.each<ToolbarMode>([ 'floating', 'sliding' ], (toolbarMode) => {
+      it(`TINY-9271: Emits 'ToggleToolbarDrawer' in ${toolbarMode} via execCommand`, async () => {
+        await pTestEvent(toolbarMode, (editor) => editor.execCommand('ToggleToolbarDrawer'));
+      });
+      it(`TINY-9271: Emits 'ToggleToolbarDrawer' in ${toolbarMode} via user click`, async () => {
```
lnewson, line 106:
```suggestion

      it(`TINY-9271: Emits 'ToggleToolbarDrawer' in ${toolbarMode} via user click`, async () => {
```
Just some "fluff" that's normally included for readability 🤷

modules/tinymce/src/themes/silver/test/ts/browser/editor/toolbar/ToolbarDrawerToggleTest.ts
```diff
@@ -68,4 +69,43 @@ describe('browser.tinymce.themes.silver.editor.toolbar.ToolbarDrawerToggleTest',
...
+    lastState.set(false);
+    await UiUtils.pWaitForEditorToRender();
+
+    editor.on('ToggleToolbarDrawer', (options: { isToolbarDrawerToggled: boolean }) => {
+      lastState.set(options.isToolbarDrawerToggled);
+    });
+
+    assert.isFalse(lastState.get().getOrDie('Should be set'));
+    command(editor);
+    await Waiter.pWait(10);
```
ephox-mogran, line 93: What are we waiting for here? Can it be more precise? Can we wait for a predicate?

modules/tinymce/src/themes/silver/test/ts/browser/editor/toolbar/ToolbarDrawerToggleTest.ts
```diff
@@ -68,4 +69,43 @@ describe('browser.tinymce.themes.silver.editor.toolbar.ToolbarDrawerToggleTest',
...
+      menubar: false,
+      statusbar: false,
+      width: 200,
+      toolbar_mode: toolbarMode,
+      base_url: '/project/tinymce/js/tinymce'
+    });
+    editor.focus();
+
+    const lastState = Singleton.value<boolean>();
+    lastState.set(false);
```
ephox-mogran, line 84: By using a single value here, we aren't checking how many times the event is firing ... just what the last value was. Is that a problem? Should we consider using something closer to an array like TestStore?

modules/tinymce/src/themes/silver/main/ts/ui/toolbar/CommonToolbar.ts
```diff
@@ -199,7 +201,8 @@ const renderSlidingMoreToolbar = (toolbarSpec: MoreDrawerToolbarSpec): SketchSpe
     onClosed: (comp) => {
       // TINY-9223: This will only broadcast to the same mothership as the toolbar
       comp.getSystem().broadcastOn([ Channels.toolbarHeightChange() ], { type: 'closed' });
-    }
+    },
+    onToggled: toolbarSpec.onToggled,
```
ephox-mogran, line 205: It's pretty minor, but I'm not sure "toggling" a toolbar would mean expanding / shrinking it to me. It seems like it would be more about showing/hiding. Can we make this configuration name more descriptive? I don't have any better names by the way :)

modules/alloy/src/main/ts/ephox/alloy/ui/schema/SplitSlidingToolbarSchema.ts
```diff
@@ -58,7 +59,11 @@ const parts: () => PartType.PartTypeAdt[] = Fun.constant([
             },
             onStartGrow: (comp) => {
               AlloyParts.getPart(comp, detail, 'overflow-button').each(Toggling.on);
-            }
+              detail.onToggled(true);
```
ephox-mogran, line 62: It's a bit inconsistent that we directly call `onToggled` here for Sliding, but for Floating, we rely on the Toggling behaviour's handler being defined. The Toggling behaviour for the button for sliding is just below (about 20-30 lines below this). Could we just put `onToggled` into that config like we do for Floating? Is there a reason for the different approach between the two?

ephox-mogran: Because now we've kind of combined approaches. Due to the animation aspects of sliding we have got this: 

* onShrunk calls `Toggling.off` but doesn't fire handler
* onStartGrow calls `Toggling.on` *and* does fire handler
* onStartShrink fires handler but doesn't call `Toggling.off` yet. 

So the handler isn't mapped to Toggling, which I guess is why you didn't use the Toggling config approach.

If this extra complexity is all created by the need to make sure the handler fires as soon as the close is issued, rather than after the animation is finished, maybe we should revisit that requirement. By decoupling Toggling and the handler like this, we're arguably making the interactions even harder to understand.

ephox-mogran: - [ ] if we go with this approach, let's put thorough comments in the code as to why. Otherwise, the next developer who looks at this code is probably just going to change it to make it seem more consistent.

And I think I'm unconvinced that we want to introduce so much complexity just so the event doesn't wait for the sliding animation to complete before firing ... but I might be in the minority on that.

ephox-mogran: Or if we stick to this, then maybe we should decouple the other one as well. Not sure.

modules/tinymce/src/themes/silver/test/ts/browser/editor/toolbar/ToolbarDrawerToggleTest.ts
yacodes: Updated [here](https://github.com/tinymce/tinymce/pull/8238/commits/bf58856c776bbe02e4ba4b99159eea982121e23b)

modules/tinymce/src/themes/silver/test/ts/browser/editor/toolbar/ToolbarDrawerToggleTest.ts
yacodes: For some reason the state was not toggled instantly, not sure why. Updated [here](https://github.com/tinymce/tinymce/pull/8238/commits/6a399604a6153909c68b5b2f28a8396b4aa37990) with waiting for the floating toolbar to appear. Hoping this is a bit more clear.

modules/tinymce/src/themes/silver/test/ts/browser/editor/toolbar/ToolbarDrawerToggleTest.ts
yacodes: Yeah, probably counting the amount of triggered events is a good idea. Updated with `TestStore` [here](https://github.com/tinymce/tinymce/pull/8238/commits/a900ff7e91cd86ca15912a4a076d09a1d12af975).

modules/tinymce/src/themes/silver/main/ts/ui/toolbar/CommonToolbar.ts
yacodes: Don't know about that one. In case of the sliding toolbar maybe it makes sense, but for the floating it's exactly like toggling, no? 

We can reword it as `onMoreButtonClick`. However it's verbose and do not really underline the intentions. I'd prefer to leave `onToggled` here.

modules/alloy/src/main/ts/ephox/alloy/ui/schema/SplitSlidingToolbarSchema.ts
yacodes: @ephox-mogran Added comments to the `SplitSlidingToolbarSchema` [here](https://github.com/tinymce/tinymce/pull/8238/commits/d16d3768492e9013664463f92ee70b1b68e024d4).

I would prefer to stick to the current implementation, because the behaviour is much more clear for the end-user of the `onToggled` callback. In the current case it calls immediately after the button click.

Otherwise, if we will add `onToggled` to the `Toggling` behaviour, the `ToggleToolbarEvent` will be called in the beginning of the toolbar's opening, but in the end of the toolbar's closing. That's extremely confusing. 

Would discuss it with the team, but, again, personally I would prefer the current implementation.

yacodes: Maybe it is also a good idea to add `OpenToggleToolbarDrawer` and `CloseToggleToolbarDrawer` events, which correspond to the `Toggling` behaviour of the `SplitSlidingToolbarSchema`.

modules/alloy/CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
+- Added `onToggled` event for `SplitFloatingToolbar` and `SplitSlidingToolbar` components. #TINY-9271
```
lnewson, line 10:
```suggestion
- Added the `onToggled` callback for `SplitFloatingToolbar` and `SplitSlidingToolbar` components. #TINY-9271
```
Just correcting the terminology here is no event here that's fired, it's a callback.

modules/tinymce/src/themes/silver/main/ts/Render.ts
```diff
@@ -165,6 +166,9 @@ const setup = (editor: Editor): RenderInfo => {
       onEscape: () => {
         editor.focus();
       },
+      onToolbarToggled: (state: boolean) => {
+        fireToggleToolbarDrawer(editor, state);
```
lnewson, line 170: This is almost there, but still doesn't follow the TinyMCE style guide sorry. You should be using `import * as Events` and `Events.fireToggleToolbarDrawer` here so that it improves readability and makes it clear it's not a local function.

modules/tinymce/src/themes/silver/main/ts/api/Events.ts
```diff
@@ -35,6 +35,10 @@ const fireResolveName = (editor: Editor, node: Node): EditorEvent<{ name: string
     target: node
   });
 
+const fireToggleToolbarDrawer = (editor: Editor, state: boolean): void => {
+  editor.dispatch('ToggleToolbarDrawer', { isToolbarDrawerToggled: state });
```
lnewson, line 39: I wonder if this should just be called `state`? As it'd make it consistent with other events, for example: https://github.com/tinymce/tinymce/blob/a5cddb59ab810756e436940708cc94ba075c95c6/modules/tinymce/src/plugins/fullscreen/main/ts/api/Events.ts#L4
Additionally we already know it's about the toolbar drawer so it's doubling up. If not `state` then maybe it should just be `toggled` or `isToggled` instead?

modules/tinymce/src/themes/silver/main/ts/ui/toolbar/CommonToolbar.ts
ephox-mogran: Yeah, as I said, I don't have any better ideas.

modules/alloy/src/main/ts/ephox/alloy/ui/schema/SplitSlidingToolbarSchema.ts
ephox-mogran: Yeah, it's messy either way. I wonder if the problem here was more that we initially made those calls animation dependent. And we probably did that because our actual use case needed it to be animation dependent. I'm happy to defer to whatever @TheSpyder , @ltrouton , and @spocke think, as the official TinyMCE reviewers. As long as we have an explanation for why we've done it this way in the code and the commit message, I don't mind so much.

modules/alloy/src/main/ts/ephox/alloy/ui/schema/FloatingToolbarButtonSchema.ts
```diff
@@ -42,7 +43,8 @@ const parts: () => PartType.PartTypeAdt[] = Fun.constant([
           aria: {
             mode: 'expanded'
           },
-          toggleOnExecute: false
+          toggleOnExecute: false,
+          onToggled: (_api, state) => detail.onToggled(state)
```
ephox-mogran, line 47: In terms of adding a comment, I probably meant something like 

Here, we would put

```
// For FloatingToolbars, we can hook up our `onToggled` handler directly to the Toggling because we don't have to worry about any animations. Unfortunately, for SlidingToolbars, Toggling is more directly hooked into the animation for growing, so to have an event `onToggled` that doesn't care about the animation, we can't just hook into the Toggling config.
```

And then in the SlidingToolbar button Toggling config, either copy the same comment, or just say "See comment in ___" (which might get out-of-date)

modules/alloy/src/main/ts/ephox/alloy/ui/schema/SplitSlidingToolbarSchema.ts
```diff
@@ -58,7 +59,19 @@ const parts: () => PartType.PartTypeAdt[] = Fun.constant([
...
               AlloyParts.getPart(comp, detail, 'overflow-button').each(Toggling.on);
-            }
+              /**
+               * Calling the `onToggled` for the consistency with `onStartShrink` callback.
+               */
+              detail.onToggled(true);
+            },
+            onStartShrink: () => {
+              /**
+               * Calling the `onToggled` callback here instead of `Toggling` behaviour,
```
ephox-mogran, line 69:
```suggestion
               * Calling the `onToggled` callback here instead of `Toggling` behaviour's configuration
```

ephox-mogran (commented):
It would be ideal to add some tests cases in alloy, as well.

modules/alloy/src/main/ts/ephox/alloy/ui/schema/SplitSlidingToolbarSchema.ts
TheSpyder: I would've expected the event to not fire until the animation is complete but I can see valid reasons either way.

I'm fine with just explaining it as Morgan suggested.

modules/alloy/CHANGELOG.md
yacodes: Updated the changelog [here](https://github.com/tinymce/tinymce/pull/8238/commits/2d3044613f2f55c8065be436df1a284376972b91)

modules/tinymce/src/themes/silver/main/ts/api/Events.ts
yacodes: Updated to `isToolbarDrawerToggled` to match the already existing API [here](https://github.com/tinymce/tinymce/pull/8238/commits/691d7d20e19563bf30884ca99cba3aed9f66c90c).

modules/alloy/src/main/ts/ephox/alloy/ui/schema/FloatingToolbarButtonSchema.ts
yacodes: Added the comments [here](https://github.com/tinymce/tinymce/pull/8238/commits/76b67657d40baa31fd7443df16fb080bb2edd9ff).

modules/tinymce/src/themes/silver/main/ts/Render.ts
yacodes: @lnewson The issue is that there is already an `Events` variable in this scope, which setups document and other events. L456

modules/alloy/src/main/ts/ephox/alloy/ui/schema/SplitSlidingToolbarSchema.ts
yacodes: @ephox-mogran @TheSpyder I've discussed when to call the callback with the team and we've arrived to a conclusion to call `onToggled` **after** the animation is completed. Updated the code accordingly [here](https://github.com/tinymce/tinymce/pull/8238/commits/8d8ddca29583d055293eda91a0b852e8a1d9c00c). What do you think?

modules/tinymce/src/themes/silver/main/ts/api/Events.ts
lnewson: @yacodes you appear to have gone the wrong direction here? You're link isn't working, so I'm not sure what you were linking to, but as noted above we shouldn't be using `ToolbarDrawer` in the property name as that's already the event name. Also it would be more consistent to use `state` here instead as we already do elsewhere.

modules/tinymce/src/themes/silver/main/ts/Render.ts
yacodes: Updated [here](https://github.com/tinymce/tinymce/pull/8238/commits/41f53bd6542c55a41658022a49469695e27efdb5).

yacodes:
@ephox-mogran Added a test for `FloatingToolbarButton` [here](https://github.com/tinymce/tinymce/pull/8238/commits/4af2e1106e2ef3967f852e556b23969cc44a9df1). Not sure if there is more to add.

modules/alloy/src/main/ts/ephox/alloy/ui/schema/SplitSlidingToolbarSchema.ts
ephox-mogran: So I've looked at the recent changes, and I have a few more comments. It's probably easier to discuss these via a call, but I don't know if that will suit you time-wise. Anyway..

* When declaring handlers in alloy, try not to remove the component from the arguments. It's so hard to get built components in alloy. Your current code may not need it, but someone else's code might. Always get into the habit of all handlers having component as their first argument
* The SlidingToolbar now takes a configuration setting `onToggled` that it doesn't appear to use. I don't think it's necessary any more.
* However, instead of making an `onToggled` setting in FloatingToolbar, it might be an idea to mimic what Sliding has and have a onOpened and onClosed for consistency.
* The test you've added doesn't test the state being passed through the handler, and also only tests the button. You'd want something that checks that the FloatingToolbar is correctly configuring its button also. I'm not sure if there's an existing test.

It looks good otherwise.

yacodes: @ephox-mogran 
1. Updated all callbacks with the `comp` variable containing `AlloyComponent`;
2. Removed `onToggled` handler from the `SlidingToolbar`;
3. Renamed `onToggled` handler to `onOpened` and `onClosed` in `FloatingToolbar`;
4. Updated `FloatingButtonTest` to check the state variable;
5. Updated `FloatingToolbarTest` with `onOpened` and `onClosed` callbacks.

modules/alloy/CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
 
+### Added
+- Added the `onToggled` callback for `SplitFloatingToolbar` and `SplitSlidingToolbar` components. #TINY-9271
```
ephox-mogran, line 10: I don't think is accurate any more. It seems more like the FloatingToolbar added onOpened and onClosed, and the FloatingToolbarButton added an onToggled

ephox-mogran (approved):
This looks pretty good. The changelog for alloy looks incorrect, though, so just check exactly what you added. Other than that, approved!

modules/alloy/CHANGELOG.md
yacodes: Updated the changelog [here](https://github.com/tinymce/tinymce/pull/8238/commits/521c870d2e05af4fae59b2dd1a3c66cd80715894)

PR #8246, opened 2022-11-08
TINY-8827: Update ContextToolbar to close Quicklink toolbar on a single ESC keypress

yacodes:
Related Ticket: TINY-8827

Description of Changes:
* Updated `ContextToolbar.ts` to close Quicklink toolbar on a single ESC keypress. It is not very elegant, but I couldn't figure out a better solution. Let me know if there is one.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/plugins/link/test/ts/browser/QuickLinkTest.ts
+++ b/modules/tinymce/src/plugins/link/test/ts/browser/QuickLinkTest.ts
@@ -182,4 +182,15 @@ describe('browser.tinymce.plugins.link.QuickLinkTest', () => {
     TinyUiActions.keydown(editor, Keys.enter());
     UiFinder.notExists(SugarBody.body(), '.tox-pop__dialog');
   });
+
+  it('TINY-8827: Closes Quicklink toolbar on escape keyup', async () => {
+    const editor = hook.editor();
+    editor.setContent('');
+    TinyContentActions.keystroke(editor, 'K'.charCodeAt(0), metaKey);
+    await TinyUiActions.pWaitForPopup(editor, '.tox-pop__dialog .tox-toolbar');
+    await FocusTools.pTryOnSelector('Selector should be in context form input', doc, '.tox-toolbar input');
+    await Waiter.pWait(100);
+    TinyUiActions.keyup(editor, Keys.escape());
+    UiFinder.notExists(SugarBody.body(), '.tox-pop__dialog');
+  });
 });
--- a/modules/tinymce/src/themes/silver/main/ts/ui/context/ContextToolbar.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/ui/context/ContextToolbar.ts
@@ -47,11 +47,15 @@ const register = (editor: Editor, registryContextToolbars: Record<string, Contex
   const lastElement = Singleton.value<SugarElement<Element>>();
   const lastTrigger = Singleton.value<TriggerCause>();
   const lastContextPosition = Singleton.value<InlineContent.ContextPosition>();
+  const lastToolbarKey = Singleton.value<string>();
 
   const contextbar = GuiFactory.build(
     renderContextToolbar({
       sink,
       onEscape: () => {
+        if (lastToolbarKey.get().getOr('default') === 'quicklink') {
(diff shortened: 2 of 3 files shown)


modules/tinymce/src/plugins/link/test/ts/browser/QuickLinkTest.ts
```diff
@@ -182,4 +182,15 @@ describe('browser.tinymce.plugins.link.QuickLinkTest', () => {
...
     UiFinder.notExists(SugarBody.body(), '.tox-pop__dialog');
   });
+
+  it('TINY-8827: Closes Quicklink toolbar on escape keyup', async () => {
+    const editor = hook.editor();
+    editor.setContent('');
+    TinyContentActions.keystroke(editor, 'K'.charCodeAt(0), metaKey);
+    await TinyUiActions.pWaitForPopup(editor, '.tox-pop__dialog .tox-toolbar');
+    await FocusTools.pTryOnSelector('Selector should be in context form input', doc, '.tox-toolbar input');
+    await Waiter.pWait(100);
```
ephox-mogran, line 192: Generally avoid random waits. We should always know *why* we are waiting, otherwise we become even more subject to race conditions.

modules/tinymce/src/themes/silver/main/ts/ui/context/ContextToolbar.ts
```diff
@@ -47,11 +47,15 @@ const register = (editor: Editor, registryContextToolbars: Record<string, Contex
   const lastElement = Singleton.value<SugarElement<Element>>();
   const lastTrigger = Singleton.value<TriggerCause>();
   const lastContextPosition = Singleton.value<InlineContent.ContextPosition>();
+  const lastToolbarKey = Singleton.value<string>();
 
   const contextbar = GuiFactory.build(
     renderContextToolbar({
       sink,
       onEscape: () => {
+        if (lastToolbarKey.get().getOr('default') === 'quicklink') {
```
ephox-mogran, line 56: It's kind of strange to default a value just to fail a predicate check. I would have expected that we'd have something on Optional that takes a value, and only returns true if it is some, and the some is that value, however, I can't find it.

Normally, this function is called `is(v: V)` and returns false for None, and true for Some if the value contained is v. But maybe it only exists in languages with an easier way of stating `Eq` instances for equality. 

I guess because we clearly don't have this function, it's probably *marginally* better to use `.exists((v) => v === 'quicklink')`, though I'm not sure if I can justify that.

modules/tinymce/src/themes/silver/main/ts/ui/context/ContextToolbar.ts
```diff
@@ -257,6 +262,7 @@ const register = (editor: Editor, registryContextToolbars: Record<string, Contex
       const scopes = getScopes();
       // TODO: Have this stored in a better structure
       Obj.get(scopes.lookupTable, e.toolbarKey).each((ctx) => {
+        lastToolbarKey.set(e.toolbarKey);
```
ephox-mogran, line 265: This does look like the wrong approach, but I haven't looked into it. It does seem to be consistent with an approach we are using throughout this, however, because there are *lots* of clear statements in the above code. Sometimes, when that is the case, it can be better to group together some of them into one object ...but that all depends on whether they are set / cleared at the same time.

So, I'll have a look and see if I can find a cleaner way of achieving what you want.

ephox-mogran (changes requested):
I've had a look at this. There are a lot of things going on here. I'll put them in slack.

modules/tinymce/src/themes/silver/main/ts/ui/context/ContextToolbar.ts
TheSpyder: That was moved to `Optionals`
https://github.com/tinymce/tinymce/blob/ba6994f39b09ecd2c28cef68c3e8a85a4815e235/modules/katamari/src/main/ts/ephox/katamari/api/Optionals.ts#L8-L9

yacodes:
Closing because this issue is way bigger than was originally thought.

PR #8257, opened 2022-11-11
TINY-6037: Add `content_css_cors` setting to global `tinymce.dom.styleSheetLoader`

MitchCTiny:
Related Ticket: TINY-6037

Description of Changes:
* Setting the `content_css_cors` option will now update the global `tinymce.dom.styleSheetLoader` with the given setting
* This allows the `crossOrigin="anonymous"` attribute to be added to stylesheets loaded in inline mode
* This was not possible before as inline editors share the global `styleSheetLoader`, rather than separate ones for each editor, which never has this setting updated.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -42,7 +42,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Dragging a noneditable element before or after another noneditable element now works correctly. #TINY-9253
 - The restored selection after a redo or undo action was not scrolled into view. #TINY-9222
 - A newline could not be inserted when the selection was restored from a bookmark after an inline `contenteditable="false"` element. #TINY-9194
-- The global `tinymce.dom.styleSheetLoader` was not affected by the `content_css_cors` setting, preventing `crossOrigin="anonymous"` from being applied to loaded stylesheets when in inline mode. #TINY-6037
+- The global `tinymce.dom.styleSheetLoader` was not affected by the `content_css_cors` option. #TINY-6037
 
 ## 6.2.0 - 2022-09-08
 
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -42,6 +42,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Dragging a noneditable element before or after another noneditable element now works correctly. #TINY-9253
 - The restored selection after a redo or undo action was not scrolled into view. #TINY-9222
 - A newline could not be inserted when the selection was restored from a bookmark after an inline `contenteditable="false"` element. #TINY-9194
+- The global `tinymce.dom.styleSheetLoader` was not affected by the `content_css_cors` setting, preventing `crossOrigin="anonymous"` from being applied to loaded stylesheets when in inline mode. #TINY-6037
 
 ## 6.2.0 - 2022-09-08
 
--- a/modules/tinymce/src/core/main/ts/api/Editor.ts
+++ b/modules/tinymce/src/core/main/ts/api/Editor.ts
@@ -291,6 +291,11 @@ class Editor implements EditorObservable {
       DOMUtils.DOM.styleSheetLoader._setReferrerPolicy(referrerPolicy);
     }
 
+    const contentCssCors = Options.hasContentCssCors(self);
+    if (contentCssCors) {
+      DOMUtils.DOM.styleSheetLoader._setContentCssCors(contentCssCors);
+    }
+
     AddOnManager.languageLoad = getOption('language_load');
     AddOnManager.baseURL = editorManager.baseURL;
 
--- a/modules/tinymce/src/core/main/ts/api/Editor.ts
+++ b/modules/tinymce/src/core/main/ts/api/Editor.ts
@@ -292,7 +292,7 @@ class Editor implements EditorObservable {
     }
 
     const contentCssCors = Options.hasContentCssCors(self);
-    if (contentCssCors !== undefined) {
+    if (Type.isNonNullable(contentCssCors)) {
       DOMUtils.DOM.styleSheetLoader._setContentCssCors(contentCssCors);
     }
 
--- a/modules/tinymce/src/core/main/ts/api/Editor.ts
+++ b/modules/tinymce/src/core/main/ts/api/Editor.ts
@@ -292,7 +292,7 @@ class Editor implements EditorObservable {
     }
 
     const contentCssCors = Options.hasContentCssCors(self);
-    if (contentCssCors) {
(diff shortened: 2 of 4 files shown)


modules/tinymce/src/core/main/ts/api/Editor.ts
```diff
@@ -291,6 +291,11 @@ class Editor implements EditorObservable {
       DOMUtils.DOM.styleSheetLoader._setReferrerPolicy(referrerPolicy);
     }
 
+    const contentCssCors = Options.hasContentCssCors(self);
+    if (contentCssCors !== undefined) {
```
MitchCTiny, line 295: Should this condition also check that the mode is inline? Or would this be fine to include in classic mode as well?

modules/tinymce/CHANGELOG.md
```diff
@@ -42,6 +42,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Dragging a noneditable element before or after another noneditable element now works correctly. #TINY-9253
 - The restored selection after a redo or undo action was not scrolled into view. #TINY-9222
 - A newline could not be inserted when the selection was restored from a bookmark after an inline `contenteditable="false"` element. #TINY-9194
+- The global `tinymce.dom.styleSheetLoader` was not affected by the `content_css_cors` setting, preventing `crossOrigin="anonymous"` from being applied to loaded stylesheets when in inline mode. #TINY-6037
```
spocke, line 45: Maybe just keep it short the reader can look up what the content_css_cors option is doing. Also it's called options now.

```suggestion
- The global `tinymce.dom.styleSheetLoader` was not affected by the `content_css_cors` option. #TINY-6037
```

spocke (approved):

modules/tinymce/src/core/main/ts/api/Editor.ts
ltrouton: I think it is fine to not check the editor mode as we don't have any specific checks for `referrerPolicy` above either. Could we maybe do `Type.isNonNullable` instead?

modules/tinymce/src/core/test/ts/browser/dom/ContentCssCorsTest.ts
```diff
@@ -11,20 +12,118 @@ describe('browser.tinymce.core.dom.ContentCssCorsTest', () => {
...
     const corsLinks = editor.getDoc().querySelectorAll('link[crossorigin="anonymous"]');
-    assert.equal(corsLinks.length > 0, expected, 'should have link with crossorigin="anonymous"');
+    assert.equal(corsLinks.length, expected, `should have ${expected} link(s) with crossorigin="anonymous"`);
   };
 
-  it('assert crossorigin link presence with setting set', async () => {
-    const editor = await McEditor.pFromSettings<Editor>({ ...settings, content_css_cors: true });
-    assertCorsLinkPresence(editor, true);
-    McEditor.remove(editor);
+  afterEach(() => {
```
ltrouton, line 27: Does this have to be `afterEach` or can it be `after`?

ltrouton (approved):
Nice work on the tests.

modules/tinymce/src/core/test/ts/browser/dom/ContentCssCorsTest.ts
MitchCTiny: `afterEach` since the global `styleSheetLoader` is affected each time `content_css_cors` is set in an inline editor. I wasn't able to find another way to reload `tinymce` in each test, so this was the next best thing.

modules/tinymce/src/core/main/ts/api/Editor.ts
MitchCTiny: Good idea 👍

modules/tinymce/src/core/main/ts/api/Editor.ts
```diff
@@ -291,6 +291,11 @@ class Editor implements EditorObservable {
       DOMUtils.DOM.styleSheetLoader._setReferrerPolicy(referrerPolicy);
     }
 
+    const contentCssCors = Options.hasContentCssCors(self);
+    if (Type.isNonNullable(contentCssCors)) {
+      DOMUtils.DOM.styleSheetLoader._setContentCssCors(contentCssCors);
```
ephox-mogran, line 296: So I guess the only reason we are comfortable doing this is we think there is no valid reason for someone to have two different css_cors settings on the same page? I can think of a bizarre reason someone might, but is that a limitation we are OK with? Because with this change, interleaving editor loads could have the wrong behaviour if one of them is supposed to have css_cors support, and the other isn't (if I'm understanding the concept properly).

I imagine it's fine ... but it's just something we have to be aware of. We clearly already do this for referrer policy, but you *could* make a case that makes even less sense to not share across all editors. Once again, though, you could find a bizarre edge case where you don't want it to be shared.

ephox-mogran: The other approach was making the editor instantiate its own stylesheetLoader with the editor's own settings/options, but I assume we decided against that?

MitchCTiny: The `content_css_cors` option will now be updated whenever an editor is loaded, and affect any style sheets loaded after it is set, until the option is set to anything different when loading another editor. Inline mode editors share this `styleSheetLoader`, and the document and any stylesheets loaded on it. Now, the stylesheets needed for any editor will be loaded with the correct cors setting, unless it is an inline editor with a different cors setting to a previously initialised inline editor, in which case the stylesheets used by all inline editors will already be loaded.

MitchCTiny: Classic editors have their own `styleSheetLoaders` for stylesheets loaded into the iframe document, however stylesheets loaded onto the page document were not affected by the cors setting.

ephox-mogran: I understand that part. I worry that there is a situation where the stylesheet loading can interleave, and due to the sequencing of promises, the global might be changed before the next stream of stylesheets loads on the first editor. But it's probably unfounded. And as I said before, an edge case. It would depend on the exact ordering of stylesheet loading, and waiting in the editor, to whether there are race conditions here.

Having said that, even if there are race conditions, they would only matter if the editors had different settings, which isn't likely to be the case. I just wanted to know if we were accepting this as a limitation, or we are just confident that there should be no reason due to waits / promises / when stylesheets load that the setting could change before all the stylesheets have loaded for a particular editor.

And my other question was more about did we consider making inline *not* share the global loader, or was that just going to be too big a change?

ephox-mogran: Yeah, that makes some sense, given that the stylesheet will be inherited by each editor regardless of whether it's specified as part of the editor configuration.

PR #8258, opened 2022-11-11
TINY-9232: Parser needs to handle nested anchor cases better

spocke:
Related Ticket: TINY-9232

Description of Changes:
* Splits out invalid elements that gets contained in nested structures in transparent elements like `p -> a -> p` for example

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/content/TransparentElements.ts
+++ b/modules/tinymce/src/core/main/ts/content/TransparentElements.ts
@@ -37,9 +37,9 @@ const updateBlockStateOnChildren = (schema: Schema, scope: Element): Element[] =
 };
 
 const trimEdge = (el: DocumentFragment, leftSide: boolean) => {
-  const childName = leftSide ? 'lastChild' : 'firstChild';
+  const childPropertyName = leftSide ? 'lastChild' : 'firstChild';
 
-  for (let child = el[childName]; child; child = child[childName]) {
+  for (let child = el[childPropertyName]; child; child = child[childPropertyName]) {
     if (Empty.isEmpty(SugarElement.fromDom(child))) {
       child.parentNode?.removeChild(child);
       return;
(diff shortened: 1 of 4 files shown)


vpyshnenko (approved):

modules/tinymce/src/core/test/ts/browser/html/DomParserTest.ts
```diff
@@ -1448,6 +1448,43 @@ describe('browser.tinymce.core.html.DomParserTest', () => {
 
         assert.equal(serializedHtml, expectedHtml);
       });
+
+      const testSplitInvalidBlocksOut = (testCase: { input: string; expected: string }) => {
```
HAFRMO, line 1452: Should this function be moved to the top of the context-block or top of the description? Additionally, should it be named 'assert' instead of 'test'?

modules/tinymce/src/core/main/ts/content/TransparentElements.ts
```diff
@@ -20,16 +21,97 @@ const updateTransparent = (blocksSelector: string, transparent: Element) => {
...
+const updateBlockStateOnChildren = (schema: Schema, scope: Element): Element[] => {
   const transparentSelector = makeSelectorFromSchemaMap(schema.getTransparentElements());
   const blocksSelector = makeSelectorFromSchemaMap(schema.getBlockElements());
 
-  Arr.each(scope.querySelectorAll(transparentSelector), (transparent) => updateTransparent(blocksSelector, transparent));
+  return Arr.filter(scope.querySelectorAll(transparentSelector), (transparent) => updateTransparent(blocksSelector, transparent));
+};
+
+const trimEdge = (el: DocumentFragment, leftSide: boolean) => {
+  const childName = leftSide ? 'lastChild' : 'firstChild';
```
HAFRMO, line 40: Could we pick a different name for this variable? `childname` makes it seem like we're looking for specific, named nodes instead of if we're going to the left or right?

HAFRMO (approved):

modules/tinymce/src/core/main/ts/content/TransparentElements.ts
spocke: Good point, maybe `childPropertyName`.

TheSpyder (approved):

PR #8267, opened 2022-11-15
TINY-6853: Lists applied incorrectly in table cells

hamza0867:
Related Ticket: TINY-6853

Description of Changes:
* Toggling list inside a table cell with that has an anchor element in it, now includes the anchor element in the produced list

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Color picker dialog now starts on the appropriate color for the cursor position. #TINY-9213
 
 ### Fixed
+- Clicking on list toolbar button inside a table cell that has anchor element when the cursor is outside the anchor element would not include the anchor element in the produce list. #TINY-6853
 - Parsing media content would cause a memory leak, which for example occurred when using the `getContent` API. #TINY-9186
 - Dragging a noneditable element toward the bottom edge would cause the page to scroll up. #TINY-9025
 - Range expanding capabilities would behave inconsistently depending on where the cursor was placed. #TINY-9029
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -25,7 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Color picker dialog now starts on the appropriate color for the cursor position. #TINY-9213
 
 ### Fixed
-- Clicking on list toolbar button inside a table cell that has anchor element when the cursor is outside the anchor element would not include the anchor element in the produce list. #TINY-6853
+- Clicking on list toolbar button inside a table cell that has anchor element when the cursor is outside the anchor element would not include the anchor element in the produced list. #TINY-6853
 - Parsing media content would cause a memory leak, which for example occurred when using the `getContent` API. #TINY-9186
 - Dragging a noneditable element toward the bottom edge would cause the page to scroll up. #TINY-9025
 - Range expanding capabilities would behave inconsistently depending on where the cursor was placed. #TINY-9029
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -25,7 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Color picker dialog now starts on the appropriate color for the cursor position. #TINY-9213
 
 ### Fixed
-- Clicking on list toolbar button inside a table cell that has anchor element when the cursor is outside the anchor element would not include the anchor element in the produced list. #TINY-6853
+- Creating a list in a table cell when the caret is in front of an anchor element would not properly include the anchor in the list. #TINY-6853
 - Parsing media content would cause a memory leak, which for example occurred when using the `getContent` API. #TINY-9186
 - Dragging a noneditable element toward the bottom edge would cause the page to scroll up. #TINY-9025
 - Range expanding capabilities would behave inconsistently depending on where the cursor was placed. #TINY-9029
--- a/modules/tinymce/src/plugins/lists/main/ts/actions/ToggleList.ts
+++ b/modules/tinymce/src/plugins/lists/main/ts/actions/ToggleList.ts
@@ -1,5 +1,4 @@
-import { Type, Unicode, Obj, Fun, Arr } from '@ephox/katamari';
-import { SugarElement, SugarNode } from '@ephox/sugar';
+import { Type, Unicode } from '@ephox/katamari';
 
 import BookmarkManager from 'tinymce/core/api/dom/BookmarkManager';
 import DOMUtils from 'tinymce/core/api/dom/DOMUtils';
@@ -47,25 +46,7 @@ const removeStyles = (dom: DOMUtils, element: HTMLElement, styles: string[]): vo
   Tools.each(styles, (style) => dom.setStyle(element, style, ''));
 };
 
-const blocks = [
-  'article', 'aside', 'details', 'div', 'dt', 'figcaption', 'footer',
-  'form', 'fieldset', 'header', 'hgroup', 'html', 'main', 'nav',
-  'section', 'summary', 'body', 'p', 'dl', 'multicol', 'dd', 'figure',
-  'address', 'center', 'blockquote', 'h1', 'h2', 'h3', 'h4', 'h5', 'h6',
-  'listing', 'xmp', 'pre', 'plaintext', 'menu', 'dir', 'ul', 'ol', 'li', 'hr',
-  'table', 'tbody', 'thead', 'tfoot', 'th', 'tr', 'td', 'caption'
-];
-
-const lazyLookup = <T extends Node = HTMLElement>(items: string[]) => {
-  let lookup: Record<string, boolean> | undefined;
-  return (node: SugarElement<Node>): node is SugarElement<T> => {
-    lookup = lookup ? lookup : Arr.mapToObject(items, Fun.always);
-    return Obj.has(lookup, SugarNode.name(node));
-  };
-};
-
-const isBlock = lazyLookup(blocks);
-const isInline = (node: SugarElement<Node>): node is SugarElement<HTMLElement> => SugarNode.isElement(node) && !isBlock(node);
+const isInline = (editor: Editor, node: Node): boolean => Type.isNonNullable(node) && !NodeType.isBlock(node, editor.schema.getBlockElements());
 
 const getEndPointNode = (editor: Editor, rng: Range, start: Boolean, root: Node): Node => {
   let container = rng[start ? 'startContainer' : 'endContainer'];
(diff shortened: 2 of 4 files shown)


modules/tinymce/src/plugins/lists/main/ts/actions/ToggleList.ts
```diff
@@ -59,6 +80,42 @@ const getEndPointNode = (editor: Editor, rng: Range, start: Boolean, root: Node)
...
+  if (start && NodeType.isTextNode(container)) {
+    if (Unicode.isZwsp(container.textContent as string)) {
+      if (container.previousSibling !== null && isInline(SugarElement.fromDom(container.previousSibling))) {
+        container = container.previousSibling.firstChild || container.previousSibling;
+      } else if (container.nextSibling !== null && isInline(SugarElement.fromDom(container.nextSibling))) {
+        container = container.nextSibling.firstChild || container.nextSibling;
+      }
+    }
+    if (!Unicode.isZwsp(container.textContent as string)) {
+      if ( container.parentNode !== null && isInline(SugarElement.fromDom(container.parentNode))) {
```
yacodes, line 92: It's an unnecessary space right after the round bracket, I believe. (Why eslint hasn't figured this out?)

yacodes (changes requested):

modules/tinymce/src/plugins/lists/main/ts/actions/ToggleList.ts
```diff
@@ -59,6 +80,42 @@ const getEndPointNode = (editor: Editor, rng: Range, start: Boolean, root: Node)
     container = container.nextSibling;
   }
 
+  if (start && NodeType.isTextNode(container)) {
```
ephox-mogran, line 83: So this is a lot of code targeting what looks like a relatively specific case. Often, this is what code often starts as, but ideally, it's better if we can make it more general. So here are some questions I'd ask myself as I was going through the process of getting this code ready for merging:

i) "why" are we doing this? What exactly is it trying to achieve. Could an explanatory comment about "why" help here? 
ii) is this handling the general case? Is this making assumptions about the DOM structure? If so, what are those assumptions based on? 
iii) if we are traversing left with previous sibling, I'd normally expect us to descend with `lastChild`, not `firstChild`. Is there a reason that we have picked `firstChild` for both previous and next traversals? 
iv) a lot of existing code in TinyMCE uses reassignment and lots of `if` statements, but ultimately, we want to get away from doing that. Could we rewrite this in a more intuitive way? Could we make it clearer how each movement of what 'container' is progresses? 
v) I think ultimately with a DOM traversal this complex, we'd want to test it in isolation. Could this be pulled out and tested separately?
vi) when using `textContent` we are essentially losing information about the DOM descendants and structure of the node. Is this a known limitation? Would this code still work if the original container had multiple empty elements, as well as a text node with a unicode zero width cursor.
vii) mixing `ifs` with `while` statements can make it even harder to follow what's going on. Can we pull these things out?
viii) when mixing Sugar and non-sugar APIs, things can be more confusing. Should they be mixed? And if so, is there a better boundary to mix them?

modules/tinymce/src/plugins/lists/main/ts/actions/ToggleList.ts
hamza0867: Good catch! fixed 👍

modules/tinymce/src/plugins/lists/main/ts/actions/ToggleList.ts
spocke: I think you can use the TextSeeker api here so find a suitable text node. It will traverse the dom and find text nodes within the current block scope from a point that is what you need. It also takes case of complex things like noneditables etc.
However there are more complex situations here say there is a image instead of a text node `<td><a href="#"></a></td>` there are internal apis that would find a better point here however they are not publicly exposed so the list plugin can't use them but maybe they should be though the editor.selection api so that you can do editor.selection.findNextCaretPosition etc those apis does only check for the next character in the logical order not the visual order so it might not always do what you expect in bidi text but in this case where it goes from a zwnbsp spot to the beginning of a text node they should work.

Other possible solutions could be to have the inline boundaries remove the zwnbsp point spot before the insert list  command is executed some how either by listening to the plugin command or having some public core api to do it. Having it listening for list commands makes the core dependent on the plugin in a way we usually try avoid however sometimes that is a better and easier solution also in this case the plugin is a mono repo plugin and possibly even a candidate for being moved into the core.

modules/tinymce/src/plugins/lists/main/ts/actions/ToggleList.ts
```diff
@@ -59,6 +61,53 @@ const getEndPointNode = (editor: Editor, rng: Range, start: Boolean, root: Node)
     container = container.nextSibling;
   }
 
+  // The reason why the next two if statements exist is because when the root node is a table cell (possibly some other node types)
```
ephox-mogran, line 64: Thanks for this comment. I'm starting to see what you're doing now, which helps with the review process.

modules/tinymce/src/plugins/lists/main/ts/actions/ToggleList.ts
```diff
@@ -59,6 +61,53 @@ const getEndPointNode = (editor: Editor, rng: Range, start: Boolean, root: Node)
...
+  // and when the selection is collapsed inside one of the inline nodes then only that inline node (or text node) will be included
+  // in the created list because that would be one level below td node and the other inline nodes won't be included.
+  // So the fix proposed is to traverse left when looking for start node (and traverse right when looking for end node)
+  // and keep traversing as long as we have an inline or text node (same for traversing right).
+  // This way we end up including all the inline elements in the created list.
+  // For more info look at #TINY-6853
+
+  // Traverse left to include inline/text nodes
+  if (start && NodeType.isTextNode(container)) {
+    if (Unicode.isZwsp(container.textContent as string)) {
```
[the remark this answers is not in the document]
ephox-mogran: I think I'm just a little unsure in general of what this use case is trying to cover. Zero width cursors can appear in a variety of structures ... or are we assuming that tinymce has prevented that at this point somehow? If this code has an assumption on the DOM structure based on some tinymce process, we might need to outline that here.

modules/tinymce/src/plugins/lists/main/ts/actions/ToggleList.ts
```diff
@@ -46,6 +47,26 @@ const removeStyles = (dom: DOMUtils, element: HTMLElement, styles: string[]): vo
...
 };
 
+const blocks = [
+  'article', 'aside', 'details', 'div', 'dt', 'figcaption', 'footer',
+  'form', 'fieldset', 'header', 'hgroup', 'html', 'main', 'nav',
+  'section', 'summary', 'body', 'p', 'dl', 'multicol', 'dd', 'figure',
+  'address', 'center', 'blockquote', 'h1', 'h2', 'h3', 'h4', 'h5', 'h6',
+  'listing', 'xmp', 'pre', 'plaintext', 'menu', 'dir', 'ul', 'ol', 'li', 'hr',
+  'table', 'tbody', 'thead', 'tfoot', 'th', 'tr', 'td', 'caption'
+];
```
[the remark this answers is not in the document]
hamza0867: Since I no longer use this array, should this conversation be resolved ?

modules/tinymce/src/plugins/lists/main/ts/actions/ToggleList.ts
```diff
@@ -46,6 +47,26 @@ const removeStyles = (dom: DOMUtils, element: HTMLElement, styles: string[]): vo
...
+
+const lazyLookup = <T extends Node = HTMLElement>(items: string[]) => {
+  let lookup: Record<string, boolean> | undefined;
```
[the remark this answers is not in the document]
hamza0867: Since I no longer use this function, should this conversation be resolved ?

modules/tinymce/src/plugins/lists/main/ts/actions/ToggleList.ts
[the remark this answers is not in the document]
hamza0867: Yes there is an assumption which is: you can't have multiple ZWSP nodes next to each other ( You can have one text node with multiple ZWSP characters but not multiple nodes each having ZWSP ). That is why I don't loop and use an if instead. Is this assumption correct ?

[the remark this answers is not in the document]
ephox-mogran: I'm not sure. I think @spocke would know. My knowledge of zero width spaces mostly comes from textbox.io, and we put them in all the time to make things simpler. I think tinymce is a lot more controlled than that.

modules/tinymce/src/plugins/lists/test/ts/browser/ToggleListInTableCell.ts
```diff
@@ -0,0 +1,79 @@
...
+  const wrapInsideTableWithList = (content: string): string =>
+    `<table style="border-collapse: collapse; width: 100%;" border="1"><colgroup><col style="width: 50%;"><col style="width: 50%;"></colgroup><tbody><tr><td><ul><li>${content}</li></ul></td></tr></tbody></table>`;
+
+  it('TINY-6853: toggle list inside table cell with anchor outside of anchor to the left of the anchor should include the anchor in the list', () => {
+    const editor = hook.editor();
+    const content = '<a href="#">link</a>';
+    editor.setContent(wrapInsideTable(content));
+    TinySelections.setCursor(editor, [ 0, 1, 0, 0, 0 ], 0);
+    // Needed to move the cursor outside of the link
+    TinyContentActions.keydown(editor, Keys.left());
```
spocke, line 28: This can use keystroke I think it fires both key events.

modules/tinymce/src/plugins/lists/test/ts/browser/ToggleListInTableCell.ts
```diff
@@ -0,0 +1,79 @@
...
+    TinyUiActions.clickOnToolbar(editor, 'button[aria-label="Bullet list"]');
+    TinyAssertions.assertContent(editor, wrapInsideTableWithList(content));
+  });
+  it('TINY-6853: toggle list inside table cell with cursor between two anchors should include both anchors in the list', () => {
+    const editor = hook.editor();
+    const content = '<a href="#">a</a><a href="#">b</a>';
+    editor.setContent(wrapInsideTable(content));
+    TinySelections.setCursor(editor, [ 0, 1, 0, 0, 1 ], 0);
+    // Needed to move the cursor outside of the link
+    TinyContentActions.keydown(editor, Keys.left());
```
spocke, line 74: Same here keystroke

modules/tinymce/src/plugins/lists/test/ts/browser/ToggleListInTableCell.ts
```diff
@@ -0,0 +1,79 @@
...
+    const content = '<a href="#">link</a>';
+    editor.setContent(wrapInsideTable(content));
+    TinySelections.setCursor(editor, [ 0, 1, 0, 0, 0 ], 0);
+    // Needed to move the cursor outside of the link
+    TinyContentActions.keydown(editor, Keys.left());
+    TinyContentActions.keyup(editor, Keys.left());
+    TinyUiActions.clickOnToolbar(editor, 'button[aria-label="Bullet list"]');
+    TinyAssertions.assertContent(editor, wrapInsideTableWithList(content));
+  });
+  it('TINY-6853: toggle list inside table cell with anchor outside of anchor to the right of the anchor should include the anchor in the list', () => {
```
spocke, line 33: Maybe some newlines between the it calls just to fluff it up a bit. To me at least that is easier to read but feel free to ignore if you don't agree.

spocke (changes requested):

modules/tinymce/src/plugins/lists/test/ts/browser/ToggleListInTableCell.ts
hamza0867: Done!

modules/tinymce/src/plugins/lists/test/ts/browser/ToggleListInTableCell.ts
hamza0867: Done!

modules/tinymce/src/plugins/lists/main/ts/actions/ToggleList.ts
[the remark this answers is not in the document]
spocke: Nice, Should be able to handle some edge cases like super nested things. Maybe worth adding a test for such a case like for example. `<a href="#"><strong><em></em></strong></a>`

modules/tinymce/CHANGELOG.md
```diff
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Color picker dialog now starts on the appropriate color for the cursor position. #TINY-9213
 
 ### Fixed
+- Clicking on list toolbar button inside a table cell that has anchor element when the cursor is outside the anchor element would not include the anchor element in the produced list. #TINY-6853
```
spocke, line 28: Not sure if it's any better writing changelog items is hard. Just noticed that the one you have is unusually long.

```suggestion
- Creating a list in a table cell when the caret is in front of an anchor element would not properly include the anchor in the list. #TINY-6853
```

spocke (approved):

modules/tinymce/src/plugins/lists/main/ts/actions/ToggleList.ts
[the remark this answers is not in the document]
hamza0867: Done!

HAFRMO (approved):

PR #8272, opened 2022-11-16
TINY-9337: Remove focusing after toggling the toolbar

yacodes:
Related Ticket: TINY-9337

Description of Changes:
* Removed focusing after toggling the toolbar in `FloatingToolbarButton`.

The focusing was added in the initial `FloatingToolbarButton` commit in 2019. I am not sure what are the consequences of removing it, but it definitely conflicts with `skip_focus` argument in `execCommend`.

Other solutions would be to refactor the `execCommand` in order to pass down `skip_focus` argument to all commands, but that feels like a major public API change.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

[TINY-9337]: https://ephocks.atlassian.net/browse/TINY-9337?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/alloy/src/main/ts/ephox/alloy/api/ui/FloatingToolbarButton.ts
+++ b/modules/alloy/src/main/ts/ephox/alloy/api/ui/FloatingToolbarButton.ts
@@ -26,12 +26,11 @@ import * as Sketcher from './Sketcher';
 import { Toolbar } from './Toolbar';
 import { CompositeSketchFactory } from './UiSketcher';
 
-
 const onClose = (
   button: AlloyComponent,
   toolbar: AlloyComponent,
   ariaControls: AriaControls.AriaManager,
-  options?: { skipFocus: boolean },
+  options?: { skipFocus: boolean }
 ) => {
   Sandboxing.close(toolbar);
   if (!options?.skipFocus) {
@@ -39,7 +38,7 @@ const onClose = (
     Focusing.focus(button);
   }
   ariaControls.unlink(button.element);
-}
+};
 
 const toggle = (
   button: AlloyComponent,
@@ -93,7 +92,7 @@ const makeSandbox = (
   button: AlloyComponent,
   spec: FloatingToolbarButtonSpec,
   detail: FloatingToolbarButtonDetail,
-  ariaControls: AriaControls.AriaManager,
+  ariaControls: AriaControls.AriaManager
 ) => ({
   dom: {
     tag: 'div',
@@ -141,46 +140,47 @@ const factory: CompositeSketchFactory<FloatingToolbarButtonDetail, FloatingToolb
   const ariaControls = AriaControls.manager();
 
   return {
-  ...Button.sketch({
-    ...externals.button(),
-    action: (button) => {
-      toggle(button, externals, detail, spec, ariaControls);
-    },
-    buttonBehaviours: SketchBehaviours.augment(
-      { dump: externals.button().buttonBehaviours },
-      [
-        Coupling.config({
-          others: {
-            toolbarSandbox: (button) => {
-              return makeSandbox(button, spec, detail, ariaControls);
+    ...Button.sketch({
+      ...externals.button(),
+      action: (button) => {
+        toggle(button, externals, detail, spec, ariaControls);
+      },
+      buttonBehaviours: SketchBehaviours.augment(
+        { dump: externals.button().buttonBehaviours },
+        [
+          Coupling.config({
+            others: {
+              toolbarSandbox: (button) => {
+                return makeSandbox(button, spec, detail, ariaControls);
+              }
             }
-          }
-        })
-      ]
-    )
-  }),
-  apis: {
-    setGroups: (button: AlloyComponent, groups: AlloySpec[]) => {
-      Sandboxing.getState(Coupling.getCoupled(button, 'toolbarSandbox')).each((toolbar) => {
-        setGroups(button, toolbar, detail, spec.layouts, groups);
-      });
-    },
-    reposition: (button: AlloyComponent) => {
-      Sandboxing.getState(Coupling.getCoupled(button, 'toolbarSandbox')).each((toolbar) => {
-        position(button, toolbar, detail, spec.layouts);
-      });
-    },
-    toggle: (button: AlloyComponent, options: { skipFocus: boolean }) => {
-      toggle(button, externals, detail, spec, ariaControls, options);
-    },
-    getToolbar: (button: AlloyComponent) => {
-      return Sandboxing.getState(Coupling.getCoupled(button, 'toolbarSandbox'));
-    },
-    isOpen: (button: AlloyComponent) => {
-      return Sandboxing.isOpen(Coupling.getCoupled(button, 'toolbarSandbox'));
+          })
+        ]
+      )
+    }),
+    apis: {
+      setGroups: (button: AlloyComponent, groups: AlloySpec[]) => {
+        Sandboxing.getState(Coupling.getCoupled(button, 'toolbarSandbox')).each((toolbar) => {
+          setGroups(button, toolbar, detail, spec.layouts, groups);
+        });
+      },
+      reposition: (button: AlloyComponent) => {
+        Sandboxing.getState(Coupling.getCoupled(button, 'toolbarSandbox')).each((toolbar) => {
+          position(button, toolbar, detail, spec.layouts);
+        });
+      },
+      toggle: (button: AlloyComponent, options: { skipFocus: boolean }) => {
+        toggle(button, externals, detail, spec, ariaControls, options);
+      },
+      getToolbar: (button: AlloyComponent) => {
+        return Sandboxing.getState(Coupling.getCoupled(button, 'toolbarSandbox'));
+      },
+      isOpen: (button: AlloyComponent) => {
+        return Sandboxing.isOpen(Coupling.getCoupled(button, 'toolbarSandbox'));
+      }
     }
-  }
-  }};
+  };
+};
 
 const FloatingToolbarButton: FloatingToolbarButtonSketcher = Sketcher.composite<FloatingToolbarButtonSpec, FloatingToolbarButtonDetail, FloatingToolbarButtonApis>({
   name: 'FloatingToolbarButton',
--- a/modules/alloy/src/main/ts/ephox/alloy/api/ui/FloatingToolbarButton.ts
+++ b/modules/alloy/src/main/ts/ephox/alloy/api/ui/FloatingToolbarButton.ts
@@ -33,8 +33,8 @@ const onClose = (
   options?: { skipFocus: boolean }
 ) => {
   Sandboxing.close(toolbar);
+  Toggling.off(button);
   if (!options?.skipFocus) {
-    Toggling.off(button);
     Focusing.focus(button);
   }
   ariaControls.unlink(button.element);
--- a/modules/alloy/src/main/ts/ephox/alloy/api/ui/FloatingToolbarButton.ts
+++ b/modules/alloy/src/main/ts/ephox/alloy/api/ui/FloatingToolbarButton.ts
@@ -1,4 +1,4 @@
-import { Fun, Optional } from '@ephox/katamari';
+import { Fun, Optional, Singleton } from '@ephox/katamari';
 
 import * as ComponentStructure from '../../alien/ComponentStructure';
 import * as AriaControls from '../../aria/AriaControls';
@@ -26,44 +26,17 @@ import * as Sketcher from './Sketcher';
 import { Toolbar } from './Toolbar';
 import { CompositeSketchFactory } from './UiSketcher';
 
-const onClose = (
-  button: AlloyComponent,
-  toolbar: AlloyComponent,
-  ariaControls: AriaControls.AriaManager,
-  options?: { skipFocus: boolean }
-) => {
-  Sandboxing.close(toolbar);
-  Toggling.off(button);
-  if (!options?.skipFocus) {
-    Focusing.focus(button);
-  }
-  ariaControls.unlink(button.element);
-};
+const shouldSkipFocus = Singleton.value<boolean>();
 
-const toggle = (
-  button: AlloyComponent,
-  externals: Record<string, any>,
-  detail: FloatingToolbarButtonDetail,
-  spec: FloatingToolbarButtonSpec,
-  ariaControls: AriaControls.AriaManager,
-  options?: { skipFocus: boolean }
-) => {
+const toggle = (button: AlloyComponent, externals: Record<string, any>, options?: { skipFocus: boolean }) => {
+  shouldSkipFocus.set(options?.skipFocus || false);
   const toolbarSandbox = Coupling.getCoupled(button, 'toolbarSandbox');
-
   if (Sandboxing.isOpen(toolbarSandbox)) {
-    onClose(button, toolbarSandbox, ariaControls, options);
+    Sandboxing.close(toolbarSandbox);
   } else {
     Sandboxing.open(toolbarSandbox, externals.toolbar());
-    detail.fetch().get((groups) => {
-      Sandboxing.getState(toolbarSandbox).each((toolbar) => {
-        setGroups(button, toolbar, detail, spec.layouts, groups);
-        ariaControls.link(button.element);
-        if (!options?.skipFocus) {
-          Keying.focusIn(toolbar);
-        }
-      });
-    });
   }
+  shouldSkipFocus.clear();
 };
 
 const position = (button: AlloyComponent, toolbar: AlloyComponent, detail: FloatingToolbarButtonDetail, layouts: Layouts | undefined) => {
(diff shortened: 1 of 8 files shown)


ephox-mogran (changes requested):
You need to be really careful when removing focus calls. Remember, that the editor needs to be completely keyboard accessible. If you remove this focus, then a keyboard user can't start navigating inside the floating toolbar more section once they open it with an `<enter>` key.

Similarly, you need to check that pressing `<escape>` inside the floating toolbar section will close the section and focus the triggering button. It seems this is still working, because it is probably being done by something else as well ... but would that mean that when closing `skip_focus` is being ignored? 

You'll also need to update the test: `modules/alloy/src/test/ts/browser/ui/button/FloatingToolbarButtonTest.ts`. It looks like it should probably have another test to explicitly check focus has transferred into the floating section after being opened (by the keyboard or mouse), to more clearly identify that this is a requirement.

spocke:
I'm with @ephox-mogran this probably need to be that harder path of supporting the `skip_focus` parameter for the command. I think we can string that though by just passing in the extra args at the end of the function that you register for a command. Currently when you register a command you provide a function with 2 parameters if you can have 3 and the new one is option so basically (ui, value?, args?). But not sure how to handle the not focusing part how to string that though all of silver.

yacodes:
I was thinking about an additional parameter too. We do have a couple of callbacks which react to `grabFocus` key argument. Maybe we should add the similar behaviour here as well.

yacodes:
@ephox-mogran @spocke Updated the PR with the additional `skipFocus: true` option for the `ToggleToolbarDrawer` command. It took a bit of refactoring to pass it all the way down to the Alloy components, making the changes a bit big.

modules/tinymce/src/themes/silver/test/ts/browser/editor/toolbar/ToolbarDrawerToggleTest.ts
```diff
@@ -107,4 +107,27 @@ describe('browser.tinymce.themes.silver.editor.toolbar.ToolbarDrawerToggleTest',
       });
     });
   });
+
+  context(`Should preserve focus if skipFocus: true option was passed`, () => {
```
ephox-mogran, line 111: I think you'll need to add tests for when `skipFocus` is explicitly set to false, also.

modules/tinymce/src/themes/silver/main/ts/ui/general/OuterContainer.ts
```diff
@@ -81,7 +81,7 @@ interface OuterContainerApis {
 interface ToolbarApis {
   readonly setGroups: (toolbar: AlloyComponent, groups: SketchSpec[]) => void;
   readonly refresh: (toolbar: AlloyComponent) => void;
-  readonly toggle?: (toolbar: AlloyComponent) => void;
+  readonly toggle?: (toolbar: AlloyComponent, options?: { skipFocus: boolean }) => void;
```
ephox-mogran, line 84: Generally, alloy has avoided variable arguments, and relied on boulder etc. to standardise. This one is a bit different, though, as it's not just the spec coming in, but an argument to an API.

I think many of my issues with defaulted arguments stem from pre-Typescript days, where it made it much more likely that a typo wouldn't be picked up. However, I _think_ I'd still prefer to see something more like: 

`toggle`
`toggleWithoutFocusing`

And then _internally_, they can use an argument passed to a common function, but the APIs would be separate. I don't really know if I can justify that, though ... and the naming I have chosen there is horrible. Just something to think about.

modules/alloy/src/main/ts/ephox/alloy/api/ui/FloatingToolbarButton.ts
```diff
@@ -26,12 +26,43 @@ import * as Sketcher from './Sketcher';
...
+  options?: { skipFocus: boolean }
+) => {
   const toolbarSandbox = Coupling.getCoupled(button, 'toolbarSandbox');
+
   if (Sandboxing.isOpen(toolbarSandbox)) {
-    Sandboxing.close(toolbarSandbox);
+    onClose(button, toolbarSandbox, ariaControls, options);
   } else {
     Sandboxing.open(toolbarSandbox, externals.toolbar());
+    detail.fetch().get((groups) => {
```
ephox-mogran, line 57: What's the main reason you inlined this, rather than just pulling out an `onOpen` to match the `onClose`?

modules/alloy/src/main/ts/ephox/alloy/api/ui/FloatingToolbarButton.ts
```diff
@@ -26,12 +26,43 @@ import * as Sketcher from './Sketcher';
...
+      Sandboxing.getState(toolbarSandbox).each((toolbar) => {
+        setGroups(button, toolbar, detail, spec.layouts, groups);
```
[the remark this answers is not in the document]
yacodes: Fixed [here](https://github.com/tinymce/tinymce/pull/8272/commits/a8afe989fe2daaada95fa531a863d81d5b65512f).

modules/alloy/src/main/ts/ephox/alloy/api/ui/FloatingToolbarButton.ts
yacodes: The reason for this to get access to the `skipFocus` parameter. I couldn't find a way to get it's value from `onOpen` and `onClose` callbacks.

modules/tinymce/src/themes/silver/test/ts/browser/editor/toolbar/ToolbarDrawerToggleTest.ts
yacodes: Added the additional test [here](https://github.com/tinymce/tinymce/pull/8272/commits/cc1abae83b76cf99918426cb99b0a854a59208df).

modules/tinymce/src/themes/silver/main/ts/ui/general/OuterContainer.ts
yacodes: I am not sure I can understand which problem this API splitting can help us solve. :(

modules/alloy/src/main/ts/ephox/alloy/api/ui/FloatingToolbarButton.ts
```diff
@@ -57,111 +88,99 @@ const setGroups = (button: AlloyComponent, toolbar: AlloyComponent, detail: Floa
...
+    [
+      Keying.config({
+        mode: 'special',
+        onEscape: (toolbar) => {
+          onClose(button, toolbar, ariaControls);
+          return Optional.some<boolean>(true);
+        }
+      }),
+      Sandboxing.config({
+        isPartOf: (container, data, queryElem): boolean => {
```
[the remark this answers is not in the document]
ephox-mogran: > Is there absolutely no way to let these callbacks know of the skipFocus option from the external API?

These are all bad options, but they exist :)

a) mark the DOM in some way temporarily, that is removed after the function call
b) store the state in some way that is removed after the function call

The problem with approaches like this is that it's very easy for the "remove" part not to fire, and then they just stay in their "won't transfer focus state" any more. Having said that, maybe we could just explicitly remove them again when interacting with the keyboard. It might be the best bad option.

modules/tinymce/src/themes/silver/main/ts/ui/general/OuterContainer.ts
ephox-mogran: Primarily just that alloy doesn't really use defaulted parameters anywhere. At least it didn't before. But it's up to you. It's mainly a thing of consistency.

yacodes: @ephox-mogran Removed the `options` parameter from alloy components in favor for `toggle` and `toggleWithoutFocusing` functions.

modules/alloy/src/main/ts/ephox/alloy/api/ui/FloatingToolbarButton.ts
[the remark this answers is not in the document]
yacodes: @ephox-mogran Updated the solution by storing a `shouldSkipFocus` boolean in a singleton before `onOpen` and `onClose` callbacks are called. This works since the callbacks are synchronous. This singleton value is cleared after the call. Let me know what you think! :)

modules/alloy/src/main/ts/ephox/alloy/api/ui/FloatingToolbarButton.ts
```diff
@@ -61,17 +69,22 @@ const makeSandbox = (button: AlloyComponent, spec: FloatingToolbarButtonSpec, de
   const ariaControls = AriaControls.manager();
 
   const onOpen = (sandbox: AlloyComponent, toolbar: AlloyComponent) => {
+    const skipFocus = shouldSkipFocus.get().getOr(false);
```
hamza0867, line 72: Is seems this variable can be computed outside of the `onOpen` callback. You can move this line out of the callback and use it inside `onOpen` and `onClose` callback as well

modules/alloy/src/main/ts/ephox/alloy/api/ui/FloatingToolbarButton.ts
```diff
@@ -61,17 +69,22 @@ const makeSandbox = (button: AlloyComponent, spec: FloatingToolbarButtonSpec, de
...
+        Keying.focusIn(toolbar);
+      }
     });
   };
 
   const onClose = () => {
     // Toggle and focus the button
     Toggling.off(button);
-    Focusing.focus(button);
+    if (!shouldSkipFocus.get().getOr(false)) {
```
hamza0867, line 85: Here you can use the `skipFocus` variable as suggested in the previous comment

hamza0867 (approved):

modules/tinymce/src/themes/silver/main/ts/Render.ts
```diff
@@ -425,8 +425,8 @@ const setup = (editor: Editor): RenderInfo => {
       OuterContainer.focusToolbar(outerContainer);
     });
 
-    editor.addCommand('ToggleToolbarDrawer', () => {
-      OuterContainer.toggleToolbarDrawer(outerContainer);
+    editor.addCommand('ToggleToolbarDrawer', (_ui, options?: { skipFocus: boolean }) => {
+      OuterContainer[options?.skipFocus ? 'toggleToolbarDrawerWithoutFocusing' : 'toggleToolbarDrawer'](outerContainer);
```
TheSpyder, line 429: This is a really confusing code style, please just reference the methods directly. In most cases (I realise not this one) this sort of thing actually breaks tree shaking, because we use `import * as` which doesn't actually create the object it just simulates it at compile time. Unless code like this is used; then the object is created.

- [ ] fix

TheSpyder (commented):
I think it's ok? If morgan is happy I'm happy (so I'm not approving, since I won't be around to follow up for 3 weeks)

modules/alloy/src/main/ts/ephox/alloy/api/ui/FloatingToolbarButton.ts
yacodes: Unfortunately, this won't work. Since this variable can be changed *just* before the `onOpen` and `onClose` callbacks are called. It will be always the same if we will move this line out and this is not the intention.

modules/tinymce/src/themes/silver/main/ts/Render.ts
yacodes: Updated [here](https://github.com/tinymce/tinymce/pull/8272/commits/d7dac2d8efe212d68eb2c0911f9b300a9dc37386).

modules/alloy/src/main/ts/ephox/alloy/api/ui/FloatingToolbarButton.ts
```diff
@@ -26,6 +26,14 @@ import * as Sketcher from './Sketcher';
 import { Toolbar } from './Toolbar';
 import { CompositeSketchFactory } from './UiSketcher';
 
+const shouldSkipFocus = Singleton.value<boolean>();
+
+const toggleWithoutFocusing = (button: AlloyComponent, externals: Record<string, any>) => {
+  shouldSkipFocus.set(true);
```
ephox-mogran, line 32: We can try it, but this nearly always ends badly :)

modules/alloy/src/main/ts/ephox/alloy/api/ui/SplitFloatingToolbar.ts
```diff
@@ -87,6 +87,11 @@ const factory: CompositeSketchFactory<SplitFloatingToolbarDetail, SplitFloatingT
           FloatingToolbarButton.toggle(floatingToolbarButton);
         });
       },
+      toggleWithoutFocusing: (toolbar: AlloyComponent) => {
+        memFloatingToolbarButton.getOpt(toolbar).each((floatingToolbarButton) => {
+          FloatingToolbarButton.toggleWithoutFocusing(floatingToolbarButton);
```
ephox-mogran, line 92: You can inline this. It might be a bit nicer:

```
memFloatingToolbarButton.getOpt(toolbar).each(FloatingToolbarButton.toggleWithoutFocusing)
```

modules/tinymce/src/themes/silver/main/ts/Render.ts
```diff
@@ -425,8 +425,12 @@ const setup = (editor: Editor): RenderInfo => {
       OuterContainer.focusToolbar(outerContainer);
     });
 
-    editor.addCommand('ToggleToolbarDrawer', () => {
-      OuterContainer.toggleToolbarDrawer(outerContainer);
+    editor.addCommand('ToggleToolbarDrawer', (_ui, options?: { skipFocus: boolean }) => {
+      if (options?.skipFocus) {
```
ephox-mogran, line 429: You could also use a ternary here. I like ternaries because they force a value regardless of the branching. But I might be in the minority there. Up to you.

modules/tinymce/src/themes/silver/main/ts/ui/general/OuterContainer.ts
```diff
@@ -134,6 +136,11 @@ const factory: UiSketcher.CompositeSketchFactory<OuterContainerSketchDetail, Out
         Optionals.mapFrom(toolbar.getApis<ToolbarApis>().toggle, (toggle) => toggle(toolbar));
       });
     },
+    toggleToolbarDrawerWithoutFocusing: (comp) => {
+      Composite.parts.getPart(comp, detail, 'toolbar').each((toolbar) => {
+        Optionals.mapFrom(toolbar.getApis<ToolbarApis>().toggleWithoutFocusing, (toggleWithoutFocusing) => toggleWithoutFocusing(toolbar));
```
ephox-mogran, line 141: Generally, you would always use the Sketcher API directly, like 

```
SplitFloatingToolbar.toggleWithoutFocusing
```

So this could become 

```
Composite.parts.getPart(comp, detail, 'toolbar').each(SplitFloatingToolbar.toggleWithoutFocusing)
```

Is the reason you avoided using that you wanted to handle when `toggleWithoutFocusing` was not defined because the `toolbar` part component wasn't a `SplitFloatingToolbar` ?

modules/tinymce/src/themes/silver/test/ts/browser/editor/toolbar/ToolbarDrawerToggleTest.ts
```diff
@@ -107,4 +107,50 @@ describe('browser.tinymce.themes.silver.editor.toolbar.ToolbarDrawerToggleTest',
...
+          width: 200,
+          toolbar_mode: toolbarMode,
+          base_url: '/project/tinymce/js/tinymce'
+        });
+        await UiUtils.pWaitForEditorToRender();
+        editor.focus();
+        const initialFocusedElement = document.activeElement;
+        editor.execCommand('ToggleToolbarDrawer', false, { skipFocus: true });
+        await TinyUiActions.pWaitForUi(editor, '.tox-toolbar__overflow');
+        assert.equal(initialFocusedElement, document.activeElement, 'Focus should be preserved');
```
ephox-mogran, line 126: Generally, you would use `FocusTools` instead of `document.activeElement` directly. This is especially true as it simplifies the path forward to more compete shadow DOM Support. There are also convenience functions (e.g. `pTryOnSelector`) that can make this a bit more streamlined.

- [ ] switch to using FocusTools where possible

ephox-mogran (approved):
Some minor things, but I'll approve it.

The singleton will bite us, though :P

modules/alloy/src/main/ts/ephox/alloy/api/ui/FloatingToolbarButton.ts
yacodes: I'm out of alternative options here, unfortunately. This seems like the most readable solution. Marking the DOM will provide similar results. It's just DOM instead of the singleton.

modules/alloy/src/main/ts/ephox/alloy/api/ui/SplitFloatingToolbar.ts
yacodes: Thanks! Simplified [here](https://github.com/tinymce/tinymce/pull/8272/commits/6234a9905af406f5fa199f372df6408589fbf5b0).

modules/tinymce/src/themes/silver/main/ts/Render.ts
yacodes: I like ternaries too :) However we are not going to return any value here, so I'd leave the if statement.

modules/tinymce/src/themes/silver/main/ts/ui/general/OuterContainer.ts
yacodes: > Is the reason you avoided using that you wanted to handle when `toggleWithoutFocusing` was not defined because the `toolbar` part component wasn't a `SplitFloatingToolbar` ?

Yes, I've followed the `toggleToolbarDrawer` function example.

modules/tinymce/src/themes/silver/test/ts/browser/editor/toolbar/ToolbarDrawerToggleTest.ts
yacodes: Updated [here](https://github.com/tinymce/tinymce/pull/8272/commits/3eb16a0e34d89b19e5ace1e84413596938211b85).

PR #8294, opened 2022-11-23
TINY-9148: Fix scrolling to the bottom after closing a dialog in Safari

yacodes:
Related Ticket: TINY-9148

Description of Changes:
* Changes `position` style value from `relative` to `absolute` in alloy's positioning behaviour.

The issue is that after removing the dialog from the sink Safari tries to focus the sink if it's positioning is `relative`, thus scrolling the page down. Using `absolute` value for positioning helps prevent the scrolling and keeps the sink in the same place. 

That's a bit of a radical change though. Maybe use it only in Safari?

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/alloy/CHANGELOG.md
+++ b/modules/alloy/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
 
+### Changed
+- Changed positioning behaviour position style from `relative` to `absolute` to avoid Safari autoscroll. #TINY-9148
+
 ## 12.0.0 - 2022-11-23
 
 ### Added
--- a/modules/alloy/CHANGELOG.md
+++ b/modules/alloy/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ## Unreleased
 
 ### Changed
-- Changed positioning behaviour position style from `relative` to `absolute` to avoid Safari autoscroll. #TINY-9148
+- Changed `useFixed` function to `usePositioningType` to allow usage of `absolute` positioning which fixes autoscrolling in Safari. #TINY-9148
 
 ## 12.0.0 - 2022-11-23
 
--- a/modules/alloy/src/main/ts/ephox/alloy/behaviour/positioning/ActivePosition.ts
+++ b/modules/alloy/src/main/ts/ephox/alloy/behaviour/positioning/ActivePosition.ts
@@ -1,12 +1,12 @@
 import { DomDefinitionDetail } from '../../dom/DomDefinition';
 import * as DomModification from '../../dom/DomModification';
-import { PositioningConfig, PositioningType  } from './PositioningTypes';
+import { PositioningConfig, PositioningType } from './PositioningTypes';
 
 const getStyles = (positioningType: PositioningType): Record<string, string> => {
   if (positioningType === 'relative') {
     return { position: 'relative' };
   } else if (positioningType === 'absolute') {
-    // Elements with `position: absolute` and without any content collapse, 
+    // Elements with `position: absolute` and without any content collapse,
     // so theirs width equals 0. That differs from `position: relative` behaviour,
     // which preserves containers width.
     return { position: 'absolute', width: '100%' };
@@ -15,7 +15,7 @@ const getStyles = (positioningType: PositioningType): Record<string, string> => 
   }
 };
 
-const exhibit = (base: DomDefinitionDetail, posConfig: PositioningConfig): DomModification.DomModification => 
+const exhibit = (base: DomDefinitionDetail, posConfig: PositioningConfig): DomModification.DomModification =>
   DomModification.nu({
     classes: [ ],
     styles: getStyles(posConfig.usePositioningType()),
--- a/modules/alloy/src/main/ts/ephox/alloy/behaviour/positioning/ActivePosition.ts
+++ b/modules/alloy/src/main/ts/ephox/alloy/behaviour/positioning/ActivePosition.ts
@@ -5,7 +5,7 @@ import { PositioningConfig } from './PositioningTypes';
 const exhibit = (base: DomDefinitionDetail, posConfig: PositioningConfig): DomModification.DomModification =>
   DomModification.nu({
     classes: [ ],
-    styles: posConfig.useFixed() ? { } : { position: 'relative' }
+    styles: posConfig.useFixed() ? { } : { position: 'absolute', width: '100%' }
   });
 
 export {
--- a/modules/alloy/src/main/ts/ephox/alloy/behaviour/positioning/ActivePosition.ts
+++ b/modules/alloy/src/main/ts/ephox/alloy/behaviour/positioning/ActivePosition.ts
@@ -6,6 +6,9 @@ const getStyles = (positioningType: PositioningType): Record<string, string> => 
   if (positioningType === 'relative') {
     return { position: 'relative' };
   } else if (positioningType === 'absolute') {
+    // Elements with `position: absolute` and without any content collapse, 
+    // so theirs width equals 0. That differs from `position: relative` behaviour,
+    // which preserves containers width.
     return { position: 'absolute', width: '100%' };
   } else {
     return { };
--- a/modules/alloy/src/main/ts/ephox/alloy/behaviour/positioning/ActivePosition.ts
+++ b/modules/alloy/src/main/ts/ephox/alloy/behaviour/positioning/ActivePosition.ts
@@ -1,15 +1,21 @@
 import { DomDefinitionDetail } from '../../dom/DomDefinition';
 import * as DomModification from '../../dom/DomModification';
-import { PositioningConfig } from './PositioningTypes';
+import { PositioningConfig, PositioningType  } from './PositioningTypes';
 
-const exhibit = (base: DomDefinitionDetail, posConfig: PositioningConfig): DomModification.DomModification =>
+const getStyles = (positioningType: PositioningType): Record<string, string> => {
+  if (positioningType === 'relative') {
+    return { position: 'relative' };
+  } else if (positioningType === 'absolute') {
+    return { position: 'absolute', width: '100%' };
+  } else {
+    return { };
+  }
+};
+
+const exhibit = (base: DomDefinitionDetail, posConfig: PositioningConfig): DomModification.DomModification => 
   DomModification.nu({
     classes: [ ],
-    styles: {
-      fixed: { },
-      relative: { position: 'relative' },
-      absolute: { position: 'absolute', width: '100%' }
-    }[posConfig.usePositioningType()]
+    styles: getStyles(posConfig.usePositioningType()),
   });
 
 export {
--- a/modules/alloy/src/main/ts/ephox/alloy/behaviour/positioning/ActivePosition.ts
+++ b/modules/alloy/src/main/ts/ephox/alloy/behaviour/positioning/ActivePosition.ts
@@ -5,7 +5,11 @@ import { PositioningConfig } from './PositioningTypes';
 const exhibit = (base: DomDefinitionDetail, posConfig: PositioningConfig): DomModification.DomModification =>
   DomModification.nu({
     classes: [ ],
-    styles: posConfig.useFixed() ? { } : { position: 'absolute', width: '100%' }
+    styles: {
+      fixed: { },
+      relative: { position: 'relative' },
+      absolute: { position: 'absolute', width: '100%' }
+    }[posConfig.usePositioningType()]
   });
 
 export {
--- a/modules/alloy/src/main/ts/ephox/alloy/behaviour/positioning/PositionApis.ts
+++ b/modules/alloy/src/main/ts/ephox/alloy/behaviour/positioning/PositionApis.ts
@@ -15,6 +15,9 @@ import { Transition } from '../../positioning/view/Transitions';
 import { PlacementDetail, PlacementSpec, PositioningConfig, PositioningState } from './PositioningTypes';
 import { PlacementSchema } from './PositionSchema';
 
+const isFixed = (posConfig: PositioningConfig): boolean =>
+  posConfig.usePositioningType() === 'fixed';
+
 const getFixedOrigin = (): Origins.OriginAdt => {
   // Don't use window.innerWidth/innerHeight here, as we don't want to include scrollbars
   // since the right/bottom position is based on the edge of the scrollbar not the window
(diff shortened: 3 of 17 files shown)


ephox-mogran:
> That's a bit of a radical change though. Maybe use it only in Safari?

I'd tend to advise against having such a potentially significant source of edge cases branch on browser.

This one is interesting. We'd always just used relative -> absolute to position with respect to the sink, but it turns out you are correct, it will also happen if there is position: absolute on the sink. So looking at MDN (https://developer.mozilla.org/en-US/docs/Web/CSS/position):

relative
**The element is positioned according to the normal flow of the document**, and then offset relative to itself based on the values of top, right, bottom, and left. The offset does not affect the position of any other elements; thus, the space given for the element in the page layout is the same as if position were static.

This value creates a new [stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context) when the value of z-index is not auto. Its effect on table-*-group, table-row, table-column, table-cell, and table-caption elements is undefined.

absolute
**The element is removed from the normal document flow, and no space is created for the element in the page layout**. It is positioned relative to its closest positioned ancestor, if any; otherwise, it is placed relative to the initial [containing block](https://developer.mozilla.org/en-US/docs/Web/CSS/Containing_block). Its final position is determined by the values of top, right, bottom, and left.

This value creates a new [stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context) when the value of z-index is not auto. The margins of absolutely positioned boxes do not [collapse](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Box_Model/Mastering_margin_collapsing) with other margins.

---

The main difference seems to be whether or not it is in the normal document flow. I'm not convinced switching relative to absolute everywhere is going to be a good solution, but our usage of the UI sink at least really fits more in with the idea of "not being in the document flow". Maybe we should change `useFixed` to something that lets you set: 'fixed', 'relative', or 'absolute'. That way, alloy can support both depending on the use case.

yacodes:
@ephox-mogran Thanks for such an elaborated reply! Updated `useFixed` function to `usePositioningType` which now should return `'fixed' | 'relative' | 'absolute'`. This allows to use different positions in different cases.

modules/alloy/CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
 
+### Changed
```
ephox-mogran, line 9: This will be a breaking change, so you'll need to put a major bump in `versions.txt`

modules/alloy/src/main/ts/ephox/alloy/behaviour/positioning/ActivePosition.ts
```diff
@@ -5,7 +5,11 @@ import { PositioningConfig } from './PositioningTypes';
 const exhibit = (base: DomDefinitionDetail, posConfig: PositioningConfig): DomModification.DomModification =>
   DomModification.nu({
     classes: [ ],
-    styles: posConfig.useFixed() ? { } : { position: 'relative' }
+    styles: {
```
ephox-mogran, line 8: This is pretty difficult to follow and a bit risky. I think now that we can't use a ternary because there's more than 2 things, because as a general rule we avoid nested ternaries, we should just use `if / else` etc. Here, if `usePositioningType` has an invalid value (which is definitely possible because boulder types gets coerced), then this would be undefined. So something like: 

```
const getStyles = (positioningType: 'fixed' | 'relative' | 'absolute'): Record<string, string> => {
  if (...) {
    return ...
  } else if (...) {
  
  } else {
    return { }
  }
};
```
might be better.

modules/alloy/src/main/ts/ephox/alloy/behaviour/positioning/PositionApis.ts
```diff
@@ -63,7 +63,9 @@ const positionWithinBounds = (component: AlloyComponent, posConfig: PositioningC
     // We need to calculate the origin (esp. the bounding client rect) *after* we have done
     // all the preprocessing of the component and placee. Otherwise, the relative positions
     // (bottom and right) will be using the wrong dimensions
-    const origin = posConfig.useFixed() ? getFixedOrigin() : getRelativeOrigin(component);
+    const origin = posConfig.usePositioningType() === 'fixed'
```
ephox-mogran, line 66: Instead of doing an `=== 'fixed'` check everywhere, we should be an `isFixed` function somewhere. It might make sense to put it in `Positioning` as one of its ExtraApis that just takes in config

modules/alloy/src/main/ts/ephox/alloy/behaviour/positioning/PositionSchema.ts
```diff
@@ -9,7 +9,7 @@ const TransitionSchema = [
 ];
 
 export const PositionSchema = [
-  FieldSchema.defaulted('useFixed', Fun.never),
+  FieldSchema.defaulted('usePositioningType', Fun.constant('relative')),
```
ephox-mogran, line 12: This type needs to be more restrictive. Boulder should fail if we pass through something random like "usePositioningType: random-guess". Look at what other functions there are in `FieldSchema`.

ephox-mogran: (especially because of how the other part was implemented where if the positioningType wasn't one of a set of possible values, styles became undefined)

ephox-mogran: - [ ] lock down possible values more through boulder.

modules/alloy/src/main/ts/ephox/alloy/behaviour/positioning/ActivePosition.ts
```diff
@@ -5,7 +5,11 @@ import { PositioningConfig } from './PositioningTypes';
...
+      fixed: { },
+      relative: { position: 'relative' },
+      absolute: { position: 'absolute', width: '100%' }
```
ephox-mogran, line 11: Hmm. The `width: 100%` seems a bit special-cased. What's the reason for it? Can we put an explanatory comment here about why position: absolute needs a width: 100% and other things don't.

Remember, there is supposed to be nothing about `Positioning` that makes it a tinymce sink. That's just how we use it. But Positioning could just be on something inside. In fact, if you look at some of the sketchers, you'll see that some have the option of an "internal sink". The idea there is that it would let you show the menu inside the DOM hierarchy itself, which we wanted to keep as an option if we needed it for accessibility.

modules/tinymce/src/themes/silver/main/ts/Render.ts
```diff
@@ -279,7 +279,9 @@ const setup = (editor: Editor): RenderInfo => {
       },
       behaviours: Behaviour.derive([
         Positioning.config({
-          useFixed: () => header.isDocked(lazyHeader)
+          // TINY-9148: Using absolute positioning here because Safari
+          // tries to focus the sink after it's content is changed with relative positioning.
```
ephox-mogran, line 283: Minor: "its" :)

modules/alloy/src/main/ts/ephox/alloy/behaviour/positioning/PositionApis.ts
ephox-mogran: We would be able to call it in the `getMode` function below as well.

ephox-mogran (commented):
Almost there. I'm still less than convinced by the solution, but that's just because it seems like it will have far-reaching implications, and seems risky to do without a lot of QA. We know that having position: absolute fixes this case, but what cases will we encounter because we aren't doing relative any more.

Having said all that, as I mentioned before, the tinymce sinks probably shouldn't be in the layout. So it's hard to say.

The other times we were addressing this problem, we've just talked mostly about transferring focus/selection away or even explicitly blurring things. That has its own side-effects, but did you explore that as an option as well when trying to fix this? Did it work? Or were you concerned there would be more problems with that approach?

modules/alloy/CHANGELOG.md
yacodes: Bumped in `versions.txt` and in `alloy/package.json` as well [here](https://github.com/tinymce/tinymce/pull/8294/commits/c709ad0dd777e3f481d7ed976384ccd6ca1ad63a).

modules/alloy/src/main/ts/ephox/alloy/behaviour/positioning/ActivePosition.ts
yacodes: Refactored [here](https://github.com/tinymce/tinymce/pull/8294/commits/a49621cb68f9160ca0a31e740078186ae89718a8).

modules/alloy/src/main/ts/ephox/alloy/behaviour/positioning/PositionApis.ts
yacodes: Updated [here](https://github.com/tinymce/tinymce/pull/8294/commits/3f69566aaac7634750c0a703b7eff61a9e091b16)

modules/alloy/src/main/ts/ephox/alloy/behaviour/positioning/PositionSchema.ts
yacodes: Came up with something like [this](https://github.com/tinymce/tinymce/pull/8294/commits/25526a655436472e32aaebbea9ed7dd3e9c9d8e6).

modules/tinymce/src/themes/silver/main/ts/Render.ts
yacodes: Updated [here](https://github.com/tinymce/tinymce/pull/8294/commits/2ff1f0ca036bc59ad59b39d80451ef0445333047).

modules/alloy/src/main/ts/ephox/alloy/behaviour/positioning/ActivePosition.ts
yacodes: Yeah, the reason for the `width: 100%` is that absolute-positioned elements' width collapse to 0 or its contents. This makes quite a lot of dropdowns look ugly and squished, cause they do not have enough width. This can be fixed in the CSS as well but a lot more hardly.

yacodes:
> Almost there. I'm still less than convinced by the solution, but that's just because it seems like it will have far-reaching implications, and seems risky to do without a lot of QA. We know that having position: absolute fixes this case, but what cases will we encounter because we aren't doing relative any more.
> 
> Having said all that, as I mentioned before, the tinymce sinks probably shouldn't be in the layout. So it's hard to say.
> 
> The other times we were addressing this problem, we've just talked mostly about transferring focus/selection away or even explicitly blurring things. That has its own side-effects, but did you explore that as an option as well when trying to fix this? Did it work? Or were you concerned there would be more problems with that approach?

To eliminate the riskiness of this we can apply `position: absolute` just in Safari browser. This way the implication of the change will affect only Safari users, who already have issues with focusing :)

I've tried blurring but it doesn't really help, because focusing is being executed before any blurring. I couldn't find out when exactly the focusing is happening. 

Saving focus and then restoring it felt like a lot of work as well. Will look into it one more time though.

Another solution is to try `float: left`, which extracts the element from the normal flow as well. But I'm not sure that it is that much different from the solution in this PR. Will look into it as well.

modules/alloy/src/main/ts/ephox/alloy/behaviour/positioning/ActivePosition.ts
yacodes: Added a comment [here](https://github.com/tinymce/tinymce/pull/8294/commits/d345b25bde70cbfcdf6cbe658928b7ea6dca3cc5).

ephox-mogran:
> To eliminate the riskiness of this we can apply position: absolute just in Safari browser. This way the implication of the change will affect only Safari users, who already have issues with focusing :)

The main problem with doing something like this is it creates more institutional knowledge that a developer has to learn / keep in their head about why Safari might be behaving drastically differently from other browsers. I would strongly advise against such a large deviation in behaviour across browsers.

ephox-mogran (commented):
The code changes are very clean. Nice work.

I'm still worried about the change, but it's easily reversible now, so if problems arise in QA, we can resolve them, and you don't even need to revert anything in alloy ... just change the Positioning config of a sink.

There is a test failing which is stopping me Approving currently: 

modules/tinymce/src/themes/silver/test/ts/browser/editor/sizing/ResizeNotInRootTest.ts

modules/alloy/src/main/ts/ephox/alloy/behaviour/positioning/PositionSchema.ts
```diff
@@ -9,7 +9,14 @@ const TransitionSchema = [
...
 
 export const PositionSchema = [
-  FieldSchema.defaulted('useFixed', Fun.never),
+  FieldSchema.defaultedOf(
+    'usePositioningType',
+    Fun.constant('relative'),
+    StructureSchema.valueOf((val) =>
+      Arr.contains([ 'fixed', 'relative', 'absolute' ], val())
+        ? Result.value(val)
+        : Result.error('Invalid value for usePositioningType'))),
```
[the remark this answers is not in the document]
TheSpyder: I agree the invalid value needs to be there - adding the valid ones might be a bit verbose though?

- [ ] add invalid value to error message

TheSpyder (approved):
Conditionally approved, pending the test fix

modules/alloy/src/main/ts/ephox/alloy/behaviour/positioning/PositionSchema.ts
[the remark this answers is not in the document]
yacodes: Add the value [here](https://github.com/tinymce/tinymce/pull/8294/commits/5d4dfd01b04ce23738bdf3547e658fb513745cba).

We expect a function which returns a specific string though. Ain't this error message showing an exact value and not a function itself a bit misleading? Are there other ways to check the function's return value here?

modules/tinymce/src/themes/silver/main/ts/Render.ts
```diff
@@ -279,7 +279,9 @@ const setup = (editor: Editor): RenderInfo => {
       },
       behaviours: Behaviour.derive([
         Positioning.config({
-          useFixed: () => header.isDocked(lazyHeader)
+          // TINY-9148: Using absolute positioning here because Safari
```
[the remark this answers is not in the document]
yacodes: Not exactly sure what you mean and what should've been done. :(

yacodes:
@ephox-mogran @TheSpyder I've tried other solutions to this and found out that:
1. Adding `float: left` won't help with the Safari issue;
2. Explicitly blurring the active element before removing it from the sink actually helps and is pretty straightforward. Created a new branch [here](https://github.com/tinymce/tinymce/commit/c3de2deda60b7842d05c89108cc27fd0a1c086d0). What do you think about this solution?

yacodes:
Closing the PR in favor of https://github.com/tinymce/tinymce/pull/8336.

PR #8295, opened 2022-11-24
TINY-4818: Wrap invalid list child nodes within list items

vpyshnenko:
Related Ticket: TINY-4818

Description of Changes:
*  added content filter for lists to wrap invalid child nodes in `li` elements

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/plugins/lists/main/ts/core/FilterContent.ts
+++ b/modules/tinymce/src/plugins/lists/main/ts/core/FilterContent.ts
@@ -1,0 +1,46 @@
+import { Arr } from '@ephox/katamari';
+
+import Editor from 'tinymce/core/api/Editor';
+import AstNode from 'tinymce/core/api/html/Node';
+
+const isListItem = (node: AstNode) => node.name === 'li';
+const isEmpty = (nodeBuffer: AstNode[]) => nodeBuffer.length === 0;
+
+const wrapInvalidChildren = (list: AstNode) => {
+  const insertListItem = (buffer: AstNode[], refNode?: AstNode) => {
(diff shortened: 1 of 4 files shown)


modules/tinymce/src/plugins/lists/main/ts/core/FilterContent.ts
```diff
@@ -0,0 +1,46 @@
...
+  const restBuffer = Arr.foldl(list.children(), reducer, []);
+  if (!isEmpty(restBuffer)) {
+    insertListItem(restBuffer);
+  }
+};
+
+const setup = (editor: Editor): void => {
+  editor.on('PreInit', () => {
+    const { parser } = editor;
+    parser.addNodeFilter('ul, ol', (nodes) => Arr.each(nodes, wrapInvalidChildren ));
```
spocke, line 40:
```suggestion
    parser.addNodeFilter('ul,ol', (nodes) => Arr.each(nodes, wrapInvalidChildren));
```

spocke (approved):
Usually we try to optimize the heck out of these filters since they are some what performance sensitive on large documents. But the code is clean and readable now and I would only be slow for documents with tons and tons of lists. So maybe we wait with optimization until we hit a performance support case. Optimization in this case would be to use tight loops, not create that empty array until we really need to etc.

yacodes (approved):

hamza0867 (approved):

HusamJaber:
this pull request cause us problems please check the issue https://github.com/tinymce/tinymce/issues/9751

PR #8300, opened 2022-11-28
TINY-9302: Add key overrides to prevent insertion of interpreted tags after ranged deletion

mkzhx:
Related Ticket: TINY-9302

Description of Changes:
* Add keydown override for ranged selection deletion when selection is at the start of a container to insert a formatted caret container containing original formats at the cursor after deletion. This retains the original formats independent of native browser behaviour.
* Add keyup override for ranged deletion using keyboard shortcuts to insert an unformatted caret container after deletion to ensure the browser doesn't add interpreted format tags.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

[TINY-9302]: https://ephocks.atlassian.net/browse/TINY-9302?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


modules/tinymce/src/core/main/ts/keyboard/DeleteBackspaceKeys.ts
```diff
@@ -50,24 +51,54 @@ const executeKeydownOverride = (editor: Editor, caret: Cell<Text | null>, evt: K
...
   MatchKeys.execute([
     { keyCode: VK.BACKSPACE, action: MatchKeys.action(CefDelete.paddEmptyElement, editor) },
-    { keyCode: VK.DELETE, action: MatchKeys.action(CefDelete.paddEmptyElement, editor) }
+    { keyCode: VK.DELETE, action: MatchKeys.action(CefDelete.paddEmptyElement, editor) },
+    ...multiDeleteKeyPatterns
   ], evt);
 };
 
 const setup = (editor: Editor, caret: Cell<Text | null>): void => {
+  // global backspace keydown state to emulate Meta + Backspace keyup detection on macOS
```
spocke, line 85: This isn't really a global state that would be for the hole page in my mind.

modules/tinymce/src/core/main/ts/keyboard/DeleteBackspaceKeys.ts
```diff
@@ -50,24 +51,54 @@ const executeKeydownOverride = (editor: Editor, caret: Cell<Text | null>, evt: K
...
+    { keyCode: VK.DELETE, ctrlKey: true, action: MatchKeys.action(InlineFormatDelete.refreshCaret, editor) }
+  ];
+
+  // macOS surpresses keyup events for most keys including Backspace when Meta key is engaged
+  // To emulate Meta + Backspace on macOS, add a pattern for the meta key when backspace was
+  // detected on keydown
+  if (os.isMacOS() && isBackspaceKeydown) {
+    multiDeleteKeyPatterns.push({
+      // firefox detects Cmd as "Command" not "Meta"
+      keyCode: browser.isFirefox() ? 224 : 91,
```
spocke, line 72: And there are no modern things we can use here like `.key` etc they are all incorrect? If firefox ever fixes this bug this code will break but not sure if we can avoid that we can't really feature detect that.

modules/tinymce/src/core/main/ts/delete/InlineFormatDelete.ts
```diff
@@ -47,9 +58,63 @@ const deleteCaret = (editor: Editor, forward: boolean): Optional<() => void> =>
...
+    return Optional.none();
+  }
+};
+
 const backspaceDelete = (editor: Editor, forward: boolean): Optional<() => void> =>
-  editor.selection.isCollapsed() ? deleteCaret(editor, forward) : Optional.none();
+  editor.selection.isCollapsed() ? deleteCaret(editor, forward) : deleteRange(editor);
+
+const hasAncestorInlineCaret = (elm: SugarElement<Node>): boolean =>
+  PredicateFilter.ancestors(elm, (node) => isCaretNode(node.dom), (node) => ElementType.isBlock(node)).length > 0;
```
spocke, line 105: ETA reduce and use PredicateExists instead since you are only interested if it exists or not.

```suggestion
  PredicateExists.ancestor(elm, (node) => isCaretNode(node.dom), ElementType.isBlock);
```

modules/tinymce/src/core/test/ts/browser/delete/InlineFormatDeleteTest.ts
```diff
@@ -283,6 +286,182 @@ describe('browser.tinymce.core.delete.InlineFormatDelete', () => {
...
+                  }),
+                  ...Arr.map(secondOuterText, (text) => s.text(str.is(text))),
+                ]
+              })
+            ]
+          });
+        })
+      );
+      const selPath = browser.isFirefox() ? [ 0, 2, 0, 0, 0, 0 ] : [ 0, 1, 0, 0, 0, 0 ];
+      TinyAssertions.assertSelection(editor, selPath, 0, selPath, 0);
```
spocke, line 464:
```suggestion
      TinyAssertions.assertCursor(editor, selPath, 0);
```

modules/tinymce/src/core/test/ts/webdriver/delete/ShortcutDeleteTest.ts
```diff
@@ -0,0 +1,116 @@
...
+  };
+
+  const assertContentDeletionThenTyping = (editor: Editor): void =>
+    TinyAssertions.assertContent(editor, browser.isFirefox()
+      ? '<p><span style="text-decoration: underline;">d</span></p>'
+      : '<p>d</p>');
+
+  const ctrlModifier: BackspaceDeleteModifier = os.isMacOS() ? { alt: true } : { ctrl: true };
+
+  it('Ctrl/Alt + Backspace at the end of a formatted word', async () => {
```
spocke, line 83: Shouldn't these also have the ticket id prefix?

modules/tinymce/src/core/test/ts/webdriver/delete/ShortcutDeleteTest.ts
```diff
@@ -0,0 +1,116 @@
...
+    indent: false
+  }, [], true);
+  const platform = PlatformDetection.detect();
+  const os = platform.os;
+  const browser = platform.browser;
+
+  const doBackspaceDelete = async (editor: Editor, deletionKey: 'Backspace' | 'Delete', modifiers: BackspaceDeleteModifier): Promise<void> => {
+    await RealKeys.pSendKeysOn('iframe => body', [ RealKeys.combo(modifiers, deletionKey) ]);
+    if (browser.isSafari()) {
+      // safari doesn't trigger this override automatically in test
```
spocke, line 23: Should we then skip the test for safari since it's not really testing the right thing?

spocke (approved):

modules/tinymce/src/core/test/ts/webdriver/delete/ShortcutDeleteTest.ts
mkzhx: I agree, I've skipped the entire test for Safari.

modules/tinymce/src/core/test/ts/webdriver/delete/ShortcutDeleteTest.ts
mkzhx: I'm not sure what the best practice here is. The ticket is on some specific behaviours, which is why I added the ticket id prefixes for the tests that check that the browser no longer adds rogue tags after ranged deletion and then entering text.

The fix resulted in implementing a new override that has a larger scope than the ticket specified. I didn't mark these tests with the ticket ids as they are not directly related to the ticket requirements, but rather related to the fix.

modules/tinymce/src/core/main/ts/keyboard/DeleteBackspaceKeys.ts
mkzhx: True, I clarified the comment to
```
// track backspace keydown state for emulating Meta + Backspace keyup detection on macOS
let isBackspaceKeydown = false;
```

modules/tinymce/src/core/main/ts/keyboard/DeleteBackspaceKeys.ts
mkzhx: `.key` shows `Meta` for both Firefox and Chrome/Safari. However, I think we are reliant on `keyCode` because that is what is accepted by the type `MatchKeys.KeyPattern[]` used in `MatchKeys.execute`. 

Not sure if it is safe to skip browser detection and just add patterns for both `keyCode: 224` and `keyCode: 91`.

modules/tinymce/CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Improved
 - Direct invalid child text nodes of list elements will be wrapped in list item elements. #TINY-4818
 
+### Fixed
+- Ranged deletion of formatted text using selection or keyboard shortcut would sometimes cause Blink and Webkit browsers to insert interpreted tags upon typing, which may affect document semantics. #TINY-9302
```
ephox-mogran, line 13: We might want to choose something a bit more straightforward than "affect document semantics"

modules/tinymce/src/core/main/ts/delete/InlineFormatDelete.ts
```diff
@@ -1,28 +1,42 @@
...
+
+const getParentInlines = (editor: Editor): SugarElement<Node>[] => {
+  const rootElm = SugarElement.fromDom(editor.getBody());
+  const startElm = SugarElement.fromDom(editor.selection.getStart());
   const parents = Parents.parentsAndSelf(startElm, rootElm);
   return Arr.findIndex(parents, ElementType.isBlock).fold(
     Fun.constant(parents),
     (index) => parents.slice(0, index)
-  );
+  ).filter(hasOnlyOneChild);
```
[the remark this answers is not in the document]
ephox-mogran: I imagine it's something to do with inline formatting elements working that way. But any text node fragmentation, zero width cursors etc. might interfere with that. Is that a problem? Or does tinymce just normalise the issue away?

ephox-mogran (approved):
I'm not going to pretend to understand the implications of these changes, but if @spocke has approved it, that's good enough for me.

Just had some minor comments about the code itself.

tiny-ben-tran (approved):

modules/tinymce/CHANGELOG.md
mkzhx: Do you mean I should give the example of `<u>` replacing `<span style="text-decoration: underline;>`?

modules/tinymce/src/core/main/ts/delete/InlineFormatDelete.ts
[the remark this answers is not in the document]
mkzhx: This also raises the issue of why we are applying the one-child filter to 'self' as well (`parents = Parents.parentsAndSelf`). I guess this doesn't apply to the case when `deleteCaret` is triggered like before, but it does introduce a bug in the new `deleteRange`.

Ex:
```
<p><span style="text-decoration: underline;">a<em><strong>b</strong></em></span></p>
```
`getParentInlines` will return `[]` if cursor is at 'a', as selection start (and therefore 'self') is considered to be `<span style="text-decoration: underline;">a<em><strong>b</strong></em></span>`, which contains more than one child. This doesn't make sense as 'a' is clearly formatted with `<span style="text-decoration: underline;">`.

modules/tinymce/CHANGELOG.md
ephox-mogran: I'm not sure exactly. It's just that "affect document semantics" just sounds very formal, and isn't really all that clear. It's hard because it doesn't create deprecated tags ... maybe inconsistent tags?

modules/tinymce/src/core/main/ts/delete/InlineFormatDelete.ts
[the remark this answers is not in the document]
ephox-mogran: Hmm. As discussed, the concern I have with this is that it doesn't make much sense to include elements that are parents of elements that have more than one children, either. Previously, the `deleteCaret` code only looked at the last element, so it wasn't important, but we're working with all of them for `deleteRange` so we probably need to exclude anything that is an ancestor of something with more than one child.

I'm also not sure if that's really accurate, either - but I don't know enough about what formatting structures we'd expect.

modules/tinymce/src/core/main/ts/delete/InlineFormatDelete.ts
```diff
@@ -47,9 +58,63 @@ const deleteCaret = (editor: Editor, forward: boolean): Optional<() => void> =>
...
+
+const updateCaretFormat = (editor: Editor, updateFormats: Node[]): void => {
+  const missingFormats = Arr.difference(updateFormats, getFormatNodesAtStart(editor));
+  if (missingFormats.length > 0) {
+    createCaretFormatAtStart(editor, missingFormats);
+  }
+};
+
+const isSelectionAtStartOfContainer = (editor: Editor): boolean =>
+  editor.selection.getRng().startOffset === 0;
```
ephox-mogran, line 87: This condition seems rather general. The `startOffset` will mean different things depending on what the `startContainer` is. Given that this is the condition that we are using to determine whether or not we should override the delete, we'll need to confirm that anything that passes this condition can run `updateCaretFormat` without unexpected impact.

modules/tinymce/src/core/test/ts/webdriver/delete/ShortcutDeleteTest.ts
mkzhx: Ticket id prefixed all tests added.

modules/tinymce/CHANGELOG.md
mkzhx: Wording clarified:
~affect document semantics~ -> result in inconsistent tags

modules/tinymce/src/core/main/ts/delete/InlineFormatDelete.ts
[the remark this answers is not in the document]
mkzhx: Looking at the tests, I think the goal of `getParentInlines` is to find all single-child inline parents _until_ a multi-child inline parent or block. This is something the `filter` statement doesn't exactly achieve.

For example, with cursor set to `[ 0, 0, 0, 0, 0 ], 1` (between 'a' and 'b') in
```
<p><s><span style="text-decoration: underline;"><em>a</em></span>b</s></p>
```
`getParentInlines` returns 
image
Then, we take the last parent inline, the `<span ...>`, and check `DeleteUtils.willDeleteLastPositionInElement` on it. If it returns true (which it does since there is only a single character within), then the `<span ...>` gets replaced by a caret with `<span ...>` and `<em>` formats.

However, like you said Morgan, if we add an outer ancestor `<strong>` such that it has a single child,
```
<p><strong><s><span style="text-decoration: underline;"><em>a</em></span>b</s></strong></p>
```
`getParentInlines` returns
image
skipping the `<s>` which contains two children.
We check the last parent inline, the `<strong>` in `DeleteUtils.willDeleteLastPositionInElement`, which returns false as there is still a 'b' after deletion within `<strong>`, and the override is not executed. However, we ended up deleting `a` with the `<span ...>` underline and `<em>` formats, which are not retained and typing again will produce the interpreted tags like `<u>`.

This is why I propose we replace the return statement with something such as:
```
  return Arr.findIndex(parents, (x) => ElementType.isBlock(x) || !hasOnlyOneChild(x)).fold(
    Fun.constant(parents),
    (index) => parents.slice(0, index)
  );
```
so that we find all single-child inline parents _until_ a multi-child inline parent or block as required.

This is just to fix up `deleteCaret`, and my `deleteRange` will need to use something that omits `hasOnlyOneChild` completely.

modules/tinymce/src/core/main/ts/delete/InlineFormatDelete.ts
mkzhx: Tightened condition to 
```ts
rangeStartsAtStartOfTextContainer(rng) && rangeStartContainerParentIsFormatElement(editor, rng) && rangeEndsAtOrAfterEndOfStartContainer(rng);
```
This was what I intended target the override for. Tests are added in InlineFormatDeleteOverrideTest.

modules/tinymce/src/core/main/ts/delete/InlineFormatDelete.ts
```diff
@@ -1,28 +1,42 @@
...
 };
 
-const hasOnlyOneChild = (elm: SugarElement<Node>): boolean =>
-  Traverse.childNodesCount(elm) === 1;
-
-const deleteLastPosition = (forward: boolean, editor: Editor, target: SugarElement<Node>, parentInlines: SugarElement<Node>[]) => {
+const getFormatNodes = (editor: Editor, parentInlines: SugarElement<Node>[]): Node[] => {
   const isFormatElement = Fun.curry(CaretFormat.isFormatElement, editor);
-  const formatNodes = Arr.map(Arr.filter(parentInlines, isFormatElement), (elm) => elm.dom);
+  return Arr.map(Arr.filter(parentInlines, isFormatElement), (elm) => elm.dom);
```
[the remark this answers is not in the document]
mkzhx: I'm not sure how this can be converted to `Arr.bind` without making it more complicated.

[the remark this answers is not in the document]
ephox-mogran:
```
Arr.bind(parentInlines, (elem) => isFormatElement(elem) ? [ elm.dom ] : [ ])
```

As I said, YMMV.

modules/tinymce/src/core/main/ts/delete/InlineFormatDelete.ts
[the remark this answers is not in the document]
ephox-mogran: That sounds really good @mkzhx, and thanks for being so thorough with your investigation!

modules/tinymce/src/core/main/ts/delete/InlineFormatDelete.ts
mkzhx: Also tightened override condition for `refreshCaret` in `requiresRefreshCaretOverride`:
```
rng.collapsed && (rangeStartsAtTextContainer(rng) || editor.dom.isEmpty(rng.startContainer)) && !hasAncestorInlineCaretAtStart(editor);
```

modules/tinymce/src/core/main/ts/delete/InlineFormatDelete.ts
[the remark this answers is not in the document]
mkzhx: Added `getParentInlinesUntilMultichildInline` (used by `deleteCaret`) to differentiate from `getParentInlines` (used by `deleteRange`).

modules/tinymce/src/core/main/ts/delete/InlineFormatDelete.ts
mkzhx: Integrated tests in InlineFormatDeleteOverrideTest into InlineFormatDeleteTest and a new InlineFormatPostShortcutDeleteTest to increase specificity, make new tests consistent with existing ones, and not require exporting tests just for testing purposes.

ephox-mogran (approved):

modules/tinymce/src/core/main/ts/keyboard/DeleteBackspaceKeys.ts
```diff
@@ -50,24 +51,54 @@ const executeKeydownOverride = (editor: Editor, caret: Cell<Text | null>, evt: K
...
+    { keyCode: VK.BACKSPACE, altKey: true, action: MatchKeys.action(InlineFormatDelete.refreshCaret, editor) },
+    { keyCode: VK.DELETE, altKey: true, action: MatchKeys.action(InlineFormatDelete.refreshCaret, editor) }
+  ] : [
+    { keyCode: VK.BACKSPACE, ctrlKey: true, action: MatchKeys.action(InlineFormatDelete.refreshCaret, editor) },
+    { keyCode: VK.DELETE, ctrlKey: true, action: MatchKeys.action(InlineFormatDelete.refreshCaret, editor) }
+  ];
+
+  // macOS surpresses keyup events for most keys including Backspace when Meta key is engaged
+  // To emulate Meta + Backspace on macOS, add a pattern for the meta key when backspace was
+  // detected on keydown
```
TheSpyder, line 68: Do we have some research to back this up? I checked in Slack and nobody responded to your question, sorry about that.

(I also wrote this comment in December and forgot to submit my review, oops)

modules/tinymce/src/core/main/ts/keyboard/DeleteBackspaceKeys.ts
TheSpyder: `key` is the long-term solution, I think there's a ticket in the backlog about what we need to switch to it but it has probably been shelved by now.

modules/tinymce/src/core/main/ts/keyboard/DeleteBackspaceKeys.ts
mkzhx: I think Cmd + Backspace, like most other Cmd-modified keys, is transformed into a "key equivalent" on `keydown`, and so Cmd + Backspace does not fire a `keyup`. See "Handling Key Equivalents" in https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/EventOverview/HandlingKeyEvents/HandlingKeyEvents.html

mkzhx: @TheSpyder I think Cmd + Backspace, like most other Cmd-modified keys, is transformed into a "key equivalent" on `keydown`, and so Cmd + Backspace does not fire a `keyup`. See "Handling Key Equivalents" in https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/EventOverview/HandlingKeyEvents/HandlingKeyEvents.html

TheSpyder: well TIL, I guess. I didn't think keydown without keyup was possible, but I've confirmed in an [event tester](https://javascript.info/keyboard-events)

```
keydown key=Meta code=MetaLeft metaKey
--------------------------------------------------------------------------------
keydown key=Backspace code=Backspace metaKey
--------------------------------------------------------------------------------
keyup key=Meta code=MetaLeft
```

PR #8304, opened 2022-11-29
TINY-5167: Toolbar split buttons in advlist plugin to show the correct state when the cursor is in a checklist

shanmen-tiny:
Related Ticket: TINY-5167

Description of Changes:
* Toolbar split buttons in advlist plugin to show the correct state when the cursor is in a checklist

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/plugins/advlist/main/ts/core/ListUtils.ts
+++ b/modules/tinymce/src/plugins/advlist/main/ts/core/ListUtils.ts
@@ -36,8 +36,8 @@ const isWithinNonEditableList = (editor: Editor, element: Element | null): boole
 };
 
 export {
-  isTableCellNode,
-  isListNode,
+  isTableCellNode, // Exported for testing
+  isListNode, // Exported for testing
   inList,
   getSelectedStyleType,
   isWithinNonEditableList
--- a/modules/tinymce/src/plugins/advlist/main/ts/core/ListUtils.ts
+++ b/modules/tinymce/src/plugins/advlist/main/ts/core/ListUtils.ts
@@ -2,7 +2,7 @@ import { Arr, Optional, Type } from '@ephox/katamari';
 
 import Editor from 'tinymce/core/api/Editor';
 
-const isCustomList = (list: Element): boolean =>
+const isCustomList = (list: HTMLElement): boolean =>
   /\btox\-/.test(list.className);
 
 const isChildOfBody = (editor: Editor, elm: Node): boolean => {
@@ -16,9 +16,9 @@ const isListNode = matchNodeNames<HTMLOListElement | HTMLUListElement | HTMLDLis
 
 const isTableCellNode = matchNodeNames<HTMLTableHeaderCellElement | HTMLTableCellElement>(/^(TH|TD)$/);
 
-const inList = (editor: Editor, e: Node[], nodeName: string): boolean =>
-  Arr.findUntil(e, isListNode, isTableCellNode)
-    .exists((list) => list.nodeName === nodeName && !isCustomList(list) && isChildOfBody(editor, list));
+const inList = (editor: Editor, parents: Node[], nodeName: string): boolean =>
+  Arr.findUntil(parents, (parent) => isListNode(parent) && !isCustomList(parent), isTableCellNode)
+    .exists((list) => list.nodeName === nodeName && isChildOfBody(editor, list));
 
 const getSelectedStyleType = (editor: Editor): Optional<string> => {
   const listElm = editor.dom.getParent(editor.selection.getNode(), 'ol,ul');
--- a/modules/tinymce/src/plugins/advlist/main/ts/core/ListUtils.ts
+++ b/modules/tinymce/src/plugins/advlist/main/ts/core/ListUtils.ts
@@ -37,6 +37,7 @@ const isWithinNonEditableList = (editor: Editor, element: Element | null): boole
 
 export {
   isTableCellNode,
+  isListNode,
   inList,
   getSelectedStyleType,
   isWithinNonEditableList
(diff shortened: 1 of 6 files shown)


modules/tinymce/src/plugins/advlist/test/ts/browser/AdvlistOptionsAndToolbarTest.ts
```diff
@@ -77,158 +72,233 @@ describe('browser.tinymce.plugins.advlist.AdvlistOptionsAndToolbarTest', () => {
...
-      label: 'TINY-8721: Test not including default in advlist_number_styles option',
-      options: {
-        advlist_number_styles: 'lower-alpha,lower-greek'
+  const pAssertButtonToggledState = (name: string, state: boolean) =>
+    Waiter.pTryUntil('Wait for toolbar button state', () => {
+      const button = UiFinder.findIn(SugarBody.body(), `div.tox-split-button[aria-label="${name}"]`).getOrDie();
+      return Assertions.assertStructure('', ApproxStructure.build((s, _, __) => s.element('div', {
+        children: [
+          s.element('span', {
+            exactClasses: [ 'tox-tbtn', ...(state ? [ 'tox-tbtn--enabled' ] : [] ) ]
```
ephox-mogran, line 81: You can also use `arr.has` and `arr.not`, but I don't want to stand in the way of you using your new API :)

modules/tinymce/src/plugins/advlist/main/ts/core/ListUtils.ts
```diff
@@ -1,18 +1,24 @@
...
 
-const isListNode = (editor: Editor) => (node: Node | null): boolean => {
-  return Type.isNonNullable(node) && (/^(OL|UL|DL)$/).test(node.nodeName) && isChildOfBody(editor, node);
-};
+const isListNode = matchNodeNames<HTMLOListElement | HTMLUListElement | HTMLDListElement>(/^(OL|UL|DL)$/);
+
+const isTableCellNode = matchNodeNames<HTMLTableHeaderCellElement | HTMLTableCellElement>(/^(TH|TD)$/);
+
+const inList = (editor: Editor, e: Node[], nodeName: string): boolean =>
+  Arr.findUntil(e, isListNode, isTableCellNode)
```
ephox-mogran, line 20: This code is a bit complex, so an explanatory comment about why we stop if we find a table node or exactly why we are doing this might help the reader.

modules/tinymce/src/plugins/advlist/main/ts/core/ListUtils.ts
```diff
@@ -1,18 +1,24 @@
...
+    .exists((list) => list.nodeName === nodeName && !isCustomList(list) && isChildOfBody(editor, list));
```
ephox-mogran, line 21: Minor: should the `isCustomList` check be moved so that it's earlier? Should it make it fail the `Arr.findUntil` for example?

modules/tinymce/src/plugins/advlist/main/ts/core/ListUtils.ts
ephox-mogran: Ah having looked at the code in `Buttons` that this was adapted from, this might be a good case for keeping the parameter name as `parents` so that it's clearer that we are traversing *up* DOM structures rather than *across*.

ephox-mogran: Although I see that it's mostly the same code as the lists plugin, so there is value in keeping it consistent. Just things to think about in general, then. Lists called it `parents`  though, which supports that argument.

ephox-mogran (approved):
Looks good.

tiny-ben-tran (approved):
Is there tests which cover `isTableCellNode()`?

modules/tinymce/src/plugins/advlist/main/ts/core/ListUtils.ts
shanmen-tiny: Yeah, made a mistake, was passing the event initially. Changed it to passing the parents but didn't rename the parameter. Fixed, and thank you.

modules/tinymce/src/plugins/advlist/main/ts/core/ListUtils.ts
shanmen-tiny: Make sense to move the `isCustomList` earlier. Made the suggested changes, thanks.

modules/tinymce/src/plugins/advlist/test/ts/atomic/ListUtilsTest.ts
```diff
@@ -0,0 +1,22 @@
+import { describe, it } from '@ephox/bedrock-client';
+import { SugarElement } from '@ephox/sugar';
+import { assert } from 'chai';
+
+import { isTableCellNode, isListNode } from 'tinymce/plugins/advlist/core/ListUtils';
+
+describe('atomic.tinymce.plugins.advlist.ListUtilsTest', () => {
+  it('isTableCellNode', () => {
+    assert.isTrue(isTableCellNode(SugarElement.fromTag('td').dom), 'td');
```
spocke, line 9: Is atomic tests supposed to fiddle with the dom? Maybe it has a different meaning now days.

spocke (approved):

modules/tinymce/src/plugins/advlist/test/ts/atomic/ListUtilsTest.ts
shanmen-tiny: To be honest, I'm not sure if it's appropriate to put these in the atomic test. Happy to move it to the browser test and include testing the `inList`.

PR #8307, opened 2022-11-30
TINY-9428: Code Scanning

ephox-mogran:
Related Ticket: TINY-9428

Description of Changes:
* Setting up CodeQL to replace the soon-to-be-decommissioned LGTM

Pre-checks:
* [x] Changelog entry added
* [N/A] Tests have been added (if applicable)

Review:
* [X] Milestone set
* [N/A] Docs ticket created (if applicable)

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,42 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ "feature/**", "hotfix/**", "develop", "main" ]
+  pull_request:
+    branches: [ "feature/**", "hotfix/**", "develop", "main" ]
+## Add schedule if needed
+#  schedule:
+#    - cron: "* * * * 0"
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ javascript ]
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Initialize CodeQL
+        uses: github/codeql-action/init@v2
+        with:
+          languages: ${{ matrix.language }}
+          queries: +security-and-quality
+
+      - name: Autobuild
+        uses: github/codeql-action/autobuild@v2
+
+      - name: Perform CodeQL Analysis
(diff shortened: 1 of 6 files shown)


.github/workflows/codeql.yml
```diff
@@ -0,0 +1,42 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ "feature/**", "hotfix/**", "develop", "main" ]
```
ephox-mogran, line 5: I don't think this quite right. If someone has more experience, feel free to fix it. 

I haven't read this yet: https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-code-scanning#configuring-frequency

spocke: Why wouldn't it be correct? We can always tweak this later if we notice that we need to expand it.

spocke (approved):

.github/workflows/codeql.yml
jscasca: I did not know before that this rpeo doesn't have a `main` branch so instead obviously we have to target `master`. Other than that we can probably avoid `feature` and `hotfix` on push and only scan when making PRs? Additionally, definitely suggest to add `release` branches.

Over all, I would suggest:
```
on:
  push:
    branches: [ "release/**", "develop", "master" ]
  pull_request:
    branches: [ "release/**", "develop", "master" ]
 ```

jscasca (changes requested):
Target `master` branch instead of `main`, and `release` branches

.github/workflows/codeql.yml
spocke: So I guess we don't care about running this on feature branches? Should we then apply these patter to all repos running beehive?

ephox-mogran: Reading the docs, I *think* this is supposed to be the target branch of the PR, not the source branch. So any PRs that will merge into develop, master, release. So feature branches, theoretically, would be included because they are going to merge into develop.

spocke: Right I was thinking of long living ones where you have PRs targeting long living feature branches but we don't use that style of development that often and this will then run once you make the final PR and that is probably good enough.

TheSpyder (approved):

jscasca (approved):

TheSpyder:
Build failure isn't related to code scanning, so merging

PR #8336, opened 2022-12-08
TINY-9148: Add active element blurring before closing a form

yacodes:
Related Ticket: TINY-9148

Description of Changes:
* Blurs active element before closing a form.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


HAFRMO (approved):

modules/tinymce/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Checkmark did not show in menu colorswatches. #TINY-9395
 - Toolbar split buttons in advlist plugin to show the correct state when the cursor is in a checklist. #TINY-5167
 - Dragging transparent elements into transparent blocks elements could produce invalid nesting of transparents. #TINY-9231
+- Closing the dialog does not scroll to the sink in Safari anymore. #TINY-9148
```
spocke, line 14: I think the changelog items should be written how they where before not how they are now at least that is the style I've seen lately in the changelog if you compare to the other items. Also is this specific to the search dialog? And I think the sink is a bit to internal for anyone external to understand so maybe better to describe what the bug looks like for the user.

```suggestion
- Closing the search/replace dialog would scroll down the document in Safari. #TINY-9148
```

Or if it's generic for all dialogs:

```suggestion
- Closing a dialog would scroll down the document in Safari. #TINY-9148
```

Or something like that.

modules/tinymce/src/plugins/searchreplace/test/ts/browser/SearchReplaceDialogTest.ts
```diff
@@ -112,4 +114,14 @@ describe('browser.tinymce.plugins.searchreplace.SearchReplaceDialogTest', () =>
     findAndAssertFound(editor, 3);
     TinyUiActions.closeDialog(editor);
   });
+
+  it('TINY-9148: Scroll should remain in place after the dialog is closed', async () => {
```
spocke, line 118: Are we sure that this test works? Does it fail if you remove your fix. Since the fiddle scrolls the parent window pane not the iframe contents and the sink is in the parent document not the iframe document. So I would expect that the editor container would need say margin of 2000px or something to produce a scroll.

- [ ] Confirm that this test works if you remove the fix

spocke (approved):

modules/tinymce/src/plugins/searchreplace/test/ts/browser/SearchReplaceDialogTest.ts
yacodes: You're right. I was testing with the `autoresize` plugin on. Thanks for pointing it out!

PR #8339, opened 2022-12-13
DOC-1487: Add EditorUpload API documentation

MitchCTiny:
Related Ticket: DOC-1487

Description of Changes:
* Adds documentation for each property and method in the EditorUpload API

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/api/EditorUpload.ts
+++ b/modules/tinymce/src/core/main/ts/api/EditorUpload.ts
@@ -18,21 +18,18 @@ import { createUploader, openNotification } from './util/ImageUploader';
  *
  * @class tinymce.EditorUpload
  * @example
- * // Apply a filter to all images as uploaded..
+ * // Apply a new filter within the image scan.
  * tinymce.activeEditor.EditorUpload.addFilter((image) => {
  *   const maxSize = 1920 * 1080;
  *   const imageSize = image.width * image.height;
  *   return imageSize < maxSize;
  * });
  *
- * // UploadImages
+ * // Upload all valid images in the editor
  * tinymce.activeEditor.EditorUpload.uploadImages();
  *
- * // ScanForImages
+ * // Scan the editor for valid images
  * tinymce.activeEditor.EditorUpload.scanForImages();
- *
- * // Destroy
- * tinymce.activeEditor.EditorUpload.destroy();
  */
 
 export interface UploadResult {
@@ -53,8 +50,8 @@ interface EditorUpload {
   blobCache: BlobCache;
 
   /**
-   * Adds a custom filter to be executed before any images are uploaded to the server.
-   * Images must return true on every added filter to be uploaded, else they are filtered out.
+   * Adds a custom filter that controls the images which are included in the scan.
+   * Images must return true on every added filter to be considered valid.
    *
    * @method addFilter
    * @param {Function} filter Function which filters each image upload.
@@ -69,7 +66,7 @@ interface EditorUpload {
   addFilter: (filter: (img: HTMLImageElement) => boolean) => void;
 
   /**
-   * Uploads all data uri/blob uri images in the editor contents to server.
+   * Uploads all the data uri/blob uri images scanned from the editor content to the server.
    *
    * @method uploadImages
    * @return {Promise} Promise instance with images and status for each image.
@@ -77,7 +74,7 @@ interface EditorUpload {
   uploadImages: () => Promise<UploadResult[]>;
 
   /**
-   * Uploads all data uri/blob uri images in the editor contents to server only when automatic uploads are enabled.
+   * Uploads all data uri/blob uri images to the server only when automatic uploads are enabled.
    *
    * @method uploadImagesAuto
    * @return {Promise} Promise instance with images and status for each image.
@@ -85,7 +82,7 @@ interface EditorUpload {
   uploadImagesAuto: () => Promise<UploadResult[]>;
 
   /**
-   * Retrieves each valid image element in the editor, and blob information for each image.
+   * Scans the editor content for valid image elements and generates blob information for each image.
    *
    * @method scanForImages
    * @return {Promise} Promise instance with element object and blob information for each image.
@@ -93,7 +90,8 @@ interface EditorUpload {
   scanForImages: () => Promise<BlobInfoImagePair[]>;
 
   /**
-   * Resets the blob cache, uploads and image scanner.
+   * Resets the blob cache, uploads and image scanner. Called automatically on <code>editor.remove</code>.
+   * This method is not recommended to be used by integrators.
    *
    * @method destroy
    */
--- a/modules/tinymce/src/core/main/ts/api/EditorUpload.ts
+++ b/modules/tinymce/src/core/main/ts/api/EditorUpload.ts
@@ -90,8 +90,8 @@ interface EditorUpload {
   scanForImages: () => Promise<BlobInfoImagePair[]>;
 
   /**
-   * Resets the blob cache, uploads and image scanner. Called automatically on <code>editor.remove</code>.
-   * This method is not recommended to be used by integrators.
+   * Resets the blob data and upload status of all uploaded images. Called automatically on <code>editor.remove</code>.
+   * This method is not recommended for integration.
    *
    * @method destroy
    */
--- a/modules/tinymce/src/core/main/ts/api/EditorUpload.ts
+++ b/modules/tinymce/src/core/main/ts/api/EditorUpload.ts
@@ -13,10 +13,26 @@ import * as Options from './Options';
 import { createUploader, openNotification } from './util/ImageUploader';
 
 /**
+ * TinyMCE Editor Upload API
  * Handles image uploads, updates undo stack and patches over various internal functions.
  *
- * @private
  * @class tinymce.EditorUpload
+ * @example
+ * // Apply a filter to all images as uploaded..
+ * tinymce.activeEditor.EditorUpload.addFilter((image) => {
+ *   const maxSize = 1920 * 1080;
+ *   const imageSize = image.width * image.height;
+ *   return imageSize < maxSize;
+ * });
+ *
+ * // UploadImages
+ * tinymce.activeEditor.EditorUpload.uploadImages();
+ *
+ * // ScanForImages
+ * tinymce.activeEditor.EditorUpload.scanForImages();
+ *
+ * // Destroy
+ * tinymce.activeEditor.EditorUpload.destroy();
  */
 
 export interface UploadResult {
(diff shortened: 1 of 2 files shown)


modules/tinymce/src/core/main/ts/api/EditorUpload.ts
```diff
@@ -13,10 +13,26 @@ import * as Options from './Options';
...
+ *   return imageSize < maxSize;
+ * });
+ *
+ * // UploadImages
+ * tinymce.activeEditor.EditorUpload.uploadImages();
+ *
+ * // ScanForImages
+ * tinymce.activeEditor.EditorUpload.scanForImages();
+ *
+ * // Destroy
```
spocke, line 34: Destroy is not meant to be called by developers it's just an internal thing that just happens to be exposed. So maybe not add an example of that.

modules/tinymce/src/core/main/ts/api/EditorUpload.ts
```diff
@@ -28,11 +44,59 @@ export interface UploadResult {
...
+  /**
+   * Retrieves each valid image element in the editor, and blob information for each image.
+   *
+   * @method scanForImages
+   * @return {Promise} Promise instance with element object and blob information for each image.
+   */
   scanForImages: () => Promise<BlobInfoImagePair[]>;
+
+  /**
+   * Resets the blob cache, uploads and image scanner.
```
spocke, line 96: Maybe add a note that this function is called automatically on editor.remove and should generally not be called by integrators.

spocke (approved):

modules/tinymce/src/core/main/ts/api/EditorUpload.ts
```diff
@@ -13,10 +13,26 @@ import * as Options from './Options';
 import { createUploader, openNotification } from './util/ImageUploader';
 
 /**
+ * TinyMCE Editor Upload API
  * Handles image uploads, updates undo stack and patches over various internal functions.
  *
- * @private
  * @class tinymce.EditorUpload
+ * @example
+ * // Apply a filter to all images as uploaded..
```
brian-forte, line 21:
```suggestion
 * // Apply a filter to all images as they are uploaded.
```
1. Typo correction: s/.././
2. Suggested copy-edit: s/images as uploaded/images as they are uploaded/

modules/tinymce/src/core/main/ts/api/EditorUpload.ts
```diff
@@ -13,10 +13,26 @@ import * as Options from './Options';
...
+ *   const maxSize = 1920 * 1080;
+ *   const imageSize = image.width * image.height;
+ *   return imageSize < maxSize;
+ * });
+ *
+ * // UploadImages
+ * tinymce.activeEditor.EditorUpload.uploadImages();
+ *
+ * // ScanForImages
+ * tinymce.activeEditor.EditorUpload.scanForImages();
```
brian-forte, line 32:
```suggestion
 * // Upload images
 * tinymce.activeEditor.EditorUpload.uploadImages();
 *
 * // Scan for images
 * tinymce.activeEditor.EditorUpload.scanForImages();
```

If the associated comment is going to be just the API name, may as well edit said name for readability and legibility.

modules/tinymce/src/core/main/ts/api/EditorUpload.ts
```diff
@@ -28,11 +44,59 @@ export interface UploadResult {
...
+   * Cache of blob elements created in an editor instance.
+   *
+   * @property blobCache
+   * @type Object
+   */
   blobCache: BlobCache;
+
+  /**
+   * Adds a custom filter to be executed before any images are uploaded to the server.
+   * Images must return true on every added filter to be uploaded, else they are filtered out.
```
brian-forte, line 57:
```suggestion
   * Adds a custom filter that is executed before any images are uploaded to the server.
   * Images must return true on every added filter to be uploaded.
   * An image that returns false on any added filter is filtered out (ie, not uploaded).
```

1. Switched from the quasi–future-tense infinitive — _to be_ — to the present tense indicative — _is_.
2. Went all in on setting out what happens on returning either *true* and *false*.

NB: As it stands, both the original and edited text infer something that, if true, is probably worth making absolutely explicit.

They both infer that, if you have x images to upload, no uploading of any images begins until all x images run through all added filters.

For large values of x, I’m guessing that could result in a UX worth being explicit about.

If this is the case, I’d argue for something like the following:

```suggestion
   * Adds a custom filter that is executed before any images are uploaded to the server.
   * Images must return true on every added filter to be uploaded.
   * An image that returns false on any added filter is filtered out (ie, not uploaded).
   * NB: no images are uploaded until all added filters are executed over all supplied images.
```

OTOH, if the process is

1. run filters over image-x.
2. if true upload image-x.
3. if false do not upload image-x.
4. run filters over image-x+1
5. if true, upload image-x
6. if false, do not upload image-x+1.
7. repeat until image-x+n (where x+n = last image in the list).

that would be a different UX. And, if this latter is the case, I’d argue for something like the following:

```suggestion
   * Adds a custom filter that is executed before an image is uploaded to the server.
   * Images must return true on every added filter to be uploaded.
   * An image that returns false on any added filter is filtered out (ie, not uploaded).
   * NB: individual images are uploaded as soon as they return true to every added filter.
```

modules/tinymce/src/core/main/ts/api/EditorUpload.ts
```diff
@@ -28,11 +44,59 @@ export interface UploadResult {
...
+   * tinymce.activeEditor.EditorUpload.addFilter((image) => {
+   *   const maxSize = 1920 * 1080;
+   *   const imageSize = image.width * image.height;
+   *   return imageSize < maxSize;
+   * });
+   */
   addFilter: (filter: (img: HTMLImageElement) => boolean) => void;
+
+  /**
+   * Uploads all data uri/blob uri images in the editor contents to server.
```
brian-forte, line 72:
```suggestion
   * Upload all data uri/blob uri images in the editor contents to server.
```

Sticking to simple present transitive both for consistency’s sake and to keep the declarative approach clearer.

modules/tinymce/src/core/main/ts/api/EditorUpload.ts
```diff
@@ -28,11 +44,59 @@ export interface UploadResult {
...
+   *
+   * @method uploadImages
+   * @return {Promise} Promise instance with images and status for each image.
+   */
   uploadImages: () => Promise<UploadResult[]>;
+
+  /**
+   * Uploads all data uri/blob uri images in the editor contents to server only when automatic uploads are enabled.
```
brian-forte, line 80:
```suggestion
   * Upload all data uri/blob uri images in the editor contents to server only when automatic uploads are enabled.
```
Sticking to simple present transitive both for consistency’s sake and to keep the declarative approach clearer.

modules/tinymce/src/core/main/ts/api/EditorUpload.ts
```diff
@@ -28,11 +44,59 @@ export interface UploadResult {
...
+   *
+   * @method uploadImagesAuto
+   * @return {Promise} Promise instance with images and status for each image.
+   */
   uploadImagesAuto: () => Promise<UploadResult[]>;
+
+  /**
+   * Retrieves each valid image element in the editor, and blob information for each image.
```
brian-forte, line 88:
```suggestion
   * Retrieve each valid image element in the editor, and blob information for each image.
```
Sticking to simple present transitive both for consistency’s sake and to keep the declarative approach clearer.

modules/tinymce/src/core/main/ts/api/EditorUpload.ts
brian-forte, line 96:
```suggestion
   * Reset the blob cache, uploads and image scanner.
```
Sticking to simple present transitive both for consistency’s sake and to keep the declarative approach clearer.

modules/tinymce/src/core/main/ts/api/EditorUpload.ts
MitchCTiny: Apologies for the confusion, the filters are applied during the scanForImages method, which is called from uploadImages. I'll fix the wording now

modules/tinymce/src/core/main/ts/api/EditorUpload.ts
MitchCTiny: This is inconsistent with the style used in many other API documentation comments

modules/tinymce/src/core/main/ts/api/EditorUpload.ts
MitchCTiny: These should be more descriptive, I'll fix them

modules/tinymce/src/core/main/ts/api/EditorUpload.ts
MitchCTiny: Done, thanks

modules/tinymce/src/core/main/ts/api/EditorUpload.ts
MitchCTiny: Done

modules/tinymce/src/core/main/ts/api/EditorUpload.ts
MitchCTiny: Correcting with reference to scan, rather than upload

modules/tinymce/src/core/main/ts/api/EditorUpload.ts
```diff
@@ -28,11 +41,60 @@ export interface UploadResult {
...
+   * Scans the editor content for valid image elements and generates blob information for each image.
+   *
+   * @method scanForImages
+   * @return {Promise} Promise instance with element object and blob information for each image.
+   */
   scanForImages: () => Promise<BlobInfoImagePair[]>;
+
+  /**
+   * Resets the blob cache, uploads and image scanner. Called automatically on <code>editor.remove</code>.
+   * This method is not recommended to be used by integrators.
```
brian-forte, line 94:
```suggestion
   * Resets uploads, the image scanner, and the blob cache. Called automatically on <code>editor.remove</code>.
   * This method is not recommended for integrator use.
```
1. Changes the presentation order of items this method resets. To my ear, placing the nouns in definite form after the noun in indefinite form sounds better. (Also, _image scanner_ reads as absolutely needing to be in the definite form.)
2. Switching in a preposition for the infinitive present tense verb reads less awkwardly, at least to me.

MitchCTiny: Should make more sense now, higher level description.

brian-forte (approved):

ephox-mogran (approved):

modules/tinymce/src/core/main/ts/api/EditorUpload.ts
```diff
@@ -28,11 +41,60 @@ export interface UploadResult {
...
+   * Cache of blob elements created in an editor instance.
+   *
+   * @property blobCache
+   * @type Object
+   */
   blobCache: BlobCache;
+
+  /**
+   * Adds a custom filter that controls the images which are included in the scan.
+   * Images must return true on every added filter to be considered valid.
```
tiny-ben-tran, line 54:
```suggestion
   * Adds a custom filter that will be called all images being uploaded.
   * Return `true` for an image that is considered to be valid
```
Not sure how this would sound compared to the original.

modules/tinymce/src/core/main/ts/api/EditorUpload.ts
```diff
@@ -28,11 +41,60 @@ export interface UploadResult {
...
+   *
+   * @method addFilter
+   * @param {Function} filter Function which filters each image upload.
+   * @example
+   * // Filter which images are uploaded.
```
tiny-ben-tran, line 59:
```suggestion
   * // Filter images by their size
```

Ignore if this doesn't sound right

modules/tinymce/src/core/main/ts/api/EditorUpload.ts
```diff
@@ -28,11 +41,60 @@ export interface UploadResult {
...
+  /**
+   * Uploads all the data uri/blob uri images scanned from the editor content to the server.
+   *
+   * @method uploadImages
+   * @return {Promise} Promise instance with images and status for each image.
+   */
   uploadImages: () => Promise<UploadResult[]>;
+
+  /**
+   * Uploads all data uri/blob uri images to the server only when automatic uploads are enabled.
```
tiny-ben-tran, line 77: Is it called automatically as well? if the logic for automatic uploading images is elsewhere, then this function looks like a duplicate of `uploadImages`

tiny-ben-tran (approved):

modules/tinymce/src/core/main/ts/api/EditorUpload.ts
MitchCTiny: Filters being included in the scan rather than upon uploading is more accurate, as the scan is performed to find all images when upload is called.

modules/tinymce/src/core/main/ts/api/EditorUpload.ts
MitchCTiny: That's correct, I was following the style for other examples in APIs where the comment describes the API rather than the functionality in the example.

modules/tinymce/src/core/main/ts/api/EditorUpload.ts
MitchCTiny: When automatic uploads are enabled, this is called on `setContent`, so every time the content is changed. It is essentially a duplicate of `uploadImages` which only runs when automatic uploads are enabled, if you were to only describe the API and not it's use. In time I can see this API being removed as the conditional within the `editor.on('setContent')` call already checks if automatic uploads are enabled and could call `uploadImages` instead, and `uploadImagesAuto` should only be called then anyway.

PR #8348, opened 2022-12-16
TINY-9474: Visualchars applies to noneditable content

spocke:
Related Ticket: TINY-9474

Description of Changes:
* Prevents visual characters spans from being rendered within a noneditable context.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

[TINY-9474]: https://ephocks.atlassian.net/browse/TINY-9474?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/tinymce/src/plugins/visualchars/main/ts/core/Nodes.ts
+++ b/modules/tinymce/src/plugins/visualchars/main/ts/core/Nodes.ts
@@ -26,8 +26,8 @@ const isChildEditable = (node: SugarElement<Node>, currentState: boolean) => {
   return currentState;
 };
 
-// inlined sugars PredicateFilter.descendants for file size but also make it only act on editable nodes
-const filterDescendants = <T extends Node>(scope: SugarElement<Node>, predicate: (x: SugarElement<Node>) => x is SugarElement<T>, editable: boolean): SugarElement<T>[] => {
+// inlined sugars PredicateFilter.descendants for file size but also make it only act on editable nodes it changes the current editable state when it traveses down
+const filterEditableDescendants = <T extends Node>(scope: SugarElement<Node>, predicate: (x: SugarElement<Node>) => x is SugarElement<T>, editable: boolean): SugarElement<T>[] => {
   let result: SugarElement<T>[] = [];
   const dom = scope.dom;
   const children = Arr.map(dom.childNodes, SugarElement.fromDom);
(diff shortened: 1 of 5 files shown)


lorenzo-pomili (approved):

modules/tinymce/src/plugins/visualchars/main/ts/core/Nodes.ts
```diff
@@ -11,17 +11,32 @@ const isMatch = (n: SugarElement<Node>): n is SugarElement<Text> => {
...
+
+const isChildEditable = (node: SugarElement<Node>, currentState: boolean) => {
+  if (SugarNode.isHTMLElement(node)) {
+    const value = ContentEditable.getRaw(node);
+    if (value === 'true') {
+      return true;
+    } else if (value === 'false') {
+      return false;
+    }
+  }
```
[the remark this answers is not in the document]
spocke: The problem is that the isEditable does a closest check and since we are walking the dom as a tree level by level I only need to check if the state has changed when I traverse down. So it becomes a O(1) instead of a O(n) on the depth of the tree. The code refactor you suggest is smaller but does have potential performance implications.

modules/tinymce/src/plugins/visualchars/main/ts/core/Nodes.ts
```diff
@@ -11,17 +11,32 @@ const isMatch = (n: SugarElement<Node>): n is SugarElement<Text> => {
...
+
+  return currentState;
+};
+
+// inlined sugars PredicateFilter.descendants for file size but also make it only act on editable nodes
+const filterDescendants = <T extends Node>(scope: SugarElement<Node>, predicate: (x: SugarElement<Node>) => x is SugarElement<T>, editable: boolean): SugarElement<T>[] => {
```
ephox-mogran, line 30: It took me a while to process this, so a comment might be useful for the future. Is the basic idea 

* it's just normal PredicateFilter.descendants except 
** it skips over including any elements that are CEF, but doesn't automatically skip their children
** it skips over their children only if they don't have a contenteditable=true

So we can't just filter it afterwards, because we need to know the ancestry of each one, so instead we just manually skip anything that has not overridden a CEF ancestor by passing through the current "editable" state. Is that right?

ephox-mogran (approved):

modules/tinymce/src/plugins/visualchars/main/ts/core/Nodes.ts
spocke: I thought the test for this function and the `only act on editable nodes` would be enough. Yeah it checks for editable nodes by tracking the contenteditable state if it changes between true/false if it doesn't then it uses what ever state that is on the stack. It's only when it goes down or up the state change matters down it's done by reading what that node we are going down by has. Up it just pops the stack.

I renamed the function to `filterEditableDescendants` instead so that it more clear what the function is doing and I also added more details to the comment.

PR #8353, opened 2022-12-20
TINY-9268: Extend menu button and toolbar button APIs to update the menu button icon and toolbar button icon respectively

hamza0867:
Related Ticket: TINY-9268

Description of Changes:
* Extend menu button and toolbar button APIs to update the menu button icon and toolbar button icon respectively

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

[TINY-9268]: https://ephocks.atlassian.net/browse/TINY-9268?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


modules/tinymce/src/themes/silver/main/ts/ui/dropdown/CommonDropdown.ts
```diff
@@ -162,6 +162,11 @@ const renderCommonDropdown = <T>(
           onControlAttached(spec, editorOffCell),
           onControlDetached(spec, editorOffCell)
         ]),
+        AddEventsBehaviour.config('common-button-display-events', [
```
spocke, line 165: Should this be `menubutton-display-events` I think the name here is used for tooling.

modules/tinymce/src/themes/silver/main/ts/ui/toolbar/button/ToolbarButtons.ts
```diff
@@ -60,7 +62,15 @@ interface ChoiceFetcher {
 
 const getButtonApi = (component: AlloyComponent): Toolbar.ToolbarButtonInstanceApi => ({
   isEnabled: () => !Disabling.isDisabled(component),
-  setEnabled: (state: boolean) => Disabling.set(component, !state)
+  setEnabled: (state: boolean) => Disabling.set(component, !state),
+  setText: (text: string) => {
```
spocke, line 66: One here is a block the other one is not should the setText and setIcon be the same both just lambdas without block since they have just one single statement in them?

modules/tinymce/src/themes/silver/main/ts/ui/toolbar/button/ToolbarButtons.ts
```diff
@@ -120,9 +156,11 @@ const renderCommonStructure = (
         receiver.map((r) => Reflecting.config({
           channel: r,
           initialData: { icon, text },
-          renderComponents: (data: ButtonState, _state) => componentRenderPipeline([
-            data.icon.map((iconName) => renderIconFromPack(iconName, providersBackstage.icons)),
-            data.text.map((text) => renderLabel(text, ToolbarButtonClasses.Button, providersBackstage))
+          renderComponents: (_data: ButtonState, _state) => componentRenderPipeline([
+            // data.icon.map((iconName) => renderIconFromPack(iconName, providersBackstage.icons)),
```
spocke, line 160: Remove the old code here.

modules/tinymce/src/themes/silver/main/ts/ui/toolbar/button/ToolbarButtons.ts
```diff
@@ -87,30 +103,50 @@ const renderCommonStructure = (
...
+      [SystemEvents.attachedToDom()]: [ 'common-button-display-events', 'reflecting' ]
     },
 
     buttonBehaviours: Behaviour.derive(
       [
         DisablingConfigs.toolbarButton(providersBackstage.isDisabled),
         ReadOnly.receivingConfig(),
         AddEventsBehaviour.config('common-button-display-events', [
+          AlloyEvents.runOnAttached((comp, _se) => {
+            Css.set(comp.element, 'width', Css.get(comp.element, 'width') );
```
spocke, line 138: Maybe add a comment why this is required since we want to keep the width when `setText` is executed but that might not be obvious if you read this out of context. Or maybe describe it with code since this is used in a few places maybe it's better to have say a function that does `ToolbarUtils.forceInitialSize(comp)` or similar that does the code and also describes why it's there.

```suggestion
            Css.set(comp.element, 'width', Css.get(comp.element, 'width'));
```

modules/tinymce/src/themes/silver/main/ts/ui/dropdown/CommonDropdown.ts
```diff
@@ -162,6 +162,11 @@ const renderCommonDropdown = <T>(
...
+          AlloyEvents.runOnAttached((comp, _se) => {
+            Css.set(comp.element, 'width', Css.get(comp.element, 'width') );
```
spocke, line 167:
```suggestion
            Css.set(comp.element, 'width', Css.get(comp.element, 'width'));
```

spocke (approved):

modules/tinymce/src/themes/silver/main/ts/ui/toolbar/button/ToolbarButtons.ts
hamza0867: I am waiting for @ephox-mogran to review this code and see if this approach is the right approach or not. That is why I am keeping these comments until I get a review from him.

modules/tinymce/src/themes/silver/main/ts/ui/toolbar/button/ToolbarButtons.ts
hamza0867: Agreed. Done!

modules/tinymce/src/themes/silver/main/ts/ui/toolbar/button/ToolbarButtons.ts
hamza0867: Done. I used UiUtils since it is not used only in toolbar but also in dropdown.

modules/tinymce/src/themes/silver/main/ts/ui/toolbar/button/ToolbarButtons.ts
```diff
@@ -263,7 +301,16 @@ const renderSplitButton = (spec: Toolbar.ToolbarSplitButton, sharedBackstage: Ui
         comp.getSystem().getByDom(button).each((buttonComp) => Toggling.set(buttonComp, state));
       });
     },
-    isActive: () => SelectorFind.descendant(comp.element, 'span').exists((button) => comp.getSystem().getByDom(button).exists(Toggling.isOn))
+    isActive: () => SelectorFind.descendant(comp.element, 'span').exists((button) => comp.getSystem().getByDom(button).exists(Toggling.isOn)),
+    setText: (text: string) =>
+      AlloyTriggers.emitWith(comp.components()[0], updateMenuText, {
```
ephox-mogran, line 306: Try to exist using direct child relationships for identifying descendants - it makes everything too coupled to the existing DOM structure. Is `comp.components()[0]` the same as the `descendant('span')` above? Could a similar approach be used instead?

modules/tinymce/src/plugins/autosave/main/ts/ui/Buttons.ts
```diff
@@ -1,9 +1,9 @@
 import Editor from 'tinymce/core/api/Editor';
-import { Toolbar } from 'tinymce/core/api/ui/Ui';
+import { Menu } from 'tinymce/core/api/ui/Ui';
 
 import * as Storage from '../core/Storage';
 
-const makeSetupHandler = (editor: Editor) => (api: Toolbar.ToolbarButtonInstanceApi) => {
+const makeSetupHandler = (editor: Editor) => (api: Menu.MenuItemInstanceApi) => {
```
ephox-mogran, line 6: This change seems unexpected. `Menu.MenuItemInstanceApi` sounds like the API of a single menu item, whereas this is in Buttons. Is this really intended?

ephox-mogran (commented):
Seems like it's almost there, but there are a few questions I have. We might try and organise a session to go through them.

modules/tinymce/src/themes/silver/main/ts/ui/dropdown/CommonDropdown.ts
ephox-mogran: This also looks like a duplicate name with ToolbarButtons

modules/tinymce/src/themes/silver/main/ts/ui/toolbar/button/ToolbarButtons.ts
ephox-mogran: I think we can see two options here 

* if we still need to respond to the channel `r` broadcasts, then instead of being reflecting, this should be receiving, and it should respond to the message by emitting the custom events: `updateMenuText` and `updateMenuIcon` with the data passed as part of the message
* if we don't need to respond to the channel `r` broadcasts, then we should just remove this reflecting altogether.

Either way, what we have here isn't likely to be correct. If it receives a broadcast, it's going to reset the text and icon back to the original, and clobber any of the subsequent replacing calls, based on no information (it isn't using the information from the reflecting message)

modules/tinymce/src/themes/silver/main/ts/ui/toolbar/button/ToolbarButtons.ts
```diff
@@ -162,21 +197,24 @@ const renderCommonToolbarButton = <T>(spec: GeneralToolbarButton<T>, specialisat
...
+            }),
+            onControlAttached(specialisation, editorOffCell),
+            onControlDetached(specialisation, editorOffCell)
+          ]),
+          // Enable toolbar buttons by default
+          DisablingConfigs.toolbarButton(() => !spec.enabled || providersBackstage.isDisabled()),
+          ReadOnly.receivingConfig()
+        ].concat(specialisation.toolbarButtonBehaviours)
+      ),
+      ...structure.buttonBehaviours
```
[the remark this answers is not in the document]
ephox-mogran: Previously, these were the behaviours defined in buttonBehaviours
 
 * disabling 
 * receiving for read-only (if we switch to using receiving for updating menu text etc., we'll need to combine these configs)
 * mousedown handler for focusing the button
 * reflecting configuration if required 
 
 But this code path wasn't using them. It is now. So in addition to the new events you've defined, these behaviours will be added to the button. You need to ensure that that change is intentional. Was it a bug before that readonly wasn't respected, or that disabled wasn't expected etc. 
 
 Also, try to track down what is passing `receiver`, and see if the receiving / reflecting configs are still necessary.

modules/tinymce/src/themes/silver/main/ts/ui/toolbar/button/ToolbarButtons.ts
hamza0867: Agreed. Changed to use SelectorFind.descendant

vpyshnenko (approved):

modules/tinymce/src/themes/silver/main/ts/ui/toolbar/button/ToolbarButtons.ts
```diff
@@ -168,27 +186,30 @@ const renderFloatingToolbarButton = (spec: Toolbar.GroupToolbarButton, backstage
...
+            }),
+            onControlAttached(specialisation, editorOffCell),
+            onControlDetached(specialisation, editorOffCell)
+          ]),
+          // Enable toolbar buttons by default
+          DisablingConfigs.toolbarButton(() => !spec.enabled || providersBackstage.isDisabled()),
+          ReadOnly.receivingConfig()
+        ].concat(specialisation.toolbarButtonBehaviours)
+      ),
+      [commonButtonDisplayEvent]: structure.buttonBehaviours?.[commonButtonDisplayEvent],
```
TheSpyder, line 211: This is a red flag for me - behaviours should always go inside the `derive` call. Not doing that means there's probably a bug lurking somewhere.

modules/tinymce/src/themes/silver/test/ts/browser/editor/toolbar/ToolbarDrawerToggleTest.ts
```diff
@@ -37,30 +38,33 @@ describe('browser.tinymce.themes.silver.editor.toolbar.ToolbarDrawerToggleTest',
   };
 
   context(`Using the 'ToggleToolbarDrawer' command should toggle the toolbar if applicable`, () => {
-    Arr.each<{ mode: ToolbarMode; shouldToggle: boolean }>([
-      { mode: 'floating', shouldToggle: true },
-      { mode: 'sliding', shouldToggle: true },
-      { mode: 'wrap', shouldToggle: false },
-      { mode: 'scrolling', shouldToggle: false }
-    ], (test) => {
+    const isWindowsAndEdge = Env.os.isWindows() && Env.browser.isChromiumEdge();
```
TheSpyder, line 41: What's going on here? Why do we suddenly need to know the difference between Chrome and Edge?

modules/tinymce/src/themes/silver/test/ts/module/ToolbarUtils.ts
```diff
@@ -30,7 +30,7 @@ const pAssertFloatingToolbarHeight = async (editor: Editor, expectedHeight: numb
 
 const pOpenFloatingToolbarAndAssertPosition = async (editor: Editor, getTop: () => number, pActions?: () => Promise<void>): Promise<void> => {
   await pOpenMore(ToolbarMode.floating);
-  await pAssertFloatingToolbarPosition(editor, getTop, 105, 531);
+  await pAssertFloatingToolbarPosition(editor, getTop, 105, 520);
```
TheSpyder, line 33: This doesn't seem related to the feature you added 🤔

TheSpyder (changes requested):

modules/tinymce/src/themes/silver/test/ts/browser/editor/toolbar/ToolbarDrawerToggleTest.ts
spocke: We discussed a weird flake that is only happening on Edge. So trying to skip it for now it's unrelated to the test. Not sure what to do with these flakes it takes lots of time to investigate them. I would probably not be something a user will run into.

modules/tinymce/src/themes/silver/main/ts/ui/toolbar/button/ToolbarButtons.ts
```diff
@@ -275,7 +294,21 @@ const renderSplitButton = (spec: Toolbar.ToolbarSplitButton, sharedBackstage: Ui
         comp.getSystem().getByDom(button).each((buttonComp) => Toggling.set(buttonComp, state));
       });
     },
-    isActive: () => SelectorFind.descendant(comp.element, 'span').exists((button) => comp.getSystem().getByDom(button).exists(Toggling.isOn))
+    isActive: () => SelectorFind.descendant(comp.element, 'span').exists((button) => comp.getSystem().getByDom(button).exists(Toggling.isOn)),
+    setText: (text: string) =>
+      SelectorFind.descendant(comp.element, 'span').each((button) =>
```
spocke, line 299: I wonder if this could just be emitted on the `comp` instead of the `buttonComp` the event broadcast should be sent down to all the descendants. Or do you think there is a risk that something on the actual comp would also listen to the same broadcasts?

spocke (approved):

modules/tinymce/src/themes/silver/main/ts/ui/toolbar/button/ToolbarButtons.ts
hamza0867: Unfortunately that doesn't work, I know it because a test case fails, but I don't know exactly why it doesn't work.

modules/tinymce/src/themes/silver/test/ts/module/ToolbarUtils.ts
hamza0867: Indeed it seems I needed to make this change to pass some tests locally and forgot to clean it up. I just cleaned it up in a new commit

modules/tinymce/src/themes/silver/main/ts/ui/toolbar/button/ToolbarButtons.ts
TheSpyder: That wasn't clear - please add a comment about why we're doing it this way, because (as far as I can tell) it's different to every other behaviour specification we have.

TheSpyder (approved):

modules/tinymce/src/themes/silver/main/ts/ui/toolbar/button/ToolbarButtons.ts
hamza0867: Done!

TheSpyder:
I'm pretty sure the failing test is not caused by this PR, develop is passing so let's see what happens

PR #8371, opened 2023-01-05
TINY-9277: Highlight the content area if the editor is focused

tiny-ben-tran:
Related Ticket: TINY-9277

Description of Changes:
* Add an option `highlight_on_focus` which supports highlighting the content area if the editor is focused
* Add new variables for the colour and border width so they can be overridden in the custom skin
* Also support dark mode

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

[TINY-9277]: https://ephocks.atlassian.net/browse/TINY-9277?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/oxide/src/less/theme/components/edit-area/edit-area.less
+++ b/modules/oxide/src/less/theme/components/edit-area/edit-area.less
@@ -16,15 +16,11 @@
     &::before {
       border: @edit-area-border-width solid @edit-area-border-color;
       border-radius: 4px;
-      bottom: 0;
       content: '';
-      inset: 2px;
-      left: 0;
+      inset: 0;
       opacity: 0;
       pointer-events: none;
       position: absolute;
-      right: 0;
-      top: 0;
       transition: opacity .15s;
       z-index: 1;
     }
--- a/modules/oxide/src/less/theme/components/edit-area/edit-area.less
+++ b/modules/oxide/src/less/theme/components/edit-area/edit-area.less
@@ -16,11 +16,15 @@
     &::before {
       border: @edit-area-border-width solid @edit-area-border-color;
       border-radius: 4px;
+      bottom: 0;
       content: '';
       inset: 2px;
+      left: 0;
       opacity: 0;
       pointer-events: none;
       position: absolute;
+      right: 0;
+      top: 0;
       transition: opacity .15s;
       z-index: 1;
     }
--- a/modules/oxide/src/less/theme/components/edit-area/edit-area.less
+++ b/modules/oxide/src/less/theme/components/edit-area/edit-area.less
@@ -4,6 +4,7 @@
 
 @edit-area-iframe-background-color: @color-white;
 @edit-area-border: 0;
+@edit-area-border-color: #2D6ADF;
 
 .tox {
   .tox-edit-area {
@@ -11,6 +12,17 @@
     flex: 1;
     overflow: hidden;
     position: relative; // Required for the iframe to resize in IE and Edge
+    &::before {
+      border: 2px solid @edit-area-border-color;
+      border-radius: 4px;
+      content: '';
+      inset: 2px;
+      opacity: 0;
+      pointer-events: none;
+      position: absolute;
+      transition: opacity .15s;
+      z-index: 10;
+    }
   }
 
   .tox-edit-area__iframe {
@@ -22,4 +34,8 @@
     position: absolute; // Required for the iframe to resize in IE and Edge
     width: 100%;
   }
+}
+
+.tox.tox-edit-focus .tox-edit-area::before {
+  opacity: 1;
 }
--- a/modules/oxide/src/less/theme/components/edit-area/edit-area.less
+++ b/modules/oxide/src/less/theme/components/edit-area/edit-area.less
@@ -5,6 +5,7 @@
 @edit-area-iframe-background-color: @color-white;
 @edit-area-border: 0;
 @edit-area-border-color: #2D6ADF;
+@edit-area-border-width: 2px;
 
 .tox {
   .tox-edit-area {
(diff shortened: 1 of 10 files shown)


modules/tinymce/src/core/test/ts/webdriver/keyboard/IframeTabfocusTest.ts
```diff
@@ -1,37 +1,74 @@
...
+      if (inputElem) {
+        inputElem.parentNode?.removeChild(inputElem);
+      }
+    });
+
+    it('TINY-8315: Add an input field outside the editor, focus on the editor, press the tab key and assert focus shifts to the input field', async () => {
+      FocusTools.isOnSelector('iframe is focused', SugarDocument.getDocument(), 'iframe');
+      await RealKeys.pSendKeysOn('iframe => body', [ RealKeys.text('\t') ]);
+      await FocusTools.pTryOnSelector('Wait for focus to be on input', SugarDocument.getDocument(), '#inputElem');
+    });
```
tiny-ben-tran, line 38: Moved the previous test into a context

modules/tinymce/src/core/test/ts/webdriver/keyboard/IframeTabfocusTest.ts
```diff
@@ -1,37 +1,74 @@
...
   });
 
-  it('TINY-8315: Add an input field outside the editor, focus on the editor, press the tab key and assert focus shifts to the input field', async () => {
-    FocusTools.isOnSelector('iframe is focused', SugarDocument.getDocument(), 'iframe');
-    await RealKeys.pSendKeysOn('iframe => body', [ RealKeys.text('\t') ]);
-    await FocusTools.pTryOnSelector('Wait for focus to be on input', SugarDocument.getDocument(), '#inputElem');
+  context('Highlight editor content area on focus', () => {
```
spocke, line 41: This could be tested without using webdriver since the focus border should be available on api calls like editor.focus() also it should also test the default that the does not have a focus border when the option is not specified.

I think we should only use webdriver tests when really necessary. So I would make a separate test for just the api focus tests. I guess we could still keep this one so make sure that keyboard focus does the right thing as well.

modules/tinymce/src/core/test/ts/webdriver/keyboard/IframeTabfocusTest.ts
```diff
@@ -1,37 +1,74 @@
...
+    const asssertIsHighlighted = (editor: Editor) => assert.isTrue(Class.has(TinyDom.container(editor), 'tox-edit-focus'), 'Editor container should has tox-edit-focus class');
```
spocke, line 42:
```suggestion
    const assertIsHighlighted = (editor: Editor) => assert.isTrue(Class.has(TinyDom.container(editor), 'tox-edit-focus'), 'Editor container should has tox-edit-focus class');
```

modules/tinymce/src/core/test/ts/webdriver/keyboard/IframeTabfocusTest.ts
```diff
@@ -1,37 +1,74 @@
...
+    const asssertIsNotHighlighted = (editor: Editor) => assert.isFalse(Class.has(TinyDom.container(editor), 'tox-edit-focus'), 'Editor container should not has tox-edit-focus class');
```
spocke, line 43:
```suggestion
    const assertIsNotHighlighted = (editor: Editor) => assert.isFalse(Class.has(TinyDom.container(editor), 'tox-edit-focus'), 'Editor container should not has tox-edit-focus class');
```

modules/oxide/src/less/theme/components/edit-area/edit-area.less
```diff
@@ -4,13 +4,25 @@
...
+@edit-area-border-color: #2D6ADF;
 
 .tox {
   .tox-edit-area {
     display: flex;
     flex: 1;
     overflow: hidden;
     position: relative; // Required for the iframe to resize in IE and Edge
+    &::before {
+      border: 2px solid @edit-area-border-color;
```
spocke, line 16: I think these pixel values needs to be variables. Say you want to decrease the border with to 1px this is not currently possible with variable overrides.

modules/oxide/src/less/theme/components/edit-area/edit-area.less
```diff
@@ -4,13 +4,25 @@
...
+      border-radius: 4px;
+      content: '';
+      inset: 2px;
+      opacity: 0;
+      pointer-events: none;
+      position: absolute;
+      transition: opacity .15s;
+      z-index: 10;
```
spocke, line 24: Why was the z-index needed here?

tiny-ben-tran: It makes the border visible by bringing the pseudo element forward. The value can be 1 though

modules/oxide/src/less/theme/components/edit-area/edit-area.less
tiny-ben-tran: I added a variable for the border width.

modules/tinymce/src/core/test/ts/webdriver/keyboard/IframeTabfocusTest.ts
tiny-ben-tran: I added API tests for the new option

shanmen-tiny (approved):

TheSpyder (approved):

PR #8386, opened 2023-01-12
TINY-7993: Fix selection after inserting a link

yacodes:
Related Ticket: TINY-7993

Description of Changes:
* Fixing selection right after inserting a link, so the context menu and dialog are behaving correctly straight away. The solution is a bit hacky, bit simple. Left a descriptive comment as well.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -40,7 +40,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Clicking on a disabled split button will no longer call the `onAction` callback. #TINY-9504
 - Selection after inserting a link is now within the link so the context menu and dialog behave correctly straight away. #TINY-7993
 - Fix determining if the current selection contains a link and retrieving its href value. #TINY-7993
-- The Edit Link dialog incorrectly retrieved the URL value when opened immediately after the link insertion. #TINY-7993
+- The "Edit Link" dialog incorrectly retrieved the URL value when opened immediately after the link insertion. #TINY-7993
 
 ## 6.3.1 - 2022-12-06
 
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -39,6 +39,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Inline boundary was rendered for noneditable inline boundary elements. #TINY-9471
 - Clicking on a disabled split button will no longer call the `onAction` callback. #TINY-9504
 - Selection after inserting a link is now within the link so the context menu and dialog behave correctly straight away. #TINY-7993
+- Fix determining if the current selection contains a link and retrieving its href value. #TINY-7993
 
 ## 6.3.1 - 2022-12-06
 
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -40,6 +40,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Clicking on a disabled split button will no longer call the `onAction` callback. #TINY-9504
 - Selection after inserting a link is now within the link so the context menu and dialog behave correctly straight away. #TINY-7993
 - Fix determining if the current selection contains a link and retrieving its href value. #TINY-7993
+- The Edit Link dialog incorrectly retrieved the URL value when opened immediately after the link insertion. #TINY-7993
 
 ## 6.3.1 - 2022-12-06
 
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -38,8 +38,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - It was possible to open links inside the editor if the editor root was noneditable. #TINY-9470
 - Inline boundary was rendered for noneditable inline boundary elements. #TINY-9471
 - Clicking on a disabled split button will no longer call the `onAction` callback. #TINY-9504
-- Selection after inserting a link is now within the link so the context menu and dialog behave correctly straight away. #TINY-7993
-- Fix determining if the current selection contains a link and retrieving its href value. #TINY-7993
 - The "Edit Link" dialog incorrectly retrieved the URL value when opened immediately after the link insertion. #TINY-7993
 
 ## 6.3.1 - 2022-12-06
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -38,6 +38,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - It was possible to open links inside the editor if the editor root was noneditable. #TINY-9470
 - Inline boundary was rendered for noneditable inline boundary elements. #TINY-9471
 - Clicking on a disabled split button will no longer call the `onAction` callback. #TINY-9504
+- Selection after inserting a link is now within the link so the context menu and dialog behave correctly straight away. #TINY-7993
 
 ## 6.3.1 - 2022-12-06
 
--- a/modules/tinymce/src/core/main/ts/api/commands/LinkCommands.ts
+++ b/modules/tinymce/src/core/main/ts/api/commands/LinkCommands.ts
@@ -8,14 +8,6 @@ export const registerCommands = (editor: Editor): void => {
     const anchor = editor.dom.getParent(editor.selection.getNode(), 'a');
 
     if (Type.isObject(linkDetails) && Type.isString(linkDetails.href)) {
-      /**
-       * TINY-7993: Tiny's formatter sets an undesirable selection after applying a link.
-       * The caret is positioned right after the link, so `editor.selection.getNode()` returns a paragraph instead of the link.
-       * Consequently, the context menu and the link dialog do not work properly.
-       * We're using a bookmark here to restore the desirable selection.
-       */
-      const bookmark = editor.selection.getBookmark();
-
       // Spaces are never valid in URLs and it's a very common mistake for people to make so we fix it here.
       linkDetails.href = linkDetails.href.replace(/ /g, '%20');
 
@@ -28,8 +20,6 @@ export const registerCommands = (editor: Editor): void => {
       if (linkDetails.href) {
         editor.formatter.apply('link', linkDetails, anchor);
       }
-
-      editor.selection.moveToBookmark(bookmark);
     }
   };
 
--- a/modules/tinymce/src/core/main/ts/api/commands/LinkCommands.ts
+++ b/modules/tinymce/src/core/main/ts/api/commands/LinkCommands.ts
@@ -8,6 +8,14 @@ export const registerCommands = (editor: Editor): void => {
     const anchor = editor.dom.getParent(editor.selection.getNode(), 'a');
 
     if (Type.isObject(linkDetails) && Type.isString(linkDetails.href)) {
+      /**
+       * TINY-7993: Tiny's formatter sets an undesirable selection after applying a link.
+       * The caret is positioned right after the link, so `editor.selection.getNode()` returns a paragraph instead of the link.
+       * Consequently, the context menu and the link dialog do not work properly.
+       * We're using a bookmark here to restore the desirable selection.
+       */
+      const bookmark = editor.selection.getBookmark();
+
       // Spaces are never valid in URLs and it's a very common mistake for people to make so we fix it here.
       linkDetails.href = linkDetails.href.replace(/ /g, '%20');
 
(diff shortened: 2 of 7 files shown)


modules/tinymce/src/core/main/ts/api/commands/LinkCommands.ts
```diff
@@ -8,6 +8,14 @@ export const registerCommands = (editor: Editor): void => {
     const anchor = editor.dom.getParent(editor.selection.getNode(), 'a');
 
     if (Type.isObject(linkDetails) && Type.isString(linkDetails.href)) {
+      /**
+       * TINY-7993: Tiny's formatter sets an undesirable selection after applying a link.
+       * The caret is positioned right after the link, so `editor.selection.getNode()` returns a paragraph instead of the link.
+       * Consequently, the context menu and the link dialog do not work properly.
+       * We're using a bookmark here to restore the desirable selection.
+       */
+      const bookmark = editor.selection.getBookmark();
```
TheSpyder, line 17: This is definitely the wrong approach. If you look at the source JIRA comments, a simple selection collapse should suffice rather than using bookmarks - that's also the workaround used in the support case.

And depending on who you ask, customers inserting a link might not want the cursor to be in the link after closing the dialog (for example if they continue typing they probably don't want it to be inside the link).

TheSpyder (changes requested):

modules/tinymce/src/core/main/ts/api/commands/LinkCommands.ts
vpyshnenko: I found different approaches of how selection should behave after hyperlink insertion. For example, Jira moves collapsed selection after hyperlink, but CK editor lefts the entire hyperlink selected. It looks like TinyMCE choose the way similar to CK editor. So this PR tries to fix intended functionality but not to change it to the alternative approach.

yacodes: > This is definitely the wrong approach. If you look at the source JIRA comments, a simple selection collapse should suffice rather than using bookmarks - that's also the workaround used in the support case.

Unless I'm misunderstanding something, the function `applyLinkToSelection` operates only on the selection here. So bookmarking makes perfect sense here, no? 

> And depending on who you ask, customers inserting a link might not want the cursor to be in the link after closing the dialog (for example if they continue typing they probably don't want it to be inside the link).

As implied above, I believe that `applyLinkToSelection` works only within the selection. If a user inserts a link while writing text the link won't be selected, so one can proceed the writings. I've tested that manually and it's like so now.



> So this PR tries to fix intended functionality but not to change it to the alternative approach.

Viktor is 100% right here. We are not planning to change the current behavior, just fix the issue with the dialog and context menu.

modules/tinymce/src/plugins/link/test/ts/browser/LinkContextMenuTest.ts
```diff
@@ -18,14 +18,32 @@ describe('browser.tinymce.plugins.link.LinkContextMenuTest', () => {
...
+    const editor = hook.editor();
+    editor.setContent('<p><a href="http://tiny.cloud">tiny</a></p>');
+    await TinyUiActions.pTriggerContextMenu(editor, 'a[href="http://tiny.cloud"]', '.tox-silver-sink [role="menuitem"]');
+    await pTestContextMenuItems(editor);
+  });
+
+  it('TINY-7993: Context menu shows correctly right after inserting a link', async () => {
+    const editor = hook.editor();
+    editor.setContent('<p>aaa bbb ccc</p>');
+    TinySelections.setSelection(editor, [ 0, 0 ], 'aaa '.length, [ 0, 0 ], 'aaa bbb'.length);
```
vpyshnenko, line 39: If you try this case against the `aaa` part or `ccc` part the issue is still there - opened again link dialog still misses the url field value

modules/tinymce/src/plugins/link/test/ts/browser/DialogOpenTest.ts
```diff
@@ -69,18 +69,69 @@ describe('browser.tinymce.plugins.link.DialogOpenTest', () => {
...
-  });
+  const testDialogOnSelection = (
+    startPath: number[],
+    soffset: number,
+    finishPath: number[],
+    foffset: number,
+    result: string,
+    message: string
+  ) => {
+    it(`TINY-7993: Dialog shows correctly right after inserting a link: ${message}`, async () => {
```
vpyshnenko, line 80: Don't put `it(...)` call in the function body. It is hard to track down what test exactly failed. Instead, wrap every `testDialgOnSelection` call within `it()`

modules/tinymce/src/plugins/link/test/ts/browser/LinkContextMenuTest.ts
yacodes: Re-implemented the solution and the tests, so this is obsolete now, I believe.

modules/tinymce/src/plugins/link/test/ts/browser/DialogOpenTest.ts
yacodes: But it has a distinct message for each `it` block. It should be quite obvious which test has failed, no?

yacodes:
It was a bit naive to thinks that bookmarking will solve the issue. Re-implemented the solution by updating the way the current anchor element is determined within the dialog (`Utils.getAnchorElement`) as well as context menu and toolbar (`Actions.toggleEnabledState`).

Also wrote several test suites with different selections:
- selection at the beginning of the paragraph;
- selection in the middle of the paragraph;
- selection at the end of the paragraph;
- selection across several paragraphs.

Looking forward to hearing some comments!

modules/tinymce/src/plugins/link/test/ts/browser/DialogOpenTest.ts
vpyshnenko: I meant exploring the error stack  on line numbers in module

modules/tinymce/src/plugins/link/test/ts/browser/DialogOpenTest.ts
```diff
@@ -66,4 +68,76 @@ describe('browser.tinymce.plugins.link.DialogOpenTest', () => {
...
+  const pTestDialogOnSelection = async (
+    startPath: number[],
+    soffset: number,
+    finishPath: number[],
+    foffset: number,
+    result: string
+  ) => {
+    const editor = hook.editor();
+    editor.setContent('<p>aaa bbb ccc</p><p>ddd</p>');
+    TinySelections.setSelection.apply(null, [ editor, startPath, soffset, finishPath, foffset ]);
```
spocke, line 81: I assume that this is part of some refactor change that was made but not fully restored?

```suggestion
    TinySelections.setSelection(editor, startPath, soffset, finishPath, foffset);
```

modules/tinymce/src/plugins/link/main/ts/core/Actions.ts
```diff
@@ -70,7 +70,12 @@ const toggleActiveState = (editor: Editor) => (api: Toolbar.ToolbarToggleButtonI
 };
 
 const toggleEnabledState = (editor: Editor) => (api: Toolbar.ToolbarButtonInstanceApi | Menu.MenuItemInstanceApi): () => void => {
-  const updateState = () => api.setEnabled(Utils.isInAnchor(editor, editor.selection.getNode()));
+  const updateState = () => {
```
spocke, line 73: I wonder if the isInAnchor should be changed here instead so that it checks for this extra case. The original code here what pretty simple just a predicate that checks that state then sets the enabled api. If there is a risk of breaking other things maybe it could at least be a local predicate in this module or something so that the updateState function still is this simple one liner.

spocke (changes requested):

modules/tinymce/src/plugins/link/test/ts/browser/LinkContextMenuTest.ts
```diff
@@ -18,14 +18,32 @@ describe('browser.tinymce.plugins.link.LinkContextMenuTest', () => {
...
+    editor.execCommand('mceLink');
+    await TinyUiActions.pWaitForDialog(editor);
+    FocusTools.setActiveValue(SugarDocument.getDocument(), 'http://tiny.cloud');
+    TinyUiActions.submitDialog(editor);
+    await Waiter.pTryUntil('Wait for content to change', () =>
+      TinyAssertions.assertContent(editor, '<p>aaa <a href="http://tiny.cloud">bbb</a> ccc</p>'));
+    await TinyUiActions.pTriggerContextMenu(editor, 'a[href="http://tiny.cloud"]', '.tox-silver-sink [role="menuitem"]');
+    await pTestContextMenuItems(editor);
```
vpyshnenko, line 47: also, you need to assure that `Open Link` context menu item is not disabled

modules/tinymce/CHANGELOG.md
```diff
@@ -32,6 +32,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Removed a workaround for ensuring stylesheets are loaded in an outdated version of webkit. #TINY-9433
 - Lists in a noneditable root were incorrectly editable using list API commands, toolbar buttons and menu items. #TINY-9458
 - Color picker dialog would not update the preview color if the hex input value was prefixed with `#` symbol. #TINY-9457
+- Fix determining if the current selection contains a link and retrieving its href value. #TINY-7993
```
vpyshnenko, line 35: As you can see in the entries for the Fixed section we avoid using redundant 'Fix' word. Instead, we explain what the bug was.
```suggestion
- The Edit Link dialog incorrectly retrieved the URL value when opened immediately after the link insertion. #TINY-7993
```

modules/tinymce/src/plugins/link/test/ts/browser/DialogOpenTest.ts
yacodes: Yeah, forgot to change this. Thanks for pointing it out! Fixed [here](https://github.com/tinymce/tinymce/pull/8386/commits/86b37c56b64208ea0065ad8183fb75a356241919).

modules/tinymce/CHANGELOG.md
yacodes: Thanks for suggesting this! Fixed [here](https://github.com/tinymce/tinymce/pull/8386/commits/5ffe360f5139ff1f7e80faaa952457b4af5b8660).

modules/tinymce/src/plugins/link/main/ts/core/Actions.ts
yacodes: `Utils.isInAnchor` is used in a several places like `toggleActiveState` above and in `ui/Controls` as well. I'm afraid that moving the selection logic there will bloat it, since we need this logic only within `toggleEnabledState` function. It's a quite specific case and shouldn't be generalized I believe.

modules/tinymce/src/plugins/link/test/ts/browser/LinkContextMenuTest.ts
yacodes: You are right, thanks! Added the test [here](https://github.com/tinymce/tinymce/pull/8386/commits/1dffd34dde59c4fd60630aadb5dbcbdb303ed13d).

vpyshnenko (approved):

modules/tinymce/src/plugins/link/main/ts/core/Actions.ts
spocke: Ok maybe just pull it out in a local function then. That function name could then describe the thing it's doing in the ternary. Like `hasLinksInSelection` or something?

spocke (approved):

modules/tinymce/CHANGELOG.md
```diff
@@ -32,6 +32,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Removed a workaround for ensuring stylesheets are loaded in an outdated version of webkit. #TINY-9433
 - Lists in a noneditable root were incorrectly editable using list API commands, toolbar buttons and menu items. #TINY-9458
 - Color picker dialog would not update the preview color if the hex input value was prefixed with `#` symbol. #TINY-9457
+- The Edit Link dialog incorrectly retrieved the URL value when opened immediately after the link insertion. #TINY-7993
```
spocke, line 35: Maybe add ticks or quote around `Edit Link` not sure what the process is there. Maybe look for examples in the changelog.

yacodes: Looked through the CHANGELOG and it varies from italic to using quotes. Feels like the quotes is the way to go. Fixed.

modules/tinymce/src/plugins/link/main/ts/core/Actions.ts
yacodes: Extracted all the code to the `hasExactlyOneLinkInSelection` function. The name is a bit verbose, but it is exactly what it does.

TheSpyder (approved):
That's a much better solution 👍

PR #8391, opened 2023-01-16
TINY-9516: InlineFormatDelete element with multiple children selection test fails on latest Firefox v108.0.2

mkzhx:
Related Ticket: TINY-9516

Description of Changes:
* Updated structure and selection assertion to match Firefox's updated `contenteditable` behaviour on v108.0.2

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/test/ts/browser/delete/InlineFormatDeleteTest.ts
+++ b/modules/tinymce/src/core/test/ts/browser/delete/InlineFormatDeleteTest.ts
@@ -868,15 +868,7 @@ describe('browser.tinymce.core.delete.InlineFormatDelete', () => {
                         style: str.is('text-decoration: underline;')
                       },
                       children: [
-                        s.element('em', {
-                          children: [
-                            s.element('strong', {
-                              children: [
-                                s.element('br', {})
-                              ]
-                            })
-                          ]
-                        })
+                        s.element('br', {})
                       ]
                     })
                   ]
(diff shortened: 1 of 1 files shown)


spocke (approved):

TheSpyder:
Remember to add team members otherwise only spocke and I will see it

tiny-ben-tran (approved):

shanmen-tiny (approved):

modules/tinymce/src/core/test/ts/browser/delete/InlineFormatDeleteTest.ts
```diff
@@ -904,7 +896,8 @@ describe('browser.tinymce.core.delete.InlineFormatDelete', () => {
           });
         })
       );
-      TinyAssertions.assertCursor(editor, [ 0, 0, 0, 0 ], 0);
+      const selPath = browser.isFirefox() ? [ 0, 0 ] : [ 0, 0, 0, 0 ];
```
TheSpyder, line 899: well at least the formats match now, even if the cursor position doesn't.

TheSpyder (approved):

modules/tinymce/src/core/test/ts/browser/delete/InlineFormatDeleteTest.ts
mkzhx: Yeah, the cursor positions matching before was purely a coincidence. The difference in cursor positions is because of the difference in structures, which is due to Firefox not needing any custom caret formats being inserted as it natively preserves the original formats, unlike Chrome or Safari.

PR #8392, opened 2023-01-16
TINY-9380: Accessibility fixes for Shortcuts Table in Help dialog

shanmen-tiny:
Related Ticket: TINY-9380

Description of Changes:
* Added optional `classes` property to dialog table (to target the table in help dialog for adding right padding)
* Added right padding to action column in the Help dialog 
* Changed default color of `@dialog-table-border-color`

Screenshot:
image

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/bridge/CHANGELOG.md
+++ b/modules/bridge/CHANGELOG.md
@@ -5,7 +5,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+
+### Added
 - Added optional `select`-property to `colorswatch`-type `fancymenuitem` to allow setting colors as selected. #TINY-9395
+- Added optional `classes` property to dialog table. #TINY-9380
 
 ## 4.2.0 - 2022-11-23
 
--- a/modules/bridge/CHANGELOG.md
+++ b/modules/bridge/CHANGELOG.md
@@ -5,10 +5,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
-
-### Added
 - Added optional `select`-property to `colorswatch`-type `fancymenuitem` to allow setting colors as selected. #TINY-9395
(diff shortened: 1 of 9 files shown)


modules/bridge/CHANGELOG.md
```diff
@@ -5,7 +5,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+
+### Added
 - Added optional `select`-property to `colorswatch`-type `fancymenuitem` to allow setting colors as selected. #TINY-9395
+- Added optional `classes` property to dialog table. #TINY-9380
```
TheSpyder, line 11: I'd like to discuss whether this is the correct approach, as it adds a public API (meaning it should go in the TinyMCE changelog as well). Using a smaller dialog size was presented as an option, and we might be able to style the help dialog table without using a custom property?

Happy to discuss in slack rather than PR comments.

TheSpyder (changes requested):

modules/bridge/CHANGELOG.md
spocke: Yes, we kind of have this idea of having the public declarative api not knowing about how things are represented visually with classes as such. So adding a classes property to the api and specifically using a `tox-` prefixed class kind of goes against that idea. Also it seems that the word count plugin also has padding issues so then it more of a generic thing. So I wonder if there is some form of generic change we instead can do in the oxide skin to avoid having special cases in dialogs like making all tables in a html panel have this extra padding?

spocke (changes requested):

modules/bridge/CHANGELOG.md
shanmen-tiny: @TheSpyder and @spocke, I have reverted my change to adding a new property to the public API, and now tables in an html panel have this extra padding.

TheSpyder: I appreciate that - I've pinged Paul in slack to double-check this still meets his requirements.

TheSpyder (approved):
Have confirmed with Paul that this matches his design

MitchCTiny (approved):

spocke (approved):

PR #8393, opened 2023-01-16
TINY-9491: Added some restrictions to link plugin.

HAFRMO:
Related Ticket: 

Description of Changes:
Link plugin opened the context menu on table of content, added a check to make it stop doing that.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Direct invalid child text nodes of list elements will be wrapped in list item elements. #TINY-4818
 
 ### Changed
+- The `link` plugin will no longer try to add, change, or remove links generated in the `tableofcontent` plugin. #TINY-9491
 - The formatting of `contenteditable="false"` elements are no longer cloned to new cells while creating new table rows. #TINY-9449
 
 ### Fixed
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Direct invalid child text nodes of list elements will be wrapped in list item elements. #TINY-4818
 
 ### Changed
-- The `link` plugin will no longer try to add, change, or remove links generated in the `tableofcontent` plugin. #TINY-9491
+- The `link` plugins context menu items will no longer appear for noneditable links. #TINY-9491
 - The formatting of `contenteditable="false"` elements are no longer cloned to new cells while creating new table rows. #TINY-9449
 
 ### Fixed
--- a/modules/tinymce/src/plugins/image/test/ts/browser/ContextMenuTest.ts
+++ b/modules/tinymce/src/plugins/image/test/ts/browser/ContextMenuTest.ts
@@ -1,5 +1,6 @@
 import { Keys, Mouse, UiFinder, Waiter } from '@ephox/agar';
 import { describe, it } from '@ephox/bedrock-client';
+import { Fun } from '@ephox/katamari';
 import { SugarShadowDom } from '@ephox/sugar';
 import { TinyAssertions, TinyDom, TinyHooks, TinySelections, TinyUiActions } from '@ephox/wrap-mcagar';
 
@@ -12,6 +13,18 @@ describe('browser.tinymce.plugins.image.ContextMenuTest', () => {
     toolbar: 'image',
     indent: false,
     base_url: '/project/tinymce/js/tinymce',
+    setup: (editor: Editor) => {
+      editor.ui.registry.addMenuItem('contextfiller', {
+        icon: 'link',
+        text: 'Context Filler',
+        onAction: Fun.noop
+      });
+
+      editor.ui.registry.addContextMenu('contextfiller', {
+        update: Fun.constant('contextfiller')
+      });
+    },
+    contextmenu: 'image link contextfiller',
     image_caption: true
   }, [ Plugin ], true);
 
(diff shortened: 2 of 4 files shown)


tiny-ben-tran (approved):

modules/tinymce/src/plugins/link/main/ts/ui/Controls.ts
```diff
@@ -57,7 +58,14 @@ const setupContextMenu = (editor: Editor): void => {
   const inLink = 'link unlink openlink';
   const noLink = 'link';
   editor.ui.registry.addContextMenu('link', {
-    update: (element) => Utils.hasLinks(editor.dom.getParents(element, 'a') as HTMLAnchorElement[]) ? inLink : noLink
+    update: (element) => {
+      const isInToC = PredicateFind.ancestor(SugarElement.fromDom(element), (ancestor) => Class.has(ancestor, 'mce-toc')).isSome();
```
TheSpyder, line 62: Are there any other classes we should check for? Or perhaps a list of them somewhere we can reuse?

TheSpyder (approved):

modules/tinymce/src/plugins/link/main/ts/ui/Controls.ts
HAFRMO: Not that I know of, the issue was just with regards to the table of contents. I did consider adding both a list of internal ones ( Which at the moment would just be the one ) and an option to add others by the user, but I thought it'd be best to start off simple. But I can add one or both of those if we want to.

modules/tinymce/src/plugins/image/test/ts/browser/ContextMenuTest.ts
```diff
@@ -36,6 +37,25 @@ describe('browser.tinymce.plugins.image.ContextMenuTest', () => {
...
+          '<li>' +
+            '<a href="#mcetoc_1gmi3gu5b11" data-mce-href="#mcetoc_1gmi3gu5b11">point </a>' +
+          '</li>' +
+        '</ul>' +
+      '</div>' +
+      '<h2 id="mcetoc_1gmi3gu5b12">point</h2>',
+      { format: 'raw' }
+    );
+    Mouse.contextMenuOn(TinyDom.body(editor), 'a');
+    await Waiter.pWait(250); // Give plenty of time for the context menu to appear
```
spocke, line 55: Please use a proper waiter here that waits for a state to appear this just screams flaky tests to me.

spocke (changes requested):

modules/tinymce/src/plugins/image/test/ts/browser/ContextMenuTest.ts
spocke: There is a pTriggerContext menu that can likely be used in this case.

HAFRMO: The test here is to make sure that nothing happens. Since nothing changes, we can't wait for a change. So I set it up to wait for the context menu to have a chance to appear, and then make sure it's not there. Not sure how to improve that otherwise.

spocke: The menu item should be disabled or maybe even not showing at all since it makes little sense to have a button that does nothing. So then one could check for that have a waiter that waits for the context menu then verifies that the menu item is disabled or not there.

HAFRMO: Yes. The context menu does not appear. I can't wait for it to appear, because it won't, and I can't wait for it not to appear, because that will immediately pass the test, even if it was about to show up. So I set a period to wait for it to appear, and then check that it doesn't. I can't really see a better way of checking that nothing happens.

vpyshnenko: Maybe for this test one can configure two context menu items, one should be presented there always (for ex. separator) and the Link item sometimes. Then wait for the context menu to always appeared and check for the availability of the Link item in it.

modules/tinymce/src/plugins/image/test/ts/browser/ContextMenuTest.ts
```diff
@@ -36,6 +50,25 @@ describe('browser.tinymce.plugins.image.ContextMenuTest', () => {
...
+            '<a href="#mcetoc_1gmi3gu5b11" data-mce-href="#mcetoc_1gmi3gu5b11">point </a>' +
+          '</li>' +
+        '</ul>' +
+      '</div>' +
+      '<h2 id="mcetoc_1gmi3gu5b12">point</h2>',
+      { format: 'raw' }
+    );
+    Mouse.contextMenuOn(TinyDom.body(editor), 'a');
+    await pOpenContextMenu(editor, 'a');
+    UiFinder.notExists(SugarShadowDom.getContentContainer(SugarShadowDom.getRootNode(TinyDom.targetElement(editor))), 'div[title="Link..."');
```
spocke, line 69: Curious why what all the shadow dom magic needed here? From what I can see the editor is not rendered in a shadow dom. Couldn't you just look for the selector on SugarBody.body() and maybe make it more specific the selector if possible to include that the title is not part of some other structure since in theory that div could be in the toolbar.

modules/tinymce/CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Direct invalid child text nodes of list elements will be wrapped in list item elements. #TINY-4818
 
 ### Changed
+- The `link` plugin will no longer try to add, change, or remove links generated in the `tableofcontent` plugin. #TINY-9491
```
spocke, line 18: Not sure how we should describe this but the change is now more general. The fix is going in to the core and `tableofcontent` is now a commercial plugin so I don't think we should be referencing this directly.

What about:
```suggestion
- The `link` plugins context menu items will no longer appear for noneditable links. #TINY-9491
```

modules/tinymce/src/plugins/image/test/ts/browser/ContextMenuTest.ts
```diff
@@ -36,6 +50,25 @@ describe('browser.tinymce.plugins.image.ContextMenuTest', () => {
     TinyAssertions.assertSelection(editor, [], 0, [], 1);
   });
 
+  it('TINY-9491: Opening context menus on a selected figure', async () => {
```
spocke, line 53: This test should be in the link plugin also it's more generic than the toc change.

So maybe just setup a div with noneditable links inside.

- [ ] Move test to `plugins/link/test/ts/browser/ContextMenuTest.ts`
- [ ] Make test more generic

spocke (approved):

modules/tinymce/src/plugins/image/test/ts/browser/ContextMenuTest.ts
HAFRMO: I don't know, honestly. When I went going for stuff I went and fetched the function calls we have in the functions ( such as the opening of the context toolbar ) and used those. Since those do work I thought it'd be best to just use them.

modules/tinymce/src/plugins/image/test/ts/browser/ContextMenuTest.ts
HAFRMO: Did as such.

PR #8396, opened 2023-01-17
TINY-9176: Change the focus style of buttons

tiny-ben-tran:
Description of Changes:
* Updated the focus style of menu, toolbar, split and collection buttons
* Included styles for `focus-visible`

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/oxide/src/less/theme/components/toolbar-button/toolbar-split-button.less
```diff
@@ -15,25 +17,29 @@
...
 
   .tox-split-button:focus {
-    background: @toolbar-button-focus-background-color;
+    background: @toolbar-split-button-focus-background-color;
     box-shadow: @toolbar-button-focus-box-shadow;
-    color: @toolbar-button-focus-text-color;
+    color: @toolbar-split-button-focus-text-color;
+
+    svg {
+      fill: @toolbar-split-button-focus-text-color;
```
TheSpyder, line 33: This might be why your changes break the "current colour" line in the forecolor and backcolor toolbar buttons. Those use SVG `fill` styles to change dynamically:
https://github.com/tinymce/tinymce/blob/a81d5d3b7e3d9efd1126b202a1858ae2a9022640/modules/tinymce/src/themes/silver/main/ts/ui/toolbar/button/ToolbarButtons.ts#L265-L269

Maybe your styles are impacting more elements than you expected? Try using the DOM inspector to see what's overriding the runtime fill styles.

TheSpyder (approved):
Seems ok as a general concept, happy to merge and wait for QA to see if any style bugs crop up as a result of this.

modules/oxide/src/less/theme/components/toolbar-button/toolbar-split-button.less
tiny-ben-tran: This also happens on 6-stable when using 'oxide-dark' skin

tiny-ben-tran: I just figured this out. The underline bar is for showing the currently selected color (by calling setIconFill). So it works as expected

modules/oxide/src/less/theme/components/collection/collection.less
```diff
@@ -17,18 +17,21 @@
 @collection-item-label-disabled-text-color: fade(@collection-item-label-text-color, 50%);
 @collection-item-caret-disabled-text-color: fade(@collection-item-label-text-color, 50%);
 
-@collection-list-item-active-background-color: @toolbar-button-hover-background-color;
-@collection-list-item-label-active-text-color: contrast(@collection-list-item-active-background-color, @color-black, @color-white);
+@collection-item-hover-background-color: #285ec7;
```
spocke, line 20: This variable needs to be inherited from something else. It shouldn't be hard coded here in the component but be assigned to some other variable so that a skin author just needs to alter the background color at one place.

modules/oxide/src/less/theme/components/collection/collection.less
```diff
@@ -17,18 +17,21 @@
...
+@collection-item-hover-text-color: #fff;
```
spocke, line 21: Same here if you look at the other things in the file this needs to use some other type of variable not sure what.

modules/oxide/src/less/theme/components/collection/collection.less
```diff
@@ -116,6 +119,10 @@
     user-select: none;
   }
 
+  .tox-collection__item--active {
+    fill: #fff;
```
spocke, line 123: This should not be hard coded it needs to be a variable and probably have it be inherited from something else.

modules/oxide/src/less/theme/components/collection/collection.less
```diff
@@ -223,6 +230,21 @@
     }
   }
 
+  .tox-collection__item--active .tox-collection__item-accessory,
+  .tox-collection__item--active .tox-collection__item-caret {
+    color: rgba(255, 255, 255, .7);
```
spocke, line 235: These properties should not be hard coded.

modules/oxide/src/less/theme/components/menubar/menubar-button.less
```diff
@@ -22,20 +22,20 @@
 @menubar-select-disabled-box-shadow: @toolbar-button-box-shadow;
 @menubar-select-disabled-text-color: fade(@toolbar-button-text-color, 50%);
 
-@menubar-select-hover-background-color: @toolbar-button-hover-background-color;
+@menubar-select-hover-background-color: #285ec7;
```
spocke, line 25: Why where these changed to a hard coded value. If they need altering could they have been altered with fade or something instead.

modules/oxide/src/less/theme/components/menubar/menubar-button.less
```diff
@@ -22,20 +22,20 @@
...
+@menubar-select-hover-text-color: #fff;
```
spocke, line 26: This needs to be calculated some how.

spocke (changes requested):
There needs to be more variables in general and they need to inherit or be altered from existing values. Verify that you don't break the dark skin for example when changing this stuff should not be white when they should be black etc.

tiny-ben-tran:
> There needs to be more variables in general and they need to inherit or be altered from existing values. Verify that you don't break the dark skin for example when changing this stuff should not be white when they should be black etc.

@spocke, this is partly my fault for not marking this PR as a draft, as my plan is to get Paul K to approve the visual changes first. However, this comment is still helpful for me to figure out how these variables work

modules/oxide/src/less/theme/components/collection/collection.less
```diff
@@ -136,37 +144,38 @@
...
+    background-color: @collection-toolbar-item-active-background-color;
     color: @collection-toolbar-item-label-active-text-color;
   }
 
+  .tox-collection__item-caret {
+    align-items: center;
+    display: flex;
+    min-height: @collection-item-list-icon-height;
+
+    // Workaround for the above not centering items in IE11. See https://github.com/philipwalton/flexbugs/issues/231#issuecomment-362790042
```
noxuhax, line 167: This seems to be unnecessary since we’ve dropped IE11 support

PR #8398, opened 2023-01-20
TINY-8977: Inline header fixup.

HAFRMO:
Related Ticket: 

Description of Changes:
Minimum width changes

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/themes/silver/main/ts/modes/Inline.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/modes/Inline.ts
@@ -1,6 +1,6 @@
 import { AlloyComponent, Attachment, Boxes, Disabling } from '@ephox/alloy';
 import { Cell, Singleton } from '@ephox/katamari';
-import { DomEvent, SugarElement } from '@ephox/sugar';
+import { DomEvent, Scroll, SugarElement } from '@ephox/sugar';
 
 import Editor from 'tinymce/core/api/Editor';
 import { NodeChangeEvent } from 'tinymce/core/api/EventTypes';
@@ -61,7 +61,16 @@ const setupEvents = (editor: Editor, targetElm: SugarElement, ui: InlineHeader, 
     requestAnimationFrame(() => resizeContent(e));
   });
 
-  editor.on('ScrollWindow', () => ui.updateMode());
+  let lastScrollX = 0;
+  editor.on('ScrollWindow', () => {
+    const newScrollX = Scroll.get().left;
+    if (newScrollX !== lastScrollX) {
+      lastScrollX = newScrollX;
+      ui.update();
+    }
+
+    ui.updateMode();
+  });
 
   // Bind to async load events and trigger a content resize event if the size has changed
   const elementLoad = Singleton.unbindable();
--- a/modules/tinymce/src/themes/silver/main/ts/modes/Inline.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/modes/Inline.ts
@@ -1,5 +1,5 @@
 import { AlloyComponent, Attachment, Boxes, Disabling } from '@ephox/alloy';
-import { Cell, Singleton } from '@ephox/katamari';
+import { Cell, Singleton, Throttler } from '@ephox/katamari';
 import { DomEvent, Scroll, SugarElement } from '@ephox/sugar';
 
 import Editor from 'tinymce/core/api/Editor';
@@ -62,11 +62,12 @@ const setupEvents = (editor: Editor, targetElm: SugarElement, ui: InlineHeader, 
   });
 
   let lastScrollX = 0;
+  const updateUi = Throttler.last(() => ui.update(), 33);
   editor.on('ScrollWindow', () => {
     const newScrollX = Scroll.get().left;
     if (newScrollX !== lastScrollX) {
       lastScrollX = newScrollX;
-      ui.update();
+      updateUi.throttle();
     }
 
     ui.updateMode();
--- a/modules/tinymce/src/themes/silver/main/ts/ui/header/InlineHeader.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/ui/header/InlineHeader.ts
@@ -24,6 +24,8 @@ export interface InlineHeader {
 
 const { ToolbarLocation, ToolbarMode } = Options;
 
+const scanIfWithinDistance = 40;
+
 export const InlineHeader = (
   editor: Editor,
   targetElm: SugarElement<HTMLElement>,
@@ -133,7 +135,7 @@ export const InlineHeader = (
       };
 
       const widthProperties = optToolbarWidth.map(
-        (toolbarWidth: number) => {
+        (toolbarWidth: number): {} | { width: string } => {
           const scroll = Scroll.get();
 
           /*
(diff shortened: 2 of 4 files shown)


spocke (approved):

yacodes (approved):

modules/tinymce/src/themes/silver/main/ts/ui/header/InlineHeader.ts
```diff
@@ -140,6 +176,22 @@ export const InlineHeader = (
...
+  const restoreAndGetCompleteOuterContainerWidth = (): Optional<number> => {
+    /*
+    Editors can be placed so far to the right that their left position is beyond the window width. This causes problems with flex-wrap. To solve this, set a width style on the container.
+    Natural width of the container needs to be calculated first.
+    */
+    if (!useFixedToolbarContainer) {
+      Css.set(mainUi.outerContainer.element, 'position', 'absolute');
+      Css.set(mainUi.outerContainer.element, 'left', '0px');
+      Css.remove(mainUi.outerContainer.element, 'width');
+      const w = Width.getOuter(mainUi.outerContainer.element);
```
TheSpyder, line 188: This is a fairly expensive process, particularly if we do it a lot while horizontal scrolling, and the results won't change very often. Is there a way to cache the value?

modules/tinymce/src/themes/silver/main/ts/modes/Inline.ts
```diff
@@ -61,7 +61,16 @@ const setupEvents = (editor: Editor, targetElm: SugarElement, ui: InlineHeader,
     requestAnimationFrame(() => resizeContent(e));
   });
 
-  editor.on('ScrollWindow', () => ui.updateMode());
+  let lastScrollX = 0;
+  editor.on('ScrollWindow', () => {
+    const newScrollX = Scroll.get().left;
+    if (newScrollX !== lastScrollX) {
+      lastScrollX = newScrollX;
+      ui.update();
```
TheSpyder, line 69: Does `ScrollWindow` fire for every pixel? We might need to debounce this so it's not using a lot of CPU. `ui.update()` is quite a thorough and complex operation.

TheSpyder (approved):

modules/tinymce/src/themes/silver/main/ts/modes/Inline.ts
HAFRMO: How would you prefer to do it? I don't know if we have any existing functionality for it.

modules/tinymce/src/themes/silver/main/ts/ui/header/InlineHeader.ts
HAFRMO: We probably could cache it, but the issue is knowing for sure it's updated frequently enough, and that we don't end up with the wrong value.

TheSpyder: Yeah we'd need some way to compare what we have to the new state of the UI and determine whether it needs updating.

A simple fix is don't set a width when one isn't needed - that will reduce the impact of scrolling that doesn't trigger the bug (since you then won't be removing the width and repositioning the toolbar constantly during scroll).

From there, we could do things like reduce how much we measure unless the right side of the toolbar is within some area of pixels on the right side of the viewport.

modules/tinymce/src/themes/silver/main/ts/modes/Inline.ts
TheSpyder: A debounce? That's `Throttler.last()`. Set it to 33ms and see how it looks, if it's too janky maybe lower it a bit but not below 16.

modules/tinymce/src/themes/silver/main/ts/ui/header/InlineHeader.ts
```diff
@@ -24,6 +24,8 @@ export interface InlineHeader {
 
 const { ToolbarLocation, ToolbarMode } = Options;
 
+const scanIfWithinDistance = 40;
```
hamza0867, line 27: The name of this variable is a verb. I think it would be more appropriate to name it with a name indicative of what the variable is used for, like `minimumDistanceToChange`

modules/tinymce/src/themes/silver/main/ts/ui/header/InlineHeader.ts
```diff
@@ -126,10 +128,50 @@ export const InlineHeader = (
...
+      const baseProperties = {
         position: 'absolute',
-        top: Math.round(top) + 'px',
-        left: Math.round(targetBounds.x) + 'px'
+        left: Math.round(targetBounds.x) + 'px',
+        top: Math.round(top) + 'px'
+      };
+
+      const widthProperties = optToolbarWidth.map(
+        (toolbarWidth: number): {} | { width: string } => {
```
hamza0867, line 138: This pattern to me looks like where an optional should be instead of `{} | { width: string }` I would suggest the return of this function be `Optional<{width: string}>` and use `bind` function instead of `map`

spocke (approved):

modules/tinymce/src/themes/silver/main/ts/ui/header/InlineHeader.ts
HAFRMO: A good name is a bit tricky, though I agree with your point.

modules/tinymce/src/themes/silver/main/ts/ui/header/InlineHeader.ts
HAFRMO: I considered doing about that, but I feel it's a very messy way of doing it. Not that mine feels the best either. Not sure which one people would consider the easiest to read.

hamza0867 (approved):

modules/tinymce/src/themes/silver/main/ts/ui/header/InlineHeader.ts
TheSpyder: You're already inside an optional, so you've just created `Optional<{} | { width: string }>`

But when I see code like this
```
          if (availableWidth > width) { // If there's already enough space, don't add a width for performance reasons.
            return {};
          }

          return {
            width: width + 'px'
          };
```

That's an optional. Please change `return {}` to `return Optional.none()` and combine this with the outer optional - the empty case is already handled at the end via `.getOr({ });`

Personally I'd consider keeping it as an `Optional<number>` and using better techniques to apply it than object spread inside `Css.setAll`, but I suspect it's a little too late for that.

TheSpyder (changes requested):

modules/tinymce/src/themes/silver/main/ts/ui/header/InlineHeader.ts
HAFRMO: I am aware that it's in an optional. I just didn't think it'd help readability to add that there. But since there seems to be strong preference for this, I'll do that.

TheSpyder: `Optional<{} | { width: string }>` is _much_ harder to follow than `toolbarWidth.bind()`, I would hope?

TheSpyder (approved):

PR #8399, opened 2023-01-23
TINY-9443: 'Find and Replace' modal closes incorrectly when there is no match found

hamza0867:
Related Ticket: TINY-9443

Description of Changes:
* Instead of opening an alert on error, we now just show an error banner in the search replace modal

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/tinymce/src/plugins/searchreplace/main/ts/ui/Dialog.ts
```diff
@@ -101,47 +99,60 @@ const open = (editor: Editor, currentSearchState: Cell<Actions.SearchState>): vo
...
+        type: 'alertbanner',
+        level: 'error',
+        text: 'Could not find the specified string.',
+        icon: 'warning',
+      });
+    }
+    return items;
+  };
+
+  const getDialogSpec = (error: boolean, initialData: DialogData): Dialog.DialogSpec<DialogData> => ({
```
yacodes, line 150:
```suggestion
  const getDialogSpec = (initialData: DialogData, error?: boolean): Dialog.DialogSpec<DialogData> => ({
```

Consider making the `error` variable optional and second in the parameters list. This will allow calling the function without providing a boolean argument first. 

``` ts
getDialogSpec(dialogData)

// instead of
getDialogSpec(false, dialogData)
```

I feel like this is quite more straightforward, unless I am missing some common practice of using `error` as a first argument.

yacodes (changes requested):

modules/tinymce/src/plugins/searchreplace/main/ts/ui/Dialog.ts
yacodes: Yeah, I know see what you've meant. Maybe it will be a bit more readable to rename the variable to `hasError` to underline the boolean nature of the variable. Object literal is a great idea as well! :)

spocke: Yeah maybe call it `showNoMatchesAlertBanner` or `showNoMatchesMessage' or something that describes what it does though out the code. Or maybe make it an Optional string with a error message and pass that down the chain not sure what that would be at this point.

spocke (approved):

yacodes (approved):

modules/tinymce/src/plugins/searchreplace/main/ts/ui/Dialog.ts
hamza0867: Done!

PR #8405, opened 2023-01-24
TINY-9534: Add template icon

vpyshnenko:
Related Ticket: TINY-9534

Description of Changes:
*  Add icon for `Add template` toolbar button

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [x] Milestone set


yacodes (approved):

lnewson:
Don't forget that when doing something that is a minor/major change in the monorepo libraries you need to update `versions.txt` as well. That was missed for the typography plugin and it was accidentally released as a patch instead.

vpyshnenko:
> Don't forget that when doing something that is a minor/major change in the monorepo libraries you need to update `versions.txt` as well. That was missed for the typography icon and it was accidentally released as a patch instead.

Done. It is easy to forget not least due `oxide-icons-default` does not have its own changelog like other icon packages.

spocke (approved):
I'm surprised that the oxide-icons-default doesn't have a change log.

PR #8408, opened 2023-01-26
TINY-9454: Preserve formatting when backspacing from one empty block to another

yacodes:
Related Ticket: TINY-9454

Description of Changes:
* Preserve formatting when backspacing from one empty block to another. Otherwise block's formatting is cleared by inserting `br[data-mce-bogus]`.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/delete/MergeBlocks.ts
+++ b/modules/tinymce/src/core/main/ts/delete/MergeBlocks.ts
@@ -1,5 +1,5 @@
 import { Arr, Fun, Optional } from '@ephox/katamari';
-import { Compare, Insert, PredicateFilter, Remove, SugarElement, Traverse } from '@ephox/sugar';
+import { Compare, Insert, Replication, PredicateFind, Remove, SugarElement, Traverse } from '@ephox/sugar';
 
 import * as CaretFinder from '../caret/CaretFinder';
 import CaretPosition from '../caret/CaretPosition';
@@ -55,22 +55,23 @@ const nestedBlockMerge = (
 
 const sidelongBlockMerge = (rootNode: SugarElement<Node>, fromBlock: SugarElement<Element>, toBlock: SugarElement<Element>): Optional<CaretPosition> => {
   if (Empty.isEmpty(toBlock)) {
-    const inlineDescendants = Traverse.firstChild(toBlock).fold(
-      Fun.constant([]),
-      (child) => {
-        const descendants = PredicateFilter.descendants(child, ElementType.isInline);
-        return ElementType.isInline(child) ? [ child, ...descendants ] : descendants;
-      }
-    );
+    if (Empty.isEmpty(fromBlock)) {
+      const bogus = PaddingBr.createPaddingBr();
+      const child = Traverse.firstChild(toBlock).fold(
+        Fun.constant(bogus),
+        (firstChild) => {
+          const firstChildCopy = Replication.deep(firstChild);
+          PredicateFind
+            .descendant(firstChildCopy, (el) => !Traverse.hasChildNodes(el))
+            .each((deepestChild) => Insert.append(deepestChild, bogus));
+          return firstChildCopy;
+        }
+      );
+      Remove.empty(fromBlock);
+      Insert.append(fromBlock, child);
+    }
 
     Remove.remove(toBlock);
-    if (Empty.isEmpty(fromBlock)) {
-      const bogus = PaddingBr.fillWithPaddingBr(fromBlock);
-      Arr.foldr(inlineDescendants, (descendant: SugarElement<HTMLElement>, wrapper) => {
-        Insert.wrap(descendant, wrapper);
-        return wrapper;
-      }, bogus);
-    }
     return CaretFinder.firstPositionIn(fromBlock.dom);
   }
 
--- a/modules/tinymce/src/core/main/ts/delete/MergeBlocks.ts
+++ b/modules/tinymce/src/core/main/ts/delete/MergeBlocks.ts
@@ -1,5 +1,5 @@
 import { Arr, Fun, Optional } from '@ephox/katamari';
-import { Compare, Insert, PredicateFilter, Replication, Remove, SugarElement, Traverse } from '@ephox/sugar';
+import { Compare, Insert, Replication, Remove, SugarElement, Traverse } from '@ephox/sugar';
 
 import * as CaretFinder from '../caret/CaretFinder';
 import CaretPosition from '../caret/CaretPosition';
@@ -56,17 +56,18 @@ const nestedBlockMerge = (
 const sidelongBlockMerge = (rootNode: SugarElement<Node>, fromBlock: SugarElement<Element>, toBlock: SugarElement<Element>): Optional<CaretPosition> => {
   if (Empty.isEmpty(toBlock)) {
     if (Empty.isEmpty(fromBlock)) {
-      const inlineToBlockDescendants = Traverse.firstChild(toBlock).fold(
-        Fun.constant([]),
-        (child) => {
-          const descendants = PredicateFilter.descendants(child, ElementType.isInline);
-          return Arr.map(ElementType.isInline(child) ? [ child, ...descendants ] : descendants, Replication.shallow);
-        }
-      );
+      const getInlineToBlockDescendants = (el: SugarElement<Element>) => {
+        const helper = (node: SugarElement<Element>, elements: SugarElement<Element>[]): SugarElement<Element>[] =>
+          Traverse.firstChild(node).fold(
+            () => elements,
+            (child) => ElementType.isInline(child) ? helper(child, elements.concat(Replication.shallow(child))) : elements
+          );
+        return helper(el, []);
+      };
 
       const newFromBlockDescendants = Arr.foldr(
-        inlineToBlockDescendants,
-        (element: SugarElement<HTMLElement>, wrapper) => {
+        getInlineToBlockDescendants(toBlock),
+        (element: SugarElement<Element>, wrapper) => {
           Insert.wrap(element, wrapper);
           return wrapper;
         },
--- a/modules/tinymce/src/core/main/ts/delete/MergeBlocks.ts
+++ b/modules/tinymce/src/core/main/ts/delete/MergeBlocks.ts
@@ -67,9 +67,9 @@ const sidelongBlockMerge = (rootNode: SugarElement<Node>, fromBlock: SugarElemen
 
       const newFromBlockDescendants = Arr.foldr(
         getInlineToBlockDescendants(toBlock),
-        (element: SugarElement<Element>, wrapper) => {
-          Insert.wrap(element, wrapper);
-          return wrapper;
+        (element: SugarElement<Element>, descendant) => {
+          Insert.wrap(element, descendant);
+          return descendant;
         },
         PaddingBr.createPaddingBr()
       );
--- a/modules/tinymce/src/core/main/ts/delete/MergeBlocks.ts
+++ b/modules/tinymce/src/core/main/ts/delete/MergeBlocks.ts
@@ -59,8 +59,7 @@ const sidelongBlockMerge = (rootNode: SugarElement<Node>, fromBlock: SugarElemen
     Remove.remove(toBlock);
     if (Empty.isEmpty(fromBlock)) {
       PaddingBr.fillWithPaddingBr(fromBlock);
-      SelectorFind.child(fromBlock, '[data-mce-bogus]').fold(
-        Fun.noop,
+      SelectorFind.child(fromBlock, '[data-mce-bogus]').each(
         (bogus) => Arr.each(descendants, (descendant) => Insert.wrap(bogus, descendant))
       );
     }
--- a/modules/tinymce/src/core/main/ts/delete/MergeBlocks.ts
+++ b/modules/tinymce/src/core/main/ts/delete/MergeBlocks.ts
@@ -1,5 +1,5 @@
 import { Arr, Fun, Optional } from '@ephox/katamari';
-import { Compare, Insert, Replication, PredicateFind, Remove, SugarElement, Traverse } from '@ephox/sugar';
+import { Compare, Insert, PredicateFilter, Remove, SugarElement, Traverse } from '@ephox/sugar';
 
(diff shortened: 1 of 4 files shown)


lorenzo-pomili (approved):

modules/tinymce/src/core/main/ts/delete/MergeBlocks.ts
```diff
@@ -55,9 +55,14 @@ const nestedBlockMerge = (
 
 const sidelongBlockMerge = (rootNode: SugarElement<Node>, fromBlock: SugarElement<Element>, toBlock: SugarElement<Element>): Optional<CaretPosition> => {
   if (Empty.isEmpty(toBlock)) {
+    const descendants = SelectorFilter.descendants(toBlock, '*');
     Remove.remove(toBlock);
     if (Empty.isEmpty(fromBlock)) {
       PaddingBr.fillWithPaddingBr(fromBlock);
+      SelectorFind.child(fromBlock, '[data-mce-bogus]').fold(
+        Fun.noop,
+        (bogus) => Arr.each(descendants, (descendant) => Insert.wrap(bogus, descendant))
```
TheSpyder, line 64: This probably isn't doing what you think it's doing, either that or it's a weird way to do it, I'd like to know what you're thinking here.

`Insert.wrap` starts by inserting the first argument into the parent of the second argument, which will move `bogus` out of `fromBlock` and put it into `toBlock`, but `toBlock` has been removed from the DOM. I'm not sure how your test is passing. Even if it works in some scenarios this seems like it will lead to weird edge cases.

`InsertAll.append(bogus, descendants)` is a better way to append an array of elements into an element, and then if you want the `bogus` element to move please do so more clearly.

TheSpyder (changes requested):

modules/tinymce/src/core/main/ts/delete/MergeBlocks.ts
```diff
@@ -55,9 +55,14 @@ const nestedBlockMerge = (
 
 const sidelongBlockMerge = (rootNode: SugarElement<Node>, fromBlock: SugarElement<Element>, toBlock: SugarElement<Element>): Optional<CaretPosition> => {
   if (Empty.isEmpty(toBlock)) {
+    const descendants = SelectorFilter.descendants(toBlock, '*');
```
spocke, line 58: This would break if the structure is complex say it shouldn't wrap the BR with both B and I in this case.
```
<p>
<b></b>
<i></i>
</p>
```

So we are likely only interested in the first or last branch of the tree also only inline elements. So I suggest doing a recursive function that traverses down on the first child and collects them and maybe filters out blocks.

So if the structure is `<p><b><i></i></b><span></span>` it would clone the B, I but not the SPAN.
Also if the structure is <div><p><b><i>...` then I think it should ignore the P element.

modules/tinymce/src/core/main/ts/delete/MergeBlocks.ts
spocke: I would expect this to be an Arr.foldl where it wraps on each iteration and returns the newly wrapped element for the next iteration.

spocke (changes requested):

modules/tinymce/src/core/main/ts/delete/MergeBlocks.ts
yacodes: `InsertAll.append(bogus, descendants)` is not exactly what I want. The idea is to wrap the `bogus` with `toBlock` descendants. 

Update the code with more intuitive solution by @spocke. using `Arr.foldr` and returning the newly wrapped element.

https://github.com/tinymce/tinymce/blob/5d854d3e550ce06fca195f6cac6b2f3ce42d4b40/modules/tinymce/src/core/main/ts/delete/MergeBlocks.ts#L68-L72

modules/tinymce/src/core/main/ts/delete/MergeBlocks.ts
```diff
@@ -55,9 +55,14 @@ const nestedBlockMerge = (
...
     Remove.remove(toBlock);
     if (Empty.isEmpty(fromBlock)) {
       PaddingBr.fillWithPaddingBr(fromBlock);
```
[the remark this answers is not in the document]
yacodes: Supposing the wrong line is selected here. Updated `PaddingBr.fillWithPaddingBr(fromBlock);` to return the `br` element.

https://github.com/tinymce/tinymce/blob/5d854d3e550ce06fca195f6cac6b2f3ce42d4b40/modules/tinymce/src/core/main/ts/dom/PaddingBr.ts#L32-L37

modules/tinymce/src/core/main/ts/delete/MergeBlocks.ts
yacodes: How can this kind of markup appear, besides editing HTML manually? Just curious.

Updated the code with the traversing.

https://github.com/tinymce/tinymce/blob/5d854d3e550ce06fca195f6cac6b2f3ce42d4b40/modules/tinymce/src/core/main/ts/delete/MergeBlocks.ts#L58-L64

yacodes: And added an additional inline element to the test as well to handle the case with several children.

TheSpyder: If you look at the implementation of `Empty.isEmpty`, Spocke's example should return true. How we might get there is another question :)
https://github.com/tinymce/tinymce/blob/ece3bf774b9ae20910d073af5351f162c4fb52d5/modules/tinymce/src/core/main/ts/dom/Empty.ts#L31-L71

modules/tinymce/src/core/main/ts/delete/MergeBlocks.ts
```diff
@@ -55,9 +55,21 @@ const nestedBlockMerge = (
...
+        return ElementType.isInline(child) ? [ child, ...descendants ] : descendants;
+      }
+    );
+
     Remove.remove(toBlock);
     if (Empty.isEmpty(fromBlock)) {
-      PaddingBr.fillWithPaddingBr(fromBlock);
+      const bogus = PaddingBr.fillWithPaddingBr(fromBlock);
+      Arr.foldr(inlineDescendants, (descendant: SugarElement<HTMLElement>, wrapper) => {
+        Insert.wrap(descendant, wrapper);
```
TheSpyder, line 70: This still doesn't look right. You said `The idea is to wrap the bogus with toBlock descendants`, but now you're wrapping descendants with the `br` tag.

My main point with the code yesterday was that your result remains in `toBlock`; I would expect something like
* move descendant into `fromBlock`
* move the BR into the descendant
* something in this process that ensures the order of the nested structure is retained

yacodes: I'm utterly confused. If this code worked as you're describing there will be nothing in the editor. As well as the tests won't pass. Everything works as expected though. Also how cna one wrap something with a `br` tag? `br` is always the deepest descendant in the `Insert.wrap` call.

I've updated the code with an alternative solution, explicitly cloning children from the `toBlock` with `Replication` API and inserting the bogus to the deepest child. Hoping this updated version will clarify things a bit.

https://github.com/tinymce/tinymce/blob/1c83d5b1e2668bb0b56985b949bafc0bbad39fc1/modules/tinymce/src/core/main/ts/delete/MergeBlocks.ts#L58-L72

TheSpyder: A couple of things happened here. I was wrong, your choice of variable names and a cursory glance at the modules it references lead me to an incorrect conclusion. But after further investigation _it's even more confusing_.

Please revert the latest commit and let's try again. You were actually quite close to what I expected.

What happened is I misunderstood the `bogus` const. We usually reserve that variable name for references to the actual bogus BR tag, and that's what it appeared to be looking at the PR, but that's not what you did. Perhaps you thought it was `fromBlock`, which would make your correct. But it was undefined - `fillWithPaddingBr` _does not return a value_.
https://github.com/tinymce/tinymce/blob/ece3bf774b9ae20910d073af5351f162c4fb52d5/modules/tinymce/src/core/main/ts/dom/PaddingBr.ts#L32-L35

I'm not sure how your code did anything, because it should've resulted in references to `bogus.dom` which is a runtime error when `bogus` is `undefined`.

So, here's what I think we should do, rather than your changes to meet my incorrect advice:
* Remove the `bogus` variable
* pass `fromBlock` to `Arr.foldr` as the wrapper
* write more tests, because I suspect the similarity of the two empty paragraphs is why it passes incorrectly. And make sure to do the red/green cycle - disable your fix and see if the tests fail in the way you expect.

That should be better.

TheSpyder: Wait, wait, you modified `fillWithPaddingBr` and reversed it in that commit. So I'm not going crazy. But your variable names were wrong - due to how `Arr.foldr` works `descendant` is the bogus element, `wrapper` is the descendant. And so we return to your code being confusing.

I'm not sure what to recommend now. Still revert the commit, but try renaming the variables and instead of modifying `fillPaddingBr` just put that code here. Yes your changes didn't break existing usages of the function, but I as someone who expects the function to do a certain thing was confused about how your code operated and _that_ is the problem. Code should not be surprising.

modules/tinymce/src/core/main/ts/delete/MergeBlocks.ts
```diff
@@ -55,10 +55,29 @@ const nestedBlockMerge = (
...
-    Remove.remove(toBlock);
     if (Empty.isEmpty(fromBlock)) {
-      PaddingBr.fillWithPaddingBr(fromBlock);
+      const inlineToBlockDescendants = Traverse.firstChild(toBlock).fold(
+        Fun.constant([]),
+        (child) => {
+          const descendants = PredicateFilter.descendants(child, ElementType.isInline);
+          return ElementType.isInline(child) ? [ child, ...descendants ] : descendants;
+        }
+      );
```
spocke, line 65: Thinking something like this just grab the first child down the tree. Not sure if these structures are common since they would all need to be empty for this to fail. Like `<p><b></b><i></i></p>` the code you got would still get the descendants of the child so `<p><b><i></i><u></u></b></p>` would grab B, I, U while it should only grab B and I.

```suggestion
      const getInlineToBlockDescendants = (el: SugarElement<Element>) => {
         const helper = (node: SugarElement<Node>, elements: SugarElement<Element>[]) => {
            Traverse.firstChild(node).fold(
              () => elements,
              (child) => SugarNode.isElement(child) ? helper(child, elements.concat(child)) : elements
            );
         };

         return helper(el, []);
      };
      const inlineToBlockDescendants = getInlineToBlockDescendants(toBlock);
```

spocke (approved):

modules/tinymce/src/core/main/ts/delete/MergeBlocks.ts
yacodes: Thanks, missed that. Updated the code and added the related test.

modules/tinymce/src/core/main/ts/delete/MergeBlocks.ts
TheSpyder: > But `bogus` is the **descendant** and `toBlock`'s descendant is the **wrapper** of the `bogus` in this context. 

That makes sense _inside_ the function. But looking at the full context, when you're iterating over an array called `inlineDescendants`, we normally use words like `descendant` in the function argument to refer to elements of the array. Suddenly reversing that, while technically correct, is logically confusing.

One way I deal with this (naming is hard) is by using the generic `acc` name for the accumulator, as the implementation of `foldr` does. Then I don't need to describe what the accumulator is holding, only that it's the accumulator.

modules/tinymce/src/core/main/ts/delete/MergeBlocks.ts
```diff
@@ -55,10 +55,30 @@ const nestedBlockMerge = (
...
 const sidelongBlockMerge = (rootNode: SugarElement<Node>, fromBlock: SugarElement<Element>, toBlock: SugarElement<Element>): Optional<CaretPosition> => {
   if (Empty.isEmpty(toBlock)) {
-    Remove.remove(toBlock);
     if (Empty.isEmpty(fromBlock)) {
-      PaddingBr.fillWithPaddingBr(fromBlock);
+      const getInlineToBlockDescendants = (el: SugarElement<Element>) => {
+        const helper = (node: SugarElement<Element>, elements: SugarElement<Element>[]): SugarElement<Element>[] =>
+          Traverse.firstChild(node).fold(
+            () => elements,
+            (child) => ElementType.isInline(child) ? helper(child, elements.concat(Replication.shallow(child))) : elements
```
TheSpyder, line 63: I thought I followed up but clearly not (I've had a crazy day). Why did you re-add the `Replication.shallow`? I don't mind if we move the elements, or was there a bug this resolved?

modules/tinymce/src/core/main/ts/delete/MergeBlocks.ts
```diff
@@ -55,10 +55,30 @@ const nestedBlockMerge = (
...
+          );
+        return helper(el, []);
+      };
+
+      const newFromBlockDescendants = Arr.foldr(
+        getInlineToBlockDescendants(toBlock),
+        (element: SugarElement<Element>, wrapper) => {
+          Insert.wrap(element, wrapper);
+          return wrapper;
```
TheSpyder, line 72: To make this clearer, as per my other comments, I suggest
```suggestion
        (element: SugarElement<Element>, descendant) => {
          Insert.wrap(element, descendant);
          return descendant;
```

TheSpyder (approved):
That looks good, thanks for sticking with me 👍

modules/tinymce/src/core/main/ts/delete/MergeBlocks.ts
yacodes: Since we've decided to filter all block type elements just moving the child throws `The new child is an ancestor of the parent`, which is understandable, since the element still contains its children. With `Replication.shallow` we map all the descendants, filter block type elements and making a shallow copy without children.

HanKienjd:
@TheSpyder 
I tried with version 6.4.1 , for tinymce-react which is the corresponding version ...

PR #8412, opened 2023-01-29
TINY-9277: Remove the editor focus highlight when not in focus

tiny-ben-tran:
Related Ticket: TINY-9277

Description of Changes:
* Removed the editor focus highlight when not in focus

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/focus/FocusController.ts
+++ b/modules/tinymce/src/core/main/ts/focus/FocusController.ts
@@ -59,11 +59,11 @@ const registerEvents = (editorManager: EditorManager, e: { editor: Editor }) => 
 
   SelectionRestore.register(editor);
 
-  const toggleContainerFocus = (editor: Editor) => {
+  const toggleContentAreaOnFocus = (editor: Editor, fn: (element: SugarElement<Element>, clazz: string) => void) => {
     // Inline editors have a different approach to highlight the content area on focus
     if (Options.shouldHighlightOnFocus(editor) && editor.inline !== true) {
       const contentArea = SugarElement.fromDom(editor.getContainer());
-      Class.toggle(contentArea, 'tox-edit-focus');
+      fn(contentArea, 'tox-edit-focus');
     }
   };
 
@@ -77,7 +77,7 @@ const registerEvents = (editorManager: EditorManager, e: { editor: Editor }) => 
 
       editorManager.setActive(editor);
       editorManager.focusedEditor = editor;
-      toggleContainerFocus(editor);
+      toggleContentAreaOnFocus(editor, Class.add);
       editor.dispatch('focus', { blurredEditor: focusedEditor });
       editor.focus(true);
     }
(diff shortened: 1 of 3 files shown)


modules/tinymce/src/core/main/ts/focus/FocusController.ts
```diff
@@ -59,11 +59,11 @@ const registerEvents = (editorManager: EditorManager, e: { editor: Editor }) =>
 
   SelectionRestore.register(editor);
 
-  const toggleContainerFocus = (editor: Editor) => {
+  const toggleContentAreaOnFocus = (editor: Editor, fn: (element: SugarElement<Element>, clazz: string) => void) => {
```
TheSpyder, line 62: I'm not sure exactly how, but I feel like `Class.toggler()` will make this easier. Or maybe just using the `Toggler` module directly, since there's extra processing required each time.

TheSpyder (approved):

modules/tinymce/src/core/main/ts/focus/FocusController.ts
spocke: I don't think it will ever fire focusin twice in a row. Also the the `classList.add` will only add the class if it's not there a toggler would first read the value to check if it's there then toggle it on so that would be two dom operations so likely slower. I think the code is pretty easy to understand as it is.

modules/tinymce/src/core/test/ts/webdriver/keyboard/IframeTabfocusTest.ts
```diff
@@ -70,5 +69,15 @@ describe('webdriver.tinymce.core.keyboard.IframeTabfocusTest', () => {
       await RealMouse.pClickOn('iframe => body');
       assertIsHighlighted(editor);
     });
+
+    it('TINY-9277: Remove highlight when not in focus', async () => {
```
spocke, line 73: This could also be done without using webdriver just focus something else on the page with element.focus() or do editor.focus(); you could keep this test as well but I prefer to avoid having webdriver tests if we don't need to.

spocke (approved):

modules/tinymce/src/core/test/ts/webdriver/keyboard/IframeTabfocusTest.ts
tiny-ben-tran: Agreed that we should avoid writing webdriver tests if possible. I have converted this into a browser test.

modules/tinymce/src/core/main/ts/focus/FocusController.ts
tiny-ben-tran: I would choose the current code because it is slightly faster

TheSpyder: I just don't like the style of passing `Class.add` and `Class.remove` as first-class arguments

PR #8414, opened 2023-01-31
TINY-9556: fix flaky test ToolbarDrawerToggleTest.ts

hamza0867:
Related Ticket: TINY-9556

Description of Changes:
* Fix for flaky test ToolbarDrawerToggleTest.ts

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/api/ThemeManager.ts
+++ b/modules/tinymce/src/core/main/ts/api/ThemeManager.ts
@@ -16,7 +16,7 @@ export interface Theme {
   execCommand?: (command: string, ui?: boolean, value?: any) => boolean;
   destroy?: () => void;
   init?: (editor: Editor, url: string) => void;
-  renderUI?: () => RenderResult;
+  renderUI?: () => Promise<RenderResult>;
   getNotificationManagerImpl?: () => NotificationManagerImpl;
   getWindowManagerImpl?: () => WindowManagerImpl;
 }
--- a/modules/tinymce/src/core/main/ts/api/ThemeManager.ts
+++ b/modules/tinymce/src/core/main/ts/api/ThemeManager.ts
@@ -16,7 +16,7 @@ export interface Theme {
   execCommand?: (command: string, ui?: boolean, value?: any) => boolean;
   destroy?: () => void;
   init?: (editor: Editor, url: string) => void;
-  renderUI?: () => Promise<RenderResult>;
+  renderUI?: () => Promise<RenderResult> | RenderResult;
   getNotificationManagerImpl?: () => NotificationManagerImpl;
   getWindowManagerImpl?: () => WindowManagerImpl;
 }
--- a/modules/tinymce/src/core/main/ts/init/Init.ts
+++ b/modules/tinymce/src/core/main/ts/init/Init.ts
@@ -168,14 +168,14 @@ const augmentEditorUiApi = (editor: Editor, api: Partial<EditorUiApi>) => {
   editor.ui = { ...editor.ui, ...uiApiFacade };
 };
 
-const init = (editor: Editor): void => {
+const init = async (editor: Editor): Promise<void> => {
   editor.dispatch('ScriptsLoaded');
 
   initIcons(editor);
   initTheme(editor);
   initModel(editor);
   initPlugins(editor);
-  const renderInfo = renderThemeUi(editor);
+  const renderInfo = await renderThemeUi(editor);
   augmentEditorUiApi(editor, Optional.from(renderInfo.api).getOr({}));
   editor.editorContainer = renderInfo.editorContainer as HTMLElement;
   appendContentCssFromSettings(editor);
--- a/modules/tinymce/src/core/test/ts/browser/init/EditorCustomThemeTest.ts
+++ b/modules/tinymce/src/core/test/ts/browser/init/EditorCustomThemeTest.ts
@@ -1,47 +1,141 @@
-import { describe, it } from '@ephox/bedrock-client';
+import { context, describe, it } from '@ephox/bedrock-client';
 import { TinyHooks } from '@ephox/wrap-mcagar';
 import { assert } from 'chai';
 
 import Editor from 'tinymce/core/api/Editor';
+import { Theme } from 'tinymce/core/api/ThemeManager';
+import { TinyMCE } from 'tinymce/core/api/Tinymce';
+
+declare const tinymce: TinyMCE;
 
 describe('browser.tinymce.core.init.EditorCustomThemeTest', () => {
-  const hook = TinyHooks.bddSetupLight<Editor>({
-    add_unload_trigger: false,
-    disable_nodechange: true,
-    automatic_uploads: false,
-    entities: 'raw',
-    indent: false,
-    base_url: '/project/tinymce/js/tinymce',
-    theme: (editor: Editor, targetnode: HTMLElement) => {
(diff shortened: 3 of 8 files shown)


modules/tinymce/src/core/main/ts/api/ThemeManager.ts
```diff
@@ -16,7 +16,7 @@ export interface Theme {
   execCommand?: (command: string, ui?: boolean, value?: any) => boolean;
   destroy?: () => void;
   init?: (editor: Editor, url: string) => void;
-  renderUI?: () => RenderResult;
+  renderUI?: () => Promise<RenderResult>;
```
spocke, line 19: This needs to be sync and async to not break the public api but it can just be wrapped in a promise if it's a sync value.

```suggestion
  renderUI?: () => Promise<RenderResult> | RenderResult;
```

- [ ] Add fallback to sync api
- [ ] Add test for this new async and sync api by creating a custom themes in the tests and init with those
- [ ] Add changelog items for API change and load/render order change

spocke (approved):

TheSpyder (changes requested):
Sorry, adding async in this way is a breaking change which we can't do outside of a major release. I don't know if any of our customers _actually_ implement a custom theme, but it's quite clearly a public API.

Even if this is the best way forward we need to find some other way to improve the test consistency for now.

modules/tinymce/src/core/main/ts/init/Init.ts
```diff
@@ -168,14 +168,14 @@ const augmentEditorUiApi = (editor: Editor, api: Partial<EditorUiApi>) => {
...
-const init = (editor: Editor): void => {
+const init = async (editor: Editor): Promise<void> => {
   editor.dispatch('ScriptsLoaded');
 
   initIcons(editor);
   initTheme(editor);
   initModel(editor);
   initPlugins(editor);
-  const renderInfo = renderThemeUi(editor);
+  const renderInfo = await renderThemeUi(editor);
```
spocke, line 178: I would have expected the `renderThemeUi` to have some form of Promise check with branching and wrapping it in a promise, does await work on sync functions?

hamza0867: Yes it does 😆

spocke: Such a bizarre language.

modules/tinymce/src/core/test/ts/browser/init/EditorCustomThemeTest.ts
```diff
@@ -1,47 +1,141 @@
...
-    }
-  }, []);
-
-  it('getContainer/getContentAreaContainer', () => {
-    const editor = hook.editor();
-    assert.equal(editor.getContainer().id, 'editorContainer', 'Should be the new editorContainer element');
-    assert.equal(editor.getContainer().nodeType, 1, 'Should be an element');
-    assert.equal(editor.getContentAreaContainer().id, 'iframeContainer', 'Should be the new iframeContainer element');
-    assert.equal(editor.getContentAreaContainer().nodeType, 1, 'Should be an element');
+    it('The skin loads before editor finshes initializing', () => {
```
spocke, line 135:
```suggestion
    it('The skin loads before editor finishes initializing', () => {
```

modules/tinymce/src/core/test/ts/browser/init/EditorCustomThemeTest.ts
```diff
@@ -1,47 +1,141 @@
...
+  context('named asynchronous custom theme', () => {
+
+    const AsyncTheme = () => {
+      tinymce.ThemeManager.add('sync_theme', (editor): Theme => {
+        const theme: Theme = {
+          renderUI: async () => {
+            const editorContainer = document.createElement('div');
+            editor.getElement().after(editorContainer);
+            const promise = new Promise((resolve, _reject) => {
+              setTimeout(() => resolve(editor.dispatch('SkinLoaded')), 1000);
```
spocke, line 105: Maybe lower this to 10 or even 0 just needs a bit of a delay to make un async.

hamza0867: Done!

modules/tinymce/src/core/main/ts/api/ThemeManager.ts
```diff
@@ -16,7 +16,7 @@ export interface Theme {
   execCommand?: (command: string, ui?: boolean, value?: any) => boolean;
   destroy?: () => void;
   init?: (editor: Editor, url: string) => void;
-  renderUI?: () => RenderResult;
+  renderUI?: () => Promise<RenderResult> | RenderResult;
```
TheSpyder, line 19: Perhaps I wasn't clear before. I'm not changing my mind about this without a much wider discussion. It's breaking a public API which we can't do until a major release.

Our `tinymce.d.ts` currently contains:
```
interface Theme {
    ui?: any;
    inline?: any;
    execCommand?: (command: string, ui?: boolean, value?: any) => boolean;
    destroy?: () => void;
    init?: (editor: Editor, url: string) => void;
    renderUI?: () => {
        iframeContainer?: HTMLIFrameElement;
        editorContainer: HTMLElement;
        api?: Partial<EditorUiApi>;
    };
    getNotificationManagerImpl?: () => NotificationManagerImpl;
    getWindowManagerImpl?: () => WindowManagerImpl;
}
```

And after this change:
```
interface RenderResult {
    iframeContainer?: HTMLElement;
    editorContainer: HTMLElement;
    api?: Partial<EditorUiApi>;
}
interface Theme {
    ui?: any;
    inline?: any;
    execCommand?: (command: string, ui?: boolean, value?: any) => boolean;
    destroy?: () => void;
    init?: (editor: Editor, url: string) => void;
    renderUI?: () => Promise<RenderResult> | RenderResult;
    getNotificationManagerImpl?: () => NotificationManagerImpl;
    getWindowManagerImpl?: () => WindowManagerImpl;
}
```

Will this break any customer integrations? Almost certainly not. I would be very surprised if _anyone_ is relying on the old type. But it's still a breaking change under the strict semver rules.

spocke: Should we risk it here since it needs to be a customer who uses typescript and has custom themes that can't be that many people. Also it will not break cloud since both sync and async would work.

But if they have some predicate that acts on the output of the renderUI and expecting that to be only a RenderResult then it will fail to compile. But it would be fairly obvious how to fix that.

spocke: I think we are fine we are breaking types every minor release it seems. Sometimes we go from 'something' to 'something | null' etc in public apis that is what was breaking the `casechange` plugin for example that was bumped from 6.0.0 to 6.3.1 and it seems that we are doing that quite often not seen any complaints so I think it's ok to do this change here it's even less risky than the other things we broken before in other minors since it's a fringe api.

TheSpyder: At the end of the day semver is best described as an agreement, not an enforceable rule. The definition of what constitutes a breaking change is also debatable.

If we've decided to relax a bit based on what we think will actually break customer integrations that's fine, but since December 2019 we've practiced strict semver. I just wanted to point out that this doesn't fit that definition.

TheSpyder (approved):

PR #8424, opened 2023-02-06
TINY-9244: Parse templates before rendering

mkzhx:
Related Ticket: TINY-9244

Description of Changes:
* Parse html when retrieving template content for preview and insertion

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/tinymce/src/plugins/template/test/ts/module/InsertTemplate.ts
```diff
@@ -8,16 +8,37 @@ import Editor from 'tinymce/core/api/Editor';
 const toolbarButtonSelector = '[role="toolbar"] button[aria-label="Insert template"]';
 const dialogSelector = 'div.tox-dialog';
 
-const pInsertTemplate = async (editor: Editor, assertFn?: (elm: SugarElement<Node>) => Promise<void>): Promise<void> => {
+const waitUntilIframeLoaded = async (dialogEl: SugarElement<Node>): Promise<void> => {
+  await UiFinder.pWaitForState<HTMLIFrameElement>('iframe is loaded', dialogEl, 'iframe', (elm) => {
+    const iframeDoc = elm.dom.contentDocument || elm.dom.contentWindow?.document;
```
spocke, line 13: Why are there two cases here will not `contentDocument` work on all browsers? If not add a comment why and on what.

modules/tinymce/src/plugins/template/test/ts/browser/TemplateSanityTest.ts
```diff
@@ -78,4 +75,27 @@ describe('browser.tinymce.plugins.template.TemplateSanityTest', () => {
...
+      templates: [{ title: 'a', description: 'b', content: unparsedHtml }],
+    });
+
+    const unParsedPreviewHtmlSelector = 'p > img[src="error"][onerror="console.log(`Unparsed html`);"]';
+    const parsedPreviewHtmlSelector = 'p > img[src="error"][data-mce-src="error"]';
+    const assertPreviewContent = (dialogEl: SugarElement<Node>, existsSelector: string, notExistsSelector: string): void => {
+      UiFinder.findIn<HTMLIFrameElement>(dialogEl, 'iframe').fold(
+        () => assert.fail('Preview iframe not found'),
+        (iframe) => {
+          const iframeDoc = iframe.dom.contentDocument || iframe.dom.contentWindow?.document;
```
spocke, line 92: Why are there two cases here will not contentDocument work on all browsers? If not add a comment why and on what.

spocke (approved):

modules/tinymce/src/plugins/template/test/ts/module/InsertTemplate.ts
mkzhx: The code was there before my change. I assume this is because contentDocument isn't supported on pre- IE8 versions of IE. Added comment to document that.

modules/tinymce/src/plugins/template/test/ts/browser/TemplateSanityTest.ts
mkzhx: The code was there before my change. I assume this is because contentDocument isn't supported on pre- IE8 versions of IE. Added comment to document that.

spocke: Oh is it that old then get rid of it we don't support IE at all anymore.

modules/tinymce/src/plugins/template/test/ts/browser/TemplateSanityTest.ts
```diff
@@ -78,4 +75,65 @@ describe('browser.tinymce.plugins.template.TemplateSanityTest', () => {
...
+  const assertFnDoesNotReadUnsanitizedHtmlInDom = async (editor: Editor, fn: (unsanitizedHtml: string) => void | Promise<void>): Promise<void> => {
+    const fnReadsUnsanitizedHtmlInDom = async () => {
+      const unsanitizedHtml = '<img src="error" onerror="window.document.unsanitizedHtmlFn();">';
+      let isUnsanitizedHtmlRead = false;
+      (editor.getDoc() as any).unsanitizedHtmlFn = () => {
+        isUnsanitizedHtmlRead = true;
+      };
+      await fn(unsanitizedHtml);
+      // wait for unsanitized html to be read and error to be thrown
+      await Waiter.pWait(1000);
```
spocke, line 116: Change this to a waiter. Delaying here for 1 second seems wasteful.

spocke: No wait this is one of those wait for something not to happen the worst case. :cry:
I wonder is it super important to have this test I don't think we have similar tests for the parser for example that could in theory do similar things.

mkzhx: This function is used by 2 only tests that verifies that unsanitized HTML is not read during template insertion, so my concern is that our testing may not catch any potential regressions if I remove these 2 tests. On the plus side, I find that as long as the `pWait` time is non-zero, ex: 1ms, the function is reliable enough to catch any error thrown by the unsanitized HTML. Is `pWait(1)` good enough?

MitchCTiny (approved):

PR #8468, opened 2023-02-20
TINY-9399: Fix removing annotation immediately after creation

yacodes:
Related Ticket: TINY-9399

Description of Changes:
* It is possible to remove an annotation immediately after creation without any selection changes.

The bug was introduced in TINY-8195 and connected to [this](https://github.com/tinymce/tinymce/pull/7485/files#diff-3882b4c135273323573f94b4f8e6a895594701e68d80588eb3749f9567e5502cL89-L92) bookmark, which adjusts the selection in the way that it is impossible to retain the selected word. 

Everything works fine by using offset bookmark `getBookmark(2)`, but I do not really understand the differences between different bookmark types (and only the default one is used in Tiny's code). Other possible solution is in this PR.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -66,6 +66,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Inserting newlines inside an editable element inside a noneditable root would sometimes try to split the editable element. #TINY-9461
 - Creating a list in a table cell when the caret is in front of an anchor element would not properly include the anchor in the list. #TINY-6853
 - Formatting could be applied or removed on noneditable list items inside a noneditable root. #TINY-9563
+- Removing annotation is possible immediately after creation without any selection changes. #TINY-9399
 
 ## 6.3.1 - 2022-12-06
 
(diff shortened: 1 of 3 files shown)


lorenzo-pomili (approved):

modules/tinymce/CHANGELOG.md
```diff
@@ -66,6 +66,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Inserting newlines inside an editable element inside a noneditable root would sometimes try to split the editable element. #TINY-9461
 - Creating a list in a table cell when the caret is in front of an anchor element would not properly include the anchor in the list. #TINY-6853
 - Formatting could be applied or removed on noneditable list items inside a noneditable root. #TINY-9563
+- Removing annotation is possible immediately after creation without any selection changes. #TINY-9399
```
vpyshnenko, line 69:
```suggestion
- Annotation would not be removed if the conversation is immediately deleted after the creation.. #TINY-9399
```

vpyshnenko (approved):

TheSpyder (approved):

modules/tinymce/CHANGELOG.md
spocke: Not sure we should mention `conversation` in the changelog of tinymce since that is comment specific.

Maybe something like this:
```suggestion
- Annotation would not be removed if the annotation was immediately deleted after being created. #TINY-9399
```

spocke (approved):

modules/tinymce/CHANGELOG.md
yacodes: Thanks! Updated with the suggestions.

PR #8480, opened 2023-02-23
TINY-9600: Provide configuration to turn off XSS sanitization

mkzhx:
Related Ticket: TINY-9600

Description of Changes:
* Add `xss_sanitization: boolean` option (default: `true`) to enable/disable XSS sanitization in the parser
* Make all content-related behaviour, including `setContent`, `getContent`, initialising content, and pasting content, respect `xss_sanitization` option

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/api/html/DomParser.ts
+++ b/modules/tinymce/src/core/main/ts/api/html/DomParser.ts
@@ -56,18 +56,19 @@ export interface DomParserSettings {
   allow_html_in_named_anchor?: boolean;
   allow_script_urls?: boolean;
   allow_unsafe_link_target?: boolean;
+  blob_cache?: BlobCache;
   convert_fonts_to_spans?: boolean;
+  document?: Document;
   fix_list_elements?: boolean;
   font_size_legacy_values?: string;
   forced_root_block?: boolean | string;
   forced_root_block_attrs?: Record<string, string>;
+  inline_styles?: boolean;
   preserve_cdata?: boolean;
   remove_trailing_brs?: boolean;
   root_name?: string;
+  sanitize?: boolean;
   validate?: boolean;
-  inline_styles?: boolean;
-  blob_cache?: BlobCache;
-  document?: Document;
 }
 
 interface DomParser {
@@ -408,11 +409,11 @@ const DomParser = (settings: DomParserSettings = {}, schema = Schema()): DomPars
   const defaultedSettings = {
     validate: true,
     root_name: 'body',
+    sanitize: true,
     ...settings
   };
 
   const parser = new DOMParser();
-  const purify = setupPurify(defaultedSettings, schema);
 
   const parseAndSanitizeWithContext = (html: string, rootName: string, format: string = 'html'): Element => {
     const mimeType = format === 'xhtml' ? 'application/xhtml+xml' : 'text/html';
@@ -425,8 +426,11 @@ const DomParser = (settings: DomParserSettings = {}, schema = Schema()): DomPars
     const body = parser.parseFromString(wrappedHtml, mimeType).body;
 
     // Sanitize the content
-    purify.sanitize(body, getPurifyConfig(defaultedSettings, mimeType));
-    purify.removed = [];
+    if (defaultedSettings.sanitize) {
+      const purify = setupPurify(defaultedSettings, schema);
+      purify.sanitize(body, getPurifyConfig(defaultedSettings, mimeType));
+      purify.removed = [];
+    }
 
     return isSpecialRoot ? body.firstChild as Element : body;
   };
--- a/modules/tinymce/src/core/main/ts/api/html/DomParser.ts
+++ b/modules/tinymce/src/core/main/ts/api/html/DomParser.ts
@@ -414,6 +414,7 @@ const DomParser = (settings: DomParserSettings = {}, schema = Schema()): DomPars
   };
 
   const parser = new DOMParser();
+  const purify = setupPurify(defaultedSettings, schema);
 
   const parseAndSanitizeWithContext = (html: string, rootName: string, format: string = 'html'): Element => {
     const mimeType = format === 'xhtml' ? 'application/xhtml+xml' : 'text/html';
@@ -427,7 +428,6 @@ const DomParser = (settings: DomParserSettings = {}, schema = Schema()): DomPars
 
     // Sanitize the content
     if (defaultedSettings.sanitize) {
-      const purify = setupPurify(defaultedSettings, schema);
       purify.sanitize(body, getPurifyConfig(defaultedSettings, mimeType));
       purify.removed = [];
     }
--- a/modules/tinymce/src/core/main/ts/content/PrePostProcess.ts
+++ b/modules/tinymce/src/core/main/ts/content/PrePostProcess.ts
@@ -11,7 +11,7 @@ import { Content, GetContentArgs, isTreeNode, SetContentArgs } from './ContentTy
 const serializeContent = (content: Content): string =>
   isTreeNode(content) ? HtmlSerializer({ validate: false }).serialize(content) : content;
 
-const withSerializedContent = <R extends EditorEvent<{ content: string }>>(editor: Editor, content: Content, fireEvent: (content: string) => R): R & { content: Content } => {
+const withSerializedContent = <R extends EditorEvent<{ content: string }>>(content: Content, fireEvent: (content: string) => R, sanitize: boolean): R & { content: Content } => {
   const serializedContent = serializeContent(content);
   const eventArgs = fireEvent(serializedContent);
   if (eventArgs.isDefaultPrevented()) {
@@ -20,7 +20,7 @@ const withSerializedContent = <R extends EditorEvent<{ content: string }>>(edito
     // Restore the content type back to being an AstNode. If the content has changed we need to
     // re-parse the new content, otherwise we can return the input.
     if (eventArgs.content !== serializedContent) {
-      const rootNode = DomParser({ validate: false, forced_root_block: false, sanitize: Options.shouldSanitizeXss(editor) }).parse(eventArgs.content, { context: content.name });
+      const rootNode = DomParser({ validate: false, forced_root_block: false, sanitize }).parse(eventArgs.content, { context: content.name });
       return { ...eventArgs, content: rootNode };
     } else {
       return { ...eventArgs, content };
@@ -47,7 +47,7 @@ const postProcessGetContent = <T extends GetContentArgs>(editor: Editor, content
   if (args.no_events) {
     return content;
   } else {
-    const processedEventArgs = withSerializedContent(editor, content, (c) => Events.fireGetContent(editor, { ...args, content: c }));
+    const processedEventArgs = withSerializedContent(content, (content) => Events.fireGetContent(editor, { ...args, content }), Options.shouldSanitizeXss(editor));
     return processedEventArgs.content;
   }
 };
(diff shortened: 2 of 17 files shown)


modules/tinymce/src/core/main/ts/api/html/DomParser.ts
```diff
@@ -425,8 +426,11 @@ const DomParser = (settings: DomParserSettings = {}, schema = Schema()): DomPars
     const body = parser.parseFromString(wrappedHtml, mimeType).body;
 
     // Sanitize the content
-    purify.sanitize(body, getPurifyConfig(defaultedSettings, mimeType));
-    purify.removed = [];
+    if (defaultedSettings.sanitize) {
+      const purify = setupPurify(defaultedSettings, schema);
```
TheSpyder, line 430: I'm not sure about moving the setup code in here. I guess if it causes performance issues we can deal with it later.

modules/tinymce/src/core/main/ts/content/PrePostProcess.ts
```diff
@@ -4,13 +4,14 @@ import Editor from '../api/Editor';
...
 import HtmlSerializer from '../api/html/Serializer';
+import * as Options from '../api/Options';
 import { EditorEvent } from '../api/util/EventDispatcher';
 import { Content, GetContentArgs, isTreeNode, SetContentArgs } from './ContentTypes';
 
 const serializeContent = (content: Content): string =>
   isTreeNode(content) ? HtmlSerializer({ validate: false }).serialize(content) : content;
 
-const withSerializedContent = <R extends EditorEvent<{ content: string }>>(content: Content, fireEvent: (content: string) => R): R & { content: Content } => {
+const withSerializedContent = <R extends EditorEvent<{ content: string }>>(editor: Editor, content: Content, fireEvent: (content: string) => R): R & { content: Content } => {
```
TheSpyder, line 14: Please don't pass the editor to a method unless it's _really_ needed - keeping the scope of method arguments to a minimum is a technique that can help to reduce accidental dependencies from within the method, aka spaghetti code. It also helps with testing (although I'm not sure that really matters here, just a general point).

In this case, just pass the sanitize option value - it might look like more code to do that, but after minification the difference is negligible.

- [ ] Change method arguments

TheSpyder (approved):

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
spocke: Maybe it could be pulled out to a function like `runDomPurify` or something would make it easier to test that in isolation if we wanted to.

modules/tinymce/src/core/test/ts/browser/content/EditorContentTest.ts
```diff
@@ -17,16 +18,66 @@ const defaultExpectedEvents = [
 ];
 
 describe('browser.tinymce.core.content.EditorContentTest', () => {
+  const toHtml = (node: AstNode): string => {
+    const htmlSerializer = HtmlSerializer({});
+    return htmlSerializer.serialize(node);
+  };
```
spocke, line 24: Do we need the extra variable for something. I think this is shorter and readable? I know that the original function had it as well. So up to you what you prefer.

```suggestion
  const toHtml = (node: AstNode): string => HtmlSerializer({}).serialize(node);
```

modules/tinymce/src/core/test/ts/browser/content/EditorContentTest.ts
```diff
@@ -17,16 +18,66 @@ const defaultExpectedEvents = [
...
+        inline: true,
+        xss_sanitization: false
+      },
+      {
+        inline: false,
+        xss_sanitization: false
       }
     ], (options) => {
-      context('Test with inline option ' + options.inline, () => {
+      context('Test with inline: ' + options.inline + ' and xss_sanitization: ' + options.xss_sanitization, () => {
```
spocke, line 80: Maybe convert to template string while you are at it? But feel free to skip it since it's mostly code style.

```suggestion
      context(`Test with inline: ${options.inline} and xss_sanitization: ${options.xss_sanitization}`, () => {
```

modules/tinymce/src/core/test/ts/browser/content/EditorContentTest.ts
```diff
@@ -291,4 +309,112 @@ describe('browser.tinymce.core.content.EditorContentTest', () => {
...
     }
   );
+
+  Arr.each([
+    { inline: true },
+    { inline: false }
+  ], (options) => {
+    const unsanitizedHtml = '<p><a href="javascript:alert(1)">XSS</a></p>';
+
+    context('TINY-9600: Test unsanitized content with inline: ' + options.inline + ' and xss_sanitization: true', () => {
```
spocke, line 319:
```suggestion
    context(`TINY-9600: Test unsanitized content with inline: ${options.inline} + ' and xss_sanitization: true`, () => {
```

modules/tinymce/src/core/test/ts/browser/content/EditorContentTest.ts
```diff
@@ -291,4 +309,112 @@ describe('browser.tinymce.core.content.EditorContentTest', () => {
...
+        const content = editor.getContent();
+        assert.equal(content, unsanitizedHtml, 'Unsanitized html should not be altered');
+      });
+
+      it('getContent html with iframe with child node should get the content as expected and not error', () => {
+        const editor = hook.editor();
+        editor.setContent('<p><iframe><p>test</p></iframe></p>');
+        const content = editor.getContent();
+        assert.equal(content,
+          PlatformDetection.detect().browser.isSafari()
```
spocke, line 390: This is pretty strange this should be investigated in a follow up. If the editors don't produce the same HTML on save it could produce very strange errors if say you diff the initial value with the current value and you made no changes depending on platform it would be dirty vs not dirty. This should be normalized in the parser/serializer code. Maybe with a custom filter. Log a follow up ticket.

- [ ] Log ticket to investigate safari specific html output

modules/tinymce/src/core/test/ts/browser/html/DomParserTest.ts
```diff
@@ -1487,4 +1487,34 @@ describe('browser.tinymce.core.html.DomParserTest', () => {
...
+        assert.equal(serializedHtml, testCase.input);
+      });
+    };
+
+    Arr.each([{
+      name: 'should not remove script tags',
+      input: '<script>alert(1)</script>'
+    }, {
+      name: 'should not remove iframe tags with child nodes',
+      input: '<iframe src="https://example.com"><p>Lorem ipsum</p></iframe>'
```
spocke, line 1506: Hmm, the Safari entity encoding branch case is not happening here. So is the editor somehow doing special filtering only for Safari. Or is dompurifier doing something here since it's disabled for this test.

spocke (approved):
Good work on this one. :+1:

modules/tinymce/src/core/main/ts/content/PrePostProcess.ts
mkzhx: Thanks for the detailed feedback, I've made the change and noted it for the future.

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
mkzhx: Agreed, I should've seen that the setupPurify function is now being run every single time content is parsed instead of just when the DomParser is defined like before.

modules/tinymce/src/core/test/ts/browser/content/EditorContentTest.ts
mkzhx: Thanks for catching that, definitely prefer the shortened version.

modules/tinymce/src/core/test/ts/browser/content/EditorContentTest.ts
mkzhx: Converted all concatenated strings to template strings.

modules/tinymce/src/core/test/ts/browser/content/EditorContentTest.ts
mkzhx: Created ticket TINY-9624 to investigate Safari-specific HTML output.

modules/tinymce/src/core/test/ts/browser/html/DomParserTest.ts
mkzhx: To be investigated in TINY-9624.

modules/tinymce/src/core/test/ts/browser/content/InsertUnsanitizedContentTest.ts
```diff
@@ -0,0 +1,40 @@
...
+    editor.setContent(initialContent);
+    TinySelections.setCursor(editor, [ 0, 0 ], 0);
+    editor.insertContent(content);
+    TinyAssertions.assertContent(editor, `${expected}\n${initialContent}`);
+  };
+
+  context('TINY-9600: Inserting unsanitized html with xss_sanitization: true', () => {
+    const hook = TinyHooks.bddSetupLight<Editor>({
+      base_url: '/project/tinymce/js/tinymce',
+      xss_sanitization: true
```
tiny-ben-tran, line 20:
```suggestion
```
Since this is `true` by default?

modules/tinymce/src/core/test/ts/browser/content/InsertUnsanitizedContentTest.ts
```diff
@@ -0,0 +1,40 @@
...
+    }, []);
+
+    it('insertContent should sanitize inserted html', () => {
+      const editor = hook.editor();
+      testInsertContent(editor, unsanitizedHtml, sanitizedHtml);
```
tiny-ben-tran, line 25:
```suggestion
      testInsertContent(hook.editor(), unsanitizedHtml, sanitizedHtml);
```

Very minor

modules/tinymce/src/core/test/ts/browser/content/InsertUnsanitizedContentTest.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  context('TINY-9600: Inserting unsanitized html with xss_sanitization: false', () => {
+    const hook = TinyHooks.bddSetupLight<Editor>({
+      base_url: '/project/tinymce/js/tinymce',
+      xss_sanitization: false
+    }, []);
+
+    it('insertContent should not alter inserted unsanitized html', () => {
+      const editor = hook.editor();
+      testInsertContent(editor, unsanitizedHtml, unsanitizedHtml);
```
tiny-ben-tran, line 37:
```suggestion
      testInsertContent(hook.editor(), unsanitizedHtml, unsanitizedHtml);
```
I guess less is better

modules/tinymce/src/core/test/ts/browser/content/InsertUnsanitizedContentTest.ts
mkzhx: Good point, it's useful to have the test ensure that `xss_sanitization` is actually set to true by default.

modules/tinymce/src/core/test/ts/browser/content/InsertUnsanitizedContentTest.ts
mkzhx: Addressed in another PR.

modules/tinymce/src/core/test/ts/browser/content/InsertUnsanitizedContentTest.ts
mkzhx: Addressed in followup PR.

modules/tinymce/src/core/test/ts/browser/content/InsertUnsanitizedContentTest.ts
mkzhx: @tiny-ben-tran On second thoughts, not setting `xss_sanitization` would have the same effect as existing tests that test editor content sanitization. This test is here to ensure that the new option works as it is intended, so that we don't have logic that disables sanitization when `xss_sanitization` is set to `true` as well.

PR #8481, opened 2023-02-23
TINY-9520: Add top-right close 'x' button of the modal dialogs to keyboard navigation

shanmen-tiny:
Related Ticket: TINY-9520

Description of Changes:
* Added `Tabbing.config` to top right close 'x' button
* Added `firstTabstop` optional property to specify the index of elements to focus in when dialog shows
	* To address issues when adding `Tabbing.config` to the top right close 'x' button, the focus is transferred to the close button when dialog shows, for dialogs with a close button and without a close button.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/themes/silver/main/ts/ui/dialog/Dialogs.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/ui/dialog/Dialogs.ts
@@ -128,7 +128,7 @@ const renderDialogWithHeader = (spec: DialogSpec): SketchSpec => {
   return renderDialog({ ...spec, firstTabstop: 1 });
 };
 
-const renderDialogWithoutHeader = (spec: DialogSpec): SketchSpec => {
+const renderDialogWithHiddenHeader = (spec: DialogSpec): SketchSpec => {
   return renderDialog(spec);
 };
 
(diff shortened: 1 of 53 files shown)


modules/tinymce/src/themes/silver/main/ts/ui/dialog/Dialogs.ts
```diff
@@ -121,8 +121,17 @@ export interface DialogSpec {
...
 }
 
+const renderDialogWithHeader = (spec: DialogSpec): SketchSpec => {
+  return renderDialog({ ...spec, firstTabstop: 1 });
+};
+
+const renderDialogWithHiddenHeader = (spec: DialogSpec): SketchSpec => {
+  return renderDialog(spec);
+};
+
```
TheSpyder, line 134: Can't we just add `firstTabstop: 1` to `SilverDialogCommon` instead of changing `Dialog` directly?

Nothing about these methods guarantees there is a close button or not, as the name implies. So we might as well leave control in the hands of the calling function.

- [ ] consider simplifying

TheSpyder (approved):
Seems ok to me. I'm keen to see other opinions though.

modules/tinymce/src/themes/silver/main/ts/ui/dialog/Dialogs.ts
shanmen-tiny: Oh yeah, thanks for the suggestion. We could pass `firstTabstop: 1` from `SilverDialogCommon` and omit it if it was called from a function that doesn't require setting `firstTabstop`. Will make the changes and revert back to a single function in `Dialogs.ts`.

spocke (approved):

PR #8490, opened 2023-02-27
TINY-8363: Fix tests previously skipped due to false-positive mXSS detection by DOMPurify

mkzhx:
Related Ticket: TINY-8363

Description of Changes:
* Some tests were previously skipped due to false-positive mXSS detection by DOMPurify
* These tests are fixed by disabling DOMPurify sanitization with new option introduced in TINY-9600

Pre-checks:
* [x] ~Changelog entry added~ - only tests changed

Review:
* [x] Milestone set


modules/tinymce/src/core/test/ts/browser/dom/SerializerTest.ts
```diff
@@ -334,27 +337,27 @@ describe('browser.tinymce.core.dom.SerializerTest', () => {
...
     );
   });
 
-  // TODO: TINY-4627/TINY-8363
-  it.skip('Script with tags inside a comment', () => {
-    const ser = DomSerializer({ fix_list_elements: true });
-    ser.setRules('script[type|language|src]');
+  it('Script with tags inside a comment with sanitize: false', () => {
+    // TINY-8363: Disable sanitization to avoid DOMPurify false positive affecting expected output
+    const ser = DomSerializer({ fix_list_elements: true, sanitize: false });
```
spocke, line 354: So just to be clear the schema now always allows script elements but it's the dompurify that decides to strip them. Since this test seems to be altered from the original to that.

But is seems to still be deleted here:
https://github.com/tinymce/tinymce/blob/develop/modules/tinymce/src/core/main/ts/api/html/Schema.ts#L406

So why was the `ser.setRules('script[type|language|src]');` line removed I would expect that to still be required?

I would expect that disabling dompurify should have not effect on the schema and the test in this form should fail.

I don't think it should work like this setting `xss_sanitization` to false now allows invalid attributes or elements to be added. To me only scripts and event attributes should be allowed not any type of attribute or elements. See fiddle: https://fiddle.tiny.cloud/Pziaab compare dev to staging. So it would need some form of fallback filtering towards the schema if scripts and things are allowed if purifier was doing this. Or the option should be renamed to `validate` or something since it now bypassing the schema rules.

- [ ] Verify that this is the new behavior we want

mkzhx: @spocke Thanks for your input, this ticket also made me wary of such issues - currently a large portion of node manipulation and schema validation code is in DOMPurify hook functions, which are not run if DOMPurify sanitization is disabled.
https://github.com/tinymce/tinymce/blob/develop/modules/tinymce/src/core/main/ts/api/html/DomParser.ts#L124
https://github.com/tinymce/tinymce/blob/develop/modules/tinymce/src/core/main/ts/api/html/DomParser.ts#L200
The limitations of disabling DOMPurify is currently in being discussed by engineering and product. Some limitations such as not removing bogus elements is quite significant. Drafting this PR until a decision is reached.

mkzhx: Schema validation and content filtering option issues addressed in TINY-9635 (merged 06/03/2023).

mkzhx: Reverted unintended test changes.

ArvinJ-H (approved):

spocke (approved):

MitchCTiny (approved):

PR #8499, opened 2023-02-28
TINY-9176: Improve contrast and focus styles of UI buttons

MitchCTiny:
Related Ticket: TINY-9176

Description of Changes:
* Improved contrast of toolbar and menubar buttons in the focus, hover, active, and enabled states

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


MitchCTiny:
Closed https://github.com/tinymce/tinymce/pull/8396 to reopen as owner.

modules/oxide/src/less/theme/components/form/textfield.less
```diff
@@ -21,7 +21,7 @@
 @textfield-disabled-text-color: fade(@textfield-text-color, 85%);
 
 @textfield-focus-background-color: @textfield-background-color;
-@textfield-focus-border-color: @color-tint;
+@textfield-focus-border-color: #2d6adf;
```
spocke, line 24: This should likely gets it's value from somewhere else not be specific to the textfield.

modules/oxide/src/less/theme/components/form/textfield.less
```diff
@@ -56,7 +56,8 @@
   .tox-textfield:focus {
     background-color: @textfield-focus-background-color;
     border-color: @textfield-focus-border-color;
-    box-shadow: @textfield-focus-box-shadow;
+    border-radius: 6px;
```
spocke, line 59: Don't we have variables for this. Isn't there skins that have sharp borders? See if there is some variable that this could be using.

modules/oxide/src/less/theme/components/form/textfield.less
```diff
@@ -56,7 +56,8 @@
...
+    box-shadow: 0 0 0 1px @textfield-focus-border-color;
```
spocke, line 60: Why was this changed from @textfield-focus-box-shadow if we want to change the border here from 2px to 1px then it should alter the variable not a style directly.

spocke (changes requested):

modules/oxide/src/less/theme/components/form/textfield.less
MitchCTiny: I noticed this after making the PR, since I inherited the branch. I made a commit to revert some of these unnecessary changes but was trying to fix a test failure before pushing. I've pushed it now but I'll re-request your review once the tests are passing.

modules/oxide/src/less/theme/components/form/textfield.less
MitchCTiny: Reverted

modules/oxide/src/less/theme/components/form/textfield.less
MitchCTiny: Reverted

TheSpyder (approved):

spocke (approved):

PR #8500, opened 2023-03-01
TINY-9408: Fix sticky-toolbar transition in iframe mode

MitchCTiny:
Related Ticket: TINY-9408

Description of Changes:
* Fix transition of sticky-toolbar in inline mode
* Toolbar would disappear immediately, when it is meant to fade out

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/oxide/src/less/theme/components/editor/editor.less
+++ b/modules/oxide/src/less/theme/components/editor/editor.less
@@ -34,7 +34,7 @@
     box-shadow: @editor-header-box-shadow;
     padding: @pad-xs 0;
 
-    :not(.tox-editor-dock-transition) {
+    &:not(.tox-editor-dock-transition) {
       transition: box-shadow .5s;
     }
   }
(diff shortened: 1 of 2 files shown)


shanmen-tiny (approved):
Thanks for the quick PR. I have tested this locally and the sticky toolbar fades out in iframe mode.

spocke (approved):

spocke:
Does this require a changelog item? It would be something that a customer would notice?

MitchCTiny:
> Does this require a changelog item? It would be something that a customer would notice?
I'll push a changelog commit, I just need to figure out this test failure.

tiny-ben-tran (approved):

TheSpyder (approved):
I wonder if this breaks the box shadow transition. I guess if it does that'll only be during the fade out, so it shouldn't be particularly visible.

PR #8502, opened 2023-03-01
TINY-9603: Fix an invisible space after applying heading text_patterns

yacodes:
Related Ticket: TINY-9603

Description of Changes:
If there is a single space between pattern.start and text (e.g. `# 1`) then it will be left in the text content as an invisible one and then can be transformed to nbsp in certain circumstances. This is not an issue with multiple spaces because they are transformed to non-breaking ones straight away. I've decided to remove this single space whatsoever as it feels to be the expected behavior.

Other solution is to preserve this space by replacing it with `&nbsp;` so it will be visible to the user.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/textpatterns/core/BlockPattern.ts
+++ b/modules/tinymce/src/core/main/ts/textpatterns/core/BlockPattern.ts
@@ -1,4 +1,5 @@
 import { Arr, Obj, Optional, Type, Unicode } from '@ephox/katamari';
+import { SugarText, SugarElement } from '@ephox/sugar';
 
 import * as TextSearch from '../../alien/TextSearch';
 import DOMUtils from '../../api/dom/DOMUtils';
@@ -9,6 +10,8 @@ import Tools from '../../api/util/Tools';
 import { generatePathRange, resolvePathRange } from '../utils/PathRange';
 import * as Utils from '../utils/Utils';
 import { BlockPattern, BlockPatternMatch, Pattern, PatternSet } from './PatternTypes';
+
+const startsWithSingleSpace = (s: string): boolean => /^\s[^\s]/.test(s);
 
 const stripPattern = (dom: DOMUtils, block: Node, pattern: BlockPattern): void => {
   // The pattern could be across fragmented text nodes, so we need to find the end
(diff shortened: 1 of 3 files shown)


modules/tinymce/src/core/main/ts/textpatterns/core/BlockPattern.ts
```diff
@@ -32,7 +32,7 @@ const stripPattern = (dom: DOMUtils, block: Node, pattern: BlockPattern): void =
      * then it will be left in the text content and then can appear in certain circumstances.
      * This is not an issue with multiple spaces because they are transformed to non-breaking ones.
      *
-     * In this specific case we've decided to remove this single space whatsoever 
+     * In this specific case we've decided to remove this single space whatsoever
```
vpyshnenko, line 35: this self-describing correction looks funny  :)

modules/tinymce/src/core/main/ts/textpatterns/core/BlockPattern.ts
```diff
@@ -23,6 +26,20 @@ const stripPattern = (dom: DOMUtils, block: Node, pattern: BlockPattern): void =
...
+     * then it will be left in the text content and then can appear in certain circumstances.
+     * This is not an issue with multiple spaces because they are transformed to non-breaking ones.
+     *
+     * In this specific case we've decided to remove this single space whatsoever
+     * as it feels to be the expected behavior.
+     */
+    const text = SugarElement.fromDom(node);
+    const textContent = SugarText.get(text);
+    if (startsWithSingleSpace(textContent)) {
+      SugarText.set(text, textContent.replace(/^\s/, ''));
```
TheSpyder, line 41: Do we need regex for this, or can we just exclude the first character? `textContent.slice(1)` for example?

TheSpyder (approved):

modules/tinymce/src/core/main/ts/textpatterns/core/BlockPattern.ts
yacodes: Yeah, we can simplify it with `slice(`)`. Thanks! Updated.

lorenzo-pomili (approved):

spocke (approved):

PR #8513, opened 2023-03-03
TINY-9635: Use NodeIterator to perform node and attribute validation when sanitization is disabled

mkzhx:
Related Ticket: TINY-9635

Description of Changes:
Previously node and attribute validation code (incl. bogus element removal, schema validation, etc.) is attached to DOMPurify hooks
* With the new option `xss_sanitization` introduced in TINY-9600, this code is no longer run when DOMPurify sanitization is disabled with `xss_sanitization: false`

This PR reintroduces the functionality by using NodeIterator to iterate through the body to validate nodes and attributes when DOMPurify is disabled (`xss_sanitization: false`)
* We reuse the same logic as what was previously defined in the DOMPurify hooks where possible (attribute validation had to be slightly adapted)
* Sanitization code is also refactored into a new `Sanitization` module separate from `DomParser`
* Address bug in `isInvalidUri` where whitespaces in URIs prevented script/data regex from matching properly 

Pre-checks:
* [x] ~Changelog entry added~ - a follow-up to TINY-9600

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/api/html/DomParser.ts
+++ b/modules/tinymce/src/core/main/ts/api/html/DomParser.ts
@@ -458,9 +458,10 @@ const DomParser = (settings: DomParserSettings = {}, schema = Schema()): DomPars
       let uid = 0;
       let node;
       while ((node = nodeIterator.nextNode())) {
-        uid = processNode(node, defaultedSettings, schema, uid);
+        // disable schema validation to allow all elements and attributes including those not in the schema spec
+        uid = processNode(node, { ...defaultedSettings, validate: false }, schema, uid);
         if (NodeType.isElement(node)) {
-          filterAttributes(node, defaultedSettings.validate, schema);
+          filterAttributes(node, false, schema);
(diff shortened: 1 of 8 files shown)


modules/tinymce/src/core/main/ts/api/html/DomParser.ts
```diff
@@ -508,7 +357,7 @@ const DomParser = (settings: DomParserSettings = {}, schema = Schema()): DomPars
   };
 
   const isWrappableNode = (blockElements: SchemaMap, node: AstNode) => {
-    const isInternalElement = Type.isString(node.attr(internalElementAttr));
+    const isInternalElement = Type.isString(node.attr('data-mce-type'));
```
TheSpyder, line 360: Perhaps export this as a constant from `Sanitization` instead of dupliating it?

TheSpyder (approved):

MitchCTiny (approved):

PR #8521, opened 2023-03-06
TINY-9599: Incorrect event target for dragend and drop events when dragging and dropping noneditable elements within editor

spocke:
Related Ticket: TINY-9599

Description of Changes:
* Added target and other properties to dragstart, dragend and drop events that we fake.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


yacodes (approved):

vpyshnenko (approved):

lorenzo-pomili (approved):

modules/tinymce/src/core/main/ts/events/DragEvents.ts
```diff
@@ -0,0 +1,90 @@
...
+import { EditorEvent } from '../api/util/EventDispatcher';
+
+const makeDndEventFromMouseEvent = <K extends keyof MouseEvent>(type: string, mouseEvent: EditorEvent<MouseEvent>, extra: Record<K, MouseEvent[K]>): DragEvent => ({
+  ...mouseEvent,
+  dataTransfer: null, // We are not supporting dataTransfer yet but DragEvent is MouseEvent + dataTransfer so the property should exist
+  type,
+  ...extra
+});
+
+const makeDndEvent = <K extends keyof DragEvent>(type: string, props: Record<K, DragEvent[K]>): DragEvent => {
```
TheSpyder, line 12: This looks like code we normally only use in tests. In modern browsers, isn't it easier than this to create new events?

TheSpyder (approved):

modules/tinymce/src/core/main/ts/events/DragEvents.ts
spocke: Yes I did that at first but only some properties are readonly so I had to clone all properties by doing a ugly for in loop and do some unsafe casting to treat it as a record. Also there is a risk of the browsers warning about accessing deprecated properties when you access the values like that. How we do it on agar is that we make a custom event for dnd not dragevent instances and manually fill in some properties not all properties so doing it like agar would basically end up as the same thing as this, a big list of properties to fill in. Then later all these properties gets cloned anyway back to an object in the event dispatcher code. So having all functions returning the same EditorEvent type seems like the best idea. Make it a module that can make fake editor simulated event objects not fake native drag and drop events. This is also similar to how we fake the input events.

PR #8428, opened 2023-02-07
TINY-9414: Scrollable Containers - All Phases

shanmen-tiny:
Related Ticket: TINY-9414

Description of Changes:
* Contains changes from all 6 phases: 
  	* https://github.com/tinymce/tinymce/pull/8226
	* https://github.com/tinymce/tinymce/pull/8228
	* https://github.com/tinymce/tinymce/pull/8285
	* https://github.com/tinymce/tinymce/pull/8288
	* https://github.com/tinymce/tinymce/pull/8296
	* https://github.com/tinymce/tinymce/pull/8302
* Also, renamed `ui_of_tomorrow` in this branch (OptionTypes.ts)
* Updated the calculation of absolute position for the toolbar container with test cases (InlineToolbarPositionTest.ts, InlineHeader.ts)

Pre-checks:
* [x] Changelog entry added
* [x] Tests have been added (if applicable)
* [x] Branch prefixed with `feature/`, `hotfix/` or `spike/`

Review:
* [x] Milestone set
* [x] Docs ticket created (if applicable)

GitHub issues (if applicable):

PR #8526, opened 2023-03-08
TINY-9652: Hide toolbar when editor is out of view in a scrollable container

shanmen-tiny:
Related Ticket: TINY-9652

Description of Changes:
* Changed the logic of `updateVisibility` of the toolbar when `ui_mode: 'split'`
	* 	In a scrollable container, for toolbar_location: top and if the toolbar is shown and the scrollable container is scrolled to the top, the toolbar should be invisible (in this case, the toolbar isn't docked, so the code updating the visibility is not triggered)
	* 	In a scrollable container, for toolbar_location: bottom and if the toolbar is shown and the scrollable container is scrolled to the furthest down to the bottom, the toolbar should be invisible 

Pre-checks:
* [x] ~Changelog entry added~ found in TINY-9414

Review:
* [x] Milestone set

### DIFF
--- a/modules/alloy/src/main/ts/ephox/alloy/behaviour/docking/DockingApis.ts
+++ b/modules/alloy/src/main/ts/ephox/alloy/behaviour/docking/DockingApis.ts
@@ -82,17 +82,8 @@ const applyMorph = (
 const refreshInternal = (component: AlloyComponent, config: DockingConfig, state: DockingState): void => {
   // Absolute coordinates (considers scroll)
   const viewport: DockingViewport = config.lazyViewport(component);
-  // If docked then check if we need to hide/show the component
-  const isDocked = state.isDocked();
 
-  viewport.optScrollEnv.fold(() => {
-    if (isDocked) {
-      updateVisibility(component, config, state, viewport);
-    }
-  }, (_scrollEnv) => {
-    // If scrolling environment is set, we want to hide the toolbar if editor is out of the viewport
-    updateVisibility(component, config, state, viewport);
-  });
+  updateVisibility(component, config, state, viewport);
 
   Dockables.tryMorph(component, viewport, state).each((morph) => {
     applyMorph(component, config, state, viewport, morph);
--- a/modules/alloy/src/main/ts/ephox/alloy/behaviour/docking/DockingApis.ts
+++ b/modules/alloy/src/main/ts/ephox/alloy/behaviour/docking/DockingApis.ts
@@ -85,11 +85,14 @@ const refreshInternal = (component: AlloyComponent, config: DockingConfig, state
   // If docked then check if we need to hide/show the component
   const isDocked = state.isDocked();
 
-  // Checking isDocked works for only, toolbar_location: top mode, as you scroll, it would be undocked, docked then hidden
-  // But for toobar_location, it would be from docked, undocked then hidden
-  if (isDocked || (!isDocked && Arr.contains(state.getModes(), 'bottom'))) {
+  viewport.optScrollEnv.fold(() => {
+    if (isDocked) {
+      updateVisibility(component, config, state, viewport);
+    }
+  }, (_scrollEnv) => {
+    // If scrolling environment is set, we want to hide the toolbar if editor is out of the viewport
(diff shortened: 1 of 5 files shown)


modules/alloy/src/main/ts/ephox/alloy/behaviour/docking/DockingApis.ts
```diff
@@ -85,11 +85,14 @@ const refreshInternal = (component: AlloyComponent, config: DockingConfig, state
   // If docked then check if we need to hide/show the component
   const isDocked = state.isDocked();
 
-  // Checking isDocked works for only, toolbar_location: top mode, as you scroll, it would be undocked, docked then hidden
-  // But for toobar_location, it would be from docked, undocked then hidden
-  if (isDocked || (!isDocked && Arr.contains(state.getModes(), 'bottom'))) {
+  viewport.optScrollEnv.fold(() => {
```
shanmen-tiny, line 88: Not sure if this is the appropriate way? `optScrollEnv` would only be set when `ui_mode: 'split'`.

modules/tinymce/src/themes/silver/main/ts/modes/Inline.ts
```diff
@@ -78,7 +78,7 @@ const setupEvents = (editor: Editor, targetElm: SugarElement, ui: InlineHeader,
   // When the page has been scrolled, we need to update any docking positions. We also
   // want to reposition all the Ui elements if required.
   let lastScrollX = 0;
-  const updateUi = Throttler.last(() => ui.update(), 33);
+  const updateUi = Throttler.last(() => ui.update(Docking.refresh), 33);
```
shanmen-tiny, line 81: This merged before #8428 was merged, and later in 9414 we have made changes to `ui.update` to call `Docking.reset`, changing it to what it was before. 

https://github.com/tinymce/tinymce/blob/4c7c3d133a1789e797efe80d2f5a91c94adef2c3/modules/tinymce/src/themes/silver/main/ts/modes/Inline.ts#LL65C48-L65C48

modules/alloy/src/main/ts/ephox/alloy/behaviour/docking/DockingApis.ts
MitchCTiny:
```suggestion
 if ((ui_mode === 'split' && viewport.optScrollEnv.isSome()) || isDocked) {
   updateVisibility(component, config, state, viewport);
 }
```
I'm happy with either way. This is assuming `optScrollEnv` is an option and you only want to `updateVisibility` when `isDocked` or within a scrolling environment.

MitchCTiny (approved):

modules/alloy/src/main/ts/ephox/alloy/behaviour/docking/DockingApis.ts
shanmen-tiny: Ah yeah, `isSome()` works as well. Not sure if we want `ui_mode` option in Alloy?

MitchCTiny: Hmm, if this is only relevant in a scrollable container, and doesn't have to do with the sink, perhaps it can be included without depending on the option?

shanmen-tiny: I think the main intention is to reduce unnecessary calls. In `ui_mode: combined` mode, the docking only happens at the top/bottom, so updating the visibility occurs only when it's docked, e.g. for docking at the top:
1. Toolbar location top, Scrolling to the bottom of the screen, the toolbar will be docked (fixed position). Scrolling further down the page, it's not partly visible; it will be hidden (changing from docked mode [fixed], needs to be hidden cause docking at the top has a fixed position). Scrolling from bottom to top, the editor and toolbar won't be in the viewport. Because it is statically positioned (undocked mode), so the toolbar won't be visible even if it's not hidden anyway. 
But I think it makes sense to remove the `isDocked` checking and update visibility on scroll.

TheSpyder (approved):

PR #8531, opened 2023-03-09
TINY-9670: Move iframe init code sideways from InitContentBody to InitIframe

TheSpyder:
Related Ticket: TINY-9670

Description of Changes:
* While explaining the Init process to @kemister85, I realised that the iframe-specific things `InitContentBody` does no longer need to be there. The `skipWrite` parameter is never used.
* So I moved all the code to `InitIframe`, which simplifies the explanation considerably.

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/init/InitIframe.ts
+++ b/modules/tinymce/src/core/main/ts/init/InitIframe.ts
@@ -1,8 +1,9 @@
 import { Optional, Strings } from '@ephox/katamari';
-import { Attribute, Class, SugarElement } from '@ephox/sugar';
+import { Attribute, Class, DomEvent, SugarElement } from '@ephox/sugar';
 
 import DOMUtils from '../api/dom/DOMUtils';
 import Editor from '../api/Editor';
+import Env from '../api/Env';
 import * as Options from '../api/Options';
 import { TranslatedString } from '../api/util/I18n';
 import * as InitContentBody from './InitContentBody';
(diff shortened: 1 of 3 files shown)


modules/tinymce/src/core/main/ts/init/InitIframe.ts
```diff
@@ -88,7 +89,32 @@ const init = (editor: Editor, boxInfo: BoxInfo): void => {
   editor.getElement().style.display = 'none';
   DOM.setAttrib(editor.id, 'aria-hidden', 'true');
 
-  InitContentBody.initContentBody(editor);
+  // Restore visibility on target element
```
spocke, line 92: This makes the init function rather fat. Could you make this a local function in this module. I think it still separate thing since it deals with the inside of the iframe. The rest of the init creates the iframe and sets aria hidden on the target element so different things. Maybe rename it to say `setupIframeBody` or something?

- [ ] Consider keeping the function but making it an internal function of the InitIframe module

spocke (approved):

modules/tinymce/src/core/main/ts/init/InitIframe.ts
TheSpyder: Ah yes I just inlined the function call but that's a good idea, thanks

ltrouton (approved):

PR #8538, opened 2023-03-10
TINY-9642: Added template to deprecated list.

HAFRMO:
Related Ticket: 

Description of Changes:
Added deprecation texts.

Pre-checks:
* ~~[x] Changelog entry added~~
* ~~[x] Tests have been added (if applicable)~~

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/Deprecations.ts
+++ b/modules/tinymce/src/core/main/ts/Deprecations.ts
@@ -12,10 +12,26 @@ const removedOptions = (
   'paste_enable_default_filters,paste_filter_drop,paste_word_valid_elements,paste_retain_style_properties,paste_convert_word_fake_lists'
 ).split(',');
 
+const deprecatedOptions = (
+  'template_cdate_classes,template_mdate_classes,template_selected_content_classes,template_preview_replace_values,template_replace_values,templates,template_cdate_format,template_mdate_format'
+).split(',');
+
 const removedPlugins = 'bbcode,colorpicker,contextmenu,fullpage,legacyoutput,spellchecker,textcolor'.split(',');
 
+const deprecatedPlugins = [
+  {
+    name: 'template',
+    replacedWith: 'Advanced Template',
+  },
+];
+
+const getMatchingOptions = (options: RawEditorOptions, searchingFor: string[]): string[] => {
+  const settingNames = Arr.filter(searchingFor, (setting) => Obj.has(options, setting));
+  return Arr.sort(settingNames);
+};
+
 const getRemovedOptions = (options: RawEditorOptions): string[] => {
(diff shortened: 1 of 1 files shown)


vpyshnenko (approved):

modules/tinymce/src/core/main/ts/Deprecations.ts
```diff
@@ -12,10 +12,26 @@ const removedOptions = (
...
+const deprecatedOptions = (
+  'template_cdate_classes,template_mdate_classes,template_selected_content_classes,template_preview_replace_values,template_replace_values,templates,template_cdate_format,template_mdate_format'
+).split(',');
+
 const removedPlugins = 'bbcode,colorpicker,contextmenu,fullpage,legacyoutput,spellchecker,textcolor'.split(',');
 
+const deprecatedPlugins = [
+  {
+    name: 'template',
+    replacedWith: 'Advanced Template',
```
spocke, line 24: Should this be the plugin name `advtemplate`?

spocke (approved):

modules/tinymce/src/core/main/ts/Deprecations.ts
HAFRMO: Ticket description just said to note that it was replaced with Advanced Template. I'm guessing it's easier to look up the documentation that way?

TheSpyder: I agree, `advtemplate` won't make a lot of sense to customers at first.

However all of this is not how we did it in v5; we used a generic `moved to premium` message for `imagetools` and `toc` rather than explicitly name the new plugin. We then pointed to the release notes and said 'look here for more information'.
https://github.com/tinymce/tinymce/blob/8005d93353196045057f2a3e54a9149ef4f4b441/modules/tinymce/src/core/main/ts/Deprecations.ts#L62-L63

I'd prefer to do things in a similar way here for consistency.

Also, note, we didn't deprecate the settings for imagetools or toc. We just assumed as customers migrated away from those plugins they'd have to update the settings as well.

HAFRMO: I added the name to it since the issue encouraged it, and it might help the user actually know what it's replaced by ( even if this particular name might not be the hardest to figure out ) instead of some generic page. But if it's preferred I can change it more towards the generic ( though do we have an 'upcoming changes'-page for this? ).

As for the settings, I don't think it'd hurt to mark the options as deprecated?

Afraithe: Lets just get the message in now, we can tweak the message in a patch or followup release if needed, maybe talk to marketing. Ideally this would link to 7-upcoming changes as that would be when it will be removed etc.

PR #8543, opened 2023-03-10
TINY-9419: `More...` button disapear if it is triggered and view mode is switched

lorenzo-pomili:
Related Ticket: TINY-9419

Description of Changes:
The problem is that the [toggleToolbarDrawer](https://github.com/tinymce/tinymce/blob/develop/modules/tinymce/src/themes/silver/main/ts/ui/general/OuterContainer.ts#L200) uses [this refresh function](https://github.com/tinymce/tinymce/blob/develop/modules/alloy/src/main/ts/ephox/alloy/toolbar/SplitToolbarUtils.ts#L22), but it needs the width of the toolbar to render the buttons.

An easy fix is to show the toolbar before `toggleToolbarDrawer`, but this cause the user will see the animation again

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/themes/silver/main/ts/ui/general/OuterContainer.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/ui/general/OuterContainer.ts
@@ -196,6 +196,10 @@ const factory: UiSketcher.CompositeSketchFactory<OuterContainerSketchDetail, Out
       });
     },
     showMainView: (comp: AlloyComponent) => {
+      if (toolbarDrawerOpenState) {
+        apis.toggleToolbarDrawer(comp);
+      }
+
       Composite.parts.getPart(comp, detail, 'editorContainer').each((editorContainer) => {
         const element = editorContainer.element;
 
@@ -203,9 +207,7 @@ const factory: UiSketcher.CompositeSketchFactory<OuterContainerSketchDetail, Out
         Attribute.remove(element, 'aria-hidden');
       });
 
-      if (toolbarDrawerOpenState) {
-        apis.toggleToolbarDrawer(comp);
-      }
+      apis.refreshToolbar(comp);
     }
   };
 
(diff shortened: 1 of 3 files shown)


modules/tinymce/src/themes/silver/main/ts/ui/general/OuterContainer.ts
```diff
@@ -196,16 +196,16 @@ const factory: UiSketcher.CompositeSketchFactory<OuterContainerSketchDetail, Out
...
-      }
-
       Composite.parts.getPart(comp, detail, 'editorContainer').each((editorContainer) => {
         const element = editorContainer.element;
 
         Css.remove(element, 'display');
         Attribute.remove(element, 'aria-hidden');
       });
+
+      if (toolbarDrawerOpenState) {
```
spocke, line 206: A some what hacky way to avoid the animation would be to first do toggle when it's hidden. Then do refresh when it visible that would re-measure things. The better way would be to add a option or separate function to toggleToolbarDrawer that immediately goes to the animation final state. Not sure what is animating that is it's sliding or transitioning but there should be a way to disable the animation for both.

spocke (approved):
[ ] Add changlog item since this bug is in the current released version as well

HAFRMO (approved):

modules/tinymce/src/themes/silver/main/ts/ui/general/OuterContainer.ts
lorenzo-pomili: I did the hacky way to have something to merge in case we want this feature for the next release, but I'm digging to check how to get no animation

yacodes (approved):

modules/tinymce/src/themes/silver/main/ts/ui/general/OuterContainer.ts
lorenzo-pomili: I'll merge for now, maybe we can do a follow up to refactor it

PR #8558, opened 2023-03-15
TINY-9653: Add defaultExpandedIds and onToggleExpand options to Tree component

hamza0867:
Related Ticket: TINY-9653

Description of Changes:
* Add defaultExpandedKeys and onToggleExpand options to Tree component

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/bridge/src/main/ts/ephox/bridge/components/dialog/Tree.ts
+++ b/modules/bridge/src/main/ts/ephox/bridge/components/dialog/Tree.ts
@@ -1,4 +1,4 @@
-import { FieldSchema, StructureSchema } from '@ephox/boulder';
+import { FieldSchema, StructureSchema, ValueType } from '@ephox/boulder';
 import { Optional, Result } from '@ephox/katamari';
 
 import { ToolbarMenuButtonSpec, ToolbarMenuButton } from '../../api/Toolbar';
@@ -11,12 +11,23 @@ export interface TreeSpec {
   type: 'tree';
   items: TreeItemSpec[];
   onLeafAction?: (id: Id) => void;
+  defaultExpandedKeys?: Id[];
+  onExpand?: (
+    expandedKeys: Id[],
+    { expanded, node }: { expanded: boolean; node: Id }
+  ) => void;
 }
 
 export interface Tree {
   type: 'tree';
   items: TreeItem[];
+  defaultExpandedKeys: Optional<Id[]>;
   onLeafAction: Optional<(id: Id) => void>;
+  onExpand: Optional<(
+    expandedKeys: Id[],
+    { expanded, node }: { expanded: boolean; node: Id }
+  ) => void
+  >;
 }
 
 interface BaseTreeItemSpec {
@@ -57,7 +68,7 @@ const baseTreeItemFields = [
   FieldSchema.requiredStringEnum('type', [ 'directory', 'leaf' ]),
   ComponentSchema.title,
   FieldSchema.requiredString('id'),
-  FieldSchema.optionOf('menu', MenuButtonSchema ),
+  FieldSchema.optionOf('menu', MenuButtonSchema),
 ];
 
 const treeItemLeafFields = baseTreeItemFields;
@@ -83,7 +94,9 @@ const treeItemSchema = StructureSchema.chooseProcessor('type', {
 const treeFields = [
   ComponentSchema.type,
   FieldSchema.requiredArrayOf('items', treeItemSchema),
-  FieldSchema.optionFunction('onLeafAction')
+  FieldSchema.optionFunction('onLeafAction'),
+  FieldSchema.optionFunction('onExpand'),
+  FieldSchema.optionArrayOf('defaultExpandedKeys', ValueType.string),
 ];
 
 export const treeSchema = StructureSchema.objOf(treeFields);
--- a/modules/bridge/src/main/ts/ephox/bridge/components/dialog/Tree.ts
+++ b/modules/bridge/src/main/ts/ephox/bridge/components/dialog/Tree.ts
@@ -21,7 +21,7 @@ export interface TreeSpec {
 export interface Tree {
   type: 'tree';
   items: TreeItem[];
-  defaultExpandedIds: Optional<Id[]>;
+  defaultExpandedIds: Id[];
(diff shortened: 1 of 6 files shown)


vpyshnenko (approved):

modules/tinymce/src/themes/silver/demo/ts/demo/TreeDemo.ts
```diff
@@ -1,12 +1,12 @@
 import { Dialog } from '@ephox/bridge';
-import { Id } from '@ephox/katamari';
 
 import { TinyMCE } from 'tinymce/core/api/PublicApi';
 
 declare let tinymce: TinyMCE;
 
 interface Data {
   search: string;
+  expandedKeys: string[];
```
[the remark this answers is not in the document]
TheSpyder: Agreed. Data is there for fields in the dialog, it's not a great place to store extra information.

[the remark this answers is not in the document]
hamza0867: Done!

modules/bridge/src/main/ts/ephox/bridge/components/dialog/Tree.ts
```diff
@@ -11,12 +11,23 @@ export interface TreeSpec {
...
+    { expanded, node }: { expanded: boolean; node: Id }
+  ) => void;
 }
 
 export interface Tree {
   type: 'tree';
   items: TreeItem[];
+  defaultExpandedIds: Optional<Id[]>;
   onLeafAction: Optional<(id: Id) => void>;
+  onExpand: Optional<(
```
spocke, line 26: Since this is both expanding and collapsing maybe it would be better to just have to functions here also let the tracking of these things be handled by the integrator instead it makes it harder to integrate but makes the api and code complexity simpler from our side. Having it be called when nodes are collapsing and calling it expand is confusing to me at least.
```
onExpand: (id: Id),
onCollapse: (id: Id)
```

modules/tinymce/src/themes/silver/main/ts/ui/dialog/Tree.ts
```diff
@@ -9,6 +11,10 @@ import * as Icons from '../icons/Icons';
 
 type TreeSpec = Omit<Dialog.Tree, 'type'>;
 type OnLeafAction = (id: string) => void;
+interface ExpandTreeNodeEventArgs extends EventFormat {
```
spocke, line 14: Maybe call it ToggleExpand since it's both expand and collapse at the same time?

Also add a blank line before the interface. The other interfaces in this file all have blank lines between them. It kind of a block of code. I think the type declarations are fine to have next to each other. I think we should have a lint rule for this. Maybe it's just me nit picking. :smile: 

```suggestion

interface ToggleExpandTreeNodeEventArgs extends EventFormat {
```

spocke (changes requested):

modules/bridge/src/main/ts/ephox/bridge/components/dialog/Tree.ts
```diff
@@ -11,12 +11,23 @@ export interface TreeSpec {
...
+  onExpand?: (
+    expandedKeys: Id[],
+    { expanded, node }: { expanded: boolean; node: Id }
+  ) => void;
 }
 
 export interface Tree {
   type: 'tree';
   items: TreeItem[];
+  defaultExpandedIds: Optional<Id[]>;
```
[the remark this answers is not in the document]
hamza0867: The way I thought about it, is that it is easier for the user of this component to not provide the option instead of having to provide an empty array. What do you think ?

modules/bridge/src/main/ts/ephox/bridge/components/dialog/Tree.ts
hamza0867: Didn't we change this policy to use only one function since tiny 6. For example in the forms we no longer have `api.disable(name)` and `api.enable(name)`, instead we have `api.setEnabled(boolean, name)` ? Also this functions provides the array of expanded ids so it is not just a boolean.

modules/bridge/src/main/ts/ephox/bridge/components/dialog/Tree.ts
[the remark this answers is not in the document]
spocke: I mean internally wouldn't it be easier to have an array since you do a getOr([]) on it anyway. So just default it to empty array in the schema. But for the integrator it's optional.

modules/bridge/src/main/ts/ephox/bridge/components/dialog/Tree.ts
spocke: That was because `disable` is politically incorrect. We could have a single function but then it should be called something like `onToggleExpand` or something since it's called when it collapsed as well.

modules/tinymce/src/themes/silver/main/ts/ui/dialog/Tree.ts
hamza0867: Done!

modules/bridge/src/main/ts/ephox/bridge/components/dialog/Tree.ts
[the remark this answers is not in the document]
hamza0867: Ah I see now

modules/bridge/src/main/ts/ephox/bridge/components/dialog/Tree.ts
hamza0867: I see. I would rather call it `onToggleExpand`. I will change that. Thanks

modules/bridge/src/main/ts/ephox/bridge/components/dialog/Tree.ts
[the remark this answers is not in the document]
hamza0867: Done!

modules/bridge/src/main/ts/ephox/bridge/components/dialog/Tree.ts
hamza0867: Done!

spocke (approved):

PR #8565, opened 2023-03-17
TINY-4189: Checkbox and containing label now both disabled/enabled

HAFRMO:
Related Ticket: 

Description of Changes:
The checkbox would not trigger disable on the label. The checkbox now calls it.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/themes/silver/main/ts/ui/general/Checkbox.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/ui/general/Checkbox.ts
@@ -4,7 +4,7 @@ import {
 } from '@ephox/alloy';
 import { Dialog } from '@ephox/bridge';
 import { Fun, Optional } from '@ephox/katamari';
-import { Checked } from '@ephox/sugar';
+import { Checked, Traverse } from '@ephox/sugar';
 
 import { UiFactoryBackstageProviders } from '../../backstage/Backstage';
 import * as ReadOnly from '../../ReadOnly';
@@ -21,6 +21,7 @@ export const renderCheckbox = (spec: CheckboxSpec, providerBackstage: UiFactoryB
     return Optional.some(true);
   };
 
+  let isEnabled = spec.enabled;
   const pField = AlloyFormField.parts.field({
     factory: { sketch: Fun.identity },
     dom: {
@@ -34,7 +35,19 @@ export const renderCheckbox = (spec: CheckboxSpec, providerBackstage: UiFactoryB
     behaviours: Behaviour.derive([
       ComposingConfigs.self(),
       Disabling.config({
-        disabled: () => !spec.enabled || providerBackstage.isDisabled()
+        disabled: () => !spec.enabled || providerBackstage.isDisabled(),
+        onDisabled: (component) => {
+          if (isEnabled) { // Prevent recursive loop
+            isEnabled = false;
+            Traverse.parent(component.element).each((element) => component.getSystem().getByDom(element).each(Disabling.disable));
+          }
+        },
+        onEnabled: (component) => {
+          if (!isEnabled) { // Prevent recursive loop
+            isEnabled = true;
+            Traverse.parent(component.element).each((element) => component.getSystem().getByDom(element).each(Disabling.enable));
+          }
+        }
       }),
       Tabstopping.config({}),
       Focusing.config({ }),
--- a/modules/tinymce/src/themes/silver/main/ts/ui/general/Checkbox.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/ui/general/Checkbox.ts
@@ -4,7 +4,7 @@ import {
 } from '@ephox/alloy';
 import { Dialog } from '@ephox/bridge';
 import { Fun, Optional } from '@ephox/katamari';
-import { Checked, Traverse } from '@ephox/sugar';
+import { Checked, Class, Traverse } from '@ephox/sugar';
 
 import { UiFactoryBackstageProviders } from '../../backstage/Backstage';
 import * as ReadOnly from '../../ReadOnly';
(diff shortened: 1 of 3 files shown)


hamza0867 (approved):

modules/tinymce/src/themes/silver/main/ts/ui/general/Checkbox.ts
```diff
@@ -34,7 +35,19 @@ export const renderCheckbox = (spec: CheckboxSpec, providerBackstage: UiFactoryB
     behaviours: Behaviour.derive([
       ComposingConfigs.self(),
       Disabling.config({
-        disabled: () => !spec.enabled || providerBackstage.isDisabled()
+        disabled: () => !spec.enabled || providerBackstage.isDisabled(),
+        onDisabled: (component) => {
+          if (isEnabled) { // Prevent recursive loop
```
hamza0867, line 40: I don't like this. Even though you prevent the recursive loop this way, I'd rather not have both the parent and the child control each other's states. The control should be in one component either the parent or the child. So I suggest you remove the onDisabled, onEnabled from the parent component and leave only the ones in the child component. This is my opinion, let's see what others think.

hamza0867 (changes requested):

modules/tinymce/src/themes/silver/main/ts/ui/general/Checkbox.ts
spocke: Yes, this is flipped around the parent component has onDisabled and onEnabled I think instead that code should go in and also disable/enable the input. And the disabling call where ever that is now should act on the parent control not the input.

spocke: The whole thing needs to be disabled since both the label and the checkbox should have a disabled aria state.

spocke: Yes, you are right the disabled state is only set on the actual input element. However since it's a checkbox you can't see that it's disabled so the label also needs to be disabled if it's in a disabled state. However that can be done with css using the ~ combinator so something like `input:disabled ~ span.tox-checkbox__label { color: gray }` or we could manually mutate the disabled class only on the parent. I don't think we need a `isEnabled` state the onDisabled/onEnabled already handles that state managment.

So I think as @hamza0867 said the disabling behavior should only be on the input. But the onDisabled/onEnabled could add/remove the class on the parent would likely be the easiest solution.

I think the original bug where the class doesn't get removed is no longer the case however now it has no visual indicator of when the checkbox is disabled or not if it in say the unchecked case.

spocke: The combinator is not going to work so ignore that since the svg updates as well if you add the disabled class to the form field. So adding/removing the class on the parent seems like the quickest fix here. The `aira-diabled` would be removed if you remove the parent disabling behavior. I think something fundamental broke here the intent I think is that all form fields checkboxes, inputs etc should toggle the class on/off on the whole field.

So I would add this to the pField:
```
        onDisabled: (component) => {
          Traverse.parentElement(component.element).each((el) => Class.add(el, 'tox-checkbox--disabled'));
        },
        onEnabled: (component) => {
          Traverse.parentElement(component.element).each((el) => Class.remove(el, 'tox-checkbox--disabled'));
        }
```

Then remove the Disabling behavior for the AlloyFormField the `fieldBehaviors` since the aria state there is a lie.

HAFRMO: Seems like I'd misunderstood a critical part of the code, so my previous comment is entirely wrong, thank you hamza for taking the time to point it out.
I'll run with the above suggestion. We'd made a solution during the explanation that relied on the :not-selector in css, which was a bit more elegant but which was a bit more likely to be a breaking change.

modules/tinymce/src/themes/silver/main/ts/ui/general/Checkbox.ts
```diff
@@ -37,16 +36,10 @@ export const renderCheckbox = (spec: CheckboxSpec, providerBackstage: UiFactoryB
...
-            Traverse.parent(component.element).each((element) => component.getSystem().getByDom(element).each(Disabling.disable));
-          }
+          Traverse.parentElement(component.element).each((element) => Class.add(element, 'tox-checkbox--disabled'));
         },
         onEnabled: (component) => {
-          if (!isEnabled) { // Prevent recursive loop
-            isEnabled = true;
-            Traverse.parent(component.element).each((element) => component.getSystem().getByDom(element).each(Disabling.enable));
-          }
+          Traverse.parentElement(component.element).each((element) => Class.remove(element, 'tox-checkbox--disabled'));
```
hamza0867: I would even say there is no need for this. We can accomplish the same thing by changing the css, instead of using `tox-checkbox--disabled` we can use `:has(input:disabled)` and we won't need to run extra js to toggle css classes.

hamza0867 (approved):

modules/tinymce/src/themes/silver/main/ts/ui/general/Checkbox.ts
lorenzo-pomili: I think `:has` selector could give us some trouble with FireFox: https://caniuse.com/?search=%3Ahas

TheSpyder (approved):
I agree with the thread that it's not my preferred solution, but it'll do

modules/tinymce/src/themes/silver/main/ts/ui/general/Checkbox.ts
HAFRMO: Seems like using the current method, and let the css stay in place, might be the best option for the time being.

PR #8586, opened 2023-03-23
TINY-9730: Add command, toolbar, and menu bar button to insert accordion

yacodes:
Related Ticket: TINY-9730

Description of Changes:
* Added initial `accordion` plugin structure;
* Added the `InsertAccordion` command;
* Added the toolbar and menu bar buttons;
* Added tests for inserting an accordion.

This new `accordion` plugin development is split in several tasks. This is the first one.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/tinymce/src/plugins/accordion/main/ts/core/Actions.ts
```diff
@@ -0,0 +1,29 @@
+import Editor from 'tinymce/core/api/Editor';
+
+const insertAccordion = (editor: Editor): void => {
+  const dom = editor.dom;
+
+  const rng = editor.selection.getRng();
+  const summaryText = rng.collapsed ? 'Accordion summary...' : rng.toString();
+
+  const target = dom.getParent(rng.commonAncestorContainer, dom.isBlock);
```
spocke, line 9: This is to wide a block can be things like tables and lists so this breaks the structure if you start wrapping parts of the table or list structure in details elements. It think this should either be a smaller set of block types, the schema has a text blocks list, or it would have to be smarter on the types of blocks that do break. Also in TDs the elements might not have any blocks so it would have to be smart to first take the contents in the cell and convert that into a details body elements.

modules/tinymce/src/plugins/accordion/test/ts/browser/AccordionPluginTest.ts
```diff
@@ -0,0 +1,61 @@
...
+    plugins: 'accordion',
+    add_unload_trigger: false,
+    disable_nodechange: true,
+    indent: false,
+    entities: 'raw',
+    extended_valid_elements: 'details[class|open|data-mce-open],summary[class],div[class],p',
+    base_url: '/project/tinymce/js/tinymce'
+  }, [ AccordionPlugin ]);
+
+  it('TINY-9730: Insert an accordion element into a single paragraph', () => {
```
spocke, line 20: I think testing needs to be more expanded due to the potential breaking of wrapping inside or on blocks that are special. So maybe make this a function that can be reused instead. All the tests seems to have common things going on. Setting up some html, selecting a piece of content, applying the accordion, then verifying the content and the selection.

spocke (changes requested):

modules/tinymce/src/plugins/accordion/demo/ts/demo/Demo.ts
```diff
@@ -0,0 +1,12 @@
+import { TinyMCE } from 'tinymce/core/api/PublicApi';
+
+declare let tinymce: TinyMCE;
+
+tinymce.init({
+  selector: 'textarea.tinymce',
+  plugins: 'lists image accordion code',
+  toolbar: 'numlist bullist | image | accordion | code',
+  menu: { tools: { title: 'Tools', items: 'accordion code' }},
```
vpyshnenko, line 9: Shouldn't accordion be in `Insert` menu

modules/tinymce/src/plugins/accordion/main/ts/core/Actions.ts
```diff
@@ -0,0 +1,38 @@
...
+
+const validAncestorContainers = [ 'P', 'H1', 'H2', 'H3', 'H4', 'H5', 'H6', 'UL', 'OL', 'TABLE', 'FIGURE', 'DETAILS', 'DIV', 'DL', 'BODY' ];
+const isValidAncestorContainer = (node: Node): node is HTMLElement => 
+  validAncestorContainers.includes(node.nodeName);
+
+const insertAccordion = (editor: Editor): void => {
+  const dom = editor.dom;
+
+  const rng = editor.selection.getRng();
+  const summaryText = rng.toString() || 'Accordion summary...';
```
TheSpyder, line 13: I don't think we include english strings in content, even the placeholder has to be configured. I'm not sure how we achieve the feature goals of accordions without these strings, though.

Please at least make sure they can be translated.

TheSpyder (changes requested):

modules/tinymce/src/plugins/accordion/test/ts/browser/AccordionPluginTest.ts
yacodes: Added the `testInsertingAccordion` function and a couple of tests to test inserting an accordion into lists and tables.

lorenzo-pomili (approved):

modules/tinymce/CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - New optional `defaultExpandedIds` and `onToggleExpand` options to the `tree` component config. #TINY-9653
+- New `accordion` plugin with the `InsertAccordion` command, toolbar, and menu bar button to insert an accordion. #TINY-9730
```
vpyshnenko, line 11: Maybe make it less verbose. Hardly to expect we can describe all those things added with the new plugin in one change log entry.
```suggestion
- New `accordion` plugin with the `InsertAccordion` command. #TINY-9730
```

vpyshnenko (approved):

modules/tinymce/CHANGELOG.md
yacodes: Thanks, shortened the CHANGELOG entry!

yacodes:
@TheSpyder @spocke Can I ask you to take one more look and adjust the requested changes, please? Thanks!

modules/tinymce/src/plugins/accordion/main/ts/core/Actions.ts
```diff
@@ -0,0 +1,52 @@
+import Editor from 'tinymce/core/api/Editor';
+
+import { fireInsertAccordionEvent } from '../api/Events';
+
+const nonRemovableContainers = [ 'BODY', 'TD', 'TH', 'LI', 'DIV', 'DT', 'DD' ];
+const isNonRemovableContainer = (node: Node): boolean =>
+  nonRemovableContainers.includes(node.nodeName);
+
+const validAncestorContainers =
```
spocke, line 9: Is table really a correct parent. You shouldn't be able to insert an accordion next to a `tbody` ?

modules/tinymce/src/plugins/accordion/main/ts/core/Actions.ts
```diff
@@ -0,0 +1,52 @@
...
+  const summaryText = rng.toString() || editor.translate('Accordion summary...');
+  const bodyText = editor.translate('Accordion body...');
+
+  const target = dom.getParent(rng.commonAncestorContainer, isValidAncestorContainer);
+
+  if (!target) {
+    return;
+  }
+
+  const details = dom.create('details', { class: 'mce-accordion', open: 'open' });
```
spocke, line 33: I wonder if this whole insert thing could be done with the insertContent API instead since that is schema aware. You can get the inserted summary location by doing a temporary data attribute.

```
const uid = Id.generate('acc');
const summaryText = editor.dom.encode(editor.selection.getRng().toString() || editor.translate('Accordion summary...'));
const bodyText = editor.dom.encode(editor.translate('Accordion body...'));
editor.insertContent(`<details mce-accordion open><summary class="mce-accordion-summary" data-mce-id="${uid}">${summaryText}<div class="mce-accordion-body"><p>${bodyText}</p></div></summary>`);
const summary = editor.dom.select(`[data-mce-id="${uid"]`);
if (summary) {
  editor.selection.setCursorLocation(summary, 1);
  summary.removeAttribute('data-mce-id');
}
```

I think we should consider that since it would reduce the complexity here a ton since the insert code is already handling a lot of these things. At least try that apporach if it failes then we can do something else.

modules/tinymce/src/plugins/accordion/main/ts/core/Actions.ts
```diff
@@ -0,0 +1,52 @@
+import Editor from 'tinymce/core/api/Editor';
+
+import { fireInsertAccordionEvent } from '../api/Events';
+
+const nonRemovableContainers = [ 'BODY', 'TD', 'TH', 'LI', 'DIV', 'DT', 'DD' ];
```
spocke, line 5: Not sure body should be in the valid set.

spocke (changes requested):

modules/tinymce/src/plugins/accordion/main/ts/core/Actions.ts
yacodes: Simplified the naming a bit and added a more explicit `getInsertPosition` function. Let me know if this clarifies this bit and please share your ideas about the naming!

modules/tinymce/src/plugins/accordion/main/ts/core/Actions.ts
```diff
@@ -0,0 +1,52 @@
+import Editor from 'tinymce/core/api/Editor';
+
+import { fireInsertAccordionEvent } from '../api/Events';
```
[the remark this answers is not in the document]
yacodes: Updated to the named import.

modules/tinymce/src/plugins/accordion/main/ts/core/Actions.ts
yacodes: `BODY` element can occur here as a `rng.commonAncestorContainer` when several paragraphs are selected. In this case I am inserting the `details` element at the end of the `BODY` element. Seems like the correct solution to me.


Without the `BODY` element in this list the `target` variable will be `null` and the accordion won't be inserted, which is a weird behavior.

spocke: Right but not sure we should have the schema re-implemented in the accordion plugin. I think using the insert api is the right approach here since smif you say insert a table in the middle of a paragraph with text that gets split apart and the accordion is inserted the same thing should happen with the accordion inserting anywhere should just split the parent block and insert it in it's place. Since details is a block element and blocks can't contain blocks unless the block is a div or a td then it will split.

yacodes: Obsolete due to the latest changes.

modules/tinymce/src/plugins/accordion/main/ts/core/Actions.ts
yacodes: Obsolete due to the latest changes.

modules/tinymce/src/plugins/accordion/main/ts/core/Actions.ts
TheSpyder: Can you toggle the enabled state of the insert button based on the node changed event? This is how other features block nested inserting, I believe.

modules/tinymce/src/plugins/accordion/main/ts/core/Actions.ts
```diff
@@ -0,0 +1,39 @@
...
+    if (!body) {
+      return;
+    }
+    if (!body.lastChild) {
+      return;
+    }
+    editor.selection.setCursorLocation(body.lastChild, 1);
+  }
+
+  const uid = Id.generate('acc');
```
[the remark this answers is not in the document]
TheSpyder: One of these days we'll figure out how to stop codeql complaining about `Id.generate`

modules/tinymce/src/plugins/accordion/main/ts/core/Actions.ts
```diff
@@ -0,0 +1,39 @@
...
+  const summaryText = editor.dom.encode(editor.selection.getRng().toString() || editor.translate('Accordion summary...'));
+  const bodyText = editor.dom.encode(editor.translate('Accordion body...'));
```
TheSpyder, line 22: Wild idea, and this is a huge scope change, could we have the summary and body act like placeholder and be removed when the user focuses them? And also when saving?

We'd probably still need to translate the string, but it would avoid potentially having the text included on save.

This would be a new ticket, not an extension of the current one.

modules/tinymce/src/plugins/accordion/main/ts/core/Actions.ts
```diff
@@ -0,0 +1,39 @@
...
+    return;
+  }
+  details.removeAttribute('data-mce-id');
+
+  const summary = editor.dom.select(`summary`, details)[0];
+  if (summary) {
+    editor.selection.setCursorLocation(summary, 1);
+  }
+
+  Events.fireInsertAccordionEvent(editor, details);
```
TheSpyder, line 36: That's not quite how TinyMCE events work. If the event name is present tense, like "InsertAccordion", it should give the integrator a chance to adjust what happens. A notification event must be past tense, e.g. "InsertedAccordion".

For example, the `insertCustomChar` event inserts the character returned by the event:
https://github.com/tinymce/tinymce/blob/c12fb6af3b6b72886946dd3fb1bb1b8be249949f/modules/tinymce/src/plugins/charmap/main/ts/core/Actions.ts#L5-L8

I think, given our goals for this feature, it should remain a notification event and be renamed.

- [ ] rename event (I realise this will require discussion, as you've followed the spec exactly)

TheSpyder (approved):

modules/tinymce/src/plugins/accordion/main/ts/core/Actions.ts
yacodes: We've had a discussion about the placeholders. The desired solution is to show them only if summary and body have no text within and hide otherwise. But we've decided to implement it as a separate task after implementing the accordion plugin.

modules/tinymce/src/plugins/accordion/main/ts/core/Actions.ts
yacodes: It is a good point, thanks! Will rename the event and will trigger the discussion of the necessity of the `InsertAccordion` event to manipulate its contents before inserting or something.

modules/tinymce/src/plugins/accordion/main/ts/core/Actions.ts
yacodes: Thanks for the tip! Disabled the button if the cursor is within the <summary> element. I've tried using `sugar` but I do not feel that it is cleaner, all the SugarElement conversions make this verbose too.

modules/tinymce/src/plugins/accordion/main/ts/ui/Buttons.ts
```diff
@@ -0,0 +1,25 @@
...
+const isSummary = (node: Node): node is HTMLElement =>
+  node.nodeName === 'SUMMARY';
+
+const isInSummary = (editor: Editor): boolean => {
+  const node = editor.selection.getNode();
+  return isSummary(node) || Boolean(editor.dom.getParent(node, isSummary));
+};
+
+const onSetup = (editor: Editor) => (buttonApi: Toolbar.ToolbarButtonInstanceApi | Menu.MenuItemInstanceApi) => {
+  const f = () => buttonApi.setEnabled(!isInSummary(editor));
```
spocke, line 14: Maybe describe this one with a name since if we need to have another event listener here later. That is likely if we toggle say content editable false on/off on the body and want to disable the button then the `f` would need to be changed. So better to change it now. I think in general it better to have a describing name unless the function is some random function that is not known say in a compose or map or something.

```suggestion
  const nodeChanged = () => buttonApi.setEnabled(!isInSummary(editor));
```

spocke (approved):

modules/tinymce/src/plugins/accordion/main/ts/core/Actions.ts
```diff
@@ -0,0 +1,39 @@
...
+const insertAccordion = (editor: Editor): void => {
+  const container = editor.selection.getNode();
+  if (container.nodeName === 'SUMMARY') {
+    const body = editor.dom.getNext(container, 'div.mce-accordion-body');
+    if (!body) {
+      return;
+    }
+    if (!body.lastChild) {
+      return;
+    }
```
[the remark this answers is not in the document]
yacodes: Totally forgot about the optional chaining operator, looks great, thanks!

modules/tinymce/src/plugins/accordion/main/ts/core/Actions.ts
```diff
@@ -0,0 +1,39 @@
...
+  const bodyText = editor.dom.encode(editor.translate('Accordion body...'));
+  editor.insertContent(`<details data-mce-id="${uid}" class="mce-accordion" open="open"><summary class="mce-accordion-summary">${summaryText}</summary><div class="mce-accordion-body"><p>${bodyText}</p></div></details>`);
+
+  const details = editor.dom.select(`[data-mce-id="${uid}"]`)[0];
+  if (!details) {
+    return;
+  }
+  details.removeAttribute('data-mce-id');
+
+  const summary = editor.dom.select(`summary`, details)[0];
```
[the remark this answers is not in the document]
yacodes: Yeah, I've used a selector with the `id` attr previously. Forgot to change.

modules/tinymce/src/plugins/accordion/main/ts/ui/Buttons.ts
yacodes: I think I've copied it from somewhere, sorry! Changed the name to `onNodeChange`.

modules/tinymce/src/plugins/accordion/main/ts/core/Actions.ts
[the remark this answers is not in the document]
yacodes: We've decided to remove the `InsertedAccordion` event.

PR #8591, opened 2023-03-27
TINY-9680: context toolbar not showing the correct status for `advlist`

lorenzo-pomili:
Related Ticket: TINY-9680

Description of Changes:
the problem seems to was caused by the different way that we manage the `makeSetupHandler` in [advlist](https://github.com/tinymce/tinymce/blob/develop/modules/tinymce/src/plugins/advlist/main/ts/ui/Buttons.ts) versus in the [lists](https://github.com/tinymce/tinymce/blob/develop/modules/tinymce/src/plugins/lists/main/ts/ui/Buttons.ts#L7-L13), so I simply aligned the two behaviors.

About the test, at the moment there is some strange thing that is going on with the test about the open of the `context toolbar`

- I had to put it as the first `context` because somehow the other test block the opening of the `context toolbar`
- I had to put `inline: true` otherwise it does not work
- I tried to use: `editor.dispatch('contexttoolbar-show', { toolbarKey });` but it won't reproduce the bug

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Table toolbar was visible even if the table was within a noneditable host element. #TINY-9664
 - Quickbar toolbars was shown for noneditable contents in a noneditable root. #TINY-9460
 - Inline alert in the "Search and Replace" dialog persisted when it wasn't necessary. #TINY-9704
+- Context toolbars was not showning the correct status for the `advlist` buttons. #TINY-9680
 
 ## 6.4.0 - 2023-03-15
 
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Table toolbar was visible even if the table was within a noneditable host element. #TINY-9664
 - Quickbar toolbars was shown for noneditable contents in a noneditable root. #TINY-9460
 - Inline alert in the "Search and Replace" dialog persisted when it wasn't necessary. #TINY-9704
-- Context toolbars was not showning the correct status for the `advlist` buttons. #TINY-9680
+- Context toolbars displayed the incorrect status for the `advlist` plugin buttons. #TINY-9680
 
 ## 6.4.0 - 2023-03-15
 
--- a/modules/tinymce/src/plugins/advlist/main/ts/ui/Buttons.ts
+++ b/modules/tinymce/src/plugins/advlist/main/ts/ui/Buttons.ts
@@ -3,7 +3,6 @@ import { Type } from '@ephox/katamari';
 import Editor from 'tinymce/core/api/Editor';
 import { NodeChangeEvent } from 'tinymce/core/api/EventTypes';
 import { Menu, Toolbar } from 'tinymce/core/api/ui/Ui';
-import { EditorEvent } from 'tinymce/core/api/util/EventDispatcher';
 import Tools from 'tinymce/core/api/util/Tools';
 
 import * as Options from '../api/Options';
@@ -26,13 +25,11 @@ const normalizeStyleValue = (styleValue: string | undefined): string =>
   Type.isNullable(styleValue) || styleValue === 'default' ? '' : styleValue;
 
 const makeSetupHandler = (editor: Editor, nodeName: ListType) => (api: Toolbar.ToolbarSplitButtonInstanceApi | Toolbar.ToolbarToggleButtonInstanceApi) => {
-  const nodeChangeHandler = (e: EditorEvent<NodeChangeEvent>) => {
+  const nodeChangeHandler = (e: NodeChangeEvent) => {
     api.setActive(ListUtils.inList(editor, e.parents, nodeName));
     api.setEnabled(!ListUtils.isWithinNonEditableList(editor, e.element));
   };
-  editor.on('NodeChange', nodeChangeHandler);
-
-  return () => editor.off('NodeChange', nodeChangeHandler);
+  return ListUtils.setNodeChangeHandler(editor, nodeChangeHandler);
 };
 
 const addSplitButton = (editor: Editor, id: string, tooltip: string, cmd: string, nodeName: ListType, styles: string[]): void => {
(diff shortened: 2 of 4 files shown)


modules/tinymce/src/plugins/advlist/main/ts/ui/Buttons.ts
```diff
@@ -26,13 +25,11 @@ const normalizeStyleValue = (styleValue: string | undefined): string =>
   Type.isNullable(styleValue) || styleValue === 'default' ? '' : styleValue;
 
 const makeSetupHandler = (editor: Editor, nodeName: ListType) => (api: Toolbar.ToolbarSplitButtonInstanceApi | Toolbar.ToolbarToggleButtonInstanceApi) => {
-  const nodeChangeHandler = (e: EditorEvent<NodeChangeEvent>) => {
+  const nodeChangeHandler = (e: NodeChangeEvent) => {
     api.setActive(ListUtils.inList(editor, e.parents, nodeName));
     api.setEnabled(!ListUtils.isWithinNonEditableList(editor, e.element));
   };
```
spocke, line 31: Maybe just instead of simulating events you just break up the state change into a new function. Also it needs to use `getStart(true)` to do the same thing as nodeChange.

```suggestion
  const updateButtonState = (element: Element, parents: Node[]) => {
    api.setActive(ListUtils.inList(editor, parents, nodeName));
    api.setEnabled(!ListUtils.isWithinNonEditableList(editor, element));
  };

  const nodeChangeHandler = (e: NodeChangeEvent) => updateButtonState(e.element, e.parent);
  const element = editor.selection.getStart(true);
  updateButtonState(element, editor.dom.getParents(element));
```

modules/tinymce/src/plugins/advlist/test/ts/browser/AdvlistOptionsAndToolbarTest.ts
```diff
@@ -85,6 +85,59 @@ describe('browser.tinymce.plugins.advlist.AdvlistOptionsAndToolbarTest', () => {
       })), button);
     });
 
+  context('Context Toolbar Split buttons active state', () => {
+    const hook = TinyHooks.bddSetup<Editor>({
+      ...baseOptions,
+      inline: true, // this is needed because otherwise `TinySelections.setSelection` does not trigger the context toolbar correctly
```
spocke, line 91: Why would that be different I see the bug happening in iframe editors as well? Does it need more waiters?

lorenzo-pomili: yes the bug is in both, but I'm not able to open the `context toolbar` in tests without `inline: true`, I tried to add waiter etc. but nothing seems to work :/

spocke: Might be that the editor doesn't have focus. Try adding `,true` to the bddSetup

lorenzo-pomili: yes, it solves the problem, thanks :)

modules/tinymce/src/plugins/advlist/main/ts/ui/Buttons.ts
lorenzo-pomili: changed

modules/tinymce/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Table toolbar was visible even if the table was within a noneditable host element. #TINY-9664
 - Quickbar toolbars was shown for noneditable contents in a noneditable root. #TINY-9460
 - Inline alert in the "Search and Replace" dialog persisted when it wasn't necessary. #TINY-9704
+- Context toolbars was not showning the correct status for the `advlist` buttons. #TINY-9680
```
yacodes, line 14:
```suggestion
- Context toolbars displayed the incorrect status for the `advlist` plugin buttons. #TINY-9680
```

modules/tinymce/src/plugins/advlist/main/ts/ui/Buttons.ts
```diff
@@ -26,13 +25,15 @@ const normalizeStyleValue = (styleValue: string | undefined): string =>
...
-    api.setActive(ListUtils.inList(editor, e.parents, nodeName));
-    api.setEnabled(!ListUtils.isWithinNonEditableList(editor, e.element));
+  const updateButtonState = (element: Element, parents: Node[]) => {
+    api.setActive(ListUtils.inList(editor, parents, nodeName));
+    api.setEnabled(!ListUtils.isWithinNonEditableList(editor, element));
   };
-  editor.on('NodeChange', nodeChangeHandler);
+  const nodeChangeHandler = (e: NodeChangeEvent) => updateButtonState(e.element, e.parents);
+  const element = editor.selection.getStart(true);
+  updateButtonState(element, editor.dom.getParents(element));
```
yacodes, line 34: Why do we need to call the `updateButtonState` function explicitly here? It will be called one more time right after this call from `ListUtils.setNodeChangeHandler(editor, nodeChangeHandler)`. This call feels a bit redundant.

yacodes (approved):

modules/tinymce/src/plugins/advlist/main/ts/ui/Buttons.ts
yacodes: Or the `ListUtils.setNodeChangeHandler`. Unless I'm missing something, the `updateButtonState` is unnecessary called twice.

TheSpyder: No, you're right. The new code calls it twice - using different inputs (`editor.selection.getStart(true)` and then `editor.selection.getNode()`)

I think inside `setNodeChangeHandler` is the correct place, but it should use `getStart` as spocke said in the other thread.

- [ ] only set the initial state once

TheSpyder (approved):

modules/tinymce/src/plugins/advlist/main/ts/ui/Buttons.ts
spocke: You don't need the `setNodeChangeHandler` since line 36 should just be. `return () => editor.off('NodeChange', nodeChangeHandler);` but you could pull this apart into that module if you like but still use the non fake event approach.

lorenzo-pomili: What about something like this:
```javascript
const updateButtonState = (editor: Editor, parents: Node[]) => {
    const element = editor.selection.getStart(true);
    api.setActive(ListUtils.inList(editor, parents, nodeName));
    api.setEnabled(!ListUtils.isWithinNonEditableList(editor, element));
  };
  const nodeChangeHandler = (e: NodeChangeEvent) => updateButtonState(editor, e.parents);

  return ListUtils.setNodeChangeHandler(editor, nodeChangeHandler);
```
it does not use the element in the fake event and also does not call `updateButtonState`

spocke: Yes the easy fix is to just get rid of the extra `updateButtonState(element, editor.dom.getParents(element));` and keep the separate function. As suggested.

lorenzo-pomili: done, let me know if it is ok

spocke (approved):

PR #8592, opened 2023-03-27
TINY-9668: fix dialog's components `max-height` in order to make the content scroll

lorenzo-pomili:
Related Ticket: TINY-9668

Description of Changes:
~to fix this I added a `max-height` as `vh` to each component of the `dialog`:~
~`tox-dialog__header` -> 15vh - its padding~
~`tox-dialog__content-js` -> 70vh~
~`tox-dialog__footer` -> 15vh - its padding~

~this seems to work, but I'm not sure that we still need this [other fix](https://github.com/tinymce/tinymce/blob/develop/modules/oxide/src/less/theme/components/dialog/dialog.less#L233) for `tox-dialog__body-content`~

to fix scroll problem, the `max-height` of `tox-dialog__body-content` as been changed into a `min` between `650px` (the current value) and `calc(100vh - 110px)`

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/oxide/src/less/theme/alien/unknowns.less
+++ b/modules/oxide/src/less/theme/alien/unknowns.less
@@ -35,6 +35,7 @@
   .tox-dialog__content-js {
     display: flex;
     flex: 1;
+    max-height: 70vh;
   }
 
   // TODO: How is this used? This rule quite drastically changes
--- a/modules/oxide/src/less/theme/alien/unknowns.less
+++ b/modules/oxide/src/less/theme/alien/unknowns.less
@@ -35,7 +35,6 @@
   .tox-dialog__content-js {
     display: flex;
     flex: 1;
-    max-height: 70vh;
   }
 
   // TODO: How is this used? This rule quite drastically changes
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Quickbar toolbars was shown for noneditable contents in a noneditable root. #TINY-9460
 - Inline alert in the "Search and Replace" dialog persisted when it wasn't necessary. #TINY-9704
 - Context toolbars displayed the incorrect status for the `advlist` plugin buttons. #TINY-9680
+- The dialog's conent doesn't scroll. #TINY-9668
 
 ## 6.4.0 - 2023-03-15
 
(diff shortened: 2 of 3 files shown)


modules/oxide/src/less/theme/alien/unknowns.less
```diff
@@ -35,6 +35,7 @@
   .tox-dialog__content-js {
     display: flex;
     flex: 1;
+    max-height: 70vh;
```
TheSpyder, line 38: I don't think this is going to work, because `body-content` has 650px.
https://github.com/tinymce/tinymce/blob/17cb12bb10649b868717e0e42f1cd41a807a9f69/modules/oxide/src/less/theme/components/dialog/dialog.less#L233

The reason `tabpanel` dialogs work and regular `panel` dialogs don't is this resize code:
https://github.com/tinymce/tinymce/blob/3a6455aaebf6034e122ef6cf86beeecef2c7103b/modules/tinymce/src/themes/silver/main/ts/ui/alien/DialogTabHeight.ts#L116-L121

It adjusts the tab height when the window resizes based on the window height minus the height of the dialog "chrome" (header and button rows). Once the height is set, `overflow: auto` kicks in and the scrollbar appears.

Is there a way we can adapt or build equivalent code for regular panels?

TheSpyder (changes requested):
See inline comments

modules/oxide/src/less/theme/alien/unknowns.less
lorenzo-pomili: yes, I saw both, the resize code and that `max-height: 650px`.

About the `max-height: 650px` it is inside the `.tox-dialog__content-js` more specifically in `.tox-dialog__body-content`, so:
  · if the content exceed `650px` it will scroll because of `.tox-dialog__body-content`
  · if the content doesn't exceed `650px` but exceeds the height of the window it will scroll because of `.tox-dialog__content-js`
  · if the content doesn't exceed `650px` and also doesn't exceed the height of the window it will not scroll

About the resize calculation, I think there is a way to do so (I was checking how the height of the panel in tabpanel is calculated), but I thought that a CSS solution would be more solid, but since I'm not a CSS expert I 'd like also to have an opinion from @noxuhax too to understand if my fix is correct and also if solve all via CSS is the best solution in this case or not

lorenzo-pomili: @noxuhax I removed the `max-height` from `tox-dialog__content-js` and also from `header` and `footer`, and I added that `max-height` with `min` and `calc`, let me know if it is ok

modules/tinymce/CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Quickbar toolbars was shown for noneditable contents in a noneditable root. #TINY-9460
 - Inline alert in the "Search and Replace" dialog persisted when it wasn't necessary. #TINY-9704
 - Context toolbars displayed the incorrect status for the `advlist` plugin buttons. #TINY-9680
+- The dialog's conent doesn't scroll. #TINY-9668
```
vpyshnenko, line 18:
```suggestion
- The content of the dialog body could not be scrolled. #TINY-9668
```

vpyshnenko (approved):

PR #8616, opened 2023-04-05
TINY-9251: Use K8s for TinyMCE build process

jscasca:
Migrate TinyMCE build process to containers

Related Ticket: TINY-9251

Description of Changes:
* Use a node container for building. Keep headless-test on mac-headless (`headless-macos`)

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -129,13 +129,15 @@ timestamps {
         // TODO: Determine if this should be run on a container instead
         node('headless-macos') {
           // Prepare the branch on the node
-          checkout(scm)
-          gitMerge(primaryBranch)
-          cleanAndInstall()
-          exec("yarn ci -s tinymce-rollup")
+          lock('headless tests') {
+            checkout(scm)
+            gitMerge(primaryBranch)
+            cleanAndInstall()
+            exec("yarn ci -s tinymce-rollup")
 
-          echo "Platform: chrome-headless tests on node: $NODE_NAME"
-          runHeadlessTests(runAllTests)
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -121,11 +126,14 @@ node("headless-macos") {
 
     processes["headless-and-archive"] = {
       stage("headless tests") {
-        // Prevent multiple headless tests running at once
-        lock("headless tests") {
-          // chrome-headless tests run on the same node as the pipeline
-          // we are re-using the state prepared by `ci-all` below
-          // if we ever change these tests to run on a different node, rollup is required in addition to the normal CI command
+        // TODO: Determine if this should be run on a container instead
```
TheSpyder, line 129: Where we do any of this will depend on the performance. We started using headless macos because it was way, way faster.

You'll still need the lock, inside the `headless-macos` node, because it has multiple executors. Or change it to not use general macos, not headless macos, for these tests.

TheSpyder (commented):
I just did a rebuild and nothing has changed since the commit Lee made in January - the `type check` stage takes 90 seconds on the container vs 50-55 on the Apple Silicon macos node.

The tests run faster, but that's because this branch is 4 months out of date.

I'm not convinced this is a clear win.

Jenkinsfile
jscasca: Used lock for `headless-macos`

TheSpyder (approved):

PR #8627, opened 2023-04-07
TINY-9754: now `fontsizeinput` revert to old value instead of setting a default value 

lorenzo-pomili:
Related Ticket: TINY-9754

Description of Changes:

possible solutions to avoid that inserting a value with an invalid unit the value will be set to `0pt` (or any other default unit)

1. a mechanism that reverses it to the previous value.
2. check what value the user is trying to insert and add the default unit

I implemented the first one because the second one has these 2 problems:
- value: `abc` -> no number so back to the previous default.
- from `16pt` to `1emm` -> probably here the intention of the user would have been to go to `1em` but we cannot know that so we would convert it to `1pt` which is far away from the user intention.

**Other Changes:**

- remove duplication in the `CHANGELOG`
- put a `defautlValue` different from `0` to cover other possible edge cases with a default that makes "more sense"

Pre-checks:
* [x] Changelog entry added

Review:
* [ ] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -12,9 +12,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Improved
 - Screen readers are now able to announce the highlighted menu item of link comboboxes. #TINY-9280
-
-### Improved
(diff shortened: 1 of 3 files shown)


lorenzo-pomili:
got here for the right one: https://github.com/tinymce/tinymce/pull/8628

PR #8632, opened 2023-04-11
TINY-9467: Annotations should not apply to noneditable text in the root

spocke:
Related Ticket: TINY-9467

Description of Changes:
* Blocking annotator from adding and removing annotations on the current selection.

* [x] Discuss if we should do this or not first

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


yacodes (approved):

TheSpyder (approved):
I would've thought commenting on noneditable sections was fine. But I guess this is fine too :)

spocke:
> I would've thought commenting on noneditable sections was fine. But I guess this is fine too :)

The weird part is that in the noneditable root scenario you probably don't want to annotate things that are noneditable since there is no way to edit the thing unless you go into source mode but in say comments only mode then you want to be able to comment on things but not edit things. Maybe I should skip blocking this one. It will also be hard for us to see the difference between noneditable root mode and comment mode since in theory both could be used at the same time. I will check with other people tomorrow see what we think I'm leaning towards closing this PR and ticket.

lorenzo-pomili (approved):

PR #8652, opened 2023-04-14
TINY-9731: Add accessibility features to the accordion plugin

yacodes:
Related Ticket: TINY-9731

Description of Changes:
* Added toggling features and events;
* Added keyboard handlers;
* Added context menu.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/tinymce/src/plugins/accordion/main/ts/core/Actions.ts
```diff
@@ -29,4 +32,57 @@ const insertAccordion = (editor: Editor): void => {
...
+
+const removeAccordion = (editor: Editor): void => {
+  const details = Utils.getSelectedDetails(editor);
+  if (!details) {
+    return;
+  }
+
+  const { nextSibling } = details;
+  if (nextSibling) {
+    editor.selection.setCursorLocation(nextSibling, 0);
```
spocke, line 62: What happens if there is not nextSibling where does the selection go. Is (nextSibling, 0) always a valid spot? What if the next sibling is a table. I don't think `<table>|<tbody>` is a valid spot. You could maybe use:
```
editor.selection.select(nextSibling, true);
editor.selection.collapse(true);
```
That will find the first text node or BR spot then move the selection to that and then collapse. It would be good if we had better public apis to do this find the first suitable spot for a node.

modules/tinymce/src/plugins/accordion/test/ts/browser/AccordionPluginTest.ts
```diff
@@ -1,43 +1,63 @@
...
+  `<summary class="mce-accordion-summary">${summary}</summary>` +
+  `<div class="mce-accordion-body">${body}</div></details>`;
+
+interface InsertAccordionTest {
+  initialContent: string;
+  initialCursor: [number[], number];
+  assertContent: string;
+  assertCursor: [number[], number];
+}
+const testInsertingAccordion = (editor: Editor, test: InsertAccordionTest): void => {
```
spocke, line 23:
```suggestion

const testInsertingAccordion = (editor: Editor, test: InsertAccordionTest): void => {
```

modules/tinymce/src/plugins/accordion/test/ts/browser/AccordionPluginTest.ts
```diff
@@ -1,43 +1,63 @@
...
+  editor.setContent(test.initialContent);
+  TinySelections.setCursor(editor, ...test.initialCursor);
+  editor.execCommand('InsertAccordion');
+  TinyAssertions.assertContent(editor, test.assertContent);
+  assert.equal(editor.selection.getNode().nodeName, 'SUMMARY');
+  TinyAssertions.assertCursor(editor, ...test.assertCursor);
+};
+
+const testEvent = (editor: Editor, eventName: string, cmd: string, callback: (event: any) => void): void => {
```
spocke, line 32: Use a proper type here not `any` either a concrete one or T at least so that it can flow though this function.

modules/tinymce/CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - New optional `defaultExpandedIds` and `onToggleExpand` options to the `tree` component config. #TINY-9653
 - New optional `defaultSelectedId` option to the `tree` component config. #TINY-9715
 - New `accordion` plugin with the `InsertAccordion` command. #TINY-9730
+- Accessibility features to the `accordion` plugin. #TINY-9731
```
spocke, line 13: Since this has not bee released yet you don't need to add a changelog item.

- [ ] Remove changelog item

modules/tinymce/src/plugins/accordion/main/ts/api/Events.ts
```diff
@@ -0,0 +1,10 @@
+import Editor from 'tinymce/core/api/Editor';
+import { EditorEvent } from 'tinymce/core/api/util/EventDispatcher';
+
+export const fireToggleAccordionEvent =
+  (editor: Editor, element: HTMLDetailsElement, state: boolean): EditorEvent<{ element: HTMLDetailsElement; state: boolean }> =>
```
spocke, line 5: Make this a AccordionEvent type so that you can re-use them in tests etc.

spocke (changes requested):

modules/tinymce/src/plugins/accordion/main/ts/core/Utils.ts
```diff
@@ -0,0 +1,28 @@
+import Editor from 'tinymce/core/api/Editor';
+
+export const isSummary = (node?: Node): node is HTMLElement =>
+  node ? node.nodeName === 'SUMMARY' : false;
```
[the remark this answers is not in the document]
yacodes: That's cleaner, thanks!

modules/tinymce/src/plugins/accordion/main/ts/core/Keyboard.ts
```diff
@@ -0,0 +1,64 @@
...
+  editor.on('keydown', (event): void => {
+    if (event.shiftKey) {
+      return;
+    }
+    if (event.keyCode !== VK.ENTER) {
+      return;
+    }
+
+    const node = editor.selection.getNode();
+    if (!editor.dom.isEmpty(node)) {
```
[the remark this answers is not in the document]
yacodes: That's correct. I've updated the code to look for the nearest block parent, but it is not working correctly either. After applying formatting for the empty paragraph, this markup appears:

```html
<span id="_mce_caret" data-mce-bogus="1" data-mce-type="format-caret"><span style="text-decoration: underline;" data-mce-style="text-decoration: underline;">﻿</span></span>
```

And `editor.dom.isEmpty` returns `false` for this structure for some reason. Is it expected behavior or a bug in the `isEmpty` function?

modules/tinymce/src/plugins/accordion/main/ts/core/Keyboard.ts
```diff
@@ -0,0 +1,64 @@
...
+    const details = editor.dom.getParent(node, Utils.isDetails);
+    if (!details) {
+      return;
+    }
+
+    const paragraph = editor.dom.create('p');
+    paragraph.innerHTML = '<br data-mce-bogus="1" />';
+    details.insertAdjacentElement('afterend', paragraph);
+    editor.selection.setCursorLocation(paragraph, 0);
+    node.remove();
```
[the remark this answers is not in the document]
yacodes: Haven't thought about this state. Remove the `node.remove()` call to avoid this state.

Is there a better way to insert the paragraph here? Or this one is fine?

modules/tinymce/src/plugins/accordion/main/ts/core/Keyboard.ts
```diff
@@ -0,0 +1,64 @@
...
+    }
+    if (!Utils.isInSummary(editor)) {
+      return;
+    }
+    event.preventDefault();
+    editor.execCommand('ToggleAccordion');
+  });
+};
+
+const setupEnterKeyInAccordionBody = (editor: Editor): void => {
```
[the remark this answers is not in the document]
yacodes: Is it really necessary? I do not understand why it should be called on `insertParagraph` as well. Also is there really any other application for this function except the keyboard keypress?

Also it would make the code a bit messier, since we need to prevent the default event and thus this function must accept a callback or return a boolean or something like that.

modules/tinymce/src/plugins/accordion/test/ts/browser/AccordionPluginTest.ts
```diff
@@ -1,43 +1,63 @@
...
+  { open?: boolean; summary?: string; body?: string } = {}): string =>
+  `<details class="mce-accordion"${open ? ` open="open"` : ''}>` +
+  `<summary class="mce-accordion-summary">${summary}</summary>` +
+  `<div class="mce-accordion-body">${body}</div></details>`;
+
+interface InsertAccordionTest {
+  initialContent: string;
+  initialCursor: [number[], number];
+  assertContent: string;
+  assertCursor: [number[], number];
```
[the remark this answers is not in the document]
yacodes: Updated.

modules/tinymce/src/plugins/accordion/main/ts/core/Actions.ts
```diff
@@ -29,4 +32,57 @@ const insertAccordion = (editor: Editor): void => {
...
+  } else {
+    const paragraph = editor.dom.create('p');
```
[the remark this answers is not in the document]
yacodes: Extracted the function to the `Utils.ts` file.

modules/tinymce/src/plugins/accordion/main/ts/core/Actions.ts
```diff
@@ -1,15 +1,18 @@
...
 import Editor from 'tinymce/core/api/Editor';
 
+import * as Events from '../api/Events';
+import * as Utils from './Utils';
+
 const insertAccordion = (editor: Editor): void => {
-  const container = editor.selection.getNode();
-  if (container.nodeName === 'SUMMARY') {
-    const body = editor.dom.getNext(container, 'div.mce-accordion-body');
+  if (Utils.isInSummary(editor)) {
```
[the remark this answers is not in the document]
yacodes: For consistency sake it is correct. Fixed.

modules/tinymce/src/plugins/accordion/main/ts/core/Actions.ts
yacodes: Works great, thanks!

modules/tinymce/src/plugins/accordion/main/ts/api/Events.ts
yacodes: Exported the event types for the tests here.

modules/tinymce/src/plugins/accordion/main/ts/core/Keyboard.ts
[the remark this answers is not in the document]
spocke: I guess we could discuss this but currently there is similar code in the core and that code can be simulated using the commands. See this fiddle for an example: https://fiddle.tiny.cloud/A5iaab/1 the blockquote will be exited if you hit enter twice or if you issue the command twice the same code is running in the core.

So I guess we have four options:
1. Don't do the command logic. Then `InsertParagraph` will not be the same as a `enter` keypress
2. Do the command override, would involve some prevent defaults and hack in the plugin for those commands effected.
3. Move the enter code to the core so that any enter in the last empty block in a `details` element exits the `details` element similar to `blockquote`.
4. At least prepare for 2 but do 1 so that the code is separated and isolated from keyboard logic so that it can in theory be executed as a command.

It would have been nice if we had some form of enter registry of functions for this but it's so seldom we need to override the enter key other than in the core.

modules/tinymce/src/plugins/accordion/main/ts/core/Keyboard.ts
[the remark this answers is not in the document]
spocke: I wonder if that is a bug I think it's because it is filled with a zwnbsp character and that is treated as contents. I guess in theory we could look specifically for this element in the domUtils.isEmpty code and if it's just containing that zwnbsp character we treat that as nothing but I wonder if that would have other side effects.

modules/tinymce/src/plugins/accordion/main/ts/core/Keyboard.ts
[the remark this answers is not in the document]
yacodes: Moved the enter code to the core, to the `newline/InsertAccordion.ts` to be exactly. Hoping this is a correct way to go.

modules/tinymce/src/core/main/ts/newline/InsertBlock.ts
```diff
@@ -420,11 +421,15 @@ const insert = (editor: Editor, evt?: EditorEvent<KeyboardEvent>): void => {
   // Enter inside block contained within a LI then split or insert before/after LI
   if (containerBlockName === 'LI' && !ctrlKey) {
     const liBlock = containerBlock as HTMLLIElement;
-    parentBlock = liBlock;
+
     containerBlock = liBlock.parentNode;
     parentBlockName = containerBlockName;
   }
 
+  if (NodeType.isElement(containerBlock) && InsertAccordion.shouldPreventInsertParagraph(editor, shiftKey, parentBlock)) {
```
spocke, line 429: This is not really preventing inserting a paragraph since it later inserts a paragraph. So maybe it should be called `isAtEmptyBlockAtEndOfDetails` or something.

modules/tinymce/src/core/main/ts/newline/InsertAccordion.ts
```diff
@@ -0,0 +1,37 @@
...
+export const isAccordionBody = (node?: Node | null): node is HTMLDivElement =>
+  node?.nodeName === 'DIV' && (node as HTMLDivElement).classList.contains('mce-accordion-body');
+
+export const getAccordionRoot = (editor: Editor, element: Element): HTMLDetailsElement | null =>
+  editor.dom.getParent(element, isDetails);
+
+export const isInAccordionBody = (editor: Editor, element: Element): boolean =>
+  isAccordionBody(element) || Boolean(editor.dom.getParent(element, isAccordionBody));
+
+export const shouldPreventInsertParagraph = (editor: Editor, shiftKey: boolean, parentBlock: Element): boolean => {
```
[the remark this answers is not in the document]
yacodes: But this is the behavior intentional for accordion plugin. We've named the plugin like that exactly to distinguish it from the details element. Also won't adding this keyboard behavior for all details elements be a major change then?

modules/tinymce/src/core/main/ts/newline/InsertBlock.ts
yacodes: Renamed to `isLastEmptyBlockInAccordion`. If we'll decide to rename everything to `details` instead of `accordion`, then will rename it once again.

modules/tinymce/src/core/main/ts/newline/InsertAccordion.ts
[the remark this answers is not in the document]
spocke: In theory it could be wider I guess we could discuss this at the stand up. I think it's a bit unlikely that you want to create multiple empty paragraphs at the end of a details element. So if it would work as blockquotes I think that new behavior would be fine.

yacodes:
Added BACKSPACE behavior to prevent removal of the accordion's body element.

modules/tinymce/src/core/main/ts/newline/InsertAccordion.ts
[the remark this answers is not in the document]
yacodes: @spocke Updated the file to a more generic `InsertDetails`. It works just for simple cases. However this update made me think — do we really need a `div.mce-accordion-body` within the plugin? What purpose does it serve? Maybe we should stick to the standard and remove the custom accordion body whatsoever?

yacodes:
Added the selection to this PR as well. Now accordion is outlined with CSS if the selection is within it.

modules/tinymce/src/core/main/ts/api/dom/ControlSelection.ts
```diff
@@ -40,7 +40,7 @@ interface SelectedResizeHandle extends ResizeHandle {
 }
 
 const elementSelectionAttr = 'data-mce-selected';
-const controlElmSelector = 'table,img,figure.image,hr,video,span.mce-preview-object';
+const controlElmSelector = 'table,img,figure.image,hr,video,span.mce-preview-object,details';
```
spocke, line 43: Could you add test for this change. There is a ControlSelectionTest

modules/tinymce/src/core/main/ts/newline/InsertDetails.ts
spocke, line 1: Maybe name this module DetailsNewline or InsertDetailsNewline or something since it's about inserting blocks or newlines in details not inserting details elements. The other modules in the newline dir insert the thing that it describes.

modules/tinymce/src/core/main/ts/api/dom/ControlSelection.ts
yacodes: Added the test for `data-mce-selected` attribute for details element.

modules/tinymce/src/core/main/ts/newline/InsertDetails.ts
yacodes: Updated.

modules/tinymce/src/core/main/ts/newline/InsertAccordion.ts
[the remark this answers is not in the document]
TheSpyder: I'm in favour of generic `details` support, but product has been very specific that we want a crafted UI just for accordions inserted by our plugin

modules/tinymce/src/core/main/ts/newline/InsertDetailsNewLine.ts
```diff
@@ -0,0 +1,34 @@
...
+import * as NewLineUtils from './NewLineUtils';
+
+export const isDetails = (node?: Node | null): node is HTMLDetailsElement =>
+  node?.nodeName === 'DETAILS';
+
+export const getDetailsRoot = (editor: Editor, element: Element): HTMLDetailsElement | null =>
+  editor.dom.getParent(element, isDetails);
+
+export const isLastEmptyBlockInDetails = (editor: Editor, shiftKey: boolean, element: Element): boolean => {
+  switch (true) {
```
TheSpyder, line 11: What's wrong with a simple boolean expression?

TheSpyder (changes requested):

modules/tinymce/src/plugins/accordion/main/ts/core/Utils.ts
```diff
@@ -0,0 +1,54 @@
...
+
+export const isOpen = (details: HTMLDetailsElement): boolean =>
+  details.hasAttribute('open');
+
+export const isInSummary = (editor: Editor): boolean => {
+  const node = editor.selection.getNode();
+  return isSummary(node) || Boolean(editor.dom.getParent(node, isSummary));
+};
+
+export const getSelectedDetails = (editor: Editor): HTMLDetailsElement | null => {
```
vpyshnenko, line 18: consider to return `Optional<HTMLDetailsElement>`

modules/tinymce/src/plugins/accordion/main/ts/core/Actions.ts
```diff
@@ -29,4 +27,55 @@ const insertAccordion = (editor: Editor): void => {
...
+
+  details.remove();
+};
+
+const toggleAllAccordions = (editor: Editor, state?: boolean): void => {
+  const accordions = Array.from(editor.getBody().querySelectorAll('details'));
+  if (accordions.length === 0) {
+    return;
+  }
+  const shouldOpen = state ?? !Utils.isOpen(accordions[0]);
```
vpyshnenko, line 71: Shouldn't each accordion invert the current state when no state is applied instead of depending on the state of the first one?

modules/tinymce/src/plugins/accordion/test/ts/browser/AccordionPluginTest.ts
```diff
@@ -121,12 +130,154 @@ describe('browser.tinymce.plugins.accordion.AccordionPluginTest', () => {
...
+  });
+
+  it('TINY-9731: Toggle all accordion elements with an argument', () => {
+    const editor = hook.editor();
+    editor.setContent([ createAccordion({ open: true }), createAccordion({ open: true }) ].join(''));
+    TinySelections.setCursor(editor, [ 0, 0, 0 ], 0);
+    editor.execCommand('ToggleAllAccordions', false, false);
+    TinyAssertions.assertContentPresence(editor, { 'details:not([open="open"])': 2 });
+    TinyAssertions.assertCursor(editor, [ 0, 0, 0 ], 0);
+    editor.execCommand('ToggleAllAccordions', true, true);
```
vpyshnenko, line 193: Apparently no UI involved for this command
```suggestion
    editor.execCommand('ToggleAllAccordions', false, true);
```

modules/tinymce/src/plugins/accordion/test/ts/browser/AccordionPluginTest.ts
```diff
@@ -121,12 +130,154 @@ describe('browser.tinymce.plugins.accordion.AccordionPluginTest', () => {
...
+    TinyAssertions.assertContentPresence(editor, { 'details[open="open"]': 2 });
+    TinyAssertions.assertCursor(editor, [ 0, 0, 0 ], 0);
+    editor.execCommand('ToggleAllAccordions', true, true);
```
vpyshnenko, line 196:
```suggestion
    editor.execCommand('ToggleAllAccordions', false, true);
```

modules/tinymce/src/plugins/accordion/test/ts/browser/AccordionPluginTest.ts
```diff
@@ -121,12 +130,154 @@ describe('browser.tinymce.plugins.accordion.AccordionPluginTest', () => {
...
+    });
+    testEvent(editor, 'ToggledAccordion', 'ToggleAccordion', (event: ToggledAccordionEvent) => {
+      assert.equal(event.element.nodeName, 'DETAILS');
+      assert.isTrue(event.state);
+    });
+  });
+
+  it('TINY-9731: Emit the "ToggledAllAccordions" event', () => {
+    const editor = hook.editor();
+    editor.setContent(createAccordion({ summary: 'tiny' }));
```
vpyshnenko, line 217: add a case for several accordions to be in the content

modules/tinymce/src/plugins/accordion/main/ts/core/Utils.ts
yacodes: It is much better with Optional. Thanks for the suggestion!

modules/tinymce/src/plugins/accordion/main/ts/core/Actions.ts
yacodes: Makes sense. Triggering the desired state is still possible with the additional argument, so by default it can toggle each accordion's state separately.

modules/tinymce/src/plugins/accordion/test/ts/browser/AccordionPluginTest.ts
yacodes: Fixed.

modules/tinymce/src/plugins/accordion/test/ts/browser/AccordionPluginTest.ts
yacodes: Updated the test with 2 accordions. Added state checks as well.

vpyshnenko:
If the accordion is located at the beginning of the document, I'm having trouble finding a way to add a new line above it.

yacodes:
> If the accordion is located at the beginning of the document, I'm having trouble finding a way to add a new line above it.

Ain't it the same with lists? I tried modifier keys and seems like it is impossible with lists as well. Am I missing something here?

vpyshnenko:
In the case of the list located at the beginning of the document, the sequence of [Enter, Up, Backspace] should add a new line above the list. However, we lack the equivalent of VS Code Cmd+Shift+Enter shortcut

modules/tinymce/src/core/main/ts/newline/InsertDetailsNewLine.ts
TheSpyder: It's not a style we use anywhere else. More readable for you may be a confusing mess for someone looking at this months or years from now. If you want to propose a new style, that's a team decision.

Injecting some personality is fine, but at a general level I try to encourage a similar coding style across the entire product for ease of maintenance. Every time you create a special unique snowflake it leaves a headache for someone to find later.

yacodes:
Added an additional task to handle inserting a paragraph before the first accordion TINY-9827.

modules/tinymce/src/plugins/accordion/main/ts/core/Actions.ts
```diff
@@ -13,18 +13,21 @@ const insertAccordion = (editor: Editor): void => {
...
   const summaryText = editor.dom.encode(editor.selection.getRng().toString() || editor.translate('Accordion summary...'));
   const bodyText = editor.dom.encode(editor.translate('Accordion body...'));
-  editor.insertContent(`<details data-mce-id="${uid}" class="mce-accordion" open="open"><summary class="mce-accordion-summary">${summaryText}</summary><p>${bodyText}</p></details>`);
 
-  const details = editor.dom.select(`[data-mce-id="${uid}"]`)[0];
-  if (!details) {
-    return;
-  }
-  details.removeAttribute('data-mce-id');
+  editor.undoManager.transact(() => {
```
vpyshnenko, line 17: `insertContent` just calls the `mceInsertContent` command and all commands are in undo transactions, look topic in [Slack](https://cksource.slack.com/archives/C04KXE6PMRB/p1683319674177149?thread_ts=1683309133.416439&cid=C04KXE6PMRB
)

TheSpyder (approved):

modules/tinymce/src/plugins/accordion/main/ts/ui/Buttons.ts
```diff
@@ -20,6 +14,26 @@ const register = (editor: Editor): void => {
...
   editor.ui.registry.addMenuItem('accordion', { icon: 'accordion', text: 'Accordion', onSetup: onSetup(editor), onAction });
+
+  editor.ui.registry.addToggleButton('accordiontoggle', {
+    text: 'Toggle accordion',
+    tooltip: 'Toggle accordion',
+    onAction: () => editor.execCommand('ToggleAccordion')
+  });
+
+  editor.ui.registry.addToggleButton('accordionremove', {
+    text: 'Remove accordion',
```
vpyshnenko, line 25: I noticed in strings for translations that it should be "Delete accordion"
https://github.com/tinymce/tinymce-enterprise-language-packs/blob/main/src/text/pot/core/tinymce_core.pot#L1258

yacodes: Nice point! Updated.

modules/tinymce/src/core/main/ts/newline/InsertDetailsNewLine.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export const isDetails = (node?: Node | null): node is HTMLDetailsElement =>
+  node?.nodeName === 'DETAILS';
+
+export const getDetailsRoot = (editor: Editor, element: Element): HTMLDetailsElement | null =>
+  editor.dom.getParent(element, isDetails);
+
+export const isLastEmptyBlockInDetails = (editor: Editor, shiftKey: boolean, element: Element): boolean =>
+  !shiftKey &&
+  element.nodeName === 'P' &&
```
spocke, line 12: Use the forced_root_block option for this `Options.getForcedRootBlock`.

modules/tinymce/src/core/main/ts/newline/InsertDetailsNewLine.ts
```diff
@@ -0,0 +1,28 @@
...
+  element.parentNode?.lastChild === element &&
+  editor.dom.isEmpty(element) &&
+  isDetails(element.parentNode);
+
+export const insertNewLine = (editor: Editor, createNewBlock: (name: string) => Element, parentBlock: Element): void => {
+  const newBlock = createNewBlock('p');
```
spocke, line 18: Use the forced_root_block option for this `Options.getForcedRootBlock`.

spocke (approved):

spocke (approved):

PR #8681, opened 2023-04-25
TINY-9696: Add padding when removing elements if required.

HAFRMO:
Related Ticket: TINY-9696

Description of Changes:
Avoid leaving p-blocks empty and add padding if needed.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -26,7 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Table toolbar was visible even if the table was within a noneditable host element. #TINY-9664
 - Quickbar toolbars was shown for noneditable contents in a noneditable root. #TINY-9460
 - Show the calculated height and width of media embed elements in the `media` plugin dialog. #TINY-8714
-- Removing an image that failed to upload from an empty paragraph will not leave the paragraph without the correct padding. #TINY-9696
+- Removing an image that failed to upload from an empty paragraph would leave the paragraph without the correct padding. #TINY-9696
 - Allow a media embed element to be correctly resized when using the `media` plugin dialog by converting the media embed to a standalone iframe. #TINY-8714
 - Inline alert in the "Search and Replace" dialog persisted when it wasn't necessary. #TINY-9704
 - Context toolbars displayed the incorrect status for the `advlist` plugin buttons. #TINY-9680
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Table toolbar was visible even if the table was within a noneditable host element. #TINY-9664
 - Quickbar toolbars was shown for noneditable contents in a noneditable root. #TINY-9460
 - Show the calculated height and width of media embed elements in the `media` plugin dialog. #TINY-8714
+- Removing an image that failed to upload from an empty paragraph will not leave the paragraph without the correct padding. #TINY-9696
 - Allow a media embed element to be correctly resized when using the `media` plugin dialog by converting the media embed to a standalone iframe. #TINY-8714
 - Inline alert in the "Search and Replace" dialog persisted when it wasn't necessary. #TINY-9704
 - Context toolbars displayed the incorrect status for the `advlist` plugin buttons. #TINY-9680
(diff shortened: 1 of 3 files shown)


modules/tinymce/CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Table toolbar was visible even if the table was within a noneditable host element. #TINY-9664
 - Quickbar toolbars was shown for noneditable contents in a noneditable root. #TINY-9460
 - Show the calculated height and width of media embed elements in the `media` plugin dialog. #TINY-8714
+- Removing an image that failed to upload from an empty paragraph will not leave the paragraph without the correct padding. #TINY-9696
```
vpyshnenko, line 29:
```suggestion
- Removing an image that failed to upload from an empty paragraph would leave the paragraph without the correct padding. #TINY-9696
```

vpyshnenko (approved):

modules/tinymce/src/core/main/ts/api/EditorUpload.ts
```diff
@@ -98,6 +98,15 @@ interface EditorUpload {
   destroy: () => void;
 }
 
+const isEmptyForPadding = (element: SugarElement<any>) =>
+  !Traverse.hasChildNodes(element) && SugarNode.isTag('p')(element);
```
spocke, line 102: There is more cases than `P` here basically any text block element. Things like TD or LI will be hard since they are part of a bigger structure. So you need the schema here and use schema.getTextBlocks() then check if the SugarNode.name is in that map. Then this will work for H1, DIV etc etc. Add tests for a few other types as well.

- [ ] Expand to more than just paragraphs

modules/tinymce/src/core/main/ts/api/EditorUpload.ts
```diff
@@ -213,7 +222,11 @@ const EditorUpload = (editor: Editor): EditorUpload => {
         if (imagesToRemove.length > 0 && !Rtc.isRtc(editor)) {
           editor.undoManager.transact(() => {
             Arr.each(imagesToRemove, (element) => {
-              editor.dom.remove(element);
+              const sugarElement = SugarElement.fromDom(element);
```
spocke, line 225: You can use SugarElements.fromDom here that and pass that into Arr.each then the element will be sugar. Then use `element.dom.src` not Attr.get since if I recall src is always the full path while attributes is the value in the dom.

modules/tinymce/src/core/test/ts/browser/EditorUploadTest.ts
```diff
@@ -429,6 +429,22 @@ describe('browser.tinymce.core.EditorUploadTest', () => {
     return editor.uploadImages().then(() => assertEventsLength(1));
   });
 
+  it('TINY-9696: Removing an image should not leave the containing block without bogus element.', async () => {
+    const editor = hook.editor();
+    setInitialContent(editor, `<p>A</p><p></p><p>${imageHtml(testBlobDataUri)}</p>`);
+
+    editor.options.set('images_upload_handler', () => {
```
spocke, line 436: Are we sure setting this doesn't effect other tests after this is they don't set this. So we don't get a pivot flake here.

spocke (approved):

modules/tinymce/src/core/test/ts/browser/EditorUploadTest.ts
HAFRMO: Both the previous and subsequent tests set this value on their own, so it should not cause any more issues that wouldn't already be caused.

modules/tinymce/CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Table toolbar was visible even if the table was within a noneditable host element. #TINY-9664
 - Quickbar toolbars was shown for noneditable contents in a noneditable root. #TINY-9460
 - Show the calculated height and width of media embed elements in the `media` plugin dialog. #TINY-8714
+- Removing an image that failed to upload from an empty paragraph would leave the paragraph without the correct padding. #TINY-9696
```
ltrouton, line 29: I initially thought this PR was fixing something to do with style padding when reading the entry but that could just be where my brain went. Might be worth mentioning `br` specifically for clarity.
```suggestion
- Removing an image that failed to upload from an empty paragraph would leave the paragraph without a padding br. #TINY-9696
```

ltrouton (approved):

PR #8690, opened 2023-04-28
TINY-9314: Make Directionality plugin work with direction CSS property

mkzhx:
Related Ticket: TINY-9314

Description of Changes:
* Directionality plugin adds an inline `direction` style if the `dir` attribute has been taken precedence by either an inline `direction` style, or a `direction` style from the stylesheet
* Clean up `dir` attribute removal logic

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/plugins/directionality/test/ts/browser/DirectionStyleTest.ts
+++ b/modules/tinymce/src/plugins/directionality/test/ts/browser/DirectionStyleTest.ts
@@ -17,95 +17,97 @@ describe('browser.tinymce.plugins.directionality.DirectionStyleTest', () => {
     TinyAssertions.assertContent(editor, expected);
   };
 
+  const baseSettings = {
+    plugins: 'directionality',
+    toolbar: 'ltr rtl',
+    base_url: '/project/tinymce/js/tinymce'
+  };
+
   context('TINY-9314: Inline direction style', () => {
-    const hook = TinyHooks.bddSetup<Editor>({
-      plugins: 'directionality',
-      toolbar: 'ltr rtl',
-      base_url: '/project/tinymce/js/tinymce'
-    }, [ Plugin ]);
+    const hook = TinyHooks.bddSetup<Editor>(baseSettings, [ Plugin ]);
 
     it('Applying ltr with dir="rtl" and direction: rtl', () => testDirectionStyle(
       hook.editor(),
-      '<p dir="rtl" style="direction: rtl;">RTL Style</p>',
+      '<p dir="rtl" style="direction: rtl;">Lorem ipsum</p>',
       'ltr',
-      '<p style="direction: ltr;">RTL Style</p>'
+      '<p style="direction: ltr;">Lorem ipsum</p>'
     ));
 
     it('Applying rtl with dir="ltr" and direction: ltr', () => testDirectionStyle(
       hook.editor(),
-      '<p dir="ltr" style="direction: ltr;">LTR Style</p>',
+      '<p dir="ltr" style="direction: ltr;">Lorem ipsum</p>',
       'rtl',
-      '<p dir="rtl" style="direction: rtl;">LTR Style</p>'
+      '<p dir="rtl" style="direction: rtl;">Lorem ipsum</p>'
     ));
 
     it('Applying ltr with dir="ltr" and direction: ltr', () => testDirectionStyle(
       hook.editor(),
-      '<p dir="ltr" style="direction: ltr;">RTL Style</p>',
+      '<p dir="ltr" style="direction: ltr;">Lorem ipsum</p>',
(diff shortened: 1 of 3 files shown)


MitchCTiny (approved):

modules/tinymce/src/plugins/directionality/test/ts/browser/DirectionStyleTest.ts
```diff
@@ -0,0 +1,271 @@
...
+      '<p dir="rtl" style="direction: rtl;">Lorem ipsum</p>',
+      'ltr',
+      '<p style="direction: ltr;">Lorem ipsum</p>'
+    ));
+
+    it('Applying rtl with dir="ltr" and direction: ltr', () => testDirectionStyle(
+      hook.editor(),
+      '<p dir="ltr" style="direction: ltr;">Lorem ipsum</p>',
+      'rtl',
+      '<p dir="rtl" style="direction: rtl;">Lorem ipsum</p>'
```
ltrouton, line 40: In the above `it` block the `dir` attribute is removed, but for this test case it isn't. I am curious what is causing that discrepancy. My guess it is the change here: https://github.com/tinymce/tinymce/pull/8690/files#diff-6399166c6f9c744f6fa49c73dfc28d1557920af20e49abdf0b842d8c7f40a455R28

I would expect using the ltr or rtl buttons to have equivalent HTML structure but maybe I am missing something.

ltrouton (commented):
Nice work on the tests.

modules/tinymce/src/plugins/directionality/test/ts/browser/DirectionStyleTest.ts
mkzhx: @ltrouton This is expected and aligns with previous behaviour. Since the parent (`body`) is in `ltr`, the `dir` attribute of the previous test was removed as it is unnecessary. In this case it is kept as the element has a different `dir` to its parent. The change you linked is a refactoring that should not change existing behaviour.

ArvinJ-H (approved):

modules/tinymce/src/plugins/directionality/test/ts/browser/DirectionStyleTest.ts
```diff
@@ -0,0 +1,271 @@
...
+  };
+
+  context('Inline direction style', () => {
+    const hook = TinyHooks.bddSetup<Editor>(baseSettings, [ Plugin ]);
+
+    it('TINY-9314: Applying ltr with dir="rtl" and direction: rtl', () => testDirectionStyle(
+      hook.editor(),
+      '<p dir="rtl" style="direction: rtl;">Lorem ipsum</p>',
+      'ltr',
+      '<p style="direction: ltr;">Lorem ipsum</p>'
```
spocke, line 33: In this case it could unset the style and favor the attribute.

modules/tinymce/src/plugins/directionality/test/ts/browser/DirectionStyleTest.ts
```diff
@@ -0,0 +1,271 @@
...
+    ));
+
+    it('TINY-9314: Applying rtl with dir="ltr" and direction: ltr', () => testDirectionStyle(
+      hook.editor(),
+      '<p dir="ltr" style="direction: ltr;">Lorem ipsum</p>',
+      'rtl',
+      '<p dir="rtl" style="direction: rtl;">Lorem ipsum</p>'
```
spocke, line 40: Seems a little clunky to have both a style and a dir attribute basically double information. If we set the style we should unset the attribute since it has lower presence normally but I guess we could detect that make a smart setDir that checks if it needs to set the attribute, style by checking the runtime style value after setting either and unsetting the other.

spocke (approved):

modules/tinymce/src/plugins/directionality/test/ts/browser/DirectionStyleTest.ts
mkzhx: This was a side effect favouring to keep the `direction` property if it was present initially. However, I am going to change the implementation to remove the `direction` property if it is not necessary, as detailed in the below conversation. The `dir` attribute will be removed regardless as its parent is in `ltr` already, to keep things minimal.

modules/tinymce/src/plugins/directionality/test/ts/browser/DirectionStyleTest.ts
mkzhx: @spocke Yes I agree, the aim of the plugin is to keep the attributes/styles the minimal it needs to be, so I will make the change in my implementation to remove the `direction` style property if it is not needed.

PR #8699, opened 2023-05-04
TINY-9731: Add details_initial_state and details_serialized_state options

yacodes:
Related Ticket: TINY-9731

Description of Changes:
* Add new `details_initial_state` and `details_serialized_state` options.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/tinymce/src/core/main/ts/selection/DetailsElement.ts
```diff
@@ -10,20 +11,30 @@ const preventSummaryToggle = (editor: Editor): void => {
...
   editor.parser.addNodeFilter('details', (elms) => {
+    const initialStateOption = Options.getDetailsInitialState(editor);
     Arr.each(elms, (details) => {
-      details.attr('data-mce-open', details.attr('open'));
-      details.attr('open', 'open');
+      if (initialStateOption === 'expanded') {
+        details.attr('open', 'open');
+      } else if (initialStateOption === 'collapsed') {
+        details.attr('open', null);
+      } else if (initialStateOption === 'inherited') {
```
vpyshnenko, line 22: Could we just skip reassigning the same value?

vpyshnenko (approved):

modules/tinymce/src/core/main/ts/selection/DetailsElement.ts
yacodes: Yeah, it is totally redundant :)

modules/tinymce/src/core/main/ts/api/Options.ts
```diff
@@ -790,6 +790,22 @@ const register = (editor: Editor): void => {
     default: true
   });
 
+  registerOption('details_initial_state', {
+    processor: (value) => {
+      const valid = Arr.contains([ 'inherited', 'collapsed', 'expanded' ], value);
```
spocke, line 795: I wonder if `inherited` could be called `keep` since it doesn't force one or the other it just keeps what ever value is currently present. However I guess that might be confusing if you open a details and then save then it didn't keep the original state. Another alternative could be to call it `current` since it serializes or uses the current value. Not sure if these suggestions are better just throwing them out here.

spocke (approved):
Looks great. :+1:

modules/tinymce/src/core/main/ts/api/Options.ts
vpyshnenko: Perhaps we could implement the `inherited` behavior as the default when this option is unspecified. Consequently, this option would allow either 'collapsed' or 'expanded'.

HAFRMO (approved):

modules/tinymce/src/core/main/ts/api/Options.ts
yacodes: I'd prefer to keep the `inherited` name as a default one rather then omitting it at all. I can imagine the hassle to remove a key from the options object just to force the default behavior. Providing `undefined` or `null` to access this behavior feels wrong as well. It is much easier to use a string here in my opinion.

Speaking of `keep` or `current` I am not really sure what is the best naming here...

PR #8708, opened 2023-05-09
TINY-9601:  Create `DataTransfer` object for `DragEvent.dataTransfer` property of drag and drop events

mkzhx:
Related Ticket: TINY-9601

Description of Changes:
* Code quality changes to agar's DataTransfer modules (originally intended for implementation but later abandoned in favour of new implementation based on native constructor).
* New DataTransfer and related modules in dragster to create `DataTransfer` objects. Related modules simulate:
  * an [event-dependent permission mode](https://html.spec.whatwg.org/dev/dnd.html#concept-dnd-rw)
  * [dragstart-only modifications to `effectAllowed`](https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer/effectAllowed); and
  * the need to create clones of `DataTransfer` objects with separate references and scopes when dispatching them for different drag events in the same drag-drop operation to prevent collateral mutations.
* Refactor `DragEvents` module and replace multiple event creation functions with single `makeDragEvent` function.
* Create and dispatch `dataTransfer` objects in drag-and-drop events and use it as the source-of-truth for content insertion on drop, as users can manipulate the data in the `dataTransfer` object on `dragstart`.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/dragster/src/main/ts/ephox/dragster/file/FileList.ts
```diff
@@ -1,7 +1,11 @@
 import { Arr } from '@ephox/katamari';
 
 const createFileList = (inputFiles: File[]): FileList => {
-  const files = {
+  const files: {
```
spocke, line 4: Can this be just FileList since the object is created it should not clash with the readonly property of the length. Or is the null stuff from that type annoying here?

mkzhx:
> It seems that browser now days support constructing [DataTransfer](https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer/DataTransfer) instances. I wonder if we could use native support instead. I guess the `mode` part is not exposed but maybe these types of dataTransfer instance is all writable all the time?

@spocke I think it is a good idea to use the native browser constructor. I did some testing and found that on native events, the dataTransfer on `drop` is read-only and on `dragend` is protected. However, I think I can emulate this behavior if I create a new `DataTransferInstance` module in Dragster. My plan is:
1. Move the custom dataTransfer code back to agar and remove the references to dragster in agar.
2. Create new `DataTransferInstance` module in dragster which under the hood mutates a natively constructed`DataTransfer` object.

modules/agar/tsconfig.json
```diff
@@ -18,6 +18,7 @@
     "src/test/ts"
   ],
   "references": [
+    { "path": "../dragster" },
```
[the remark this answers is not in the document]
mkzhx: Reference to dragster removed as the original `DataTransfer` module is moved back to agar.

mkzhx:
@spocke Thanks for the suggestions. The spec states that the mode should be respected depending on the event type -  https://html.spec.whatwg.org/dev/dnd.html#concept-dnd-rw. Another consideration is that `effectAllowed` can only be modified in a `dragstart` event https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer/effectAllowed. 

So I created a new module in dragster which mostly uses native implementation from the `DataTransfer()` constructor but also respects the event-dependent mode and `effectAllowed` considerations. I also replaced `DataTransferManager` with functional-style dragster modules.

modules/agar/src/test/ts/browser/datatransfer/DataTransferItemTest.ts
```diff
@@ -1,42 +1,39 @@
...
-import { Pipeline } from 'ephox/agar/api/Pipeline';
-import { Step } from 'ephox/agar/api/Step';
 import { createDataTransfer } from 'ephox/agar/datatransfer/DataTransfer';
 import { createDataTransferItemFromFile, createDataTransferItemFromString, getData } from 'ephox/agar/datatransfer/DataTransferItem';
 
-UnitTest.asynctest('DataTransferItemTest', (success, failure) => {
-  Pipeline.async({}, [
-    Logger.t('Create transfer item from file', Step.sync(() => {
-      const fileItem = createDataTransferItemFromFile(createDataTransfer(), createFile('a.txt', 1234, new Blob([ '123' ], { type: 'text/plain' })));
+describe('atomic.dragster.datatransfer.DataTransferItemTest', () => {
```
spocke, line 9:
```suggestion
describe('atomic.datatransfer.DataTransferItemTest', () => {
```

modules/dragster/src/main/ts/ephox/dragster/datatransfer/DataTransfer.ts
```diff
@@ -0,0 +1,131 @@
...
+  Arr.each(original.files, (file) => clone.items.add(file));
+
+  // Set event type
+  getEvent(original).each((type) => {
+    setEvent(clone, type);
+  });
+
+  // Set mode
+  originalMode.each((mode) => {
+    setMode(original, mode);
```
spocke, line 124: Is this correct why is it setting the mode on the original?

modules/dragster/src/main/ts/ephox/dragster/datatransfer/DataTransfer.ts
```diff
@@ -0,0 +1,131 @@
...
+  getDragImage(original).each((imageData) => clone.setDragImage(imageData.image, imageData.x, imageData.y));
+
+  // Copy string data
+  Arr.each(original.types, (type) => {
+    if (type !== 'Files') {
+      clone.setData(type, original.getData(type));
+    }
+  });
+
+  // Copy files
```
spocke, line 114: Is some of these just stating the obvious?

modules/dragster/src/main/ts/ephox/dragster/datatransfer/DragImage.ts
```diff
@@ -0,0 +1,24 @@
+import { Id, Optional } from '@ephox/katamari';
+
+interface DragImageData {
+  image: Element;
+  x: number;
+  y: number;
+}
```
spocke, line 7: Maybe indicate that this is to be an immutable structure.

```suggestion
interface DragImageData {
  readonly image: Element;
  readonly x: number;
  readonly y: number;
}
```

modules/dragster/src/main/ts/ephox/dragster/datatransfer/Event.ts
```diff
@@ -0,0 +1,45 @@
...
+};
+
+const setEvent = (transfer: DataTransfer, type: Event): void => mkSetEventFn(type)(transfer);
+
+const setDragstartEvent = mkSetEventFn(Event.Dragstart);
+const setDropEvent = mkSetEventFn(Event.Drop);
+const setDragendEvent = mkSetEventFn(Event.Dragend);
+
+const checkEvent = (expectedType: Event) => (transfer: DataTransfer): boolean => {
+  const dt: any = transfer;
```
spocke, line 28: Maybe try to avoid `any` you treat this as a record so should be `Record<string, any>` or `Record<string, unknown>`.

modules/dragster/src/main/ts/ephox/dragster/datatransfer/Items.ts
```diff
@@ -0,0 +1,37 @@
+import { Type } from '@ephox/katamari';
+
+import { isInReadWriteMode } from './Mode';
```
spocke, line 3: Maybe namespace import this since `Mode.isInReadWriteMode` is a bit more clear that it's a external module function. This module is small now but just a good practice that I've been told that we should follow that I sometimes forget to follow my self. :smile:

modules/dragster/src/main/ts/ephox/dragster/datatransfer/Mode.ts
```diff
@@ -0,0 +1,45 @@
...
+
+const modeId = Id.generate('mode');
+
+const getMode = (transfer: DataTransfer): Optional<Mode> => {
+  const dt: any = transfer;
+  return Optional.from(dt[modeId]);
+};
+
+const mkSetModeFn = (mode: Mode) => (transfer: DataTransfer): void => {
+  const dt: any = transfer;
```
spocke, line 17: `Record<string, any>` casting.

modules/dragster/src/main/ts/ephox/dragster/datatransfer/Mode.ts
```diff
@@ -0,0 +1,45 @@
...
+};
+
+const setMode = (transfer: DataTransfer, mode: Mode): void => mkSetModeFn(mode)(transfer);
+
+const setReadWriteMode = mkSetModeFn(Mode.ReadWrite);
+const setReadOnlyMode = mkSetModeFn(Mode.ReadOnly);
+const setProtectedMode = mkSetModeFn(Mode.Protected);
+
+const checkMode = (expectedMode: Mode) => (transfer: DataTransfer): boolean => {
+  const dt: any = transfer;
```
spocke, line 28: `Record<string, any>` casting.

modules/dragster/src/test/ts/atomic/datatransfer/DataTransferCloneTest.ts
```diff
@@ -0,0 +1,92 @@
...
+import { KAssert } from '@ephox/katamari-assertions';
+import { PlatformDetection } from '@ephox/sand';
+import { assert } from 'chai';
+
+import { cloneDataTransfer, createDataTransfer, getDragImage } from 'ephox/dragster/datatransfer/DataTransfer';
+import { Event, getEvent, setDragstartEvent } from 'ephox/dragster/datatransfer/Event';
+import { getMode, Mode, setMode, setReadOnlyMode, setReadWriteMode } from 'ephox/dragster/datatransfer/Mode';
+
+describe('atomic.dragster.datatransfer.DataTransferCloneTest', () => {
+  const dragImage = {
```
spocke, line 11: If you expose the DragImage type then that can be used in this test so anytime we add some new property, or list references you would benefit from having a proper type on it.

modules/dragster/src/test/ts/atomic/datatransfer/DataTransferCloneTest.ts
```diff
@@ -0,0 +1,92 @@
...
+    setDragstartEvent(dataTransfer);
+    dataTransfer.dropEffect = 'copy';
+    dataTransfer.effectAllowed = 'copy';
+    dataTransfer.setData('text/plain', '123');
+    dataTransfer.setData('text/html', '<p>123</p>');
+    dataTransfer.setDragImage(dragImage.image, dragImage.x, dragImage.y);
+    dataTransfer.items.add(testFile1);
+    dataTransfer.items.add(testFile2);
+
+    // set mode
```
spocke, line 35: Stating the obvious if the code and the comment is basically the same the comment is not really needed.

modules/dragster/src/test/ts/atomic/datatransfer/DataTransferCloneTest.ts
```diff
@@ -0,0 +1,92 @@
+import { describe, it } from '@ephox/bedrock-client';
+import { KAssert } from '@ephox/katamari-assertions';
+import { PlatformDetection } from '@ephox/sand';
+import { assert } from 'chai';
+
+import { cloneDataTransfer, createDataTransfer, getDragImage } from 'ephox/dragster/datatransfer/DataTransfer';
+import { Event, getEvent, setDragstartEvent } from 'ephox/dragster/datatransfer/Event';
+import { getMode, Mode, setMode, setReadOnlyMode, setReadWriteMode } from 'ephox/dragster/datatransfer/Mode';
+
+describe('atomic.dragster.datatransfer.DataTransferCloneTest', () => {
```
spocke, line 10: Not sure these tests are technically atomic since it uses the dom. The intent at least in the past was that atomic tests where tests that was just plain JS and could in theory be executed without a browser context like in node.

modules/tinymce/src/core/test/ts/browser/DragDropOverridesTest.ts
```diff
@@ -194,6 +280,189 @@ describe('browser.tinymce.core.DragDropOverridesTest', () => {
       await pAssertNotification('Dropped file type is not supported');
     });
 
+    it('TINY-9601: dataTransfer data modified in dragstart event should persist in drop and dragend events and in content inserted on drop', async () => {
+      const editor = hook.editor();
+
+      const testImage = {
```
spocke, line 286: Is there a type we could use for these if not then add one just to the test. Like TestImage or something.

modules/tinymce/src/core/test/ts/browser/DragDropOverridesTest.ts
```diff
@@ -30,14 +41,61 @@ describe('browser.tinymce.core.DragDropOverridesTest', () => {
...
+          assert.equal(dataTransfer.effectAllowed, spec.effectAllowed, `Expected dataTransfer on "${eventType}" event to have effectAllowed`);
+          KAssert.eqOptional(`Expected dataTransfer on "${eventType}" event to have dragImage`, Optional.from(spec.dragImage), DataTransfer.getDragImage(dataTransfer));
+
+          const dtFiles = dataTransfer.files;
+          if (spec.files.length === 0) {
+            assert.isNull(dtFiles.item(0), `Expected dataTransfer on "${eventType}" event to have no files`);
+          } else {
+            Arr.each(spec.files, (specFile) => {
+              let fileExists = false;
+              for (let i = 0, l = dtFiles.length; i < l; ++i) {
```
spocke, line 80: dtFiles is an array like so you should be able to use use Arr.each on that as well. But since this checks for a file if it exists `Arr.find` might be better not sure.

modules/tinymce/src/core/test/ts/browser/DragDropOverridesTest.ts
```diff
@@ -30,14 +41,61 @@ describe('browser.tinymce.core.DragDropOverridesTest', () => {
...
+            // Temporarily set to read-only to allow checking of data
+            DataTransferMode.setReadOnlyMode(dataTransfer);
+          }
+
+          Arr.each(spec.data, ({ type, value }) => assert.equal(dataTransfer.getData(type), value, `Expected dataTransfer on "${eventType}" event to have ${type} data`));
+          assert.equal(dataTransfer.dropEffect, spec.dropEffect, `Expected dataTransfer on "${eventType}" event to have dropEffect`);
+          assert.equal(dataTransfer.effectAllowed, spec.effectAllowed, `Expected dataTransfer on "${eventType}" event to have effectAllowed`);
+          KAssert.eqOptional(`Expected dataTransfer on "${eventType}" event to have dragImage`, Optional.from(spec.dragImage), DataTransfer.getDragImage(dataTransfer));
+
+          const dtFiles = dataTransfer.files;
```
spocke, line 74: Maybe split this out to an assertFiles function the whole thing is rather large?

spocke (approved):
Looks really nice. Good work on the research into specs and everything. :+1:

MitchCTiny (approved):

modules/agar/src/test/ts/browser/datatransfer/DataTransferItemTest.ts
mkzhx: Changed `dragster` to `agar`.

modules/dragster/src/main/ts/ephox/dragster/datatransfer/DataTransfer.ts
mkzhx: This is because it was set to read-only earlier for data copying purposes, so now we want to reset the mode of the original to what it was before https://github.com/tinymce/tinymce/blob/e2062fd1a8bef54a3d99e597ec5875fc586fe2b5/modules/dragster/src/main/ts/ephox/dragster/datatransfer/DataTransfer.ts#L98

modules/dragster/src/main/ts/ephox/dragster/datatransfer/DataTransfer.ts
mkzhx: Agreed, removed such comments.

modules/dragster/src/main/ts/ephox/dragster/datatransfer/Event.ts
mkzhx: Thanks for the suggestion, will keep this in mind for next time.

modules/dragster/src/main/ts/ephox/dragster/datatransfer/DataTransfer.ts
spocke: But shouldn't you always be able to read data regardless of what mode it's in? My memory on the spec if fuzzy.

modules/dragster/src/main/ts/ephox/dragster/datatransfer/DataTransfer.ts
```diff
@@ -0,0 +1,131 @@
...
+import { getEvent, isInDragStartEvent, setDragstartEvent, setEvent } from './Event';
+import { createEmptyFileList } from './Files';
+import { normalizeItems } from './Items';
+import { getMode, isInProtectedMode, isInReadWriteMode, setMode, setReadOnlyMode, setReadWriteMode } from './Mode';
+
+type DropEffect = DataTransfer['dropEffect'];
+type EffectAllowed = DataTransfer['effectAllowed'];
+
+const validDropEffects: DropEffect[] = [ 'none', 'copy', 'link', 'move' ];
+const validEffectAlloweds: EffectAllowed[] = [ 'none', 'copy', 'copyLink', 'copyMove', 'link', 'linkMove', 'move', 'all', 'uninitialized' ];
```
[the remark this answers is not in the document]
mkzhx: The property is called `effectAllowed` so the naming of the type follows that.

modules/agar/src/test/ts/browser/datatransfer/DataTransferItemTest.ts
```diff
@@ -1,42 +1,39 @@
...
+          resolve(undefined);
         } catch (e) {
-          die(e);
+          reject(e);
         }
       });
-    }))
-  ], success, failure);
+    });
+  });
```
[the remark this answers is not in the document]
mkzhx: I don't think `getAsString` can be awaited on, because even though it is asynchronous it doesn't return a Promise.

modules/dragster/src/main/ts/ephox/dragster/datatransfer/Items.ts
mkzhx: Good to know, also did it for `DataTransfer` and `DragImage` modules in dragster.

modules/dragster/src/test/ts/atomic/datatransfer/DataTransferCloneTest.ts
mkzhx: Good to know, made the dataTransfer tests browser tests.

modules/tinymce/src/core/test/ts/browser/DragDropOverridesTest.ts
mkzhx: Exposed DragImageData type from dragster to be used here.

modules/tinymce/src/core/test/ts/browser/DragDropOverridesTest.ts
mkzhx: I was originally intending to use the `item` method, which was the reason for the `for` loop. But since the `files` property is designed to be iterated anyway I can refactor it using `Arr.find`.

PR #8715, opened 2023-05-10
TINY-9813: Newlines should be added when pressing enter after a table.

HAFRMO:
Related Ticket: 

Description of Changes:
Added newlines where they should be added.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -33,6 +33,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Redial would in some situations cause select elements not to have an initial value selected when they should have. #TINY-9679
 - Table toolbar was visible even if the table was within a noneditable host element. #TINY-9664
 - Quickbar toolbars was shown for noneditable contents in a noneditable root. #TINY-9460
+- Adding a newline in Chrome when the cursor was placed behind a table could in some specific situations not generate the expected newline. #TINY-9813
 - Show the calculated height and width of media embed elements in the `media` plugin dialog. #TINY-8714
 - Removing an image that failed to upload from an empty paragraph would leave the paragraph without a padding br. #TINY-9696
 - Allow a media embed element to be correctly resized when using the `media` plugin dialog by converting the media embed to a standalone iframe. #TINY-8714
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -33,7 +33,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Redial would in some situations cause select elements not to have an initial value selected when they should have. #TINY-9679
 - Table toolbar was visible even if the table was within a noneditable host element. #TINY-9664
 - Quickbar toolbars was shown for noneditable contents in a noneditable root. #TINY-9460
-- Adding a newline in Chrome when the cursor was placed behind a table could in some specific situations not generate the expected newline. #TINY-9813
+- Adding a newline in Chrome when the cursor was placed after a table could in some specific situations not generate the expected newline. #TINY-9813
 - Show the calculated height and width of media embed elements in the `media` plugin dialog. #TINY-8714
 - Removing an image that failed to upload from an empty paragraph would leave the paragraph without a padding br. #TINY-9696
 - Allow a media embed element to be correctly resized when using the `media` plugin dialog by converting the media embed to a standalone iframe. #TINY-8714
--- a/modules/tinymce/src/core/main/ts/newline/InsertBlock.ts
+++ b/modules/tinymce/src/core/main/ts/newline/InsertBlock.ts
@@ -1,5 +1,5 @@
 import { Arr, Obj, Optional, Optionals, Type } from '@ephox/katamari';
-import { Css, PredicateFilter, SugarElement, SugarNode } from '@ephox/sugar';
+import { ContentEditable, Css, Insert, PredicateFilter, SugarElement, SugarNode, Traverse } from '@ephox/sugar';
 
 import DOMUtils from '../api/dom/DOMUtils';
 import DomTreeWalker from '../api/dom/TreeWalker';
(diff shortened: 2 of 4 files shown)


modules/tinymce/src/core/test/ts/browser/newline/InsertNewLineTest.ts
```diff
@@ -18,6 +19,21 @@ describe('browser.tinymce.core.newline.InsertNewLineTest', () => {
     InsertNewLine.insert(editor, args as EditorEvent<KeyboardEvent>);
   };
 
+  const setSelectionToBody = (editor: Editor, offset: number) => {
+    const sel = editor.selection.getSel();
+
+    if (Type.isNullable(sel)) {
+      throw new Error();
```
vpyshnenko, line 26: consider adding some error message

modules/tinymce/src/core/main/ts/newline/InsertBlock.ts
```diff
@@ -241,6 +241,14 @@ const insert = (editor: Editor, evt?: EditorEvent<KeyboardEvent>): void => {
   const rng = editor.selection.getRng();
   const newBlockName = Options.getForcedRootBlock(editor);
 
+  const isInRoot = rng.collapsed && rng.startContainer === editor.dom.getRoot();
```
TheSpyder, line 244: What about inside a `div` tag, rather than in the root? Does the bug not happen then?

modules/tinymce/src/core/main/ts/newline/InsertBlock.ts
```diff
@@ -435,14 +443,25 @@ const insert = (editor: Editor, evt?: EditorEvent<KeyboardEvent>): void => {
...
+    newBlock = createNewBlock(newBlockName);
+    child.fold(
+      () => {
+        Insert.append(start, SugarElement.fromDom(newBlock));
+      },
+      (child) => {
+        Insert.before(child, SugarElement.fromDom(newBlock));
+      }
+    );
+    editor.selection.setCursorLocation(newBlock, 0);
```
TheSpyder, line 463: What's this code doing that the code further down this `if` block doesn't already? Can we just adjust the order of the conditions somehow to achieve the same effect?
https://github.com/tinymce/tinymce/blob/873133a8529034434a414225dd2d88522d3bad8d/modules/tinymce/src/core/main/ts/newline/InsertBlock.ts#L452-L457

TheSpyder (approved):

modules/tinymce/src/core/main/ts/newline/InsertBlock.ts
HAFRMO: It does seem to occur there, I did not check for that. But I am wondering if the solution should not be slightly different for that as the div-block can be split using existing code whereas the body really can't be.

modules/tinymce/src/core/main/ts/newline/InsertBlock.ts
HAFRMO: It uses the newBlockName instead of going on default, which is important as otherwise it attempts to clone the next element when one exist, which is a CEF, giving the new element that same property.

modules/tinymce/CHANGELOG.md
```diff
@@ -33,6 +33,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Redial would in some situations cause select elements not to have an initial value selected when they should have. #TINY-9679
 - Table toolbar was visible even if the table was within a noneditable host element. #TINY-9664
 - Quickbar toolbars was shown for noneditable contents in a noneditable root. #TINY-9460
+- Adding a newline in Chrome when the cursor was placed behind a table could in some specific situations not generate the expected newline. #TINY-9813
```
vpyshnenko, line 36:
```suggestion
- Adding a newline in Chrome when the cursor was placed after a table could in some specific situations not generate the expected newline. #TINY-9813
```

modules/tinymce/src/core/test/ts/browser/newline/InsertNewLineTest.ts
```diff
@@ -18,6 +19,21 @@ describe('browser.tinymce.core.newline.InsertNewLineTest', () => {
     InsertNewLine.insert(editor, args as EditorEvent<KeyboardEvent>);
   };
 
+  const setSelectionToBody = (editor: Editor, offset: number) => {
```
spocke, line 22: This function is not really needed use `TinySelection.setCursor(editor, [], 1)` should do the same thing.

- [ ] Replace with mcagar function

modules/tinymce/src/core/main/ts/newline/InsertBlock.ts
```diff
@@ -241,6 +241,14 @@ const insert = (editor: Editor, evt?: EditorEvent<KeyboardEvent>): void => {
   const rng = editor.selection.getRng();
   const newBlockName = Options.getForcedRootBlock(editor);
 
+  const isInRoot = rng.collapsed && rng.startContainer === editor.dom.getRoot();
+  const start = SugarElement.fromDom(rng.startContainer);
+
+  const child = Traverse.child(start, rng.startOffset);
+  const isCefOpt = child.map((element) => SugarNode.isHTMLElement(element) && !ContentEditable.isEditable(element));
```
spocke, line 248: Use exists or forall those use a predicate and return a boolean true or false by default.

modules/tinymce/src/core/test/ts/browser/newline/InsertNewLineTest.ts
HAFRMO: The setCursor function does not do the same thing, no, I started with it and used this version only because there was no other choice. It moves the cursor up and away from the body in this situation, invalidating the whole test.

spocke: Then add a comment why it's not the same thing. The only difference I can think of is that Firefox renders a fake caret after that table.

vpyshnenko (approved):

modules/tinymce/src/core/main/ts/newline/InsertBlock.ts
TheSpyder: you're right. New ticket please.

PR #8734, opened 2023-05-16
TINY-9814: fix toolbar refresh with `toolbar_sticky`

lorenzo-pomili:
Related Ticket: TINY-9814

Description of Changes:
Having `toolbar_sticky` and `toolbar_sticky_offset` and `autoresize` (to have the editor scrolling), cause a render problem when you switch back from a view to the editor.

the cause is similar to the one in this [PR](https://github.com/tinymce/tinymce/pull/8543).

without the `toolbar_sticky_offset` when you switch back from a view the toolbar is not rendered so it works as intended, but with `toolbar_sticky_offset` the toolbar is rendered immediately so, since the action before refresh is to [show the editor](https://github.com/tinymce/tinymce/blob/develop/modules/tinymce/src/themes/silver/main/ts/ui/general/OuterContainer.ts#L206-L207), trying to refresh it immediately doesn't take the correct width.

However I see that removing entirely make this case works, but it break the one for the other [PR](https://github.com/tinymce/tinymce/pull/8543)

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -73,7 +73,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Typing after deleting formatted content could remove a space at the start of the typing. #TINY-9310
 - Invalid markup in Notification and Dialog close buttons. #TINY-9849
 - In dialogs, an incorrect `aria-describedby` attribute caused the body of the dialog to be announced when using a screen reader. #TINY-9816
-- Switching from a `view` to an `editor` with a `toolbar_sticky` and `toolbar_sticky_offset` and is scrolled, only the more button is showed. #TINY-9814
+- The sticky toolbar would not be rendered correctly when transitioning from the custom editor view to the main view. #TINY-9814
 
 ## 6.4.2 - 2023-04-26
 
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -73,6 +73,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Typing after deleting formatted content could remove a space at the start of the typing. #TINY-9310
 - Invalid markup in Notification and Dialog close buttons. #TINY-9849
 - In dialogs, an incorrect `aria-describedby` attribute caused the body of the dialog to be announced when using a screen reader. #TINY-9816
+- Switching from a `view` to an `editor` with a `toolbar_sticky` and `toolbar_sticky_offset` and is scrolled, only the more button is showed. #TINY-9814
 
 ## 6.4.2 - 2023-04-26
 
--- a/modules/tinymce/src/themes/silver/main/ts/modes/Iframe.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/modes/Iframe.ts
@@ -167,6 +167,7 @@ const render = async (editor: Editor, uiRefs: ReadyUiReferences, rawUiConfig: Re
       if (Type.isNull(OuterContainer.whichView(outerContainer))) {
         editor.focus();
         editor.nodeChanged();
+        OuterContainer.refreshToolbar(outerContainer);
       }
     }
   });
--- a/modules/tinymce/src/themes/silver/main/ts/ui/general/OuterContainer.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/ui/general/OuterContainer.ts
@@ -207,7 +207,9 @@ const factory: UiSketcher.CompositeSketchFactory<OuterContainerSketchDetail, Out
         Attribute.remove(element, 'aria-hidden');
       });
 
-      apis.refreshToolbar(comp);
+      setTimeout(() => {
+        apis.refreshToolbar(comp);
+      }, 0);
     }
   };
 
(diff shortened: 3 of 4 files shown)


modules/tinymce/src/themes/silver/test/ts/browser/view/ViewTest.ts
```diff
@@ -416,7 +416,8 @@ describe('browser.tinymce.themes.silver.view.ViewTest', () => {
       assert.equal(Html.get(editorContainer), expectedHtml);
     };
 
-    it('TINY-9419: "Expand or collapse" button should not be removed if the toolbar is opened and view is opened and close', () => {
+    // HERE
```
vpyshnenko, line 419: it looks like unnecessary comments

lorenzo-pomili: removed, thanks :)

modules/tinymce/CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Typing after deleting formatted content could remove a space at the start of the typing. #TINY-9310
+- Switching from a `view` to an `editor` with a `toolbar_sticky` and `toolbar_sticky_offset` and is scrolled, only the more button is showed. #TINY-9814
```
vpyshnenko, line 11:
```suggestion
- The sticky toolbar would not be rendered correctly when transitioning from the custom editor view to the main view. #TINY-9814
```

vpyshnenko (approved):

modules/tinymce/CHANGELOG.md

modules/tinymce/src/themes/silver/main/ts/ui/general/OuterContainer.ts
```diff
@@ -207,7 +207,9 @@ const factory: UiSketcher.CompositeSketchFactory<OuterContainerSketchDetail, Out
         Attribute.remove(element, 'aria-hidden');
       });
 
-      apis.refreshToolbar(comp);
+      setTimeout(() => {
```
spocke, line 210: I think it would be better if we some how call this refresh when the view is being updated. So in the ViewWrapper toggleView call somehow execute the outer container refresh. The toggleView is called from the OuterContainer module so it should have access to that component in there.

spocke: No wait miss reading this so is it we need to wait for the autoresize to happen then maybe change the Iframe code that waits for events then calls a refresh on the OuterContainer on various events but only on sliding and floating modes maybe it should do so for this mode as well.

lorenzo-pomili: ok, I maybe have found a different solution

spocke: I poked around with this couldn't find a better way to solve this I think the real bug is in DockingApis and that it updates and measures things while they are `display: none`. I tried adding a `component.element.dom.offsetParent` nul check in the refresh code in DockingApis and that seems to fix it as well that just ignores the update from happening if the element is or is within a `display: none`. This is wider than just views this will happen if scroll happens while the editor is hidden say in a tab view etc.

lorenzo-pomili: sadly add `component.element.dom.offsetParent` make a regression for [TINY-9419](https://github.com/tinymce/tinymce/blob/develop/modules/tinymce/src/themes/silver/test/ts/browser/view/ViewTest.ts#L419-L431), so I would stay to the current solution

modules/tinymce/src/themes/silver/test/ts/browser/view/ViewTest.ts
```diff
@@ -426,8 +426,80 @@ describe('browser.tinymce.themes.silver.view.ViewTest', () => {
...
+
+      await Waiter.pTryUntil('Wait for scroll top to be before the toolbar', () => Scroll.get().top < toolbar.dom.getBoundingClientRect().top);
+      elementToScrollTo.dom.scrollIntoView();
+      await Waiter.pTryUntil('Wait for scroll top to be after the toolbar', () => Scroll.get().top > toolbar.dom.getBoundingClientRect().top);
+
+      editor.execCommand('ToggleView', true, 'myview1');
+      assertViewHtml(0, '<button>myview1</button>');
+
+      // this is needed because otherwise the bug is not reproduced
+      await Waiter.pWait(0);
```
spocke, line 496: If I uncomment this and run the test it still passes is that expected?

modules/tinymce/src/themes/silver/main/ts/modes/Iframe.ts
```diff
@@ -167,6 +167,7 @@ const render = async (editor: Editor, uiRefs: ReadyUiReferences, rawUiConfig: Re
       if (Type.isNull(OuterContainer.whichView(outerContainer))) {
         editor.focus();
         editor.nodeChanged();
+        OuterContainer.refreshToolbar(outerContainer);
```
spocke, line 170: Even if I remove this the test still passes. Does it need to be a manual test it is really hard to automatically test?

lorenzo-pomili: ok, to make the test fail you need to comment this but also to revert the change in `OuterContainer` (put the `apis.refreshToolbar(comp);`), because the change in `OuterContainer` is needed to fix the case in the test:
```
TINY-9419: "Expand or collapse" button should not be removed if the toolbar is opened and view is opened and close
```
but this create the regression of this PR

spocke: Right makes sense I can make the test failing now thanks.

PR #8733, opened 2023-05-16
TINY-9829: Dispatch `insertFromPaste` input event upon pasting

mkzhx:
Related Ticket: TINY-9829

Description of Changes:
* Rename `FakeInputEvents` module to `InputEvents` and place it under `modules/tinymce/src/core/main/ts/events`
* Dispatch an `insertFromPaste` input event with relevant `data` property after a paste action (incl. image paste) occurs

Pre-checks:
* [x] Changelog entry added
* [x] Tests have been added (if applicable)
* [x] Branch prefixed with `feature/`, `hotfix/` or `spike/`

Review:
* [x] Milestone set
* [x] ~Docs ticket created (if applicable)~

GitHub issues (if applicable):

PR #8747, opened 2023-05-22
TINY-9829: Dispatch "insertFromPaste" `beforeinput` event with `dataTransfer` object on paste

mkzhx:
Related Ticket: TINY-9829

Description of Changes:
* [Previous PR](https://github.com/tinymce/tinymce/pull/8733) dealt with dispatching "insertFromPaste" `input` event on paste
* This PR dispatches an "insertFromPaste" `beforeinput` event with a dataTransfer property (as per native behaviour on Chrome) before dispatching that `input` event
* Refactor PasteEventUtils module

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/paste/Clipboard.ts
+++ b/modules/tinymce/src/core/main/ts/paste/Clipboard.ts
@@ -1,4 +1,4 @@
-import { DataTransfer } from '@ephox/dragster';
+import { DataTransfer, DataTransferContent } from '@ephox/dragster';
 import { Arr, Cell, Strings, Type } from '@ephox/katamari';
 
 import Editor from '../api/Editor';
@@ -36,7 +36,7 @@ const doPaste = (editor: Editor, content: string, internal: boolean, pasteAsText
     const content = args.content;
     SmartPaste.insertContent(editor, content, pasteAsText);
     const dataTransfer = DataTransfer.createDataTransfer();
-    dataTransfer.setData('text/html', content);
+    DataTransferContent.setHtmlData(dataTransfer, content);
     InputEvents.fireBeforeInputEvent(editor, 'insertFromPaste', { dataTransfer });
     InputEvents.fireInputEvent(editor, 'insertFromPaste');
   }
--- a/modules/tinymce/src/core/main/ts/paste/Clipboard.ts
+++ b/modules/tinymce/src/core/main/ts/paste/Clipboard.ts
@@ -1,3 +1,4 @@
+import { DataTransfer } from '@ephox/dragster';
 import { Arr, Cell, Strings, Type } from '@ephox/katamari';
 
 import Editor from '../api/Editor';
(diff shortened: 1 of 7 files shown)


spocke (approved):
Maybe also add a test that checks the order of the events so that it dispatches `paste`, `beforeinput`, `input` in that order.

mkzhx:
> Maybe also add a test that checks the order of the events so that it dispatches `paste`, `beforeinput`, `input` in that order.

@spocke Good idea, I added an assertion that checks the events are fired in order `paste` -> `pastepreprocess` -> `pastepostprocess` -> `beforeinput` -> `input`. 

Since the paste overrides (which fire the other events) happen in a `paste` event handler
https://github.com/tinymce/tinymce/blob/54a41e53ec2662939c9c5e4c0887b2ecab86d665/modules/tinymce/src/core/main/ts/paste/Clipboard.ts#L228
I had to change the callback that sets up the test handlers in `InternalClipboard.ts` to use `setup` option instead of `init_instance_callback` to ensure that the test handlers are run first before that handler with the paste overrides. Otherwise, it would seem that the `paste` event is fired after the other events when in reality it is fired first.

modules/tinymce/src/core/test/ts/browser/paste/InternalClipboardTest.ts
```diff
@@ -17,26 +17,29 @@ describe('browser.tinymce.core.paste.InternalClipboardTest', () => {
   const dataTransfer = Singleton.value<DataTransfer>();
   const lastPreProcessEvent = Singleton.value<EditorEvent<PastePreProcessEvent>>();
   const lastPostProcessEvent = Singleton.value<EditorEvent<PastePostProcessEvent>>();
+  const lastBeforeInputEvent = Singleton.value<EditorEvent<InputEvent>>();
   const lastInputEvent = Singleton.value<EditorEvent<InputEvent>>();
+  let eventTypes: string[] = [];
+  const setEventSingletonAndAddType = (singleton: Singleton.Value<EditorEvent<any>>, event: EditorEvent<any>) => {
```
spocke, line 23: Isn't the any here T since they should be the same. You should be able to feed a `EditorEvent<T>` to a `Singleton.Value<EditorEvent<U>>`.

mkzhx: Thanks for the suggestion! Will use generics over `any` where possible next time.

MitchCTiny (approved):

modules/tinymce/src/core/main/ts/paste/Clipboard.ts
```diff
@@ -34,6 +35,9 @@ const doPaste = (editor: Editor, content: string, internal: boolean, pasteAsText
   if (!args.cancelled) {
     const content = args.content;
     SmartPaste.insertContent(editor, content, pasteAsText);
+    const dataTransfer = DataTransfer.createDataTransfer();
+    DataTransferContent.setHtmlData(dataTransfer, content);
+    InputEvents.fireBeforeInputEvent(editor, 'insertFromPaste', { dataTransfer });
     InputEvents.fireInputEvent(editor, 'insertFromPaste');
```
TheSpyder, line 41: isn't the `beforeInput` event normally cancellable, which prevents the `input` event?

mkzhx: Nice observation, resolved in https://github.com/tinymce/tinymce/pull/8769

PR #8754, opened 2023-05-24
TINY-9887: Fixed formatter API regressions

spocke:
Related Ticket: TINY-9887

Description of Changes:
This is a follow up to TINY-9678 added the `selection.isEditable` check to the wrong function since they were very confusingly named `applyFormat` in the `ApplyFormat` module but `remove` in the `RemoveFormat` module but the `RemoveFormat` had a function that was called `removeFormat` but was not the opposite of the `applyFormat` function. So a refactored it now so that it makes more sense and is the same internal API. But also fixed a few regressions.

* Fixed regression with remove format being a noop if custom node was specified but selection was non editable.
* Fixed regression with apply format being a noop if custom node was specified but selection was non editable.
* Fixed regression with remove format not working in the first or last child of a CET see bug report not covering this though since it's not the real bug.
* Renamed the `RemoveFormat.remove` to `RemoveFormat.removeFormat` so that it's similar to `ApplyFormat.applyFormat`.
* Renamed the `RemoveFormat.removeFormat` to `RemoveFormat.removeNodeFormat` since it's node specific.
* Renamed the internal `removeFormatInternal` to `removeNodeFormatInternal`
* Named the internal `removeFormat` the `removeFormatInternal`

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/test/ts/browser/FormatterApplyTest.ts
+++ b/modules/tinymce/src/core/test/ts/browser/FormatterApplyTest.ts
@@ -1,7 +1,8 @@
 import { Assertions } from '@ephox/agar';
 import { describe, it } from '@ephox/bedrock-client';
 import { Obj } from '@ephox/katamari';
-import { LegacyUnit, TinyAssertions, TinyHooks, TinySelections, TinyState } from '@ephox/wrap-mcagar';
+import { Hierarchy } from '@ephox/sugar';
+import { LegacyUnit, TinyAssertions, TinyDom, TinyHooks, TinySelections, TinyState } from '@ephox/wrap-mcagar';
 import { assert } from 'chai';
 
 import Editor from 'tinymce/core/api/Editor';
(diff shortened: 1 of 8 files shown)


vpyshnenko:
Does this need to update the changelog?

vpyshnenko (approved):

spocke:
> Does this need to update the changelog?

No since it's a regression made by a fix that has not yet been released.

modules/tinymce/src/core/test/ts/browser/FormatterApplyTest.ts
```diff
@@ -2515,4 +2516,13 @@ describe('browser.tinymce.core.FormatterApplyTest', () => {
       TinyAssertions.assertContent(editor, initialContent);
     });
   });
+
+  it('TINY-9887: Should be not be noop if selection is not in an editable context but a custom editable node is specified', () => {
```
hamza0867, line 2520: I think you meant:
```suggestion
  it('TINY-9887: Should not be noop if selection is not in an editable context but a custom editable node is specified', () => {
```

hamza0867 (approved):

modules/tinymce/src/core/test/ts/browser/FormatterApplyTest.ts
spocke: Oops thanks for noticing that.

PR #8763, opened 2023-05-29
TINY-9896: Add option to run sequential and limit memory

jscasca:
Related Ticket: TINY-9896

Description of Changes:
* Add the option run parts of the build process sequentially to avoid running into memory issues. Running only the `oxide` parts sequentially should not impact the build time. Running `eslint` sequentially with `ci` will increase the build time by around 40 seconds. Running all the `oxide` process in parallel does not appear to decrease build time either

Pre-checks:
* [ ] Changelog entry added

Review:
* [ ] Milestone set


spocke (approved):

mkzhx (approved):

TheSpyder:
ah yes this makes sense, because `eslint` takes so long it dominates all other tasks and they can just run sequentially.

To be honest, at this point we could move `tinymce-rollup` to `ci-seq` and it would probably lower the overall build time. Having the bundle task wait for linting to finish probably doesn't make sense with how long linting takes now.

PR #8776, opened 2023-06-01
TINY-9920: Add test to check fallback Keyboard Navigation html matches English html

mkzhx:
Related Ticket: TINY-9920

Description of Changes:
* Related to TINY-9633 PR https://github.com/tinymce/tinymce/pull/8764
* Update fallback html string to match the string loaded from en.html exactly (make newlines and spacing consistent)
* Add test to check fallback html string matches string loaded from en.html exactly

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/plugins/help/main/ts/ui/KeyboardNavTabI18n.ts
+++ b/modules/tinymce/src/plugins/help/main/ts/ui/KeyboardNavTabI18n.ts
@@ -9,13 +9,13 @@ const fallback = `<h1>Begin keyboard navigation</h1>
 <dl>
   <dt>Focus the Menu bar</dt>
   <dd>Windows or Linux: Alt+F9</dd>
-  <dd>macOS: ⌥F9</dd>
+  <dd>macOS: &#x2325;F9</dd>
   <dt>Focus the Toolbar</dt>
   <dd>Windows or Linux: Alt+F10</dd>
-  <dd>macOS: ⌥F10</dd>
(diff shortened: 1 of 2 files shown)


TheSpyder (commented):
I posted this in the earlier PR but apparently github sucks on mobile and it left my comment as pending when I submitted my approval.

I don’t think we need the fallback 🤔 yes it can technically fail but much like the emoji database on the emoticons plugin it’s a critical piece and the dialog just won’t work without it.

I would rather not have the string in the codebase twice, even with tests to verify, apologies for needing to do all this work unnecessarily

mkzhx:
> I posted this in the earlier PR but apparently github sucks on mobile and it left my comment as pending when I submitted my approval.
> 
> I don’t think we need the fallback 🤔 yes it can technically fail but much like the emoji database on the emoticons plugin it’s a critical piece and the dialog just won’t work without it.
> 
> I would rather not have the string in the codebase twice, even with tests to verify, apologies for needing to do all this work unnecessarily

@TheSpyder We would need the fallback for the situation where an invalid language code is specified in the `language` option, since we rely on a valid language code to retrieve the corresponding HTML file.

TheSpyder:
I didn't think of that, but I like the idea of hard-coding `en.html` in the plugin as the fallback.

mkzhx:
Closed to implement loading from `en.html` as a fallback in TINY-9928.

PR #8780, opened 2023-06-01
TINY-9884: Fix pressing BACKSPACE in details body and summary

yacodes:
Related Ticket: TINY-9884

Description of Changes:
* Fixed pressing BACKSPACE in details body and summary.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -81,7 +81,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Saving the Table Properties dialog after changing properties unrelated to cells would overwrite cell properties set by the Cell Properties dialog. #TINY-9837
 - Fixed the constrained bounds calculation for dismissal of toolbar when using `toolbar_location: 'bottom'`. #TINY-9718
 - The floating toolbar did not occupy the entire available space when the page had a flexbox layout. #TINY-9847
-- Pressing BACKSPACE within accordion's body and summary broke the markup structure. #TINY-9884
 
 ## 6.4.2 - 2023-04-26
 
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -81,6 +81,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Saving the Table Properties dialog after changing properties unrelated to cells would overwrite cell properties set by the Cell Properties dialog. #TINY-9837
 - Fixed the constrained bounds calculation for dismissal of toolbar when using `toolbar_location: 'bottom'`. #TINY-9718
 - The floating toolbar did not occupy the entire available space when the page had a flexbox layout. #TINY-9847
+- Fixed backspace and delete keypresses within details elements. #TINY-9884
 
 ## 6.4.2 - 2023-04-26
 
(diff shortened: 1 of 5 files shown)


modules/tinymce/CHANGELOG.md
```diff
@@ -80,6 +80,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The sticky toolbar would not be rendered correctly when transitioning from the custom editor view to the main view. #TINY-9814
 - Saving the Table Properties dialog after changing properties unrelated to cells would overwrite cell properties set by the Cell Properties dialog. #TINY-9837
 - Fixed the constrained bounds calculation for dismissal of toolbar when using `toolbar_location: 'bottom'`. #TINY-9718
+- Pressing BACKSPACE within accordion's body and summary broke the markup structure. #TINY-9884
```
spocke, line 83: This has not yet been released so no need for a changelog item.

modules/tinymce/src/plugins/accordion/main/ts/core/Keyboard.ts
```diff
@@ -25,11 +25,14 @@ const setup = (editor: Editor): void => {
   });
 
   editor.on('keydown', (e) => {
-    if (e.keyCode === VK.BACKSPACE) {
+    if (e.keyCode === VK.BACKSPACE || e.keyCode === VK.DELETE) {
```
spocke, line 28: This breaks Ctrl-Delete or Ctrl-Backspace so you can't delete words inside a details or summary element.

spocke: The only time you need to run this custom code is when you are in the intersection between summary and the details body so I guess you could try to detect that and only execute the delete when that happens since when you hit ctrl-backspace in say `<summary>foo</summary><p>|bar</p>` then the delete acts like a regular delete since the block boundary is being treated as a word boundary. You could make some form of edge function but all that could be handled with the CaretFinder stuff but that is in the core.

spocke (changes requested):

modules/tinymce/src/plugins/accordion/main/ts/core/Keyboard.ts
yacodes: I've moved the BACKSPACE and DELETE handlers to the core (`selection/DetailsElement.ts`). Ended up using `CaretFinder` and a couple of `e.preventDefault()` methods. Let me know what you think, please!

spocke: Delete doesn't seem to work in the branch and the backspace still merges things into the summary. I think for checking if the caret is in the first or last position you could do:
```
// Is caret at beginning of summary
CaretFinder.firstPositionIn(summary).exists((pos) => pos.isEqual(CaretPosition.fromRangeStart(editor.selection.getRng())));

// Is caret at end of summary
CaretFinder.lastPostionIn(summary).exists((pos) => pos.isEqual(CaretPosition.fromRangeStart(editor.selection.getRng())));

// Is caret just before summary in details element
CaretFinder.lastPostionIn(summary).exists((pos) => pos.isEqual(CaretPostion.prevPosition(editor.selection.getRng()).exists((next) => next.isEqual(pos))));
```
The CaretPosition `isAtStart` will just check if the position is the first in the current element or text node not the first in the current block scope.

yacodes: Weirdly, but everything worked in FF. I can reproduce the problem in Chrome though. Fixed the code with more peculiar `CaretFinder` checks.

spocke: Ok seems to work now in Chrome nice. :+1:

vpyshnenko (approved):

PR #8799, opened 2023-06-06
TINY-6888: now delete `li` in particular situation doesn't cause error

lorenzo-pomili:
Related Ticket: TINY-6888

Description of Changes:
The problem was that when we were in a `li` and the parent of the new caret container is the same `li` we remove the `li` and then we try to use `ToggleList.mergeWithAdjacentLists` we use it on `otherLi.parentNode`, but at that point we already removed `otherLi`.

To solve this now the `otherLi.parentNode` reference is stored in a `cost` before the remove

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -12,7 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Returning an empty string in custom context menu update function would result in a small white line appearing on right click and the native browser context menu would not show up. #TINY-9842
 - Tab navigation incorrectly stopped around `iframe` dialog component. #TINY-9815
-- `backspaceDeleteIntoListCaret` try to use a pointer to a removed element. #TINY-6888
+- deleting `li` with only `br`s in it sometimes caused a crush. #TINY-6888
 
 ## 6.5.0 - 2023-06-12
 
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Returning an empty string in custom context menu update function would result in a small white line appearing on right click and the native browser context menu would not show up. #TINY-9842
 - Tab navigation incorrectly stopped around `iframe` dialog component. #TINY-9815
+- `backspaceDeleteIntoListCaret` try to use a pointer to a removed element. #TINY-6888
 
 ## 6.5.0 - 2023-06-12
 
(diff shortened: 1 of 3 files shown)


vpyshnenko (approved):

modules/tinymce/src/plugins/lists/test/ts/browser/BackspaceDeleteFromBlockIntoLiTest.ts
```diff
@@ -126,4 +126,12 @@ describe('Browser Test: .RemoveTrailingBlockquoteTest', () => {
     TinyContentActions.keystroke(editor, Keys.backspace());
     TinyAssertions.assertContent(editor, '<ul><li>a</li></ul>');
   });
+
+  it('TINY-6888: delete a `li` with a `br` and a `br` with `data-mce-bogus', () => {
```
spocke, line 130: Could you add some tests with non bogus BRs as well so that we don't break `<li>foo<br><br>` or `<li>foo<br><br><br>` cases? Or any other cases you can think of.

spocke (approved):

modules/tinymce/src/plugins/lists/test/ts/browser/BackspaceDeleteFromBlockIntoLiTest.ts
lorenzo-pomili: added

modules/tinymce/CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - The pattern replacement removes spaces if they were contained in a tag that only contains a space and the text to replace. #TINY-9744
 
+### Fixed
+- `backspaceDeleteIntoListCaret` try to use a pointer to a removed element. #TINY-6888
```
spocke, line 13: This is an internal function describe the error instead that it threw an exception when trying to delete the last BR element in a list element or similar. Think of the reader here they won't know or care where the fix was made more what the fix was for.

- [ ] Update changelog to reflect the change from the user perspective

lorenzo-pomili: @spocke done, let me know if it is ok

modules/tinymce/CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Returning an empty string in custom context menu update function would result in a small white line appearing on right click and the native browser context menu would not show up. #TINY-9842
 - Tab navigation incorrectly stopped around `iframe` dialog component. #TINY-9815
+- deleting `li` with only `br`s in it sometimes caused a crush. #TINY-6888
```
TheSpyder, line 15: crash? 😅

PR #8808, opened 2023-06-07
TINY-9815: Skip fake tab stops from tab navigation flow

vpyshnenko:
Related Ticket: TINY-9815

Description of Changes:
* There are two fake tab stop divs wrapping the iframe element. When the focus is on the component before the iframe the Tab press successfully skips the first fake tab stop div and moves focus to the iframe. However next Tab press does not skip the second tab stop div because there is only keyup event fired on it, as such keydown Tab key rules do not work. The solution is to bind key rules to keyup event but check that focus is currently on tab stop div to avoid duplicated tab handling.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/alloy/src/main/ts/ephox/alloy/keying/TabbingTypes.ts
+++ b/modules/alloy/src/main/ts/ephox/alloy/keying/TabbingTypes.ts
@@ -9,7 +9,7 @@ import { NativeSimulatedEvent } from '../events/SimulatedEvent';
 import * as ArrNavigation from '../navigation/ArrNavigation';
 import * as KeyMatch from '../navigation/KeyMatch';
 import * as KeyRules from '../navigation/KeyRules';
-import { KeyRuleHandler, TabbingConfig } from './KeyingModeTypes';
+import { KeyRuleStatelessHandler, TabbingConfig } from './KeyingModeTypes';
 import * as KeyingType from './KeyingType';
 
 const create = (cyclicField: FieldProcessor): KeyingType.KeyingType<TabbingConfig, Stateless> => {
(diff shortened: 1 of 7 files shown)


modules/alloy/src/main/ts/ephox/alloy/keying/TabbingTypes.ts
```diff
@@ -116,7 +125,9 @@ const create = (cyclicField: FieldProcessor): KeyingType.KeyingType<TabbingConfi
   ]);
 
   const getKeyupRules = Fun.constant([
-    KeyRules.rule(KeyMatch.inSet(Keys.ESCAPE), exit)
+    KeyRules.rule(KeyMatch.inSet(Keys.ESCAPE), exit),
+    KeyRules.rule(KeyMatch.inSet(Keys.TAB), goFromPseudoTabstop),
+
```
lorenzo-pomili, line 130: this line maybe should be removed

lorenzo-pomili (approved):

modules/alloy/src/main/ts/ephox/alloy/keying/TabbingTypes.ts
```diff
@@ -93,20 +93,29 @@ const create = (cyclicField: FieldProcessor): KeyingType.KeyingType<TabbingConfi
...
     const navigate = tabbingConfig.cyclic ? ArrNavigation.cycleNext : ArrNavigation.tryNext;
     return go(component, simulatedEvent, tabbingConfig, navigate);
   };
 
-  const execute: KeyRuleHandler<TabbingConfig, Stateless> = (component, simulatedEvent, tabbingConfig) =>
+  const isFirstChild = (node: Node): boolean => node.parentNode?.firstChild === node;
+
+  const goFromPseudoTabstop: KeyRuleStatelessHandler<TabbingConfig> = (component, simulatedEvent, tabbingConfig) =>
+    findCurrent(component, tabbingConfig).filter((elem) => !tabbingConfig.useTabstopAt(elem))
+      .fold(
```
spocke, line 110: Isn't this a `bind` since the none case is none and the some case is some or none?

PR #8815, opened 2023-06-09
TINY-9776: add the possibility to paste images with `mceInsertClipboardContent`

lorenzo-pomili:
Related Ticket: TINY-9776

Description of Changes:
this is needed to test the fix in the plugin

P.S. do we need changelog's entry for this?

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - The pattern replacement removes spaces if they were contained in a tag that only contains a space and the text to replace. #TINY-9744
-- Now `mceInsertClipboardContent` also manage image pasting. #TINY-9776
+- The `mceInsertClipboardContent` command will now also accept a `files` property to insert images. #TINY-9776
 
 ## 6.5.0 - 2023-06-12
 
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - The pattern replacement removes spaces if they were contained in a tag that only contains a space and the text to replace. #TINY-9744
+- Now `mceInsertClipboardContent` also manage image pasting. #TINY-9776
 
 ## 6.5.0 - 2023-06-12
 
--- a/modules/tinymce/src/core/main/ts/paste/Clipboard.ts
+++ b/modules/tinymce/src/core/main/ts/paste/Clipboard.ts
@@ -142,17 +142,20 @@ const isImage = (editor: Editor) => {
   });
 };
 
-const getImagesFromDataTransfer = (editor: Editor, dataTransfer: DataTransfer): File[] => {
+const getFilesFromDataTransfer = (editor: Editor, dataTransfer: DataTransfer): File[] => {
   const items = dataTransfer.items ? Arr.bind(Arr.from(dataTransfer.items), (item) => {
     return item.kind === 'file' ? [ item.getAsFile() as File ] : [];
   }) : [];
   const files = dataTransfer.files ? Arr.from(dataTransfer.files) : [];
-  return Arr.filter(items.length > 0 ? items : files, isImage(editor));
+  return items.length > 0 ? items : files;
 };
+
+const getImagesFromFiles = (editor: Editor, files: File[]): File[] =>
+  Arr.filter(files, isImage(editor));
 
 const pasteImageFiles = (editor: Editor, files: File[], rng: Range | undefined): boolean => {
   if (Options.shouldPasteDataImages(editor) && files.length > 0) {
-    readFilesAsDataUris(files).then((fileResults) => {
+    readFilesAsDataUris(getImagesFromFiles(editor, files)).then((fileResults) => {
       if (rng) {
         editor.selection.setRng(rng);
       }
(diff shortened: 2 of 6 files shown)


modules/tinymce/src/core/main/ts/paste/Commands.ts
```diff
@@ -28,6 +28,10 @@ const register = (editor: Editor, pasteFormat: Cell<string>): void => {
     if (value.text) {
       Clipboard.pasteText(editor, value.text);
     }
+
+    if (value.imageEvent) {
```
spocke, line 32: I think this could be refactored to be `value.files` if you separate the `Clipboard.pasteImageData` into `Clipboard.pasteFiles` and `Clipboard.pasteImageDataFromEvent`. Then have the `pasteFiles` take an array of files instead that goes though an image filter as it does when extracting files from the clipboard dataTransfer instances. The the `pasteImageDataFromEvent` would just check what event it is grab the dataTransfer, preventDefault and call the `pasteFiles` function. It would be a bit strange to have a command accepting an event and preventing that event. Commands should be executable outside of events.

Also this needs tests and a changelog since it's an API change. There are multiple tests for this command.

spocke (changes requested):

modules/tinymce/src/core/main/ts/paste/Commands.ts
spocke: Maybe I wasn't good at explaining it I don't think we should make the `mceInsertClipboardContent` command event specific. Instead with a bit of refactoring it could take `html` `text` or `files` since we want to be able to retain file names it needs to be files not blobs. If it accepts an array like of files. Then inside the Clipboard module we make a new function that accepts files instead of the events.

So then you could do this:

```
editor.on('paste', (e) => {
  e.preventDefault();
  editor.execCommand('mceInsertClipboardContent', { files: e.clipboardData.files } );
});
```

Or:
```
editor.on('drop', (e) => {
  e.preventDefault();
  editor.execCommand('mceInsertClipboardContent', { files: e.dataTransfer.files } );
});
```

Or for testing:
editor.execCommand('mceInsertClipboardContent', { files: [
  new File([new Blob([], { type: 'text/html' })], 'test.html')
] } );

lorenzo-pomili: refactored

modules/tinymce/CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - The pattern replacement removes spaces if they were contained in a tag that only contains a space and the text to replace. #TINY-9744
+- Now `mceInsertClipboardContent` also manage image pasting. #TINY-9776
```
spocke, line 11: Mayeb change it to:

```suggestion
- The `mceInsertClipboardContent` command will now also accept a `files` property to insert images. #TINY-9776
```

modules/tinymce/src/core/main/ts/paste/Clipboard.ts
```diff
@@ -150,6 +150,24 @@ const getImagesFromDataTransfer = (editor: Editor, dataTransfer: DataTransfer):
   return Arr.filter(items.length > 0 ? items : files, isImage(editor));
 };
 
+const pasteImageFiles = (editor: Editor, files: File[], rng: Range | undefined): boolean => {
```
spocke, line 153: I think more parts of `pasteImageData` needs to be moved into this function to do what the function name says. The `pasteImageData` will first grab the images from `getImagesFromDataTransfer` so it filters out images only. This function accepts any file type it should only take the images out so the `getImagesFromDataTransfer` should be split up into so that you can filter files here and in the `pasteImageData`. Also the boolean that this returns could be used to determine if the event should be prevented in `pasteImageData`. So some more refactoring is required here.

I would do it like this more or less:
* So I would do change the `getImagesFromDataTransfer` to `getFilesFromDataTransfer`
* Create a new function `getImagesFromFiles` that filters a collection of files based on isImage predicate
* Change `pasteImageFiles` to take files then run the `getImagesFromFiles`
* Then change the `pasteImageData` to prevent default if the `pasteImageFiles` returns true since then the clipboard has image data in it.

spocke (changes requested):

modules/tinymce/src/core/test/ts/browser/paste/PasteTest.ts
```diff
@@ -343,6 +364,22 @@ describe('browser.tinymce.core.paste.PasteTest', () => {
...
 
+  it('TINY-9776: it should be possible to paste an image via mceInsertClipboardContent', async () => {
+    const editor = hook.editor();
+    editor.setContent('<p>text</p>');
+    TinySelections.setSelection(editor, [ 0, 0 ], 1, [ 0, 0 ], 2);
+
+    TinyAssertions.assertContentPresence(editor, { img: 0 });
+
+    editor.execCommand('mceInsertClipboardContent', false, {
+      files: [ base64ToBlob(base64ImgSrc, 'image/gif', 'image.gif') ]
```
spocke, line 375: Also feed it more than one image and also other mime types. It should not try to process those.

modules/tinymce/src/core/main/ts/paste/Clipboard.ts
lorenzo-pomili: refactored

modules/tinymce/CHANGELOG.md

modules/tinymce/src/core/test/ts/browser/paste/PasteTest.ts
lorenzo-pomili: added

spocke (approved):

HAFRMO (approved):

PR #8827, opened 2023-06-15
TINY-9872: List now better detect a suitable parent.

HAFRMO:
Related Ticket: 

Description of Changes:
It would previous detect what will be the list item as the stopper, preventing the list from being longer than one.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -16,7 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Returning an empty string in custom context menu update function would result in a small white line appearing on right click and the native browser context menu would not show up. #TINY-9842
-- When making lists and selecting divs the lists would in some situations only be partially applied. #TINY-9872
+- Creating a list from multiple div elements would only create a partial list. #TINY-9872
 - Tab navigation incorrectly stopped around `iframe` dialog component. #TINY-9815
 - It was possible to delete the sole empty block just before a details element if it was nested within another details element. #TINY-9965
 - deleting `li` with only `br`s in it sometimes caused a crush. #TINY-6888
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - The pattern replacement removes spaces if they were contained in a tag that only contains a space and the text to replace. #TINY-9744
 
 ### Fixed
+- When making lists and selecting divs the lists would in some situations only be partially applied. #TINY-9872
 - Tab navigation incorrectly stopped around `iframe` dialog component. #TINY-9815
 
 ## 6.5.0 - 2023-06-12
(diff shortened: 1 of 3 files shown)


modules/tinymce/src/plugins/lists/main/ts/actions/ToggleList.ts
```diff
@@ -191,10 +191,19 @@ const hasCompatibleStyle = (dom: DOMUtils, sib: Element, detail: ListDetail): bo
   return sibStyle === detailStyle;
 };
 
+const getRootSearchStart = (editor: Editor, range: Range): Node => {
```
spocke, line 194: This seems to not work for cases with `<div><b>f[oo</b></div><div><b>bar</b></div><div><b>b]az</b></div>` so it should work with formatted nodes as well not just plain text nodes.

modules/tinymce/src/plugins/lists/test/ts/browser/ApplyToDivTest.ts
```diff
@@ -0,0 +1,36 @@
...
+    editor.setContent('<div>Line 1</div><div>Line 2</div><div>Line 3</div>');
+
+    TinySelections.setSelection(editor, [ 0, 0 ], 2, [ 2, 0 ], 2);
+    editor.execCommand('InsertUnorderedList');
+
+    TinyAssertions.assertContent(editor, '<ul>\n<li>Line 1</li>\n<li>Line 2</li>\n<li>Line 3</li>\n</ul>');
+    TinyAssertions.assertSelection(editor, [ 0, 0, 0 ], 2, [ 0, 2, 0 ], 2);
+  });
+
+  it('TINY-9872: Apply to divs in forced_root_block:div, but not when widely separated', () => {
```
spocke, line 26: The `forced_root_block` option doesn't seem to have an affect to this issue? Should you specifically check for that?

HAFRMO: Well, true, the forced div isn't actually the cause of the issue, that is a poorly written name on my part. My bad.

modules/tinymce/src/plugins/lists/main/ts/actions/ToggleList.ts
HAFRMO: Checking for forced_root_block is not a good idea, since it only acts as an easy trigger, but isn't actually the cause of the issue. The cause is the divs themselves, whether they  come from force_root_block, setContent, or anything else. Just checking for the body is also not a good idea, since this issue can occur in other places too, such as nested divs. It'd be easier, as you point out, but I don't think it'd be a good solution. I'll look into our code for detecting the block to move into an li and use that for how to step above it.

vpyshnenko (approved):

modules/tinymce/CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Returning an empty string in custom context menu update function would result in a small white line appearing on right click and the native browser context menu would not show up. #TINY-9842
+- When making lists and selecting divs the lists would in some situations only be partially applied. #TINY-9872
```
TheSpyder, line 19:
```suggestion
- Creating a list from multiple div elements would only create a partial list. #TINY-9872
```

TheSpyder (approved):

PR #8829, opened 2023-06-15
TINY-9997: Restore behavior that converts image url to image tag on paste

mkzhx:
Related Ticket: TINY-9997

Description of Changes:
* Restore behavior that converts image url to image tag on paste while still allowing firing of simulated input events when default is prevented, as introduced in https://github.com/tinymce/tinymce/pull/8733

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/paste/Commands.ts
+++ b/modules/tinymce/src/core/main/ts/paste/Commands.ts
@@ -22,6 +22,8 @@ const register = (editor: Editor, pasteFormat: Cell<string>): void => {
 
   editor.addCommand('mceInsertClipboardContent', (ui, value) => {
     if (value.html) {
+      // TINY-9997: Input events are not simulated when using paste commands, similar to how the 'mceInsertContent'
+      // and 'Delete' commands work.
       Clipboard.pasteHtml(editor, value.html, value.internal, false);
     }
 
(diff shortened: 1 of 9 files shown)


TheSpyder (approved):

tiny-ben-tran (approved):

modules/tinymce/src/core/main/ts/paste/Commands.ts
```diff
@@ -22,11 +22,11 @@ const register = (editor: Editor, pasteFormat: Cell<string>): void => {
 
   editor.addCommand('mceInsertClipboardContent', (ui, value) => {
     if (value.html) {
-      Clipboard.pasteHtml(editor, value.html, value.internal);
+      Clipboard.pasteHtml(editor, value.html, value.internal, false);
```
ltrouton, line 25: I can't quite wrap my head around why for the commands, the simulated events argument is false.   Probably missing something obvious so just thought I would check

MitchCTiny (approved):

modules/tinymce/src/core/main/ts/paste/Commands.ts
mkzhx: @ltrouton I remember having a conversation with @spocke, where we noticed that while commands such as `mceInsertContent` and `Delete` do not fire any `input` events, we worked on getting the events to fire on keyboard actions. The conclusion was that we specifically only simulate these native `input` events if the user is doing the actions, not when API calls are made.

ltrouton: Right, that makes sense. Do you think it would it be worth adding a comment to clarify that?

mkzhx: Added 👍

PR #8854, opened 2023-06-24
TINY-9987: Spike how to add validation to input field

hamza0867:
Related Ticket: TINY-9987

Description of Changes:
* Adds form validation to textfields

Pre-checks:


TheSpyder (commented):
The bridge API to set the validator seems fine, although it's not clear what the `validateOnLoad` is for (it's never used).

The implementation however is definitely wrong. It's not entirely clear to me where in the Alloy/Silver boundary this needs to sit, but digging all over the Alloy rendered dom from within the Silver dialog is definitely wrong.

TheSpyder:
I think there will be follow-up tickets?

PR #8896, opened 2023-07-11
TINY-10032: Round scrollTop to prevent isScrollAtBottom being unexpectedly false

mkzhx:
Related Ticket: TINY-10032

Description of Changes:
* Round scrollTop to prevent isScrollAtBottom being unexpectedly false

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/themes/silver/main/ts/ui/dialog/IFrame.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/ui/dialog/IFrame.ts
@@ -33,7 +33,7 @@ const getDynamicSource = (initialData: Optional<string>, stream: boolean): IFram
           Optional.from(iframe.contentDocument).fold(
             setSrcdocValue,
             (doc) => {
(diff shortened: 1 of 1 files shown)


TheSpyder (approved):
I wonder if there are edge cases where `ceil` is needed instead of `round` 🤔

ltrouton (approved):

mkzhx:
> I wonder if there are edge cases where `ceil` is needed instead of `round` 🤔

@TheSpyder Yeah looking at this StackOverflow answer https://stackoverflow.com/a/71076218 `Math.round` doesn't seem to handle all edge cases. 

MDN recommends using Math.abs and a threshold https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollHeight#determine_if_an_element_has_been_totally_scrolled, but I think `Math.ceil` is safe enough without an arbitrary threshold.

tiny-ben-tran (approved):

PR #8898, opened 2023-07-11
TINY-9863: improved newline break detection in divs before a cef

HAFRMO:
Related Ticket: 

Description of Changes:
Let the newline code merge with the previously added work. I initially thought that since the code was very eager to split the parent div, that's what we should do ( and we didn't do it on the previous issue because the body shouldn't be split ). Discussion about it made us agree that even when in divs they should not be split, so we can use the same code.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


TheSpyder (approved):
nice!

lorenzo-pomili (approved):

PR #8901, opened 2023-07-12
TINY-10003: add refresh for `Toolbar`

lorenzo-pomili:
Related Ticket: TINY-10003

Description of Changes:
Here the problem was that switching from a view to the editor calls [this refresh](https://github.com/tinymce/tinymce/blob/develop/modules/tinymce/src/themes/silver/main/ts/ui/general/OuterContainer.ts#L132) , but in this case we were using [alloy/api/ui/Toolbar](https://github.com/tinymce/tinymce/blob/develop/modules/alloy/src/main/ts/ephox/alloy/api/ui/Toolbar.ts#L45) that didn't have the `refresh` API, and this cause the error.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/alloy/CHANGELOG.md
+++ b/modules/alloy/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
 
+### Fixed
+- Going back from a view to the editor in mobile caused an error. #TINY-10003
+
 ## 13.0.0 - 2023-07-12
 
 ### Added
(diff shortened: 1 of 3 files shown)


HAFRMO (commented):
Should there be some sort of test for this?

TheSpyder (approved):
Weird that we haven't come across this in the last 3 years. I'm curious about why it wasn't there, but it seems harmless to have either way.

vpyshnenko (approved):

HAFRMO (approved):

PR #8926, opened 2023-07-24
TINY-10078, TINY-10097, TINY-10109, TINY-10110: Maintain non-end scroll positions, reduce flickering, and fix inconsistent autoscrolling for streamContent: true iframe dialog components

mkzhx:
Related Ticket: TINY-10078, TINY-10097, TINY-10109, TINY-10110

Description of Changes:

TINY-10078
* Manually maintain non-end scroll positions on Firefox and Safari

TINY-10097
* Throttle iframe content updates to 500ms intervals on Safari to reduce visual impact of flickering (document.write() flickers on Safari but not Chrome or Firefox)

TINY-10109
* Fix Safari not consistently scrolling to the bottom by waiting for iframe load

TINY-10110
* Dynamically determine the scrolling element of an iframe based on content doctype declaration and other factors

Pre-checks:
* [x] Changelog entry added
* [x] Tests have been added (if applicable)
* [x] Branch prefixed with `feature/`, `hotfix/` or `spike/`

Review:
* [x] Milestone set
* [x] ~Docs ticket created (if applicable)~

GitHub issues (if applicable):

PR #8932, opened 2023-07-26
TINY-10078: Throttle Firefox update interval to 200ms to improve freescrolling

mkzhx:
Related Ticket: TINY-10078

Description of Changes:
* Follow-up to https://github.com/tinymce/tinymce/pull/8926
* Since we manually maintain non-end scroll positions on Firefox and Safari on each update, having rapid updates would cause freescrolling to feel stuck. Safari is already throttled to 500ms as part of TINY-10097, so here we throttle Firefox to 200ms to improve freescrolling.

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/themes/silver/main/ts/ui/dialog/IFrame.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/ui/dialog/IFrame.ts
@@ -18,7 +18,7 @@ type IframeSpec = Omit<Dialog.Iframe, 'type'>;
 
 const browser = PlatformDetection.detect().browser;
 const isSafari = browser.isSafari();
-const isFirefox = browser.isFirefox();
+const isSafariOrFirefox = isSafari || browser.isFirefox();
 const isChromium = browser.isChromium();
 
 const getDynamicSource = (initialData: Optional<string>, stream: boolean): IFrameSourcing => {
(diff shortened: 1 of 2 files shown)


modules/tinymce/src/themes/silver/main/ts/ui/dialog/IFrame.ts
```diff
@@ -81,8 +81,11 @@ const getDynamicSource = (initialData: Optional<string>, stream: boolean): IFram
       });
   };
 
-  // TINY-10097: Throttle to reduce flickering, as the document.write() method still observes significant flickering on Safari.
-  const writeValueThrottler = isSafari ? Optional.some(Throttler.first(writeValue, 500)) : Optional.none();
+  // TINY-10078: On Firefox, throttle to 200ms allow improve scrolling experience. Since we are manually maintaining previous scroll
+  // position on each update, when updating too rapidly, attempting to scroll around the iframe can feel stuck.
+  // TINY-10097: On Safari, throttle to 500ms reduce flickering as the document.write() method still observes significant flickering.
+  // Also improves scrolling, as scroll positions are maintained manually similar to Firefox.
+  const writeValueThrottler = isSafariOrFirefox ? Optional.some(Throttler.first(writeValue, isSafari ? 500 : 200)) : Optional.none();
```
TheSpyder, line 88: Looking at this in context I'd like to simplify the code a bit.

For starters, perhaps `isSafariOrFirefox` could be an optional number. Then the `500` and `200` values are defined along with other impactful constants at the top, and this becomes a much nicer `Optional.map`.

Ideally I'd like to avoid `writeValueThrottler.fold` below, but I'm not sure how we do that while retaining the single `Throttler` instance.

TheSpyder (approved):

modules/tinymce/src/themes/silver/main/ts/ui/dialog/IFrame.ts
mkzhx: That's a great suggestion, thanks 👍

TheSpyder: I actually meant replacing the `isSafariOrFirefox` boolean with the optional number, but that works 😂

PR #8952, opened 2023-08-01
TINY-10022: insert new block commands

vpyshnenko:
Related Ticket: TINY-10022

Description of Changes:
* Added two new commands for inserting new empty block before / after current position.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -8,8 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - New `help_accessibility` option which displays the keyboard shortcut to access the help functionality in the statusbar. #TINY-9379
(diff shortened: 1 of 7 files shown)


TheSpyder (commented):
I haven't had a chance to review it fully, but I will say - I was expecting what is now `NewBlock` to either use or refactor code from `InsertNewLine` and `InsertBlock` 🤔

vpyshnenko:
@TheSpyder, the logic of these commands is way simpler. There is not much we need to reuse from `InsertNewLine`.  We don't need to delete the current ranged selection, choose linebreak/block,  inherit parent styles, or split the current block. In these commands, we want just to create an empty block before/after the current one.

TheSpyder (approved):
Ah I didn't look closely enough to see that `createNewBlock` had been refactored. Thank you for your patience.

TheSpyder (approved):
Ah I didn't look closely enough to see that `createNewBlock` had been refactored. Thank you for your patience.

PR #8953, opened 2023-08-01
DOC-2158: edits to bring changelog.md into sync with changelog.adoc, and with the 6.6.1 Release Notes sub-heads

brian-forte:
Related Ticket: DOC-2158

Description of Changes:
* Edits to bring changelog.md into sync with changelog.adoc in the docs and with the sub-heads in the forthcoming 6.6.1 Release Notes.

Pre-checks:

Review:

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Added
-- New filtering option, `pad_empty_with_br`. Set to `true` to pad empty block elements with `<br>` tags instead of the `&nbsp;` character entity. #TINY-9861
+- Restored filtering option, `pad_empty_with_br`. Set to `true` to pad empty block elements with `<br>` tags instead of the `&nbsp;` character entity. #TINY-9861
 
 ### Improved
 - When Safari is the host browser, content updates for iframe dialog components with `streamContent: true` set are now throttled to 500ms intervals. #TINY-10097
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -7,18 +7,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Added
-- New `pad_empty_with_br` option that can be set to `true` to pad empty block elements with `<br>` tag instead of nbsp character. #TINY-9861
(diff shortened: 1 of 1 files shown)


modules/tinymce/CHANGELOG.md
```diff
@@ -7,18 +7,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Added
-- New `pad_empty_with_br` option that can be set to `true` to pad empty block elements with `<br>` tag instead of nbsp character. #TINY-9861
+- New filtering option, `pad_empty_with_br`. Set to `true` to pad empty block elements with `<br>` tags instead of the `&nbsp;` character entity. #TINY-9861
```
tiny-ben-tran, line 10: Didn't we agree on rewording this entry so it doesn't sound like a new option for a minor release?

tiny-ben-tran (approved):

modules/tinymce/CHANGELOG.md
brian-forte: > Didn't we agree on rewording this entry so it doesn't sound like a new option for a minor release?

We did, yes. Thanks for the reminder.

I’ve been tossing up between _Renewed_ and _Restored_ as the first word in the sentence. As in:

* **Renewed** filtering option, `pad_empty_with_br`. Set to `true`…; or
* **Restored** filtering option, `pad_empty_with_br`. Set to `true`….

TheSpyder: Call it _restored_, because we never documented the previous option so nobody other than that one customer will know we renamed it :)

Should we also move from `Added` to perhaps `Improved`?

TheSpyder (approved):

modules/tinymce/CHANGELOG.md
brian-forte, line 10:
```suggestion
- Restored filtering option, `pad_empty_with_br`. Set to `true` to pad empty block elements with `<br>` tags instead of the `&nbsp;` character entity. #TINY-9861
```

brian-forte: > Call it restored

Done.

> Should we also move from _Added_ to, perhaps, _Improved_?

I’ve gone back and forth on this one.

It is absolutely an addition but being able to do something new is generally considered an improvement. (And, to almost the entire user base, this option does enable doing something new.)

Against that, it is the return of previously extant capability. And that argues for it being not an improvement but, as we now note, a restoration. Which is an argument for putting the entry (and the release note write-up) into the _Fixed_ section.

_Addition_ feels, to me, like putting it into a category that is simultaneously not-best-fit and best-fit-we’ve-got.

PR #8986, opened 2023-08-15
TINY-10136: Prevent list generation to go outside of expected child nodes.

HAFRMO:
Related Ticket: https://ephocks.atlassian.net/browse/TINY-10136

Description of Changes:
Restrict the DOMtreeWalker to the correct parent.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -26,7 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Autocompletion of url for links would on sufficiently long links and wide screens leave out middle parts of the url from view. #TINY-10017
 - Toolbar number input would not properly disable upon changing mode. #TINY-10129
 - Quick Toolbar plugin would show text alignment buttons on pagebreaks. #TINY-10054
-- Creating lists would in some situations include unexpected elements into the list. #TINY-10136
+- Creating lists in empty blocks would sometimes include adjacent blocks. #TINY-10136
 - Creating a list from multiple div elements would only create a partial list. #TINY-9872
 - Tab navigation incorrectly stopped around `iframe` dialog component. #TINY-9815
 - It was possible to delete the sole empty block just before a details element if it was nested within another details element. #TINY-9965
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Autocompletion of url for links would on sufficiently long links and wide screens leave out middle parts of the url from view. #TINY-10017
 - Toolbar number input would not properly disable upon changing mode. #TINY-10129
 - Quick Toolbar plugin would show text alignment buttons on pagebreaks. #TINY-10054
+- Creating lists would in some situations include unexpected elements into the list. #TINY-10136
 - Creating a list from multiple div elements would only create a partial list. #TINY-9872
 - Tab navigation incorrectly stopped around `iframe` dialog component. #TINY-9815
 - It was possible to delete the sole empty block just before a details element if it was nested within another details element. #TINY-9965
--- a/modules/tinymce/src/plugins/lists/main/ts/actions/ToggleList.ts
+++ b/modules/tinymce/src/plugins/lists/main/ts/actions/ToggleList.ts
@@ -64,7 +64,7 @@ const getEndPointNode = (editor: Editor, rng: Range, start: Boolean, root: Node)
   }
 
   const findBlockAncestor = (node: Node) => {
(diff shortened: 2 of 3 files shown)


vpyshnenko (approved):

modules/tinymce/src/plugins/lists/main/ts/actions/ToggleList.ts
```diff
@@ -76,9 +84,12 @@ const getEndPointNode = (editor: Editor, rng: Range, start: Boolean, root: Node)
   const findBetterContainer = (container: Node, forward: boolean): Optional<Node> => {
     const walker = new DomTreeWalker(container, root);
     const dir = forward ? 'next' : 'prev';
+    const blockParent = SugarElement.fromDom(findBlockAncestor(container));
     let node;
     while ((node = walker[dir]())) {
-      if (!(NodeType.isVoid(editor, node) || Unicode.isZwsp(node.textContent as string) || node.textContent?.length === 0)) {
+      const suitableType = !(NodeType.isVoid(editor, node) || Unicode.isZwsp(node.textContent as string) || node.textContent?.length === 0);
+      const isWithinParent = Has.ancestor(SugarElement.fromDom(node), blockParent);
```
spocke, line 91: Since this seems to just make sure that the node we find isn't outside the scope of the parent block then you could just feed the DomTreeWalker the parent node as a scope if it exists otherwise fallback to the root. So this function then becomes:

```
  const findBetterContainer = (container: Node, forward: boolean): Optional<Node> => {
    const scope = editor.dom.getParent(container, editor.dom.isBlock) ?? root;
    const walker = new DomTreeWalker(container, scope);
    const dir = forward ? 'next' : 'prev';
    let node;
    while ((node = walker[dir]())) {
      if (!(NodeType.isVoid(editor, node) || Unicode.isZwsp(node.textContent as string) || node.textContent?.length === 0)) {
        return Optional.some(node);
      }
    }

    return Optional.none();
  };
```

Where the actual two lines changes are:
```
    const scope = editor.dom.getParent(container, editor.dom.isBlock) ?? root;
    const walker = new DomTreeWalker(container, scope);
```

That seems to the same thing an the tests seems to pass at least the `ApplyTest.ts`.

modules/tinymce/src/plugins/lists/test/ts/browser/ApplyTest.ts
```diff
@@ -900,6 +900,17 @@ describe('browser.tinymce.plugins.lists.ApplyTest', () => {
     TinyAssertions.assertRawContent(editor, expected);
   });
 
+  it('TINY-10136: Do not list-convert too many elements', () => {
```
spocke, line 903: Could we describe the scenario a bit better here. Not sure of the exact wording but it basically a problem it it trying to move the start and end container position outside of the block if it's applied to an empty inline element inside that block.

modules/tinymce/CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Autocompletion of url for links would on sufficiently long links and wide screens leave out middle parts of the url from view. #TINY-10017
 - Toolbar number input would not properly disable upon changing mode. #TINY-10129
 - Quick Toolbar plugin would show text alignment buttons on pagebreaks. #TINY-10054
+- Creating lists would in some situations include unexpected elements into the list. #TINY-10136
```
spocke, line 29: Maybe:

```suggestion
- Creating lists in empty blocks would sometimes include adjacent blocks. #TINY-10136
```

spocke (changes requested):

modules/tinymce/src/plugins/lists/main/ts/actions/ToggleList.ts
HAFRMO: I'd still say we should step up until we find the block, just to prevent a cases where inlines are stacked. But yes, this is a much neater solution which I should have thought of.

modules/tinymce/src/plugins/lists/test/ts/browser/ApplyTest.ts
HAFRMO: Sure, I was trying to keep it short, but might have been too short.

spocke (approved):

modules/tinymce/src/plugins/lists/main/ts/actions/ToggleList.ts
spocke: Right I guess it should be `const scope = editor.dom.getParent(container, editor.dom.isBlock) ?? container;` since the function you have either finds a block parent or falls back to the container. But I guess we could keep the function to if that is clearer.

vpyshnenko (approved):

PR #8995, opened 2023-08-21
TINY-10115: Add support for case sensitive translations

mkzhx:
Related Ticket: TINY-10115

Description of Changes:
* Where there are multiple case sensitive variants of a translation key, they will now all be kept in the translation object instead of just the lowercase variant
* If there is no lowercase variant present, then we manually add one to ensure that case insensitive matching still works

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/api/util/I18n.ts
+++ b/modules/tinymce/src/core/main/ts/api/util/I18n.ts
@@ -118,7 +118,7 @@ const translate = (text: Untranslated): TranslatedString => {
     // make sure we work on a string and return a string
     const textStr = toString(text);
     return Obj.has(langData, textStr)
-      ? langData[textStr].toString()
+      ? toString(langData[textStr])
       : Obj.get(langData, textStr.toLowerCase()).map(toString).getOr(textStr);
   };
 
--- a/modules/tinymce/src/core/main/ts/api/util/I18n.ts
+++ b/modules/tinymce/src/core/main/ts/api/util/I18n.ts
@@ -20,8 +20,8 @@ export type Untranslated = Primitive | TokenisedString | RawString | null | unde
 export type TranslatedString = string;
 
 const isDuplicated = (items: string[], item: string) => {
-  let count = 0;
-  return Arr.exists(items, (x) => (x === item ? ++count : count) > 1);
+  const firstIndex = items.indexOf(item);
+  return firstIndex !== -1 && items.indexOf(item, firstIndex + 1) > firstIndex;
 };
 
 const isRaw = (str: any): str is RawString => Type.isObject(str) && Obj.has(str, 'raw');
--- a/modules/tinymce/src/core/main/ts/api/util/I18n.ts
+++ b/modules/tinymce/src/core/main/ts/api/util/I18n.ts
@@ -19,6 +19,11 @@ export type Untranslated = Primitive | TokenisedString | RawString | null | unde
 
 export type TranslatedString = string;
 
+const isDuplicated = (items: string[], item: string) => {
+  let count = 0;
+  return Arr.exists(items, (x) => (x === item ? ++count : count) > 1);
+};
+
 const isRaw = (str: any): str is RawString => Type.isObject(str) && Obj.has(str, 'raw');
 
 const isTokenised = (str: any): str is TokenisedString => Type.isArray(str) && str.length > 1;
@@ -68,8 +73,10 @@ const add = (code: string, items: Record<string, string>): void => {
   const lcNames = Arr.map(Obj.keys(items), (name) => name.toLowerCase());
   Obj.each(items, (translation, name) => {
     const lcName = name.toLowerCase();
-    let count = 0;
-    if (lcName !== name && Arr.exists(lcNames, (lc) => (lc === lcName ? ++count : count) > 1)) {
+    if (lcName !== name && isDuplicated(lcNames, lcName)) {
+      if (!Obj.has(items, lcName)) {
+        langData[lcName] = translation;
+      }
       langData[name] = translation;
     } else {
       langData[lcName] = translation;
--- a/modules/tinymce/src/core/main/ts/api/util/I18n.ts
+++ b/modules/tinymce/src/core/main/ts/api/util/I18n.ts
@@ -117,7 +117,9 @@ const translate = (text: Untranslated): TranslatedString => {
   const getLangData = (text: Untranslated) => {
     // make sure we work on a string and return a string
     const textStr = toString(text);
-    return Obj.get(langData, Obj.has(langData, textStr) ? textStr : textStr.toLowerCase()).map(toString).getOr(textStr);
+    return Obj.has(langData, textStr)
+      ? langData[textStr].toString()
+      : Obj.get(langData, textStr.toLowerCase()).map(toString).getOr(textStr);
   };
 
   const removeContext = (str: string) => str.replace(/{context:\w+}$/, '');
--- a/modules/tinymce/src/core/main/ts/api/util/I18n.ts
+++ b/modules/tinymce/src/core/main/ts/api/util/I18n.ts
@@ -1,4 +1,4 @@
-import { Cell, Obj, Type } from '@ephox/katamari';
+import { Arr, Cell, Obj, Type } from '@ephox/katamari';
 
 /**
  * I18n class that handles translation of TinyMCE UI.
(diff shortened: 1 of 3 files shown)


modules/tinymce/src/core/main/ts/api/util/I18n.ts
```diff
@@ -19,6 +19,11 @@ export type Untranslated = Primitive | TokenisedString | RawString | null | unde
 
 export type TranslatedString = string;
 
+const isDuplicated = (items: string[], item: string) => {
+  let count = 0;
+  return Arr.exists(items, (x) => (x === item ? ++count : count) > 1);
+};
```
TheSpyder, line 25: This just feels slow. Here's a cool trick that [I found on google](https://www.jsowl.com/check-for-duplicate-strings-in-an-array-in-javascript/):

```suggestion
const isDuplicated = (items: string[], item: string) => items.indexOf(item) !== items.lastIndexOf(item);
```

modules/tinymce/src/core/main/ts/api/util/I18n.ts
```diff
@@ -103,7 +117,7 @@ const translate = (text: Untranslated): TranslatedString => {
   const getLangData = (text: Untranslated) => {
     // make sure we work on a string and return a string
     const textstr = toString(text);
-    return Obj.get(langData, textstr.toLowerCase()).map(toString).getOr(textstr);
+    return Obj.get(langData, textstr).or(Obj.get(langData, textstr.toLowerCase())).map(toString).getOr(textstr);
```
TheSpyder, line 120: I know this is my code, but it just occurred to me that because JS isn't a lazy language this will double our calls to `Obj.get`. We might need to unroll this (`Obj.get` is just an `if` block with `Obj.has`) to avoid impacting startup performance (when we will hit this function a _lot_).

TheSpyder (approved):

modules/tinymce/src/core/main/ts/api/util/I18n.ts
mkzhx: If I'm not wrong, my algorithm stops when it has encountered a duplicate when searching through the array sequentially, while your algorithm checks through the array from the front and back. So I feel the speed is roughly the same but depends on the order of the array (unless browsers have implemented some optimisations for `indexOf`).

I feel that in most cases the duplicate keys will be placed next to each other so searching through the array sequentially will yield a faster result (assuming the browser doesn't arbitrarily reorder the keys when calling `Object.keys`).

What do you reckon?

modules/tinymce/src/core/main/ts/api/util/I18n.ts
mkzhx: That's a good point, is something like this what you had in mind?
```
return Obj.get(langData, Obj.has(langData, textStr) ? textStr : textStr.toLowerCase()).map(toString).getOr(textStr);

```

modules/tinymce/src/core/main/ts/api/util/I18n.ts
TheSpyder: What I'm trying to avoid is using a function to do `===` instead of leveraging the underlying `indexOf` implementation. It would take a gigantic array, with duplicates very close together, for your implementation to be faster. Unless you feel like switching to a `for` loop, those are well optimised :)

The other option (also from that link) is something like

```
const afterFound = items.indexOf(item) + 1;
return items.indexOf(item, afterFound) > afterFound;
```

Which I think will achieve what you're after, continue searching sequentially for duplicates.

modules/tinymce/src/core/main/ts/api/util/I18n.ts
TheSpyder: It's not quite what I was after - `Obj.get` will call `Obj.has` again - but it's fast enough. `hasOwnProperty` calls should be cheap.

TheSpyder (approved):

modules/tinymce/src/core/main/ts/api/util/I18n.ts
```diff
@@ -102,8 +116,8 @@ const translate = (text: Untranslated): TranslatedString => {
 
   const getLangData = (text: Untranslated) => {
     // make sure we work on a string and return a string
-    const textstr = toString(text);
-    return Obj.get(langData, textstr.toLowerCase()).map(toString).getOr(textstr);
+    const textStr = toString(text);
+    return Obj.get(langData, Obj.has(langData, textStr) ? textStr : textStr.toLowerCase()).map(toString).getOr(textStr);
```
ltrouton, line 120: Can we have `Obj.has(langData, textStr) ? textStr : textStr.toLowerCase()` as separate variable above. Seems like the line is getting bit long  

This is another alternative that might be a bit clearer

```suggestion
    return Obj.get(langData, textStr)
    	.orThunk(() => Obj.get(langData, textStr.toLowerCase())
    	.map(toString)
    	.getOr(textStr);
```

modules/tinymce/src/core/test/ts/browser/util/I18nTest.ts
```diff
@@ -88,4 +88,37 @@ describe('browser.tinymce.core.util.I18nTest', () => {
...
+      assertTranslation('TEST', 'TEST');
+    });
+
+    it('All case sensitive variants of duplicate keys should be preserved and a lowercase variant should be added where a lowercase variant is not present', () => {
+      addTranslationsAndSetCode('code-ci1', {
+        Test: 'Test',
+        TEST: 'TEST'
+      });
+
+      assert.oneOf(I18n.translate('test'), [ 'Test', 'TEST' ]);
```
ltrouton, line 119: Can we be more specific than `assert.oneOf`? The translation for the lowercase key is expected to be the last entry i.e. 'TEST' wouldn't it?

ltrouton (approved):

modules/tinymce/src/core/main/ts/api/util/I18n.ts
mkzhx: Yes, this one would be a more optimised version of my idea using the `indexOf` method, thanks for the suggestion.

modules/tinymce/src/core/main/ts/api/util/I18n.ts
mkzhx: Would this be the performant unroll you had in mind? Doing it this way means we don't have to create an Optional each time as well.
```ts
    return Obj.has(langData, textStr)
      ? langData[textStr].toString()
      : Obj.get(langData, textStr.toLowerCase()).map(toString).getOr(textStr);
```

To further improve performance we can even get rid of the need to create an Optional as well:
```
    if (Obj.has(langData, textStr)) {
      return langData[textStr].toString();
    } else {
      const lcTextStr = textStr.toLowerCase();
      if (Obj.has(langData, lcTextStr)) {
        return langData[lcTextStr].toString();
      } else {
      return textStr;
      }
    }
```

Thoughts?

modules/tinymce/src/core/main/ts/api/util/I18n.ts
mkzhx: The goal of rewriting this was to remove duplicated calls to `Obj.has` which happens as part of `Obj.get` to improve performance. I've have a couple of ideas doing it: https://github.com/tinymce/tinymce/pull/8995#discussion_r1305360182

modules/tinymce/src/core/test/ts/browser/util/I18nTest.ts
mkzhx: In the case where a lowercase variant of duplicated keys is not specified, we would create a lowercase variant using the value of whichever variant of the duplicate keys that come first depending on how the browser ordered the object. Since object order is not guaranteed to be retained, it can be either of those two values.

modules/tinymce/src/core/main/ts/api/util/I18n.ts
TheSpyder: Sorry for not following up earlier. I don't quite remember exactly what I had planned, but the unrolled optional seems like the best answer (this might become a hot codepath with a lot of plugins running).

Do note that `toString` is a function, here, not just calling `.toString()` on the value. I don't know why it's using a custom toString, but I'd rather not investigate that right now 😂

modules/tinymce/src/core/main/ts/api/util/I18n.ts
TheSpyder: Thunk was one of the options I considered, but we generate enough short-lived objects as it is and this code will run close to 1000 times with all the plugins loaded

TheSpyder (approved):
Just fix that `toString` and then we're good

spocke (approved):

PR #8996, opened 2023-08-22
TINY-10124: Improved the keyboard focus visibility for links inside dialogs

lostkeys:
Related Ticket: TINY-10124

Description of Changes:
* Improved the color contrast between normal, hover and active states
* Added a focus outline around links when using keyboard navigation
* Updated the link colors in the dark skin to be AA compliant.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/oxide/src/less/theme/components/dialog/dialog.less
+++ b/modules/oxide/src/less/theme/components/dialog/dialog.less
@@ -41,11 +41,11 @@
 
 @dialog-body-link-color: @color-tint;
 @dialog-body-link-text-decoration: underline;
-@dialog-body-link-hover-color: darken(@color-tint, 20%);
+@dialog-body-link-hover-color: darken(@dialog-body-link-color, 20%);
 @dialog-body-link-hover-text-decoration: underline;
-@dialog-body-link-active-color: darken(@color-tint, 30%);
+@dialog-body-link-active-color: darken(@dialog-body-link-color, 30%);
 @dialog-body-link-active-text-decoration: underline;
-@dialog-body-link-focus-outline-color: @color-tint;
+@dialog-body-link-focus-outline-color: @dialog-body-link-color;
 @dialog-body-link-focus-outline-border-radius: 1px;
 
 @dialog-body-h1-font-size: @font-size-lg;
@@ -235,8 +235,8 @@
   }
 
   .tox-dialog__body-nav-item--active {
-    border-bottom: 2px solid @color-tint;
-    color: @color-tint;
+    border-bottom: 2px solid @dialog-body-link-color;
+    color: @dialog-body-link-color;
   }
 
   .tox-dialog__body-content {
--- a/modules/oxide/src/less/theme/components/dialog/dialog.less
+++ b/modules/oxide/src/less/theme/components/dialog/dialog.less
@@ -40,11 +40,13 @@
 @dialog-body-text-transform: none;
 
 @dialog-body-link-color: @color-tint;
-@dialog-body-link-text-decoration: none;
-@dialog-body-link-hover-color: darken(@color-tint, 10%);
-@dialog-body-link-hover-text-decoration: none;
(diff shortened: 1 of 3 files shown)


lostkeys:
I made a tradeoff between implementation speed, ease LESS variable customization and the visual looks. Using `outline` was the most practical approach even if it means the outline isn't as nice looking as the buttons. I also deliberately limited the amount of LESS styling the developer can do to not go overboard with introducing new variables. Below is an example of how it will look:

Focused links

modules/oxide/src/less/theme/components/dialog/dialog.less
```diff
@@ -40,11 +40,13 @@
 @dialog-body-text-transform: none;
 
 @dialog-body-link-color: @color-tint;
-@dialog-body-link-text-decoration: none;
-@dialog-body-link-hover-color: darken(@color-tint, 10%);
-@dialog-body-link-hover-text-decoration: none;
-@dialog-body-link-active-color: darken(@color-tint, 10%);
-@dialog-body-link-active-text-decoration: none;
+@dialog-body-link-text-decoration: underline;
```
noxuhax, line 43: That’s totally up to you, but imo having links permanently underlined would make plugins list usecase mentioned in the Jira look rather busy. Though I agree that having links underlined before hover (and possibly removing the underline on hover) is a better for the a11y perspective, so I guess the final decision would depend on which situation is more likely yo be encountered: lists of links or a single link inside a paragraph as in the example in the PR.

modules/oxide/src/less/theme/components/dialog/dialog.less
```diff
@@ -40,11 +40,13 @@
...
+@dialog-body-link-hover-color: darken(@color-tint, 20%);
+@dialog-body-link-hover-text-decoration: underline;
+@dialog-body-link-active-color: darken(@color-tint, 30%);
```
noxuhax, line 46: Same as with hovers, probably shouldn’t blanket make it all darker. Honestly, I think we’re good with having hover and active the same color since unlike the filled buttons, it is harder to notice that change on a thin lines of a font.

noxuhax (changes requested):

spocke (approved):

modules/oxide/src/less/theme/components/dialog/dialog.less
```diff
@@ -40,11 +40,13 @@
 @dialog-body-text-transform: none;
 
 @dialog-body-link-color: @color-tint;
-@dialog-body-link-text-decoration: none;
-@dialog-body-link-hover-color: darken(@color-tint, 10%);
-@dialog-body-link-hover-text-decoration: none;
-@dialog-body-link-active-color: darken(@color-tint, 10%);
-@dialog-body-link-active-text-decoration: none;
+@dialog-body-link-text-decoration: underline;
+@dialog-body-link-hover-color: darken(@color-tint, 20%);
```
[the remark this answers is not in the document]
lostkeys: I've updated the link colors for the dark skin in general. I don't think we commit making it fully AA compliant, but it was fairly bad. This is much better.

Screenshot 2023-08-25 at 11 20 34

modules/oxide/src/less/theme/components/dialog/dialog.less
lostkeys: I think we end on the side of being more accessible with the default skin. It makes it easier to identify them as links even if you don't see color at all.

noxuhax (approved):

modules/oxide/src/less/theme/components/dialog/dialog.less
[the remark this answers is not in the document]
noxuhax: The updated color looks much better, good call! Overall, I’d prefer this color to be dynamically generated for all ugc skins and not baked in just in our own oxide-dark, I understand the importance of the implementation speed given the resources we have.

PR #9002, opened 2023-08-24
TINY-9998: Applying bullist to a custom unordered list would turn list into paragraphs instead of a bulleted list.

hamza0867:
Related Ticket: TINY-9998 

Description of Changes:
* Applying bullist to a custom unordered list would turn list into paragraphs instead of a bulleted list.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - When loading content CSS takes more than half a second, the editor will be set to the "in progress" state until the CSS is ready. #TINY-10008
 
 ### Fixed
+- Clicking in the bullist toolbar button on a selected checklist would turn the list into paragraphs instead of a bulleted list. #TINY-9975
 - Returning an empty string in custom context menu update function would result in a small white line appearing on right click and the native browser context menu would not show up. #TINY-9842
 - Autocompletion of url for links would on sufficiently long links and wide screens leave out middle parts of the url from view. #TINY-10017
 - Toolbar number input would not properly disable upon changing mode. #TINY-10129
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -22,7 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - When loading content CSS takes more than half a second, the editor will be set to the "in progress" state until the CSS is ready. #TINY-10008
 
 ### Fixed
-- Clicking in the bullist toolbar button on a selected checklist would turn the list into paragraphs instead of a bulleted list. #TINY-9975
+- Applying an ordered or unordered list to a selected checklist would incorrectly turn the list into paragraphs. #TINY-9975
 - Returning an empty string in custom context menu update function would result in a small white line appearing on right click and the native browser context menu would not show up. #TINY-9842
 - Autocompletion of url for links would on sufficiently long links and wide screens leave out middle parts of the url from view. #TINY-10017
 - Toolbar number input would not properly disable upon changing mode. #TINY-10129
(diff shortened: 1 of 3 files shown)


modules/tinymce/src/plugins/lists/main/ts/actions/ToggleList.ts
```diff
@@ -316,17 +316,34 @@ const updateList = (editor: Editor, list: Element, listName: 'UL' | 'OL' | 'DL',
   }
 };
 
+const updateCustomList = (editor: Editor, list: Element, listName: 'UL' | 'OL' | 'DL', detail: ListDetail): void => {
+  list.className = list.className.split(' ').filter((cls) => !/\btox\-/.test(cls)).join(' ');
```
TheSpyder, line 320: Surely there's a nicer way to do this by iterating over the `classList` rather than split+filter+join+trim 🤔 
https://developer.mozilla.org/en-US/docs/Web/API/Element/classList

- [ ] update both places

TheSpyder (approved):

modules/tinymce/src/plugins/lists/main/ts/actions/ToggleList.ts
hamza0867: Done!

modules/tinymce/src/plugins/lists/main/ts/actions/ToggleList.ts
```diff
@@ -347,6 +365,13 @@ const toggleSingleList = (editor: Editor, parentList: HTMLElement | null, listNa
       flattenListSelection(editor);
     } else {
       const bookmark = Bookmark.createBookmark(editor.selection.getRng());
+      if (isCustomList(parentList)) {
+        parentList.classList.forEach((cls, _, classList) => {
+          if (!/\btox\-/.test(cls)) {
```
spocke, line 370: I assume this is some rewrite from being a className match on `\b` since `\b` is a word boundary a single class name should never have it. So this would likely be better written as `cls.startsWith('tox-')`. Since we want to remove all classes matching `tox-<something>` right.

modules/tinymce/src/plugins/lists/test/ts/browser/ApplyListOnCustomList.ts
```diff
@@ -0,0 +1,42 @@
...
+describe('browser.tinymce.plugins.lists.ApplyListOnParagraphWithStylesTest', () => {
+
+  const hook = TinyHooks.bddSetupLight<Editor>({
+    indent: false,
+    plugins: 'lists',
+    toolbar: 'numlist bullist',
+    base_url: '/project/tinymce/js/tinymce'
+  }, [ Plugin ]);
+
+  it('TINY-9998: Apply list to custom list should keep the original list structure', () => {
```
spocke, line 17: This test is quite dense could it be separated into the different cases being tested here?

modules/tinymce/CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - When loading content CSS takes more than half a second, the editor will be set to the "in progress" state until the CSS is ready. #TINY-10008
 
 ### Fixed
+- Clicking in the bullist toolbar button on a selected checklist would turn the list into paragraphs instead of a bulleted list. #TINY-9975
```
spocke, line 25: Maybe something like:

```suggestion
- Applying an ordered or unordered list to a selected checklist would incorrectly turn the list into paragraphs. #TINY-9975
```

Since it's more broad than just the toolbar button it could in theory be applied with commands, menus etc.

spocke (approved):

modules/tinymce/src/plugins/lists/main/ts/actions/ToggleList.ts
hamza0867: Done!

modules/tinymce/src/plugins/lists/test/ts/browser/ApplyListOnCustomList.ts
hamza0867: I split this test into two cases, nested lists and flat lists. How do you like the tests now ? Should they split even more or is this okay ? @spocke

spocke (approved):

PR #9007, opened 2023-08-28
TINY-10165: fix for toolbar underfined case

lorenzo-pomili:
Related Ticket: TINY-10165

Description of Changes:
it seems that in [this PR](https://github.com/tinymce/tinymce/pull/8901) I did not consider the case when the toolbar is undefined, I'm not sure how it is possible that this case didn't pop up while I was debugging the previous PR (maybe a new version of the browser? :thinking: )

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/alloy/src/main/ts/ephox/alloy/api/ui/Toolbar.ts
+++ b/modules/alloy/src/main/ts/ephox/alloy/api/ui/Toolbar.ts
@@ -1,4 +1,4 @@
-import { Optional } from '@ephox/katamari';
+import { Fun, Optional } from '@ephox/katamari';
 
 import * as AlloyParts from '../../parts/AlloyParts';
 import * as ToolbarSchema from '../../ui/schema/ToolbarSchema';
@@ -42,7 +42,8 @@ const factory: CompositeSketchFactory<ToolbarDetail, ToolbarSpec> = (detail, com
       extra.behaviours
     ),
     apis: {
-      setGroups
+      setGroups,
+      refresh: Fun.noop
     },
     domModification: {
       attributes: {
--- a/modules/alloy/src/main/ts/ephox/alloy/api/ui/Toolbar.ts
+++ b/modules/alloy/src/main/ts/ephox/alloy/api/ui/Toolbar.ts
@@ -26,9 +26,9 @@ const factory: CompositeSketchFactory<ToolbarDetail, ToolbarSpec> = (detail, com
 
   const getGroupContainer = (component: AlloyComponent) => detail.shell ? Optional.some(component) : AlloyParts.getPart(component, detail, 'groups');
 
-  const refresh = (toolbar: AlloyComponent) => {
-    const toolbarApi: ToolbarApis = toolbar.getApis();
-    if (toolbarApi.refresh) {
+  const refresh = (toolbar: AlloyComponent | undefined) => {
+    const toolbarApi: ToolbarApis | undefined = toolbar?.getApis();
+    if (toolbarApi?.refresh) {
       toolbarApi.refresh();
     }
   };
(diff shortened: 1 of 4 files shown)


modules/alloy/src/main/ts/ephox/alloy/api/ui/Toolbar.ts
```diff
@@ -26,9 +26,9 @@ const factory: CompositeSketchFactory<ToolbarDetail, ToolbarSpec> = (detail, com
 
   const getGroupContainer = (component: AlloyComponent) => detail.shell ? Optional.some(component) : AlloyParts.getPart(component, detail, 'groups');
 
-  const refresh = (toolbar: AlloyComponent) => {
-    const toolbarApi: ToolbarApis = toolbar.getApis();
-    if (toolbarApi.refresh) {
+  const refresh = (toolbar: AlloyComponent | undefined) => {
```
spocke, line 29: From were is this called with no arguments? Is that a mistake since then this function would be an noop?

spocke (approved):

modules/alloy/src/main/ts/ephox/alloy/api/ui/Toolbar.ts
lorenzo-pomili: I ended up replacing it with a `Fun.noop`, since that `refresh` was referencing itself, let me know if you think it could make problem

spocke (approved):

vpyshnenko (approved):

modules/tinymce/src/themes/silver/main/ts/ui/general/OuterContainer.ts
```diff
@@ -81,9 +81,9 @@ interface OuterContainerApis {
 
 interface ToolbarApis {
   readonly setGroups: (toolbar: AlloyComponent, groups: SketchSpec[]) => void;
-  readonly refresh: (toolbar: AlloyComponent) => void;
-  readonly toggle: (toolbar: AlloyComponent) => void;
-  readonly toggleWithoutFocusing: (toolbar: AlloyComponent) => void;
+  readonly refresh?: (toolbar: AlloyComponent) => void;
+  readonly toggle?: (toolbar: AlloyComponent) => void;
+  readonly toggleWithoutFocusing?: (toolbar: AlloyComponent) => void;
```
TheSpyder, line 86: I prefer setting refresh to `Fun.noop` to making the methods optional

- [ ] revert the last commit

TheSpyder (approved):

modules/tinymce/src/themes/silver/main/ts/ui/general/OuterContainer.ts
spocke: This type is wrong and all the other calls to Toolbar apis is wrapped in that mapFrom null check. I think we should be fine with this change.

spocke: But I guess we could do it two stage. Revert this then make another PR that we marge after the release that does this proper.

lorenzo-pomili: done

lorenzo-pomili:
a test failed and I lunched the CI again.

I tried to put a `console.log` in local to see if the problem was that we use that `noop` function, but it seems not to be called, anyway I'll try to do more investigation meanwhile the CI is running to see if maybe the `console.log` was not called for other reasons

cc @spocke , @TheSpyder

lorenzo-pomili:
I get the log with `yarn start` but not with the failing test, I think the 2 things are not related even though they are on the same element

PR #9008, opened 2023-08-28
TINY-9911: Load from javascript instead of css url links.

HAFRMO:
Related Ticket: https://ephocks.atlassian.net/browse/TINY-9911

Description of Changes:
Adding css with the help of js-files ( gulp-file modifications not included as it's only for customer use, we got our Webpack ). No tests, they will be added separately, ideally.

Pre-checks:
* [x] Changelog entry added
* ~Tests have been added (if applicable)~

Review:
* [x] Milestone set


modules/tinymce/src/themes/silver/main/ts/ui/skin/Loader.ts
```diff
@@ -40,14 +40,71 @@ const loadSkin = (isInline: boolean, editor: Editor): Promise<void> => {
...
+    }
+  }
+};
+
+const loadKeySkin = (isInline: boolean, editor: Editor): Promise<void | [void, void]> => {
+  const skinKey = Options.getSkinKey(editor);
+
+  if (skinKey) {
+    // eslint-disable-next-line no-debugger
+    debugger;
```
TheSpyder, line 83: Probably needs more testing, if this is getting into the PR

modules/tinymce/src/themes/silver/main/ts/ui/skin/Loader.ts
```diff
@@ -40,14 +40,71 @@ const loadSkin = (isInline: boolean, editor: Editor): Promise<void> => {
...
+  if (!Options.isSkinDisabled(editor) && Type.isString(skinKey)) {
+    loadRawUiSkins(editor, skinKey);
+    loadShadowDomRawUiSkins(editor, skinKey);
+  }
+
+  return Promise.resolve();
+};
+
+const loadSkin = (isInline: boolean, editor: Editor): Promise<void> => {
+  return Promise.all([
```
TheSpyder, line 102: This will run the promises in parallel, which doesn't seem like the right answer. IMO it should check whether we want to load from url or key, and then do the one thing it's supposed to.

modules/tinymce/src/themes/silver/main/ts/ui/skin/Loader.ts
```diff
@@ -29,9 +30,8 @@ const loadShadowDomUiSkins = (editor: Editor, skinUrl: string): Promise<void> =>
   }
 };
 
-const loadSkin = (isInline: boolean, editor: Editor): Promise<void> => {
+const loadUrlSkin = (isInline: boolean, editor: Editor): Promise<void | [void, void]> => {
```
TheSpyder, line 33: You could stick an empty `.then()` on the end to replace the one you removed and avoid adding the second promise type

modules/tinymce/src/themes/silver/main/ts/ui/skin/Loader.ts
```diff
@@ -40,14 +40,71 @@ const loadSkin = (isInline: boolean, editor: Editor): Promise<void> => {
...
+  if (Type.isString(css)) {
+    return loadRawCss(editor, skinUiCss, css, editor.ui.styleSheetLoader);
+  }
+};
+
+const loadShadowDomRawUiSkins = (editor: Editor, skinKey: string): void => {
+  const isInShadowRoot = SugarShadowDom.isInShadowRoot(SugarElement.fromDom(editor.getElement()));
+  if (isInShadowRoot) {
+    const shadowDomSkinCss = skinKey + '/skin.shadowdom.css';
+    const css = EditorManager.resources.get(shadowDomSkinCss);
```
TheSpyder, line 70: This duplication makes me cringe. I wonder if we could re-use `skin_url` rather than making the new `skin_key` with all the dancing around how it's similar but different.

It's a big change, but then we'd check if the URL was in the editor resources, if it is load as raw CSS if not load as URL.

Does that make sense? Would we have time for a change that big?

TheSpyder (changes requested):

modules/tinymce/src/themes/silver/main/ts/ui/skin/Loader.ts
HAFRMO: We could do that, if you prefer, but I had some concerns that people might always want to use the normal url one, and the other was a suppliment. So just using one wasn't ideal. But I might have been wrong.

modules/tinymce/src/themes/silver/main/ts/ui/skin/Loader.ts
HAFRMO: Sure, if that's preferrable.

modules/tinymce/src/themes/silver/main/ts/ui/skin/Loader.ts
HAFRMO: When I made this I actively did not want to reuse anything existing due to the exciting risks that can come from overloading an option.
Not saying I'm a big fan of the duplication either, so not like I'd be opposed to removing it somehow.
And the time-limit I don't know, it seems like we're moving this to... patch version? Next version? I think?

TheSpyder: I think it's a valid risk to take - it would also allow developers to cache their custom resources by simply adding the contents with the full URL as the key.

modules/tinymce/src/themes/silver/main/ts/ui/skin/Loader.ts
TheSpyder: I think loading the resource as a string should always take priority over loading from the URL, even if the key is a URL.

modules/tinymce/src/core/main/ts/api/EditorManager.ts
```diff
@@ -89,6 +90,7 @@ const purgeDestroyedEditor = (editor: Editor | null): void => {
 
 interface EditorManager extends Observable<EditorManagerEventMap> {
   defaultOptions: RawEditorOptions;
+  resources: EditorResources;
```
TheSpyder, line 93: I don't know how I missed this last time, apologies for not bringing it up earlier.

We might need a better name for this; now that I'm not focused on the skin loader code I was about to say "why do we need `EditorManager.resources` when we already have `tinymce.resource`"

I understand that this one is only for pre-caching, where `tinymce.resource` is caching by URL, but they are _very_ similar concepts particularly now that we overloaded the CSS URL in the skin loader.

We can't simply do `tinymce.resources.load(cssurl)`, because if it's not cached it will attempt to download it as JS. But if we add a new method to check if a resource has loaded, that might replace the need for `EditorManager.resources` 🤔 

I'm not sure which path to suggest here. I would prefer to delete this and find a way to use the existing resource cache, but it might not be an easy change.

- [ ] Rename, or replace with a tweaked`tinymce.resource`

TheSpyder (changes requested):

modules/tinymce/src/themes/silver/main/ts/ui/skin/Loader.ts
```diff
@@ -14,40 +15,72 @@ const loadStylesheet = (editor: Editor, stylesheetUrl: string, styleSheetLoader:
...
 
   // In Modern Inline, this is explicitly called in editor.on('focus', ...) as well as in render().
   // Seems to work without, but adding a note in case things break later
   if (!Options.isSkinDisabled(editor) && Type.isString(skinUrl)) {
     return Promise.all([
-      loadUiSkins(editor, skinUrl),
-      loadShadowDomUiSkins(editor, skinUrl)
-    ]).then(SkinLoaded.fireSkinLoaded(editor), SkinLoaded.fireSkinLoadError(editor, 'Skin could not be loaded'));
+      loadUrlUiSkins(editor, skinUrl),
+      loadShadowDomUrlUiSkins(editor, skinUrl)
```
TheSpyder, line 73: Do they need to be renamed now? Perhaps keep the old names so the diff is easier to read 🤔

TheSpyder:
I am happy with the skin loader architecture now, thank you for that

modules/tinymce/src/core/main/ts/api/EditorManager.ts
HAFRMO: The problem I see with that is that the 'outside' resources aren't loaded due to activity from the editor, 'adds themselves' when they are run. They'd need to load themselves to each editor as it boots up ( presumably by listening to events in EditorManager... ) which doesn't seem that practical. `EditorManager.resources` means there's a set target for resources to be loaded into ( and loaded once ).

HAFRMO: Unless I have missed something. If you have a working idea I'd be glad to see if we can add it.

modules/tinymce/src/themes/silver/main/ts/ui/skin/Loader.ts
HAFRMO: Probably not. I'll change them back.

modules/tinymce/src/core/main/ts/api/EditorManager.ts
spocke: My idea was just to use the existing resource API since skins but maybe add a get function. So the .js would do something like:
tinymce.resources.add('skin/dark/whatever'. 'bunch of css');

The you would have the new get function:
const cssData = tinymce.resouces.get('skin/dark.whatever');
if (cssData) {
   addStyleTagToPage(editor, cssData);
} else {
  loadCssFromUrl(someUrl);
}

I really thought we had a `get` function but I see now that we only have a load that would try to load the resource in from an url if not found. Get could be a sync function since it would not do anything async like loading.

HAFRMO: Maybe I am just missing something obvious, but how would those call get a link to the editor? The closest I can think of would be to use `tinymce.activeEditor`, but that'd leave all other instances in the dirt. Or is `tinymce.resources` an instance shared between all editors?

TheSpyder: Yes `tinymce.resource` is a global pool shared across the entire page, the same way your `EditorManager.resources` is. The `tinymce.resource.load` api requires a URL, but `tinymce.resource.add` doesn't - it just uses a string key.

Spocke and I both believe that if you make a small tweak to the existing resource pool - I agree with his suggestion of adding the `get` method - it will be equivalent to the resource pool you created.

At that point it would be a more or less direct translation of the code you have today to the global resource api; continue using the CSS URL as the key, and it will be up to the bundle code to make sure that URL is added as a string key to the global resource pool before the editor loads.

HAFRMO: Alright.

modules/tinymce/src/themes/silver/main/ts/ui/skin/Loader.ts
```diff
@@ -14,26 +15,54 @@ const loadStylesheet = (editor: Editor, stylesheetUrl: string, styleSheetLoader:
...
+  return styleSheetLoader.loadRawCss(key, css);
+};
+
+const loadUiSkins = async (editor: Editor, skinUrl: string): Promise<void> => {
+  const skinUiCss = skinUrl + '/skin.css';
+  const css = await tinymce.Resource.get(skinUiCss);
+
+  if (Type.isString(css)) {
+    loadRawCss(editor, skinUiCss, css, editor.ui.styleSheetLoader);
+    return Promise.resolve();
```
TheSpyder, line 30: You can probably just `return` now that this is an async function 🤔 

`styleSheetLoader.loadRawCss()` doesn't return anything, so I'm surprised this worked before 🤔

TheSpyder (approved):
That's looking really good!

vpyshnenko (approved):

PR #9019, opened 2023-09-01
TINY-9287: Added brightness/saturation 2D selector and hue slider to tabstop targets, added speed multiplier for pressing shift.

HAFRMO:
Related Ticket: https://ephocks.atlassian.net/browse/TINY-9287

Description of Changes:
Included the Brightness/Saturation selection area and the Hue slider in the tabbing rotation.
Also added a speed multiplier when pressing shift, as requested.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/acid/src/main/ts/ephox/acid/gui/components/HueSlider.ts
+++ b/modules/acid/src/main/ts/ephox/acid/gui/components/HueSlider.ts
@@ -30,8 +30,7 @@ const sliderFactory = (translate: (key: string) => string, getClass: (key: strin
       tag: 'div',
       classes: [ getClass('hue-slider') ],
       attributes: {
-        'role': 'presentation',
-        'aria-role': 'slider',
+        'role': 'slider',
         'aria-valuemin': 0,
         'aria-valuemax': 360,
         'aria-valuenow': 120,
--- a/modules/acid/src/main/ts/ephox/acid/gui/components/HueSlider.ts
+++ b/modules/acid/src/main/ts/ephox/acid/gui/components/HueSlider.ts
@@ -1,5 +1,6 @@
-import { AlloyComponent, AlloyTriggers, Behaviour, Focusing, SketchSpec, Slider } from '@ephox/alloy';
+import { AlloyComponent, AlloyTriggers, Behaviour, Focusing, SketchSpec, Slider, Tabstopping } from '@ephox/alloy';
 import { Fun } from '@ephox/katamari';
+import { Attribute } from '@ephox/sugar';
 
 import { sliderUpdate } from '../ColourEvents';
 
@@ -29,7 +30,11 @@ const sliderFactory = (translate: (key: string) => string, getClass: (key: strin
       tag: 'div',
       classes: [ getClass('hue-slider') ],
       attributes: {
-        role: 'presentation'
+        'role': 'presentation',
+        'aria-role': 'slider',
+        'aria-valuemin': 0,
+        'aria-valuemax': 360,
+        'aria-valuenow': 120,
       }
     },
     rounded: false,
@@ -42,10 +47,12 @@ const sliderFactory = (translate: (key: string) => string, getClass: (key: strin
       thumb
     ],
     sliderBehaviours: Behaviour.derive([
+      Tabstopping.config({ }),
       Focusing.config({ })
     ]),
 
     onChange: (slider: AlloyComponent, _thumb: any, value: any) => {
+      Attribute.set(slider.element, 'aria-valuenow', Math.floor(360 - (value * 3.6)));
       AlloyTriggers.emitWith(slider, sliderUpdate, {
         value
       });
--- a/modules/acid/src/main/ts/ephox/acid/gui/components/SaturationBrightnessPalette.ts
+++ b/modules/acid/src/main/ts/ephox/acid/gui/components/SaturationBrightnessPalette.ts
@@ -1,5 +1,6 @@
 import { AlloyComponent, AlloyTriggers, Behaviour, Composing, Focusing, Sketcher, SketchSpec, Slider, SliderTypes, UiSketcher } from '@ephox/alloy';
-import { Fun, Optional } from '@ephox/katamari';
+import { Fun, Optional, Type } from '@ephox/katamari';
+import { Attribute } from '@ephox/sugar';
 
 import { Hex } from '../../api/colour/ColourTypes';
 import * as HsvColour from '../../api/colour/HsvColour';
@@ -74,6 +75,7 @@ const paletteFactory = (_translate: (key: string) => string, getClass: (key: str
   const setPaletteThumb = (slider: AlloyComponent, hex: Hex): void => {
     const hsv = HsvColour.fromRgb(RgbaColour.fromHex(hex));
     Slider.setValue(slider, { x: hsv.saturation, y: 100 - hsv.value });
+    Attribute.set(slider.element, 'aria-valuetext', `Saturation ${hsv.saturation}%, Brightness ${hsv.value}%`);
   };
 
   const factory: UiSketcher.SingleSketchFactory<SaturationBrightnessPaletteDetail, SaturationBrightnessPaletteSpec> = (_detail): SketchSpec => {
@@ -83,6 +85,9 @@ const paletteFactory = (_translate: (key: string) => string, getClass: (key: str
     });
 
     const onChange = (slider: AlloyComponent, _thumb: AlloyComponent, value: number | SliderTypes.SliderValue) => {
+      if (!Type.isNumber(value)) {
+        Attribute.set(slider.element, 'aria-valuetext', `Saturation ${Math.floor(value.x)}%, Brightness ${Math.floor(100 - value.y)}%`);
+      }
       AlloyTriggers.emitWith(slider, ColourEvents.paletteUpdate, {
         value
       });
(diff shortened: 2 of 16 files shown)


modules/acid/src/main/ts/ephox/acid/gui/components/SaturationBrightnessPalette.ts
```diff
@@ -104,7 +109,8 @@ const paletteFactory = (_translate: (key: string) => string, getClass: (key: str
       dom: {
         tag: 'div',
         attributes: {
-          role: 'presentation'
+          'role': 'presentation',
+          'aria-valuetext': 'Saturation 0%, Brightness 0%'
```
spocke, line 113: Since this has role presentation no screen reader will ever read this. But there really is no role for this looking at what CK is doing they use a slider role here I guess the screen reader will just read the aria-valuetext anyway but it's a bit weird since the user would then never know that they can move in all four directions. Not sure there is a best practices here for this exotic component.

modules/acid/src/main/ts/ephox/acid/gui/components/SaturationBrightnessPalette.ts
```diff
@@ -74,6 +75,7 @@ const paletteFactory = (_translate: (key: string) => string, getClass: (key: str
   const setPaletteThumb = (slider: AlloyComponent, hex: Hex): void => {
     const hsv = HsvColour.fromRgb(RgbaColour.fromHex(hex));
     Slider.setValue(slider, { x: hsv.saturation, y: 100 - hsv.value });
+    Attribute.set(slider.element, 'aria-valuetext', `Saturation ${hsv.saturation}%, Brightness ${hsv.value}%`);
```
spocke, line 78: We need to pass these though I18n there is variables that you can use so can do this: https://fiddle.tiny.cloud/EGiaab

The component gets a translate function passed in that takes a key and returns a value that will not really work in this case. So it needs to take the key and the numbers to produce a full string maybe we can change the translate function to do an array instead for the color picker component that way you could pass N args like we do for the translate function in the backstage.

spocke (changes requested):

modules/acid/src/main/ts/ephox/acid/gui/components/SaturationBrightnessPalette.ts
HAFRMO: Looking at the guidelines they all use three sliders. We do have the manual input fields which serve a similar purpose, so there is still something.

vpyshnenko:
Found that with Shift+Tab revert movement the navigation stops on the 2D panel and could not pass further.

modules/acid/src/main/ts/ephox/acid/gui/components/SaturationBrightnessPalette.ts
```diff
@@ -19,7 +20,11 @@ export interface SaturationBrightnessPaletteSketcher extends Sketcher.SingleSket
   setThumb: (slider: AlloyComponent, hex: Hex) => void;
 }
 
-const paletteFactory = (_translate: (key: string) => string, getClass: (key: string) => string): SaturationBrightnessPaletteSketcher => {
+const paletteFactory = (translate: (key: string) => string, getClass: (key: string) => string): SaturationBrightnessPaletteSketcher => {
+  const translateReplace = (key: string, value1: number, value2: number): string => {
+    return translate(key).replace('{0}', value1 + '').replace('{1}', value2 + '');
+  };
```
TheSpyder, line 26: You don't need to replace manually, translate actually takes an array - although the `ColorPicker` translate function will need to be updated to leverage this, that's where the `key: string` type is enforced.
https://github.com/tinymce/tinymce/blob/ece3bf774b9ae20910d073af5351f162c4fb52d5/modules/tinymce/src/core/main/ts/api/util/I18n.ts#L122-L126

Here's how it's used:
https://github.com/tinymce/tinymce/blob/46ae40b9e7dbab275fd0731ca34ed0ea0633e475/modules/tinymce/src/themes/silver/main/ts/ui/statusbar/Statusbar.ts#L30

TheSpyder (changes requested):

modules/acid/src/main/ts/ephox/acid/gui/components/SaturationBrightnessPalette.ts
TheSpyder: Oh you'd already mentioned this. right.

Part of the complexity here is the component `translate` function actually maps to codes, because it was ported from TBIO verbatim, and it has a custom translate wrapper function that enforces strings.
https://github.com/tinymce/tinymce/blob/92f53fe893ca049293c02e5fe82a9ffa3056afc3/modules/tinymce/src/themes/silver/main/ts/ui/dialog/ColorPicker.ts#L11-L27

TheSpyder (approved):

modules/acid/src/main/ts/ephox/acid/gui/components/SaturationBrightnessPalette.ts
spocke: This needs to change somehow so either:

1) Change the role to `slider` and pass the hard coded `aira-valuetext` though the translation function and verify that it works on screen readers.
2) Remove the `aria-valuetext` and remove the tab stop for this thing. Tabbing to a presentation element should not be done since the screenreader will likely not pick anything up.

modules/acid/src/main/ts/ephox/acid/gui/components/HueSlider.ts
```diff
@@ -29,7 +30,11 @@ const sliderFactory = (translate: (key: string) => string, getClass: (key: strin
       tag: 'div',
       classes: [ getClass('hue-slider') ],
       attributes: {
-        role: 'presentation'
+        'role': 'presentation',
+        'aria-role': 'slider',
```
spocke, line 34: This should just be `role` not sure there is a `aira-role`?

- [ ] Remove `aira-role` replace `role` with `slider`

modules/acid/src/main/ts/ephox/acid/gui/components/SaturationBrightnessPalette.ts
```diff
@@ -104,7 +111,8 @@ const paletteFactory = (_translate: (key: string) => string, getClass: (key: str
       dom: {
         tag: 'div',
         attributes: {
-          role: 'presentation'
+          'role': 'presentation',
```
spocke, line 114: This is still not fixed. If this is `presentation` I think the screen reader will just ignore any aria properties on it and thus this PR is not fixing what the ticket is about.

- [ ] Set proper slider role or something that the screen reader will pick up

modules/acid/src/main/ts/ephox/acid/gui/components/SaturationBrightnessPalette.ts
```diff
@@ -104,7 +111,8 @@ const paletteFactory = (_translate: (key: string) => string, getClass: (key: str
...
+          'aria-valuetext': 'Saturation 0%, Brightness 0%'
```
spocke, line 115: Is this translated by a setAttribute call later, if that is the case remove this string. Serves no purpose. If not then translate this as well though the translate function with 0 and 0 as it's parameter inputs.

modules/alloy/src/main/ts/ephox/alloy/ui/slider/SliderParts.ts
```diff
@@ -90,6 +91,8 @@ const thumbPart = PartType.required<SliderDetail, { dom: OptionalDomSchema; even
   }
 });
 
+const isShift = (event: any) => event.raw ? event.raw.shiftKey : false;
```
spocke, line 94: Never use `any` in production code if there is a proper type. In this case you can do the following:

1. Use the `KeyMatch.isShift` function from the KeyMatch module import it as a named import
2. Change this to `const isShift = (evt: NativeSimulatedEvent<KeyboardEvent>) => evt.event.raw.shiftKey;` and remove the multiple calls to `event.event` (DRY)
3. Or do both make a isShift function that just calls `KeyMatch.isShift` but unpacks the event.event inside it.

- [ ] Remove the use of the `any` type or remove the predicate

spocke (approved):

modules/acid/src/main/ts/ephox/acid/gui/components/SaturationBrightnessPalette.ts
HAFRMO: This ticket is actually not at all about this, it's about keyboard navigation explicitly for those who can see. The aria screenreader is something we're adding since we're changing things in the area. But sure, I can change that.

modules/acid/src/main/ts/ephox/acid/gui/components/SaturationBrightnessPalette.ts
HAFRMO: No, I forgot to translate this one. That's on me.

modules/acid/src/main/ts/ephox/acid/gui/components/SaturationBrightnessPalette.ts
spocke: But this would be the first thing we ever have in the UI that you can tab to that doesn't have any description at all in a screen reader. So if you are blind and accidentally tab to that thing you wouldn't know what it is. We need to think about the bigger picture not degrade a11y in one part while improving it in another part.

HAFRMO: No, I agree with you that it needs improvement, not arguing that at all. It is undeniably not as good as it should be. Or arguably 'good' at all. It just wasn't what the ticket is about ( and quite frankly this component isn't exactly accommodating either )

PR #9020, opened 2023-09-01
TINY-10088: update dompurify

lorenzo-pomili:
Related Ticket: TINY-10088

Description of Changes:

- [x] update DOMPurify
- [x] create a DOMPurify's hook to preserve the original attributes order
- [x] remove the local DOMPurify's patch

About the patch there were 3 changes:

- move `_removeAttribute` inside `!hookEvent.keepAttr` if
- add `_removeAttribute(name, currentNode)` to `!_isValidAttribute(lcTag, lcName, value)` if
- add a check of different between the current and the initial value before the handler for invalid data-* attributes

testing the previous version with the first 2 changes I see that some test fails because of the order, so the new hook covers that case.

~About the 3rd one I cannot find a case for that I'm continuing to dig to find it, but maybe we can remove it~
The 3rd one seems like a performance optimization not so impactful, I'll remove it entirely

### **EDIT:**

The patch was meant also to improve the performance in particular for `src` attributes that can cause prefetches and since to avoid useless prefetches we need the 3rd change but to make it works we also need the other 2 changes the solution is to re-implement the whole patch for the new version

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/tinymce/src/core/main/ts/api/html/Sanitization.ts
```diff
@@ -152,6 +157,25 @@ const setupPurify = (settings: DomParserSettings, schema: Schema): DOMPurifyI =>
     }
   });
 
+  purify.addHook('afterSanitizeAttributes', (node) => {
+    const { attributes } = node;
+    const currentOrder = Arr.from(attributes);
```
TheSpyder, line 162: Why did you copy the attributes to an array, and then search it?

There's no need, you can just do an object lookup as the original suggestion did.

TheSpyder (approved):
Upgrading all the way to v3, that's bold!

(Given that the patch has been deleted, I was going to suggest it if you hadn't 😂 )

TheSpyder (approved):
Upgrading all the way to v3, that's bold!

(Given that the patch has been deleted, I was going to suggest it if you hadn't 😂 )

HAFRMO (approved):

mkzhx (approved):

modules/tinymce/CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+- Setting the content with an attribute that contains a self closed HTML tag did not preserve the tag. #TINY-10088
```
mkzhx, line 10:
```suggestion
- Setting the content with an attribute that contains a self-closing HTML tag did not preserve the tag. #TINY-10088
```

modules/tinymce/src/core/main/ts/api/html/Sanitization.ts
lorenzo-pomili: changed

modules/tinymce/CHANGELOG.md

spocke (approved):

TheSpyder (approved):

PR #9026, opened 2023-09-05
Fix applying heading to `summary` content

vpyshnenko:
Related Ticket: TINY-10154

Description of Changes:
* Updated schema to make heading elements as a valid child for summary
* fixed `expandRng` function by adding `summary` element to the list of scopeRoot elements

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/api/html/DomParser.ts
+++ b/modules/tinymce/src/core/main/ts/api/html/DomParser.ts
@@ -353,8 +353,8 @@ const DomParser = (settings: DomParserSettings = {}, schema = Schema()): DomPars
    */
   const removeAttributeFilter = attributeFilterRegistry.removeFilter;
 
-  const findInvalidChildren = (node: AstNode, invalidChildren: AstNode[], isInsert: boolean): void => {
-    if (InvalidNodes.isInvalid(schema, node, isInsert)) {
+  const findInvalidChildren = (node: AstNode, invalidChildren: AstNode[]): void => {
+    if (InvalidNodes.isInvalid(schema, node)) {
       invalidChildren.push(node);
     }
   };
@@ -429,7 +429,7 @@ const DomParser = (settings: DomParserSettings = {}, schema = Schema()): DomPars
    * @example
    * const rootNode = tinymce.html.DomParser({...}).parse('<b>text</b>');
    */
-  const parse = (html: string, args: ParserArgs = {}, isInsert: boolean = false): AstNode => {
+  const parse = (html: string, args: ParserArgs = {}): AstNode => {
     const validate = defaultedSettings.validate;
     const rootName = args.context ?? defaultedSettings.root_name;
 
@@ -451,7 +451,7 @@ const DomParser = (settings: DomParserSettings = {}, schema = Schema()): DomPars
 
     // Find the invalid children in the tree
     const invalidChildren: AstNode[] = [];
-    const invalidFinder = validate ? (node: AstNode) => findInvalidChildren(node, invalidChildren, isInsert) : Fun.noop;
+    const invalidFinder = validate ? (node: AstNode) => findInvalidChildren(node, invalidChildren) : Fun.noop;
 
     // Set up attribute and node matching
     const matches: FilterNode.FilterMatches = { nodes: {}, attributes: {}};
--- a/modules/tinymce/src/core/main/ts/api/html/DomParser.ts
+++ b/modules/tinymce/src/core/main/ts/api/html/DomParser.ts
@@ -80,7 +80,7 @@ interface DomParser {
   addNodeFilter: (name: string, callback: ParserFilterCallback) => void;
   getNodeFilters: () => ParserFilter[];
   removeNodeFilter: (name: string, callback?: ParserFilterCallback) => void;
-  parse: (html: string, args?: ParserArgs, isInsert?: boolean) => AstNode;
+  parse: (html: string, args?: ParserArgs) => AstNode;
 }
 
 type WalkerCallback = (node: AstNode) => void;
@@ -424,7 +424,6 @@ const DomParser = (settings: DomParserSettings = {}, schema = Schema()): DomPars
    * @method parse
    * @param {String} html Html string to sax parse.
    * @param {Object} args Optional args object that gets passed to all filter functions.
-   * @param {Boolean} isInsert Optional denoting whether the current action involves inserting content into the editor.
    * @return {tinymce.html.Node} Root node containing the tree.
    * @example
    * const rootNode = tinymce.html.DomParser({...}).parse('<b>text</b>');
--- a/modules/tinymce/src/core/main/ts/api/html/DomParser.ts
+++ b/modules/tinymce/src/core/main/ts/api/html/DomParser.ts
@@ -80,7 +80,7 @@ interface DomParser {
   addNodeFilter: (name: string, callback: ParserFilterCallback) => void;
   getNodeFilters: () => ParserFilter[];
   removeNodeFilter: (name: string, callback?: ParserFilterCallback) => void;
-  parse: (html: string, args?: ParserArgs) => AstNode;
+  parse: (html: string, args?: ParserArgs, isInsert?: boolean) => AstNode;
 }
 
 type WalkerCallback = (node: AstNode) => void;
@@ -353,8 +353,8 @@ const DomParser = (settings: DomParserSettings = {}, schema = Schema()): DomPars
    */
   const removeAttributeFilter = attributeFilterRegistry.removeFilter;
 
-  const findInvalidChildren = (node: AstNode, invalidChildren: AstNode[]): void => {
-    if (InvalidNodes.isInvalid(schema, node)) {
+  const findInvalidChildren = (node: AstNode, invalidChildren: AstNode[], isInsert: boolean): void => {
+    if (InvalidNodes.isInvalid(schema, node, isInsert)) {
       invalidChildren.push(node);
     }
   };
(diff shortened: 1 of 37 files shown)


modules/tinymce/src/core/test/ts/browser/fmt/ExpandRangeTest.ts
```diff
@@ -270,6 +270,14 @@ describe('browser.tinymce.core.fmt.ExpandRangeTest', () => {
       const rng = expandRng(editor, [ 0, 0 ], 1, [ 0, 0 ], 2, selectorFormat, false);
       assertRange(editor, rng, [], 0, [], 1);
     });
+
+    it('TINY-10154: should expand over the nested summary content instead of parent details body element', () => {
```
spocke, line 274: Maybe also add a test to the formatter since you could still run the apply H1 on a summary without having the accordion plugin enabled so the change is in core but affects the accordion. Just a simple tests that test this on a regular `<details><summary>text</summary>blah</details>` element.

modules/tinymce/src/plugins/accordion/test/ts/browser/AccordionFormattingTest.ts
```diff
@@ -0,0 +1,36 @@
...
+  const hook = TinyHooks.bddSetup<Editor>(
+    {
+      plugins: 'accordion',
+      base_url: '/project/tinymce/js/tinymce',
+      indent: false
+    },
+    [ Plugin ]
+  );
+
+  it('TINY-10154: should apply heading to the  summary', () => {
```
spocke, line 19:
```suggestion
  it('TINY-10154: should apply heading to the summary', () => {
```

spocke (approved):

modules/tinymce/src/core/test/ts/browser/fmt/ExpandRangeTest.ts
vpyshnenko: done

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
```diff
@@ -80,7 +80,7 @@ interface DomParser {
   addNodeFilter: (name: string, callback: ParserFilterCallback) => void;
   getNodeFilters: () => ParserFilter[];
   removeNodeFilter: (name: string, callback?: ParserFilterCallback) => void;
-  parse: (html: string, args?: ParserArgs) => AstNode;
+  parse: (html: string, args?: ParserArgs, isInsert?: boolean) => AstNode;
```
spocke, line 83: I don't think we need to track inserts as separate things and if we need to it should really be a `ParserArgs` property in fact there seems to already be a `insert` boolean there might not be typed properly. The multiple headings in a summary is never valid regardless when you parse so I really think we should avoid making it a special thing for insert. If the editor some how manages to produce a incorrect structure by some other operation than insert then that will not be filtered up if we have this extra check in there.

- [ ] Remove extra insert tracking

modules/tinymce/src/core/main/ts/html/InvalidNodes.ts
```diff
@@ -159,17 +159,31 @@ const hasClosest = (node: AstNode, parentName: string): boolean => {
...
   // Anchors are a special case and cannot be nested
-  } else if (parent && node.name === 'a' && hasClosest(parent, 'a')) {
+  if (node.name === 'a' && hasClosest(parent, 'a')) {
     return true;
-  } else {
-    return false;
   }
+
+  // heading element is valid if it is the only one child of summary (though on inserting it is still invalid)
+  if ([ 'h1', 'h2', 'h3', 'h4', 'h5', 'h6' ].includes(node.name) && hasClosest(node, 'summary')) {
```
spocke, line 179: This is the same as the call in `DetailsElements.ts` I think it would be cleaner to make a `main/ts/html/AstType` module that has the `isHeading` function here and use it here and in the details elements.

modules/tinymce/src/core/main/ts/selection/DetailsElement.ts
```diff
@@ -11,6 +12,23 @@ const preventSummaryToggle = (editor: Editor): void => {
   });
 };
 
+const isHeading = (node: AstNode) => [ 'h1', 'h2', 'h3', 'h4', 'h5', 'h6' ].includes(node.name);
```
spocke, line 15: So move this to a new module like `AstNodeType`. Then use is in this module as a named import like `AstNodeType.isHeading` there are likely other candidates to become a generic predicate like that later on.

modules/tinymce/src/core/main/ts/selection/DetailsElement.ts
```diff
@@ -11,6 +12,23 @@ const preventSummaryToggle = (editor: Editor): void => {
...
+
+// If the `summary` element contains multiple heading elements,
+// merge their contents into the first heading element.
+const handleSummaryHeadings = (details: AstNode) =>
+  Arr.findMap(details.children(), (node: AstNode) => Optionals.someIf(node.name === 'summary', node))
```
spocke, line 20: Isn't this the same as also maybe move the summary predicate to the `AstNodeType` if you create such a module.

```suggestion
  Arr.find(details.children(), (node) => node.name === 'summary'))
```

modules/tinymce/src/core/main/ts/selection/DetailsElement.ts
```diff
@@ -11,6 +12,23 @@ const preventSummaryToggle = (editor: Editor): void => {
   });
 };
 
+const isHeading = (node: AstNode) => [ 'h1', 'h2', 'h3', 'h4', 'h5', 'h6' ].includes(node.name);
+
+// If the `summary` element contains multiple heading elements,
+// merge their contents into the first heading element.
+const handleSummaryHeadings = (details: AstNode) =>
```
spocke, line 19: Maybe rename this to `mergeSummaryHeadings` if that is what it's doing?

spocke (approved):

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
vpyshnenko: If I use `ParserArgs` `insert` property for that purpose InsertContetTest.ts would fail, due to [SetContentString](https://github.com/tinymce/tinymce/blob/develop/modules/tinymce/src/core/main/ts/content/SetContentImpl.ts#L63) use it while set (not insert) content.

vpyshnenko: When we have `h1` element as the first child of `summary] on `setContent` it is a valid child whereas on `inserContent` it is an invalid child and we unwrap it.

spocke: It might be better that the single child rule is just a filter that you add. So the schema would make any h1 valid but the filter does a separate pass after the parsing has been done. That way you wouldn't need to modify the parser at all.

spocke: No wait a simple filter would not work since you want to avoid the splitting. We just need to figure out where to add this, adding an extra insert tracking though the whole parser makes the entire thing more complex. You could in theory have a separate cleanup step in the insert code similar to how transparent elements are handled.

spocke: Ok played around with this a bit. What you could do is make sure that it always parses with a full context if you insert headings into a summary element and do that in the InsertContentImpl. The problem here now is that the first child heading rule is not going to get applied if you paste simple content into a simple context since it tries to optimize that case in the insert content code. But this is a complex case so with a special extra check in the insert code you can force it to always take the full context code path and then the first child rule always gets applied properly since then it does it on a bigger fragment including the summary parent context and it's siblings.

modules/tinymce/src/core/main/ts/html/InvalidNodes.ts
```diff
@@ -159,17 +160,31 @@ const hasClosest = (node: AstNode, parentName: string): boolean => {
...
   // Anchors are a special case and cannot be nested
-  } else if (parent && node.name === 'a' && hasClosest(parent, 'a')) {
+  if (node.name === 'a' && hasClosest(parent, 'a')) {
     return true;
-  } else {
-    return false;
   }
+
+  // heading element is valid if it is the only one child of summary (though on inserting it is still invalid)
+  if (AstNodeType.isHeading(node) && hasClosest(node, 'summary')) {
```
spocke, line 180: Maybe be specific here and only check for headings as a direct child of the summary not any parent that is summary that the `hasClosest` will match. However not sure if that is ever going to happen.

vpyshnenko: When we insert `h2` in the `summary` that already contains `h1`, then the closest parent would be `h1` but we still need to handle this case to correctly merge headings.

lorenzo-pomili (approved):

modules/tinymce/src/core/main/ts/html/AstNodeType.ts
```diff
@@ -0,0 +1,10 @@
+import AstNode from '../api/html/Node';
+
+const isHeading = (node: AstNode): boolean => [ 'h1', 'h2', 'h3', 'h4', 'h5', 'h6' ].includes(node.name);
```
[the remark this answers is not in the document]
spocke: Yeah, maybe it should all go though the schema but I guess headings is unlikely that they can be reconfigured. Maybe we hard code it now in this PR then make a new feature PR that adds predicates to the schema for all the things that make sense then we can migrate to that over time. I've been contemplating if we should have node and astnode predicates in the schema or not. The problem with adding such a thing to the schema is that we kind of make the schema DOM dependent. But then again the parser is know DOM dependant anyway and that is a heavy user of the schema. Maybe it's better to just have the schema have predicates for strings then we have abtractions on that internally that takes those predicates and makes new ones for Dom nodes, sugar element nodes and parser astnodes.

PR #9027, opened 2023-09-05
TINY-9911: Changed gulp-file to generate javascript files instead of css.

HAFRMO:
Related Ticket: 

Description of Changes:
Used for testing withing the epic. Not intended for production release at this time. Will not pass tests due to removing the css files.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/oxide/gulpfile.js
+++ b/modules/oxide/gulpfile.js
@@ -94,7 +94,7 @@ gulp.task('minifyCss', function() {
     .pipe(cleanCSS({ rebase: false }))
     .pipe(through2.obj(function(file, _, cb) {
       if (file.isBuffer()) {
-        const contents = `tinymce.resources.add('${file.relative}', '${file.contents.toString().replace(/'/g, "\\'")}')`;
+        const contents = `tinymce.resources.add('${file.relative}', '${JSON.stringify(file.contents.toString())}')`;
         file.contents = Buffer.from(contents)
       }
       cb(null, file);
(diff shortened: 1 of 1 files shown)


modules/oxide/gulpfile.js
```diff
@@ -90,7 +92,14 @@ gulp.task('minifyCss', function() {
   return gulp.src(['./build/skins/**/*.css', '!**/*.min.css'])
     .pipe(sourcemaps.init())
     .pipe(cleanCSS({ rebase: false }))
-    .pipe(rename({ extname: '.min.css' }))
+    .pipe(through2.obj(function(file, _, cb) {
+      if (file.isBuffer()) {
+        const contents = `tinymce.resources.add('${file.relative}', '${file.contents.toString().replace(/'/g, "\\'")}')`;
```
spocke, line 97: I think it's better to use JSON.stringify here so make sure that all things gets escaped properly.

HAFRMO: Which part do you want to put stringify on? The file is css, so it shouldn't have anything to escape in the first place?

spocke: CSS has plenty of things that might need escaping.

All this is valid css:
```
[foo="bar"], [foo='bar'] {}
.foo { background: url('bar') };
.foo { background: url("bar") };
```

spocke: Might be other characters in there as well. Just replacing the quotes just seems a bit weak that all. JSON.stringify will make sure that it's always a valid string.

HAFRMO: Alright. Can do that.

vpyshnenko (approved):

PR #9031, opened 2023-09-11
TINY-10209: Added support for large inline dialogs and updating dialog size on redial

MitchCTiny:
Related Ticket: TINY-10209

Description of Changes:
* Added large dialog class to inline dialogs when large size is specified
* Redialling dialogs with a new size now updates the size class
* Repositioning and re-docking was also necessary after the size of inline dialogs are updated.
* Modal dialogs only need the class to be updated to display correctly
* URL dialogs still do not support updating the size on redial
* Dialog sizes are retained and restored when fullscreen is toggled (api.toggleFullscreen)
* Some cleanup with reused functionality across modal, url, and inline dialogs, now that all sizes are supported and updated
  * Fullscreen toggle functionality
  * Retrieving and updating the size class.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/themes/silver/main/ts/ui/window/SilverDialogCommon.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/ui/window/SilverDialogCommon.ts
@@ -73,7 +73,7 @@ const fullscreenClass = 'tox-dialog--fullscreen';
 const largeDialogClass = 'tox-dialog--width-lg';
 const mediumDialogClass = 'tox-dialog--width-md';
 
-const getDialogSizeClasses = (size: Dialog.DialogSize): Optional<string> => {
+const getDialogSizeClass = (size: Dialog.DialogSize): Optional<string> => {
   switch (size) {
     case 'large':
       return Optional.some(largeDialogClass);
@@ -90,14 +90,14 @@ const updateDialogSizeClass = (size: Dialog.DialogSize, component: AlloyComponen
   if (!Arr.contains(classes, fullscreenClass)) {
     const currentSizeClass = Arr.find(classes, (c) => c === largeDialogClass || c === mediumDialogClass);
     currentSizeClass.map((c) => Classes.remove(sugarBody, [ c ]));
-    getDialogSizeClasses(size).map((dialogSizeClass) => Classes.add(sugarBody, [ dialogSizeClass ]));
+    getDialogSizeClass(size).map((dialogSizeClass) => Classes.add(sugarBody, [ dialogSizeClass ]));
   }
 };
 
 const toggleFullscreen = (comp: AlloyComponent, currentSize: Dialog.DialogSize): void => {
   const sugarBody = SugarElement.fromDom(comp.element.dom);
   const classes = Classes.get(sugarBody);
-  const currentSizeClass = Arr.find(classes, (c) => c === largeDialogClass || c === mediumDialogClass).or(getDialogSizeClasses(currentSize));
+  const currentSizeClass = Arr.find(classes, (c) => c === largeDialogClass || c === mediumDialogClass).or(getDialogSizeClass(currentSize));
   Classes.toggle(sugarBody, [ fullscreenClass, ...currentSizeClass.toArray() ]);
 };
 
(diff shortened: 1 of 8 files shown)


modules/tinymce/src/themes/silver/main/ts/ui/window/SilverDialogCommon.ts
```diff
@@ -70,37 +69,57 @@ const getEventExtras = (lazyDialog: () => AlloyComponent, providers: UiFactoryBa
...
+      return Optional.some(largeDialogClass);
+    case 'medium':
+      return Optional.some(mediumDialogClass);
+    default:
+      return Optional.none();
+  }
+};
+
+const updateDialogSizeClass = (size: Dialog.DialogSize, component: AlloyComponent): void => {
+  const sugarBody = SugarElement.fromDom(component.element.dom);
```
shanmen-tiny, line 88:
```suggestion
  const dialog = SugarElement.fromDom(component.element.dom);
```
Maybe?

modules/tinymce/src/themes/silver/main/ts/ui/window/SilverDialogCommon.ts
```diff
@@ -70,37 +69,57 @@ const getEventExtras = (lazyDialog: () => AlloyComponent, providers: UiFactoryBa
...
+  const classes = Classes.get(sugarBody);
+  if (!Arr.contains(classes, fullscreenClass)) {
+    const currentSizeClass = Arr.find(classes, (c) => c === largeDialogClass || c === mediumDialogClass);
+    currentSizeClass.map((c) => Classes.remove(sugarBody, [ c ]));
+    getDialogSizeClass(size).map((dialogSizeClass) => Classes.add(sugarBody, [ dialogSizeClass ]));
+  }
```
shanmen-tiny, line 94:
```suggestion
  if (!Class.has(sugarBody, fullscreenClass)) {
    Classes.remove(sugarBody, [ largeDialogClass, mediumDialogClass ]);
    getDialogSizeClass(size).each((dialogSizeClass) => Class.add(sugarBody, dialogSizeClass));
  }
```
Would this be better if we just want to strip `largeDialogClass` or `mediumDialogClass` and add the desired dialog size class?

shanmen-tiny (approved):
Nice work figuring out the docking part! 🚀

modules/tinymce/src/themes/silver/main/ts/ui/window/SilverDialogCommon.ts
MitchCTiny: That looks much better, thanks!

modules/tinymce/src/themes/silver/main/ts/ui/window/SilverDialogCommon.ts
```diff
@@ -70,37 +69,57 @@ const getEventExtras = (lazyDialog: () => AlloyComponent, providers: UiFactoryBa
...
+};
+
+const updateDialogSizeClass = (size: Dialog.DialogSize, component: AlloyComponent): void => {
+  const sugarBody = SugarElement.fromDom(component.element.dom);
+  const classes = Classes.get(sugarBody);
+  if (!Arr.contains(classes, fullscreenClass)) {
+    const currentSizeClass = Arr.find(classes, (c) => c === largeDialogClass || c === mediumDialogClass);
+    currentSizeClass.map((c) => Classes.remove(sugarBody, [ c ]));
+    getDialogSizeClass(size).map((dialogSizeClass) => Classes.add(sugarBody, [ dialogSizeClass ]));
```
TheSpyder, line 93: Sugar puts singular and plural in different places - in fact the plural generally delegates to the singular - so this can be done far more easily
(also please don't use `map` with functions that don't return a value, it makes a new `Optional<undefined>` that is then ignored)
```suggestion
    currentSizeClass.each((c) => Class.remove(sugarBody, c));
    getDialogSizeClass(size).each((dialogSizeClass) => Class.add(sugarBody, dialogSizeClass));
```

modules/tinymce/src/themes/silver/main/ts/ui/window/SilverDialogCommon.ts
TheSpyder: Ah I was wondering if we'd exposed `classList.contains`
https://developer.mozilla.org/en-US/docs/Web/API/Element/classList

yes this is probably the better solution

TheSpyder (approved):

PR #9035, opened 2023-09-14
TINY-10213: manage li with a list and then other elements

lorenzo-pomili:
Related Ticket: TINY-10213

Description of Changes:
the problem here was that in order to merge on Indent a `li` with the successive `li` if that has a list as the first element, we parse that list but ignore the other elements inside the `li`.

to solve this issue I added the siblings in the parsing of that list.

But since our `Entry`'s structure wasn't design only to have entries that are `li` or `lists` so I added a kind of `Entry` that is neither a `li` or a `list` to manage the issue.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Setting the content with an attribute that contains a self-closing HTML tag did not preserve the tag. #TINY-10088
 - Resize handles would not appear on editable images in a non-editable context. #TINY-10118
 - The dialog size was not updated when the `size` argument was changed when redialling a dialog. #TINY-10209
+- Indent/Undent a list that contains a not `li` with a `ul`/`ol` as first element make disappear all others elements inside the `li`. #TINY-10213
 
 ### Improved
 - Improved the tooltips of picker buttons for the urlinput components in the "Insert/Edit Image" and "Insert/Edit Link" dialogs. #TINY-10155
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Setting the content with an attribute that contains a self-closing HTML tag did not preserve the tag. #TINY-10088
 - Resize handles would not appear on editable images in a non-editable context. #TINY-10118
 - The dialog size was not updated when the `size` argument was changed when redialling a dialog. #TINY-10209
-- Indent/Undent a list that contains a not `li` with a `ul`/`ol` as first element make disappear all others elements inside the `li`. #TINY-10213
+- Toggling a list that contains an LI element having another list as its first child would remove the remaining content within that LI element. #TINY-10213
 
 ### Improved
 - Improved the tooltips of picker buttons for the urlinput components in the "Insert/Edit Image" and "Insert/Edit Link" dialogs. #TINY-10155
--- a/modules/tinymce/src/plugins/lists/main/ts/listmodel/ComposeList.ts
+++ b/modules/tinymce/src/plugins/lists/main/ts/listmodel/ComposeList.ts
@@ -1,7 +1,7 @@
 import { Arr, Optional, Optionals } from '@ephox/katamari';
 import { Attribute, Css, Insert, InsertAll, Replication, SugarElement, SugarNode } from '@ephox/sugar';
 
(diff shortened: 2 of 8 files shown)


modules/tinymce/src/plugins/lists/main/ts/listmodel/ComposeList.ts
```diff
@@ -92,10 +109,28 @@ const writeDeep = (scope: Document, cast: Segment[], entry: Entry): Segment[] =>
...
-  const cast = Arr.foldl(entries, (cast, entry) => {
-    return entry.depth > cast.length ? writeDeep(scope, cast, entry) : writeShallow(scope, cast, entry);
+  let firstCommentEntryOpt: Optional<Entry> = Optional.none();
+
+  const cast = Arr.foldl(entries, (cast, entry, i) => {
+    if (isEntryList(entry)) {
+      return entry.depth > cast.length ? writeDeep(scope, cast, entry) : writeShallow(scope, cast, entry);
+    } else {
+      if (i === 0 && isEntryComment(entry)) {
+        firstCommentEntryOpt = Optional.some(entry);
```
HAFRMO, line 119: Having a side-effect like this feels very fragile. Is there no other way to do this? Even if it's a bit less efficient in compute-time?
And if there is no better way could a comment perhaps be added, to make it easier to understand what's going on for future developers.

TheSpyder: I don't understand what this comment is doing at all. It feels like a patch rather than addressing the underlying issue.

lorenzo-pomili: this is actually not strictly related to the issue but since for this issue we need to manage not-list element like `p`, I thought that it was correct to also fix the case with `comments`.

in particular this part is used to manage the case when the `comment` is the first element so something like this:
```
<ol>
  <!-- a -->
  <li>b</li>
</ol>
```

in particular this exception is needed because otherwise we are not able to generate a list to put the `comment` in

I added a comment

modules/tinymce/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Setting the content with an attribute that contains a self-closing HTML tag did not preserve the tag. #TINY-10088
 - Resize handles would not appear on editable images in a non-editable context. #TINY-10118
+- Indent/Undent a list that contains a not `li` with a `ul`/`ol` as first element make disappear all others elements inside the `li`. #TINY-10213
```
vpyshnenko, line 12: Toggling a list that contains an LI element having another list as its first child would remove the remaining content within that LI element.

modules/tinymce/src/plugins/lists/test/ts/browser/IndentTest.ts
```diff
@@ -449,6 +449,79 @@ describe('browser.tinymce.plugins.lists.IndentTest', () => {
...
+      '</li>' +
+      '<!-- c2 -->' +
+      '</ol>'
+    );
+
+    LegacyUnit.setSelection(editor, 'li', 0);
+    editor.execCommand('Outdent');
+
+    TinyAssertions.assertContent(editor,
+      '<p>a</p>' +
```
vpyshnenko, line 473: How this paragraph ends up converted from the LI?

lorenzo-pomili: here the action is `'Outdent'` not `'Indent'`, you can try in fiddle:
```
<ol>
  <li>a</li>
  <li>
    <ol>
      <li>b</li>
    </ol>
  </li>
</ol>
```
and press `shift+tab` (or click on the `toolbar`'s button) in the first `li` since it is at the first level it will be go out of the table

modules/tinymce/CHANGELOG.md

modules/tinymce/src/plugins/lists/test/ts/browser/IndentTest.ts
vpyshnenko: Right, turns out CK editor behaves the same way. I expected nothing would happen if there was no room for outdenting.

vpyshnenko (approved):

hamza0867 (approved):

modules/tinymce/src/plugins/lists/main/ts/listmodel/ComposeList.ts
TheSpyder: That seems like an edge case, not something that's worth adding a mutable tracking variable for

PR #9043, opened 2023-09-20
TINY-10144: Adopt changie for changelogs

ArvinJ-H:
Related Ticket: 
https://ephocks.atlassian.net/browse/TINY-10144

Description of Changes:
- added changie command
- moved 6.8.0 change under changie/unreleased

Pre-checks:
* [x] Changelog entry added
* [x] <s>Tests have been added (if applicable)</s>

Review:
* [x] Milestone set


ArvinJ-H:
before merging this ticket, all current changelog must first be migrated under `.changes`.

modules/tinymce/CHANGELOG.md
```diff
@@ -4,15 +4,6 @@ All notable changes to this project will be documented in this file.
...
-## Unreleased
-
-### Fixed
-- Setting the content with an attribute that contains a self-closing HTML tag did not preserve the tag. #TINY-10088
-- Resize handles would not appear on editable images in a non-editable context. #TINY-10118
-
-### Improved
-- Improved the tooltips of picker buttons for the urlinput components in the "Insert/Edit Image", "Insert/Edit Link" and "Insert/Edit Media" dialogs. #TINY-10155
-
 ## 6.7.0 - 2023-08-30
```
TheSpyder, line 7: Why is it both here and in the `.changes` folder? 🤔  If we want to keep this file we should keep all changes in it. If we don't, it should be removed and git ignored.

package.json
```diff
@@ -33,7 +33,9 @@
     "headless-test-firefox": "yarn -s grunt headless-auto --bedrock-browser=firefox-headless",
     "headless-test-manual": "yarn -s grunt headless-manual",
     "bedrock": "bedrock --customRoutes modules/tinymce/src/core/test/json/routes.json",
-    "test-one": "yarn tsc && yarn bedrock-auto -b chrome-headless -f"
+    "test-one": "yarn tsc && yarn bedrock-auto -b chrome-headless -f",
+    "change": "changie new",
+    "change-merge": "changie batch auto && changie merge"
```
TheSpyder, line 38: We should, at the very least, merge unreleased changes in CI to check there are no errors.

TheSpyder (changes requested):
You missed one crucial requirement - there are multiple changelogs in this project. How do we deal with that? I'm guessing changie doesn't support multiple changelog output files, so we might need a `.changes` folder for each module.

.changes/header.tpl.md
```diff
@@ -0,0 +1,6 @@
+# Changelog
```
spocke, line 1: I think this file should be moved in some other central place so that they can be shared by multiple `.changie.yml` config files. Also the .changes should be one per project since we want to use changie for all the projects in the monorepo. There are just way to easy to make mistakes in changelogs today and getting the conflicts between PRs. They are far less in libraries since we don't change them that often still having a tool would help us having the right format.

modules/tinymce/CHANGELOG.md
spocke: I think we need to have the changelog.md file that gets generated commited in the tinymce repo so that we can point to that and review it easily. I think the different .md files in the .changes need to exist and be commited to so that you can merge that with new changes based on the previous version. There will be a lot of files but it would be more clear on what we have here. But I wonder if we should split apart all of the changelog into separate files here even though we wouldn't use it like that I think the idea is that it would combine all the different versions into one changelog. Having just one file containing all the changes like it is now if kind of not how it supposed to work but I guess there is nothing really blocking us from using it that way.

spocke (dismissed):

package.json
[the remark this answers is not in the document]
spocke: If we need to expand that to other projects that like `sugar-changelog-add` etc? Or should that all be commands inside each package.

modules/tinymce/CHANGELOG.md
TheSpyder: We can't commit the generated file or we'll continue to have conflicts 😞 

I agree we need a way to review it and compare, but it will need to be in local (or CI) compiled output.

Splitting the old changelog into individual files could be interesting, but it sounds like a fair chunk of work to script up and I'm not sure what the benefit would be.

.changes/unreleased/Changed-20230925-111345.yaml
```diff
@@ -0,0 +1,5 @@
+kind: Changed
+body: Changelog is now maintained by changie
```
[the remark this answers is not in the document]
TheSpyder: No, I agree we can delete this one.

package.json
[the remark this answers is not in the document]
TheSpyder: I hope changie is fast enough to compile every changelog in one `yarn` command, have it run all of the individual changie builds one after the other, we wouldn't need to split them.

modules/tinymce/CHANGELOG.md
spocke: No it would only be generated when you release so there would not be any conflicts then. It's all just separate files until you do that step.
Regarding the splitting it just a bit weird to have a 6.7.0 release having all the data. Going forward it will not be like that for future things then every file would be just that release data but I'm fine with it being weird if everyone understands why it's weird. But you are right it doesn't bring us any value beyond that to split it.

package.json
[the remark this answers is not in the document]
ArvinJ-H: new script has been added to handle multiple changelog change and batch merge

modules/tinymce/CHANGELOG.md
ArvinJ-H: the `CHANGELOG.md` when finished should only have released log, unreleased will be kept inside `.changes/unreleased`, each entry will have a separate file, conflict should not be a concern.

The only issue right now, is when running `changie batch/merge`, all the unreleased changes will be merged into the changelog, from what I see there is no config that can be used to select which change should be added, the only way to do so is manually add the changes into a new change log then run `changie merge`

TheSpyder: That still doesn't explain why there is a duplicate of the release changelog in `.changes`. Is that just how changie works? I see it's been added for every other project as well.

modules/tinymce/.changes/unreleased/Changed-20230925-111345.yaml
TheSpyder, line 1: I don't think we want this in the TinyMCE changelog. It's not a user-facing change (it would end up on the docs release notes).

changie-runner.js
```diff
@@ -0,0 +1,39 @@
+const { execSync } = require('child_process');
+const fs = require('fs');
+const path = require('path');
+
+const modules = process.argv[2];
+const command = process.argv.slice(3).join(' ');
+const changieConfigSrc = './.changie.yaml';
+
+if (!modules || !command) {
+    console.error('Module and command are required.');
```
TheSpyder, line 10: You need better help text than that.

TheSpyder (changes requested):
I appreciate the work you put into that script, but I wish you'd asked for help before writing it. A Makefile would be much better suited to this task, or even a shell script instead of a nodejs script.

I'm also not convinced this is the correct approach. Changie _does_ have monorepo support (admittedly it's new, released mid August)
https://changie.dev/config/#config-projects

modules/tinymce/.changes/unreleased/Changed-20230925-111345.yaml
ArvinJ-H: In `.changes` folder, with each new change log entry, changie will create a temporary file for it, the general flow is 
```
1. Developer makes a change 
2. Creates a `.change` file using `changie new`
3. Multiple `.change` files accumulate.
4. run `changie batch {option}` to consolidate unreleased changes into a new changelog under `.change` (eg. `6.8.0.md`)
5. Run Changie to consolidate `.change` changelog into CHANGELOG.md (at root or `/module`)
```
The multiple changelogs under `.changes` folder, assuming they are `6.7.0.md` and `6.8.0.md`.  This modular approach helps avoid merge conflicts when multiple devs add entries concurrently.

After the merge, the new changelog generated by Changie will still exist under `.changes`, from experimenting and testing, the newest changelog is essential for Changie to generate the changelog with the correct version number. The older version however can be deleted.

Regarding the runner, I ran into some issues with the batch and merged all modules at the same time during initial testing. I need to look back on this and see if I missed a config.

The ability to batch and merge all together came up in a discussion of making sure we don't miss any changes modules and also speed up the release process.

modules/tinymce/CHANGELOG.md
spocke: Yes, that is how changie works it basically just concats things. So building a complete change log is to concat all the separate version changelogs together. In this case however we haven't separated the changelogs into those different files so 6.7.0 contains all the history not only the changes that where in the 6.7.0 release so that is why I suggested splitting it since then it would concat like it's supposed to work and make more sense.

ArvinJ-H:
the fragment files are still placeholder for now, I have updated the naming structure to 
```
fragmentFileFormat: "{{.Kind}}-{{.Custom.Issue}}-{{.Project}}"
```
ie. `Added-TINY-10100-tinymce.ymal`

modules/tinymce/CHANGELOG.md
TheSpyder: Right, I see now in the changie quick start guide it says to do this. OK, makes sense.

.eslintrc.json
```diff
@@ -24,5 +24,6 @@
     "prefer-rest-params": "off",
     "prefer-spread": "off",
     "max-len": ["warn", 260]
-  }
+  },
+  "ignorePatterns":["changie-runner.js"]
```
TheSpyder, line 28: No longer used
```suggestion
  }
```

.changes/acid/5.0.8.md
TheSpyder, line 1: File is empty

- [ ] include actual changelog

TheSpyder (changes requested):
Thank you for bearing with us while we figure out how we want this to operate. It's looking really good after the project config change!

Aside from the comments I've made, the big thing this is missing now is documentation. I think at a minimum we need 3 things:

1. Without the package scripts you wrote earlier, we need a section in `CONTRIBUTING.md` to explain how to add changelog entries.
2. Our release process needs to be updated to note that `changie merge` needs to be run. Alternatively just add that step into the existing automated release scripts (I think it's the community build repo).
3. To prevent mistakes from old habits, we need the build to fail if changelogs (which are now generated files) are modified in a PR. Ideally this would be a pre-commit hook but that might be out of scope (although [husky](https://www.npmjs.com/package/husky) does make those easy).

TheSpyder (approved):
Just waiting on changie to be installed on the build pod. I'm really starting to wonder about having a docker container available for our native binary tooling...

.changie.yaml
```diff
@@ -0,0 +1,95 @@
...
+  - label: snooker
+    key: snooker
+    changelog: modules/snooker/CHANGELOG.md
+  - label: sugar
+    key: sugar
+    changelog: modules/sugar/CHANGELOG.md
+  - label: tinymce
+    key: tinymce
+    changelog: modules/tinymce/CHANGELOG.md
+kinds:
```
spocke, line 70: I think the order of the kinds might be wrong here:
https://www.notion.so/tinycloud/Changelog-Standard-3027cf188f104370984b291578b673d3

I think Improved should be before `Fixed`

TheSpyder: Ah I thought this was just for the `changie new` prompt. But you're right, this is the order the kinds are emitted into the changelog file as well.

@Skylite73 looks like you copied our list for the repositories you used changie on, and we were wrong...

.changie.yaml
```diff
@@ -0,0 +1,96 @@
...
+- label: Changed
+  auto: major
+- label: Deprecated
+  auto: minor
+- label: Removed
+  auto: major
+- label: Fixed
+  auto: patch
+- label: Security
+  auto: patch
```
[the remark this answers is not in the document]
TheSpyder: Not initially, but I'd like to, and this file will be the template for other uses of changie

modules/bridge/CHANGELOG.md
```diff
@@ -1,16 +1,14 @@
...
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html),
+and is generated by [Changie](https://github.com/miniscruff/changie).
 
 ## 4.7.0 - 2023-11-22
 
 ### Added
-- Added `setTooltip` function to `ToolbarSplitButtonInstanceApi` and `NestedMenuItemInstanceApi`. #TINY-9796
-- Added `picker_text` optional property to `UrlInputSpec`. #TINY-10155
+- Added `picker_text` optional property to `UrlInputSpec` #TINY-10155
```
TheSpyder, line 11: You're missing changes

ArvinJ-H: done

MitchC1999 (approved):

danoaky-tiny (approved):

modules/alloy/CHANGELOG.md
```diff
@@ -437,221 +436,278 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## 4.7.3 - 2019-01-24
 ### Added
- - Added bubble support for hotspot anchors
+
```
ArvinJ-H, line 439: this white space is github error, it does not appear on actual file

shanmen-tiny (approved):

mkzhx (approved):

PR #9046, opened 2023-09-20
TINY-10237: Enable SVG support

spocke:
Related Ticket: TINY-10237

This is a bit on the big side but you should be able to review it commit by commit.

Description of Changes:
* Adds opt-in support for SVG elements by just allowing SVG then all children of a SVG is also valid
* Adds support for XLINK in SVGs since that is how you do links in previous version of SVG
* Adds opt-out XSS filtering of SVG elements.
* SVG elements are purified as a separate step since it can have SVG, HTML mixed.
* Treats SVG elements as black boxes anything inside them is just retained as it but filtered though sanitation. 
* Changes a bunch of predicate checks to be isHTMLElement before we could cheat by treating all elements in the editor as HTMLElements after this it's no longer the case it can now contain SVGElement nodes.
* Right now it just supports SVGs but this is kind of opening for other sub specs like MathML.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Colorpicker now includes the Brightness/Saturation selector and hue slider in the keyboard navigable items. #TINY-9287
 - Improved the tooltips of picker buttons for the urlinput components in the "Insert/Edit Image" and "Insert/Edit Link" dialogs. #TINY-10155
 - Inline dialog will now respect `size: 'large'` argument in the dialog spec. #TINY-10209
+- SVG elements and it's children is now retained if you configure in SVG elements as a valid element. #TINY-10237
 
 ## 6.7.0 - 2023-08-30
 
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -16,7 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Colorpicker now includes the Brightness/Saturation selector and hue slider in the keyboard navigable items. #TINY-9287
 - Improved the tooltips of picker buttons for the urlinput components in the "Insert/Edit Image" and "Insert/Edit Link" dialogs. #TINY-10155
 - Inline dialog will now respect `size: 'large'` argument in the dialog spec. #TINY-10209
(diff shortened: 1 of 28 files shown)


modules/tinymce/CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Colorpicker now includes the Brightness/Saturation selector and hue slider in the keyboard navigable items. #TINY-9287
 - Improved the tooltips of picker buttons for the urlinput components in the "Insert/Edit Image" and "Insert/Edit Link" dialogs. #TINY-10155
 - Inline dialog will now respect `size: 'large'` argument in the dialog spec. #TINY-10209
+- SVG elements and it's children is now retained if you configure in SVG elements as a valid element. #TINY-10237
```
vpyshnenko, line 19: SVG elements and their children are now retained when configured as valid elements.

vpyshnenko, line 19:
```suggestion
- SVG elements and their children are now retained when configured as valid elements. #TINY-10237
```

vpyshnenko (approved):

modules/tinymce/CHANGELOG.md
spocke: Yes that is much better than what I had. Thanks

modules/tinymce/src/core/main/ts/api/dom/DOMUtils.ts
```diff
@@ -464,7 +464,7 @@ const DOMUtils = (doc: Document, settings: Partial<DOMUtilsSettings> = {}): DOMU
   const getStyle = (elm: string | Element | null, name: string, computed?: boolean): string | undefined => {
     const $elm = get(elm);
 
-    if (Type.isNullable($elm) || !NodeType.isElement($elm)) {
+    if (Type.isNullable($elm) || (!NodeType.isHTMLElement($elm) && !NodeType.isSVGElement($elm))) {
```
lorenzo-pomili, line 467: why not continue using `isElement` here?

lorenzo-pomili (approved):

modules/tinymce/src/core/main/ts/api/dom/DOMUtils.ts
spocke: Because Element doesn't have the `style` property. SVGElement and HTMLElement has that property. So any custom element or things like mathml might not have that property.

PR #9051, opened 2023-09-23
TINY-10010: Empty.isEmpty should respect empty CET

vpyshnenko:
Related Ticket: TINY-10010

Description of Changes:
* Replace in `Quirks.ts` `DomUtils.isEmpty` on `Empty.isEmtpy`
* add to `Empty.isEmpty` function the condition to treat empty CET as a non-empty content

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -16,7 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Resize handles would not appear on editable images in a non-editable context. #TINY-10118
 - The dialog size was not updated when the `size` argument was changed when redialling a dialog. #TINY-10209
 - Toggling a list that contains an LI element having another list as its first child would remove the remaining content within that LI element. #TINY-10213
-- An empty element with a `contenteditable="true"` attribute within a table cell and noneditable root was deleted when the Backspace key was pressed. #TINY-10010
+- An empty element with a `contenteditable="true"` attribute within a table cell would not be treated as content and get removed if backspace or delete was being pressed. #TINY-10010
 
 ### Improved
 - Colorpicker now includes the Brightness/Saturation selector and hue slider in the keyboard navigable items. #TINY-9287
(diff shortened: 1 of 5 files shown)


modules/tinymce/src/core/main/ts/dom/Empty.ts
```diff
@@ -52,6 +52,10 @@ const isEmptyNode = (targetNode: Node, skipBogus: boolean): boolean => {
           continue;
         }
       }
+      if (NodeType.isContentEditableTrue(node)
+        && Traverse.parentElement(SugarElement.fromDom(node)).exists((parent) => !ContentEditable.isEditable(parent))) {
+        return false;
+      }
```
spocke, line 58: I don't think you need to go though sugar and options in this case. I think this should do the same thing.
```suggestion
      if (NodeType.isContentEditableTrue(node) && !ContentEditable.isEditable(node.parentNode)) {
        return false;
      }
```

modules/tinymce/CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Resize handles would not appear on editable images in a non-editable context. #TINY-10118
 - The dialog size was not updated when the `size` argument was changed when redialling a dialog. #TINY-10209
 - Toggling a list that contains an LI element having another list as its first child would remove the remaining content within that LI element. #TINY-10213
+- An empty element with a `contenteditable="true"` attribute within a table cell and noneditable root was deleted when the Backspace key was pressed. #TINY-10010
```
spocke, line 19: The issue seems wider than just non-editable root so we don't need to mention that. However it does happen on both backspace and delete. So maybe this:
```suggestion
- An empty element with a `contenteditable="true"` attribute within a table cell would not be treated as content and get removed if backspace or delete was being pressed. #TINY-10010
```

spocke (approved):

modules/tinymce/src/core/main/ts/dom/Empty.ts
vpyshnenko: `ContentEditable.isEditable` requires the first argument to be SugarElement.

spocke: Right I read that as `dom.isEditable` some how. I think at least the `if` statement should be split up into having a that long statement in there looks a bit odd. Maybe make a isEditableParent predicate of the inline thing just to split it up.

vpyshnenko: added `hasNonEditableParent` predicate

lorenzo-pomili (approved):

PR #9058, opened 2023-09-27
TINY-10154: Improve validation of heading elements in summary

vpyshnenko:
Related Ticket:  TINY-10154

Description of Changes:
* Handle the case when heading elements are inserted to a `summary` already containing the heading element of the same level.
* Changelog added in the previous [PR](https://github.com/tinymce/tinymce/pull/9026)

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/html/InvalidNodes.ts
+++ b/modules/tinymce/src/core/main/ts/html/InvalidNodes.ts
@@ -177,7 +177,7 @@ const isInvalid = (schema: Schema, node: AstNode, isInsert: boolean, parent: Ast
   }
 
   // heading element is valid if it is the only one child of summary (though on inserting it is still invalid)
-  if (AstNodeType.isHeading(node) && hasClosest(node, 'summary')) {
+  if (AstNodeType.isSummary(parent) && AstNodeType.isHeading(node)) {
     if (parent.firstChild === node) {
       return isInsert;
     }
--- a/modules/tinymce/src/core/main/ts/html/InvalidNodes.ts
+++ b/modules/tinymce/src/core/main/ts/html/InvalidNodes.ts
@@ -3,6 +3,7 @@ import { Fun } from '@ephox/katamari';
 import AstNode from '../api/html/Node';
 import Schema from '../api/html/Schema';
 import Tools from '../api/util/Tools';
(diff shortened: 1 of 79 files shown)


lorenzo-pomili (approved):

modules/tinymce/src/core/test/ts/browser/selection/DetailsElementTest.ts
```diff
@@ -135,5 +143,13 @@ describe('browser.tinymce.selection.DetailsElementTest', () => {
       editor.setContent(content);
       TinyAssertions.assertContent(editor, '<details><summary>helloworldGoodBye</summary><div>body</div></details>');
     });
+
+    it('TINY-10154: should unwrap the heading element while pasting in summary', () => {
+      const editor = hook.editor();
+      editor.setContent('<details><summary><h4>helloworld</h4></summary><div>body</div></details>');
+      TinySelections.setCursor(editor, [ 0, 0, 0, 0 ], 'hello'.length);
+      editor.execCommand('mceInsertClipboardContent', false, { html: '<meta charset="utf-8/"><!-- x-tinymce/html --><h2>wonderful</h2>' });
```
spocke, line 151: Maybe better to use agar clipboard stuff here instead of a command.

modules/tinymce/src/core/main/ts/html/InvalidNodes.ts
```diff
@@ -178,7 +178,7 @@ const isInvalid = (schema: Schema, node: AstNode, parent: AstNode | null | undef
 
   // heading element is valid if it is the only one child of summary
   if (AstNodeType.isSummary(parent) && AstNodeType.isHeading(node)) {
-    return !hasOnlyChild(parent, node.name);
+    return !(parent?.firstChild === node && parent?.lastChild === node);
```
spocke, line 181: Right so the parent here is not a direct parent but the closest valid parent element and the node is the invalid node within that context. Maybe we should add some form of comment that explains this better. I think I was assuming that the parent here would be the parent of the node in this case and that is why I wrongly suggested that we could use the `hasOnlyChild` function.

spocke (approved):

PR #9070, opened 2023-10-05
TINY-10133: Don't unwrap `details` inside `li`

vpyshnenko:
Related Ticket: TINY-10133

Description of Changes:
* Add `isUnwrappable` predicate to prevent the `details` from unwrapping when merging the `li` content on Backspace pressing.
* Noticed the same issue for CEF `li` child, so prevent unwrapping CEF as well.
Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/plugins/lists/main/ts/core/Delete.ts
+++ b/modules/tinymce/src/plugins/lists/main/ts/core/Delete.ts
@@ -1,5 +1,5 @@
 import { Arr, Optionals } from '@ephox/katamari';
-import { Compare, PredicateFind, Remove, SugarElement, SugarNode } from '@ephox/sugar';
+import { Compare, ContentEditable, PredicateFind, Remove, SugarElement, SugarNode } from '@ephox/sugar';
 
 import DOMUtils from 'tinymce/core/api/dom/DOMUtils';
 import RangeUtils from 'tinymce/core/api/dom/RangeUtils';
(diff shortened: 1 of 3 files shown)


modules/tinymce/src/plugins/lists/main/ts/core/Delete.ts
```diff
@@ -61,8 +61,12 @@ const hasOnlyOneBlockChild = (dom: DOMUtils, elm: Element): boolean => {
...
+const isUnwrappable = (elm: HTMLElement): boolean =>
+  ContentEditable.isEditable(SugarElement.fromDom(elm))
+    && !Arr.contains([ 'DETAILS' ], elm.nodeName);
+
 const unwrapSingleBlockChild = (dom: DOMUtils, elm: Element): void => {
-  if (hasOnlyOneBlockChild(dom, elm)) {
+  if (hasOnlyOneBlockChild(dom, elm) && isUnwrappable(elm.firstChild as HTMLElement)) {
     dom.remove(elm.firstChild as HTMLElement, true);
   }
 };
```
spocke, line 72: Maybe we should avoid dangerous type casts and do something like this.

```
const isUnwrappable = (node: Node | null): node is HTMLElement =>
  Optional.from(node)
    .map(SugarElement.fromDom)
    .filter(SugarNode.isHTMLElement)
    .exists((el) => ContentEditable.isEditable(el) && Arr.contains([ 'details' ], SugarNode.name(el)));

const unwrapSingleBlockChild = (dom: DOMUtils, elm: Element): void => {
  if (hasOnlyOneBlockChild(dom, elm) && isUnwrappable(elm.firstChild)) {
    dom.remove(elm.firstChild, true);
  }
};

```

spocke (approved):

lorenzo-pomili (approved):

modules/tinymce/src/plugins/lists/main/ts/core/Delete.ts
vpyshnenko: Thanks, that looks much better

Issue #8657, opened 2023-04-17
TinyMCE 6.4.1 | RGB format Issue in SSRS | How to use color in hexadecimal


dixitrohit17:
**Migration - TinyMCE Version 4.9.2 to 6.4.1**

TinyMCE editor is used in MVC 5 Application, where user have option to create the template using editor. Once template designed/created, Editor's complete HTML text will pass to SSRS for generating report. 
Due to color formatting which is in RGB color code, text colors are not reflecting on SSRS generated report.
In earlier version color code is in hexadecimal, with hex text color displayed properly in SSRS.

**Is there any way to set color code in hexadecimal instead of RGB or any other way to make editor html work in SSRS.**

**Note:** Other styling is visible on SSRS report except Text Color.

**TinyMCE Version -** 6.4.1, 
 **Browser -** Google chrome and MS Edge 
 With hexadecimal color code functionality was working in **TinyMCE Version 4.9.2**

dixitrohit17:
Hello Everyone,

Is there any update on this item.

TheSpyder:
This may have been an unintended consequence of a change in TinyMCE 6 where we avoid rewriting RGB colours to hex when the input is RGB. I'll check with our product team, but the fact that our own colour picker which defaults to hex configuration is applying RGB seems like a bug.

dixitrohit17:
@TheSpyder Thank you for addressing the concern. Please let me know if anything requires from my side or any progress on this issue.

PR #9076, opened 2023-10-09
TINY-9819: Added force_hex_color option

danoaky-tiny:
Related Ticket: [TINY-9819](https://ephocks.atlassian.net/browse/TINY-9819)

Description of Changes:
Added an editor option called force_hex_color. Defaults to false. Setting to true will cause applied colors on the "data-mce-style" attribute to be set as HEX values instead of the default RGB (as of a change in v6). Setting to true reflects the behavior of applying colors in v5, as every style color was forced to HEX then.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

[TINY-9819]: https://ephocks.atlassian.net/browse/TINY-9819?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


modules/tinymce/src/core/main/ts/api/html/Styles.ts
```diff
@@ -39,6 +41,7 @@ const Styles = (settings: StylesSettings = {}, schema?: Schema): Styles => {
   /* jshint maxlen:255 */
   /* eslint max-len:0 */
   const urlOrStrRegExp = /(?:url(?:(?:\(\s*\"([^\"]+)\"\s*\))|(?:\(\s*\'([^\']+)\'\s*\))|(?:\(\s*([^)\s]+)\s*\))))|(?:\'([^\']+)\')|(?:\"([^\"]+)\")/gi;
+  const rgbRegExp = /rgb\s*\(\s*([0-9]+)\s*,\s*([0-9]+)\s*,\s*([0-9]+)\s*\)/gi;
```
TheSpyder, line 44: You missed `rgba`, and this already exists:
https://github.com/tinymce/tinymce/blob/f61d071878146aee69f57bdb712c1a6d0e8c8def/modules/acid/src/main/ts/ephox/acid/api/colour/RgbaColour.ts#L10-L11

In fact it's probably better to use `RgbaColour.fromString` the same way the `Transformations` code does.

TheSpyder (approved):
I really wish we could call this a regression and turn the option on by default. Sadly we touted it as intentional:
https://github.com/tinymce/tinymce/blob/5ee5a70c5d7f79c0d1eb1ae7ab5992ae2916584d/modules/tinymce/CHANGELOG.md#L600

Please make a note to change the default in v7 (perhaps even remove the option and just always force it on the way v5 did).

TheSpyder:
Fixes #8657 (I can't seem to link it for some reason)

modules/tinymce/src/core/main/ts/api/OptionTypes.ts
```diff
@@ -300,6 +300,7 @@ export interface EditorOptions extends NormalizedEditorOptions {
   font_size_style_values: string;
   forced_root_block: string;
   forced_root_block_attrs: Record<string, string>;
+  force_hex_color: boolean;
```
ltrouton, line 303: We should also add `force_hex_color?: boolean;` to the BaseOptions interface around line 45 of this file

modules/tinymce/src/themes/silver/test/ts/headless/components/colorpicker/ColorPickerTest.ts
```diff
@@ -162,4 +164,110 @@ describe('headless.tinymce.themes.silver.components.colorpicker.ColorPickerTest'
...
+
+      const hexColor = '#E03E2D';
+      editor.execCommand('mceApplyTextcolor', 'forecolor' as any, hexColor);
+
+      const outerHTML = editor.dom.getOuterHTML(editor.getBody());
+      const dataMceStyle = outerHTML.match(/data-mce-style="[^"]+"/)?.[0];
+      assert.isString(dataMceStyle, `Can't find data-mce-style in "${outerHTML}"`);
+      const expectedColor = 'color: rgb(224, 62, 45);';
+      expect(dataMceStyle).to.include(expectedColor, `"${dataMceStyle}" does not include expected color: "${expectedColor}".`);
+    });
```
ltrouton, line 188: I am wondering if we can do something like this. I am not sure if the selector is exactly right but I believe something like that should work

```suggestion
		TinyAssertions.assertContentPresence(editor, { [`span[data-mce-style="color: ${hexColor}"]`]: 1 });
```

ltrouton (commented):
Nice work 👍

tiny-ben-tran (approved):

modules/tinymce/src/core/main/ts/api/OptionTypes.ts
danoaky-tiny: Added it to BaseEditorOptions as well

modules/tinymce/src/themes/silver/test/ts/headless/components/colorpicker/ColorPickerTest.ts
danoaky-tiny: Oh that's much nicer, just put that in now

modules/tinymce/src/core/main/ts/api/html/Styles.ts
danoaky-tiny: Ahh I see. I've updated it to now use the methods from Transformations instead of adding that regex

modules/tinymce/src/core/main/ts/api/html/Styles.ts
```diff
@@ -263,6 +265,13 @@ const Styles = (settings: StylesSettings = {}, schema?: Schema): Styles => {
               value = value.toLowerCase();
             }
 
+            // Convert RGB colors to HEX
+            if (settings.force_hex_color) {
+              RgbaColour.fromString(value).each((rgba) => {
+                value = Fun.pipe(rgba, RgbaColour.toString, Transformations.rgbaToHexString);
```
spocke, line 271: I don't think you need to use `Fun.pipe` here I think that was mostly created for the `Maybe` that we are not using anyware to make it easier to work with partially applied functions. So if you have functions that looks like this: `const foo = (something) => (a) => ..` then it would look nice if you pipe them like `Fun.pipe(a, foo(something))` instead of `foo(something)(a)` and if you have many of them its even more ugly.
But in this case I don't think this reads all that bad:
```suggestion
                value = Transformations.rgbaToHexString(RgbaColour.toString(rgba));
```

spocke: Another issue is this looks like it's throwing away the alpha channel. I wonder if we should keep those so we only force hex colors if we really can force hex colors. Or should there be a mode to handle that as as well like a less forceful mode? If the alpha channel is fully opaque then we can do this operation safely without information loss.

spocke: There are specific tests for this module in the StylesTest.ts test for this test should be there as well since the ColorPickerTest seems more like a integration test more higher up at UI level. Also add a case for the alpha channels so that it shows clearly that it does throw away that channel and for what cases.

spocke (changes requested):
This would need a changelog item as well as a new feature I guess even though we are just bringing an old one back to existence.

Also not sure why this is under a private fork?

modules/tinymce/src/core/main/ts/api/html/Styles.ts
danoaky-tiny: > I don't think you need to use `Fun.pipe` here I think that was mostly created for the `Maybe` that we are not using anyware to make it easier to work with partially applied functions. So if you have functions that looks like this: `const foo = (something) => (a) => ..` then it would look nice if you pipe them like `Fun.pipe(a, foo(something))` instead of `foo(something)(a)` and if you have many of them its even more ugly. But in this case I don't think this reads all that bad:

Makes sense, I removed the `Fun.pipe`

danoaky-tiny: > Another issue is this looks like it's throwing away the alpha channel. I wonder if we should keep those so we only force hex colors if we really can force hex colors. Or should there be a mode to handle that as as well like a less forceful mode? If the alpha channel is fully opaque then we can do this operation safely without information loss.

So would you maybe prefer a union type for the option, like: `force_hex_color: 'always' | 'rgb_only' | 'off';`?
Or split it into two options like `force_hex_color` & `coerce_hex_color`, or something else?

spocke: Yes I think we should bring this back in a better version that what we had before. When the original option was created that forced hex colors rgba didn't exist, only rgb, as css concept so then there was no need to consider that.

Also not having a relax option could then cause new bug reports of users pasting something then the colors change because we strip the alpha channel. Not that I remember ever seeing that when we had that option so might be rare.

I think it's better to have one single option since they would otherwise be hard to explain what happens if you set one or the other or both etc.

spocke (approved):

spocke:
Title has the wrong ticket ID it should be TINY-9819 that will be used in the squashed commit message.

- [ ] Switch ticket ID to TINY-9819 in title

modules/tinymce/src/core/main/ts/api/html/Styles.ts
danoaky-tiny: Ok, I've made those changes. There's some tests that attempt to cover using RGBA too

TheSpyder: So much for my idea of removing this option in TinyMCE 7 😂 

We need to update this code:
https://github.com/tinymce/tinymce/blob/14da33f3d1e666bea45a4bd392073a355803e4ca/modules/acid/src/main/ts/ephox/acid/api/colour/HexColour.ts#L39

To support hex with alpha, which all supported browsers have had for many years:
https://developer.mozilla.org/en-US/docs/Web/CSS/hex-color

So I guess in v7 the ticket should be "support alpha in the rgb to hex colour code, _then_ delete the force hex colour option"

TheSpyder:
I didn't notice this was done from a private fork, we definitely needed to give you contributor access before you began development that needed to become a PR.

Would it be too annoying to open a new PR? I think we're all happy with the code, it just means if tests fail a new PR would be required. I guess a potential new PR is better than a guaranteed new PR 🤷‍♂️

modules/tinymce/src/core/main/ts/api/html/Styles.ts
spocke: It doesn't work on IE and we don't support IE for tinymce anymore, however do we or should we support IE as an output target. I can't of the top of my head think of any html that we produce tight now that wouldn't work on IE but I guess that will change overtime since we don't test on IE. I guess we could go with hex rgba and see if we get complaints and then make some backwards compatible option since this is an opt in we could document this as a good to know kind of thing like don't enable this if you have an target audience that uses IE.

danoaky-tiny:
Closed because we're swapping to a PR not created from a private fork

modules/tinymce/src/core/main/ts/api/html/Styles.ts
TheSpyder: Ah, I see what you mean. I think for TinyMCE 7 we should be safe saying we don't create IE-compatible HTML anymore. But that's a discussion for another day 😂

PR #9082, opened 2023-10-16
TINY-10147: Improve screen reader output for bespoke dropdown toolbar buttons

mkzhx:
Related Ticket: TINY-10147

Description of Changes:
* Dynamically append visible child text of bespoke dropdown toolbar buttons to its aria-label

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


MitchCTiny (approved):
I see the fix is to announce the name of the button then the current selection. LGTM

modules/tinymce/src/themes/silver/test/ts/browser/editor/bespoke/DropdownAriaLabelTest.ts
```diff
@@ -0,0 +1,113 @@
...
+}
+
+describe('browser.tinymce.themes.silver.editor.bespoke.DropdownAriaLabelTest', () => {
+  const hook = TinyHooks.bddSetup<Editor>({
+    base_url: '/project/tinymce/js/tinymce',
+    content_css: '/project/tinymce/src/themes/silver/test/css/content.css',
+    toolbar: 'align fontfamily fontsize blocks styles'
+  });
+
+  const testDropdownAriaLabel = (scenario: Scenario) => async () => {
```
TheSpyder, line 24: There are a whole lot of `if` checks and branches here that will make it harder to understand which path the code has followed if a test fails.

I think it's been refactored too heavily. Instead of a single function, please make a series of functions (it looks like you'll need 3?) so there are no branches in the code. This will initially be heavily duplicated, which is fine, you can then look at where it makes sense to extract pieces into helper functions.

Some duplication is fine - I've grown fond of saying "duplication is documentation". This doesn't mean avoid refactoring, but strategic duplication will greatly aid readability which is essential to tracking down problems when things fail.

It's particularly important for tests like this where you can't have good error messages. If a menu name is wrong all you get is `waited for element blah to appear, and it didn't`. This doesn't tell you what the incorrect value was - it can't - so the code needs to be easy to follow otherwise failures can require debugging the test.

modules/tinymce/src/themes/silver/main/ts/api/Events.ts
```diff
@@ -39,6 +39,26 @@ const fireToggleToolbarDrawer = (editor: Editor, state: boolean): void => {
...
+};
+
+const fireBlocksTextUpdate = (editor: Editor, data: { value: string }): void => {
+  editor.dispatch('BlocksTextUpdate', data);
+};
+
+const fireFontFamilyTextUpdate = (editor: Editor, data: { value: string }): void => {
+  editor.dispatch('FontFamilyTextUpdate', data);
+};
+
```
TheSpyder, line 61: Does this cover all of the bespokes? We're effectively making an API here, I can imagine some integrators will be happy to see this information broadcast, so we need to make sure it's right the first time.

TheSpyder (approved):

modules/tinymce/src/themes/silver/main/ts/api/Events.ts
shanmen-tiny: Do we need one for `FontSizeInput`? https://github.com/tinymce/tinymce/blob/82f8b3cd850d5ff58f4b5f327f56d47773743a80/modules/tinymce/src/themes/silver/main/ts/ui/toolbar/Integration.ts#L118

shanmen-tiny (approved):

modules/tinymce/src/themes/silver/test/ts/browser/editor/bespoke/DropdownAriaLabelTest.ts
mkzhx: That's a good point, I've kind of had a feeling that having so many branches impacted readability. I've refactored it into 3 functions that pass callback functions to a common testing function which has no branches.

modules/tinymce/src/themes/silver/main/ts/api/Events.ts
mkzhx: Yes `fontsizeinput` looks to be the only bespoke button I've missed, I've added that. I've also added tests for all the bespoke text update events.

shanmen-tiny (approved):

modules/tinymce/src/themes/silver/main/ts/api/Events.ts
```diff
@@ -39,6 +39,30 @@ const fireToggleToolbarDrawer = (editor: Editor, state: boolean): void => {
...
+  editor.dispatch('FontSizeInputTextUpdate', data);
+};
+
+const fireBlocksTextUpdate = (editor: Editor, data: { value: string }): void => {
+  editor.dispatch('BlocksTextUpdate', data);
+};
+
+const fireFontFamilyTextUpdate = (editor: Editor, data: { value: string }): void => {
+  editor.dispatch('FontFamilyTextUpdate', data);
+};
```
TheSpyder, line 64: Please make a docs ticket to describe the new events. They might prove useful to integrators.

@spocke are you happy with these event names?

mkzhx: Good catch, I've created [DOC-2202](https://ephocks.atlassian.net/browse/DOC-2202).

PR #9085, opened 2023-10-17
TINY-10132: All applied inline formats are lost when format is removed.

ArvinJ-H:
Related Ticket: 
https://ephocks.atlassian.net/browse/TINY-10132

Description of Changes:
- Update the filtering logic to include all caret format node
- When there is no content in the editor, `DeleteElement` will ignore caret

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/tinymce/src/core/main/ts/fmt/CaretFormat.ts
```diff
@@ -270,18 +270,21 @@ const removeCaretFormat = (editor: Editor, name: string, vars?: FormatVars, simi
...
-  while (node) {
+  while (node && isFormatElement(editor, SugarElement.fromDom(node))) {
     if (MatchFormat.matchNode(editor, node, name, vars, similar)) {
-      formatNode = node as Element;
-      break;
-    }
-
-    if (node.nextSibling) {
-      hasContentAfter = true;
+      if (matchCount === 0) {
```
TheSpyder, line 277: Since this is the only place it's checked, it should be a boolean not a number 🤨

modules/tinymce/src/core/main/ts/delete/DeleteElement.ts
```diff
@@ -87,7 +88,12 @@ const isBlock = (editor: Editor, elm: SugarElement<Node>): boolean =>
 const paddEmptyBlock = (elm: SugarElement<Node>): Optional<CaretPosition> => {
   if (Empty.isEmpty(elm)) {
     const br = SugarElement.fromHtml('<br data-mce-bogus="1">');
-    Remove.empty(elm);
+    // Remove all bogus elements except caret
+    Arr.each(Traverse.children(elm), (node) => {
+      if (node.dom.textContent !== ZWSP) {
```
TheSpyder, line 93: You did well to get this far without help, but `data-mce-type="format-caret"` or `id="_mce_caret"` is a better way to detect this. It's applied to the outer element in this "temporary style" mode, which is used when applying styles in a scenario where it's impossible to persist those styles to a rendered HTML document.
https://github.com/tinymce/tinymce/blob/14dd42f7015cd5573e1da14a3b725e213e3483d0/modules/tinymce/src/core/main/ts/fmt/ApplyFormat.ts#L344

There's even a method to help detect it:
https://github.com/tinymce/tinymce/blob/9dbb1e7742b0835382abc11aa2edd6c33a71f625/modules/tinymce/src/core/main/ts/fmt/CaretFormat.ts#L381

TheSpyder (changes requested):

modules/tinymce/src/core/main/ts/fmt/CaretFormat.ts
ArvinJ-H: oh... yeah I thought I changed it. Fixing it now

modules/tinymce/src/core/main/ts/fmt/CaretFormat.ts
```diff
@@ -270,18 +270,21 @@ const removeCaretFormat = (editor: Editor, name: string, vars?: FormatVars, simi
...
-      break;
-    }
-
-    if (node.nextSibling) {
-      hasContentAfter = true;
+      if (noMatch) {
+        formatNode = node as Element;
+        noMatch = false;
+      }
+    } else {
```
MitchCTiny, line 281: You could move the `noMatch` condition to before the `MatchFormat.matchNode` call and then it wouldn't run once a match is found.

MitchCTiny (approved):

modules/tinymce/src/core/test/ts/browser/FormatterRemoveTest.ts
```diff
@@ -306,7 +306,8 @@ describe('browser.tinymce.core.FormatterRemoveTest', () => {
     const editor = hook.editor();
     editor.setContent('<p><em><b>abc</b></em></p>');
     editor.formatter.register('format', { inline: 'b' });
-    LegacyUnit.setSelection(editor, 'b', 3, 'b', 3);
+    // Replaced LegacyUnit with TinySelections due to legacy unit producing a different result
```
spocke, line 309: Drop these comments since if this breaks things then we go back and review the PR so see why that was done. The source wouldn't need to mention this. We are going to migrate away from LegacyUnit for other reasons overtime so when that happens this comment would not make any sense anymore.

modules/tinymce/CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Search and replace plugin would incorrectly find matching text inside non-editable root elements. #TINY-10162
 - Search and replace plugin would incorrectly find matching text inside SVG elements. #TINY-10162
 - Removed use of `async` for editor rendering which caused visual blinking when reloading the editor in-place. #TINY-10249
+- Toggling the formatting on the caret at the start of the editor, which initially has multiple styles applied, results in the incorrect removal of other formatting styles. #TINY-10132
```
spocke, line 25:
```suggestion
- Toggling off one format on the caret when multiple formats was toggled on would toggle all of them off. #TINY-10132
```

This doesn't only affect start of the editor. It can be end of a paragraph beginning of a word.

All these have this behavior all of them seems covered by you fix:
```
<p>foo bar|</p>
<p>foo| bar</p>
<p>|</p>
```

modules/tinymce/src/core/test/ts/browser/fmt/CaretFormatTest.ts
```diff
@@ -565,4 +565,80 @@ describe('browser.tinymce.core.fmt.CaretFormatTest', () => {
...
+              }),
+              s.text(str.is('abc'))
+            ]
+          })
+        ]
+      })));
+    TinyAssertions.assertCursor(editor, [ 0, 0, 0, 0, 0 ], 2);
+  });
+
+  it('TINY-10132: Apply and remove multiple format to caret in an empty editor', () => {
```
spocke, line 607: Seems like there are more scenarios here.

spocke (approved):

modules/tinymce/src/core/main/ts/delete/DeleteElement.ts
ArvinJ-H: I used `isCaretNode` rather than `isEmptyCaretFormatElement`, to avoid introducing circular dependency

modules/tinymce/src/core/main/ts/fmt/CaretFormat.ts
ArvinJ-H: nah that won't work when the format are nested like this
```<p><span style="text-decoration: underline;"><span style="color: red; text-decoration: underline;"><br></span></span></p>```
when toggling underline, `noMatch` will be set to false once the first `span` is matched, then the second `underline` will get pushed into `parents`then concat to the new caret.

Comment (MitchCTiny) [reply modules/tinymce/src/core/main/ts/fmt/CaretFormat.ts:281]: I just mean this condition doesn't do anything when noMatch is false, so why bother with the call to matchNode first if you can skip it by including noMatch before it.

Comment (ArvinJ-H) [reply modules/tinymce/src/core/main/ts/fmt/CaretFormat.ts]: gonna split this into a different ticket, accidentally out of scoop

File: modules/tinymce/src/core/main/ts/delete/DeleteElement.ts (hunk 2 above)
Comment (TheSpyder) [reply modules/tinymce/src/core/main/ts/delete/DeleteElement.ts]: Ah I forgot that you made changes to CaretFormat so of course you knew about it.

If it's just importing constant function references, TypeScript can (much to my disgust) deal with circular dependencies.

**Review by TheSpyder — APPROVED** [review]

File: modules/tinymce/src/core/main/ts/delete/DeleteElement.ts [hunk 7]
```diff
@@ -8,6 +8,7 @@ import * as CaretFinder from '../caret/CaretFinder';
 import CaretPosition from '../caret/CaretPosition';
 import * as Empty from '../dom/Empty';
 import * as NodeType from '../dom/NodeType';
+import { isEmptyCaretFormatElement } from '../fmt/FormatUtils';
```
mkzhx, line 11: Scope this import
```suggestion
import * as FormatUtils from '../fmt/FormatUtils';
```

modules/tinymce/src/core/main/ts/delete/InlineFormatDelete.ts
```diff
@@ -8,6 +8,7 @@ import * as NodeType from '../dom/NodeType';
 import * as Parents from '../dom/Parents';
 import * as CaretFormat from '../fmt/CaretFormat';
 import * as FormatContainer from '../fmt/FormatContainer';
+import { isEmptyCaretFormatElement } from '../fmt/FormatUtils';
```
mkzhx, line 11: Scope this import

modules/tinymce/src/core/main/ts/delete/DeleteElement.ts
```diff
@@ -87,7 +88,12 @@ const isBlock = (editor: Editor, elm: SugarElement<Node>): boolean =>
 const paddEmptyBlock = (elm: SugarElement<Node>): Optional<CaretPosition> => {
   if (Empty.isEmpty(elm)) {
     const br = SugarElement.fromHtml('<br data-mce-bogus="1">');
-    Remove.empty(elm);
+    // Remove all bogus elements except caret
+    Arr.each(Traverse.children(elm), (node) => {
+      if (!isEmptyCaretFormatElement(node)) {
+        Remove.remove(node);
+      }
+    });
```
mkzhx, line 96: It might be good to introduce a new parameter for `DeleteElement.deleteElement` called `preserve_empty_caret_format` or something, so you can just call `Remove.empty` for all other cases. Seems to unnecessarily slow all cases down when this is only needed for 1.

mkzhx (approved):

modules/tinymce/src/core/main/ts/fmt/CaretFormat.ts
```diff
@@ -307,15 +282,15 @@ const removeCaretFormat = (editor: Editor, name: string, vars?: FormatVars, simi
...
     const caretContainer = getParentCaretContainer(editor.getBody(), formatNode);
+    const parentsAfter = Type.isNonNullable(caretContainer) ? dom.getParents(formatNode.parentNode, Fun.always, caretContainer) : [];
     const newCaretContainer = createCaretContainer(false).dom;
 
     insertCaretContainerNode(editor, newCaretContainer, caretContainer ?? formatNode);
 
     const cleanedFormatNode = cleanFormatNode(editor, newCaretContainer, formatNode, name, vars, similar);
-    const caretTextNode = insertFormatNodesIntoCaretContainer(parents.concat(cleanedFormatNode.toArray()), newCaretContainer);
-
+    const caretTextNode = insertFormatNodesIntoCaretContainer(parents.concat(cleanedFormatNode.toArray()).concat(parentsAfter), newCaretContainer);
```
MitchC1999, line 291:
```suggestion
    const caretTextNode = insertFormatNodesIntoCaretContainer([
    ...parents,
    ...cleanedFormatNode.toArray(),
    ...parentsAfter ], newCaretContainer);
```
This would be more readable as a spread.

modules/tinymce/src/core/main/ts/fmt/CaretFormat.ts
```diff
@@ -378,8 +353,9 @@ const isFormatElement = (editor: Editor, element: SugarElement<Node>): boolean =
   return Obj.has(inlineElements, SugarNode.name(element)) && !isCaretNode(element.dom) && !NodeType.isBogus(element.dom);
 };
 
-const isEmptyCaretFormatElement = (element: SugarElement<Node>): boolean => {
-  return isCaretNode(element.dom) && isCaretContainerEmpty(element.dom);
+const isFormatCaret = (editor: Editor, element: SugarElement<Node>): boolean => {
```
MitchC1999, line 356: Is this function just to match when the current node contains a caret and is a text node? This name is a little confusing. You could also return the `isCaretNode` match first, then call and match the inline elements second to run faster when not in a caret node.

modules/tinymce/src/core/test/ts/browser/fmt/CaretFormatTest.ts
```diff
@@ -44,6 +44,22 @@ describe('browser.tinymce.core.fmt.CaretFormatTest', () => {
...
   };
 
+  interface Action {
+    type: 'apply' | 'remove';
+    format: string;
+    options: Record<string, any>;
+  }
+
+  const handleCaretFormat = (editor: Editor, actions: Action[]) => {
+    actions.forEach(({ type, format, options }) => {
```
MitchC1999, line 54:
```suggestion
    Arr.each(actions, ({ type, format, options }) => {
```

MitchC1999 (approved):

modules/tinymce/src/core/main/ts/fmt/CaretFormat.ts
ArvinJ-H: nope its the reverse, checking if the element is not a bogus or caret, and is a format element such as`b/i/u/span`

ArvinJ-H: ah i see what u mean, true that should speed thing up a bit

MitchC1999 (approved):

PR #9086, opened 2023-10-17
TINY-10268: manage `list` inside an `li` surrounded by 2 not `li` elements

lorenzo-pomili:
Related Ticket: TINY-10268

Description of Changes:
Since our current model for table doesn't allow us to represent this situation:
```
<li>
  <something>
  <list>
  <something-else>
</li>
```
except managing it as a single `li` with some content ignoring the fact that there is a list in it.

to allow us to manage this case we have to change how the `selectedListRoots` is working no more returning the whole tree of lists until the end but stopping at the first list which is in that particular state

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -27,6 +27,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Search and replace plugin would incorrectly find matching text inside non-editable root elements. #TINY-10162
 - Search and replace plugin would incorrectly find matching text inside SVG elements. #TINY-10162
 - Removed use of `async` for editor rendering which caused visual blinking when reloading the editor in-place. #TINY-10249
+- A `list` inside a `li` surrounded by 2 not `li` elements was just considered a `content` of that `li`. #TINY-10268
 
 ## 6.7.1 - 2023-10-19
 
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -27,7 +27,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Search and replace plugin would incorrectly find matching text inside non-editable root elements. #TINY-10162
 - Search and replace plugin would incorrectly find matching text inside SVG elements. #TINY-10162
 - Removed use of `async` for editor rendering which caused visual blinking when reloading the editor in-place. #TINY-10249
-- A `list` inside a `li` surrounded by 2 not `li` elements was just considered a `content` of that `li`. #TINY-10268
+- List items containing the first and the last element as text and a nested list caused some list operations to fail. #TINY-10268
 
 ## 6.7.1 - 2023-10-19
 
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -27,7 +27,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Search and replace plugin would incorrectly find matching text inside non-editable root elements. #TINY-10162
 - Search and replace plugin would incorrectly find matching text inside SVG elements. #TINY-10162
 - Removed use of `async` for editor rendering which caused visual blinking when reloading the editor in-place. #TINY-10249
-- List items containing the first and the last element as text and a nested list caused some list operations to fail. #TINY-10268
+- List items containing a list element surrounded by non list nodes would cause some list operations to fail. #TINY-10268
 
 ## 6.7.1 - 2023-10-19
 
--- a/modules/tinymce/src/plugins/lists/main/ts/core/Selection.ts
+++ b/modules/tinymce/src/plugins/lists/main/ts/core/Selection.ts
@@ -69,9 +69,9 @@ const getClosestListHost = (editor: Editor, elm: Node): HTMLElement => {
 };
 
 const isListInsideLiWithSiblings = (list: SugarElement<Node>): boolean =>
-  Traverse.prevSibling(list).filter((sibling) => !NodeType.isListItemNode(sibling.dom)).isSome() &&
-  Traverse.nextSibling(list).filter((sibling) => !NodeType.isListItemNode(sibling.dom)).isSome() &&
-  Traverse.parent(list).filter((parent) => NodeType.isListItemNode(parent.dom)).isSome();
+  Traverse.prevSibling(list).exists((sibling) => !NodeType.isListItemNode(sibling.dom))
+  && Traverse.nextSibling(list).exists((sibling) => !NodeType.isListItemNode(sibling.dom))
+  && Traverse.parent(list).exists((parent) => NodeType.isListItemNode(parent.dom));
 
 const findLastParentListNode = (editor: Editor, elm: Element): Optional<HTMLOListElement | HTMLUListElement> => {
   const parentLists = editor.dom.getParents<HTMLOListElement | HTMLUListElement>(elm, 'ol,ul', getClosestListHost(editor, elm));
--- a/modules/tinymce/src/plugins/lists/main/ts/core/Selection.ts
+++ b/modules/tinymce/src/plugins/lists/main/ts/core/Selection.ts
@@ -68,10 +68,11 @@ const getClosestListHost = (editor: Editor, elm: Node): HTMLElement => {
   return parentBlock.getOr(editor.getBody());
 };
 
-const isListInsideLiWithSiblings = (list: SugarElement<Node>): boolean =>
-  Traverse.prevSibling(list).exists((sibling) => !NodeType.isListItemNode(sibling.dom))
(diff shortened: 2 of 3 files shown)


modules/tinymce/CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Search and replace plugin would incorrectly find matching text inside non-editable root elements. #TINY-10162
 - Search and replace plugin would incorrectly find matching text inside SVG elements. #TINY-10162
 - Removed use of `async` for editor rendering which caused visual blinking when reloading the editor in-place. #TINY-10249
+- A `list` inside a `li` surrounded by 2 not `li` elements was just considered a `content` of that `li`. #TINY-10268
```
TheSpyder, line 25:
```suggestion
- List items containing both text and a nested list caused some list operations to fail. #TINY-10268
```

I'm not sure if "fail" is the correct word, but it's close enough I guess?

modules/tinymce/src/plugins/lists/main/ts/core/Selection.ts
```diff
@@ -67,6 +68,11 @@ const getClosestListHost = (editor: Editor, elm: Node): HTMLElement => {
   return parentBlock.getOr(editor.getBody());
 };
 
+const isListInsideLiWithSiblings = (list: SugarElement<Node>): boolean =>
```
TheSpyder, line 71: A couple of points

* There's code somewhere to find a parent list. I suggest using it instead of doing `Traverse.parent` and filtering.
* Once you've found a parent list, could you get the children array and check if the first / last elements are not list items? Would that give the same result? It might clean the code up a bit.

And on a wider point, thinking about the algorithm, doesn't this bug occur if there is one text element above _or_ below? It doesn't need both to break? The code requires both to activate.

TheSpyder (approved):

modules/tinymce/src/plugins/lists/main/ts/core/Selection.ts
```diff
@@ -67,6 +68,11 @@ const getClosestListHost = (editor: Editor, elm: Node): HTMLElement => {
...
+  Traverse.prevSibling(list).filter((sibling) => !NodeType.isListItemNode(sibling.dom)).isSome() &&
```
spocke, line 72: You can reduce this by using `exists` instead of `filter` and `isSome`.

modules/tinymce/src/plugins/lists/main/ts/core/Selection.ts
spocke: Above only would be a normal list so that would not trigger the bug:
```
<li>
  foo
  <ul>
    ..
  </ul>
</li>
```

However below ones does trigger it:
```
<li>
  <ul>
    ..
  </ul>
  foo
</li>
```

modules/tinymce/src/plugins/lists/test/ts/browser/IndentTest.ts
```diff
@@ -522,6 +522,48 @@ describe('browser.tinymce.plugins.lists.IndentTest', () => {
     assert.equal(editor.selection.getNode().nodeName, 'LI');
   });
 
+  it('TINY-10268: a `list` inside a `li` surrounded by 2 not element should allow the user to intend/outdent its `li`', () => {
```
spocke, line 525:
```suggestion
  it('TINY-10268: a `list` inside a `li` surrounded by 2 not element should allow the user to indent/outdent its `li`', () => {
```

modules/tinymce/src/plugins/lists/test/ts/browser/IndentTest.ts
```diff
@@ -522,6 +522,48 @@ describe('browser.tinymce.plugins.lists.IndentTest', () => {
...
+        Dedent me
+        <ul>
+          <li class="to-indent">or try to dedent me 1</li>
+          <li>or try to dedent me 2</li>
+        </ul>
+        <span>abc</span>
+      </li>
+    </ul>`);
+
+    LegacyUnit.setSelection(editor, '.to-indent', 0);
```
spocke, line 538: Avoid adding new tests that use LegacyUnit I want to get rid of that some day. :smile: `TinySelection.setCursor` is the new API.

spocke (approved):

modules/tinymce/src/plugins/lists/main/ts/core/Selection.ts
lorenzo-pomili: done

modules/tinymce/CHANGELOG.md
lorenzo-pomili: @TheSpyder since we are managing a nested list between 2 text, could it be better something like this?
```
- List items containing a nested list between two text caused some list operations to fail. #TINY-10268
```

modules/tinymce/src/plugins/lists/test/ts/browser/IndentTest.ts
lorenzo-pomili: done, sorry :sweat_smile:

modules/tinymce/src/plugins/lists/test/ts/browser/IndentTest.ts
lorenzo-pomili: changed

modules/tinymce/src/plugins/lists/main/ts/core/Selection.ts
lorenzo-pomili: >There's code somewhere to find a parent list. I suggest using it instead of doing Traverse.parent and filtering.
about this, the code here is intended only to check if the parent is a `li`, I refactored it after the suggestion of @spocke now it should be more clear

modules/tinymce/CHANGELOG.md
TheSpyder: Well that was my point, the bug isn't just for between two text items. But if that's all you're going to address in this ticket that seems fine - just seems like a waste to make it a whole new ticket when the fix looks like only one or two lines of code different to what's in this PR.

modules/tinymce/src/plugins/lists/main/ts/core/Selection.ts
TheSpyder: Right, yes. So the code is still wrong, it needs to handle the other case.

TheSpyder: >  the code here is intended only to check if the parent is a li

Which is fine, but once you have the parent, if it is a `li`, then checking if the first or last children are text nodes is a very simple operation rather than doing `Traverse.prevSibling` and `Traverse.nextSibling` first.

lorenzo-pomili: wait we can change the check in something like this:

```
Traverse.parent(list).exists((parent) => {
  return NodeType.isListItemNode(parent.dom)
  && (Traverse.firstChild(parent).exists((firstChild) => !NodeType.isListNode(firstChild.dom))
  || Traverse.lastChild(parent).exists((lastChild) => !NodeType.isListNode(lastChild.dom)));
});
```

and this avoid the crash for this scenario:

```
<ul>
  <li>
    <ul>
      <li>A</li>
      <li>B</li>
    </ul>
    C
  </li>
</ul>
```
but only if you try to indent the `li`s if you try to indent `C` you still have the crash

and it breaks this scenario:

```
<ul>
  <li>
    <ul>
      <li>A</li>
      <li>B</li>
    </ul>
    <p>C</p>
  </li>
</ul>
```

here doing `Indent` and then `Undent` on `B` it make it merge with `A`

the case that we want to manage here is the one with both first and last child of the `li` are not `lists` I refactored the code to fit this condition and I added a test

TheSpyder: That seems like an acceptable compromise. The remaining bugs can be a new ticket.

modules/tinymce/CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Search and replace plugin would incorrectly find matching text inside non-editable root elements. #TINY-10162
 - Search and replace plugin would incorrectly find matching text inside SVG elements. #TINY-10162
 - Removed use of `async` for editor rendering which caused visual blinking when reloading the editor in-place. #TINY-10249
+- List items containing the first and the last element as text and a nested list caused some list operations to fail. #TINY-10268
```
spocke, line 30: Maybe not sure if its better:
```suggestion
- List items containing a list element surrounded by non list nodes would cause some list operations to fail. #TINY-10268
```
Writing changelog items is harder than naming things. :smile:

spocke (approved):

modules/tinymce/CHANGELOG.md

modules/tinymce/CHANGELOG.md
lorenzo-pomili: changed based to this comment: https://github.com/tinymce/tinymce/pull/9086#discussion_r1367003088

PR #9087, opened 2023-10-18
TINY-10263: We now keep getModifierState properly accessible.

HAFRMO:
Related Ticket: https://ephocks.atlassian.net/browse/TINY-10263

Description of Changes:
We did not correctly treat the function `getModifierState` when cloning an event.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/tinymce/src/core/main/ts/events/EventUtils.ts
```diff
@@ -69,6 +70,12 @@ const clone = <T extends PartialEvent>(originalEvent: T, data?: T): T => {
     event.composedPath = () => originalEvent.composedPath!();
   }
 
+  // The composed path can't be cloned, so delegate instead
```
TheSpyder, line 73: Please be careful when copy/pasting code, this comment is now incorrect 😛 

- [ ] change comment

modules/tinymce/src/core/test/ts/browser/keyboard/KeyboardEventTest.ts
```diff
@@ -0,0 +1,26 @@
...
+
+import Editor from 'tinymce/core/api/Editor';
+
+describe('browser.tinymce.core.keyboard.KeyboardEventTest', () => {
+  const hook = TinyHooks.bddSetupLight<Editor>({
+    indent: false,
+    base_url: '/project/tinymce/js/tinymce'
+  }, []);
+
+  it('TINY-10263: getModifierState exists and does not crash the editor', () => {
```
TheSpyder, line 14: Does this actually fail if you remove the fix? I thought the issue was only in certain frameworks like React.

Please delete the test if it can't be made to fail in automated testing, that's an actively harmful test to have.

TheSpyder (approved):

modules/tinymce/src/core/test/ts/browser/keyboard/KeyboardEventTest.ts
HAFRMO: Yes, this does fail if I remove the fix. I checked it before and I checked it again right now.
If you check the issue there's a reproduction that just uses a plain fiddle too. It might have been that it's particularly prominent in react for some reason?

vpyshnenko (approved):

vpyshnenko:
On Safari the `getModifierState` is not available on the event, so the test is failed

HAFRMO:
> On Safari the `getModifierState` is not available on the event, so the test is failed

So it seems. I'm not sure why though, Safari supports the function and it should be generated. I'll look at our event-generation.

modules/tinymce/src/core/test/ts/browser/keyboard/KeyboardEventTest.ts
TheSpyder: makes sense, all good 👍

PR #9096, opened 2023-10-20
TINY-10295: Fix errors related to JS bundling

hamza0867:
Related Ticket: TINY-10295

Description of Changes:
* Fix errors related to JS bundling

Pre-checks:
* [ ] Changelog entry added

Review:
* [ ] Milestone set


modules/tinymce/src/themes/silver/main/ts/api/Options.ts
```diff
@@ -348,6 +348,8 @@ const getSkinUrl = (editor: Editor): string | undefined => {
   }
 };
 
+const getSkinUrlOpt = (editor: Editor): Optional<string> => Optional.from(editor.options.get('skin_url'));
```
TheSpyder, line 351:
```suggestion
const getSkinUrlOpt = wrapOptional(option('skin_url'));
```

Also, following other uses in this file I suggest renaming it to `getSkinUrlOption` but that would break your code if I made it a suggestion.

modules/tinymce/src/themes/silver/main/ts/ui/skin/Loader.ts
TheSpyder, line 1: This async code for loading the editor isn't going to fly when the PR to `develop` is made, we only just finished getting rid of it. If you think carefully, a bundle scenario can and should be able to load instantly without waiting for _anything_.

I think the new `Resource.get` API was implemented incorrectly. It shouldn't be a simple pointer to `Resource.load` tasks. It should return undefined (or I guess null) if the resource isn't cached. This is much harder, given the architecture, but we should take the opportunity this feature presents to make big improvements.

We added `Resource.has` to check if a resource is bundled, calling `get` in a bundle scenario should more or less assume `has` was already checked and therefore it's going to return a value.

modules/tinymce/src/core/main/ts/init/ContentCss.ts
```diff
@@ -2,9 +2,14 @@ import { Arr } from '@ephox/katamari';
 
 import Editor from '../api/Editor';
 import * as Options from '../api/Options';
+import { TinyMCE } from '../api/Tinymce';
+
+declare let tinymce: TinyMCE;
 
 const isContentCssSkinName = (url: string) => /^[a-z0-9\-]+$/i.test(url);
 
+const isBundledCssSkinName = (url: string) => tinymce.Resource.has('content/' + url + '/content.css');
```
TheSpyder, line 11: This `'content/' + name + '/content.css'` concept probably needs a reusable home, it's hard coded 3 times in this PR alone

TheSpyder (approved):

modules/tinymce/src/themes/silver/main/ts/ui/skin/Loader.ts
hamza0867: Done. How do you like it now ?

modules/tinymce/src/core/main/ts/init/ContentCss.ts
hamza0867: Done!

modules/tinymce/src/themes/silver/main/ts/api/Options.ts
hamza0867: Done!

modules/tinymce/Gruntfile.js
```diff
@@ -1,5 +1,5 @@
 /*eslint-env node */
-const { string: PluginString } = require('rollup-plugin-string');
```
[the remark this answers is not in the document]
hamza0867: Done!

PR #9097, opened 2023-10-20
TINY-10271: Backspace/delete can be performed on selected SVG elements in Firefox

spocke:
Related Ticket: TINY-10271

Description of Changes:
* Fixed delete/backspace but also type over inside SVG elements on Firefox.
* Didn't add a changelog since this is part of the overall support for SVG elements.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/keyboard/PreventNoneditableInput.ts
+++ b/modules/tinymce/src/core/main/ts/keyboard/PreventNoneditableInput.ts
@@ -1,0 +1,12 @@
+import Editor from '../api/Editor';
+
+export const setup = (editor: Editor): void => {
+  editor.on('BeforeInput', (e) => {
+    // Normally input is blocked on non-editable elements that have contenteditable="false" however we are also treating
+    // SVG elements as non-editable and Firefox lets you delete parts of the SVG if you manage to select a part. So this
+    // checks for that scenario and prevents the default behaviour.
+    if (!editor.selection.isEditable()) {
+      e.preventDefault();
+    }
+  });
+};
--- a/modules/tinymce/src/core/main/ts/keyboard/KeyboardOverrides.ts
+++ b/modules/tinymce/src/core/main/ts/keyboard/KeyboardOverrides.ts
@@ -11,6 +11,7 @@ import * as EnterKey from './EnterKey';
 import * as HomeEndKeys from './HomeEndKeys';
 import * as InputKeys from './InputKeys';
 import * as PageUpDownKeys from './PageUpDownKeys';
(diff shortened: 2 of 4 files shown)


modules/tinymce/src/core/main/ts/keyboard/PreventNoneditableInput.ts
```diff
@@ -0,0 +1,12 @@
+import Editor from '../api/Editor';
+
+export const setup = (editor: Editor): void => {
+  editor.on('BeforeInput', (e) => {
+    // Normally input is blocked on non-editable elements that have contenteditable="false" however we are also treating
+    // SVG elements as non-editable and Firefox lets you delete parts of the SVG if you manage to select a part. So this
+    // checks for that scenario and prevents the default behaviour.
+    if (!editor.selection.isEditable()) {
+      e.preventDefault();
```
TheSpyder, line 9: could we potentially remove other checks from the code because this now blocks those scenarios?

TheSpyder (approved):

ltrouton (approved):

lorenzo-pomili (approved):

modules/tinymce/src/core/main/ts/keyboard/PreventNoneditableInput.ts
spocke: We might but might be harder to test those things in isolation.

So theoretically we could get rid of some code but that might then make it harder to develop and test those things in isolation if you want the code to be non-editable aware. This just checks for the scenario of when the selection is within a non-editable context but there are other scenarios around non-editables where the caret is in a editable context but say deleting into a non-editable structure this wouldn't be blocked by this.

TheSpyder: I think reducing how much individual code needs to be aware of non-editable is good, so the ways we check for it are centralised. Just look at how many places you've had to update this year to stabilise non-editable sections.

spocke: Yes but bit this change doesn't do that. What would do that is to have a centralized dom model api that Is non editable aware. But at least now we have centralized way of checking if something is editable that is editor centric not dom centric only so that it could base it result on options and schema rules. So slowly we are moving towards better apis here.

TheSpyder: ok. I do want a centralised dom model api, for the record 😂

PR #9098, opened 2023-10-20
TINY-10273: Adjacent SVG block level elements adds extra paragraph

spocke:
Related Ticket: TINY-10273

Description of Changes:
* Removed empty paragraphs created between SVG root elements.
* Didn't add a changelog since we didn't have official SVG support before.

Pre-checks:

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/ForceBlocks.ts
+++ b/modules/tinymce/src/core/main/ts/ForceBlocks.ts
@@ -44,13 +44,7 @@ const shouldRemoveTextNode = (blockElements: SchemaMap, node: Node) => {
     if (node.data.length === 0) {
       return true;
     } else if (/^\s+$/.test(node.data)) {
-      if (!node.nextSibling) {
-        return true;
-      } else if (isBlockElement(blockElements, node.nextSibling)) {
-        return true;
-      } else if (Namespace.isNonHtmlElementRoot(node.nextSibling)) {
-        return true;
-      }
+      return !node.nextSibling || isBlockElement(blockElements, node.nextSibling) || Namespace.isNonHtmlElementRoot(node.nextSibling);
     }
   }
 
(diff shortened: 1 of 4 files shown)


TheSpyder (approved):
There's an issue that just opened around this being a problem for comments, too 🤔

modules/tinymce/src/core/main/ts/ForceBlocks.ts
```diff
@@ -43,8 +43,14 @@ const shouldRemoveTextNode = (blockElements: SchemaMap, node: Node) => {
...
-    } else if (/^\s+$/.test(node.data) && (!node.nextSibling || isBlockElement(blockElements, node.nextSibling))) {
-      return true;
+    } else if (/^\s+$/.test(node.data)) {
+      if (!node.nextSibling) {
+        return true;
+      } else if (isBlockElement(blockElements, node.nextSibling)) {
+        return true;
+      } else if (Namespace.isNonHtmlElementRoot(node.nextSibling)) {
+        return true;
+      }
```
lorenzo-pomili, line 53:
```suggestion
return !node.nextSibling || isBlockElement(blockElements, node.nextSibling) || Namespace.isNonHtmlElementRoot(node.nextSibling);
```

since they are all true and pretty self explanatory couldn't be better to just return the 3 condition in `or`?

lorenzo-pomili (approved):

modules/tinymce/src/core/main/ts/ForceBlocks.ts
spocke: Yeah it's hard to know how and when it makes sense to split that long expression out. But you are right that doesn't look to bad to have that expression there as s a one liner.

spocke:
> There's an issue that just opened around this being a problem for comments, too 🤔

Yeah I saw that but I think I want to make a separate ticket about that since that should go into changelogs/release notes since that affects existing versions.

PR #9104, opened 2023-10-23
TINY-10304: Fixed broken non-editable root advlist test (#9091)

spocke:
Related Ticket: TINY-10304

Description of Changes:
* Backporting fixes to tests causing issues as we merge things in.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/mcagar/src/main/ts/ephox/mcagar/alien/EditorTypes.ts
+++ b/modules/mcagar/src/main/ts/ephox/mcagar/alien/EditorTypes.ts
@@ -47,6 +47,7 @@ export interface Editor {
   setContent: (content: string) => void;
 
   execCommand: (command: string, ui?: boolean, value?: any, args?: any) => boolean;
+  setEditableRoot: (state: boolean) => void;
 
   nodeChanged: () => void;
   focus: () => void;
--- a/modules/mcagar/src/main/ts/ephox/mcagar/alien/EditorTypes.ts
+++ b/modules/mcagar/src/main/ts/ephox/mcagar/alien/EditorTypes.ts
@@ -47,7 +47,7 @@ export interface Editor {
   setContent: (content: string) => void;
 
   execCommand: (command: string, ui?: boolean, value?: any, args?: any) => boolean;
-  setEditableRoot: (state: boolean) => void;
+  setEditableRoot?: (state: boolean) => void; // Introduced in v6.5
 
   nodeChanged: () => void;
   focus: () => void;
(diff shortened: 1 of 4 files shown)


lorenzo-pomili (approved):

modules/mcagar/src/main/ts/ephox/mcagar/api/bdd/TinyState.ts
```diff
@@ -1,14 +1,14 @@
 import { Editor as EditorType } from '../../alien/EditorTypes';
 
 export const withNoneditableRootEditor = <T extends EditorType = EditorType>(editor: T, f: (editor: T) => void): void => {
-  editor.getBody().contentEditable = 'false';
+  editor.setEditableRoot(false);
```
ltrouton, line 4: I have concerns about switching to `setEditableRoot` without having a fallback. Mcagar is used in other repos and I don't think we want a patch version of mcagar to have to rely on a specific version of TinyMCE, v6.5 in this case

modules/mcagar/src/main/ts/ephox/mcagar/alien/EditorTypes.ts
```diff
@@ -47,6 +47,7 @@ export interface Editor {
   setContent: (content: string) => void;
 
   execCommand: (command: string, ui?: boolean, value?: any, args?: any) => boolean;
+  setEditableRoot: (state: boolean) => void;
```
ltrouton, line 50: I believe this should be optional as we don't know for sure a v6 editor actually has it
```suggestion
  setEditableRoot?: (state: boolean) => void;
```

modules/mcagar/src/main/ts/ephox/mcagar/api/bdd/TinyState.ts
spocke: The TinyState API would fail only when being executed on pre v6.5 versions not sure why one would call that function on a version that doesn't have that feature.

modules/mcagar/src/main/ts/ephox/mcagar/alien/EditorTypes.ts
spocke: I can make this optional I guess and document why this is optional and add some form of check. This is a back port but since master will be merged into develop on release I can make the change here in the backport.

modules/mcagar/src/main/ts/ephox/mcagar/api/bdd/TinyState.ts
spocke: I added a check if that API is available if it's not then an error message about requiring at least 6.5 will be presented.

ltrouton (approved):

modules/mcagar/src/main/ts/ephox/mcagar/api/bdd/TinyState.ts
TheSpyder: I'm less concerned about this. We tend to use new mcagar on old TinyMCE, not the other way around.

PR #9105, opened 2023-10-24
TINY-9819: Added force_hex_color option

danoaky-tiny:
Related Ticket: [TINY-9819](https://ephocks.atlassian.net/browse/TINY-9819)

Description of Changes:
Added an editor option called force_hex_color. Defaults to "off".
Setting to "always" will cause applied colors on the "data-mce-style" attribute to be set as HEX values instead of the default RGB (as of a change in v6). This reflects the behavior of applying colors in v5, as every style color was forced to HEX then.
Setting to "rgb_only" will only convert RGB colors, and _not_ RGBA.

Pre-checks:
* [X] Changelog entry added

Review:
* [x] Milestone set

Original [pull request](https://github.com/tinymce/tinymce/pull/9076) was closed because I didn't have contributor status yet.

[TINY-9819]: https://ephocks.atlassian.net/browse/TINY-9819?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/tinymce/src/core/main/ts/api/OptionTypes.ts
+++ b/modules/tinymce/src/core/main/ts/api/OptionTypes.ts
@@ -44,6 +44,8 @@ export interface ToolbarGroup {
 export type ToolbarMode = 'floating' | 'sliding' | 'scrolling' | 'wrap';
 export type ToolbarLocation = 'top' | 'bottom' | 'auto';
 
+export type ForceHexColor = 'always' | 'rgb_only' | 'off';
+
 interface BaseEditorOptions {
   a11y_advanced_options?: boolean;
   add_form_submit_trigger?: boolean;
@@ -117,7 +119,7 @@ interface BaseEditorOptions {
   font_size_style_values?: string;
   font_size_formats?: string;
   font_size_input_default_unit?: string;
-  force_hex_color?: 'always' | 'rgb_only' | 'off';
+  force_hex_color?: ForceHexColor;
   forced_root_block?: string;
   forced_root_block_attrs?: Record<string, string>;
   formats?: Formats;
@@ -301,7 +303,7 @@ export interface EditorOptions extends NormalizedEditorOptions {
   font_size_style_values: string;
   forced_root_block: string;
   forced_root_block_attrs: Record<string, string>;
-  force_hex_color: 'always' | 'rgb_only' | 'off';
+  force_hex_color: ForceHexColor;
   format_noneditable_selector: string;
   height: number | string;
   highlight_on_focus: boolean;
--- a/modules/tinymce/src/core/main/ts/api/OptionTypes.ts
+++ b/modules/tinymce/src/core/main/ts/api/OptionTypes.ts
@@ -117,6 +117,7 @@ interface BaseEditorOptions {
   font_size_style_values?: string;
   font_size_formats?: string;
   font_size_input_default_unit?: string;
+  force_hex_color?: 'always' | 'rgb_only' | 'off';
   forced_root_block?: string;
   forced_root_block_attrs?: Record<string, string>;
   formats?: Formats;
@@ -300,6 +301,7 @@ export interface EditorOptions extends NormalizedEditorOptions {
   font_size_style_values: string;
   forced_root_block: string;
   forced_root_block_attrs: Record<string, string>;
+  force_hex_color: 'always' | 'rgb_only' | 'off';
   format_noneditable_selector: string;
   height: number | string;
   highlight_on_focus: boolean;
--- a/modules/tinymce/src/core/main/ts/api/Options.ts
+++ b/modules/tinymce/src/core/main/ts/api/Options.ts
@@ -5,7 +5,7 @@ import * as Pattern from '../textpatterns/core/Pattern';
 import * as PatternTypes from '../textpatterns/core/PatternTypes';
 import DOMUtils from './dom/DOMUtils';
 import Editor from './Editor';
-import { EditorOptions } from './OptionTypes';
+import { EditorOptions, ForceHexColor } from './OptionTypes';
 import I18n from './util/I18n';
 import Tools from './util/Tools';
 
@@ -829,8 +829,9 @@ const register = (editor: Editor): void => {
 
   registerOption('force_hex_color', {
     processor: (value) => {
-      const valid = Arr.contains([ 'always', 'rgb_only', 'off' ], value);
-      return valid ? { value, valid } : { valid: false, message: 'Must be one of: always, rgb_only, or off.' };
+      const options: ForceHexColor[] = [ 'always', 'rgb_only', 'off' ];
+      const valid = Arr.contains(options, value);
+      return valid ? { value, valid } : { valid: false, message: `Must be one of: ${options.join(', ')}.` };
     },
     default: 'off',
   });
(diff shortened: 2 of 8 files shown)


TheSpyder (approved):

modules/tinymce/src/core/test/ts/browser/html/StylesTest.ts
```diff
@@ -234,4 +237,124 @@ describe('browser.tinymce.core.html.StylesTest', () => {
     assertStyles(styles, 'background:url(javascript:alert(1)', `background: url('javascript:alert(1');`);
     assertStyles(styles, 'background:url(vbscript:alert(1)', `background: url('vbscript:alert(1');`);
   });
+
+  const applyAndAssertForeColor = (editor: Editor, appliedColor: string, expectedColor: string) => {
```
spocke, line 241: I think all these tests should be moved to a new test file for that particular option maybe a `ForceHexColorTest.ts` file. Since these tests do use a full editor so they are more kind of higher up integration tests of testing the option and the feature at an editor level.

However this file should also have tests but specific to that module the so tests that would run for example `Styles({ force_hex_color: 'always' }` and then assert that the parse and serialize functions properly convert the colors. So a few `it` test for each of the option cases. This way if you work on that part of the code you could easily test things in isolation from the editor and more easily debug things as well since the code path would be more shallow.

- [ ] Split the tests and add some module specific ones

modules/tinymce/src/core/test/ts/browser/html/StylesTest.ts
```diff
@@ -234,4 +237,124 @@ describe('browser.tinymce.core.html.StylesTest', () => {
...
+        TinySelections.setSelection(editor, [ 0, 0 ], 0, [ 0, 0 ], 'color me'.length);
+        editor.execCommand('mceApplyTextcolor', 'forecolor' as any, appliedColor);
+        TinyAssertions.assertContentPresence(editor, { [`span[data-mce-style="color: ${expectedColor};"]`]: 1 });
+      },
+      /** Apply color using the 'forecolor' part of the toolbar. */
+      usingToolbar: async () => {
+        TinyUiActions.clickOnToolbar(editor, '[aria-label^="Text color"] > .tox-tbtn + .tox-split-button__chevron');
+        await TinyUiActions.pWaitForUi(editor, '.tox-swatches');
+        TinyUiActions.clickOnUi(editor, `div[data-mce-color="${appliedColor}"]`);
+        TinyAssertions.assertContentPresence(editor, { [`span[data-mce-style="color: ${expectedColor};"]`]: 1 });
```
spocke, line 257: Maybe assert both the presence that is basically the current view state and the content using assertContent that is the serialized output state and that should be the expected forced color as well.

spocke (approved):

modules/tinymce/src/core/test/ts/browser/html/StylesTest.ts
danoaky-tiny: I've moved the Editor UI related tests to their own file in `browser/editor/ForceHexColorTest.ts`. Let me know if there's a better spot to put in.
Also added in some tests for `Styles({ force_hex_color: '...' })`

modules/tinymce/src/core/test/ts/browser/html/StylesTest.ts
danoaky-tiny: Added `assertContent` calls as well

modules/tinymce/src/core/main/ts/api/OptionTypes.ts
```diff
@@ -117,6 +117,7 @@ interface BaseEditorOptions {
   font_size_style_values?: string;
   font_size_formats?: string;
   font_size_input_default_unit?: string;
+  force_hex_color?: 'always' | 'rgb_only' | 'off';
```
ltrouton, line 120: Probably worth having an exported type at the top of the file for these values. e.g. `export type ForceHexColor = 'always' | 'rgb_only' | 'off'`

That way it can be shared and possible used in the Options.ts file as well

modules/tinymce/src/core/main/ts/api/Options.ts
```diff
@@ -827,6 +827,14 @@ const register = (editor: Editor): void => {
     default: ''
   });
 
+  registerOption('force_hex_color', {
+    processor: (value) => {
+      const valid = Arr.contains([ 'always', 'rgb_only', 'off' ], value);
```
ltrouton, line 832: Based on my other comment, we might be able to add a type to the array so we can catch future changes easier

modules/tinymce/src/core/main/ts/api/html/Styles.ts
```diff
@@ -263,6 +265,16 @@ const Styles = (settings: StylesSettings = {}, schema?: Schema): Styles => {
               value = value.toLowerCase();
             }
 
+            // Convert RGB/RGBA colors to HEX
+            if (typeof settings.force_hex_color === 'string' && settings.force_hex_color !== 'off') {
```
ltrouton, line 269: Should probably use the `Type` module from Katamari instead. Do we need the string check actually?
```suggestion
            if (Type.isString(settings.force_hex_color) && settings.force_hex_color !== 'off') {
```

modules/tinymce/src/core/main/ts/api/html/Styles.ts
```diff
@@ -263,6 +265,16 @@ const Styles = (settings: StylesSettings = {}, schema?: Schema): Styles => {
...
+              RgbaColour.fromString(value).each((rgba) => {
+                //  Always convert or only convert if there will be no loss of information from the alpha channel
+                if (settings.force_hex_color === 'always' || rgba.alpha === 1) {
```
ltrouton, line 272: What does `1` mean here? Might be worth making it a variable or adding a comment to avoid it being a magic number

ltrouton (approved):

modules/tinymce/src/core/main/ts/api/html/Styles.ts
TheSpyder: It's not a magic number; alpha is a fractional value where 1 is opaque and 0 is transparent. I don't know it's worth trying to explain this, we don't explain what `rgba.red` `rgba.green` or `rgba.blue` numbers mean (0 to 255). It's just something you need to know when dealing with colours, much like `hue` being a range from 0 to 360.

modules/tinymce/src/core/main/ts/api/html/Styles.ts
TheSpyder: We do because it's optional. If it was made non-optional then the string check could be removed.

TheSpyder (approved):

modules/tinymce/src/core/main/ts/api/OptionTypes.ts
danoaky-tiny: Added a ForceHexColor export to OptionTypes.ts

modules/tinymce/src/core/main/ts/api/Options.ts
danoaky-tiny: The force_hex_color processor now uses the ForceHexColor type

modules/tinymce/src/core/main/ts/api/html/Styles.ts
danoaky-tiny: I swapped it over to use `Type.isString`

PR #9110, opened 2023-10-24
TINY-10254: manage the case when the selection is in multiple table inside a table

lorenzo-pomili:
Related Ticket: TINY-10254

Description of Changes:
The problem was that when we have a situation like this:

```
...
<td>
  <table id="table1">
    <tbody>
      <tr>
        <td>table</td>
        <td>abc</td>
      </tr>
    </tbody>
  </table>
  <p>content</p>
</td>
...
```

selecting `content` and `abc` our [getTableDetailsFromRange](https://github.com/tinymce/tinymce/blob/develop/modules/tinymce/src/core/main/ts/delete/TableDeleteUtils.ts#L23) returns `isSameTable: false` and `isMultiTable: true` because both `content` and `abc` result in a table, but `content` resutls in the parent table and `abc` in `table1` so not the same table.

but doing it when we go to delete the selection we manage it with [this code](https://github.com/tinymce/tinymce/blob/develop/modules/tinymce/src/core/main/ts/delete/TableDeleteAction.ts#L123-L124) but in this case we intercept the `getClosestCell` of the `rng.endContainer` is the parent cell so we delect also `table1`.

To solve it I changed [getTableDetailsFromRange](https://github.com/tinymce/tinymce/blob/develop/modules/tinymce/src/core/main/ts/delete/TableDeleteUtils.ts#L23) checking if they are 2 different tables but at least one of them is the table releated to the `rng.commonAncestorContainer`

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -40,6 +40,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Toggling a list that contains an LI element having another list as its first child would remove the remaining content within that LI element. #TINY-10213
 - List items containing a list element surrounded by non list nodes would cause some list operations to fail. #TINY-10268
 - The `accordion` toggling with the Enter key press would behave incorrectly on Safari. #TINY-10177
+- Deleting a range that includes a part of a `table` and a `text`, both inside a `td` removed the entire content of the `td`. #TINY-10254
 
 ## 6.7.1 - 2023-10-19
 
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -40,7 +40,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Toggling a list that contains an LI element having another list as its first child would remove the remaining content within that LI element. #TINY-10213
 - List items containing a list element surrounded by non list nodes would cause some list operations to fail. #TINY-10268
 - The `accordion` toggling with the Enter key press would behave incorrectly on Safari. #TINY-10177
-- Deleting a range that includes a part of a `table` and a `text`, both inside a `td` removed the entire content of the `td`. #TINY-10254
+- Deleting a range that included both text of a cell and a nested table in that cell removed the entire content of the cell. #TINY-10254
 
 ## 6.7.1 - 2023-10-19
 
--- a/modules/tinymce/src/core/main/ts/delete/TableDeleteUtils.ts
+++ b/modules/tinymce/src/core/main/ts/delete/TableDeleteUtils.ts
@@ -1,4 +1,4 @@
-import { Fun, Optional, Optionals } from '@ephox/katamari';
+import { Optional, Optionals } from '@ephox/katamari';
 import { Compare, SelectorFilter, SugarElement } from '@ephox/sugar';
 
 import * as TableCellSelection from '../selection/TableCellSelection';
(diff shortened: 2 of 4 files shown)


modules/tinymce/CHANGELOG.md
```diff
@@ -30,6 +30,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Search and replace plugin would incorrectly find matching text inside SVG elements. #TINY-10162
 - Removed use of `async` for editor rendering which caused visual blinking when reloading the editor in-place. #TINY-10249
 - List items containing a list element surrounded by non list nodes would cause some list operations to fail. #TINY-10268
+- Deleting a range that includes a part of a `table` and a `text`, both inside a `td` removed the entire content of the `td`. #TINY-10254
```
TheSpyder, line 33: Needs to be more user-focussed and past tense. Also it probably applies to `th` as well so just call it "cell"
```suggestion
- Deleting a range that included both text of a cell and a nested table in that cell removed the entire content of the cell. #TINY-10254
```

modules/tinymce/src/core/main/ts/delete/TableDeleteUtils.ts
```diff
@@ -20,15 +20,29 @@ const isRootFromElement = (root: SugarElement<Node>): IsRootFn =>
 const getTableCells = (table: SugarElement<HTMLTableElement>): SugarElement<HTMLTableCellElement>[] =>
   SelectorFilter.descendants<HTMLTableCellElement>(table, 'td,th');
 
+const areSameTable = (table1: Optional<SugarElement<HTMLTableElement>>, table2: Optional<SugarElement<HTMLTableElement>>): boolean =>
+  Optionals.lift2(table1, table2, Compare.eq).getOr(false);
```
TheSpyder, line 24: You're having to do a lot of dancing around the logic due to the optional values. Please adjust `getTableDetailsFromRange` to make a new branch of logic that only runs if both start and end table are `some`.

This may also allow you to switch from `let` back to `const`, I'm not sure - perhaps make a complete separate branch for this case and have two `return` blocks.

TheSpyder (changes requested):

modules/tinymce/CHANGELOG.md

modules/tinymce/src/core/main/ts/delete/TableDeleteUtils.ts
lorenzo-pomili: @TheSpyder I refactored it making a function that manage the quirks for details and making the specific case for this issue, let me know what do you think about it

TheSpyder: It's certainly improved, but I encourage you to try for something better. I was hinting at something important - optional values are getting in your way 🙂 

Your logic requires start and end to be `some`, so start by unwrapping them. This would be along the lines of `Optionals.lift2(maybeStart, maybeEnd, (start, end) => { // logic here }).getOr(details)`. The `lift2` function is for much more than just comparison.

That will greatly simplify the code and remove the need for this `areSameTable` function completely; you'll be able to just use `Compare.eq` in your code.

TheSpyder (approved):

modules/tinymce/src/core/main/ts/delete/TableDeleteUtils.ts
```diff
@@ -20,24 +20,45 @@ const isRootFromElement = (root: SugarElement<Node>): IsRootFn =>
 const getTableCells = (table: SugarElement<HTMLTableElement>): SugarElement<HTMLTableCellElement>[] =>
   SelectorFilter.descendants<HTMLTableCellElement>(table, 'td,th');
 
+const getTable = (node: Node, isRoot: IsRootFn) => TableCellSelection.getClosestTable(SugarElement.fromDom(node), isRoot);
+
+const selectionInTableWithNestedTable = (details: TableSelectionDetails, rng: Range, isRoot: IsRootFn): TableSelectionDetails => {
+  return Optionals.lift3(details.startTable, details.endTable, getTable(rng.commonAncestorContainer, isRoot), (startTable, endTable, ancestorTable) => {
```
spocke, line 26: Could this be simplified by simply checking if startTable contains endTable or endTable constains startTable or is that just more confusing?

spocke (approved):

modules/tinymce/src/core/main/ts/delete/TableDeleteUtils.ts
```diff
@@ -20,24 +20,45 @@ const isRootFromElement = (root: SugarElement<Node>): IsRootFn =>
...
+    const areTheSameTableAtTheStart = Compare.eq(startTable, endTable);
```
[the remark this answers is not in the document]
lorenzo-pomili: yes, I changed it

modules/tinymce/src/core/main/ts/delete/TableDeleteUtils.ts
lorenzo-pomili: @spocke I changed it but it still seems a very confusing thing (anyway is better that my previous implementation :sweat_smile: ), if I find a better approach I'll change it otherwise I'll merge with this version

lorenzo-pomili: ok, I removed the check on the same table since is implicit using `PredicateExists.descendant` and also moved the other way around the condition in the return to don't use the negation, I think I'll be not able to improve anymore, I'll merge it as soon as the CI finish

TheSpyder: `PredicateExists.descendant` is quite slow. `Compare.contains` is what Spocke was talking about. I wish I'd seen this earlier 😞

PR #9123, opened 2023-10-30
TINY-10289: manage delete/backspace involving a nested list

lorenzo-pomili:
Related Ticket: TINY-10289

Description of Changes:
when we are in a situation like this:
```
<ol>
  <li>
    <span>List 1</span>
  </li>
  <li>
    <h2>Header</h2>
    <ol>
      <li>List 1-1</li>
    </ol>
    <p>Place custor at the start of this line and hit backspace.1</p>
  </li>
</ol>
```
and we try to use delete or backspace to merge the `p` with the `li` we get an error since we went in a case where we are trying to merge the parent `li` inside the child `li`.

to manage this I added a check that in case of `backspace` from the `p` it moves the current selection container inside the `otherLi` which is the nested `li`

in case of `delete` from `li` we just have to return false since that case is well managed from other part of the code

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -31,7 +31,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Search and replace plugin would incorrectly find matching text inside non-editable root elements. #TINY-10162
 - Search and replace plugin would incorrectly find matching text inside SVG elements. #TINY-10162
 - Removed use of `async` for editor rendering which caused visual blinking when reloading the editor in-place. #TINY-10249
-- Merging an exernal `p` inside a `list` via `delete`/`backspace` would incorreclty try to move a parent element inside a child element. #TINY-10289
+- Merging an external `p` inside a `list` via delete or backspace would incorrectly try to move a parent element inside a child element. #TINY-10289
 
 ## 6.7.2 - 2023-10-25
 
--- a/modules/tinymce/CHANGELOG.md
+++ b/modules/tinymce/CHANGELOG.md
@@ -31,6 +31,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Search and replace plugin would incorrectly find matching text inside non-editable root elements. #TINY-10162
 - Search and replace plugin would incorrectly find matching text inside SVG elements. #TINY-10162
 - Removed use of `async` for editor rendering which caused visual blinking when reloading the editor in-place. #TINY-10249
+- Merging an exernal `p` inside a `list` via `delete`/`backspace` would incorreclty try to move a parent element inside a child element. #TINY-10289
 
 ## 6.7.2 - 2023-10-25
 
--- a/modules/tinymce/src/plugins/lists/test/ts/webdriver/DeleteTest.ts
+++ b/modules/tinymce/src/plugins/lists/test/ts/webdriver/DeleteTest.ts
@@ -52,7 +52,7 @@ describe('webdriver.tinymce.plugins.lists.DeleteTest', () => {
 
       editor.setContent(initialContent, { format: 'raw' });
       TinySelections.setCursor(editor, [ 0, 1, 1, 0, 0 ], 'List 1-1'.length);
-      await RealKeys.pSendKeysOn('iframe => body', [ RealKeys.deleteKey() ]);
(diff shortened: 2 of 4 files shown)


modules/tinymce/CHANGELOG.md
```diff
@@ -30,6 +30,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Search and replace plugin would incorrectly find matching text inside non-editable root elements. #TINY-10162
 - Search and replace plugin would incorrectly find matching text inside SVG elements. #TINY-10162
 - Removed use of `async` for editor rendering which caused visual blinking when reloading the editor in-place. #TINY-10249
+- Merging an exernal `p` inside a `list` via `delete`/`backspace` would incorreclty try to move a parent element inside a child element. #TINY-10289
```
spocke, line 33: Apperently / is hard to understand for some readers I always get feed back on that since I use that all the time so use or instead. Fixed some typos as well and unwrapped the ticks around delete and backspace since they are more like words then say variables.

```suggestion
- Merging an external `p` inside a `list` via delete or backspace would incorrectly try to move a parent element inside a child element. #TINY-10289
```

modules/tinymce/src/plugins/lists/test/ts/webdriver/DeleteTest.ts
```diff
@@ -0,0 +1,59 @@
...
+    base_url: '/project/tinymce/js/tinymce'
+  }, [ Plugin ], true);
+
+  context('backspace and delete', () => {
+    it('TINY-10289: backspace from a `p` after an `ol` or delete from the last `li` inside the `ol` should merge the `p` inside the `li`', async () => {
+      const editor = hook.editor();
+      editor.on('keyup keydown', (e) => {
+        // eslint-disable-next-line no-console
+        console.log('e: ', e.isDefaultPrevented());
+      });
```
spocke, line 21: Remove this.

modules/tinymce/src/plugins/lists/test/ts/webdriver/DeleteTest.ts
```diff
@@ -0,0 +1,59 @@
...
+          '</li>' +
+          '<li>' +
+            '<h2>Header</h2>' +
+            '<ol>' +
+              '<li>List 1-1Place custor at the start of this line and hit backspace.</li>' +
+            '</ol>' +
+          '</li>' +
+        '</ol>';
+
+      editor.setContent(initialContent, { format: 'raw' });
```
spocke, line 48: Why raw here? Avoid using raw unless it's really necessary. I'm guessing the span is being removed but then replace that with for example a `strong` or something that doesn't get removed.

modules/tinymce/src/plugins/lists/test/ts/webdriver/DeleteTest.ts
```diff
@@ -0,0 +1,59 @@
...
+      TinySelections.setCursor(editor, [ 0, 1, 2 ], 0);
+      await RealKeys.pSendKeysOn('iframe => body', [ RealKeys.backspace() ]);
+      TinyAssertions.assertContent(editor, expectedContent, { format: 'raw' });
```
spocke, line 51: Asserting on raw might have consequences since that doesn't go though the serializer attribute order can't be guaranteed. However in this case there is no other attributes but still avoid it in general.

spocke (approved):

modules/tinymce/CHANGELOG.md

modules/tinymce/src/plugins/lists/test/ts/webdriver/DeleteTest.ts
lorenzo-pomili: done

modules/tinymce/src/plugins/lists/test/ts/webdriver/DeleteTest.ts
lorenzo-pomili: the `span` wasn't needed so I completely removed it

modules/tinymce/src/plugins/lists/test/ts/webdriver/DeleteTest.ts
lorenzo-pomili: ok, I also removed the `format: 'raw'` from the assertion

hamza0867 (approved):

PR #9134, opened 2023-11-03
TINY-10320: Generate css files to make feature compatible with develop

hamza0867:
Related Ticket: TINY-10320

Description of Changes:
* Generate css files to make feature compatible with develop


modules/tinymce/src/themes/silver/main/ts/ui/skin/Loader.ts
```diff
@@ -53,15 +53,20 @@ const loadShadowDomUiSkins = async (editor: Editor, skinUrl: string): Promise<vo
 };
 
 const loadUrlSkin = async (isInline: boolean, editor: Editor): Promise<void> => {
-  Options.getSkinUrlOption(editor).fold(Fun.noop, (skinUrl) => {
+  Options.getSkinUrlOption(editor).fold(() => {
```
spocke, line 56: Will this mean that content css will be loaded even if the skin url is not set? It used to be a noop?

spocke (approved):

modules/oxide/gulpfile.js
```diff
@@ -89,12 +89,25 @@ gulp.task('less', function() {
...
+    .pipe(rename({ extname: '.min.css' }))
+    .pipe(sourcemaps.write('.'))
+    .pipe(gulp.dest('./build/skins'))
+    .pipe(connect.reload());
+});
+
+//
+// Generate JS
+//
+gulp.task('generateJs', function() {
```
HAFRMO, line 104: Why are we adding this to make it compatible with develop? With this won't we generate the css twice and bundling it all up? Seems like it'd take more time and space.

hamza0867: No we generate them only once

modules/tinymce/src/themes/silver/main/ts/ui/skin/Loader.ts
hamza0867: No it was not used to be a noop. I just forgot to add this part when the use does not specify a skin url. It is just the default behaviour.

PR #9141, opened 2023-11-08
TINY-10316: inserting `li` via `enter` makes it inherit styles

lorenzo-pomili:
Related Ticket: TINY-10316

Description of Changes:
Here the problem was that `InsertLi.insert` was creating a new `li` but ignoring the `styles` of the "source li", to solve this now the `new li` will inherit styles of the `li` inside that we pressed enter to create a new li.

NOTE:
since in the case when we are in a case where the "source li" is between to `li` we split the list and we use a new `li` with inside a new `list` this case need to be managed differently

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/newline/InsertLi.ts
+++ b/modules/tinymce/src/core/main/ts/newline/InsertLi.ts
@@ -1,4 +1,5 @@
-import { Type } from '@ephox/katamari';
+import { Arr, Obj, Type } from '@ephox/katamari';
+import { Css, SugarElement, SugarNode } from '@ephox/sugar';
 
 import Editor from '../api/Editor';
 import * as NodeType from '../dom/NodeType';
@@ -49,8 +50,13 @@ const isFirstOrLastLi = (containerBlock: Element, parentBlock: Element, first: b
   return node === parentBlock;
 };
 
+const getStyles = (elm: HTMLElement): string => Arr.foldl(
+  Obj.mapToArray(Css.getAllRaw(SugarElement.fromDom(elm)), (style, styleName) => `${styleName}: ${style};`),
+  (acc, s) => acc + s,
+  '');
+
 // Inserts a block or br before/after or in the middle of a split list of the LI is empty
-const insert = (editor: Editor, createNewBlock: (name: string) => Element, containerBlock: Element, parentBlock: Element, newBlockName: string): void => {
+const insert = (editor: Editor, createNewBlock: (name: string, styles?: Record<string, string>) => Element, containerBlock: Element, parentBlock: Element, newBlockName: string): void => {
   const dom = editor.dom;
   const rng = editor.selection.getRng();
   const containerParent = containerBlock.parentNode;
@@ -64,6 +70,10 @@ const insert = (editor: Editor, createNewBlock: (name: string) => Element, conta
   }
 
   let newBlock = createNewBlock(newBlockName);
+  const parentBlockStyles = isListItem(parentBlock) ? getStyles(parentBlock) : undefined;
+  if (parentBlockStyles) {
+    newBlock = createNewBlock(newBlockName, { style: parentBlockStyles });
+  }
 
   if (isFirstOrLastLi(containerBlock, parentBlock, true) && isFirstOrLastLi(containerBlock, parentBlock, false)) {
     if (hasParent(containerBlock, 'LI')) {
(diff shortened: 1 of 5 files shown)


modules/tinymce/src/core/main/ts/newline/InsertLi.ts
```diff
@@ -105,8 +115,18 @@ const insert = (editor: Editor, createNewBlock: (name: string) => Element, conta
     const fragment = tmpRng.extractContents();
 
     if (newBlockName === 'LI' && hasFirstChild(fragment, 'LI')) {
+      const previousChildren = Arr.filter(
+        Arr.map(Arr.from(newBlock.children), SugarElement.fromDom),
+        (child) => !SugarNode.isTag('br')(child)
```
spocke, line 120:
```suggestion
        Fun.negate(SugarNode.isTag('br'))
```

modules/tinymce/src/core/main/ts/newline/InsertLi.ts
```diff
@@ -64,6 +70,10 @@ const insert = (editor: Editor, createNewBlock: (name: string) => Element, conta
   }
 
   let newBlock = createNewBlock(newBlockName);
+  const parentBlockStyles = isListItem(parentBlock) ? getStyles(parentBlock) : undefined;
```
spocke, line 73: If the get styles return an empty string if there is no styles or using an object instead of a string then you should be able to do this in one line like this.
```
const newBlock = isListItem(parentBlock) ? createNewBlock(newBlockName, { style: getStyles(parentBlock) }) ? createNewBlock(blockName);
```
But I guess it's a matter of style. I try to avoid using local let statements for mutatin or if branching if possible.

modules/tinymce/src/plugins/lists/test/ts/browser/StyleTest.ts
```diff
@@ -0,0 +1,67 @@
...
+
+describe('browser.tinymce.plugins.lists.StyleTest', () => {
+  const hook = TinyHooks.bddSetup<Editor>({
+    plugins: 'lists',
+    toolbar: false,
+    indent: false,
+    base_url: '/project/tinymce/js/tinymce'
+  }, [ Plugin ], true);
+
+  it('TINY-10316: Create a new `li` from a nexted list should preserve the style', () => {
```
spocke, line 16: Is `nexted` a word? Also since this is altering the core not the list plugin I expect this test to be added in the core under the newline tests or as a separate test file for list new lines. Also it shouldn't have to enable the `lists` plugin in that test. I guess we could still have a test here for the list plugin just to make sure that the enter key works as expected in the list plugin as well.

lorenzo-pomili: sorry it was `nested`, I refactored to have 2 different tests and one time use `lists` and the other without plugins

modules/tinymce/src/core/main/ts/newline/InsertLi.ts
lorenzo-pomili: changed

modules/tinymce/src/core/main/ts/newline/InsertLi.ts
```diff
@@ -105,8 +115,18 @@ const insert = (editor: Editor, createNewBlock: (name: string) => Element, conta
...
     if (newBlockName === 'LI' && hasFirstChild(fragment, 'LI')) {
+      const previousChildren = Arr.filter(
+        Arr.map(Arr.from(newBlock.children), SugarElement.fromDom),
+        (child) => !SugarNode.isTag('br')(child)
+      );
+
       newBlock = fragment.firstChild as HTMLLIElement;
       dom.insertAfter(fragment, containerBlock);
+
+      Arr.each(previousChildren, (child) => newBlock.prepend(child.dom));
```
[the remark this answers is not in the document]
lorenzo-pomili: changed

modules/tinymce/src/core/main/ts/newline/InsertLi.ts
lorenzo-pomili: changed

modules/tinymce/src/core/main/ts/newline/InsertLi.ts
```diff
@@ -105,8 +115,18 @@ const insert = (editor: Editor, createNewBlock: (name: string) => Element, conta
     const fragment = tmpRng.extractContents();
 
     if (newBlockName === 'LI' && hasFirstChild(fragment, 'LI')) {
+      const previousChildren = Arr.filter(
+        Arr.map(Arr.from(newBlock.children), SugarElement.fromDom),
```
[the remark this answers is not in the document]
lorenzo-pomili: removed

vpyshnenko (approved):

Issue #5952, opened 2020-08-07
The caret jumps to first cell of first row when deleting last row/column


fredure:
**What is the current behavior? Describe the bug**
The caret jumps to first cell of first row when deleting last row/column.

**Please provide the steps to reproduce and if possible a minimal demo of the problem via [fiddle.tiny.cloud](https://fiddle.tiny.cloud/) or similar.**
1. Place the caret in the cell of the last deleted row/column;
2. Delete row/column.

Codepen: https://codepen.io/fredure/pen/WNwbMjb

**What is the expected behavior?**
The caret is positioned in the cell of the previous row/column.

**Which versions of TinyMCE, and which browser / OS are affected by this issue? Did this work in previous versions of TinyMCE?**
Version of TinyMCE - 5.2.0 (2020-03-14).
Browser - all.
OS - all.

This didn't work in previous version.

lnewson:
Thanks for the report @fredure! I discussed this with the team and we agree this is unexpected behaviour, so I've logged an internal ticket for us to fix this. We'll let you know once a fix is available.

PR #9173, opened 2023-11-20
TINY-6309: Caret moves to first cell when deleting last row in colgroup table

danoaky-tiny:
Related Ticket: https://ephocks.atlassian.net/browse/TINY-6309

Description of Changes:
Caret/cursor now correctly moves to the directly adjacent cell when a row or column is deleted.
The Jira ticket states that both row and column deletion are affected. But as of right now, only deleting the last row within a table that has a colgroup still has this bug. So this change only puts in a fix for row deletions.
I've also put in tests for column deletion as well.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

#5952


versions.txt
```diff
@@ -7,3 +7,4 @@ mcagar@9.0.0
 alloy@14.0.0
 bridge@4.7.0
 sugar@9.3.0
+snooker@11.1.1
```
spocke, line 10: Does this need to be a minor since this is a bug fix only it could go into a patch.

spocke (approved):

versions.txt
danoaky-tiny: Oh right, is the current version of snooker not already 11.1.0?
https://github.com/tinymce/tinymce/blob/693fef6287304759bc2f3dc31f76e9e74a68d8d1/modules/snooker/package.json#L4
Or am I misunderstanding?

modules/tinymce/src/models/dom/test/ts/browser/table/command/TableDeleteRowTest.ts
```diff
@@ -80,4 +80,43 @@ describe('browser.tinymce.models.dom.table.command.TableDeleteRowTest', () => {
...
+        const originalTBody = tr(4, 3);
+        editor.setContent(createTable(originalTBody.join('')));
+
+        // Delete last row:
+        TinySelections.setCursor(editor, [ ...pathToBody, 3, 1, 0 ], 0);
+        editor.execCommand('mceTableDeleteRow');
+        TinyAssertions.assertCursor(editor, [ ...pathToBody, 2, 1, 0 ], 3);
+        TinyAssertions.assertContent(editor, createTable(originalTBody.slice(0, -1).join('')));
+
+        // Delete first row:
```
ltrouton, line 112: I would probably have deleting the first row and deleting the last row as separate `it` block test cases

modules/tinymce/src/models/dom/test/ts/browser/table/command/TableDeleteRowTest.ts
```diff
@@ -80,4 +80,43 @@ describe('browser.tinymce.models.dom.table.command.TableDeleteRowTest', () => {
...
+  Arr.each([
+    {
+      label: 'no colgroup',
+      pathToBody: [ 0, 0 ],
+      createTable: (tbody: string) => `<table><tbody>${tbody}</tbody></table>`,
+    },
+    {
+      label: 'with colgroup',
+      pathToBody: [ 0, 1 ],
+      createTable: (tbody: string) => `<table><colgroup><col><col><col></colgroup><tbody>${tbody}</tbody></table>`,
```
ltrouton, line 97: Just have to be mindful that the number of `col` elements is fixed here and that could be incorrect from changing `const originalTBody = tr(4, 3);`

ltrouton (approved):

modules/tinymce/src/models/dom/test/ts/browser/table/command/TableDeleteRowTest.ts
danoaky-tiny: I've now split them into separate `it` blocks

modules/tinymce/src/models/dom/test/ts/browser/table/command/TableDeleteRowTest.ts
danoaky-tiny: Good point, I've now refactored it to account for this

spocke:
So this seems to be for 7.0

- [ ] Merge this after we make the 6.8 release

TheSpyder (approved):

versions.txt
danoaky-tiny: Bumped snooker version to 11.2.1 now after the 6.8 release

mkzhx (approved):

MitchC1999 (approved):

Issue #6176, opened 2020-11-03
Table Plugin: table_class_list Property Dosen't Allow the Removal of a Table Class


VonSagan:
Using the table plugin with a table_class_list property defined with the first element being {title: 'None', value: ''} does not allow me to set the class to none once the class is set to some value when using the table properties dialog. It will allow me to select none and save but when I look at the code or open up the properties again it is still the previous value.

To reproduce this go to demo here: http://fiddle.tiny.cloud/gxhaab/1

1. Add a table using table menu, it doesn't matter how many rows or columns.
2. Open the table properties dialog from the table menu.
3. Select a table class other then none and then save.
5. Reopen the table properties dialog and select the none table class and save.
6. Reopen the table properties dialog and you will see it is the previous class you selected and not none.
7. Checking the code also confirms that the table still has the previous class set.

The expected behavior was that the class would now be none after it was selected and saved and that the HTML code would show that the table had no class.

I tested this on TinyMCE 5.5.1 using Firefox 82.0.2 (64-bit) on Windows 10 Version 20H2 64-bit.

I can confirm this worked in TinyMCE 4.9.11

ashfordneil:
Thanks for reporting this, and thanks for the incredibly detailed bug report. We've added this to our internal backlog, and you'll be alerted when a fix is available.

Gappa:
I've ran into this issue as well (TinyMCE 5.10.0) and there is a workaround. Since class switching works just fine, let's make a fake class, that will represent the "no class" option:

``` javascript
table_class_list: [
    	{title: 'None', value: 'fake-no-class'},
    	{title: 'No Borders', value: 'table_no_borders'},
    	{title: 'Red borders', value: 'table_red_borders'},
    	{title: 'Blue borders', value: 'table_blue_borders'},
    	{title: 'Green borders', value: 'table_green_borders'}
]
```

http://fiddle.tiny.cloud/AWhaab/5

kkrvch:
@ashfordneil  I just ran into the issue described in this BR and I think that we can overcome this by removing the rows 96-98 in this file [TableDialog.ts](https://github.com/tinymce/tinymce/blob/develop/modules/tinymce/src/plugins/table/main/ts/ui/TableDialog.ts).

PR #9179, opened 2023-11-22
TINY-6653: Allow removing table classes with an empty class in table_class_list

danoaky-tiny:
Related Ticket: https://ephocks.atlassian.net/browse/TINY-6653

Description of Changes:
Removed an if statement within the table properties dialog submit callback that deleted any classes that were to be applied if they were an empty string. I can't seem to find the original reason for this if statement, but the tests still pass with removing it.
Added a test for this that follows the replication steps the original github issue has.
As well as this I've added a default "Selected..." option to the class select for the table, row and cell properties dialog. This is always added as the first option and will not change the selection's class.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

GitHub issues (if applicable): #6176


modules/tinymce/src/plugins/table/main/ts/ui/TableDialog.ts
```diff
@@ -107,10 +107,6 @@ const onSubmitTableForm = (editor: Editor, tableElm: HTMLTableElement | null | u
 
   api.close();
 
-  if (data.class === '') {
-    delete data.class;
-  }
```
ltrouton, line 112: I am not sure what should happen if the table has a class already on it that is not in the dialog class list. By removing the above code, `<table class="random" ...>` would be removed despite not explicitly changing the class dropdown in the dialog at all to 'None'.

I am pretty sure we have run into this situation before with the cell dialog. We'll have to see what we did there

ltrouton: Maybe we need to look at having a base `Select...` option that the dropdown defaults to if the class attribute on the table does not match any of the predefined options.
e.g. const noMatchClass: Dialog.ListBoxItemSpec[] = [{ text: 'Select...', value: 'select' }];

TheSpyder: @ltrouton table cell dialog deletes the class in that scenario as Dan has done here:
https://fiddle.tiny.cloud/Ky7rRJzEPV/0

Maybe log a new ticket to suggest improving both 🤔

TheSpyder (approved):

modules/tinymce/src/plugins/table/main/ts/ui/TableDialog.ts
mkzhx: @TheSpyder I believe what Luke is saying is that if we add a class not in the `table_class_list` like `class="random"`. Then, we open the Table dialog, make no changes, and then save:
* On 6-stable, the content is not modified as expected;
* On this branch, the `class="random"` is removed.

This is because any class not in the class list would default to the first None value, which would remove the class even though no changes have been made. The suggestion is to make another placeholder value like `Select...` instead of the first value as the fallback when the Table class is not recognised to be in the class list.

ltrouton: I think there might be a misunderstanding of what I was referring to in my comment as I think the new behaviour is causing a potential regression from stable. With the change, it is possible for the class attribute to be removed from a table by opening the table dialog, making no changes then clicking save which doesn't seem right.

I'll demo it to Daniel next week and hopefully that will clear up any confusion

TheSpyder: I do understand that - but what I'm trying to say is this is how the row and cell dialogs already work. Try my fiddle  to confirm.

All we're doing here is applying that behaviour to the table dialog. This is likely expected behaviour based on the bug report where it isn't applying correctly.

If it's undesirable behaviour that seems like something we should discuss for v7?

TheSpyder: > This scenario is likely what those lines of code were guarding against

I suspect what those lines were trying to do was _implement_ this behaviour.

In the row and cell dialogs, there's a system that compares the input to the dialog value when the dialog is saved - so deleting the value results in it being deleted it from the DOM:
https://github.com/tinymce/tinymce/blob/9c486f7c951e55ae34bb548efe897eec705ac04d/modules/tinymce/src/plugins/table/main/ts/ui/CellDialog.ts#L45-L47

Unfortunately the table dialog was implemented differently so instead these lines of code made the empty string case have no effect.

modules/tinymce/CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+- Allow removing table classes via the table props dialog to keep consistent with how the cell and row class selecting works. #TINY-6653
```
spocke, line 10: Can this somehow describe that the actual change is here. Like "It was not possible to remove the class attribute from table elements by selecting empty string items configured by the `table_class_list` option." or something. Since just referring to what the cell and row dialog then you need to know how those work.

spocke (approved):

modules/tinymce/src/plugins/table/main/ts/ui/TableDialog.ts
danoaky-tiny: Implemented changes outlined by Luke in the Jira ticket

modules/tinymce/src/plugins/table/test/ts/browser/settings/TableCellClassListTest.ts
```diff
@@ -35,8 +40,48 @@ describe('browser.tinymce.plugins.table.TableCellClassListTest', () => {
...
+      { title: 'none', value: '' },
+      { title: 'test', value: 'test' }
+    ]);
+    editor.setContent('<table><tbody><tr><td class="something">x</td></tr></tbody></table>');
+    TinySelections.setSelection(editor, [ 0, 0, 0, 0, 0 ], 0, [ 0, 0, 0, 0, 0 ], 1);
+    await TableTestUtils.selectClassViaPropsDialog(editor, 'mceTableCellProps', 'none');
+    TinyAssertions.assertContentPresence(editor, { 'td[class]': 0, 'td': 1 });
+  });
+
+  it('TINY-6653: Selecting "Unchanged" will do nothing', async () => {
```
ltrouton, line 75:
```suggestion
  it('TINY-6653: Selecting "Select..." will do nothing', async () => {
```

modules/tinymce/src/plugins/table/test/ts/browser/settings/TableCellClassListTest.ts
```diff
@@ -35,8 +40,48 @@ describe('browser.tinymce.plugins.table.TableCellClassListTest', () => {
     editor.setContent(tableHtml);
     TinySelections.setSelection(editor, [ 0, 0, 0, 0, 0 ], 0, [ 0, 0, 0, 0, 0 ], 1);
     editor.execCommand('mceTableCellProps');
-    await TableTestUtils.pAssertListBoxValue('Select class', editor, 'Class', 'test');
+    await TableTestUtils.pAssertListBoxValue('Select class', editor, 'Class', 'mce-no-match');
```
ltrouton, line 43: Is there any way we can assert that the list box text is `Select...` as well?

modules/tinymce/src/plugins/table/test/ts/browser/settings/TableCellClassListTest.ts
```diff
@@ -35,8 +40,48 @@ describe('browser.tinymce.plugins.table.TableCellClassListTest', () => {
...
+  });
+
+  it('TINY-6653: Selecting a cell class that has no value should remove a previously applied class', async () => {
+    const editor = hook.editor();
+    editor.options.set('table_cell_class_list', [
+      { title: 'none', value: '' },
+      { title: 'test', value: 'test' }
+    ]);
+    editor.setContent(tableHtml);
+    TinySelections.setSelection(editor, [ 0, 0, 0, 0, 0 ], 0, [ 0, 0, 0, 0, 0 ], 1);
```
ltrouton, line 55: It probably would be good to assert that the `Select...` list box item is selected when there is no match and the `{ title: 'none', value: '' }` is in the `table_cell_class_list`.

Now that I think about it, could we still run into an issue where the 'None' item is selected for a multi-cell selection?

modules/tinymce/src/plugins/table/test/ts/module/test/TableTestUtils.ts
```diff
@@ -239,6 +239,16 @@ const createRow = (cellContents: string[]): SugarElement<HTMLTableRowElement> =>
   return tr;
 };
 
+/** Opens the specified properties dialog, selects the class called `classTitle`, confirms changes and closes dialog. */
```
ltrouton, line 242:
```suggestion
/** Opens the specified properties dialog, selects the class called `classTitle`, confirms changes and submits dialog. */
```

modules/tinymce/src/plugins/table/test/ts/module/test/TableTestUtils.ts
```diff
@@ -239,6 +239,16 @@ const createRow = (cellContents: string[]): SugarElement<HTMLTableRowElement> =>
   return tr;
 };
 
+/** Opens the specified properties dialog, selects the class called `classTitle`, confirms changes and submits dialog. */
```
spocke, line 242: I think we should try to avoid comments like this if we can if the code is not clear enough then break things out as named functions that document the code instead. For example the execCommand and pWaitForDialog could be one function called `pOpenDialogByCommand(editor, cmd)` then the click on select and wait and click on item could be `pSelectListBoxValue(editor, classTitle)` then the close thing could be `pSubmitDialog(editor)` or something.

modules/tinymce/src/plugins/table/main/ts/ui/CellDialog.ts
```diff
@@ -42,7 +42,7 @@ const updateSimpleProps = (modifier: DomModifier, colModifier: DomModifier, data
   if (shouldUpdate('scope')) {
     modifier.setAttrib('scope', data.scope);
   }
-  if (shouldUpdate('class')) {
+  if (shouldUpdate('class') && data.class !== 'mce-no-match') {
```
spocke, line 45: I guess this new value needs to be documented right if you configure a custom class list you might want to have this one as one of the options where it doesn't change the class value.

spocke (approved):

modules/tinymce/src/plugins/table/main/ts/ui/UiUtils.ts
```diff
@@ -69,6 +69,14 @@ const buildListItems = (items: UserListItem[]): Dialog.ListBoxItemSpec[] =>
...
   });
 
+export const buildClassList = (classList: UserListItem[]): Dialog.ListBoxItemSpec[] => {
+  if (!classList.length) {
+    return [];
+  }
+
+  return buildListItems([{ text: 'Select...', value: 'mce-no-match' }, ...classList ]);
+};
+
```
MitchC1999, line 79:
```suggestion
export const buildClassList = (classList: UserListItem[]): Dialog.ListBoxItemSpec[] => 
  Optionals.someIf(classList.length !== 0, buildListItems([{ text: 'Select...', value: 'mce-no-match' }, ...classList ]));

```
`buildClassList` could return an optional, and then where it's used you could use:
```
const classes = UiUtils.buildClassList(Options.getTableClassList(editor));

  if (classes.isSome()) {
    if (data.class) {
      data.class = data.class.replace(/\s*mce\-item\-table\s*/g, '');
    }
  }
```
and 
```
const getClassList = (editor: Editor): Optional<Dialog.ListBoxSpec> =>
  UiUtils.buildClassList(Options.getRowClassList(editor)).map((classes) => ({
      name: 'class',
      type: 'listbox',
      label: 'Class',
      items: classes
    }));
```

MitchC1999 (approved):

modules/tinymce/src/plugins/table/test/ts/browser/settings/TableCellClassListTest.ts
danoaky-tiny: "None" is selected if at least one cell has no class yes

modules/tinymce/src/plugins/table/main/ts/ui/UiUtils.ts
danoaky-tiny: Good point. I refactored `buildClassList` to return an `Optional` instead but didn't use `someIf`, since it would call `buildListItems` regardless of `classList.length` is. Maybe if there was some lazy version of `someIf` or something similar, but kind of seems a bit over engineered for what should just be one if statement to early return.

modules/tinymce/src/plugins/table/test/ts/browser/settings/TableCellClassListTest.ts
danoaky-tiny: I've addressed this now by using `'mce-no-match'` when there's conflicting class values for multi cell/row selections

ltrouton (approved):

shanmen-tiny (approved):

PR #9211, opened 2023-11-30
Consistent use of single quote entity

Jojoshua:
Using &#39 is problematic with certain frameworks. We used to get around this using the following code in v4

setup: function (editor: Editor) {
    editor.on("SaveContent", function (i) {
      i.content = i.content.replace(/&#39/g, "&apos");
    });

However this doesn't work in v6 anymore.

### DIFF
--- a/modules/tinymce/src/core/main/ts/api/html/Entities.ts
+++ b/modules/tinymce/src/core/main/ts/api/html/Entities.ts
@@ -40,7 +40,7 @@ const asciiMap: Record<number, string> = {
 // Raw entities
 const baseEntities: EntitiesMap = {
   '\"': '&quot;', // Needs to be escaped since the YUI compressor would otherwise break the code
-  '\'': '&#39;',
+  '\'': '&apos;',
   '<': '&lt;',
   '>': '&gt;',
   '&': '&amp;',


Jojoshua:
> The `SaveContent` event has a very specific use case. Try `GetContent`, that should work for _any_ scenario where content is retrieved from the editor.

The scenario is that I am triggering save on all the TinyMCe's on the page and I need this replacement to happen at that time before I post it to the server. Is GetContent going to work when save is triggered? If the single quote doesn't get converted to apos, the .net framework thinks it's being attacked.

TheSpyder:
> Is GetContent going to work when save is triggered?

Yes. The event is dispatched every time `getContent()` is called, and the event is able to mutate the `content` as a string.

https://github.com/tinymce/tinymce/blob/bcdea2ad14e3c2cea40743fb48c63bba067ae6d1/modules/tinymce/src/core/main/ts/content/PrePostProcess.ts#L50-L51

spocke:
The reason why don't use apos here is that it's not a valid HTML 4 entity. So at the time of writing that code there where browsers that didn't support the apos entity for example IE. However I failed to find out what exact IE versions doesn't support it.

Here is the entities in the HTML4 spec notice that apos is not there:
https://www.w3.org/TR/REC-html40/sgml/entities.html

However it is in this HTML5 reference:
https://html.spec.whatwg.org/multipage/named-characters.html#named-character-references

Since this is such a central thing and the HTML that the editor generates could in theory be still be displayed on old IE based HTML renderers I'm not confident that we can just change this without having some form of option to configure it back to the old IE legacy mode.

Jojoshua:
I figured it was an IE thing. I will try a flag.

spocke:
Curious what dotnet sanitization framework requires apos over numeric entities? Want to see if I can reproduce the issue and estimate how common this problem is.

Jojoshua:
> Curious what dotnet sanitization framework requires apos over numeric entities? Want to see if I can reproduce the issue and estimate how common this problem is.

In my scenario it is ASP.NET MVC. If you google search something like _tinymce single quote .net_ you can find posts of the issue.

To reproduce just set tinymce option **encoding: 'xml'** then try to post it to the server.

Jojoshua:
@spocke 

Can this be revisited? I am not seeing TinyMCE is supported by IE https://www.tiny.cloud/docs/tinymce/latest/support/#supported-web-browsers

This is still a problem for us for each upgrade. Here are other examples with the same framework issue.
- https://stackoverflow.com/questions/8496573/tinymce-net-error-when-posting-with-a-single-quote-in-the-text
- https://stackoverflow.com/questions/2473972/a-potentially-dangerous-request-form-value-was-detected-but-validateinputfalse
- https://stackoverflow.com/questions/72353172/tinymce-6-replace-text-issue/75068670#75068670

Jojoshua:
Yeah I would think even if this did cause a problem in pre-IE11,those users should be the ones bypassing to use a callback to change their content, not those using supported browsers.

We have been using ```&apos;``` in the tinymce for many years and it did work in IE11. I tried to prove this and show a fiddle on https://fiddle.tiny.cloud/ where you can make Edge render the page in Internet Explorer mode but that site won't even load in IE mode. 

image


The editor won't even show up on https://www.tiny.cloud/ in Internet Explorer mode either (https://www.tiny.cloud/)

image

TheSpyder:
Yes, TinyMCE dropped support for IE11 in version 6. Maybe an older release would still operate in fiddle but it's not something we test for. Good to know you've had success using `apos`.

Jojoshua:
> Yes, TinyMCE dropped support for IE11 in version 6. Maybe an older release would still operate in fiddle but it's not something we test for. Good to know you've had success using `apos`.

Yeah but note that's using apos in having to make this replacement in the tinymce source code. The PR would fix this for good.

PR #9212, opened 2023-11-30
TINY-10312: Fix expand range for block format

vpyshnenko:
Related Ticket: TINY-10312

Description of Changes:
* When applying block format and selection start/end container is a text node, then expand the range to the beginning/end of the container

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


TheSpyder (approved):
Can you please add a test for applying a block format as well? That was the purpose of this change and I'd like to make sure it doesn't regress by accident in the future.

modules/tinymce/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - When deleting the last row in a table, the cursor would jump to the first cell (top left), instead of moving to the next adjacent cell in some cases. #TINY-6309
 - The functions `schema.isWrapper` and `schema.isInline` didn't exclude element names that started with `#` those should not be considered elements. #TINY-10385
+- Heading formatting would be partially applied to the content within the `summary` element when the caret is positioned between words. #TINY-10312
```
spocke, line 12:
```suggestion
- Heading formatting would be partially applied to the content within the `summary` element when the caret was positioned between words. #TINY-10312
```

modules/tinymce/src/core/test/ts/browser/fmt/ExpandRangeTest.ts
```diff
@@ -226,6 +226,21 @@ describe('browser.tinymce.core.fmt.ExpandRangeTest', () => {
       const rng = expandRng(editor, [ 0, 0, 0, 0, 0 ], 0, [ 0, 0, 0, 0, 0 ], 1, blockFormat, false);
       assertRange(editor, rng, [ 0, 0, 0 ], 0, [ 0, 0, 0 ], 1);
     });
+
+    it('TINY-10154: should expand over the nested summary content instead of parent details body element', () => {
```
spocke, line 230: You also talked about table cases etc should we add some more tests here to cover all of them in case we ever refactor that code we want it to work for those cases as well.

spocke (approved):

modules/tinymce/src/core/test/ts/browser/fmt/ExpandRangeTest.ts
vpyshnenko: done

PR #9213, opened 2023-12-04
TINY-10426: Singularize accessible labels of bespoke buttons that append the current value

mkzhx:
Related Ticket: TINY-10426

Description of Changes:
* Singularize labels of `fontfamily`, `fontsize`, `blocks`, and `styles` bespoke toolbar buttons
* Keep menu item text plural
* Remove string concatenation in Tooltip module
* Add translation tests for bespoke button aria labels

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/themes/silver/main/ts/ui/core/color/ColorSwatch.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/ui/core/color/ColorSwatch.ts
@@ -138,11 +138,10 @@ const getToolTipText = (editor: Editor, format: ColorFormat, lastColor: string) 
     return format === 'forecolor' ? 'Text color' : 'Background color';
   }
 
-  const tooltipPrefix = format === 'forecolor' ? 'Text color {0}' : 'Background color {0}';
   const colors = getColors(Options.getColors(editor, format), format, false);
   const colorText = Arr.find(colors, (c) => c.value === lastColor).getOr({ text: '' }).text;
 
-  return editor.translate([ tooltipPrefix, editor.translate(colorText) ]);
+  return editor.translate([ `${format === 'forecolor' ? 'Text color' : 'Background color'} {0}`, colorText ]);
 };
 
 const registerTextColorButton = (editor: Editor, name: string, format: ColorFormat, lastColor: Cell<string>) => {
--- a/modules/tinymce/src/themes/silver/main/ts/ui/core/color/ColorSwatch.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/ui/core/color/ColorSwatch.ts
@@ -138,10 +138,11 @@ const getToolTipText = (editor: Editor, format: ColorFormat, lastColor: string) 
     return format === 'forecolor' ? 'Text color' : 'Background color';
   }
 
+  const tooltipPrefix = format === 'forecolor' ? 'Text color {0}' : 'Background color {0}';
   const colors = getColors(Options.getColors(editor, format), format, false);
   const colorText = Arr.find(colors, (c) => c.value === lastColor).getOr({ text: '' }).text;
 
-  return editor.translate([ `${format === 'forecolor' ? 'Text color' : 'Background color'} {0}`, colorText ]);
+  return editor.translate([ tooltipPrefix, editor.translate(colorText) ]);
 };
 
 const registerTextColorButton = (editor: Editor, name: string, format: ColorFormat, lastColor: Cell<string>) => {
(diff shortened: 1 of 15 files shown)


modules/tinymce/src/themes/silver/main/ts/ui/core/color/ColorSwatch.ts
```diff
@@ -138,11 +138,10 @@ const getToolTipText = (editor: Editor, format: ColorFormat, lastColor: string)
     return format === 'forecolor' ? 'Text color' : 'Background color';
   }
 
-  const tooltipPrefix = format === 'forecolor' ? 'Text color {0}' : 'Background color {0}';
   const colors = getColors(Options.getColors(editor, format), format, false);
   const colorText = Arr.find(colors, (c) => c.value === lastColor).getOr({ text: '' }).text;
 
-  return editor.translate([ tooltipPrefix, editor.translate(colorText) ]);
+  return editor.translate([ `${format === 'forecolor' ? 'Text color' : 'Background color'} {0}`, colorText ]);
 };
```
shanmen-tiny, line 145: The reason for doing this `return editor.translate([ tooltipPrefix, editor.translate(colorText) ]);` was colorText would be translated first before being substituted into the placeholder, changing this would lead to the `Text color` translated, but the color left untranslated.

TheSpyder: And that's probably true of the `Tooltip.getTooltipText()` function as well. It shouldn't hurt anything to have the double-translate call, it's a fairly simple object lookup. I suggest pulling this commit out of the PR.

mkzhx: That's a good catch, reverted the change.

spocke:
This fails CI there are tests that need updated selectors.

mkzhx:
> This fails CI there are tests that need updated selectors.
The local tests were somehow running on a cached build so I thought they would pass. I've mistakenly changed a menu item selector in a test that didn't need to be changed, so I fixed it. The other failures were quite strange as they were happening only on Safari, which doesn't seem to change the language to the one set by `I18n.setCode` in the `before` hook. I changed the structure of the test to use the editor's `setup` function and it now works as expected in all 3 browsers.

shanmen-tiny (approved):

modules/tinymce/src/themes/silver/main/ts/ui/core/complex/BlocksBespoke.ts
```diff
@@ -44,7 +45,7 @@ const getSpec = (editor: Editor): SelectSpec => {
   const dataset = buildBasicSettingsDataset(editor, 'block_formats', Delimiter.SemiColon);
 
   return {
-    tooltip: Tooltip.getTooltipText(editor, title, fallbackFormat),
+    tooltip: Tooltip.makeTooltipText(editor, `${btnTitle} {0}`, fallbackFormat),
```
TheSpyder, line 48: That's not quite what I meant (I would prefer to have an explicit string that is literally `Block {0}` with no string append _at all_) but this is better than it was.

TheSpyder (approved):

modules/tinymce/src/themes/silver/main/ts/ui/core/complex/BlocksBespoke.ts
mkzhx: I feel this string append is necessary to make the code more maintainable. I am trying to make sure that the label is of the format "<name of component> <current value>", and the name of the component is `btnTitle`. If we were to change the name in the future we would have to remember to also update the tooltip manually if we had an explicit string instead.

TheSpyder: I understand, there are pros and cons to every approach, for example this could run into the plural issue for the same reason the 6.8 release did. But at least each case is all in one file now.

mkzhx: Yeah that's a fair point that the tooltip should make sense on its own regardless of any slight changes to the name. I'm happy to make the change by replacing the `btnTitle` usages with a `btnTooltip` variable that is the explicit string. 

I'll then rename the `title` parameter to `tooltip` in the common `createSelectButton` function, since it is only used for that purpose, to remove the string append there as well:
https://github.com/tinymce/tinymce/blob/29a335850b6a0531498a559a380f439e8a1b069e/modules/tinymce/src/themes/silver/main/ts/ui/core/complex/BespokeSelect.ts#L190-L191

mkzhx: Changes applied

TheSpyder (approved):

PR #9249, opened 2023-12-14
TINY-10459: Empty.isEmpty (DomUtils.isEmpty) unification

vpyshnenko:
Related Ticket: TINY-10459

Description of Changes:
* The common logic moved to the one `Empty.isEmptyNode` function.
* Now, `Empty.isEmptyNode` always requires schema as a first argument (to simplify currying)
* Added 4 options, to adjust  the behavior
  * `skipBogus`, default is `true`
  * `includeZwsp`, default is `false`
  * `checkRootAsContent`, default is `false` (includes the target node, not only its children for checking)
  * `isContent`  predicate function to overwrite the meaning of what should be considered a content
* Now, `Empty.isEmpty` is a simple sugar wrapper on `Empty.isEmptyNode` with `{checkRootAsContent: true}` set.
* Now, `DomUtils.isEmpty` calls the `Empty.isEmptyNode` with `{checkRootAsContent: false}` set.

* Changed
  * Previously nodes with just comments were treated as non-empty. Now, comment nodes are skipped (like bogus) from checking.
    

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/tinymce/src/core/main/ts/api/dom/DOMUtils.ts
```diff
@@ -1745,10 +1672,18 @@ const DOMUtils = (doc: Document, settings: Partial<DOMUtilsSettings> = {}): DOMU
...
      * @method isEmpty
-     * @param {Object} elements Optional name/value object with elements that are automatically treated as non-empty elements.
+     * @param {Node} node The node to check if it's empty.
+     * @param {Object} options An optional object with additional options.
+     *
+     * The options object can have the following properties:
+     * - skipBogus: If true, the function will skip over bogus nodes during the check. Optional, default is true.
+     * - includeZwsp: If true, the function will include zero-width space characters in the check. Optional, default is false.
+     * - check_root_as_content: If true, the function will check if the target node (instead of its children) is considered "content". Optional, default is false.
+     * - isContent: An optional function that takes a node as a parameter and returns a boolean indicating whether the node is considered "content".
```
spocke, line 1682: Do we need to expose all these at this time? Do we need that outside the core?
Also this breaks the old api since you can no longer specify the element names record. Also the old code would use that record instead of the set of that the `isNonEmptyElement` is now checking. 
I would suggest having a `isContentElement(el: Element)` or a `isNonEmptyElement(el: Element)` predicate option in the internal structure that would default to using the Empty `isNonEmptyElement` that checks the record string.
I wonder if this wasn't properly covered with tests. I wonder if we could possibly also handle the nbzwsp case with a `isNonEmptyTextNode` predicate internally? Predicates makes things way more flexible but one could always generate predicates from options or have predefined predicates that one can stick in.

- [ ] Restore the old API and see if we can make it compatible with the old elements behaviour

spocke (changes requested):

modules/tinymce/src/core/main/ts/dom/Empty.ts
```diff
@@ -1,85 +1,114 @@
...
+  includeZwsp?: boolean;
+  check_root_as_content?: boolean;
+  isContent?: (node: Node) => boolean;
+}
+
+const defaultOptionValues: AdjustOptions = {
+  skipBogus: true,
+  includeZwsp: false,
+  check_root_as_content: false,
 };
```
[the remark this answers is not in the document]
vpyshnenko: The `isContent` fucntion is also exported from `Empty` module and is used separately, for ex in [TrimNode](Empty.isC). I don't like the idea that the user of the function should always figure out the all needed optional keys.

modules/tinymce/src/core/main/ts/dom/Empty.ts
```diff
@@ -1,85 +1,114 @@
...
-import { isWhitespaceText } from '../text/Whitespace';
+import Schema from '../api/html/Schema';
+import { isWhitespaceText, isZwsp } from '../text/Whitespace';
 import * as NodeType from './NodeType';
 
-const hasWhitespacePreserveParent = (node: Node, rootNode: Node): boolean => {
-  const rootElement = SugarElement.fromDom(rootNode);
-  const startNode = SugarElement.fromDom(node);
-  return SelectorExists.ancestor(startNode, 'pre,code', Fun.curry(Compare.eq, rootElement));
+interface AdjustOptions {
```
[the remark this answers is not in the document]
vpyshnenko: done

modules/tinymce/src/core/main/ts/delete/BlockMergeBoundary.ts
```diff
@@ -1,6 +1,7 @@
 import { Optional, Optionals } from '@ephox/katamari';
 import { Compare, PredicateFind, SugarElement, SugarNode } from '@ephox/sugar';
 
+import Schema from '../api/html/Schema';
```
TheSpyder, line 4: off-topic, but I wonder if we can start using `import type` 🤔

modules/tinymce/src/core/main/ts/dom/Empty.ts
```diff
@@ -1,85 +1,119 @@
...
+  || isNonEmptyElement(node, schema)
+  || isBookmark(node)
+  || isNamedAnchor(node)
+  || isText(node, rootNode, schema, options)
+  || NodeType.isContentEditableFalse(node)
+  || NodeType.isContentEditableTrue(node) && hasNonEditableParent(node);
+};
+
+const isEmptyNode = (schema: Schema, targetNode: Node, options?: IsEmptyOptions): boolean => {
+  options = { ...defaultOptionValues, ...options };
```
TheSpyder, line 65: Please don't assign to function arguments. I'm surprised our linter allows this.

I think this ES6 trick is a nicer way to do it than double-spread
`const { skipBogus = true, includeZwsp = false, checkRootAsContent: false } = options`

Although you'd have to rebuild the `options` object later on, I suppose, but that still seems better.

modules/tinymce/src/core/main/ts/api/dom/DOMUtils.ts
TheSpyder: I suspect the element names is not required now that we use the schema, but yes we need to be careful about breaking this API.

modules/tinymce/src/core/main/ts/dom/Empty.ts
```diff
@@ -1,85 +1,114 @@
...
 import * as NodeType from './NodeType';
 
-const hasWhitespacePreserveParent = (node: Node, rootNode: Node): boolean => {
-  const rootElement = SugarElement.fromDom(rootNode);
-  const startNode = SugarElement.fromDom(node);
-  return SelectorExists.ancestor(startNode, 'pre,code', Fun.curry(Compare.eq, rootElement));
+interface AdjustOptions {
+  skipBogus?: boolean;
+  includeZwsp?: boolean;
+  check_root_as_content?: boolean;
```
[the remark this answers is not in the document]
vpyshnenko: done

modules/tinymce/src/core/main/ts/delete/BlockMergeBoundary.ts
vpyshnenko: I'd like to not import run time code if I don't need it.

modules/tinymce/src/core/main/ts/dom/Empty.ts
vpyshnenko: I keep the default values outside of function since it is used in several places. I will just use another name for the argument. IMO, `options.someKey` is easy to get where the thing comes from.

modules/tinymce/src/core/main/ts/api/dom/DOMUtils.ts
vpyshnenko: I've updated the `DomUtils.isEmpty` function for compatibility with the old API.

modules/tinymce/src/core/main/ts/delete/BlockMergeBoundary.ts
spocke: I remember I tried to use that and Lee told me that since it was a modern TS feature older TS versions wouldn't be able to consume our types. Not sure if that is true or not haven't tested to import type with `type` and compare the .d.ts files to see if anything changed since that is the only thing that could be breaking things for third party.

modules/tinymce/src/core/main/ts/dom/Empty.ts
```diff
@@ -1,85 +1,122 @@
...
-import { isWhitespaceText } from '../text/Whitespace';
+import Schema from '../api/html/Schema';
+import { isWhitespaceText, isZwsp } from '../text/Whitespace';
 import * as NodeType from './NodeType';
 
-const hasWhitespacePreserveParent = (node: Node, rootNode: Node): boolean => {
-  const rootElement = SugarElement.fromDom(rootNode);
-  const startNode = SugarElement.fromDom(node);
-  return SelectorExists.ancestor(startNode, 'pre,code', Fun.curry(Compare.eq, rootElement));
+interface IsEmptyOptions {
```
spocke, line 9: So this gets exposed to the public DOMUtils.isEmpty API. Is that something we want here or do we just want to expose a subset of these things? Before only `includeZwsp` was exposed as a option.
If we expose them then maybe we need to describe that we added them in the changelog or possibly even document them.

spocke (approved):

modules/tinymce/src/core/test/ts/browser/dom/DomUtilsTest.ts
```diff
@@ -721,7 +721,7 @@ describe('browser.tinymce.core.dom.DOMUtilsTest', () => {
     assert.isTrue(DOM.isEmpty(getTestElement()));
 
     DOM.setHTML('test', '<div><!-- comment --></div>');
-    assert.isFalse(DOM.isEmpty(getTestElement()));
+    assert.isTrue(DOM.isEmpty(getTestElement()));
```
lorenzo-pomili, line 724: doesn't this change impact on some behaviors?

lorenzo-pomili (approved):

modules/tinymce/src/core/test/ts/browser/dom/DomUtilsTest.ts
vpyshnenko: yes, it is. that's why it is introduced in the major version.

PR #9256, opened 2023-12-15
TINY-10470: Worked on isolated test runs

spocke:
Related Ticket: TINY-10470

Description of Changes:
* Changed paths for jax so that it passes when you run the tests locally inside that project with say `yarn test`
* Changed the grunt task imports so they pass if you run tests locally inside tinymce using `yarn test`
* Disabled a test for sugar since it flaked on Chrome in manual mode and always fails in headless mode. It checks if the element is invisible by looking at offsetHeight and offsetWidth sometimes it would have that sometimes it wouldn't in firefox it never has that.
* Added `test-manual` to tinymce so that you can run the tests for tinymce only manually like all other things

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/jax/package.json
+++ b/modules/jax/package.json
@@ -12,7 +12,7 @@
     "lint": "eslint --config ../../.eslintrc.json src/**/*.ts",
     "build": "tsc -b",
     "test": "bedrock-auto -b chrome-headless --customRoutes ../tinymce/src/core/test/json/routes.json -d src/test/ts/",
-    "test-manual": "bedrock --customRoutes ../tinymce/src/json/routes.json -d src/test/ts/",
+    "test-manual": "bedrock --customRoutes --customRoutes ../tinymce/src/core/test/json/routes.json -d src/test/ts/",
     "start": "webpack-dev-server --open-page './src/demo/html'",
     "build:demo": "webpack"
   },
--- a/modules/jax/package.json
+++ b/modules/jax/package.json
@@ -11,8 +11,8 @@
     "prepublishOnly": "yarn run lint && yarn run build",
     "lint": "eslint --config ../../.eslintrc.json src/**/*.ts",
     "build": "tsc -b",
-    "test": "bedrock-auto -b chrome-headless --customRoutes src/test/json/routes.json -d src/test/ts/",
-    "test-manual": "bedrock --customRoutes src/test/json/routes.json -d src/test/ts/",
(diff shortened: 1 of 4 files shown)


modules/jax/package.json
```diff
@@ -11,8 +11,8 @@
     "prepublishOnly": "yarn run lint && yarn run build",
     "lint": "eslint --config ../../.eslintrc.json src/**/*.ts",
     "build": "tsc -b",
-    "test": "bedrock-auto -b chrome-headless --customRoutes src/test/json/routes.json -d src/test/ts/",
-    "test-manual": "bedrock --customRoutes src/test/json/routes.json -d src/test/ts/",
+    "test": "bedrock-auto -b chrome-headless --customRoutes ../tinymce/src/core/test/json/routes.json -d src/test/ts/",
+    "test-manual": "bedrock --customRoutes ../tinymce/src/json/routes.json -d src/test/ts/",
```
ltrouton, line 15: I assume they are meant to be the same.

```suggestion
    "test-manual": "bedrock --customRoutes ../tinymce/src/core/test/json/routes.json -d src/test/ts/",
```

modules/sugar/src/test/ts/browser/VisibilityTest.ts
```diff
@@ -19,12 +17,13 @@ UnitTest.test('VisibilityTest', () => {
   Css.set(c, 'display', 'none');
   Assert.eq('', false, Visibility.isVisible(c));
 
-  const s = SugarElement.fromTag('span');
-  Assert.eq('', false, Visibility.isVisible(s));
-
-  Insert.append(SugarBody.body(), s);
-  const expected = PlatformDetection.detect().browser.isFirefox();
-  Assert.eq('', expected, Visibility.isVisible(s)); // tricked you! height and width are zero == hidden
+  // TODO: Disabled since it is flaking on Chrome sometimes it's hidden sometimes its not
```
ltrouton, line 20: Do we have a Notion or JIra we can link this to?

modules/jax/package.json
spocke: Good catch changed this.

modules/sugar/src/test/ts/browser/VisibilityTest.ts
spocke: Added a ticket reference so that we can investigate this further.

ltrouton (approved):

vpyshnenko (approved):

PR #9268, opened 2023-12-21
TINY-10335: Handle immediate sink on fullscreen toggle

vpyshnenko:
Related Ticket: TINY-10335

Description of Changes:
* When the TinyMCE editor operates within a scrollable container, the UI sink – the area displaying editor's interface elements – is constrained by the boundaries of this container. A notable issue arises when toggling to fullscreen mode: while the editor container adopts a 'fixed' position, its UI sink does not, leading to display inconsistencies. This PR addresses the identified issue by dynamically adjusting the UI sink's positioning. When fullscreen mode is activated, the position of the sink is set to 'fixed'.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/plugins/fullscreen/main/ts/core/Actions.ts
+++ b/modules/tinymce/src/plugins/fullscreen/main/ts/core/Actions.ts
@@ -1,5 +1,5 @@
 import { Cell, Fun, Optional, Singleton, Throttler } from '@ephox/katamari';
-import { Css, DomEvent, EventUnbinder, SugarElement, SugarShadowDom, Traverse, WindowVisualViewport } from '@ephox/sugar';
+import { Class, Css, DomEvent, EventUnbinder, SugarElement, SugarShadowDom, Traverse, WindowVisualViewport, Width, SugarNode } from '@ephox/sugar';
 
 import DOMUtils from 'tinymce/core/api/dom/DOMUtils';
 import Editor from 'tinymce/core/api/Editor';
@@ -89,6 +89,9 @@ const toggleFullscreen = (editor: Editor, fullscreenState: Cell<ScrollInfo | nul
   const documentElement = document.documentElement;
   const editorContainer = editor.getContainer();
   const editorContainerS = SugarElement.fromDom(editorContainer);
+  const sinkContainerS = Traverse.nextSibling(editorContainerS)
+    .filter((elm): elm is SugarElement<HTMLElement> => SugarNode.isHTMLElement(elm) && Class.has(elm, 'tox-silver-sink'));
+
   const fullscreenRoot = getFullscreenRoot(editor);
 
   const fullscreenInfo: ScrollInfo | null = fullscreenState.get();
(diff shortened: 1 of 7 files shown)


modules/tinymce/src/plugins/fullscreen/main/ts/core/Actions.ts
```diff
@@ -153,6 +156,9 @@ const toggleFullscreen = (editor: Editor, fullscreenState: Cell<ScrollInfo | nul
     editorContainerStyle.width = editorContainerStyle.height = '';
 
     handleClasses(DOM.addClass);
+    sinkContainerS.each((elm) => {
+      elm.dom.style.position = 'fixed';
```
spocke, line 160:
```suggestion
      Css.set(element, 'position', 'fixed');
```

modules/tinymce/src/plugins/fullscreen/main/ts/core/Actions.ts
```diff
@@ -176,6 +182,10 @@ const toggleFullscreen = (editor: Editor, fullscreenState: Cell<ScrollInfo | nul
     editorContainerStyle.top = fullscreenInfo.containerTop;
     editorContainerStyle.left = fullscreenInfo.containerLeft;
 
+    sinkContainerS.each((elm) => {
+      elm.dom.style.position = 'relative';
```
spocke, line 186:
```suggestion
      Css.set(element, 'position', 'relative');
```

spocke (approved):
I wonder if this should be in the silver theme instead in theory we should be able to listen to fullscreen event state changes and update the sink that way but then the theme would depend on a event that the plugin produces. 🤔

modules/tinymce/src/plugins/fullscreen/main/ts/core/Actions.ts
spocke: I wonder if it would be more safe if this would restore the state to what ever it was before we changed it rather than hard coding it.

- [ ] Also this needs a test could be a simple did the sink get the right state or not if the fullscreen was toggled with the `ui_mode` being set. Or go all advanced on it and measure things on screen.

modules/tinymce/src/plugins/fullscreen/main/ts/core/Actions.ts
```diff
@@ -176,6 +182,10 @@ const toggleFullscreen = (editor: Editor, fullscreenState: Cell<ScrollInfo | nul
     editorContainerStyle.top = fullscreenInfo.containerTop;
     editorContainerStyle.left = fullscreenInfo.containerLeft;
 
+    sinkContainerS.each((elm) => {
+      Css.set(elm, 'position', 'relative');
```
lorenzo-pomili, line 186: could it make sense to store the `position` before line 160, and then put that position back instead of using always `relative`?

lorenzo-pomili (approved):

modules/tinymce/src/plugins/fullscreen/main/ts/core/Actions.ts
vpyshnenko: A simple test was added. The previous sink css position value is preserved within the FullScreenInfo state.

modules/tinymce/src/plugins/fullscreen/main/ts/core/Actions.ts
vpyshnenko: Yes. The previous sink css position value is preserved within the FullScreenInfo state on [line 150](https://github.com/tinymce/tinymce/blob/c0ed9f3c4a50ee519df6c68365b9f3191b53bf91/modules/tinymce/src/plugins/fullscreen/main/ts/core/Actions.ts#L150)

PR #9276, opened 2024-01-02
TINY-10490: direct right click on a `ol`'s `li` now enbale the contextmenu voices as expected also in Firefox

lorenzo-pomili:
Related Ticket: TINY-10490

Description of Changes:
the problem here is that in Firefox we create a bookmark when there is no selection ([check](https://github.com/tinymce/tinymce/blob/develop/modules/tinymce/src/core/main/ts/selection/SelectionBookmark.ts#L34), via [this function](https://github.com/tinymce/tinymce/blob/develop/modules/tinymce/src/core/main/ts/selection/SelectionBookmark.ts#L71-L75)) in [this handler](https://github.com/tinymce/tinymce/blob/develop/modules/tinymce/src/core/main/ts/selection/SelectionRestore.ts#L36-L40).

So what when we open the context menu, it takes the focus so the [getRng](https://github.com/tinymce/tinymce/blob/develop/modules/tinymce/src/core/main/ts/api/dom/Selection.ts#L307) fallbacks on the [initial bookmark](https://github.com/tinymce/tinymce/blob/develop/modules/tinymce/src/core/main/ts/api/dom/Selection.ts#L325-L331) since the editor has not the focus anymore

in this way the on `mouseup` we call a [nodeChanged](https://github.com/tinymce/tinymce/blob/develop/modules/tinymce/src/core/main/ts/NodeChange.ts#L79) that get as a [new node](https://github.com/tinymce/tinymce/blob/develop/modules/tinymce/src/core/main/ts/NodeChange.ts#L100) the bookmark.

As result the selection results to be on the initial bookmark instead of the current one.

To fix this the initial idea was to update the selection on `contextmenu`, but this cause problems when the context menu was openede on an image or a `figure.image`, [here an attempt to fix it](https://github.com/tinymce/tinymce/pull/9276/commits/e370876863f3ad7bd5f8d11da16b5d6cf445bbb4), but after that attempt I discovered that it has also problem if 2 table's cell are selected (I have similar problems replacing `editor.selection.setCursorLocation` with `editor.selection.placeCaretAt`).

So instead of update the selection exlcuding all the exceptions, I tried to update the bookmark using `SelectionBookmark.store(editor)` and it seems to work, I see that is also used for example when [we open the dialogs](https://github.com/tinymce/tinymce/blob/develop/modules/tinymce/src/core/main/ts/api/WindowManager.ts#L103) and this seems to me a very similar case, ~should I entirely remove the browser check and simple do it for all the browsers?~ I decided to remove that check to not have two different behaviours

Before marking as ready for review:
* [x] Merge https://github.com/tinymce/bedrock/pull/138
* [x] update Bedrock with the new resulting version

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/tinymce/src/plugins/lists/test/ts/webdriver/ContextMenuTest.ts
```diff
@@ -0,0 +1,30 @@
...
+
+describe('webdriver.tinymce.plugins.lists.ContextMenuTest', () => {
+  const hook = TinyHooks.bddSetup<Editor>({
+    plugins: 'lists',
+    toolbar: 'numlist bullist',
+    base_url: '/project/tinymce/js/tinymce',
+    contextmenu: 'lists'
+  }, [ Plugin ], true);
+
+  it('TINY-10490: open contextmenu direclty on a li of a ol should have `List properties...` enabled', async () => {
```
spocke, line 17:
```suggestion
  it('TINY-10490: open contextmenu directly on a li of a ol should have `List properties...` enabled', async () => {
```

modules/tinymce/CHANGELOG.md
```diff
@@ -5,6 +5,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html),
 and is generated by [Changie](https://github.com/miniscruff/changie).
 
+## Unreleased
```
spocke, line 8: We shouldn't add this to the changelog now since we now generate the changelogs using changie. So add a new changelog items using that instead of modifying this file directly.

- [ ] Use changie to add a new changelog item

spocke (changes requested):

modules/tinymce/src/plugins/lists/test/ts/webdriver/ContextMenuTest.ts
lorenzo-pomili: done

modules/tinymce/CHANGELOG.md

spocke (approved):

vpyshnenko (approved):

PR #9280, opened 2024-01-04
TINY-10350: Implement iframe sandbox exclusions and default sandbox_iframes to true

mkzhx:
Related Ticket: TINY-10350

Description of Changes:

Core:
* Implement iframe sandbox exclusions via new `sandbox_iframe_exclusions` option. This option holds "safe" domains whose iframes don't need to be sandboxed and prevent common safe embeds like YouTube and Codepen from breaking.
* The default list for `sandbox_iframe_exclusions` will likely need to undergo further discussions to be improved to include more domains that our users often use in iframe embeds. Linked ticket logged as TINY-10535.
* Default `sandbox_iframes` to `true`

Polaris:
* Implement new `extractUrlHost` function under Strings API. Uses regex to capture the host. The regex is adapted from the existing Regexes.link regex.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/polaris/src/test/ts/atomic/api/UrlTest.ts
```diff
@@ -0,0 +1,73 @@
...
+      'p72internal://foo.com': 'foo.com',
+      'but15characters://foo.com': 'foo.com'
+    };
+
+    it('TINY-10350: Should correctly extract host name from urls', () => {
+      Obj.each(hostMap, (v, k) => {
+        const host = Url.extractHost(k).getOrDie();
+        Assert.eq(`expected host to be ${v} but was ${host}`, v, host);
+      });
+    });
```
MitchC1999, line 71:
```suggestion
    Obj.each(hostMap, (v, k) => {
      it(`TINY-10350: Should correctly extract ${v} from ${k}`, () => {
        const host = Url.extractHost(k).getOrDie();
        Assert.eq(`expected host to be ${v} but was ${host}`, v, host);
      });
    });
```
Each test case should be its own test. This will allow us to tell what is failing.

MitchC1999 (approved):

modules/tinymce/src/core/main/ts/content/PrePostProcess.ts
```diff
@@ -30,6 +30,12 @@ const withSerializedContent = <R extends EditorEvent<{ content: string }>>(conte
   }
 };
 
+const makeParserSettings = (editor: Editor): Partial<DomParserSettings> => ({
```
spocke, line 33: I don't think we need to use partial here since the interface it already partial since all options are optional.

modules/tinymce/src/core/main/ts/html/Embed.ts
```diff
@@ -0,0 +1,52 @@
...
+    name = 'iframe';
+  } else if (isMimeType(type, 'image')) {
+    name = 'img';
+  } else if (isMimeType(type, 'video')) {
+    name = 'video';
+  } else if (isMimeType(type, 'audio')) {
+    name = 'audio';
+  } else {
+    name = 'iframe';
+  }
```
spocke, line 33: Could we extract this to a function and not use mutation:

```
const typeToName = (type: string): 'iframe' | 'img' | 'video' | 'audio' => {
  if (Type.isUndefined(type)) {
    return 'iframe';
  } else if (isMimeType(type, 'image')) {
    return 'img';
  } else if (isMimeType(type, 'video')) {
    return 'video';
  } else if (isMimeType(type, 'audio')) {
    return 'audio';
  } else {
    return 'iframe';
  }
};
```

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
```diff
@@ -71,6 +71,7 @@ export interface DomParserSettings {
   remove_trailing_brs?: boolean;
   root_name?: string;
   sandbox_iframes?: boolean;
+  sandbox_iframes_exclusions?: Record<string, {}>;
```
spocke, line 74: I think it's better to use an array here and the micro optimize this later maybe use a set instead.

modules/tinymce/src/core/main/ts/html/ParserFilters.ts
```diff
@@ -184,21 +155,24 @@ const register = (parser: DomParser, settings: DomParserSettings): void => {
 
   registerBase64ImageFilter(parser, settings);
 
+  const shouldSandboxIframes = settings.sandbox_iframes ?? false;
+  const sandboxIframesExclusions = settings.sandbox_iframes_exclusions ?? {};
```
spocke, line 159: So this would be the spot where I would convert the array of items into a `Set` and then use that to check for host names.

modules/tinymce/src/plugins/media/test/ts/browser/PlaceholderTest.ts
```diff
@@ -12,7 +12,8 @@ describe('browser.tinymce.plugins.media.core.PlaceholderTest', () => {
   const baseSettings = {
     plugins: [ 'media' ],
     toolbar: 'media',
-    extended_valid_elements: 'script[src|type]'
+    extended_valid_elements: 'script[src|type]',
+    sandbox_iframes_exclusions: []
```
spocke, line 16: Should we also have a dedicated test that checks all the hard coded whitelisted items?

spocke (approved):

TheSpyder:
@MitchC1999 our "current and previous" policy means any bugs on Safari 15 or older are not something we're going to fix. If our testing nodes are out of date, we fix the nodes not the tests.

modules/tinymce/src/plugins/media/test/ts/browser/PlaceholderTest.ts
MitchC1999: That will be done as part of [TINY-10535](https://ephocks.atlassian.net/browse/TINY-10535)

PR #9297, opened 2024-01-12
TINY-10454: remove deprecated `remove_trailing_brs` option

lorenzo-pomili:
Related Ticket: TINY-10454

Description of Changes:
removed `remove_trailing_brs` option

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set
* [x] Docs ticket created (DOC-2239)


TheSpyder (changes requested):
This is incorrect.

All this ticket was specified to do was remove the deprecated `DomParser` setting which no longer has any effect. The config setting and `DomSerializer` implementation must remain.

modules/tinymce/src/core/main/ts/html/ParserFilters.ts
```diff
@@ -65,10 +64,6 @@ const createSafeEmbed = (mime?: string, src?: string, width?: string, height?: s
 const register = (parser: DomParser, settings: DomParserSettings): void => {
   const schema = parser.schema;
 
-  if (settings.remove_trailing_brs) {
```
spocke, line 68: If we are just removing the `DomParserSetting` but keeping the editor option then this code needs to be moved somewhere else. Maybe move it to the `InitContentBody` since that is the only other spot where filters are added based on editor options. Then move the tests to something that acts on editors instead of the DOMParser.

spocke: No wait it also registered in the `DomSerializerFilters` so we could just remove these lines and. move the test to the dom serializer.

spocke: It seems that there are already tests for it it the `browser.tinymce.core.dom.SerializerTest` test but we need to make sure that it covers all the cases that the removed test does.

lorenzo-pomili: I moved the tests from `modules/tinymce/src/core/main/ts/html/ParserFilters.ts` to `modules/tinymce/src/core/test/ts/browser/dom/SerializerTest.ts`

hamza0867 (approved):

.changes/unreleased/tinymce-TINY-10454-2024-01-12.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Removed
+body: remove deprecated `remove_trailing_brs` option
```
spocke, line 3:
```suggestion
body: Removed the deprecated `remove_trailing_brs` option from DomParser.
```

modules/tinymce/src/core/test/ts/browser/dom/SerializerTest.ts
```diff
@@ -922,5 +922,65 @@ describe('browser.tinymce.core.dom.SerializerTest', () => {
       '',
       'Should remove br');
   });
+
+  it('Remove redundant br elements', () => {
```
spocke, line 926: It seems that these are the same tests that is already in this file? Could you just go though them to double check that they are all the same. I'm guessing this was copied over when that deprecation PR was made or something but if we made changes or added tests to the DomParser tests since then we might be missing things.

- [ ] Remove duplicate tests

spocke (approved):

HAFRMO (approved):

.changes/unreleased/tinymce-TINY-10454-2024-01-12.yaml
lorenzo-pomili: done

modules/tinymce/src/core/test/ts/browser/dom/SerializerTest.ts
lorenzo-pomili: they were all duplicated :sweat_smile:

TheSpyder (approved):
Much better 👍

PR #9312, opened 2024-01-21
TINY-10453: Remove `title` attribute for menu items, add `data-mce-label` for toolbar and dialog buttons

shanmen-tiny:
Related Ticket: TINY-10453

Description of Changes:
* Attribute changes to buttons are made in this commit: https://github.com/tinymce/tinymce/commit/37c5adf55ecd972e0bc975ebc5f0c85b6843499f
	* Added `data-mce-label` for toolbar and dialog buttons
	* Removed `title` attribute for buttons that have visible label, see notion link for more detail
	* `renderCommonDropdown` now takes `tooltip` and `ariaLabel` properties, `listBox` requires the `aria-label` attribute but not the tooltip
* Tests changes for attribute changes (removal of title attribute and `data-mce-label` are made in this commit: https://github.com/tinymce/tinymce/commit/49ee66a1af030f33d9e4e54ebd5bf6e8082cfb41
* Tootip tests for different buttons: https://github.com/tinymce/tinymce/commit/83c10ed0fe42a4382e135cbe1aeacf88c5aedb9e
* Remove the `title` attribute from directory and tree item: https://github.com/tinymce/tinymce/pull/9312/commits/98e9eb383072dee91c8a5633342291a828792c46
* Removed the `title` attribute for `ViewButton` as it already has a visible label. Also removed title attribute when dynamically updating the `aria-label` for `MenuItem`  when state changes . https://github.com/tinymce/tinymce/pull/9312/commits/77a0f1a94410d1b72cfc227b01f922597c72d38c

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/plugins/table/test/ts/module/test/TableModifiersTestUtils.ts
+++ b/modules/tinymce/src/plugins/table/test/ts/module/test/TableModifiersTestUtils.ts
@@ -96,7 +96,7 @@ const closeMenu = (editor: Editor): void => {
 };
 
 const clickOnButton = (editor: Editor, title: string): void => {
-  TinyUiActions.clickOnToolbar(editor, `button[data-mce-label="${title}"]`);
+  TinyUiActions.clickOnToolbar(editor, `button[data-mce-name="${title}"]`);
 };
 
 const pClickOnMenuItem = async (editor: Editor, title: string): Promise<void> => {
--- a/modules/tinymce/src/plugins/table/test/ts/module/test/TableModifiersTestUtils.ts
+++ b/modules/tinymce/src/plugins/table/test/ts/module/test/TableModifiersTestUtils.ts
@@ -96,7 +96,7 @@ const closeMenu = (editor: Editor): void => {
 };
 
 const clickOnButton = (editor: Editor, title: string): void => {
-  TinyUiActions.clickOnToolbar(editor, `button[data-mce-btn="${title}"]`);
+  TinyUiActions.clickOnToolbar(editor, `button[data-mce-label="${title}"]`);
 };
 
 const pClickOnMenuItem = async (editor: Editor, title: string): Promise<void> => {
--- a/modules/tinymce/src/plugins/table/test/ts/module/test/TableModifiersTestUtils.ts
+++ b/modules/tinymce/src/plugins/table/test/ts/module/test/TableModifiersTestUtils.ts
@@ -14,6 +14,16 @@ interface AssertStyleOptions {
   readonly customStyle: string;
 }
 
+const menuTitleToToolbarBarMap: Record<string, string> = {
+  'Border style': 'tablecellborderstyle',
+  'Border width': 'tablecellborderwidth',
+  'Vertical align': 'tablecellvalign',
+  'Background color': 'tablecellbackgroundcolor',
+  'Border color': 'tablecellbordercolor',
+  'Table styles': 'tableclass',
+  'Cell styles': 'tablecellclass',
+};
+
 interface AssertStyleOptionsWithCheckmarks extends AssertStyleOptions {
   readonly checkMarkEntries: number;
 }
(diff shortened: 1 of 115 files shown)


modules/tinymce/src/themes/silver/main/ts/ui/dialog/SizeInput.ts
```diff
@@ -34,8 +34,7 @@ export const renderSizeInput = (spec: SizeInputSpec, providersBackstage: UiFacto
       classes: [ 'tox-lock', 'tox-button', 'tox-button--naked', 'tox-button--icon' ],
       attributes: {
         'aria-label': translatedLabel,
-        // TINY-10453: Remove this tooltip, we don't want duplicate tooltips and until we figured a better way to test, it's here now so that tests would pass
-        'title': translatedLabel
+        'data-mce-btn': spec.label.getOr('Constrain proportions')
```
ltrouton, line 37: Probably worth having a label variable like `const label = spec.label.getOr('Constrain proportions')`; to avoid the duplication

ltrouton (approved):

spocke:
Could the property be called `data-mce-title` or `data-mce-label` since `btn` kind of makes it specific to just buttons a custom label could be on other things as well like checkboxes, input fields and thing that can have focus.

modules/tinymce/src/themes/silver/main/ts/ui/core/complex/BespokeSelect.ts
```diff
@@ -179,14 +179,15 @@ const createMenuItems = (backstage: UiFactoryBackstage, spec: SelectSpec): Bespo
   };
 };
 
-const createSelectButton = (editor: Editor, backstage: UiFactoryBackstage, spec: SelectSpec, tooltipWithPlaceholder: string, textUpdateEventName: string): SketchSpec => {
+const createSelectButton = (editor: Editor, backstage: UiFactoryBackstage, spec: SelectSpec, tooltipWithPlaceholder: string, textUpdateEventName: string, btnName?: string): SketchSpec => {
```
MitchC1999, line 182: Seeing as no bespoke buttons can be made via public API, and we are now always passing in the btnName, should the btnName not be optional?

shanmen-tiny:
> Could the property be called `data-mce-title` or `data-mce-label` since `btn` kind of makes it specific to just buttons a custom label could be on other things as well like checkboxes, input fields and thing that can have focus.

Yeah not a bad idea, I'm inclined towards `data-mce-label` so I've just renamed that. Hopefully, this attribute will also be used in other places in the future.

modules/tinymce/src/themes/silver/main/ts/ui/dialog/SizeInput.ts
shanmen-tiny: Fixed 👍  https://github.com/tinymce/tinymce/pull/9312/commits/61bd86bad49010bf1144e64bf6b967b19e5ed11b

TheSpyder:
`label` makes it sound like a text label, though. This is the codename used to register the button in code.

Perhaps `data-mce-name`?

modules/tinymce/src/plugins/table/test/ts/module/test/TableModifiersTestUtils.ts
```diff
@@ -103,10 +113,11 @@ const pAssertMenuPresence = async (
   container: SugarElement<HTMLElement>,
   useMenuOrToolbar: 'toolbar' | 'menuitem'
 ): Promise<void> => {
+  const title = useMenuOrToolbar === 'toolbar' ? menuTitleToToolbarBarMap[menuTitle] : menuTitle;
```
TheSpyder, line 116: This is the sort of reason why I wanted to apply the `data-mce-name` concept everywhere and never use `aria-label` in tests 🙃 it would make the tests so much easier and more consistent

But there isn't scope for that now.

TheSpyder (approved):

shanmen-tiny:
> `label` makes it sound like a text label, though. This is the codename used to register the button in code.
> 
> Perhaps `data-mce-name`?
@TheSpyder Renamed to `data-mce-name`. Tagging @spocke, tick if you're happy with that as well. 👍

spocke:
> `label` makes it sound like a text label, though. This is the codename used to register the button in code.
> 
> Perhaps `data-mce-name`?

Right, I thought it was just a renamed title to not show it. Looks like it's a button name of what the control id is so yeah then name is more correct. But I wonder if this will now be abused to custom style individual buttons if all controls have them but maybe that is fine. Technically one could have done this with bth-child or matching the title before but less elegant and probably more error prone.

Also this PR needs a similar PR for the monorepo before this one can be merged, since now when we constant pull in changes in that from develop we need to fix it all at once.

TheSpyder:
> I wonder if this will now be abused to custom style individual buttons if all controls have them but maybe that is fine

It's a risk, for sure, but it's better than using the english text to locate buttons in tests.

spocke:
> > I wonder if this will now be abused to custom style individual buttons if all controls have them but maybe that is fine
> 
> It's a risk, for sure, but IMO it's better than using the english text to locate buttons in tests.

Yes people who in the past raced concern about this has left. I remember there where concerns of adding extra properties just used for tests etc. Looks nice from a dev perspective now we can make mcagar simpler and reduce duplicate code of querying for buttons using selectors in a million places.

PR #9314, opened 2024-01-22
TINY-10378: Added confidence tests for the resource css loading functionality

HAFRMO:
Related Ticket: https://ephocks.atlassian.net/browse/TINY-10378

Description of Changes:
Minor confidence test. No changelog added as it's internal only.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/api/Resource.ts
+++ b/modules/tinymce/src/core/main/ts/api/Resource.ts
@@ -6,6 +6,7 @@ interface Resource {
   has: (id: string) => boolean;
   get: (id: string) => any;
   unload: (id: string) => void;
+  remove: (id: string) => void;
 }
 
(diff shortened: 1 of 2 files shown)


hamza0867 (approved):

modules/tinymce/src/core/main/ts/api/Resource.ts
```diff
@@ -65,6 +66,11 @@ const create = (): Resource => {
     resources[id] = data;
   };
 
+  const remove = (id: string) => {
```
spocke, line 69: Unload is supposed to remove the ID if that is not the case then the unload function has. a bug since we introduced the resources collection. So add resources to the `unload` function instead.

- [ ] Unload should be the remove feature

spocke (approved):

PR #9325, opened 2024-01-24
Add `summary` to non-empty elements in schema

vpyshnenko:
Related Ticket: TINY-10303

Description of Changes:
* Add `summary` to non-empty elements in the schema
* Updated existing tests according to the new behavior
* Now `summary` is treated as an editable host and can not be deleted from inside (like a table cell)

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


.changes/unreleased/tinymce-TINY-10303-2024-01-24.yaml
```diff
@@ -0,0 +1,7 @@
+project: tinymce
+kind: Fixed
+body: Pressing Backspace at the start of an empty `summary` element within a `details`
+  element nested in a list item no longer incorrectly removes the `summary` element.
```
TheSpyder, line 4: This is long enough already, we can lose that word 😂 
```suggestion
body: Pressing Backspace at the start of an empty `summary` element within a `details`
  element nested in a list item no longer removes the `summary` element.
```

modules/tinymce/src/plugins/lists/test/ts/browser/BackspaceDeleteTest.ts
```diff
@@ -7,7 +7,7 @@ import Plugin from 'tinymce/plugins/lists/Plugin';
 
 describe('browser.tinymce.plugins.lists.BackspaceDeleteTest', () => {
   const hook = TinyHooks.bddSetupLight<Editor>({
-    plugins: 'lists',
+    plugins: 'lists accordion',
```
TheSpyder, line 10: Is `accordion` really required? This test seems like it would work without that. I don't like having cross-plugin dependencies, even if it's in tests.

TheSpyder (approved):

modules/tinymce/src/plugins/lists/test/ts/browser/BackspaceDeleteTest.ts
vpyshnenko: Fixed.

hamza0867 (approved):

Issue #5506, opened 2020-03-16
Every time after clicking "More"(triple dots) button, the first button in extended bar is in active status


zzs1020:
**What is the current behavior? Describe the bug**
Every time after clicking "More"(triple dots) button, the first button in the extended bar is in activated status
![image](https://user-images.githubusercontent.com/23032993/76772031-9ca3be00-6776-11ea-948a-00ab718ea3ca.png)

**Please provide the steps to reproduce and if possible a minimal demo of the problem via [fiddle.tinymce.com](http://fiddle.tinymce.com/) or similar.**
try on your demo https://www.tiny.cloud/docs/demo/full-featured/, after click triple dots button, then change align to right, then close dots button, then open again

**What is the expected behavior?**
should in inactive status if not clicked before.

**Which versions of TinyMCE, and which browser / OS are affected by this issue? Did this work in previous versions of TinyMCE?**
tinymce 5.10.0, and your cloud version

lostkeys:
Hi, It's not active, it gets the keyboard focus state. Unfortunately it's very similar to the enabled/active state, however a blue focus outline would be too visually disruptive so that's why it is not changed yet. Personally I'd like TinyMCE to have an invisible "button zero" that had the keyboard focus state so the keyboard focus state is invisible until an arrow key is pressed. I'll log this as a feature request.

zzs1020:
@lostkeys thanks!

PR #9333, opened 2024-01-26
TINY-9176: improved showing which ui element has focus

lostkeys:
Related Ticket: TINY-9176

Description of Changes:
* This accessibility improvement adds a clear outline around the element that has focus, and does not rely solely on color.
* The statusbar uses :focus-visible which makes the outline only show up when navigating using the keyboard. This does not work optimally on Safari, but I deem it an acceptable trade-off to not have the outline show up on click.
* Some ui elements like menu items and list menu has a different technical implementation, interlocking the focus and active states. Therefore they behave slightly differently.

Some examples
focus-outlines

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

https://github.com/tinymce/tinymce/issues/5506

### DIFF
--- a/.changes/unreleased/tinymce-TINY-9176-2024-01-26.yaml
+++ b/.changes/unreleased/tinymce-TINY-9176-2024-01-26.yaml
@@ -1,0 +1,6 @@
+project: tinymce
+kind: Improved
+body: Improved showing which element has focus for keyboard navigation
+time: 2024-01-26T10:57:16.397607+01:00
+custom:
+  Issue: TINY-9176
--- a/.changes/unreleased/tinymce-TINY-9176-2024-01-26.yaml
+++ b/.changes/unreleased/tinymce-TINY-9176-2024-01-26.yaml
@@ -1,6 +1,6 @@
 project: tinymce
 kind: Improved
-body: Improved showing which element has focus for keyboard navigation
+body: Improved showing which element has focus for keyboard navigation.
 time: 2024-01-26T10:57:16.397607+01:00
 custom:
   Issue: TINY-9176
--- a/modules/oxide/src/less/theme/components/toolbar-button/toolbar-button.less
+++ b/modules/oxide/src/less/theme/components/toolbar-button/toolbar-button.less
@@ -20,7 +20,7 @@
 @toolbar-button-text-transform: none;
 @toolbar-button-width: 34px;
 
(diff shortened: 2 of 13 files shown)


modules/oxide/src/less/theme/components/toolbar-button/toolbar-button.less
```diff
@@ -20,15 +20,15 @@
 @toolbar-button-text-transform: none;
 @toolbar-button-width: 34px;
 
-@toolbar-button-hover-background-color: multiply(white, contrast(@background-color, fade(@color-tint, 20), fade(@color-tint, 80)));
+@toolbar-button-hover-background-color: contrast(@background-color, darken(@background-color, 6%), lighten(@background-color, 7%)); // Differentiating against the enabled color //multiply(white, contrast(@background-color, fade(@color-tint, 20), fade(@color-tint, 80)));
```
noxuhax, line 23: Probably should remove the old value, must be a leftover?

modules/oxide/src/less/theme/globals/global-variables.less
```diff
@@ -32,6 +32,10 @@
 @panel-border-radius: 6px; // Dialogs, menues etc.
 @control-border-radius: 6px; // Buttons, input fields etc.
 
+// Focus outline
+@keyboard-focus-outline: 0 0 0 2px @color-tint; // Uses box shadow property
```
noxuhax, line 36: Maybe it would be a better idea to only set `@keyboard-focus-outline-width` and `@keyboard-focus-outline-color` as global variables and leave the rest of the shadow properties out of configurable values? 

That would reduce the repetition, simplify the dark skin (a bit) and would remove the need for a separate `inset` variable — inset could be applied in the mixin itself if we pass a bool variable to it.

noxuhax (changes requested):

TheSpyder:
Converting to draft until @lostkeys returns from his trip

.changes/unreleased/tinymce-TINY-9176-2024-01-26.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Improved
+body: Improved showing which element has focus for keyboard navigation
```
spocke, line 3: I think we try to add a `.` at the end of all changelog items these days.

spocke (approved):

modules/oxide/src/less/theme/globals/global-variables.less
lostkeys: I made the requested changes!

noxuhax (approved):
🥂

PR #9337, opened 2024-01-29
TINY-10561: fix CSS for advcode dialog and view mode

lorenzo-pomili:
Related Ticket: TINY-10561

Description of Changes:

- added correct styles for `mce-codemirror`
- changed `tox-view__pane` to `position: 'relative'` since the position was the default (`static`), this shouldn't impact other implementation of the view

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/oxide/src/less/theme/components/advcode/advcode.less
+++ b/modules/oxide/src/less/theme/components/advcode/advcode.less
@@ -1,0 +1,20 @@
+//
+// Advance code
+//
+.tox {
+  .mce-codemirror {
+    position: 'absolute';
+    top: '0';
+    left: '0';
+    bottom: '0';
+    right: '0';
+    z-index: '1';
+    background: #fff;
+    font-size: '13px';
+    margin: '8px';
+
+    &.tox-inline-codemirror {
+      position: relative;
+    }
+  }
+}
--- a/modules/oxide/src/less/theme/components/advcode/advcode.less
+++ b/modules/oxide/src/less/theme/components/advcode/advcode.less
@@ -3,15 +3,15 @@
 //
 .tox {
   .mce-codemirror {
-    position: 'absolute';
-    top: '0';
-    left: '0';
-    bottom: '0';
-    right: '0';
-    z-index: '1';
     background: #fff;
+    bottom: '0';
     font-size: '13px';
+    left: '0';
     margin: '8px';
+    position: 'absolute';
+    right: '0';
+    top: '0';
+    z-index: '1';
 
     &.tox-inline-codemirror {
       position: relative;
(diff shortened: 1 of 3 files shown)


modules/oxide/src/less/theme/components/view/view.less
```diff
@@ -78,6 +78,7 @@
   .tox-view__pane {
     height: 100%;
     padding: @view-pane-padding;
+    position: relative;
```
TheSpyder, line 81: @tiny-ben-tran does this impact the work you're doing?

modules/oxide/src/less/theme/components/advcode/advcode.less
TheSpyder, line 1: Now that we have better options for premium styles, can this not live in the plugin?

TheSpyder (approved):

modules/oxide/src/less/theme/components/view/view.less
tiny-ben-tran: It doesn't seem to

modules/oxide/src/less/theme/components/advcode/advcode.less
spocke: Actually we are probably getting rid of the css in advcode plugin. It used to include the css for the syntax highlighting and some layout overrides not in v6 of codemirror the sytqxg highlighting is done though a style tag that it generates so there is nearly nothing left in the css.

This would make it easier to consume for bundle people as well one less thing to bundle.

vpyshnenko (approved):

modules/oxide/src/less/theme/components/advcode/advcode.less
spocke: I'm resolving this. We already have a bunch of css specific to plugins in oxide I think it would be hard to do skinning otherwise. This is less of an issue than cross dependencies between plugins and or core at a JS API level the only cost here is that the css bundle size is slightly larger but adding 20 lines here and avoiding having to lazy load a CSS file and deal with the building problems would be a win for the customer it would be easier to integrate.

TheSpyder: Sorry my github inbox is a mess.

I disagree on the bundle cost, once we get the bundles done properly with `index.js` files it won't be any harder to bundle with or without CSS.

Skinning I can understand, until we rewrite the lot with CSS variables it'll be difficult for plugins to ship things that might change by the skin variables. Which was intentional, but ended up making our own features more difficult.

spocke: Good point about the index.js making bundling easier. But we are not there yet so this is better for the customer until we do. Also it depends a bit on how we think about css going forward we will likely be more component based in the plugins on the ui side. Then the css should be more of a framework of reusable things that we can apply to things in the components and in this case codemirror might be a reusable component that just happens to be used by the advcode plugin but could theoretically be used by other plugins.

PR #9358, opened 2024-02-05
TINY-10433: Fixed incorect object processor for `event_root` option

### LINKED ISSUE #9218: event_root option appears to be misconfigured
## 📝 Provide detailed reproduction steps (if any)
Please provide the steps to reproduce and if possible a minimal demo of the problem via [fiddle.tiny.cloud](fiddle.tiny.cloud) or similar.
1. Use `event_root` option
2. See a warning when Editor is initialized saying `event_root` should be an `object`

### ✔️ Expected result
*What is the expected result of the above steps?*
Able to use the `event_root` option without seeing a warning on initialization of the Editor

### ❌ Actual result
*What is the actual result of the above steps?*
We see a warning that `event_root` needs to be configured as an `object` on initialization of the Editor

### ❓ Possible solution
*If you have ideas, you can list them here. Otherwise, you can delete this section.*
I believe that we just need to change the configuration of the `event_root` option from `object` to `string` over here:
https://github.com/tinymce/tinymce/blob/454de9339c02b958ad451ac7daa2afd417bf5cb5/modules/tinymce/src/core/main/ts/api/Options.ts#L358-L360

## 📃 Other details
- Browser: Chrome
- OS: MacOS
- First affected version: Since introduced?
- Worked in version: Never?
---

If you'd like to see this fixed sooner, add a 👍 reaction to this post.

shanmen-tiny:
Related Ticket: TINY-10433

Description of Changes:
* Resolves #9218 

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

#9218

### DIFF
--- a/modules/tinymce/src/core/main/ts/api/Options.ts
+++ b/modules/tinymce/src/core/main/ts/api/Options.ts
@@ -356,7 +356,7 @@ const register = (editor: Editor): void => {
   });
 
   registerOption('event_root', {
-    processor: 'object'
+    processor: 'string'
   });
 
   registerOption('service_message', {
--- a/.changes/unreleased/tinymce-TINY-10433-2024-02-05.yaml
+++ b/.changes/unreleased/tinymce-TINY-10433-2024-02-05.yaml
@@ -1,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: Fixed incorrect object processor for `event_root` option.
+time: 2024-02-05T08:24:59.073602+08:00
+custom:
+  Issue: TINY-10433


tiny-ben-tran (approved):

modules/tinymce/src/core/main/ts/api/Options.ts
```diff
@@ -356,7 +356,7 @@ const register = (editor: Editor): void => {
   });
 
   registerOption('event_root', {
```
spocke, line 358: If this has been broken since v6 and no one has complained in all that time do we really need to have this option, maybe we just get rid of it? Since it has pretty much been gone since v6 just not form the docs since it has not been possible to be used right?

spocke (approved):

modules/tinymce/src/core/main/ts/api/Options.ts
shanmen-tiny: I think we should just keep it, at least we know there are one or two of them that are using this option 😅

PR #9359, opened 2024-02-05
TINY-10596: add `Tabstopping` to customeditor component

lorenzo-pomili:
Related Ticket: TINY-10596

Description of Changes:
I added an `focusBehaviour` for the `customEditor` in dialog, this relay on a `focusSelector` in settings, I tried to use a callback instead of a selector but I get an error, and if I remember correctly we can pass a callback in the specs.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/.changes/unreleased/tinymce-TINY-10596-2024-02-07.yaml
+++ b/.changes/unreleased/tinymce-TINY-10596-2024-02-07.yaml
@@ -1,0 +1,6 @@
+project: tinymce
+kind: Added
+body: Added focus behavior for CustomEditor.
+time: 2024-02-07T12:29:51.857930182+01:00
+custom:
+  Issue: TINY-10596
--- a/.changes/unreleased/tinymce-TINY-10596-2024-02-07.yaml
+++ b/.changes/unreleased/tinymce-TINY-10596-2024-02-07.yaml
@@ -1,6 +1,6 @@
 project: tinymce
 kind: Added
-body: Added focus behavior for CustomEditor.
+body: Added onFocus callback for CustomEditor dialog component.
 time: 2024-02-07T12:29:51.857930182+01:00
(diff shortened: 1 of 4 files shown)


modules/tinymce/src/themes/silver/main/ts/ui/dialog/CustomEditor.ts
```diff
@@ -23,6 +24,16 @@ export const renderCustomEditor = (spec: CustomEditorSpec): SimpleSpec => {
   });
 
   const initialValue = Singleton.value<string>();
+  const focusBehaviour = !isOldCustomEditor(spec) && spec.settings?.focusSelector ? [
```
spocke, line 27: I think the `settings` was supposed to be a anything object. I think this should go into the `CustomEditorNew` structure instead as an optional thing or even make it a function like `onFocus: () => void` that way you can do anything when the component gets focus.

- [ ] Move this to the actual spec so that it's properly validated and documented in the types

spocke (changes requested):

.changes/unreleased/tinymce-TINY-10596-2024-02-07.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Added
+body: Added focus behavior for CustomEditor.
```
spocke, line 3: If you move this into the spec you can change this to be `Added onFocus callback for CustomEditor dialog component.` or something like that.

lorenzo-pomili: changed

spocke (approved):

HAFRMO (approved):

PR #9376, opened 2024-02-09
TINY-10637: Replaced autocompleter spans with ranges

spocke:
Related Ticket: TINY-10637

Description of Changes:
* Removes the need for a span to be inserted for autocompleters to enable IMEs to properly function
* Breaks compatibility with RTC but that is no longer supported and the auto completer feature was never finished anyway
* It's now only reading from the dom so composition is never interrupted by us mutating the DOM.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


TheSpyder (approved):
I would say no longer adding elements to the DOM is _more_ compatible with RTC, not less 😛

spocke:
> I would say no longer adding elements to the DOM is _more_ compatible with RTC, not less 😛

Yes but the RTC project not expects that the autocompleter produces a span container so it's likely busted now if we merge this but since it's no longer maintained. But yeah this is what I wanted to do before to get RTC working and was at that time shot down.

vpyshnenko (approved):

PR #9386, opened 2024-02-13
TINY-9980: Added new custom elements spec

spocke:
Related Ticket: TINY-9980

Description of Changes:
* Adds a more powerful custom elements spec to the `custom_elements` option and `addCustomElements` API
* Wraps the setup of the schema into a function that gets invoked so that all the functions are defined before that is executed.
* Changes the custom_elements parser slightly so that it doesn't use the inline property instead it bases the block/inline rules based on what it extends so it if it extends a div then it's a block if it extends a span then it's an inline.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/schema/ValidChildrenRuleParser.ts
+++ b/modules/tinymce/src/core/main/ts/schema/ValidChildrenRuleParser.ts
@@ -1,4 +1,4 @@
-import { Arr } from '@ephox/katamari';
+import { Arr, Fun, Optional } from '@ephox/katamari';
 
 import * as SchemaUtils from './SchemaUtils';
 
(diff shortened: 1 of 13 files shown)


lorenzo-pomili (approved):

modules/tinymce/src/core/main/ts/schema/ValidChildrenRuleParser.ts
```diff
@@ -29,17 +38,9 @@ export const parseValidChildrenRules = (value: string): ValidChildrenRule[] => {
...
-          const preset = nameMatches[1] === '@';
-          const name = nameMatches[2];
-
-          return [{ preset, name }];
-        } else {
-          return [];
-        }
-      });
+      const validChildren = Arr.bind(SchemaUtils.split(matches[3], '|'), (validChild) =>
+        parseValidChild(validChild).fold(Fun.constant([]), Arr.pure)
```
ltrouton, line 42: I think this is equivalent
```suggestion
        parseValidChild(validChild).toArray();
```

ltrouton (approved):

modules/tinymce/src/core/main/ts/schema/ValidChildrenRuleParser.ts
spocke: Right forgot about that one.

PR #9389, opened 2024-02-15
TINY-10560: fix newline after using `selection.setContent` with a `<pre>`

lorenzo-pomili:
Related Ticket: TINY-10560

Description of Changes:
using `editor.selection.setContent('<pre><code>hello</code></pre>')` and then press enter, the problem was that we have this [newBlock](https://github.com/tinymce/tinymce/blob/develop/modules/tinymce/src/core/main/ts/newline/InsertBlock.ts#L410) undefined.

even though if we are at the end of line [this check](https://github.com/tinymce/tinymce/blob/develop/modules/tinymce/src/core/main/ts/newline/InsertBlock.ts#L396) returns false.

This is cuased by the fact that [this check](https://github.com/tinymce/tinymce/blob/develop/modules/tinymce/src/core/main/ts/newline/InsertBlock.ts#L231) doesn't pass because the `container.parentNode` and the `parentBlock` are differend in particular `container.parentNode` is the body and the `parentBlock` is the `<pre>`.

That is the result of this [range normalization](https://github.com/tinymce/tinymce/blob/develop/modules/tinymce/src/core/main/ts/newline/InsertBlock.ts#L304) behave different for the [pre](https://github.com/tinymce/tinymce/blob/develop/modules/tinymce/src/core/main/ts/selection/NormalizeRange.ts#L203).

Since that normalizzation seems to be intented to work in that way my idea was to modify the comparison of the [container.parentNode === parentBlock](https://github.com/tinymce/tinymce/blob/develop/modules/tinymce/src/core/main/ts/newline/InsertBlock.ts#L231) putting it in `or` with `container === parentBlock`

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/.changes/unreleased/tinymce-TINY-10560-2024-02-15.yaml
+++ b/.changes/unreleased/tinymce-TINY-10560-2024-02-15.yaml
@@ -1,6 +1,6 @@
 project: tinymce
 kind: Fixed
-body: Adding newline after using `selection.setContent` with a `<pre>` caused a crash
+body: Adding newline after using `selection.setContent` to insert a block element would throw an unhandled exception.
 time: 2024-02-15T11:41:14.948838122+01:00
 custom:
   Issue: TINY-10560
--- a/.changes/unreleased/tinymce-TINY-10560-2024-02-15.yaml
+++ b/.changes/unreleased/tinymce-TINY-10560-2024-02-15.yaml
@@ -1,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: Adding newline after using `selection.setContent` with a `<pre>` caused a crash
+time: 2024-02-15T11:41:14.948838122+01:00
+custom:
+  Issue: TINY-10560
--- a/modules/tinymce/src/core/main/ts/newline/InsertBlock.ts
+++ b/modules/tinymce/src/core/main/ts/newline/InsertBlock.ts
@@ -228,7 +228,7 @@ const insert = (editor: Editor, evt?: EditorEvent<KeyboardEvent>): void => {
     }
 
     // If after the last element in block node edge case for #5091
-    if (container.parentNode === parentBlock && isAfterLastNodeInContainer && !start) {
+    if ((container.parentNode === parentBlock || container === parentBlock) && isAfterLastNodeInContainer && !start) {
       return true;
     }
 
--- a/modules/tinymce/src/core/test/ts/browser/newline/InsertNewLineTest.ts
+++ b/modules/tinymce/src/core/test/ts/browser/newline/InsertNewLineTest.ts
@@ -836,4 +836,15 @@ describe('browser.tinymce.core.newline.InsertNewLineTest', () => {
     TinyAssertions.assertContent(editor, '<blockquote><p>A</p></blockquote><p>&nbsp;</p>');
     TinyAssertions.assertCursor(editor, [ 1, 0, 0, 0 ], 0);
   });
+
+  it('TINY-10560: Press Enter after a `selection.setContent` should create a new line and not throw errors', () => {
+    const editor = hook.editor();
+    editor.setContent('abc');
+    TinySelections.setSelection(editor, [ 0, 0 ], 0, [ 0, 0 ], 3);
+
+    editor.selection.setContent('<pre><code>hello</code></pre>');
+    TinyAssertions.assertContent(editor, '<p>&nbsp;</p><pre><code>hello</code></pre><p>&nbsp;</p>');
+    insertNewline(editor, { });
+    TinyAssertions.assertContent(editor, '<p>&nbsp;</p><pre><code>hello</code></pre><p>&nbsp;</p><p>&nbsp;</p>');
+  });
 });
--- a/modules/tinymce/src/core/test/ts/browser/newline/InsertNewLineTest.ts
+++ b/modules/tinymce/src/core/test/ts/browser/newline/InsertNewLineTest.ts
@@ -3,6 +3,7 @@ import { after, before, context, describe, it } from '@ephox/bedrock-client';
 import { Type } from '@ephox/katamari';
 import { PlatformDetection } from '@ephox/sand';
 import { TinyAssertions, TinyDom, TinyHooks, TinySelections, TinyState } from '@ephox/wrap-mcagar';
+import { assert } from 'chai';
 
 import Editor from 'tinymce/core/api/Editor';
 import { EditorEvent } from 'tinymce/core/api/util/EventDispatcher';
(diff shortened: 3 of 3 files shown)


modules/tinymce/src/core/main/ts/newline/InsertBlock.ts
```diff
@@ -228,7 +228,7 @@ const insert = (editor: Editor, evt?: EditorEvent<KeyboardEvent>): void => {
     }
 
     // If after the last element in block node edge case for #5091
-    if (container.parentNode === parentBlock && isAfterLastNodeInContainer && !start) {
+    if ((container.parentNode === parentBlock || container === parentBlock) && isAfterLastNodeInContainer && !start) {
```
spocke, line 231: So this change will alter any block element it seems that the bug also happens on h1 etc.

.changes/unreleased/tinymce-TINY-10560-2024-02-15.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: Adding newline after using `selection.setContent` with a `<pre>` caused a crash
```
spocke, line 3: Since this affects more than PRE we should expand it.

```suggestion
body: Adding newline after using `selection.setContent` insert a block element would throw an unhandled excception.
```

modules/tinymce/src/core/test/ts/browser/newline/InsertNewLineTest.ts
```diff
@@ -836,4 +836,15 @@ describe('browser.tinymce.core.newline.InsertNewLineTest', () => {
     TinyAssertions.assertContent(editor, '<blockquote><p>A</p></blockquote><p>&nbsp;</p>');
     TinyAssertions.assertCursor(editor, [ 1, 0, 0, 0 ], 0);
   });
+
+  it('TINY-10560: Press Enter after a `selection.setContent` should create a new line and not throw errors', () => {
```
spocke, line 840: Could you stick this into a test function then feed it more scenarios. H1 etc seems that more blocks can produce the error.

You could also simply set the selection to the spot where this happens the insert code is really not part of the bug the bug is that if you press enter if the selection range is after the text node then it blows up.

You can still have one specific test case of the pre and insert so that we cover the scenario described in the ticket.

So this should also exhibit the bug:
```
editor.setContent('<p>abc</p>');
TinySelections.setCursor(editor, [ 0 ], 1);
insertNewline(editor, { });
```

modules/tinymce/src/core/test/ts/browser/newline/InsertNewLineTest.ts
```diff
@@ -836,4 +836,15 @@ describe('browser.tinymce.core.newline.InsertNewLineTest', () => {
...
+    const editor = hook.editor();
+    editor.setContent('abc');
+    TinySelections.setSelection(editor, [ 0, 0 ], 0, [ 0, 0 ], 3);
+
+    editor.selection.setContent('<pre><code>hello</code></pre>');
+    TinyAssertions.assertContent(editor, '<p>&nbsp;</p><pre><code>hello</code></pre><p>&nbsp;</p>');
+    insertNewline(editor, { });
```
spocke, line 847: There is a assert.doesNotThrow function you can use to signal test intent.

modules/tinymce/src/core/test/ts/browser/newline/InsertNewLineTest.ts
```diff
@@ -836,4 +836,15 @@ describe('browser.tinymce.core.newline.InsertNewLineTest', () => {
     TinyAssertions.assertContent(editor, '<blockquote><p>A</p></blockquote><p>&nbsp;</p>');
     TinyAssertions.assertCursor(editor, [ 1, 0, 0, 0 ], 0);
   });
+
+  it('TINY-10560: Press Enter after a `selection.setContent` should create a new line and not throw errors', () => {
+    const editor = hook.editor();
+    editor.setContent('abc');
+    TinySelections.setSelection(editor, [ 0, 0 ], 0, [ 0, 0 ], 3);
+
+    editor.selection.setContent('<pre><code>hello</code></pre>');
```
spocke, line 845: I don't think the code is needed the minimal case seems to be just a plain text node.

spocke, line 845: It's interesting that `editor.insertContent` doesn't fail so maybe the command some how fixes the selection first.

spocke (approved):

modules/tinymce/src/core/main/ts/newline/InsertBlock.ts
lorenzo-pomili: no it doesn't happen for `h1`.
that `container` is [container = rng.startContainer](https://github.com/tinymce/tinymce/blob/develop/modules/tinymce/src/core/main/ts/newline/InsertBlock.ts#L309), but for `h1` that `rng.startContainer` is the `h1` for `pre` is the `body`.

this is because before the [normalization](https://github.com/tinymce/tinymce/blob/develop/modules/tinymce/src/core/main/ts/newline/InsertBlock.ts#L304), `rng.startContainer` is the `body`.

That normalization when you have a `pre` you get a different behavior from `h1` etc. because of [this check](https://github.com/tinymce/tinymce/blob/develop/modules/tinymce/src/core/main/ts/selection/NormalizeRange.ts#L203), the `nonEmptyElementsMap` have these tags:
```
"summary", "textarea", "svg", "pre", "code", "script", "object", "audio", "video", "iframe", "th", "td", "track", "wbr", "source", "embed", "param", "meta", "link", "isindex", "input", "img", "hr", "frame", "col", "br", "basefont", "base", "area"
```

spocke: I get an error on Chrome Mac if I run this fiddle that inserts a H1 with a text node and hit enter inside that H1: https://fiddle.tiny.cloud/sbZUY0pIPm/0

lorenzo-pomili: I don't get any error in Chrome Linux :/ , does this branch solve the problem with `h1` on your machine?

modules/tinymce/src/core/test/ts/browser/newline/InsertNewLineTest.ts
lorenzo-pomili: ok, I wrapped the `insertNewLine`

modules/tinymce/src/core/test/ts/browser/newline/InsertNewLineTest.ts
lorenzo-pomili: tag `code` removed

TheSpyder (approved):
Once those conversations are resolved I'm happy with this

modules/tinymce/src/core/test/ts/browser/newline/InsertNewLineTest.ts
lorenzo-pomili: I refactored the test, now I do the check for different tags and removed the content expectations, I didn't change the selection to put the caret, let me know if it is ok in this way or if I should change something else

modules/tinymce/src/core/test/ts/browser/newline/InsertNewLineTest.ts
lorenzo-pomili: it seems that `editor.insertContent` is using `selection.setContent` but with a [bookmark](https://github.com/tinymce/tinymce/blob/develop/modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts#L330) and then [move the selection](https://github.com/tinymce/tinymce/blob/develop/modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts#L372) to that bookmark

PR #9194, opened 2023-11-27
TINY-10339: Add `label` component to View `buttons` header

MitchC1999:
Related Ticket: TINY-10339

Description of Changes:
* New `label` component available in the `buttons` array in the View spec.
* Label component comes with two sizes, `normal` (default) and `large`.
* Added CSS for the `label` component and each size.

Pre-checks:
* [x] Changelog entry added
* [x] Tests have been added (if applicable)
* [x] Branch prefixed with `feature/`, `hotfix/` or `spike/`

Review:
* [x] Milestone set
* [x] Docs ticket created (if applicable)

GitHub issues (if applicable):

PR #9404, opened 2024-02-20
TINY-10609: Revert changes made in `a4ec376`

shanmen-tiny:
Related Ticket: TINY-10609

Description of Changes:
* Revert changes made in #9194

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [x] Milestone set


modules/oxide/src/less/theme/components/view/view.less
```diff
@@ -30,22 +30,6 @@
...
-    text-align: @button-text-align;
-  }
-
-  .tox-view__label--normal {
-    font-size: @font-size-md;
-  }
-
-  .tox-view__label--large {
-    font-size: @font-size-lg;
-  }
```
shanmen-tiny, line 47: Wonder if these should be kept here, not sure if we can use the variables within the plugin itself?

spocke (approved):

modules/oxide/src/less/theme/components/view/view.less
ltrouton: Hmmm, I am not sure how we have handled this in the past. I don't think we can use the variables but I think it is acceptable to have a little bit of duplication and hardcode the font-size values in the plugin if needed

ltrouton (approved):

modules/oxide/src/less/theme/components/view/view.less
shanmen-tiny: 👍, will move the normal/large label classes into the plugin itself.

MitchC1999 (approved):

tiny-ben-tran (approved):

PR #9410, opened 2024-02-22
TINY-10578: License change from MIT to GPL 2 or later

spocke:
Related Ticket: TINY-10578

Description of Changes:
* Replaced `LICENSE.txt` with `LICENSE.md`
* Updated the grunt scripts to include `LICENSE.md` instead of `LICENSE.txt` in distribution packages
* Updated the `LICENSE.md` with new information about the license change
* Updated all package.json files that had `MIT` to `GPL-2.0-or-later`
* Updated all distribution packages by changing the grunt script so that they use `GPL-2.0-or-later`
* Updated `grunt-nuget-pack` to latest 1.0.1 version to support the new `license` field.
* Removed the `iconUrl` field from nuget since that is now deprecated. There is a replacement `icon` field that needs to be a local file path. Not super important to have a icon in the nuget registry we can re-visit this later if it is. https://learn.microsoft.com/en-us/nuget/reference/nuspec#iconurl
* Added `license` to nuget to be `GPL-2.0-or-later` and the `licenseUrl` to be `https://licenses.nuget.org/GPL-2.0-or-later` this to avoid the redirect that will point to an 404 after the license rename and also just setting `licenseUrl` is deprecated. It's standard practise to have this `licenseUrl` fallback since `nuget pack` will generate this if omitted. https://learn.microsoft.com/en-us/nuget/reference/nuspec#licenseurl

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/agar/package.json
```diff
@@ -23,7 +23,7 @@
     "test"
   ],
   "author": "Ephox Corporation DBA Tiny Technologies, Inc",
-  "license": "MIT",
+  "license": "GPL-2.0-or-later",
```
TheSpyder, line 26: We will need new major versions for every library

- [ ] bump the majors (should be an easy lerna command)

TheSpyder (approved):

modules/agar/package.json
spocke: Bumped in the version.txt file that is how we normally do it unless that has now changed. Since doing a `lerna version` will treat what is there now as the release and start tagging things.

hamza0867 (approved):

LICENSE.md
```diff
@@ -0,0 +1,6 @@
+# Software License Agreement
+
+**TinyMCE** – <https://github.com/tinymce/tinymce>](https://github.com/tinymce/tinymce)
```
TheSpyder, line 3: missing a `[` at the start of the markdown link

oops, I should've looked at the rendered file 🙃

spocke: Yeah I just copied in what I got from Fredrik didn't verify it closely. I fixed this pushed straight to master seemed silly to take time from people to review a one character change.

PR #9451, opened 2024-02-29
TINY-10642: Removed deprecated items.

HAFRMO:
Related Ticket: https://ephocks.atlassian.net/browse/TINY-10642

Description of Changes:
Removed things due to deprecation

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/bridge/src/main/ts/ephox/bridge/components/dialog/DialogFooterButton.ts
```diff
@@ -77,10 +73,7 @@ const baseFooterButtonFields = [
   ComponentSchema.generatedName('button'),
   ComponentSchema.optionalIcon,
   FieldSchema.defaultedStringEnum('align', 'end', [ 'start', 'end' ]),
-  // this should be removed, but must live here because FieldSchema doesn't have a way to manage deprecated fields
-  ComponentSchema.primary,
   ComponentSchema.enabled,
-  // this should be defaulted to `secondary` but the implementation needs to manage the deprecation
   FieldSchema.optionStringEnum('buttonType', [ 'primary', 'secondary' ])
```
ltrouton, line 77: Is the comment above saying this should be changed to the following?
```suggestion
  FieldSchema.defaultedStringEnum('buttonType', 'secondary', [ 'primary', 'secondary' ])
```

.changes/unreleased/sugar-TINY-10642-2024-02-29.yaml
```diff
@@ -0,0 +1,6 @@
+project: sugar
+kind: Removed
+body: Removed deprecated things
```
ltrouton, line 3: I feel like we should at least mention the APIs that were removed in the entry

ltrouton (commented):
I feel like we should do a better job for the changelog entry messages, especially for the tinymce one. Just leaving some initial comments

modules/tinymce/src/core/main/ts/html/LegacyFilter.ts
ltrouton, line 1: Just want to sanity check. I know we are removing the options but those options have always been defaulted to `true`. By removing this file, 7.0 will essentially act like setting those removed options to `false`. Is that what we want?

modules/tinymce/src/themes/silver/main/ts/ui/general/Button.ts
```diff
@@ -126,7 +121,7 @@ const renderButtonSpec = (
   const components = [ icon.getOrThunk(() => GuiFactory.text(translatedText)) ];
 
   // The old default is based on the now-deprecated 'primary' property. `buttonType` takes precedence now.
-  const buttonType = spec.buttonType.getOr(!spec.primary && !spec.borderless ? 'secondary' : 'primary');
+  const buttonType = spec.buttonType.getOr(!Optionals.equals(spec.buttonType, Optional.some('primary')) && !spec.borderless ? 'secondary' : 'primary');
```
ltrouton, line 124: Can remove the comment above now too. Also, having an `Optionals.equals` inside a `getOr` doesn't seem like the right way to do the check

modules/tinymce/src/core/main/ts/paste/ProcessFilters.ts
```diff
@@ -25,7 +25,7 @@ const preProcess = (editor: Editor, html: string): string => {
     });
   });
 
-  const fragment = parser.parse(html, { forced_root_block: false, isRootContent: true });
+  const fragment = parser.parse(html, { isRootContent: true });
```
ltrouton, line 28: Just want to sanity check that this is correct. 

With the change made at https://github.com/tinymce/tinymce/pull/9451/files#diff-6f02c7b2bc90a7443c69cbcc401505fa683e4e4c09192fa3eb36629373bdfb4dR258 and not explicitly setting `forced_root_block: ''` it seems like the forced root block could either undefined or default to 'p'?

Maybe I am missing some other code somewhere that makes not specificing `forced_root_block` safe

modules/tinymce/src/themes/silver/main/ts/ui/general/Button.ts
TheSpyder: Luke is more than just correct, `spec.buttonType` will by definition not equal `primary` (because it's guaranteed to be `none` in the `getOr`).

modules/tinymce/src/core/main/ts/api/Options.ts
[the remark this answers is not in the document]
TheSpyder: absolutely

TheSpyder (commented):
It seems very late to be doing a change of this scale. Or is this intended for TinyMCE 8?

If this is intended for 7, the only really important stuff is in `tinymce`. The rest we can do any any time as they're not customer-facing API. Perhaps separate them.

The `forced_root_block` and `LegacyFilter` changes I'm really not sure about, as Luke has already commented.

modules/bridge/src/main/ts/ephox/bridge/components/dialog/DialogFooterButton.ts
HAFRMO: That is very possible. I'll change it.

modules/mcagar/src/main/ts/ephox/mcagar/alien/Options.ts
```diff
@@ -1,11 +1,7 @@
 import { Editor } from './EditorTypes';
 
 const get = <R>(editor: Editor, name: string): R => {
```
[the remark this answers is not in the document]
HAFRMO: So we'll keep it deprecated in perpetuity?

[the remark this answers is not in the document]
ltrouton: `editor.getParam(name)` is not deprecated in v4, v5 and we want newer versions of mcagar to still be able to work with the older versions.

modules/tinymce/src/core/main/ts/html/LegacyFilter.ts
HAFRMO: Good question. If you think it's more appropriate we can keep the file and add a task to look into it? Or at the very least, move it somewhere more appropriate than something named 'Legacy'.

modules/mcagar/src/main/ts/ephox/mcagar/alien/Options.ts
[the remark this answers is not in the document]
TheSpyder: That's the key. We don't want to keep deprecated things, but we need mcagar to work with as many versions of TinyMCE as possible. Which can mean talking to now-removed APIs.

modules/bridge/src/main/ts/ephox/bridge/components/dialog/Button.ts
```diff
@@ -34,10 +30,7 @@ const buttonFields = [
   ComponentSchema.generatedName('button'),
   ComponentSchema.optionalIcon,
   ComponentSchema.borderless,
-  // this should be defaulted to `secondary` but the implementation needs to manage the deprecation
-  FieldSchema.optionStringEnum('buttonType', [ 'primary', 'secondary', 'toolbar' ]),
-  // this should be removed, but must live here because FieldSchema doesn't have a way to manage deprecated fields
-  ComponentSchema.primary,
+  FieldSchema.defaultedStringEnum('buttonType', 'secondary', [ 'primary', 'secondary' ])
```
ltrouton, line 33: I think the `Button` interface also needs to be updated based on this change.

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
```diff
@@ -258,16 +255,8 @@ const whitespaceCleaner = (root: AstNode, schema: Schema, settings: DomParserSet
...
-  if (name === false) {
-    return '';
-  } else if (name === true) {
-    return 'p';
-  } else {
-    return name;
-  }
-};
+const getRootBlockName = (settings: DomParserSettings, args: ParserArgs) =>
+  args.forced_root_block ?? settings.forced_root_block ?? '';
```
ltrouton, line 259: I am not convinced adding `?? ''` is actually safe. Around line 487 where `getRootBlockName` is used, there is a check for if the block name is `undefined`. I don't know whether allowing the block name to be undefined is intentional or not but making this change seems risky

ltrouton (changes requested):
Out of all the items removed in the PR which I think is,
- editor.fire() API
- `primary` property removed for all configurable buttons
- `convert_fonts_to_spans` config option (LegacyFilter.ts removed as a result)
- `inline_styles` config option (LegacyFilter.ts removed as a result)
- forced_root_block: <boolean> is no longer valid if using the DomParser() or DomSerializer() APIs directly
- Agar APIs
- Sugar APIs

Are there any items noted above that must be removed for 7.0? As Andy mentioned as well, making these changes, most notably removing the LegacyFilter.ts and associated options and the `forced_root_block` change seems risky past code freeze.

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
ltrouton: I realised that where `getRootBlockName` is used around line 487, it is a falsy check so `''` will act the same as `undefined`. In any case, wherever we previously set `forced_root_block: false` in tests or other files, I would probably prefer for it to be changed to `forced_root_block: ''` rather than just removed.  Others might have different opinion.

modules/tinymce/src/core/test/ts/browser/content/EditorContentNotInitializedTest.ts
```diff
@@ -34,14 +29,14 @@ describe('browser.tinymce.core.content.EditorContentNotInitializedTest', () => {
 
   it('set content on editor without initializing it', () => {
     const editor = createEditor();
-    setContentAndAssertReturn(editor, 'hello');
+    editor.setContent('hello');
```
spocke, line 32: This isn't testing the same thing anymore. The test would check if setContent on an editor would return the same value on getContent if the editor was removed or not yet initialized.

- [ ] Revert test change since this is no longer asserting anything

modules/tinymce/src/plugins/accordion/test/ts/webdriver/AccordionBackspaceDeleteTest.ts
```diff
@@ -105,7 +105,7 @@ describe('webdriver.tinymce.plugins.accordion.AccordionBackspaceDeleteTest', ()
 
     const testPreventBodyDeletion = (deletionKey: DeletionKey) => async () => {
       const editor = hook.editor();
-      editor.setContent(createAccordion(editor, { body: '<p><br></p>' }));
+      createAccordion(editor, { body: '<p><br></p>' });
```
spocke, line 108: Is this correct the old code would do a `editor.setContent(editor.setContent(getAccordionContent(spec))` the new one will be: editor.setContent(getAccordionContent(spec) so skipped one extra setContent filtering. Is that intended, pretty strange test.

spocke (changes requested):

.changes/unreleased/tinymce-TINY-10642-2024-02-29.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Removed
+body: Removed deprecated things
```
spocke, line 3: We need to describe in more detail what we removed here maybe multiple log items even.

modules/tinymce/src/core/test/ts/browser/content/EditorContentNotInitializedTest.ts
HAFRMO: Yes. The deprecated thing here was that the returned value was removed. Checking the return value after that would always be `void`.

spocke: Right I see that the type is changed and the return. Where did we list that as something we want to break?

modules/mcagar/src/main/ts/ephox/mcagar/alien/Options.ts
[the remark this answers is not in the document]
HAFRMO: Keeping it in Mcagar caused some issues with the typing, but I've worked around it. The result is not elegant, but should work.

modules/tinymce/src/plugins/accordion/test/ts/webdriver/AccordionBackspaceDeleteTest.ts
HAFRMO: Yes, this is correct. `editor.setContent` no longer returns a value, so this is no longer possible to do. If you want a similar effect you'd have to do a `getContent` as well, and I wasn't sure we wanted to do that.

modules/tinymce/src/core/test/ts/browser/content/EditorContentNotInitializedTest.ts
HAFRMO: editor.ts, line 809 `<em>Note: The content return value was deprecated in TinyMCE 6.0 and has been marked for removal in TinyMCE 7.0.</em>`

spocke: Right yes that and the setContent API change needs to be reverted. We probably should update that incorrect comment to mention 8.0 since the 7.0 ship has sailed.

PR #9474, opened 2024-03-08
TINY-10590: Tree dom structure could confuse block merge boundary det…

HAFRMO:
Related Ticket: https://ephocks.atlassian.net/browse/TINY-10590

Description of Changes:
We incorrectly detected the placement of a cursor inside a div but outside a block to be a merge between two blocks ( the parent block and the contained block. This would result in incorrect behavior.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


vpyshnenko (approved):

modules/tinymce/src/core/test/ts/browser/delete/BlockMergeBoundaryTest.ts
```diff
@@ -72,6 +72,12 @@ describe('browser.tinymce.core.delete.BlockMergeBoundary', () => {
       const boundaryOpt = readBlockBoundary(true, [ 1, 0 ], 1);
       assertBlockBoundaryNone(boundaryOpt);
     });
+
+    it('TINY-10590: If the cursor is between an internal block and a text-node it should not count as a block boundary', () => {
```
spocke, line 76: Flush this out with more tests. Like backspace on: `<div>A<p>|B</p>C</div>` or delete on `<div>A<p>B|</p>C</div>` so the bug is more wide that just this case. So think the `isNotAncestorial` will need to be expanded a bit to handle the other cases here.

spocke (changes requested):

modules/tinymce/src/core/test/ts/browser/delete/BlockMergeBoundaryTest.ts
spocke: I think you could use `Compare.contains` here so if from contains to or to contains from then it should not do anything.

spocke (approved):

PR #9482, opened 2024-03-14
TINY-10370: Cover `{$caret}` placeholder feature of editor.insertContent API with tests.

vpyshnenko:
Related Ticket: TINY-10370

Description of Changes:
* Cover undocumented {$caret} placeholder feature of editor.insertContent API.

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [x] ~Milestone set~

### DIFF
--- a/modules/tinymce/src/core/test/ts/browser/content/InsertContentTest.ts
+++ b/modules/tinymce/src/core/test/ts/browser/content/InsertContentTest.ts
@@ -859,4 +859,65 @@ describe('browser.tinymce.core.content.InsertContentTest', () => {
       TinyAssertions.assertContent(editor, '<p>a<svg><circle cx="50" cy="50" r="40" stroke="black" stroke-width="3" fill="red"><desc><p>hello</p></desc></circle></svg>b</p>');
     });
   });
+
+  context('${caret) placeholder of caret position after insert', () => {
+    const hook = TinyHooks.bddSetupLight<Editor>({
+      indent: false,
+      base_url: '/project/tinymce/js/tinymce',
+    }, [], true);
+
+    it('TINY-10370: should position the cursor at the beginning', () => {
+      const editor = hook.editor();
+      editor.setContent('<p>cd</p>');
+      TinySelections.setCursor(editor, [ 0, 0 ], 0);
+      editor.insertContent('{$caret}ab');
+      TinyAssertions.assertContent(editor, '<p>abcd</p>');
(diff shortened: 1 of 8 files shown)


modules/tinymce/src/core/test/ts/browser/content/InsertContentTest.ts
```diff
@@ -859,4 +859,65 @@ describe('browser.tinymce.core.content.InsertContentTest', () => {
       TinyAssertions.assertContent(editor, '<p>a<svg><circle cx="50" cy="50" r="40" stroke="black" stroke-width="3" fill="red"><desc><p>hello</p></desc></circle></svg>b</p>');
     });
   });
+
+  context('${caret) placeholder of caret position after insert', () => {
+    const hook = TinyHooks.bddSetupLight<Editor>({
```
spocke, line 864: I don't think you need this one also if you have one hook in the parent scope then one in the sub scope I think there will be two editors at the same time I realize now that I did the same mistake on the svg element I could have put the extended_valid_elements on the parent editor and use that.
So I suggest re-using the hook from the parent scope it has indent: false but some custom elements.

spocke (approved):

modules/tinymce/src/core/test/ts/browser/content/InsertContentTest.ts
vpyshnenko: If I remove this TinyHooks.bddSetupLight call the tests start failing since hook.editor().getBody() returned null.
image

But if remove that one for the upper SVG tests as well, the tests are passed. But I don't like to move  extended_valid_elements for svg to the parent scope since it looks like something specific for svg test context which creeps out and clutters the parent scope, reducing readability.

spocke: Right maybe we should split these out then into separate files. The file is beginning to become quite large. Like a new InsertContentCaretPlaceholderTest.ts then we might move out the SVG test to InsertContentSvgTest.ts or something.

spocke: We could just add the new placeholder test file now then have a code cleanup task to split it apart more.

vpyshnenko: I've created a separate `insert` folder for InsertContent related tests.

lorenzo-pomili (approved):

PR #9508, opened 2024-03-23
TINY-10684: Fix inline header `max-width` calculations

shanmen-tiny:
Related Ticket: TINY-10684

Description of Changes:
* Fixed the `max-width` calculation to use the viewport width instead of body width and use `SugarLocation.viewport` instead of `SugarLocation.absolute`, we want to exclude scrolls and calculate the location of the `targetElement`
* Tests are written to assert at least x number of toolbar groups instead of width since it has been flakey, and that should work across different screen sizes.
* Previous behaviour

https://github.com/tinymce/tinymce/assets/111734091/54720d9e-6b84-4afb-983d-6415194ede8a

*  `width: max-content` is set because the toolbar is absolute positioned, and they will `shrink-wrap` to fit their contents unless the dimensions are specified. See: [link](https://www.w3.org/wiki/CSS_absolute_and_fixed_positioning?source=post_page#:~:text=Absolutely%20positioned%20elements%20will%20shrink,by%20setting%20the%20height%20property)

https://github.com/tinymce/tinymce/assets/111734091/24b5ad08-f7f7-45d2-8de5-89c2c727336f

* Behaviour after fixing

https://github.com/tinymce/tinymce/assets/111734091/1c35312e-b849-4f7d-a7f7-7520e3840797

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/.changes/unreleased/tinymce-TINY-10684-2024-03-23.yaml
+++ b/.changes/unreleased/tinymce-TINY-10684-2024-03-23.yaml
@@ -1,0 +1,7 @@
+project: tinymce
+kind: Fixed
+body: Inline header now calculates `max-width` correctly using the viewport-relative
+  positioning.
+time: 2024-03-23T21:42:30.928799+08:00
+custom:
+  Issue: TINY-10684
--- a/.changes/unreleased/tinymce-TINY-10684-2024-03-23.yaml
+++ b/.changes/unreleased/tinymce-TINY-10684-2024-03-23.yaml
@@ -1,7 +1,6 @@
 project: tinymce
 kind: Fixed
-body: Inline header now calculates `max-width` correctly using the viewport-relative
-  positioning.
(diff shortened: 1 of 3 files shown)


.changes/unreleased/tinymce-TINY-10684-2024-03-23.yaml
```diff
@@ -0,0 +1,7 @@
+project: tinymce
+kind: Fixed
+body: Inline header now calculates `max-width` correctly using the viewport-relative
+  positioning.
```
ltrouton, line 4: Something like this is probably more useful for integrators
```suggestion
body: An inline editor toolbar now behaves correctly in horizontally scrolled containers.
```

ltrouton (approved):

lorenzo-pomili (approved):

tiny-ben-tran (approved):

PR #9521, opened 2024-04-03
TINY-10808 & TINY-10800: Dialog accessibility fixes, dialog title markup changed to use an `h1` element

shanmen-tiny:
Related Ticket: TINY-10800 & TINY-10808

Description of Changes:
* Dialog title markup changed to use an `h1` element instead of `div`
* Don't think we've used platform detection in alloy, but it requires since the `aria-labelledby` is not announced with  `role="dialog"` so that has been changed to `aria-label` only on MacOS
* Ideally we should still use `aria-labelledby` since the dialog already has a visible label, this is a workaround for dialog title to be announced on MacOS VoiceOver. (See: [Dialog Modal Pattern](https://arc.net/l/quote/pchpomfw))
* Changes only impact MacOS, other screen reader like JAWS and NVDA do support usage of `aria-labelledby` in `role="dialog"`. [Assistive Technology Support](https://arc.net/l/quote/jmovypin)

https://github.com/tinymce/tinymce/assets/111734091/dae9d813-bbc3-4fa2-a053-320a40ff7849

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/.changes/unreleased/tinymce-TINY-10808-2024-04-02.yaml
+++ b/.changes/unreleased/tinymce-TINY-10808-2024-04-02.yaml
@@ -1,0 +1,7 @@
+project: tinymce
+kind: Fixed
+body: Dialog title not being announced VoiceOver MacOS, dialog contains `aria-label`
+  instead of `aria-labelledby` on MacOS.
+time: 2024-04-02T21:42:19.730734+08:00
+custom:
+  Issue: TINY-10808
--- a/.changes/unreleased/tinymce-TINY-10808-2024-04-02.yaml
+++ b/.changes/unreleased/tinymce-TINY-10808-2024-04-02.yaml
@@ -1,7 +1,7 @@
 project: tinymce
 kind: Fixed
-body: Dialog title not being announced VoiceOver MacOS, dialog contains `aria-label`
-  instead of `aria-labelledby` on MacOS.
+body: Dialog title was not announced in macOS VoiceOver, dialogs now use `aria-label`
+ instead of `aria-labelledby` on macOS.
 time: 2024-04-02T21:42:19.730734+08:00
 custom:
   Issue: TINY-10808
--- a/modules/alloy/src/main/ts/ephox/alloy/api/ui/ModalDialog.ts
+++ b/modules/alloy/src/main/ts/ephox/alloy/api/ui/ModalDialog.ts
@@ -117,11 +117,12 @@ const factory: CompositeSketchFactory<ModalDialogDetail, ModalDialogSpec> = (det
         AddEventsBehaviour.config(modalEventsId, [
           AlloyEvents.runOnAttached((c) => {
             // TINY-10808 - Workaround to address the dialog header not being announced on VoiceOver with aria-labelledby, ideally we should use the aria-labelledby
-            const title = TextContent.get(AlloyParts.getPartOrDie(c, detail, 'title').element);
+            const titleElm = AlloyParts.getPartOrDie(c, detail, 'title').element;
+            const title = TextContent.get(titleElm);
             if (browser.os.isMacOS() && Type.isNonNullable(title)) {
               Attribute.set(c.element, 'aria-label', title);
             } else {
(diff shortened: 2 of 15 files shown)


.changes/unreleased/tinymce-TINY-10808-2024-04-02.yaml
```diff
@@ -0,0 +1,7 @@
+project: tinymce
+kind: Fixed
+body: Dialog title not being announced VoiceOver MacOS, dialog contains `aria-label`
+  instead of `aria-labelledby` on MacOS.
```
TheSpyder, line 4:
```suggestion
body: Dialog title was not announced in macOS VoiceOver, dialogs now use `aria-label`
 instead of `aria-labelledby` on macOS.
```

And similarly for the alloy changelog (although you can make it more technical, only the tinymce changelog needs to be customer-focused)

modules/alloy/src/main/ts/ephox/alloy/api/ui/ModalDialog.ts
```diff
@@ -114,7 +116,13 @@ const factory: CompositeSketchFactory<ModalDialogDetail, ModalDialogSpec> = (det
         }),
         AddEventsBehaviour.config(modalEventsId, [
           AlloyEvents.runOnAttached((c) => {
-            AriaLabel.labelledBy(c.element, AlloyParts.getPartOrDie(c, detail, 'title').element);
+            // TINY-10808 - Workaround to address the dialog header not being announced on VoiceOver with aria-labelledby, ideally we should use the aria-labelledby
+            const title = TextContent.get(AlloyParts.getPartOrDie(c, detail, 'title').element);
+            if (browser.os.isMacOS() && Type.isNonNullable(title)) {
+              Attribute.set(c.element, 'aria-label', title);
+            } else {
+              AriaLabel.labelledBy(c.element, AlloyParts.getPartOrDie(c, detail, 'title').element);
```
TheSpyder, line 124: Is this macOS bug true of all `aria-labelledby` connections? If so you could move the fix into the `AriaLabel.labelledBy` function 🤔 

Also:
```suggestion
              AriaLabel.labelledBy(c.element, title);
```

modules/tinymce/src/themes/silver/main/ts/ui/window/SilverDialogHeader.ts
```diff
@@ -52,7 +52,7 @@ const renderTitle = (
 
   return {
     dom: {
-      tag: 'div',
+      tag: 'h1',
```
TheSpyder, line 55: have you checked this has no impact on our skins? there are more skins than just the default 🤔 

maybe make a special note for QA to check that.

TheSpyder (approved):

modules/tinymce/src/themes/silver/main/ts/ui/window/SilverDialogHeader.ts
shanmen-tiny: Will do that and make a note, the styles for the heading title use class selector.

modules/alloy/src/main/ts/ephox/alloy/api/ui/ModalDialog.ts
shanmen-tiny: > Is this macOS bug true of all aria-labelledby connections

The other usage is in tabpanel, but they don't use this `AriaLabel`, only the dialog does.

The second parameter takes an element to set the id attribute instead of taking a string.

ltrouton, line 124:
```suggestion
					 const titleElm = AlloyParts.getPartOrDie(c, detail, 'title').element;
            const title = TextContent.get(titleElm);
            if (browser.os.isMacOS() && Type.isNonNullable(title)) {
              Attribute.set(c.element, 'aria-label', title);
            } else {
              AriaLabel.labelledBy(c.element, titleElm);
```

ltrouton (approved):

tiny-ben-tran (approved):

PR #9523, opened 2024-04-04
TINY-10602: Theme loader now respects the suffix when loading skin CSS files

TheSpyder:
Related Ticket: TINY-10602

Description of Changes:
* I swapped premium tests to load the unminified TinyMCE, but it turns out a lot of the skin files still load minified when this happens.
* Some tests had to be updated.
* I wrote a new test for using the suffix in skin URLs.
* I also applied some minor refactoring to the theme loader.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/.changes/unreleased/tinymce-TINY-10602-2024-04-04.yaml
+++ b/.changes/unreleased/tinymce-TINY-10602-2024-04-04.yaml
@@ -1,6 +1,6 @@
 project: tinymce
 kind: Fixed
-body: Updated theme loader respects the suffix when loading skin CSS files
+body: Theme loader now respects the suffix when loading skin CSS files
 time: 2024-04-04T16:35:26.405336+11:00
 custom:
   Issue: TINY-10602
--- a/.changes/unreleased/tinymce-TINY-10602-2024-04-04.yaml
+++ b/.changes/unreleased/tinymce-TINY-10602-2024-04-04.yaml
@@ -1,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: Updated theme loader respects the suffix when loading skin CSS files
+time: 2024-04-04T16:35:26.405336+11:00
+custom:
+  Issue: TINY-10602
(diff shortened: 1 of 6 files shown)


modules/tinymce/src/core/test/ts/browser/init/ShadowDomEditorTest.ts
```diff
@@ -14,8 +14,8 @@ describe('browser.tinymce.core.init.ShadowDomEditorTest', () => {
     }
   });
 
-  const isSkin = (ss: StyleSheet) => ss.href !== null && Strings.contains(ss.href, 'skin.min.css');
-  const isShadowDomSkin = (ss: StyleSheet) => ss.href !== null && Strings.contains(ss.href, 'skin.shadowdom.min.css');
+  const isSkin = (ss: StyleSheet) => ss.href !== null && Strings.contains(ss.href, 'skin.css');
```
spocke, line 17: Wouldn't these need to check both skin.min.css and skin.css not it only matches the latter.

spocke (approved):

.changes/unreleased/tinymce-TINY-10602-2024-04-04.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: Theme loader now respects the suffix when loading skin CSS files
```
spocke, line 3: Looks like a bug fix so maybe past tense is better:
```suggestion
body: Theme loader did not respect the suffix when it was loading skin CSS files.
```

Also I think we should have a `.` a the end.

modules/tinymce/src/core/test/ts/browser/init/ShadowDomEditorTest.ts
TheSpyder: in code, yes. In tests there is no suffix unless we set one - that's the new test I added.

ltrouton (approved):
I was wondering where the `suffix` editor option was set but I now realise that it is one of the cloud override options that is set in the `tinymce.min.js` script

TheSpyder:
@ltrouton for non-cloud users it's here
https://github.com/tinymce/tinymce/blob/4ea0e3f39d79054a054b6dda8614fc9282bc0ab5/modules/tinymce/src/core/main/ts/api/EditorManager.ts#L222-L225

PR #9555, opened 2024-04-12
TINY-10806: Remove `role` attribute from Color Swatch menu item.

vpyshnenko:
Related Ticket: TINY-10806

Description of Changes:
* Added the `menuRole` boolean field to Alloy Menu Schema defaulted to true. Set it to false for the Color  Swatch  menu item.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/alloy/src/main/ts/ephox/alloy/ui/schema/MenuSchema.ts
+++ b/modules/alloy/src/main/ts/ephox/alloy/ui/schema/MenuSchema.ts
@@ -124,7 +124,8 @@ const schema = Fun.constant([
   FieldSchema.defaulted('fakeFocus', false),
   FieldSchema.defaulted('focusManager', FocusManagers.dom()),
   Fields.onHandler('onHighlight'),
-  Fields.onHandler('onDehighlight')
+  Fields.onHandler('onDehighlight'),
+  FieldSchema.defaulted('menuRole', true),
 ]);
 
 const name = Fun.constant('menu');
--- a/modules/tinymce/src/themes/silver/main/ts/ui/menus/item/build/ColorSwatchItem.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/ui/menus/item/build/ColorSwatchItem.ts
@@ -1,6 +1,6 @@
-import { ItemTypes, ItemWidget, Menu as AlloyMenu, MenuTypes, SketchSpec } from '@ephox/alloy';
+import { ItemTypes, ItemWidget, Menu as AlloyMenu, MenuTypes } from '@ephox/alloy';
 import { Menu } from '@ephox/bridge';
-import { Fun, Id, Optional, Resolve } from '@ephox/katamari';
+import { Fun, Id } from '@ephox/katamari';
 
 import { UiFactoryBackstage } from 'tinymce/themes/silver/backstage/Backstage';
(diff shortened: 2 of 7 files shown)


modules/tinymce/src/themes/silver/main/ts/ui/menus/item/build/ColorSwatchItem.ts
```diff
@@ -34,6 +42,9 @@ export const renderColorSwatchItem = (spec: Menu.ColorSwatchMenuItem, backstage:
     movement: deriveMenuMovement(columns, presets)
   };
 
+  // TINY-10806: Avoid duplication of ARIA role="menu" in the accessibility tree for Color Swatch menu item.
+  const widgetSketchSpec = removeRoleAttr(AlloyMenu.sketch(widgetSpec));
```
TheSpyder, line 46: Why are we post-mutating the spec, and not just avoiding the role going into the spec to begin with?

spocke: You probably need to modify alloy to allow a menu to not have the menu role. This color swatch menu item is very special it kind of works like a menu since it has child elements with the role=menuitemradio but since the parent is already a role=menu it shouldn't have role=menu on it's container. Maybe add some weird boolean `menuRole` that defaults to true then you can set it to false and if you do then you. don't add the role=menu in the MenuSpec.ts. That way it's all immutable and passed down to the thing.

vpyshnenko: Initially, I'd like to avoid extending the menu spec for such an exceptional case. But now I see it is the wrong approach, so I added the `menuRole` boolean field defaulted to `true`.

TheSpyder: If it was a public API that might've been worth considering, but when it's all internal like this then yes adjusting the API is fine - particularly if it avoids mutating what is basically a config object.

Side effects, implicit dependencies, spaghetti code, these are all problems that come up in a few years if architecture decisions aren't taken carefully :)

TheSpyder (approved):

modules/alloy/src/main/ts/ephox/alloy/ui/schema/MenuSchema.ts
```diff
@@ -124,7 +124,8 @@ const schema = Fun.constant([
   FieldSchema.defaulted('fakeFocus', false),
   FieldSchema.defaulted('focusManager', FocusManagers.dom()),
   Fields.onHandler('onHighlight'),
-  Fields.onHandler('onDehighlight')
+  Fields.onHandler('onDehighlight'),
+  FieldSchema.defaulted('menuRole', true),
```
ltrouton, line 128: We should add an alloy changelog entry as well

ltrouton (approved):

shanmen-tiny (approved):

modules/alloy/src/main/ts/ephox/alloy/ui/schema/MenuSchema.ts
vpyshnenko: done

PR #9560, opened 2024-04-16
TINY-10597: Focus traps in the editor when in fullscreen mode

shanmen-tiny:
Related Ticket: TINY-10597

Description of Changes:
* `fullscreen_trap_focus` that prevents editor from being blurred when tabbing out of the editor
* Existing flow of to the address bar, or browser UIs are maintained, the activeElement is `<body>` when the browser buttons/menus are in focus

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/.changes/unreleased/tinymce-TINY-10597-2024-04-14.yaml
+++ b/.changes/unreleased/tinymce-TINY-10597-2024-04-14.yaml
@@ -1,0 +1,7 @@
+project: tinymce
+kind: Added
+body: New `fullscreen_trap_focus` option, that prevents focus from going out of the editor
+  when in fullscreen mode.
+time: 2024-04-14T21:23:35.919442+08:00
+custom:
+  Issue: TINY-10597
--- a/.changes/unreleased/tinymce-TINY-10597-2024-04-14.yaml
+++ b/.changes/unreleased/tinymce-TINY-10597-2024-04-14.yaml
@@ -1,7 +1,6 @@
 project: tinymce
 kind: Added
-body: New `fullscreen_trap_focus` option, that prevents focus from going out of the editor
-  when in fullscreen mode.
+body: Traps focus from going out of the editor when in fullscreen mode.
 time: 2024-04-14T21:23:35.919442+08:00
 custom:
   Issue: TINY-10597
--- a/.changes/unreleased/tinymce-TINY-10597-2024-04-14.yaml
+++ b/.changes/unreleased/tinymce-TINY-10597-2024-04-14.yaml
@@ -1,6 +1,6 @@
 project: tinymce
 kind: Fixed
-body: Fullscreen mode now traps focus in the editor, prevents focus from leaving the editor.
+body: Fullscreen mode now prevents focus from leaving the editor.
 time: 2024-04-14T21:23:35.919442+08:00
 custom:
   Issue: TINY-10597
--- a/.changes/unreleased/tinymce-TINY-10597-2024-04-14.yaml
+++ b/.changes/unreleased/tinymce-TINY-10597-2024-04-14.yaml
@@ -1,6 +1,6 @@
 project: tinymce
-kind: Added
-body: Traps focus from going out of the editor when in fullscreen mode.
+kind: Fixed
+body: Fullscreen mode now traps focus in the editor, prevents focus from leaving the editor.
 time: 2024-04-14T21:23:35.919442+08:00
 custom:
   Issue: TINY-10597
--- a/modules/tinymce/src/core/main/ts/api/Options.ts
+++ b/modules/tinymce/src/core/main/ts/api/Options.ts
@@ -859,6 +859,11 @@ const register = (editor: Editor): void => {
     default: true
   });
 
+  registerOption('fullscreen_trap_focus', {
+    processor: 'boolean',
+    default: false
+  });
+
   // These options must be registered later in the init sequence due to their default values
   editor.on('ScriptsLoaded', () => {
     registerOption('directionality', {
@@ -980,6 +985,7 @@ const getSandboxIframesExclusions = (editor: Editor): string[] => editor.options
 const shouldConvertUnsafeEmbeds = option('convert_unsafe_embeds');
 const getLicenseKey = option('license_key');
 const getApiKey = option('api_key');
+const shouldFullScreenTrapFocus = option('fullscreen_trap_focus');
 
 export {
   register,
@@ -1089,5 +1095,6 @@ export {
   getLicenseKey,
   getSandboxIframesExclusions,
   shouldConvertUnsafeEmbeds,
-  getApiKey
+  getApiKey,
+  shouldFullScreenTrapFocus
 };
--- a/modules/tinymce/src/core/main/ts/api/Options.ts
+++ b/modules/tinymce/src/core/main/ts/api/Options.ts
@@ -859,11 +859,6 @@ const register = (editor: Editor): void => {
     default: true
   });
 
-  registerOption('trap_focus', {
-    processor: 'boolean',
-    default: false
-  });
-
   // These options must be registered later in the init sequence due to their default values
   editor.on('ScriptsLoaded', () => {
     registerOption('directionality', {
@@ -985,7 +980,6 @@ const getSandboxIframesExclusions = (editor: Editor): string[] => editor.options
 const shouldConvertUnsafeEmbeds = option('convert_unsafe_embeds');
 const getLicenseKey = option('license_key');
 const getApiKey = option('api_key');
-const shouldFullScreenTrapFocus = option('trap_focus');
 
 export {
   register,
@@ -1095,6 +1089,5 @@ export {
   getLicenseKey,
   getSandboxIframesExclusions,
   shouldConvertUnsafeEmbeds,
-  getApiKey,
-  shouldFullScreenTrapFocus
+  getApiKey
 };
--- a/modules/tinymce/src/core/main/ts/api/Options.ts
+++ b/modules/tinymce/src/core/main/ts/api/Options.ts
@@ -859,7 +859,7 @@ const register = (editor: Editor): void => {
     default: true
   });
 
-  registerOption('fullscreen_trap_focus', {
+  registerOption('trap_focus', {
     processor: 'boolean',
     default: false
   });
@@ -985,7 +985,7 @@ const getSandboxIframesExclusions = (editor: Editor): string[] => editor.options
 const shouldConvertUnsafeEmbeds = option('convert_unsafe_embeds');
 const getLicenseKey = option('license_key');
 const getApiKey = option('api_key');
-const shouldFullScreenTrapFocus = option('fullscreen_trap_focus');
+const shouldFullScreenTrapFocus = option('trap_focus');
 
 export {
   register,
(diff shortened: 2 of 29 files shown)


modules/tinymce/src/core/main/ts/focus/FocusController.ts
```diff
@@ -18,6 +18,10 @@ const isEditorUIElement = (elm: Node): boolean => {
   return NodeType.isElement(elm) && FocusManager.isEditorUIElement(elm);
 };
 
+const isEditorUIElementWithoutFullscreen = (elm: Node): boolean => {
+  return NodeType.isElement(elm) && FocusManager.isEditorUIElement(elm) && !elm.className.toString().includes('tox-fullscreen');
+};
+
```
shanmen-tiny, line 24: Not sure if this is permitted to be here. In fullscreen mode, the `tox-fullscreen` class is added to the body and the document element. When the elements inside the body (not part of the editor) receives focus, the editor blur is not dispatched since the ancestor/parent of the element contains the `tox` class.

spocke: The `tox-` is also hard coded into the core ideally the theme should export the predicate but that never happened. So I don't think it's the end of the world to hard code more things here.

modules/tinymce/src/plugins/fullscreen/main/ts/api/Events.ts
```diff
@@ -1,10 +1,25 @@
...
 
 const fireFullscreenStateChanged = (editor: Editor, state: boolean): void => {
   editor.dispatch('FullscreenStateChanged', { state });
   editor.dispatch('ResizeEditor');
 };
 
+const setup = (editor: Editor, fullscreenState: Cell<Actions.FullScreenInfo | null>): void => {
+  editor.on('blur', () => {
+    if (fullscreenState.get() && Options.shouldFullScreenTrapFocus(editor)) {
+      Delay.setEditorTimeout(editor, () => editor.focus());
```
spocke, line 17: Will this flicker. Can this be done instead by preventing tab key default behavior in the editor if the editor is in a fullscreen state instead?

modules/tinymce/src/plugins/fullscreen/test/ts/webdriver/FullscreenTrapFocusTest.ts
```diff
@@ -0,0 +1,111 @@
...
+        const pDoShiftTab = async (selector: string = 'iframe => body'): Promise<void> => {
+          await RealKeys.pSendKeysOn(selector, [ RealKeys.combo({ shift: true }, 'tab') ]);
+        };
+
+        const pDoTab = async (selector: string = 'iframe => body'): Promise<void> => {
+          await RealKeys.pSendKeysOn(selector, [ RealKeys.text('tab') ]);
+        };
+
+        const setupInput = (insert: (marker: SugarElement<Node>, element: SugarElement<Node>) => void, editor: Editor) => {
+          const input = SugarElement.fromHtml('<input />');
```
spocke, line 39: I don't think you need XHTML style here also you can use `fromTag('input')`.

modules/tinymce/src/core/main/ts/api/Options.ts
```diff
@@ -859,6 +859,11 @@ const register = (editor: Editor): void => {
     default: true
   });
 
+  registerOption('fullscreen_trap_focus', {
+    processor: 'boolean',
+    default: false
```
spocke, line 864: Maybe we should consider this be true for v8 seems like something you normally don't want to happen.

spocke (approved):

modules/tinymce/src/plugins/fullscreen/main/ts/api/Events.ts
shanmen-tiny: @spocke I was thinking the same initially, but won't preventing tab key default behaviour would also interfere with the list indentation, nonbreaking plugin tab and table navigation? Not sure if I'm missing something.

spocke: It would but I guess it depends on when you register that block if it's after all those then the defulat native behavior would be prevented. Would have been nice if we had a preventDefaultNative so that we would only prevent default behavior but not our behavior. 😊

TheSpyder: Yeah if you wait for the `beforeInit` or `init` events to register a key handler it will happen after the normal handlers (core and other plugins) are registered.

But `blur` only happens when tab causes focus to leave the editor, so there should be no need to delay this focus call?

modules/tinymce/src/plugins/fullscreen/main/ts/Plugin.ts
```diff
@@ -19,6 +20,7 @@ export default (): void => {
     Options.register(editor);
     Commands.register(editor, fullscreenState);
     Buttons.register(editor, fullscreenState);
+    Events.setup(editor, fullscreenState);
```
TheSpyder, line 23: Why do we need this as well as the core feature? I expected one or the other, not both?

modules/tinymce/src/core/main/ts/focus/FocusController.ts
```diff
@@ -54,6 +58,11 @@ const getActiveElement = (editor: Editor): Element => {
   }
 };
 
+const shouldTrapFocusInFullScreen = (editor: Editor): boolean =>
+  editor.plugins.fullscreen
+    && editor.plugins.fullscreen.isFullscreen()
```
TheSpyder, line 63: This is specifically what I asked you not to do, we can't have core assuming things about plugin capabilities 😛

modules/tinymce/src/core/main/ts/api/Options.ts
TheSpyder: I asked for this to be an undocumented option that the fullscreen plugin can set in order to avoid core needing to know the plugin exists and what it's capable of. The alternative would be to pull fullscreen into the core, which while more convenient for fixes like this probably isn't the best idea.

A better name would probably be `trap_focus`, then it's a capability that can be enabled and disabled as appropriate.

TheSpyder (changes requested):

modules/tinymce/src/plugins/fullscreen/main/ts/api/Events.ts
shanmen-tiny: Fixed that now the logic resides in core, so there's no need of this event here in the plugin 👍

modules/tinymce/src/core/main/ts/api/Options.ts
shanmen-tiny: Fixed that, now it has been renamed to `trap_focus`.

.changes/unreleased/tinymce-TINY-10597-2024-04-14.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: Fullscreen mode now traps focus in the editor, prevents focus from leaving the editor.
```
TheSpyder, line 3:
```suggestion
body: Fullscreen mode now prevents focus from leaving the editor.
```

TheSpyder (approved):
Much better!

spocke:
I see no reason why this couldn't all live in the fullscreen plugin and just be a prevent on the default behavior? Made a fiddle here: https://fiddle.tiny.cloud/oMPOxS5GAV/0 I can't get out of the iframe in that using tab and the UI already traps tab so not sure we need any special there.

TheSpyder:
I think I was hinting at that in my comment about the event handler
https://github.com/tinymce/tinymce/pull/9560#discussion_r1568058034

But I got lost in the chat about having a feature flag on the core editor for this

shanmen-tiny:
> I see no reason why this couldn't all live in the fullscreen plugin and just be a prevent on the default behavior? Made a fiddle here: [fiddle.tiny.cloud/oMPOxS5GAV/0](https://fiddle.tiny.cloud/oMPOxS5GAV/0) I can't get out of the iframe in that using tab and the UI already traps tab so not sure we need any special there.

@spocke Yeah registering after the `init` would place the keydown after everything has registered, this would ensure that the other keydown are not blocked. I've just got one concern on this as it also blocks shift-tabbing into the browser UI buttons since tabbing are all are blocked. 🤔 
Would that be something that we need to consider?

spocke:
> @spocke Yeah registering after the `init` would place the keydown after everything has registered, this would ensure that the other keydown are not blocked. I've just got one concern on this as it also blocks shift-tabbing into the browser UI buttons since tabbing are all are blocked. 🤔 Would that be something that we need to consider?

I think shift-tab should also be blocked since you get to the tab index elements above before coming to the browser also we do this in dialogs as well so I guess one could treat fullscreen as a dialog kind of state where the tabbing becomes centered inside the editor.

modules/tinymce/src/plugins/fullscreen/main/ts/core/Keyboard.ts
```diff
@@ -0,0 +1,21 @@
...
+
+import Editor from 'tinymce/core/api/Editor';
+import VK from 'tinymce/core/api/util/VK';
+
+import { FullScreenInfo } from './Actions';
+
+const setup = (editor: Editor, fullscreenState: Cell<FullScreenInfo | null>): void => {
+  editor.on('init', () => {
+    editor.on('keydown', (e) => {
+      if ((e.keyCode === VK.TAB || (e.keyCode === VK.TAB && e.shiftKey)) && fullscreenState.get()) {
```
TheSpyder, line 11: Your `|| (e.keyCode === VK.TAB && e.shiftKey)` did nothing, because it would only run if `e.keyCode === VK.TAB` was false.

Instead, what you probably want is to only handle tab with no modifiers other than shift.

```suggestion
      if (e.keyCode === VK.TAB && !(e.metaKey || e.ctrlKey) && fullscreenState.get()) {
```

TheSpyder (approved):
shift+tab was blocked by the `blur` handler, so yes it makes sense to block it this way as well.

modules/tinymce/src/plugins/fullscreen/test/ts/webdriver/FullscreenTrapFocusTest.ts
```diff
@@ -0,0 +1,143 @@
...
+    }
+  });
+
+  const pToggleFullscreen = async (editor: Editor, nativeMode: boolean, fullscreen: boolean) => {
+    TinyContentActions.keystroke(editor, 121, { alt: true });
+    await FocusTools.pTryOnSelector('Assert toolbar is focused', SugarDocument.getDocument(), 'div[role=toolbar] .tox-tbtn');
+    await RealKeys.pSendKeysOn('div[role=toolbar] .tox-tbtn', [ RealKeys.text('enter') ]);
+    if (nativeMode) {
+      await pIsFullscreen(fullscreen);
+      await Waiter.pWait(500);
```
ltrouton, line 45: What is this for? If we need to wait for the next execution cycle we could do `Waiter.pWait(0);`

ltrouton (approved):

modules/tinymce/src/plugins/fullscreen/test/ts/webdriver/FullscreenTrapFocusTest.ts
shanmen-tiny: I was getting some weird failures with animations locally when going in and out from fullscreen mode with native 😅

PR #9562, opened 2024-04-17
TINY-10931: Fix open link menu item in link context menu not working with selection expanded over link to the outside content

tiny-ben-tran:
Related Ticket: TINY-10391

Description of Changes:
* Improved the logic of finding links for open link menu action so it would be able to find the targeted link inside the selection

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


.changes/unreleased/tinymce-TINY-10391-2024-04-17.yaml
```diff
@@ -0,0 +1,7 @@
+project: tinymce
+kind: Fixed
+body: Open link menu in link context menu did not work with the selection over link
+  to the outside content
```
TheSpyder, line 4:
```suggestion
body: Open link context menu action did not work with selection surrounding a link
```

modules/tinymce/src/core/test/ts/browser/dom/ControlSelectionTest.ts
```diff
@@ -32,7 +32,7 @@ describe('browser.tinymce.core.dom.ControlSelectionTest', () => {
     const clientX = (rect.left + rect.width / 2), clientY = (rect.top + rect.height / 2);
     editor.dispatch('mousedown', { target, clientX, clientY, button: 2 } as MouseEvent);
     editor.dispatch('mouseup', { target, clientX, clientY, button: 2 } as MouseEvent);
-    editor.dispatch('contextmenu', { target, clientX, clientY, button: 2 } as PointerEvent);
+    editor.dispatch('contextmenu', { target, clientX, clientY, button: 2, x: clientX, y: clientY } as PointerEvent);
```
TheSpyder, line 35: um, `clientX` and `clientY` are already in the event attributes?

modules/tinymce/src/core/main/ts/NodeChange.ts
```diff
@@ -49,9 +50,15 @@ class NodeChange {
 
     // IE has a bug where it fires a selectionchange on right click that has a range at the start of the body
     // When the contextmenu event fires the selection is located at the right location
-    editor.on('contextmenu', () => {
-      SelectionBookmark.store(editor);
+    editor.on('contextmenu', (event) => {
+      const elementFromPoint = SugarElement.fromPoint(SugarElement.fromDom(editor.contentDocument), event.x, event.y);
+      elementFromPoint.each((elm) => {
+        if (elm.dom.nodeName === 'A') {
+          editor.selection.select(elm.dom);
```
TheSpyder, line 57: I don't think modifying the selection just because the context menu opened is a good idea.

With a bit of digging it turns out the code that determines whether `open link` should be enabled can deal with selected content that includes a link.
https://github.com/tinymce/tinymce/blob/f6a87ed7a6d32c627cf8ac6c1bd44ce36d69b14b/modules/tinymce/src/plugins/link/main/ts/core/Actions.ts#L83-L87

But the `open link` code itself doesn't use that - it has it's own code to find the link to open, and it's less flexible.
https://github.com/tinymce/tinymce/blob/f6a87ed7a6d32c627cf8ac6c1bd44ce36d69b14b/modules/tinymce/src/plugins/link/main/ts/core/Actions.ts#L9-L13

Just make them both use the more flexible code.

TheSpyder (changes requested):

modules/tinymce/src/core/main/ts/NodeChange.ts
tiny-ben-tran: Done!

modules/tinymce/src/plugins/link/main/ts/core/Actions.ts
```diff
@@ -6,11 +8,11 @@ import VK from 'tinymce/core/api/util/VK';
...
 
-const getLink = (editor: Editor, elm: Node): HTMLAnchorElement | null =>
-  editor.dom.getParent<HTMLAnchorElement>(elm, 'a[href]');
+const getLinks = (editor: Editor) => editor.selection.isCollapsed() ?
+  Utils.getLinks(editor.dom.getParents(editor.selection.getStart())) as [HTMLAnchorElement] :
+  Utils.getLinksInSelection(editor.selection.getRng());
 
-const getSelectedLink = (editor: Editor): HTMLAnchorElement | null =>
-  getLink(editor, editor.selection.getStart());
+const getSelectedLink = (editor: Editor): HTMLAnchorElement | null => Arr.head(getLinks(editor)).getOrNull();
```
TheSpyder, line 15: If you look at where this is used, the strictly `null` return type was just a factor from `editor.dom.getParent`. Not strictly necessary.
```suggestion
const getSelectedLink = (editor: Editor): HTMLAnchorElement | undefined => getLinks(editor)[0];
```

modules/tinymce/src/plugins/link/main/ts/core/Actions.ts
```diff
@@ -39,11 +41,12 @@ const gotoSelectedLink = (editor: Editor) => (): void => {
 };
 
 const setupGotoLinks = (editor: Editor): void => {
-  editor.on('click', (e) => {
-    const link = getLink(editor, e.target);
-    if (link && VK.metaKeyPressed(e)) {
+  editor.on('click', (e: any) => {
+    const links = Utils.getLinks(editor.dom.getParents(e.target)) as [HTMLAnchorElement];
```
TheSpyder, line 45: I'm a little surprised you refactored the code above without using it here. Is there a test for the refactored code at least?

TheSpyder (approved):

modules/tinymce/src/plugins/link/main/ts/core/Actions.ts
tiny-ben-tran: It would require the event target `e.target` to be passed into the refactored code, which makes harder to understand so I decided not to use it.

I will add few more tests for the refactored code

.changes/unreleased/tinymce-TINY-10391-2024-04-17.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: Open link context menu action did not work with selection surrounding a link
```
spocke, line 3:
```suggestion
body: Open link context menu action did not work with selection surrounding a link.
```

modules/tinymce/src/plugins/link/main/ts/core/Actions.ts
```diff
@@ -39,11 +41,12 @@ const gotoSelectedLink = (editor: Editor) => (): void => {
 };
 
 const setupGotoLinks = (editor: Editor): void => {
-  editor.on('click', (e) => {
-    const link = getLink(editor, e.target);
-    if (link && VK.metaKeyPressed(e)) {
+  editor.on('click', (e: any) => {
```
spocke, line 44: Why the any here. It is a proper type so `t.target` should exist.The metaKeyPressed takes keyboard like events and just checks for the metaKey, ctrlKey etc so that should also be on this event object.

- [ ] Remove any cast

modules/tinymce/src/plugins/link/test/ts/browser/OpenLinkContextMenuTest.ts
```diff
@@ -0,0 +1,60 @@
...
+  const hook = TinyHooks.bddSetupLight<Editor>({
+    plugins: 'link',
+    toolbar: 'link',
+    indent: false,
+    base_url: '/project/tinymce/js/tinymce',
+    image_caption: true
+  }, [ LinkPlugin ], true);
+
+  // Capture the href links being clicked
+  document.addEventListener('click', (e) => {
```
spocke, line 21: This isn't properly removed you could use sugar here it's DomEvent and use the unbinder of that. THis should also be done in a hook so before/after not just like this for everything regardless if the test is executed or not.

- [ ] Make event registration proper on before/after hooks and with a cleanup

spocke (approved):

ArvinJ-H (approved):

shanmen-tiny (approved):

PR #9567, opened 2024-04-18
TINY-10853: Added math plugin to default menu, help and content css

spocke:
Related Ticket: TINY-10853

Description of Changes:
* Added the `math` plugin to the default insert menu.
* Added css to display the `tiny-math-block` element as `flex` this makes it possible to later have alignment on it using `justify-content` defaults to center since that is what a math element in a regular block would do and more of the norm.
* Had to disable a csslint rule since it doesn't seem to like custom elements. Might investigate that further in some other ticket.
* Added the math plugin to the help dialog.

Pre-checks:
* [-] Changelog entry added
* [-] Tests have been added (if applicable)

Review:
* [x] Milestone set
* [-] Docs ticket created (if applicable)

### DIFF
--- a/modules/oxide/src/less/theme/content/math/math.less
+++ b/modules/oxide/src/less/theme/content/math/math.less
@@ -2,9 +2,12 @@
 // Math
 //
 
+@math-margin: 16px 0 16px 0;
+
 /* stylelint-disable selector-type-no-unknown */
 tiny-math-block {
   display: flex;
   justify-content: center;
+  margin: @math-margin;
 }
 /* stylelint-enable selector-type-no-unknown */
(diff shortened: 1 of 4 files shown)


modules/oxide/src/less/theme/content/math/math.less
```diff
@@ -0,0 +1,13 @@
+//
+// Math
+//
+
+@math-margin: 16px 0 16px 0;
+
+/* stylelint-disable selector-type-no-unknown */
```
TheSpyder, line 7: Our stylelint is _very_ old, upgrading it is the first thing to try to fix this. I think I was going to do it as a part of switching oxide to the latest `less` version, and from gulp to grunt, but that fell through the cracks in the lead-up to 7.0

TheSpyder (approved):

PR #9577, opened 2024-04-19
TINY-10837: Fix for style attribute not being retained when toggling lists

spocke:
Related Ticket: TINY-10837

Description of Changes:
* Fixes so the `style` attribute gets applied from the LI model element to the block element only styles right now not sure if all attributes should be retained or not maybe we expand it later.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set
* [-] Docs ticket created (if applicable)

### DIFF
--- a/.changes/unreleased/tinymce-TINY-10837-2024-04-18.yaml
+++ b/.changes/unreleased/tinymce-TINY-10837-2024-04-18.yaml
@@ -1,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: Styles was not retained when toggling a list on and off.
+time: 2024-04-18T10:57:01.817028+02:00
+custom:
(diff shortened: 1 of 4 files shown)


.changes/unreleased/tinymce-TINY-10837-2024-04-18.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: Styles was not retained when toggling a list on and off.
```
TheSpyder, line 3:
```suggestion
body: Styles were not retained when toggling a list on and off.
```

TheSpyder (approved):

vpyshnenko (approved):

PR #9580, opened 2024-04-22
TINY-10741: Improve the visibility of focus outline on the view toolbar buttons

tiny-ben-tran:
Related Ticket: TINY-10741

Description of Changes:
* Increased the padding of the view toolbar so the focus outline of the buttons are not obscured by the scrollbar

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [x] Milestone set

### DIFF
--- a/modules/oxide/src/less/theme/components/view/view.less
+++ b/modules/oxide/src/less/theme/components/view/view.less
@@ -1,4 +1,4 @@
-@view-header-padding: @pad-sm @pad-sm 0 @pad-sm;
+@view-header-padding: 10px 10px 2px 10px;
 @view-header-padding-mobile: @pad-sm;
 @view-pane-padding: @pad-sm;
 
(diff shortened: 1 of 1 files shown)


modules/oxide/src/less/theme/components/view/view.less
```diff
@@ -64,7 +64,7 @@
     gap: @view-toolbar-button-spacing-x; // spacing between buttons within a group
     justify-content: space-between;
     overflow-x: auto;           // Show horizontal scrollbar on small screens
-    padding: @view-header-padding;
+    padding: 10px 10px 2px 10px;
```
spocke, line 67: Why the hardcoded value shouldn't you change the variable here?
What effects does this have on other things that use this like the inline code view?

ArvinJ-H (approved):

modules/oxide/src/less/theme/components/view/view.less
tiny-ben-tran: My intention was to limit this change to what revision history is using but it is not consistent when the toolbar has no button groups.  To keep the consistency, I changed the variable instead. It would affect toolbar's padding with/without button groups.

modules/oxide/src/less/theme/components/view/view.less
```diff
@@ -1,4 +1,4 @@
-@view-header-padding: @pad-sm @pad-sm 0 @pad-sm;
+@view-header-padding: (@pad-sm + 2) (@pad-sm + 2) 2px (@pad-sm + 2);    // Add 2px to account for outline
```
TheSpyder, line 1: 2 is great until we change the outline 😂 

but I don't know if the outline width is a variable.

TheSpyder (approved):

shanmen-tiny (approved):

modules/oxide/src/less/theme/components/view/view.less
tiny-ben-tran: There are two variables that could be used: `@keyboard-focus-outline-width` and `@tinymce-border-width`. Decide to go with the first one because it is also used in style for focus outline on the buttons

PR #9594, opened 2024-04-26
TINY-10391: Fix open link context menu not enabled for image with embedded link

tiny-ben-tran:
Related Ticket: TINY-10391

Description of Changes:
* Updated the logic of getting links so the open link menu can be enabled for image with embedded link
* Moved some common assertions into `Utils.ts`

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/.changes/unreleased/tinymce-TINY-10391-2024-04-26.yaml
+++ b/.changes/unreleased/tinymce-TINY-10391-2024-04-26.yaml
@@ -1,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: Open link menu not enabled for image with embedded link.
+time: 2024-04-26T17:26:20.867461+09:30
+custom:
+  Issue: TINY-10391
--- a/.changes/unreleased/tinymce-TINY-10391-2024-04-26.yaml
+++ b/.changes/unreleased/tinymce-TINY-10391-2024-04-26.yaml
@@ -1,6 +1,6 @@
 project: tinymce
 kind: Fixed
-body: Open link menu not enabled for image with embedded link.
+body: Open link context menu action was not enabled for links on images.
 time: 2024-04-26T17:26:20.867461+09:30
 custom:
   Issue: TINY-10391
--- a/modules/tinymce/src/plugins/link/main/ts/core/Actions.ts
+++ b/modules/tinymce/src/plugins/link/main/ts/core/Actions.ts
@@ -6,9 +6,16 @@ import VK from 'tinymce/core/api/util/VK';
 import * as OpenUrl from './OpenUrl';
 import * as Utils from './Utils';
 
-const getLinks = (editor: Editor) => editor.selection.isCollapsed() ?
-  Utils.getLinks(editor.dom.getParents(editor.selection.getStart())) as [HTMLAnchorElement] :
-  Utils.getLinksInSelection(editor.selection.getRng());
+const isSelectionOnImageWithEmbeddedLink = (editor: Editor) => {
+  const rng = editor.selection.getRng();
+  const node = rng.startContainer;
+  // Handle a case where an image embedded with a link is selected
+  return Utils.isLink(node) && rng.startContainer === rng.endContainer && editor.dom.select('img', node).length === 1;
+};
+
+const getLinks = (editor: Editor) => editor.selection.isCollapsed() || isSelectionOnImageWithEmbeddedLink(editor)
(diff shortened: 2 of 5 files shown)


shanmen-tiny (approved):

modules/tinymce/src/plugins/link/test/ts/browser/OpenLinkContextMenuTest.ts
```diff
@@ -92,4 +92,18 @@ describe('browser.tinymce.plugins.link.OpenLinkContextMenuTest', () => {
       'https://www.one.com/'
     ]);
   });
+
+  it('TINY-10391: Should open a new tab when triggering open link on an image with embedded link', async () => {
+    const editor = hook.editor();
+    editor.setContent('<p><a href="https://www.exampleimage.com"><img src="http://www.example.image/image"></a></p>');
+    // Open link context menu on the link
+    await TinyUiActions.pTriggerContextMenu(editor, 'a[href="https://www.exampleimage.com"]', '.tox-silver-sink [role="menuitem"]');
```
ArvinJ-H, line 100: should you wait for editor to focus on the first item of the context menu before keydown?
`pTriggerContextMenu` only wait for the menu to show up, doesn't check for focus

ArvinJ-H (approved):

modules/tinymce/src/plugins/link/test/ts/browser/OpenLinkContextMenuTest.ts
tiny-ben-tran: Should not matter because the context menu handles the keydowns

.changes/unreleased/tinymce-TINY-10391-2024-04-26.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: Open link menu not enabled for image with embedded link.
```
TheSpyder, line 3: To keep the wording consistent with the related change in this release
```suggestion
body: Open link context menu action was not enabled for links on images.
```

modules/tinymce/src/plugins/link/main/ts/core/Actions.ts
```diff
@@ -6,9 +6,16 @@ import VK from 'tinymce/core/api/util/VK';
...
 import * as Utils from './Utils';
 
-const getLinks = (editor: Editor) => editor.selection.isCollapsed() ?
-  Utils.getLinks(editor.dom.getParents(editor.selection.getStart())) as [HTMLAnchorElement] :
-  Utils.getLinksInSelection(editor.selection.getRng());
+const isSelectionOnImageWithEmbeddedLink = (editor: Editor) => {
+  const rng = editor.selection.getRng();
+  const node = rng.startContainer;
+  // Handle a case where an image embedded with a link is selected
+  return Utils.isLink(node) && rng.startContainer === rng.endContainer && editor.dom.select('img', node).length === 1;
```
TheSpyder, line 13: I'm a little surprised `getLinksInSelection` doesn't work when this image select code does 🤔 

`Utils.isLink(node) && rng.startContainer === rng.endContainer` should be enough to pass something called "get links in selection". It's a little misleading for it to only consider the selection contents, not the selection itself.

Not saying we need to change it now - but something to consider.

TheSpyder (approved):

modules/tinymce/src/plugins/link/main/ts/core/Actions.ts
```diff
@@ -6,9 +6,16 @@ import VK from 'tinymce/core/api/util/VK';
...
+};
+
+const getLinks = (editor: Editor) => editor.selection.isCollapsed() || isSelectionOnImageWithEmbeddedLink(editor)
```
ltrouton, line 16: I am wondering if we can keep this more generic so that it covers any selectable node that is wrapped in an `a` tag

```suggestion
const getLinks = (editor: Editor) => editor.selection.isCollapsed() || (editor.selection.getStart() === editor.selection.getEnd())
```

ltrouton (approved):

modules/tinymce/src/plugins/link/main/ts/core/Actions.ts
tiny-ben-tran: It would fail a test case where it executes open link action on a link inside selection. In that case, `getLinks` should trigger `Utils.getLinksInSelection` instead

Issue #9598, opened 2024-04-29
block_formats options not translated if using documented approach


andrewnicols:
## 📝 Provide detailed reproduction steps (if any)
https://fiddle.tiny.cloud/Fug3lEC28d/0

1. Set a `block_formats` option of (exact copy from docs):
```js
Paragraph=p; Heading 1=h1; Heading 2=h2; Heading 3=h3; Heading 4=h4; Heading 5=h5; Heading 6=h6;
```
1. Set language to German
2. Ensure that the `format` menubar is present
3. Open the Format menu -> Blocks

### ✔️ Expected result
All headings translated

### ❌ Actual result
Only paragraph is translated. Others are not

### ❓ Possible solution
I believe the following patch will solve the issue:
```diff
diff --git a/modules/tinymce/src/themes/silver/main/ts/ui/core/complex/SelectDatasets.ts b/modules/tinymce/src/themes/silver/main/ts/ui/core/complex/SelectDatasets.ts
index 01de005aa2..39a7892112 100644
--- a/modules/tinymce/src/themes/silver/main/ts/ui/core/complex/SelectDatasets.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/ui/core/complex/SelectDatasets.ts
@@ -26,7 +26,7 @@ const process = (rawFormats: string[]): BasicSelectItem[] => Arr.map(rawFormats,
   // Allow text=value block formats
   const values = item.split('=');
   if (values.length > 1) {
-    title = values[0];
+    title = values[0].trim();
     format = values[1];
   }
 
```

I haven't worked out where to write a unit test for this but happy to provide one.

## 📃 Other details
- Browser: Any
- OS: Any.
- First affected version: Unsure
- Worked in version: Unsure

---

PR #9599, opened 2024-04-29
Trim titles in basic settings (#9598)

andrewnicols:
The `buildBasicSettingsDataset` call is used for:
- `block_formats`
- `font_family_formats`
- `font_size_formats`

The documented values for both `block_formats` and `font_family_formats` include a space between the `;` and the title, for example the default value in `block_formats` is:

```
Paragraph=p; Heading 1=h1; Heading 2=h2; Heading 3=h3; Heading 4=h4; Heading 5=h5; Heading 6=h6;
```

However the `split()` method in `SelectDatasets.ts` splits on the delimter, leading to values such as ` Heading 1=h1` which are then passed into the `process()` method which just split them on `=` leading to a title of ` Heading 1` (with leading whitespace) which does not match any valid translation.

This patch modifies the `split()` method to trim the title component.

Happy to add tests if you can tell me where best to add them. I had a look but couldn't find unit tests for any of this functionality.

Related Ticket: 

Description of Changes:
* Placeholder text

Pre-checks:
* [ ] Changelog entry added

Review:
* [ ] Milestone set

#9598


andrewnicols:
Note: I considered whether the docs should be fixed instead, but there are existing configurations out there which are broken if the language is changed.

TheSpyder:
Thank you for the PR! Looks like the best test we have is one for `font_family_formats`, which coincidentally had to trim the string before comparing it to the expected. Your fix should make that trim unnecessary?

https://github.com/tinymce/tinymce/blob/c47600fc916f20c2ee96e7b9cfb7f36ebab37b6a/modules/tinymce/src/core/test/ts/browser/FontSelectCustomTest.ts#L25

Ideally we'd have a smaller test for `SelectDatasets` in `modules/tinymce/src/themes/silver/test/ts/atomic` somewhere but that would likely require refactoring `SelectDatasets` so the test doesn't need a full editor just to get the option value to process.

TheSpyder:
For the changelog requirement, we use [changie](https://changie.dev/) - you don't need it installed, just run `yarn changie new` and select `tinymce` as the project. It will ask for a `TINY-` code, that's from our internal JIRA, just use the github issue number and replace the `TINY-` prefix in the file it creates with `GH-` before committing.

EkimChau:
Hello [andrewnicols](https://github.com/andrewnicols),

Thank you for your code contribution, in order for Tiny Inc to accept your code, we will need permission to use your code contribution, this can be done through the Tiny Contributor License Agreement (Tiny CLA). Please contact [legal@tiny.cloud](mailto:legal@tiny.cloud) to sign the Tiny CLA agreement.

andrewnicols (commented):
Thanks folks,

@ekimChau I have pinged legal for information on signing the CLA

@TheSpyder:

I've created the changelog entry with changie
I've removed the existing Trim from `modules/tinymce/src/core/test/ts/browser/FontSelectCustomTest.ts` and confirmed that the test fails before this change, and passes afterwards.

I've also pulled out the relevant parts of SelectDatasets.ts into a new ./utils/Select.ts location whcih seemed to be the most appropriate location, and written unit tests for this.

In doing so I made some additional changes to better handle invalid configuration, including:
- spaces before around the value; and multiple spaces in a space-delimited string.

TypeScript isn't my daily langauge so I do have a question regarding the use of the `Delimiters.SemiColon`.  In the unit test I wanted to use the literal semicolon (`;`), but tsc complains because it doens't match the Delimiters. What should I do here?

Thanks,

Andrew

andrewnicols:
Oh, and I've also switched from `title.trim()` to using `Strings.trim(title)` which seems to be more in-keeping with the rest of the codebase.

TheSpyder:
> TypeScript isn't my daily langauge so I do have a question regarding the use of the `Delimiters.SemiColon`. In the unit test I wanted to use the literal semicolon (`;`), but tsc complains because it doens't match the Delimiters. What should I do here?

You can't use literal strings, it's designed that way. You must reference the enum directly - which it looks like you did.

modules/tinymce/src/themes/silver/main/ts/ui/core/complex/SelectDatasets.ts
```diff
@@ -21,47 +14,24 @@ export interface AdvancedSelectDataset extends UiFactoryBackstageForStyleFormats
...
+    data,
   };
 };
 
 export {
   buildBasicSettingsDataset,
-  buildBasicStaticDataset
+  buildBasicStaticDataset,
+  Delimiter,
+  BasicSelectItem
```
TheSpyder, line 36: Since the refactored code is in the core, not moved to tests, you can just change where code that imports `Delimiter ` and `BasicSelectItem ` to `utils/Select` rather than re-export it here? There doesn't look like too many of them.

Alternatively, we do have places that export methods from modules only for testing (with a comment to that effect). So you could undo the refactoring if you prefer.

modules/tinymce/src/themes/silver/main/ts/ui/core/complex/utils/Select.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export enum Delimiter {
+  SemiColon = ';',
+  Space = ' '
+}
+
+const process = (rawFormats: string[]): BasicSelectItem[] => Arr.map(rawFormats, (item) => {
+  let title = item, format = item;
+  // Allow text=value block formats
+  const values = Arr.filter(item.split('='), (value) => Strings.isNotEmpty(value));
```
TheSpyder, line 19: Nicely done - but that's not quite all the edge cases. We do use `Strings` most of the time, but I think what would be best here is to filter out strings that are empty or contain only whitespace. Maybe someone makes a typo like `Heading = = h1;` (that's probably a good test to add, actually).

Code to check whitespace is in a different module for.... reasons.
https://github.com/tinymce/tinymce/blob/b4c1a77689f2cd7cdd65bd4456f7de405430f658/modules/tinymce/src/core/main/ts/text/Whitespace.ts#L5

modules/tinymce/src/themes/silver/test/ts/atomic/ui/core/complex/utils/SelectTest.ts
```diff
@@ -0,0 +1,47 @@
...
+
+import { Delimiter, makeBasicSelectItems } from 'tinymce/themes/silver/ui/core/complex/utils/Select';
+
+describe('atomic.tinymce.themes.silver.ui.core.complex.utils.Select', () => {
+  context('makeBasicSelectItems', () => {
+    it('Works with a semicolon delimiter', () => {
+      const config = 'Paragraph=p; Heading 1=h1;Heading 2=h2; Heading 3 =h3;Pre =pre';
+      const items = makeBasicSelectItems(config, Delimiter.SemiColon);
+      assert.isArray(items);
+      assert.lengthOf(items, 5);
```
TheSpyder, line 12: Length doesn't need to be checked if you're doing a `deepEqual` - the length will be part of that check already.

It should also print a much nicer error message than "expected length of 5 was 6" or whatever

TheSpyder (commented):
Looking pretty good. Please make sure you run `yarn lint` at some point - I can see what looks like one or two lint errors (we don't have public CI so that won't run until we merge it).

craig-helpshift:
Does anyone know by when this will be shipped?

PR #9605, opened 2024-04-30
TINY-10885: Keep focus on Image Dialog after image upload.

vpyshnenko:
Related Ticket: TINY-10885

Description of Changes:
* Set focus on the image Source input field after image upload

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/.changes/unreleased/tinymce-TINY-10885-2024-04-30.yaml
+++ b/.changes/unreleased/tinymce-TINY-10885-2024-04-30.yaml
@@ -1,6 +1,6 @@
 project: tinymce
 kind: Fixed
-body: Edit image modal did lose focus after image upload.
+body: Insert/Edit image dialog lost focus after the image upload completed.
 time: 2024-04-30T16:02:20.600725+03:00
 custom:
   Issue: TINY-10885
--- a/.changes/unreleased/tinymce-TINY-10885-2024-04-30.yaml
+++ b/.changes/unreleased/tinymce-TINY-10885-2024-04-30.yaml
@@ -1,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: Edit image modal did lose focus after image upload.
+time: 2024-04-30T16:02:20.600725+03:00
+custom:
+  Issue: TINY-10885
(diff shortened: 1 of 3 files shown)


.changes/unreleased/tinymce-TINY-10885-2024-04-30.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: Edit image modal did lose focus after image upload.
```
TheSpyder, line 3: "edit image" is a feature now, so we probably need to be more specific
```suggestion
body: Insert/Edit image dialog lost focus after the image upload completed.
```

TheSpyder (approved):

spocke (approved):

hamza0867 (approved):

.changes/unreleased/tinymce-TINY-10885-2024-04-30.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: Insert/Edit image dialog lost focus after the image upload completed.
```
hamza0867, line 3: I think we would usually use the `would` word in here so something like:
```suggestion
body: Insert/Edit image dialog would lose focus after the image upload completes.
```

TheSpyder: The only real requirement is that it's past tense - so what I wrote and your suggestion are both valid. It comes down to personal choice.

vpyshnenko: @TheSpyder, could you please revise this improvement as a native speaker?

TheSpyder: @vpyshnenko I prefer my suggestion, that's why I wrote it :)

If we think it needs rewording, how about
```suggestion
body: The Insert/Edit image dialog no longer loses focus after image uploads complete.
```

vpyshnenko: I'll stick your original one since it is in the past tense

PR #9616, opened 2024-05-03
TINY-10605: Added EditorOptions debug API

danoaky-tiny:
Related Ticket: TINY-10605

Description of Changes:
- Added a `debug` method to `EditorOptions`. It's functionally the same as what's described in the investigation ticket [TINY-10862](https://ephocks.atlassian.net/browse/TINY-10862).
- Added an extra optional parameter to `createOptions` to pass in the initial non-normalised options.
- `normalizeOptions` now doesn't modify its `options` param as a side effect.

Example:
```console
> tinymce.activeEditor.options.debug()
{
    "selector": "textarea.tinymce",
    "plugins": "table lists image accordion code",
    "toolbar": "table | numlist bullist | image | accordion | code",
    "menu": {
        "insert": {
            "title": "Insert",
            "items": "table | image | accordion"
        }
    },
    "details_initial_state": "inherited",
    "details_serialize_state": "inherited",
    "height": 500,
    "paste_as_text": true,
    "inline": null,
    "setup": "[object Function]",
    "encoding": "[object Undefined]",
    "protect": "[object RegExp]",
    "target": "[object HTMLDivElement]"
}

```

Pre-checks:
* [-] Changelog entry added
* [-] Branch prefixed with `feature/`, `hotfix/` or `spike/`

Review:
* [-] Milestone set

[TINY-10862]: https://ephocks.atlassian.net/browse/TINY-10862?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changes/unreleased/tinymce-TINY-10605-2024-05-03.yaml
+++ b/.changes/unreleased/tinymce-TINY-10605-2024-05-03.yaml
@@ -1,6 +1,6 @@
 project: tinymce
 kind: Added
-body: Added `options.debug` API
+body: Added `options.debug` API that logs the initial raw editor options to console.
 time: 2024-05-03T14:26:42.74417921+10:00
 custom:
   Issue: TINY-10605
--- a/.changes/unreleased/tinymce-TINY-10605-2024-05-03.yaml
+++ b/.changes/unreleased/tinymce-TINY-10605-2024-05-03.yaml
@@ -1,0 +1,6 @@
+project: tinymce
+kind: Added
+body: Added `options.debug` API
+time: 2024-05-03T14:26:42.74417921+10:00
+custom:
+  Issue: TINY-10605
--- a/modules/tinymce/src/core/main/ts/api/EditorOptions.ts
+++ b/modules/tinymce/src/core/main/ts/api/EditorOptions.ts
@@ -152,7 +152,6 @@ export interface Options {
    * Logs the initial raw editor options to the console.
    *
    * @method debug
-   * @returns {void}
    */
   debug: () => void;
 }
--- a/modules/tinymce/src/core/main/ts/api/EditorOptions.ts
+++ b/modules/tinymce/src/core/main/ts/api/EditorOptions.ts
@@ -147,6 +147,14 @@ export interface Options {
    * @return {Boolean} True if the option has a value set, otherwise false.
    */
   isSet: (name: string) => boolean;
+
+  /**
+   * Logs the initial raw editor options to the console.
+   *
+   * @method debug
+   * @returns {void}
+   */
+  debug: () => void;
 }
 
 // A string array allows comma/space separated values as well for ease of use
@@ -224,7 +232,7 @@ const processDefaultValue = <T, U>(name: string, defaultValue: T, processor: Pro
   return undefined;
 };
 
-const create = (editor: Editor, initialOptions: Record<string, unknown>): Options => {
+const create = (editor: Editor, initialOptions: Record<string, unknown>, rawInitialOptions: Record<string, unknown> = initialOptions): Options => {
   const registry: Record<string, OptionSpec<any, any>> = {};
   const values: Record<string, any> = {};
 
@@ -294,13 +302,38 @@ const create = (editor: Editor, initialOptions: Record<string, unknown>): Option
   const isSet = (name: string) =>
     Obj.has(values, name);
 
+  const debug = () => {
+    try {
+      // eslint-disable-next-line no-console
+      console.log(
+        JSON.parse(JSON.stringify(rawInitialOptions, (_key, value: unknown) => {
+          if (
+            Type.isBoolean(value) ||
+            Type.isNumber(value) ||
+            Type.isString(value) ||
+            Type.isNull(value) ||
+            Type.isArray(value) ||
+            Type.isPlainObject(value)
+          ) {
+            return value;
+          }
+          return Object.prototype.toString.call(value);
+        }))
+      );
+    } catch (error) {
+      // eslint-disable-next-line no-console
+      console.error(error);
+    }
+  };
+
   return {
     register,
     isRegistered,
     get,
     set,
     unset,
-    isSet
+    isSet,
+    debug,
   };
 };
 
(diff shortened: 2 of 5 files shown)


danoaky-tiny:
Unsure if this needed tests. I could wrap `console.log` & `console.error` and test against that. Let me know.

modules/tinymce/src/core/main/ts/api/EditorOptions.ts
```diff
@@ -147,6 +147,14 @@ export interface Options {
    * @return {Boolean} True if the option has a value set, otherwise false.
    */
   isSet: (name: string) => boolean;
+
+  /**
+   * Logs the initial raw editor options to the console.
+   *
+   * @method debug
+   * @returns {void}
```
spocke, line 155: I think we just omit the return if it's void.

.changes/unreleased/tinymce-TINY-10605-2024-05-03.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Added
+body: Added `options.debug` API
```
spocke, line 3:
```suggestion
body: Added `options.debug` API.
```
Or maybe something like:
```suggestion
body: Added `options.debug` API that logs the initial raw editor options to console.
```
I guess we flush it out in the release notes anyway.

spocke (approved):
I think we should have tests on this I think we do some console.log monkey patching for some other things seems like this is something want to do here as well since we might need to tweak the output.

TheSpyder:
I agree, we will use this in support so we need to make sure it continues to include everything we expect it to.

danoaky-tiny:
Added tests @TheSpyder @spocke

modules/tinymce/src/core/test/ts/browser/EditorOptionsDebugTest.ts
```diff
@@ -0,0 +1,103 @@
...
+import Editor from 'tinymce/core/api/Editor';
+
+describe('browser.tinymce.core.EditorOptionsDebugTest', () => {
+  const createHook = (options: Record<string, unknown>) =>
+    TinyHooks.bddSetupLight<Editor>({
+      ...options,
+      base_url: '/project/tinymce/js/tinymce',
+    }, []);
+
+  const METHODS = [ 'log', 'error' ] as const;
```
spocke, line 16: Seems overly complicated. Should be possible to test this with just:
```
const log: any[] = [];
const orgConsoleLog = console.log;
console.log = (...args) => log.push(args);
editor.options.dump();
console.log = orgConsoleLog;
assert.deepEqual(log, [['foo']])
```

modules/tinymce/src/core/test/ts/browser/EditorOptionsDebugTest.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  context('With empty options', () => {
+    const hook = createHook({});
+    it('TINY-10605: should log empty object', () => {
+      hook.editor().options.debug();
+      assertLastLog({});
+    });
+  });
+
+  context('With function', () => {
```
ltrouton, line 66: I know we can't be exhaustive but could we test setting a random option to `undefined`, `null`, Regexp, and a DOM node. That should cover the majority of the possible option types I would think.

ltrouton (approved):

modules/tinymce/src/core/test/ts/browser/EditorOptionsDebugTest.ts
danoaky-tiny: Added test coverage for undefined, null, RegExp and a div element.

modules/tinymce/src/core/test/ts/browser/EditorOptionsDebugTest.ts
```diff
@@ -0,0 +1,106 @@
...
+      ...options,
+      base_url: '/project/tinymce/js/tinymce',
+    }, []);
+  const ignoredKeys = new Set([ 'promotion', 'toolbar', 'menubar', 'statusbar', 'base_url', 'selector', 'setup' ]);
+  const assertDebugLog = (editor: Editor, expectedLog: unknown) => {
+    const logs: unknown[] = [];
+    const originalLog = console.log;
+    console.log = (arg: unknown) => logs.push(arg);
+    editor.options.debug();
+    console.log = originalLog;
```
TheSpyder, line 22: Please put some sort of try/finally block here, so if `.debug()` throws an exception it doesn't break `console.log` for the entire page

TheSpyder (approved):

modules/tinymce/src/core/test/ts/browser/EditorOptionsDebugTest.ts
spocke: We have a few spots of these types of collections maybe we should add a transact type collect function in agar. Then we would be able to add that in a central place.

TheSpyder:
I must've submitted my review seconds before it was merged, the timestamp still says my comment was before it merged even though it's below the merge in the timeline 😂

modules/tinymce/src/core/test/ts/browser/EditorOptionsDebugTest.ts
TheSpyder: I was expecting an `after` hook to restore the global values, which is exactly what would be good to do with that sort of collect function

spocke: That would be really nice something like `const { getMessages, reset } = ConsoleHooks.logs({ levels: [ 'error', 'info' ] })` or something.

spocke: Maybe a training ticket we could make out of that seems simple enough.

danoaky-tiny: @TheSpyder  The debug method itself has a try catch encompassing the whole function, so it should be fine as for how it's implemented right now.
I had an after hook in a previous commit, but simplified it later.
Something in agar sounds good to me.

PR #9621, opened 2024-05-07
TINY-10840: Fixed errors thrown by ForceBlocks and element selections

spocke:
Related Ticket: TINY-10840

Description of Changes:
* Added a new way to store/restore selections when wrapping elements in the root so that it doesn't throw if the element indexes doesn't match after it's been wrapped.
* Had to add special code for Firefox that does strange things when the body is empty or filled.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/test/ts/browser/delete/DivDeleteTest.ts
+++ b/modules/tinymce/src/core/test/ts/browser/delete/DivDeleteTest.ts
@@ -1,31 +1,92 @@
-import { before, describe, it } from '@ephox/bedrock-client';
-import { PlatformDetection } from '@ephox/sand';
-import { TinyAssertions, TinyHooks, TinySelections } from '@ephox/wrap-mcagar';
+import { Keys } from '@ephox/agar';
+import { context, describe, it } from '@ephox/bedrock-client';
+import { TinyAssertions, TinyContentActions, TinyHooks, TinySelections } from '@ephox/wrap-mcagar';
+import { assert } from 'chai';
 
 import Editor from 'tinymce/core/api/Editor';
 
 describe('browser.tinymce.core.delete.DivDeleteTest', () => {
-  const isFirefox = PlatformDetection.detect().browser.isFirefox();
-
   const hook = TinyHooks.bddSetup<Editor>({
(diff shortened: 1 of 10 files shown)


modules/tinymce/src/core/test/ts/browser/delete/DivDeleteTest.ts
```diff
@@ -0,0 +1,31 @@
...
+describe('browser.tinymce.core.delete.DivDeleteTest', () => {
+  const isFirefox = PlatformDetection.detect().browser.isFirefox();
+
+  const hook = TinyHooks.bddSetup<Editor>({
+    base_url: '/project/tinymce/js/tinymce',
+    indent: false
+  }, [], true);
+
+  before(function () {
+    // Firefox doesn't unwrap DIV elements if you delete inside them and it's the only child element in the BODY
```
TheSpyder, line 16: Don't we want our behaviour to be as similar as possible across browsers?

TheSpyder (approved):

modules/tinymce/src/core/test/ts/browser/delete/DivDeleteTest.ts
spocke: Yes at some point I would want to create a custom delete not a bunch of patches on the native delete. Not sure this is the time to override this as well. With the method to are applying now it's hard to know if you catched all the scenarios. It will be a bigger task to rewrite this to custom code.

spocke: I guess maybe we could first detect what the block is then execute native delete then wrap it back in that block if it was deleted. That would solve this particular case.

spocke: Yeah that approach seems to work but yes another patch over the delete. 😄

TheSpyder (approved):
Very nice!

lorenzo-pomili (approved):

mikkorantalainen:
Will this be fixed in 6.x branch?

PR #9636, opened 2024-05-09
TINY-10854: Now insert format also manage `nbsp`s correctly

lorenzo-pomili:
Related Ticket: TINY-10854

Description of Changes:
Since to normalize the `nbsp`s we use [normalizeNbsps](https://github.com/tinymce/tinymce/blob/main/modules/tinymce/src/core/main/ts/keyboard/Nbsps.ts#L179) but this mechanism is based on the text elements and to fix our case we would need to find the parent and parse all the children element which could require way more time.

Since when we add a format we apply a [CaretFormat](https://github.com/tinymce/tinymce/blob/main/modules/tinymce/src/core/main/ts/fmt/CaretFormat.ts#L168), so to fix this case my idea is to apply an `nbsp` normalization with on `input` (like for `normalizeNbspsInEditor` in [InputKeys](https://github.com/tinymce/tinymce/blob/main/modules/tinymce/src/core/main/ts/keyboard/InputKeys.ts)), but only for the next insertion (so using `editor.once`) since this is needed only after the first insertion.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/.changes/unreleased/tinymce-TINY-10854-2024-05-10.yaml
+++ b/.changes/unreleased/tinymce-TINY-10854-2024-05-10.yaml
@@ -1,6 +1,6 @@
 project: tinymce
 kind: Fixed
-body: 'Extra unnecessary nbsp entities was inserted before and after inline elements created by input of formatted text.'
+body: 'Unnecessary nbsp entities were inserted when typing at the edges of inline elements.'
 time: 2024-05-10T11:38:17.292535402+02:00
 custom:
   Issue: TINY-10854
--- a/.changes/unreleased/tinymce-TINY-10854-2024-05-10.yaml
+++ b/.changes/unreleased/tinymce-TINY-10854-2024-05-10.yaml
@@ -1,0 +1,7 @@
+project: tinymce
+kind: Fixed
+body: '`nbsp`s wasn''t replaced correctly if the next character is wrapped in an inline
+  element.'
+time: 2024-05-10T11:38:17.292535402+02:00
+custom:
+  Issue: TINY-10854
--- a/.changes/unreleased/tinymce-TINY-10854-2024-05-10.yaml
+++ b/.changes/unreleased/tinymce-TINY-10854-2024-05-10.yaml
@@ -1,7 +1,6 @@
 project: tinymce
 kind: Fixed
-body: '`nbsp`s wasn''t replaced correctly if the next character is wrapped in an inline
-  element.'
+body: 'Extra unnecessary nbsp entities was inserted before and after inline elements created by input of formatted text.'
 time: 2024-05-10T11:38:17.292535402+02:00
 custom:
   Issue: TINY-10854
--- a/modules/tinymce/src/core/main/ts/fmt/CaretFormat.ts
+++ b/modules/tinymce/src/core/main/ts/fmt/CaretFormat.ts
@@ -1,5 +1,5 @@
 import { Arr, Fun, Obj, Optional, Strings, Type, Unicode } from '@ephox/katamari';
-import { Attribute, Insert, Remove, SugarElement, SugarNode } from '@ephox/sugar';
+import { Attribute, Insert, PredicateFind, Remove, SugarElement, SugarNode, Traverse } from '@ephox/sugar';
 
 import DomTreeWalker from '../api/dom/TreeWalker';
 import Editor from '../api/Editor';
@@ -165,6 +165,27 @@ const cleanFormatNode = (editor: Editor, caretContainer: Node, formatNode: Eleme
   }
 };
 
+const normalizeNbspsBetween = (editor: Editor, caretContainer: Node | null) => {
+  const fake = true;
+  if (fake) {
+    editor.once('input', (_e) => {
+      if (caretContainer && !editor.dom.isEmpty(caretContainer)) {
+        Traverse.prevSibling(SugarElement.fromDom(caretContainer)).each((node) => {
+          if (NodeType.isText(node.dom)) {
+            node.dom.data = node.dom.data.replace(Unicode.nbsp, ' ');
+          } else {
+            PredicateFind.descendant(node, (e) => NodeType.isText(e.dom)).each((textNode) => {
+              if (NodeType.isText(textNode.dom)) {
+                textNode.dom.data = textNode.dom.data.replace(Unicode.nbsp, ' ');
+              }
+            });
+          }
+        });
+      }
+    });
+  }
+};
+
 const applyCaretFormat = (editor: Editor, name: string, vars?: FormatVars): void => {
   let caretContainer: Node | null;
   const selection = editor.selection;
@@ -210,6 +231,8 @@ const applyCaretFormat = (editor: Editor, name: string, vars?: FormatVars): void
 
       selectionRng.insertNode(caretContainer);
       offset = 1;
+
+      normalizeNbspsBetween(editor, caretContainer);
 
       editor.formatter.apply(name, vars, caretContainer);
     } else {
@@ -296,6 +319,8 @@ const removeCaretFormat = (editor: Editor, name: string, vars?: FormatVars, simi
       removeCaretContainerNode(editor, caretContainer, Type.isNonNullable(caretContainer));
     }
     selection.setCursorLocation(caretTextNode, 1);
+    // TODO: is this correct that we use caretTextNode.parentElement instead of direct go to the correct element?
+    normalizeNbspsBetween(editor, caretTextNode.parentElement);
 
     if (dom.isEmpty(formatNode)) {
       dom.remove(formatNode);
--- a/modules/tinymce/src/core/main/ts/fmt/CaretFormat.ts
+++ b/modules/tinymce/src/core/main/ts/fmt/CaretFormat.ts
@@ -169,7 +169,7 @@ const normalizeNbsps = (node: SugarElement<Text>) => SugarText.set(node, SugarTe
 
 const normalizeNbspsBetween = (editor: Editor, caretContainer: Node | null) => {
   const handler = () => {
-    if (caretContainer && !editor.dom.isEmpty(caretContainer)) {
+    if (caretContainer !== null && !editor.dom.isEmpty(caretContainer)) {
       Traverse.prevSibling(SugarElement.fromDom(caretContainer)).each((node) => {
         if (SugarNode.isText(node)) {
           normalizeNbsps(node);
--- a/modules/tinymce/src/core/main/ts/fmt/CaretFormat.ts
+++ b/modules/tinymce/src/core/main/ts/fmt/CaretFormat.ts
@@ -166,8 +166,8 @@ const cleanFormatNode = (editor: Editor, caretContainer: Node, formatNode: Eleme
 };
 
 const normalizeNbspsBetween = (editor: Editor, caretContainer: Node | null) => {
-  editor.once('input', (e) => {
-    if (!e.isComposing && caretContainer && !editor.dom.isEmpty(caretContainer)) {
+  const handler = () => {
+    if (caretContainer && !editor.dom.isEmpty(caretContainer)) {
       Traverse.prevSibling(SugarElement.fromDom(caretContainer)).each((node) => {
         if (NodeType.isText(node.dom)) {
           node.dom.data = node.dom.data.replace(Unicode.nbsp, ' ');
@@ -180,6 +180,11 @@ const normalizeNbspsBetween = (editor: Editor, caretContainer: Node | null) => {
         }
       });
     }
+  };
+  editor.once('input', (e) => {
+    !e.isComposing ? handler() : editor.once('compositionend', () => {
+      handler();
+    });
   });
 };
 
--- a/modules/tinymce/src/core/main/ts/fmt/CaretFormat.ts
+++ b/modules/tinymce/src/core/main/ts/fmt/CaretFormat.ts
@@ -166,8 +166,8 @@ const cleanFormatNode = (editor: Editor, caretContainer: Node, formatNode: Eleme
 };
 
 const normalizeNbspsBetween = (editor: Editor, caretContainer: Node | null) => {
-  editor.once('input', (_e) => {
-    if (caretContainer && !editor.dom.isEmpty(caretContainer)) {
+  editor.once('input', (e) => {
+    if (!e.isComposing && caretContainer && !editor.dom.isEmpty(caretContainer)) {
       Traverse.prevSibling(SugarElement.fromDom(caretContainer)).each((node) => {
         if (NodeType.isText(node.dom)) {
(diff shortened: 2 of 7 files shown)


modules/tinymce/src/core/main/ts/fmt/CaretFormat.ts
```diff
@@ -165,6 +165,29 @@ const cleanFormatNode = (editor: Editor, caretContainer: Node, formatNode: Eleme
   }
 };
 
+const normalizeNbspsBetween = (editor: Editor, caretContainer: Node | null) => {
+  const handler = () => {
+    if (caretContainer && !editor.dom.isEmpty(caretContainer)) {
+      Traverse.prevSibling(SugarElement.fromDom(caretContainer)).each((node) => {
+        if (NodeType.isText(node.dom)) {
```
spocke, line 172: There is a sugar predicate for this `SugarNode.isText`.

modules/tinymce/src/core/main/ts/fmt/CaretFormat.ts
```diff
@@ -165,6 +165,29 @@ const cleanFormatNode = (editor: Editor, caretContainer: Node, formatNode: Eleme
...
+          node.dom.data = node.dom.data.replace(Unicode.nbsp, ' ');
```
spocke, line 173: You can use sugars Text.get/set here I think.

modules/tinymce/src/core/main/ts/fmt/CaretFormat.ts
```diff
@@ -165,6 +165,29 @@ const cleanFormatNode = (editor: Editor, caretContainer: Node, formatNode: Eleme
...
+            if (NodeType.isText(textNode.dom)) {
+              textNode.dom.data = textNode.dom.data.replace(Unicode.nbsp, ' ');
+            }
+          });
+        }
+      });
+    }
+  };
+  editor.once('input', (e) => {
+    !e.isComposing ? handler() : editor.once('compositionend', () => {
```
spocke, line 185: Use an if statement I think we should only use ternaries for expressions that return something other than void.

.changes/unreleased/tinymce-TINY-10854-2024-05-10.yaml
```diff
@@ -0,0 +1,7 @@
+project: tinymce
+kind: Fixed
+body: '`nbsp`s wasn''t replaced correctly if the next character is wrapped in an inline
+  element.'
```
spocke, line 4: Very hard do describe this, my attempt. 😄 

```suggestion
body: 'Extra unnecessary nbsp entities was inserted before and after inline elements created by input of formatted text.'
```

spocke (approved):
- [ ] Fix typos in PR title npsp -> nbsp and correclty -> correctly

modules/tinymce/src/core/main/ts/fmt/CaretFormat.ts
lorenzo-pomili: thanks, I changed them

modules/tinymce/src/core/main/ts/fmt/CaretFormat.ts
lorenzo-pomili: changed using: `SugarText.get/set`

modules/tinymce/src/core/main/ts/fmt/CaretFormat.ts
lorenzo-pomili: changed to a normal if, I did with a ternary operator just to make it less "ugly", but in the end is "ugly" anyway, so switch to if is a good idea :sweat_smile:

.changes/unreleased/tinymce-TINY-10854-2024-05-10.yaml
lorenzo-pomili: ok, changed with yours

modules/tinymce/src/core/main/ts/fmt/CaretFormat.ts
TheSpyder: I thought TypeScript could be configured to error on this. Maybe it was eslint.

.changes/unreleased/tinymce-TINY-10854-2024-05-10.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: 'Extra unnecessary nbsp entities was inserted before and after inline elements created by input of formatted text.'
```
TheSpyder, line 3:
```suggestion
body: Unnecessary nbsp entities were inserted when typing at the edges of inline elements.
```

modules/tinymce/src/core/main/ts/fmt/CaretFormat.ts
```diff
@@ -165,6 +165,35 @@ const cleanFormatNode = (editor: Editor, caretContainer: Node, formatNode: Eleme
   }
 };
 
+const normalizeNbsps = (node: SugarElement<Text>) => SugarText.set(node, SugarText.get(node).replace(Unicode.nbsp, ' '));
+
+const normalizeNbspsBetween = (editor: Editor, caretContainer: Node | null) => {
+  const handler = () => {
+    if (caretContainer && !editor.dom.isEmpty(caretContainer)) {
```
TheSpyder, line 172: I'm surprised that removing _every_ nbsp passes the tests. But our tests for NBSP normalisation probably don't cover this new `caretContainer` condition well enough. Speaking of which...
```suggestion
    if (caretContainer !== null && !editor.dom.isEmpty(caretContainer)) {
```

Try something like `<p>text <strong>&nbsp;more text</strong></p>`. Your code will probably remove the nbsp, which is wrong.

- [ ] do some wider checks of this logic

lorenzo-pomili: ok, I'm writing the logic for it, I'd like to make an abstraction starting from [Nbsps.ts](https://github.com/tinymce/tinymce/blob/main/modules/tinymce/src/core/main/ts/keyboard/Nbsps.ts) to use it in both cases, but I'm not sure how possible is since it's strong relation with the elements of the DOM.

Since the logic in `Nbsps.ts` isn't working for this cases, what logic do we want to implement?
by now we have these tests: [NbspsTest.ts](https://github.com/tinymce/tinymce/blob/main/modules/tinymce/src/core/test/ts/atomic/keyboard/NbspsTest.ts), should I make these cases work considering that instead of a simple text that could be an inline element?

spocke: I would break out the everything inside the isEmpty check. Then make a separate function that takes a target node that checks the sibling before target, first child of target, last child of target, sibling after target. Then checks if any of those nodes has a non whitespace character followed by a nbsp since then you can replace that nbsp with a space.
Maybe feed those siblings into an array if that makes things easier. We might need to expand this in the future for other cases like `<strong>foo&nbsp;<span>bar</span>` etc.
Maybe it makes sense to stick that new function into the Nbsps module?

spocke (approved):

.changes/unreleased/tinymce-TINY-10854-2024-05-10.yaml
lorenzo-pomili: done

modules/tinymce/src/core/test/ts/webdriver/keyboard/SpaceKeyTest.ts
```diff
@@ -151,5 +169,45 @@ describe('webdriver.tinymce.core.keyboard.SpaceKeyTest', () => {
         TinyAssertions.assertContent(editor, '<p>a c</p>');
       }
     });
+
+    it('TINY-10854: `&nbsp;`s should be converted to spaces when bedore or after there is an inline element', async () => {
```
spocke, line 173:
```suggestion
    it('TINY-10854: `&nbsp;`s should be converted to spaces when before or after there is an inline element', async () => {
```

modules/tinymce/src/core/test/ts/webdriver/keyboard/SpaceKeyTest.ts
```diff
@@ -151,5 +169,45 @@ describe('webdriver.tinymce.core.keyboard.SpaceKeyTest', () => {
...
+      await RealKeys.pSendKeysOn('iframe => body', [ RealKeys.text(' ') ]);
+
+      removeCaretFormat(editor, 'bold', {});
+      await RealKeys.pSendKeysOn('iframe => body', [ RealKeys.text(' ') ]);
+      await RealKeys.pSendKeysOn('iframe => body', [ RealKeys.text('c') ]);
+      await RealKeys.pSendKeysOn('iframe => body', [ RealKeys.text(' ') ]);
+      TinyAssertions.assertContent(editor, '<p>a <strong>b </strong> c&nbsp;</p>');
+    });
+
+    it('TINY-10854: `&nbsp;`s should be converted to spaces when bedore or after there is an inline element (with composition)', async () => {
```
spocke, line 190:
```suggestion
    it('TINY-10854: `&nbsp;`s should be converted to spaces when before or after there is an inline element (with composition)', async () => {
```

spocke (approved):

modules/tinymce/src/core/test/ts/webdriver/keyboard/SpaceKeyTest.ts
lorenzo-pomili: changed

modules/tinymce/src/core/test/ts/webdriver/keyboard/SpaceKeyTest.ts
lorenzo-pomili: changed

modules/tinymce/src/core/main/ts/fmt/CaretFormat.ts
TheSpyder: Typing in TinyMCE won't reproduce all the edge cases because we do normalisation on space.

Browsers don't take inline tags into account when collapsing whitespace. If there are two spaces in a row, and one of them isn't an nbsp, it will render as one space.

lorenzo-pomili: at the moment the `handler` is triggered only on typing to not have performance issues, I'll try to check where the `Nbsps.ts` functions are used and see it is possible somehow distinguish if we need to also check for siblings

Afraithe: @TheSpyder that seems out of scope for this issue reported by client? The scenario outlined cause a great deal of &nbsp's if you typed in the certain way, that is the issue we want to fix. There can be other edge cases, but also cases where you might want nbsp's, we should be careful expanding on this to much imo.

TheSpyder: This change removes all `nbsp` characters in the previous sibling of the cursor. I'm not the one who expanded the scope, I'm simply responding with potential issues given that new scope.

When I said this
> Typing in TinyMCE won't reproduce all the edge cases because we do normalisation on space.

I didn't mean to imply the fix should run more often than when typing. Only that the scenarios it has to deal with include things that can't be created within the editor. Users copy and paste content from all sorts of places.

lorenzo-pomili: wait about the remove of all `nbsp` last version of the code shouldn't do it, for the paste part, the behavior is different between simple text and text with inline elements at the moment, I'll try to understand how to solve it

modules/tinymce/src/core/main/ts/fmt/CaretFormat.ts
```diff
@@ -165,6 +166,37 @@ const cleanFormatNode = (editor: Editor, caretContainer: Node, formatNode: Eleme
...
+      Traverse.prevSibling(SugarElement.fromDom(caretContainer)).each((node) => {
+        if (SugarNode.isText(node)) {
+          normalizeNbsps(node);
+        } else {
+          PredicateFind.descendant(node, (e) => SugarNode.isText(e)).each((textNode) => {
+            if (SugarNode.isText(textNode)) {
+              normalizeNbsps(textNode);
+            }
+          });
+        }
```
TheSpyder, line 183: This is the code that removes every nbsp in the previous sibling, and is what I was concerned about in the other thread.

modules/tinymce/src/core/main/ts/keyboard/Nbsps.ts
```diff
@@ -203,11 +203,75 @@ const normalizeNbspsInEditor = (editor: Editor): void => {
...
+const startsWithSpaceOrNbsp = (text: string) => /^(&nbsp;|\s)(.*?)/.test(text);
+const endsWithSpaceOrNbsp = (text: string) => /(.*?)(&nbsp;|\s)$/.test(text);
+
+const endsWithDoubleSpaceOrNbsp = (text: string) => /^(.*?)(&nbsp;|\s){2}$/.test(text);
+
+const startsWithDoubleSpaceOrNbsp = (text: string) => /^(&nbsp;|\s){2}(.*?)$/.test(text);
+
+const normalizeNbspWithElements = (content: string, schema: Schema): string => {
+  const wrappedContent = document.createElement('div');
+  wrappedContent.innerHTML = content;
```
TheSpyder, line 215: This doesn't seem like a very performant way to implement filtering. We already have filter processing for paste after content has been parsed into an AST node tree (the logic that `paste_postprocess` uses).

TheSpyder (commented):
I'm not sure how this got so off track, or how my comments lead to the huge new chunk of code (which, by the way, despite being in the keyboard nbsp folder is only used by the paste code).

I'm sorry for not communicating my concerns more clearly. Perhaps I should take a step back from reviews - this one in particular - and let someone else who is better able to focus on the high level problem here guide you to the best solution.

modules/tinymce/src/core/main/ts/fmt/CaretFormat.ts
lorenzo-pomili: but `normalizeNbsps` removes only the last `&nbsp;`, it use this replace: `.replace(new RegExp(`${Unicode.nbsp}$`), ' ')`

`'a\u00a0b\u00a0\u00a0'.replace(new RegExp(`${Unicode.nbsp}$`), ' ')` -> `'a\u00a0b\u00a0 '`

TheSpyder: Oh _that_ was what you did in the latest refactor. Sorry, github makes it really hard to see what changed between looking at a review.

modules/tinymce/src/core/main/ts/keyboard/Nbsps.ts
lorenzo-pomili: removed

TheSpyder (approved):

PR #9638, opened 2024-05-12
TINY-10892: Fix deleting from P with IMG into LI would create unwanted font-styles on Chrome

hamza0867:
Related Ticket:  TINY-10892

Description of Changes:
*  Fix deleting from P with IMG into LI would create unwanted font-styles on Chrome

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/.changes/unreleased/tinymce-TINY-10892-2024-05-12.yaml
+++ b/.changes/unreleased/tinymce-TINY-10892-2024-05-12.yaml
@@ -1,7 +1,6 @@
 project: tinymce
 kind: Fixed
-body: Backsapcing from a paragraph that has an `img` tag into a list would cause the
-  `img` tag to create unwanted inline styles
+body: Deleting into a list from a paragraph that has an `img` tag  could cause extra inline styles to be added
 time: 2024-05-12T23:00:56.388965+01:00
 custom:
   Issue: TINY-10892
--- a/.changes/unreleased/tinymce-TINY-10892-2024-05-12.yaml
+++ b/.changes/unreleased/tinymce-TINY-10892-2024-05-12.yaml
@@ -1,0 +1,7 @@
+project: tinymce
+kind: Fixed
+body: Backsapcing from a paragraph that has an `img` tag into a list would cause the
+  `img` tag to create unwanted inline styles
+time: 2024-05-12T23:00:56.388965+01:00
+custom:
+  Issue: TINY-10892
--- a/modules/tinymce/src/core/main/ts/util/Quirks.ts
+++ b/modules/tinymce/src/core/main/ts/util/Quirks.ts
@@ -465,7 +465,6 @@ const Quirks = (editor: Editor): Quirks => {
     const specialDelete = Cell(false);
     const content = Cell<Optional<string>>(Optional.none());
     editor.on('keydown', (e) => {
-      debugger;
       if (isDefaultPrevented(e) || e.key !== 'Delete') {
         return;
       }
--- a/modules/tinymce/src/core/main/ts/util/Quirks.ts
+++ b/modules/tinymce/src/core/main/ts/util/Quirks.ts
@@ -1,5 +1,4 @@
-import { Cell, Fun, Optional, Singleton } from '@ephox/katamari';
-import { PredicateFind, SugarElement } from '@ephox/sugar';
+import { Fun } from '@ephox/katamari';
 
 import Editor from '../api/Editor';
 import Env from '../api/Env';
@@ -9,7 +8,6 @@ import { EditorEvent } from '../api/util/EventDispatcher';
 import Tools from '../api/util/Tools';
 import VK from '../api/util/VK';
 import * as CaretContainer from '../caret/CaretContainer';
-import * as ElementType from '../dom/ElementType';
 import * as Empty from '../dom/Empty';
 import * as Rtc from '../Rtc';
 
@@ -33,7 +31,6 @@ const Quirks = (editor: Editor): Quirks => {
   const isWebKit = browser.isChromium() || browser.isSafari();
   const isiOS = Env.deviceType.isiPhone() || Env.deviceType.isiPad();
   const isMac = Env.os.isMacOS() || Env.os.isiOS();
-  const isChrome = browser.isChromium();
 
   /**
    * Executes a command with a specific state this can be to enable/disable browser editing features.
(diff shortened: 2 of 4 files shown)


modules/tinymce/src/core/test/ts/webdriver/delete/QuirksChromeTest.ts
```diff
@@ -0,0 +1,83 @@
...
+import Editor from 'tinymce/core/api/Editor';
+import Env from 'tinymce/core/api/Env';
+
+describe('webdriver.tinymce.core.delete.QuirksChromeTest', () => {
+  before(function () {
+    if (!Env.browser.isChromium() && !Env.browser.isSafari()) {
+      this.skip();
+    }
+  });
+
```
TheSpyder, line 14: Just because the fix is for specific browsers doesn't mean we should only test on those browsers. Based on your description I assume this behaviour works on firefox - we should make sure that continues to be the case.

`skip` should be used to avoid browsers with problems we don't want to deal with, not to write tests just for certain browsers. Every `skip` is tech debt we take on as a risk and probably should deal with one day.

```suggestion
```

modules/tinymce/src/core/main/ts/util/Quirks.ts
TheSpyder, line 1: What happened in this file? The lines don't seem long enough for eslint to reformat them - if that was a problem it would be failing the build.

.changes/unreleased/tinymce-TINY-10892-2024-05-12.yaml
```diff
@@ -0,0 +1,7 @@
+project: tinymce
+kind: Fixed
+body: Backsapcing from a paragraph that has an `img` tag into a list would cause the
+  `img` tag to create unwanted inline styles
```
TheSpyder, line 4:
```suggestion
body: Deleting into a list from a paragraph that has an `img` tag  could cause extra inline styles to be added
```

TheSpyder (changes requested):
I'll review it when the unnecessary format changes are reversed.

modules/tinymce/src/core/main/ts/util/Quirks.ts
hamza0867: I believe I tried to edit it using the Zed editor and it made all of those changes 🤷‍♂️

.changes/unreleased/tinymce-TINY-10892-2024-05-12.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: Deleting into a list from a paragraph that has an `img` tag  could cause extra inline styles to be added
```
spocke, line 3:
```suggestion
body: Deleting into a list from a paragraph that has an `img` tag could cause extra inline styles to be added.
```

modules/tinymce/src/core/main/ts/util/Quirks.ts
```diff
@@ -327,6 +329,160 @@ const Quirks = (editor: Editor): Quirks => {
...
+  const removeExtraFontFamilyOnKeyup = (editor: Editor, specialDelete: Cell<boolean>, content: Cell<Optional<string>>) => {
+    editor.on('keyup', (e) => {
+      if (isDefaultPrevented(e) || e.key !== 'Backspace' && e.key !== 'Delete' || !specialDelete.get()) {
+        return;
+      }
+
+      const rng = selection.getRng();
+      const container = rng.startContainer;
+      const root = dom.getRoot();
+      const parent = Optional.from(Traverse.parents(SugarElement.fromDom(container)).find((node) => node.dom.nodeName.toLowerCase() === 'li'));
```
spocke, line 343: You can use PredicateFind.ancestor or PredicateFind.closest here also this should likely be constrained to the root so it doesn't try to find parents outside the root element in inline mode.

modules/tinymce/src/core/main/ts/util/Quirks.ts
```diff
@@ -327,6 +329,160 @@ const Quirks = (editor: Editor): Quirks => {
...
+    });
+  };
+
+  /**
+   * Removes font-family style added when pressing backspace when the cursor is just before an image
+   * and there is a list before the image. #TINY-10892
+   */
+  const removeExtraFontFamilyOnBackspace = () => {
+    const specialDelete = Cell(false);
+    const content = Cell<Optional<string>>(Optional.none());
```
spocke, line 382: You can use a Singleton.value here.

modules/tinymce/src/core/main/ts/util/Quirks.ts
```diff
@@ -327,6 +329,160 @@ const Quirks = (editor: Editor): Quirks => {
...
+        parent = parent.parentNode;
+      }
+      let hasImgNode = false;
+      parent.childNodes.forEach((node) => {
+        if (node.nodeName.toLowerCase() === 'img') {
+          hasImgNode = true;
+          return;
+        }
+      });
+      if (!hasImgNode || parent.previousSibling?.nodeName.toLowerCase() !== 'ol') {
```
spocke, line 412: This doesn't work with `UL` should handle both cases. There is a ElementType.isList that takes sugar elements. Maybe if the code was all sugar then one could use that.

spocke (changes requested):
This fix adds some weird undo levels. I really should have looked into this earlier since I think we just need to add a `ElementType.isListItem` here: https://github.com/tinymce/tinymce/blob/main/modules/tinymce/src/core/main/ts/delete/BlockMergeBoundary.ts#L60 that is whats preventing the the merge blocks code from working with lists. There might be issues with merging list items into list items since some of that is overrided in the lists plugin so that you would outdent the list instead. But a complete override is better than trying to let the browser do it's thing, have a bunch of tracking states on that easier to test and easier to predict the behavior.

modules/tinymce/src/core/main/ts/util/Quirks.ts
hamza0867: Done!

modules/tinymce/src/core/main/ts/util/Quirks.ts
hamza0867: Done!

modules/tinymce/src/core/main/ts/util/Quirks.ts
hamza0867: Done!

modules/tinymce/src/core/test/ts/webdriver/delete/QuirksChromeTest.ts
hamza0867: Done!

spocke (approved):

lorenzo-pomili (approved):

modules/tinymce/src/core/test/ts/webdriver/delete/QuirksChromeTest.ts
TheSpyder, line 1: It's no longer really a chrome test, nor it is a quirks test 🤔 

I'm happy to keep the test, just rename it (and remember the describe call includes the name).

TheSpyder (approved):
Well! That was quite a lesson in learning what the editor is capable of before building something that already exists 😂 

I'm not sure I knew it existed either, to be fair.

TheSpyder (changes requested):
I just saw #9648. We don't need two PRs for the same fix, as `release/7` will be merged to `main` after release.

hamza0867:
Closing this PR as #9648 is the PR that should have existed to begin with.

PR #9642, opened 2024-05-13
TINY-10973: Resize bottom margin not always applied.

HAFRMO:
Related Ticket: https://ephocks.atlassian.net/browse/TINY-10793

Description of Changes:
`autoresize_bottom_margin` was not applied in some situations.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/plugins/autoresize/main/ts/core/Resize.ts
+++ b/modules/tinymce/src/plugins/autoresize/main/ts/core/Resize.ts
@@ -29,7 +29,7 @@ const toggleScrolling = (editor: Editor, state: boolean): void => {
   }
 };
 
-interface PreviousData {
+interface ResizeData {
   readonly totalHeight: number;
   readonly contentHeight: number;
   readonly initiated: boolean;
@@ -56,7 +56,7 @@ const shouldScrollIntoView = (trigger: EditorEvent<unknown> | undefined) => {
 /**
  * This method gets executed each time the editor needs to resize.
  */
-const resize = (editor: Editor, oldSize: Cell<PreviousData>, trigger?: EditorEvent<unknown>, getExtraMarginBottom?: () => number): void => {
+const resize = (editor: Editor, oldSize: Cell<ResizeData>, trigger?: EditorEvent<unknown>, getExtraMarginBottom?: () => number): void => {
   const dom = editor.dom;
 
   const doc = editor.getDoc();
@@ -141,7 +141,7 @@ const resize = (editor: Editor, oldSize: Cell<PreviousData>, trigger?: EditorEve
   }
 };
 
-const setup = (editor: Editor, oldSize: Cell<PreviousData>): void => {
+const setup = (editor: Editor, oldSize: Cell<ResizeData>): void => {
   const getExtraMarginBottom = () => Options.getAutoResizeBottomMargin(editor);
 
   editor.on('init', (e) => {
@@ -176,8 +176,8 @@ const setup = (editor: Editor, oldSize: Cell<PreviousData>): void => {
 };
 
 export {
-  PreviousData,
   resize,
+  ResizeData,
   setup
 };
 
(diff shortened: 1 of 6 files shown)


modules/tinymce/src/plugins/autoresize/test/ts/browser/AutoresizePluginDocumentTest.ts
```diff
@@ -0,0 +1,78 @@
...
+          resizeEventsCount.set(resizeEventsCount.get() + 1);
+        });
+      }
+    }, [ AutoresizePlugin, FullscreenPlugin ], true);
+
+    beforeEach(() => {
+      resizeEventsCount.set(0);
+    });
+
+    it('TBA: Should not have a resize handle visible by default', async () => {
```
spocke, line 46: I think these TBA is is just annoying if they never get changed to anything then why have them. Either use the ticket ID here too or just remove the `TBA:` I never seen anyone ever change from that value.

modules/tinymce/src/plugins/autoresize/test/ts/browser/AutoresizePluginTest.ts
```diff
@@ -186,7 +186,7 @@ describe('browser.tinymce.plugins.autoresize.AutoresizePluginTest', () => {
     });
   });
 
-  context('TINY-9123', () => {
+  context('TINY-8872', () => {
```
spocke, line 189: Was this ticket ID wrong?

modules/tinymce/src/plugins/autoresize/main/ts/core/Resize.ts
```diff
@@ -29,6 +29,12 @@ const toggleScrolling = (editor: Editor, state: boolean): void => {
   }
 };
 
+interface PreviousData {
```
spocke, line 32: Maybe name this ResizeData or something using the word Previous is a bit odd since you normally don't name types for states like that it's more a implementation detail of how you use that type that is current or previous.

spocke (approved):

modules/tinymce/src/plugins/autoresize/test/ts/browser/AutoresizePluginTest.ts
HAFRMO: Yes. Neither of us could figure out why it was this number, so we updated it since we were in the area anyway.

modules/tinymce/src/plugins/autoresize/test/ts/browser/AutoresizePluginDocumentTest.ts
HAFRMO: Wups, forgot to change those.

TheSpyder: Originally these were testrail case numbers, and QA would come through later to make sure automated tests were tracked in testrail (so we didn't manually test something that was covered by automated testing).

Some time later, whether intentionally or just because it made sense, we started using JIRA ticket numbers - and we rarely make PRs without JIRA tickets now.

modules/tinymce/src/plugins/autoresize/main/ts/core/Resize.ts
```diff
@@ -29,6 +29,12 @@ const toggleScrolling = (editor: Editor, state: boolean): void => {
   }
 };
 
+interface ResizeData {
+  readonly totalHeight: number;
+  readonly contentHeight: number;
+  readonly initiated: boolean;
```
TheSpyder, line 35: This seems poorly named. If it's a flag that the height has been set on the editor already, how about calling it `forced` or `set`?

- [ ] consider renaming

TheSpyder (approved):

modules/tinymce/src/plugins/autoresize/main/ts/core/Resize.ts
HAFRMO: Something like that, which is why I thought `initated` would be a descriptive name. But if you prefer I can rename it, though both 'forced' and `set` also seem a bit lacking in detail as to what it refers to ( and forced would invert the current values, wouldn't it? ).

PR #9645, opened 2024-05-14
TINY-10942: Show autocompleter regardless of match length

danoaky-tiny:
Related Ticket: TINY-10942

Description of Changes:
Removed the `context.text.length - ac.matchLength >= 10` check.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/keyboard/Autocompleter.ts
+++ b/modules/tinymce/src/core/main/ts/keyboard/Autocompleter.ts
@@ -100,7 +100,6 @@ export const setup = (editor: Editor): void => {
             // Ensure the active autocompleter trigger matches, as the old one may have closed
             // and a new one may have opened. If it doesn't match, then do nothing.
             if (ac.trigger !== context.trigger) {
-              cancelIfNecessary();
               return;
             }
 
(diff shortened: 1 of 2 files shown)


modules/tinymce/src/core/main/ts/keyboard/Autocompleter.ts
```diff
@@ -99,25 +99,23 @@ export const setup = (editor: Editor): void => {
...
-                  Events.fireAutocompleterUpdateActiveRange(editor, { range: context.range });
-                  Events.fireAutocompleterUpdate(editor, { lookupData });
-                } else {
-                  uiActive.set(true);
-                  Events.fireAutocompleterUpdateActiveRange(editor, { range: context.range });
-                  Events.fireAutocompleterStart(editor, { lookupData });
-                }
-              }
+            if (ac.trigger !== context.trigger) {
+              cancelIfNecessary();
```
spocke, line 103: Is this really just removing the max length this seems to now call cancelIfNecessary when they are different before it only called this if the max length was to big. So should this just be a return or does it even matter?

spocke (commented):
Needs a changelog as well.

modules/tinymce/src/core/main/ts/keyboard/Autocompleter.ts
danoaky-tiny: I removed it, so it's just a return now. I couldn't see a difference in testing

MitchC1999 (approved):

modules/tinymce/src/core/main/ts/keyboard/Autocompleter.ts
TheSpyder: The cancel function closes the autocompleter. If you look at the comment above this `if` block, when the `trigger` properties don't match then this autocompleter has closed and a new one opened. So yes it wasn't necessary.

TheSpyder (approved):

Issue #9391, opened 2024-02-16
👍 force_hex_color - Converts background-color:transparent  to background-color:#000000 (black)


TopCoder02:
@danoaky-tiny 

## 📝 Provide detailed reproduction steps (if any)
Please provide the steps to reproduce and if possible a minimal demo of the problem via [fiddle.tiny.cloud](fiddle.tiny.cloud) or similar.

**force_hex_color** - Converts `background-color:transparent`  to `background-color:#000000`
**force_hex_color** - Doesn't convert anything in `<style>` tags

### ✔️ Expected result
I don't know the right answer here but, maybe one of these two options will work

1.  `background-color:#0000000` (Notice the extra 0, which is the transparency layer)
2.  just leave it as `transparent` and not convert it. 

### ❌ Actual result
*What is the actual result of the above steps?*

Transparent color becomes black.

### ❓ Possible solution
*If you have ideas, you can list them here. Otherwise, you can delete this section.*

1. `background-color:#0000000` (Notice the extra 0, which is the transparency layer)
2. just leave it as `transparent` and not convert it. 

If you'd like to see this fixed sooner, add a 👍 reaction to this post.

TheSpyder:
Which version? A replication case will help a lot. Although the option is named "force hex color", the fact that it forces the color value to hex seems to be by design.

The option was added to solve a regression between TinyMCE 5 and 6, but we couldn't revert it as that would be a breaking change. We're looking at a proper fix for TinyMCE 7, so your feedback and an example would be much appreciated.

TopCoder02:
@TheSpyder 

It happens in both 6 and 7. 

Here's a fiddle:
https://fiddle.tiny.cloud/fVQGGgyZ0B/0

It shows both issues, the conversion to black and ignoring the style tags. 

**Source View**
![image](https://github.com/tinymce/tinymce/assets/119634613/036a691b-b983-4153-a45f-7f12a3d36198)

**WYSIYG View**
![image](https://github.com/tinymce/tinymce/assets/119634613/fed6215d-9a9d-4fd4-8b35-46a4f59fe659)

TheSpyder:
yes we haven't implemented the change yet so `7-dev` will be the same as `6`.

TopCoder02:
I doing this in the meantime, it only addresses the issue of transparent being converted to black.   The issue where it ignores and doesn't force HEX colors in the `style` tags, this won't address.

```
 editor.on('BeforeSetContent', function (e) {
[...]

PR #9649, opened 2024-05-15
TINY-10916: Transparent color is converted into #00000

ArvinJ-H:
Related Ticket: 
TINY-10916

Description of Changes:
- Added a seperate if statement to cover the edge case of transparent

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set
* [x] <s>Docs ticket created (if applicable)</s>

https://github.com/tinymce/tinymce/issues/9391

### DIFF
--- a/.changes/unreleased/tinymce-TINY-10916-2024-05-15.yaml
+++ b/.changes/unreleased/tinymce-TINY-10916-2024-05-15.yaml
@@ -1,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: Fixed a bug where applying a transparent color resulted in it being converted to black.
+time: 2024-05-15T17:05:18.42363+10:00
+custom:
+  Issue: TINY-10916
--- a/.changes/unreleased/tinymce-TINY-10916-2024-05-15.yaml
+++ b/.changes/unreleased/tinymce-TINY-10916-2024-05-15.yaml
@@ -1,6 +1,6 @@
 project: tinymce
 kind: Fixed
-body: Fixed a bug where transparent colors were converted to black.
+body: CSS color values set to `transparent` were incorrectly converted to '#000000`.
 time: 2024-05-15T17:05:18.42363+10:00
 custom:
   Issue: TINY-10916
--- a/.changes/unreleased/tinymce-TINY-10916-2024-05-15.yaml
+++ b/.changes/unreleased/tinymce-TINY-10916-2024-05-15.yaml
@@ -1,6 +1,6 @@
 project: tinymce
 kind: Fixed
-body: Fixed a bug where applying a transparent color resulted in it being converted to black.
+body: Fixed a bug where transparent colors were converted to black.
 time: 2024-05-15T17:05:18.42363+10:00
 custom:
   Issue: TINY-10916
--- a/modules/acid/src/main/ts/ephox/acid/api/colour/RgbaColour.ts
+++ b/modules/acid/src/main/ts/ephox/acid/api/colour/RgbaColour.ts
@@ -9,7 +9,9 @@ const max = Math.max;
 const round = Math.round;
 
 const rgbRegex = /^\s*rgb\s*\(\s*(\d+)\s*[,\s]\s*(\d+)\s*[,\s]\s*(\d+)\s*\)\s*$/i;
-const rgbaRegex = /^\s*rgba\s*\(\s*(\d+)\s*[,\s]\s*(\d+)\s*[,\s]\s*(\d+)\s*[,\s]\s*(\d?(?:\.\d+)?)\s*\)\s*$/i;
+
+// This regex will match rgba(0, 0, 0, 0.5) or rgba(0, 0, 0, 50%) or rgb(0, 0, 0, 1), or without commas
+const rgbaRegex = /^\s*(rgba|rgb)\s*\(\s*(\d{1,3})\s*[,\s]\s*(\d{1,3})\s*[,\s]\s*(\d{1,3})\s*[,\s]\s*(1|0|0?\.\d+|\d{1,3}%)\s*\)\s*$/i;
 
 const rgbaColour = (red: number, green: number, blue: number, alpha: number): Rgba => ({
   red,
@@ -113,7 +115,7 @@ const getColorFormat = (colorString: string): string => {
   } else if (rgbaRegex.test(colorString)) {
     return 'rgba';
   }
-  return 'other'; // Add a return statement for the default case
+  return 'other';
 };
 
 const fromString = (rgbaString: string): Optional<Rgba> => {
--- a/modules/acid/src/main/ts/ephox/acid/api/colour/RgbaColour.ts
+++ b/modules/acid/src/main/ts/ephox/acid/api/colour/RgbaColour.ts
@@ -10,8 +10,8 @@ const round = Math.round;
 
 const rgbRegex = /^\s*rgb\s*\(\s*(\d+)\s*[,\s]\s*(\d+)\s*[,\s]\s*(\d+)\s*\)\s*$/i;
 
-// This regex will match rgba(0, 0, 0, 0.5) or rgba(0, 0, 0, 50%) or rgb(0, 0, 0, 1), or without commas
-const rgbaRegex = /^\s*(rgba|rgb)\s*\(\s*(\d{1,3})\s*[,\s]\s*(\d{1,3})\s*[,\s]\s*(\d{1,3})\s*[,\s]\s*(1|0|0?\.\d+|\d{1,3}%)\s*\)\s*$/i;
+// This regex will match rgba(0, 0, 0, 0.5) or rgba(0, 0, 0, 50%) , or without commas
+const rgbaRegex = /^\s*rgba\s*\(\s*(\d+)\s*[,\s]\s*(\d+)\s*[,\s]\s*(\d+)\s*[,\s]\s*((?:\d?\.\d+|\d+)%?)\s*\)\s*$/i;
 
 const rgbaColour = (red: number, green: number, blue: number, alpha: number): Rgba => ({
   red,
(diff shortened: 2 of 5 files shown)


modules/tinymce/src/core/main/ts/api/html/Styles.ts
```diff
@@ -246,6 +247,9 @@ const Styles = (settings: StylesSettings = {}, schema?: Schema): Styles => {
             name = decodeHexSequences(name);
             value = decodeHexSequences(value);
 
+            console.log('name', name);
+            console.log('value', value);
```
spocke, line 251: Remove these.

```suggestion
            console.log('name', name);
            console.log('value', value);
```

.changes/unreleased/tinymce-TINY-10916-2024-05-15.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: Fixed a bug where applying a transparent color resulted in it being converted to black.
```
spocke, line 3: It converts existing colors as well so just loading up a document with `background: transparent` will change the color.

modules/tinymce/src/core/test/ts/browser/html/StylesTest.ts
```diff
@@ -264,4 +264,9 @@ describe('browser.tinymce.core.html.StylesTest', () => {
     assertStyles(styles, 'color: rgba(1, 2, 3);', 'color: rgba(1, 2, 3);');
     assertStyles(styles, 'color: rgb(1, 2, 3, 0.5);', 'color: rgb(1, 2, 3, 0.5);');
   });
+
+  it('TINY-10916: transparent should be convert to black with 0 alpha', () => {
+    const styles = Styles();
+    assertStyles(styles, 'color: transparent;', 'color: #00000000;');
```
spocke, line 270: It should not convert them at all. I guess maybe it wrote the ticket incorreclty.

spocke (changes requested):

modules/tinymce/src/core/test/ts/browser/html/StylesTest.ts
spocke: Also I would want a test on the setContent/getContent APIs as well.

modules/tinymce/src/core/main/ts/api/html/Styles.ts
```diff
@@ -265,6 +269,11 @@ const Styles = (settings: StylesSettings = {}, schema?: Schema): Styles => {
               value = value.toLowerCase();
             }
 
+            // Convert transparent to fully transparent black
+            if (value === 'transparent') {
+              value = '#00000000';
+            }
```
TheSpyder, line 275: But this isn't what we did in TinyMCE 5, we left it as `transparent`.

Can we not just leave it? The conversion to hex is only supposed to apply to `rgb()` colors, not named colors.

ArvinJ-H: Yeah, sorry had some confusing yesterday

modules/tinymce/src/core/test/ts/browser/html/StylesTest.ts
```diff
@@ -264,4 +271,16 @@ describe('browser.tinymce.core.html.StylesTest', () => {
     assertStyles(styles, 'color: rgba(1, 2, 3);', 'color: rgba(1, 2, 3);');
     assertStyles(styles, 'color: rgb(1, 2, 3, 0.5);', 'color: rgb(1, 2, 3, 0.5);');
   });
+
+  it('TINY-10916: transparent should not be converted to other format', () => {
+    const styles = Styles();
+    assertStyles(styles, 'color: transparent;', 'color: transparent;');
```
tiny-ben-tran, line 277: Should add more test cases to cover other properties that accept color values such as `border-color` and `background-color`

tiny-ben-tran (approved):

modules/tinymce/src/core/main/ts/api/html/Styles.ts
```diff
@@ -266,7 +267,7 @@ const Styles = (settings: StylesSettings = {}, schema?: Schema): Styles => {
             }
 
             // Convert RGB colors to HEX
-            if (!rgbaRegExp.test(value)) {
+            if (!rgbaRegExp.test(value) && value.toLowerCase() !== 'transparent') {
```
TheSpyder, line 270: I think there's a more fundamental problem here. If you look at the `rgbaRegExp` and this test, it's converting every value that doesn't begin with `rgba`. That's the problem, specifically whitelisting `transparent` won't help.

TheSpyder (changes requested):

modules/tinymce/src/core/main/ts/api/html/Styles.ts
TheSpyder: If you look back at TinyMCE 5, where we had the behaviour customers want, it was more or less as I described.
https://github.com/tinymce/tinymce/blob/05989b03bf56d98aaf3522630b481616799c4dc8/modules/tinymce/src/core/main/ts/api/html/Styles.ts#L60
https://github.com/tinymce/tinymce/blob/05989b03bf56d98aaf3522630b481616799c4dc8/modules/tinymce/src/core/main/ts/api/html/Styles.ts#L302

ArvinJ-H: Yes, the rgba regex need some fix up, I have updated it to match all 4 values.

The actual cause of the issue is this function
```
RgbaColour.fromString(value).each((rgba) => {}

const fromString = (rgbaString: string): Optional<Rgba> => {
  if (rgbaString === 'transparent') {
    return Optional.some(rgbaColour(0, 0, 0, 0));
  }
 }
 ```
 It recognise `transparent` as a valid string, and convert it into `rgba(0,0,0,0)`, that gets passed into this function
 ```
 const rgbaToHexString = (color: string): string =>
  RgbaColour.fromString(color)
    .map(HexColour.fromRgba)
    .map((h) => '#' + h.value)
    .getOr(color);
```
`fromRgba` function converts colour into `rgb` colour (in this case #000000), hence why the colour black.

TheSpyder: Good sleuthing! I think the regex issue is not just that it lacked detail, but swapping the condition logic means that _everything_ which isn't `rgba()` is passed for conversion, rather than just `rgb()` colors as intended.

Removing the `transparent` line looks like a much better fix than adding the `transparent` exclusion here. It's been in the code since 2018, so the exact reason will be lost in an old CI server before it was moved to open source. I can only hope tests fail if it's really needed.

But either way, I think swapping back to an `rgb` check here (perhaps using the regex from `RgbaColour.ts`) is a good idea, since that's clearly the intent of the code.

ArvinJ-H: According to documentation, all colours except `rgba` are converted to hex, so the if statement on the regex should be valid.
https://www.tiny.cloud/docs/tinymce/latest/migration-from-6x/#force-hex-color

I'll remove the `transparent` line and see if CI reports anything anywhere.

spocke: I guess we need to update the docs there since transparent was not converted into rgba in v6.

TheSpyder: That's documentation written based on the code. The intent, at least from my perspective, was to return to our v5 behaviour.

ltrouton (approved):

modules/tinymce/src/core/main/ts/api/html/Styles.ts
```diff
@@ -42,7 +42,7 @@ const Styles = (settings: StylesSettings = {}, schema?: Schema): Styles => {
   const urlOrStrRegExp = /(?:url(?:(?:\(\s*\"([^\"]+)\"\s*\))|(?:\(\s*\'([^\']+)\'\s*\))|(?:\(\s*([^)\s]+)\s*\))))|(?:\'([^\']+)\')|(?:\"([^\"]+)\")/gi;
   const styleRegExp = /\s*([^:]+):\s*([^;]+);?/g;
   const trimRightRegExp = /\s+$/;
-  const rgbaRegExp = /rgba *\(/i;
+  const rgbRegExp = /rgb\s*\(\s*([0-9]+)\s*,\s*([0-9]+)\s*,\s*([0-9]+)\s*\)/i;
```
TheSpyder, line 45: This doesn't belong here. Please make a function in `RgbaColour` to test using the regex it has instead of repeating it here.

- [ ] Remove this regex

TheSpyder (approved):
I think this will make a lot of people happy 👍

modules/acid/src/main/ts/ephox/acid/api/colour/RgbaColour.ts
```diff
@@ -130,6 +138,7 @@ export {
   fromHsv,
   fromHex,
   fromString,
+  getColorFormat,
```
spocke, line 141: Shouldn't acid also have some tests for this or the results of this change?

.changes/unreleased/tinymce-TINY-10916-2024-05-15.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: Fixed a bug where transparent colors were converted to black.
```
spocke, line 3: I think since the category is already Fixed it nice if we can avoid using fixed in the changelog message.

```suggestion
body: CSS color values set to `transparent` were incorrectly converted to '#000000`.
```

spocke (approved):

modules/acid/src/main/ts/ephox/acid/api/colour/RgbaColour.ts
```diff
@@ -102,13 +105,20 @@ const fromStringValues = (red: string, green: string, blue: string, alpha: strin
...
   const b = parseInt(blue, 10);
   const a = parseFloat(alpha);
+
   return rgbaColour(r, g, b, a);
 };
 
-const fromString = (rgbaString: string): Optional<Rgba> => {
-  if (rgbaString === 'transparent') {
-    return Optional.some(rgbaColour(0, 0, 0, 0));
+const getColorFormat = (colorString: string): string => {
```
TheSpyder, line 112: This return should've been a sum type, not a `string`.

All I really wanted was some sort of function to run `rgbRegex.test` - not a full "get format" method when we only ever want to check one of the conditions...

PR #9657, opened 2024-05-20
TINY-10428: emit slider onChange event while using arrow keys 

ztomaszyk:
Related Ticket: TINY-10428

Description of Changes:
Silver theme slider component emits onChange event now while using arrow keys 

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/.changes/unreleased/tinymce-TINY-10428-2024-05-20.yaml
+++ b/.changes/unreleased/tinymce-TINY-10428-2024-05-20.yaml
@@ -1,6 +1,6 @@
 project: tinymce
 kind: Improved
-body: 'Silver theme slider component emits onChange event now while using arrow keys '
+body: 'Silver theme slider component emits onChange event now while using arrow keys.'
 time: 2024-05-20T16:47:55.820758+02:00
 custom:
   Issue: TINY-10428
--- a/.changes/unreleased/tinymce-TINY-10428-2024-05-20.yaml
+++ b/.changes/unreleased/tinymce-TINY-10428-2024-05-20.yaml
@@ -1,6 +1,6 @@
 project: tinymce
 kind: Improved
-body: 'Silver theme slider component emits onChange event now while using arrow keys.'
+body: 'Dialog slider components now emit an onChange event when using arrow keys.'
 time: 2024-05-20T16:47:55.820758+02:00
 custom:
   Issue: TINY-10428
--- a/.changes/unreleased/tinymce-TINY-10428-2024-05-20.yaml
+++ b/.changes/unreleased/tinymce-TINY-10428-2024-05-20.yaml
@@ -1,0 +1,6 @@
+project: tinymce
+kind: Improved
+body: 'Silver theme slider component emits onChange event now while using arrow keys '
+time: 2024-05-20T16:47:55.820758+02:00
+custom:
+  Issue: TINY-10428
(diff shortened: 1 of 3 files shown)


.changes/unreleased/tinymce-TINY-10428-2024-05-20.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Improved
+body: 'Silver theme slider component emits onChange event now while using arrow keys '
```
spocke, line 3:
```suggestion
body: 'Silver theme slider component emits onChange event now while using arrow keys.'
```

hamza0867 (approved):

modules/tinymce/src/themes/silver/test/ts/browser/window/SilverDialogSliderApiTest.ts
```diff
@@ -0,0 +1,58 @@
...
+      it('TINY-10428: dialog onChange handler should fire when using arrow key on slider', async () => {
+        store.clear();
+        const editor = hook.editor();
+        DialogUtils.open(editor, dialogSpec, test.params);
+        await TinyUiActions.pWaitForDialog(editor);
+
+        await FocusTools.pTryOnSelector('Focus should be on slider rail', SugarDocument.getDocument(), '.tox-slider__rail');
+        TinyUiActions.keydown(editor, Keys.left());
+
+        await Waiter.pTryUntil('', () => store.assertEq('Check if on change event is logged after using arrow key', [ 'logOnChangeEvent' ]));
```
spocke, line 54: Not sure this needs to be a waiter and if it needs to be one then add a label for when then waiter fails like `Waited for change event` or something.

ztomaszyk: It passed locally without the waiter, so I am removing it. Thanks!

.changes/unreleased/tinymce-TINY-10428-2024-05-20.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Improved
+body: 'Silver theme slider component emits onChange event now while using arrow keys.'
```
TheSpyder, line 3:
```suggestion
body: 'Dialog slider components now emit an onChange event when using arrow keys.'
```

modules/tinymce/src/themes/silver/test/ts/browser/window/SilverDialogSliderApiTest.ts
```diff
@@ -0,0 +1,58 @@
...
+      it('TINY-10428: dialog onChange handler should fire when using arrow key on slider', async () => {
+        store.clear();
+        const editor = hook.editor();
+        DialogUtils.open(editor, dialogSpec, test.params);
+        await TinyUiActions.pWaitForDialog(editor);
+
+        await FocusTools.pTryOnSelector('Focus should be on slider rail', SugarDocument.getDocument(), '.tox-slider__rail');
+        TinyUiActions.keydown(editor, Keys.left());
+
+        store.assertEq('Check if on change event is logged after using arrow key', [ 'logOnChangeEvent' ]);
```
TheSpyder, line 54: Please check the value changed, rather than just that an event fired. Also check both left and right key directions for the correct value change.

- [ ] improve tests

TheSpyder (approved):

modules/tinymce/src/themes/silver/test/ts/browser/window/SilverDialogSliderApiTest.ts
ztomaszyk: There was no issue with value change. Slider just didn't emit `formChangeEvent` on that change. Testing if the value changed unfortunately doesn't cover my changes. Testing if the event was logged in the dialog is what actually tests my changes. 
All of that aside, I think it is good to assert both value and event to make sure arrow keys work as expected. Adding assertion to test both value change and event fired for left and right key 👍

TheSpyder: yes, while it's not technically part of your change it is the reason the event fires, so I think it's worth checking here.

PR #9658, opened 2024-05-21
TINY-10869: Improve merging styled spans by recursively stripping matched styled spans on insert

MitchC1999:
Related Ticket: TINY-10869

Description of Changes:
* Allow merging of identical "non-inheritable" styled spans, including background-color
* Recursively remove any inserted spans which are identical to a parent span of the insertion node

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/tinymce/src/core/test/ts/browser/content/insert/MergeInsertedSpansTest.ts
```diff
@@ -0,0 +1,208 @@
...
+  const hook = TinyHooks.bddSetupLight<Editor>({
+    base_url: '/project/tinymce/js/tinymce',
+  }, [], true);
+
+  const red = '#ff0000';
+  const green = '#00ff00';
+  const blue = '#0000ff';
+
+  // const buildNestedSpans = (style: string, values: string[], text: string): string => Arr.foldr(values, (acc, v) => `<span style="${style}: ${v};">${acc}</span>`, text);
+  const buildNestedSpans = (spans: SpanStyle[][], text: string): string => Arr.foldr(spans, (acc, spanStyles) => `<span style="${Arr.foldr(spanStyles, (acc, span) => ` ${span.style}: ${span.value};${acc}`, '').trim()}">${acc}</span>`, text);
```
spocke, line 23: Could you break this apart having a function call with a predicate lambda defined inside a template string makes it very hard to read.

modules/tinymce/src/core/test/ts/browser/content/insert/MergeInsertedSpansTest.ts
```diff
@@ -0,0 +1,208 @@
...
+
+  const testMergeNestedSpans = (initial: string, inserted: string, expected: string, path: number[], offset: number) => {
+    const editor = hook.editor();
+    editor.setContent('<p>' + initial + '</p>');
```
spocke, line 27: Why not template strings in the simple case when it's used in the complex case above.

```suggestion
    editor.setContent(`<p>${initial}</p>`);
```

modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
```diff
@@ -90,20 +89,22 @@ const reduceInlineTextElements = (editor: Editor, merge: boolean | undefined): v
...
+          if (currentNode.childNodes.length === 1) {
+            stripNodes(currentNode.childNodes.item(0) as HTMLElement);
           }
-
-          if (elementUtils.compare(parentNode, node)) {
-            dom.remove(node, true);
-            break;
+          // Check recursively if the current node has the same attributes and styles as any parent
+          const conflictWithParent = (parentNode: HTMLElement | null): boolean => Type.isNonNullable(parentNode) && parentNode !== root
+            && (elementUtils.compare(currentNode, parentNode) || conflictWithParent(parentNode.parentElement));
```
spocke, line 102: Could we add more tests for this not just styles since compare will not merge any similar element that is a textInlineElement so strong, b, mark etc also any attribute so spans with data attributes only or classes etc. I think that is a good thing to reduce those but we should have tests for that since if we ever refactor this then things might break. It would also be good to just have a test that verifies that `a` elements are not touched even if they have styles etc. Right now they wouldn't be since for one it's not a textInlineElement and the compare function would see that the element name is not the same as the parent and links can't be nested. But since this is such a central feature of the editor I would want that to be covered as well.

modules/tinymce/src/core/test/ts/browser/content/insert/MergeInsertedSpansTest.ts
```diff
@@ -0,0 +1,208 @@
+import { describe, it } from '@ephox/bedrock-client';
+import { Arr } from '@ephox/katamari';
+import { TinyAssertions, TinyHooks, TinySelections } from '@ephox/wrap-mcagar';
+
+import Editor from 'tinymce/core/api/Editor';
+import * as InsertContent from 'tinymce/core/content/InsertContent';
+
+interface SpanStyle {
+  style: string;
+  value: string;
```
spocke, line 10: Readonly

modules/tinymce/src/core/main/ts/content/InsertContentImpl.ts
```diff
@@ -90,20 +89,22 @@ const reduceInlineTextElements = (editor: Editor, merge: boolean | undefined): v
...
       const isInline = Type.isNonNullable(textInlineElements[node.nodeName.toLowerCase()]);
-      if (isInline && StyleUtils.hasInheritableStyles(dom, node)) {
-        for (let parentNode = node.parentElement; Type.isNonNullable(parentNode) && parentNode !== root; parentNode = parentNode.parentElement) {
-          // Check if the parent has a style conflict that would prevent the child node from being safely removed,
-          // even if a exact node match could be found further up the tree
-          const styleConflict = StyleUtils.hasStyleConflict(dom, node, parentNode);
-          if (styleConflict) {
-            break;
+      if (isInline) {
+        // Remove nodes recursively if identical to any parent node
```
spocke, line 93: Could this comment be removed if you renamed the function to `unwrapAllEqualChildElements` or something that captures what it's actually doing.

spocke (approved):
Looks really good. 👍

ArvinJ-H (approved):

shanmen-tiny (approved):
Nice work 🚀

modules/tinymce/src/core/test/ts/browser/content/insert/InsertContentTest.ts
ltrouton, line 1: For some of the test cases you updated, we probably need to also update the `it` block title as it is now inaccurate. Would it be worth adding more test cases with different children counts as well?

ltrouton (approved):

PR #9663, opened 2024-05-22
TINY-10904: now polaris manage acronyms correctly

lorenzo-pomili:
Related Ticket: TINY-10904

Description of Changes:
to manage correctly the acronyms when we get the words via polaris, I changed the check inside `findWordsWithIndices` instead of inside `isWordBoundary`.

This is because `isWordBoundary` is based on single "characters" so to avoid performance and readability issues I create an external check that is triggered only at the end of a word and on dots at the end of the word or after it.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/.changes/unreleased/polaris-TINY-10904-2024-05-22.yaml
+++ b/.changes/unreleased/polaris-TINY-10904-2024-05-22.yaml
@@ -1,0 +1,6 @@
+project: polaris
+kind: Fixed
+body: Acronyms where not managed correctly.
+time: 2024-05-22T16:02:02.052181985+02:00
+custom:
+  Issue: TINY-10904
--- a/.changes/unreleased/polaris-TINY-10904-2024-05-22.yaml
+++ b/.changes/unreleased/polaris-TINY-10904-2024-05-22.yaml
@@ -1,6 +1,6 @@
 project: polaris
 kind: Fixed
-body: Acronyms where not managed correctly.
+body: Acronyms were not managed correctly.
 time: 2024-05-22T16:02:02.052181985+02:00
 custom:
   Issue: TINY-10904
--- a/modules/polaris/src/main/ts/ephox/polaris/words/Words.ts
+++ b/modules/polaris/src/main/ts/ephox/polaris/words/Words.ts
@@ -51,7 +51,6 @@ const findWordsWithIndices = <T>(chars: Word<T>, sChars: string[], characterMap:
     word.push(chars[i]);
 
     // If there's a word boundary between the current character and the next character,
-    // (and this boundary doesn't depend from a dot at the end of an acronym)
     // append the current word to the words array and start building a new word.
     if (isWordBoundary(characterMap, i)) {
       const ch = sChars[i];
--- a/modules/polaris/src/main/ts/ephox/polaris/words/Words.ts
+++ b/modules/polaris/src/main/ts/ephox/polaris/words/Words.ts
@@ -107,9 +107,6 @@ const findWordsWithIndices = <T>(chars: Word<T>, sChars: string[], characterMap:
           start: startOfWord,
           end: endOfWord
         });
-        couldBeAnAcronym.prev = undefined;
-        couldBeAnAcronym.check = undefined;
-        couldBeAnAcronym.isTheLast = false;
       }
 
       word = [];
--- a/modules/polaris/src/main/ts/ephox/polaris/words/Words.ts
+++ b/modules/polaris/src/main/ts/ephox/polaris/words/Words.ts
@@ -38,6 +38,8 @@ export interface WordsWithIndices<T> {
   readonly indices: WordIndex[];
 }
 
+const isAnAcronym = (str: string) => new RegExp(/^(?:[A-Z]\.)+$/gm).test(str);
+
 const findWordsWithIndices = <T>(chars: Word<T>, sChars: string[], characterMap: CharacterMap, options: WordOptions): WordsWithIndices<T> => {
   const words: Word<T>[] = [];
   const indices: WordIndex[] = [];
@@ -51,12 +53,13 @@ const findWordsWithIndices = <T>(chars: Word<T>, sChars: string[], characterMap:
     word.push(chars[i]);
 
     // If there's a word boundary between the current character and the next character,
+    // (and this boundary doesn't depend from a dot at the end of an acronym)
     // append the current word to the words array and start building a new word.
-    if (isWordBoundary(characterMap, i)) {
+    if (isWordBoundary(characterMap, i) && !isAnAcronym(word.join('') + sChars[i + 1])) {
       const ch = sChars[i];
       if (
         (options.includeWhitespace || !WHITESPACE.test(ch)) &&
-        (options.includePunctuation || !PUNCTUATION.test(ch))
+        (options.includePunctuation || !PUNCTUATION.test(ch) || isAnAcronym(word.join('')))
       ) {
         const startOfWord = i - word.length + 1;
         const endOfWord = i + 1;
--- a/modules/polaris/src/main/ts/ephox/polaris/words/Words.ts
+++ b/modules/polaris/src/main/ts/ephox/polaris/words/Words.ts
@@ -2,6 +2,7 @@ import { Arr } from '@ephox/katamari';
 
 import { CharacterMap, classify } from './StringMapper';
 import * as UnicodeData from './UnicodeData';
+import { characterIndices as ci } from './UnicodeData';
 import { isWordBoundary } from './WordBoundary';
 
 const EMPTY_STRING = UnicodeData.EMPTY_STRING;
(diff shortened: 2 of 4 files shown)


.changes/unreleased/polaris-TINY-10904-2024-05-22.yaml
```diff
@@ -0,0 +1,6 @@
+project: polaris
+kind: Fixed
+body: Acronyms where not managed correctly.
```
spocke, line 3:
```suggestion
body: Acronyms were not managed correctly.
```

modules/polaris/src/main/ts/ephox/polaris/words/Words.ts
```diff
@@ -38,6 +38,8 @@ export interface WordsWithIndices<T> {
   readonly indices: WordIndex[];
 }
 
+const isAnAcronym = (str: string) => new RegExp(/^(?:[A-Z]\.)+$/gm).test(str);
```
spocke, line 41: I wonder if this could cause issues in some languages. Maybe we should limit this to single letter so it doesn't pass in. `SOMETHING.` as one word. If a user types in all caps for some odd reason we shouldn't treat that as an acronym for example. Also I think this can be simplified to: `/^([A-Z]\.)+$/` since the capture group ignore shouldn't matter since we are just using test we are not extracting things also the multiline should not be needed since we only want to match on letter and dot not \n stuff anyway and since we are only matching one thing I don't think we need to make it global either.

modules/polaris/src/main/ts/ephox/polaris/words/Words.ts
```diff
@@ -51,12 +53,13 @@ const findWordsWithIndices = <T>(chars: Word<T>, sChars: string[], characterMap:
...
     // If there's a word boundary between the current character and the next character,
+    // (and this boundary doesn't depend from a dot at the end of an acronym)
     // append the current word to the words array and start building a new word.
-    if (isWordBoundary(characterMap, i)) {
+    if (isWordBoundary(characterMap, i) && !isAnAcronym(word.join('') + sChars[i + 1])) {
       const ch = sChars[i];
       if (
         (options.includeWhitespace || !WHITESPACE.test(ch)) &&
-        (options.includePunctuation || !PUNCTUATION.test(ch))
+        (options.includePunctuation || !PUNCTUATION.test(ch) || isAnAcronym(word.join('')))
```
spocke, line 62: I wonder if this could cause performance issues on larger documents? This is also used for word count for example.

spocke (approved):

spocke:
This should have a `hotfix` branch prefix not sure how important that really is though since it's targeting a release branch.

modules/polaris/src/main/ts/ephox/polaris/words/Words.ts
lorenzo-pomili: > I wonder if this could cause issues in some languages.
Yes this could be a matter to manage, but I think the best solution would be to check if someone complain about that, since check all the way to write could be not possible in short period

> Maybe we should limit this to single letter so it doesn't pass in. SOMETHING. as one word.
is already limited to 1 letter `SOMETHING.` doesn't pass, to make it pass you have to change it in `S.O.M.E.T.H.I.N.G.`

> Also I think this can be simplified to: /^([A-Z]\.)+$/ since the capture group ignore shouldn't matter since we are just using test we are not extracting things also the multiline should not be needed since we only want to match on letter and dot not \n stuff anyway and since we are only matching one thing I don't think we need to make it global either.

agree removing `gm` and `?:`, but I wonder if we should extend this check to the lower case too, since I don't see any reason to have it only for upper case.

modules/polaris/src/main/ts/ephox/polaris/words/Words.ts
lorenzo-pomili: the problem here as that I did not find any alternative, I putted the check at the end and in the lowest level to make it be triggered only on boundaries and if we are checking a `dot`

modules/polaris/src/main/ts/ephox/polaris/words/Words.ts
lorenzo-pomili: my proposal would be to change it in:
```js
const isAnAcronym = (str: string) => new RegExp(/^([A-Z]\.)+$/).test(str);
```
or
```js
const isAnAcronym = (str: string) => new RegExp(/^([a-zA-Z]\.)+$/).test(str);
```

spocke: Right I forgot to remove the part about it checking the SOMETHING I initially misread that regex. I guess lowercase would be fine too.

lorenzo-pomili: I made a [test commit](https://github.com/tinymce/tinymce/pull/9663/commits/ba6d8ee03f47863eea4eaf8ad5903c7e8e0df297) with code that should be more optimized, but sadly is way harder to read, let me know what do you think about it

modules/polaris/src/main/ts/ephox/polaris/words/Words.ts
```diff
@@ -74,9 +107,15 @@ const findWordsWithIndices = <T>(chars: Word<T>, sChars: string[], characterMap:
           start: startOfWord,
           end: endOfWord
         });
+        couldBeAnAcronym.prev = undefined;
+        couldBeAnAcronym.check = undefined;
+        couldBeAnAcronym.isTheLast = false;
```
TheSpyder, line 112: This doesn't seem necessary, you're in an `if` block and the only other code before repeating the reset is `word = []` 🤔

modules/polaris/src/main/ts/ephox/polaris/words/Words.ts
```diff
@@ -38,25 +39,57 @@ export interface WordsWithIndices<T> {
...
   // Loop through each character in the classification map and determine whether
   // it precedes a word boundary, building an array of distinct words as we go.
   for (let i = 0; i < characterMap.length; ++i) {
 
     // Append this character to the current word.
     word.push(chars[i]);
 
+    const onWordBoundary = isWordBoundary(characterMap, i);
+    const dotAfterLetter = sChars[i] === '.' && couldBeAnAcronym.prev === ci.ALETTER;
+    const letterAfterDot = characterMap[i] === ci.ALETTER && (couldBeAnAcronym.prev === '.' || couldBeAnAcronym.prev === undefined);
```
TheSpyder, line 70: This should let you avoid keeping track of `prev` at all? Or am I misreading the algorithm? It's really confusing how you've put both characters and mapped characters into a single `undefined | number | string` variable.
```suggestion
    const dotAfterLetter = sChars[i] === '.' && characterMap[i - 1] === ci.ALETTER;
    const letterAfterDot = characterMap[i] === ci.ALETTER && (sChars[i - 1] === '.');
```

modules/polaris/src/main/ts/ephox/polaris/words/Words.ts
```diff
@@ -38,25 +39,57 @@ export interface WordsWithIndices<T> {
...
+        couldBeAnAcronym.check = true;
+      } else {
+        couldBeAnAcronym.check = false;
+      }
+    }
     // If there's a word boundary between the current character and the next character,
+    // (and this boundary doesn't depend from a dot at the end of an acronym)
     // append the current word to the words array and start building a new word.
-    if (isWordBoundary(characterMap, i)) {
+    if (onWordBoundary && (couldBeAnAcronym.isTheLast || !couldBeAnAcronym.check)) {
```
TheSpyder, line 88: Since `isTheLast` is only set to true when `onWordBoundary` is true, and all this code only runs when `onWordBoundary` is true why does `isTheLast` exist? Looking at how it's set, it could probably be replaced with a check for `dotAfterLetter`.

But please make sure your tests capture the edge cases so they can confirm whether I'm wrong.

With this and the `prev` change your mutable additions collapse to a `check` boolean.

TheSpyder (approved):
- [ ] Check my suggestions

modules/polaris/src/main/ts/ephox/polaris/words/Words.ts
lorenzo-pomili: you are right, removed it

modules/polaris/src/main/ts/ephox/polaris/words/Words.ts
spocke: Yeah this became hard to read.

It needs to check more than just the previous character so we can't use a simple - 1 index. Since this could be things like `U.S.A.` but we could do a loop backwards but that would mean that it could be slow if we have a lot of these patterns. Maybe we limit how many steps backwards it can take. Also looping would break this early for the happy path most words will not be single letters separated with dots. So checking every word agains a regexp should be slower for normal documents since this will likely never backtrack more than two letters.

I wonder if this check could then be moved to the isWordBoundary function it would need to get the chars array passed in as well to check for `.` characters since, we don't have that explicit in the map, it could then be a extra check at the end that checks for patterns backwards. Loops from the current index and checks for ALETTER and dots. Then you make a similar comment as the ones that are already there for edge cases like braking in numbers etc since this is similar to that.

spocke: Thinking on this some more the only time this will trigger is if the word boundary character is a `.` so the happy path is even happier it would only execute at the end of sentences or on acronyms. So you could just make a if statement to check if the current char is `.` then then check I guess forward and backwards for single letters interspersed with dots.

lorenzo-pomili: I simplified it, I try to refactor it more to see if I can make it more readable, let me know what do you think about the current version

lorenzo-pomili: ok, thanks to @spocke we should have a final version

lorenzo-pomili:
close in favor of the a PR with the correct branch name: https://github.com/tinymce/tinymce/pull/9671

PR #9675, opened 2024-05-29
TINY-6887: Change the action order

ArvinJ-H:
Related Ticket: 
TINY-6887

Description of Changes:
- Changed the order actions, sandbox is closed before action is triggered

Pre-checks:
* [x] Changelog entry added
* [x] <s>Tests have been added (if applicable)</s>

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/plugins/table/test/ts/browser/TableContextMenuInsertTest.ts
+++ b/modules/tinymce/src/plugins/table/test/ts/browser/TableContextMenuInsertTest.ts
@@ -39,7 +39,7 @@ describe('browser.tinymce.plugins.table.TableCellPropsStyleTest', () => {
     assert.notEqual(warnSanitized, expectedSanitized, `Warning matched expected: ${expectedSanitized}`);
   };
 
-  it('TINY-6887', async () => {
+  it('TINY-6887: should have no console warning when inserting table via context menu', async () => {
     const editor = hook.editor();
     editor.setContent('<p></p>');
     await pOpenContextMenu(editor, 'p');
--- a/modules/tinymce/src/plugins/table/test/ts/browser/TableContextMenuInsertTest.ts
+++ b/modules/tinymce/src/plugins/table/test/ts/browser/TableContextMenuInsertTest.ts
@@ -1,0 +1,51 @@
+/* eslint-disable no-console */
+import { describe, it } from '@ephox/bedrock-client';
+import { Arr } from '@ephox/katamari';
+import { TinyHooks, TinyUiActions } from '@ephox/wrap-mcagar';
+import { assert } from 'chai';
+
+import Editor from 'tinymce/core/api/Editor';
+import AutoResizePlugin from 'tinymce/plugins/autoresize/Plugin';
+import TablePlugin from 'tinymce/plugins/table/Plugin';
+
+import { pOpenContextMenu, sanitizeString } from '../module/test/TableTestUtils';
+
+describe('browser.tinymce.plugins.table.TableCellPropsStyleTest', () => {
+  const hook = TinyHooks.bddSetupLight<Editor>({
+    plugins: 'table autoresize',
+    indent: false,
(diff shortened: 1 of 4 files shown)


spocke (commented):
Can we add tests for it so it doesn't come back?

TheSpyder:
If `onAction` causes the menu to close, do we even need the close sandbox line at all?

TheSpyder:
> If `onAction` causes the menu to close, do we even need the close sandbox line at all?

After digging in a bit, the answer is yes. Why the menu closes is an interesting question. As near as I can tell, it's because `spec.onAction` is this `editor.execCommand` call, and exec command will focus the editor by default. The change in focus closes the menu.
https://github.com/tinymce/tinymce/blob/f6a87ed7a6d32c627cf8ac6c1bd44ce36d69b14b/modules/tinymce/src/plugins/table/main/ts/ui/MenuItems.ts#L61-L64

However this only happens if the table plugin is active. In practice the table grid picker won't be rendered without the table plugin, but in theory it could be used for something else.

TheSpyder (approved):

modules/tinymce/src/plugins/table/test/ts/browser/TableContextMenuInsertTest.ts
```diff
@@ -0,0 +1,50 @@
...
+
+    const warnSanitized = sanitizeString(lastWarn);
+    const expectedSanitized = sanitizeString(expectedLog);
+
+    if (warnSanitized === expectedSanitized) {
+      throw new Error(`Warning matched expected: ${expectedSanitized}`);
+    }
+  };
+
+  it('TINY-6887', async () => {
```
shanmen-tiny, line 42: Description

modules/tinymce/src/plugins/table/test/ts/browser/TableContextMenuInsertTest.ts
```diff
@@ -0,0 +1,50 @@
...
+    console.warn = originalWarn;
+
+    const lastWarn = Arr.last(warns).getOr('none');
+
+    const warnSanitized = sanitizeString(lastWarn);
+    const expectedSanitized = sanitizeString(expectedLog);
+
+    if (warnSanitized === expectedSanitized) {
+      throw new Error(`Warning matched expected: ${expectedSanitized}`);
+    }
```
shanmen-tiny, line 39:
```suggestion
    assert.notEqual(warnSanitized, expectedSanitized, `Warning matched expected: ${expectedSanitized}`);
```

modules/tinymce/src/plugins/table/test/ts/browser/TableContextMenuInsertTest.ts
```diff
@@ -0,0 +1,50 @@
...
+  };
+
+  it('TINY-6887', async () => {
+    const editor = hook.editor();
+    editor.setContent('<p></p>');
+    await pOpenContextMenu(editor, 'p');
+    await TinyUiActions.pWaitForPopup(editor, '.tox-collection--list');
+    TinyUiActions.clickOnUi(editor, 'div[role="menuitem"]:contains("Table")');
+    await assertDebugLog( editor, 'The component must be in a context to execute: triggerEvent <div class="tox-insert-table-picker"></div> is not in context.');
```
shanmen-tiny, line 48: Maybe rename to `pAssertNoConsoleWarn` or something and move the second parameter into the function? This looks like it's expecting the string but it's throwing error in that function when it matches 🤪

shanmen-tiny (approved):

modules/tinymce/src/plugins/table/test/ts/browser/TableContextMenuInsertTest.ts
ArvinJ-H: oh yeah that function name needs to be updated, i'll move the sting into the function

PR #9683, opened 2024-06-05
TINY-11001: Replaced tiny branding logo

shanmen-tiny:
Related Ticket: TINY-11001

Description of Changes:
* Replaced Tiny logo with new logo with text
* Sync hover styles to tinymce-5 for premium skins
* Updated `aria-label`
* Screenshots for premium skin will be uploaded in the comment section of the ticket
* Icon is minified through `oxide-icon-tools` down to 1.32k before gzipped

<details>
<summary>Hover state</summary>

image

</details>

<details>
<summary>Screen max-width: 767px</summary>

SCR-20240605-qsg

</details>

<details>
<summary>Screen min-width: 767px ~ max-width: 1000px</summary>
- ~The logo svg is set to smaller and has less margin in the span due to all items (element path, help text, word count and branding are all shown, so sometimes it will overlap with the resize button)~ No longer required, arrow is removed.

SCR-20240605-qsi

</details>

<details>
<summary>Screen min-width: 1000px</summary>

SCR-20240605-qsm

</details>

<details>
<summary>RTL</summary>

SCR-20240605-qs2

</details>

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


.changes/unreleased/tinymce-TINY-11001-2024-06-05.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Changed
+body: Replaced tiny branding logo with `Build with TinyMCE` text and logo.
```
MitchC1999, line 3: Does this need to be documented? It's more of a maintenance task than a change and won't have a release notes entry

MitchC1999 (approved):

TheSpyder:
So this is where I realise that if we're in RTL the arrow is likely backwards 🫣

If there isn't a unicode arrow we can use they really need to be two images (so we can transform-flip the arrow).

shanmen-tiny:
> So this is where I realise that if we're in RTL the arrow is going to be backwards 🫣
> 
> If there isn't a unicode arrow we can use they really need to be two images (so we can transform-flip the arrow).

We could use: `editor.translate([ 'Build with {0} {1}', Logo.trim(), I18n.isRtl() ? '↖' : '↗︎' ])`
Would using image a better way? Now we require 2 strings 🤪

SCR-20240605-iv9

.changes/unreleased/tinymce-TINY-11001-2024-06-05.yaml
shanmen-tiny: One was added when changing from the text from v5 to v6 🤪 Happy to remove if no objections.

spocke: Yes this needs to be logged since it does these things:
* Changes the CSS that could break custom skins
* Changes the I18n that could break custom language packs.
* The logo and text is wider than before so might cause rendering issues on smaller editor instances.

spocke (approved):

PR #9687, opened 2024-06-05
TINY-10946: replace `Math.random` with `crypto.getRandomValues`

lorenzo-pomili:
Related Ticket: TINY-10946

Description of Changes:
replaced `Math.random` with `crypto.getRandomValues`

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/katamari/src/main/ts/ephox/katamari/api/Num.ts
+++ b/modules/katamari/src/main/ts/ephox/katamari/api/Num.ts
@@ -17,3 +17,7 @@ export const cycleBy = (value: number, delta: number, min: number, max: number):
 // ASSUMPTION: Max will always be larger than min
 export const clamp = (value: number, min: number, max: number): number =>
   Math.min(Math.max(value, min), max);
+
+// this is from this example: https://github.com/gkouziik/eslint-plugin-security-node/blob/master/docs/rules/detect-insecure-randomness.md
+// quoting the link above: "Divide a random UInt32 by the maximum value (2^32 -1) to get a result between 0 and 1"
+export const random = (): number => window.self.crypto.getRandomValues(new Uint32Array(1))[0] / 4294967295;
--- a/modules/katamari/src/main/ts/ephox/katamari/api/Num.ts
+++ b/modules/katamari/src/main/ts/ephox/katamari/api/Num.ts
@@ -18,6 +18,5 @@ export const cycleBy = (value: number, delta: number, min: number, max: number):
 export const clamp = (value: number, min: number, max: number): number =>
   Math.min(Math.max(value, min), max);
 
-// this is from this example: https://github.com/gkouziik/eslint-plugin-security-node/blob/master/docs/rules/detect-insecure-randomness.md
-// quoting the link above: "Divide a random UInt32 by the maximum value (2^32 -1) to get a result between 0 and 1"
+// the division is meant to get a number between 0 and 1 for more information check this discussion: https://stackoverflow.com/questions/58285941/how-to-replace-math-random-with-crypto-getrandomvalues-and-keep-same-result
 export const random = (): number => window.self.crypto.getRandomValues(new Uint32Array(1))[0] / 4294967295;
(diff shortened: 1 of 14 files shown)


modules/katamari/src/main/ts/ephox/katamari/api/Num.ts
```diff
@@ -17,3 +17,7 @@ export const cycleBy = (value: number, delta: number, min: number, max: number):
 // ASSUMPTION: Max will always be larger than min
 export const clamp = (value: number, min: number, max: number): number =>
   Math.min(Math.max(value, min), max);
+
+// this is from this example: https://github.com/gkouziik/eslint-plugin-security-node/blob/master/docs/rules/detect-insecure-randomness.md
+// quoting the link above: "Divide a random UInt32 by the maximum value (2^32 -1) to get a result between 0 and 1"
+export const random = (): number => window.self.crypto.getRandomValues(new Uint32Array(1))[0] / 4294967295;
```
TheSpyder, line 23: Oh very nice. This is better than I expected when I asked for a central random method!

modules/tinymce/src/core/main/ts/util/Uuid.ts
TheSpyder, line 1: I really don't think we need this file at all, it's only use in `BlobCache` and only for an internal reference that I'm fairly sure we never expose to users.

TheSpyder (approved):

modules/katamari/src/main/ts/ephox/katamari/api/Num.ts
spocke: I wonder if we should drop the reference to that docs example since that project seems to be ISC however googling around that type of method to get random floats seems to be commonly available on stackoverflow.

modules/tinymce/src/core/main/ts/util/Uuid.ts
spocke: We could do that in a cleanup task?

spocke (approved):

modules/tinymce/src/core/main/ts/util/Uuid.ts
lorenzo-pomili: created: TINY-11027

modules/katamari/src/main/ts/ephox/katamari/api/Num.ts
lorenzo-pomili: I was quoting it also because I copied the their comment: `Divide a random UInt32 by the maximum value (2^32 -1) to get a result between 0 and 1` that it clarify why there is a division for such odd number

TheSpyder: Spocke is concerned about whether we might get an IP concern - although ISC allows us to use this, it would require acknowledgement.

An equivalent post is available on Stack Overflow you could link to instead.
https://stackoverflow.com/questions/58285941/how-to-replace-math-random-with-crypto-getrandomvalues-and-keep-same-result

lorenzo-pomili: what about?
```
// the division is meant to get a number between 0 and 1 for more information check this discussion: https://stackoverflow.com/questions/58285941/how-to-replace-math-random-with-crypto-getrandomvalues-and-keep-same-result
```

lorenzo-pomili: I changed it

TheSpyder: I was so busy looking at the comment I didn't realise it uses `window.self.crypto`... that's a bit unnecessary when just `crypto` would suffice 😂

spocke: Don't we have a eslint rule to not use globals directly. But window.crypto should be the same as window.self.crypto.

TheSpyder: I thought we dropped that rule after `dom-globals` was retired. But yes it's `.self` that I was really highlighting as unnecessary.

lorenzo-pomili: I removed the `.self`

PR #9725, opened 2024-06-20
TINY-10973: fix split buttons popup positioning with scroll and `fullscreen`

lorenzo-pomili:
Related Ticket: TINY-10973

Description of Changes:
the problem was that the `popupSinkBounds` with a scrollable container returned the container bounds even if the editor is fullscreen

~EDIT: I moved the check inside `ScrollingContext.getBoundsFrom` since is correct that if we are in fullscreen that boundaries are the `Boxes.win()`~ (changed again)

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/.changes/unreleased/tinymce-TINY-10973-2024-06-20.yaml
+++ b/.changes/unreleased/tinymce-TINY-10973-2024-06-20.yaml
@@ -1,0 +1,7 @@
+project: tinymce
+kind: Fixed
+body: Split buttons popup were placed in the wrong positon when the container is scrollable
+  and the editor use fullscreen
+time: 2024-06-20T14:03:06.529094655+02:00
+custom:
+  Issue: TINY-10973
--- a/.changes/unreleased/tinymce-TINY-10973-2024-06-20.yaml
+++ b/.changes/unreleased/tinymce-TINY-10973-2024-06-20.yaml
@@ -1,7 +1,6 @@
 project: tinymce
 kind: Fixed
-body: Split buttons popup were placed in the wrong positon when the container is scrollable
-  and the editor use fullscreen
+body: Split buttons popup were incorrectly positioned when switching to fullscreen mode if the editor was inside a scrollable container.
 time: 2024-06-20T14:03:06.529094655+02:00
 custom:
   Issue: TINY-10973
--- a/.changes/unreleased/tinymce-TINY-10973-2024-06-20.yaml
+++ b/.changes/unreleased/tinymce-TINY-10973-2024-06-20.yaml
@@ -1,6 +1,6 @@
 project: tinymce
 kind: Fixed
-body: Split buttons popup were incorrectly positioned when switching to fullscreen mode if the editor was inside a scrollable container.
+body: Split button popups were incorrectly positioned when switching to fullscreen mode if the editor was inside a scrollable container.
 time: 2024-06-20T14:03:06.529094655+02:00
 custom:
   Issue: TINY-10973
--- a/modules/tinymce/src/themes/silver/main/ts/Theme.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/Theme.ts
@@ -21,6 +21,9 @@ const registerOptions = (editor: Editor) => {
   ContextMenuOptions.register(editor);
 };
 
+const isFullscreen = (editor: Editor): boolean =>
+  editor.plugins.fullscreen && editor.plugins.fullscreen.isFullscreen();
+
 export default (): void => {
   ThemeManager.add('silver', (editor): Theme => {
     registerOptions(editor);
@@ -54,7 +57,8 @@ export default (): void => {
             // At this stage, it looks like we need to calculate the bounds each time, just in
             // case the scrolling context details have changed since the last time. The bounds considers
             // the Boxes.box sizes, which might change over time.
-            return ScrollingContext.getBoundsFrom(sc);
+            // TINY-10973: if the editor is in fullscreen we could not consider the `ScrollingContext`
+            return isFullscreen(editor) ? Boxes.win() : ScrollingContext.getBoundsFrom(sc);
           };
         }
       );
--- a/modules/tinymce/src/themes/silver/main/ts/Theme.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/Theme.ts
@@ -21,9 +21,6 @@ const registerOptions = (editor: Editor) => {
   ContextMenuOptions.register(editor);
 };
 
-const isFullscreen = (editor: Editor): boolean =>
-  editor.plugins.fullscreen && editor.plugins.fullscreen.isFullscreen();
-
 export default (): void => {
   ThemeManager.add('silver', (editor): Theme => {
     registerOptions(editor);
@@ -57,8 +54,7 @@ export default (): void => {
             // At this stage, it looks like we need to calculate the bounds each time, just in
             // case the scrolling context details have changed since the last time. The bounds considers
             // the Boxes.box sizes, which might change over time.
-            // TINY-10973: if the editor is in fullscreen we could not consider the `ScrollingContext`
-            return isFullscreen(editor) ? Boxes.win() : ScrollingContext.getBoundsFrom(sc);
+            return ScrollingContext.getBoundsFrom(editor, sc);
           };
         }
       );
--- a/modules/tinymce/src/themes/silver/main/ts/Theme.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/Theme.ts
@@ -44,13 +44,20 @@ export default (): void => {
     const renderUI = (): RenderResult => {
       const renderResult = renderModeUI();
 
-      // If we have a scroll container, it looks like we need to calculate the bounds each time, just in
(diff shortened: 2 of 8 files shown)


.changes/unreleased/tinymce-TINY-10973-2024-06-20.yaml
```diff
@@ -0,0 +1,7 @@
+project: tinymce
+kind: Fixed
+body: Split buttons popup were placed in the wrong positon when the container is scrollable
+  and the editor use fullscreen
```
spocke, line 4: Maybe:
```suggestion
body: Split buttons popup were incorrectly positioned when switching to fullscreen mode if the editor was inside a scrollable container.
```

modules/tinymce/src/themes/silver/test/ts/browser/editor/toolbar/SplitButtonPopupPositionWithFullscreenTest.ts
```diff
@@ -0,0 +1,59 @@
...
+      base_url: '/project/tinymce/js/tinymce',
+      plugins: [
+        'advlist', 'lists', 'fullscreen'
+      ],
+      toolbar: 'fullscreen bullist numlist',
+      toolbar_mode: 'sliding',
+      ui_mode: 'split',
+    }, setupElement, []);
+
+    it('TINY-10973: the split button popups should be rendered close by the corrispective split button', async () => {
```
spocke, line 40:
```suggestion
    it('TINY-10973: the split button popups should be rendered close by the corrispective split button', async () => {
```
```suggestion
    it('TINY-10973: the split button popups should be rendered close to the split button', async () => {
```

modules/tinymce/src/themes/silver/main/ts/modes/ScrollingContext.ts
```diff
@@ -52,19 +52,24 @@ export const detect = (popupSinkElem: SugarElement<HTMLElement>): Optional<Scrol
 export const detectWhenSplitUiMode = (editor: Editor, popupSinkElem: SugarElement<HTMLElement>): Optional<ScrollingContext> =>
   Options.isSplitUiMode(editor) ? detect(popupSinkElem) : Optional.none();
 
+const isFullscreen = (editor: Editor): boolean =>
```
spocke, line 55: I wonder if we could make this detection more broad it more that if the `editor.getContainer()` is inside a `position: fixed` element then we should use the window bounds? I think cross plugin API calls should be kept at a minimum if possible. I guess one could also try to check for the `tox-fullscreen` class however that is also kind of plugin specific.

modules/tinymce/src/themes/silver/test/ts/browser/editor/toolbar/SplitButtonPopupPositionWithFullscreenTest.ts
```diff
@@ -0,0 +1,59 @@
+import { context, describe, it } from '@ephox/bedrock-client';
+import { Css, Insert, Remove, SugarBody, SugarElement } from '@ephox/sugar';
+import { TinyHooks, TinyUiActions } from '@ephox/wrap-mcagar';
+import { assert } from 'chai';
+
+import Editor from 'tinymce/core/api/Editor';
+
+describe('browser.tinymce.themes.silver.editor.toolbar.SplitButtonPopupPositionWithFullscreenTest', () => {
+  context('in a scroll content', () => {
```
spocke, line 9:
```suggestion
  context('in scrollable container', () => {
```

spocke (approved):

modules/tinymce/src/themes/silver/test/ts/browser/editor/toolbar/SplitButtonPopupPositionWithFullscreenTest.ts
lorenzo-pomili: done

modules/tinymce/src/themes/silver/main/ts/modes/ScrollingContext.ts
lorenzo-pomili: initially I was thinking to create a new editor's api like `editor.isFullscreen`, but probably trying to put the logic inside `editor.getContainer()` could have more sense, I'll try it

lorenzo-pomili: umh, maybe is not a good idea change `editor.getContainer()`, also I don't see how to do a different check because the check is done in [popupSinkBounds](https://github.com/tinymce/tinymce/blob/main/modules/tinymce/src/themes/silver/main/ts/Theme.ts#L57) that make sense to have the check for a fixed container with `ScrollingContext.detectWhenSplitUiMode`

spocke: Maybe the check is fine then.

modules/tinymce/src/themes/silver/main/ts/modes/ScrollingContext.ts
```diff
@@ -52,19 +52,24 @@ export const detect = (popupSinkElem: SugarElement<HTMLElement>): Optional<Scrol
...
+  editor.plugins.fullscreen && editor.plugins.fullscreen.isFullscreen();
+
 // Using all the scrolling viewports in the ancestry, limit the absolute
 // coordinates of window so that the bounds are limited by all the scrolling
 // viewports.
-export const getBoundsFrom = (sc: ScrollingContext): Bounds => {
+export const getBoundsFrom = (editor: Editor, sc: ScrollingContext): Bounds => {
```
TheSpyder, line 61: It looks like the only reason you have `editor` here is for the `fullscreen` call? That's too tightly coupled.

Please take some inspiration from other parts of our UI architecture and pass an `isFullscreen` function to this method instead of the full editor object. Reducing the capability of function arguments is an easy way to avoid spaghetti code dependencies.

Or, if you look really closely, there's a way to completely avoid adding the extra method argument. I'm happy to chat if you can't figure it out.

- [ ] Improve method architecture

TheSpyder (changes requested):

ztomaszyk (approved):

modules/tinymce/src/themes/silver/main/ts/modes/ScrollingContext.ts
lorenzo-pomili: ok, I think I got it, I moved the check inside `detectWhenSplitUiMode`, but after that I had to refactor the code in `Theme.ts` (and also the comment of that code :sweat_smile: ), let me know this is what you had in mind or I'm off track

TheSpyder: That's not what I expected, but it's probably cleaner.

What I was thinking is the `sc` parameter here is created in a context that has an editor, so you could add an `isFullscreen` property to it:
https://github.com/tinymce/tinymce/blob/64333acf0e9e93aa1853c824405654fb663e2e28/modules/tinymce/src/themes/silver/main/ts/modes/ScrollingContext.ts#L9-L12

I'm not entirely convinced that checking for a scrollable parent every time render is called is a good performance decision, but so long as it only happens when a popup opens it should be fine.

.changes/unreleased/tinymce-TINY-10973-2024-06-20.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: Split buttons popup were incorrectly positioned when switching to fullscreen mode if the editor was inside a scrollable container.
```
TheSpyder, line 3:
```suggestion
body: Split button popups were incorrectly positioned when switching to fullscreen mode if the editor was inside a scrollable container.
```

modules/tinymce/src/themes/silver/main/ts/Theme.ts
```diff
@@ -44,20 +44,13 @@ export default (): void => {
...
-        (sc) => {
-          popupSinkBounds = () => {
-            // At this stage, it looks like we need to calculate the bounds each time, just in
-            // case the scrolling context details have changed since the last time. The bounds considers
-            // the Boxes.box sizes, which might change over time.
-            return ScrollingContext.getBoundsFrom(sc);
-          };
-        }
-      );
+      ).map(ScrollingContext.getBoundsFrom).getOr(Boxes.win());
```
TheSpyder, line 53:
```suggestion
      ).fold(Boxes.win, ScrollingContext.getBoundsFrom);
```

TheSpyder (approved):

modules/tinymce/src/themes/silver/main/ts/Theme.ts
lorenzo-pomili: @TheSpyder  the code has been changed with the suggestion you wrote in this comment: https://github.com/tinymce/tinymce/pull/9725#discussion_r1665093904 so this part no long exist, let me know if the new implementation is ok

shanmen-tiny (approved):

TheSpyder (approved):

PR #9740, opened 2024-07-01
TINY-10978: Exported variables

lorenzo-pomili:
Related Ticket: TINY-10978

Description of Changes:
needed for the PR in the plugin

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [x] ~Milestone set~

### DIFF
--- a/modules/oxide/src/less/theme/globals/properties.less
+++ b/modules/oxide/src/less/theme/globals/properties.less
@@ -1,0 +1,31 @@
+.tox {
+  --tox-background-color: @background-color;
+  --tox-base-value: @base-value;
+  --tox-border-color: @border-color;
+  --tox-button-background-color: @button-background-color;
+  --tox-button-border-color: @button-border-color;
+  --tox-button-disabled-background-color: @button-disabled-background-color;
+  --tox-button-disabled-text-color: @button-disabled-text-color;
+  --tox-button-font-weight: @button-font-weight;
+  --tox-button-hover-background-color: @button-hover-background-color;
+  --tox-button-hover-text-color: @button-hover-text-color;
+  --tox-button-secondary-background-color: @button-secondary-background-color;
+  --tox-button-secondary-disabled-background-color: @button-secondary-disabled-background-color;
+  --tox-button-secondary-disabled-text-color: @button-secondary-disabled-text-color;
+  --tox-button-secondary-hover-background-color: @button-secondary-hover-background-color;
+  --tox-button-secondary-hover-text-color: @button-secondary-hover-text-color;
+  --tox-button-secondary-text-color: @button-secondary-text-color;
+  --tox-color-black: @color-black;
+  --tox-color-error: @color-error;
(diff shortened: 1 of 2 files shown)


modules/oxide/src/less/theme/globals/properties.less
TheSpyder, line 1: I know CSS variables are supported by all of our browsers now, but I don't know if mixing and matching LESS with CSS variables is a good precedent.

TheSpyder: Turns out we already do some of this with comments. Maybe there's a migration discussion to be had; I would prefer CSS variables, but I don't want to sit on the fence with them and LESS. It would be one more thing for new developers to learn.

spocke: The less variables gets compiled to css so not sure how that would ever be raw less values unless the less compiler is broken or you use raw encoded strings. Either way we would need to have visual regression testing in general if we are afraid of things breaking. This besomes a new API and I think this is the way to go just because it's less right now doesn't mean it needs to be less in the future. We need someway to map our skin variables to uploadcare variables and this is the way to do it. The alternative would be more css file bloat and/or hardcode uploadcare specific stuff in the skin.

TheSpyder: I wasn't talking about raw values. What I meant was that CSS variables is something we've discussed migrating our entire UI to. They're a better solution, in my opinion, but they are a different solution.

Our approach in the past has been to put plugin CSS in the core so it can be skinned appropriately. I don't agree with that either, but we can't make up new ways to develop our product just because we disagree with how it's been done for years.

I'd rather schedule a complete swap from LESS to CSS variables in TinyMCE 8, and then we can finally start removing plugin CSS from the core.

spocke: Can't really be done if you use shadow dom unless we load the CSS into that shadow element that would either make it a new request to the entire skin or a separate file. Exporting variables seems like a simpler approach just export everything as variables then later when we actually switch to CSS variables then we could just use these exported properties as the values you can change. I ran into the same problem with the mathml preview I wanted to get the text color so that I could render white on black if the skin is a dark skin.

spocke: The documented way of skinning uploadcare components is with CSS variables so somehow we need to get our variables to their variables this seems like the easiest way to do that. I guess in theory we could scope this to a `uploadcare` class then compute the colors of that using these variable names if that is less dangerous.

spocke: We won't ever be building our own pickers that is to much work. We just skin what uploadcare provide. So the variables approach will stay for the release in one form or another. The custom UI will be for image manipulation.

TheSpyder (approved):

tiny-ben-tran (approved):

TheSpyder:
I'm told this is 7.5 or later so marking as draft

PR #9669, opened 2024-05-28
TINY-10590: Merge blocks now no longer relying on browser behavior for edgecase.

HAFRMO:
Related Ticket: https://ephocks.atlassian.net/browse/TINY-10590

Description of Changes:
Merge block would not always consider nested elements correctly.

Pre-checks:
* [x] Changelog entry added
* [x] Tests have been added (if applicable)
* [x] Branch prefixed with `feature/`, `hotfix/` or `spike/`

Review:
* [x] Milestone set
* [x] Docs ticket created (if applicable)

GitHub issues (if applicable):

PR #9746, opened 2024-07-04
Revert "TINY-10590: Merge blocks now no longer relying on browser behavior for edgecase."

tiny-ben-tran:
After an investigation, it was determined that the change in #9669 incorrectly modifies the DOM when pressing backspace on a `br`. The following image shows the result between before (left) and after (right) the change has been reverted. 

Screenshot 2024-07-04 at 11 21 50 am


TheSpyder:
It's not clear why this is being reverted, even after reading JIRA. Some comment in the PR description would be nice.

TheSpyder (approved):

MitchC1999 (approved):

shanmen-tiny (approved):

PR #9753, opened 2024-07-05
TINY-11053: Now handling whole html documents in the text area.

HAFRMO:
Related Ticket: https://ephocks.atlassian.net/browse/TINY-11057

Description of Changes:
Previously whole html-blocks were put into a body-tag and merging the head and body before we could do anything about them.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/.changes/unreleased/tinymce-TINY-11053-2024-07-04.yaml
+++ b/.changes/unreleased/tinymce-TINY-11053-2024-07-04.yaml
@@ -1,0 +1,7 @@
+project: tinymce
+kind: Improved
+body: When adding a full document the content of the head element is no longer added
+  to the body.
+time: 2024-07-04T17:12:09.757923664+02:00
+custom:
+  Issue: TINY-11053
--- a/.changes/unreleased/tinymce-TINY-11053-2024-07-04.yaml
+++ b/.changes/unreleased/tinymce-TINY-11053-2024-07-04.yaml
@@ -1,7 +1,6 @@
 project: tinymce
 kind: Improved
-body: When adding a full document the content of the head element is no longer added
-  to the body.
+body: When a full document was loaded as editor content the head elements were added to the body.
 time: 2024-07-04T17:12:09.757923664+02:00
 custom:
   Issue: TINY-11053
--- a/modules/tinymce/src/core/main/ts/api/html/DomParser.ts
+++ b/modules/tinymce/src/core/main/ts/api/html/DomParser.ts
@@ -297,7 +297,7 @@ const DomParser = (settings: DomParserSettings = {}, schema = Schema()): DomPars
       if (format === 'xhtml') {
         // If parsing XHTML then the content must contain the xmlns declaration, see https://www.w3.org/TR/xhtml1/normative.html#strict
         return `<html xmlns="http://www.w3.org/1999/xhtml"><head></head><body>${content}</body></html>`;
-      } else if (/^[\s]*<head/i.test(html) || /^[\s]*<html/i.test(html)) {
+      } else if (/^[\s]*<head/i.test(html) || /^[\s]*<html/i.test(html) || /^[\s]*<!DOCTYPE/i.test(html)) {
         return `<html>${content}</html>`;
       } else {
         return `<body>${content}</body>`;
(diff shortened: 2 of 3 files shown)


modules/tinymce/src/core/main/ts/api/html/DomParser.ts
```diff
@@ -293,9 +293,18 @@ const DomParser = (settings: DomParserSettings = {}, schema = Schema()): DomPars
...
     const isSpecialRoot = Obj.has(schema.getSpecialElements(), rootName.toLowerCase());
     const content = isSpecialRoot ? `<${rootName}>${html}</${rootName}>` : html;
-    // If parsing XHTML then the content must contain the xmlns declaration, see https://www.w3.org/TR/xhtml1/normative.html#strict
-    const wrappedHtml = format === 'xhtml' ? `<html xmlns="http://www.w3.org/1999/xhtml"><head></head><body>${content}</body></html>` : `<body>${content}</body>`;
-    const body = parser.parseFromString(wrappedHtml, mimeType).body;
+    const makeWrap = () => {
+      if (format === 'xhtml') {
+        // If parsing XHTML then the content must contain the xmlns declaration, see https://www.w3.org/TR/xhtml1/normative.html#strict
+        return `<html xmlns="http://www.w3.org/1999/xhtml"><head></head><body>${content}</body></html>`;
+      } else if (/^[\s]*<head/i.test(html) || /^[\s]*<html/i.test(html)) {
```
TheSpyder, line 300: It's probably safer to check for a `<body` tag, IMO that's the most reliable indicator of whether the content is a body contents or a full document.
```suggestion
      } else if (/^[\s]*<body/i.test(html)) {
```

.changes/unreleased/tinymce-TINY-11053-2024-07-04.yaml
```diff
@@ -0,0 +1,7 @@
+project: tinymce
+kind: Improved
+body: When adding a full document the content of the head element is no longer added
+  to the body.
```
TheSpyder, line 4:
```suggestion
kind: Fixed
body: When a full document was loaded as editor content the head elements were added to the body.
```

TheSpyder (approved):
The ticket number in your description is wrong. You only need to supply the ticket number, not the full URL, it will be auto-linked.

ltrouton (approved):

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
lorenzo-pomili: I found a couple of edge cases:
```HTML
<!DOCTYPE html>
<html>
  <head>
    <!-- c 1 -->
    <!-- c 2 -->
  </head>
  <body>
    abc
  </body>
</html>
```
and
```HTML
<!-- <!DOCTYPE html> -->
<html>
  <head>
    <!-- c 1 -->
    <!-- c 2 -->
  </head>
  <body>
    abc
  </body>
</html>
```
@TheSpyder solution seems to solve it but we should do a check on the entire string not only at the start, something like this:
```suggestion
      } else if (/<body/i.test(html)) {
```

I don't see a case when the string `<body` could be inserted not as tag, but I could be wrong

- [ ] manage the edge cases in this comment

lorenzo-pomili (approved):

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
HAFRMO: Yeah, there are edge cases, and I'm not sure how to solve it. We can't quite make the regex just a pure `/<body/i` as suggested because newlines very easily are included when pasting from anywhere else, that is why I include the space-ignoring at the start.
As for checking for a body tag, we can't do that since the content that is fed in here does not always include it. The need to use the html-tag instead of the body-tag is an exception to the rule, and kind-of needs to be treated as it. As if we use the html-tag when it's not needed we risk dataloss ( spaces, maybe, but still ).
As for how to handle the presented cases, I don't know. This is a very poor situation.

lorenzo-pomili: ok, if the check on the body is not possible, what fixing at last the `<!DOCTYPE html>` doing something like:
```suggestion
      } else if (/^[\s]*<head/i.test(html) || /^[\s]*<html/i.test(html) || /^[\s]*<!DOCTYPE/i.test(html)) {
```
?

I understand that if it starts with a comment we can't understand if we should manage in a different way, but maybe is worth to manage the `DOCTYPE` case

TheSpyder: This whole change is about removing data, anything outside the `body` tag needs to be discarded. I'mnot sure how a `<body` check would lose data inside the body.

HAFRMO: It can happen ( we actually have tests for it ) if the wrong content is placed inside of just a pure html tag instead of a body tag. So if our check for `<body` were to trigger on the wrong content and wrap it in the <html> instead of <body>, some content can be lost ( Whitespace as far as I know, yes, but still ) and I don't want to add the situation where such things can disappear with seemingly random content. It'd be a pain to discover why later.

PR #9758, opened 2024-07-09
TINY-10884: remove fixed width for listbox

ztomaszyk:
Related Ticket: TINY-10884

Description of Changes:
Removed setting fixed width for listbox dropdown.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/themes/silver/main/ts/ui/dropdown/CommonDropdown.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/ui/dropdown/CommonDropdown.ts
@@ -178,7 +178,7 @@ const renderCommonDropdown = <T>(
           onControlDetached(spec, editorOffCell)
         ]),
         AddEventsBehaviour.config(fixWidthBehaviourName, [
-          AlloyEvents.runOnAttached((comp, _se) => UiUtils.forceInitialSize(comp)),
+          AlloyEvents.runOnAttached((comp, _se) => spec.listRole === 'listbox' ? Fun.noop : UiUtils.forceInitialSize(comp)),
         ]),
         AddEventsBehaviour.config('menubutton-update-display-text', [
           // These handlers are just using Replacing to replace either the menu
--- a/.changes/unreleased/tinymce-TINY-10884-2024-07-09.yaml
+++ b/.changes/unreleased/tinymce-TINY-10884-2024-07-09.yaml
@@ -1,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: The listbox component had a fixed width and was not a responsive ui element.
+time: 2024-07-09T17:43:03.579836+02:00
+custom:
+  Issue: TINY-10884


modules/tinymce/src/themes/silver/main/ts/ui/dropdown/CommonDropdown.ts
```diff
@@ -178,7 +178,7 @@ const renderCommonDropdown = <T>(
           onControlDetached(spec, editorOffCell)
         ]),
         AddEventsBehaviour.config(fixWidthBehaviourName, [
-          AlloyEvents.runOnAttached((comp, _se) => UiUtils.forceInitialSize(comp)),
+          AlloyEvents.runOnAttached((comp, _se) => spec.listRole === 'listbox' ? Fun.noop : UiUtils.forceInitialSize(comp)),
```
TheSpyder, line 181: I'm concerned this might impact more than where you intended it to. Have you checked everywhere listboxes are used? I see 31 references across open source and premium features.

Most of those aren't in resizable dialogs, I admit, but still.

ztomaszyk: I am not sure if I checked all of the places, but this bug is not specific for just math dialog. Fixed width was added in the CommonDropdown for menu and toolbar buttons two years ago [here](https://github.com/tinymce/tinymce/pull/8353), but it also affected listboxes. Like you said, most of the dialogs are not resizable so I think no one noticed before. 
Here is an example of this bug in other dialog: 

https://github.com/tinymce/tinymce/assets/91727075/a00e399e-3b0c-46d4-b5b5-bd096a6dd964

Is it possible to test for window resize? 🤔

TheSpyder: Ah, I see now. I wonder if this code is over-simplified, other "common dropdown" components probably don't want a fixed width either (they should be using flex to expand to the size of their parent element). Tracking that down seems out of scope for this fix, though.

It's not easy to test window resize but you don't need to - just patch the width of the `.tox-dialog` element after it loads. Or any element the listbox is in, really.

The existing tests for listbox are headless, not in a dialog context, but you might be able to add a test that puts the listbox into a fixed width container and then checks what happens if the container resizes. It turn out to be be too hard, in which case don't worry.

ztomaszyk: I tried adding test for listbox and putting it in the fixed container. It doesn't work as we don't load all the css at this point so it doesn't behave as it does in TinyMCE. I also tried adding a test in the SilverThemeDialog but same thing with the css. If I was to write a test for it it needs to be in the editor context. In general tinyMCE editor tests or some plugin. What do you think? Tbh, I don't like it as it is really far away from the change.

TheSpyder: I think it would be a valuable test to have but yes let's not waste time on this ticket trying to get that right. Perhaps log a tech debt ticket in JIRA about it?

TheSpyder (approved):

lorenzo-pomili (approved):

PR #9773, opened 2024-07-17
TINY-10981: Extend editor mode and readonly to allow selection

shanmen-tiny:
Related Ticket: TINY-10981

Description of Changes:
* Extending mode to allow selection in read only mode
* When deactivating the old mode, if the deactivation callback throws an exception, it will prevent the mode from being deactivated

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/tinymce/src/core/main/ts/api/Options.ts
```diff
@@ -859,6 +859,18 @@ const register = (editor: Editor): void => {
     default: true
   });
 
+  registerOption('selection_mode_allowed_commands', {
```
shanmen-tiny, line 862: This may not be a good idea 🤔 I was thinking since the ui's are all disabled in the upcoming PR for uiEnabled, if we might just want to remove this.

MitchC1999: Maybe in the future we can add an "allowed in readonly" property to commands as well

MitchC1999 (approved):

modules/tinymce/src/core/main/ts/api/Options.ts
TheSpyder: This feels like an implementation detail we don't want to hard code into a setting just yet. Once a setting has been released it's hard to change.

I'm also a bit concerned that we have premium features listed here. I'd rather the premium feature remain in control of when the commands do and don't work.

modules/tinymce/src/core/main/ts/mode/Readonly.ts
```diff
@@ -156,14 +176,33 @@ const processReadonlyEvents = (editor: Editor, e: Event): void => {
 };
 
 const registerReadOnlySelectionBlockers = (editor: Editor): void => {
+  editor.on('beforeinput paste', (e) => {
+    if (isReadOnly(editor) && editor.mode.isSelectionEnabled()) {
```
TheSpyder, line 180: I feel like with these new blockers it doesn't matter if selection is enabled. They should always block in readonly mode.

```suggestion
    if (isReadOnly(editor)) {
```

modules/tinymce/src/core/main/ts/mode/Readonly.ts
```diff
@@ -156,14 +176,33 @@ const processReadonlyEvents = (editor: Editor, e: Event): void => {
...
+  });
+
   editor.on('ShowCaret', (e) => {
-    if (isReadOnly(editor)) {
+    if (isReadOnly(editor) && !editor.mode.isSelectionEnabled()) {
       e.preventDefault();
     }
   });
 
   editor.on('ObjectSelected', (e) => {
```
TheSpyder, line 204: For a fractional saving in bundle size, identical listeners can be registered with a single function. This was true before your change, but it's a nice improvement to include.
`editor.on('ShowCaret ObjectSelected', (e) => {`

modules/tinymce/src/core/main/ts/mode/Readonly.ts
```diff
@@ -54,32 +56,50 @@ const restoreFakeSelection = (editor: Editor) => {
...
-    setEditorCommandState(editor, 'StyleWithCSS', false);
-    setEditorCommandState(editor, 'enableInlineTableEditing', false);
-    setEditorCommandState(editor, 'enableObjectResizing', false);
-    if (EditorFocus.hasEditorOrUiFocus(editor)) {
-      editor.focus();
+  const shouldSetContentEditableTrue = () => Type.isBoolean(readOnlyMode) ? readOnlyMode : !Obj.get(readOnlyMode, 'selectionEnabled').getOr(true);
+
+  if (shouldSetReadOnly) {
+    setEditorReadonly(editor);
+    if (shouldSetContentEditableTrue()) {
```
TheSpyder, line 97: This doesn't need to be a function if you do it inside the first `if`. I also think `Obj.get` is overkill for this scenario? Just `readonlyMode.selectionEnabled` would work for the `if` condition where this is used.
```suggestion
  if (shouldSetReadOnly) {
    const shouldSetContentEditableTrue = Type.isBoolean(readOnlyMode) ? readOnlyMode : readonlyMode.selectionEnabled;
    setEditorReadonly(editor);
    if (shouldSetContentEditableTrue) {
```

modules/tinymce/src/core/main/ts/api/Mode.ts
```diff
@@ -1,16 +1,38 @@
...
+import { Cell, Fun, Obj, Type } from '@ephox/katamari';
 
 import { registerMode, setMode } from '../mode/Mode';
 import { isReadOnly, registerReadOnlyContentFilters, registerReadOnlySelectionBlockers } from '../mode/Readonly';
 import Editor from './Editor';
 
+export type ReadonlyProperty = Readonly<'uiEnabled' | 'selectionEnabled'>;
+
+export type EditorReadOnlyType = boolean | {
+  [K in ReadonlyProperty]?: boolean;
```
TheSpyder, line 10: Wouldn't it be easier to just list the two of them directly? We're not likely to have a lot of these 🤔

modules/tinymce/src/core/main/ts/api/Mode.ts
```diff
@@ -1,16 +1,38 @@
...
  * TinyMCE Editor Mode API.
  *
  * @class tinymce.EditorMode
  */
 
 export interface EditorMode {
+  /**
+   * Checks if the editor content can be selected.
+   *
+   * @method selectionIsReadOnly
```
TheSpyder, line 23: This and the ui enabled setting have incorrect names in the doc comment

modules/tinymce/src/core/main/ts/api/Mode.ts
```diff
@@ -1,16 +1,38 @@
...
+  /**
+   * Checks if the editor user interface is in a readonly state.
+   *
+   * @method uiIsReadOnly
+   * @return {boolean} true if the editor user interface is enabled.
+   */
+  isUiEnabled: () => boolean;
+
   /**
    * Checks if the editor is in a readonly state.
```
TheSpyder, line 37: I think the meaning of this has changed a bit now

```suggestion
   * Checks if the editor content is in a readonly state.
```

modules/tinymce/src/core/main/ts/api/Mode.ts
```diff
@@ -86,7 +108,14 @@ export const create = (editor: Editor): EditorMode => {
   registerReadOnlyContentFilters(editor);
   registerReadOnlySelectionBlockers(editor);
 
+  const getReadonlyFromProperty = (property: ReadonlyProperty) => {
+    const mode = availableModes.get()[activeMode.get()].editorReadOnly;
```
TheSpyder, line 112: That's a bit messy. We might need to change the `activeMode` cell to have a bit more information.

I also wonder whether some of this should move into the `mode/Mode.ts` file, the `api/Mode.ts` file is trying to be very simple and avoid the complexities of mode management. That's why `setMode` is over there.

- [ ] I think we can do better, and in particular avoid `Obj.get`

TheSpyder (approved):

TheSpyder:
I suspected when this work started that I'd need to do an architecture API review, sorry I didn't make time for it earlier.

.changes/unreleased/tinymce-TINY-10981-2024-07-17.yaml
```diff
@@ -0,0 +1,7 @@
+project: tinymce
+kind: Added
+body: '`selectionEnabled` to read only mode, which allow selection and cursor to be
+  placed within the editor in read only mode.'
```
TheSpyder, line 4:
```suggestion
body: new '`selectionEnabled` property of the read only mode API, which allows normal cursor movement
  within a read-only editor.'
```

modules/tinymce/src/core/main/ts/api/Options.ts
shanmen-tiny: I've removed this for this version, maybe we can explore different opportunities in the upcoming version.

modules/tinymce/src/core/main/ts/api/Mode.ts
shanmen-tiny: I've simplified that and moved to `mode/Mode.ts` instead of placing that in `api/Mode.ts` that should make the api and the function cleaner.

TheSpyder (approved):

modules/tinymce/src/core/main/ts/NodeChange.ts
```diff
@@ -97,7 +97,7 @@ class NodeChange {
     let node: Element | undefined;
 
     // Fix for bug #1896577 it seems that this can not be fired while the editor is loading
-    if (this.editor.initialized && selection && !Options.shouldDisableNodeChange(this.editor) && !this.editor.mode.isReadOnly()) {
+    if ((this.editor.initialized && selection && !Options.shouldDisableNodeChange(this.editor)) || (this.editor.mode.isReadOnly() && this.editor.mode.isSelectionEnabled())) {
```
ltrouton, line 100: Maybe I am reading the `if` block incorrectly but it looks like `(this.editor.mode.isReadOnly() && this.editor.mode.isSelectionEnabled())` can override the previous `(this.editor.initialized && selection && !Options.shouldDisableNodeChange(this.editor))` check. Is this intentional?

modules/tinymce/src/core/main/ts/api/dom/ControlSelection.ts
```diff
@@ -430,7 +430,12 @@ const ControlSelection = (selection: EditorSelection, editor: Editor): ControlSe
     const targetElm = e.type === 'mousedown' ? e.target : selection.getNode();
     const controlElm = SelectorFind.closest<HTMLElement>(SugarElement.fromDom(targetElm), controlElmSelector)
       .map((e) => e.dom)
-      .filter((e) => dom.isEditable(e.parentElement) || (e.nodeName === 'IMG' && dom.isEditable(e)))
+      .filter((e) => {
+        if (editor.mode.isReadOnly()) {
+          return editor.mode.isSelectionEnabled() && e.nodeName !== 'TABLE';
+        }
```
ltrouton, line 436: I am curious where this came from?

modules/tinymce/src/core/main/ts/mode/Readonly.ts
```diff
@@ -156,14 +173,20 @@ const processReadonlyEvents = (editor: Editor, e: Event): void => {
...
+  editor.on('beforeinput paste', (e) => {
     if (isReadOnly(editor)) {
       e.preventDefault();
     }
   });
 
-  editor.on('ObjectSelected', (e) => {
-    if (isReadOnly(editor)) {
+  editor.on('BeforeSetContent', (e) => {
+    if (isReadOnly(editor) && editor.mode.isSelectionEnabled() && !e.initial) {
```
ltrouton, line 183: Just trying to determine if `editor.mode.isSelectionEnabled()` is required in the check here? What is this new code trying to achieve?

modules/tinymce/src/core/main/ts/NodeChange.ts
shanmen-tiny: I think we should also still respect if `shoulDisableNodeChange` option if the integrator decide to switch that off, I'll push a change to fix that.

modules/tinymce/src/core/main/ts/api/dom/ControlSelection.ts
shanmen-tiny: This should be removed.

modules/tinymce/src/core/main/ts/mode/Readonly.ts
shanmen-tiny: We switched from `contentEditable=true` when the editor is in this `selectionEnabled` mode, so you can place a cursor and move around the editor with only keyboard. What I'm trying to achieve is blocking `setContent/insertContent` when the cursor is placed in the editor, so that only annotating is permitted but not altering the content of the editor. The previous readonly (or full readonly mode) disables the editor body `contentEditable='false'` to the body, so inserting content is blocked by default.

modules/tinymce/src/core/main/ts/NodeChange.ts
ltrouton: The change looks better 👍

modules/tinymce/src/core/main/ts/mode/Readonly.ts
ltrouton: Ah right. I think I know what you are trying to do now. Could add a comment clarifying the purpose of the check maybe?

modules/tinymce/src/core/test/ts/webdriver/delete/SelectionEnabledDeleteContentTest.ts
```diff
@@ -0,0 +1,386 @@
...
+      '</div>' ].join('\n'));
+
+    setMode(editor, 'design');
+    TinySelections.select(editor, 'div', []);
+    assertFakeSelection(editor, true);
+    await RealKeys.pSendKeysOn('iframe => body', [ RealKeys.backspace() ]);
+    TinyAssertions.assertContent(editor, '');
+  });
+
+  it('TINY-10891: Deleting table content should not be permitted in cusorEnabled mode', async () => {
```
ltrouton, line 185: There appears to be a few other test cases with the old cursorEnabled name
```suggestion
  it('TINY-10891: Deleting table content should not be permitted in selectionEnabled mode', async () => {
```

ltrouton (approved):
Nice work on the test 👍

PR #9783, opened 2024-07-23
TINY-11123: close dropdowns while resizing window

ztomaszyk:
Related Ticket: TINY-11123

Description of Changes:
Dropdown would not close or adjust its width while resizing the window. We decided to close the dropdown to solve that issue.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/.changes/unreleased/tinymce-TINY-11123-2024-08-26.yaml
+++ b/.changes/unreleased/tinymce-TINY-11123-2024-08-26.yaml
@@ -1,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: Dropdowns would not adjust width after window resize.
+time: 2024-08-26T17:39:32.903555+02:00
+custom:
+  Issue: TINY-11123
--- a/.changes/unreleased/tinymce-TINY-11123-2024-08-26.yaml
+++ b/.changes/unreleased/tinymce-TINY-11123-2024-08-26.yaml
@@ -1,6 +1,6 @@
 project: tinymce
-kind: Fixed
-body: Dropdowns would not adjust width after window resize.
+kind: Improved
+body: Dropdowns would not close after window resize.
 time: 2024-08-26T17:39:32.903555+02:00
 custom:
   Issue: TINY-11123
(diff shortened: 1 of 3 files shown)


.changes/unreleased/tinymce-TINY-11123-2024-08-26.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Improved
+body: Dropdowns would not close after window resize.
```
TheSpyder, line 3:
```suggestion
body: Dialog list dropdown menus now close when the browser window resizes.
```

modules/tinymce/src/themes/silver/main/ts/ui/dropdown/CommonDropdown.ts
```diff
@@ -178,7 +178,10 @@ const renderCommonDropdown = <T>(
           onControlDetached(spec, editorOffCell)
         ]),
         AddEventsBehaviour.config(fixWidthBehaviourName, [
-          AlloyEvents.runOnAttached((comp, _se) => spec.listRole === 'listbox' ? Fun.noop : UiUtils.forceInitialSize(comp)),
+          AlloyEvents.runOnAttached((comp, _se) => spec.listRole === 'listbox' ? Fun.noop() : UiUtils.forceInitialSize(comp)),
```
TheSpyder, line 181: I doubt `Fun.noop()` is the correct answer here. If the types want a `void` return, just make it `undefined`?

TheSpyder (approved):

modules/tinymce/src/themes/silver/main/ts/ui/dropdown/CommonDropdown.ts
ztomaszyk: I find `Fun.noop` quite readable here. If it is listbox - no operation. If it is not force initial size. 
I don't like how it looks with undefined, so maybe something like this:
```
if (spec.listRole !== 'listbox') {
  UiUtils.forceInitialSize(comp)
}
```
What do you think?

TheSpyder: yes, that's also a good way to do it 👍

lorenzo-pomili (approved):

Issue #9788, opened 2024-07-25
Add support for CSS Variables in the Colour map


andrewnicols:
## 📝 Provide a description of the new feature or improvement
Add support for CSS Colours in the `color_map`, for example:

```js
tinymce.init({
  selector: 'textarea',  // change this value according to your HTML
  toolbar: 'forecolor backcolor',
  color_map: [
    'var(--black)', 'Black',
    'var(--gray)', 'Gray',
    'var(--white)', 'White',
    'var(--red)', 'Red',
    'var(--yellow)', 'Yellow',
    'var(--green)', 'Green',
    'var(--blue)', 'Blue'
  ]
});
```

### 🫶 What is the motivation?
The colour of content is intrinsically linked to the style of the page. If an organisation changes its branding, there should be a way to update content colouring accordingly. This also applies to content created on pages supporting a dark mode.

With support for CSS colour maps, it becomes possible for the page to override the colours for different themes, and cases such as dark mode (e.g. to invert colours).

### 🔗 What is the consequence of not having this feature?
Content colouring is hard-coded into the content, and difficult to manage and update.

### 🚦 How important would you rate the requested feature or improvement?
Critical.

We have a [tracker issue](https://tracker.moodle.org/browse/MDL-76517) for our product with a huge amount of interest. We have had to disable the colour picker until we can support updating content colouring in a maintainable fashion.

---

If you'd like to see this implemented sooner, add a 👍 reaction to this post.

sluebbert:
At this point it may be worth a new issue to keep things clean, but I am very much interested in getting this feature to work for the background colors separately from the foreground colors.

The commits seem to just make the same `color_map_raw` assigned to both. I think what I'm looking for is a way to provide a `color_map_foreground_raw` and `color_map_background_raw`.

Let me know if this is outside the scope of this conversation and I can create a different one with the specific use case I have.

TheSpyder:
Oh! It was released in TinyMCE 7.5 but we forgot to close this ticket. Yes, expanding it to separate background and foreground would be a new ticket.

There is more interest in this capability than I expected. Maybe we should've made some sort of object overload for color maps instead of adding a new `_raw` option... oh well, too late now. It wouldn't be difficult to make a new PR adding background and foreground raw options, similar to the PR @andrewnicols did, if you're interested.

PR #9789, opened 2024-07-25
Support the use of CSS variables in `color_map` (#9788)

andrewnicols:
Related Ticket: TINY-11385

Description of Changes:
Add support for CSS Variables in colour maps
https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

- GH-9788

### DIFF
--- a/.changes/unreleased/tinymce-GH-9599-2024-07-25.yaml
+++ b/.changes/unreleased/tinymce-GH-9599-2024-07-25.yaml
@@ -1,6 +1,6 @@
 project: tinymce
 kind: Added
-body: Added support for CSS Variables in the `color_map`
+body: Added support for using raw CSS in the list of possible colours, using the `color_map_raw` property.
 time: 2024-07-25T10:42:28.314139+08:00
 custom:
   Issue: GH-9788
--- a/.changes/unreleased/tinymce-GH-9599-2024-07-25.yaml
+++ b/.changes/unreleased/tinymce-GH-9599-2024-07-25.yaml
@@ -1,0 +1,6 @@
+project: tinymce
+kind: Added
+body: Added support for CSS Variables in the `color_map`
+time: 2024-07-25T10:42:28.314139+08:00
+custom:
+  Issue: GH-9788
--- a/modules/tinymce/src/themes/silver/main/ts/ui/core/color/Options.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/ui/core/color/Options.ts
@@ -14,12 +14,21 @@ const mapColors = (colorMap: string[]): Menu.ChoiceMenuItemSpec[] => {
   const colors: Menu.ChoiceMenuItemSpec[] = [];
 
   for (let i = 0; i < colorMap.length; i += 2) {
-    colors.push({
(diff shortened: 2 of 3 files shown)


TheSpyder:
I wonder if we might be better with a setting to treat the color map values as raw instead of having a specific workaround for `var` 🤔

andrewnicols:
> I wonder if we might be better with a setting to treat the color map values as raw instead of having a specific workaround for `var` 🤔

What kind of thing did you have in mind? At the moment the `color_map` is an `Array<string>`. Are you thinking of making the first value in each pair an object or something maybe?

As a related note, it would be great to be able to disable the Custom color option too.

TheSpyder:
I'll need to check with the team, but I mean something like a new `color_map_raw: true` option. This would be used instead of the `var` regex in your code, with the setting on we would just do `value: colorMap[i]` all the time.


> As a related note, it would be great to be able to disable the Custom color option too.

Like this? 😉 
https://www.tiny.cloud/docs/tinymce/latest/user-formatting-options/#custom_colors

andrewnicols:
> I'll need to check with the team, but I mean something like a new `color_map_raw: true` option. This would be used instead of the `var` regex in your code, with the setting on we would just do `value: colorMap[i]` all the time.

Sounds pretty reasonable. Let me know if this is the way you'd like to go and I'll see if I can do this.
 
> > As a related note, it would be great to be able to disable the Custom color option too.
> 
> Like this? 😉 https://www.tiny.cloud/docs/tinymce/latest/user-formatting-options/#custom_colors

Ah perfect - I hadn't read down far enough apparently.

andrewnicols:
I've put together a quick PoC of that change on the existing branch. Currently as a separate commit for comparison.

andrewnicols:
Hi @TheSpyder ,

Did you have any thoughts on this approach?

TheSpyder:
@andrewnicols sorry for the delay. I think that's what I expected.

I did flag it with the team but we're having some challenges with the 7.3 release that is taking up our time. Hopefully we can take a look at it for 7.4.

andrewnicols:
Hi @TheSpyder ,

I was just wondering if there's any chance we can progress this now that Tiny 7.3 is out and established.

Thanks

TheSpyder:
I realised too late that it hadn't been included in 7.4. There should be a short turnaround to 7.5, and it looks like we'll be able to fit it in.

andrewnicols:
Thanks - that's great to know. We're about 4.5 months from our next code freeze so that gives us a bit of time to play with.

tiny-ben-tran (approved):
👍

andrewnicols:
I’ve made that change. Tha ns

ArvinJ-H (approved):

ltrouton (approved):

ArvinJ-H:
Added a doc ticket
https://ephocks.atlassian.net/browse/DOC-2557

ltrouton:
We have been able to verify that CI passes in a separate test branch so I am going to merge this. Thanks @andrewnicols

PR #9796, opened 2024-07-29
TINY-10890: Fix menu part of split button to be wrongly disabled

shanmen-tiny:
Related Ticket: TINY-10890

Description of Changes:
* Follow up PR, the button part of a split button was incorrectly disabled.
* Tests have been added, but only asserting the menu part of button and only when it's enabled, when it's disabled the whole div (button) structure gets disabled but the menu part and the arrow are missing those attributes (but they are still disabled) 
* Missed out re-enabling `contenteditable="true"` when switching from readonly -> selectionEnabled mode.

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/mode/Readonly.ts
+++ b/modules/tinymce/src/core/main/ts/mode/Readonly.ts
@@ -94,6 +94,10 @@ const toggleReadOnly = (editor: Editor, readOnlyMode: EditorReadOnlyType): void 
     if (Type.isBoolean(readOnlyMode) ? readOnlyMode : !readOnlyMode.selectionEnabled) {
       setContentEditable(body, false);
       switchOffContentEditableTrue(body);
+    } else {
+      if (editor.hasEditableRoot()) {
+        setContentEditable(body, true);
+      }
     }
   } else {
     unsetEditorReadonly(editor, body);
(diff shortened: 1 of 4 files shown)


MitchC1999 (approved):

modules/tinymce/src/core/main/ts/mode/Readonly.ts
```diff
@@ -94,6 +94,10 @@ const toggleReadOnly = (editor: Editor, readOnlyMode: EditorReadOnlyType): void
     if (Type.isBoolean(readOnlyMode) ? readOnlyMode : !readOnlyMode.selectionEnabled) {
       setContentEditable(body, false);
       switchOffContentEditableTrue(body);
+    } else {
+      if (editor.hasEditableRoot()) {
+        setContentEditable(body, true);
```
TheSpyder, line 99: Why isn't this just a call to `unsetEditorReadonly`?

TheSpyder (approved):

modules/tinymce/src/core/main/ts/mode/Readonly.ts
shanmen-tiny: We still want to keep the editor in "readonly" mode. But only allowing the cursor to be placed, so reverting the contenteditable to true.

PR #9800, opened 2024-07-30
TINY-11095: add inline label to toolbars

lorenzo-pomili:
Related Ticket: TINY-11095

Description of Changes:
example: 
Screenshot from 2024-09-02 07-12-24

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set
* [x] Docs ticket created (DOC-2509)


.changes/unreleased/tinymce-TINY-11095-2024-08-27.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Added
+body: New `addContextToolbar` to toolbar.
```
TheSpyder, line 3: That's incorrect?

TheSpyder (commented):
Can you please post a screenshot of this? I want to make sure it matches the designs.

.changes/unreleased/tinymce-TINY-11095-2024-08-27.yaml
lorenzo-pomili: It wasn't, I fixed it

lorenzo-pomili:
@TheSpyder I put a screenshot in the description

lorenzo-pomili:
I just notice that I did a mess, trying to rebase this branch and some not needed commits were in it, now it should be correct

modules/bridge/src/main/ts/ephox/bridge/components/toolbar/ToolbarLabel.ts
```diff
@@ -0,0 +1,50 @@
+import { StructureSchema } from '@ephox/boulder';
+import { Optional, Result } from '@ephox/katamari';
+
+import * as ComponentSchema from '../../core/ComponentSchema';
+
+export interface BaseToolbarLabelSpec<I extends BaseToolbarLabelInstanceApi> {
+  tooltip?: string;
+  icon?: string;
+  text?: string;
+  onSetup?: (api: I) => (api: I) => void;
```
spocke, line 10: Not sure we need a `onSetup` if there is no API we can add that later if we need an api.

modules/bridge/src/main/ts/ephox/bridge/components/toolbar/ToolbarLabel.ts
```diff
@@ -0,0 +1,50 @@
+import { StructureSchema } from '@ephox/boulder';
+import { Optional, Result } from '@ephox/katamari';
+
+import * as ComponentSchema from '../../core/ComponentSchema';
+
+export interface BaseToolbarLabelSpec<I extends BaseToolbarLabelInstanceApi> {
+  tooltip?: string;
+  icon?: string;
```
spocke, line 8: Icon for a label? Does that even make sense?

spocke:
How does this work with RTL?

Could we instead make the addContextToolbar take labeled groups and render those kind of what we have for the general toolbar. Kind of like this: https://www.tiny.cloud/docs/tinymce/latest/toolbar-configuration-options/#example-adding-toolbar-group-labels so it would be
```
editor.ui.registry.addContextToolbar('foo', [
  {
    label: 'Foo',
    items: [ 'bold', 'italic' ]
  },
  {
    label: 'Bar',
    items: [ 'underline', 'link' ]
  },
])
```

The benefit of that is that the label gets tied to the group it's for not just loose text things that is not tied to anything and rendering that in reverse order would be easier not sure how hard that is to add to silver though.

spocke (approved):

TheSpyder (approved):
I appreciate the effort you've put in here, and the way you're rendering the labels seems fine, but now that I've seen a screenshot I agree with @spocke about the configuration. This can't be a new registry component, it needs to be a label config added to toolbar groups.

What you've created is probably focusable in keyboard navigation, and note how you've had to add tooltips for a thing that is plain text. Both of these things seem incorrect.

We could arguably offer a setting to display the group `name` we already support, but that's there for accessibility and keeping the two separate is probably a good idea.

lorenzo-pomili:
@spocke @TheSpyder  about the current configuration:

`onSetup`: I copied the `onSetup` from the others `registry` thinking that it could be useful to conditional update the label or similar, but no strong option on it, we can remove it
`icon`: that's one other thing that I preserve coping the other implementations, I don't have a strong opinion on it so we can remove it
`tooltip`: needed because of the `icon` if we remove that we can remove both

about the keyboard navigation is not possible since there is no focusable behavior for the label component

I'd suggest to remove `icon`, `tooltip` and maybe `onSetup` but continue with this approach.

If we want to switch to the @spocke  suggested signature, we can try to fix the experimental branch that I made before this one, that at the moment it has another field that [assign label at the same index of the group](https://github.com/tinymce/tinymce/pull/9763/files#diff-68bd6ebbef3184d840bd60a3af8bc6b42ac4b479d9ff3cf79a4937794b03638aR197), there are some leftover and it needed to be styled but we can take the `css` made by @ztomaszyk  for this branch

lorenzo-pomili:
@spocke it seems that your implementation doesn't announce the label via screen reader, I'd try to improve my other spike [with this lines of code](https://github.com/tinymce/tinymce/pull/9763/files#diff-529da62212ff91b52ac173d981b37a9922118e14f12cc4449759ae654647b978R65-R69) now it has `fieldset`s for groups and the screen reader seems to work.

The problem is that I defined the title as: `title: 'label: ' + label` that's because it seems not to work if I do `title: label` the label appear in the DOM, but it's not announced, adding a fixed string it's also announced from my screen reader (I know it doesn't make any sense) :sweat_smile: , but I think maybe having the `fieldset`s for the accessibility could be the right way

TheSpyder:
@lorenzo-pomili that's a bug in the `title` implementation, it wasn't changed to render with `aria-label` as everything else was in 7.

lorenzo-pomili:
@spocke and @TheSpyder starting from the @spocke 's spike: https://github.com/tinymce/tinymce/compare/spike/toolbar-group-labels?expand=1 I create a new branch with the label, here the draft PR: https://github.com/tinymce/tinymce/pull/9900

if you think that could be the right way let me know, and I'll finish to fix that branch and open the PR

lorenzo-pomili:
closed, since [this one](https://github.com/tinymce/tinymce/pull/9900) is the correct version of this

PR #9802, opened 2024-07-30
TINY-11082: place the caret after the table when pasting 

ztomaszyk:
Related Ticket: TINY-11082

Description of Changes:
Caret would be placed inside the last cell of the pasted table. Now is placed after the table.

Pre-checks:
* [x] Changelog entry added
* [x] Tests have been added (if applicable) <---- adjusted existing tests

Review:
* [x] Milestone set

### DIFF
--- a/.changes/unreleased/tinymce-TINY-11082-2024-07-30.yaml
+++ b/.changes/unreleased/tinymce-TINY-11082-2024-07-30.yaml
@@ -1,7 +1,6 @@
 project: tinymce
 kind: Improved
-body: Caret would be placed inside the last cell of the pasted table. Now is placed
-  after the table.
+body: Pasting a table now places the cursor after the table instead of into the last cell.
 time: 2024-07-30T13:39:58.930774+02:00
 custom:
   Issue: TINY-11082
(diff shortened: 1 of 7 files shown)


.changes/unreleased/tinymce-TINY-11082-2024-07-30.yaml
```diff
@@ -0,0 +1,7 @@
+project: tinymce
+kind: Improved
+body: Caret would be placed inside the last cell of the pasted table. Now is placed
+  after the table.
```
TheSpyder, line 4:
```suggestion
body: Pasting a table now places the cursor after the table instead of into the last cell.
```

TheSpyder (approved):

modules/tinymce/src/core/test/ts/browser/content/insert/InsertContentTest.ts
```diff
@@ -604,7 +604,8 @@ describe('browser.tinymce.core.content.insert.InsertContentTest', () => {
             '</td>' +
           '</tr>' +
         '</tbody>' +
-      '</table>'
+      '</table>' +
+      `<p>${String.fromCharCode(160)}</p>` // UTF-16 char code '160' is NO-BREAK SPACE (U+00A0)
```
spocke, line 608: There is a Unicode module in Katamari that has a constant for this. Then you don't need the comment since the code would be explaining it. I wonder why the editor is configured as `raw` entities here. Ideally it would be better to be able to use entities but maybe there is a reason for it being configured that way.

spocke (approved):

HAFRMO (approved):

PR #9807, opened 2024-08-01
TINY-11131: Added custom state to tree structure

HAFRMO:
Related Ticket: TINY-11131

Description of Changes:
To display some sort of locked/readonly icon for locked/readonly templates and categories we needed to add the ability to add these.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/.changes/unreleased/tinymce-TINY-11131-2024-07-31.yaml
+++ b/.changes/unreleased/tinymce-TINY-11131-2024-07-31.yaml
@@ -1,6 +1,6 @@
 project: tinymce
 kind: Added
-body: Tree component now allows the addition of a custom icon
+body: Tree component now allows the addition of a custom icon.
 time: 2024-07-31T10:02:00.266569402+02:00
 custom:
   Issue: TINY-11131
--- a/.changes/unreleased/tinymce-TINY-11131-2024-07-31.yaml
+++ b/.changes/unreleased/tinymce-TINY-11131-2024-07-31.yaml
@@ -1,0 +1,6 @@
+project: tinymce
+kind: Added
+body: Tree component now allows the addition of a custom icon
+time: 2024-07-31T10:02:00.266569402+02:00
+custom:
+  Issue: TINY-11131
--- a/modules/bridge/src/main/ts/ephox/bridge/components/dialog/Tree.ts
+++ b/modules/bridge/src/main/ts/ephox/bridge/components/dialog/Tree.ts
@@ -32,23 +32,20 @@ export interface Tree {
   defaultSelectedId: Optional<Id>;
 }
 
-interface SpecialTreeState {
-  icon: string;
-  tooltip: string;
-}
-
 interface BaseTreeItemSpec {
   title: string;
   id: Id;
   menu?: ToolbarMenuButtonSpec;
-  specialState?: SpecialTreeState;
+  customStateIcon?: string;
+  customStateIconTooltip?: string;
 }
 
 interface BaseTreeItem {
   title: string;
   id: string;
   menu: Optional<ToolbarMenuButton>;
-  specialState: Optional<SpecialTreeState>;
+  customStateIcon: Optional<string>;
+  customStateIconTooltip: Optional<string>;
 }
 
 export interface DirectorySpec extends BaseTreeItemSpec {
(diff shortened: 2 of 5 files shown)


modules/bridge/src/main/ts/ephox/bridge/components/dialog/Tree.ts
```diff
@@ -71,6 +78,10 @@ const baseTreeItemFields = [
   ComponentSchema.title,
   FieldSchema.requiredString('id'),
   FieldSchema.optionOf('menu', MenuButtonSchema),
+  FieldSchema.optionObjOf('specialState', [
```
TheSpyder, line 81: This is not an acceptable config API name. Do you have any other suggestions for it?

HAFRMO: Not really, I genuinely wasn't sure what to call it. If you have any good idea I'd be happy to hear it.

spocke: Does it need to be it's own object. I assume that it's this way since the tooltip would require the icon but I think this could be just properties on the same object.

spocke: Or maybe it should just be a `readonly` flag then it gets the icon the tooltip could apply to the whole element so if you omit the icon then it would still display for text only nodes.

HAFRMO: I  made it a separate object since as you mention the tooltip needs the icon, and the icon does also need the tooltip. If they are two separately optional values it'd be somewhat of a weird mix. What'd we make to be a default icon? Default tooltip?
And while making this a readonly flag does seem easier in general this is a generic element and any tooltips would be equally generic. While we might be able to refer to the leaves as 'item' in the tooltip ( so `This Item is Locked` ) the directories would have to be directories, which would read strange when they are otherwise referred to as, for example, categories. Providing a tooltip is needed to make it read properly when used.

spocke: We already have that on other buttons you can create a toolbar button with just an icon and no tooltip. Here is an example: https://fiddle.tiny.cloud/OTqhev3xnu/0 so I think the easiest way is to just flatten this so it's just one object then is some user wants to add a icon without a tooltip then it's on them.

HAFRMO: Alright, sure, we can do that if you think it's best. What'll be the expected outcome for providing a tooltip but no icon? And do you have any suggestion for what you want to call them?

spocke: Can the tooltip be for the entire item I think that is how buttons work in the toolbar even if you provide a text to the button you get the tooltip on the whole thing. But would a screen reader read out both the text and the tooltip in that case can you focus on the locked item if so then it should tell the user that it's locked.

HAFRMO: So if you provide an icon, the tooltip is on the icon, and if you don't provide an icon the tooltip is added to the whole element itself? That seems a bit weird. If we want a tooltip for the whole item I think that should be a dedicated property, not a result from not having an icon?

spocke: I didn't fully understand what this PR was adding now I do. If the tooltip applies to the whole item it should be called `tooltip` but the `stateIcon` make more sense now. Can't figure out a better name for `state` though.

HAFRMO: The tooltip will not be to the whole item, this issue was specifically about an icon. If we wish to add a tooltip to the whole item that should be a separate issue. We do have until 7.4 to make whatever changes we want without it being breaking.

.changes/unreleased/tinymce-TINY-11131-2024-07-31.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Added
+body: Tree component now allows the addition of a custom icon
```
spocke, line 3:
```suggestion
body: Tree component now allows the addition of a custom icon.
```

spocke (approved):

modules/tinymce/src/themes/silver/test/ts/headless/components/tree/TreeTest.ts
```diff
@@ -119,6 +135,12 @@ describe('headless.tinymce.themes.silver.tree.TreeTest', () => {
     assertLeafSelectedState('File 3', true, file3);
   });
 
+  it('TINY-11131: Check that custom icon is correct', () => {
```
lorenzo-pomili, line 138: maybe add an assertion that says that the other elements don't have the icon

lorenzo-pomili (approved):

PR #9817, opened 2024-08-06
TINY-11122: add focus to view button api

ztomaszyk:
Related Ticket: TINY-11122

Description of Changes:
Added focus function to view button api. This PR is to be merged before https://github.com/tinymce/tinymce-premium/pull/280

Pre-checks:
* [ ] Changelog entry added
* [ ] ~~Tests have been added (if applicable)~~

Review:
* [ ] Milestone set
* [ ] ~~Docs ticket created (if applicable)~~

### DIFF
--- a/modules/bridge/src/main/ts/ephox/bridge/components/view/ViewButton.ts
+++ b/modules/bridge/src/main/ts/ephox/bridge/components/view/ViewButton.ts
@@ -10,6 +10,7 @@ interface ViewButtonApi {
 interface ViewToggleButtonApi extends ViewButtonApi {
   isActive: () => boolean;
   setActive: (state: boolean) => void;
+  focus: () => void;
 }
 
 interface BaseButtonSpec<Api extends ViewButtonApi> {
--- a/modules/tinymce/src/themes/silver/main/ts/ui/view/ViewButtons.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/ui/view/ViewButtons.ts
@@ -1,6 +1,6 @@
 import { AlloyComponent, Behaviour, Button as AlloyButton, GuiFactory, Memento, Replacing, SimpleOrSketchSpec } from '@ephox/alloy';
 import { Optional } from '@ephox/katamari';
-import { Attribute, Class } from '@ephox/sugar';
+import { Attribute, Class, Focus } from '@ephox/sugar';
 
 import { UiFactoryBackstageProviders } from '../../backstage/Backstage';
 import { renderReplaceableIconFromPack } from '../button/ButtonSlices';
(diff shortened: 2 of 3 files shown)


modules/tinymce/src/themes/silver/main/ts/ui/view/ViewButtons.ts
```diff
@@ -38,8 +38,13 @@ export const renderButton = (spec: ViewButtonWithoutGroup, providers: UiFactoryB
     };
     const isActive = () => Class.has(comp.element, ViewButtonClasses.Ticked);
 
+    const focus = () => {
+      const elm = comp.element;
+      Focus.focus(elm);
+    };
```
spocke, line 44:
```suggestion
    const focus = () => Focus.focus(comp.element);
```

spocke (approved):

spocke (approved):
Is it possible to make a test for this API?

lorenzo-pomili (approved):
+1 to this: https://github.com/tinymce/tinymce/pull/9817#pullrequestreview-2224362285

HAFRMO (approved):

ztomaszyk:
> Is it possible to make a test for this API?

It probably is, but I don't see the value in such a test. It would test if the focus from sugar library works? Or if the function is available in the onAction api? I tested the whole behavior in the https://github.com/tinymce/tinymce-premium/pull/280 so I thought it is sufficient enough.  

@spocke @lorenzo-pomili do you have an idea for a test that would make sense here?

lorenzo-pomili:
@ztomaszyk considering the feature it could be a little redundant, but you could add a test in [ViewTest.ts](https://github.com/tinymce/tinymce/blob/main/modules/tinymce/src/themes/silver/test/ts/browser/view/ViewTest.ts), so if for some reason the behavior will be removed from the plugin implementation it is tested anyway at the source

modules/bridge/src/main/ts/ephox/bridge/components/view/ViewButton.ts
```diff
@@ -10,6 +10,7 @@ interface ViewButtonApi {
 interface ViewToggleButtonApi extends ViewButtonApi {
   isActive: () => boolean;
   setActive: (state: boolean) => void;
+  focus: () => void;
```
TheSpyder, line 13: I really regret that views were implemented differently to dialogs, so methods like this have to go here instead of on `View`, but it's too late to change it now 😞

spocke: Yes it was supposed to be more like the sidebar just a div where we could dump things and the UI would be very minimal but since that things have been added. But I guess since we are going all react now we could just rewrite the UI in react if we want to add more things to it. It's not a complex UI in the advcode plugin, but then again I thought a comment react UI wouldn't as complex as it turned out to be. 😄

TheSpyder: React won't solve API issues, though. It might be a good opportunity to redo the View API in TinyMCE 8; I doubt anyone is using it other than us.

spocke: It would since this api would not be needed. We can remove the whole buttons api for views on tinymce v8. Since we only use it for the buttons in the advcode plugin the revisionhistory already does the whole UI in the view api with react.

PR #9821, opened 2024-08-07
TINY-11100: Fire input event during delete if one is not already fired.

HAFRMO:
Related Ticket: TINY-11100

Description of Changes:
The list plugin will fire an input event if none is fired by the delete command.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/.changes/unreleased/tinymce-TINY-11100-2024-08-05.yaml
+++ b/.changes/unreleased/tinymce-TINY-11100-2024-08-05.yaml
@@ -1,0 +1,7 @@
+project: tinymce
+kind: Fixed
+body: Deleting a selection in a list element would sometimes prevent the `input` event
+  from firing
+time: 2024-08-05T05:07:18.131117218+02:00
+custom:
+  Issue: TINY-11100
--- a/.changes/unreleased/tinymce-TINY-11100-2024-08-05.yaml
+++ b/.changes/unreleased/tinymce-TINY-11100-2024-08-05.yaml
@@ -1,7 +1,7 @@
 project: tinymce
 kind: Fixed
 body: Deleting a selection in a list element would sometimes prevent the `input` event
-  from firing
+  from being dispatched.
 time: 2024-08-05T05:07:18.131117218+02:00
 custom:
   Issue: TINY-11100
--- a/modules/tinymce/src/plugins/lists/main/ts/core/Delete.ts
+++ b/modules/tinymce/src/plugins/lists/main/ts/core/Delete.ts
@@ -274,14 +274,14 @@ const backspaceDeleteRange = (editor: Editor): boolean => {
   if (hasListSelection(editor)) {
     editor.undoManager.transact(() => {
       // Some delete actions may prevent the input event from being fired. If we do not detect it, we fire it ourselves.
-      let fireInput = true;
-      const inputHandler = () => fireInput = false;
+      let shouldFireInput = true;
+      const inputHandler = () => shouldFireInput = false;
 
       editor.on('input', inputHandler);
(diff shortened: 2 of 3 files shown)


modules/tinymce/src/plugins/lists/main/ts/core/Delete.ts
```diff
@@ -273,7 +273,18 @@ const hasListSelection = (editor: Editor) => {
 const backspaceDeleteRange = (editor: Editor): boolean => {
   if (hasListSelection(editor)) {
     editor.undoManager.transact(() => {
+      // Some delete actions may prevent the input event from being fired. If we do not detect it, we fire it ourselves.
+      let fireInput = true;
```
lorenzo-pomili, line 277: could it make sense to change this in something like: `shouldFireInput`?

lorenzo-pomili (approved):

modules/tinymce/src/plugins/lists/main/ts/core/Delete.ts
```diff
@@ -273,7 +273,18 @@ const hasListSelection = (editor: Editor) => {
...
+      const inputHandler = () => fireInput = false;
+
+      editor.on('input', inputHandler);
```
spocke, line 280: Since the `delete` command is our own command here is that we missed firing input on some override we make? So it fires when it's native only or something.

spocke: Right we made that input patch only for keyboard commands. Since this is also a keyboard action but not in core we don't fire one. I guess some browser do and some others dont?

HAFRMO: We default to the browser's native delete action in the delete command, where we very explicitly block the `beforeInput` and `input` events ( See `tinymce/modules/tinymce/src/core/main/ts/delete/DeleteUtils.ts:13` at `execCommandIgnoreInputEvents` which we get to from `execNativeDeleteCommand` )

modules/tinymce/src/plugins/lists/main/ts/core/Delete.ts
HAFRMO: probably a good idea.

TheSpyder (approved):

.changes/unreleased/tinymce-TINY-11100-2024-08-05.yaml
```diff
@@ -0,0 +1,7 @@
+project: tinymce
+kind: Fixed
+body: Deleting a selection in a list element would sometimes prevent the `input` event
+  from firing
```
spocke, line 4: Apparently fire is a bad word.

```suggestion
  from being dispatched.
```

modules/tinymce/src/plugins/lists/test/ts/browser/BackspaceDeleteTest.ts
```diff
@@ -61,6 +61,28 @@ describe('browser.tinymce.plugins.lists.BackspaceDeleteTest', () => {
...
+
+    editor.focus();
+    let inputEventCounter = 0;
+    const inputEventHandler = () => inputEventCounter++;
+
+    const li = editor.dom.select('li')[0].childNodes.item(0);
+    const rng = editor.dom.createRng();
+    rng.setStart(li, 0);
+    rng.setEnd(li, 1);
+    editor.selection.setRng(rng);
```
spocke, line 76: Replace this with `TinySelection.setSelection`.

spocke (approved):

Issue #9798, opened 2024-07-29
[media plugin] media_url_resolver bug


axww:
Sorry for the issue re-posting as there's no one follow up before.
I've tried code review however unable to solve in my ability.
Need your help, many thanks!

**Sample code on https://fiddle.tiny.cloud/**
```
<script type="text/javascript">
  tinymce.init({
    selector: "textarea",
    plugins: "media code",
    toolbar: "media code",
    media_url_resolver: (data) => {
      return new Promise((resolve) => {
        if (data.url.indexOf('a.com') !== -1) {
          resolve({ html: `<iframe src="z.com" width="400" height="400" ></iframe>` });
        } else {
          resolve({ html: '' });
        }
      });
    }
  });
</script>
<form method="post" action="dump.php"><textarea name="content"></textarea></form>
```

## 📝 Provide detailed reproduction steps (if any)

1. Click "Media" button.
2. Input "**a.com**" in "Source" box.

3. Switch to "Embed" card, the code in "Embed" should be: resolved to: (✔️ Expected result)
`<iframe src="z.com" width="400" height="400" ></iframe>`

4. Back to "General" card, change dimensions to different value. (Eg. width/height: 500)
6. Switch to "Embed" card again, the code in "Embed" become: (❌ Issue happens)
`<iframe src="a.com" width="500" height="500"></iframe>`

**Issue here:**
Only the first time input **without** dimensions change did the correct url resolve.
media_url_resolver did not work correctly after dimensions changed.

## 📃 Other details
- Browser: Microsoft Edge 127.0.2651.74
- OS: Windows 11
- Version: TinyMCE 7.2.1

---

If you'd like to see this fixed sooner, add a 👍 reaction to this post.

axww:
Note:
getEmbedHtml
handleUpdate
dataToHtml
updateHtml

axww:
```
case 'dimensions':
case 'altsource':
case 'poster':
  handleUpdate(api, detail.name, currentData.get());
  break;
}
```
handleUpdate function was only used by these cases.
Did not change source URL, should keep embed.

axww:
This BUG has been 2 years old, still not fixed.

PR #9824, opened 2024-08-09
Fix MediaResolver invalid after dimensions change

axww:
**GitHub issues:**
https://github.com/tinymce/tinymce/issues/9798

**Description of Changes:**

This is an issue of `Media` plugin, you can try following above issue page.
When convert rule settled using `media_url_resolver`, input source URL should be converted to other format.

Currently only the first time I input source URL the embed converted normally.
If I change dimensions, the embed src will become origin source URL.

Therefore I checked `modules\tinymce\src\plugins\media\main\ts\ui\Dialog.ts`
When `dimensions`, `altsource`, or `poster` `onChange`, it call function `handleUpdate`.

`handleUpdate` was only used here by these cases, so we modify it:
Change `data.source` from origin source to MediaResolver converted URL extract from `data.embed`.

In this way, converted URL in embed will not be replaced by origin source while wrapping new HTML. 
The problem will be solved. I've tested all scenes include dimensions and source change, all works well.

**Pre-checks:**
* [ ] Changelog entry added

**Review:**
* [ ] Milestone set

### DIFF
--- a/modules/tinymce/src/plugins/media/main/ts/ui/Dialog.ts
+++ b/modules/tinymce/src/plugins/media/main/ts/ui/Dialog.ts
@@ -210,7 +210,7 @@ const showDialog = (editor: Editor): void => {
         ? { ...dialogData, embed: '' }
         : dialogData;
 
-    const embed = dataToHtml(editor, data);
+    const embed = dataToHtml(editor, data.embed ? { ...data, source: HtmlToData.htmlToData(data.embed, editor.schema).source } : data);
     api.setData(wrap({
       ...data,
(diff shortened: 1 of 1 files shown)


TheSpyder:
@coderabbitai ignore

TheSpyder:
Sorry about that.

0w0c:
Really sorry I'm new to Git, this PR has some mistakes on branch and submit process.
Please just consider this PR as a possible fix solution I tried on my fork.
There should be better ways fixing the MediaResolver issue.

PR #9835, opened 2024-08-21
TINY-11155: Move creation of sidebars back to after `PostRender` event

MitchC1999:
Related Ticket: TINY-11155

Description of Changes:
* Sidebar creation was moved to the `SkinLoaded` event, causing the `ToggleSidebar` command to not work in an `init` event listener.
* Moved it back to inside the `PostRender` event

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/themes/silver/test/ts/browser/sidebar/SidebarTest.ts
+++ b/modules/tinymce/src/themes/silver/test/ts/browser/sidebar/SidebarTest.ts
@@ -3,7 +3,7 @@ import { context, describe, it } from '@ephox/bedrock-client';
 import { Fun } from '@ephox/katamari';
 import { McEditor } from '@ephox/mcagar';
 import { SugarBody, SugarElement, Traverse } from '@ephox/sugar';
-import { TinyHooks, TinyUiActions } from '@ephox/wrap-mcagar';
+import { TinyUiActions } from '@ephox/wrap-mcagar';
 
 import Editor from 'tinymce/core/api/Editor';
 import { Sidebar } from 'tinymce/core/api/ui/Ui';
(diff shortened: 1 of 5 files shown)


shanmen-tiny (approved):

modules/tinymce/src/themes/silver/test/ts/browser/sidebar/SidebarTest.ts
```diff
@@ -14,7 +15,7 @@ interface EventLog {
 
 describe('browser.tinymce.themes.silver.sidebar.SidebarTest', () => {
   const store = TestStore<EventLog>();
-  const hook = TinyHooks.bddSetupLight<Editor>({
+  const createEditor = async () => await McEditor.pFromSettings<Editor>({
```
TheSpyder, line 18: If adding and removing editors is important, the design intent of `TinyHooks` is you would have multiple `context` blocks inside the `describe` block and a new `hook` for each one. `bddSetupLight` uses `before` and `after` to create and remove the editor like you're doing manually here.

This also avoids having `it` blocks as siblings of `context` blocks, which will probably cause weird state bugs at some point.

TheSpyder (changes requested):

modules/tinymce/src/themes/silver/test/ts/browser/sidebar/SidebarTest.ts
MitchC1999: Thanks for the knowledge. I've so far only worked with McEditor when initializing editors with different configs for testing, so I appreciate this alternative.

TheSpyder: I'm glad you worked it out! I'm sure this stuff is documented somewhere, we should make sure it's in the onboarding process.

TheSpyder (approved):

PR #9839, opened 2024-08-22
TINY-10980: Fix issues with `uiEnabled` mode

shanmen-tiny:
Related Ticket: TINY-10980

Description of Changes:
* Renamed from `allowedModes` that takes a `string[]` to `allowedinReadonlyUiMode` bool
* Instead of checking of the mode and allowing the API to set the button to be enabled/disabled, now uses `onEnabled` or `onDisabled` to make sure that the buttons are in the correct state.
* Configuration that could change the button state:
  *  `DisablingConfigs` - on load, checks if there's enabled/disabled property for the button, or the buttons has `allowedinReadonlyUiMode` property
  * `buttonAPI` - if the `onSetup` button API is used, then will store the state. Example of this is used is: when `setEnabled(false)` is set to button, when switching to `uiEnabled` mode, even though the button has  `allowedinReadonlyUiMode`, we will still want it to be disabled, even after switching back to design, it should still use the store state.
  * `ReadOnly.receivingConfigConditional` - this is used to listen for mode switches, and to enable/disable the buttons as appropriately.
* Fixes this issue when the button is set to disabled with (`enabled: false` or `setEnabled(false)`), the button gets re-enabled. https://fiddle.tiny.cloud/u3YZQU1LS7/1
* Split out `Disabling.config` from modules/tinymce/src/themes/silver/main/ts/ui/dropdown/CommonDropdown.ts 

Pre-checks:
* [x] ~Changelog entry added~ will be added in epic/EPIC-114

Review:
* [x] Milestone set

### DIFF
--- a/modules/alloy/src/main/ts/ephox/alloy/behaviour/disabling/DisableState.ts
+++ b/modules/alloy/src/main/ts/ephox/alloy/behaviour/disabling/DisableState.ts
@@ -1,10 +1,10 @@
-import { Cell } from '@ephox/katamari';
+import { Singleton } from '@ephox/katamari';
 
 import { nuState } from '../common/BehaviourState';
 import { DisableState } from './DisableTypes';
 
 const init = (): DisableState => {
-  const lastDisabledState = Cell<boolean | undefined>(undefined);
+  const lastDisabledState = Singleton.value<boolean>();
 
   const readState = () => 'last disabled state:' + lastDisabledState.get();
 
--- a/modules/alloy/src/main/ts/ephox/alloy/behaviour/disabling/DisableState.ts
+++ b/modules/alloy/src/main/ts/ephox/alloy/behaviour/disabling/DisableState.ts
@@ -4,12 +4,12 @@ import { nuState } from '../common/BehaviourState';
 import { DisableState } from './DisableTypes';
 
 const init = (): DisableState => {
-  const lastDisabledState = Cell(false);
+  const lastDisabledState = Cell<boolean | undefined>(undefined);
 
   const readState = () => 'last disabled state:' + lastDisabledState.get();
 
   return nuState({
-    getLastDisabledState: () => lastDisabledState.get() === true,
+    getLastDisabledState: () => lastDisabledState.get(),
     setLastDisabledState: lastDisabledState.set,
     readState
   });
--- a/modules/bridge/src/main/ts/ephox/bridge/components/toolbar/ToolbarButton.ts
+++ b/modules/bridge/src/main/ts/ephox/bridge/components/toolbar/ToolbarButton.ts
@@ -1,4 +1,4 @@
-import { FieldSchema, StructureSchema, ValueType } from '@ephox/boulder';
+import { FieldSchema, StructureSchema } from '@ephox/boulder';
 import { Optional, Result } from '@ephox/katamari';
 
 import * as ComponentSchema from '../../core/ComponentSchema';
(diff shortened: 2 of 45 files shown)


modules/alloy/src/main/ts/ephox/alloy/behaviour/disabling/DisableState.ts
```diff
@@ -4,12 +4,12 @@ import { nuState } from '../common/BehaviourState';
 import { DisableState } from './DisableTypes';
 
 const init = (): DisableState => {
-  const lastDisabledState = Cell(false);
+  const lastDisabledState = Cell<boolean | undefined>(undefined);
```
MitchC1999, line 7: This could be a singleton
```suggestion
  const lastDisabledState = Singleton.value<boolean>();
```
Singetons don't need to be initialized. Their `get` method returns an optional though, so you could use `get().getOr(false)`. You can also see if it contains a value with `isSet`

MitchC1999 (approved):

modules/alloy/src/main/ts/ephox/alloy/behaviour/disabling/DisableState.ts
shanmen-tiny: Thanks, made changes to the `DisableApi` to return `Optional<boolean>` too.

modules/alloy/src/main/ts/ephox/alloy/behaviour/disabling/DisableState.ts
```diff
@@ -1,15 +1,15 @@
-import { Cell } from '@ephox/katamari';
+import { Singleton } from '@ephox/katamari';
 
 import { nuState } from '../common/BehaviourState';
 import { DisableState } from './DisableTypes';
 
 const init = (): DisableState => {
-  const lastDisabledState = Cell(false);
+  const lastDisabledState = Singleton.value<boolean>();
```
spocke, line 7: Why is this now a singleton can this be `disabled/enabled/<unknown>` ? It's either disabled or enabled it can't really be in a third state?

modules/bridge/src/main/ts/ephox/bridge/components/toolbar/ToolbarButton.ts
```diff
@@ -9,7 +9,7 @@ export interface BaseToolbarButtonSpec<I extends BaseToolbarButtonInstanceApi> {
   icon?: string;
   text?: string;
   onSetup?: (api: I) => (api: I) => void;
-  allowedModes?: string[];
+  allowedInReadonlyUiMode?: boolean;
```
spocke, line 12: I'm a bit concerned over how this new option will impact future requirements to disable/enable buttons on other editor states like the selection. That is why I specced out the context thing that it would just be predicates that return true/false if the toolbar should be enabled or not. Those predicates could then run when the UI element is rendered and when the UI needs to be refreshed that could be on modeChange and nodeChange and be cached for performance. The readonly channel could then just be replaced with a update disabled state channel that just runs the predicate to determine the new state.
I guess one could implement that on top having this simply be mapped to `enabledContext: 'any'` or something since setting this new option to true means that it will always be enabled. But what would happen if you set both say `allowedInReadonlyUiMode: true` and `enabledContext: 'editable'` or something. Then you have two conflicting options.

shanmen-tiny:
Closing PR in favour of TINY-11211.

PR #9847, opened 2024-08-29
TINY-11211: Add context for UI components

shanmen-tiny:
Related Ticket:  TINY-11211

Description of Changes:
* First phase, adding the base for `addContext` to the registry and using them in the backstage.
* Added the `context` spec to the Toolbar buttons for now, will be expanding to other type of buttons in the upcoming PRs.

Pre-checks:
* [x] ~Changelog entry added~ Will be consolidated into epic/EPIC-114

Review:
* [x] Milestone set

### DIFF
--- a/modules/bridge/src/main/ts/ephox/bridge/api/Registry.ts
+++ b/modules/bridge/src/main/ts/ephox/bridge/api/Registry.ts
@@ -32,6 +32,7 @@ export interface Registry {
   addAutocompleter: (name: string, spec: AutocompleterSpec) => void;
   addSidebar: (name: string, spec: SidebarSpec) => void;
   addView: (name: string, spec: ViewSpec) => void;
+  addContext: (name: string, func: (...args: any[]) => boolean) => void;
 
   getAll: () => {
     buttons: Record<string, ToolbarButtonSpec | GroupToolbarButtonSpec | ToolbarMenuButtonSpec | ToolbarSplitButtonSpec | ToolbarToggleButtonSpec>;
@@ -42,6 +43,7 @@ export interface Registry {
     icons: Record<string, string>;
     sidebars: Record<string, SidebarSpec>;
     views: Record<string, ViewSpec>;
+    contexts: Record<string, (...args: any[]) => boolean>;
   };
 }
 
@@ -52,12 +54,14 @@ export const create = (): Registry => {
   const icons: Record<string, string> = {};
   const contextMenus: Record<string, ContextMenuApi> = {};
   const contextToolbars: Record<string, ContextToolbarSpec | ContextFormSpec> = {};
+  const contexts: Record<string, (...args: any[]) => boolean> = {};
   const sidebars: Record<string, SidebarSpec> = {};
   const views: Record<string, ViewSpec> = {};
   const add = <T, S extends T>(collection: Record<string, T>, type: string) => (name: string, spec: S): void => {
     collection[name.toLowerCase()] = { ...spec, type };
   };
   const addIcon = (name: string, svgData: string) => icons[name.toLowerCase()] = svgData;
+  const addContext = (name: string, func: (...args: any[]) => boolean) => contexts[name.toLowerCase()] = func;
 
   return {
     addButton: add(buttons, 'button'),
@@ -75,6 +79,7 @@ export const create = (): Registry => {
     addSidebar: add(sidebars, 'sidebar'),
     addView: add(views, 'views'),
     addIcon,
+    addContext,
 
     getAll: () => ({
       buttons,
@@ -89,7 +94,8 @@ export const create = (): Registry => {
 
       contextToolbars,
       sidebars,
-      views
+      views,
+      contexts
     })
   };
 };
--- a/modules/bridge/src/main/ts/ephox/bridge/api/Registry.ts
+++ b/modules/bridge/src/main/ts/ephox/bridge/api/Registry.ts
@@ -32,7 +32,7 @@ export interface Registry {
   addAutocompleter: (name: string, spec: AutocompleterSpec) => void;
   addSidebar: (name: string, spec: SidebarSpec) => void;
   addView: (name: string, spec: ViewSpec) => void;
-  addContext: (name: string, pred: (...args: any[]) => boolean) => void;
+  addContext: (name: string, pred: (args: string) => boolean) => void;
 
   getAll: () => {
     buttons: Record<string, ToolbarButtonSpec | GroupToolbarButtonSpec | ToolbarMenuButtonSpec | ToolbarSplitButtonSpec | ToolbarToggleButtonSpec>;
@@ -43,7 +43,7 @@ export interface Registry {
     icons: Record<string, string>;
     sidebars: Record<string, SidebarSpec>;
     views: Record<string, ViewSpec>;
-    contexts: Record<string, (...args: any[]) => boolean>;
+    contexts: Record<string, (args: string) => boolean>;
   };
 }
 
@@ -54,14 +54,14 @@ export const create = (): Registry => {
   const icons: Record<string, string> = {};
   const contextMenus: Record<string, ContextMenuApi> = {};
   const contextToolbars: Record<string, ContextToolbarSpec | ContextFormSpec> = {};
-  const contexts: Record<string, (...args: any[]) => boolean> = {};
+  const contexts: Record<string, (args: string) => boolean> = {};
   const sidebars: Record<string, SidebarSpec> = {};
   const views: Record<string, ViewSpec> = {};
   const add = <T, S extends T>(collection: Record<string, T>, type: string) => (name: string, spec: S): void => {
     collection[name.toLowerCase()] = { ...spec, type };
   };
   const addIcon = (name: string, svgData: string) => icons[name.toLowerCase()] = svgData;
-  const addContext = (name: string, pred: (...args: any[]) => boolean) => contexts[name.toLowerCase()] = pred;
+  const addContext = (name: string, pred: (args: string) => boolean) => contexts[name.toLowerCase()] = pred;
 
   return {
     addButton: add(buttons, 'button'),
--- a/modules/bridge/src/main/ts/ephox/bridge/api/Registry.ts
+++ b/modules/bridge/src/main/ts/ephox/bridge/api/Registry.ts
@@ -32,7 +32,7 @@ export interface Registry {
   addAutocompleter: (name: string, spec: AutocompleterSpec) => void;
   addSidebar: (name: string, spec: SidebarSpec) => void;
   addView: (name: string, spec: ViewSpec) => void;
-  addContext: (name: string, func: (...args: any[]) => boolean) => void;
+  addContext: (name: string, pred: (...args: any[]) => boolean) => void;
 
   getAll: () => {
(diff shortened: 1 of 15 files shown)


modules/tinymce/src/themes/silver/test/ts/browser/editor/toolbar/ToolbarButtonContextTest.ts
```diff
@@ -0,0 +1,603 @@
...
+
+            scenario.buttonSetupEight(ed);
+          }
+        }, [], true);
+
+        it(`TINY-11211: Toolbar ${scenario.label} should be enabled in uiEnabled mode`, async () => {
+          const editor = hook.editor();
+          editor.mode.set('design');
+          scenario.assertButtonEnabled('t8');
+
```
shanmen-tiny, line 500: For this button configuration without a `NodeChange` or any other events, just `setEnabled(false)`. I'm not sure if we need to care about this? With this change, we always run on `NodeChange` or `SwitchMode`, however the `setEnabled` is only called once, so this button gets enabled. 
```
    ed.ui.registry.addButton('t8', {
      icon: 'italic',
      text: 'Test Menu Item 8',
      context: 'mode:design',
      onAction: Fun.noop,
      onSetup: (api) => {
        api.setEnabled(false);
        return Fun.noop;
      }
    });
```

modules/tinymce/src/themes/silver/main/ts/ButtonState.ts
```diff
@@ -0,0 +1,44 @@
...
+    broadcastEvents(uiRefs);
+  });
+
+  editor.on('SwitchMode', () => {
+    broadcastEvents(uiRefs);
+  });
+
+  editor.on('NodeChange', () => {
+    broadcastEvents(uiRefs);
+  });
```
MitchC1999, line 28:
```suggestion
  editor.on('init SwitchMode NodeChange', () => {
    broadcastEvents(uiRefs);
  });
```

modules/tinymce/src/themes/silver/main/ts/ButtonState.ts
```diff
@@ -0,0 +1,44 @@
...
+};
+
+const toggleButtonStateOnReceive = (shouldDisable: () => boolean): Behaviour.NamedConfiguredBehaviour<any, any> => Receiving.config({
```
MitchC1999, line 31:
```suggestion
const buttonStateToggleOnReceive = (shouldDisable: () => boolean): Behaviour.NamedConfiguredBehaviour<any, any> => Receiving.config({
```
`toggleButtonStateOnReceive` sounded in reference to "Toggle Button" to me. Suggestion is just opinion.

modules/tinymce/src/themes/silver/main/ts/backstage/Backstage.ts
```diff
@@ -58,7 +59,27 @@ const init = (lazySinks: { popup: () => Result<AlloyComponent, string>; dialog:
...
-    tooltips: TooltipsBackstage(lazySinks.dialog)
+    tooltips: TooltipsBackstage(lazySinks.dialog),
+    checkButtonContext: (specContext: string) => {
+      const resolveContext = () => {
+        if (Strings.contains(specContext, ':')) {
+          return specContext.split(':');
+        } else {
+          return [ specContext ];
+        }
+      };
```
MitchC1999, line 70:
```suggestion
      const resolveContext = () => Strings.contains(specContext, ':'))
        ? specContext.split(':')
        : [ specContext ];
```

modules/tinymce/src/themes/silver/main/ts/backstage/Backstage.ts
```diff
@@ -58,7 +59,27 @@ const init = (lazySinks: { popup: () => Result<AlloyComponent, string>; dialog:
...
+
+      const [ key, value ] = resolveContext();
+
+      return Obj.get(editor.ui.registry.getAll().contexts, key)
+        .fold(() => {
+          // Fallback to 'mode:design' if key is not found
+          return Obj.get(editor.ui.registry.getAll().contexts, 'mode').map((pred) => pred(editor, 'design')).getOr(false);
+        },
```
MitchC1999, line 78:
```suggestion
      const [ key, value ] = resolveContext();

      const contexts = editor.ui.registry.getAll().contexts;
      return Obj.get(contexts, key)
        .fold(() => {
          // Fallback to 'mode:design' if key is not found
          return Obj.get(contexts, 'mode').exists((pred) => pred(editor, 'design'));
        },
```

modules/tinymce/src/themes/silver/test/ts/browser/editor/toolbar/ToolbarButtonContextTest.ts
MitchC1999: The current behaviour would enable a button that is disabled when the mode is switched to design, and disable a button that is enabled when the mode is switched to readonly. We would expect that any button that needs to be enabled or disabled in a specific mode would be setup to trigger the api on the 'SwitchMode' event. Supporting anything else would be unexpected. With the new context API, that workaround is solved as well 👍

MitchC1999 (approved):

modules/tinymce/src/themes/silver/main/ts/ButtonState.ts
shanmen-tiny: Maybe just `ButtonState.toggleOnReceive` since we are importing `ButtonState` module and to call that it would be `ButtonState.toggleOnReceive`? Would that be clearer?

MitchC1999: Sounds good to me 👍

shanmen-tiny:
The tests are failing for two main reasons:
- Mainly the table buttons, they have their own `NodeChange` listeners, and the contexts' `NodeChange` is triggered after the listeners, hence the buttons are enabled.
  - https://github.com/tinymce/tinymce/blob/8b66cf289a1b5b39526ed9a5b5d9c8b2c539e3e5/modules/tinymce/src/plugins/table/main/ts/selection/SelectionTargets.ts#L178
- Undo/Redo buttons, they are listening for types of events as well. 
  - https://github.com/tinymce/tinymce/blob/8b66cf289a1b5b39526ed9a5b5d9c8b2c539e3e5/modules/tinymce/src/themes/silver/main/ts/ui/core/UndoRedo.ts#L6

modules/bridge/src/main/ts/ephox/bridge/api/Registry.ts
```diff
@@ -32,6 +32,7 @@ export interface Registry {
   addAutocompleter: (name: string, spec: AutocompleterSpec) => void;
   addSidebar: (name: string, spec: SidebarSpec) => void;
   addView: (name: string, spec: ViewSpec) => void;
+  addContext: (name: string, pred: (...args: any[]) => boolean) => void;
```
spocke, line 35: Wouldn't this be `string[]` instead of any since it's something we produce? Also if that is the case then we should change all the other occurrences of any. I think we should try to avoid the `any` type if we can.

modules/tinymce/src/themes/silver/main/ts/backstage/Backstage.ts
spocke: Not sure you need a if or a ternary splitting a string with `:` that doesn't have a `:` will produce an array with just the original item.

modules/tinymce/src/themes/silver/test/ts/browser/editor/toolbar/ToolbarButtonContextTest.ts
```diff
@@ -0,0 +1,585 @@
...
+          const editor = hook.editor();
+          editor.setContent('<p>a</p>');
+          scenario.assertButtonEnabled('t1');
+
+          editor.setEditableRoot(false);
+          scenario.assertButtonEnabled('t1');
+        });
+      });
+
+      context('Toolbar button spec with context: any, switching modes in editor setup ', () => {
```
spocke, line 135:
```suggestion
      context('Toolbar button spec with context: any, switching modes in editor setup', () => {
```

modules/tinymce/src/themes/silver/test/ts/browser/editor/toolbar/ToolbarButtonContextTest.ts
```diff
@@ -0,0 +1,585 @@
...
+          const editor = hook.editor();
+          editor.setContent('<p>a</p>');
+          scenario.assertButtonDisabled('t3');
+
+          editor.setEditableRoot(false);
+          scenario.assertButtonDisabled('t3');
+        });
+      });
+
+      context('Toolbar button spec with context: mode:readonly, switching modes in editor setup ', () => {
```
spocke, line 258:
```suggestion
      context('Toolbar button spec with context: mode:readonly, switching modes in editor setup', () => {
```

spocke (approved):
Since this is new API it needs a changelog item(s) about the new APIs.

modules/bridge/src/main/ts/ephox/bridge/api/Registry.ts
shanmen-tiny: The first argument would be the Editor, but since this is in bridge. It doesn't have the type for Editor since it only deals with `specs` 😞 I was thinking of (editor: Editor, args: string[])

spocke: Hmm, do we need the editor since the registry is on the editor you would have that in scope in the closure anyway. We should log a tech dept to get tid of bridge.

shanmen-tiny: Right 🤦‍♂️ Pushed a commit to fix that.

shanmen-tiny:
> This is a little concerning since it now acts on nodeChange and that disabled state update will interfere with the disabled state change in any custom setup function. Previously it would only interfere when there was a mode change not it's way more often. I think we need to discuss how that is supposed to work to that this doesn't become a breaking change.

Tests are now passing. Now it shouldn't change from the previous behaviour, for example: t8 and t12. https://github.com/tinymce/tinymce/commit/c47a73f9d7dceade86333f0af74359eba8a1c635

So the button would be disabled onLoad for t8, and `NodeChange` would no longer be executed. For `t12` (button spec without onSetup but just `enabled: false`), one example of this is being used in `UndoRedo`, I guess we can keep the same behaviour if there's nothing in the `onSetup` the `enabled: false` would only be only used on init.

modules/tinymce/src/themes/silver/main/ts/backstage/Backstage.ts
```diff
@@ -58,7 +59,23 @@ const init = (lazySinks: { popup: () => Result<AlloyComponent, string>; dialog:
...
+      const resolveContext = () => Strings.contains(specContext, ':') ? specContext.split(':') : [ specContext ];
+
+      const [ key, value ] = resolveContext();
+      const contexts = editor.ui.registry.getAll().contexts;
+      const enabledInContext = Obj.get(contexts, key)
+        .fold(
+          // Fallback to 'mode:design' if key is not found
+          () => Obj.get(contexts, 'mode').map((pred) => pred('design')).getOr(false),
+          (pred) => pred(value)
+        );
```
MitchC1999, line 73:
```suggestion
      const enabledInContext = Obj.get(contexts, key).exists((pred) => pred(value))
      // Fallback to 'mode:design' if key is not found
      || Obj.get(contexts, 'mode').exists((pred) => pred('design'));
```

MitchC1999 (approved):

PR #9852, opened 2024-09-03
TINY-11211: Add context for UI components - 2 

shanmen-tiny:
Related Ticket: TINY-11211

Description of Changes:
* Changes for menubar, menu items and toolbar menu button
* `addContext('mode:!readonly')` for menu bar buttons, which should be disabled in `readonly` and allowed in other modes.

Pre-checks:
* [x] ~Changelog entry added~ To be added in epic/EPIC-114

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/themes/silver/main/ts/ui/core/Context.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/ui/core/Context.ts
@@ -1,0 +1,27 @@
+import { Fun } from '@ephox/katamari';
+
+import Editor from 'tinymce/core/api/Editor';
+
+const register = (editor: Editor): void => {
+  editor.ui.registry.addContext('editable', (editor: Editor) => {
+    return editor.selection.isEditable();
+  });
+
+  editor.ui.registry.addContext('mode', (editor: Editor, mode: string) => {
+    return editor.mode.get() === mode;
+  });
+
+  editor.ui.registry.addContext('any', Fun.always);
+
+  editor.ui.registry.addContext('formatting', (editor: Editor, format: string) => {
+    return editor.formatter.canApply(format);
+  });
+
+  editor.ui.registry.addContext('insert', (editor: Editor, child: string) => {
+    return editor.schema.isValidChild(editor.selection.getNode().tagName, child);
+  });
+};
+
+export {
+  register
+};
--- a/modules/tinymce/src/themes/silver/main/ts/ui/core/Context.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/ui/core/Context.ts
@@ -3,21 +3,21 @@ import { Fun } from '@ephox/katamari';
 import Editor from 'tinymce/core/api/Editor';
 
 const register = (editor: Editor): void => {
-  editor.ui.registry.addContext('editable', (editor: Editor) => {
+  editor.ui.registry.addContext('editable', () => {
     return editor.selection.isEditable();
   });
 
-  editor.ui.registry.addContext('mode', (editor: Editor, mode: string) => {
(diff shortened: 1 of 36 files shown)


modules/tinymce/src/themes/silver/main/ts/ui/menus/menubar/SilverMenubar.ts
```diff
@@ -49,7 +49,8 @@ const factory: UiSketcher.SingleSketchFactory<SilverMenubarDetail, SilverMenubar
         text: m.text,
         fetch: (callback) => {
           callback(m.getItems());
-        }
+        },
+        context: '!readonly'
```
spocke, line 53: Maybe something we need to discuss but we might need to have the menubar enabled in readonly but disable the menu items instead since if comments in readonly mode is enabled and the editor is in readonly and there is no add comment toolbar button and the user wants to add a comment using the menu item then there is no way to get to it?

spocke (approved):

modules/tinymce/src/themes/silver/main/ts/ui/menus/menubar/SilverMenubar.ts
shanmen-tiny: Wouldn't that be a behavioural change for readonly mode, as previously everything is disabled but now the buttons in the menubar are enabled?  However, for existing integrations, since the default menu items have `mode:design` only, no buttons should be enabled when it's in readonly mode.

modules/tinymce/src/themes/silver/main/ts/ui/core/Context.ts
```diff
@@ -20,6 +20,10 @@ const register = (editor: Editor): void => {
   editor.ui.registry.addContext('insert', (child: string) => {
     return editor.schema.isValidChild(editor.selection.getNode().tagName, child);
   });
+
+  editor.ui.registry.addContext('!readonly', () => {
+    return editor.mode.get() !== 'readonly';
+  });
```
MitchC1999, line 26:
```suggestion
  editor.ui.registry.addContext('mode:!readonly', () => {
    return editor.mode.get() !== 'readonly';
  });
```
Would this fit into mode contexts?

MitchC1999 (approved):

modules/tinymce/src/themes/silver/main/ts/ui/menus/menubar/SilverMenubar.ts
spocke: Yes it is a big change however we are not fully supporting comments in readonly mode if only the toolbar buttons works but menu items doesn't. If we are not fully supporting both UI elements then we need to make sure that people are aware that we only support comments though toolbar in readonly/comment only mode.

shanmen-tiny: Ah, sorry. The menubar buttons will be enabled when it's not in readonly, so the menu items in comments will have the same context `!readonly` and since we are going to register a new mode, so as long as it's not in `readonly` mode the menu items will be enabled. This applies to the toolbar buttons as well.

spocke: Ok cool just wanted to double check. Right it checking the mode specifically not the readonly state. 👍

modules/tinymce/src/themes/silver/main/ts/ui/core/Context.ts
shanmen-tiny: Pushed changes 👍  appreciate if this can be looked into!

MitchC1999 (approved):

PR #9859, opened 2024-09-09
TINY-11217: Prevent child `li` inherit parent list item style

ArvinJ-H:
Related Ticket: 
TINY-11217

Description of Changes:
- Added a default styling for `li` element to prevent child list item inheriting parent element

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


TheSpyder:
Moving to draft until Product decides what they want to happen with this

.changes/unreleased/tinymce-TINY-11217-2024-09-10.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Added
+body: Added `apply_list_formatting` config to disable the list item styling
```
spocke, line 3: This needs to be updated to the new default styles for list elements.

modules/oxide/src/less/skins/content/default/content.less
```diff
@@ -79,3 +79,12 @@ code {
     padding-right: 1rem;
   }
 }
+
+li {
+  color: black;
```
spocke, line 84: Should this be a variable, and should this be done in all the content skins?

modules/tinymce/src/core/main/ts/fmt/ApplyFormat.ts
```diff
@@ -41,6 +41,7 @@ const canFormatBR = (editor: Editor, format: ApplyFormat, node: HTMLBRElement, p
 };
 
 const applyStyles = (dom: DOMUtils, elm: Element, format: ApplyFormat, vars: FormatVars | undefined) => {
+
```
spocke, line 44: I guess this line is not needed anymore?

modules/tinymce/src/core/test/ts/browser/fmt/ListIndentationFormatTest.ts
```diff
@@ -0,0 +1,32 @@
...
+describe('browser.tinymce.core.fmt.ListItemFormatTest', () => {
+  const hook = TinyHooks.bddSetup<Editor>({
+    apply_list_formatting: false,
+    base_url: '/project/tinymce/js/tinymce',
+    plugins: 'lists'
+  }, [], true);
+
+  context('Apply list formatting', () => {
+
+    it('TINY-8961: Apply multiple formats sequentially to entire list', () => {
```
spocke, line 16: Maybe we can inspect the computed styles with Css.get on the LI elements to see if this works correctly?

modules/tinymce/src/core/test/ts/browser/fmt/ListItemFormatTest.ts
```diff
@@ -374,4 +374,116 @@ describe('browser.tinymce.core.fmt.ListItemFormatTest', () => {
       );
     });
   });
+
+  context('Apply list formatting', () => {
+    const hook = TinyHooks.bddSetupLight<Editor>({
+      apply_list_formatting: false,
```
spocke, line 380: This is no longer needed? Does the test make sense anymore?

spocke (dismissed):

modules/tinymce/src/core/test/ts/browser/fmt/ListItemFormatTest.ts
ArvinJ-H: this test is testing for list format with indentation, should still be applicable, but in the wrong place. I'll move this test to the `ListIndentationFormatTest.ts`

modules/oxide/src/less/skins/content/default/content.less
ArvinJ-H: no, the variables are used internally, we don't want to draw any correlation with the `content.less`. I have removed all the variable references

modules/oxide/src/less/skins/content/dark/content.less
```diff
@@ -85,3 +85,11 @@ code {
     padding-right: 1rem;
   }
 }
+
+li {
+  color: #ffffff;
```
kimwoodfield, line 90: Hey @ArvinJ-H  I _think_ we could potentially reduce some of the code duplication here by using mixins and imports.

Could we define a `li` style with everything from `background-color` to `font-family` and then just update the `color` in each file as needed while just referencing the mixin for all of the remaining styles that we need? I'm not sure if it's a convention we follow here but could help with that.

kimwoodfield (approved):

tiny-ben-tran (approved):
It looks like the tests can be applied to both unorder and order lists. And there is only one tests for the order list. 

As far as I can understand the functionality of this change is that a list item should have the same formats after indentation If that is correct, please add a test for it

modules/tinymce/src/core/test/ts/browser/fmt/ListIndentationFormatTest.ts
```diff
@@ -0,0 +1,133 @@
...
+    const fontWeight = Css.get(sugarItem2, 'font-weight') === '400';
+    const fontFamily = Css.get(sugarItem2, 'font-family') ===
+      (platform.isChromium() ?
+        '-apple-system, "system-ui", "Segoe UI", Roboto, Oxygen-Sans, Ubuntu, Cantarell, "Helvetica Neue", sans-serif' :
+        '-apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Oxygen-Sans, Ubuntu, Cantarell, "Helvetica Neue", sans-serif');
+
+    return color && backgroundColor && fontStyle && fontWeight && fontFamily;
+  };
+
+  context('Apply list formatting', () => {
```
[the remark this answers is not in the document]
spocke: Context is also about clumping together different thing say a module has three functions A, B, C I tend to use context to wrap A, B, C and have multiple it's for each of those functions since then you can set only on A if you are working on fixing some bug in A and just run the tests for that. So if the context here is applying list formatting you can wrap that in a context and focus on that if you have more than one test for it.

shanmen-tiny (approved):

modules/tinymce/src/core/test/ts/browser/fmt/ListIndentationFormatTest.ts
```diff
@@ -0,0 +1,133 @@
...
+    const color = Css.get(sugarItem2, 'color') === 'rgb(34, 47, 62)';
+    const backgroundColor = Css.get(sugarItem2, 'background-color') === 'rgba(0, 0, 0, 0)';
+    const fontStyle = Css.get(sugarItem2, 'font-style') === 'normal';
+    const fontWeight = Css.get(sugarItem2, 'font-weight') === '400';
+    const fontFamily = Css.get(sugarItem2, 'font-family') ===
+      (platform.isChromium() ?
+        '-apple-system, "system-ui", "Segoe UI", Roboto, Oxygen-Sans, Ubuntu, Cantarell, "Helvetica Neue", sans-serif' :
+        '-apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Oxygen-Sans, Ubuntu, Cantarell, "Helvetica Neue", sans-serif');
+
+    assert.isTrue(color && backgroundColor && fontStyle && fontWeight && fontFamily);
```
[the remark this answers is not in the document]
ArvinJ-H: done

modules/tinymce/src/core/test/ts/browser/fmt/ListIndentationFormatTest.ts
ltrouton, line 1: I don't see any test cases for color and background color. Can this be added?

modules/tinymce/src/core/test/ts/browser/fmt/ListIndentationFormatTest.ts
```diff
@@ -0,0 +1,131 @@
...
+      const editor = hook.editor();
+
+      editor.setContent('<ul><li>Item 1</li><li class="testClass">Item 2</li></ul>');
+      TinySelections.setSelection(editor, [ 0 ], 0, [ 0 ], 1);
+      editor.formatter.apply('bold');
+      editor.formatter.apply('italic');
+      editor.formatter.apply('underline');
+      TinySelections.setCursor(editor, [ 0, 1 ], 1);
+
+      TinyContentActions.keydown(editor, Keys.tab());
```
ltrouton, line 44: Can we assert the content after this as well. We want to make sure the `Tab` is actually indenting the list item

.changes/unreleased/tinymce-TINY-11217-2024-09-10.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Added
+body: Added new default styles for list elements.
```
ltrouton, line 3:
```suggestion
body: Added new default `li` element styles to the content CSS to prevent child list elements from inheriting certain parent list item styles.
```

modules/tinymce/src/core/test/ts/browser/fmt/ListIndentationFormatTest.ts
ArvinJ-H: added

modules/tinymce/src/core/test/ts/browser/fmt/ListIndentationFormatTest.ts
ArvinJ-H: added

ltrouton (approved):

TheSpyder:
The problem with where this fix ended up is it won't help customers who already customised our defaults, and it may have unwanted side effects for customers that do use our defaults.

I guess we just have to wait and see whether my concerns are valid.

ArvinJ-H:
> The problem with where this fix ended up is it won't help customers who already customised our defaults, and it may have unwanted side effects for customers that do use our defaults.
> 
> I guess we just have to wait and see whether my concerns are valid.

It should be fine, if they are using `content_style`, this setting will have higher precedence over the changes in `content.less`.

This is the same with `importcss` plugin

spocke:
> > The problem with where this fix ended up is it won't help customers who already customised our defaults, and it may have unwanted side effects for customers that do use our defaults.
> > I guess we just have to wait and see whether my concerns are valid.
> 
> It should be fine, if they are using `content_style`, this setting will have higher precedence over the changes in `content.less`.
> 
> This is the same with `importcss` plugin

It's some what breaking since before you only needed to add a font styles to the body and the li elements would inheir that style from the body. Now you need to define both but it would be pretty apperent that you need to add that since the list is not rendered properly.

ArvinJ-H:
> > > The problem with where this fix ended up is it won't help customers who already customised our defaults, and it may have unwanted side effects for customers that do use our defaults.
> 
> > > I guess we just have to wait and see whether my concerns are valid.
> 
> > 
> 
> > It should be fine, if they are using `content_style`, this setting will have higher precedence over the changes in `content.less`.
> 
> > 
> 
> > This is the same with `importcss` plugin
> 
> 
> 
> It's some what breaking since before you only needed to add a font styles to the body and the li elements would inheir that style from the body. Now you need to define both but it would be pretty apperent that you need to add that since the list is not rendered properly.

Yeah true, maybe we should just remove `content.less` all together to avoid more confusion?

PR #9860, opened 2024-09-09
TINY-10847: Fix clipboard text content when copying tables

ltrouton:
Related Ticket: TINY-10847

Description of Changes:
* Update the table clipboard handler so that the replicated table is added to the DOM. We can use the browser to give us the `innerText` of the table element which separates cells with a `\t` character and rows with a `\n` character

Pre-checks:
* [X] Changelog entry added

Review:
* [X] Milestone set

### DIFF
--- a/modules/tinymce/src/core/test/ts/browser/paste/InternalClipboardTest.ts
+++ b/modules/tinymce/src/core/test/ts/browser/paste/InternalClipboardTest.ts
@@ -141,7 +141,7 @@ describe('browser.tinymce.core.paste.InternalClipboardTest', () => {
                 '<td>b</td>\n' +
               '</tr>\n' +
             '</tbody>\n' +
-          '</table>', 'ab');
+          '</table>', 'a\tb');
       TinyAssertions.assertSelection(editor, [ 0, 0, 0, 1, 0 ], 0, [ 0, 0, 0, 1, 0 ], 0);
     });
 
--- a/modules/tinymce/src/models/dom/main/ts/table/actions/Clipboard.ts
+++ b/modules/tinymce/src/models/dom/main/ts/table/actions/Clipboard.ts
@@ -1,6 +1,6 @@
 import { Arr, Optional } from '@ephox/katamari';
 import { CopySelected, TableFill, TableLookup } from '@ephox/snooker';
-import { SugarElement, SugarElements, SugarNode } from '@ephox/sugar';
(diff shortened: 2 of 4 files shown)


modules/tinymce/src/models/dom/main/ts/table/actions/Clipboard.ts
```diff
@@ -25,15 +25,40 @@ const extractSelected = (cells: SugarElement<HTMLTableCellElement>[]): Optional<
...
+const getTextContent = (editor: Editor, replicaElements: SugarElement<HTMLTableElement>[]): string => {
+  const doc = editor.getDoc();
+  const dos = SugarShadowDom.getRootNode(SugarElement.fromDom(editor.getBody()));
+
+  // Set up offscreen div so that the extracted table element can be inserted into the DOM
+  // TINY-10847: If the table element is detached from the DOM, calling innerText is equivalent to calling
+  // textContent which does not include '\n' and '\t' characters to separate rows and cells respectively
+  const offscreenDiv = SugarElement.fromTag('div', doc);
+  Attribute.set(offscreenDiv, 'data-mce-bogus', 'all');
+  Css.setAll(offscreenDiv, {
```
spocke, line 37: Not sure you need this since you add and remove synchronously there wouldn't be time for the browser do do a paint cycle.

spocke (approved):
There is a extra `i` in `copying` in the PR title.

modules/tinymce/src/models/dom/main/ts/table/actions/Clipboard.ts
ltrouton: That's true. Although, I would prefer to just follow usual offscreen div convention of setting the styles on it to be safe

shanmen-tiny (approved):

modules/tinymce/src/core/test/ts/browser/paste/InternalClipboardTest.ts
```diff
@@ -141,7 +141,7 @@ describe('browser.tinymce.core.paste.InternalClipboardTest', () => {
                 '<td>b</td>\n' +
               '</tr>\n' +
             '</tbody>\n' +
-          '</table>', 'ab');
+          '</table>', 'a\tb');
```
TheSpyder, line 144: but of _course_ we had tests that asserted the incorrect behaviour 🫣 🤣

TheSpyder (approved):

PR #9876, opened 2024-09-16
TINY-11257: fix resize handle placing

ztomaszyk:
Related Ticket: TINY-11257

Description of Changes:
Resize handle would be incorrectly located in the right corner when no other components were added to status bar.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/.changes/unreleased/tinymce-TINY-11257-2024-09-16.yaml
+++ b/.changes/unreleased/tinymce-TINY-11257-2024-09-16.yaml
@@ -1,0 +1,7 @@
+project: tinymce
+kind: Fixed
+body: Resize handle would be incorrectly located when no other components were added
+  to status bar.
+time: 2024-09-16T17:08:32.128063+02:00
+custom:
+  Issue: TINY-11257
--- a/.changes/unreleased/tinymce-TINY-11257-2024-09-16.yaml
+++ b/.changes/unreleased/tinymce-TINY-11257-2024-09-16.yaml
@@ -1,7 +1,7 @@
 project: tinymce
 kind: Fixed
-body: Resize handle would be incorrectly located when no other components were added
-  to status bar.
+body: The editor resize handle was incorrectly rendered when all components were
+  removed from the status bar.
 time: 2024-09-16T17:08:32.128063+02:00
 custom:
   Issue: TINY-11257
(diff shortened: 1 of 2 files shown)


modules/oxide/src/less/theme/components/statusbar/statusbar.less
```diff
@@ -190,6 +190,10 @@
       }
     }
   }
+
+  .tox-statusbar__resize-handle:only-child {
+    flex: 1;
```
spocke, line 195: This will make the resize handle be the width of the editor not just the corner of the editor. There are other way we could fix this one is to set justify-content: end on the parent container.
Or you could use `margin-left: auto` so that it has a margin of the full width in the container probably the simplest one.
Another is to keep a spacing element in there the `tox-statusbar__text-container` could be kept in there and be empty to push out the area so that the resize element is at the end.
Also remember to test with RTL it seem that the `margin-left: auto` works better than the `flex` approach here not sure why. Copy over the ar.js file into `js/tinymce/langs` then set `language: 'ar`` to test RTL.

spocke (changes requested):

modules/oxide/src/less/theme/components/statusbar/statusbar.less
ztomaszyk: I went with `margin-left: auto`, tested with RTL ✅

.changes/unreleased/tinymce-TINY-11257-2024-09-16.yaml
```diff
@@ -0,0 +1,7 @@
+project: tinymce
+kind: Fixed
+body: Resize handle would be incorrectly located when no other components were added
+  to status bar.
```
TheSpyder, line 4: The bar has things on it by default, so the integrator action to trigger the issue was removing things.

```suggestion
body: The editor resize handle was incorrectly rendered when all components were
  removed from the status bar.
```

TheSpyder (approved):

modules/oxide/src/less/theme/components/statusbar/statusbar.less
TheSpyder: That suggests there wasn't a bug in RTL 🤔 

If there was, you'd have to use `margin-inline-start` instead

ztomaszyk: It wasn't, the bug occurred for both LTR and RTL

spocke (approved):

PR #9906, opened 2024-10-07
TINY-11335: add placeholder styles and make it resizable

ztomaszyk:
Related Ticket: TINY-11335

Description of Changes:
Upploadcare placeholder styles

Pre-checks:
* [x] Changelog entry added

Review:
* [ ] Milestone set


TheSpyder:
Should this wait until it can be targeted at an epic branch? It's going to get confusing if we have multiple nested layers of feature branches without using automated stacked commits.

modules/oxide/src/less/theme/content-ui.less
```diff
@@ -24,5 +24,6 @@
 @import 'content/table-of-contents/table-of-contents';
 @import 'content/transparent/transparent';
 @import 'content/table/table';
+@import 'content/upploadcare/upploadcare';
```
TheSpyder, line 27: typo in the folder and filename
```suggestion
@import 'content/uploadcare/uploadcare';
```

.changes/unreleased/tinymce-TINY-11335-2024-10-08.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Improved
+body: Elements with `data-mce-resize` set to true are now resizable.
```
TheSpyder, line 3:
```suggestion
body: All elements with `data-mce-resize` set to true are now resizable.
```

modules/tinymce/src/core/main/ts/api/dom/ControlSelection.ts
```diff
@@ -123,6 +123,10 @@ const ControlSelection = (selection: EditorSelection, editor: Editor): ControlSe
       return false;
     }
 
+    if (elm.getAttribute('data-mce-resize') === 'true') {
+      return true;
+    }
```
TheSpyder, line 128: I suspect this is a little too general of a feature, for example the body check below is probably important. Maybe limit it to `span` and/or `div` tags?

spocke: Maybe we also restrict it to non editable elements only since you can't control select anything. But I wonder if these placeholders should really be resizable. You could add two new predicates to dom/NodeType that is `isDiv` and `isSpan` you can then also the the NodeType.isContentEditableFalse to check if it's non editable.

- [ ] Add test that selects a div and a span that has this attribute set and validate that it has resize handles and possibly even try to resize if it we have some code that does that fake drag/dropping.

.changes/unreleased/tinymce-TINY-11335-2024-10-08.yaml
spocke: I wonder if we should hide this it's really not a documented feature also if we document it then we kind of need to support the way it works today. It also a editor internal temporary attribute so explaining all that would require a docs section on it's own. Also it's oddly named I really should have named it `data-mce-resizable` 😄

modules/tinymce/src/core/main/ts/api/dom/ControlSelection.ts
TheSpyder: Update: I have confirmed with Fredrik that this is not intentional. The resize should only apply to our placeholder, not any element with the resize attribute.

- [ ] Change to something like `span.data-mce-resize`, similar to the preview object selector

spocke (changes requested):

modules/tinymce/src/core/main/ts/api/dom/ControlSelection.ts
ztomaszyk: I could do `.tox-uploadcare-placeholder` to be extra specific, but I went with more general selector because I thought the core code should be dummy here and not aware of the uploadcare plugin features 🤔

lorenzo-pomili:
@TheSpyder about the target in `mrgit.json` at the moment to change it we have to remove the labels, and since the PR of `feature/TINY-11095-2` will be not merged (this is the version that should be merged: https://github.com/tinymce/tinymce/pull/9900), I propose to leave the target for now, and when we merge `feature/TINY-11095-3` we will do a PR to fix all the labels

modules/tinymce/src/core/main/ts/api/dom/ControlSelection.ts
TheSpyder: That's fair enough, if we're too general it will encourage integrators to enable resize on things that can't be resized.

How about `span.mce-placeholder`, so it's at least somewhat similar to `span.mce-preview-object`

.changes/unreleased/tinymce-TINY-11335-2024-10-08.yaml
TheSpyder: yes, I forgot to update this after my other comment

- [ ] delete changelog entry

lorenzo-pomili:
@TheSpyder I see, sorry I misunderstood that, yes in that case the target should be `epic/EPIC-117`

PR #9895, opened 2024-10-01
TINY-11332: Bump dependencies

jscasca:
Related Ticket: TINY-11332

Description of Changes:
* Bump dependencies

Pre-checks:
* [x] Changelog entry added
* [x] Tests have been added (if applicable)
* [x] Branch prefixed with `feature/`, `hotfix/` or `spike/`

Review:
* [x] Milestone set
* [x] Docs ticket created (if applicable)

GitHub issues (if applicable):

PR #9907, opened 2024-10-08
TINY-11331: Update dependencies

ltrouton:
Related Ticket: 

Description of Changes:
*  Update dompurify 

Pre-checks:
* [X] Changelog entry added

Review:
* [X] Milestone set


modules/tinymce/CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+- Upgraded dependencies. #TINY-11331
+- Inavlid HTML elements within an `svg` element are removed. #TINY-11331
```
TheSpyder, line 11:
```suggestion
- Invalid HTML elements within an `svg` element are removed. #TINY-11331
```

TheSpyder (approved):

shanmen-tiny (approved):
nitpick: Comparing the patch for DOMPurify here and in #9895, I noticed two continue statements added. I'm assuming they should both behave the same. Should we keep them consistent? 👀

ltrouton:
> nitpick: Comparing the patch for DOMPurify here and in #9895, I noticed two continue statements added. I'm assuming they should both behave the same. Should we keep them consistent?

I removed the `/* Remove attribute */` in the is PR which I don't think was removed in https://github.com/tinymce/tinymce/pull/9895 so that probably explains the slight difference. The `+` and `-` on the far left of the patch file should all be exactly the same between the two

jscasca (approved):

PR #9916, opened 2024-10-09
TINY-11338: add uploadcare loader styles to oxide

ztomaszyk:
Related Ticket: TINY-11338

Description of Changes:
Loader and overlay styles.

Pre-checks:

Review:
* [x] Milestone set

### DIFF
--- a/modules/oxide/src/less/theme/content/uploadcare/uploadcare.less
+++ b/modules/oxide/src/less/theme/content/uploadcare/uploadcare.less
@@ -1,7 +1,7 @@
-@overlay-background-color: fadeout(@background-color);
+@overlay-background-color: rgba(255, 255, 255, 0.5);;
 
 @loading-element-border: 3px solid @color-tint;
-@loading-element-width: min(30%, 24px);
+@loading-element-width: ~"min(24px, 30%)";
 
 .tox-uc-image-overlay {
   position: absolute;
--- a/modules/oxide/src/less/theme/content/uploadcare/uploadcare.less
+++ b/modules/oxide/src/less/theme/content/uploadcare/uploadcare.less
@@ -19,15 +19,5 @@
   border: @loading-element-border;
   border-bottom-color: transparent;
 
-  animation: rotation 1s linear infinite;
-
-  @keyframes rotation {
-    0% {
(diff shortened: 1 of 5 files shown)


modules/oxide/src/less/theme/content/uploadcare/uploadcare.less
```diff
@@ -0,0 +1,33 @@
...
+.tox-uc-loading-element {
+  width: @loading-element-width;
+  aspect-ratio: 1 / 1;
+  border-radius: 50%;
+  border: @loading-element-border;
+  border-bottom-color: transparent;
+
+  animation: rotation 1s linear infinite;
+
+  @keyframes rotation {
```
TheSpyder, line 24: We already have a `rotation`? Nesting them in other declarations doesn't make a difference, they compile to the top level of our CSS file.
https://github.com/tinymce/tinymce/blob/fcec6e97353d4bbabb5af08bdaa043423d77c224/modules/oxide/src/less/theme/components/comments/comment.less#L358-L366

Maybe we should just have a collection of `@keyframes` declarations in an `animation.less` or similar 🤔 and we should probably namespace them to avoid conflicts with the page. That's a bit too much scope creep, so maybe just do it for this specific keyframe declaration and log a ticket for the rest.

TheSpyder (approved):

modules/oxide/src/less/theme/content/uploadcare/uploadcare.less
spocke: Making a animations file makes sense since it's likely we will have spinners in more places but I wonder is the key frames name a page global rule or a css file global if it's page global then we need to prefix them better since a integrator might already have a animation called rotation this should then be named tox-rotation or something.

noxuhax: Yep, `@keyframe` names are page-scoped, so it’s important to have a unique name

spocke: - [ ] Namespace the keyframes name to avoid name clashes and move it to a `animations.less` file

spocke (approved):

PR #9923, opened 2024-10-15
TINY-11335: add placeholder custom element styles and make it resizable

ztomaszyk:
Related Ticket: TINY-11335

Description of Changes:
Add placeholder custom element styles and make it resizable

Pre-checks:
* [ ] Changelog entry added

Review:
* [ ] Milestone set


lorenzo-pomili (approved):

modules/tinymce/src/core/main/ts/fmt/DefaultFormats.ts
```diff
@@ -40,10 +40,17 @@ const get = (editor: Editor): Formats => {
         selector: 'img,audio,video',
         collapsed: false,
         styles: {
-          float: 'left'
+          float: 'left',
         },
         preview: 'font-family font-size'
       },
+      {
+        selector: '.mce-placeholder',
```
spocke, line 48: Should these selectors be the custom element name then we don't need the extra class attribute on it? The class has the benefit of being generic so if we say have two types of placeholders they could share the same class and then get the same alignment behaviour. But that assumes that any new placeholder will also be a inline block type.

modules/tinymce/src/core/main/ts/api/dom/ControlSelection.ts
```diff
@@ -127,6 +127,10 @@ const ControlSelection = (selection: EditorSelection, editor: Editor): ControlSe
       return false;
     }
 
+    if (dom.hasClass(elm, 'mce-placeholder')) {
```
spocke, line 130: I guess we decided to remove this but if we want to have it optional or experimental I wonder if we should use the method you had before where it specifically checked if the `data-mce-resize` was set to true it could do that and then also check if that element is the `tiny-uploadcare-placeholder` element maybe?
It becomes hard coded plugin specific things in the core but we already do that for other things so maybe that ship has already sailed.

modules/tinymce/src/core/main/ts/fmt/DefaultFormats.ts
spocke: I really wish we had some api to register these from the plugin side.

modules/oxide/src/less/theme/content/uploadcare/uploadcare.less
```diff
@@ -0,0 +1,29 @@
+@placeholder-initial-width: 600px;
```
spocke, line 1: I wonder if these should be `uploadcare-` prefixed they become quite long though but in case we add some other placeholder this name would then clash. Or do we split then up once we do that into a generic and a specific version of the placeholder?

spocke (approved):

modules/oxide/src/less/theme/content/uploadcare/uploadcare.less
ztomaszyk: For now these are uploadcare specific and it does make sense to have a prefix so I've decided to add it. If we decide to add more general placeholder with the same styles we can change that

modules/tinymce/src/core/main/ts/fmt/DefaultFormats.ts
ztomaszyk: I like the class/attribute approach because of that generic value. This way we don't have core aware of some plugin specific stuff. It is just a behavior that you can add in any plugin/placeholder element. 
If we add another not inline block placeholder we could have two different classes for alignment behavior like `.mce-placeholder-inline` and `.mce-placeholder-block`. Or even more generic ones for not just for placeholders

modules/tinymce/src/core/main/ts/api/dom/ControlSelection.ts
ztomaszyk: I will remove it for now, but we can revisit it if we decide to go with an optional resize. To be honest, I don't like adding more plugin specific stuff to core

PR #9926, opened 2024-10-16
TINY-6023: Update Clipboard API

ArvinJ-H:
Related Ticket: 
TINY-6023

Description of Changes:
- Updated the Copy, Cut, Paste button to use newest clipboard API
- Added a fallback mode for unsupported browser. 

- [x] Don't merge until 7.5 is out

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/tinymce/src/core/main/ts/api/commands/ClipboardCommands.ts
```diff
@@ -1,37 +1,94 @@
...
+      msg = msg.replace(/Ctrl\+/g, '\u2318+');
+    }
+
+    editor.notificationManager.open({ text: msg, type: 'error' });
+  };
+
+  editor.editorCommands.addCommands({
+    'Cut,Copy,Paste': (command) => {
+      const clipboard = navigator.clipboard;
+      console.log('command', navigator.clipboard);
```
TheSpyder, line 76: Please add a [secure context check](https://developer.mozilla.org/en-US/docs/Web/API/Window/isSecureContext) before poking the async clipboard API to avoid fatal errors in the console.

modules/tinymce/src/core/main/ts/api/commands/ClipboardCommands.ts
```diff
@@ -1,37 +1,94 @@
...
+    if (failed || !doc.queryCommandSupported(command)) {
+      handleClipboardError();
+    }
+  };
+
+  const handleClipboardError = () => {
+
+    let msg = editor.translate(
+      `Your browser doesn't support direct access to the clipboard. ` +
+        'Please use the Ctrl+X/C/V keyboard shortcuts instead.'
```
TheSpyder, line 63: This error is now incorrect, we need two new strings... and I guess we do need to add the permission API here after all.

* if the page is not in a secure context, something about direct clipboard access is not allowed by the page
* if the user denied the permission, say something about clipboard permission not granted

modules/tinymce/src/plugins/codesample/test/ts/webdriver/CodeSampleCopyAndPasteTest.ts
```diff
@@ -48,7 +48,8 @@ describe('webdriver.tinymce.plugins.codesample.CodeSampleCopyAndPasteTest', () =
     hook.editor().setContent('');
   });
 
-  (browser.isSafari() ? it.skip : it)('TINY-8861: press enter after pasting a code sample should not add a newline inside the code', async () => {
+  // Chromium requires clipboard permissions to be granted, so we can't run the test (and there is no bug)
+  (browser.isChromium() || browser.isSafari() ? it.skip : it)('TINY-8861: press enter after pasting a code sample should not add a newline inside the code', async () => {
```
TheSpyder, line 52: Isn't there a webdriver setting we can use to bypass this? Surely other developers who need the permission API have tests...

modules/tinymce/src/core/main/ts/api/commands/ClipboardCommands.ts
```diff
@@ -1,37 +1,94 @@
...
+      handleClipboardError();
+    }
+  };
+
+  const fallbackClipboardCommand = (editor: Editor, command: string) => {
+    const doc = editor.getDoc();
+    let failed = false;
+
+    try {
+      doc.execCommand(command);
```
TheSpyder, line 44: There is no point to this code anymore, the only browser it ever worked on was IE.

TheSpyder (changes requested):

modules/tinymce/src/plugins/codesample/test/ts/webdriver/CodeSampleCopyAndPasteTest.ts
ArvinJ-H: the permission for `clipboard-write` was removed as a valid entry quite a while ago... I'm still trying to return the correct state with `navigator.permissions.query`, it seems it is always returning `true` even when I denied the prompt 
https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/1245

modules/tinymce/src/core/main/ts/api/commands/ClipboardCommands.ts
ArvinJ-H: the permission for clipboard-write was removed as a valid entry quite a while ago... I'm still trying to return the correct state with navigator.permissions.query, it seems it is always returning true even when I denied the prompt
https://github.com/microsoft/TypeScript-DOM-lib-generator/issues/1245

Also the permission API (`clipboard-write') is only supported by Chrome
https://developer.mozilla.org/en-US/docs/Web/API/Permissions_API

modules/tinymce/src/core/main/ts/api/commands/ClipboardCommands.ts
```diff
@@ -1,38 +1,102 @@
...
+  const browser = PlatformDetection.detect().browser;
+
+  const copyToClipboard = async (text: string) => {
+    await navigator.clipboard.writeText(text);
+  };
+
+  const pasteFromClipboard = async (): Promise<void> => {
+    let text = '';
+    text = await navigator.clipboard.readText();
+    editor.selection.setContent(text);
```
TheSpyder, line 16: This can be a `const` now

modules/tinymce/src/core/main/ts/api/commands/ClipboardCommands.ts
```diff
@@ -1,38 +1,102 @@
...
+  const handleSecureContentError = () => {
+
+    let msg = editor.translate(
+      'Error: This operation requires a secure context (HTTPS)' +
+        'Direct access to the clipboard is not allowed in insecure contexts.'
+    );
+
+    if (Env.os.isMacOS() || Env.os.isiOS()) {
+      msg = msg.replace(/Ctrl\+/g, '\u2318+');
+    }
```
TheSpyder, line 33: This code no longer does anything, and `msg` can be a const

- [ ] remove

modules/tinymce/src/core/main/ts/api/commands/ClipboardCommands.ts
```diff
@@ -1,38 +1,102 @@
...
+  const cutToClipboard = async (text: string) => {
+    await navigator.clipboard.writeText(text);
+    editor.selection.setContent('');
+  };
+
+  const handleSecureContentError = () => {
+
+    let msg = editor.translate(
+      'Error: This operation requires a secure context (HTTPS)' +
+        'Direct access to the clipboard is not allowed in insecure contexts.'
```
TheSpyder, line 28: It's not on two lines in the rendering, so punctuation is neccessary
```suggestion
      'Error: This operation requires a secure context (HTTPS).' +
        'Direct access to the clipboard is not allowed in insecure contexts.'
```

modules/tinymce/src/core/main/ts/api/commands/ClipboardCommands.ts
```diff
@@ -1,38 +1,102 @@
...
+
 import Editor from '../Editor';
 import Env from '../Env';
 
 export const registerCommands = (editor: Editor): void => {
-  editor.editorCommands.addCommands({
-    'Cut,Copy,Paste': (command) => {
-      const doc = editor.getDoc();
-      let failed;
+  const browser = PlatformDetection.detect().browser;
```
TheSpyder, line 7: `PlatformDetection.detect()` is cached, you can just call it every time rather than cache the browser value. Or just use `Env.browser`, since that's already imported.

modules/tinymce/src/core/main/ts/api/commands/ClipboardCommands.ts
```diff
@@ -1,38 +1,102 @@
...
+
+  const handlePermissionError = () => {
+
+    let msg = editor.translate(
+      'Clipboard permission denied. Please allow clipboard access to use this feature.'
+    );
+
+    if (Env.os.isMacOS() || Env.os.isiOS()) {
+      msg = msg.replace(/Ctrl\+/g, '\u2318+');
+    }
```
TheSpyder, line 67: This code no longer does anything, and `msg` can be a const

- [ ] remove

modules/tinymce/src/core/main/ts/api/commands/ClipboardCommands.ts
```diff
@@ -1,38 +1,102 @@
...
+    editor.notificationManager.open({ text: msg, type: 'error' });
+
+    return false;
+  };
 
-      // Try executing the native command
+  const checkPermission = async () => {
+    const permission = browser.isChromium() ? await navigator.permissions.query({ name: 'clipboard-read' as PermissionName }) : { state: 'granted' };
+    if (permission.state === 'granted') {
+      return true;
```
TheSpyder, line 43: Now that you've rewritten this file, these checks are already in the only code that calls this function. Maybe the `denied` state should move there too?

- [ ] de-duplicate and simplify

modules/tinymce/src/plugins/codesample/test/ts/webdriver/CodeSampleCopyAndPasteTest.ts
TheSpyder: As discussed in Slack, the answer is "yes but bedrock needs to be updated"

PR #9957, opened 2024-10-29
TINY-11383: When trying to change row type, check if the row is CET, not the selected cell(s)

HAFRMO:
Related Ticket: TINY-11383

Description of Changes:
Previously just checked if the cell was CEF, now check the whole row since it is on the row we act.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


.changes/unreleased/tinymce-TINY-11383-2024-10-28.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: Changing table row type from a CEF cell would not work as expected.
```
TheSpyder, line 3: people don't know what CEF is.
```suggestion
body: Changing table row type from a contenteditable=false cell would not work as expected.
```

TheSpyder (approved):

modules/snooker/src/main/ts/ephox/snooker/model/RunOperation.ts
```diff
@@ -169,6 +169,15 @@ const onCells = (warehouse: Warehouse, target: TargetSelection): Optional<Struct
 
 // Custom unlocked extractors
 
+const onUnlockedRows = (warehouse: Warehouse, target: TargetSelection): Optional<Structs.DetailExt[]> =>
```
ltrouton, line 172: This is a nitpick but in Snooker, putting `contenteditable="false"` doesn't mean the cell/column/row is locked. Snooker only has the concept of locked columns (`data-snooker-locked-cols`) but not locked rows at this stage.

If we make it so that a row cannot have its type changed if the `tr` has `contenteditable="false"` then we would also have to consider what should happen when a `col` has `contenteditable="false"` for various operations as well.

For the row type table operations, I would be more inclined to just use `RunOperation.onCell`  and `RunOperation.onCells` and re-evaluate when locked rows are properly implemented some day.

ltrouton (commented):
I would add a test case using the row dialog as well to set the row type as that is what the Jira ticket was mainly about and is more of e2e test

modules/snooker/src/main/ts/ephox/snooker/model/RunOperation.ts
HAFRMO: I had some concerns that this might then allow changes on content which is CEF entirely, rather than where the cell is CEF but the table/row as a whole is CET.
As for when we need to consider td's being CEF, isn't that what the whole isLocked check is for?

ltrouton: `isLocked` is tied to the locked columns. For example, if we have `data-snooker-locked-cols="0"` then all cells in the first column are considered locked but the cells are not necessarily CEF

HAFRMO: Alright, I'll change it to onCell.

modules/tinymce/src/core/test/ts/browser/caret/TableRowNonEditableTest.ts
ltrouton, line 1: Can we add a test to `TableRowDialogTest.ts` as well

ltrouton (approved):

modules/tinymce/src/core/test/ts/browser/caret/TableRowNonEditableTest.ts
HAFRMO: Alright. The issue lay in the command, but a test for the dialog can't hurt.

Issue #5127, opened 2019-08-21
Adding a link to an image with caption makes the checkbox "Show caption" disappear


M-Reimer:
**What is the current behavior? Describe the bug**
If link is added to an image with caption, then the checkbox "Show caption" in the image settings disappears. Re-checking it adds a second caption to the image.

**Please provide the steps to reproduce and if possible a minimal demo of the problem via [fiddle.tinymce.com](http://fiddle.tinymce.com/) or similar.**
I've copied a similar fiddle and edited it:
http://fiddle.tinymce.com/ELfaab/3

Right click on the image, choose "Image". Verify that "Show caption" is checked. Close popup window.

Right click on the image again, choose "Link". Add a link to whatever URL you like.

Right click on the image and choose "Image" again. Now the "Show caption" checkbox is no longer checked.

**What is the expected behavior?**

The checkbox should still be checked.

**Which versions of TinyMCE, and which browser / OS are affected by this issue? Did this work in previous versions of TinyMCE?**

5.0.13

PR #9964, opened 2024-10-31
Hotfix/5127 Improve handling of captioned images where <img> is not direct descendent of <figure>

benryves:
Related Ticket: 
https://github.com/tinymce/tinymce/issues/5127

Description of Changes:
* Images can be given a caption by ticking a box in the image editor which wraps them in a `<figure>`. This works as long as the image is kept as a direct descendant of the `<figure>`, but things go awry if the path is changed e.g. if the image is made a link, so the path is now figure > a > img. The caption checkbox in the image editor is cleared, and ticking it adds a second caption to the image (for example).
* Similarly, adding a caption to an existing image with a link removes the `` from the `<a>`, making the image link no longer work and leaving an empty `<a>` behind.
* This fix works by changing the way image and figure are selected.
  * Rather than directly access the `<figure>` from the `` by accessing `image.parentNode`, a `getFigure(image)` function is used which will access any parent `<figure>` of the `` (so `hasCaption(image)` now works if there's an `<a>` between the two).
  * Rather than directly manipulate the `` when wrapping the element in a `<figure>` a `getInlineContainer(image)` function is used which searches for the topmost non-block-level element containing the `` (e.g. its containing `<a>`) which means that that those elements are retained when the image is moved inside the containing figure (links no longer get lost and left behind as empty elements).

Pre-checks:
* [ ] Changelog entry added

Review:
* [ ] Milestone set

https://github.com/tinymce/tinymce/issues/5127

### DIFF
--- a/modules/tinymce/src/plugins/image/main/ts/core/ImageData.ts
+++ b/modules/tinymce/src/plugins/image/main/ts/core/ImageData.ts
@@ -58,8 +58,11 @@ const getAttrib = (image: HTMLElement, name: string): string => {
   }
 };
 
+const getFigure = (image: HTMLElement): HTMLElement | null =>
+  DOM.getParent(image, 'figure');
+
 const hasCaption = (image: HTMLElement): boolean =>
-  image.parentNode !== null && image.parentNode.nodeName === 'FIGURE';
+  getFigure(image) !== null;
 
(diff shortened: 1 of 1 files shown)


TheSpyder:
internal reference TINY-4070

TheSpyder:
@coderabbitai ignore

PR #9970, opened 2024-11-05
TINY-11459: add placeholder for context form input

ztomaszyk:
Related Ticket: TINY-11459

Description of Changes:
Add placeholder for context form input

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/themes/silver/main/ts/ui/context/ContextFormTextInput.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/ui/context/ContextFormTextInput.ts
@@ -1,6 +1,6 @@
 import { AddEventsBehaviour, AlloyComponent, AlloyEvents, Behaviour, Disabling, FormField, GuiFactory, Input, Keying, NativeEvents, SketchSpec } from '@ephox/alloy';
 import { InlineContent } from '@ephox/bridge';
-import { Optional } from '@ephox/katamari';
+import { Fun, Optional } from '@ephox/katamari';
 
 import { UiFactoryBackstageProviders } from '../../backstage/Backstage';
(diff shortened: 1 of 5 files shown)


lorenzo-pomili (approved):

modules/tinymce/src/themes/silver/main/ts/ui/context/ContextFormTextInput.ts
```diff
@@ -17,9 +17,16 @@ export const renderContextFormTextInput = (
     components: [ GuiFactory.text(label) ]
   }));
 
+  const placeholder = ctx.placeholder.fold( Fun.constant({}), (p) => ({ placeholder: providers.translate(p) }));
```
TheSpyder, line 20: Just for future reference
```suggestion
  const placeholder = ctx.placeholder.map((p) => ({ placeholder: providers.translate(p) })).getOr({});
```

spocke: One could argue that both these approaches are fine but the suggested one doesn't need Fun to be imported also I think we use the suggested pattern more often.

TheSpyder: The original avoids creating an extra wrapping `Optional` instance from the `map` result, but it adds a throwaway function.

We use the `map().getOr()` pattern often because it's more concise, at least IMO. The only reason we avoid it is if calculating the fallback is expensive; that's when we either go with `fold` or perhaps `.map().getOrThunk()`.

PR #9974, opened 2024-11-06
TINY-11399: now is possible to overwrite `quickbars` `QuickbarInsertImage` behavior

lorenzo-pomili:
Related Ticket: TINY-11399

Description of Changes:
now if we want to overwrite the behavior of the `quickbars` insert image button we can just overwrite the command: `QuickbarInsertImage`, the `quickbars` plugin have a check that allow us to define that command also before the load of the plugin

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set
* [x] Docs ticket created (DOC-2572)


modules/tinymce/src/plugins/quickbars/main/ts/api/Commands.ts
```diff
@@ -0,0 +1,25 @@
+import Editor from 'tinymce/core/api/Editor';
+
+import * as Actions from '../insert/Actions';
+import * as Conversions from '../insert/Conversions';
+import * as Picker from '../insert/Picker';
+
+const register = (editor: Editor): void => {
+  if (!editor.queryCommandSupported('PickFile')) {
```
spocke, line 8: I think we need to make the command quickbar specific so maybe `QuickbarInsertImage`? It does more than just pick a file it also inserts the image.

- [ ] Make the command more specific to the quickbar plugin

spocke (approved):

modules/tinymce/src/plugins/quickbars/main/ts/api/Commands.ts
lorenzo-pomili: updated

modules/tinymce/src/plugins/quickbars/test/ts/browser/CommandsTest.ts
```diff
@@ -0,0 +1,36 @@
...
+import { Arr } from '@ephox/katamari';
+import { TinyHooks } from '@ephox/wrap-mcagar';
+import { assert } from 'chai';
+
+import Editor from 'tinymce/core/api/Editor';
+import QuickbarsPlugin from 'tinymce/plugins/quickbars/Plugin';
+
+describe('browser.tinymce.plugins.quickbars.CommandsTest', () => {
+  Arr.each([ true, false ], (withPresetedCommand) => {
+    const store: string[] = [];
```
shanmen-tiny, line 11: FYI there's `TestStore` in agar that you can use 👍

lorenzo-pomili: refactored, thanks :)

.changes/unreleased/tinymce-TINY-11399-2024-11-06.yaml
```diff
@@ -0,0 +1,7 @@
+project: tinymce
+kind: Added
+body: '`QuickbarInsertImage` command which can be used to open the file picker or
+  overwrite to get the new behavior when `quickimage` is clicked.'
```
TheSpyder, line 4:
```suggestion
body: ‘The `quickimage` button now uses a`QuickbarInsertImage` command, allowing it to be registered by other plugins and replace the default action'
```

modules/tinymce/src/plugins/quickbars/main/ts/api/Commands.ts
```diff
@@ -0,0 +1,25 @@
+import Editor from 'tinymce/core/api/Editor';
+
+import * as Actions from '../insert/Actions';
+import * as Conversions from '../insert/Conversions';
+import * as Picker from '../insert/Picker';
+
+const register = (editor: Editor): void => {
+  if (!editor.queryCommandSupported('QuickbarInsertImage')) {
+    editor.addCommand('QuickbarInsertImage', () => {
```
TheSpyder, line 9: I suggest delaying registration of this command to `preinit` - otherwise we have another race condition based on plugin load order.

- [ ] delay to right before the editor finishes loading

TheSpyder (approved):

tiny-ben-tran (approved):

modules/tinymce/src/plugins/quickbars/main/ts/api/Commands.ts
tiny-ben-tran: It might not help if another plugin wants to override the command in `preinit`, right?

TheSpyder: There will always be _some_ sort of race condition, but our examples all register event handlers and commands when the plugin loads - which is long before `preinit`.

.changes/unreleased/tinymce-TINY-11399-2024-11-06.yaml
lorenzo-pomili: changed

modules/tinymce/src/plugins/quickbars/main/ts/api/Commands.ts
lorenzo-pomili: wait, I added [a setup implementation to the test case](https://github.com/tinymce/tinymce/pull/9974/files#diff-693762719f6da463204bea6c874fd4becd9ebe19fabea4321128807de33cd2b8R18-R20) and I did it in this way to be sure that if another plugin implement `QuickbarInsertImage` the new implementation always wins.

at the moment the behavior is:
`newImplementation` -then> `quickbars` loads => the `if` check is false
`quickbars` loads -then> `newImplementation` => new implementation will overwrite the `quickbars`' one

the only way to have a race condition that I see is if that `addCommand` is async, Am I missing something?

TheSpyder: Oh you did `preinit` in the new plugin implementation to avoid the race condition. I'd reverse that and have the `preinit` here instead.

.changes/unreleased/tinymce-TINY-11399-2024-11-06.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Added
```
spocke, line 2:
```suggestion
kind: Changed
```

.changes/unreleased/tinymce-TINY-11399-2024-11-06.yaml
```diff
@@ -0,0 +1,6 @@
...
+body: 'The `quickimage` button now uses a`QuickbarInsertImage` command, allowing it to be registered by other plugins and replace the default action'
```
spocke, line 3: Not sure we need to state why we added it, it's mostly an internal concern?

```suggestion
body: 'The `quickimage` button now uses a `QuickbarInsertImage` command.'
```

modules/tinymce/src/plugins/quickbars/test/ts/browser/CommandsTest.ts
```diff
@@ -0,0 +1,36 @@
...
+      menubar: false,
+      setup: (ed: Editor) => {
+        if (withPresetedCommand) {
+          ed.addCommand('QuickbarInsertImage', () => store.add('QII'));
+        }
+      },
+      base_url: '/project/tinymce/js/tinymce'
+    }, [ QuickbarsPlugin ], true);
+
+    it(`TINY-11399: QuickbarInsertImage should be not overwrite by the plugin if it already exists${withPresetedCommand ? ' (withPresetedCommand)' : ''}`, async () => {
```
spocke, line 25:
```suggestion
    it(`TINY-11399: QuickbarInsertImage should be not overwritten by the plugin if it already exists ${withPresetedCommand ? ' (withPresetedCommand)' : ''}`, async () => {
```

modules/tinymce/src/plugins/quickbars/test/ts/browser/CommandsTest.ts
```diff
@@ -0,0 +1,36 @@
+import { TestStore } from '@ephox/agar';
+import { describe, it } from '@ephox/bedrock-client';
+import { Arr } from '@ephox/katamari';
+import { TinyHooks } from '@ephox/wrap-mcagar';
+
+import Editor from 'tinymce/core/api/Editor';
+import QuickbarsPlugin from 'tinymce/plugins/quickbars/Plugin';
+
+describe('browser.tinymce.plugins.quickbars.CommandsTest', () => {
+  Arr.each([ true, false ], (withPresetedCommand) => {
```
spocke, line 10: Maybe pre-defined is better?

```suggestion
  Arr.each([ true, false ], (withPredefinedCommand) => {
```

modules/tinymce/src/plugins/quickbars/test/ts/browser/CommandsTest.ts
```diff
@@ -0,0 +1,36 @@
...
+        }
+      },
+      base_url: '/project/tinymce/js/tinymce'
+    }, [ QuickbarsPlugin ], true);
+
+    it(`TINY-11399: QuickbarInsertImage should be not overwrite by the plugin if it already exists${withPresetedCommand ? ' (withPresetedCommand)' : ''}`, async () => {
+      const editor = hook.editor();
+      store.assertEq('it should be empty at start', []);
+      editor.execCommand('QuickbarInsertImage');
+      store.assertEq('if overwritte it should exec the overwritten command', withPresetedCommand ? [ 'QII' ] : []);
```
spocke, line 29:
```suggestion
      store.assertEq('should use predefined command if it exists', withPresetedCommand ? [ 'QII' ] : []);
```

spocke (approved):

modules/tinymce/src/plugins/quickbars/test/ts/browser/CommandsTest.ts
spocke, line 25: I think if you use the test function pattern here instead then the labels would be less confusing since then you could do. It makes it easier to just define two separate strings. With a loop and a state boolean then you need to use a ternary to me at least that is harder to read.

```
it('Should execute predefined command', () => testQuickInsertImage(true));
it('Should execute overwritten command', () => testQuickInsertImage(false));
```

spocke: It also adds more nesting if you have context in content and a loop it becomes pretty deep after a while not the case here but in general.

modules/tinymce/src/plugins/quickbars/main/ts/api/Commands.ts
lorenzo-pomili: moved inside the `PreInit`

.changes/unreleased/tinymce-TINY-11399-2024-11-06.yaml
lorenzo-pomili: done

modules/tinymce/src/plugins/quickbars/test/ts/browser/CommandsTest.ts
lorenzo-pomili: done

modules/tinymce/src/plugins/quickbars/test/ts/browser/CommandsTest.ts
lorenzo-pomili: done

modules/tinymce/src/plugins/quickbars/test/ts/browser/CommandsTest.ts
lorenzo-pomili: done

modules/tinymce/src/plugins/quickbars/test/ts/browser/CommandsTest.ts
lorenzo-pomili: the function would need to have the definition of the hook inside it, is there a way to define it and change the setup when you call: `const editor = hook.editor();`?

PR #9985, opened 2024-11-11
Fix placeholder not disappearing when typing special chars

### LINKED ISSUE #9873: Placeholder not disappearing when typing special chars.
### Problem

The placeholder text does not disappear when typing certain special characters on an Italian keyboard layout (e.g., `@`, `#`, `¶`, `[]`). To type these character you need to press the `Alt` (or `Option` on macOS) key and then the actual keyboard key.

### Reproduction Steps

1. Configure a TinyMCE editor instance with a placeholder text.
2. Use an Italian keyboard layout.
3. Begin typing any of the following special characters: `@`, `#`, `¶`, `[]`.
   
### Expected Behavior

The placeholder should disappear as soon as the user types any character, including special characters.

### Observed Behavior

When typing the specified special characters, the placeholder remains visible, overlapping with the input text.

### Potential Cause

It seems the issue might be related to the check performed in the following code snippet: [Placeholder.ts, Line 78](https://github.com/tinymce/tinymce/blob/main/modules/tinymce/src/core/main/ts/content/Placeholder.ts#L78).

### Environment

- TinyMCE version: any
- Browser: any
- Operating System: any

### Additional Notes

The issue can be reproduced on any instance of TinyMCE with a configured placeholder text.

MythicalMAxX:
Fixes #9873

Fix the placeholder not disappearing when typing special characters on an Italian keyboard layout.

* Update `isNonTypingKeyboardEvent` function in `modules/tinymce/src/core/main/ts/content/Placeholder.ts` to treat events with `Alt` key pressed as typing events.
* Remove the check for non-typing events in the `updatePlaceholder` function in `modules/tinymce/src/core/main/ts/content/Placeholder.ts` to always update the placeholder.
* Add a test case in `modules/tinymce/src/core/test/ts/webdriver/content/PlaceholderTest.ts` to check that the placeholder disappears when typing special characters with the `Alt` key.

---

For more details, open the [Copilot Workspace session](https://copilot-workspace.githubnext.com/tinymce/tinymce/pull/9985?shareId=9217b320-51a4-48b5-9017-dab476ff8aed).

### DIFF
--- a/modules/tinymce/src/core/main/ts/content/Placeholder.ts
+++ b/modules/tinymce/src/core/main/ts/content/Placeholder.ts
@@ -17,7 +17,7 @@ const nonTypingKeycodes = [
   // page up/down, insert
   33, 34, 45,
   // alt, shift, ctrl
-  16, 17, 18,
+  16, 17,
   // meta/windows key
   91, 92, 93,
   // direction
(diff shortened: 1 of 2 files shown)


TheSpyder:
Did you just accept the AI recommendations without checking them? I'm not impressed.

You've updated the non-typing function to allow alt, which seems plausible, but then you removed the only place that function is called.

PR #9988, opened 2024-11-12
TINY-11488: Added new option to disable the editor

shanmen-tiny:
Related Ticket: TINY-11488 & TINY-11490

Description of Changes:
* Reverted previous `readonly` mode to `disabled` option
* Added demo (feel free to try and compare)
* Added support for dynamically changing the option in runtime, `firing` a `DisabledStateChangeEvent` when the value of setting the option and `editor.disabled` is different
* Disabled the ability to switch mode, switch to non-editable root and enable UIs when the editor is disabled
* Restored some tests from the previous readonly mode to disabled mode.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/.changes/unreleased/tinymce-TINY-11488-2024-11-11.yaml
+++ b/.changes/unreleased/tinymce-TINY-11488-2024-11-11.yaml
@@ -1,0 +1,7 @@
+project: tinymce
+kind: Added
+body: New `disabled` option to restore the previous `readonly` mode behavior, allowing
+  the editor to be displayed in a disabled state.
+time: 2024-11-11T22:35:34.883095+08:00
+custom:
+  Issue: TINY-11488
--- a/modules/tinymce/src/core/demo/ts/demo/DisabledDemo.ts
+++ b/modules/tinymce/src/core/demo/ts/demo/DisabledDemo.ts
@@ -1,0 +1,42 @@
+import { DomEvent, Html, Insert, SelectorFind, SugarBody, SugarElement, TextContent } from '@ephox/sugar';
+
+import { TinyMCE } from 'tinymce/core/api/PublicApi';
+
+declare let tinymce: TinyMCE;
+
+export default (): void => {
+  const container = SelectorFind.descendant(SugarBody.body(), '#ephox-ui').getOrDie();
+
+  const statusDiv = SugarElement.fromTag('p');
+  Insert.before(container, statusDiv);
+
+  const toggleModeBtn = SugarElement.fromTag('button');
+  Html.set(toggleModeBtn, 'Toggle readonly mode');
+  DomEvent.bind(toggleModeBtn, 'click', () => {
+    tinymce.activeEditor?.mode.set(tinymce.activeEditor.mode.get() === 'readonly' ? 'design' : 'readonly');
+    TextContent.set(statusDiv, `Mode: ${tinymce.activeEditor?.mode.get()}, Disabled: ${tinymce.activeEditor?.options.get('disabled')}`);
+  });
+  Insert.before(container, toggleModeBtn);
+
+  const toggleDisabledBtn = SugarElement.fromTag('button');
+  Html.set(toggleDisabledBtn, 'Toggle disabled mode');
+  DomEvent.bind(toggleDisabledBtn, 'click', () => {
+    tinymce.activeEditor?.options.set('disabled', !tinymce.activeEditor.options.get('disabled'));
+    TextContent.set(statusDiv, `Mode: ${tinymce.activeEditor?.mode.get()}, Disabled: ${tinymce.activeEditor?.options.get('disabled')}`);
+  });
+  Insert.before(container, toggleDisabledBtn);
+
+  tinymce.init({
+    skin_url: '../../../../js/tinymce/skins/ui/oxide',
+    selector: 'div.tinymce',
+    height: 1000,
+    plugins: 'accordion image table emoticons charmap codesample insertdatetime',
+    // eslint-disable-next-line max-len
+    toolbar: 'bold italic underline strikethrough subscript superscript accordion addtemplate inserttemplate| fontfamily fontsize fontsizeinput | numlist bullist checklist mergetags footnotes footnotesupdate| typography permanentpen formatpainter removeformat forecolor backcolor | blockquote nonbreaking hr pagebreak | casechange styles blocks lineheight | ltr rtl outdent indent | align alignleft aligncenter alignright alignjustify alignnone | h1 h2 h3 h4 h5 h6 h7 | copy cut paste pastetext selectall remove newdocument wordcount searchreplace | undo redo | save cancel restoredraft | fullscreen print preview export code help | template insertdatetime codesample emoticons charmap | anchor link unlink image media pageembed insertfile | visualblocks visualchars | table advtablerownumbering tableclass tablecellclass tablecellvalign tablecellborderwidth tablecellborderstyle tablecaption tablecellbackgroundcolor tablecellbordercolor tablerowheader tablecolheader',
+    setup: (ed) => {
+      ed.on('init', () => {
+        TextContent.set(statusDiv, `Mode: ${ed.mode.get()}, Disabled: ${ed.options.get('disabled')}`);
+      });
+    }
+  });
+};
--- a/modules/tinymce/src/core/main/ts/api/EditorObservable.ts
+++ b/modules/tinymce/src/core/main/ts/api/EditorObservable.ts
@@ -1,6 +1,6 @@
 import { Obj } from '@ephox/katamari';
 
-import { isReadOnly, processReadonlyEvents } from '../mode/Readonly';
+import * as Disabled from '../mode/Disabled';
 import DOMUtils from './dom/DOMUtils';
 import { EventUtilsCallback } from './dom/EventUtils';
 import Editor from './Editor';
@@ -55,12 +55,12 @@ const getEventTarget = (editor: Editor, eventName: string): Node => {
   return editor.getBody();
 };
 
-const isListening = (editor: Editor) => !editor.hidden;
+const isListening = (editor: Editor) => !editor.hidden && !editor.disabled;
 const fireEvent = (editor: Editor, eventName: string, e: Event) => {
   if (isListening(editor)) {
     editor.dispatch(eventName, e);
-  } else if (isReadOnly(editor)) {
-    processReadonlyEvents(editor, e);
+  } else if (Disabled.isDisabled(editor)) {
+    Disabled.processDisabledEvents(editor, e);
   }
 };
 
--- a/modules/tinymce/src/core/main/ts/api/EventTypes.ts
+++ b/modules/tinymce/src/core/main/ts/api/EventTypes.ts
@@ -173,6 +173,10 @@ export interface OpenNotificationEvent {
   notification: NotificationApi;
 }
 
+export interface DisabledStateChangeEvent {
+  state: boolean;
+}
+
 export interface EditorEventMap extends Omit<NativeEventMap, 'blur' | 'focus'> {
   'activate': { relatedTarget: Editor | null };
   'deactivate': { relatedTarget: Editor };
--- a/modules/tinymce/src/core/main/ts/api/Options.ts
+++ b/modules/tinymce/src/core/main/ts/api/Options.ts
@@ -5,6 +5,7 @@ import * as Pattern from '../textpatterns/core/Pattern';
 import * as PatternTypes from '../textpatterns/core/PatternTypes';
 import DOMUtils from './dom/DOMUtils';
 import Editor from './Editor';
+import { fireDisabledStateChange } from './Events';
 import { EditorOptions } from './OptionTypes';
 import I18n from './util/I18n';
 import Tools from './util/Tools';
@@ -434,6 +435,20 @@ const register = (editor: Editor): void => {
 
   registerOption('disable_nodechange', {
     processor: 'boolean',
+    default: false
+  });
+
+  registerOption('disabled', {
+    processor: (value) => {
+      if (Type.isBoolean(value)) {
+        if (editor.disabled !== value) {
+          fireDisabledStateChange(editor, value);
+        }
+        return { valid: true, value };
+      }
+
+      return { valid: false, message: 'The value must be a boolean.' };
+    },
     default: false
   });
 
@@ -988,6 +1003,7 @@ const getSandboxIframesExclusions = (editor: Editor): string[] => editor.options
 const shouldConvertUnsafeEmbeds = option('convert_unsafe_embeds');
 const getLicenseKey = option('license_key');
 const getApiKey = option('api_key');
+const isDisabled = option('disabled');
 
 export {
   register,
@@ -1097,5 +1113,6 @@ export {
   getLicenseKey,
   getSandboxIframesExclusions,
   shouldConvertUnsafeEmbeds,
-  getApiKey
+  getApiKey,
+  isDisabled
 };
--- a/modules/tinymce/src/core/main/ts/mode/Disabled.ts
+++ b/modules/tinymce/src/core/main/ts/mode/Disabled.ts
@@ -1,0 +1,144 @@
+import { Arr, Optional, Strings } from '@ephox/katamari';
+import { Attribute, Compare, SelectorFilter, SelectorFind, SugarElement } from '@ephox/sugar';
+
+import Editor from '../api/Editor';
+import VK from '../api/util/VK';
+import * as ModeUtils from '../util/ModeUtils';
+
+const isDisabled = (editor: Editor): boolean => editor.disabled;
+
+const internalContentEditableAttr = 'data-mce-contenteditable';
+
+const setContentEditable = (elm: SugarElement<HTMLElement>, state: boolean) => {
+  elm.dom.contentEditable = state ? 'true' : 'false';
+};
+
+const switchOffContentEditableTrue = (elm: SugarElement<Node>) => {
+  Arr.each(SelectorFilter.descendants<HTMLElement>(elm, '*[contenteditable="true"]'), (elm) => {
+    Attribute.set(elm, internalContentEditableAttr, 'true');
+    setContentEditable(elm, false);
+  });
+};
+
+const switchOnContentEditableTrue = (elm: SugarElement<Node>) => {
+  Arr.each(SelectorFilter.descendants<HTMLElement>(elm, `*[${internalContentEditableAttr}="true"]`), (elm) => {
+    Attribute.remove(elm, internalContentEditableAttr);
+    setContentEditable(elm, true);
+  });
+};
+
+const toggleDisabled = (editor: Editor, state: boolean): void => {
+  const body = SugarElement.fromDom(editor.getBody());
+
+  if (state) {
+    editor.disabled = true;
+    ModeUtils.disableEditor(editor);
+    setContentEditable(body, false);
+    switchOffContentEditableTrue(body);
+  } else {
+    editor.disabled = false;
+    switchOnContentEditableTrue(body);
+    ModeUtils.enableEditor(editor);
+  }
+};
+
+const registerDisabledContentFilters = (editor: Editor): void => {
+  if (editor.serializer) {
+    registerFilters(editor);
+  } else {
+    editor.on('PreInit', () => {
+      registerFilters(editor);
+    });
+  }
+};
+
+const registerFilters = (editor: Editor): void => {
+  editor.parser.addAttributeFilter('contenteditable', (nodes) => {
+    if (isDisabled(editor)) {
+      Arr.each(nodes, (node) => {
+        node.attr(internalContentEditableAttr, node.attr('contenteditable'));
+        node.attr('contenteditable', 'false');
+      });
+    }
+  });
+
+  editor.serializer.addAttributeFilter(internalContentEditableAttr, (nodes) => {
+    if (isDisabled(editor)) {
+      Arr.each(nodes, (node) => {
+        node.attr('contenteditable', node.attr(internalContentEditableAttr));
+      });
+    }
+  });
+
+  editor.serializer.addTempAttr(internalContentEditableAttr);
+};
+
+const isClickEvent = (e: Event): e is MouseEvent => e.type === 'click';
+
+const allowedEvents: ReadonlyArray<string> = [ 'copy' ];
+
+const isDisabledAllowedEvent = (e: Event) => Arr.contains(allowedEvents, e.type);
+
+const getAnchorHrefOpt = (editor: Editor, elm: SugarElement<Node>): Optional<string> => {
+  const isRoot = (elm: SugarElement<Node>) => Compare.eq(elm, SugarElement.fromDom(editor.getBody()));
+  return SelectorFind.closest<HTMLAnchorElement>(elm, 'a', isRoot).bind((a) => Attribute.getOpt(a, 'href'));
+};
+
+const processDisabledEvents = (editor: Editor, e: Event): void => {
+  /*
+    If an event is a click event on or within an anchor, and the CMD/CTRL key is
+    not held, then we want to prevent default behaviour and either:
+      a) scroll to the relevant bookmark
+      b) open the link using default browser behaviour
+  */
+  if (isClickEvent(e) && !VK.metaKeyPressed(e)) {
+    const elm = SugarElement.fromDom(e.target as Node);
+    getAnchorHrefOpt(editor, elm).each((href) => {
+      e.preventDefault();
+      if (/^#/.test(href)) {
+        const targetEl = editor.dom.select(`${href},[name="${Strings.removeLeading(href, '#')}"]`);
+        if (targetEl.length) {
+          editor.selection.scrollIntoView(targetEl[0], true);
+        }
+      } else {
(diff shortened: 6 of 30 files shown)


MitchC1999 (approved):

modules/tinymce/src/core/main/ts/mode/Disabled.ts
```diff
@@ -0,0 +1,144 @@
...
+import Editor from '../api/Editor';
+import VK from '../api/util/VK';
+import * as ModeUtils from '../util/ModeUtils';
+
+const isDisabled = (editor: Editor): boolean => editor.disabled;
+
+const internalContentEditableAttr = 'data-mce-contenteditable';
+
+const setContentEditable = (elm: SugarElement<HTMLElement>, state: boolean) => {
+  elm.dom.contentEditable = state ? 'true' : 'false';
```
kimwoodfield, line 13: Do we need that condition? We know that `state` is always going to be a boolean so we could just use `toString`:

`
const setContentEditable = (elm: SugarElement<HTMLElement>, state: boolean) => {
  elm.dom.contentEditable = state.toString();
`

modules/tinymce/src/core/main/ts/mode/Disabled.ts
```diff
@@ -0,0 +1,144 @@
...
+  Arr.each(SelectorFilter.descendants<HTMLElement>(elm, `*[${internalContentEditableAttr}="true"]`), (elm) => {
+    Attribute.remove(elm, internalContentEditableAttr);
+    setContentEditable(elm, true);
+  });
+};
+
+const toggleDisabled = (editor: Editor, state: boolean): void => {
+  const body = SugarElement.fromDom(editor.getBody());
+
+  if (state) {
```
kimwoodfield, line 33: I know this will work but is it against convention to rely on truthiness? I thought we had to be explicit (e.g. `state === true`)? 

Just a nitpick in any case 😅

spocke: I don't think you need to do that for booleans since you are not hiding weird behaviour it's either true or false a truthy check on other types is more vague what counts as true/false and then I think it's better to use type checks or similar o be very specific on what you are checking for.

modules/tinymce/src/core/main/ts/mode/Disabled.ts
```diff
@@ -0,0 +1,144 @@
...
+      e.preventDefault();
+    }
+  });
+
+  // Preprend to the handlers as this should be the first to fire
+  editor.on('DisabledStateChange', (e) => {
+    if (!e.isDefaultPrevented()) {
+      toggleDisabled(editor, e.state);
+    }
+  }, true);
```
shanmen-tiny, line 130: This seems a bit fragile. Should it be moved to `Options`, or should it be checked by `e.state` instead of `editor.disabled` in the rest of the places?

shanmen-tiny: Previously, the `editor.readonly` state is updated before the event is fired, so we can use `editor.readonly` in the event listener for the rest of the places. 
https://github.com/tinymce/tinymce/blob/908f8c9dcb409b0ff44fcc6cb5cda52cf52b86f0/modules/tinymce/src/core/main/ts/mode/Mode.ts#L24
https://github.com/tinymce/tinymce/blob/908f8c9dcb409b0ff44fcc6cb5cda52cf52b86f0/modules/tinymce/src/core/main/ts/mode/Readonly.ts#L57

modules/tinymce/src/core/main/ts/mode/Disabled.ts
kimwoodfield: @spocke Thanks for clarifying!

.changes/unreleased/tinymce-TINY-11488-2024-11-11.yaml
```diff
@@ -0,0 +1,7 @@
+project: tinymce
+kind: Added
+body: New `disabled` option to restore the previous `readonly` mode behavior, allowing
```
spocke, line 3: `disabled` was one of those words that was considered offensive so everything was swapped to `enabled` for all apis etc. So I think we need to flip this is we really still care about that see TINY-8101 for details.

modules/tinymce/src/core/main/ts/mode/Disabled.ts
```diff
@@ -0,0 +1,144 @@
...
+  });
+
+  editor.serializer.addTempAttr(internalContentEditableAttr);
+};
+
+const isClickEvent = (e: Event): e is MouseEvent => e.type === 'click';
+
+const allowedEvents: ReadonlyArray<string> = [ 'copy' ];
+
+const isDisabledAllowedEvent = (e: Event) => Arr.contains(allowedEvents, e.type);
```
spocke, line 80:
```suggestion
const isAllowedEventInDisabledMode = (e: Event) => Arr.contains(allowedEvents, e.type);
```

modules/tinymce/src/core/main/ts/util/ModeUtils.ts
```diff
@@ -0,0 +1,72 @@
...
+  }
+};
+
+const setCommonEditorCommands = (editor: Editor, state: boolean): void => {
+  setEditorCommandState(editor, 'StyleWithCSS', state);
+  setEditorCommandState(editor, 'enableInlineTableEditing', state);
+  setEditorCommandState(editor, 'enableObjectResizing', state);
+};
+
+const setContentEditable = (elm: SugarElement<HTMLElement>, state: boolean) => {
```
spocke, line 29: This function already exists in sugar as `ContentEditable.set`

modules/tinymce/src/core/test/ts/browser/DisabledModeTest.ts
```diff
@@ -0,0 +1,523 @@
...
+      assert.equal(Class.has(elm, 'tox-tbtn--disabled'), expectedState, 'Button should have expected disabled state');
+    };
+
+    const assertHrefOpt = (editor: Editor, selector: string, expectedHref: Optional<string>) => {
+      const elm = SugarElement.fromDom(editor.dom.select(selector)[0]);
+      const hrefOpt = Disabled.getAnchorHrefOpt(editor, elm);
+      Assert.eq('href options match', expectedHref, hrefOpt, tOptional());
+    };
+
+    it('TBA: Switching to disabled mode while having cef selection should remove fake selection', () => {
```
spocke, line 94: You don't need TBA here use the real TINY-<N> ticket number or omit it but I assume this is copied from some other test when we still used that odd pattern.

modules/tinymce/src/core/main/ts/mode/Disabled.ts
```diff
@@ -0,0 +1,144 @@
+import { Arr, Optional, Strings } from '@ephox/katamari';
+import { Attribute, Compare, SelectorFilter, SelectorFind, SugarElement } from '@ephox/sugar';
+
+import Editor from '../api/Editor';
+import VK from '../api/util/VK';
+import * as ModeUtils from '../util/ModeUtils';
+
+const isDisabled = (editor: Editor): boolean => editor.disabled;
```
spocke, line 8: Do we really need to expose that disabled state on the editor instance? Setting that manually will not work anyway since it doesn't dispatch events etc. Can it be hidden by using the option since that seems to be the mutating state being used here.

```suggestion
const isDisabled = (editor: Editor): boolean => Options.isDisabled(editor);
```

.changes/unreleased/tinymce-TINY-11488-2024-11-11.yaml
spocke: Do we really need to make this one so complex couldn't this simply be a setting that controls what the readonly mode does. Like a `readonly_disabled` option that can't be mutated independently it just controls what the readonly mode toggle does if it disables everything or just somethings and leave caret movement etc?

spocke (approved):

.changes/unreleased/tinymce-TINY-11488-2024-11-11.yaml
shanmen-tiny: The requirement was made to make the `disabled` option like the previous readonly mode, which includes updating the option at runtime. I'm wondering if introducing a separate `disabled` option might be clearer here 🤔

kimwoodfield (approved):

modules/tinymce/src/core/main/ts/mode/Disabled.ts
shanmen-tiny: 👍  Changed to use `ContentEditable.set` instead.

.changes/unreleased/tinymce-TINY-11488-2024-11-11.yaml
TheSpyder: I don't think anyone is particularly attached to the `disabled` name. But that would be worth double-checking.

We can't really invert it to `editor.enabled = false` like we did with the other `disabled` APIs. I wonder if it could be something like `editor.lockdown` or `editor.blocked` 🤔

modules/tinymce/src/core/main/ts/mode/Disabled.ts
TheSpyder: I can see why this is here - it's a convenient way for any internal component to get the current state of the value - but I think Spocke is right, we can't confuse integrators by offering a potential second way to "set" the value. We need to rely on `editor.options.get` for internal use.

- [ ] remove editor property

TheSpyder (approved):

modules/tinymce/src/core/main/ts/mode/Disabled.ts
```diff
@@ -0,0 +1,140 @@
...
+  });
+};
+
+const toggleDisabled = (editor: Editor, state: boolean): void => {
+  const body = SugarElement.fromDom(editor.getBody());
+
+  if (state) {
+    editor.disabled = true;
+    ModeUtils.disableEditor(editor);
+    ContentEditable.set(body, false);
```
ltrouton, line 32: Do we have to undo this in the `else` block or that done somewhere else?

.changes/unreleased/tinymce-TINY-11488-2024-11-11.yaml
spocke: Why can't we invert it if it all goes though a isDisabled function intent ally just flip the logic in that and call the option enabled and make it on by default?

spocke: We could also expose the isEnabled on either mode or maybe even editor that way you can't mutate it since it's just a getter predicate.

modules/tinymce/src/core/main/ts/mode/Disabled.ts
shanmen-tiny: I've moved that to the `ModeUtils.enableEditor` since some of the functionalities are shared in the disabled mode and readonly mode.

modules/tinymce/src/core/main/ts/mode/Disabled.ts
shanmen-tiny: @TheSpyder @spocke  Requesting re-review, one problem I’m facing is the `DisabledStateChangeEvent` is fired before the option is changed, so when getting the value of the option with `isDisabled: () => editor.options.get('disabled')` returns false. And that is not returning the correct result when I use `isDisabled` check in things like `ObjectSelected` and in the UI backstage, because updating the option only done after the listeners have processed all the handlers. I've wrapped the dispatching of the event in a `promise.resolve().then` so that it's queued in the next queue. Let me know if it works.

shanmen-tiny: Or maybe a `postProcess` callback that executes after the value differs from the original value? 
https://github.com/tinymce/tinymce/compare/feature/TINY-11488...feature/TINY-114888_spike?expand=1

modules/tinymce/src/core/main/ts/api/EventTypes.ts
```diff
@@ -173,6 +173,10 @@ export interface OpenNotificationEvent {
   notification: NotificationApi;
 }
 
+export interface DisabledStateChangeEvent {
+  state: boolean;
```
kimwoodfield, line 177: Should `state` be `readonly`?

modules/tinymce/src/core/main/ts/api/Options.ts
```diff
@@ -437,6 +438,23 @@ const register = (editor: Editor): void => {
     default: false
   });
 
+  registerOption('disabled', {
```
spocke, line 441: So can we name this `enabled` and default that to `true`? I mean we put a lot of effort getting rid of disabled though out the whole UI and options in previous work and at least I tried to continue that path. It can still be called isDisabled internally I think. At least then we don't have `disabled` in docs and other places publicly.

modules/tinymce/src/core/main/ts/api/EventTypes.ts
spocke: It depends the events objects are mutable we don't Object.seal them for example and we do even mutate them sometimes so setting them as readonly would be a bit of a lie. But I guess it's always a lie in type unsafe language like TS.

spocke (approved):

modules/tinymce/src/core/main/ts/api/Options.ts
shanmen-tiny: Moved this to Jira.

shanmen-tiny: Thanks for raising this point, I've moved the discussion to Jira, I will update the ticket once I get inputs 🙂

modules/tinymce/src/core/demo/ts/demo/DisabledDemo.ts
```diff
@@ -0,0 +1,42 @@
...
+  });
+  Insert.before(container, toggleModeBtn);
+
+  const toggleDisabledBtn = SugarElement.fromTag('button');
+  Html.set(toggleDisabledBtn, 'Toggle disabled mode');
+  DomEvent.bind(toggleDisabledBtn, 'click', () => {
+    tinymce.activeEditor?.options.set('disabled', !tinymce.activeEditor.options.get('disabled'));
+    TextContent.set(statusDiv, `Mode: ${tinymce.activeEditor?.mode.get()}, Disabled: ${tinymce.activeEditor?.options.get('disabled')}`);
+  });
+  Insert.before(container, toggleDisabledBtn);
```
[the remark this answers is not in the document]
shanmen-tiny: No, it should be fired when the value of the option changes.

modules/tinymce/src/themes/silver/test/ts/browser/editor/DisableTest.ts
```diff
@@ -48,34 +36,78 @@ describe('browser.tinymce.themes.silver.editor.DisableTest', () => {
...
-    it('TINY-11211: Should not enable button on NodeChange with context: any when ui is disabled', async () => {
-      const editor = hook.editor();
       editor.ui.setEnabled(false);
-      assertButtonDisabled('t1');
-      editor.nodeChanged();
-      assertButtonDisabled('t1');
-      editor.ui.setEnabled(true);
-      assertButtonEnabled('t1');
+      await pAssertUiDisabled(editor, true);
     });
```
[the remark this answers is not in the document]
shanmen-tiny: This should be covered in browser/DisabledModeTest.ts.

modules/tinymce/src/core/main/ts/mode/Disabled.ts
```diff
@@ -0,0 +1,139 @@
...
+    const elm = SugarElement.fromDom(e.target as Node);
+    getAnchorHrefOpt(editor, elm).each((href) => {
+      e.preventDefault();
+      if (/^#/.test(href)) {
+        const targetEl = editor.dom.select(`${href},[name="${Strings.removeLeading(href, '#')}"]`);
+        if (targetEl.length) {
+          editor.selection.scrollIntoView(targetEl[0], true);
+        }
+      } else {
+        window.open(href, '_blank', 'rel=noopener noreferrer,menubar=yes,toolbar=yes,location=yes,status=yes,resizable=yes,scrollbars=yes');
```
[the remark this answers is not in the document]
shanmen-tiny: `noopener noreferrer` seems to only be available in the `windowFeatures` the third parameter in the `window.open` api. https://developer.mozilla.org/en-US/docs/Web/API/Window/open#Window_features 🤔 
Also I've tested this in jsfiddle: https://jsfiddle.net/fb90dmsn/, with the provided parameter. `Referrer` header is not present and logging `window.opener` returns null.

.changes/unreleased/tinymce-TINY-11488-2024-11-11.yaml
shanmen-tiny: We are going with `disabled` option as proposed, see comments in EPIC-165

modules/tinymce/src/core/main/ts/mode/Disabled.ts
```diff
@@ -0,0 +1,139 @@
...
+};
+
+const registerDisabledModeEventHandlers = (editor: Editor): void => {
+  editor.on('ShowCaret', (e) => {
+    if (isDisabled(editor)) {
+      e.preventDefault();
+    }
+  });
+
+  editor.on('ObjectSelected', (e) => {
```
MitchC1999, line 114:
```suggestion
  editor.on('ShowCaret ObjectSelected', (e) => {
```

MitchC1999 (approved):

modules/tinymce/src/core/main/ts/api/Options.ts
```diff
@@ -437,6 +438,23 @@ const register = (editor: Editor): void => {
...
+            fireDisabledStateChange(editor, value);
+          });
+        }
+        return { valid: true, value };
+      }
+
+      return { valid: false, message: 'The value must be a boolean.' };
+    },
+    default: false
+  });
```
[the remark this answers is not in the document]
shanmen-tiny: Yep, it's possible to use `options.set('disabled', true/false)` at runtime to enable or disable the editor. The UI elements and editor properties will update automatically. The processor is called every time this option is set.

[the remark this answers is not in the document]
MitchC1999: Would this be better exposed as an API instead? We don't generally expect initialization options to be updated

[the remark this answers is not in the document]
shanmen-tiny: Yea, I see your point. In this case, the dynamic `options.set()` gives us simpler way of handling editor state changes without needing separate enable/disable APIs (for future options as well). I'm happy to discuss if you see advantages to the API approach.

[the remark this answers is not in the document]
MitchC1999: I'm happy for us to handle it this way for this release. I don't think we should document that this can be changed dynamically, so if we get requests for it we can discuss the best way to expose it then.

modules/tinymce/src/core/demo/ts/demo/DisabledDemo.ts
```diff
@@ -0,0 +1,42 @@
...
+    selector: 'div.tinymce',
+    height: 1000,
+    plugins: 'accordion image table emoticons charmap codesample insertdatetime',
+    // eslint-disable-next-line max-len
+    toolbar: 'bold italic underline strikethrough subscript superscript accordion addtemplate inserttemplate| fontfamily fontsize fontsizeinput | numlist bullist checklist mergetags footnotes footnotesupdate| typography permanentpen formatpainter removeformat forecolor backcolor | blockquote nonbreaking hr pagebreak | casechange styles blocks lineheight | ltr rtl outdent indent | align alignleft aligncenter alignright alignjustify alignnone | h1 h2 h3 h4 h5 h6 h7 | copy cut paste pastetext selectall remove newdocument wordcount searchreplace | undo redo | save cancel restoredraft | fullscreen print preview export code help | template insertdatetime codesample emoticons charmap | anchor link unlink image media pageembed insertfile | visualblocks visualchars | table advtablerownumbering tableclass tablecellclass tablecellvalign tablecellborderwidth tablecellborderstyle tablecaption tablecellbackgroundcolor tablecellbordercolor tablerowheader tablecolheader',
```
ArvinJ-H, line 35:
```suggestion
    toolbar: 'bold italic underline strikethrough subscript superscript accordion addtemplate inserttemplate| fontfamily fontsize fontsizeinput | numlist bullist checklist mergetags footnotes footnotesupdate| typography permanentpen formatpainter removeformat forecolor backcolor | blockquote nonbreaking hr pagebreak | casechange styles blocks lineheight | ltr rtl outdent indent | align alignleft aligncenter alignright alignjustify alignnone | h1 h2 h3 h4 h5 h6 | copy cut paste pastetext selectall remove newdocument wordcount searchreplace | undo redo | save cancel restoredraft | fullscreen print preview export code help | template insertdatetime codesample emoticons charmap | anchor link unlink image media pageembed insertfile | visualblocks visualchars | table advtablerownumbering tableclass tablecellclass tablecellvalign tablecellborderwidth tablecellborderstyle tablecaption tablecellbackgroundcolor tablecellbordercolor tablerowheader tablecolheader',
```

im not sure h7 exist...

ArvinJ-H (approved):

modules/tinymce/src/core/main/ts/api/Options.ts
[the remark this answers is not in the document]
TheSpyder: Having it be dynamic is what made this PR difficult, a customer confirmed they wanted exactly what `editor.mode.set('readonly')` used to do including runtime toggle. I decided I didn't want a dedicated API for it.

Yes it's not something we have done in the past, but we've had a few requests for this sort of dynamic options behaviour. Features that call `editor.options.get()` every time they are used are already dynamic.

While some of these "locked at load time" options are extremely difficult to make dynamic, it would be nice if we could look at ways to make this easy. Perhaps we'll end up supporting a post-processor in option registration, I'd prefer to avoid a wide-scale "option changed" event.

modules/tinymce/src/core/main/ts/mode/Disabled.ts
TheSpyder: call it `onChange` instead of `postProcess`, because it won't run unless the set option is different

log as a ticket for 7.7 or 8.0

modules/tinymce/src/core/main/ts/api/Options.ts
TheSpyder: Where was the JIRA discussion? I've had a few days to think about it and I'm really tempted to agree. I was on the fence at first, but I like the way `editor.options.set('enabled', false)` reads.

TheSpyder: nevermind, found it

modules/tinymce/src/core/main/ts/api/Options.ts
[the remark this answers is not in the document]
spocke: The mode api is the thing that should be used for flipping between UI states like this. So not sure why this couldn't be done with flipping readonly on/off and having a additional option that just controls what the readonly mode does like readonly_disabled_ui or something that is enabled by default and that the customer can disable that one doesn't need to be dynamic.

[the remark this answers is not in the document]
TheSpyder: This whole thing started that way, along these lines:

```
editor.mode.set({
  readonly_ui: true,
  readonly_content: true
});
```

But it seemed overly complex, both to implement and explain.

modules/tinymce/src/core/main/ts/api/Options.ts
ArvinJ-H: just on the `editor.options.set('enabled', false)` part, in this particular case, if we call it `isDisabled` internally but using `enabled` publicly, this is an inverted logic (for us)?

modules/tinymce/src/core/main/ts/api/Options.ts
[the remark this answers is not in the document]
shanmen-tiny: Right so @spocke's suggestion is to add something on top of the readonly mode, that controls that whether we should disable everything or not. That way we don't need new events nor dynamically switching the option at run time and can reuse the `SwitchMode` as well since we have the modes API for that.

modules/tinymce/src/core/main/ts/api/EditorObservable.ts
```diff
@@ -55,12 +55,12 @@ const getEventTarget = (editor: Editor, eventName: string): Node => {
...
-const isListening = (editor: Editor) => !editor.hidden;
+const isListening = (editor: Editor) => !editor.hidden && EditorState.isEnabled(editor);
 const fireEvent = (editor: Editor, eventName: string, e: Event) => {
   if (isListening(editor)) {
     editor.dispatch(eventName, e);
-  } else if (isReadOnly(editor)) {
-    processReadonlyEvents(editor, e);
+  } else if (EditorState.isDisabled(editor)) {
+    EditorState.processDisabledEvents(editor, e);
   }
```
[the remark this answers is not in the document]
shanmen-tiny: There's no need to process them when it's hidden.

.changes/unreleased/tinymce-TINY-11488-2024-11-11.yaml
TheSpyder: well that's confusing, because after you wrote it the decision changed 😂

TheSpyder (approved):

lostkeys (approved):

PR #9990, opened 2024-11-13
TINY-11515: Update `advlist` plugin to recognise the `latin` list style type as an alias for the `alpha` list style type.

ltrouton:
Related Ticket: TINY-11515

Description of Changes:
* When a list is either `lower-latin` or `upper-latin`, the "Lower Alpha" or "Upper Alpha" buttons will be highlighted in the split button dropdown

Pre-checks:
* [X] Changelog entry added

Review:
* [X] Milestone set

### DIFF
--- a/modules/tinymce/src/plugins/advlist/main/ts/ui/Buttons.ts
+++ b/modules/tinymce/src/plugins/advlist/main/ts/ui/Buttons.ts
@@ -1,4 +1,4 @@
-import { Type } from '@ephox/katamari';
+import { Arr, Obj, Type } from '@ephox/katamari';
 
 import Editor from 'tinymce/core/api/Editor';
 import { NodeChangeEvent } from 'tinymce/core/api/EventTypes';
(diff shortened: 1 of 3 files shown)


modules/tinymce/src/plugins/advlist/main/ts/ui/Buttons.ts
```diff
@@ -36,6 +36,14 @@ const makeSetupHandler = (editor: Editor, nodeName: ListType) => (api: Toolbar.T
 };
 
 const addSplitButton = (editor: Editor, id: string, tooltip: string, cmd: string, nodeName: ListType, styles: string[]): void => {
+  const listStyleTypeAliases: Record<string, string> = {
+    'lower-latin': 'lower-alpha',
+    'upper-latin': 'upper-alpha',
+    'lower-alpha': 'lower-latin',
+    'upper-alpha': 'upper-latin'
```
MitchC1999, line 43:
```suggestion
  const listStyleTypeAliases: Record<string, string[]> = {
    'lower-latin': [ 'lower-alpha' ],
    'upper-latin': [ 'upper-alpha' ],
    'lower-alpha': [ 'lower-latin' ],
    'upper-alpha': [ 'upper-latin' ]
```
Would there be any benefit to making these string arrays? In case there are any more aliases added in the future?

MitchC1999 (approved):

spocke (approved):

modules/tinymce/src/plugins/advlist/main/ts/ui/Buttons.ts
spocke: Maybe deal with that if it ever happens also web standard these days is more strict so it will likely not happen as often as before when browsers just implemented what ever they felt like.

PR #9998, opened 2024-11-16
Move caret behind newly inserted links

Philipp91:
Related Ticket: #6711

Description of Changes:
* This is a rebased version of #6982. Since there was no movement on the original bug for 3 years, I'm trying again to fix it myself. I'll try to figure out the failing tests this time. I'm sending this out to see how CI behaves, as I wasn't able to run the test locally per the CONTRIBUTING.md instructions.
* Add unit test that succeeds _before_ the changes, for documentation purposes.
* Change behavior to move caret behind newly inserted links.
* Update the unit tests accordingly. Add a new test case to illustrate with an extreme case (adding a link on `b` within `abc`, then typing `d`, which should result in `a<a>b</a>dc` -- note how the `d` is _not_ inside the link).

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

GitHub issues (if applicable): #6711

## Summary by CodeRabbit

- **New Features**
	- Enhanced cursor placement after updating and creating links for a smoother typing experience.
	- Added new methods for link management, improving functionality within the editor.

- **Bug Fixes**
	- Fixed cursor positioning issues after link updates and creations.

- **Tests**
	- Introduced new test cases to verify correct cursor positioning after link creation and updates, enhancing test coverage.


TheSpyder:
Apologies for the lack of movement; a ticket was logged at the time but never prioritised.

If you can't get the whole test suite working (and I don't blame you) try `yarn test-one <filename>`.

Looking back at that PR, I'm glad you moved it to only apply to links created by the plugin. However, there are still two things from Lee's comment that are relevant:
* Please pull the entire fix into the plugin instead of adding a new core selection API. The `dom` object you used is also available as `editor.dom` so you should still be able to write the fix in the plugin.
* We need you to sign a CLA. We're currently between automated systems, so it's a manual process - please email legal@tiny.cloud with your github username and request a CLA.

Philipp91:
> try yarn test-one <filename>.

Actually, not even the one test works:

```
No local chromedriver for chrome. Searching system path...
Error creating webdriver Error: Unable to find a suitable driver for chrome
```

> I'm glad you moved it to only apply to links created by the plugin.

Did I? I didn't make conscious changes. A file I had touched went away and I reimplemented it elsewhere, but I don't see `plugin` in the filepaths I'm touching at all, rather they all are under `core` somehow. Could you perhaps explain what's the difference?

TheSpyder:
> Actually, not even the one test works:

Oh I forgot `test-one` uses the webdriver. It's looking for the `chromedriver` binary, installed either [manually](https://developer.chrome.com/docs/chromedriver) or through NPM. If you don't want to install that use the dev server by changing `bedrock-auto` to `bedrock`
```
yarn tsc && yarn bedrock -f <filename>
```

> Did I? I didn't make conscious changes.

Apologies, I misread the file's location. If you move the code to the end of this `createLink` function it will only change the behaviour of the dialog, not the API in general.
https://github.com/tinymce/tinymce/blob/0805c74c8f45c699ea5c5c9dfb683c4b9c4caa1b/modules/tinymce/src/plugins/link/main/ts/core/Actions.ts#L73-L87

modules/tinymce/src/plugins/link/main/ts/core/Actions.ts
```diff
@@ -78,6 +82,10 @@ const createLink = (editor: Editor, selectedElm: Element, text: Optional<string>
     text.fold(
       () => {
         editor.execCommand('mceInsertLink', false, linkAttrs);
+        // Now the newly inserted link is selected. Move the cursor behind the new link, so the user can go on typing.
+        const rng = dom.createRng();
+        rng.setStartAfter(editor.selection.getEnd());
+        editor.selection.setRng(rng);
```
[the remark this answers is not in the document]
Philipp91: I don't think the "more robust implementation" is desirable.

I'm not sure if the concern about `editor.selection.getEnd()` returning empty is justified.

Philipp91:
I have now updated the change to implement it only within the `link` plugin. It now also affects updated links, to make the behavior consistent with the `quickbars` plugin and with Google Docs, all of which move the cursor behind newly inserted _and_ updated links.

While testing this manually, I observed some weird behavior, reported it as #6711.

I'm not sure where to insert unit tests for this. Augment `modules/tinymce/src/plugins/link/test/ts/browser/DialogFlowTest.ts`? Or another file in that directory? Or insert an entirely new test file?

modules/tinymce/src/plugins/link/main/ts/core/Actions.ts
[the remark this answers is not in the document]
TheSpyder: It's a valid concern, there's no inherent guarantee that the `mceInsertLink` event will create the link requested, but it usually would.

However the code you added doesn't actually require a link (despite what the comment suggests). It just moves the cursor to the end of whatever was selected. This will _always_ succeed.

TheSpyder:
Thank you, and apologies for the coderabbit intervention. It's a new tool we're trialling and it should have been excluded from activating on community contributions.

You don't need a specific unit test for this; duplicating an existing "insert link with the dialog" test and adding a selection position check would be fine.

modules/tinymce/src/plugins/link/main/ts/core/Actions.ts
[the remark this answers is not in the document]
EkimChau: @coderabbitai: ignore

Philipp91:
> an existing "insert link with the dialog" test

Oddly enough, I can't identify such a test case. There is `UpdateLinkTest.ts`, but not `InsertLinkTest.ts`. And the other files are a tad more specialized. I've now added it into `SelectedTextLinkTest.ts`, which I believe comes closest to asserting the link insertion behavior, though rather tangentially. Please let me know if you have a better place to put it or if I should add a new `InsertLinkTest.ts`.

Like in the previous attempt of this pull request, I have now pushed two commits, one that documents the current behavior in the form of unit test cases, and another that fixes the behavior and updates the test cases accordingly.

-----

> please email legal@tiny.cloud with your github username and request a CLA.

I haven't heard back yet.

modules/tinymce/src/plugins/link/test/ts/browser/SelectedTextLinkTest.ts
```diff
@@ -40,6 +40,16 @@ describe('browser.tinymce.plugins.link.SelectedTextLinkTest', () => {
     existence(SugarBody.body(), '.tox-label:contains("Text to display")');
   };
 
+  it('TBA: inserting new link should move cursor behind the new link', async () => {
+    const editor = hook.editor();
+    editor.setContent('<p>word</p>');
+    TinySelections.setSelection(editor, [ 0 ], 0, [ 0 ], 1); // <p>[word]</p>
```
TheSpyder, line 46: Actually this selects `[<p>word</p>]` - a more realistic user selection would be
```suggestion
    TinySelections.setSelection(editor, [ 0, 0 ], 0, [ 0, 0 ], 4); // <p>[word]</p>
```

TheSpyder (approved):
I have followed up with our team about the CLA. You should receive an email within 24hrs.

> Oddly enough, I can't identify such a test case. There is `UpdateLinkTest.ts`, but not `InsertLinkTest.ts`. And the other files are a tad more specialized. I've now added it into `SelectedTextLinkTest.ts`, which I believe comes closest to asserting the link insertion behavior, though rather tangentially. Please let me know if you have a better place to put it or if I should add a new `InsertLinkTest.ts`.

This is why I used a wordy description instead of a filename - we have many tests that perform the action I suggested. `SelectedTextLinkTest` is a good choice, `DialogOpenTest` is probably what you were looking for specifically but what you've done is all we need.

 
> Like in the previous attempt of this pull request, I have now pushed two commits, one that documents the current behavior in the form of unit test cases, and another that fixes the behavior and updates the test cases accordingly.

That's all good - we squash merge PRs so the branch commits don't actually matter.

modules/tinymce/src/plugins/link/test/ts/browser/SelectedTextLinkTest.ts
Philipp91: Ah yes, I was wondering about that. Both whether it's correct (and it wasn't) and which way round would be the "realistic" way, i.e. what happens if someone manually selects the word. Thanks for the suggestion, I've updated the PR.

Philipp91:
I've sent back the signed CLA. I haven't done these two things:

* Changelog entry added
* Branch prefixed with `feature/`, `hotfix/` or `spike/`

Should I?

TheSpyder:
Feel free to add the changelog; we use changie (`yarn changie new`). You're welcome to credit yourself (there should be some examples of this in `modules/tinymce/changelog.md`). The prefixes are for our git-flow branches, don't worry about those.

modules/tinymce/src/plugins/link/main/ts/core/Actions.ts
ltrouton, line 88: Just to explicitly make sure the new rng is collapsed
```suggestion
        const end = editor.selection.getEnd();
        const rng = dom.createRng();
        rng.setStartAfter(end);
        rng.setEndAfter(end);

```

modules/tinymce/src/plugins/link/main/ts/core/Actions.ts
```diff
@@ -67,7 +67,11 @@ const updateLink = (editor: Editor, anchorElm: HTMLAnchorElement, text: Optional
   });
 
   editor.dom.setAttribs(anchorElm, linkAttrs);
-  editor.selection.select(anchorElm);
+
+  // Move the cursor behind the updated link, so the user can go on typing.
+  const rng = editor.dom.createRng();
+  rng.setStartAfter(anchorElm);
```
ltrouton, line 73:
```suggestion
  rng.setStartAfter(anchorElm);
  rng.setEndAfter(anchorElm);
```

ltrouton (commented):
Thanks for the contribution.

When I run the link plugin tests locally via `yarn bedrock -d modules/tinymce/src/plugins/link/test/ts/` I am getting a number of existing test case failures. I am getting errors in the following files:
- https://github.com/tinymce/tinymce/blob/main/modules/tinymce/src/plugins/link/test/ts/browser/DefaultLinkTargetTest.ts
- https://github.com/tinymce/tinymce/blob/main/modules/tinymce/src/plugins/link/test/ts/browser/DialogOpenTest.ts
- https://github.com/tinymce/tinymce/blob/main/modules/tinymce/src/plugins/link/test/ts/browser/LinkContextMenuTest.ts

The tests look to be failing as they either try to open the dialog or context menu again after a link has been created. To fix these test cases, we should first assert the selection is after the link then move the selection back into the link before opening a dialog or context menu again. 

If you would like us to help with this at all, let us know.

modules/tinymce/src/plugins/link/main/ts/core/Actions.ts
Philipp91: Did you intentionally delete the `editor.selection.setRng(rng);` line? I've added it back.

Philipp91:
I ran into weird test behavior and sent #10144. I'm rebasing the changes here onto that fix.

modules/tinymce/src/plugins/link/main/ts/core/Actions.ts
ltrouton: That was an accident, thanks for picking that up

ltrouton:
I have pushed a couple of commits to this PR if that is alright. We would like to try and get this into the 7.7 release and the code freeze for that today.

PR #10004, opened 2024-11-20
TINY-11558: Fix bundling for skin UI content CSS

ltrouton:
Related Ticket: TINY-11558

Description of Changes:
* Change Loader.ts to use `skin` option rather than the `skin_url` option for checking if a resource exists for bundling.
* Fix loading UI skin content CSS in iframe mode as it would have never correctly inserted it in the iframe

Pre-checks:
* [X] Changelog entry added

Review:
* [X] Milestone set

## Release Notes

- **New Features**
  - Introduced a demo page for showcasing bundled CSS functionality in TinyMCE.
  - Added a new function to retrieve the skin option from the editor's settings.

- **Bug Fixes**
  - Fixed issues related to CSS bundling for skin UI content.
  - Corrected resource keys used for CSS bundling of JavaScript files.

- **Tests**
  - Added a comprehensive test suite for validating bundled CSS functionality within TinyMCE.
  - Updated tests to reflect changes in resource paths for CSS loading.

- **Chores**
  - Enhanced the handling of skin resources and loading processes in the TinyMCE editor.

### DIFF
--- a/modules/tinymce/src/core/demo/html/bundled_css_demo.html
+++ b/modules/tinymce/src/core/demo/html/bundled_css_demo.html
@@ -1,0 +1,35 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <title>Bundled CSS Demo Page</title>
+  </head>
+
+  <body>
+  <h2>Bundled CSS Demo Page</h2>
+    <div id="ephox-ui">
+      <textarea class="tinymce">
+        <p>Plain paragraph</p>
+        <p><a id="abc"></a></p>
+        <table style="border-collapse: collapse; width: 99.9851%;" border="1"><colgroup><col style="width: 50%;"><col style="width: 50%;"></colgroup>
+        <tbody>
+        <tr>
+        <td>a</td>
+        <td>b</td>
+        </tr>
+        <tr>
+        <td>c</td>
+        <td>d</td>
+        </tr>
+        </tbody>
+        </table>
+        <p><img src="https://www.w3schools.com/w3css/img_lights.jpg" alt="" width="600" height="400"></p>
+      </textarea>
+    </div>
+    <script src="../../../../js/tinymce/tinymce.js"></script>
+    <script src="../../../../js/tinymce/skins/ui/oxide/skin.js"></script>
+    <script src="../../../../js/tinymce/skins/content/default/content.js"></script>
+    <script src="../../../../js/tinymce/skins/ui/oxide/content.js"></script>
+    <script src="../../../../scratch/demos/core/demo.js"></script>
+    <script>demos.BundledCssDemo();</script>
+  </body>
+</html>
--- a/modules/tinymce/src/core/demo/html/bundled_css_demo.html
+++ b/modules/tinymce/src/core/demo/html/bundled_css_demo.html
@@ -29,6 +29,9 @@
     <script src="../../../../js/tinymce/skins/ui/oxide/skin.js"></script>
     <script src="../../../../js/tinymce/skins/content/default/content.js"></script>
     <script src="../../../../js/tinymce/skins/ui/oxide/content.js"></script>
+    <!-- <script src="../../../../js/tinymce/skins/ui/oxide-dark/skin.js"></script>
+    <script src="../../../../js/tinymce/skins/content/dark/content.js"></script>
+    <script src="../../../../js/tinymce/skins/ui/oxide-dark/content.js"></script> -->
     <script src="../../../../scratch/demos/core/demo.js"></script>
     <script>demos.BundledCssDemo();</script>
   </body>
--- a/modules/tinymce/src/core/main/ts/api/dom/StyleSheetLoader.ts
+++ b/modules/tinymce/src/core/main/ts/api/dom/StyleSheetLoader.ts
@@ -166,9 +166,10 @@ const StyleSheetLoader = (documentOrShadowRoot: Document | ShadowRoot, settings:
     // Start loading
     const styleElem = SugarElement.fromTag('style', doc.dom);
     Attribute.setAll(styleElem, {
-      rel: 'stylesheet',
-      type: 'text/css',
-      id: state.id
+      'rel': 'stylesheet',
+      'type': 'text/css',
+      'id': state.id,
+      'data-mce-key': key
     });
 
     styleElem.dom.innerHTML = css;
--- a/modules/tinymce/src/core/test/ts/browser/dom/BundledCssTest.ts
+++ b/modules/tinymce/src/core/test/ts/browser/dom/BundledCssTest.ts
@@ -91,7 +91,7 @@ describe('browser.tinymce.core.dom.BundledCssTest', () => {
               for (const skin of [ 'oxide', 'oxide-dark', 'tinymce-5', 'tinymce-5-dark' ] as UiSkin[]) {
                 context('skin: ' + skin, () => {
                   context('no bundling', () => {
-                    it('TINY-11558: Load CSS styesheets as links', async () => {
+                    it('TINY-11558: Load CSS stylesheets as links', async () => {
                       const [ editor, cleanup ] = await pCreateEditor(
                         shadowdom,
                         {
--- a/modules/tinymce/src/core/test/ts/browser/dom/BundledCssTest.ts
+++ b/modules/tinymce/src/core/test/ts/browser/dom/BundledCssTest.ts
@@ -26,25 +26,39 @@ describe('browser.tinymce.core.dom.BundledCssTest', () => {
   const skinUiContentCss = (skin: UiSkin, inline: boolean, withSuffix: boolean) => skinUiDir(skin) + `content${inline ? '.inline' : ''}${withSuffix ? '.min' : ''}.css`;
   const skinUiContentJs = (skin: UiSkin, inline: boolean) => skinUiDir(skin) + `content${inline ? '.inline' : ''}.js`;
 
-  const styleTagsToKeys = (styleTags: SugarElement<HTMLStyleElement>[]) => {
-    return Arr.map(styleTags, (tag) => tag.dom.dataset.mceKey);
-  };
+  const styleTagToKey = (styleTag: SugarElement<HTMLStyleElement>) =>
+    styleTag.dom.dataset.mceKey || styleTag.dom.id;
+  const styleTagsToKeys = (styleTags: SugarElement<HTMLStyleElement>[]) =>
+    Arr.map(styleTags, styleTagToKey);
 
-  const linkTagsToKeys = (linkTags: SugarElement<HTMLLinkElement>[], basePath: string) => {
-    return Arr.map(linkTags, (tag) => tag.dom.href.replace(basePath, ''));
-  };
+  const linkTagToKey = (linkTag: SugarElement<HTMLLinkElement>, basePath: string) =>
+    linkTag.dom.href.replace(basePath, '');
+  const linkTagsToKeys = (linkTags: SugarElement<HTMLLinkElement>[], basePath: string) =>
+    Arr.map(linkTags, (tag) => linkTagToKey(tag, basePath));
 
-  const getCssTags = (container: SugarElement<HTMLHeadElement | ShadowRoot>) => {
-    const style =
+  const getCssTags = (container: SugarElement<HTMLHeadElement | ShadowRoot>, basePath: string) => {
+    const styleTags =
       SelectorFilter.descendants<HTMLStyleElement>(
         container,
         'style'
       );
-    const link =
+    // An unrelated style tag is included in Firefox so fitler it out
+    const filterKeys = new Set([ 'mceDefaultStyles' ]);
+    const style = Arr.filter(styleTags, (tag) => {
+      const key = styleTagToKey(tag);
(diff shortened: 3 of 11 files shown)


modules/tinymce/src/core/test/ts/browser/dom/BundledCssTest.ts
```diff
@@ -0,0 +1,256 @@
...
+
+  for (const inline of [ false, true ]) {
+    context('inline: ' + inline, () => {
+      for (const shadowdom of [ false, true ]) {
+        context('shadowdom: ' + shadowdom, () => {
+          for (const contentCss of [ 'default', 'dark' ] as const) {
+            context('contentCSS: ' + contentCss, () => {
+              for (const skin of [ 'oxide', 'oxide-dark' ] as const) {
+                context('skin: ' + skin, () => {
+                  const oxideUiSkinMap: Record<string, string> = {
```
kimwoodfield, line 90: Hey Luke, is it important that we define `oxideUiSkinMap` inside of the nested loops like this?
If we define it outside then we can just create the object once and reference it. Unless I'm mistaken, we should have it in the scope and it doesn't look like we are modifying it.

modules/tinymce/src/core/test/ts/browser/dom/BundledCssTest.ts
```diff
@@ -0,0 +1,256 @@
...
+                    'oxide-dark': 'dark',
+                    'oxide': 'default',
+                    'tinymce-5': 'tinymce-5',
+                    'tinymce-5-dark': 'tinymce-5-dark'
+                  };
+
+                  context('no bundling', () => {
+                    it('TINY-11558: Load CSS styesheets as links', async () => {
+                      const [ editor, cleanup ] = await pCreateEditor(
```
kimwoodfield, line 99: What is the convention behind naming specific variables like this with `p`? 
Is it short for something like procedure? I'd be curious to know when or where we should use it!

modules/tinymce/src/themes/silver/main/ts/ui/skin/Loader.ts
```diff
@@ -11,6 +11,24 @@ declare let tinymce: TinyMCE;
...
+    'tinymce-5-dark': 'tinymce-5-dark'
+  };
+
+  const skin = Options.getSkin(editor);
+  if (skin === false) {
+    return Optional.none();
+  } else if (skin === true || Type.isNullable(skin)) {
+    return Obj.get(oxideUiSkinMap, 'oxide');
+  } else {
+    return Obj.get(oxideUiSkinMap, skin).or(Optional.from(skin));
```
kimwoodfield, line 28: What kind of scenario might trigger this `else` statement to happen?

kimwoodfield (approved):

modules/tinymce/src/core/test/ts/browser/dom/BundledCssTest.ts
ltrouton: The `p` prefix indicates that the function returns a promise. We only use that convention in tests and it mainly makes it easier to know when we need to use `await` etc.

modules/tinymce/src/core/test/ts/browser/dom/BundledCssTest.ts
ltrouton: If we were outside a test file, I would be more worried about this. Although, for test files, I think visual readability is more important than absolute importance.
In any case, I can make the change 👍

modules/tinymce/src/core/demo/html/bundled_css_demo.html
```diff
@@ -0,0 +1,35 @@
...
+    <title>Bundled CSS Demo Page</title>
+  </head>
+
+  <body>
+  <h2>Bundled CSS Demo Page</h2>
+    <div id="ephox-ui">
+      <textarea class="tinymce">
+        <p>Plain paragraph</p>
+        <p><a id="abc"></a></p>
+        <table style="border-collapse: collapse; width: 99.9851%;" border="1"><colgroup><col style="width: 50%;"><col style="width: 50%;"></colgroup>
```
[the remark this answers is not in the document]
ltrouton: Depending on the CSS `box-sizing`, the border widths and padding may not be considered in the width which can lead to the width not being perfectly 100%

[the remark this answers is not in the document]
TheSpyder: Browsers have ridiculous sub-pixel positioning 😆 

At one point we tried capping the number of decimal places but it just lead to rounding errors.

modules/tinymce/src/core/demo/html/bundled_css_demo.html
```diff
@@ -0,0 +1,35 @@
...
+        </tbody>
+        </table>
+        <p><img src="https://www.w3schools.com/w3css/img_lights.jpg" alt="" width="600" height="400"></p>
+      </textarea>
+    </div>
+    <script src="../../../../js/tinymce/tinymce.js"></script>
+    <script src="../../../../js/tinymce/skins/ui/oxide/skin.js"></script>
+    <script src="../../../../js/tinymce/skins/content/default/content.js"></script>
+    <script src="../../../../js/tinymce/skins/ui/oxide/content.js"></script>
+    <script src="../../../../scratch/demos/core/demo.js"></script>
```
TheSpyder, line 32: ooh that's smart, nicely done

modules/tinymce/src/core/main/ts/api/dom/StyleSheetLoader.ts
```diff
@@ -166,9 +166,10 @@ const StyleSheetLoader = (documentOrShadowRoot: Document | ShadowRoot, settings:
     // Start loading
     const styleElem = SugarElement.fromTag('style', doc.dom);
     Attribute.setAll(styleElem, {
-      rel: 'stylesheet',
-      type: 'text/css',
-      id: state.id
+      'rel': 'stylesheet',
+      'type': 'text/css',
+      'id': state.id,
```
TheSpyder, line 171: These don't need to be string keys, only the key with `'` in it 🤔 
```suggestion
      rel: 'stylesheet',
      type: 'text/css',
      id: state.id,
```

modules/tinymce/src/themes/silver/main/ts/ui/skin/Loader.ts
```diff
@@ -11,6 +11,22 @@ declare let tinymce: TinyMCE;
 import * as Options from '../../api/Options';
 import * as SkinLoaded from './SkinLoaded';
 
+const getSkinResourceIdentifier = (editor: Editor): Optional<string> => {
+  const oxideUiSkinMap: Record<string, string> = {
+    'oxide-dark': 'dark',
+    'oxide': 'default',
+    'tinymce-5': 'tinymce-5',
+    'tinymce-5-dark': 'tinymce-5-dark'
```
TheSpyder, line 19: Since these last two are the same, can we get rid of them?

Auto-translating skin `oxide-dark` to content `dark` seems like a new feature? I can't find any reference to it - all the docs say both settings are required. I like the idea, but maybe this is not the best time 🤔

modules/tinymce/src/themes/silver/main/ts/ui/skin/Loader.ts
```diff
@@ -23,58 +39,97 @@ const loadRawCss = (editor: Editor, key: string, css: string, styleSheetLoader:
...
+  'ui/' + identifier + '/' + filename;
+
+const getResourceValue = (resourceKey: string): Optional<string> =>
+  Optional.from(tinymce.Resource.get(resourceKey)).filter(Type.isString);
+
+const loadCSS = (
+  editor: Editor,
+  filenameBase: string,
+  onLoadRaw: (key: string, css: string) => Promise<void>,
+  onLoadStyleSheet: (styleSheetUrl: string) => Promise<void>,
```
TheSpyder, line 52: There was a time when we'd use a proper `ADT` for this instead of approximating it with optional some/none 🥲

TheSpyder (approved):
Very nice!

modules/tinymce/src/core/main/ts/api/dom/StyleSheetLoader.ts
ltrouton: eslint complained at me as it either wants the keys to be all wrapped in `'` or none of them

modules/tinymce/src/themes/silver/main/ts/ui/skin/Loader.ts
ltrouton: The issue is that when you specify `skin: 'oxide-dark'`, the `skin.js` file has `tinymce.Resource.add('ui/dark/skin.css'` so I needed to do a mapping to get the correct resource key value.

Now that I think about it, maybe that is where the fix should be made and the resource key should be  `tinymce.Resource.add('ui/oxide-dark/skin.css'` instead so a mapping isn't needed

modules/tinymce/src/themes/silver/main/ts/ui/skin/Loader.ts
ltrouton: A typescript-based ADT is doable. I am not sure if it is worth it in this case though. I'll have a think.

modules/tinymce/src/themes/silver/main/ts/ui/skin/Loader.ts
TheSpyder: Oh! you've discovered a flaw in our theme build process. I knew I'd seen this mapping before. In the source folders, the skin is called `dark`.
https://github.com/tinymce/tinymce/blob/3eb6af8a9de284a67ec677fc825ef437bb6394f2/modules/oxide/src/less/skins/ui/dark/skin.less

We rename it in the tinymce bundle, but we don't replace the contents of the new JS files that are based on the source skin name.
https://github.com/tinymce/tinymce/blob/48e4df0806fd0ac73a88fffc4220b89f53712ba7/modules/tinymce/Gruntfile.js#L28-L33

We can either apply the rename to the content during the build, or rename the source skin to match the one we ship.

ltrouton: I would be hesitant to change the source skin names at the 'oxide'module level given we rely on it in the premium skins.
I'll look at replacing the resource key it in the tinymce Gruntfile

modules/tinymce/src/core/test/ts/browser/dom/BundledCssTest.ts
```diff
@@ -0,0 +1,252 @@
...
+              for (const skin of [ 'oxide', 'oxide-dark', 'tinymce-5', 'tinymce-5-dark' ] as UiSkin[]) {
+                context('skin: ' + skin, () => {
+                  context('no bundling', () => {
+                    it('TINY-11558: Load CSS styesheets as links', async () => {
+                      const [ editor, cleanup ] = await pCreateEditor(
+                        shadowdom,
+                        {
+                          inline,
+                          skin,
+                          ...!inline ? { content_css: contentCss } : {},
```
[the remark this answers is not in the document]
TheSpyder: Spread is syntax, not an operator, and therefore does not have precedence.
https://stackoverflow.com/a/48656377/7546

modules/tinymce/src/themes/silver/main/ts/ui/skin/Loader.ts
TheSpyder: @ltrouton don't think too hard, we already did that for you 😂 
https://github.com/tinymce/tinymce/blob/ba6994f39b09ecd2c28cef68c3e8a85a4815e235/modules/katamari/src/main/ts/ephox/katamari/api/Adt.ts

modules/tinymce/src/themes/silver/main/ts/ui/skin/Loader.ts
```diff
@@ -23,58 +33,97 @@ const loadRawCss = (editor: Editor, key: string, css: string, styleSheetLoader:
...
+  );
 
+const loadUrlSkin = async (isInline: boolean, editor: Editor): Promise<void> => {
   const skinUrl = Options.getSkinUrl(editor);
 
+  await loadUiContentCSS(editor, isInline, skinUrl);
+
   // In Modern Inline, this is explicitly called in editor.on('focus', ...) as well as in render().
   // Seems to work without, but adding a note in case things break later
   if (!Options.isSkinDisabled(editor) && Type.isString(skinUrl)) {
```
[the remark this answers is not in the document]
TheSpyder: This suggestion is invalid because the function is declared `async`. That means any non-promise return value will be wrapped in a promise; since all functions have an implicit void return at the end this guarantees a `Promise<void>` return value.

modules/tinymce/src/themes/silver/main/ts/ui/skin/Loader.ts
ltrouton: I am aware that exists in Katamari 😄. I know we largely moved away from using Adt in katamari in new code as it can be a bit un-wieldy to use and typed unions help a lot now

ltrouton: I have done some refactoring to hopefully make everything cleaner

PR #10033, opened 2024-12-04
TINY-11549: improve context toolbar positioning

lorenzo-pomili:
Related Ticket: TINY-11549

Description of Changes:
it seems that the problem was that the context toolbar insertion didn't consider the bounds so they were centered based on entire size of the element and not just based on the visible part.

At the moment I just add some unit test, if you confirm me that the fix make sense I'd like to add some "browser" test for the context toolbar since they are the focus of this Jira

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

## Summary by CodeRabbit

- **New Features**
  - Enhanced context toolbar positioning with improved horizontal placement data for a more precise and refined user experience.
  - Introduced a comprehensive suite of unit tests for context toolbar functionality in the TinyMCE editor.

- **Bug Fixes**
  - Resolved an issue where the toolbar did not utilize the full available space during horizontal scrolling, ensuring consistent and predictable layout behavior.

### DIFF
--- a/.changes/unreleased/alloy-TINY-11549-2024-12-04.yaml
+++ b/.changes/unreleased/alloy-TINY-11549-2024-12-04.yaml
@@ -1,6 +1,6 @@
 project: alloy
 kind: Improved
-body: Anchor center position now also consider boundaries.
+body: Centering an element now also considers boundaries.
 time: 2024-12-04T17:27:08.025517432+01:00
 custom:
   Issue: TINY-11549
--- a/.changes/unreleased/alloy-TINY-11549-2024-12-04.yaml
+++ b/.changes/unreleased/alloy-TINY-11549-2024-12-04.yaml
@@ -1,0 +1,6 @@
+project: alloy
+kind: Improved
+body: Anchor center position now also consider boundaries.
+time: 2024-12-04T17:27:08.025517432+01:00
+custom:
+  Issue: TINY-11549
--- a/.changes/unreleased/alloy-TINY-11549-2024-12-04.yaml
+++ b/.changes/unreleased/alloy-TINY-11549-2024-12-04.yaml
@@ -1,6 +1,0 @@
-project: alloy
-kind: Improved
-body: Centering an element now also considers boundaries.
-time: 2024-12-04T17:27:08.025517432+01:00
-custom:
-  Issue: TINY-11549
--- a/modules/alloy/src/main/ts/ephox/alloy/positioning/layout/LayoutInset.ts
+++ b/modules/alloy/src/main/ts/ephox/alloy/positioning/layout/LayoutInset.ts
@@ -34,7 +34,7 @@ const northY = (anchor: AnchorBox): number => anchor.y;
 const southY = (anchor: AnchorBox, element: AnchorElement): number => anchor.y + anchor.height - element.height;
 
 // returns centre of anchor minus half the element height - used to vertically centre element to the anchor
-const centreY = (anchor: AnchorBox, element: AnchorElement): number => anchor.y + (anchor.height / 2) - (element.height / 2);
+const centreY = (anchor: AnchorBox, element: AnchorElement, boundsHeight: number): number => anchor.y + (Math.min(anchor.height, boundsHeight) / 2) - (element.height / 2);
 
 // positions element relative to the bottom right of the anchor
 const southwest: AnchorLayout = (anchor: AnchorBox, element: AnchorElement, bubbles: Bubble) => NuSpotInfo(
@@ -103,9 +103,9 @@ const south: AnchorLayout = (anchor: AnchorBox, element: AnchorElement, bubbles:
 );
 
 // positions element with the right edge against the anchor, vertically centered
-const east: AnchorLayout = (anchor: AnchorBox, element: AnchorElement, bubbles: Bubble) => NuSpotInfo(
+const east: AnchorLayout = (anchor: AnchorBox, element: AnchorElement, bubbles: Bubble, _placee: SugarElement<HTMLElement>, bounds: Boxes.Bounds) => NuSpotInfo(
   eastEdgeX(anchor, element),
-  centreY(anchor, element),
+  centreY(anchor, element, bounds.height),
   bubbles.insetEast(),
   Direction.west(),
   Placement.East,
@@ -114,9 +114,9 @@ const east: AnchorLayout = (anchor: AnchorBox, element: AnchorElement, bubbles: 
 );
 
 // positions element with the left each against the anchor, vertically centered
-const west: AnchorLayout = (anchor: AnchorBox, element: AnchorElement, bubbles: Bubble) => NuSpotInfo(
+const west: AnchorLayout = (anchor: AnchorBox, element: AnchorElement, bubbles: Bubble, _placee: SugarElement<HTMLElement>, bounds: Boxes.Bounds) => NuSpotInfo(
   westEdgeX(anchor),
-  centreY(anchor, element),
+  centreY(anchor, element, bounds.height),
   bubbles.insetWest(),
   Direction.east(),
   Placement.West,
--- a/modules/alloy/src/main/ts/ephox/alloy/positioning/layout/LayoutInset.ts
+++ b/modules/alloy/src/main/ts/ephox/alloy/positioning/layout/LayoutInset.ts
@@ -1,3 +1,6 @@
+import { SugarElement } from '@ephox/sugar';
+
+import * as Boxes from '../../alien/Boxes';
 import { nu as NuSpotInfo } from '../view/SpotInfo';
 import { Bubble } from './Bubble';
 import * as Direction from './Direction';
@@ -19,7 +22,7 @@ const labelPrefix = 'layout-inset';
 const westEdgeX = (anchor: AnchorBox): number => anchor.x;
 
 // returns middle of anchor minus half the element width - used to horizontally centre element to the anchor
-const middleX = (anchor: AnchorBox, element: AnchorElement): number => anchor.x + (anchor.width / 2) - (element.width / 2);
+const middleX = (anchor: AnchorBox, element: AnchorElement, boundsWidth: number): number => anchor.x + (Math.min(anchor.width, boundsWidth) / 2) - (element.width / 2);
 
 // returns right edge of anchor minus element width - used to display element to the right, right edge against the anchor
 const eastEdgeX = (anchor: AnchorBox, element: AnchorElement): number => anchor.x + anchor.width - element.width;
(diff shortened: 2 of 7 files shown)


modules/alloy/src/test/ts/browser/position/layout/LayoutInsetTest.ts
```diff
@@ -65,5 +65,35 @@ describe('LayoutInsetTest', () => {
       const newLayout = LayoutInset.flip(Boxes.bounds(0, 0, 50, 50), { width: 10, height: 10 }, Bubble.fallback(), element, Boxes.win());
       assert.equal(newLayout.placement, Placement.North);
     });
+
+    it('TINY-11549: when the boudaries are smaller than the anchor element the position sohuld respect the middle of the boundaries and not the anchor element', () => {
```
tiny-ben-tran, line 69:
```suggestion
    it('TINY-11549: when the boundaries are smaller than the anchor element, the position should respect the middle of the boundaries and not the anchor element', () => {
```

tiny-ben-tran (commented):
I am not sure why such low level change is required. I think you can override the boundary to the positioning config for the toolbar positioning instead.

modules/alloy/src/test/ts/browser/position/layout/LayoutInsetTest.ts
lorenzo-pomili: done

lorenzo-pomili:
@tiny-ben-tran do you mean something like [renderFloatingMoreToolbar](https://github.com/tinymce/tinymce/blob/main/modules/tinymce/src/themes/silver/main/ts/ui/toolbar/CommonToolbar.ts#L161)?

if so I'm not sure it could work, there are already the correct bounds with the `width` of the visible area but the function that center the toolbar center it based on the element width and not the visible area

modules/alloy/src/main/ts/ephox/alloy/positioning/layout/LayoutInset.ts
```diff
@@ -19,7 +22,7 @@ const labelPrefix = 'layout-inset';
 const westEdgeX = (anchor: AnchorBox): number => anchor.x;
 
 // returns middle of anchor minus half the element width - used to horizontally centre element to the anchor
-const middleX = (anchor: AnchorBox, element: AnchorElement): number => anchor.x + (anchor.width / 2) - (element.width / 2);
+const middleX = (anchor: AnchorBox, element: AnchorElement, boundsWidth: number): number => anchor.x + (Math.min(anchor.width, boundsWidth) / 2) - (element.width / 2);
```
spocke, line 25: I wonder if there is ever a situation where we want to be sure that the middle is calculated to the middle of the anchor regardless of what the bound is so the old behavior. Since now it's not really a middle since it's clamped by the bounds of the view. So do we need both, like a clampedMiddleX but then you need to special mode all over the place to switch between things is that just over complicating things. But I know to little about the positioning apis to know what this could break.

spocke (approved):

modules/alloy/src/main/ts/ephox/alloy/positioning/layout/LayoutInset.ts
lorenzo-pomili: ok, maybe I found a problem, I found a strange behavior in inline editor, that seems better that the previous one, but I need to make some more test

tiny-ben-tran:
> if so I'm not sure it could work, there are already the correct bounds with the `width` of the visible area but the function that center the toolbar center it based on the element width and not the visible area

I see.

.changes/unreleased/alloy-TINY-11549-2024-12-04.yaml
```diff
@@ -0,0 +1,6 @@
+project: alloy
+kind: Improved
+body: Anchor center position now also consider boundaries.
```
tiny-ben-tran, line 3:
```suggestion
body: Centering an element now also considers boundaries.
```
??

modules/alloy/src/main/ts/ephox/alloy/positioning/layout/LayoutInset.ts
lorenzo-pomili: ok, it wasn't better in the previous version but the problem seems to resent only when you scroll horizontally and you have an image larger that the visible area, I think it could be something out of scope for this one

.changes/unreleased/alloy-TINY-11549-2024-12-04.yaml
lorenzo-pomili: I changed, but I have a doubt now :sweat_smile: , is not that the action of centering weren't consider the boundaries, but that the center position weren't considering the boundaries but only the element.

Should I revert the change fixing the conjugation of the verb?
```
body: Anchor center position now also considers boundaries.
```

spocke: What about says what the change is but very long. 😊
Anchor center X position was calculated based on the anchor width instead of the anchor width clamped by the bounds width.

TheSpyder: Alloy changelogs don't matter too much - they're not published to customers.

Or at all, until we get around to rebuilding the open source publishing scripts.

TheSpyder (commented):
This seems like the wrong approach. The AnchorBox is just a rectangle; if you want to clamp it to the visible area that should be done by the code that calls layout, not the layout itself. It will then apply to all layouts.

This is just complicating the layout code while adding restrictions that we might want to change later, similar to Spocke's `clampedMiddleX` suggestion.

My initial guess at where to look for clamping the anchor is here:
https://github.com/tinymce/tinymce/blob/1eaf8d7fbd7e323cd9b793094de39c70e4628dcd/modules/alloy/src/main/ts/ephox/alloy/behaviour/positioning/PositionApis.ts#L35

You could make a new `Anchor` method for it?

spocke:
> This seems like the wrong approach. The AnchorBox is just a rectangle; if you want to clamp it to the visible area that should be done by the code that calls layout, not the layout itself. It will then apply to all layouts.
> 

It depends on if we consider not clamping by bounds a oversight in the original code or a specific problem to toolbar positioning. Like why are we passing on the bounds into those calculations if we are not even using them.

TheSpyder:
> Like why are we passing on the bounds into those calculations if we are not even using them.

Probably confusion about what layouts were designed to do. `LayoutInsets` didn't even exist in the old TBIO code, but it still had inline toolbars just fine. Looks like the `bounds` addition happened in this giant change when we added transitions to smooth the toolbar moving between elements:
https://github.com/tinymce/tinymce/pull/7118

I didn't block the PR, I don't have time to come up with a better solution, I just hoped there would be one.

.changes/unreleased/alloy-TINY-11549-2024-12-04.yaml
lorenzo-pomili: changed with: `Anchor center X position was calculated based on the anchor width instead of the anchor width clamped by the bounds width.`

ztomaszyk (approved):

TheSpyder:
I'll probably need some time to get my head back in this space before I can comment on the latest change. With the APAC kickoff next week that might not happen until Thursday, sorry.

.changes/unreleased/alloy-TINY-11549-2024-12-04.yaml
```diff
@@ -0,0 +1,6 @@
+project: alloy
+kind: Improved
+body: Centering an element now also considers boundaries.
```
spocke, line 3: So this is no longer a change to alloy right? So remove this? We also need a changelog for the `tinymce` project since this is new behaviour.

modules/alloy/src/test/ts/browser/position/layout/LayoutInsetTest.ts
```diff
@@ -65,5 +65,35 @@ describe('LayoutInsetTest', () => {
       const newLayout = LayoutInset.flip(Boxes.bounds(0, 0, 50, 50), { width: 10, height: 10 }, Bubble.fallback(), element, Boxes.win());
       assert.equal(newLayout.placement, Placement.North);
     });
+
+    it.skip('TINY-11549: when the boundaries are smaller than the anchor element, the position should respect the middle of the boundaries and not the anchor element', () => {
```
spocke, line 69: We need to tweak this to handle the new behavior right.

modules/tinymce/src/themes/silver/main/ts/ui/context/ContextToolbarAnchor.ts
```diff
@@ -94,7 +94,9 @@ const getAnchorSpec = (editor: Editor, mobile: boolean, data: PositionData, posi
     const newAnchor = {
       ...anchor,
       y: bounds.y,
-      height: bounds.height
+      height: bounds.height,
+      x: bounds.x,
```
spocke, line 98: So this fix will be specific to the context toolbars that is nice. Can we add test for that?

spocke (approved):

modules/tinymce/src/themes/silver/main/ts/ui/context/ContextToolbarAnchor.ts
lorenzo-pomili: I didn't add a test or removed the changelog's entry and the old test, because I wanted an opinion from you and @TheSpyder to understand if the new version of the fix is better then the previous one (checkout to `f5d376226c220a49389e3de00d35a8bfe8f66f84` to check it), I still think that `LayoutInsert` should consider the correct boundaries, but if you tell me this solution is better I'll go with the test and the changelog if you think it was better before I'll switch back to the previous version

lorenzo-pomili: I was assuming that the `bounds` were by definition something more strict that the `anchor` or something that define the max available area, but now that you mentioned I'm realizing that it was just an assumption.

I'll check it better and see if I need a couple of `Math.min` if I can take the `bounds` as is or some other option

modules/alloy/src/test/ts/browser/position/layout/LayoutInsetTest.ts
lorenzo-pomili: done

.changes/unreleased/alloy-TINY-11549-2024-12-04.yaml
lorenzo-pomili: done

modules/tinymce/src/themes/silver/main/ts/ui/context/ContextToolbarAnchor.ts
lorenzo-pomili: @TheSpyder  the `Math.min` was needed so I staid with the overwrite of anchor fields, I also added a comment that I'm not sure is needed/clear, let me know what do you think about it

TheSpyder: @lorenzo-pomili what this suggests to me is you could use `...bounds` instead of `...anchor`, and then keep the Math.min for `x` and `width`.

lorenzo-pomili: ok, I see, I changed it

PR #10038, opened 2024-12-09
TINY-11500: When validating, also remove invalid attribute names

HAFRMO:
Related Ticket: TINY-11500

Description of Changes:
A regex-check on attribute names to ensure that they are valid. The entirety of the valid regexes could not be added due to javascript regex being limited to four-digit hex ( Remaining still in a comment if this can be updated in the future ).
Running some efficiency tests suggests that this regex-check has a minor impact on speed even larger files with a lot of attributes.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/.changes/unreleased/tinymce-TINY-11500-2024-12-09.yaml
+++ b/.changes/unreleased/tinymce-TINY-11500-2024-12-09.yaml
@@ -1,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: Attribute names with invalid characters were not properly filtered out.
+time: 2024-12-09T17:29:29.84038+01:00
+custom:
+    Issue: TINY-11500
--- a/modules/sugar/src/main/ts/ephox/sugar/api/dom/Replication.ts
+++ b/modules/sugar/src/main/ts/ephox/sugar/api/dom/Replication.ts
@@ -1,3 +1,5 @@
+import { Arr, Obj } from '@ephox/katamari';
+
 import { HTMLElementFullTagNameMap } from '../../alien/DomTypes';
 import { SugarElement } from '../node/SugarElement';
 import * as Attribute from '../properties/Attribute';
@@ -5,6 +7,11 @@ import * as Traverse from '../search/Traverse';
 import * as Insert from './Insert';
 import * as InsertAll from './InsertAll';
 import * as Remove from './Remove';
+
+// eslint-disable-next-line max-len
+const startAttributeCharacterLayout = '[A-Za-z:_]|[\u{C0}-\u{D6}]|[\u{D8}-\u{F6}]|[\u{F8}-\u{2FF}]|[\u{370}-\u{37D}]|[\u{37F}-\u{1FFF}]|[\u{200C}-\u{200D}]|[\u{2070}-\u{218F}]|[\u{2C00}-\u{2FEF}]|[\u{3001}-\u{D7FF}]|[\u{F900}-\u{FDCF}]|[\u{FDF0}-\u{FFFD}]';// |[\u{10000}-\u{EFFFF}]';
+const followingAttributeCharacterLayout = startAttributeCharacterLayout + '|"-"|"."|[0-9]|\u{B7}|[\u{0300}-\u{036F}]|[\u{203F}-\u{2040}]';
+const attributeNameRegex = RegExp(`^(${startAttributeCharacterLayout})(${followingAttributeCharacterLayout})*$`);
 
 const clone = <E extends Node> (original: SugarElement<E>, isDeep: boolean): SugarElement<E> =>
   SugarElement.fromDom(original.dom.cloneNode(isDeep) as E);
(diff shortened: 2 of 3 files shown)


.changes/unreleased/tinymce-TINY-11500-2024-12-09.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: Attribute names with invalid characters were not properly filtered out.
```
TheSpyder, line 3:
```suggestion
body: Malformed attribute names caused a parse error when loading content.
```

modules/sugar/src/main/ts/ephox/sugar/api/dom/Replication.ts
```diff
@@ -6,6 +8,11 @@ import * as Insert from './Insert';
 import * as InsertAll from './InsertAll';
 import * as Remove from './Remove';
 
+// eslint-disable-next-line max-len
+const startAttributeCharacterLayout = '[A-Za-z:_]|[\u{C0}-\u{D6}]|[\u{D8}-\u{F6}]|[\u{F8}-\u{2FF}]|[\u{370}-\u{37D}]|[\u{37F}-\u{1FFF}]|[\u{200C}-\u{200D}]|[\u{2070}-\u{218F}]|[\u{2C00}-\u{2FEF}]|[\u{3001}-\u{D7FF}]|[\u{F900}-\u{FDCF}]|[\u{FDF0}-\u{FFFD}]';// |[\u{10000}-\u{EFFFF}]';
```
TheSpyder, line 12: Please include notes about where this regex came from, so we have a better chance of maintaining this section of code in the future.

modules/sugar/src/main/ts/ephox/sugar/api/dom/Replication.ts
```diff
@@ -6,6 +8,11 @@ import * as Insert from './Insert';
...
+const followingAttributeCharacterLayout = startAttributeCharacterLayout + '|"-"|"."|[0-9]|\u{B7}|[\u{0300}-\u{036F}]|[\u{203F}-\u{2040}]';
+const attributeNameRegex = RegExp(`^(${startAttributeCharacterLayout})(${followingAttributeCharacterLayout})*$`);
```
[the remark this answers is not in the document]
TheSpyder: Also, since you don't use the grouping, make them non-capturing groups e.g. `(?..)` which might help with performance 🤔 

In fact I don't think you need the first group at all.

modules/sugar/src/main/ts/ephox/sugar/api/dom/Replication.ts
```diff
@@ -22,6 +29,11 @@ const shallowAs = <K extends keyof HTMLElementFullTagNameMap> (original: SugarEl
   const nu = SugarElement.fromTag(tag);
 
   const attributes = Attribute.clone(original);
+  Arr.each(Obj.keys(attributes), (key) => {
+    if (!attributeNameRegex.test(key)) {
+      delete attributes[key];
+    }
+  });
```
TheSpyder, line 36: Makes me wonder if this should go in `Attribute.clone` 🤔

But then looking at _that_ code, perhaps this should inline `clone` rather than do what amounts to 3 iterations over the attribute list (once you look at what `setAll()` does)...

Maybe the whole function could be a single loop over `element.dom.attributes` that only calls `Attribute.set()` if the name is valid. Less pressure on the garbage collector, less looping.

modules/tinymce/src/core/test/ts/browser/html/SanitizationTest.ts
```diff
@@ -22,6 +22,13 @@ describe('browser.tinymce.core.html.SanitizationTest', () => {
       mimeType: 'text/html'
     }));
 
+    it('TINY-11500: Sanitize with invalid attribute name in content', () => testHtmlSanitizer({
```
spocke, line 25: This isn't testing the validation properly. You need to also pass in `valid_elements: '*[*]'` to the Schema int he `testHtmlSanitizer` otherwise it will just filter out anything that doesn't have valid names. If you do the above these invalid attributes are still there.
Also we should have a higher up test where it tests this with a editor instance as well. These should be similar tests for SVGs/MathML etc that you can look at.

- [ ] Add proper filtering in the parser likely in the attribute filter that the DOMPurify runs

spocke (changes requested):

modules/tinymce/src/core/test/ts/browser/html/SanitizationTest.ts
spocke: Added a fiddle here that shows what is expected to work: https://fiddle.tiny.cloud/sUlDxSfbZs/0
Initial parsing and inserting of contents with these invalid attributes should not be possible if we fix that then they are filtered away for all other operations in the editor.

modules/sugar/src/main/ts/ephox/sugar/api/dom/Replication.ts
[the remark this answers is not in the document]
HAFRMO: The first group is unfortunately needed. But I'll turn them both into non-capture.

PR #10043, opened 2024-12-13
TINY-11377: Converted `isEmpty` to an assertion function, improving cases with optional object properties

TheSpyder:
Related Ticket: TINY-11377

Description of Changes:
* Not actually related to the ticket, something I came across while working on it. The code in the test won't compile without the assertion return type and it's a pattern I was trying to use.

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [ ] Milestone set

## Summary by CodeRabbit

- **New Features**
  - Introduced a new function to create potentially empty objects for testing.
  
- **Bug Fixes**
  - Enhanced type safety of the `isEmpty` function, ensuring accurate type inference for empty objects.

- **Tests**
  - Expanded the test suite for `isEmpty` with additional cases to validate TypeScript's understanding of empty object types.

### DIFF
--- a/modules/katamari/src/test/ts/atomic/api/obj/ObjIsEmptyTest.ts
+++ b/modules/katamari/src/test/ts/atomic/api/obj/ObjIsEmptyTest.ts
@@ -4,7 +4,7 @@ import fc from 'fast-check';
 
 import * as Obj from 'ephox/katamari/api/Obj';
 
-const makePossiblyEmptyValue = (): { hasKey: boolean } | {} => ({ hasKey: true });
+const makePossiblyEmptyValue = (isEmpty: boolean): { hasKey: boolean } | {} => isEmpty ? {} : { hasKey: true };
 
 describe('atomic.katamari.api.obj.ObjIsEmptyTest', () => {
   it('unit tests', () => {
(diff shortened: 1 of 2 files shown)


modules/katamari/src/test/ts/atomic/api/obj/ObjIsEmptyTest.ts
```diff
@@ -20,4 +22,12 @@ describe('atomic.katamari.api.obj.ObjIsEmptyTest', () => {
       }
     ));
   });
+
+  it('helps TypeScript understand the empty type', () => {
```
spocke, line 26: Could we also add some form of tests that checks when it's empty. Like `if (isEmpty({})) { assert.deepEqual({}, {}) }` or something os that it checks that result type became an empty object.

spocke (approved):

modules/katamari/src/test/ts/atomic/api/obj/ObjIsEmptyTest.ts
TheSpyder: TypeScript allows assigning any values to a `{}`	type, so it's mostly a pointless test, but sure.

kimwoodfield (approved):

TheSpyder:
Test failure appears to be an unrelated flake, at least it was on `main`

PR #10075, opened 2025-01-07
TINY-11498: Fix error with notifications when toolbar is at the bottom

shanmen-tiny:
Related Ticket:  TINY-11498

Description of Changes:
* Fixed condition to apply docking and call Docking.refresh when not using a sticky toolbar or when the toolbar is at the bottom
	* When using a sticky toolbar that's located at the top, since we are attaching it to the anchor bar, there's no need to apply docking. It will stay attached as we scroll.
	* When not using a sticky toolbar or when the toolbar is located at the bottom, we need the notification to be docked at the top of the window

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/.changes/unreleased/tinymce-TINY-11498-2025-01-06.yaml
+++ b/.changes/unreleased/tinymce-TINY-11498-2025-01-06.yaml
@@ -1,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: Fixed error with notifications when toolbar is at the bottom.
+time: 2025-01-06T10:12:46.807235+08:00
+custom:
+  Issue: TINY-11498
--- a/.changes/unreleased/tinymce-TINY-11498-2025-01-06.yaml
+++ b/.changes/unreleased/tinymce-TINY-11498-2025-01-06.yaml
@@ -1,6 +1,6 @@
 project: tinymce
 kind: Fixed
-body: Fixed error with notifications when toolbar is at the bottom.
+body: Opening a notification when the toolbar is at the bottom threw an error.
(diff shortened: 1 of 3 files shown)


.changes/unreleased/tinymce-TINY-11498-2025-01-06.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: Fixed error with notifications when toolbar is at the bottom.
```
kimwoodfield, line 3: Just a super small nitpick but do we need `Fixed` in the `body` when it's already in `kind`?

kimwoodfield (approved):

.changes/unreleased/tinymce-TINY-11498-2025-01-06.yaml
shanmen-tiny: Reworded 👍

modules/tinymce/src/themes/silver/main/ts/alien/NotificationManagerImpl.ts
```diff
@@ -79,6 +79,8 @@ export default (
       });
     };
 
+    const shouldApplyDocking = () => !Options.isStickyToolbar(editor) || !sharedBackstage.header.isPositionedAtTop();
```
spocke, line 82: Could this use `comp.hasConfigured(Docking)` instead? We always want to refresh if it's docked right?

spocke (approved):

PR #10077, opened 2025-01-07
TINY-11707: Add option link_attributes_postprocess that allows overriding attributes of inserted link through link plugin dialog

hamza0867:
Related Ticket: TINY-11707

Description of Changes:
* Adds option link_attributes_postprocess that allows overriding attributes of inserted link through link plugin dialog

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

## Summary by CodeRabbit

- **New Features**
  - Introduced `link_attributes_postprocess` option for TinyMCE.
  - Allows users to override link attributes when inserting links through the link dialog.

- **Documentation**
  - Added documentation for the new link attributes post-processing feature.

- **Tests**
  - Created a comprehensive test suite to verify the functionality of link attributes post-processing.

### DIFF
--- a/.changes/unreleased/tinymce-TINY-11707-2025-01-07.yaml
+++ b/.changes/unreleased/tinymce-TINY-11707-2025-01-07.yaml
@@ -1,0 +1,7 @@
+project: tinymce
+kind: Added
+body: '`link_attributes_postprocess` option that allows overriding attributes of a
+  link that would be inserted through the link dialog.'
+time: 2025-01-07T20:12:31.813482+01:00
+custom:
+  Issue: TINY-11707
--- a/modules/tinymce/src/plugins/link/test/ts/browser/AttributesPostprocessTest.ts
+++ b/modules/tinymce/src/plugins/link/test/ts/browser/AttributesPostprocessTest.ts
@@ -9,7 +9,7 @@ import Plugin from 'tinymce/plugins/link/Plugin';
 
 import { TestLinkUi } from '../module/TestLinkUi';
 
-describe('browser.tinymce.plugins.link.AttributesPostprocess', () => {
+describe('browser.tinymce.plugins.link.AttributesPostprocessTest', () => {
   const newRel = 'noopener noreferrer';
   const newHref = 'https://www.google.com';
   const attrsOverride = (attrs: any) => {
--- a/modules/tinymce/src/plugins/link/test/ts/browser/AttributesPostprocessTest.ts
+++ b/modules/tinymce/src/plugins/link/test/ts/browser/AttributesPostprocessTest.ts
@@ -1,0 +1,64 @@
+import { ApproxStructure, FocusTools, Waiter } from '@ephox/agar';
+import { describe, it, before, after } from '@ephox/bedrock-client';
+import { Fun } from '@ephox/katamari';
+import { SugarDocument } from '@ephox/sugar';
+import { TinyAssertions, TinyHooks, TinySelections } from '@ephox/wrap-mcagar';
+
+import Editor from 'tinymce/core/api/Editor';
+import Plugin from 'tinymce/plugins/link/Plugin';
+
+import { TestLinkUi } from '../module/TestLinkUi';
+
+describe('browser.tinymce.plugins.link.AttributesPostprocess', () => {
+  const newRel = Fun.constant('noopener noreferrer');
+  const attrsOverride = (attrs: any) => {
+    attrs.rel = newRel();
+  };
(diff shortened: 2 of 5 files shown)


hamza0867:
* **#10077** <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10077?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10077?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

.changes/unreleased/tinymce-TINY-11707-2025-01-07.yaml
```diff
@@ -0,0 +1,7 @@
+project: tinymce
+kind: Added
+body: '`link_attributes_postprocess` option that allows overriding attributes of a
+  link that would be inserted through the link dialog.'
```
spocke, line 4: I think it looks a bit weird if we just have the backticks at the beginning of the sentence. We used `New` before.

```suggestion
body: 'New `link_attributes_postprocess` option that allows overriding attributes of a
  link that would be inserted through the link dialog.'
```

modules/tinymce/src/plugins/link/test/ts/browser/AttributesPostprocessTest.ts
```diff
@@ -0,0 +1,64 @@
...
+import { Fun } from '@ephox/katamari';
+import { SugarDocument } from '@ephox/sugar';
+import { TinyAssertions, TinyHooks, TinySelections } from '@ephox/wrap-mcagar';
+
+import Editor from 'tinymce/core/api/Editor';
+import Plugin from 'tinymce/plugins/link/Plugin';
+
+import { TestLinkUi } from '../module/TestLinkUi';
+
+describe('browser.tinymce.plugins.link.AttributesPostprocess', () => {
```
spocke, line 12: I think we use `Test` suffix on these labels.

```suggestion
describe('browser.tinymce.plugins.link.AttributesPostprocessTest', () => {
```

modules/tinymce/src/plugins/link/test/ts/browser/AttributesPostprocessTest.ts
```diff
@@ -0,0 +1,64 @@
...
+  const newRel = Fun.constant('noopener noreferrer');
```
spocke, line 13: Why the constant function if it's invoked to get the value out at both places? Some refactor that was done and this was left there? You could just use a string constant? Function constants is useful if you need to pass it to some function that is higher order.

modules/tinymce/src/plugins/link/test/ts/browser/AttributesPostprocessTest.ts
```diff
@@ -0,0 +1,64 @@
...
+  const attrsOverride = (attrs: any) => {
+    attrs.rel = newRel();
```
spocke, line 15: This only tests `rel` we should make sure that we can overrride all the attributes that it has so maybe a separate test that say adds something to every obj property.

- [ ] Add tests for more properties like `href` etc and make sure that the `data-mce-href` gets properly updated

modules/tinymce/src/plugins/link/test/ts/browser/AttributesPostprocessTest.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    FocusTools.setActiveValue(SugarDocument.getDocument(), 'http://www.tiny.cloud');
+    TestLinkUi.assertDialogContents({
+      href: 'http://www.tiny.cloud',
+      text: 'Something',
+      title: '',
+      target: ''
+    });
+    await TestLinkUi.pClickSave(editor);
+    await Waiter.pTryUntil('Waiting for content to have correct structure', () => {
```
spocke, line 46: Is this really async?
Can we use assertContent instead?

There is a risk at least on the `href` attribute that we don't properly assert the right value since this assertContentStructure should really have been named `assertRawContentStructure` since that is what it does it takes the current DOM state and asserts that but for the `href` attribute for example there might be the cached `data-mce-href` value and that is what will be used when serializing. So does the postprocess properly update the `href` so that both the `href` and the `data-mce-href` attributes gets the right value?

We can still use the `assertContentStructure` for asserting both values but we then must remember to do so, so that is why the raw assertions like this is a bit more risky. It's easy to forget.

spocke (approved):

PR #10079, opened 2025-01-08
TINY-11727: Fix floating promise eslint errors

ltrouton:
Related Ticket: TINY-11727

Description of Changes:
* Change `@typescript-eslint/no-floating-promises` from `'off'` to `'error'`
* Add override for `demo` files to ignore `tinymce.init` calls
* Helps catch un-awaited promises in tests
* Created https://ephocks.atlassian.net/browse/TINY-11735 as a follow up to try and resolve more of the floating promises if possible

Pre-checks:
* ~~[X] Changelog entry added~~
* ~~[X] Tests have been added (if applicable)~~

Review:
* [X] Milestone set

## Summary by CodeRabbit

# Release Notes

- **Linting and Code Quality**
  - Enhanced ESLint configuration to improve promise handling and code consistency
  - Added rules to manage unhandled promises across multiple modules

- **Asynchronous Improvements**
  - Updated various functions to use async/await for better promise handling
  - Improved control flow in test cases and editor initialization processes

- **Test Suite Enhancements**
  - Standardized asynchronous test function signatures
  - Improved reliability of test assertions by properly awaiting async operations

- **Editor Initialization**
  - Added more robust skin and resource loading mechanisms
  - Improved handling of editor setup and plugin loading

- **Bug Fixes**
  - Minor improvements to promise handling in various editor components
  - Enhanced error management in asynchronous operations

### DIFF
--- a/modules/agar/src/test/ts/browser/api/ClipboardTest.ts
+++ b/modules/agar/src/test/ts/browser/api/ClipboardTest.ts
@@ -23,7 +23,7 @@ describe('ClipboardTest', () => {
     }
   };
 
-  const assertStringItem = async (item: DataTransferItem, expected: { type: string; data: string }): Promise<void> => {
+  const pAssertStringItem = async (item: DataTransferItem, expected: { type: string; data: string }): Promise<void> => {
     assert.equal(item.type, expected.type);
     assert.equal(item.kind, 'string');
     assert.equal(await getItemData(item), expected.data);
@@ -88,8 +88,8 @@ describe('ClipboardTest', () => {
     const dataTransfer = pasteState.get().getOrDie('Could not get dataTransfer from state');
 
     assert.equal(dataTransfer.items.length, 2);
-    await assertStringItem(dataTransfer.items[0], { type: 'text/plain', data: 'Hello world!' });
-    await assertStringItem(dataTransfer.items[1], { type: 'text/html', data: '<b>Hello world!</b>' });
+    await pAssertStringItem(dataTransfer.items[0], { type: 'text/plain', data: 'Hello world!' });
+    await pAssertStringItem(dataTransfer.items[1], { type: 'text/html', data: '<b>Hello world!</b>' });
   });
 
   it('Paste text and html files', async () => {
@@ -119,7 +119,7 @@ describe('ClipboardTest', () => {
 
     assert.equal(dataTransfer.items.length, 2);
     await assertFileItem(dataTransfer.items[0], { type: 'text/plain', name: 'a.txt', data: 'Hello world!' });
-    await assertStringItem(dataTransfer.items[1], { type: 'text/html', data: '<b>Hello world!</b>' });
+    await pAssertStringItem(dataTransfer.items[1], { type: 'text/html', data: '<b>Hello world!</b>' });
   });
 
   it('Cut', async () => {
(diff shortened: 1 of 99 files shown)


spocke (approved):

spocke (approved):

modules/agar/src/test/ts/browser/api/ClipboardTest.ts
```diff
@@ -23,7 +23,7 @@ describe('ClipboardTest', () => {
     }
   };
 
-  const assertStringItem = async (item: DataTransferItem, expected: { type: string; data: string }): Promise<void> => {
+  const pAssertStringItem = async (item: DataTransferItem, expected: { type: string; data: string }): Promise<void> => {
```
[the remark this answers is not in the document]
ltrouton: That's true. I don't mind the convention though

tiny-ben-tran (approved):

modules/tinymce/src/core/test/ts/browser/EditorManagerCommandsTest.ts
```diff
@@ -36,6 +36,7 @@ describe('browser.tinymce.core.EditorManagerCommandsTest', () => {
   ], (test) => {
     it(`mceToggleEditor (${test.label})`, (done) => {
       viewBlock.update('<textarea id="ed_1" class="tinymce"></textarea>');
+      // eslint-disable-next-line @typescript-eslint/no-floating-promises
```
TheSpyder, line 39: Many of these are incorrect - what we should instead do is `await` the line below and make the test async. Which we did a month ago on the TINY-11177 branch 🫣 

At least now I can just search for the disable comment to find the rest of the places to fix...

TheSpyder: wait, my apologies, I was looking at the conflict before I properly understood the code.

These are callback tests, so the change to `await EditorManager.init()` is now duplicated by the call to `done()` in the init instance callback.

The question now is which test style will throw the correct error if the asserts fail 🤔

TheSpyder: yeaaaaah this never failed correctly because `init_instance_callback` isn't called during the `EditorManager.init()` promise. Doesn't matter if we use the new one or the old one.

It at least does fail the build with the new bedrock, because it finally catches the unhandled promise rejection correctly. The error is nonsense but at least there is _an_ error.

To fail correctly, we'd need some sort of helper that
* made a promise
* called `EditorManager.init()`
* resolved inside `init_instance_callback`
(I think I've seen code like that somewhere recently, so it shouldn't be a difficult pattern to extract

Then we just `await initInstance()` and run the assertions afterwards.

cc @shanmen-tiny

shanmen-tiny: Oops, I'll fix this in the TINY-11177 branch. 

I think this should be the pattern: https://github.com/tinymce/tinymce/blob/3bfbda75dd7ceb63e166f04e2564b5f5b3e32324/modules/tinymce/src/core/test/ts/browser/EditorAutoFocusTest.ts#L40

Resolving within `init_instance_callback` instead of on focus.

TheSpyder: Sounds like something that should go in mcagar 😂

PR #10089, opened 2025-01-10
TINY-11215 Fix table resize bars to remain properly aligned in scrollable container

shanmen-tiny:
Related Ticket: TINY-11215

Description of Changes:
* For inline mode when using `ui_mode: 'split'`, the resize container is now rendered adjacent to the editor body so that when scrolling the container, the bars are still attached to the table
* Instead of changing the `z-index` of the dragster blocker when dragging, insert the dragster blocker into the body.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/.changes/unreleased/tinymce-TINY-11215-2025-01-10.yaml
+++ b/.changes/unreleased/tinymce-TINY-11215-2025-01-10.yaml
@@ -1,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: Table resize bars now remain properly aligned when the inline editor is used inside a scrollable container.
+time: 2025-01-10T10:02:26.35372+08:00
+custom:
+  Issue: TINY-11215
--- a/.changes/unreleased/tinymce-TINY-11215-2025-01-10.yaml
+++ b/.changes/unreleased/tinymce-TINY-11215-2025-01-10.yaml
@@ -1,6 +1,6 @@
 project: tinymce
 kind: Fixed
-body: Table resize bars now remain properly aligned when the inline editor is used inside a scrollable container.
+body: Table resize bars was not properly aligned for inline editors inside scrollable containers.
 time: 2025-01-10T10:02:26.35372+08:00
 custom:
   Issue: TINY-11215
--- a/modules/dragster/src/main/ts/ephox/dragster/detect/Blocker.ts
+++ b/modules/dragster/src/main/ts/ephox/dragster/detect/Blocker.ts
@@ -21,11 +21,12 @@ export const Blocker = (options: Partial<BlockerOptions>): Blocker => {
   const div = SugarElement.fromTag('div');
   Attribute.set(div, 'role', 'presentation');
   Css.setAll(div, {
-    position: 'fixed',
-    left: '0px',
-    top: '0px',
-    width: '100%',
-    height: '100%'
+    'position': 'fixed',
+    'left': '0px',
+    'top': '0px',
+    'width': '100%',
+    'height': '100%',
+    'z-index': '10000'
   });
 
   Class.add(div, Styles.resolve('blocker'));
--- a/modules/dragster/src/main/ts/ephox/dragster/detect/Blocker.ts
+++ b/modules/dragster/src/main/ts/ephox/dragster/detect/Blocker.ts
@@ -21,12 +21,11 @@ export const Blocker = (options: Partial<BlockerOptions>): Blocker => {
   const div = SugarElement.fromTag('div');
   Attribute.set(div, 'role', 'presentation');
   Css.setAll(div, {
-    'position': 'fixed',
-    'left': '0px',
-    'top': '0px',
-    'width': '100%',
-    'height': '100%',
-    'z-index': '10000'
+    position: 'fixed',
+    left: '0px',
+    top: '0px',
+    width: '100%',
+    height: '100%'
   });
 
   Class.add(div, Styles.resolve('blocker'));
--- a/modules/snooker/src/main/ts/ephox/snooker/resize/BarManager.ts
+++ b/modules/snooker/src/main/ts/ephox/snooker/resize/BarManager.ts
@@ -1,5 +1,5 @@
 import { Dragger } from '@ephox/dragster';
-import { Cell, Fun, Optional, Singleton } from '@ephox/katamari';
+import { Fun, Optional } from '@ephox/katamari';
 import { Bindable, Event, Events } from '@ephox/porkbun';
 import { Attribute, Class, Compare, ContentEditable, Css, DomEvent, SelectorFind, SugarBody, SugarElement } from '@ephox/sugar';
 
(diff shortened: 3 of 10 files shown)


modules/tinymce/src/models/dom/test/ts/browser/table/ResizeTableTest.ts
```diff
@@ -1157,4 +1163,258 @@ describe('browser.tinymce.models.dom.table.ResizeTableTest', () => {
...
+      const rowRect = row.getBoundingClientRect();
+      const rowTopRelativeToEditor = rowRect.top - editorRect.top;
+
+      // Get table's border spacing
+      const borderSpacing = parseInt(Css.get(SugarElement.fromDom(table), 'border-spacing'), 10);
+
+      // For all rows except last, offset includes the border spacing
+      const isLastRow = rowIndex === table.rows.length - 1;
+      const expectedOffset = isLastRow ?
+        19 : // Last row: no need to account for spacing
```
kimwoodfield, line 1213: Small nitpick but why did we go with `19`? 

Maybe we can store that in a separate variable with a name explaining what it is or why we are using it or at least provide a comment why it was chosen.

kimwoodfield (approved):

.changes/unreleased/tinymce-TINY-11215-2025-01-10.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: Table resize bars now remain properly aligned when the inline editor is used inside a scrollable container.
```
spocke, line 3: I think we tend to use past tense so something like:

```suggestion
body: Table resize bars was not properly aligned for inline editors inside scrollable containers.
```

modules/tinymce/src/models/dom/main/ts/table/api/TableResizeHandler.ts
```diff
@@ -120,14 +121,53 @@ export const TableResizeHandler = (editor: Editor): TableResizeHandler => {
...
+      resizeWire.clear();
     });
   };
 
-  editor.on('init', () => {
+  const bindEvents = (sz: TableResize) => {
+    sz.events.startDrag.bind((_event) => {
+      isDragging.set(true);
+      selectionRng.set(editor.selection.getRng());
+      editor.dispatch('blur', { focusedEditor: null });
```
spocke, line 137: I don't think this is correct interacting with editor UI should not dispatch a blur event of the editor. Blur should only happen if you move focus out side the editor content or editor UI. The resize bars are part of the editor content UI. Oddly enough it does trigger blur/focus on inline today but not in iframe mode but I think that is a bug and I think we should continue on that path. Both iframe and inline mode should work the same here ideally.

modules/dragster/src/main/ts/ephox/dragster/detect/Blocker.ts
```diff
@@ -21,11 +21,12 @@ export const Blocker = (options: Partial<BlockerOptions>): Blocker => {
...
-    left: '0px',
-    top: '0px',
-    width: '100%',
-    height: '100%'
+    'position': 'fixed',
+    'left': '0px',
+    'top': '0px',
+    'width': '100%',
+    'height': '100%',
+    'z-index': '10000'
```
spocke, line 29: Are we confident that this value is high enough. I looks like the floating UI In inline mode is hidden while resizing can we add some comment what it's higher than. I guess at some point we should have these things defined. It looks like the resize corners have indexes around 1000 and that the current version has a bug where the blocker has lower z-index that than so resizing stops if you hit that corner I wonder if there are other things like that.

modules/snooker/src/main/ts/ephox/snooker/resize/BarManager.ts
```diff
@@ -133,34 +140,42 @@ export const BarManager = (wire: ResizeWire): BarManager => {
...
   });
 
   const destroy = () => {
-    mousedown.unbind();
-    mouseover.unbind();
     resizing.destroy();
     Bars.destroy(wire);
   };
 
+  const unbindEvents = () => {
```
spocke, line 162: Why was this separated, having a destroy that doesn't unbind the events seems odd? Maybe `destroy` needs to be renamed `hide` or something and a new `destroy` just calls `hide` and `unbindEvents`. Destroy to me at least means remove everything about the resize bars, the DOM structure, the events etc. Like go back to what it was before we added it.

modules/tinymce/src/models/dom/main/ts/table/api/TableResizeHandler.ts
```diff
@@ -138,32 +178,32 @@ export const TableResizeHandler = (editor: Editor): TableResizeHandler => {
...
-        Events.fireObjectResizeStart(editor, rawTable, Utils.getPixelWidth(rawTable), Utils.getPixelHeight(rawTable), barResizerPrefix + event.type);
-      });
+  editor.on('init', () => createResizeHandler());
 
-      sz.events.afterResize.bind((event) => {
-        const table = event.table;
-        const rawTable = table.dom;
-        Utils.removeDataStyle(table);
+  if (editor.inline) {
+    editor.on('NodeChange', () => {
```
spocke, line 189: I'm not that confident that `NodeChange` will not trigger while you are dragging. Since that can be triggered by calling `editor.nodeChange()` and we some times call that our self to update UI state. So things like auto saving, async operations like spellchecking or link checking could potentially call that. Can't we use `mouseup` instead or similar since that will only happen when the user has finished the drag operation and I guess that mouse up would be on the blocker or bar not in the editor event like this? Doesn't dragster handle all that drag state, like startDrag, dragging and dragEnd?

modules/tinymce/src/models/dom/main/ts/table/api/TableResizeHandler.ts
```diff
@@ -138,32 +178,32 @@ export const TableResizeHandler = (editor: Editor): TableResizeHandler => {
...
-        });
+      const targetElm = editor.selection.getNode();
+      const controlElm = SelectorFind.closest<HTMLElement>(SugarElement.fromDom(targetElm), 'table')
+        .map((e) => e.dom)
+        .getOrUndefined();
 
-        Events.fireObjectResized(editor, rawTable, Utils.getPixelWidth(rawTable), Utils.getPixelHeight(rawTable), barResizerPrefix + event.type);
-        editor.undoManager.add();
-      });
+      const isChildOrEqual = (node: Node, parent: Node): boolean =>
```
spocke, line 199: Heh yeah maybe I should have named that function `contains` since it acts like the browser contains function.

spocke (changes requested):

modules/tinymce/src/models/dom/main/ts/table/api/TableResizeHandler.ts
shanmen-tiny: Thanks for highlighting. Now I wonder if this is the correct approach to tackle the bug. I'm working on another way of handling all the issues highlighted, which doesn't require changing the `z-index` or using `NodeChange` to remove the container and add that back.

shanmen-tiny:
@spocke Would you mind taking another look at the implementation? The changes include:
* Removed `NodeChange` listener to add/remove the resize container
* Modified the container rendering location to synchronize with scrollable editor containers when scrolled
* z-index issues by moving the blocker to root during drag operations

ltrouton (approved):
Nice work

modules/tinymce/src/models/dom/test/ts/browser/table/ResizeTableTest.ts
shanmen-tiny: Reworked! I've used variables instead of magic numbers.

PR #10096, opened 2025-01-16
TINY-11740: Update `preventClicksOnLinksScript` to handle anchor links appropriately

kimwoodfield:
Related Ticket: [TINY-11740](https://ephocks.atlassian.net/browse/TINY-11740)

Description of Changes:
With this change, we are just adding a check to see if a link is an anchor before preventing the default browser (opening a link in a new tab) and then scrolling to the target element using `scrollIntoView()` to ensure the behaviour is consistent between the content body of the Editor and the preview dialog when navigating to anchor elements.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set
* ~[ ] Docs ticket created (if applicable)~

[TINY-11740]: https://ephocks.atlassian.net/browse/TINY-11740?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

- **Bug Fixes**
  - Resolved an issue in the Preview Dialog where anchor links were incorrectly opening in a new tab.
  - Improved link click prevention mechanism for macOS and iOS platforms.

- **New Features**
  - Added platform detection for macOS and iOS operating systems.

The update enhances the TinyMCE Preview Dialog's link handling and cross-platform compatibility.


modules/tinymce/src/plugins/preview/main/ts/core/IframeContent.ts
```diff
@@ -29,9 +29,17 @@ const getPreviewHtml = (editor: Editor): string => {
   const preventClicksOnLinksScript = (
     '<script>' +
     'document.addEventListener && document.addEventListener("click", function(e) {' +
-    'for (var elm = e.target; elm; elm = elm.parentNode) {' +
-    'if (elm.nodeName === "A" && !(' + isMetaKeyPressed + ')) {' +
+    'for (let elm = e.target; elm; elm = elm.parentNode) {' +
+    'if (elm.nodeName === "A") {' +
+    'const href = elm.getAttribute("href");' +
+    'if (href && href.startsWith("#") && (' + isMetaKeyPressed + ')) {' +
```
tiny-ben-tran, line 35: Why was the negation operator `!` removed?

tiny-ben-tran: I didn't see that it was moved down below. All good

shanmen-tiny (approved):

modules/tinymce/src/plugins/preview/main/ts/core/IframeContent.ts
```diff
@@ -29,9 +29,17 @@ const getPreviewHtml = (editor: Editor): string => {
...
     'e.preventDefault();' +
+    'document.getElementById(href.substring(1)).scrollIntoView();' +
```
TheSpyder, line 37:
```suggestion
    'document.getElementById(href.substring(1))?.scrollIntoView();' +
```
Just in case

modules/tinymce/src/plugins/preview/main/ts/core/IframeContent.ts
TheSpyder, line 35:
```suggestion
    'if (href && href.startsWith("#")) {' +
```

I don't remember where I said this, but I think we should scroll whether meta is pressed or not.

TheSpyder (approved):
It's 2025, I think we can convert this to a template string now so it's easier to read 😉

modules/sugar/src/main/ts/ephox/sugar/api/dom/Link.ts
```diff
@@ -23,6 +25,28 @@ const addStylesheet = (url: string, scope?: SugarElement<Document>): SugarElemen
   return link;
 };
 
+const generatePreventClicksOnLinksScript = (
+  fn: (isMetaKeyPressed: (e: MouseEvent) => boolean) => void,
+  isMetaKeyPressed: (e: MouseEvent) => boolean
+): string => {
+  return `<script>(${fn.toString()})(${isMetaKeyPressed.toString()})</script>`;
```
ltrouton, line 32: I am probably missing something but what is the `$isMetaKeyPressed.toString()}` for? I can't see a reference to that in IframeContent.ts

TheSpyder: It's used to avoid needing the `Platform` module code in the function after it's been converted to a string - or at least that's the intent, he's not quite there yet 😅

kimwoodfield: @ltrouton It was being passed in for the purpose of using inside of the function which registers the click event listener so that we could determine whether the meta key had been pressed, allowing us to prevent the default behavior of the browser for links.

As Andy mentioned, I was still working on that.

With the most recent commit, rather than serializing that `isMetaKeyPressed` function and including it in the script, we evaluate whether the platform is iOS/MacOS and then serialize that boolean output in its place so that when we inject the script tags into the head of the iframe, we can simply rely on this boolean value rather than needing to bring along another function with logic from `Platform`.

ltrouton (approved):

modules/sugar/src/main/ts/ephox/sugar/api/dom/Link.ts
ltrouton: Thanks for the clarification 👍

modules/tinymce/src/plugins/preview/main/ts/core/IframeContent.ts
```diff
@@ -49,7 +36,7 @@ const getPreviewHtml = (editor: Editor): string => {
     '</head>' +
     '<body id="' + encode(bodyId) + '" class="mce-content-body ' + encode(bodyClass) + '"' + dirAttr + '>' +
     editor.getContent() +
-    preventClicksOnLinksScript +
+    Link.getPreventClicksOnLinksScript() +
```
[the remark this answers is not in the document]
TheSpyder: The code does do this, it's just not obvious because we pass in platform-specific checks at runtime.

modules/sugar/src/main/ts/ephox/sugar/impl/ClickBehavior.ts
```diff
@@ -0,0 +1,23 @@
...
+          return;
+        }
+
+        if (!isMetaKeyPressed(e)) {
+          e.preventDefault();
+        }
+      }
+    }
+  }, false);
+};
```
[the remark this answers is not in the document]
TheSpyder: This advice would normally be correct but is wrong in the specific scenario where we use the code in this PR.

The removeEventListener code has been reverted now so there's nothing to action, just wanted to make the note.

PR #10104, opened 2025-01-21
TINY-11177: Performance improvements to CI testing

TheSpyder:
Related Ticket: TINY-11177

Description of Changes:
* Upgrade to Bedrock 15, which shortens the time between tests and enforces better test habits
* Reduced some of our default timeouts that were designed for old browsers and very slow VMs
* Fixed a lot of tests that were not waiting for a promise to resolve correctly
* Fixed a few tests that did not clean up elements they added to the test page
* Added some empty binary assets to avoid uncacheable 404 pings to the test server (which is slow in remote testing)

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [x] Milestone set
* [x] Depend on bedrock 15 release once https://github.com/tinymce/bedrock/pull/145 has merged

## Release Notes

- **New Features**
	- Added new dialog measurement verification function in image plugin.
	- Enhanced bookmark assertion capabilities in TinyMCE.
	- Improved test infrastructure for various components.

- **Performance Improvements**
	- Reduced wait times in multiple test scenarios from 100ms to a dynamic wait based on user actions.
	- Updated chunk processing in test runner.
	- Optimized asynchronous test handling.

- **Dependency Updates**
	- Updated `@ephox/bedrock-client` to version 15.0.0.
	- Updated `@ephox/bedrock-server` to version 15.0.0.
	- Updated various package dependencies.

- **Test Enhancements**
	- Improved test case organization.
	- Added more robust async/await handling.
	- Enhanced event and interaction testing across multiple plugins.
	- Introduced cleanup processes to ensure proper DOM management after tests.
	- Improved error handling in `TinyLoader` tests.
	- Enhanced context menu interaction tests with dynamic wait adjustments.
	- Updated multiple tests to replace fixed wait times with dynamic user action waits.
	- Enhanced testing for toolbar button functionality and menu interactions.
	- Improved handling of asynchronous operations in various test cases.

- **Minor Adjustments**
	- Refined toolbar and menu interaction tests.
	- Updated import statements and code structure.
	- Improved error handling and logging in test utilities.
	- Adjusted test descriptions for clarity and specificity.


modules/mcagar/src/test/ts/browser/api/TinyLoaderErrorTest.ts
```diff
@@ -1,9 +1,25 @@
...
+      throw new Error('boo!');
+    },
+    { base_url: '/project/tinymce/js/tinymce' },
+    targetElement,
+    failure,
+    () => {
+      cleanupEditor?.remove();
+      Remove.remove(targetElement);
+      success();
+    });
```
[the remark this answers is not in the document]
TheSpyder: We do not clean up error cases, we want to see the failure to analyse it.

modules/tinymce/src/core/test/ts/browser/annotate/AnnotateBlocksTest.ts
```diff
@@ -615,6 +615,7 @@ describe('browser.tinymce.core.annotate.AnnotateBlocksTest', () => {
               selectionPath([ 0 ], 1, [], 3),
               { span: isRootBlock ? 2 : 3, block: isRootBlock ? 1 : 0 }
             );
+            // TODO: Selection doesn't work correctly if we don't wait 100ms here
```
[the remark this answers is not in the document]
TheSpyder: That's a good suggestion, but there are two problems
1. The path doesn't match (the selection is `[ 0 ]`, not `[ 0, 0 ]`, to match the `selectionPath()` call on line 615)
2. `TinyAssertions.assertSelection` is already checked, through the `expectedSelection` parameter of the `testApplyAnnotationOnSelection` function. The suggestion is a redundant check that will pass immediately (and thus triggers the failure that caused someone to add the 100ms wait).

modules/tinymce/src/core/test/ts/browser/NotificationManagerTest.ts
```diff
@@ -219,59 +219,60 @@ describe('browser.tinymce.core.NotificationManagerTest', () => {
...
+          add_unload_trigger: false,
+          disable_nodechange: true,
+          indent: false,
+          entities: 'raw',
+          base_url: '/project/tinymce/js/tinymce',
+          setup: (editor: Editor) => {
+            editor.on('BeforeOpenNotification', (event) => beforeOpenEvents.push(event));
+            editor.on('OpenNotification', (event) => openEvents.push(event));
+          }
+        }, []);
```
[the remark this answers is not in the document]
TheSpyder: Oh! I was so focused on the incorrect nesting of `context` inside `it` that I didn't even notice. Good pickup!

modules/agar/package.json
```diff
@@ -25,8 +25,8 @@
   "author": "Ephox Corporation DBA Tiny Technologies, Inc",
   "license": "GPL-2.0-or-later",
   "dependencies": {
-    "@ephox/bedrock-client": "11 || 12 || 13",
-    "@ephox/bedrock-common": "11 || 12 || 13",
+    "@ephox/bedrock-client": "15.0.0-alpha.10",
+    "@ephox/bedrock-common": "*",
```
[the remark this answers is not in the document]
TheSpyder: Will be fixed before merge, I added a checkbox in the description for it

modules/tinymce/src/core/test/ts/browser/annotate/AnnotateBlocksTest.ts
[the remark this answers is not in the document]
TheSpyder: I tried 17ms while developing this PR and it didn't help. After your prompting to find a proper fix I decided it was worth investigating. Turns out the thing we're waiting for here has a 30ms timer (a throttler in `modules/tinymce/src/core/main/ts/annotate/AnnotationChanges.ts`).

I tried adding a verification wait check similar to what you originally suggested, but the problem is that it wasn't consistently needing to wait. This turned out to be because the event has a de-duplication check so that it doesn't notify listeners for the same element twice. We share the editor instance between tests to speed them up, but these tests reset their annotation ID to zero on every loop. Due to this state leaking between tests the event system thought subsequent loops were duplicate events and only the first test had to wait.

I fixed this for good by adding an `after()` block that moves the selection out of the annotation, meaning every pass through the loop is consistent and has something to wait for.

shanmen-tiny (approved):

modules/tinymce/src/models/dom/test/ts/browser/table/DragResizeTest.ts
```diff
@@ -113,7 +116,10 @@ describe('browser.tinymce.models.dom.table.DragResizeTest', () => {
     Attribute.set(elm, 'data-mce-resize', 'false');
   };
 
-  afterEach(() => {
+  afterEach(async () => {
+    // Bedrock used to have a big wait between tests for reporting, that's now gone
+    // turns out the table code needs some time to settle in between tests
+    await Waiter.pWait(17);
```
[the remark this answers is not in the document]
TheSpyder: I had a quick look but it wasn't obvious, so I put the wait in.

The test suite still passes with this commented out, because we have the `retry` bedrock mechanism that reloads the page every time a test fails. Something about the state within the editor doesn't settle until a frame or two after each test finishes.

Gruntfile.js
```diff
@@ -53,34 +54,26 @@ const bedrockHeadless = (tests, browser, auto, opts) => {
...
         name: name,
         browser: browserName,
         testfiles: testFolders(tests, auto),
         bucket: bucket,
         buckets: buckets,
         chunk: chunk,
-        remote: remote,
-
-        // we have a few tests that don't play nicely when combined together in the monorepo
-        retries: 3,
```
spocke, line 83: So we getting rid of this one, I guess it makes sense since it's been an escape hatch also it has caused confusion when the test say passes on remote but not locally since locally it doesn't retry.

modules/tinymce/src/core/test/ts/browser/DragDropOverridesTest.ts
```diff
@@ -31,11 +31,11 @@ describe.skip('browser.tinymce.core.DragDropOverridesTest', () => {
     }, [], true);
 
     before(async function () {
-      this.timeout(10000);
+      this.timeout(7000);
```
spocke, line 34: Why the decrease here?

modules/tinymce/src/core/test/ts/browser/dom/ControlSelectionTest.ts
```diff
@@ -75,18 +75,24 @@ describe('browser.tinymce.core.dom.ControlSelectionTest', () => {
     const resizeHandle = await UiFinder.pWaitForVisible('Wait for resize handlers to show', editorBody, resizeSelector);
     const target = UiFinder.findIn(editorBody, targetSelector).getOrDie();
     Mouse.mouseDown(resizeHandle);
+    // a real user won't drag within a single frame, so we shouldn't either
+    await Waiter.pWait(17);
```
spocke, line 79: So all these 17ms is that to simulate a frame wait. Why not use requestAnimationFrame? Who knows what the framerate is on the remote test machines. Or maybe we should have a Waiter.pWaitForFrame that we can tweak that does this or requestAnimationFrame.

modules/tinymce/src/models/dom/test/ts/module/table/TableTestUtils.ts
```diff
@@ -132,12 +135,15 @@ const pDragResizeBar = async (editor: Editor, rowOrCol: 'row' | 'column', index:
   const resizeBar = await Waiter.pTryUntil('wait for resize bars',
     () => UiFinder.findIn(docElem, `div[data-${rowOrCol}='${index}']`).getOrDie()
   );
+  // a real user won't drag within a single frame, so we shouldn't either
   Mouse.mouseDown(resizeBar);
+  await Waiter.pWait(0);
```
spocke, line 140: Why then why are these not 17ms, 0 ms is not a frame either. But adding that would slow things down.

modules/tinymce/src/plugins/fullscreen/test/ts/webdriver/FullscreenTrapFocusTest.ts
```diff
@@ -36,12 +36,13 @@ describe('webdriver.tinymce.plugins.fullscreen.FullscreenTrapFocusTest', () => {
...
-    TinyContentActions.keystroke(editor, 121, { alt: true });
-    await FocusTools.pTryOnSelector('Assert toolbar is focused', SugarDocument.getDocument(), 'div[role=toolbar] .tox-tbtn');
-    await RealKeys.pSendKeysOn('div[role=toolbar] .tox-tbtn', [ RealKeys.text('enter') ]);
+  const pToggleFullscreen = async (nativeMode: boolean, fullscreen: boolean) => {
+    await RealMouse.pClickOn('button[data-mce-name="fullscreen"]');
     if (nativeMode) {
       await pIsFullscreen(fullscreen);
+      // Wait for 0.5 seconds to allow animations to complete
+      // The delay is arbitary and may need to be adjusted
+      await Waiter.pWait(300);
```
spocke, line 45: 300ms is not 0.5s so remove the 0.5 above and just keep the code value

modules/tinymce/src/plugins/media/test/ts/module/test/Utils.ts
```diff
@@ -165,7 +165,7 @@ const pTestEmbedContentFromUrl = async (editor: Editor, url: string, content: st
 const pSetFormItemNoEvent = pSetSourceInput;
 
 const pAssertEditorContent = (editor: Editor, expected: string): Promise<void> =>
-  Waiter.pTryUntil('Wait for editor value',
+  Waiter.pTryUntil('Waiting for editor value: ' + expected,
```
spocke, line 168:
```suggestion
  Waiter.pTryUntil(`Waiting for editor value: ${expected}`,
```

modules/tinymce/src/plugins/table/test/ts/browser/TableRemoveTrailingBrTest.ts
```diff
@@ -9,13 +9,16 @@ import * as TableTestUtils from '../module/test/TableTestUtils';
 describe('browser.tinymce.plugins.table.TableRemoveTrailingBrTest', () => {
   const baseSettings = {
     plugins: 'table code',
-    base_url: '/project/tinymce/js/tinymce'
+    base_url: '/project/tinymce/js/tinymce',
+    table_default_attributes: { },
+    table_default_styles: { width: '200px' },
```
spocke, line 14: What is going on here exactly. This doesn't seem to have anything to do with the test perf so we are not reverting something here are we?

modules/tinymce/src/themes/silver/test/ts/browser/editor/bespoke/SilverBespokeButtonsTest.ts
```diff
@@ -157,11 +157,11 @@ describe('browser.tinymce.themes.silver.editor.bespoke.SilverBespokeButtonsTest'
       TinySelections.setCursor(editor, [ 0, 0 ], 'Fi'.length);
       assertEvent(1, 'Verdana');
       await MenuUtils.pOpenMenu('FontSelect', 'Verdana');
-      assertEvent(2, 'Verdana');
+      assertEvent(1, 'Verdana');
```
spocke, line 160: Why did these need to change?

modules/tinymce/src/themes/silver/test/ts/browser/editor/menubar/EditorMenuItemContextTest.ts
```diff
@@ -618,7 +618,7 @@ describe('browser.tinymce.themes.silver.editor.toolbar.EditorMenuItemContextTest
           await scenario.pAssertMenuItemEnabled(editor, 't11');
           TinyUiActions.keystroke(editor, Keys.escape());
 
-          editor.setContent('<img src="https://picsum.photos/200/300"/>');
+          editor.setContent('<img src="data:image/png;base64, iVBORw0KGgoAAAANSUhEUgAAAAUAAAAFCAYAAACNbyblAAAAHElEQVQI12P4//8/w38GIAXDIBKE0DHxgljNBAAO9TXL0Y4OHwAAAABJRU5ErkJggg==" />');
```
spocke, line 621: Ouch remote url calls. Good catch. 😮

modules/tinymce/src/themes/silver/test/ts/module/StickyHeaderStep.ts
```diff
@@ -35,10 +35,8 @@ const testStickyHeader = (toolbarMode: ToolbarMode, toolbarLocation: ToolbarLoca
 
     PageScroll.bddSetup(hook.editor, 5000);
 
-    before(async () => {
-      // Need to wait for a fraction for some reason on safari,
-      // otherwise the initial scrolling doesn't work
-      await Waiter.pWait(100);
+    beforeEach(async () => {
+      await Waiter.pWait(17);
```
spocke, line 39: Yeah this should definitely be a requestAnimationFrame if we are waiting for UI state to change.

spocke (approved):

Gruntfile.js
```diff
@@ -53,34 +54,26 @@ const bedrockHeadless = (tests, browser, auto, opts) => {
...
         ...opts
       }
     }
   }
 };
 
-const bedrockBrowser = (tests, browserName, osName, bucket, buckets, chunk, remote, auto, opts) => {
+const bedrockBrowser = (tests, browserName, osName, bucket, buckets, chunk, auto, opts) => {
```
kimwoodfield, line 63: Unrelated to the actual change here but I wonder why we wrote this to accept both `bucket` and `buckets` as individual parameters. Couldn't we just take a single parameter `buckets` which could be a list of one or many buckets? I'm not aware of the implementation details of the object we return so I'm sure there is a reason but I was just curious.

modules/tinymce/src/core/test/ts/browser/file/ConversionsTest.ts
```diff
@@ -12,13 +12,14 @@ describe('browser.tinymce.core.file.ConversionsTest', () => {
...
   });
 
-  it('uriToBlob with invalid src', () => {
-    return Conversions.uriToBlob(invalidBlobUriSrc).then(() => {
+  it('uriToBlob with invalid src', async () => {
+    try {
+      await Conversions.uriToBlob(invalidBlobUriSrc);
       assert.fail('Conversion should fail.');
-    }, (error) => {
+    } catch (error: any) {
```
kimwoodfield, line 19: I'm probably missing something here but could you share why we went with `any` here if we are expecting `error` to include `message`? 
Could it potentially throw an type error `Cannot read properties of undefined` on L21 if it's not what we expect?

modules/tinymce/src/plugins/image/main/ts/core/Utils.ts
```diff
@@ -21,7 +21,7 @@ const getImageSize = (url: string): Promise<ImageDimensions> => new Promise((cal
   const img = document.createElement('img');
 
   const done = (dimensions: Promise<ImageDimensions>): void => {
-    img.onload = img.onerror = null;
+    // img.onload = img.onerror = null;
```
kimwoodfield, line 24: Are we planning on adding this back later or is it safe to be removed?

TheSpyder: Oh I meant to delete this not comment it out. This is from back when IE used to have memory leaks if DOM listeners were not removed from elements manually.

kimwoodfield (approved):

modules/tinymce/src/core/test/ts/browser/file/ConversionsTest.ts
TheSpyder: I have absolutely no idea. Converting from `.then()` callbacks to async/wait try/catch forced me to add a type annotation, and given how the assertion works I didn't see any reason to be more specific than `any`.

Gruntfile.js
TheSpyder: Buckets are designed to be run in parallel, which means we use a separate machine for each bucket and a dedicated bedrock instance. Bedrock needs to know the total number of buckets to divide the tests into, and which of those buckets this instance should use.

modules/tinymce/src/models/dom/test/ts/browser/table/DragResizeTest.ts
```diff
@@ -262,39 +295,17 @@ describe('browser.tinymce.models.dom.table.DragResizeTest', () => {
 
         it('TBA: Resize table bigger with handle, then resize row height smaller by dragging middle border', async () => {
           const editor = hook.editor();
+          editor.setContent('');
```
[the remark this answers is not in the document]
TheSpyder: Incorrect. Setting the content to empty has side effects such as resetting the cursor which can be valuable.

Gruntfile.js
kimwoodfield: Ah right, makes sense. Thanks for clarifying!

modules/tinymce/src/themes/silver/test/ts/browser/editor/bespoke/SilverBespokeButtonsTest.ts
TheSpyder: Although... it looks like things are still a _little_ inconsistent, because we just had a test run fail due to another section of the file now only receiving one event in a similar scenario.

PR #10121, opened 2025-01-29
TINY-11108: Height setting is more examined.

HAFRMO:
Related Ticket: TINY-11108

Description of Changes:
Attempt to convert strings into px number if possible.
Converting % seemed a bit too unreliable to me to be added here due to position-styling. This covers pt/em/px, but we can easily add any other units if needed.
Placed the code in the same file since I didn't know if we had any suitable library for it, but happy to move it if we have a better choice.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

## Release Notes

- **Bug Fixes**
	- Resolved an issue with editor height settings where pt/em values were incorrectly ignoring minimum and maximum height constraints.

- **Tests**
	- Added comprehensive test suite to validate editor size enforcement across different unit types (px, em, pt, percentage).

- **New Features**
	- Introduced a new function to enhance height value conversion, allowing for more flexible handling of various measurement units.

This update improves TinyMCE's height configuration reliability by ensuring consistent sizing behavior across multiple measurement units.

### DIFF
--- a/.changes/unreleased/tinymce-TINY-11108-2025-01-28.yaml
+++ b/.changes/unreleased/tinymce-TINY-11108-2025-01-28.yaml
@@ -1,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: Setting editor height to pt/em values would ignore min/max height settings.
+time: 2025-01-28T12:11:04.210024+01:00
+custom:
+    Issue: TINY-11108
--- a/.changes/unreleased/tinymce-TINY-11108-2025-01-28.yaml
+++ b/.changes/unreleased/tinymce-TINY-11108-2025-01-28.yaml
@@ -1,6 +1,6 @@
 project: tinymce
 kind: Fixed
-body: Setting editor height to pt/em values would ignore min/max height settings.
+body: Setting editor height to a `pt` or `em` value was ignoring min/max height settings.
 time: 2025-01-28T12:11:04.210024+01:00
 custom:
     Issue: TINY-11108
--- a/modules/tinymce/src/core/test/ts/browser/EditorSizeTest.ts
+++ b/modules/tinymce/src/core/test/ts/browser/EditorSizeTest.ts
@@ -1,0 +1,76 @@
+import { context, describe, it } from '@ephox/bedrock-client';
+import { Height } from '@ephox/sugar';
+import { TinyDom, TinyHooks } from '@ephox/wrap-mcagar';
+import { assert } from 'chai';
+
+import Editor from 'tinymce/core/api/Editor';
+
+describe('browser.tinymce.core.EditorSizeTest', () => {
+  context('Minimum Size Enforcement', () => {
+    context('Unitless wrapper', () => {
+      const hook = TinyHooks.bddSetup<Editor>({
+        selector: 'textarea',
+        height: 10,
+        min_height: 500,
+        base_url: '/project/tinymce/js/tinymce'
+      }, []);
+      it('TINY-11108: unitless', () => {
+        const editor = hook.editor();
+        assert.equal(Height.get(TinyDom.container(editor)), 500);
+      });
+    });
+
+    context('em Wrapper', () => {
+      const hook = TinyHooks.bddSetup<Editor>({
+        selector: 'textarea',
+        height: '10em',
+        min_height: 500,
+        base_url: '/project/tinymce/js/tinymce'
+      }, []);
+      it('TINY-11108: using em units', () => {
+        const editor = hook.editor();
+        assert.equal(Height.get(TinyDom.container(editor)), 500);
+      });
+    });
+
+    context('pt Wrapper', () => {
+      const hook = TinyHooks.bddSetup<Editor>({
+        selector: 'textarea',
+        height: '10pt',
+        min_height: 500,
+        base_url: '/project/tinymce/js/tinymce'
+      }, []);
+      it('TINY-11108: using pt units', () => {
+        const editor = hook.editor();
+        assert.equal(Height.get(TinyDom.container(editor)), 500);
+      });
+    });
+
+    context('px Wrapper', () => {
+      const hook = TinyHooks.bddSetup<Editor>({
+        selector: 'textarea',
+        height: '10px',
+        min_height: 500,
+        base_url: '/project/tinymce/js/tinymce'
+      }, []);
+      it('TINY-11108: using px units', () => {
+        const editor = hook.editor();
+        assert.equal(Height.get(TinyDom.container(editor)), 500);
+      });
+    });
+
+    context('percentage wrapper', () => {
+      const hook = TinyHooks.bddSetup<Editor>({
+        selector: 'textarea',
+        height: '10%',
+        min_height: 500,
+        base_url: '/project/tinymce/js/tinymce'
+      }, []);
+      it('TINY-11108: using % units', () => {
+        const editor = hook.editor();
+        // % not really supported, so we do this.
+        assert.isTrue(Height.get(TinyDom.container(editor)) < 200);
+      });
+    });
+  });
+});
--- a/modules/tinymce/src/core/test/ts/browser/EditorSizeTest.ts
+++ b/modules/tinymce/src/core/test/ts/browser/EditorSizeTest.ts
@@ -14,6 +14,7 @@ describe('browser.tinymce.core.EditorSizeTest', () => {
         min_height: 500,
         base_url: '/project/tinymce/js/tinymce'
       }, []);
+
       it('TINY-11108: unitless', () => {
         const editor = hook.editor();
         assert.equal(Height.get(TinyDom.container(editor)), 500);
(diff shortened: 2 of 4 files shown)


modules/tinymce/src/themes/silver/main/ts/ui/sizing/EditorSize.ts
```diff
@@ -5,17 +5,43 @@ import Editor from 'tinymce/core/api/Editor';
...
+    const splitValue = /^([0-9.]+)(pt|em|px)$/.exec(value);
+
+    if (splitValue) {
+      const type = splitValue[2];
+      const parsed = Number.parseFloat(splitValue[1]);
+
+      if (Number.isNaN(parsed) || parsed < 0) {
+        return Optional.none();
+      } else if (type === 'em') {
+        return Optional.from(parsed * Number.parseFloat(window.getComputedStyle(editor.targetElm).fontSize));
```
spocke, line 23: This is the only time editor is used. So you could lift this out to a module that take just an element so that it's generic maybe put it in the Utils module next to parseToInt. Maybe a SugarElement then you pass that in instead. The benefit of that is that it becomes easier to test in isolation and we could in theory use this for other convert calls in the future seems like a useful thing to have.

modules/tinymce/src/themes/silver/main/ts/ui/sizing/EditorSize.ts
```diff
@@ -5,17 +5,43 @@ import Editor from 'tinymce/core/api/Editor';
 import * as Options from '../../api/Options';
 import * as Utils from './Utils';
 
+const convertValueToPx = (editor: Editor, value: number | string): Optional<number> => {
+  if (typeof value === 'number') {
+    return Optional.from(value);
+  }
+
+  return Utils.parseToInt(value.trim()).orThunk(() => {
```
spocke, line 13: I'm not sure you need this part since you already check if it's a number and return that. Then you parse the value and the unit into a float. You could also move out the parse value/unit into a separate utils function. I'm surprised that we don't have this one anywhere.

modules/tinymce/src/core/test/ts/browser/EditorSizeTest.ts
```diff
@@ -0,0 +1,76 @@
...
+describe('browser.tinymce.core.EditorSizeTest', () => {
+  context('Minimum Size Enforcement', () => {
+    context('Unitless wrapper', () => {
+      const hook = TinyHooks.bddSetup<Editor>({
+        selector: 'textarea',
+        height: 10,
+        min_height: 500,
+        base_url: '/project/tinymce/js/tinymce'
+      }, []);
+      it('TINY-11108: unitless', () => {
```
spocke, line 17: I think it's easier to read if we separate the setup of the hook from the `it` call.

```suggestion
      }, []);

      it('TINY-11108: unitless', () => {
```

modules/tinymce/src/core/test/ts/browser/EditorSizeTest.ts
```diff
@@ -0,0 +1,76 @@
...
+    context('pt Wrapper', () => {
+      const hook = TinyHooks.bddSetup<Editor>({
+        selector: 'textarea',
+        height: '10pt',
+        min_height: 500,
+        base_url: '/project/tinymce/js/tinymce'
+      }, []);
+      it('TINY-11108: using pt units', () => {
+        const editor = hook.editor();
+        assert.equal(Height.get(TinyDom.container(editor)), 500);
```
spocke, line 45: This doesn't look right it looks like the 10pt value is just ignored and it clamps it to the min_height of 500px also. I think you might need to setup a dom structure then render the editor on that structure if you want to make sure that the em calculation is correct for the target.

- [ ] Make sure the tests actually test what is supposed to test

spocke (changes requested):

modules/tinymce/src/core/test/ts/browser/EditorSizeTest.ts
HAFRMO: Well, that is what we want? We want to set the height to 10pt, it is less than 500, so it's set to 500. That is the entire use of the min-value.

modules/tinymce/src/themes/silver/main/ts/ui/sizing/EditorSize.ts
HAFRMO: I agree it'd be good to move it elsewhere, I just wasn't sure where to ( FontUtils felt the most likely but it didn't fit, so I didn't want to put it there ). This seems like a decent solution.

modules/tinymce/src/themes/silver/main/ts/ui/sizing/EditorSize.ts
HAFRMO: I'm not sure what you mean? We've checked that it's not `10`, this checks for `"10"`, before we try to go for something like `"10em"`

.changes/unreleased/tinymce-TINY-11108-2025-01-28.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: Setting editor height to pt/em values would ignore min/max height settings.
```
TheSpyder, line 3:
```suggestion
body: Setting editor height to a `pt` or `em` value was ignoring min/max height settings.
```

modules/tinymce/src/themes/silver/main/ts/ui/sizing/EditorSize.ts
TheSpyder: > I'm surprised that we don't have this one anywhere

It's likely in the table code, or perhaps the font size code. I don't know exactly where either but I'm sure we have size parsing _somewhere_.

modules/tinymce/src/core/test/ts/browser/EditorSizeTest.ts
spocke: 🤦‍♂️  Missed the `Minimum Size Enforcement` context. You are right.

modules/tinymce/src/themes/silver/main/ts/ui/sizing/EditorSize.ts
spocke: Its now doing this:
1. Check if it's a number return that so floats are allowed
2. Convert the number to a string in `parseToInt` and return that already done by 1 but floors the value
3. Parse if it's px value including fractions also in `parseToInt`
4. Then parse the `pt` `em` `px` again `px` is already handled by step 3.

So it's now allowing you to specify sub pixel values `height: 500.34` is a valid size before that would be parsed as a int so `500`. So at least put a `Math.floor` on step 1 unless we want this new behaviour.

I wonder if we instead should simply do all of it in one parse function and if there is a pixel unit parser function somewhere that we can use. Then we can do the calculations in another function that uses that parser.

It also doesn't handle NaN that is something that the original `parseToInt` never did since it does a parseInt('' + value) that will return NaN if value is not a valid number. So one could use `Strings.toInt` to harden that or check for NaN in the float that we get from `Number.parseFloat`.

I guess we could do that in some cleanup task.

spocke (approved):

modules/tinymce/src/themes/silver/main/ts/ui/sizing/EditorSize.ts
HAFRMO: @TheSpyder You may be thinking about the stuff that we do in `src/core/main/ts/fmt/FontInfo`? It converts some stuff related to font, but not what needed here.

Most of the time when we deal with sizes we seem to do it with pre-calculated by the browser ( the resizing code does that, for example, it doesn't care what the set value is, just what it's calculated to become by the browser ) but we can't do that here since the element size is not set yet.

@spocke Did not realise that parseToInt already did have a detail to handle ( well, ignore ) px-values. But we can probably skip the parseToInt-call, particularly since the new code handles NaN as you mentioned.

As for the issue with allowing subpixel, is that not already something we should support as we start properly allowing em/pt values? Since rounding that off could make them immediately wrong anyway? Or are we more interested in keeping it to firmly to int? ( and we don't seem to round min/max either? )

ltrouton: @HAFRMO Did you have a look at https://github.com/tinymce/tinymce/blob/main/modules/sugar/src/main/ts/ephox/sugar/api/view/Dimension.ts? Would that be useful at all?

lorenzo-pomili (approved):

.changes/unreleased/tinymce-TINY-11108-2025-01-28.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Fixed
```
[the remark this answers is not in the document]
HAFRMO: No, this is a fix over expected behavior, it's behavior which was expected.

PR #10132, opened 2025-02-03
TINY-11786: Improve scroll behavior for better element positioning

ArvinJ-H:
Related Ticket: 
TINY-11786

Description of Changes:
- Improved `scrollintoview` api to by default align to the top 25% where possible
- `topMargin` is to isolate the change to a11y checker only, see https://ephocks.atlassian.net/browse/EPIC-182?focusedCommentId=144527

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

## Summary by CodeRabbit

- **New Features**
  - Enhanced scrolling behavior now applies a fixed margin when moving content into view, ensuring elements do not sit flush against the viewport edges for improved readability.

- **Style**
  - Updated visual cues for accessibility indicators now feature refined shadow effects, offering clearer notifications of accessibility issues.

- **Tests**
  - Adjusted expected scroll positions in various test cases to align with the new scrolling behavior, including additions for margin scenarios.


modules/tinymce/src/core/main/ts/dom/ScrollIntoView.ts
```diff
@@ -169,22 +202,22 @@ const intoFrame = (editor: Editor, doc: SugarElement<Document>, scrollTop: numbe
...
+const scrollElementIntoView = (editor: Editor, element: HTMLElement, alignToTop?: boolean, topMargin?: boolean): void => {
   const scroller = editor.inline ? elementIntoWindow : elementIntoFrame;
-  scroller(editor, element, alignToTop);
+  scroller(editor, element, alignToTop, topMargin);
 };
 
 // This method is made to deal with the user pressing enter, it is not useful
 // if we want for example scroll in content after a paste event.
-const scrollRangeIntoView = (editor: Editor, rng: Range, alignToTop?: boolean): void => {
+const scrollRangeIntoView = (editor: Editor, rng: Range, alignToTop?: boolean, topMargin?: boolean): void => {
```
tiny-ben-tran, line 218: Having the new argument type as `boolean` reduces the API’s flexibility. Why not use `topMargin: number` instead, allowing users to specify an exact offset from the top corner?

tiny-ben-tran (dismissed):
This change potentially affects many plugins while the intention is to improve the behaviour for one.  Please consider reducing the amount of code change here as well.  For instance: the targeted plugin doesn't use `scrollRangeIntoView` therefore the change there is unnecessary.

modules/tinymce/src/core/main/ts/dom/ScrollIntoView.ts
ArvinJ-H: we don't want to let the user to specify an exact offset, the offset change branch off from scrolling behavior improvement for a11y checker.

Browser default `scrollintoView` api only supports `start`, `end`, `middle`, while it is not implemented here due to a few potential conflicts with other setting, it is something to be looked at if we still want to set the scrolling to 25%

ArvinJ-H:
> This change potentially affects many plugins while the intention is to improve the behaviour for one. Please consider reducing the amount of code change here as well. For instance: the targeted plugin doesn't use `scrollRangeIntoView` therefore the change there is unnecessary.

That's why I isolated the change with the `topMargin` argument? So this change will only apply to a11y checker

modules/tinymce/src/core/main/ts/dom/ScrollIntoView.ts
kimwoodfield: I'm not so sure about this approach either... The native `scrollIntoView` API allows an object to be provided (`scrollIntoViewOptions`) as an argument where I think something like this might be more appropriately placed if we do actually need it. Otherwise, we run the risk of creating a trend where we provide future configuration options as separate arguments, bloating interactions with this API. 

I might be wrong so I would be curious to hear more thoughts and opinions on this one too!

ltrouton: I agree with Kim that we should look at adding an optional object argument for additional configuration. Just to note, we can't remove the current `alignToTop` argument as that would be a breaking change

spocke: I wonder if we could make the alignToTop either a boolean or a config object because it would be a bit weird to have an optional then a config. We could then deprecate the boolean option with a warning message or something then remove it in v8. I would guess we want more things here like bottom margin, scolling mode so an config object seems like a better approach.

modules/tinymce/src/core/main/ts/dom/ScrollIntoView.ts
```diff
@@ -117,47 +117,79 @@ const preserveWith = (editor: Editor, f: (startElement: SugarElement<Node>, endE
...
+  const viewWidth = Traverse.defaultView(doc ?? SugarElement.fromDom(editor.getDoc())).dom.innerWidth;
+  const viewportLeft = -editor.getBody().getBoundingClientRect().left;
   if (alignToTop) {
-    Scroll.to(pos.left, pos.top, doc);
+    if (topMargin) {
+      const targetOffset = viewHeight * 0.25;
+      const y = pos.top - targetOffset;
+      // If element is off screen to the left, scroll to slightly right of its position
+      if (pos.left < viewportLeft) {
+        Scroll.to(pos.left - 20, y, doc);
```
spocke, line 130: So we are also altering the left position so the `topMargin` does special things to left position. I would expect that `topMargin` would just alter the top not other properties of the rect. And 20 pixels seems very arbitrary should this be a config in? This could be a config so `scrollIntoView(elm, { leftMargin: '20px' })` or maybe `scrollIntoView(elm, { leftMargin: 20 })`

modules/tinymce/src/core/main/ts/dom/ScrollIntoView.ts
```diff
@@ -117,47 +117,79 @@ const preserveWith = (editor: Editor, f: (startElement: SugarElement<Node>, endE
...
 
-const scrollToMarker = (editor: Editor, marker: MarkerInfo, viewHeight: number, alignToTop: boolean, doc?: SugarElement<Document>) => {
+const scrollToMarker = (editor: Editor, marker: MarkerInfo, viewHeight: number, alignToTop: boolean, doc?: SugarElement<Document>, topMargin?: boolean) => {
   const pos = marker.pos;
+  const viewWidth = Traverse.defaultView(doc ?? SugarElement.fromDom(editor.getDoc())).dom.innerWidth;
+  const viewportLeft = -editor.getBody().getBoundingClientRect().left;
   if (alignToTop) {
-    Scroll.to(pos.left, pos.top, doc);
+    if (topMargin) {
+      const targetOffset = viewHeight * 0.25;
```
spocke, line 126: This value should be configurable either to a fixed value or percentage value. You could have `scrollIntoView(elm, { topMargin: '25%' })` then allow fixed values either with `px` or just regular numbers not sure we need that now but could be added later.

ArvinJ-H: The implementation has changed to a fixed value

modules/tinymce/src/core/main/ts/dom/ScrollIntoView.ts
ArvinJ-H: for this ticket at least we no longer modify the left or right scrolling.

We don't have any accurate data on how likely the editor will scroll horizontally, but this seems to be an issue we can put into the back log?

modules/oxide/src/less/theme/content/accessibility-checker/accessibility-checker.less
```diff
@@ -4,6 +4,10 @@
 @accessibility-overlay-opacity-90: 90%;
 @accessibility-overlay-opacity-60: 60%;
 
+@accessibility-box-shadow-color: #222f3e80;
```
TheSpyder, line 7: Should this be configurable for dark mode?

TheSpyder (approved):
Please update the PR description with the new requirements you're now implementing

modules/oxide/src/less/theme/content/accessibility-checker/accessibility-checker.less
ArvinJ-H: in dark mode this box shadow will also dim the rest of the content and draw attention to the focused element, there is perhaps some improvement we can do on this, I'll discuss with product.

.changes/unreleased/tinymce-TINY-11786-2025-03-03.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Improved
+body: Improved scrolling behavior by adding margins when scrolling elements into view, ensuring content isn't at the very edge of the viewport.
```
kimwoodfield, line 3:
```suggestion
kind: Improved
body: Scrolling behavior by adding margins when scrolling elements into view, ensuring content isn't at the very edge of the viewport.
```

modules/tinymce/src/core/main/ts/dom/ScrollIntoView.ts
```diff
@@ -119,12 +119,17 @@ const preserveWith = (editor: Editor, f: (startElement: SugarElement<Node>, endE
 
 const scrollToMarker = (editor: Editor, marker: MarkerInfo, viewHeight: number, alignToTop: boolean, doc?: SugarElement<Document>) => {
   const pos = marker.pos;
+  // Use a fixed scroll margin of 30px
+  const scrollMargin = 30;
```
kimwoodfield, line 123: Just another nitpick but I think the comment would be more helpful if it explained why we chose `30`

kimwoodfield (approved):

modules/oxide/src/less/theme/content/accessibility-checker/accessibility-checker.less
```diff
@@ -38,14 +42,24 @@
 [data-ephox-foam-a11y-violation][data-ephox-foam-a11y-current-violation="info"] {
   outline-color: @accessibility-issue-info-selection-color;
   background-color: @accessibility-issue-info-selection-background-color;
+  box-shadow: 0 0 0 @accessibility-box-shadow-width @accessibility-box-shadow-color;
+  z-index: @accessibility-box-shadow-z-index;
```
noxuhax, line 46: Wouldn’t it make sense to apply these two styles to the `[data-ephox-foam-a11y-violation][data-ephox-foam-a11y-current-violation]` above (L25) to avoid the unnecessary repetition?

.changes/unreleased/tinymce-TINY-11786-2025-03-03.yaml
TheSpyder: But the word `improved` is also implied at the beginning of the sentence because it's under an "improved" heading.

```suggestion
body: `editor.selection.scrollIntoView()` now pads the target scroll area with a small margin, ensuring content doesn't sit at the very edge of the viewport.
```

modules/oxide/src/less/theme/content/accessibility-checker/accessibility-checker.less
ArvinJ-H: dark mode in general is looking good, there are other dark mode specific issues that have some effects on other plugins needs to addressed.

modules/oxide/src/less/theme/content/accessibility-checker/accessibility-checker.less
ArvinJ-H: no, it will then create x amount of box shadow on the same plain, dragging down performance and block the highlight content

noxuhax: I was under the impression that there is only one `data-ephox-foam-a11y-current-violation` at the time so I don’t see the reason why there will be multiple shadows rendered. But anyways, that’s a pretty small nitpick so either way is good.

spocke (approved):

ltrouton (approved):

PR #10135, opened 2025-02-03
TINY-11768: Prevent duplicate title element

HAFRMO:
Related Ticket: TINY-11768

Description of Changes:
Prevents the titles from showing. I'd honestly say we should remove the functionality entirely for the groups since I can't see it being very useful ( or at least change it to some aria-thing ), but for this I've just removed it from adding the titles earlier on to keep the change small.

Pre-checks:
* [x] Changelog entry added
* ~~[ ] Tests have been added (if applicable)~~

Review:
* [x] Milestone set

## Summary by CodeRabbit

- **Bug Fixes**
  - Resolved an issue with toolbar groups displaying overlapping tooltips, ensuring a cleaner and more intuitive toolbar experience.
- **Accessibility Improvements**
  - Replaced `title` attributes with `aria-label` attributes for toolbar groups, enhancing accessibility for screen readers.

### DIFF
--- a/.changes/unreleased/tinymce-TINY-11768-2025-02-03.yaml
+++ b/.changes/unreleased/tinymce-TINY-11768-2025-02-03.yaml
@@ -1,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: Toolbar groups no longer display a title to prevent overlapping tooltips
+time: 2025-02-03T15:06:52.667595+01:00
+custom:
+    Issue: TINY-11768
--- a/.changes/unreleased/tinymce-TINY-11768-2025-02-03.yaml
+++ b/.changes/unreleased/tinymce-TINY-11768-2025-02-03.yaml
@@ -1,6 +1,6 @@
 project: tinymce
 kind: Fixed
-body: Toolbar groups no longer display a title to prevent overlapping tooltips
+body: Toolbar groups had both a `title` attribute and a custom tooltip, causing overlapping tooltips
 time: 2025-02-03T15:06:52.667595+01:00
 custom:
     Issue: TINY-11768
--- a/modules/tinymce/src/themes/silver/main/ts/ui/toolbar/Integration.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/ui/toolbar/Integration.ts
@@ -199,7 +199,7 @@ const identifyButtons = (editor: Editor, toolbarConfig: RenderToolbarConfig, bac
         lookupButton(editor, toolbarConfig.buttons, toolbarItem, toolbarConfig.allowToolbarGroups, backstage, prefixes).toArray();
     });
     return {
-      title: Optional.from(editor.translate(group.name)),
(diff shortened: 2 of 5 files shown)


spocke:
So what part of the options here no longer work after this change. Does labels for groups still work? https://www.tiny.cloud/docs/tinymce/latest/toolbar-configuration-options/#toolbar_groups
Is label being rendered as a aria label?
Is this a breaking change if it is then we should re-think this as we shouldn't remove functionality in a minor.

HAFRMO:
> So what part of the options here no longer work after this change. Does labels for groups still work? https://www.tiny.cloud/docs/tinymce/latest/toolbar-configuration-options/#toolbar_groups Is label being rendered as a aria label? Is this a breaking change if it is then we should re-think this as we shouldn't remove functionality in a minor.

I just removed the title from the groups as I tried to keep the code-change as minimal as possible. An alternative would be to keep it, but change it to an aria-label ( making the label and title here functionally the same ). Or some other aria-value, if we find that more suitable.

.changes/unreleased/tinymce-TINY-11768-2025-02-03.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: Toolbar groups no longer display a title to prevent overlapping tooltips
```
TheSpyder, line 3:
```suggestion
body: Toolbar groups had both a `title` attribute and a custom tooltip, causing overlapping tooltips
```

TheSpyder (approved):

modules/tinymce/src/themes/silver/main/ts/ui/toolbar/Integration.ts
```diff
@@ -199,7 +199,7 @@ const identifyButtons = (editor: Editor, toolbarConfig: RenderToolbarConfig, bac
         lookupButton(editor, toolbarConfig.buttons, toolbarItem, toolbarConfig.allowToolbarGroups, backstage, prefixes).toArray();
     });
     return {
-      title: Optional.from(editor.translate(group.name)),
+      title: Optional.none(),
```
spocke, line 202: This needs to be rendered as something maybe a `aria-label` since with this change the `name` has no use and the screen reader will no longer read out the toolbar groups at least not VoiceOver. Before this change it would say `Entering formatting` and `Leaving history` etc when you tab into groups. With this change it says nothing. I tested adding it as a `aria-label` and at least VoiceOver will speak that. So I suggest we make it a `aria-label` then get QA to see if that works on Jaws, other browsers etc.

- [ ] Change `name` to `aria-label`

spocke (approved):

modules/tinymce/src/themes/silver/main/ts/ui/toolbar/Integration.ts
HAFRMO: Alright.

Issue #10143, opened 2025-02-08
SelectedTextLinkTest behaves unrealistically (TinyComments)


Philipp91:
## 📝 Provide detailed reproduction steps (if any)

First add this test new test case to `modules/tinymce/src/plugins/link/test/ts/browser/SelectedTextLinkTest.ts`:

```
  it('TBA: inserting new link at caret should move cursor behind the new link', async () => {
    const editor = hook.editor();
    editor.setContent('<p></p>');
    await pOpenDialog(editor);
    FocusTools.setActiveValue(doc, 'http://something');
    await TestLinkUi.pClickSave(editor);
    TinyAssertions.assertCursor(editor, [ 0 ], 1); // <p><a>http://something</a>|</p>
  });
```

(Note: I want to add such a test case in #9998.)

1. Run it in a browser.
2. Put a break point on the `FocusTools.setActiveValue(doc, 'http://something');` line.
3. Rerun the test so that the breakpoint hits.
4. Observe how the dialog state is an empty URL field, but the "Text to display" field contains `<!-- TinyComments -->`.
5. Run the line in the debugger (stepping to the next line).
6. Observe how the URL field is now filled in with the text data `http://something`, but the "Text to display" field still contains `<!-- TinyComments -->`.

### ✔️ Expected result

If you manually open https://fiddle.tiny.cloud/, press Ctrl+K and paste a URL, the "Text to display" field is initially empty and then contains a copy of whatever you typed/pasted into the URL field. This results in the link text being the URL in the end, which is the intended behavior.

### ❌ Actual result

The test ends up creating a hyperlink with no text on it. In fact, throughout the execution of this entire test file, the TinyMCE text box in the browser remains visually empty.

### ❓ Possible solution

The `TinyComments` seems to be coming from [here](https://github.com/tinymce/tinymce/blame/8f507e8c19a47b9843263ced998101edc93f61b4/modules/tinymce/src/plugins/link/test/ts/browser/SelectedTextLinkTest.ts#L16-L21). When that was added in https://github.com/tinymce/tinymce/commit/54e535ac32ab3cd091f80e9d4129257439a751ad, the whole file looked very different still. I suspect that this is emulating something that [no longer exists](https://github.com/search?q=repo%3Atinymce%2Ftinymce%20tinycomments&type=code).

Removing it doesn't make the test fail, so I'll send a PR.

## 📃 Other details
- Browser: Chrome
- OS: Kubuntu

PR #5854, opened 2020-07-05
TINY-5205: Fixed the `link` dialog not showing the "Text to display" field in some valid cases

lnewson:
Related Ticket: TINY-5205

Description of Changes:
* The way the "Text to display" field was being determined was rather hacky, as it just used a regex on the selected content. This caused problems when coupled with Comments for example, as Comments would add a HTML comment that would cause this to no longer match. This updates it to actually walk the dom and determine when it shouldn't be shown based on the elements found.
* This also fixes an inconsistency with how it handled "inline text elements" inside an anchor, as if you used a collapsed selection for something like `<a href="URL"><strong>word</strong> more</a>`, then the "Text to display" would be shown. However, if you did a range selection then the field would disappear.

Pre-checks:
* [x] Changelog entry added
* [x] Tests have been added (if applicable)
* [x] ~Branch prefixed with `feature/` for new features (if applicable)~
* [x] License headers added on new files (if applicable)

Review:
* [x] Milestone set
* [x] Review comments resolved

GitHub issues (if applicable):

PR #10144, opened 2025-02-08
Remove the TinyComments hack from SelectedTextLinkTest

Philipp91:
Description of Changes:

It's unclear to me what this comment was meant for, but it looks like it's supposed to trick the SUT into thinking that the TinyComments plugin is also active, so that the SUT would change its behavior accordingly. But the SUT does not check for the presence of TinyComments (anymore?), so this seems redundant. Additionally it's undesirable, because this HTML comment accidentally becomes the text of inserted links (rather than the URL being copied into the link text field).

Pre-checks:
* ~~Changelog entry added~~ (doesn't seem worth it when only changing test code)
* ~~[ ] Branch prefixed with `feature/`, `hotfix/` or `spike/`~~ (seems to match neither or those)

Review:
* [x] Milestone set

GitHub issues (if applicable): #10143

## Summary by CodeRabbit

- Tests
	- Revised internal testing routines for text selection and link manipulation to ensure more accurate behavior validations.

### DIFF
--- a/modules/tinymce/src/plugins/link/test/ts/browser/SelectedTextLinkTest.ts
+++ b/modules/tinymce/src/plugins/link/test/ts/browser/SelectedTextLinkTest.ts
@@ -12,14 +12,6 @@ describe('browser.tinymce.plugins.link.SelectedTextLinkTest', () => {
   const hook = TinyHooks.bddSetupLight<Editor>({
     plugins: 'link',
     toolbar: '',
-    setup: (editor: Editor) => {
-      // Simulate comments being enabled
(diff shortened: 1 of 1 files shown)


TheSpyder:
It was added in #5854, and... at best it should've been a single test, not used for all tests. It's not clear exactly why it was added and the developer who submitted it has left the company.

I think if this is important enough to keep we'd add it to the private TinyComments tests, so this is fine.

TheSpyder:
@coderabbitai ignore

TheSpyder (approved):

ltrouton (approved):

PR #10145, opened 2025-02-08
Fire input event when a unit test sets the value of an input field

Philipp91:
Description of Changes:

The link plugin's dialog [copies the URL to the text/title field](https://github.com/tinymce/tinymce/blob/8f507e8c19a47b9843263ced998101edc93f61b4/modules/tinymce/src/plugins/link/main/ts/ui/DialogChanges.ts#L63) when those haven't been manually filled by the user. This happens whenever the `change` event on the URL field is fired. So unit tests need to ensure that happens in order to be realistic.

This commit adds a parameter to `FocusTools.setActiveValue()`, which has no effect by default but allows the respective test cases to trigger the event. All matches of `setActiveValue.*http` under `modules/tinymce/src/plugins/link/test/ts/browser` are then updated to pass the new argument.

Pre-checks:

These don't seem applicable, since I'm only adjusting tests to match real browser behavior.

To test this, I ran `yarn test-one modules/tinymce/src/plugins/link/test/ts/browser/*` and it still succeeds.

* [x] ~Changelog entry added~

Review:
* [x] Milestone set

## Summary by CodeRabbit

- **Refactor**
	- Enhanced focus handling by updating internal methods to accept additional event context.
- **Tests**
	- Updated multiple test cases for link interactions to validate the new focus behavior.

These updates improve the internal consistency of event handling for link operations without altering the visible user experience.


Philipp91:
This PR is slightly incomplete, due to #10147 the corresponding change in `TestLinkUi` currently breaks the test (even though it shouldn't).

TheSpyder:
@coderabbitai ignore

TheSpyder:
ooh good pickup!

Looking at the places `FocusTools.setActiveValue` is used, [many of them manually trigger](https://github.com/search?q=repo%3Atinymce%2Ftinymce%20FocusTools.setActiveValue&type=code) `input`. I think that would work just as well as `change` for the tests you've edited?

In that case I think the underlying method can have a default event name of `input` (perhaps as a defaulted parameter so it can be changed if necessary) and not only avoid all these extra arguments, the manual triggers can be deleted.

TheSpyder:
I fiddled with this change myself and now I understand what you're talking about with `editor.options.set()`. That's why the PR is incomplete.

Tests that fiddle with the parser settings should just make a new editor for such changes. I'll comment more in #10147, and make that change in this PR to fix the tests that started failing.

TheSpyder (approved):

TheSpyder:
@ltrouton this should be good to go once you've had a chance to look at it

Philipp91:
Cool, thank you for taking this over and figuring this all out! I guess I was still a couple steps away from a proper solution and due to my incomplete development environment and unfamiliarity, my progress was unreasonably slow. So I appreciate the help!

PR #10161, opened 2025-02-13
INT-3226: Make pasting of plain text synchronous

hamza0867:
Related Ticket: INT-3226

Description of Changes:
* Removed delayed execution of paste handling when using pastebin

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

GitHub issues:

## Summary by CodeRabbit

- **Bug Fixes**
  - Enhanced paste handling to prevent unintended content insertion when pasting non-supported formats.
  - Improved functionality for pasting plain text to execute without delay, aligning it with HTML pasting performance.

- **Refactor**
  - Simplified paste processing by removing delays, ensuring immediate and accurate content retrieval.
  - Streamlined event assertion logic by removing browser-specific conditions in paste tests.

### DIFF
--- a/.changes/unreleased/tinymce-INT-3226-2025-02-15.yaml
+++ b/.changes/unreleased/tinymce-INT-3226-2025-02-15.yaml
@@ -1,0 +1,6 @@
+project: tinymce
+kind: Improved
+body: Pasting plain text now executes without delay, just like pasting html.
+time: 2025-02-15T08:59:57.575722+01:00
+custom:
+  Issue: INT-3226
--- a/modules/tinymce/src/core/test/ts/webdriver/paste/CopyAndPasteTest.ts
+++ b/modules/tinymce/src/core/test/ts/webdriver/paste/CopyAndPasteTest.ts
@@ -15,7 +15,6 @@ import * as PasteEventUtils from '../../module/test/PasteEventUtils';
 describe('webdriver.tinymce.core.paste.CopyAndPasteTest', () => {
   const platform = PlatformDetection.detect();
   const os = platform.os;
-  const browser = platform.browser;
   const isChromeHeadless = () => navigator.userAgent.includes('HeadlessChrome');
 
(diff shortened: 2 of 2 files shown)


hamza0867:
* **#10161** <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10161?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10161?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

modules/tinymce/src/core/main/ts/paste/Clipboard.ts
```diff
@@ -269,15 +268,14 @@ const registerEventHandlers = (editor: Editor, pasteBin: PasteBin, pasteFormat:
       e.preventDefault();
       insertClipboardContent(editor, clipboardContent, clipboardContent['text/plain'], plainTextMode, true);
     } else {
+      e.preventDefault();
```
spocke, line 271: I'm surprised thst this code path is executing at all this is a fallback if the clipboard ap exists or doesn't have html or plaintext mime types.
Also preventing the paste event will make this code not execute at all since why it has a delay is that we late the paste operation fill the paste bin then we extract what it pastes from the bin if you prevent that default action from happening you should have a empty bin always so basically this code with this change should not work at all. Are we sure that this works, under what condition is this code path executed?
Also powerpaste is async on regular html so does this bug happen on that, so not sure this is the solution we are looking for.

TheSpyder: The comment on the next line explains what's going on (`We can't extract...`) and why this change can never work. The comment is as clear as it can be without doing a deep-dive into browser events; this is just one of those things that has to be learnt when it comes up.

The original ticket included a video that should've made it clear this was the wrong place to tackle the problem.

TheSpyder (changes requested):

PR #10192, opened 2025-02-24
TINY-11848: add icons for image enhance features

lorenzo-pomili:
Related Ticket: TINY-11848

Description of Changes:
add icons for image enhance features

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [x] Milestone set
* [x] Docs ticket created (DOC-3146)


modules/oxide-icons-default/src/svg/uploadcare/auto-image-enhancement.svg
spocke, line 1: The directory structure in the icons directory is a mess it should all be flattened but since this is adding to that uploadcare directory lets keep doing it the wrong way then have a cleanup ticket later to flatten the icons.

The idea is that the icons are not really tied to a specific plugin they are just named this that happens to be used by a plugin. For example the `close` icons it's used by multiple things so could in theory these icons be.

spocke (approved):

modules/oxide-icons-default/src/svg/uploadcare/auto-image-enhancement.svg
kimwoodfield: @spocke Have we considered storing these assets in a CDN or a service hosted outside of the codebase?

spocke: @kimwoodfield We need to be able to ship the plugin to on-prem installations so when the product is installed locally we can't make any assumptions that we have access to the internet. It could be blocked by CSP headers, internal firewalls etc etc. Uploadcare is a bit special since it needs internet access to upload the files but we just add the icons here like any other icons.

kimwoodfield: @spocke So we still want to be able to show those images to support those using the editor while offline. That makes sense. Thanks!

kimwoodfield (approved):

PR #10194, opened 2025-02-25
TINY-11878: Downgrade `chalk` version

ArvinJ-H:
Related Ticket: 
TINY-11878

Description of Changes:
- Recent dependency upgrade cased an import error, downgrade should fix it

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/oxide/package.json
+++ b/modules/oxide/package.json
@@ -9,7 +9,7 @@
   },
   "scripts": {
     "test": "echo 'No tests'",
-    "build": "grunt build --force",
+    "build": "grunt build",
     "start": "grunt start",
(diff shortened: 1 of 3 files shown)


spocke (approved):

TheSpyder (approved):
I've been wondering when this would finally be discovered.

As a part of this change please remove the `--force` parameter from the `build` target in oxide. Our build setup has been suppressing this error for a while, but it would also suppress valid errors. I missed that we added it in the old gulp -> grunt conversion, it's not acceptable practice.

- [x] update oxide's build target

PR #10195, opened 2025-02-25
Feature/TINY-11756

HAFRMO:
Related Ticket: TINY-11756

Description of Changes:
Added new options for the sanitizer

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/.changes/unreleased/tinymce-TINY-11756-2025-02-25.yaml
+++ b/.changes/unreleased/tinymce-TINY-11756-2025-02-25.yaml
@@ -1,0 +1,6 @@
+project: tinymce
+kind: Added
+body: Added `allow_extra_mathml_attributes` and `allow_extra_mathml_elements` options for the editor.
+time: 2025-02-25T12:23:54.372384+01:00
+custom:
+    Issue: TINY-11756
--- a/modules/tinymce/src/core/main/ts/api/OptionTypes.ts
+++ b/modules/tinymce/src/core/main/ts/api/OptionTypes.ts
@@ -50,6 +50,8 @@ interface BaseEditorOptions {
   add_form_submit_trigger?: boolean;
   add_unload_trigger?: boolean;
   allow_conditional_comments?: boolean;
+  allow_extra_mathml_attributes?: string[];
+  allow_extra_mathml_elements?: string[];
   allow_html_data_urls?: boolean;
   allow_html_in_named_anchor?: boolean;
   allow_script_urls?: boolean;
--- a/modules/tinymce/src/core/main/ts/api/Options.ts
+++ b/modules/tinymce/src/core/main/ts/api/Options.ts
@@ -324,6 +324,32 @@ const register = (editor: Editor): void => {
     default: []
   });
 
+  registerOption('allow_extra_mathml_attributes', {
+    processor: (value) => {
+      const valid = Type.isArrayOf(value, Type.isString);
+
+      if (valid) {
+        return { value, valid };
+      } else {
+        return { valid: false, message: 'Must be an array of strings.' };
+      }
+    },
+    default: []
+  });
+
+  registerOption('allow_extra_mathml_elements', {
+    processor: (value) => {
+      const valid = Type.isArrayOf(value, Type.isString);
+
+      if (valid) {
+        return { value, valid };
+      } else {
+        return { valid: false, message: 'Must be an array of strings.' };
+      }
+    },
+    default: []
+  });
+
   registerOption('inline_boundaries', {
     processor: 'boolean',
     default: true
@@ -1009,6 +1035,8 @@ const shouldConvertUnsafeEmbeds = option('convert_unsafe_embeds');
 const getLicenseKey = option('license_key');
 const getApiKey = option('api_key');
 const isDisabled = option('disabled');
+const allowExtraMathmlAttributes = option('allow_extra_mathml_attributes');
+const allowExtraMathmlElements = option('allow_extra_mathml_elements');
 
 export {
   register,
@@ -1016,6 +1044,8 @@ export {
   getIframeAttrs,
   getDocType,
   getDocumentBaseUrl,
+  allowExtraMathmlAttributes,
+  allowExtraMathmlElements,
   getBodyId,
   getBodyClass,
   getContentSecurityPolicy,
--- a/modules/tinymce/src/core/main/ts/html/Sanitization.ts
+++ b/modules/tinymce/src/core/main/ts/html/Sanitization.ts
@@ -198,7 +198,7 @@ const getPurifyConfig = (settings: DomParserSettings, mimeType: MimeType): Confi
   // Allow any URI when allowing script urls
   if (settings.allow_script_urls) {
     config.ALLOWED_URI_REGEXP = /.*/;
-  // Allow anything except javascript (or similar) URIs if all html data urls are allowed
+    // Allow anything except javascript (or similar) URIs if all html data urls are allowed
   } else if (settings.allow_html_data_urls) {
     config.ALLOWED_URI_REGEXP = /^(?!(\w+script|mhtml):)/i;
   }
@@ -244,16 +244,33 @@ const sanitizeMathmlElement = (node: Element, settings: DomParserSettings) => {
     // passed an element to the purify.sanitize function below
     const elm = node as Element;
     const lcTagName = evt.tagName ?? elm.nodeName.toLowerCase();
+    let keepElement = !settings.sanitize;
 
     if (hasAllowedEncodings && lcTagName === 'semantics') {
       evt.allowedTags[lcTagName] = true;
     }
 
     if (lcTagName === 'annotation') {
-      const keepElement = hasValidEncoding(elm);
+      keepElement = hasValidEncoding(elm);
+    } else if (Type.isArray(settings.allow_mathml_elements)) {
+      keepElement = settings.allow_mathml_elements.includes(lcTagName);
+    } else {
+      keepElement = true;
+    }
+
+    if (keepElement) {
       evt.allowedTags[lcTagName] = keepElement;
-      if (!keepElement) {
-        elm.remove();
+    } else {
+      elm.remove();
+    }
+  });
+
+  purify.addHook('uponSanitizeAttribute', (_node, event) => {
+    if (Type.isArray(settings.allow_mathml_attributes)) {
+      const keepAttribute = settings.allow_mathml_attributes.includes(event.attrName);
+
+      if (keepAttribute) {
+        event.forceKeepAttr = true;
       }
     }
   });
--- a/modules/tinymce/src/core/main/ts/init/InitContentBody.ts
+++ b/modules/tinymce/src/core/main/ts/init/InitContentBody.ts
@@ -71,6 +71,8 @@ const mkParserSettings = (editor: Editor): DomParserSettings => {
   return removeUndefined<DomParserSettings>({
     allow_conditional_comments: getOption('allow_conditional_comments'),
     allow_html_data_urls: getOption('allow_html_data_urls'),
+    allow_mathml_attributes: getOption('allow_mathml_attributes'),
+    allow_mathml_elements: getOption('allow_mathml_elements'),
     allow_svg_data_urls: getOption('allow_svg_data_urls'),
     allow_html_in_named_anchor: getOption('allow_html_in_named_anchor'),
     allow_script_urls: getOption('allow_script_urls'),
--- a/modules/tinymce/src/core/test/ts/browser/html/SanitizationTest.ts
+++ b/modules/tinymce/src/core/test/ts/browser/html/SanitizationTest.ts
@@ -31,8 +31,8 @@ describe('browser.tinymce.core.html.SanitizationTest', () => {
   });
 
   context('Santitize non-html', () => {
-    const testNamespaceSanitizer = (testCase: { input: string; expected: string; sanitize?: boolean }) => {
-      const sanitizer = getSanitizer({ sanitize: testCase.sanitize ?? true }, Schema({ custom_elements: 'math' }));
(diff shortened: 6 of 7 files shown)


modules/tinymce/src/core/main/ts/html/Sanitization.ts
```diff
@@ -242,18 +242,34 @@ const sanitizeMathmlElement = (node: Element, settings: DomParserSettings) => {
   purify.addHook('uponSanitizeElement', (node, evt) => {
     // We know the node is an element as we have
     // passed an element to the purify.sanitize function below
-    const elm = node as Element;
-    const lcTagName = evt.tagName ?? elm.nodeName.toLowerCase();
+    const lcTagName = evt.tagName ?? node.nodeName.toLowerCase();
+    let keepElement;
 
     if (hasAllowedEncodings && lcTagName === 'semantics') {
```
spocke, line 248: I wonder if we could pull this out into a function like `isValidMathElement(lcTagName, node)` or something then we can make `keepElement` a `const`.

modules/tinymce/src/core/main/ts/html/Sanitization.ts
```diff
@@ -242,18 +242,34 @@ const sanitizeMathmlElement = (node: Element, settings: DomParserSettings) => {
...
     }
 
-    if (lcTagName === 'annotation') {
-      const keepElement = hasValidEncoding(elm);
-      evt.allowedTags[lcTagName] = keepElement;
-      if (!keepElement) {
-        elm.remove();
+    evt.allowedTags[lcTagName] = keepElement;
+    if (!keepElement && settings.sanitize) {
+      if (node.nodeType === NodeTypes.ELEMENT) {
```
spocke, line 261: You could use NodeType.isElement predicate then you don't need to cast node as element since that gets inferred by TS.

modules/tinymce/src/core/main/ts/api/Options.ts
```diff
@@ -1009,13 +1035,17 @@ const shouldConvertUnsafeEmbeds = option('convert_unsafe_embeds');
 const getLicenseKey = option('license_key');
 const getApiKey = option('api_key');
 const isDisabled = option('disabled');
+const allowExtraMathmlAttributes = option('allow_extra_mathml_attributes');
+const allowExtraMathmlElements = option('allow_extra_mathml_elements');
```
spocke, line 1039: Could we prefix these with `get` so they are `getAllowExtraMathmlAttributes` since the other getters are prefixed like that or `is` if they return a boolean.

modules/tinymce/src/core/test/ts/browser/html/SanitizationTest.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable mocha/no-exclusive-tests */
```
spocke, line 1: Leftover?

.changes/unreleased/tinymce-TINY-11756-2025-02-25.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Added
+body: Added `allow_extra_mathml_attributes` and `allow_extra_mathml_elements` options for the editor.
```
spocke, line 3: Not sure we need to specify `editor` it's all for the editor also the Added is kind of implied by the category.

So what about:

```suggestion
body: New `allow_extra_mathml_attributes` and `allow_extra_mathml_elements` options.
```

spocke (approved):

spocke (commented):
Also the PR title is incorrect please fix.

modules/tinymce/src/core/main/ts/api/Options.ts
```diff
@@ -324,6 +324,32 @@ const register = (editor: Editor): void => {
     default: []
   });
 
+  registerOption('allow_extra_mathml_attributes', {
+    processor: (value) => {
+      const valid = Type.isArrayOf(value, Type.isString);
```
kimwoodfield, line 329: We basically duplicated the code from within this option and the next. Could we extract it into a separate function and then just provide it to both processors??

spocke: We even implemented a `string[]` default processor type so I don't think this code is even needed.

- [ ] Consider using `string[]` as the processor

TheSpyder:
Do I really need to add "use a reasonable PR title" as a checkbox to the template?

modules/tinymce/src/core/main/ts/api/Options.ts
HAFRMO: Sure.

modules/tinymce/src/core/test/ts/browser/html/SanitizationTest.ts
HAFRMO: Yes, sorry about that.

modules/tinymce/src/core/main/ts/html/Sanitization.ts
spocke: What I mean is that you could refactor it to this:
```
  const isValidElement = (lcTagName: string) => {
    if (hasAllowedEncodings && lcTagName === 'semantics') {
      return true;
    } else if (lcTagName === 'annotation') {
      return node.nodeType === NodeTypes.ELEMENT && hasValidEncoding(node as Element);
    } else if (Type.isArray(settings.allow_mathml_elements)) {
      return Arr.contains(settings.allow_mathml_elements, lcTagName);
    } else {
      return false;
    }
  };

  purify.addHook('uponSanitizeElement', (node, evt) => {
    // We know the node is an element as we have
    // passed an element to the purify.sanitize function below
    const lcTagName = evt.tagName ?? node.nodeName.toLowerCase();
    const keepElement = isValidElement(lcTagName);

    evt.allowedTags[lcTagName] = keepElement;
    if (!keepElement && settings.sanitize) {
      if (node.nodeType === NodeTypes.ELEMENT) {
        (node as Element).remove();
      }
    }
  });
```

spocke: But I see now that we also need to then check if the `evt.allowedTags` already allow the tag. So we should only set it to true if it's already false. It becomes a bit confusing either way.

spocke: So it would be more like this:
```
  const isValidElement = (lcTagName: string, fallback: boolean) => {
    if (hasAllowedEncodings && lcTagName === 'semantics') {
      return true;
    } else if (lcTagName === 'annotation') {
      return node.nodeType === NodeTypes.ELEMENT && hasValidEncoding(node as Element);
    } else if (Type.isArray(settings.allow_mathml_elements)) {
      return Arr.contains(settings.allow_mathml_elements, lcTagName);
    } else {
      return fallback;
    }
  };

  purify.addHook('uponSanitizeElement', (node, evt) => {
    // We know the node is an element as we have
    // passed an element to the purify.sanitize function below
    const lcTagName = evt.tagName ?? node.nodeName.toLowerCase();
    const keepElement = isValidElement(lcTagName, evt.allowedTags[lcTagName]);

    evt.allowedTags[lcTagName] = keepElement;
    if (!keepElement && settings.sanitize) {
      if (node.nodeType === NodeTypes.ELEMENT) {
        (node as Element).remove();
      }
    }
  });
```

kimwoodfield (approved):

modules/tinymce/src/core/main/ts/api/OptionTypes.ts
```diff
@@ -50,6 +50,8 @@ interface BaseEditorOptions {
   add_form_submit_trigger?: boolean;
   add_unload_trigger?: boolean;
   allow_conditional_comments?: boolean;
+  allow_extra_mathml_attributes?: string[];
+  allow_extra_mathml_elements?: string[];
```
TheSpyder, line 54: I don't think word `extra` fits with our other configuration options. Some suggestions

- `allow_mathml_attributes` and `allow_mathml_elements` would fit better with the other `allow` configs
- `extended_mathml_attributes` and `extended_mathml_elements` is perhaps the best match, in the style of `extended_valid_elements`

modules/tinymce/src/core/main/ts/api/Options.ts
```diff
@@ -311,14 +311,21 @@ const register = (editor: Editor): void => {
...
       if (valid) {
-        const newValue = Type.isArray(value) ? value : Arr.map(value.split(','), Strings.trim);
-        return { value: newValue, valid };
+        return { value, valid };
       } else {
-        return { valid: false, message: 'Must be a string or an array of strings.' };
+        return { valid: false, message: 'Must be an array of strings.' };
       }
     },
     default: []
```
TheSpyder, line 331: Why is this not also `string[]`? I don't see why elements should be any different to attributes

TheSpyder (changes requested):

modules/tinymce/src/core/main/ts/api/OptionTypes.ts
spocke: It need the extra or something similar since it doesn't replace the whole check it adds additional things on top of what DOMPurify already allows. Otherwise they would need to pass in the whole spec in these options.
The original option name was this but there where complaints about that internally of being confusing and I agree with that.

modules/tinymce/src/core/main/ts/html/Sanitization.ts
```diff
@@ -239,21 +239,40 @@ const sanitizeMathmlElement = (node: Element, settings: DomParserSettings) => {
     return hasAllowedEncodings && Type.isString(encoding) && Arr.contains(allowedEncodings, encoding);
   };
 
-  purify.addHook('uponSanitizeElement', (node, evt) => {
-    const lcTagName = evt.tagName ?? node.nodeName.toLowerCase();
-
+  const isValidElementOpt = (lcTagName: string) => {
```
ltrouton, line 242: This looks like this this could just return a boolean rather than an Optional wrapped boolean

modules/tinymce/src/core/main/ts/api/OptionTypes.ts
spocke: Yes let go with that. It is a bit different to the extended_valid_elements since that allows you to define both the element and the attributes. But docs should clear that up. Option names that make sense is hard.

modules/tinymce/src/core/main/ts/html/Sanitization.ts
```diff
@@ -239,21 +239,40 @@ const sanitizeMathmlElement = (node: Element, settings: DomParserSettings) => {
...
+      }
+    });
+  });
+
+  purify.addHook('uponSanitizeAttribute', (_node, event) => {
+    if (Type.isArray(settings.allow_mathml_attributes)) {
+      const keepAttribute = settings.allow_mathml_attributes.includes(event.attrName);
+
+      if (keepAttribute) {
+        event.forceKeepAttr = true;
```
ltrouton, line 275: This doesn't seem right to me. We still want to check the node is a mathml element and also rather than `event.forceKeepAttr = true;` we should do the reverse and remove the non-allowed attribute instead. That seems safer to me at least.

modules/tinymce/src/core/main/ts/html/Sanitization.ts
HAFRMO: It's a bit of a tri-state, "Keep", "Remove", and "Don't change". I am concerned that the value on the object could be a true/false/undefined and that squashing it to true/false could make some changes to how the library act since it may go from a does-not-exist state to definitively-false state ( as might be the result when accepting the current value ). Using optionals removes this situation, which is why I picked it here.

modules/tinymce/src/core/main/ts/html/Sanitization.ts
HAFRMO: The node should already be checked by the node checking logic, We wouldn't need to check it a second time.

And if I remember right the default behavior is to remove many of the attributes, so not setting forcekeep just got them removed anyway, which wasn't very helpful.

modules/tinymce/src/core/main/ts/api/OptionTypes.ts
```diff
@@ -50,6 +50,8 @@ interface BaseEditorOptions {
   add_form_submit_trigger?: boolean;
   add_unload_trigger?: boolean;
   allow_conditional_comments?: boolean;
+  allow_extended_mathml_attributes?: string[];
+  allow_extended_mathml_elements?: string[];
```
TheSpyder, line 54: Sorry if the conversation was confusing. It's just "extended" not "allow extended". The idea is to match `extended_valid_elements`.

```suggestion
  extended_mathml_attributes?: string[];
  extended_mathml_elements?: string[];
```

spocke: Usually I want options that opt out of security be prefixed with allow to make the user think a bit of what am I allowing here exactly and why are things denied. However I guess these option names are fine since our extended_valid_elements leta you also opt out of security we just need to make sure that we conver this in docs.

modules/tinymce/src/core/main/ts/init/InitContentBody.ts
```diff
@@ -71,6 +71,8 @@ const mkParserSettings = (editor: Editor): DomParserSettings => {
   return removeUndefined<DomParserSettings>({
     allow_conditional_comments: getOption('allow_conditional_comments'),
     allow_html_data_urls: getOption('allow_html_data_urls'),
+    allow_mathml_attributes: getOption('extended_mathml_attributes'),
+    allow_mathml_elements: getOption('extended_mathml_elements'),
```
TheSpyder, line 75: This is a little bit of nitpicking, but every other parser setting matches the external option. I realise that many of them are `allow_*` and these aren't anymore, but I think keeping those matching names will help to avoid confusion in future maintenance.

TheSpyder (approved):

PR #10207, opened 2025-03-03
TINY-11822: remove Sizzle

michalnieruchalski-tiugo:
Related Ticket: TINY-11822

Description of Changes:
* Sizzle has been removed from dependencies.
* Tests has been adjusted to don't use jQuery specific features in selectors.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

## Summary by CodeRabbit

- **New Features**
  - Introduced an enhanced element matching engine and new dialog management methods based on dialog titles.
  - Added new functions for selecting elements by label in various modules, improving interaction with UI elements.
  
- **Enhancements**
  - Streamlined UI interactions across image, link, table, and color picker dialogs.
  - Improved keyboard navigation, focus behavior, and overall accessibility in menus and toolbars.
  - Enhanced error handling and robustness in toolbar button interactions and dialog management.

- **Chores**
  - Removed legacy dependencies and refactored internal tests for improved performance and maintainability.

### DIFF
--- a/modules/agar/src/main/ts/ephox/agar/alien/SelectorEngine.ts
+++ b/modules/agar/src/main/ts/ephox/agar/alien/SelectorEngine.ts
@@ -1,5 +1,4 @@
 import { Arr, Obj, Optional } from '@ephox/katamari';
-import { NodeTypes } from '@ephox/sugar';
 import * as Sizzle from 'sizzle';
 
 const sizzleEnabled = false;
@@ -41,17 +40,8 @@ const unwrapFromQuotes = (pattern: string) => {
   return matchedGroups.content ?? pattern;
 };
 
-const hasText = (element: Node, text: string) => {
-  if (element.nodeType === NodeTypes.TEXT) {
-    return element.textContent.includes(text);
-  }
-  for (const child of Array.from(element.childNodes)) {
-    if (hasText(child, text)) {
-      return true;
-    }
-  }
-  return false;
-};
+const hasText = (element: Node, text: string) =>
+  element.textContent.includes(text);
 
 const matchesSelector = (element: Element, selector: string): boolean => {
   if (sizzleEnabled) {
--- a/modules/agar/src/main/ts/ephox/agar/alien/SelectorEngine.ts
+++ b/modules/agar/src/main/ts/ephox/agar/alien/SelectorEngine.ts
@@ -1,4 +1,4 @@
-import { Arr, Obj, Optional } from '@ephox/katamari';
+import { Arr, Obj, Optional, Strings } from '@ephox/katamari';
 import * as Sizzle from 'sizzle';
 
 const sizzleEnabled = false;
@@ -41,7 +41,7 @@ const unwrapFromQuotes = (pattern: string) => {
 };
 
 const hasText = (element: Node, text: string) =>
-  element.textContent.includes(text);
+  Strings.contains(element.textContent, text);
 
 const matchesSelector = (element: Element, selector: string): boolean => {
   if (sizzleEnabled) {
--- a/modules/agar/src/main/ts/ephox/agar/alien/SelectorEngine.ts
+++ b/modules/agar/src/main/ts/ephox/agar/alien/SelectorEngine.ts
@@ -1,4 +1,4 @@
-import { Arr, Obj, Optional, Strings } from '@ephox/katamari';
+import { Arr, Obj, Optional, Result, Strings } from '@ephox/katamari';
 
 type LookupContext = Element | Document | DocumentFragment;
 interface DecodedContainsSelector {
@@ -7,18 +7,27 @@ interface DecodedContainsSelector {
 }
 
 const selectAll = (selector: string, context: LookupContext): Element[] =>
-  decodeContains(selector).fold(
-    () => queryAll(context, selector),
-    (decodedContainsSelector) => queryAllWithContains(context, decodedContainsSelector)
-  );
+  decodeContains(selector)
+    .fold(
+      (deocdeError) => {
+        throw new Error(deocdeError);
+      },
+      (result) => result.fold(
+        () => queryAll(context, selector),
+        (decodedContainsSelector) => queryAllWithContains(context, decodedContainsSelector)
+      )
+    );
 
-const decodeContains = (selector: string): Optional<DecodedContainsSelector> => {
-  const regexp = /(?<baseSelector>.*):contains\((?<text>.*)\)$/;
+const decodeContains = (selector: string): Result<Optional<DecodedContainsSelector>, string> => {
+  const regexp = /(?<baseSelector>.*):contains\((?<text>.*?)\)(?<remainings>.*)/;
   const matchedGroups = regexp.exec(selector)?.groups ?? {};
   if (!Obj.has(matchedGroups, 'baseSelector') || !Obj.has(matchedGroups, 'text')) {
-    return Optional.none();
+    return Result.value(Optional.none());
   }
-  return Optional.some({ baseSelector: matchedGroups.baseSelector, text: unwrapFromQuotes(matchedGroups.text) });
+  if (Obj.has(matchedGroups, 'remainings') && Strings.trim(matchedGroups.remainings) !== '') {
+    return Result.error(`Invalid selector '${selector}'. ':contains' in only supported at the end of the selector`);
+  }
+  return Result.value(Optional.some({ baseSelector: matchedGroups.baseSelector, text: unwrapFromQuotes(matchedGroups.text) }));
 };
 
 const unwrapFromQuotes = (pattern: string) => {
@@ -37,10 +46,16 @@ const hasText = (element: Node, text: string) =>
   Strings.contains(element.textContent, text);
 
 const matchesSelector = (element: Element, selector: string): boolean =>
-  decodeContains(selector).fold(
-    () => matches(selector, element),
-    (decodedContainsSelector) => matchesWithContains(decodedContainsSelector, element)
-  );
+  decodeContains(selector)
+    .fold(
+      (deocdeError) => {
+        throw new Error(deocdeError);
+      },
+      (result) => result.fold(
+        () => matches(selector, element),
+        (decodedContainsSelector) => matchesWithContains(decodedContainsSelector, element)
+      )
+    );
 
 const matches = (selector: string, element: Element) => element.matches(selector);
 
--- a/modules/agar/src/main/ts/ephox/agar/alien/SelectorEngine.ts
+++ b/modules/agar/src/main/ts/ephox/agar/alien/SelectorEngine.ts
@@ -1,22 +1,16 @@
 import { Arr, Obj, Optional, Strings } from '@ephox/katamari';
-import * as Sizzle from 'sizzle';
 
-const sizzleEnabled = false;
 type LookupContext = Element | Document | DocumentFragment;
 interface DecodedContainsSelector {
   baseSelector: string;
   text: string;
 }
 
-const selectAll = (selector: string, context: LookupContext): Element[] => {
-  if (sizzleEnabled) {
-    return Sizzle(selector, context);
-  }
-  return decodeContains(selector).fold(
+const selectAll = (selector: string, context: LookupContext): Element[] =>
+  decodeContains(selector).fold(
     () => queryAll(context, selector),
     (decodedContainsSelector) => queryAllWithContains(context, decodedContainsSelector)
   );
-};
 
 const decodeContains = (selector: string): Optional<DecodedContainsSelector> => {
   const regexp = /(?<baseSelector>.*):contains\((?<text>.*)\)$/;
@@ -43,15 +37,11 @@ const queryAllWithContains = (element: LookupContext, { baseSelector, text }: De
 const hasText = (element: Node, text: string) =>
   Strings.contains(element.textContent, text);
 
-const matchesSelector = (element: Element, selector: string): boolean => {
-  if (sizzleEnabled) {
-    return Sizzle.matchesSelector(element, selector);
-  }
-  return decodeContains(selector).fold(
+const matchesSelector = (element: Element, selector: string): boolean =>
+  decodeContains(selector).fold(
     () => matches(selector, element),
     (decodedContainsSelector) => matchesWithContains(decodedContainsSelector, element)
   );
-};
 
 const matches = (selector: string, element: Element) => element.matches(selector);
 
--- a/modules/agar/src/main/ts/ephox/agar/alien/SelectorEngine.ts
+++ b/modules/agar/src/main/ts/ephox/agar/alien/SelectorEngine.ts
@@ -6,6 +6,11 @@ interface DecodedContainsSelector {
   text: string;
 }
 
+const CONTAINS_REGEXP = /(?<baseSelector>.*):contains\((?<text>.*?)\)(?<remainings>.*)/;
+const CONTAINS_REGEXP_SINGLE_QUOTE = /(?<baseSelector>.*):contains\('(?<text>.*?)'\)(?<remainings>.*)/;
+const CONTAINS_REGEXP_DOUBLE_QUOTE = /(?<baseSelector>.*):contains\("(?<text>.*?)"\)(?<remainings>.*)/;
+const CONTAINS_REGEXPS = [ CONTAINS_REGEXP_DOUBLE_QUOTE, CONTAINS_REGEXP_SINGLE_QUOTE, CONTAINS_REGEXP ];
+
 const selectAll = (selector: string, context: LookupContext): Element[] =>
   decodeContains(selector)
     .fold(
@@ -19,20 +24,22 @@ const selectAll = (selector: string, context: LookupContext): Element[] =>
     );
 
 const decodeContains = (selector: string): Result<Optional<DecodedContainsSelector>, string> => {
-  const regexp = /(?<baseSelector>.*):contains\((?<text>.*?)\)(?<remainings>.*)/;
-  const matchedGroups = regexp.exec(selector)?.groups ?? {};
-  if (!Obj.has(matchedGroups, 'baseSelector') || !Obj.has(matchedGroups, 'text')) {
-    return Result.value(Optional.none());
-  }
-  if (Obj.has(matchedGroups, 'remainings') && Strings.trim(matchedGroups.remainings) !== '') {
-    return Result.error(`Invalid selector '${selector}'. ':contains' in only supported at the end of the selector`);
-  }
-  return Result.value(Optional.some({ baseSelector: matchedGroups.baseSelector, text: unwrapFromQuotes(matchedGroups.text) }));
-};
-
-const unwrapFromQuotes = (pattern: string) => {
-  const regexp = /^(?<quote>["'])(?<content>.*)\k<quote>$/;
-  return regexp.exec(pattern)?.groups?.content ?? pattern;
+  const selectorGroups = Arr.findMap(CONTAINS_REGEXPS, (regexp) => {
+    const matchedGroups = regexp.exec(selector)?.groups ?? {};
+    if (Obj.has(matchedGroups, 'baseSelector') && Obj.has(matchedGroups, 'text')) {
+      return Optional.from(matchedGroups);
+    }
+    return Optional.none();
+  });
+  return selectorGroups.fold(
+    () => Result.value(Optional.none()),
+    ({ baseSelector, text, remainings }) => {
+      if (remainings !== undefined && Strings.trim(remainings) !== '') {
+        return Result.error(`Invalid selector '${selector}'. ':contains' is only supported as the last pseudo-class.`);
+      }
+      return Result.value(Optional.some({ baseSelector, text }));
+    }
+  );
 };
 
 const queryAll = (element: LookupContext, selector: string) => Array.from(element.querySelectorAll(selector));
--- a/modules/agar/src/main/ts/ephox/agar/alien/SelectorEngine.ts
+++ b/modules/agar/src/main/ts/ephox/agar/alien/SelectorEngine.ts
@@ -12,15 +12,10 @@ const CONTAINS_REGEXP_DOUBLE_QUOTE = /(?<baseSelector>.*):contains\("(?<text>.*?
 const CONTAINS_REGEXPS = [ CONTAINS_REGEXP_DOUBLE_QUOTE, CONTAINS_REGEXP_SINGLE_QUOTE, CONTAINS_REGEXP ];
 
 const selectAll = (selector: string, context: LookupContext): Element[] =>
-  decodeContains(selector)
+  decodeContains(selector).getOrDie()
     .fold(
-      (deocdeError) => {
-        throw new Error(deocdeError);
-      },
-      (result) => result.fold(
-        () => queryAll(context, selector),
-        (decodedContainsSelector) => queryAllWithContains(context, decodedContainsSelector)
-      )
+      () => queryAll(context, selector),
+      (decodedContainsSelector) => queryAllWithContains(context, decodedContainsSelector)
     );
 
 const decodeContains = (selector: string): Result<Optional<DecodedContainsSelector>, string> => {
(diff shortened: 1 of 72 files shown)


michalnieruchalski-tiugo:
* **#10207** <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10207?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10207?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

modules/tinymce/src/plugins/lists/test/ts/browser/ListPropertiesTest.ts
```diff
@@ -38,7 +38,7 @@ describe('browser.tinymce.plugins.lists.ListPropertiesTest', () => {
 
   const updateDialog = (editor: Editor, currentValue: string, newValue: string) => {
     const doc = SugarDocument.getDocument();
-    FocusTools.isOnSelector('Check focus is on the input field', doc, inputSelector);
+    FocusTools.isOn('Check focus is on the input field', getInputSelector(doc));
```
TheSpyder, line 41: Likewise, this could've been `FocusTools.isOnByLabel`. My suggestion is to keep the agar API as easy to use as it was before, without needing to pass in elements (which often mandates `getOrDie`).

modules/tinymce/src/plugins/table/test/ts/browser/settings/TableAppearanceOptionsTest.ts
```diff
@@ -27,7 +27,7 @@ describe('browser.tinymce.plugins.table.TableAppearanceTest', () => {
       {
         'label:contains("Cell spacing")': 0,
         'label:contains("Cell padding")': 0,
-        'label:contains("Border") + input': 0,
+        'label:contains("Border")': 0,
```
TheSpyder, line 30: That's not _really_ the same thing. Maybe add a `label + input` check, which would be 4 or something like that?

modules/tinymce/src/plugins/table/test/ts/browser/ui/TableDialogKeyboardNavTest.ts
```diff
@@ -21,11 +22,13 @@ describe('browser.tinymce.plugins.table.TableDialogKeyboardNavTest', () => {
   const pressTabKey = (editor: Editor) => TinyUiActions.keydown(editor, Keys.tab());
 
   // Assert focus is on the expected form element
-  const pAssertFocusOnItem = (label: string, selector: string) => FocusTools.pTryOnSelector(
-    `Focus should be on: ${label}`,
-    SugarDocument.getDocument(),
-    selector
-  );
+  const pAssertFocusOnItem = (label: string, selectorOrElement: string | SugarElement<Node>) => {
+    if (typeof selectorOrElement === 'string') {
```
TheSpyder, line 26: Is it so hard to add a second assert method for the 7 cases that need it? Both would be a single line 🤷

modules/tinymce/src/themes/silver/test/ts/module/MenuUtils.ts
```diff
@@ -41,12 +41,25 @@ const pOpenAlignMenu = (label: string): Promise<void> => {
...
-  const menuTextParts = menuText.indexOf(':') > -1 ? menuText.split(':') : [ menuText ];
-  const btnText = menuTextParts[0];
-  const pseudo = menuTextParts.length > 1 ? ':' + menuTextParts[1] : '';
-  const selector = `button:contains(${btnText})${pseudo}`;
-  return pOpenMenuWithSelector(label, selector);
+const pOpenMenu = async (
+  label: string,
+  menuText: string,
+  resolveAmbiguity: ((elements: SugarElement[]) => SugarElement) = ([ firstElement ]) => firstElement
+): Promise<void> => {
```
TheSpyder, line 48: This is a ton of complexity and it only seems to be all just for cases of `:first-child` and `:last-child`. Could we add those into the `:contains` support of our selector engine instead?

Or if that's really too hard, can this be a simple first/last boolean instead of a function?

modules/tinymce/src/themes/silver/test/ts/headless/components/tree/TreeTest.ts
```diff
@@ -161,8 +161,9 @@ describe('headless.tinymce.themes.silver.tree.TreeTest', () => {
     store.assertEq('Dir collapsed', [ 'dir-collapsed' ]);
     store.clear();
 
+    const subdir = UiFinder.findIn(dir.element, '.tox-tree--directory').getOrDie();
     assertDirectoryExpandedState('Subdir', false, getTreeItem('.tox-tree--directory .tox-tree--directory'));
-    Mouse.clickOn(dir.element, '.tox-tree--directory .tox-trbtn.tox-tree--directory__label');
+    Mouse.clickOn(subdir, '.tox-trbtn.tox-tree--directory__label');
```
TheSpyder, line 166: Are you sure about these? They should've worked without this refactoring 🤔

modules/tinymce/src/themes/silver/test/ts/browser/skin/OxideTablePickerMenuTest.ts
```diff
@@ -84,15 +84,15 @@ describe('browser.tinymce.themes.silver.skin.OxideTablePickerMenuTest', () => {
...
+    await FocusTools.pTryOnSelector('Focus should be on first table cell', doc, '.tox-insert-table-picker__selected');
     TinyUiActions.keydown(editor, Keys.down());
     TinyUiActions.keydown(editor, Keys.right());
     Assertions.assertStructure(
       'Checking structure',
       ApproxStructure.build((s, str, arr) => insertTablePickerApprox(s, str, arr, 2, 2)),
       menu
     );
-    await FocusTools.pTryOnSelector('Focus should be on 2 down, 2 across table cell', doc, '.tox-insert-table-picker__selected:last');
+    await FocusTools.pTryOnSelector('Focus should be on 2 down, 2 across table cell', doc, '.tox-insert-table-picker > div[role="button"]:nth-child(12)');
```
TheSpyder, line 95: `:last-child` has been around as long as `:first-child`....
https://developer.mozilla.org/en-US/docs/Web/CSS/:last-child

- [ ] Keep `:last` (and I see plenty of other places in this PR)

modules/agar/src/main/ts/ephox/agar/alien/SelectorEngine.ts
```diff
@@ -0,0 +1,59 @@
...
+  if (!Obj.has(matchedGroups, 'baseSelector') || !Obj.has(matchedGroups, 'text')) {
+    return Optional.none();
+  }
+  return Optional.some({ baseSelector: matchedGroups.baseSelector, text: unwrapFromQuotes(matchedGroups.text) });
+};
+
+const unwrapFromQuotes = (pattern: string) => {
+  const regexp = /^(?<quote>["'])(?<content>.*)\k<quote>$/;
+  const matchedGroups = regexp.exec(pattern)?.groups ?? {};
+  return matchedGroups.content ?? pattern;
```
TheSpyder, line 27:
```suggestion
  return regexp.exec(pattern)?.groups?.content ?? pattern;
```

modules/agar/src/main/ts/ephox/agar/alien/SelectorEngine.ts
```diff
@@ -0,0 +1,59 @@
...
+}
+
+const selectAll = (selector: string, context: LookupContext): Element[] =>
+  decodeContains(selector).fold(
+    () => queryAll(context, selector),
+    (decodedContainsSelector) => queryAllWithContains(context, decodedContainsSelector)
+  );
+
+const decodeContains = (selector: string): Optional<DecodedContainsSelector> => {
+  const regexp = /(?<baseSelector>.*):contains\((?<text>.*)\)$/;
```
TheSpyder, line 16: The problem I have with this is it will fall back to `querySelectorAll` when `:contains` is in the middle - which will fail with `not a valid selector`.

I'd prefer to see this code capture any extra text, and then if that final group isn't empty throw an error that explains what is wrong

TheSpyder (commented):
You're doing really well! Some of this is nitpicking, but I'm also trying to make sure the result is not surprising for someone learning the codebase long after the change is made.

modules/agar/src/main/ts/ephox/agar/find/UiSearcher.ts
```diff
@@ -73,8 +73,25 @@ const findIn = <T extends Element>(container: SugarElement<Node>, selector: stri
 const findAllIn = <T extends Element>(container: SugarElement<Node>, selector: string): Array<SugarElement<T>> =>
   selectAll(container, selector);
 
+const findTargetByLabel = <T extends Element>(container: SugarElement<Node>, labelText: string): Result<SugarElement<T>, TestLabel> => {
+  const label = Arr.find(selectAll(container, 'label'), (e) => e.dom.textContent === labelText);
```
hamza0867, line 77: Since we are using sugar elements, I'd rather you use `TextContent.get` like so:
```suggestion
  const label = Arr.find(selectAll(container, 'label'), (e) => TextContent.get(e) === labelText);
```
Of course you need to import `TextContent`

modules/tinymce/src/plugins/image/test/ts/browser/UploadTabTest.ts
```diff
@@ -52,12 +52,12 @@ describe('browser.tinymce.plugins.image.UploadTabTest', () => {
...
 
   const pAssertSrcTextValue = (expectedValue: string) => Waiter.pTryUntil('Waited for input to change to expected value', () => {
-    const input = UiFinder.findIn<HTMLInputElement>(SugarBody.body(), 'label.tox-label:contains("Source") + div > div > input.tox-textfield').getOrDie();
+    const input = UiFinder.findTargetByLabel<HTMLInputElement>(SugarBody.body(), 'Source').getOrDie();
     assert.equal(Value.get(input), expectedValue, 'Assert field source value ');
   }, 10, 7000);
 
   const pAssertSrcTextValueStartsWith = (expectedValue: string) => Waiter.pTryUntil('Waited for input to change to start with expected value', () => {
-    const input = UiFinder.findIn<HTMLInputElement>(SugarBody.body(), 'label.tox-label:contains("Source") + div > div > input.tox-textfield').getOrDie();
+    const input = UiFinder.findTargetByLabel<HTMLInputElement>(SugarBody.body(), 'Source').getOrDie();
```
hamza0867, line 60: I am a bit surprised this works, shouldn't `findTargetByLabel` here return the `label` element instead of the `input` element ?

modules/tinymce/src/plugins/link/test/ts/browser/LinkContextMenuTest.ts
```diff
@@ -45,6 +45,6 @@ describe('browser.tinymce.plugins.link.LinkContextMenuTest', () => {
     TinySelections.select(editor, 'a', []);
     await TinyUiActions.pTriggerContextMenu(editor, 'a[href="http://tiny.cloud"]', '.tox-silver-sink [role="menuitem"]');
     await pTestContextMenuItems(editor);
-    await pAssertFocusOnItem('Open link', '.tox-collection__item:contains("Open link"):not([aria-disabled="true"])');
+    await pAssertFocusOnItem('Open link', '.tox-collection__item:not([aria-disabled="true"]):contains("Open link")');
```
hamza0867, line 48: I assume this was done automatically by the editor.

hamza0867 (commented):
Good job

modules/agar/src/main/ts/ephox/agar/find/UiSearcher.ts
michalnieruchalski-tiugo: ✅

modules/tinymce/src/themes/silver/test/ts/headless/components/tree/TreeTest.ts
michalnieruchalski-tiugo: I’m sure! This test was failing after my change. In this case querySelector works a bit different than sizzle. Take a look at this example:
https://codepen.io/michalnieruchalskit/pen/QwEXveZ

modules/tinymce/src/core/test/ts/browser/dom/ControlSelectionTest.ts
```diff
@@ -172,7 +172,7 @@ describe('browser.tinymce.core.dom.ControlSelectionTest', () => {
     TinySelections.select(editor, 'span.mce-object-video', []);
     await pWaitForDragHandles(editorBody, '#mceResizeHandlenw');
     TinyAssertions.assertContentPresence(editor, {
-      'span[data-mce-selected=1] video': 1,
+      'span[data-mce-selected="1"] video': 1,
```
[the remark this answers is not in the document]
michalnieruchalski-tiugo: `span[data-mce-selected=1]` is not a valid CSS selector. I’ve changed that because It was throwing an error. Take a look at this example: 
https://codepen.io/michalnieruchalskit/pen/emzwWEQ

You can see that `div[data-mce-selected=1]` does not match the div (it did not change the color).
By clicking at the button you can see that querySelector throws an error. My understanding of the documentation you provided is that: 

1. `<value>` can be `<ident>` or `<string>` 
2. `<string>` is defined as “data type composed of any number of Unicode characters surrounded by either double or single quotes."
3. `<ident>` which is an “arbitrary string used as an identifier” and it has to follows certain rules, including: 
    * “An identifier must not start with an unescaped digit” 

Therefore in this selector: `div[data-ace-selected=1]` 1 is not a `<string>` and it’s not an `<ident>` because it starts with a digit.

modules/tinymce/src/themes/silver/test/ts/browser/skin/OxideTablePickerMenuTest.ts
michalnieruchalski-tiugo: I don’t understand this comment. Could you please elaborate?
1. I cannot keep :last as it’s not a part of css
2. Sizzle’s :last and CSS :last-child are not the same thing. Therefore I can’t replace all :last with :last-child. https://api.jquery.com/last-selector/ 
    1. :last is basically querySelectorAll + get the last returned element. It’s hard to achieve with CSS (or maybe I don’t know how)
    2. :last-child represents the last element among a group of sibling elements
    3. So for example :last will always return one match, and :last-child can return multiple
3. Replacing :last with :last-child in this test-case will not work as intended

If you don’t like the :nth-child(12) here, because it assumes that we have a 10x10 grid in table selector. I can propose three different solutions:


```suggestion
const lastSelectedCell = Arr.last(UiFinder.findAllIn(doc, '.tox-insert-table-picker__selected')).getOrDie();
await FocusTools.pTryOn('Focus should be on 2 down, 2 across table cell', lastSelectedCell);
```


```suggestion
await FocusTools.pTryOnSelector('Focus should be on 2 down, 2 across table cell', doc, '.tox-insert-table-picker__selected:nth-last-child(1 of .tox-insert-table-picker__selected)');
```


```suggestion
await FocusTools.pTryOnSelector('Focus should be on 2 down, 2 across table cell', doc, '.tox-insert-table-picker__selected[aria-label="2 columns, 2 rows"]')
```

Let me know what do you think.

modules/tinymce/src/plugins/image/test/ts/browser/UploadTabTest.ts
michalnieruchalski-tiugo: `findTargetByLabel` returns `input` of the corresponding `label`. It works as Spocky has described in the Jira ticket. You can see the implementation in `SelectorEngine.ts`

modules/tinymce/src/plugins/link/test/ts/browser/LinkContextMenuTest.ts
michalnieruchalski-tiugo: Nope, I did it on purpose. Right now we only support :contains at the end of the selector. So by changing order here I was able to maintain the same returned element and comply with our new constraint.

modules/tinymce/src/plugins/table/test/ts/browser/settings/TableAppearanceOptionsTest.ts
michalnieruchalski-tiugo: I agree it’s not the same thing, but IMO it’s enough. In this test we try to verify that there’s no input labeled as “Border”. If there’s no such label, there also won’t be any input labeled as such :D  “Cell spacing” and “Cell padding” are exactly the same as “Border” (number inputs with label) and were previously verified only by the lack of existence of their label.

michalnieruchalski-tiugo: It would make more sense to add additional checks for the test below that one, where we check if given inputs exist.



Then it would be:

```
Assertions.assertPresence(
  'assert presence of spacing, padding, border and caption inputs',
  {
    'label:contains("Cell spacing")': 1,
    'label:contains("Cell padding")': 1,
    'label:contains("Border")': 1,
    'label:contains("Caption")': 1,
    'label + input': 5,
    'label + .tox-listboxfield': 1,
    'label + .tox-checkbox': 1
  },
  dialog
);
```

modules/tinymce/src/themes/silver/test/ts/module/MenuUtils.ts
michalnieruchalski-tiugo: As I mentioned in previous comment `:last` and `:last-child` is not the same. Adding support for `:last-child` wouldn't be helpful. 

In general I avoid adding boolean parameters to functions. It's not a written on the stone rule, but just a hint as it leads to messier code. I agree that passing function as parameter is too much.

Please take a look at it now, I've introduced an interface that describes the menu by it's "text" and position in the editor lastMatch: true/false. I think it looks better now. Let me know what do you think.

modules/tinymce/src/core/test/ts/browser/dom/ControlSelectionTest.ts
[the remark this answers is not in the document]
TheSpyder: ah, my apologies, I usually search for attribute values that are strings. I didn't understand the ident restriction.

modules/tinymce/src/themes/silver/test/ts/headless/components/tree/TreeTest.ts
TheSpyder: Woah! I thought sizzle was fully compliant. I wonder how many weird bugs we've had in tests due to this difference 🙃

modules/tinymce/src/themes/silver/test/ts/browser/skin/OxideTablePickerMenuTest.ts
TheSpyder: Sorry I didn't get back to this last week. What I was hoping you'd do after my comment was try to prove me wrong by looking at the MDN docs for `:last-child`... and notice there's a `:last-of-type` next to it in the sidebar which _is_, as far as I can see, the same as sizzle's `:last` 🫣 
https://developer.mozilla.org/en-US/docs/Web/CSS/:last-of-type

modules/tinymce/src/themes/silver/test/ts/webdriver/tooltip/HtmlPanelTooltipTest.ts
```diff
@@ -67,6 +67,7 @@ describe('browser.tinymce.themes.silver.editor.HtmlPanelTooltipTest', () => {
       TinyUiActions.closeDialog(editor);
     });
 
+    /* TODO: fix */
```
TheSpyder, line 70: Please either actually fix it or log a task 🙂 

- [ ] use or remove

TheSpyder (approved):
Fantastic work, this is much better!

There's just that lingering question over `matchLast` to consider.

modules/agar/src/main/ts/ephox/agar/assertions/ApproxStructures.ts
```diff
@@ -133,6 +135,7 @@ const element = (tag: string, fields: ElementFields): StructAssert => {
 
       assertHtml(optHtml, actual);
       assertValue(optValue, actual);
+      assertTextContent(optTextContent, actual);
       assertChildren(optChildren, actual);
```
TheSpyder, line 139: I missed this one. It already exists under `children`, by using `ApproxStructures.text`
https://github.com/tinymce/tinymce/blob/e9a1186a9d8397dd5c08f343135bd0bc66786004/modules/agar/src/test/ts/browser/ApproxStructureTest.ts#L139-L144

- [ ] revert (and update tests that use it)

modules/tinymce/src/themes/silver/test/ts/browser/skin/OxideTablePickerMenuTest.ts
michalnieruchalski-tiugo: Unfortunately :last-of-type is not the same as :last. I can’t replace :last with :last-of-type in this test case.
1. :last is basically querySelectorAll + get the last returned element. https://api.jquery.com/last-selector
2. :last-of-type represents the last element of its **type** among a group of sibling elements. So the last `<div>` in the container, or the last `<p>` in the container, not the last div with some class on it.

Take a look at this example:
https://codepen.io/michalnieruchalskit/pen/ogLrWyj

TheSpyder: Fair point, and thank you for sticking with me as I try to explore the best way to tackle this. Perhaps we should discuss on Slack instead of github threads.

I can see how they're not the same when multiple "types" of elements match, which isn't what I expected reading the spec. But I think this specific case would still work, `:last-of-type` should return the 12th `div[role="button"]`?

modules/tinymce/src/themes/silver/test/ts/module/MenuUtils.ts
TheSpyder: I was about to argue further, based on the other discussion thread about `:last`, but then I remembered the original selector here was `button:contains('text'):last`. With the restrictions on implementing `:contains` ourselves there's nothing else we can do in this specific case.

I agree your edits are a good improvement, but I think in this new context please rename `matchLast` to `last`.

michalnieruchalski-tiugo: I agree, I like the `last` better

modules/tinymce/src/themes/silver/test/ts/browser/skin/OxideTablePickerMenuTest.ts
michalnieruchalski-tiugo: Here's what we're dealing with in this test case. Each cell in the grid is a `div[role="button"]`. Each selected cell (blue cell) is `div[role="button"].tox-insert-table-picker__selected`

I think you're expecting `div[role="button"].tox-insert-table-picker__selected:last-of-type` to return the last blue cell, but it won't. 

We can achieve it in the css by using this selector for example (as I mentioned in my first comment in this thread)
```
div[role="button"]:nth-last-child(1 of .tox-insert-table-picker__selected)
```

TheSpyder: OK. I admit I didn't really follow the suggestions because of how graphite formats them (without highlighting the inline diff). Now that you've explained it, I went back to the github ui and realise that yes you were trying to say this all along. I like the `:nth-last-child(1 of .tox-insert-table-picker__selected)` approach.

TheSpyder (approved):
I'm happy 👍 

You'll need to resolve the coderabbit threads yourself, I didn't check if you followed the bot's suggestions (but I suppose if it hasn't complained you probably did).

michalnieruchalski-tiugo:
> I'm happy 👍 
> 
> You'll need to resolve the coderabbit threads yourself, I didn't check if you followed the bot's suggestions (but I suppose if it hasn't complained you probably did).

These comments are from linter I think. I can't resolve them for some reason, they appear when I push changes with linting issues. Would be nice to have pre-commit hook set-up.

spocke (approved):

TheSpyder:
Oh, sorry, I was looking at them in Graphite and didn't notice. It's a github bot, so in the github UI it's clear they've been auto-resolved when you fixed the issue. Chalk up another reason I don't use graphite full time...

PR #10253, opened 2025-03-26
TINY-11986: Adding simulating typing into input elements to agar

spocke:
Related Ticket: TINY-11986

Description of Changes:
* Added new `UiControls.pType` that lets you simulate typing into textarea/input fields that support it

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

## Summary by CodeRabbit

- **New Features**
	- Enhanced text input simulation now accurately mimics natural keyboard interactions in input and textarea fields.
	- New functionality added to UI controls for simulating typing text into input elements at specified speeds.
	- Introduced new methods for handling keyboard interactions, improving user experience with input fields.
	- Comprehensive mapping of keyboard keys and their respective key codes has been added for better event handling.

- **Tests**
	- Added a new test suite to validate keyboard interactions within input elements, ensuring reliability and correctness of the new features.


TheSpyder (approved):
I'm surprised we have features that depend on real simulated typing, I would've thought just setting the value and dispatching a `change` or `input` event meets most needs 🤷‍♂️

spocke:
> I'm surprised we have features that depend on real simulated typing, I would've thought just setting the value and dispatching a `change` or `input` event meets most needs 🤷‍♂️

For the most part yes but we do have auto completers for @mentions in comments for a example that open and refresh on each keystroke.
We have a similar typing key by key inside the editor content so I think it would be useful to have one for textareas/inputs that trigger all the events in the same way.
For the mathml case for example we just type one letter and that is likely the more common scenario and this can be used for that.

modules/agar/src/test/ts/browser/keyboard/TypeInInputTest.ts
```diff
@@ -0,0 +1,140 @@
...
+
+describe('browser.agar.keyboard.TypeInInputTest', () => {
+  const pTestOnTag = async (tagHtml: string) => {
+    const store = TestStore.TestStore<StoreDataItem>();
+
+    const input = SugarElement.fromHtml<HTMLInputElement | HTMLTextAreaElement>(tagHtml);
+    Insert.append(SugarBody.body(), input);
+
+    DomEvent.bind(input, 'keydown', ({ raw, kill }) => {
+      if (raw.keyCode === 68) {
```
hamza0867, line 33: I don't think we should use `keyCode` in new code since it's deprecated and less readable. I think it's better that we use the `code` property. What do you think ?

modules/agar/src/test/ts/browser/keyboard/TypeInInputTest.ts
```diff
@@ -0,0 +1,140 @@
...
+        which: raw.which,
+        key: raw.key,
+        code: raw.code,
+        keyCode: raw.keyCode,
+        charCode: raw.charCode,
+        shiftKey: raw.shiftKey
+      });
+    });
+    DomEvent.bind(input, 'keypress', ({ raw, kill }) => {
+      if (raw.charCode === 'e'.charCodeAt(0)) {
```
hamza0867, line 48: Same here.

hamza0867 (approved):

modules/agar/src/test/ts/browser/keyboard/TypeInInputTest.ts
spocke: Yes, that makes sense since at some point we might want to add some lint rule to get rid of these and keeping it around would just add work to that. I will see if I can stick in a follow up ticket to also use this more modern key handling for the `Keyboard.keydown` etc should be pretty simple to introduce it there then we can simulate modern keyboard events with agar that would be nice.

PR #10261, opened 2025-04-02
TINY-11363: [ReadOnly] Language with IME can insert content while in readonly mode

ArvinJ-H:
Related Ticket: 
TINY-11363

Description of Changes:
- Blocked all `composition` state default behavior in read only mode
- Added `mutationObserver` to revert any change that might be added by IME or otherwise

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

## Summary by CodeRabbit

- **New Features**
  - Enhanced the editor's read-only mode to prevent unintended content changes during text composition by monitoring DOM mutations and blocking additional input events.

- **Tests**
  - Expanded test coverage with simulations of IME and keyboard input events to ensure content remains unchanged in read-only mode.
  - Verified undo and redo commands have no effect when the editor is in read-only mode.


modules/tinymce/src/core/main/ts/mode/Readonly.ts
```diff
@@ -38,23 +39,50 @@ const registerReadOnlyInputBlockers = (editor: Editor): void => {
...
+          if (Type.isNonNullable(undoLevel)) {
+            editor.undoManager.undo();
+          }
+        }
+      });
     }
   });
+
+  editor.on('init', () => {
+    observer.observe(editor.getBody(), {
```
TheSpyder, line 76: Do we still need this if we're blocking the composition events? Even with the readonly check this seems very expensive to have permanently attached to every editor.

If the answer is we do need it, maybe have it only attached during composition 🤔

modules/tinymce/src/core/main/ts/mode/Readonly.ts
```diff
@@ -38,23 +39,50 @@ const registerReadOnlyInputBlockers = (editor: Editor): void => {
...
+  // Set up mutation observer to detect and revert unintended changes
+  const observer = new MutationObserver((mutations) => {
     if (isReadOnly(editor)) {
-      const undoLevel = editor.undoManager.add();
-      if (Type.isNonNullable(undoLevel)) {
-        editor.undoManager.undo();
+      // Skip if we're in an undo transaction (annotation changes)
+      if (editor.undoManager.typing || editor.undoManager.data.length > 0) {
+        return;
       }
```
TheSpyder, line 61:
```suggestion
    // Skip if we're in an undo transaction (annotation changes)
    if (isReadOnly(editor) && !(editor.undoManager.typing || editor.undoManager.data.length > 0)) {
```

modules/tinymce/src/core/main/ts/mode/Readonly.ts
```diff
@@ -38,23 +39,50 @@ const registerReadOnlyInputBlockers = (editor: Editor): void => {
...
+
+      mutations.forEach((mutation) => {
+        if (mutation.type === 'characterData' || mutation.type === 'childList') {
+          // Revert the changes by restoring the previous content
+          const undoLevel = editor.undoManager.add();
+          if (Type.isNonNullable(undoLevel)) {
+            editor.undoManager.undo();
```
TheSpyder, line 68: If there are multiple mutations this will be quite expensive. Can't we just check if any of the mutations are what we're after and revert the content once?

TheSpyder (changes requested):

modules/tinymce/src/core/main/ts/mode/Readonly.ts
ArvinJ-H: yeah we do need the observer to not show any inputed content while `composition`, I've updated this to start at `compositionstart` and disconnect at `compositionend`

modules/tinymce/src/core/main/ts/mode/Readonly.ts
ArvinJ-H: Updated to batch processing, I'm thinking to have observer at all time, it should be okay now if we doing batch processing?

modules/tinymce/src/core/main/ts/mode/Readonly.ts
spocke: Composition events can't be cancelled. See spec: https://w3c.github.io/uievents/#event-type-compositionend

However we can tigger a undo on composition end that way we might not need the observer.

Also if you use observer then you should only observe if we are in a readonly state this code seems to always observe that and it could be slow.

spocke (changes requested):

modules/tinymce/src/core/main/ts/mode/Readonly.ts
ArvinJ-H: I added the change suggested.

There's some issues with moving undo to composition end, occasionally and frequently there will be some content be inserted into the text area using Japanese. So I kept the observer, and wrapped it in `isReadOnly()`

modules/tinymce/src/core/main/ts/mode/Readonly.ts
```diff
@@ -26,7 +26,8 @@ const toggleReadOnly = (editor: Editor, state: boolean): void => {
 const isReadOnly = (editor: Editor): boolean => editor.readonly;
 
 const registerReadOnlyInputBlockers = (editor: Editor): void => {
-  editor.on('beforeinput paste cut dragend dragover draggesture dragdrop drop drag', (e) => {
+  // Block all input events in readonly mode
+  editor.on('beforeinput input paste cut dragend dragover draggesture dragdrop drop drag', (e) => {
```
spocke, line 30: The `input` event is non cancellable see: https://w3c.github.io/uievents/#event-type-input `beforeinput` is on most platforms but not Android.

modules/tinymce/src/core/main/ts/mode/Readonly.ts
TheSpyder: yes, batch processing is good 👍 as is only connecting the observer while composing in readonly mode 👍

modules/tinymce/src/core/main/ts/mode/Readonly.ts
```diff
@@ -38,23 +39,54 @@ const registerReadOnlyInputBlockers = (editor: Editor): void => {
...
+      observer.observe(editor.getBody(), {
+        characterData: true,
+        childList: true,
+        subtree: true
+      });
+    }
+  });
+
+  editor.on('compositionend', () => {
+    observer.disconnect();
```
spocke, line 88: I think it's better to do the filtering of the mutations here by calling the `takeRecords()` that call always return any pending mutations. The callback is not always executed for every mutation it has a async nature to it. Also I think that is a bit cleaner.

modules/tinymce/src/core/main/ts/mode/Readonly.ts
```diff
@@ -38,23 +39,54 @@ const registerReadOnlyInputBlockers = (editor: Editor): void => {
...
+        if (Type.isNonNullable(undoLevel)) {
+          editor.undoManager.undo();
+        }
       }
     }
   });
 
-  editor.on('compositionend', () => {
+  // Set up mutation observer to detect and revert unintended changes
+  const observer = new MutationObserver((mutations) => {
```
spocke, line 56: Pull this callback out into a function that captures intent. Like `hasContentMutations` or something that returns a boolean from an array of mutations.

modules/tinymce/src/core/main/ts/mode/Readonly.ts
```diff
@@ -38,23 +39,54 @@ const registerReadOnlyInputBlockers = (editor: Editor): void => {
...
     if (isReadOnly(editor)) {
-      const undoLevel = editor.undoManager.add();
-      if (Type.isNonNullable(undoLevel)) {
-        editor.undoManager.undo();
+      let needsRevert = false;
+
+      mutations.forEach((mutation) => {
```
spocke, line 60: Use `Arr.exists` here instead then you don't need to loop the entire thing to flip a state to true also it's a lot cleaner since you don't need any local mutation.

modules/tinymce/src/core/main/ts/mode/Readonly.ts
```diff
@@ -38,23 +39,54 @@ const registerReadOnlyInputBlockers = (editor: Editor): void => {
...
+        if (mutation.type === 'characterData' || mutation.type === 'childList') {
+          // Mark that we need to revert
+          needsRevert = true;
+        }
+      });
+
+      // Only create one undo level for all mutations in this batch
+      if (needsRevert) {
+        const undoLevel = editor.undoManager.add();
```
spocke, line 69: Since this is called two times move it to a function that captures intent like `rollbackUndo` or something.
Also when you `undo` the stack still has the extra `redo` level so ideally we should try to get rid of that since this would cause weird effects if you say do a mutation in readonly mode. Then toggle readonly off then you have a `redo` state that you can move to unless I'm missing something here.

spocke (changes requested):

TheSpyder (approved):
The bot has pointed out that using `undo` management like this leaves open the possibility of the user having access to `redo` after composition ends.

Please implement that suggestion (minus the fallback for old apis) and add a test for it.

Don't worry about the bot's `takeRecords` suggestion. That's incorrect.

spocke (approved):

ArvinJ-H:
### Merge activity

* **Apr 29, 11:14 AM EDT**: @ArvinJ-H merged this pull request with [Graphite](https://app.graphite.dev/github/pr/tinymce/tinymce/10261).

PR #10277, opened 2025-04-09
TINY-12021: Add pageUid and editorUid properties

ltrouton:
Related Ticket: TINY-12021

Description of Changes:
* Add `uuidV4` generator to katamari `Id` module
* Add `pageUid` to tinymce global (more specifically the EditorManager)
* Add `editorUId` to editor instance object

Pre-checks:
* [X] Changelog entry added

Review:
* [X] Milestone set
* ~~[X] Docs ticket created (if applicable)~~

### DIFF
--- a/modules/katamari/src/main/ts/ephox/katamari/api/Id.ts
+++ b/modules/katamari/src/main/ts/ephox/katamari/api/Id.ts
@@ -1,4 +1,6 @@
+import * as IdUtils from '../util/IdUtilts';
 import * as Num from './Num';
+import * as Type from './Type';
 
 /**
  * Generate a unique identifier.
@@ -13,7 +15,7 @@ import * as Num from './Num';
  */
 let unique = 0;
 
-export const generate = (prefix: string): string => {
+const generate = (prefix: string): string => {
   const date = new Date();
   const time = date.getTime();
   const random = Math.floor(Num.random() * 1000000000);
@@ -21,4 +23,22 @@ export const generate = (prefix: string): string => {
   unique++;
 
   return prefix + '_' + random + unique + String(time);
+};
+
+/**
+ * Generate a uuidv4 string
+ * In accordance with RFC 4122 (https://datatracker.ietf.org/doc/html/rfc4122)
+ */
+const uuidV4 = (): `${string}-${string}-${string}-${string}-${string}` => {
+  // eslint-disable-next-line @typescript-eslint/unbound-method
+  if (Type.isFunction(window.crypto.randomUUID) && window.isSecureContext) {
+    return window.crypto.randomUUID();
+  } else {
+    return IdUtils.uuidV4String();
+  }
+};
+
+export {
+  generate,
+  uuidV4
 };
--- a/modules/katamari/src/main/ts/ephox/katamari/api/Id.ts
+++ b/modules/katamari/src/main/ts/ephox/katamari/api/Id.ts
@@ -1,6 +1,5 @@
 import * as IdUtils from '../util/IdUtilts';
 import * as Num from './Num';
-import * as Type from './Type';
 
 /**
  * Generate a unique identifier.
@@ -31,7 +30,7 @@ const generate = (prefix: string): string => {
  */
 const uuidV4 = (): `${string}-${string}-${string}-${string}-${string}` => {
   // eslint-disable-next-line @typescript-eslint/unbound-method
-  if (Type.isFunction(window.crypto.randomUUID) && window.isSecureContext) {
+  if (window.isSecureContext) {
     return window.crypto.randomUUID();
   } else {
     return IdUtils.uuidV4String();
--- a/modules/katamari/src/main/ts/ephox/katamari/util/IdUtilts.ts
+++ b/modules/katamari/src/main/ts/ephox/katamari/util/IdUtilts.ts
@@ -2,8 +2,17 @@
 
 const uuidV4Bytes = (): Uint8Array<ArrayBuffer> => {
   const bytes = window.crypto.getRandomValues(new Uint8Array(16));
+
+  // https://tools.ietf.org/html/rfc4122#section-4.1.3
+  // This will first bit mask away the most significant 4 bits (version octet)
+  // then mask in the v4 number we only care about v4 random version at this point so (byte & 0b00001111 | 0b01000000)
   bytes[6] = bytes[6] & 15 | 64;
+
+  // https://tools.ietf.org/html/rfc4122#section-4.1.1
+  // This will first bit mask away the highest two bits then masks in the highest bit so (byte & 0b00111111 | 0b10000000)
+  // So it will set the Msb0=1 & Msb1=0 described by the "The variant specified in this document." row in the table
   bytes[8] = bytes[8] & 63 | 128;
+
   return bytes;
 };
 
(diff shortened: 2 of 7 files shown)


modules/tinymce/src/core/test/ts/browser/EditorManagerTest.ts
```diff
@@ -14,8 +19,24 @@ import * as ViewBlock from '../module/test/ViewBlock';
...
+    assert.isString(uuid);
+    // Need to narrow type
+    if (!Type.isString(uuid)) {
+      assert.fail();
+    }
+    assert.lengthOf(uuid, 36);
+    assert.match(uuid, v4Regex);
+    const version = parseInt(uuid.slice(14, 15), 16);
+    assert.equal(version, 4);
+  };
```
TheSpyder, line 34: Please add further validation, specifically
https://datatracker.ietf.org/doc/html/rfc4122#section-4.4

> Set the two most significant bits (bits 6 and 7) of the clock_seq_hi_and_reserved to zero and one, respectively.

The way we've done this in the past is by verifying with the bytes, not the string, using `const bits = bytes[8].toString(2).padStart(8, "0")` and then verifying `bits.substring(0, 2)` is `"10"`. So it might be easier to have that as a specific uuid test rather than trying to parse it out of the string afterwards 🤔

modules/tinymce/src/core/test/ts/browser/EditorManagerTest.ts
```diff
@@ -1,7 +1,12 @@
 import { after, afterEach, before, describe, it } from '@ephox/bedrock-client';
+import { Type } from '@ephox/katamari';
 import { Remove, Selectors } from '@ephox/sugar';
 import { assert } from 'chai';
 import 'tinymce';
+// eslint-disable-next-line @tinymce/no-main-module-imports
```
spocke, line 6: Why was this needed?

modules/tinymce/src/core/test/ts/browser/EditorManagerTest.ts
ltrouton: I'll see what I can do 👍

modules/tinymce/src/core/test/ts/browser/EditorManagerTest.ts
ltrouton: The test doesn't run in isolation otherwise. Since we are not pulling in `mcagar`, we don't get the automatic import of tinymce, theme and model

spocke: It should be lazy loaded then maybe you need to set the base url. Check other tests.
This would make it actually import all of the silver theme in the test bundle.

ltrouton: Thanks, I'll have a look 👍

modules/katamari/src/main/ts/ephox/katamari/api/Id.ts
```diff
@@ -22,3 +24,21 @@ export const generate = (prefix: string): string => {
...
   return prefix + '_' + random + unique + String(time);
 };
+
+/**
+ * Generate a uuidv4 string
+ * In accordance with RFC 4122 (https://datatracker.ietf.org/doc/html/rfc4122)
+ */
+const uuidV4 = (): `${string}-${string}-${string}-${string}-${string}` => {
+  // eslint-disable-next-line @typescript-eslint/unbound-method
+  if (Type.isFunction(window.crypto.randomUUID) && window.isSecureContext) {
```
TheSpyder, line 34: I know you like to be defensive, but the second check implies the function exists on all supported browsers. On older browsers that don't have `crypto`, `window.isSecureContext` is likely to be `undefined`.
```suggestion
  if (window.isSecureContext) {
```

modules/katamari/src/main/ts/ephox/katamari/util/IdUtilts.ts
```diff
@@ -0,0 +1,26 @@
+/* eslint-disable no-bitwise */
+
+const uuidV4Bytes = (): Uint8Array<ArrayBuffer> => {
+  const bytes = window.crypto.getRandomValues(new Uint8Array(16));
+  bytes[6] = bytes[6] & 15 | 64;
+  bytes[8] = bytes[8] & 63 | 128;
+  return bytes;
```
TheSpyder, line 7: Since we're in an open repo I'd like to include comments about what the algorithm is doing

```suggestion
  const bytes = window.crypto.getRandomValues(new Uint8Array(16));

  // https://tools.ietf.org/html/rfc4122#section-4.1.3
  // This will first bit mask away the most significant 4 bits (version octet)
  // then mask in the v4 number we only care about v4 random version at this point so (byte & 0b00001111 | 0b01000000)
  bytes[6] = bytes[6] & 15 | 64;
  
  // https://tools.ietf.org/html/rfc4122#section-4.1.1
  // This will first bit mask away the highest two bits then masks in the highest bit so (byte & 0b00111111 | 0b10000000)
  // So it will set the Msb0=1 & Msb1=0 described by the "The variant specified in this document." row in the table
  bytes[8] = bytes[8] & 63 | 128;
  
  return bytes;
```

ltrouton: Thanks 😄

modules/tinymce/src/core/test/ts/browser/EditorManagerTest.ts
ltrouton: Hmm, the baseUrl is already set here: https://github.com/tinymce/tinymce/pull/10277/files#diff-9f3c744ede87f4e7e0529225454f688bc0a477c8c9ac3d6304a8153cb733939cR29

And I tried using what mcagar does.

Maybe I am missing something

TheSpyder: This test is a bit weird because it wasn't really designed to load a full editor. But perhaps it should, since it clearly depends on it when it runs as a part of a wider test suite.

You might be able to get away with importing the actual `dom/Model` and `silver/Theme` files rather than using their `Main` imports to avoid the eslint warnings 🤔

TheSpyder (approved):

modules/tinymce/src/core/test/ts/browser/EditorManagerTest.ts
spocke: Hmm, I pulled down the branch and ran the `pageUid/editorUid` in isolation just fine? What test didn't work in isolation?

modules/tinymce/src/core/main/ts/api/Editor.ts
```diff
@@ -271,6 +279,7 @@ class Editor implements EditorObservable {
     const self = this;
 
     this.id = id;
+    this.editorUid = Id.uuidV4();
```
kimwoodfield, line 282: So every time a page is loaded with TinyMCE and we create a new instance of the Editor, we will be assigning a new UID? Are we persisting those UUIDs? If not, will that be a problem if someone just continuously keeps reloading the page and assigning themselves technically a new editor each time?

modules/tinymce/src/core/main/ts/api/EditorManager.ts
```diff
@@ -186,6 +195,7 @@ const EditorManager: EditorManager = {
     const self = this;
     let baseURL = '';
     let suffix = '';
+    self.pageUid = Id.uuidV4();
```
kimwoodfield, line 198: Similar question as before, will we be generating a new `pageUid` for each page load?

kimwoodfield (approved):

modules/tinymce/src/core/main/ts/api/Editor.ts
kimwoodfield: Discussed offline, thanks @ltrouton 👍🏻

PR #10278, opened 2025-04-09
TINY-12005: Empty styles should disable the styles button.

HAFRMO:
Related Ticket: TINY-12005

Description of Changes:
When the onSetup was added in 6.5 the previous behavior where it would disable itself was changed. This reintroduces disabling when empty.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

## Summary by CodeRabbit

- **New Features**
	- Improved the behavior of style formatting controls: if no style options are defined, the related button will now be disabled.
	- Enhanced the interface logic so that formatting menus and controls accurately reflect configuration status by considering both editability and the availability of style items.

- **Bug Fixes**
	- Added tests to ensure that the format menu and toolbar buttons are disabled when no styles are provided.


modules/tinymce/src/themes/silver/main/ts/ui/core/ControlUtils.ts
```diff
@@ -10,9 +10,9 @@ const composeUnbinders = (f: VoidFunction, g: VoidFunction): VoidFunction => ()
   g();
 };
 
-const onSetupEditableToggle = <T extends Toolbar.ToolbarButtonInstanceApi | Menu.MenuItemInstanceApi>(editor: Editor): (api: T) => VoidFunction =>
+const onSetupEditableToggle = <T extends Toolbar.ToolbarButtonInstanceApi | Menu.MenuItemInstanceApi>(editor: Editor, require?: () => boolean): (api: T) => VoidFunction =>
   onSetupEvent<T>(editor, 'NodeChange', (api) => {
-    api.setEnabled(editor.selection.isEditable());
+    api.setEnabled(editor.selection.isEditable() && (Type.isFunction(require) ? require() : true));
```
TheSpyder, line 15: I don't have a good alternative off the top of my head, but I think we need to find a better way than this.

TheSpyder (changes requested):

modules/tinymce/src/themes/silver/main/ts/ui/core/ControlUtils.ts
spocke: Maybe we could make the entire this a predicate that is passed in and defaults to `(editor: editor) => editor.selection.isEditable()` then we wouldn't need the checking if the require is a function and also the toggling could be something custom.
But the risk with that is that someone might forget to add that `editor.selection.isEditable()` call.

modules/tinymce/src/themes/silver/main/ts/ui/core/complex/StylesBespoke.ts
```diff
@@ -82,7 +82,7 @@ const createStylesMenu = (editor: Editor, backstage: UiFactoryBackstage): void =
   const menuItems = createMenuItems(backstage, getSpec(editor, dataset));
   editor.ui.registry.addNestedMenuItem('styles', {
     text: menuTitle,
-    onSetup: onSetupEditableToggle(editor),
+    onSetup: onSetupEditableToggle(editor, () => menuItems.getStyleItems().length > 0),
```
spocke, line 85: You could maybe branch here instead so that you register a `styles` menu item without a `onSetup` and a `enabled: false` flag being set.
So more like:
```
if (menuItems.getStyleItems().length > 0) {
  editor.ui.registry.addNestedMenuItem('styles', {
    text: menuTitle,
    onSetup: onSetupEditableToggle(editor, () => menuItems.getStyleItems().length > 0),
    getSubmenuItems: () => menuItems.items.validateItems(menuItems.getStyleItems())
  });
} else {
  editor.ui.registry.addNestedMenuItem('styles', {
    text: menuTitle,
    enabled: false,
    getSubmenuItems: Fun.constant([])
  });
}
```

modules/tinymce/src/themes/silver/test/ts/browser/editor/bespoke/SilverBespokeButtonsTest.ts
```diff
@@ -361,6 +361,28 @@ describe('browser.tinymce.themes.silver.editor.bespoke.SilverBespokeButtonsTest'
...
 
+  context('Empty styles provided', () => {
+    const hookStyleless = TinyHooks.bddSetup<Editor>({
+      toolbar: 'align fontfamily fontsize blocks styles',
+      base_url: '/project/tinymce/js/tinymce',
+      content_css: '/project/tinymce/src/themes/silver/test/css/content.css',
+      style_formats: [],
+    }, []);
+
+    it('TINY-12005: Empty styles should disable the buttons.', async () => {
```
spocke, line 372: Remove the `.` and make two tests one for the button and one for the menu item.

spocke (approved):

modules/tinymce/src/themes/silver/main/ts/ui/core/ControlUtils.ts
HAFRMO: Yeah, I wasn't sure how to do it better than this. As for passing in the whole predicate makes me concerned of desyncs where the default becomes something else, but some other places are never updated.

modules/tinymce/src/themes/silver/main/ts/ui/core/complex/StylesBespoke.ts
HAFRMO: That could work, if we want to do that.

HAFRMO: That'd be a problem. I'll keep it like it is until we know a better method.

modules/tinymce/src/themes/silver/main/ts/ui/core/ControlUtils.ts
TheSpyder: I have two ideas that would make me happier:
- rename the parameter to `enabledPredicate`, and perhaps made it mandatory (or defaulted to `Fun.always` since it's used 18 times).
- Stop using `onSetupEditableToggle()` for the one case where this extra predicate applies. It's a very simple function, duplicating it seems better than adding complexity.

modules/tinymce/src/themes/silver/main/ts/ui/core/complex/StylesBespoke.ts
TheSpyder: The problem is our dynamic styles menu, so the check for whether this nested item is enabled depends on the cursor context. This is really the only way to do it, unless we get rid of the `onSetupEditableToggle` refactoring and just inline that function in this one place.

Which now that I've said it twice that seems like the better approach...

TheSpyder (approved):

kimwoodfield (approved):

PR #10282, opened 2025-04-14
TINY-11997: fix fake caret when tabbing inside a caption

lorenzo-pomili:
Related Ticket: TINY-11997

Description of Changes:
the problem was that when we tab inside a `figurecaption` there were no code to manage the hide of the fake caret, now on `NodeChange` there is a check on elements contained in a CEF element

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


.changes/unreleased/tinymce-TINY-11997-2025-04-14.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: Tabbing inside a `figcaption` didn't remove the fakeCaret.
```
TheSpyder, line 3: Implementation details don't really belong in changelogs 🤔 
```suggestion
body: Tabbing inside a `figcaption` element no longer shows two text insertion carets.
```

modules/tinymce/src/core/main/ts/SelectionOverrides.ts
```diff
@@ -160,6 +160,15 @@ const SelectionOverrides = (editor: Editor): SelectionOverrides => {
       }
     });
 
+    editor.on('NodeChange', (e) => {
```
TheSpyder, line 163: Aren't there other events already used in this file that would be more appropriate? This will run a lot more often than when tab is pressed.

Or can we add a simple "is fake caret set" check to reduce how often this code does anything significant? It could probably be a wrapper around `lastVisualCaret.isSet()` 🤔

.changes/unreleased/tinymce-TINY-11997-2025-04-14.yaml
lorenzo-pomili: done

modules/tinymce/src/core/main/ts/SelectionOverrides.ts
TheSpyder: Yeah it would need to be keypress or similar. This will do fine, thank you

TheSpyder (approved):

modules/tinymce/src/core/main/ts/SelectionOverrides.ts
lorenzo-pomili: I tried to implement it in the [keypress](https://github.com/tinymce/tinymce/blob/main/modules/tinymce/src/core/main/ts/SelectionOverrides.ts#L138) of this file, but it seems prevented in this case and I don't know why :sweat_smile:

modules/tinymce/src/core/test/ts/browser/caret/FakeCaretImageCaptionTest.ts
```diff
@@ -0,0 +1,37 @@
...
+  const isFirefox = browser.isFirefox();
+  const hook = TinyHooks.bddSetupLight<Editor>({
+    base_url: '/project/tinymce/js/tinymce',
+    plugins: 'image',
+    image_caption: true
+  }, [ ImagePlugin ]);
+
+  // skiped on FireFox since `setRawSelection` seems not to work on it
+  (isFirefox ? it.skip : it)('TINY-11997: should hide after tabbing inside figcaption', async () => {
+    const editor = hook.editor();
```
hamza0867, line 20: Another version of this would be:

```suggestion
  
  it('TINY-11997: should hide after tabbing inside figcaption', async function () => {
    // skiped on FireFox since `setRawSelection` seems not to work on it
    if(isFirefox) { this.skip(); }
    const editor = hook.editor();
```

But I don't know if it's more readable or not

hamza0867 (approved):

modules/tinymce/src/core/test/ts/browser/caret/FakeCaretImageCaptionTest.ts
lorenzo-pomili: yes, it's better in your way, reafactored

modules/tinymce/src/core/main/ts/SelectionOverrides.ts
spocke: I think this should run on `focusin` that is fired when the focus is moved from outside the element to inside the element. 

This seems to fix it. But my guess is that parts of the SelectionOverride maybe `AfterSetSelectionRange` is the code that needs to be executed on `focusin` so that it re-freshes the fake caret state. Needs some digging.
```
editor.on('focusin',  () => {
    editor.selection.setRng(editor.selection.getRng());
});
```

spocke (changes requested):

modules/tinymce/src/core/main/ts/SelectionOverrides.ts
lorenzo-pomili: I changed the event from `NodeChange` to `focusin` but using `editor.selection.setRng(editor.selection.getRng())` instead of the current code, solve this bug, but it breaks many tests, so I would stay with the current code but on `focusin` instead of `NodeChange`

modules/tinymce/src/core/main/ts/SelectionOverrides.ts
```diff
@@ -160,6 +160,16 @@ const SelectionOverrides = (editor: Editor): SelectionOverrides => {
       }
     });
 
+    editor.on('focusin', (e) => {
+      if (
+        fakeCaret.isShowing() &&
+        NodeType.isContentEditableFalse(e.target.parentNode) &&
```
spocke, line 166: Not sure we can just assume that what we are tabbing into has a parent that is has content editable false at least we should check the `dom.isEditable` state that will check all the parents.
Also when we are tabbing into something that something could also have non-editables as the first item.

Like:
```
        <figure class="image">
          <img src="...">
          <figcaption><div contenteditable="false">CEF</div></figcaption>
        </figure>
```

So what about:
```
    editor.on('focusin', () => {
      if (fakeCaret.isShowing()) {
        fakeCaret.hide();

        const rng = setElementSelection(editor.selection.getRng(), true);
        if (rng) {
          editor.selection.setRng(rng);
        }
      }
    });
```

Is that breaks tests that tests?
I also noticed that Firefox is doing something weird if you tab into the field when the fake caret is there but if you tab from some other content it works correctly.

lorenzo-pomili: @spocke I couldn't find a better way to fix both tests and not having the check for the parent as CEF, since the bug is specific for caption I would say that we can go with this solution now, and if some different cases that requires to remove that check in the future appear we can remove it, what do you think about it?

modules/tinymce/src/core/main/ts/SelectionOverrides.ts
```diff
@@ -160,6 +160,17 @@ const SelectionOverrides = (editor: Editor): SelectionOverrides => {
       }
     });
 
+    editor.on('focusin', (e) => {
+      if (fakeCaret.isShowing() && !NodeType.matchNodeNames([ 'body' ])(e.target) && !editor.dom.isEditable(e.target.parentNode)) {
```
spocke, line 164: It's safer to do a `e.target == editor.getBody()` since for inline mode `body` is not the name of the root editor container.

- [ ] Change editor root container check

modules/tinymce/src/core/test/ts/browser/caret/FakeCaretImageCaptionTest.ts
```diff
@@ -0,0 +1,72 @@
...
+  const browser = PlatformDetection.detect().browser;
+  const isFirefox = browser.isFirefox();
+  const hook = TinyHooks.bddSetupLight<Editor>({
+    base_url: '/project/tinymce/js/tinymce',
+    plugins: 'image',
+    image_caption: true
+  }, [ ImagePlugin ]);
+
+  it('TINY-11997: should hide after tabbing inside figcaption', async function () {
+    // skiped on FireFox since `setRawSelection` seems not to work on it
```
spocke, line 19:
```suggestion
    // skipped on FireFox since `setRawSelection` seems not to work on it
```

spocke (approved):

modules/tinymce/src/core/test/ts/browser/caret/FakeCaretImageCaptionTest.ts
lorenzo-pomili: fixed

PR #10285, opened 2025-04-15
TINY-11974: Implement Unified UID System

kimwoodfield:

Description of Changes:
The `UserLookup` API is responsible for providing a unified user identification system across TinyMCE plugins.

When a plugin needs user information, it first checks a local cache before fetching externally for information about the user via an asynchronous function provided by the integrator. This reduces redundant network requests and provides consistent user data across the Editor instance.

Currently, multiple plugins implement their own UID systems, requiring integrators to implement multiple user lookup functions. This leads to duplicated code and inconsistent implementations.

This PR introduces the following functionality via this API:
- Centralized caching mechanism
- Standard `User` interface with `id`, `name`, `avatar`, and `description` fields
- Allowance for plugins to extend user data with custom properties to support various use-cases 
- Supports potential authentication systems through a configurable `fetch_user_by_id` option

The idea of this unified approach is to simplify integration, reduce code duplication, and provide a foundation for future features.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/tinymce/src/core/main/ts/api/Options.ts
```diff
@@ -904,6 +904,20 @@ const register = (editor: Editor): void => {
     default: true
   });
 
+  registerOption('current_user_id', {
+    processor: 'string',
+  });
+
+  registerOption('fetch_user_by_id', {
+    processor: 'function',
+    default: () => Promise.reject()
```
spocke, line 913: Adda reject reason `Promise.reject(new Error('fetch_user_by_id option not defined')` or something.

modules/tinymce/src/core/main/ts/api/Options.ts
```diff
@@ -904,6 +904,20 @@ const register = (editor: Editor): void => {
     default: true
   });
 
+  registerOption('current_user_id', {
+    processor: 'string',
```
spocke, line 908: I guess it is a good thing to enforce the user to be available before we init the editor so this is not async.

modules/tinymce/src/core/main/ts/api/Options.ts
```diff
@@ -904,6 +904,20 @@ const register = (editor: Editor): void => {
...
+  });
+
+  registerOption('fetch_user_by_id', {
+    processor: 'function',
+    default: () => Promise.reject()
+  });
+
+  registerOption('user_cache', {
```
spocke, line 916: Why do we need to expose the user cache? I think it's better that this is an internal concern then we add functions to remove users by ids or flush the cache?
If we want to support pre-populating the cache we could add a `addUser` that schema validates and puts a user into the cache.

modules/tinymce/src/core/main/ts/lookup/UserLookup.ts
```diff
@@ -0,0 +1,90 @@
...
+ */
+
+type UserId = string;
+
+export interface User {
+  id: UserId;
+  name?: string;
+  avatar?: string;
+  description?: string;
+  [key: string]: any;
```
spocke, line 32: Could this extra meta information be put in it's own property since if we add new things we could never really schema validate them. The properties we know about now we could schema validate and say we need some new property in the future then we could schema validate that as well if we separate these things.
Or could we make this all a interface of required properties do we need to extra flexibility here of allowing any property?

modules/tinymce/src/core/main/ts/lookup/UserLookup.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  const store = (user: User, userId: UserId) => {
+    userCache[userId] = user;
+  };
+
+  const fetchUserById = (userId: UserId): Promise<User> =>
+    new Promise((resolve, reject) =>
+      lookup(userId)
+        .fold(() =>
+          Options.getFetchUserById(editor)(userId)
```
spocke, line 68: This needs a lot more error handling. What if the user is not returning a promise in the function, what if the value in the promise doesn't have the expected properties. I would suggest at least doing some basic validations of the types I see that we already include boulder in the `autocompleter` in core so we could use that then we get pretty good error messages.

We have a `Type.isPromiseLike` so one could wrap that call in something like:
```
const fetchUserById = (editor: Editor, userId: UserId) {
  const fetchUserById = Options.getFetchUserById(editor);
  const result = fetchUserById(userId);
  if (Type.isPromiseLike(result)) {
    return result.then(schemaValidateWithBoulder});
  } else {
    return fail('fetch_user_by_id should return a promise';
  }
}
```

modules/tinymce/src/core/main/ts/lookup/UserLookup.ts
```diff
@@ -0,0 +1,90 @@
...
+    Obj.each(Options.getUserCache(editor), store);
+  });
+
+  return {
+    getCurrentUserId,
+    fetchUserById,
+  };
+};
+
+export default UserLookup;
```
spocke, line 90: Avoid `export default` if we don't need to.

modules/tinymce/src/core/test/ts/browser/lookup/UserLookupTest.ts
```diff
@@ -0,0 +1,157 @@
...
+    fetch_user_by_id: (userId: string): Promise<User> => {
+      return new Promise((resolve) => {
+        setTimeout(() => {
+          resolve({
+            id: userId,
+            name: 'Test User',
+            avatar: 'test-avatar.png',
+            description: 'Test Description',
+          } as User);
+        }, 100);
```
spocke, line 22: This is a bit high we just need to make sure it's async right so 0 would be fine as well and would make tests run faster.

modules/tinymce/src/core/test/ts/browser/lookup/UserLookupTest.ts
```diff
@@ -0,0 +1,157 @@
...
+    assert.equal(currentUserId, 'test-user-1', 'Should return the configured user ID');
+  });
+
+  it('Should fetch user information', async () => {
+    const editor = hook.editor();
+    const userId = 'test-user-1';
+
+    const user = await editor.userLookup
+      .fetchUserById(userId)
+      .then(Fun.identity)
```
spocke, line 39: Why the identity here?

modules/tinymce/src/core/test/ts/browser/lookup/UserLookupTest.ts
```diff
@@ -0,0 +1,157 @@
...
+    const secondCall = await editor.userLookup.fetchUserById(userId);
+
+    assert.deepEqual(firstCall, secondCall, 'Should return same user data from cache');
+  });
+
+  it('Should handle errors gracefully', async () => {
+    const editor = hook.editor();
+
+    // Override fetch_user_by_id to simulate an error
+    editor.options.set('fetch_user_by_id', () => Promise.reject('User not found'));
```
spocke, line 67: I wonder if we should `normalize` the string to an error type since promises are weird so maybe we could enforce them to not be as weird by type guards here. Since we we add validations on resolved values we could add normalization on rejected values. Something like `(err) => err instanceof Error ? err : new Error(String(err))` or something.

modules/tinymce/src/core/main/ts/api/Options.ts
kimwoodfield: Exactly and it might become a bit troublesome if we are first waiting for an `id` to be returned from a promise only to be passing it to _another_ promise. 

I'd imagine that the most common use case would be that once their page loads, they probably have some authentication or something so before the Editor is even rendered on the screen they would already have their user information available to provide it immediately.

modules/tinymce/src/core/main/ts/api/Options.ts
kimwoodfield: Exposing the user cache is not a requirement, it is just an easy-win to prevent making a potentially unnecessary HTTP request. If they choose to provide it then we can just lookup the user from that rather than having to do so externally via the function provided to `fetch_user_by_id`.

However, I think you make a good point about schema validation. It's important that this API is flexible but we already have an expectation of what we expect to be stored in the cache so I don't think adding some basic validation would harm anyone.

@TheSpyder @MitchC1999 Do you have any thoughts on this one? Happy to take it offline if needed.

modules/tinymce/src/core/test/ts/browser/lookup/UserLookupTest.ts
kimwoodfield: I left it there by mistake so thanks for spotting it.

modules/tinymce/src/core/main/ts/lookup/UserLookup.ts
kimwoodfield: I've followed the convention from other APIs like `EditorUpload` which also uses `export default` and I think that provides a cleaner type reference in the `Editor` class. Otherwise, we'd need to do something like `public userLookup!: UserLookup.UserLookup` to access the definition which looks a bit odd. If you think it's still worth doing then I'm happy to change it though.

modules/tinymce/src/core/main/ts/lookup/UserLookup.ts
kimwoodfield: I would prefer the former since we are trying to ensure that this API is flexible and can be reused across plugins that already have a mix of required and optional fields from this definition. I think with your suggestion, we could adjust it such that it appears like this:

```ts
export interface User {
  id: UserId;
  name?: string;
  avatar?: string;
  description?: string;
  metadata?: Record<string, any>;
}
```

And then just schema validate the ones we know about if that's the idea you had but I'd like to hear what the others think about this one too.

As mentioned, happy to take this one offline if needed.

kimwoodfield:
I'm going to move this to a draft until we've ironed out some more details involving cache and validation.

modules/tinymce/src/core/main/ts/lookup/UserLookup.ts
TheSpyder: The only required property is `id`, and if the other fields are wrong it will break the rendering of user information, so validation seems a bit pointless. We'd just sanitise them before rendering to avoid XSS if a user managed to save HTML in their profile.

I'm aiming for ease of use here, which to me means allowing integrators to add extra fields directly. I'm not sure what we'd call an extra property; things like `avatar` and `description` are already "metadata".

modules/tinymce/src/core/main/ts/lookup/UserLookup.ts
TheSpyder: It looks like `export default` was the style chosen when the exported types were added, but that was 6 years ago
https://github.com/tinymce/tinymce/commit/bb669417e9fc35f7822bcf655fda250e627777d9#diff-f97965c95c92e1a26bb327b205bdc67609f3090d8bfa1d012cb1f04563749580

To avoid `UserLookup.UserLookup` we would not use `import * as UserLookup from '../lookup/UserLookup'` and instead perhaps `import { UserLookup } from '../lookup/UserLookup'`?

modules/tinymce/src/core/main/ts/lookup/UserLookup.ts
```diff
@@ -0,0 +1,90 @@
...
+  const getCurrentUserId = (): string => Options.getCurrentUserId(editor);
+
+  editor.on('init', () => {
+    Obj.each(Options.getUserCache(editor), store);
+  });
+
+  return {
+    getCurrentUserId,
+    fetchUserById,
+  };
```
TheSpyder, line 87: Since we're planning to make this transition optional at first, we probably need some way for plugins to identify that the uid system has not been configured yet so they can fall back to check for the old configuration options.

modules/tinymce/src/core/test/ts/browser/lookup/UserLookupTest.ts
TheSpyder: It probably depends what we plan to do with the rejection. But the general advice seems to be yes we should do that.
https://stackoverflow.com/a/26020867/7546

modules/tinymce/src/core/main/ts/lookup/UserLookup.ts
spocke: Since this is internal you can make a function that creates the user lookup like `createUserLookup(): UserLookup` then you export the interface and the function. I think the `export default` is kind of legacy but we need it for some of the external modules I think like when we expose it as a thing on the public API but we are only exporting this as a property not as a constructor so I think it should be possible to avoid the `export default`.

modules/tinymce/src/core/main/ts/lookup/UserLookup.ts
spocke: I don't think it would be pointless would be good to know that they don't return a number or a object or null or something on properties we expect to be strings. Since we want everyone to migrate to this API it would be useful to make it stupid proof. If we have the meta data as a separate thing we could then display warnings if they specify properties we don't expect to be there and that would detect typos in their setup not sure if boulder can do that though or if it just ignores unknown stuff.

modules/tinymce/src/core/main/ts/lookup/UserLookup.ts
spocke: Wouldn't that check just be a null/undefined check on that new option? We don't need a special API for that.

modules/tinymce/src/core/main/ts/api/Options.ts
kimwoodfield: Marking as closed, we've decided not to expose the `user_cache`.

modules/tinymce/src/core/main/ts/lookup/UserLookup.ts
TheSpyder: I am against strict validation that bans extra properties. It just makes updating the API harder, if a newer object is passed to an old version of the editor it will start throwing warnings unnecessarily.

Which I suppose means I'm arguing myself into a corner where we need the extra property. Let's go with `custom`, and make sure the documentation notes that this is where integrators should place custom properties not supported by the editor.

kimwoodfield: Added an additional `custom` field to the `User` object to capture extra metadata.

TheSpyder:
I updated `EPIC-134` to the latest `main` which should remove those extra commits but github hasn't refreshed this PR yet. That's annoying.

modules/tinymce/src/core/main/ts/lookup/UserLookup.ts
```diff
@@ -0,0 +1,174 @@
...
+}
+
+export interface UserLookup {
+  /**
+   * Retrieves the current user ID from the editor.
+   *
+   * @method getCurrentUserId
+   * @return {string} The current user ID.
+   */
+  getCurrentUserId: () => UserId;
```
TheSpyder, line 41: Since we ask developers to set this synchronously, should we make it a constant reference? We could `Object.freeze()` the value returned by `createUserLookup()` to stop integrators mutating it

MitchC1999:
> I updated `EPIC-134` to the latest `main` which should remove those extra commits but github hasn't refreshed this PR yet. That's annoying.

This branch was rebased on main before the epic branch was updated, so it needs to be rebased again.

modules/tinymce/src/core/main/ts/lookup/UserLookup.ts
```diff
@@ -0,0 +1,174 @@
...
+  fetchUsers: (userIds: UserId[]) => Promise<User>[];
+}
+
+const isUserObject = (val: unknown): val is User =>
+  Type.isObject(val)
+  && val.hasOwnProperty('id')
+  && Type.isNonNullable((val as User).id) && Type.isString((val as User).id as UserId)
+  && !(Type.isNonNullable((val as User).name) && !Type.isString((val as User).name))
+  && !(Type.isNonNullable((val as User).avatar) && !Type.isString((val as User).avatar))
+  && !(Type.isNonNullable((val as User).description) && !Type.isString((val as User).description));
```
TheSpyder, line 60: We definitely need to have better documentation about how we do object validation. Boulder isn't the best solution, it probably needs a good rewrite, but it's what we have.

It's primarily used in the UI, where the usage is refactored heavily, but here's one of the demos that should be simple enough to follow
https://github.com/tinymce/tinymce/blob/16676bc48409eb440ad1e2bb3021d62c24d975fa/modules/alloy/src/demo/ts/ephox/alloy/demo/forms/DemoRenders.ts#L56-L67

Use `StructureSchema.asRaw('failure message', schema, value)` to return a `Result<Error, T>` (where `T` is the type of `value`), which you could map through an array of fetched values and then collect the failures to console log as a single error message.

We have field types that must be strings, I don't think we have a "non-empty string" field type but if someone wants to set an empty string as their userid that's their problem.

modules/tinymce/src/core/main/ts/lookup/UserLookup.ts
```diff
@@ -0,0 +1,174 @@
...
+        pending.reject(error);
+        pendingResolvers.delete(userId);
+      }
+    });
+  };
+
+  const fetchUsers = (userIds: UserId[]): Promise<User>[] => {
+    if (Arr.forall(userIds, (userId) => !isValidUserId(userId))) {
+      throw new Error('Invalid userId provided. All userIds must be a non-empty string.');
+    }
```
TheSpyder, line 106: You've written this backwards, it will only fail if _all_ of the `userId` array is invalid.

Either:
* get rid of the check - most of the time this will only be called by our plugins, where the userid comes from the content
* add a test using an array with some empty strings and some non-empty strings

modules/tinymce/src/core/main/ts/api/Options.ts
```diff
@@ -904,6 +904,14 @@ const register = (editor: Editor): void => {
     default: true
   });
 
+  registerOption('current_user_id', {
```
TheSpyder, line 907: Arguably the word `current` is superfluous
```suggestion
  registerOption('user_id', {
```

modules/tinymce/src/core/main/ts/lookup/UserLookup.ts
```diff
@@ -0,0 +1,174 @@
...
+  const users = Arr.filter(items, (item) => isUserObject(item));
+  return users;
+};
+
+const isValidUserId = (userId: unknown): userId is UserId =>
+  Type.isString(userId)
+  && userId.length > 0;
+
+const UserLookup = (editor: Editor): UserLookup => {
+  const userCache: Record<UserId, Promise<User>> = {};
```
TheSpyder, line 80: You used a `Map` for the pending resolvers, which is good, but perhaps we should use one for the cache as well?

TheSpyder:
There had been no changes to `EPIC-134` so a merge made it equal to `main`. In that state a rebase would've had no effect and reported `already up to date`.

Screenshot 2025-04-30 at 10 41 50 am

The fix was to change the PR base branch to `main`, even though that is the same commit, and then back to the epic branch.

modules/tinymce/src/core/main/ts/lookup/UserLookup.ts
```diff
@@ -0,0 +1,174 @@
...
+
+      store(newPromise, userId);
+    });
+
+    if (uncachedIds.length > 0) {
+      const fetchUsersFn = Options.getFetchUsers(editor);
+
+      if (!Type.isFunction(fetchUsersFn)) {
+        throw new Error('fetch_users option is not defined');
+      }
```
TheSpyder, line 125: Should this be an option validator instead?

If fetch is unavailable, i.e. the integrator has not configured the UID system, we should probably check that at the top of the function. If we throw an error here, all those pending promises will become memory leaks if they are never resolved.

modules/tinymce/src/core/main/ts/lookup/UserLookup.ts
```diff
@@ -0,0 +1,174 @@
...
+  && !(Type.isNonNullable((val as User).description) && !Type.isString((val as User).description));
+
+const validateResponse = (items: unknown): User[] => {
+  if (!Array.isArray(items)) {
+    throw new Error('fetch_users must return an array');
+  }
+
+  if (items.length === 0) {
+    return [];
+  }
```
TheSpyder, line 69: `Arr.filter` will have no effect if the array is empty. You can delete this.

modules/tinymce/src/core/test/ts/browser/lookup/UserLookupTest.ts
```diff
@@ -0,0 +1,296 @@
...
+
+  it('Should throw an error for empty string ids', async () => {
+    const editor = hook.editor();
+    const userIds = [ '' ];
+    try {
+      await Promise.all(editor.userLookup.fetchUsers(userIds));
+      assert.fail('Should throw error for empty string ID');
+    } catch (error) {
+      assert.isDefined(error, 'Should handle empty string ID with error');
+    }
```
TheSpyder, line 68: We have `Chai` available, I think it has helper methods for "should throw an error"? I'm not sure if we have existing tests to pull examples from.

TheSpyder (commented):
This is much better! I think it's nearly ready for a wider review.

modules/tinymce/src/core/main/ts/lookup/UserLookup.ts
kimwoodfield: I ended up using `Object.defineProperty` instead of `Object.freeze` to make only `getCurrentUserId` read-only while keeping `fetchUsers` modifiable. I took the idea from a PR I recently saw Luke make. Let me know if you'd prefer `Object.freeze`.

modules/tinymce/src/core/main/ts/api/Options.ts
kimwoodfield: Done!

modules/tinymce/src/core/main/ts/lookup/UserLookup.ts
TheSpyder: We only need to freeze if we're exposing it as a constant. A function doesn't give the impression that an integrator could change the userID by assigning to it.

But yes under the hood `freeze` uses that `defineProperty` technique on every object property (and also blocks new property additions).

modules/tinymce/src/core/main/ts/lookup/UserLookup.ts
```diff
@@ -0,0 +1,201 @@
...
+  FieldSchema.option('custom')
+]);
+
+const transformResult = (user: any): User => {
+  const customValue = user.custom?.getOrUndefined();
+  return {
+    id: user.id,
+    name: user.name?.getOrUndefined(),
+    avatar: user.avatar?.getOrUndefined(),
+    description: user.description?.getOrUndefined(),
```
TheSpyder, line 69: This is not quite the same thing. I don't know if we want properties that weren't specified to be `undefined`, they should be missing. The difference is subtle but can be important.

One benefit to your change is that it fully discards any extra properties on the returned object that we don't want. But perhaps we need to look for methods that avoid setting a property for `None` values.

Something like `Optionals.cat` but for objects 😂

modules/tinymce/src/core/main/ts/lookup/UserLookup.ts
```diff
@@ -0,0 +1,201 @@
...
+const validateResponse = (items: unknown): User[] => {
+  if (!Array.isArray(items)) {
+    throw new Error('fetch_users must return an array');
+  }
+
+  const results = Arr.map(items, (item) =>
+    StructureSchema.asRaw<User>('Invalid user object', userSchema, item)
+  );
+
+  const { pass: valid, fail: invalid } = Arr.partition(results, (result) => result.isValue());
```
TheSpyder, line 83: Would it surprise you to learn we do this enough to make a method for it 😉 

```suggestion
  const { pass: valid, fail: invalid } = Results.partition(results);
```

modules/tinymce/src/core/main/ts/lookup/UserLookup.ts
```diff
@@ -0,0 +1,201 @@
...
+
+  const errors = Arr.map(invalid, (result, idx) =>
+    result.fold(
+      (schemaError) => `User at index ${idx}: ${StructureSchema.formatError(schemaError)}`,
+      () => {
+        throw new Error('Unreachable'); // This should never happen due to partioning.
+      }
```
TheSpyder, line 90: I need to think about this some more. If you add this as a `.mapError()` after the `StructureSchema.asRaw()` call it might help, but then we still want to collect the error values at the end 🤔

TheSpyder (commented):
I didn't finish my review before the end of the day, sorry, here's what I have so far

modules/tinymce/src/core/main/ts/lookup/UserLookup.ts
kimwoodfield: Very nice!

modules/tinymce/src/core/main/ts/lookup/UserLookup.ts
kimwoodfield: Is this something that we still need to be concerned with? We were able to remove this code when we switched to `Results.partition` since we don't need to manually fold anymore and the errors are already available directly.

modules/tinymce/src/core/test/ts/browser/lookup/UserLookupTest.ts
```diff
@@ -0,0 +1,325 @@
...
+    const userIds = [ '' ];
+    const [ userPromise ] = editor.userLookup.fetchUsers(userIds);
+
+    try {
+      await userPromise;
+      expect.fail('Should throw error for empty string ID');
+    } catch (err: unknown) {
+      expect(err).to.be.instanceOf(Error);
+      expect(err).to.have.property('message', 'User  not found');
+    }
```
TheSpyder, line 77: This isn't what I asked for when I said switch to chai. Chai has an assertion to handle this for you.
https://www.chaijs.com/api/bdd/#method_throw

Although maybe it's not that simple for promises, there's an addon library for it 🤔 
https://www.chaijs.com/plugins/chai-as-promised/

modules/tinymce/src/core/main/ts/lookup/UserLookup.ts
TheSpyder: Oh! maybe that's what I meant when I started to write a comment on this code. Thank you for checking.

modules/tinymce/src/core/main/ts/lookup/UserLookup.ts
```diff
@@ -0,0 +1,196 @@
...
+  if (errors.length > 0) {
+    const formattedErrors = Arr.map(errors, (error, idx) =>
+      `User at index ${idx}: ${StructureSchema.formatError(error)}`
+    );
+
+    // eslint-disable-next-line no-console
+    console.warn('User validation errors:\n' + formattedErrors.join('\n'));
+  }
+
+  return Arr.map(values, (result) => transformResult(result));
```
TheSpyder, line 94: Now that there's no extra arguments...
```suggestion
  return Arr.map(values, transformResult);
```

modules/tinymce/src/core/main/ts/lookup/UserLookup.ts
```diff
@@ -0,0 +1,196 @@
...
+
+const UserLookup = (editor: Editor): UserLookup => {
+  const userCache = new Map<UserId, Promise<User>>();
+  const pendingResolvers = new Map<UserId, {
+    resolve: (user: User) => void;
+    reject: (error: unknown) => void;
+  }>();
+
+  const lookup = (userId: UserId) =>
+    Optional.from(userCache.get(userId)).map(Fun.identity);
```
TheSpyder, line 105: uhhh `.map(Fun.identity)` does literally nothing 🤔 what were you intending to happen here?
```suggestion
    Optional.from(userCache.get(userId));
```

modules/tinymce/src/core/test/ts/browser/lookup/UserLookupTest.ts
```diff
@@ -0,0 +1,325 @@
...
+            id.startsWith('test-user-') ? [ createMockUser(id) ] : []
+          );
+
+          resolve(users);
+        }, 0);
+      });
+    },
+  });
+
+  it('Should return the configured user ID', () => {
```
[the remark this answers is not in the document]
TheSpyder: Only if the `it` block is specific to the test. Otherwise it belongs in the `context` block, or in this case probably the `describe`.

Bedrock will print all parent descriptions when reporting test failures.

modules/tinymce/src/core/main/ts/lookup/UserLookup.ts
```diff
@@ -60,14 +62,34 @@ const userSchema = StructureSchema.objOf([
...
+  };
+
   return {
     id: user.id,
-    name: user.name?.getOrUndefined(),
-    avatar: user.avatar?.getOrUndefined(),
-    description: user.description?.getOrUndefined(),
-    ...(customValue !== undefined && { custom: customValue })
+    ...objectCat(optionalFields)
   };
```
TheSpyder, line 93: I think you got yourself a bit confused here. `extractOptionalField` is just `Optional.from()`.

But the whole point of my suggestion to create `objectCat` was to replace all this. Literally the entire `transformResult` function could just be a single call to `objectCat`. `transformResult` itself may not even be needed, just call `objectCat` at the call site instead.

modules/tinymce/src/core/test/ts/browser/lookup/UserLookupTest.ts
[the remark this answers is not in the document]
TheSpyder: I mixed that up a bit - we generally use `describe` for the test name, not JIRA id. So either add a `context` or yes put it in every `it` block.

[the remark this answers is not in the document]
kimwoodfield: I've logically split the tests into two separate contexts, one for the `getCurrentUserId` and the other for `fetchUsers` and both have the corresponding ticket ID.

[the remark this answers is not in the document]
spocke: I would not put the ID in the in the context since what happens if there is a bug that you want to put into a new it with a different ID in that context then you have conflicting ticket id's better to copy the ID to the it calls. Also having the id in the it shows the ticket id in test failures, when you search for the label and its more consistent with our other tests.

modules/tinymce/src/core/main/ts/lookup/UserLookup.ts
```diff
@@ -0,0 +1,214 @@
...
+  const store = (user: Promise<User>, userId: UserId) => {
+    userCache.set(userId, user);
+  };
+
+  const finallyReject = (userId: UserId, error: Error) => {
+    return Optional.from(pendingResolvers.get(userId)).each(({ reject }) => {
+      reject(error);
+      pendingResolvers.delete(userId);
+    });
+  };
```
MitchC1999, line 134:
```suggestion
  const finallyReject = (userId: UserId, error: Error) =>
    Optional.from(pendingResolvers.get(userId)).each(({ reject }) => {
      reject(error);
      pendingResolvers.delete(userId);
    });
```
This also returns void anyway.

modules/tinymce/src/core/main/ts/lookup/UserLookup.ts
```diff
@@ -0,0 +1,214 @@
...
+
```
MitchC1999, line 135:
```suggestion
const finallyResolve = (userId: UserId, user?: User) =>
    Optional.from(pendingResolvers.get(userId)).each(({ resolve }) => {
      resolve(Optional.from(user).getOr({ id: userId }));
      pendingResolvers.delete(userId);
    });
```
Should refactor this resolve call as well.

modules/tinymce/src/core/main/ts/lookup/UserLookup.ts
```diff
@@ -0,0 +1,214 @@
...
+              const foundUserIds = new Set(Arr.map(users, (user) => user.id));
+
+              // Resolve found users
+              Arr.each(users, (user) => {
+                const pending = pendingResolvers.get(user.id);
+                if (pending) {
+                  pending.resolve(user);
+                  pendingResolvers.delete(user.id);
+                }
+              });
```
MitchC1999, line 176:
```suggestion
              Arr.each(users, (user) => finallyResolve(user.id, user));
```
and here

modules/tinymce/src/core/main/ts/lookup/UserLookup.ts
```diff
@@ -0,0 +1,214 @@
...
+              );
+            });
+        }
+      );
+    }
+
+    return Arr.map(userIds, (userId) => {
+      const userPromise = lookup(userId);
+      return userPromise.getOr(Promise.resolve({ id: userId }));
+    });
```
MitchC1999, line 200:
```suggestion
    return Arr.map(userIds, (userId) => lookup(userId).getOr(Promise.resolve({ id: userId })));
```

modules/tinymce/src/core/test/ts/browser/lookup/UserLookupTest.ts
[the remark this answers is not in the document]
kimwoodfield: @spocke That makes sense but the main issue I have is that we don't add the original ticket ID to every `it` case in our other tests like `EditorUpload` for example or other APIs so if we start doing that here should we be updating it everywhere in the same way? I don't really mind where we do it as long as it's consistent.

https://github.com/tinymce/tinymce/blob/main/modules/tinymce/src/core/test/ts/browser/EditorUploadTest.ts#L101

modules/tinymce/src/core/main/ts/lookup/UserLookup.ts
```diff
@@ -0,0 +1,214 @@
...
+      Optional.from(Options.getFetchUsers(editor)).fold(
+        () => {
+          Arr.each(uncachedIds, (userId) => {
+            const pending = pendingResolvers.get(userId);
+            if (pending) {
+              pending.resolve({ id: userId });
+              pendingResolvers.delete(userId);
+            }
+          });
+        },
```
MitchC1999, line 162:
```suggestion
        () => Arr.each(uncachedIds, finallyResolve),
```
So it can be used here

kimwoodfield: Nice! I ended up going with `(user) => finallyResolve(user)` because `Arr.each` would pass an index number as a second parameter which isn't what `finallyResolve` expects leading to a type mismatch so we have to just pass it what it needs but still does the trick.

MitchC1999 (approved):

shanmen-tiny (approved):

modules/tinymce/src/core/test/ts/browser/lookup/UserLookupTest.ts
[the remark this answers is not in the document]
spocke: I usually tend to add it when working on say a bug or something that was not part of the original feature. So unsure if it makes sense to add that now to new code. I guess we might have different standards on when we add these things.

modules/tinymce/src/core/test/ts/browser/lookup/UserLookupTest.ts
```diff
@@ -0,0 +1,345 @@
...
+    const editor = hook.editor();
+    const userId = 'test-user-1';
+
+    const [ firstPromise ] = editor.userLookup.fetchUsers([ userId ]);
+    const [ secondPromise ] = editor.userLookup.fetchUsers([ userId ]);
+
+    await expect(firstPromise).to.eventually.deep.equal(
+      await secondPromise,
+      'Should return same data for subsequent requests'
+    );
```
TheSpyder, line 125: You don't need deep equal. This just asserts the _contents_ are the same, which it would be even if they were not cached.

The check should be just `firstPromise === secondPromise` (or the chai equivalent).

modules/tinymce/src/core/test/ts/browser/lookup/UserLookupTest.ts
```diff
@@ -0,0 +1,345 @@
...
+    },
+  });
+
+  it('TINY-11974: Should return the configured user ID', () => {
+    const editor = hook.editor();
+    const currentUserId = editor.userLookup.getUserId();
+    expect(currentUserId).to.equal('test-user-1', 'Should return the configured user ID');
+  });
+
+  it('TINY-11974: Should have frozen getUserId but mutable fetchUsers', () => {
```
TheSpyder, line 48: You don't check `fetchUsers` at all. But I don't think you need to.

I don't think freezing a function makes a lot of sense. I expected this to just be a string, since the value is required on editor startup, and _then_ it makes sense to freeze it.

modules/tinymce/src/core/test/ts/browser/lookup/UserLookupTest.ts
```diff
@@ -0,0 +1,345 @@
...
+
+  it('TINY-11974: Should fetch users and return array of promises', () => {
+    const editor = hook.editor();
+    const userIds = [ 'test-user-1' ];
+
+    const promises = editor.userLookup.fetchUsers(userIds);
+
+    expect(promises).to.be.an('array', 'Should return an array of promises');
+    expect(promises).to.have.lengthOf(1, 'Should return array with one promise');
+    expect(promises[0]).to.be.instanceOf(Promise, 'Should return a Promise');
```
TheSpyder, line 70: Testing the type checker is generally pointless. Unless someone edits the API type signature, which would be very obvious in a PR, the array and promise type checks can't fail (or the test wouldn't compile).

modules/tinymce/src/core/test/ts/browser/lookup/UserLookupTest.ts
```diff
@@ -0,0 +1,345 @@
...
+    const editor = hook.editor();
+    const nonExistentId = 'non-existent-user';
+    const errorMessage = `User ${nonExistentId} not found`;
+
+    const [ firstPromise ] = editor.userLookup.fetchUsers([ nonExistentId ]);
+    await expect(firstPromise).to.eventually.be.rejectedWith(errorMessage);
+
+    // Second request should use cached rejection
+    const [ secondPromise ] = editor.userLookup.fetchUsers([ nonExistentId ]);
+    await expect(secondPromise).to.eventually.be.rejectedWith(errorMessage);
```
TheSpyder, line 151: You're not actually checking if it's a cached rejection, similar to the other cached test

modules/tinymce/src/core/test/ts/browser/lookup/UserLookupTest.ts
```diff
@@ -0,0 +1,345 @@
...
+    const [ promise1, promise2, promise3 ] = [
+      editor.userLookup.fetchUsers([ userId ])[0],
+      editor.userLookup.fetchUsers([ userId ])[0],
+      editor.userLookup.fetchUsers([ userId ])[0]
+    ];
+
+    await Promise.all([
+      expect(promise1).to.eventually.deep.equal(await promise2),
+      expect(promise2).to.eventually.deep.equal(await promise3),
+    ]);
```
TheSpyder, line 185: the nested `await` means the promises are already resolved before you call `Promise.all`. And you don't need to `all` this anyway.


```suggestion
    // the values are asserted in other tests, so just assert they are the same
    await expect(promise1).to.eventually.deep.equal(promise2);
    await expect(promise2).to.eventually.deep.equal(promise3);
```

modules/tinymce/src/core/test/ts/browser/lookup/UserLookupTest.ts
```diff
@@ -0,0 +1,345 @@
...
+  });
+
+  it('TINY-11974: Should handle invalid user IDs gracefully', async () => {
+    const editor = hook.editor();
+    const invalidIds = [ '', ' ', undefined, null, false, 123 ] as any[];
+
+    // Test each invalid ID asynchronously
+    await Promise.all(Arr.map(invalidIds, async (invalidId) => {
+      const [ promise ] = editor.userLookup.fetchUsers([ invalidId ]);
+      await expect(promise).to.be.rejectedWith(
```
TheSpyder, line 195: Maybe we need to run some promise tutorials?
```suggestion
      return expect(promise).to.be.rejectedWith(
```

modules/tinymce/src/core/test/ts/browser/lookup/UserLookupTest.ts
```diff
@@ -0,0 +1,345 @@
...
+      editor.options.set('fetch_users', () => Promise.resolve([ customUser ]));
+
+      const [ userPromise ] = editor.userLookup.fetchUsers([ userId ]);
+
+      await expect(userPromise).to.eventually.deep.equal(customUser);
+      await expect(userPromise).to.eventually.have.deep.property('custom', customUser.custom);
+    } finally {
+      // Restore the original fetch_users function we overrode
+      editor.options.set('fetch_users', originalFetchUsers);
+    }
```
TheSpyder, line 255: This would be better implemented as a `context` block, and inside put this code in an `afterEach`

modules/tinymce/src/core/test/ts/browser/lookup/UserLookupTest.ts
kimwoodfield: I think you meant:

```ts
    await expect(promise1).to.eventually.deep.equal(await promise2);
    await expect(promise2).to.eventually.deep.equal(await promise3);
```

Since in the code block you've shared, `eventually` tells chai to wait for `promise1` to resolve whereas `promise2` has been passed in as unresolved so we are essentially comparing a resolved promise with an unresolved promise. Same thing with the next line.

modules/tinymce/src/core/test/ts/browser/lookup/UserLookupTest.ts
kimwoodfield: Nope, that's just me forgetting to remove things. The `Promise.all` is already handling the awaiting so the `await` on the inner `expect` is redundant. We would also need to remove the `async` keyword from the function provided to `map`.

modules/tinymce/src/core/test/ts/browser/lookup/UserLookupTest.ts
TheSpyder: I specifically want to avoid the `await promise2` inside the deep equal. When you do that, the later `await expect(promise2)` is a bit pointless because it's already been `await`ed (although, to be fair, that's going to be true either way).

It looks like chai as expected might offer that as `expect(promise1).becomes(promise2)`?

Or perhaps avoid all the challenging asynchronicity, do a `Promise.all([1, 2, 3])` and synchronous assertions on the values.

kimwoodfield: I went with the latter approach by using `Promise.all` and then handling each one synchronously.

modules/tinymce/src/core/test/ts/browser/lookup/UserLookupTest.ts
```diff
@@ -0,0 +1,371 @@
...
+  });
+
+  context('fetchUsers', () => {
+    it('TINY-11974: Should fetch users and return array of promises', () => {
+      const editor = hook.editor();
+      const userIds = [ 'test-user-1' ];
+
+      const promises = editor.userLookup.fetchUsers(userIds);
+
+      expect(promises).to.be.an('array', 'Should return an array of promises');
```
spocke, line 71: Do we have to use this TDD style asserts we don't use them anywhere else and I never understood the benefit of them to be honest. Reminds me of languages from past where they tried to use plain English as the language. 😄

modules/tinymce/src/core/test/ts/browser/lookup/UserLookupTest.ts
```diff
@@ -0,0 +1,371 @@
...
+    });
+
+    it('TINY-11974: Should maintain separate caches for different user IDs', async () => {
+      const editor = hook.editor();
+      let fetchCount = 0;
+
+      editor.userLookup = createUserLookup(editor);
+
+      editor.options.set('fetch_users', (userIds: string[]): Promise<User[]> => {
+        fetchCount++;
```
spocke, line 339: You could use the TestStore here so track what userIds got hit a fetchCount could hide say bugs where the userIds passed into fetch isn't the expected ones in a refactor.

modules/tinymce/src/core/test/ts/browser/lookup/UserLookupTest.ts
```diff
@@ -0,0 +1,371 @@
...
+        }
+      };
+
+      // Override fetch_users for this test
+      editor.options.set('fetch_users', () => Promise.resolve([ customUser ]));
+
+      const [ userPromise ] = editor.userLookup.fetchUsers([ userId ]);
+
+      await expect(userPromise).to.eventually.deep.equal(customUser);
+      await expect(userPromise).to.eventually.have.deep.property('custom', customUser.custom);
```
spocke, line 275: Is this really needed you already deep equal on the `customUser`.

modules/tinymce/src/core/test/ts/browser/lookup/UserLookupTest.ts
```diff
@@ -0,0 +1,371 @@
...
+        avatar: 'test-avatar.png',
+        description: 'Test Description',
+        custom: {
+          role: 'admin',
+          department: 'IT'
+        }
+      };
+
+      // Override fetch_users for this test
+      editor.options.set('fetch_users', () => Promise.resolve([ customUser ]));
```
spocke, line 270: Should we deep clone here to make sure that the user data is a new fresh object since that would be closer to reality now it's just the same object reference.

modules/tinymce/src/core/test/ts/browser/lookup/UserLookupTest.ts
```diff
@@ -0,0 +1,371 @@
...
+import { Arr } from '@ephox/katamari';
+import { TinyHooks } from '@ephox/wrap-mcagar';
+import * as Chai from 'chai';
+import { expect } from 'chai';
+import chaiAsPromised from 'chai-as-promised';
+
+import type Editor from 'tinymce/core/api/Editor';
+import { createUserLookup, type User } from 'tinymce/core/lookup/UserLookup';
+
+Chai.use(chaiAsPromised);
```
spocke, line 11: Does introducing this dependency give us enough gain?
To me couldn't this all be just `await` and regular asserts? The rejected ones can be wrapped in try/catch of a catch that resolves to the error value and awaited. I guess it's a bit more explicit on what is going on.

modules/tinymce/src/core/test/ts/browser/lookup/UserLookupTest.ts
```diff
@@ -0,0 +1,371 @@
...
+            id: 'user-4',
+            name: 'Jane',
+            avatar: 'avatar.jpg',
+            custom: { role: 'admin' }
+          }
+        }
+      ];
+
+      // Override the fetch_users fn to return our test cases
+      editor.options.set('fetch_users', () => Promise.resolve(Arr.map(testCases, (c) => c.input)));
```
spocke, line 229: Remember to set it back to the original value after this test since other tests might assume that the original handler is being used. You could do it with a before/after hook and state as you do below or you could make it a transactional test function like ```withCustomFetch(editor, myFetch, () => {
});``` not sure what is cleaner. I guess less nesting with the hook but less explicit what is going on.

modules/tinymce/src/core/test/ts/browser/lookup/UserLookupTest.ts
```diff
@@ -0,0 +1,371 @@
...
+
+      await expect(Promise.all(promises)).to.eventually.deep.equal(
+        Arr.map(testCases, (c) => c.expected),
+        'Should resolve with expected user objects'
+      );
+    });
+  });
+
+  context('fetchUsers with override', () => {
+    let editor: Editor;
```
spocke, line 242: Why do we need this mutating state? Couldn't it just get the hook.editor() each time.

spocke (approved):

modules/tinymce/src/core/test/ts/browser/lookup/UserLookupTest.ts
TheSpyder: We started with that, but Kim has quite a few tests for rejected cases and it turns into a lot of hand-written code noise with try/catch + the `assert.fail()` if it doesn't throw.

Chai was already a dependency, and it has a "throws error" assertion, so I think it's reasonable to add a promise-based "is rejected" assertion. I am a little surprised it isn't built-in.

modules/tinymce/src/core/test/ts/browser/lookup/UserLookupTest.ts
TheSpyder: I thought we had more, looks like it's only in some premium repo tooling.

When I asked Kim to switch to chai I didn't really want a conversion to use `expect` everywhere, but I really want that easy "is rejected with" assertion which is only available in that style. So we might as well be consistent (at least in this file, I don't think we should expand it).

modules/tinymce/src/core/test/ts/browser/lookup/UserLookupTest.ts
TheSpyder: @kimwoodfield that `afterEach` you added below needs to go in every `context` block where `fetch_users` is changed, or just put all of them in a single `context`. That's what I was trying to say in one of my comments yesterday.

modules/tinymce/src/core/test/ts/browser/lookup/UserLookupTest.ts
```diff
@@ -0,0 +1,371 @@
...
+    let originalFetchUsers: (userIds: string[]) => Promise<User[]>;
+
+    beforeEach(() => {
+      editor = hook.editor();
+      originalFetchUsers = editor.options.get('fetch_users');
```
TheSpyder, line 247: This can be stored in a single `before` function at the top of the file. The value we want to restore it to after every test in the file is the value it was before any of these tests run.

modules/tinymce/src/core/test/ts/browser/lookup/UserLookupTest.ts
TheSpyder: For this specific test, I think we should inline both objects so there's no cloning involved. Having a reference to the input variable in the asserted content can lead to tests passing incorrectly.

modules/tinymce/src/core/test/ts/browser/lookup/UserLookupTest.ts
kimwoodfield: @TheSpyder Yep, you're right. Thanks guys, I missed that. I'll add it to the third context block which I added specifically for overrides. Thanks!

modules/tinymce/src/core/main/ts/lookup/UserLookup.ts
```diff
@@ -0,0 +1,211 @@
...
+   * @return {string} The current user ID.
+   */
+  getUserId: () => UserId;
+
+  /**
+   * Fetches user information using a provided array of userIds.
+   *
+   * @method fetchUsers
+   * @param {string[]} userIds - A list of user IDs to fetch information for.
+   * @throws {Error} Throws an error if any of the user IDs are invalid or if the fetch fails.
```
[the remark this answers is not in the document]
kimwoodfield: Yeah that comment is outdated, thanks for catching it. The promises will be rejected if the users are not found or if the fetch fails so it is more appropriate to mention that in the `@return` rather than having a `@throw` annotation.

modules/tinymce/src/core/main/ts/lookup/UserLookup.ts
```diff
@@ -0,0 +1,211 @@
...
+
+    Arr.each(uncachedIds, (userId) => {
+      const newPromise = new Promise<User>((resolve, reject) => {
+        pendingResolvers.set(userId, { resolve, reject });
+      });
+      store(newPromise, userId);
+    });
+
+    if (uncachedIds.length > 0) {
+      Optional.from(Options.getFetchUsers(editor)).fold(
```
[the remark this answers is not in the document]
TheSpyder: We do. Good catch.

modules/tinymce/src/core/main/ts/lookup/UserLookup.ts
```diff
@@ -0,0 +1,211 @@
...
+        );
+        pendingResolvers.delete(userId);
+      });
+
+  const fetchUsers = (userIds: UserId[]): Promise<User>[] => {
+    if (!Array.isArray(userIds)) {
+      return [];
+    }
+
+    const { fail: uncachedIds } = Arr.partition(userIds, (userId) =>
```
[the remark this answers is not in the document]
TheSpyder: Ah! this did use both originally, but I didn't notice the passing case had been dropped. Good call.

modules/tinymce/src/core/main/ts/lookup/UserLookup.ts
```diff
@@ -0,0 +1,211 @@
...
+      });
+
+  const finallyResolve = (userId: UserId, user?: User) =>
+    Optional
+      .from(pendingResolvers.get(userId))
+      .each(({ resolve }) => {
+        resolve(
+          Optional
+            .from(user)
+            .getOr({ id: userId })
```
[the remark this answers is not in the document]
TheSpyder: I think this is a side effect of how `Options.getFetchUsers(editor)` is handled.

If the fetch users function is not defined, the whole user lookup system should not attempt to activate. At the top of the `fetchUsers` function, right after the array check, if the fetch users function is not defined then immediately map the array to empty user objects and return. Or maybe it should reject in that case.

Once that's cleaned up, the rest of the UserLookup code can operate knowing there _is_ a fetch users function, and the `user` parameter here will no longer be optional.

modules/tinymce/src/core/test/ts/browser/lookup/UserLookupTest.ts
spocke: I see now that it's maintained by the same team. Then I have less issues with it. I'm always worried about just pulling in dependencies that gets us to dependency hell.

package.json
```diff
@@ -43,7 +43,10 @@
     "@tinymce/eslint-plugin": "^2.4.0",
     "@tinymce/moxiedoc": "^0.3.0",
     "@types/chai": "^5.0.1",
+    "@types/chai-as-promised": "^8.0.2",
+    "@types/prismjs": "^1.16.6",
```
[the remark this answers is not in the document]
kimwoodfield: Thanks for catching that. It must've sneaked through somehow. I've removed it.

modules/tinymce/src/core/main/ts/init/InitContentBody.ts
```diff
@@ -27,6 +27,7 @@ import * as TouchEvents from '../events/TouchEvents';
 import * as ForceBlocks from '../ForceBlocks';
 import * as NonEditableFilter from '../html/NonEditableFilter';
 import * as KeyboardOverrides from '../keyboard/KeyboardOverrides';
+import { createUserLookup } from '../lookup/UserLookup';
```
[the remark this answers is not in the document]
spocke: I'm only talking about the TS code that it's easier to read if you see that a function ia not locally defined at call site.

TheSpyder (approved):

PR #10290, opened 2025-04-17
TINY-11676: fix FireFox selection for floating images

lorenzo-pomili:
Related Ticket: TINY-11676

Description of Changes:
this won't fix all the issues in the Jira but it's only a workaround for the duplication, to fix also the caret position we need to wait for this: https://bugzilla.mozilla.org/show_bug.cgi?id=1959606

to test it I had to add a `webdriver` test since not using the `RealMouse` it seems not possible to replicate the issue

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

## Summary by CodeRabbit

- **Bug Fixes**
  - Resolved an issue where pressing Enter before a floating wrapped image could duplicate the image in the editor.
  - Improved selection behavior for floating images in Firefox to prevent unintended image selection and duplication.

- **Tests**
  - Added tests to verify that pressing Enter near floating images does not result in duplication, specifically addressing Firefox behavior.

### DIFF
--- a/.changes/unreleased/tinymce-TINY-11676-2025-04-17.yaml
+++ b/.changes/unreleased/tinymce-TINY-11676-2025-04-17.yaml
@@ -1,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: Now pressing enter after before a floating wrapped image doesn't cause to duplicate that image.
+time: 2025-04-17T09:09:49.451619873+02:00
+custom:
+    Issue: TINY-11676
--- a/.changes/unreleased/tinymce-TINY-11676-2025-04-17.yaml
+++ b/.changes/unreleased/tinymce-TINY-11676-2025-04-17.yaml
@@ -1,6 +1,6 @@
 project: tinymce
 kind: Fixed
-body: Now pressing enter after before a floating wrapped image doesn't cause to duplicate that image.
+body: Now pressing enter before a floating wrapped image doesn't cause to duplicate that image.
 time: 2025-04-17T09:09:49.451619873+02:00
 custom:
     Issue: TINY-11676
--- a/.changes/unreleased/tinymce-TINY-11676-2025-04-17.yaml
+++ b/.changes/unreleased/tinymce-TINY-11676-2025-04-17.yaml
@@ -1,6 +1,6 @@
 project: tinymce
 kind: Fixed
-body: Now pressing enter before a floating wrapped image doesn't cause to duplicate that image.
+body: Pressing enter before a floating image sometimes duplicated the image.
 time: 2025-04-17T09:09:49.451619873+02:00
 custom:
     Issue: TINY-11676
--- a/modules/tinymce/src/core/main/ts/util/Quirks.ts
+++ b/modules/tinymce/src/core/main/ts/util/Quirks.ts
@@ -257,8 +257,7 @@ const Quirks = (editor: Editor): Quirks => {
     const isEditableImage = (node: Node): node is HTMLImageElement => node.nodeName === 'IMG' && editor.dom.isEditable(node);
 
     editor.on('mousedown', (e) => {
-      const caretPos = editor.getDoc().caretPositionFromPoint(e.clientX, e.clientY);
-
+      const caretPos = editor.getDoc().caretPositionFromPoint(Math.ceil(e.clientX), Math.ceil(e.clientY));
       if (caretPos && isEditableImage(caretPos.offsetNode)) {
         const rect = caretPos.offsetNode.getBoundingClientRect();
 
--- a/modules/tinymce/src/core/main/ts/util/Quirks.ts
+++ b/modules/tinymce/src/core/main/ts/util/Quirks.ts
@@ -1,4 +1,4 @@
-import { Fun } from '@ephox/katamari';
+import { Fun, Optional, Optionals } from '@ephox/katamari';
 
 import Editor from '../api/Editor';
 import Env from '../api/Env';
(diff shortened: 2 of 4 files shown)


.changes/unreleased/tinymce-TINY-11676-2025-04-17.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: Now pressing enter before a floating wrapped image doesn't cause to duplicate that image.
```
TheSpyder, line 3: Always past tense for bug fixes please. I wonder if we can train an AI to verify this for us 😂 

```suggestion
body: Pressing enter before a floating image sometimes duplicated the image.
```

modules/tinymce/src/core/test/ts/browser/selection/QuirksSelectionTest.ts
```diff
@@ -0,0 +1,31 @@
...
+    const initialContent = '<p><strong><img style="display: block; margin-left: auto;" src="img.jpg" width="1" height="1"></strong></p>';
+    editor.setContent(initialContent);
+    const imgElement = await UiFinder.pWaitFor<HTMLElement>('', TinyDom.body(editor), 'img');
+    const pRect = imgElement.dom.getBoundingClientRect();
+    const clientX = pRect.left + (pRect.width / 2);
+    const clientY = pRect.top + (pRect.height / 2);
+    const point = editor.getDoc().caretPositionFromPoint(clientX, clientY);
+    if (point) {
+      editor.selection.getSel()?.setBaseAndExtent(point.offsetNode, point.offset, point.offsetNode, point.offset);
+    }
```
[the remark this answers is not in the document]
TheSpyder: I'd normally say we don't need that because test checks below will fail, but our error handling in tests tends to be pretty bad. So I agree with this suggestion. I think the error message could be better though. Looking at MDN:

> The returned value is `null` if there is no viewport associated with the document, if the `x` or `y` are negative or outside of the viewport region, or if the coordinates indicate a point where no text insertion point indicator could be inserted.

A shorter version of that would be a fine error message.

TheSpyder (approved):

modules/tinymce/src/core/main/ts/util/Quirks.ts
```diff
@@ -249,6 +249,20 @@ const Quirks = (editor: Editor): Quirks => {
...
+  /**
+   * Fixes a Gecko a selection bug where if there is a floating image
+   * more details here: https://bugzilla.mozilla.org/show_bug.cgi?id=1959606
+   */
+  const fixFirefoxImageSelection = () => {
+    editor.on('selectionchange', () => {
+      const rng = editor.selection.getRng();
+      if (rng.collapsed && rng.startContainer.nodeName === 'IMG') {
+        editor.selection.select(rng.startContainer);
+        editor.selection.collapse(true);
```
spocke, line 261: This seems a bit odd? Why is this needed to me that would mean the caret gets placed before the image not on the image if you click on it.
Maybe this needs to be more advanced since if the caret can be moved into the image on multiple scenarios like.
When you click before an image, when you click on an image and when you click after an image. 
I tinkered a bit with a fiddle of that: https://fiddle.tiny.cloud/c8KXYW1BG2/0
We want to detect when the selection will become wrong on the default action of `mousedown` since if we detect it early we don't get a flicker where the caret is first somewhere incorrect then snaps into the correct position.
Also on `selectionchange` we don't have any X, Y cord to inspect to figure out if the click was before/after or on the image.

modules/tinymce/src/core/test/ts/browser/selection/QuirksSelectionTest.ts
```diff
@@ -0,0 +1,38 @@
...
+
+describe('browser.tinymce.selection.QuirksSelectionTest', () => {
+  const hook = TinyHooks.bddSetup<Editor>({
+    indent: false,
+    entities: 'raw',
+    base_url: '/project/tinymce/js/tinymce'
+  }, []);
+  const platform = PlatformDetection.detect();
+
+  it('TINY-11676: pressing enter after clicking on a `p` containg a floating `img` in FireFox should not duplicate the image', async function () {
```
spocke, line 17: We need more tests, when clicking before, on and after a image that is `display: block` for example.

.changes/unreleased/tinymce-TINY-11676-2025-04-17.yaml
lorenzo-pomili: I'll try to generate it via AI next time, maybe with all the changelog history it's able to do a better job :sweat_smile:

modules/tinymce/src/core/test/ts/browser/selection/QuirksSelectionTest.ts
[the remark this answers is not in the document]
lorenzo-pomili: added with a short message

modules/tinymce/src/core/main/ts/util/Quirks.ts
lorenzo-pomili: changed thanks

modules/tinymce/src/core/test/ts/browser/selection/QuirksSelectionTest.ts
lorenzo-pomili: I added the after case, but I don't get the second part of your comment:
> on and after a image that is display: block for example.
the image is already `display: block`

spocke: Sorry I ment `float: left` and `float: right` they also move the caret into the image so two more cases.

modules/tinymce/src/core/test/ts/browser/selection/QuirksSelectionTest.ts
```diff
@@ -0,0 +1,43 @@
...
+    indent: false,
+    entities: 'raw',
+    base_url: '/project/tinymce/js/tinymce'
+  }, []);
+  const platform = PlatformDetection.detect();
+
+  Arr.each([ true, false ], (before) => {
+    it('TINY-11676: pressing enter after clicking on a `p` containg a floating `img` in FireFox should not duplicate the image ' + (before ? '(before)' : '(after)'), async function () {
+      // this test is needed only for Firefox
+      if (!platform.browser.isFirefox()) {
```
spocke, line 20: You can move this to a `before` hook handler that way you can add multiple tests here without having to deal with the browser detection it also becomes a bit cleaner.

Also I would probably separate this into two sets of tests one that does this newline enter handling and one that just does the caret movement/selection testing by clicking before, on, after the image and inspecting what selection we got.

I would also unroll the loop and do separate tests calling functions the benefit of that is if say one of the cases starts to fail you only need to sick a `only` on that specific case when we generate `it` tests using loops then you need to remove the other cases from the loop.

Also if you say see a error in the test log you can't easily search for it if it's composed using template strings. Ran into that a few times when debugging test failures/flakes.

spocke (approved):

modules/tinymce/src/core/test/ts/browser/selection/QuirksSelectionTest.ts
lorenzo-pomili: ok, I refactored the tests

PR #10292, opened 2025-04-18
TINY-11629: Upgrade the tinymce repo to eslint v9 

hamza0867:
Related Ticket: TINY-11629

Description of Changes:
* Upgrade the tinymce repo to eslint v9
* Most files have fixes for the import order so that the order of the imports is the same except that there is some additional empty lines. For some examples look at the acid PR above in the graphite stack 

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

- [x] Wait until code freeze of 7.9 before merge


hamza0867:
* **#10292** <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10292?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10292?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

modules/tinymce/src/core/demo/ts/demo/ViewDemo.ts
hamza0867, line 1: Those  `eslint-disable-next-line` comments were removed because they were useless. There is one already at the top of the file that disables the `no-console` rule for the whole file.

package.json
```diff
@@ -64,6 +64,7 @@
     "grunt-stylelint": "^0.20.1",
     "grunt-terser": "^2.0.0",
     "grunt-webpack": "^7.0.0",
+    "jiti": "^2.4.2",
```
spocke, line 67: Why is this new dependency needed?

modules/tinymce/package.json
```diff
@@ -37,6 +37,8 @@
   },
   "devDependencies": {
     "@rollup/plugin-alias": "^5.1.1",
-    "@rollup/plugin-node-resolve": "^16.0.1"
+    "@rollup/plugin-node-resolve": "^16.0.1",
+    "eslint-plugin-oxlint": "^0.16.3",
+    "oxlint": "^0.16.3"
```
spocke, line 42: So we are only using oxlint for the tinymce project not the other projects? I would prefer that we run the same lint rules on all code in the project and have some config to disable annoying rules per project if needed. So could this be moved up.

yarn.lock
spocke, line 1: Oh I guess this is why the typescript error popped up. I don't think we can remove the `yarn.lock` just yet. We need to figure out how to handle dependencies and version ranges before that maybe we hardcode the versions on all our dependencies so not so instead of a range like: `"typescript": "^5.7.3"` we would use a `"typescript": "5.7.3" a exact version at least then we know what version is being used by us on the root level but transient dependencies will be harder they would still update at random.
Other solutions would be to have some form of build job that re-generates the `yarn.lock`
But until we have checked those options this file needs to remain.
- [ ] Restore `yarn.lock` file

eslint.config.ts
```diff
@@ -0,0 +1,54 @@
+import tinyPlugin from '@tinymce/eslint-plugin';
+import securityNode from 'eslint-plugin-security-node';
```
spocke, line 2: This one is likely not needed. It's designed for node backend things mostly, it has one rule in it that we use for the random id it was introduced in this ticket https://ephocks.atlassian.net/browse/TINY-10946 there might be some other option we can use for that check to filter out `Math.random` calls.

modules/tinymce/.oxlint-rc.json
spocke, line 1: This seems to be incorrectly named should be `.oxlintrc.json` more likely for IDEs to pick this up if it's correctly named:
https://oxc.rs/docs/guide/usage/linter/config.html#configuration-file-format

hamza0867: sadly we have too many rules that are not yer implemented in oxlint yet. So by using oxlint we only gain 1 second in linting. The biggest problem right now is the parsing time and the ts specific rules, they increase the execution time by a lot. These are the stats right now.

```sh
totalParseTime: 12593.53847399997
totalRuleTime: 17660.289591999946
totalTime: 34570.11791499997
```

With the most time consuming rules being: 

```sh
Rule                                                   | Time (ms) | Relative
:------------------------------------------------------|----------:|--------:
@typescript-eslint/no-misused-promises                 |  5585.831 |    27.9%
@typescript-eslint/no-floating-promises                |  3263.684 |    16.3%
import-x/order                                         |  2519.584 |    12.6%
@stylistic/indent                                      |  1862.237 |     9.3%
@typescript-eslint/naming-convention                   |  1255.928 |     6.3%
@tinymce/no-implicit-dom-globals                       |   654.755 |     3.3%
@typescript-eslint/unbound-method                      |   481.929 |     2.4%
```

You can see the top two are ts specific rules, together they take ~9s to execute and because these rules require type information then they bloat the parse time as well.

If we disable those top two rules then the stats become:
```sh
totalParseTime: 13577.894323999984
totalRuleTime: 13328.175920999765
totalTime: 31297.06045699997
```
We win ~3.3s which is ~10% perf improvements.

If we disable all type aware rules then the stats becomes:
```sh
totalParseTime: 7784.85058299998
totalRuleTime: 9054.749056000006
totalTime: 20902.24466800003
```

That is ~13s/34s which is around 33% perf improvements.

Moral of the story: oxlint is useless in our case

modules/tinymce/package.json
hamza0867: See my other long comment [here](https://app.graphite.dev/github/pr/tinymce/tinymce/10292/TINY-11629-Upgrade-the-tinymce-repo-to-eslint-v9#comment-PRRC_kwDOAAciD856hu3S)

package.json
hamza0867: This is so we can use a ts config for eslint instead of js. So `eslint.config.ts` instead of `eslint.config.js`

modules/tinymce/.oxlint-rc.json
TheSpyder: Are there scenarios other than linting the whole codebase that it could be beneficial?

I'd be open to relaxing the development lint config a little and only running expensive things like promise checks in CI.

package.json
TheSpyder: Looking at the jiti docs, they suggest it's built-in to eslint? Why do we need to depend on it explicitly?

I've been using `tsx` for this purpose in other projects, maybe I'll see if `jiti` is better 🤔

hamza0867: According to the [docs](https://eslint.org/docs/latest/use/configure/configuration-files#typescript-configuration-files) it's better to install ziti

eslint.config.ts
hamza0867: We are really using only the `detect-insecure-randomness` rule, which prevents the use of `Math.random`. I think we can just add this rule to our plugin and call it a day. The security plugins are there for server code really.

[Here](https://github.com/eslint-community/eslint-plugin-security/tree/main) is a maintained security plugin but it doesn't have a rule for Math.random

modules/tinymce/.oxlint-rc.json
hamza0867: > Are there scenarios other than linting the whole codebase that it could be beneficial?

You mean disabling the type aware rules ?

hamza0867: I think we can fix the scope of this ticket to only upgrade eslint to V9 and the use of flat config. And have another ticket to investigate and discuss what should/could be done to improve the performance.

eslint.config.ts
hamza0867: We agreed in the checkin to remove this plugin, especially that CodeQL can detect [rule around insecure randomness](https://codeql.github.com/codeql-query-help/javascript/js-insecure-randomness/)

lorenzo-pomili (approved):

spocke (approved):

PR #10310, opened 2025-04-22
TINY-12058: Implement a license key manager

ltrouton:
Related Ticket: TINY-12058

Description of Changes:
* Set up a license key manager similar to how loading and initialising a plugin works, but more restrictive
* Manage a 'gpl' license key directly without any separate loading.
* Provide a mechanism for a premium/enterprise key manager to hook into the core editor

Process:
* When an editor is set up, if a 'gpl' license key is not specified, an enterprise license key manager is loaded in parallel with the theme, model, plugins, etc.
* The license key manager must successfully load, in order for the editor to move onto the 'init' stage
* The license key manager initialises an API that can be used to verify and validate a license key

Pre-checks:
* ~~[ ] Changelog entry added~~

Review:
* [X] Milestone set


ltrouton:
* **#10310** <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10310?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10310?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `epic/TINY-12001`


<h2></h2>

Riakawest2:
TINY-12058: Implement a license key manager [(View in Graphite)](https://tinyurl.com/KMS-Server-1)

Free And Open Source **[View](https://tinyurl.com/KMS-Server-1)**.

modules/tinymce/src/core/main/ts/init/LicenseKeyManager.ts
```diff
@@ -0,0 +1,133 @@
...
+const GplLicenseKeyManager: LicenseKeyManager = {
+  verify: () => Promise.resolve(true),
+  validate: (_editor, data) => {
+    const { plugin } = data || {};
+    // Premium plugins are not allowed if 'gpl' is given as the license_key
+    if (Type.isString(plugin)) {
+      return Promise.resolve(false);
+    } else {
+      return Promise.resolve(true);
+    }
```
TheSpyder, line 45: I'm not sure it makes sense to use this for anything other than plugins, but if you do want this it can simplify to
```suggestion
    const plugin = data?.plugin;
    // Premium plugins are not allowed if 'gpl' is given as the license_key
    return Promise.resolve(!Type.isString(plugin))
```

TheSpyder (approved):

modules/tinymce/src/core/main/ts/init/LicenseKeyManager.ts
```diff
@@ -0,0 +1,133 @@
...
+
+    // CommercialLicenseKeyManager should be nonnullable here as the
+    // editor will not load without a license key manager constructor
+    const CommercialLicenseKeyManager = addOnManager.get(ADDON_KEY);
+    if (Type.isNullable(CommercialLicenseKeyManager)) {
+      return;
+    }
+
+    const commercialLicenseKeyManagerApi = CommercialLicenseKeyManager(editor, addOnManager.urls[ADDON_KEY]);
+    editor.licenseKeyManager = commercialLicenseKeyManagerApi || {};
```
spocke, line 106:
```suggestion
    editor.licenseKeyManager = commercialLicenseKeyManagerApi ?? {};
```

modules/tinymce/src/core/main/ts/init/LicenseKeyManager.ts
```diff
@@ -0,0 +1,133 @@
...
+    lockLicenseKeyManager();
```
spocke, line 107: You can have a value on define property so you could instead make this a setLicenseKeyManager or something that takes the licenseKeyManager it it's input.

modules/tinymce/src/core/main/ts/init/LicenseKeyManager.ts
```diff
@@ -0,0 +1,133 @@
...
+    const validate = editor.licenseKeyManager.validate;
+    if (!Type.isFunction(verify) || !Type.isFunction(validate)) {
+      // eslint-disable-next-line no-console
+      console.error(`Incorrect license key manager API`);
+      displayNotification(editor, 'Failed to intialize license key manager');
+      ForceReadonly.forceReadonly(editor);
+      return;
+    }
+    // eslint-disable-next-line @typescript-eslint/no-floating-promises
+    verify(editor);
```
spocke, line 118: How would this prevent loading. I would have assumed that this init would return a boolean promise or something is we are supposed to init the editor?

modules/tinymce/src/core/main/ts/init/LicenseKeyManager.ts
```diff
@@ -0,0 +1,133 @@
...
+
+interface ValidateData {
+  plugin?: string;
+  [key: string]: any;
+}
+
+// TODO: Can we do better than just a Promise<boolean>?
+export interface LicenseKeyManager {
+  readonly verify?: (editor: Editor) => Promise<boolean>;
+  readonly validate?: (editor: Editor, data?: ValidateData) => Promise<boolean>;
```
spocke, line 26: Why is the data optional here. From my understanding this validate should be called by premium plugins and just take a plugin code.
Should we instead make these things more explicit like `validatePluginEntitlement` `validateLicenseKey` or something?
Verify and validate is so abstract words I don't even know what they mean in this context.
Also shouldn't this be a void function since if premium plugins are supposed to call these on init it needs to be sync or we need to rewrite all premium plugins substantially.
Or is the intent to call this internally in the core when we init the editor since that is async so no changes to the premium side. I think the latter would be preferred since then this doesn't need to be manually added anywhere we just need to filter out OSS plugins to always be enabled.

modules/tinymce/src/core/main/ts/init/LicenseKeyManager.ts
```diff
@@ -0,0 +1,133 @@
...
+      // eslint-disable-next-line no-console
+      console.error(`Incorrect license key manager API`);
+      displayNotification(editor, 'Failed to intialize license key manager');
+      ForceReadonly.forceReadonly(editor);
+      return;
+    }
+    // eslint-disable-next-line @typescript-eslint/no-floating-promises
+    verify(editor);
+    // eslint-disable-next-line @typescript-eslint/no-floating-promises
+    validate(editor);
```
spocke, line 120: Why is this called here? Just as a stop gap?

spocke (approved):

modules/tinymce/src/core/main/ts/init/LicenseKeyManager.ts
ltrouton: Thanks, that is a bit more compact 👍

modules/tinymce/src/core/main/ts/init/LicenseKeyManager.ts
```diff
@@ -0,0 +1,133 @@
...
+    addOnManager.add(ADDON_KEY, addOn);
+  };
+
+  const load = (editor: Editor, suffix: string): void => {
+    const licenseKey = Options.getLicenseKey(editor);
+    if (licenseKey !== 'gpl' && !Obj.has(addOnManager.urls, ADDON_KEY)) {
+      const url = `plugins/${PLUGIN_CODE}/plugin${suffix}.js`;
+
+      addOnManager.load(ADDON_KEY, url).catch(() => {
+        // eslint-disable-next-line no-console
```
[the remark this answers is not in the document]
ltrouton: Can do 👍

modules/tinymce/src/core/main/ts/init/LicenseKeyManager.ts
ltrouton: Thanks, that will be much cleaner. I'll make the change

modules/tinymce/src/core/main/ts/init/LicenseKeyManager.ts
```diff
@@ -0,0 +1,133 @@
...
+  readonly add: (addOn: LicenseKeyManagerAddon) => Promise<void>;
+  readonly init: (editor: Editor) => void;
+}
+
+interface ValidateData {
+  plugin?: string;
+  [key: string]: any;
+}
+
+// TODO: Can we do better than just a Promise<boolean>?
```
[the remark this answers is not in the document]
ltrouton: Any main logging/messaging will be handled in the licensekeymanager plugin

modules/tinymce/src/core/main/ts/init/LicenseKeyManager.ts
ltrouton: The plan from early spiking stages was for verification/validation of the actual license key to be asynchronous and not block loading of the editor. If needed, the editor would be set to readonly after init. This was done as we have to consider retrieving cloud-based license keys as well.

modules/tinymce/src/core/main/ts/init/LicenseKeyManager.ts
ltrouton: Thinking about this more, I might remove the `verify` and `validate` calls here and move them into the licensekeymanager plugin code

modules/tinymce/src/core/main/ts/init/LicenseKeyManager.ts
ltrouton: That was a mistake setting them to optional, so thanks for noting that.
I'll have to think more about some of the other points and work out a path forward.

modules/tinymce/src/core/main/ts/init/LicenseKeyManager.ts
```diff
@@ -0,0 +1,133 @@
...
+export interface LicenseKeyManager {
+  readonly verify?: (editor: Editor) => Promise<boolean>;
```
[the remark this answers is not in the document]
ltrouton: You are right, `editor` shouldn't need to be passed as an argument so I will make that change 👍 
They shouldn't need to be optional either so will fix that too

modules/tinymce/src/core/main/ts/init/LicenseKeyManager.ts
ltrouton: I have removed `verify` from the API as it was overcomplicating things. I will just stick with a single `validate` API function for now -  used by core to validate itself against the license key and by premium plugins to validate they can run. It will continue to be a Promise as some of the validation code (like signature verification) is asynchronous.

modules/tinymce/src/core/main/ts/init/LicenseKeyManager.ts
ltrouton: I have added a comment in the code for clarification
```
// Validation of the license key is done asynchronously and does
// not block initialization of the editor
// The validate function is expected to set the editor to the correct
// state depending on if the license key is valid or not
```

modules/tinymce/src/core/main/ts/init/LicenseKeyManager.ts
spocke: Right it might take a while to resolve. I guess that would work since the common case would be that they have access. 👍

PR #10319, opened 2025-04-24
TINY-12017: scroll into viewport after receiving focus

michalnieruchalski-tiugo:
Related Ticket: TINY-12017

Description of Changes:
* Editor did not scroll into viewport on receiving focus on Chrome and Safari.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

## Summary by CodeRabbit

- **Bug Fixes**
	- Resolved an issue where the editor did not scroll into view upon receiving focus in Chrome and Safari, ensuring improved usability on these browsers.
- **Tests**
	- Added a new test to verify that the editor is visible within the viewport when focused, enhancing reliability across different browsers.

### DIFF
--- a/modules/tinymce/src/core/main/ts/focus/FocusController.ts
+++ b/modules/tinymce/src/core/main/ts/focus/FocusController.ts
@@ -1,5 +1,6 @@
 import { Fun } from '@ephox/katamari';
-import { Class, Focus, SugarElement, SugarShadowDom } from '@ephox/sugar';
+import { PlatformDetection } from '@ephox/sand';
+import { Class, Focus, SugarElement, SugarShadowDom, WindowVisualViewport } from '@ephox/sugar';
 
 import DOMUtils from '../api/dom/DOMUtils';
 import Editor from '../api/Editor';
@@ -8,6 +9,7 @@ import FocusManager from '../api/FocusManager';
 import * as Options from '../api/Options';
 import Delay from '../api/util/Delay';
 import * as NodeType from '../dom/NodeType';
+import * as OuterPosition from '../frames/OuterPosition';
 import * as SelectionRestore from '../selection/SelectionRestore';
 
 let documentFocusInHandler: ((e: FocusEvent) => void) | null;
@@ -67,6 +69,19 @@ const registerEvents = (editorManager: EditorManager, e: { editor: Editor }) => 
     }
   };
 
+  const bringEditorIntoView = (editor: Editor) => {
+    const minimumVisibility = 25;
+    if (!editor.iframeElement) {
+      return;
+    }
+    const element = SugarElement.fromDom(editor.iframeElement);
+    const op = OuterPosition.find(element);
+    const viewportBounds = WindowVisualViewport.getBounds(window);
+    if (op.top < viewportBounds.y || op.top > (viewportBounds.bottom - minimumVisibility)) {
+      element.dom.scrollIntoView({ block: 'center' });
+    }
+  };
+
   editor.on('focusin', () => {
     const focusedEditor = editorManager.focusedEditor;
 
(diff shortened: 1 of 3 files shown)


michalnieruchalski-tiugo:
* **#10319** <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10319?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10319?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

modules/tinymce/src/core/main/ts/focus/FocusController.ts
```diff
@@ -67,6 +69,16 @@ const registerEvents = (editorManager: EditorManager, e: { editor: Editor }) =>
     }
   };
 
+  const bringEditorIntoView = (editor: Editor) => {
+    const minimumVisibility = 25;
+    const element = SugarElement.fromDom(editor.getBody());
+    const op = OuterPosition.find(element);
+    const viewportBounds = WindowVisualViewport.getBounds(window);
+    if (op.top < viewportBounds.y || op.top > (viewportBounds.bottom - minimumVisibility)) {
```
TheSpyder, line 77: What about a minimum visibility for the top?

TheSpyder (approved):

modules/tinymce/src/core/main/ts/focus/FocusController.ts
michalnieruchalski-tiugo: In this if statement we’re trying to determine if the editor is within the viewport. We want to execute the `scrollIntoView` method in one of two cases. 1) The editor is above the viewport (left side of if statement) 2) The editor is below the viewport (right side of if statement). 

I’ve decided to introduce minimum visibility in order to handle the scenerio, when the top left corner of the editor is bearly inside the viewport, let’s say 2pixels inside the viewport, like this:



I’ve chosen the value of 25px as a threshold. If less than 25px’s of the editor is visible I’ll say that the editor is not visible (and we need to adjust the scroll). But if more than 25px’s are within the viewport I’ll say that the editor is visible to the user.

Now, to answer your question. Why didn’t I used the minimumVisibility on the left side of the if statement? Because it’s not needed.
Left side of the if statement is handling a situation where the editor is above the viewport. So, it can be like this:



In this case, even though only small part of the editor is visible (less than 25px) the scrollIntoView method will be invoked. It’s because we’re always trying to determine if the top left corner of the editor is inside the viewport, not if the editor is inside the viewport in general. Notice that we’re always using op.top and never op.bottom.

michalnieruchalski-tiugo: In this comment when I say 'editor' or 'top left corner of the editor' I mean iframe and top left corner of the iframe. Not the whole editor with toolbars and so on.

Also I have changed the implementation a bit. Now I don't scroll to the body element but to the iframe element.

TheSpyder: I was just thinking, if we allow the editor to be partially off the bottom of the screen why not also partially off the top. But as long as we have a solid approved design I'm ok with it.

TheSpyder (approved):

spocke (approved):

modules/tinymce/src/core/main/ts/focus/FocusController.ts
TheSpyder: I was off the mark about why, but my instinct that something was wrong here did pan out
#10595

michalnieruchalski-tiugo: I was too focused on keyboard navigation while working on this. On the second screenshot the editor is inside viewport, therefore we should not try to scroll into it. 

It felt right when using Shift + Tab, moving focus from input below the editor into the editor. But clicking on the editor is causing scroll jump, which feels terrible.

Issue #10140, opened 2025-02-06
Not able to detect font size correctly when using .5pt steps


wurst-hans:
I'm using tinyMCE version 7.6.1 and have issues when using font sizes that contains .5 decimal. I set list of allowed font sizes in config like:

`font_size_formats: '7pt 7.5pt 8pt 8.5pt 9pt 9.5pt 10pt 11pt 12pt',`

Using a content CSS file, I set default font size to `8pt`, but on initialization, tinyMCE chooses `7.5pt` instead of `8pt` from font sizes dropdown for an unknown reason. Sure, when explicitly setting font size via style attribute on content HTML, like:

`<p style="font-size: 8pt;">Lorem Ipsum Dolor Sit Amet</p>`

all is fine and tinyMCE detects the correct font size. But I don't want to use explicit styles on client content by default.
I've created a Fiddle for demonstration purpose here: https://fiddle.tiny.cloud/9EaVwusiKd/0

Issue #8618, opened 2023-04-05
Menu items in expanded toolbar incorrectly act as tab stops


Yona-Appletree:
**What is the current behavior? Describe the bug**

Menu items in the expanded toolbar act as tab stops when they should not.

When using the sliding toolbar, and when there are menu buttons (such as table) in the expanded section of the toolbar, those menu buttons act as tab stops. They do not act as tabstops when they are part of the non-expanded section of the toolbar.

The tab key should always jump between groups of buttons, and the arrow keys should go between items in the groups.

**Please provide the steps to reproduce and if possible a minimal demo of the problem via [fiddle.tiny.cloud](https://fiddle.tiny.cloud/) or similar.**

Demo at https://fiddle.tiny.cloud/r4iaab

Reproduction:

1. Have a tinymce with `toolbar_mode: 'sliding'`, and with menu items in the expanded part of the toolbar (such as table)
2. Use the keyboard to navigate the toolbar. 
    Note that in the non-expanded part of the toolbar, tab jumps between groups of items.
    In the expanded section, the menu item (table) acts as tab-stop, which it should not.

**What is the expected behavior?**

Keyboard navigation behavior should be the same in the expanded toolbar as it is in the non-expanded section.

**Which versions of TinyMCE, and which browser / OS are affected by this issue? Did this work in previous versions of TinyMCE?**

We discovered this in version 5.8.1, and it continues through dev today, April 5, 2023.

TheSpyder:
We discovered this one ourselves recently as well. We're hoping to get a fix in place soon as this is an accessibility issue.

Yona-Appletree:
Glad to hear you're working on it!

I created a PR that fixes it, though I'm not sure I did it right: https://github.com/tinymce/tinymce/pull/8619

For other people running into this, especially those who maybe can't upgrade easily to the newest version, you can use the following config to run code on init that works around the issue:

```
init_instance_callback: editor => {
  editor.targetElm.nextElementSibling
    .querySelectorAll('.tox-toolbar-overlord button[data-alloy-tabstop]')
    .forEach(it => it.removeAttribute('data-alloy-tabstop'))
}
```

Demo here: https://fiddle.tiny.cloud/E4iaab/1

PR #10327, opened 2025-05-02
Fix: Font size conversion from px to pt fails for certain values (#10140)

mumberrymountain:
…140)

GitHub issues: #8618

Description of Changes:
Font size conversion from px to pt fails for certain values. 10.6667px should be converted to 8pt, but is instead converted to 7.5pt, and causes #8618 problem. I changed parseInt to parseFloat to fix this bug.


TheSpyder:
@coderabbitai ignore

PR #10328, opened 2025-05-02
TINY-11912: fix `'contextform'` `onSetup` API

lorenzo-pomili:
Related Ticket: TINY-11912

Description of Changes:
while `onAction(api)` refers to the whole `tox-context-form__group` that contains the input `onSetup(api)` just refers to the input itself, this mean that if we use `api.setInputEnabled(false);` and then we have a button that checks `formApi.isInputEnabled()` it will return `true` instead of `false` because the first will disable the entire `tox-context-form__group` but the second will check just the input.

I chose to have the parent in the `onSetup` instead of taking the `input` on `onAction` because the buttons `onAction` is a shared component that is used from many forms (file: `modules/tinymce/src/themes/silver/main/ts/ui/context/ContextFormButtons.ts`), and since different forms could have more than 1 input having just it could be limiting.

P.S. I used AI to improve the changelog entry, if it's not correct anyway next time I'll try to generate it directly with AI :sweat_smile: 

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

## Summary by CodeRabbit

- **Bug Fixes**
	- Resolved an issue where the context form setup targeted the wrong element, ensuring more accurate handling during form initialization.  
- **Tests**
	- Added automated tests verifying enable/disable behavior and command synchronization for context form text inputs and sliders in the editor.

### DIFF
--- a/modules/tinymce/src/themes/silver/main/ts/ui/context/ContextFormTextInput.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/ui/context/ContextFormTextInput.ts
@@ -20,10 +20,8 @@ export const renderContextFormTextInput = (
   const getFormParentApi = (comp: AlloyComponent, focusfallbackElement?: SugarElement<HTMLElement>) => {
     const parent = Traverse.parent(comp.element);
     const parentCompOpt = parent.bind((parent) => comp.getSystem().getByDom(parent).toOptional());
-    return parentCompOpt.map((parentComp) => ContextFormApi.getFormApi<string>(parentComp, valueState, focusfallbackElement)).getOrThunk(() => {
-      // Fallback to the current component if no parent is found
-      return ContextFormApi.getFormApi<string>(comp, valueState, focusfallbackElement);
-    });
+    return parentCompOpt.map((parentComp) => ContextFormApi.getFormApi<string>(parentComp, valueState, focusfallbackElement))
+      .getOr(ContextFormApi.getFormApi<string>(comp, valueState, focusfallbackElement));
   };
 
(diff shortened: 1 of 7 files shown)


spocke (changes requested):
This needs more work if I add a log to `ContextFormApi` then it will sometimes log `input` sometimes the `div` group it should be either or.

lorenzo-pomili:
@spocke I just realized that I had used `getOr` instead of `getOrThunk` [here](https://github.com/tinymce/tinymce/pull/10328/files#diff-2f6e9d04fad67bf48e929a6e7ab5621763b03dae87d392e25a5095cab04b3c0dR24), so the `ContextFormApi.getFormApi` where called 2 times anyway, I changed it and now the problem should be solved, can you have a check?

modules/tinymce/src/themes/silver/main/ts/ui/context/ContextFormTextInput.ts
```diff
@@ -17,7 +17,11 @@ export const renderContextFormTextInput = (
   valueState: Singleton.Value<string>
 ): SketchSpec => {
   const editorOffCell = Cell(Fun.noop);
-  const getFormApi = (comp: AlloyComponent) => ContextFormApi.getFormApi<string>(comp, valueState);
+  const getFormParentApi = (comp: AlloyComponent, focusfallbackElement?: SugarElement<HTMLElement>) => {
```
spocke, line 20: Should we move this to the `ContextFormApi` is this an issue with other inputs or just text input can we add tests to those if they are not covered?

TheSpyder (approved):

spocke (approved):

PR #10333, opened 2025-05-06
TINY-11997: fix caret position on Firefox

lorenzo-pomili:
Related Ticket: TINY-11997

Description of Changes:
I fixed the bug with the selection overrides in Firefox but I couldn't test it since the tests for this override aren't working for Firefox (the `TinySelections.setRawSelection` used to simulate the `tab` doesn't work in Firefox), so I just improved a little bit the one for Chrome

EDIT: I removed the check for Firefox since the check `editor.selection.getNode() === editor.getBody()` doesn't need any additional check

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/SelectionOverrides.ts
+++ b/modules/tinymce/src/core/main/ts/SelectionOverrides.ts
@@ -165,12 +165,14 @@ const SelectionOverrides = (editor: Editor): SelectionOverrides => {
 
     editor.on('focusin', (e) => {
       if (fakeCaret.isShowing() && e.target !== editor.getBody() && !editor.dom.isEditable(e.target.parentNode)) {
-        if (isGecko) {
-          editor.selection.select(e.target);
-        }
 
         fakeCaret.hide();
 
+        if (isGecko && editor.selection.getRng().commonAncestorContainer === editor.getBody()) {
+          const rng = editor.dom.createRng();
+          rng.selectNode(e.target);
+          editor.selection.setRng(rng);
+        }
         const rng = setElementSelection(editor.selection.getRng(), true);
         if (rng) {
           editor.selection.setRng(rng);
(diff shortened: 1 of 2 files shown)


hamza0867 (approved):

modules/tinymce/src/core/main/ts/SelectionOverrides.ts
```diff
@@ -162,8 +162,14 @@ const SelectionOverrides = (editor: Editor): SelectionOverrides => {
 
     editor.on('focusin', (e) => {
       if (fakeCaret.isShowing() && e.target !== editor.getBody() && !editor.dom.isEditable(e.target.parentNode)) {
+
         fakeCaret.hide();
 
+        if (editor.selection.getNode() === editor.getBody()) {
```
spocke, line 168: Since you can have nested non-editables we can't really compare against body.

We could instead do a check like this, so we check is the selection inside the focusin target element. If it's not then we select the contents of that element and collapse to the start of it. A rng.selectNode(e.target); is a bit weird since that selects the container not the content inside the container.

```
        if (!e.target.contains(editor.selection.getNode())) {
          editor.selection.select(e.target, true);
          editor.selection.collapse(true);
        }
```

So the full version would be like this we need a extra NodeType.isElement check since the `e.target` is unknown if we don't assert the type there and then we can't do a contains on it:

```
    editor.on('focusin', (e) => {
      if (fakeCaret.isShowing() && e.target !== editor.getBody() && !editor.dom.isEditable(e.target.parentNode) && NodeType.isElement(e.target)) {
        fakeCaret.hide();

        if (!e.target.contains(editor.selection.getNode())) {
          editor.selection.select(e.target, true);
          editor.selection.collapse(true);
        }

        const rng = setElementSelection(editor.selection.getRng(), true);
        if (rng) {
          editor.selection.setRng(rng);
        }
      }
    });
```

spocke (approved):

PR #10341, opened 2025-05-08
TINY-12073: fix nested font sizes were causing too much spacing between lines

michalnieruchalski-tiugo:
Related Ticket: 

Description of Changes:
* Placeholder text

Pre-checks:
* [ ] Changelog entry added

Review:
* [ ] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/newline/NewLineUtils.ts
+++ b/modules/tinymce/src/core/main/ts/newline/NewLineUtils.ts
@@ -1,5 +1,5 @@
 import { Arr, Fun, Obj, Optional, Optionals, Unicode } from '@ephox/katamari';
-import { Css, SugarElement } from '@ephox/sugar';
+import { Attribute, Css, Insert, Remove, SugarElement, Traverse } from '@ephox/sugar';
 
 import DOMUtils from '../api/dom/DOMUtils';
 import DomTreeWalker from '../api/dom/TreeWalker';
@@ -215,6 +215,9 @@ const createNewBlock = (
         }
       }
     } while ((node = node.parentNode) && node !== editableRoot);
+
+    flattenFontSize(SugarElement.fromDom(block), SugarElement.fromDom(caretNode));
+    mergeSameTypeElements(SugarElement.fromDom(block), SugarElement.fromDom(caretNode));
   }
 
   setForcedBlockAttrs(editor, block);
@@ -222,6 +225,47 @@ const createNewBlock = (
   emptyBlock(caretNode);
 
   return block;
+};
+
+const flattenFontSize = (block: SugarElement, node: SugarElement, fontSize: string | null = null) => {
+  if (block.dom === node.dom) {
+    return;
+  }
+
+  if (fontSize !== null) {
+    Css.remove(node, 'font-size');
+    Attribute.remove(node, 'data-mce-style');
+  } else {
+    fontSize = Css.getRaw(node, 'font-size').getOr(null);
+  }
+
+  const parent = Traverse.parentElement(node);
+  if (parent.isSome()) {
+    flattenFontSize(block, parent.getOrDie(), fontSize);
+  }
+};
+
+const mergeSameTypeElements = (block: SugarElement, node: SugarElement) => {
+  if (block.dom === node.dom) {
+    return;
+  }
+
+  const parent = Traverse.parentElement(node).getOrNull();
+  if (parent === null || parent.dom === block.dom) {
+    return;
+  }
+  const grandparent = Traverse.parentElement(parent).getOrNull();
+  if (grandparent === null) {
+    return;
+  }
+
+  if (Attribute.hasNone(parent) && parent.dom.nodeName === node.dom.nodeName) {
+    Remove.remove(parent);
+    Insert.append(grandparent, node);
+    mergeSameTypeElements(block, grandparent);
+    return;
+  }
+  mergeSameTypeElements(block, parent);
 };
 
 export {
--- a/modules/tinymce/src/core/test/ts/browser/newline/InsertNewLineTest.ts
+++ b/modules/tinymce/src/core/test/ts/browser/newline/InsertNewLineTest.ts
@@ -2,7 +2,7 @@ import { ApproxStructure, Cursors } from '@ephox/agar';
 import { after, afterEach, before, context, describe, it } from '@ephox/bedrock-client';
 import { Arr, Type } from '@ephox/katamari';
 import { PlatformDetection } from '@ephox/sand';
-import { TinyAssertions, TinyDom, TinyHooks, TinySelections, TinyState } from '@ephox/wrap-mcagar';
+import { TinyAssertions, TinyContentActions, TinyDom, TinyHooks, TinySelections, TinyState } from '@ephox/wrap-mcagar';
 import { assert } from 'chai';
 
 import Editor from 'tinymce/core/api/Editor';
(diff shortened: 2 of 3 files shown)


michalnieruchalski-tiugo:
* **#10341** <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10341?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10341?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

modules/tinymce/src/core/main/ts/newline/NewLineUtils.ts
```diff
@@ -224,6 +227,47 @@ const createNewBlock = (
   return block;
 };
 
+const flattenFontSize = (block: SugarElement, node: SugarElement, fontSize: string | null = null) => {
+  if (block.dom === node.dom) {
```
TheSpyder, line 231: If you're going to use Sugar, use Sugar 😉 
```suggestion
  if (Compare.eq(block, node)) {
```

- [ ] And the other two versions of this below

modules/tinymce/src/core/main/ts/newline/NewLineUtils.ts
```diff
@@ -224,6 +227,47 @@ const createNewBlock = (
...
+    Css.remove(node, 'font-size');
+    Attribute.remove(node, 'data-mce-style');
+  } else {
+    fontSize = Css.getRaw(node, 'font-size').getOr(null);
+  }
+
+  const parent = Traverse.parentElement(node);
+  if (parent.isSome()) {
+    flattenFontSize(block, parent.getOrDie(), fontSize);
+  }
```
TheSpyder, line 245:
```suggestion
  Traverse.parentElement(node).each(parent => flattenFontSize(block, parent, fontSize);
```

modules/tinymce/src/core/main/ts/newline/NewLineUtils.ts
```diff
@@ -224,6 +227,47 @@ const createNewBlock = (
...
+
+  const parent = Traverse.parentElement(node).getOrNull();
+  if (parent === null || parent.dom === block.dom) {
+    return;
+  }
+  const grandparent = Traverse.parentElement(parent).getOrNull();
+  if (grandparent === null) {
+    return;
+  }
```
TheSpyder, line 260: I'll leave rewriting these to avoid `getOrNull` to you

modules/tinymce/src/core/main/ts/newline/NewLineUtils.ts
```diff
@@ -224,6 +227,47 @@ const createNewBlock = (
...
+
+  if (Attribute.hasNone(parent) && parent.dom.nodeName === node.dom.nodeName) {
```
TheSpyder, line 262:
```suggestion
  if (Attribute.hasNone(parent) && Node.name(parent) === Node.name(node)) {
```

TheSpyder (changes requested):

modules/tinymce/src/core/test/ts/browser/newline/InsertNewLineTest.ts
```diff
@@ -1662,4 +1662,96 @@ describe('browser.tinymce.core.newline.InsertNewLineTest', () => {
       });
     });
   });
+
+  context('TINY-12073: nested inline elements', () => {
```
spocke, line 1666: You can skip the ticket ID on the context since if we find a bug in this code then later we add a new ticket and add that to the context. Specific `it` calls can still have the ticket id.

spocke (changes requested):

modules/tinymce/src/core/main/ts/newline/NewLineUtils.ts
```diff
@@ -224,6 +227,47 @@ const createNewBlock = (
...
+const mergeSameTypeElements = (block: SugarElement, node: SugarElement) => {
+  if (block.dom === node.dom) {
+    return;
+  }
+
+  const parent = Traverse.parentElement(node).getOrNull();
+  if (parent === null || parent.dom === block.dom) {
+    return;
+  }
+  const grandparent = Traverse.parentElement(parent).getOrNull();
```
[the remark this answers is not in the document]
michalnieruchalski-tiugo: I'm not sure if I fully understand you advice, so please take a look at the current implementation. I had to add a new method to sugar's Traverse, as `parents` returns an array of `Node` and I can't remove attribute from `Node` using `Attributes.remove`

modules/sugar/src/main/ts/ephox/sugar/api/search/Traverse.ts
```diff
@@ -67,6 +67,27 @@ const parents = (element: SugarElement<Node>, isRoot?: (e: SugarElement<Node>) =
   return ret;
 };
 
+const parentsElement = (element: SugarElement<Element>, isRoot?: (e: SugarElement<Element>) => boolean): SugarElement<Element>[] => {
```
spocke, line 70: I don't think we need this since there is a `PredicateFilter.ancestors` function that filters and stops of a root. If you can everything down to a root you just apply that with `Fun.always` to get them all.

modules/tinymce/src/core/main/ts/newline/ReduceNestedFonts.ts
```diff
@@ -0,0 +1,29 @@
+import { Arr, Optional } from '@ephox/katamari';
+import { Attribute, Compare, Css, Remove, SugarElement, Traverse, SugarNode } from '@ephox/sugar';
+
+const reduceFontStyleNesting = (block: Element, node: Element): void => {
+  const blockSugar = SugarElement.fromDom(block);
+  const nodeSugar = SugarElement.fromDom(node);
+  const isEndBlock = (element: SugarElement) => Compare.eq(element, blockSugar);
+  const elements = [ nodeSugar, ...Traverse.parentsElement(nodeSugar, isEndBlock) ];
```
spocke, line 8: If you change this to `PredicateFilter.ancestors(nodeSugar, hasFontSize, isEndBlock)`  and then also do a hasFontSize check on the nodeSugar then we you have a set of elements that you need to process.

Then the first one is the one you keep and the rest if the ones you remove the font size from and unwrap if it doesn't have any attributes anymore. I think we should only unwrap `span` elements since in theory the element could be say `strong` then it has some styling.

modules/tinymce/src/core/main/ts/newline/ReduceNestedFonts.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  Arr.each(elements, (element, index) => {
+    if (fontSize.isSome()) {
+      Css.remove(element, 'font-size');
+      Attribute.remove(element, 'data-mce-style');
+    } else {
+      fontSize = Css.getRaw(element, 'font-size');
+    }
+
+    Optional.from(elements[index - 1]).each((previousElement) => {
```
spocke, line 19: I don't think you need this check and processing of the previous element if you do the filtering and array slicing that I suggested above. It becomes more of a read the data from the dom, act on that data instead of a mix of both in a loop.

spocke (changes requested):

modules/tinymce/src/core/main/ts/newline/ReduceNestedFonts.ts
michalnieruchalski-tiugo: I need previousElement to check if it had the same nodeType as current element. 
It also related to your previous comment about not removing `strong`.

```
<strong>
<strong style="font-size: 10pt;">
</strong>
<strong>
```

would become

```
<strong style="font-size: 10pt;"></strong>
```

however 
```
<strong>
<span style="font-size: 10pt;">
</span>
<strong>
```

would stay intact

michalnieruchalski-tiugo: I like the idea of separation of read the data from dom and act on that data. I will implement it as you wish. It will mean double iteration over the array of elements but it's not the issue since the array will always be rather small.

spocke: Not sure I understand if the structure is like this:
```
<span style="font-size: 10px">
  <em>
    <span style="font-size: 20px">
      <strong>
        <span style="font-size: 30px">|</span>
      </strong>
     </span>
  </em>
</span>
```

You would find the spans 10, 20, 30 keep 30 and unwrap both 10, 20 since they are both elements that will have no attributes after you remove the font size style.

Then that would result in:
```
<em>
  <strong>
    <span style="font-size: 30px">|</span>
  </strong>
</em>
```

Isn't that the structure we want. The parent is kind of irrelevant.

michalnieruchalski-tiugo: The result would be:

```
<span>
  <em>
    <span>
      <strong>
        <span style="font-size: 30px">|</span>
      </strong>
     </span>
  </em>
</span>
```

But that's a fair point. I should remove the empty spans. That's a good test case.

But I think my example should also work, so if there's

```
<strong>
    <strong style="font-size: 10pt;">
    </strong>
<strong>
```

strong tags should be merged

spocke (approved):

TheSpyder (approved):

PR #10359, opened 2025-05-15
TINY-12118: fix context form back behavior for inline editors

lorenzo-pomili:
Related Ticket: TINY-12118

Description of Changes:
Here the problem was that in non-inline editors the active element returned by `Focus.active(SugarShadowDom.getRootNode(comp.element))` was always the `iframe` and focussing it again the focus remains on the real focussed element (an image for example) but in the inline editors it returned the `div` container and focussing it back the focus goes on the `div` so if the selection was on an image the focus get lost and the toolbar get dismissed.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

## Summary by CodeRabbit

- **Bug Fixes**
  - Resolved an issue where the context form subtoolbar was dismissed unexpectedly when navigating back in inline editors, ensuring the subtoolbar remains visible as expected.

- **Tests**
  - Introduced new automated tests to verify context form subtoolbar behavior in both inline and non-inline editors, ensuring correct toolbar visibility when using navigation actions.

### DIFF
--- a/.changes/unreleased/tinymce-TINY-12118-2025-05-16.yaml
+++ b/.changes/unreleased/tinymce-TINY-12118-2025-05-16.yaml
@@ -1,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: The `contextform` subtoolbar was being dismissed when navigating back in inline editors.
+time: 2025-05-16T14:25:44.523886887+02:00
+custom:
+    Issue: TINY-12118
--- a/.changes/unreleased/tinymce-TINY-12118-2025-05-16.yaml
+++ b/.changes/unreleased/tinymce-TINY-12118-2025-05-16.yaml
@@ -1,6 +1,6 @@
 project: tinymce
 kind: Fixed
-body: The `contextform` subtoolbar was being dismissed when navigating back in inline editors.
+body: The `contextform` subtoolbar is no longer dismissed when using the back button in inline editors.
 time: 2025-05-16T14:25:44.523886887+02:00
 custom:
     Issue: TINY-12118
--- a/modules/tinymce/src/themes/silver/main/ts/ui/context/ContextUi.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/ui/context/ContextUi.ts
@@ -28,7 +28,7 @@ interface ContextToolbarSpec {
   readonly sink: AlloyComponent;
   readonly onHide: () => void;
   readonly onBack: () => void;
-  readonly focusElement: (el: SugarElement<any>) => void;
+  readonly focusElement: (el: SugarElement<HTMLElement>) => void;
 }
 
 export interface ContextToolbarRenderResult {
--- a/modules/tinymce/src/themes/silver/main/ts/ui/context/ContextUi.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/ui/context/ContextUi.ts
@@ -3,7 +3,7 @@ import {
   SketchSpec
 } from '@ephox/alloy';
 import { Arr, Cell, Id, Optional, Result } from '@ephox/katamari';
-import { Class, Compare, Css, EventArgs, Focus, SugarElement, SugarShadowDom, Width } from '@ephox/sugar';
+import { Class, Compare, Css, EventArgs, Focus, SugarElement, SugarNode, SugarShadowDom, Width } from '@ephox/sugar';
 
 import * as ContextToolbarFocus from './ContextToolbarFocus';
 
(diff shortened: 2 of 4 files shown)


.changes/unreleased/tinymce-TINY-12118-2025-05-16.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: The `contextform` subtoolbar was being dismissed when navigating back in inline editors.
```
TheSpyder, line 3:
```suggestion
body: The `contextform` subtoolbar is no longer dismissed when using the back button in inline editors.
```

modules/tinymce/src/themes/silver/main/ts/ui/context/ContextUi.ts
```diff
@@ -121,10 +121,13 @@ const renderContextToolbar = (spec: ContextToolbarSpec): ContextToolbarRenderRes
...
         AlloyEvents.run<ForwardSlideEvent>(forwardSlideEvent, (comp, se) => {
           InlineView.getContent(comp).each((oldContents) => {
+            const isIframe = SugarNode.isTag('iframe');
+            const activeElement = Focus.active(SugarShadowDom.getRootNode(comp.element));
+            const hasSelectionInside = activeElement.exists((e) => !isIframe(e) && e.dom.contains(se.event.currentSelection));
             stack.set(stack.get().concat([
               {
                 bar: oldContents,
-                focus: Focus.active(SugarShadowDom.getRootNode(comp.element))
+                focus: hasSelectionInside ? Focus.active(SugarElement.fromDom(se.event.currentSelection)) : activeElement
```
TheSpyder, line 130: Is there any reason to not just use `se.event.currentSelection` all the time?

TheSpyder (approved):

modules/tinymce/src/themes/silver/main/ts/ui/context/ContextUi.ts
lorenzo-pomili: using always `se.event.currentSelection` would makes ['TINY-11748: Launch button should be able to navigate to a context toolbars and navigateback and retain toolbar focus'](https://github.com/tinymce/tinymce/blob/main/modules/tinymce/src/themes/silver/test/ts/browser/editor/contexttoolbar/ContextToolbarTest.ts#L173) fail, this is because in that case the `se.event.currentSelection` is the `link` but the `activeElement` (`Focus.active(SugarShadowDom.getRootNode(comp.element))`) is the button in the context toolbar, so using always `se.event.currentSelection` would mean not to preserve the focus on the context toolbar's buttons

.changes/unreleased/tinymce-TINY-12118-2025-05-16.yaml
lorenzo-pomili: changed

modules/tinymce/src/themes/silver/main/ts/ui/context/ContextUi.ts
spocke, line 130: I don't think this is the right approach since the currentSelection might be something that we can't rally focus on and it might also be removed if you move back and forth or be removed by the actions performed by the form.
Selection is already stored/restored in tinymce if you use the `editor.focus()` call so I wonder if we could maybe do this by passing in a `focusElement` function to the render `ContextToolbarSpec`.

Then that would look something like:
```
    focusElement: (el) => {
      if (editor.getBody().contains(el.dom)) {
        editor.focus();
      } else {
        Focus.focus(el);
      }
    }
```

So if the element we want to focus back to is the editor we do a call to `editor.focus()` since that will restore the selection as well. If it's something else then we just focus it.

Might need some extra code for restoring focus to editable host elements within the body though hence the `contains` instead of the simple equals.

lorenzo-pomili: refactored

modules/tinymce/src/themes/silver/main/ts/ui/context/ContextUi.ts
```diff
@@ -28,6 +28,7 @@ interface ContextToolbarSpec {
   readonly sink: AlloyComponent;
   readonly onHide: () => void;
   readonly onBack: () => void;
+  readonly focusElement: (el: SugarElement<any>) => void;
```
spocke, line 31: I think we can make this HTMLElement if you change the type of ChangeSlideEvent to that type. I think using just SugarElement is incorrect since HTMLElements are the only ones that can have focus.

spocke (approved):

modules/tinymce/src/themes/silver/main/ts/ui/context/ContextUi.ts
lorenzo-pomili: what about `SugarElement<HTMLELement>`?

lorenzo-pomili: changed to `SugarElement<HTMLELement>`

PR #10363, opened 2025-05-19
TINY-8665: split button are now buttons

ArvinJ-H:
Related Ticket: 
TINY-8665
Description of Changes:
* Placeholder text

Pre-checks:
* [ ] Changelog entry added

Review:
* [ ] Milestone set

### DIFF
--- a/modules/tinymce/src/themes/silver/main/ts/ui/toolbar/Integration.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/ui/toolbar/Integration.ts
@@ -105,7 +105,7 @@ const types: Record<string, BridgeRenderFn<any>> = {
   )
 };
 
-const extractFrom = (spec: ToolbarButton & { type: string }, backstage: UiFactoryBackstage, editor: Editor, btnName: string): Optional<AlloySpec> =>
+const extractFrom = (spec: ToolbarButton & { type: string }, backstage: UiFactoryBackstage, editor: Editor, btnName: string): Optional<AlloySpec | AlloySpec[]> =>
   Obj.get(types, spec.type).fold(
     () => {
       // eslint-disable-next-line no-console
@@ -172,16 +172,11 @@ const createToolbar = (toolbarConfig: RenderToolbarConfig): ToolbarGroupOption[]
   }
 };
 
-const lookupButton = (editor: Editor, buttons: Record<string, any>, toolbarItem: string, allowToolbarGroups: boolean, backstage: UiFactoryBackstage, prefixes: Optional<string[]>): Optional<AlloySpec> =>
+const lookupButton = (editor: Editor, buttons: Record<string, any>, toolbarItem: string, allowToolbarGroups: boolean, backstage: UiFactoryBackstage, prefixes: Optional<string[]>): Optional<AlloySpec | AlloySpec[]> =>
   Obj.get(buttons, toolbarItem.toLowerCase())
     .orThunk(() => prefixes.bind((ps) => Arr.findMap(ps, (prefix) => Obj.get(buttons, prefix + toolbarItem.toLowerCase()))))
     .fold(
       () => Obj.get(bespokeButtons, toolbarItem.toLowerCase()).map((r) => r(editor, backstage)),
-      // TODO: Add back after TINY-3232 is implemented
-      // .orThunk(() => {
-      //   console.error('No representation for toolbarItem: ' + toolbarItem);
-      //   return Optional.none();
-      // ),
       (spec) => {
         if (spec.type === 'grouptoolbarbutton' && !allowToolbarGroups) {
           // TODO change this message when sliding is available
@@ -198,8 +193,12 @@ const identifyButtons = (editor: Editor, toolbarConfig: RenderToolbarConfig, bac
   const toolbarGroups = createToolbar(toolbarConfig);
   const groups = Arr.map(toolbarGroups, (group) => {
     const items = Arr.bind(group.items, (toolbarItem) => {
-      return toolbarItem.trim().length === 0 ? [] :
-        lookupButton(editor, toolbarConfig.buttons, toolbarItem, toolbarConfig.allowToolbarGroups, backstage, prefixes).toArray();
+      if (toolbarItem.trim().length === 0) {
+        return [];
+      }
+      return lookupButton(editor, toolbarConfig.buttons, toolbarItem, toolbarConfig.allowToolbarGroups, backstage, prefixes)
+        .map((spec) => Array.isArray(spec) ? spec : [ spec ])
+        .getOr([]);
     });
     return {
       title: Optional.from(editor.translate(group.name)),
--- a/modules/tinymce/src/themes/silver/main/ts/ui/toolbar/button/ToolbarButtons.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/ui/toolbar/button/ToolbarButtons.ts
@@ -3,7 +3,9 @@ import {
   GuiFactory,
   Keying, Memento, NativeEvents, Replacing, SketchSpec, SplitDropdown as AlloySplitDropdown, SystemEvents, TieredData, TieredMenuTypes, Toggling,
   Tooltipping,
-  Unselecting
+  Unselecting,
+  AlloySpec,
+  TieredMenu
 } from '@ephox/alloy';
 import { Toolbar } from '@ephox/bridge';
 import { Arr, Cell, Fun, Future, Id, Merger, Optional, Type } from '@ephox/katamari';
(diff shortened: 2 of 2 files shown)


modules/tinymce/src/themes/silver/main/ts/ui/toolbar/button/ToolbarButtons.ts
```diff
@@ -303,140 +303,145 @@ const fetchChoices = (getApi: (comp: AlloyComponent) => Toolbar.ToolbarSplitButt
...
-const renderSplitButton = (spec: Toolbar.ToolbarSplitButton, sharedBackstage: UiFactoryBackstageShared, btnName?: string): SketchSpec => {
-  const tooltipString = Cell<string>(spec.tooltip.getOr(''));
+const getSplitButtonApi = (component: AlloyComponent): Toolbar.ToolbarSplitButtonInstanceApi => ({
+  isEnabled: () => !Disabling.isDisabled(component),
+  setEnabled: (state: boolean) => Disabling.set(component, !state),
+  setText: (text: string) => AlloyTriggers.emitWith(component, updateMenuText, { text }),
+  setIcon: (icon: string) => AlloyTriggers.emitWith(component, updateMenuIcon, { icon }),
+  setIconFill: (_id: string, _value: string) => {},
+  isActive: () => Fun.never(),
+  setActive: (_state: boolean) => {},
```
spocke, line 313: Why are these no longer implemented?

ArvinJ-H: mmm, I didn't notice that would cause the button to miss active state, adding back now

modules/oxide/src/less/theme/components/toolbar-button/toolbar-split-button.less
```diff
@@ -49,6 +49,8 @@
 
   .tox-split-button__chevron {
     width: @toolbar-button-chevron-width;
+    border-left: 1px solid #e3e3e3 !important;
```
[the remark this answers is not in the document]
ArvinJ-H: This line is currently more or less pending, since the design ticket is still in review. Once the design is confirmed I'll update this

modules/tinymce/src/themes/silver/main/ts/ui/toolbar/button/ToolbarButtons.ts
```diff
@@ -303,140 +304,183 @@ const fetchChoices = (getApi: (comp: AlloyComponent) => Toolbar.ToolbarSplitButt
...
+      }
+    }
   });
+
+  return [
+    AlloyButton.sketch({
+      ...renderCommonStructure(
+        spec.icon,
+        spec.text,
+        Optional.none(),
```
kimwoodfield, line 434: Why do we pass `Optional.none()` like this?

modules/tinymce/src/themes/silver/test/ts/browser/skin/OxideColorSwatchMenuTest.ts
```diff
@@ -88,15 +88,33 @@ describe('browser.tinymce.themes.silver.skin.OxideColorSwatchMenuTest', () => {
 
   const pOpenAndGetMenu = (title: string) =>
     () => {
-      Mouse.clickOn(SugarBody.body(), `[data-mce-name="${title}"] .tox-split-button__chevron`);
+      let ariaLabelPrefix = '';
```
kimwoodfield, line 91: Might as well pull out this logic which is identical between `pOpenAndGetMenu` and `pCloseMenu` and store it in a new function called `getAriaLabelPrefix` and just call it in both places?

kimwoodfield (approved):
Looks good to me for the most part but it would be good to get some more eyes across it to be safe.

modules/tinymce/src/themes/silver/main/ts/ui/toolbar/button/ToolbarButtons.ts
```diff
@@ -303,140 +304,183 @@ const fetchChoices = (getApi: (comp: AlloyComponent) => Toolbar.ToolbarSplitButt
...
+  const isChevron = Class.has(element, 'tox-split-button__chevron');
+
+  const mainOpt = isChevron ?
+    Traverse.prevSibling(element).bind((el) => system.getByDom(el).toOptional()) :
+    Optional.some(component);
+
+  const chevronOpt = isChevron ?
+    Optional.some(component) :
+    Traverse.nextSibling(element).bind((el) => system.getByDom(el).toOptional().filter((comp) => Class.has(comp.element, 'tox-split-button__chevron')));
+
```
shanmen-tiny, line 319: I think we try to avoid using DOM positions as such to identify related components 🤔 
Should we use `Memento` instead, or maybe use an internal alloy event to tell them both to update the tooltips?

modules/tinymce/src/themes/silver/main/ts/ui/toolbar/button/ToolbarButtons.ts
```diff
@@ -303,140 +304,183 @@ const fetchChoices = (getApi: (comp: AlloyComponent) => Toolbar.ToolbarSplitButt
...
-          }))
-      ),
-    setIcon: (icon: string) =>
-      SelectorFind.descendant(comp.element, 'span').each((button) =>
-        comp.getSystem().getByDom(button).each((buttonComp) =>
-          AlloyTriggers.emitWith(buttonComp, updateMenuIcon, {
-            icon
-          }))
-      ),
+    }),
```
shanmen-tiny, line 334: Is this a descendant of the chevron 🤔

modules/tinymce/src/themes/silver/main/ts/ui/toolbar/button/ToolbarButtons.ts
```diff
@@ -303,140 +304,183 @@ const fetchChoices = (getApi: (comp: AlloyComponent) => Toolbar.ToolbarSplitButt
...
+    isActive: () => mainOpt.exists((c) => Toggling.isOn(c)),
+    setActive: (state: boolean) => mainOpt.each((c) => Toggling.set(c, state)),
     setTooltip: (tooltip: string) => {
-      const translatedTooltip = sharedBackstage.providers.translate(tooltip);
-      Attribute.set(comp.element, 'aria-label', translatedTooltip);
       tooltipString.set(tooltip);
+      mainOpt.each((c) => Attribute.set(c.element, 'aria-label', sharedBackstage.providers.translate(tooltip)));
+      chevronOpt.each((c) => Attribute.set(c.element, 'aria-label', sharedBackstage.providers.translate(`${tooltip} menu`)));
```
shanmen-tiny, line 340: Do we need full tooltips here? Or just the button's tooltip without the last used state like `Text color menu`?

modules/oxide/src/less/theme/globals/global-variables.less
```diff
@@ -30,6 +30,7 @@
 
 // Colors
 @border-color: darken(@background-color, 6.5%);
+@border-color-light: #e3e3e3;
```
TheSpyder: Everything else in this set of colours is calculated from the background colour in order to work better with a variety of skins. can we do that here as well instead of hard coding it?

modules/tinymce/src/themes/silver/main/ts/ui/toolbar/button/ToolbarButtons.ts
TheSpyder: That's correct. Memento, or a `SelectorFind.descendant` search within the parent element for the class we're looking for.

modules/tinymce/src/themes/silver/test/ts/browser/skin/OxideColorSwatchMenuTest.ts
```diff
@@ -88,15 +88,33 @@ describe('browser.tinymce.themes.silver.skin.OxideColorSwatchMenuTest', () => {
...
+      if (title === 'forecolor') {
+        ariaLabelPrefix = 'Text color';
+      } else if (title === 'backcolor') {
+        ariaLabelPrefix = 'Background color';
+      } else {
+        // For custom buttons like swatch-button, just use the title
+        ariaLabelPrefix = title;
+      }
+      Mouse.clickOn(SugarBody.body(), `button[data-mce-name="${title}-chevron"][aria-label^="${ariaLabelPrefix}"]`);
```
TheSpyder: We stopped using ARIA prefixes. It's not clear why you even need that, when you have `-chevron` on the name already.

- [x] remove the aria-label prefix check

modules/oxide/src/less/theme/components/toolbar-button/toolbar-split-button.less
```diff
@@ -49,6 +49,8 @@
 
   .tox-split-button__chevron {
     width: @toolbar-button-chevron-width;
+    border-left: 1px solid @border-color-light;
```
spocke: This border or margin/padding of 1 px needs to be added to the focus state otherwise the button or even the entire toolbar bobs back and forth if focus is moved into the split button since the default focus is `border: 0`.

- [x] Fix focus style for split button

spocke (approved):

modules/tinymce/src/core/test/ts/browser/fmt/TextDecorationColorTest.ts
```diff
@@ -32,13 +32,13 @@ describe('browser.tinymce.core.fmt.TextDecorationColorTest', () => {
...
   const pApplyForecolor = async (editor: Editor) => {
-    TinyUiActions.clickOnToolbar(editor, '[aria-label^="Text color"] > .tox-tbtn + .tox-split-button__chevron');
+    TinyUiActions.clickOnToolbar(editor, 'button.tox-split-button__chevron[aria-label^="Text color"]');
     await TinyUiActions.pWaitForUi(editor, '.tox-swatches');
     TinyUiActions.clickOnUi(editor, `div[data-mce-color="${textColorHex.toUpperCase()}"]`);
   };
 
   const pRemoveForecolor = async (editor: Editor) => {
-    TinyUiActions.clickOnToolbar(editor, '[aria-label^="Text color"] > .tox-tbtn + .tox-split-button__chevron');
+    TinyUiActions.clickOnToolbar(editor, 'button.tox-split-button__chevron[aria-label^="Text color"]');
```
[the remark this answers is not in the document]
ArvinJ-H: yep

modules/tinymce/src/core/test/ts/browser/focus/HighlightOnFocusTest.ts
```diff
@@ -72,7 +72,7 @@ describe('browser.tinymce.core.focus.HighlightOnFocus', () => {
     it('TINY-9277: Content area should be highlighted on focus and removed when shifted to toolbar', async () => {
       const editor = hook.editor();
       assertHighlightOnFocus(editor);
-      TinyUiActions.clickOnToolbar(editor, '[aria-label^="Text color"] > .tox-tbtn + .tox-split-button__chevron');
+      TinyUiActions.clickOnToolbar(editor, 'button[data-mce-name="forecolor-chevron"][aria-label^="Text color"]');
```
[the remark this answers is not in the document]
ArvinJ-H: okay, maybe not, the aria is already check at ColorAriaLabelTest.ts. I'll remove this one

modules/tinymce/src/themes/silver/main/ts/ui/toolbar/button/ToolbarButtons.ts
ArvinJ-H: The main button of a split button doesn't use a tooltip from `renderCommonStructure` because it gets its tooltip handled separately through the custom `Tooltipping.config` passed in the `behaviours` array. The split button needs special tooltip logic that updates dynamically when colors change.

modules/tinymce/src/themes/silver/main/ts/ui/toolbar/button/ToolbarButtons.ts
ArvinJ-H: ah yes, it should be a static one.

modules/tinymce/src/themes/silver/test/ts/browser/skin/OxideColorSwatchMenuTest.ts
ArvinJ-H: This was initially added to check if aria exist, similar to another test file. Looks like there was another test that handles aria. I'll remove this one.

modules/oxide/src/less/theme/components/toolbar-button/toolbar-split-button.less
ArvinJ-H: I ran into what looks like some weird cache issues, this will take a bit longer than I thought

.changes/unreleased/tinymce-TINY-8665-2025-05-21.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Changed
+body: Split buttons now render as separate action and chevron buttons
```
tiny-ben-tran:
```suggestion
body: Split buttons are now rendered as separate action and chevron buttons
```

modules/tinymce/src/themes/silver/main/ts/ui/toolbar/Integration.ts
```diff
@@ -172,34 +172,32 @@ const createToolbar = (toolbarConfig: RenderToolbarConfig): ToolbarGroupOption[]
...
-          console.warn(`Ignoring the '${toolbarItem}' toolbar button. Group toolbar buttons are only supported when using floating toolbar mode and cannot be nested.`);
-          return Optional.none();
-        } else {
-          return extractFrom(spec, backstage, editor, toolbarItem.toLowerCase());
-        }
+const lookupButton = (editor: Editor, buttons: Record<string, any>, toolbarItem: string, allowToolbarGroups: boolean, backstage: UiFactoryBackstage, prefixes: Optional<string[]>): Optional<AlloySpec | AlloySpec[]> => Obj.get(buttons, toolbarItem.toLowerCase())
+  .orThunk(() => prefixes.bind((ps) => Arr.findMap(ps, (prefix) => Obj.get(buttons, prefix + toolbarItem.toLowerCase()))))
+  .fold(
+    () => Obj.get(bespokeButtons, toolbarItem.toLowerCase()).map((r) => r(editor, backstage)),
+    (spec) => {
```
tiny-ben-tran, line 184: Why deleted this comment?

modules/tinymce/src/themes/silver/main/ts/ui/toolbar/button/ToolbarButtons.ts
```diff
@@ -303,140 +304,197 @@ const fetchChoices = (getApi: (comp: AlloyComponent) => Toolbar.ToolbarSplitButt
...
+  const system = component.getSystem();
+  const element = component.element;
+  const isChevron = Class.has(element, 'tox-split-button__chevron');
+  const parentOpt = Traverse.parent(element);
+
+  const mainOpt = isChevron ?
+    parentOpt.bind((parent) =>
+      SelectorFind.descendant(parent, 'button:not(.tox-split-button__chevron)')
+    ).bind((el) => system.getByDom(el).toOptional()) :
+    Optional.some(component);
```
shanmen-tiny: I think this always return the first matching descendant in the toolbar group. What if it's the nth button in the toolbar group? 🤔

modules/tinymce/src/themes/silver/main/ts/ui/toolbar/button/ToolbarButtons.ts
```diff
@@ -303,140 +304,197 @@ const fetchChoices = (getApi: (comp: AlloyComponent) => Toolbar.ToolbarSplitButt
...
     fetch: fetchChoices(getApi, spec, sharedBackstage.providers),
-
-    parts: {
-      // FIX: hasIcons
-      menu: MenuParts.part(false, spec.columns, spec.presets)
+    getHotspot: (comp) => {
+      const parentOpt = Traverse.parent(comp.element);
+      return parentOpt.bind((parent) =>
+        SelectorFind.descendant(parent, 'button:not(.tox-split-button__chevron)')
+      ).bind((el) => comp.getSystem().getByDom(el).toOptional());
```
shanmen-tiny: See the other comment

modules/tinymce/src/themes/silver/main/ts/ui/toolbar/button/ToolbarButtons.ts
ArvinJ-H: yeah that sounds like an issue, but in tests this never came as an issue. Maybe I should use `SelectorFind.child`?

shanmen-tiny: The `SelectorFind.child` method still returns the first matching element. Is there a way to target a unique selector instead?

We should probably include the following sanity tests, especially since these behaviors have changed:
- The dropdown is positioned close to the corresponding button, and
- The "active" state is updated correctly.

The current approach always anchors to the first matching split button and updates its active state.

ArvinJ-H: ah yeah, missed `sibling`, this should only return the siblings of the current element

modules/tinymce/src/plugins/advlist/test/ts/browser/ContentEditableFalseTest.ts
```diff
@@ -71,8 +71,8 @@ ${listContent}
   ];
 
   const checkToolbarDisabled = (editor: Editor, listType: string) => {
-    UiFinder.exists(SugarBody.body(), `[aria-label="${listType}"][aria-disabled="true"] > .tox-tbtn`);
-    TinyUiActions.clickOnToolbar(editor, `[aria-label="${listType}"][aria-disabled="true"] > .tox-tbtn`);
+    UiFinder.exists(SugarBody.body(), `[aria-label="${listType}"][aria-disabled="true"]`);
+    TinyUiActions.clickOnToolbar(editor, `[aria-label="${listType}"][aria-disabled="true"]`);
```
tiny-ben-tran:
```suggestion
		const buttonSelector = `[aria-label="${listType}"][aria-disabled="true"]`;
    UiFinder.exists(SugarBody.body(), buttonSelector);
    TinyUiActions.clickOnToolbar(editor, buttonSelector);
```

modules/tinymce/src/themes/silver/main/ts/ui/toolbar/button/ToolbarButtons.ts
tiny-ben-tran: Selector as `button:not(.tox-split-button__chevron)` is not ideal. Does `mainOpt` have a CSS class  that can be used?

modules/tinymce/src/themes/silver/main/ts/ui/toolbar/button/ToolbarButtons.ts
tiny-ben-tran: Does this function set `fill` to the icon of both `mainOpt` and `chevron`?

modules/tinymce/src/themes/silver/main/ts/ui/toolbar/button/ToolbarButtons.ts
```diff
@@ -303,140 +304,197 @@ const fetchChoices = (getApi: (comp: AlloyComponent) => Toolbar.ToolbarSplitButt
...
+    }),
+    isActive: () => mainOpt.exists((c) => Toggling.isOn(c)),
+    setActive: (state: boolean) => mainOpt.each((c) => Toggling.set(c, state)),
     setTooltip: (tooltip: string) => {
-      const translatedTooltip = sharedBackstage.providers.translate(tooltip);
-      Attribute.set(comp.element, 'aria-label', translatedTooltip);
       tooltipString.set(tooltip);
+      mainOpt.each((c) => Attribute.set(c.element, 'aria-label', sharedBackstage.providers.translate(tooltip)));
+      chevronOpt.each((c) => {
+        Attribute.set(c.element, 'aria-label', chevronTooltip);
```
tiny-ben-tran: Should the chevron tooltip be translated?

modules/tinymce/src/themes/silver/main/ts/ui/toolbar/button/ToolbarButtons.ts
```diff
@@ -303,140 +304,197 @@ const fetchChoices = (getApi: (comp: AlloyComponent) => Toolbar.ToolbarSplitButt
...
   const editorOffCell = Cell(Fun.noop);
-  const specialisation = {
-    getApi,
-    onSetup: spec.onSetup
+  const tooltipString = Cell<string>(spec.tooltip.getOr(''));
+
+  const getChevronTooltip = () => {
+    const baseTooltip = spec.tooltip.getOr('Text color');
+    let staticTooltip;
+    if (baseTooltip.includes('Background color')) {
```
tiny-ben-tran: Why is the chevron tooltip being renamed?

tiny-ben-tran (approved):

modules/tinymce/src/themes/silver/main/ts/ui/toolbar/button/ToolbarButtons.ts
ArvinJ-H: so turns out i pushed an archived stash, it is not using selectors anymore, its using `Traverse` and sibling.  `SelectorFind` had the issue you mentioned @shanmen-tiny , but i thought this wasn't really the case because it worked great... a while ago.

modules/tinymce/src/themes/silver/main/ts/ui/toolbar/button/ToolbarButtons.ts
ArvinJ-H: before we never have a separate tooltip for the chevon button, since they were essentially a single button, chevon shared the same tooltip with main button.

By default for chevon button, the tooltip is `mainButton + menu`, this won't work for the fore/back color button, they will become `Text color black menu`. This won't make any sense both visually and via AT. So setting them separately to  `Text color menu` would be a much better choice.

The implementation has changed since the restore, have a look again. They do the same thing.

modules/tinymce/src/themes/silver/main/ts/ui/toolbar/button/ToolbarButtons.ts
ArvinJ-H: yes, i think somehow i missed this step, doing another round of validation now

modules/tinymce/src/plugins/advlist/test/ts/browser/ContentEditableFalseTest.ts
ArvinJ-H: 👌

modules/tinymce/src/themes/silver/main/ts/ui/toolbar/button/ToolbarButtons.ts
ArvinJ-H: Yeah, the function applies to both the main button and the chevron button because the `applyBoth` function.

tiny-ben-tran: It would be better to have separate functions for changing the color of the icons: one for each part. However, it may not be needed yet

modules/tinymce/src/themes/silver/main/ts/ui/toolbar/button/ToolbarButtons.ts
tiny-ben-tran: It is worth including a tooltip for chevron in the spec instead of renaming like this. If not provided, it will use `mainButton + menu` format

modules/tinymce/src/themes/silver/main/ts/ui/toolbar/button/ToolbarButtons.ts
```diff
@@ -303,140 +304,200 @@ const fetchChoices = (getApi: (comp: AlloyComponent) => Toolbar.ToolbarSplitButt
...
+    setActive: (state: boolean) => mainOpt.each((c) => Toggling.set(c, state)),
     setTooltip: (tooltip: string) => {
-      const translatedTooltip = sharedBackstage.providers.translate(tooltip);
-      Attribute.set(comp.element, 'aria-label', translatedTooltip);
       tooltipString.set(tooltip);
+      mainOpt.each((c) => Attribute.set(c.element, 'aria-label', sharedBackstage.providers.translate(tooltip)));
+      // For chevron, use the explicit chevronTooltip if provided, otherwise fall back to default behavior
+      const chevronTooltipText = spec.chevronTooltip
+        .map((chevronTooltip) => sharedBackstage.providers.translate(chevronTooltip))
+        .getOr(sharedBackstage.providers.translate([ '{0} menu', sharedBackstage.providers.translate(tooltip) ]));
```
spocke: I think change into into this and remove the fallback from the POT files. That way everything becomes unique strings you need to have in the POT files and the menu word can be moved around depending on language.

```suggestion
        .getOr(sharedBackstage.providers.translate(`${tooltip} menu`));
```

- [x] Make sure that this becomes separate strings instead of a composed sentence since some languages might need to write `Menu of insert date` type of strings.

spocke (approved):

.changes/unreleased/tinymce-TINY-8665-2025-05-21.yaml
```diff
@@ -1,6 +1,6 @@
 project: tinymce
 kind: Changed
-body: Split buttons now render as separate action and chevron buttons
+body: Split buttons are now rendered as separate action and chevron buttons
```
kimwoodfield:
```suggestion
body: Split buttons are now rendered as separate action and chevron buttons.
```

modules/tinymce/src/themes/silver/main/ts/ui/toolbar/button/ToolbarButtons.ts
TheSpyder: I don't think there should be a `chevronTooltip` property at all, it should only be generated. This seems like a weird addition to our public API.

modules/tinymce/src/themes/silver/main/ts/ui/toolbar/button/ToolbarButtons.ts
```diff
@@ -303,140 +304,200 @@ const fetchChoices = (getApi: (comp: AlloyComponent) => Toolbar.ToolbarSplitButt
...
+  const element = component.element;
+  const isChevron = Class.has(element, 'tox-split-button__chevron');
+
+  const mainOpt = isChevron ?
+    Traverse.prevSibling(element).bind((el) => system.getByDom(el).toOptional()) :
+    Optional.some(component);
+
+  const chevronOpt = isChevron ?
+    Optional.some(component) :
+    Traverse.nextSibling(element).bind((el) => system.getByDom(el).toOptional().filter((comp) => Class.has(comp.element, 'tox-split-button__chevron')));
```
TheSpyder, line 318: I've just realised this changed the API from dynamic lookup to cached values.

I don't know exactly why, but this feels like it might cause problems in edge cases.

modules/tinymce/src/themes/silver/test/ts/browser/editor/color/ColorAriaLabelTest.ts
```diff
@@ -144,46 +116,72 @@ describe('browser.tinymce.themes.silver.editor.color.ColorAriaLabelTest', () =>
         await pResetColorToDefault(editor, selectors.backcolorSplitButton());
       });
 
-      it('TINY-9697: Forecolor - default color appended to aria-label', () => {
```
TheSpyder, line 147: It looks like this replaced all the tests for the icon section of the split button with chevron menu tests? Don't we need both?

modules/tinymce/src/themes/silver/main/ts/ui/toolbar/button/ToolbarButtons.ts
ArvinJ-H: yeah if DOM changes after API is created, the cache will be unreliable, this and potentially race conditions, I'll revert to dynamic loop up and see if I can add more tests

PR #10367, opened 2025-05-20
TINY-11974: Update return type of `fetchUsers` and add avatar generation logic

kimwoodfield:

Description of Changes:
* Upon using the `UserLookup` API we discovered that knowing which promise in the `Promise<User>[]` belonged to which user was difficult as we would need to wait for a promise to resolve before knowing its `id`. We've revised the return type to be a `Record<UserId, Promise<User>>` such that even if the promise hasn't been resolved, we will always know which promise is associated with a specific user. The tests were also updated accordingly to work with this new expected return type.
* We also added centralized avatar generation logic, removing the need for a custom implementation across various plugins.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/tinymce/src/core/main/ts/lookup/UserLookup.ts
+++ b/modules/tinymce/src/core/main/ts/lookup/UserLookup.ts
@@ -1,5 +1,5 @@
 import { StructureSchema, FieldSchema } from '@ephox/boulder';
-import { Arr, Optional, Results, Obj } from '@ephox/katamari';
+import { Arr, Optional, Results, Obj, Num } from '@ephox/katamari';
 
 import Editor from '../api/Editor';
 import * as Options from '../api/Options';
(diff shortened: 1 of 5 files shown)


modules/tinymce/src/core/main/ts/lookup/UserLookup.ts
```diff
@@ -134,23 +194,28 @@ const UserLookup = (editor: Editor): UserLookup => {
...
+  const finallyResolve = (userId: UserId, user: ExpectedUser) =>
     Optional
       .from(pendingResolvers.get(userId))
       .each(({ resolve }) => {
-        resolve(user);
+        resolve({
+          ...user,
+          id: user.id,
+          name: Optional.from(user.name).getOr(user.id),
+          avatar: Optional.from(user.avatar).getOr(deriveAvatar(user.id)),
```
spocke, line 205: Doing null checks with options is no the way. We use options to avoid null yes, but if your expression checks for null and returns non null then null is not spread though the whole code base it's local.

```suggestion
          name: user.name ?? user.id,
          avatar: user.avatar ?? deriveAvatar(user.id),
```

modules/tinymce/src/core/test/ts/browser/lookup/UserLookupTest.ts
```diff
@@ -11,18 +11,25 @@ import { createUserLookup, type User } from 'tinymce/core/lookup/UserLookup';
...
+    description: 'Test Description'
+  }
 });
 
 describe('browser.tinymce.core.UserLookupTest', () => {
   const hook = TinyHooks.bddSetupLight<Editor>({
     base_url: '/project/tinymce/js/tinymce',
     user_id: 'test-user-1',
-    fetch_users: (userIds: string[]): Promise<User[]> => {
+    fetch_users: (userIds: string[]): Promise<ExpectedUser[]> => {
```
spocke, line 32: If it used here as a type, wouldn't we type this in the public interface too with the same time and therefor not need the local type definition in this file.
Also I think we should be able to put that in the type for fetch_users in the options right so in OptionTypes? Then I think you might not need to annotate this function at all an the implementors of this API would be guided.

spocke (approved):

modules/tinymce/src/core/main/ts/lookup/UserLookup.ts
kimwoodfield: We don't even need to treat these as optionals or check for null values as we already handle assigning them fallbacks in `validateResponse` if they are not present so I think we can resolve this one.

modules/tinymce/src/core/main/ts/lookup/UserLookup.ts
```diff
@@ -54,17 +63,64 @@ export interface UserLookup {
...
    * @param {string[]} userIds - A list of user IDs to fetch information for.
-   * @return {Promise<User>[]} A promise that resolves to an array of users and information about them. Promises will reject if users are not found or if the fetch fails.
+   * @return {Record<UserId, Promise<User>>} An object where each key is a user ID and its value is a Promise that resolves to the user's data or rejects if the user is not found.
    * @throws {Error} When fetch_users option is not configured.
    */
-  fetchUsers: (userIds: UserId[]) => Promise<User>[];
+  fetchUsers: (userIds: UserId[]) => Record<UserId, Promise<User>>;
 }
 
+export const AvatarColors = [
```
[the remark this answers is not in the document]
MitchC1999: These may be better suited as a list of classes. Then we can set the colors for each of them in the skin. And I don't think we need to export it here.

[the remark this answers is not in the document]
kimwoodfield: @MitchC1999 Good call. I'll create a ticket for it after we flesh it out.

modules/tinymce/src/core/main/ts/lookup/UserLookup.ts
```diff
@@ -134,23 +192,28 @@ const UserLookup = (editor: Editor): UserLookup => {
...
     Optional
       .from(pendingResolvers.get(userId))
       .each(({ resolve }) => {
-        resolve(user);
+        resolve({
+          ...user,
+          id: user.id,
+          name: user.name,
+          avatar: user.avatar,
+        });
```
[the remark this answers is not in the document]
kimwoodfield: It was necessary when I swapped `User` out for `ExpectedUser` but that's correct, it is not needed now so I've removed.

modules/tinymce/src/core/test/ts/browser/lookup/UserLookupTest.ts
```diff
@@ -11,18 +11,25 @@ import { createUserLookup, type User } from 'tinymce/core/lookup/UserLookup';
 
 Chai.use(chaiAsPromised);
 
+interface ExpectedUser {
```
spocke, line 14: Do we still need this type since the other one is exposed?

spocke (approved):

TheSpyder (approved):

MitchC1999 (approved):

PR #10369, opened 2025-05-20
TINY-11927: Parse invalid content with full structure

spocke:
Related Ticket: TINY-11927

Description of Changes:
* Changes so context parsing is just validating as it was before 6
* Inserts the fragment and re-evalulates the the validness of the nodes when it's in place of the whole structure.
* Changed some tests to match the new or basically the old behaviour of 5.
* Added two changelog items since this is both fixing a bug and altering the behaviour of the parser slighlty.
* Added new tests for the customer cases.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of HTML fragments with invalid content, ensuring the editor preserves more of the original structure during insertion.
  - When parsing with a specific context, invalid nodes are now retained and the content is marked as invalid, rather than being automatically fixed.

- **New Features**
  - Enhanced detection and cleaning of invalid nodes during content insertion for more accurate content preservation.

- **Tests**
  - Expanded and reorganized test coverage for scenarios involving invalid HTML content insertion and parsing.

### DIFF
--- a/modules/tinymce/src/core/main/ts/html/ParserUtils.ts
+++ b/modules/tinymce/src/core/main/ts/html/ParserUtils.ts
@@ -52,11 +52,22 @@ const findClosestEditingHost = (scope: AstNode): Optional<AstNode> => {
   return Optional.from(editableNode);
 };
 
+const getAllDescendants = (scope: AstNode): AstNode[] => {
+  const collection: AstNode[] = [];
+
+  for (let node = scope.firstChild; node; node = node.walk()) {
+    collection.push(node);
+  }
+
+  return collection;
+};
+
 export {
   paddEmptyNode,
   isPaddedWithNbsp,
   hasOnlyChild,
   isEmpty,
   isLineBreakNode,
-  findClosestEditingHost
+  findClosestEditingHost,
+  getAllDescendants
 };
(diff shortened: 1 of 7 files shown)


modules/tinymce/src/core/main/ts/html/ParserUtils.ts
```diff
@@ -52,11 +52,22 @@ const findClosestEditingHost = (scope: AstNode): Optional<AstNode> => {
   return Optional.from(editableNode);
 };
 
+const getAllDescendants = (scope: AstNode): AstNode[] => {
+  const collection: AstNode[] = [];
+
+  for (let node = scope.firstChild; node; node = node.walk()) {
```
hamza0867, line 58: I think something like this would be more readable:

```suggestion
  for (let node = scope.firstChild; !Type.isNullable(node) ; node = node.walk()) {
```

hamza0867 (approved):

modules/tinymce/src/core/main/ts/html/ParserUtils.ts
spocke: Yes I agree I copied the loop from another old function that had that truthy check. Well spotted.

TheSpyder (approved):

PR #10372, opened 2025-05-21
TINY-12121: Move the list-content into core.

HAFRMO:
Related Ticket: TINY-12121

Description of Changes:
Moving the functional content of the list plugin into core as far as it is possible, leaving only the ui and api components in the plugin to treat it as normal.
Additionally some existing commands such as indent and delete which were previously listened and hijacked on by the list plugin now have explicit code tying the functionality in.

Pre-checks:
* [x] Changelog entry added

Review:
* [x Milestone set

## Summary by CodeRabbit

- **New Features**
  - List functionality has been fully integrated into the core editor, removing the need for a separate lists plugin.
  - Added a new editor option to control list indentation behavior when pressing the Tab key.

- **Improvements**
  - Core editor now manages list commands, keyboard interactions, and content filtering for consistent list behavior.
  - Enhanced indentation and outdentation logic specifically for list elements.
  - Lists are automatically normalized after delete operations to maintain structure.

- **Bug Fixes**
  - Improved detection and handling of non-editable and custom lists within the editor.

- **Tests**
  - Test suites updated and renamed to align with the core integration of list features.

- **Chores**
  - Internal modules and utilities reorganized and refactored to support the core integration of lists.
  - Simplified plugin initialization by removing conditional feature gating and redundant command registrations.


.changes/unreleased/tinymce-TINY-12121-2025-05-21.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Changed
+body: Moved functionality from lists-plugin into core.
```
TheSpyder, line 3:
```suggestion
body: Moved essential parts of the `lists` plugin into core.
```

modules/tinymce/src/core/main/ts/lists/actions/Delete.ts
```diff
@@ -1,21 +1,19 @@
...
-import * as ToggleList from '../actions/ToggleList';
-
-import * as Bookmark from './Bookmark';
-import * as NodeType from './NodeType';
-import * as NormalizeLists from './NormalizeLists';
-import * as ListRangeUtils from './RangeUtils';
-import * as Selection from './Selection';
-import { isWithinNonEditableList } from './Util';
+import { flattenListSelection, outdentListSelection } from './Indendation';
+import * as ToggleList from './ToggleList';
```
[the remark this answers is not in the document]
TheSpyder: looks like a very old typo in the filename from the lists plugin!

modules/tinymce/src/plugins/lists/main/ts/core/Bookmark.ts
```diff
@@ -12,17 +12,6 @@ interface Bookmark {
...
- * Returns a range bookmark. This will convert indexed bookmarks into temporary span elements with
- * index 0 so that they can be restored properly after the DOM has been modified. Text bookmarks will not have spans
- * added to them since they can be restored after a dom operation.
- *
- * So this: <p><b>|</b><b>|</b></p>
- * becomes: <p><b><span data-mce-type="bookmark">|</span></b><b data-mce-type="bookmark">|</span></b></p>
- *
- * @param  {DOMRange} rng DOM Range to get bookmark on.
- * @return {Object} Bookmark object.
- */
```
TheSpyder, line 25: Why was this deleted? It still seems useful 🤔

modules/tinymce/src/core/main/ts/api/Options.ts
```diff
@@ -1029,6 +1034,10 @@ const getApiKey = option('api_key');
 const isDisabled = option('disabled');
 const getExtendedMathmlAttributes = option('extended_mathml_attributes');
 const getExtendedMathmlElements = option('extended_mathml_elements');
+const shouldHaveListFeatures = (editor: Editor): boolean => {
+  return editor.hasPlugin('lists') || editor.hasPlugin('advlist');
+};
+const shouldIndentOnTab = (editor: Editor): boolean => option('lists_indent_on_tab')(editor) && shouldHaveListFeatures(editor);
```
TheSpyder, line 1040: It's a simple function, just inline it when this happens I think 😂 

```suggestion
const shouldIndentOnTab = (editor: Editor): boolean => editor.options.get('lists_indent_on_tab') && shouldHaveListFeatures(editor);
```

TheSpyder (approved):

modules/tinymce/src/plugins/lists/main/ts/core/Bookmark.ts
HAFRMO: Because moxidoc would fail. Though it might have been enough to remove the param-return and the rest can be kept.

modules/tinymce/src/core/main/ts/lists/lists/NodeType.ts
```diff
@@ -64,21 +62,21 @@ const isChildOfBody = (dom: DOMUtils, elm: Element): boolean =>
   dom.isChildOf(elm, dom.getRoot());
 
 export {
-  isTextNode,
+  isBlock,
```
spocke, line 65: There is duplication here between this module and the NodeType module. But we could log that as a cleanup ticket.

modules/tinymce/src/core/test/ts/browser/lists/ApplyDlTest.ts
```diff
@@ -3,9 +3,8 @@ import { LegacyUnit, TinyAssertions, TinyHooks } from '@ephox/wrap-mcagar';
 import { assert } from 'chai';
 
 import Editor from 'tinymce/core/api/Editor';
-import Plugin from 'tinymce/plugins/lists/Plugin';
 
-describe('browser.tinymce.plugins.lists.ApplyDlTest', () => {
+describe('browser.tinymce.core.lists.ApplyDlTest', () => {
   const hook = TinyHooks.bddSetupLight<Editor>({
     plugins: 'lists',
```
spocke, line 9: Since you are keeping this it will try to load the lists plugin from the base URL if it's not already loaded. Not sure what this does for tests though will it be weird. We might end up with a situation where it wouldn't load the webpack version of the plugin in tests when do you development.

spocke (approved):

modules/tinymce/src/core/main/ts/lists/Lists.ts
```diff
@@ -0,0 +1,21 @@
...
+import Editor from '../api/Editor';
+import * as Options from '../api/Options';
+
+import * as Commands from './lists/Commands';
+import * as Delete from './lists/Delete';
+import * as FilterContent from './lists/FilterContent';
+import * as Keyboard from './lists/Keyboard';
+
+const setup = (editor: Editor): void => {
+  if (Options.shouldHaveListFeatures(editor)) {
```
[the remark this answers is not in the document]
spocke: Thinking about this some more yesterday I think we should have these enabled by default.
Since the impact would only be some keyboard behavioural changes if you don't have lists enabled and that scenario is rare. If that some how prevents a user from upgrading then we can add some form of option later.
We already have some of the lists handling baked into core today like merging list items if you insert a list inside a list so this would just add more list support to the core.
It would make this so much cleaner then the core has basic list handling from an API and keyboard perspective and the lists plugin just adds the UI buttons. It's a clean cut of responsibility.

PR #10376, opened 2025-05-23
TINY-11974: Ensure that the `userId` is always available on the `UserLookup` API

kimwoodfield:

Description of Changes:
* `user_id` is an option that may or may not be included on the editor configuration but we should always assume that `userId` is available on the API. If the option has not been provided then we will auto-generate one instead using `anonymous` as the fallback value.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


MitchC1999 (approved):

modules/tinymce/src/core/main/ts/lookup/UserLookup.ts
```diff
@@ -261,10 +261,11 @@ const UserLookup = (editor: Editor): UserLookup => {
     }, {});
   };
 
-  const userId = Options.getUserId(editor);
+  const providedUserId = Options.getUserId(editor);
+  const fallbackUserId = 'anonymous';
```
MitchC1999, line 265: Could we make it `"Anonymous"` to align with the current revision history default user?

TheSpyder: yes

- [ ] rename

modules/tinymce/src/core/main/ts/lookup/UserLookup.ts
```diff
@@ -261,10 +261,11 @@ const UserLookup = (editor: Editor): UserLookup => {
...
 
   return Object.freeze({
-    userId,
+    userId: providedUserId || fallbackUserId,
```
TheSpyder, line 268: Make it a default in the `Options` registration, not here.

TheSpyder (changes requested):

tiny-ben-tran (approved):

modules/tinymce/src/core/test/ts/browser/lookup/UserLookupTest.ts
```diff
@@ -76,6 +84,15 @@ describe('browser.tinymce.core.UserLookupTest', () => {
...
       }).to.throw();
     });
+
+    it('TINY-11974: Should default to "anonymous" when no user_id is provided', () => {
+      const editor = hook.editor();
+      editor.options.unset('user_id');
+
+      editor.userLookup = createUserLookup(editor);
+
+      expect(editor.userLookup.userId).to.equal('Anonymous', 'Should default to anonymous');
```
spocke, line 94: I think all other tests we avoid the BDD asserts. Is it because chai promises assets only has this syntax or something else.
Personally I really dislike that pattern trying to do English in some weird DSL is never good. Brings me back to the last century when I had to program a strange language called Lingo that was all English. `set the thing of something to 4` that language later switched to proper syntax `something.thing = 4`. 😃 
I don't think we need to change this now but it's inconsistent with other tests.
They are all `assert.equal(editor.userLookup.userId, 'Anonymous', 'Should default to anonymous')` style.

spocke (approved):

TheSpyder (approved):
A type and a default, just how it should be 🙂

PR #10379, opened 2025-05-26
TINY-11937: Added ability to restrict indentation actions.

HAFRMO:
Related Ticket: TINY-11937

Description of Changes:
Prevent Tiny from indenting lists past a threshold if specified. This is only relevant to actions performed by tiny ( does not restrict existing lists that are further indented, or pasts ).

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

## Summary by CodeRabbit

- **New Features**
  - Added a new option to limit the maximum depth of nested lists using the `list_max_depth` setting.
  - The indent button and command now respect the configured maximum list depth, preventing further indentation when the limit is reached.

- **Bug Fixes**
  - Improved the dynamic enabling/disabling of the indent button based on list depth constraints.

- **Tests**
  - Introduced new tests to verify correct behavior of list indentation with various maximum depth settings.


spocke (approved):

modules/tinymce/src/core/main/ts/api/Options.ts
```diff
@@ -923,6 +923,14 @@ const register = (editor: Editor): void => {
     processor: 'boolean',
     default: true
   });
+
+  registerOption('list_max_depth', {
+    processor: (value) => {
+      const valid = typeof value === 'number';
+      return valid ? { value: value < 0 ? undefined : value + 1, valid } : { valid: false, message: 'Must be a number' };  // Internally we count a list as starting with one indent, so up the count to match this. If it's less than 0 just ignore it.
+    },
+    default: undefined
```
TheSpyder, line 932: `undefined` is the normal default

- [ ] remove

modules/tinymce/src/core/main/ts/api/Options.ts
```diff
@@ -1035,118 +1043,120 @@ const isDisabled = option('disabled');
...
 const getExtendedMathmlElements = option('extended_mathml_elements');
 const shouldIndentOnTab = option('lists_indent_on_tab');
+const getListMaxDepth = (editor: Editor): Optional<number> =>
+  Optional.from(editor.options.get('list_max_depth'));
 
 export {
-  register,
-
-  getIframeAttrs,
+  canFormatEmptyLines,
```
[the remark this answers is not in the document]
TheSpyder: - [ ] revert

modules/tinymce/src/core/main/ts/api/Options.ts
```diff
@@ -1035,118 +1043,120 @@ const isDisabled = option('disabled');
 const getExtendedMathmlAttributes = option('extended_mathml_attributes');
 const getExtendedMathmlElements = option('extended_mathml_elements');
 const shouldIndentOnTab = option('lists_indent_on_tab');
+const getListMaxDepth = (editor: Editor): Optional<number> =>
+  Optional.from(editor.options.get('list_max_depth'));
```
TheSpyder, line 1047: You might notice you had to import `Optional` in order to make this work. We don't use optionals in core settings.

Remove the optional, and just use `Type.isNumber` in the code that reads the option.

modules/tinymce/src/core/main/ts/api/Options.ts
```diff
@@ -923,6 +923,14 @@ const register = (editor: Editor): void => {
     processor: 'boolean',
     default: true
   });
+
+  registerOption('list_max_depth', {
+    processor: (value) => {
+      const valid = typeof value === 'number';
+      return valid ? { value: value < 0 ? undefined : value + 1, valid } : { valid: false, message: 'Must be a number' };  // Internally we count a list as starting with one indent, so up the count to match this. If it's less than 0 just ignore it.
```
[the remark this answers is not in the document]
TheSpyder: I think `< 0` should be an error, not a silent pass that does nothing.

And please don't `+1` here. Put that in the implementation, otherwise someone trying to do maintenance on this will need to track down why the number they set didn't match the number in the code.

modules/tinymce/src/core/main/ts/commands/IndentOutdent.ts
```diff
@@ -49,6 +52,18 @@ const canOutdent = (editor: Editor): boolean => {
   return !editor.mode.isReadOnly() && (blocks.length > 1 || validateBlocks(editor, blocks));
 };
 
+const canIndent = (editor: Editor): boolean => {
+  const max = Options.getListMaxDepth(editor);
+  return !editor.mode.isReadOnly() && max.map((max) => {
+    const blocks = getSelectedListRoots(editor);
+    if (blocks.length > 0) {
+      return parseLists(SugarElements.fromDom(blocks), getItemSelection(editor)).some((entry) => entry.entries.some((value) => value.depth < max));
```
TheSpyder, line 60: I was about to chide you for using the native `Array.some` method instead of `Arr.exists`

But surely you don't need to parse the entire list and iterate through every node just to pull out the depth. These `can` methods are checked after every node change; that could get quite expensive.

Think about what depth represents and how you might be able to quickly figure it out using the selected element (then if necessary, apply it to multiple selected list items, perhaps similar to how `canOutdent` does).

modules/tinymce/src/core/main/ts/api/Options.ts
spocke: Is there a particular reason why it's all internal? It it becomes cause we seldom add core options and until now not needed an option.

TheSpyder: Hard to say. The vast majority of these have their defaults encoded in this file, so they don't need to be options.

Other cases use things like `Type.isFunction()` to determine whether to perform an action instead of making it an optional value with `.each()`. But those would all have been written before we had option validation.

Maybe we side-step the question and make the default `-1` indicating no max depth 😂 

If we were going to start returning optional values I'd want to create a `wrapOptional` function as we do on some of the premium features.

modules/tinymce/src/core/main/ts/commands/IndentOutdent.ts
HAFRMO: I was uncertain of how far you can rely on a `SelectorFilter.ancestors().length` and not wanting desyncs between the actual capability and this one as it could lead to false positives and false negatives in the case of some funky setup. And at that point I didn't want to basically duplicate the code of the parseLists to just find the depth.
But we could do it the swift way and handle any edgecases if they pop up if you think that's acceptable?

modules/tinymce/src/core/main/ts/api/Options.ts
HAFRMO: I assumed we wanted to avoid not wrapping into Optionals? Not doing it just seems like I'd have to do it where we call it anyway, so why not do it here?

modules/tinymce/src/plugins/lists/test/ts/browser/MaxIndentTest.ts
```diff
@@ -0,0 +1,122 @@
...
+  const indentIsDisabled = () =>
+    UiFinder.exists(SugarBody.body(), '[aria-label="Increase indent"][aria-disabled="true"]');
+
+  const indentIsEnabled = () =>
+    UiFinder.exists(SugarBody.body(), '[aria-label="Increase indent"]:not([aria-disabled="true"])');
+
+  context('With a depth of two', () => {
+    const hook = TinyHooks.bddSetupLight<Editor>({
+      plugins: 'lists',
+      list_max_depth: 2,
```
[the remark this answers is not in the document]
HAFRMO: The code should allow it already, so sure.

modules/tinymce/src/core/main/ts/commands/IndentOutdent.ts
```diff
@@ -49,6 +50,9 @@ const canOutdent = (editor: Editor): boolean => {
   return !editor.mode.isReadOnly() && (blocks.length > 1 || validateBlocks(editor, blocks));
 };
 
+const canIndent = (editor: Editor): boolean =>
+  !editor.mode.isReadOnly() && ListIndentation.canIndent(editor);
```
TheSpyder, line 54: You wrote a `canIncreaseDepthOfList` function, but only used it for the application of indent? This check, which runs on every node change, still uses the `parseList` code you started with?

modules/tinymce/src/core/main/ts/lists/listmodel/Util.ts
```diff
@@ -17,7 +21,15 @@ const hasFirstChildList = (el: SugarElement<HTMLElement>): boolean =>
 const hasLastChildList = (el: SugarElement<HTMLElement>): boolean =>
   Traverse.lastChild(el).exists(isList);
 
+const canIncreaseDepthOfList = (editor: Editor, amount: number): boolean => {
+  return Options.getListMaxDepth(editor).fold(
+    Fun.always,
+    (max: number) => max > amount
+  );
```
TheSpyder:
```suggestion
  return Options.getListMaxDepth(editor).map((max: number) => max > amount).getOr(true);
```

modules/tinymce/src/core/main/ts/api/Options.ts
TheSpyder: I don't remember why.

modules/tinymce/src/core/main/ts/commands/IndentOutdent.ts
HAFRMO: Yes, that function does not do anything that would help with this situation.
And it did use parseList, yes, but that is no longer the case.

modules/tinymce/src/core/main/ts/lists/listmodel/ListsIndendation.ts
```diff
@@ -63,6 +64,15 @@ const listIndentation = (editor: Editor, lists: SugarElement<HTMLElement>[], ind
   });
 };
 
+const canIndent = (editor: Editor): boolean =>
+  Options.getListMaxDepth(editor).map((max) => {
```
spocke: This can be `forall`

modules/tinymce/src/core/main/ts/lists/listmodel/ListsIndendation.ts
```diff
@@ -63,6 +64,15 @@ const listIndentation = (editor: Editor, lists: SugarElement<HTMLElement>[], ind
...
+    const blocks = editor.selection.getSelectedBlocks();
+    return Arr.exists(blocks, (element) => {
+      return SelectorFind.closest(SugarElement.fromDom(element), 'li').map((sugarElement) => SelectorFilter.ancestors(sugarElement, 'ol,ul').length < max).getOr(true);
```
spocke: This can be forall.

modules/tinymce/src/core/main/ts/api/Options.ts
HAFRMO: If we don't have a reason not to have it I think it's better to keep it? I don't see a point to have all calls wrap it in an optional?

PR #10382, opened 2025-05-27
TINY-12133: setup oxide-components repo

hamza0867:
Related Ticket: TINY-12133

Description of Changes:
* Add `oxide-components` repo

Pre-checks:
* [x] Changelog entry added

Review:
* [x] ~Milestone set~

## Summary by CodeRabbit

- **New Features**
  - Introduced the `@tiny/oxide-components` React component library with a primary `Button` component, demo application, and Storybook setup for interactive documentation and testing.
  - Added Storybook stories for the `Button` component, including interactive and testable examples.
  - Implemented build, test, and CI scripts for the new module, integrated into the main project workflows.

- **Chores**
  - Added comprehensive configuration files for TypeScript, ESLint, Vite, Vitest, and project metadata to support development and quality assurance.
  - Updated root scripts to include the new module in CI and build processes.

- **Refactor**
  - Migrated ESLint configurations in existing modules to TypeScript-based config files for improved maintainability.

- **Style**
  - Added `.gitignore` and README files for better project organization and documentation.

- **Bug Fixes**
  - Removed deprecated or unnecessary ESLint plugins and rules to streamline linting.


hamza0867:
* **#10382** <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10382?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10382?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

modules/alloy/eslint.config.ts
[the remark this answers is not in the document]
hamza0867: Either I missed them or they were merged back in by mistake at some point. Like in this [PR](https://app.graphite.dev/github/pr/tinymce/tinymce-premium/716/TINY-12134-Use-Button-component-from-oxide-components-in-tinycomments-plugin) for example. I know those `.eslintrc` files were merged back in at some point because I remember deleting them.

Jenkinsfile
```diff
@@ -182,41 +182,77 @@ def cacheName = "cache_${BUILD_TAG}"
...
-    stage('Deps') {
-      // cancel build if primary branch doesn't merge cleanly
-      gitMerge(primaryBranch)
-      yarnInstall()
+    container('playwright'){
+      // Run any playwright test in the Playwright container
+      stage('Playwright') {
+        exec('yarn -s --cwd modules/oxide-components test-manual')
+      }
     }
```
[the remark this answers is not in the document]
spocke: Ok nice. But then why is it a 2 minute run?
Haven't had time to look into this. I don't want this part to slow down all builds if we can avoid it.

modules/oxide-components/package.json
```diff
@@ -0,0 +1,76 @@
...
+  "scripts": {
+    "dev": "vite",
+    "build": "tsc -b && vite build",
+    "lint": "eslint --max-warnings=0 src/**/*.tsx",
+    "preview": "vite preview",
+    "storybook": "storybook dev -p 6006",
+    "build-storybook": "storybook build",
+    "test-watch": "vitest",
+    "test": "echo 'No tests'",
+    "test-manual": "playwright install && playwright install-deps && vitest --watch=false --reporter=junit --outputFile=test-results.xml",
```
TheSpyder, line 39: Please use a scratch folder or something else that will be git ignored when used locally.

And why is it called `test-manual` when it's used for what seems to be mostly automated CI purposes?

yarn.lock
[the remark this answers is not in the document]
hamza0867: Yes it is unfortunately. I had to blow up the lock file and generate a new one. Otherwise yarn was getting confused and was installing dependencies in the wrong place and sometimes duplicate dependencies 🤷‍♂️

Jenkinsfile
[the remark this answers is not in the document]
hamza0867: It seems we didn't need to install those system deps anymore since we are using the playwright docker image that has all of those system deps already installed. So no we went from 16 seconds to 7. 👍

modules/oxide-components/package.json
hamza0867: the `test-resluts.xml` is ignored by git. Does it have to be in the scratch folder or does it suffice to be ignored by git ?

I just changed `test-manual` to `test-ci`

In this library I think we need vitest as it has good integration with storybook for testing our react components

modules/oxide-components/src/main.tsx
spocke, line 1: Could we structure this a bit more like the other libraries we have. So this would be in `modules/oxide-components/src/main/ts/Main.ts`. We don't need the `ephox/<library>` part that is legacy.

modules/oxide-components/demo/App.tsx
spocke, line 1: Could we move this to `modules/oxide-components/src/demo/ts/` for consistency with other libraries.

modules/oxide-components/index.html
spocke, line 1: Can we serve this from `modules/oxide-components/src/demo/html/demo.html` or `modules/oxide-components/src/demo/html/index.html` to be more consistent with other libraries.

modules/oxide-components/src/Button/Button.stories.tsx
spocke, line 1: Will this increase to a huge file. Can it be separated into a directory like `modules/oxide-components/src/main/ts/Button/stores/Button.stories.tsx or should we do that later as it grows. We need to have some form of pattern that we follow when we grow the stories.

modules/oxide-components/src/main.tsx
```diff
@@ -0,0 +1,6 @@
+import { Button } from './Button/Button.component';
+
+export {
+    Button
```
spocke: Bit weird indentation here.

modules/oxide-components/README.md
```diff
@@ -0,0 +1,10 @@
+# Description
+
+`oxide-components` is a collection of various react components that would be used in the TinyMCE editor.
+# Installation
```
spocke, line 5: Picky picky but maybe a newline before this title.

.changie.yaml
spocke, line 1: The reformat here is probably fine. But might be worth disabling auto reformatting on yaml since if everyone has different formatting rules here it becomes the battle of formatters everytime someone saves the file.

modules/oxide-components/package.json
```diff
@@ -0,0 +1,76 @@
...
+  ],
+  "main": "lib/main.js",
+  "module": "lib/main.js",
+  "types": "lib/main.d.ts",
+  "scripts": {
+    "dev": "vite",
+    "build": "tsc -b && vite build",
+    "lint": "eslint --max-warnings=0 src/**/*.tsx",
+    "preview": "vite preview",
+    "storybook": "storybook dev -p 6006",
```
spocke: Could we alias this to `start` since I think most other projects we have `pnpm start` and I think the storybook is likely the way we would develop these components right?

modules/oxide-components/package.json
```diff
@@ -0,0 +1,76 @@
...
+    "LICENSE.txt"
+  ],
+  "main": "lib/main.js",
+  "module": "lib/main.js",
+  "types": "lib/main.d.ts",
+  "scripts": {
+    "dev": "vite",
+    "build": "tsc -b && vite build",
+    "lint": "eslint --max-warnings=0 src/**/*.tsx",
+    "preview": "vite preview",
```
spocke: What does this do, if I run it is tarts a localhost sever but if I navigate to it's just a 404.

modules/oxide-components/package.json
```diff
@@ -0,0 +1,76 @@
...
+    "CHANGELOG.txt",
+    "LICENSE.txt"
+  ],
+  "main": "lib/main.js",
+  "module": "lib/main.js",
+  "types": "lib/main.d.ts",
+  "scripts": {
+    "dev": "vite",
+    "build": "tsc -b && vite build",
+    "lint": "eslint --max-warnings=0 src/**/*.tsx",
```
spocke: Not all files will be `tsx` if we have utility code they would be `ts` and we would easily miss that later on.

modules/oxide-components/package.json
```diff
@@ -0,0 +1,76 @@
...
+    "README.md",
+    "LEGAL.txt",
+    "CHANGELOG.txt",
+    "LICENSE.txt"
+  ],
+  "main": "lib/main.js",
+  "module": "lib/main.js",
+  "types": "lib/main.d.ts",
+  "scripts": {
+    "dev": "vite",
```
spocke, line 31: I wonder if we even need the demo and the dev build maybe it should all go though storybook. Otherwise we would have to add demos separately from story book for every UI component that seems like a lot of duplicated work. Or just keep this single demo with one button that we more or less would never use.

modules/oxide-components/src/Button/Button.stories.tsx
```diff
@@ -0,0 +1,52 @@
...
+    type: {
+      control: 'select',
+      options: [ 'button', 'submit', 'reset' ]
+    }
+  },
+  // Use `fn` to spy on the onClick arg, which will appear in the actions panel once invoked: https://storybook.js.org/docs/essentials/actions#action-args
+  args: { onClick: fn() },
+} satisfies Meta<typeof Button>;
+
+export default meta;
```
spocke: Is this a required pattern of story book to have a mixed default exports. Usually `export default` is an anti pattern something that was used for compatibility with older module systems prior to ESM.

modules/oxide-components/src/Button/Button.stories.tsx
```diff
@@ -0,0 +1,52 @@
...
+export const ButtonWithCounter: Story = {
+  render: () => {
+    const [ count, setCount ] = useState(0);
+    return (
+      <Button onClick={() => setCount(count + 1)}>
+        Clicked {count} times
+      </Button>
+    );
+  },
+  play: async ({ canvasElement }) => {
```
spocke: How will this scale with more tests? Is the idea to pump out more stories and one `play` test for each scenario.

modules/oxide-components/package.json
```diff
@@ -0,0 +1,76 @@
...
+    "build": "tsc -b && vite build",
+    "lint": "eslint --max-warnings=0 src/**/*.tsx",
+    "preview": "vite preview",
+    "storybook": "storybook dev -p 6006",
+    "build-storybook": "storybook build",
+    "test-watch": "vitest",
+    "test": "echo 'No tests'",
+    "test-ci": "vitest --watch=false --reporter=junit --outputFile=test-results.xml",
```
spocke: There isn't any tests yet here to run could we make a simple vitest so that we can see how that would work. Where that would be placed etc? I kind of want this setup to have most part a button, storybook for that button, tests for that button etc so that we can start to see what the development flow would be. If you think that is to big for this PR we can make a follow up task.

modules/oxide-components/package.json
```diff
@@ -0,0 +1,76 @@
...
+    "test-storybook": "vitest --project storybook",
```
spocke, line 39: Running this I get errors and a warning about `project` being deprecated.

spocke (changes requested):
Moving this to request changes since there seems to be some script error and warning and I think we can improve the structure a bit before we merge.

modules/oxide-components/package.json
TheSpyder: I think it should be `*.ts?` or `*.{ts,tsx}` which our other JSX projects use.

modules/oxide-components/package.json
TheSpyder: I'd still prefer to use a scratch folder for generated reports and logs to match how our other projects operate.

yarn.lock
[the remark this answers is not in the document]
TheSpyder: You can't do that on a whim. We might as well not have a lockfile. Doing this on what is planned to be a long-running epic branch also means you'll have ongoing conflicts.

Maybe using `yarn upgrade-interactive --latest` would help. That's something you could do in a PR to `main` and then the epic branch changes would be a lot smaller.

modules/alloy/eslint.config.ts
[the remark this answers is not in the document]
TheSpyder: hmm looks like you only converted the top level, not the individual projects
https://app.graphite.dev/github/pr/tinymce/tinymce/10295/TINY-11629-Upgrade-alloy-to-use-eslint-V9

I don't understand how the build (or linting in general) has worked at all for the last 3 weeks. Please fix build issues like this in a separate PR to `main` so they aren't locked away on your epic branch.

TheSpyder (changes requested):

yarn.lock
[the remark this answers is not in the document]
spocke: Maybe we can script these things it usually happens when you add package.json deps without using the yarn add commands.
In long living branches that have dependencies the lockfile will conflict a lot.
A script tool could parse the package json grab the dependencies generate calls to `yarn add` empty the dependencies then add them one by one.
That way you could always just resolve the conflict by using theirs then run the script to re-apply the dependencies.
Duplicate dependencies could maybe be due to missing calls to `npx yarn-deduplicate` I think the older yarn was not smart enough to remove duplicates.

modules/oxide-components/package.json
hamza0867: The demo page exists mainly for developers to use it when doing development so that they can test the components in the setup that makes most sense at that time. But the real demos would be the ones in the storybook. What do you think ?

modules/oxide-components/package.json
hamza0867: There is one at `Button.stories.tsx` and you can see the tests running  in the CI as well.

modules/oxide-components/package.json
hamza0867: That's weird. It's working fine for me locally, and even according to the [docs](https://vitest.dev/guide/cli.html#project) it should be fine

modules/oxide-components/src/Button/Button.stories.tsx
hamza0867: Yes this is required by storybook

modules/oxide-components/src/Button/Button.stories.tsx
hamza0867: That's the idea for now.

modules/oxide-components/src/Button/Button.stories.tsx
hamza0867: If it comes to it we can then split the stories to have multiple `stories.tsx` files for the same component. But for now there is no need to. Because it maybe possible that we won't need huge files given the scope of each component and how it is tested in isolation.

.changie.yaml
hamza0867: Indeed. That being said, I would say this formatting is the more correct one as we have proper indentation for the `kinds` and `custom` properties

modules/oxide-components/package.json
hamza0867: Done!

modules/oxide-components/package.json
hamza0867: Done!

modules/oxide-components/package.json
hamza0867: Indeed, that command exists to preview the production build for SPAs, not libraries.

modules/oxide-components/package.json
hamza0867: Done!

modules/oxide-components/src/main.tsx
hamza0867: Done!

modules/oxide-components/src/main.tsx
hamza0867: Fixed!

modules/oxide-components/demo/App.tsx
hamza0867: Done!

modules/oxide-components/index.html
hamza0867: Done!

modules/oxide-components/package.json
spocke: I noticed this when I accidentally didn't use the right `tinymce-premium` branch so I assume it installed a newer version that throws that error. So it might be an issue when we upgrade.

modules/oxide-components/package.json
```diff
@@ -0,0 +1,74 @@
...
+    "dev": "vite",
+    "build": "tsc -b && vite build",
+    "lint": "eslint --max-warnings=0 src/**/*.ts?",
+    "start": "storybook dev -p 6006",
+    "build-storybook": "storybook build",
+    "test-watch": "vitest",
+    "test": "echo 'No tests'",
+    "test-ci": "vitest --watch=false --reporter=junit --outputFile=scratch/test-results.xml",
+    "test-storybook": "vitest --project storybook",
+    "ci": "yarn lint && yarn test"
```
spocke: This needs to build it all. Both the library and the storybook otherwise we will not get a failure if these parts fail to build as it did locally.

- [x] Build everything on CI

hamza0867: What do you mean by "build storybook" ? The storybook as of now is used only locally for development, Vitest compiles the tests that exist in storybook and runs them. Is this what you mean ? As for the library I added the build command 👍

modules/oxide-components/package.json
hamza0867: I fixed this warning. The warning was thrown in the [CI](https://jenkins2.tiny.work/job/tinymce/job/feature%252FTINY-12133/35/pipeline-overview/log?nodeId=383) as well. But it's fixed now

modules/oxide-components/package.json
spocke: That the CI does `storybook build` so that we know that the storybook is not broken. But if `test-storybook` does the same thing in the playwright container then we might not need it.

hamza0867: Ah good point. Done!

modules/oxide-components/package.json
```diff
@@ -0,0 +1,81 @@
...
+    "@vitest/browser": "^3.1.4",
+    "@vitest/coverage-v8": "^3.1.4",
+    "eslint-plugin-react-hooks": "^5.2.0",
+    "eslint-plugin-react-refresh": "^0.4.19",
+    "eslint-plugin-storybook": "^9.0.10",
+    "globals": "^16.0.0",
+    "playwright": "^1.52.0",
+    "postcss": "^8.4.39",
+    "react": "^18.3.1",
+    "react-dom": "^18.3.1",
```
TheSpyder, line 68: Apologies for not noticing this earlier, but aren't we using preact?

TheSpyder (approved):

modules/oxide-components/package.json
hamza0867: No, we are using react here because react has a bigger ecosystem and integration with storybook. For example, the automatic doc generation exists when developing in react but not preact. That being said, in the plugins consuming the library, they use preact. So in the final bundle we will have only preact. Does it make sense ?

PR #10386, opened 2025-05-28
TINY-12163: fix conflict between `mouseover` and arrow navigation in menubar

lorenzo-pomili:
Related Ticket: TINY-12163

Description of Changes:
**THE BUG:**
the navigation stopped because both on `mouseover` and on `focusShifted` the focus is moved on the target (the already focussed one in case of `focusShifted` on the element with the mouse over on `mouseover`), for some reason moving the focus to another button sometime trigger the `mouseover` again, it seems releated to [this mixin](https://github.com/tinymce/tinymce/blob/558aa03340563c8e85c7c8f1d67bd9969fc847d4/modules/oxide/src/less/theme/components/menubar/menubar-button.less#L79).

**REASON OF THIS SOLUTION:**
I didn't go in deep to understand it since I preferred to avoid that triggering both could cause the bug, this because if a style can cause it and I don't think we have a good way to test it just fixing what is triggering the issue would have a high risk of regression.

**SOLUTION:**
to avoid it I implemented a new Throttler (`Throttler.withPriority`), this function works like `Throttler.first` but it is possible to indicate that a call has a priority on the others

Pre-checks:
* [ ] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/katamari/src/main/ts/ephox/katamari/api/Throttler.ts
+++ b/modules/katamari/src/main/ts/ephox/katamari/api/Throttler.ts
@@ -5,6 +5,43 @@ export interface Throttler<A extends any[]> {
   readonly throttle: (...args: A) => void;
 }
 
+export interface ThrottlerWithPriority<A extends any[]> {
+  readonly cancel: () => void;
+  readonly throttle: (hasPriority: boolean, ...args: A) => void;
+}
+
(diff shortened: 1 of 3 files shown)


lorenzo-pomili:
closed since @spocke suggested a way better solution that have nothing to do with what I did

PR #10388, opened 2025-05-29
TINY-12056: update dompurify version and remove patch

michalnieruchalski-tiugo:
Related Ticket: 

Description of Changes:
* Placeholder text

Pre-checks:
* [ ] Changelog entry added

Review:
* [ ] Milestone set


michalnieruchalski-tiugo:
* **#10388** <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10388?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10388?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

modules/tinymce/src/core/main/ts/html/Sanitization.ts
```diff
@@ -26,6 +26,36 @@ const processNode = (node: Node, settings: DomParserSettings, schema: Schema, sc
...
+    settings.sanitize &&
+    NodeType.isElement(node) &&
+    node.hasChildNodes() &&
+    !isNode(node.firstElementChild) &&
+    /<[/\w!]/g.test(node.innerHTML) &&
+    /<[/\w!]/g.test(node.textContent ?? '')
+  ) {
+    Remove.remove(SugarElement.fromDom(node));
+    return;
+  }
```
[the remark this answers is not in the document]
michalnieruchalski-tiugo: The regex is defined twice, the way I see it is that there are two separate regexp instances. Therefore they won't share state.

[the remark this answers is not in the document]
michalnieruchalski-tiugo:
```
 /<[/\w!]/.test(node.innerHTML) &&
 /<[/\w!]/.test(node.textContent ?? '')
```

I can also use `new RegExp()` syntax. Would this code run any different if this syntax was used?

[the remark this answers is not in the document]
michalnieruchalski-tiugo: > Instance creation: new RegExp() creates a new regex object every time the line executes, whereas literals are typically cached by the engine

So multiple run's of `processNode` method will use the same regexp object instance? Therefore there can be an issue with shared state.

modules/tinymce/src/core/main/ts/html/Sanitization.ts
```diff
@@ -21,11 +21,44 @@ interface Sanitizer {
...
+    as `node.textContent` will ignore Comment and CDATA nodes.
+
+    The regular expression used will match any opening or closing HTML tag, but also comments.
+  */
+  const isNode = (value: unknown): value is Node => typeof Node === 'function' && value instanceof Node;
+  return (
+    NodeType.isElement(node) &&
+    node.hasChildNodes() &&
+    !isNode(node.firstElementChild) &&
+    /<[/\w!]/.test(node.innerHTML) &&
```
michalnieruchalski-tiugo: https://github.com/cure53/DOMPurify/blob/main/src/purify.ts#L1047

I removed the g flag from the regexp as I it's not necessary in this context. And it makes regexp stateful so nasty bugs may occur in the future.

TheSpyder: is this something we should consider logging with dompurify? There doen't seem to be a particularly specific reason for the `SAFE_FOR_XML` check to be added there.

michalnieruchalski-tiugo: Looking at the release notes of 3.2.6 I think the reason is:
- Added better configuration for aggressive mXSS scrubbing behavior

https://github.com/cure53/DOMPurify/releases/tag/3.2.6

In general, hard to tell why it was added. DOMPurify don't link their changes to particular XSS attacks, so hard to tell :/

spocke (approved):
This needs a changelog to describe the slight behaviour change for scripts. So scripts, iframes, video elements with HTML like data in them will now be emptied unless you set `xss_sanitization` to `false`.

modules/tinymce/src/core/main/ts/html/Sanitization.ts
TheSpyder: The commit where it was added isn't clear either. I would say it's worth logging an issue, just to query it and explain why we would prefer to have the check even without safe for xml.

modules/tinymce/src/core/text/tinymce-license-headers.js
```diff
@@ -1,5 +1,5 @@
 /**
  * This file bundles the code of
- * DOMPurify 3.2.4 | (c) Cure53 and other contributors | Released under the Apache license 2.0 and Mozilla Public License 2.0 | github.com/cure53/DOMPurify/blob/3.2.4/LICENSE
+ * DOMPurify 3.2.6 | (c) Cure53 and other contributors | Released under the Apache license 2.0 and Mozilla Public License 2.0 | github.com/cure53/DOMPurify/blob/3.2.6/LICENSE
  * The code included in this file has been modified.
```
spocke: Is that the case anymore?

.changes/unreleased/tinymce-TINY-12056-2025-06-16.yaml
```diff
@@ -0,0 +1,7 @@
+project: tinymce
+kind: Changed
+body: Comments with HTML like content, attributes with HTML
+  like values, elements with HTML Comment like textContent are no longer retained.
```
TheSpyder:
```suggestion
body: Comments with HTML like content, attributes with HTML
  like values, elements with HTML Comment like textContent are no longer retained while content sanitization is active.
```

modules/tinymce/src/core/text/tinymce-license-headers.js
TheSpyder: We should update this in the TinyMCE 8 IP asset register as well.
- [x] remove line

TheSpyder (approved):

PR #10389, opened 2025-05-29
TINY-12120: Disable spellcheck in text areas.

HAFRMO:
Related Ticket: TINY-12120

Description of Changes:
Mirror the spellcheck-attribute on the code sample text area.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed an issue where spellcheck was not properly disabled in the Codesample dialog. Spellcheck is now correctly disabled in code input areas.

- **New Features**
  - Added support for enabling or disabling spellcheck on textarea inputs across dialogs.

- **Tests**
  - Added tests to verify the correct behavior of the spellcheck attribute in code dialogs and textarea components.

### DIFF
--- a/modules/tinymce/src/plugins/codesample/main/ts/ui/Dialog.ts
+++ b/modules/tinymce/src/plugins/codesample/main/ts/ui/Dialog.ts
@@ -2,6 +2,7 @@ import { Arr, Fun } from '@ephox/katamari';
 
 import Editor from 'tinymce/core/api/Editor';
 
+import * as Options from '../api/Options';
 import * as CodeSample from '../core/CodeSample';
 import * as Languages from '../core/Languages';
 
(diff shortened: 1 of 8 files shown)


TheSpyder (approved):

modules/tinymce/src/plugins/codesample/main/ts/ui/Dialog.ts
```diff
@@ -28,7 +29,8 @@ const open = (editor: Editor): void => {
         {
           type: 'textarea',
           name: 'code',
-          label: 'Code view'
+          label: 'Code view',
+          spellcheck: Options.shouldBrowserSpellcheck(editor),
```
spocke: I think we should hardcode this to `false` since spellchecking source code is never going to work well and the `browser_spellcheck` option also toggles spellchecking inside the rich text editor and that might be problematic to enable while also having say our premium spellchecker enabled.
I think also we should disable this for the source code editor dialog since that is also mixed source code and plain text. If we get pushback on this we can re-introduce a `sourcecode_spellcheck` option or something that both these dialogs can use.

spocke (approved):

PR #10394, opened 2025-06-03
TINY-12187: add iframe_aria_required option

zuzanna-antas:
Related Ticket: TINY-12187

Description of Changes:
* Added new `iframe_aria_required` option that sets `aria-required` attribute on iframe element
* Implemented functionality to set the attribute directly on the Editor's element

Pre-checks:
* [x] Changelog entry added

Review:
* [ ] Milestone set

## Summary by CodeRabbit

- **New Features**
  - Introduced a new configuration option to set the aria-required attribute on the editor's iframe, enhancing accessibility.
  - Developers can now specify whether the iframe or the editor element is required for ARIA compliance via configuration or direct attribute setting.

### DIFF
--- a/.changes/unreleased/tinymce-TINY-12187-2025-06-03.yaml
+++ b/.changes/unreleased/tinymce-TINY-12187-2025-06-03.yaml
@@ -1,0 +1,7 @@
+project: tinymce
+kind: Added
+body: New `iframe_aria_required` option that sets `aria-required` attribute on iframe element. It's
+  also possible to set the attribute directly on the Ediotr's element.
+time: 2025-06-03T16:54:26.100388+02:00
+custom:
+  Issue: TINY-12187
--- a/modules/tinymce/src/core/main/ts/api/OptionTypes.ts
+++ b/modules/tinymce/src/core/main/ts/api/OptionTypes.ts
@@ -135,6 +135,7 @@ interface BaseEditorOptions {
   icons_url?: string;
   id?: string;
   iframe_aria_text?: string;
+  iframe_aria_required?: boolean;
   iframe_attrs?: Record<string, string>;
   images_file_types?: string;
   images_replace_blob_uris?: boolean;
--- a/modules/tinymce/src/core/main/ts/api/Options.ts
+++ b/modules/tinymce/src/core/main/ts/api/Options.ts
@@ -532,6 +532,11 @@ const register = (editor: Editor): void => {
     default: 'Rich Text Area'.concat(editor.hasPlugin('help') ? '. Press ALT-0 for help.' : '')
   });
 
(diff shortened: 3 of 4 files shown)


zuzanna-antas:
* **#10394** <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10394?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10394?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

modules/tinymce/src/core/main/ts/api/OptionTypes.ts
```diff
@@ -135,6 +135,7 @@ interface BaseEditorOptions {
   icons_url?: string;
   id?: string;
   iframe_aria_text?: string;
+  iframe_aria_required?: boolean;
   iframe_attrs?: Record<string, string>;
```
TheSpyder, line 139: since we already have `iframe_attrs` this seems a bit pointless.

If `aria-required` is important to have for the editor, we should make it a generic `aria_required` setting and apply it to inline mode as well.

TheSpyder (changes requested):

modules/tinymce/src/core/main/ts/api/OptionTypes.ts
spocke: The `iframe_aria_text` is special it applies the attribute to the iframe body element on Firefox but on the iframe element on other browsers. Do we need to do the same for this attribute if so I think it needs to be a special option.
The `iframe_attrs` only apply to the iframe.
The `aria-required` seems to be only valid for roles that fake form elements and the closest one is `textbox` so do we need to set that one as well. Found examples on mdm and other places that it can be used with aria-multi-line.
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/textbox_role
So this can be applied to divs so it should be some form of aria_required option.
We need to do a ton of testing on this see how screen readers react.

.changes/unreleased/tinymce-TINY-12187-2025-06-03.yaml
```diff
@@ -0,0 +1,7 @@
+project: tinymce
+kind: Added
+body: New `iframe_aria_required` option that sets `aria-required` attribute on iframe element. It's
+  also possible to set the attribute directly on the Ediotr's element.
```
spocke, line 4: I think the details of this can be added to the release notes what the default value is with examples etc.

```suggestion
body: New `iframe_aria_required` option that sets `aria-required` attribute on iframe element.
```

spocke (approved):

PR #10405, opened 2025-06-12
TINY-12213: Introduce minor CSS tweaks

kimwoodfield:
Related Ticket: [TINY-12213](https://ephocks.atlassian.net/browse/TINY-12213)

Description of Changes:
* Modified `@suggestededits-dark-transparent-color` so that it results in a value of `#d3d5d8`
* The `.tox-view__pane` class has a padding of `8px` but should be `0px` so I added a new selector `.tox-suggestededits__pane` to get around it, just like we did [here](https://github.com/tinymce/tinymce/blob/main/modules/oxide/src/less/theme/components/revisionhistory/revisionhistory.less#L34). Follow-up ticket also created to address the root issue.
* The `.tox-suggestededits__card--single` has a padding of `8px`, it has now been changed to `6px`.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

[TINY-12213]: https://ephocks.atlassian.net/browse/TINY-12213?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

- **Style**
  - Updated the dark transparent color for Suggested Edits to improve visual blending.
  - Removed padding from the Suggested Edits pane for better alignment.
  - Reduced padding within single Suggested Edits cards for a more compact appearance.
  - Adjusted padding behavior in the Revision History pane for improved layout consistency.

### DIFF
--- a/modules/oxide/src/less/theme/components/suggestededits/suggestededits.less
+++ b/modules/oxide/src/less/theme/components/suggestededits/suggestededits.less
@@ -8,7 +8,7 @@
 @suggestededits-color-tint: @color-tint;
 @suggestededits-text-color: @text-color;
 
-@suggestededits-dark-transparent-color: transparentize(@color-black, 0.1);
+@suggestededits-dark-transparent-color: mix(@color-black, @background-color, 20%);
 
 @suggestededits-background-color: @background-color;
 @suggestededits-border-color: @border-color;
--- a/modules/oxide/src/less/theme/components/suggestededits/suggestededits.less
+++ b/modules/oxide/src/less/theme/components/suggestededits/suggestededits.less
@@ -29,6 +29,10 @@
 @suggestededits-card-selected-border-color: @suggestededits-color-tint;
 
 .tox {
+  .tox-suggestededits__pane {
+    padding: 0 !important;      /* Override the default padding of tox-view__pane */
+  }
+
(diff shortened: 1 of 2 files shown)


modules/oxide/src/less/theme/components/suggestededits/suggestededits.less
```diff
@@ -29,6 +29,10 @@
 @suggestededits-card-selected-border-color: @suggestededits-color-tint;
 
 .tox {
+  .tox-suggestededits__pane {
+    padding: 0 !important;      /* Override the default padding of tox-view__pane */
```
spocke: Using `!important` is an anti pattern and should be avoided. Use more stricter specificity instead or make sure that the classes are in the right order so `tox-view__pane tox-suggestededits__pane` or something.

kimwoodfield: It's identical to the way it was done for revision history but I agree that it's not ideal. I'll take another look to see if we can improve the specificity without having to rely on `!important`.

MitchC1999 (approved):

PR #10411, opened 2025-06-16
TINY-12228: Add new crossorigin option

spocke:
Related Ticket: TINY-12228

Description of Changes:
* Adds new `crossorigin` option

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

## Summary by CodeRabbit

- **New Features**
  - Added a configuration option to set the cross-origin attribute on scripts loaded by the editor, enhancing control over external script fetching.
- **Tests**
  - Added tests to ensure proper behavior of the cross-origin attribute on dynamically loaded script elements.

### DIFF
--- a/modules/tinymce/src/core/main/ts/api/Editor.ts
+++ b/modules/tinymce/src/core/main/ts/api/Editor.ts
@@ -313,10 +313,7 @@ class Editor implements EditorObservable {
       DOMUtils.DOM.styleSheetLoader._setReferrerPolicy(referrerPolicy);
     }
 
-    const crossOrigin = Options.getCrossOrigin(self);
-    if (Type.isNonNullable(crossOrigin)) {
-      ScriptLoader.ScriptLoader._setCrossOrigin(crossOrigin);
-    }
+    ScriptLoader.ScriptLoader._setCrossOrigin(Options.getCrossOrigin(self));
 
     const contentCssCors = Options.hasContentCssCors(self);
     if (Type.isNonNullable(contentCssCors)) {
--- a/modules/tinymce/src/core/main/ts/api/Editor.ts
+++ b/modules/tinymce/src/core/main/ts/api/Editor.ts
@@ -313,6 +313,11 @@ class Editor implements EditorObservable {
       DOMUtils.DOM.styleSheetLoader._setReferrerPolicy(referrerPolicy);
     }
 
+    const crossOrigin = Options.getCrossOrigin(self);
+    if (Type.isNonNullable(crossOrigin)) {
+      ScriptLoader.ScriptLoader._setCrossOrigin(crossOrigin);
+    }
+
     const contentCssCors = Options.hasContentCssCors(self);
     if (Type.isNonNullable(contentCssCors)) {
       DOMUtils.DOM.styleSheetLoader._setContentCssCors(contentCssCors);
--- a/modules/tinymce/src/core/main/ts/api/Options.ts
+++ b/modules/tinymce/src/core/main/ts/api/Options.ts
@@ -81,6 +81,11 @@ const register = (editor: Editor): void => {
     default: ''
   });
 
+  registerOption('crossorigin', {
+    processor: 'string',
+    default: ''
+  });
+
   registerOption('language_load', {
     processor: 'boolean',
     default: true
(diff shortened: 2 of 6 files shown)


modules/tinymce/src/core/test/ts/browser/dom/CrossOriginTest.ts
```diff
@@ -0,0 +1,90 @@
+import { after, context, describe, it } from '@ephox/bedrock-client';
+import { Arr } from '@ephox/katamari';
+import { Attribute, SugarElement, SugarNode } from '@ephox/sugar';
+import { McEditor } from '@ephox/wrap-mcagar';
+import { assert } from 'chai';
+
+import ScriptLoader from 'tinymce/core/api/dom/ScriptLoader';
+import Editor from 'tinymce/core/api/Editor';
+
```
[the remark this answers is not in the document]
spocke: No, ScriptLoader is a class to it needs to be a instance. The ScriptLoader.ScriptLoader is a static instance of that class.

modules/tinymce/src/core/main/ts/api/Editor.ts
```diff
@@ -313,6 +313,11 @@ class Editor implements EditorObservable {
       DOMUtils.DOM.styleSheetLoader._setReferrerPolicy(referrerPolicy);
     }
 
+    const crossOrigin = Options.getCrossOrigin(self);
+    if (Type.isNonNullable(crossOrigin)) {
+      ScriptLoader.ScriptLoader._setCrossOrigin(crossOrigin);
+    }
+
```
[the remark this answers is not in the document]
spocke: This is a known issue, but fixing it would require us to change all the script loading calls to be instance based and bound to an editor instance that is a breaking change and might not really be worth the effort.

modules/tinymce/src/core/main/ts/api/Options.ts
```diff
@@ -81,6 +81,11 @@ const register = (editor: Editor): void => {
     default: ''
   });
 
+  registerOption('crossorigin', {
+    processor: 'string',
+    default: ''
+  });
+
```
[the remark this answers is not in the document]
spocke: That would mean that we would need to add to the list when browsers adds new possible values. Not sure that the validation is worth the effort.

lorenzo-pomili (approved):

modules/tinymce/src/core/main/ts/api/Editor.ts
```diff
@@ -313,6 +313,11 @@ class Editor implements EditorObservable {
       DOMUtils.DOM.styleSheetLoader._setReferrerPolicy(referrerPolicy);
     }
 
+    const crossOrigin = Options.getCrossOrigin(self);
+    if (Type.isNonNullable(crossOrigin)) {
```
TheSpyder: if the default is empty string, it should never be nullable?

modules/tinymce/src/core/main/ts/api/dom/ScriptLoader.ts
```diff
@@ -111,6 +116,11 @@ class ScriptLoader {
         dom.setAttrib(elm, 'referrerpolicy', this.settings.referrerPolicy);
       }
 
+      const crossOrigin = this.settings.crossOrigin;
+      if (Type.isString(crossOrigin) && crossOrigin.length > 0) {
```
TheSpyder: The `length > 0` check here is the only place you acknowledge that the default value is not one of the values allowed by the option type 🤔

TheSpyder (approved):

modules/tinymce/src/core/main/ts/api/Editor.ts
spocke: This was not needed I forgot to add the `crossorigin` type to internal representation of the OptionsTypes. Without that the value returned was undefined so that is why I added the check here. I followed the referrer_policy option to closely here it has a similar issue but not going to fix that in this PR.

modules/tinymce/src/core/main/ts/api/dom/ScriptLoader.ts
spocke: That check is not needed I wanted it to be explicit that if the length is 0 then the attribute is not set. That is an internal behaviour if the `setAttrib` function. But since this is covered by tests I don't think we need to be that explicit so I removed that.

PR #10416, opened 2025-06-17
TINY-11692: Deprecated 'setContent' method in EditorSelection class

michalnieruchalski-tiugo:
Related Ticket: 

Description of Changes:
* Placeholder text

Pre-checks:
* [ ] Changelog entry added

Review:
* [ ] Milestone set

### DIFF
--- a/.changes/unreleased/tinymce-TINY-11692-2025-06-17.yaml
+++ b/.changes/unreleased/tinymce-TINY-11692-2025-06-17.yaml
@@ -1,0 +1,6 @@
+project: tinymce
+kind: Deprecated
+body: Deprecated 'setContent' method in EditorSelection class.
+time: 2025-06-17T14:26:32.309574+02:00
+custom:
+  Issue: TINY-11692
--- a/modules/tinymce/src/core/main/ts/Deprecations.ts
+++ b/modules/tinymce/src/core/main/ts/Deprecations.ts
@@ -125,7 +125,8 @@ const logWarnings = (rawOptions: RawEditorOptions, normalizedOptions: Normalized
 };
 
 const deprecatedFeatures = {
-  fire: 'The "fire" event api has been deprecated and will be removed in TinyMCE 9. Use "dispatch" instead.'
+  fire: 'The "fire" event api has been deprecated and will be removed in TinyMCE 9. Use "dispatch" instead.',
+  selectionSetContent: 'The "setContent" method has been deprecated and will be removed in TinyMCE 9.'
 };
 
 const logFeatureDeprecationWarning = (feature: keyof typeof deprecatedFeatures): void => {
--- a/modules/tinymce/src/core/main/ts/selection/SetSelectionContent.ts
+++ b/modules/tinymce/src/core/main/ts/selection/SetSelectionContent.ts
@@ -9,6 +9,7 @@ import CaretPosition from '../caret/CaretPosition';
 import { SetSelectionContentArgs } from '../content/ContentTypes';
 import { postProcessSetContent, preProcessSetContent } from '../content/PrePostProcess';
 import * as MergeText from '../delete/MergeText';
+import * as Deprecations from '../Deprecations';
 import * as ScrollIntoView from '../dom/ScrollIntoView';
 import { needsToBeNbspLeft, needsToBeNbspRight } from '../keyboard/Nbsps';
 
@@ -104,6 +105,7 @@ const cleanContent = (editor: Editor, args: SetSelectionContentArgs) => {
 };
 
 const setContent = (editor: Editor, content: string, args: Partial<SetSelectionContentArgs> = {}): void => {
+  Deprecations.logFeatureDeprecationWarning('selectionSetContent');
   const defaultedArgs = setupArgs(args, content);
   preProcessSetContent(editor, defaultedArgs).each((updatedArgs) => {
     // Sanitize the content


michalnieruchalski-tiugo:
* **#10416** <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10416?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10416?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

modules/tinymce/src/core/main/ts/Deprecations.ts
```diff
@@ -125,7 +125,8 @@ const logWarnings = (rawOptions: RawEditorOptions, normalizedOptions: Normalized
 };
 
 const deprecatedFeatures = {
-  fire: 'The "fire" event api has been deprecated and will be removed in TinyMCE 9. Use "dispatch" instead.'
+  fire: 'The "fire" event api has been deprecated and will be removed in TinyMCE 9. Use "dispatch" instead.',
+  selectionSetContent: 'The "setContent" method has been deprecated and will be removed in TinyMCE 9.'
```
michalnieruchalski-tiugo:
```suggestion
  selectionSetContent: 'The "setContent" method has been deprecated and will be removed in TinyMCE 9. Use 'mceInsertClipboardContent' event instead'
```

Can someone confirm that `mceInsertClipboardContent` is the correct alternative?

TheSpyder: I thought it was just `editor.insertContent()`. This will overwrite the selection if there is any.

modules/tinymce/src/core/main/ts/selection/SetSelectionContent.ts
```diff
@@ -104,6 +105,7 @@ const cleanContent = (editor: Editor, args: SetSelectionContentArgs) => {
 };
 
 const setContent = (editor: Editor, content: string, args: Partial<SetSelectionContentArgs> = {}): void => {
+  Deprecations.logFeatureDeprecationWarning('selectionSetContent');
```
TheSpyder: You can't just put this here. We use the method internally; it will cause this message to appear in normal use of the editor.

TheSpyder (changes requested):

modules/tinymce/src/core/main/ts/Deprecations.ts
spocke: Yes, `editor.insertContent` is what people should be using however that is not as low level so let see if we get push back on this. We don't want to expose this low level thing since if customers are using it they will report bugs on it that is workaroud in the `insertContent` method.

spocke (changes requested):
You also need to fix the call that is in the image plugin. I think that needs to be a editor.insertContent wrapped in a transact to not capture the `data-mce-id` in an undo level.
There is also a reference in the API docs to this function for createHTML we should update that example to use insertContent instead. We shouldn't suggest in docs to use deprecated things.

modules/tinymce/src/core/main/ts/selection/SetSelectionContent.ts
spocke: Maybe separate this into two functions `setContentInternal` and `setContentExternal` where external is the one we expose in public APIs and add the message to and it just calls `setContentInternal`.

modules/tinymce/src/core/main/ts/Deprecations.ts
TheSpyder:
```suggestion
  selectionSetContent: 'The "setContent" method has been deprecated and will be removed in TinyMCE 9. Use "editor.setContent instead.'
```

michalnieruchalski-tiugo: Did you mean?


```suggestion
  selectionSetContent: 'The "setContent" method has been deprecated and will be removed in TinyMCE 9. Use "editor.insertContent" instead.'
```

modules/tinymce/src/core/main/ts/api/dom/Selection.ts
```diff
@@ -66,6 +66,19 @@ interface EditorSelection {
...
+   *
+   * @method setContent
+   * @param {String} content HTML contents to set could also be other formats depending on settings.
+   * @param {Object} args Optional settings object with for example data format.
+   * @example
+   * // Inserts some HTML contents at the current selection
+   * tinymce.activeEditor.selection.setContent('<strong>Some contents</strong>');
+   * @deprecated This method has been deprecated. Use "editor.insertContent" instead.
+   */
   setContent: (content: string, args?: Partial<SetSelectionContentArgs>) => void;
```
michalnieruchalski-tiugo, line 82: I wonder if all of our JSDOC declarations should be here - inside typescript interface.

Adding @deprecated below, inside 
`const EditorSelection` declaration didn't do anything. Once I moved it to typescript interface I got strikethrough in my code using `setContent`. I also got nice function description on hover.

modules/tinymce/src/plugins/image/main/ts/core/ImageSelection.ts
```diff
@@ -58,7 +58,7 @@ const insertImageAtCaret = (editor: Editor, data: ImageData): void => {
 
   editor.dom.setAttrib(elm, 'data-mce-id', '__mcenew');
   editor.focus();
-  editor.selection.setContent(elm.outerHTML);
+  editor.insertContent(elm.outerHTML);
```
michalnieruchalski-tiugo, line 61: @spocke 

I don't think I have to put it inside transaction. As it's inside transaction. See:

https://github.com/tinymce/tinymce/blob/main/modules/tinymce/src/plugins/image/main/ts/api/Commands.ts#L12

modules/tinymce/src/plugins/image/test/ts/browser/core/ImageSelectionTest.ts
```diff
@@ -59,8 +59,7 @@ describe('browser.tinymce.plugins.image.core.ImageSelectionTest', () => {
                   width: str.is('100'),
                   height: str.is('100')
                 }
-              }),
-              s.element('br', {})
+              })
```
michalnieruchalski-tiugo, line 62: I think I can just update those test cases.
I believe lack of `br` and `p` tag not being retained are the differences between `selection.setContent` and `editor.insertContent`.

I'll make sure I can update the tests tomorrow. However I'm interested if you have any hints on that.

michalnieruchalski-tiugo:
> You also need to fix the call that is in the image plugin. I think that needs to be a editor.insertContent wrapped in a transact to not capture the `data-mce-id` in an undo level.
> There is also a reference in the API docs to this function for createHTML we should update that example to use insertContent instead. We shouldn't suggest in docs to use deprecated things.

Thanks for pointing that out. I'll create a separate ticket and PR for docs.

modules/tinymce/src/core/main/ts/Deprecations.ts
spocke: It should be `insertContent`. The editor.setContent replaces all of the editor contents the `editor.selection.setContent` replaces the selected content so and so does `editor.insertContent`.

modules/tinymce/src/core/main/ts/api/dom/Selection.ts
spocke: We might need to verify that this change works with moxiedoc. I think you have the `tinymce-docs` repo you should be able to generate some moxiedoc stuff by pointing it to your local branch. Can't remember how but there was some scripts for it if I remember correctly.
If we want to avoid moxiedoc from processing this one you could omit the `@method` part I think.

modules/tinymce/src/plugins/image/main/ts/core/ImageSelection.ts
spocke: Nice yeah then we don't need it. 👍

modules/tinymce/src/plugins/image/test/ts/browser/core/ImageSelectionTest.ts
spocke: Yes, its probably trimming unused BRs in the `insertContent` one that is the more advanced one so I would imagine that it would do more stuff. So this change seems fine to me.

.changes/unreleased/tinymce-TINY-11692-2025-06-17.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Deprecated
+body: Deprecated 'setContent' method in EditorSelection class.
```
spocke: I think the EditorSelection is a bit internal so maybe we change the wording so that it points to how a user would likely call this.

```suggestion
body: Deprecated 'editor.selection.setContent' method.
```

modules/tinymce/src/core/main/ts/Deprecations.ts
```diff
@@ -125,7 +125,8 @@ const logWarnings = (rawOptions: RawEditorOptions, normalizedOptions: Normalized
 };
 
 const deprecatedFeatures = {
-  fire: 'The "fire" event api has been deprecated and will be removed in TinyMCE 9. Use "dispatch" instead.'
+  fire: 'The "fire" event api has been deprecated and will be removed in TinyMCE 9. Use "dispatch" instead.',
+  selectionSetContent: 'The "setContent" method has been deprecated and will be removed in TinyMCE 9. Use "editor.insertContent" instead.'
```
zuzanna-antas: I wonder if we should specify that its a `editor.selection.setContent` that is deprecated. Just to be more precise

ztomaszyk (approved):

modules/tinymce/src/core/main/ts/Deprecations.ts
michalnieruchalski-tiugo: I think 
```
'The "editor.selection.setContent" method has been deprecated and will be removed in TinyMCE 9. Use "editor.insertContent" instead.'
```
would be better. @spocke what do you think?

spocke:
I'm going to force merge this since @TheSpyder is away and we are close to a code freeze. If there is any issues with that then we can hotfix this later.

PR #10421, opened 2025-06-23
TINY-12001: Add TinyMCE 8 license key implementation

ltrouton:
Related Ticket: TINY-12001

Description of Changes:
* Add TinyMCE 8 license key implementation

All main code has already been reviewed in the previous PRs

Pre-checks:
* [x] Changelog entry added

Review:
* [X] Milestone set

## Summary by CodeRabbit

* **New Features**
  * Added support for TinyMCE 8 license keys and introduced a license key manager for managing and validating license keys.
  * Introduced unique identifiers: `pageUid` for the global TinyMCE object and `editorUid` for each editor instance.
  * Exposed license key manager APIs on the editor for license validation.
  * Improved error reporting for license key manager loading issues.

* **Bug Fixes**
  * Ensured license key properties and unique identifiers are immutable and validated as UUIDs.

* **Tests**
  * Expanded test coverage for license key management, unique identifiers, and license validation scenarios.
  * Added utilities and assertions for UUID validation in tests.
  * Updated and reorganized tests to reflect new license key management logic.

* **Chores**
  * Removed legacy license key validation code and related tests.
  * Updated editor and test configurations to consistently include the `license_key` option.


.changes/unreleased/tinymce-TINY-12021-2025-04-09.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Added
+body: Added `pageUid` and `editorUid` properties to tinymce global and editor instance respectively.
```
kimwoodfield: Do you need to repeat `Added` in the `body` if it is already in the `kind`?

modules/katamari/src/main/ts/ephox/katamari/api/Id.ts
```diff
@@ -13,7 +15,7 @@ import * as Num from './Num';
  */
 let unique = 0;
 
-export const generate = (prefix: string): string => {
+const generate = (prefix: string): string => {
   const date = new Date();
   const time = date.getTime();
   const random = Math.floor(Num.random() * 1000000000);
```
kimwoodfield, line 21: I know it's not directly related to the changes in this commit but we can also use `1e9` instead of manually typing 1 billion. Just an FYI 😄

modules/katamari/src/main/ts/ephox/katamari/util/IdUtilts.ts
```diff
@@ -0,0 +1,35 @@
+/* eslint-disable no-bitwise */
+
+const uuidV4Bytes = (): Uint8Array<ArrayBuffer> => {
```
kimwoodfield: Do you need the generic here for `ArrayBuffer`?

modules/katamari/src/main/ts/ephox/katamari/util/IdUtilts.ts
```diff
@@ -0,0 +1,35 @@
...
+  return bytes;
+};
+
+const uuidV4String = (): `${string}-${string}-${string}-${string}-${string}` => {
+  const uuid = uuidV4Bytes();
+  const getHexRange = (startIndex: number, endIndex: number) => {
+    let buff = '';
+    for (let i = startIndex; i <= endIndex; ++i) {
+      const hexByte = uuid[i].toString(16).padStart(2, '0');
+      buff = buff + hexByte;
```
kimwoodfield:
```suggestion
      buff += hexByte;
```

modules/tinymce/src/core/main/ts/api/EditorManager.ts
```diff
@@ -100,6 +101,7 @@ interface EditorManager extends Observable<EditorManagerEventMap> {
   documentBaseURL: string;
   i18n: I18n;
   suffix: string;
+  pageUid: string;
```
kimwoodfield, line 104: If you wanted to be really explicit you could set UUIDs with a custom type rather than `string` but I don't think it's necessary.

modules/tinymce/src/core/main/ts/init/LicenseKeyManager.ts
```diff
@@ -0,0 +1,109 @@
...
+  };
+
+  const load = (editor: Editor, suffix: string): void => {
+    const licenseKey = Options.getLicenseKey(editor);
+    const plugins = new Set(Options.getPlugins(editor));
+    const hasApiKey = Type.isString(Options.getApiKey(editor));
+
+    // Early return if addonConstructor already exists
+    if (Obj.has(addOnManager.urls, ADDON_KEY)) {
+      return;
```
kimwoodfield, line 51: I feel like I've discussed early returns like this before and I was advised against it. Could it be reorganized so that we just use the inverse condition to process everything else?

For example, `if (!Obj.has(addOnManager.urls, ADDON_KEY)) {` 

It's not really a huge deal so totally up to you 😄

modules/tinymce/src/core/main/ts/init/LicenseKeyManager.ts
```diff
@@ -0,0 +1,109 @@
...
+    setLicenseKeyManager(licenseKeyManagerApi);
+
+    const validate = editor.licenseKeyManager.validate;
+
+    // Validation of the license key is done asynchronously and does
+    // not block initialization of the editor
+    // The validate function is expected to set the editor to the correct
+    // state depending on if the license key is valid or not
+    // eslint-disable-next-line @typescript-eslint/no-floating-promises
+    validate({});
```
kimwoodfield: I don't see any reason to store this in a variable and then call it. I think it would be more readable and explicit to call the method directly:

`editor.licenseKeyManager.validate({});`

Also I forgot the function definition and how it works but why do we need to pass an empty object? Why can't we just make the argument optional and just call it without one if we are not passing anything.

modules/tinymce/src/core/main/ts/init/Render.ts
```diff
@@ -95,6 +100,12 @@ const loadIcons = (scriptLoader: ScriptLoader, editor: Editor, suffix: string) =
 
 const loadPlugins = (editor: Editor, suffix: string) => {
   const loadPlugin = (name: string, url: string) => {
+    // If licensekeymanager is included in the plugins list
+    // or through external_plugins, skip it
+    if (name === 'licensekeymanager') {
```
kimwoodfield, line 105: Similar to the comment earlier, could we avoid the early return by just using the inverse condition? I think it is readable as it is but I have seen others mention it before so I thought it was worth raising.

modules/tinymce/src/core/main/ts/init/Render.ts
```diff
@@ -124,20 +135,30 @@ const isModelLoaded = (editor: Editor): boolean => {
   return Type.isNonNullable(ModelManager.get(model));
 };
 
+const isLicenseKeyManagerLoaded = (editor: Editor): boolean => {
```
kimwoodfield, line 138: Is this abstraction necessary? I'd say it is pretty clean to just call the method directory in-place on the instance rather than wrapping it in another function but totally up to you.

modules/tinymce/src/core/main/ts/init/Render.ts
```diff
@@ -124,20 +135,30 @@ const isModelLoaded = (editor: Editor): boolean => {
...
 const loadScripts = (editor: Editor, suffix: string) => {
   const scriptLoader = ScriptLoader.ScriptLoader;
 
   const initEditor = () => {
-    // If the editor has been destroyed or the theme and model haven't loaded then
+    // If the editor has been destroyed or the theme, model, licenseKeyManager haven't loaded then
     // don't continue to load the editor
-    if (!editor.removed && isThemeLoaded(editor) && isModelLoaded(editor)) {
+    if (
+      !editor.removed &&
```
kimwoodfield, line 149: Please correct me if I'm wrong but it looks like these conditions represent whether the editor is ready to be initialized. Could we create a separate variable, something like `isEditorReadyToInit` and then just reference that variable. That way we can keep the logic isolated and easier to manage/update later. It's a nitpick but I start to feel uneasy when I see if conditions with a bunch of logical AND like this.

kimwoodfield (approved):

.changes/unreleased/tinymce-TINY-12021-2025-04-09.yaml
ltrouton: Probably not, will fix 👍

modules/katamari/src/main/ts/ephox/katamari/api/Id.ts
ltrouton: That would be better 😄

modules/tinymce/src/core/main/ts/api/EditorManager.ts
ltrouton: Yeah, I did think of this but it locks in a specifc uuid format and is a bit overkill

modules/katamari/src/main/ts/ephox/katamari/util/IdUtilts.ts
ltrouton: I think it is better to be explicit with the native types

modules/tinymce/src/core/main/ts/init/LicenseKeyManager.ts
ltrouton: I go back and forth with early returns. I tend not to use them very much but I do believe they have their use when it improves readability and understanding of what the function is doing.

modules/tinymce/src/core/main/ts/init/Render.ts
ltrouton: Similar to my other comment. Having a early returns feels clearer and more explicity, at least to me

modules/tinymce/src/core/main/ts/init/LicenseKeyManager.ts
ltrouton: Thanks, I will make the change. There is no specific reason for the empty object other than it just being explicit that there is no extra data

modules/tinymce/src/core/main/ts/init/Render.ts
ltrouton: It's not strictly necessary but I just wanted to follow the convention with the model and theme. I'll have a think about it 👍

modules/tinymce/src/core/main/ts/init/Render.ts
ltrouton: That's correct. If we separated it into a variable, it would still have the logical ANDs, so I don't think it would add a lot of value personally. This logic is unlikely to change very often too

modules/tinymce/src/core/test/ts/browser/init/LicenseKeyTest.ts
```diff
@@ -0,0 +1,84 @@
...
+        base_url: '/project/tinymce/js/tinymce'
+      });
+
+      it('TINY-12058: editor.licenseKeyManager should be defined', () => {
+        const editor = hook.editor();
+        assert.isObject(editor.licenseKeyManager);
+        assert.isFunction(editor.licenseKeyManager.validate);
+      });
+
+      it('TINY-12058: validate should return true by default', async () => {
```
shanmen-tiny:
```suggestion
      it('TINY-12058: validate should return false by default', async () => {
```
Nitpick, the stub always returns false. Should we update test description to `false by default`? Applies to both GPL and non GPL context blocks. Otherwise, lgtm :)

shanmen-tiny (approved):

PR #10426, opened 2025-06-24
TINY-12109: Removed return type from setContent and load APIs

spocke:
Related Ticket: TINY-12109

Description of Changes:
* Removed return values from `setContent` and `load` editor APIs

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

## Summary by CodeRabbit

- **Refactor**
  - Updated editor content methods so that setting or loading content no longer return the content value.
- **Tests**
  - Adjusted tests to accommodate updated method behavior, removing assertions on return values when setting content.
- **New Features**
  - Enhanced test utilities to better capture and reuse editor content in specific plugin tests.


hamza0867 (approved):

modules/tinymce/src/core/test/ts/browser/content/EditorContentNotInitializedTest.ts
```diff
@@ -17,11 +16,6 @@ describe('browser.tinymce.core.content.EditorContentNotInitializedTest', () => {
 
   const createEditor = () => new Editor('editor', settings, EditorManager);
 
-  const setContentAndAssertReturn = (editor: Editor, content: AstNode | string) => {
-    const actual = editor.setContent(content);
-    assert.deepEqual(actual, content, 'should return what you tried to set');
-  };
```
lorenzo-pomili: wouldn't be better to change this in:
```
  const setContentAndAssertReturn = (editor: Editor, content: AstNode | string) => {
    editor.setContent(content);
    const actual = editor.getContent();
    assert.deepEqual(actual, content, 'should return what you tried to set');
  };
```
using only `setContent` in these tests seems not to test anything :/

spocke: It tests that it doesn't throw. But maybe I should wrap that so it's more apperent that it's what this is testing now.

TheSpyder (approved):

PR #10437, opened 2025-06-26
TINY-12301: add toggle component in oxide

ztomaszyk:
Related Ticket: TINY-12301

Description of Changes:
* Extracted toggle component from suggested edits into its own reusable component
* Implemented disabled state styling
* Added CSS variables and used CSS color calculations instead of using less functions (one thing to figure out later is .bg-luma-checker checker function)

Pre-checks:

Review:
* [x] Milestone set

## Summary by CodeRabbit

* **New Features**
  * Introduced a new toggle switch component with updated styles for a more modern and interactive appearance.

* **Style**
  * Removed legacy styles for the previous suggested edits toggle and switch components.
  * Integrated the new toggle switch styles into the overall theme for consistency.

### DIFF
--- a/modules/oxide/src/less/theme/components/toggle/toggle.less
+++ b/modules/oxide/src/less/theme/components/toggle/toggle.less
@@ -36,7 +36,7 @@
     border: 1px solid var(--tox-private-slider-border-color);
   }
 
-  .tox-toggle__slider:before {
+  .tox-toggle__slider::before {
     position: absolute;
     content: "";
     height: 12px;
@@ -58,14 +58,14 @@
     box-shadow: 0 0 0 1px @background-color, 0 0 0 3px @keyboard-focus-outline-color;
   }
   
-  input:checked + .tox-toggle__slider:before {
+  input:checked + .tox-toggle__slider::before {
     -webkit-transform: translateX(12px);
     -ms-transform: translateX(12px);
     transform: translateX(12px);
     --tox-private-slider-handle-background-color: @color-white;
   }
 
-  input:disabled + .tox-toggle__slider:before {
+  input:disabled + .tox-toggle__slider::before {
     opacity: 50%;
   }
 }
--- a/modules/oxide/src/less/theme/components/toggle/toggle.less
+++ b/modules/oxide/src/less/theme/components/toggle/toggle.less
@@ -18,9 +18,12 @@
     color: @text-color;
 
     input {
-      display: none;
+      position: absolute;
+      opacity: 0;
+      width: 0;
+      height: 0;
     }
-    
+
     .tox-toggle__slider {
       position: relative;
       cursor: pointer;
@@ -53,7 +56,7 @@
     }
     
     input:focus + .tox-toggle__slider {
-      box-shadow: 0 0 1px @color-tint;
+      box-shadow: 0 0 0 1px @background-color, 0 0 0 3px @keyboard-focus-outline-color;
     }
     
     input:checked + .tox-toggle__slider:before {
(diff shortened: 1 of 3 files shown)


zuzanna-antas:
* **#10437** <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10437?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10437?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

modules/oxide/src/less/theme/components/toggle/toggle.less
```diff
@@ -0,0 +1,73 @@
...
+      background-color: var(--slider-handle-background-color);
+    }
+    
+    input:checked + .tox-toggle__slider {
+      --slider-background-color: @color-tint;
+      --slider-border-color: @color-tint;
+    }
+    
+    input:focus + .tox-toggle__slider {
+      box-shadow: 0 0 0 1px @background-color, 0 0 0 3px @keyboard-focus-outline-color;
```
kimwoodfield: I'm not the biggest fan of how this makes the toggle switch look but I can understand why it was done.

kimwoodfield (approved):
Thanks for moving these styles across.

modules/oxide/src/less/theme/components/toggle/toggle.less
```diff
@@ -0,0 +1,73 @@
...
+    }
+    
+    input:checked + .tox-toggle__slider:before {
+      -webkit-transform: translateX(12px);
```
spocke: Can this be done with other css properties, left/right, margin? At least in the past translateX/Y has all kinds of issues like forcing the element to be GPU accelerated that was causing shifts in rendering or altering the position but not for the browsers layout box model so things like `getBoundingClientRect` would lie.

I know this was in the original CSS just curious of we could fix it now or if we need to fix it. I don't know if browsers have changed and if it's perfectly fine to use now.

Since this is a `:before` pseudo element it's less of an issue but if we had this applied to a regular element then it could become problematic so if we can try to avoid this property in general I think that would be good so it doesn't bleed into things where it shouldn't be used.

modules/oxide/src/less/theme/components/toggle/toggle.less
```diff
@@ -0,0 +1,73 @@
...
+      width: var(--slider-width);
+      height: var(--slider-height);
+      background-color: var(--slider-background-color);
+      border: 1px solid var(--slider-border-color);
+    }
+
+    .tox-toggle__slider:before {
+      position: absolute;
+      content: "";
+      height: 12px;
```
spocke: Should these be calculated based on the width/height of the component?

modules/oxide/src/less/theme/components/toggle/toggle.less
lostkeys: I don't think it's an issue. The GPU rendering is a positive and any computer including your smart watch wouldn't even blink running this code. And I cannot see any real reason why you would need Javascript to look at the rendered position on screen to do anything.

We already use CSS transforms in many places, including tooltips.

modules/oxide/src/less/theme/components/toggle/toggle.less
zuzanna-antas: There has to be some visual indication of focus. I've send how this looks to Piotr and he approved

modules/oxide/src/less/theme/components/toggle/toggle.less
zuzanna-antas: I removed the width/height variables for now as discussed so for now I am leaving it as is.

zuzanna-antas:
Removed the width/height variables and added a `tox-private` prefix to color variables as discussed

modules/oxide/src/less/theme/components/toggle/toggle.less
lostkeys, line 1: The pseudo element "before" should use double colons `::before` in three places below

PR #10447, opened 2025-07-03
TINY-12142: Oxide now ships with ts definitions for css files

hamza0867:
Related ticket: TINY-12136

# Add keyboard navigation hooks to oxide-components

## Description of Changes:
* Added keyboard navigation hooks to oxide-components module
* Implemented various keyboard navigation utilities:
  * `KeyboardFocusHelpers` for focusing elements
  * `KeyboardNavigationHooks` with React hooks for tab, flow, and special key navigation
  * Navigation utilities for DOM movement and traversal
* Added dependencies on `@ephox/katamari` and `@ephox/sugar` for DOM manipulation and utility functions
* Exported new keyboard navigation functionality through the main module entry point

## Pre-checks:
* [x] ~Changelog entry added~
* [x] Tests have been added (if applicable)
* [x] Branch prefixed with `feature/`, `hotfix/` or `spike/`

## Review:
* [x] Milestone set
* [x] ~Docs ticket created (if applicable)~

PR #10448, opened 2025-07-03
TINY-12136: Port the keyboard navigation code from tinycomments repo

hamza0867:
Related Ticket: 

Description of Changes:
* Placeholder text

Pre-checks:
* [ ] Changelog entry added

Review:
* [ ] Milestone set

### DIFF
--- a/modules/oxide-components/package.json
+++ b/modules/oxide-components/package.json
@@ -26,7 +26,7 @@
   ],
   "main": "lib/main.js",
   "module": "lib/main.js",
-  "types": "lib/main.d.ts",
+  "types": "lib/src/main/ts/main.d.ts",
   "scripts": {
     "dev": "vite",
     "build": "tsc -b && vite build",
@@ -77,5 +77,9 @@
       "**/@storybook/**",
       "**/vite**"
     ]
+  },
+  "dependencies": {
+    "@ephox/katamari": "^10.0.0",
+    "@ephox/sugar": "^10.0.0"
   }
 }
--- a/modules/oxide-components/src/main/ts/keynav/KeyboardFocusHelpers.ts
+++ b/modules/oxide-components/src/main/ts/keynav/KeyboardFocusHelpers.ts
@@ -1,0 +1,8 @@
+import { Focus, SelectorFind, SugarElement } from '@ephox/sugar';
+
+export const focusDescendant = (scope: SugarElement<HTMLElement>, selector: string): void =>
+  SelectorFind.descendant<HTMLElement>(scope, selector).each(Focus.focus);
+
+export const focusAncestor = (scope: SugarElement<HTMLElement>, selector: string): void =>
+  SelectorFind.ancestor<HTMLElement>(scope, selector).each(Focus.focus);
+
--- a/modules/oxide-components/src/main/ts/keynav/KeyboardNavigationHooks.ts
+++ b/modules/oxide-components/src/main/ts/keynav/KeyboardNavigationHooks.ts
@@ -1,0 +1,72 @@
+import { Fun } from '@ephox/katamari';
+import { SugarElement } from '@ephox/sugar';
+import { useEffect, type RefObject } from 'react';
+
+import * as FlowType from './keyboard/FlowType';
+import * as KeyingType from './keyboard/KeyingType';
+import * as SpecialType from './keyboard/SpecialType';
+import * as TabbingType from './keyboard/TabbingType';
+
+interface BaseProps {
+  containerRef: RefObject<HTMLElement>;
+}
+
+const bindEvents = (container: HTMLElement, handlers: KeyingType.Handlers) => {
+  container?.addEventListener('keydown', handlers.keydown);
+  container?.addEventListener('keyup', handlers.keyup);
+  container?.addEventListener('focus', handlers.focus);
+
+  return () => {
+    container?.removeEventListener('keydown', handlers.keydown);
+    container?.removeEventListener('keyup', handlers.keyup);
+    container?.removeEventListener('focus', handlers.focus);
+  };
+};
+
+export interface TabKeyingProps extends BaseProps, TabbingType.TabbingConfig { }
+
+export const useTabKeyNavigation = (props: TabKeyingProps): void => {
+  useEffect(() => {
+    const { containerRef } = props;
+
+    if (containerRef.current) {
+      const handlers = TabbingType.create(SugarElement.fromDom(containerRef.current), props);
+      return bindEvents(containerRef.current, handlers);
+    } else {
+      return Fun.noop;
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+};
+
+export interface FlowKeyingProps extends BaseProps, FlowType.FlowConfig { }
+
+export const useFlowKeyNavigation = (props: FlowKeyingProps): void => {
+  useEffect(() => {
+    const { containerRef } = props;
+
+    if (containerRef.current) {
+      const handlers = FlowType.create(SugarElement.fromDom(containerRef.current), props);
+      return bindEvents(containerRef.current, handlers);
+    } else {
+      return Fun.noop;
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [ ]);
+};
+
+export interface SpecialKeyingProps extends BaseProps, SpecialType.SpecialConfig { }
+
+export const useSpecialKeyNavigation = (props: SpecialKeyingProps): void => {
+  useEffect(() => {
+    const { containerRef } = props;
+
+    if (containerRef.current) {
+      const handlers = SpecialType.create(SugarElement.fromDom(containerRef.current), props);
+      return bindEvents(containerRef.current, handlers);
+    } else {
+      return Fun.noop;
+    }
+  }
+  , [ props ]);
+};
--- a/modules/oxide-components/src/main/ts/keynav/keyboard/FlowType.ts
+++ b/modules/oxide-components/src/main/ts/keynav/keyboard/FlowType.ts
@@ -1,0 +1,104 @@
+import { Fun, Optional, Optionals } from '@ephox/katamari';
+import { SelectorFind, SugarElement } from '@ephox/sugar';
+
+import * as DomMovement from '../navigation/DomMovement';
+import * as DomNavigation from '../navigation/DomNavigation';
+
+import type { GeneralKeyingConfig, KeyRuleHandler } from './KeyingModeTypes';
+import * as KeyingType from './KeyingType';
+import * as KeyMatch from './KeyMatch';
+import * as KeyRules from './KeyRules';
+import * as Keys from './Keys';
+
+export interface FlowConfig {
+  readonly selector: string;
+  readonly execute?: (focused: SugarElement<HTMLElement>) => Optional<boolean>;
+  readonly escape?: (focused: SugarElement<HTMLElement>) => Optional<boolean>;
+  readonly allowVertical?: boolean;
+  readonly allowHorizontal?: boolean;
+  readonly cycles?: boolean;
+  readonly focusIn?: boolean;
+  readonly closest?: boolean;
+}
+
+interface FullFlowConfig extends Required<FlowConfig>, GeneralKeyingConfig {}
+
+export const create = (source: SugarElement<HTMLElement>, config: FlowConfig): KeyingType.Handlers => {
+  const defaultExecute = (
+    focused: SugarElement<HTMLElement>
+  ): Optional<boolean> => {
+    focused.dom.dispatchEvent(new window.Event('click'));
+    return Optional.some(true);
+  };
+
+  // TODO: Remove dupe.
+  // TODO: Probably use this for not just execution.
+  const findCurrent = (component: SugarElement<HTMLElement>, flowConfig: FullFlowConfig): Optional<SugarElement<HTMLElement>> => {
+    // TODO: Figure out why we need this, without it enter inside a nested element would navigate up to the parent element if both parent and child have handlers
+    if (flowConfig.closest) {
+      return flowConfig.focusManager.get(component).bind((elem) => SelectorFind.closest(elem, flowConfig.selector));
+    } else {
+      return flowConfig.focusManager.get(component);
+    }
+  };
+
+  const execute = (component: SugarElement<HTMLElement>, _event: Event, flowConfig: FullFlowConfig) => findCurrent(component, flowConfig).bind((focused) => flowConfig.execute(focused));
+
+  const escape = (component: SugarElement<HTMLElement>, _event: Event, flowConfig: FullFlowConfig) => findCurrent(component, flowConfig).bind((focused) => flowConfig.escape(focused));
+
+  const focusIn = (component: SugarElement<HTMLElement>, flowConfig: FullFlowConfig) => {
+    SelectorFind.descendant<HTMLElement>(component, flowConfig.selector)
+      .each((first) => {
+        flowConfig.focusManager.set(component, first);
+      });
+  };
+
+  const moveDelta = (delta: number) => (element: SugarElement<HTMLElement>, focused: SugarElement<HTMLElement>, info: FullFlowConfig): Optional<SugarElement<HTMLElement>> => {
+    const move = info.cycles ? DomNavigation.horizontal : DomNavigation.horizontalWithoutCycles;
+    return move(element, info.selector, focused, delta);
+  };
+
+  const moveLeft = moveDelta(-1);
+  const moveRight = moveDelta(+1);
+
+  const doMove = (movement: KeyRuleHandler<FullFlowConfig>): KeyRuleHandler<FullFlowConfig> =>
+    (component, event, flowConfig) => movement(component, event, flowConfig);
+
+  const getKeydownRules = (
+    _component: SugarElement<HTMLElement>,
+    _se: Event,
+    flowConfig: FullFlowConfig
+  ): Array<KeyRules.KeyRule<FullFlowConfig>> => {
+    const westMovers = [ ...flowConfig.allowHorizontal ? Keys.LEFT : [] ].concat(flowConfig.allowVertical ? Keys.UP : [ ]);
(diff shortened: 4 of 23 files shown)


hamza0867:
* **#10465** <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10465?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10448** <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10448?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10448?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10447** <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10447?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `epic/EPIC-269`


<h2></h2>

TheSpyder:
This project is our chance to do things properly, not start with tech debt and a promise to fix things later. I'll be going through it with a fine-toothed comb.

modules/oxide-components/package.json
```diff
@@ -26,7 +26,7 @@
   ],
   "main": "lib/main.js",
   "module": "lib/main.js",
```
TheSpyder, line 28: don't these need changing as well? 🤔

hamza0867: No. Vite spits out the bundled file in this location. Unless you want us to change the location where the bundled file is spit out ?

TheSpyder: All good I'm just used to `tsc`. It doesn't really matter as long as it's git-ignored.

modules/oxide-components/src/main/ts/keynav/keyboard/FlowType.ts
```diff
@@ -0,0 +1,104 @@
...
+    focused: SugarElement<HTMLElement>
+  ): Optional<boolean> => {
+    focused.dom.dispatchEvent(new window.Event('click'));
+    return Optional.some(true);
+  };
+
+  // TODO: Remove dupe.
+  // TODO: Probably use this for not just execution.
+  const findCurrent = (component: SugarElement<HTMLElement>, flowConfig: FullFlowConfig): Optional<SugarElement<HTMLElement>> => {
+    // TODO: Figure out why we need this, without it enter inside a nested element would navigate up to the parent element if both parent and child have handlers
```
kimwoodfield, line 37: Where are the tests for all of this? Are you creating a separate PR to include them?

modules/oxide-components/src/main/ts/keynav/keyboard/FocusManagers.ts
```diff
@@ -0,0 +1,26 @@
+import { Optional } from '@ephox/katamari';
+import { Focus, SugarElement } from '@ephox/sugar';
+
+export interface FocusManager {
```
kimwoodfield, line 4: Can we please keep the component folder naming structure consistent? We have `Button` and now `keynav` so capitalized and non-capitalized. I don't think React [cares](FocusAssertions) whichever we choose but we should still be consistent.

kimwoodfield: After looking at it some more, I'm thinking that maybe the reason you chose lowercase was because this code contains things more like utilities than actual components? We have some hooks and helpers in here but it is not immediately obvious. I feel like we should probably at least include a README in `keynav` to explain what all of this stuff is at a high level.

modules/oxide-components/src/main/ts/keynav/keyboard/FlowType.ts
TheSpyder: That was Hamza's plan, I am looking for comments just on the implementation for now

TheSpyder: I recall this specific `.closest` flag was something that Alloy navigation handled better. It was nearly a year ago so it might take me a while to remember the details or dig up the old review.

modules/oxide-components/src/main/ts/keynav/keyboard/FocusManagers.ts
TheSpyder: I missed that in the last PR. Our folder names should always be lowercase. I can sort of see why it went uppercase, the idea being that the folder contains the `Button` component and tests, but I also disagree with that approach.

modules/oxide-components/src/main/ts/keynav/KeyboardFocusHelpers.ts
```diff
@@ -0,0 +1,8 @@
+import { Focus, SelectorFind, SugarElement } from '@ephox/sugar';
+
+export const focusDescendant = (scope: SugarElement<HTMLElement>, selector: string): void =>
+  SelectorFind.descendant<HTMLElement>(scope, selector).each(Focus.focus);
+
+export const focusAncestor = (scope: SugarElement<HTMLElement>, selector: string): void =>
```
tiny-ben-tran: These functions are not specific to the keyboard nav. Perhaps they should be put a util file/folder outside of keynav

modules/oxide-components/src/main/ts/keynav/KeyboardNavigationHooks.ts
```diff
@@ -0,0 +1,72 @@
...
+    const { containerRef } = props;
+
+    if (containerRef.current) {
+      const handlers = SpecialType.create(SugarElement.fromDom(containerRef.current), props);
+      return bindEvents(containerRef.current, handlers);
+    } else {
+      return Fun.noop;
+    }
+  }
+  , [ props ]);
```
tiny-ben-tran, line 71: In the original implementation, `props` is passed as dependencies because the handlers are updated on the dropdown changes. It is for performance reasons and to opt for using the React API to get the dropdown items.
 
 Generally, it is not always the case and will have an impact on the performance. We should make passing `props` as dependencies optional.

modules/oxide-components/src/main/ts/keynav/keyboard/Keys.ts
```diff
@@ -0,0 +1,23 @@
+const TAB: ReadonlyArray<number> = [ 9 ];
+const ENTER: ReadonlyArray<number> = [ 13 ];
+const SHIFT: ReadonlyArray<number> = [ 16 ];
+const END: ReadonlyArray<number> = [ 35 ];
+const LEFT: ReadonlyArray<number> = [ 37 ];
+const UP: ReadonlyArray<number> = [ 38 ];
+const RIGHT: ReadonlyArray<number> = [ 39 ];
+const DOWN: ReadonlyArray<number> = [ 40 ];
+const SPACE: ReadonlyArray<number> = [ 32 ];
+const ESCAPE: ReadonlyArray<number> = [ 27 ];
```
tiny-ben-tran, line 12: The value doesn't have to be an array of one number. We should look into changing it

modules/oxide-components/src/main/ts/keynav/keyboard/SpecialType.ts
```diff
@@ -0,0 +1,112 @@
...
+    focus: (event: FocusEvent) => keyingHandlers.handleFocus(source, event)
+  };
+};
+
+export const toBeHandled = (key: KeyboardEvent): boolean => Arr.exists([
+  KeyMatch.inSet(Keys.SPACE),
+  KeyMatch.inSet(Keys.ENTER),
+  KeyMatch.and([ KeyMatch.isShift, KeyMatch.inSet(Keys.ENTER) ]),
+  KeyMatch.and([ KeyMatch.isShift, KeyMatch.inSet(Keys.TAB) ]),
+  KeyMatch.and([ KeyMatch.isNotShift, KeyMatch.inSet(Keys.TAB) ]),
```
tiny-ben-tran, line 106: Then these can be simplified

modules/oxide-components/src/main/ts/keynav/keyboard/FlowType.ts
hamza0867: Yes there is a ticket for tests [here](https://ephocks.atlassian.net/browse/TINY-12468)

modules/oxide-components/src/main/ts/keynav/keyboard/Keys.ts
hamza0867: I agree I think the reason we have an array of number is because this implementation was ported straightly from Alloy. I think we should start using the supported [key](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key) property instead of the [deprecated keyCode](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode) property. What do you think ?

tiny-ben-tran: Yes. Please ensure that `key` is supported by all browsers we support.

hamza0867: If you look at the mdn website, it says it is supported across all browsers, even the mobile ones.

modules/oxide-components/src/main/ts/keynav/keyboard/FocusManagers.ts
hamza0867: The `Button/`  folder was now moved to `button`

modules/oxide-components/src/main/ts/keynav/KeyboardFocusHelpers.ts
hamza0867: Done!

modules/oxide-components/src/main/ts/keynav/KeyboardNavigationHooks.ts
TheSpyder: Could we use just `containerRef` as the dependency? I'm not sure if that is a good idea.

modules/oxide-components/src/main/ts/keynav/keyboard/Keys.ts
TheSpyder: There's an open ticket to use `key`, I think. It's long been the supported path on browsers.

These are arrays because once we add shortcuts that need to use the command or windows key, there are 3 we need to check
https://github.com/tinymce/tinymce/blob/ca9a89cc67cf8ab4eeab3cc11c74c947eb63cdd3/modules/alloy/src/main/ts/ephox/alloy/alien/Keys.ts#L20

modules/oxide-components/src/main/ts/keynav/KeyboardNavigationHooks.ts
tiny-ben-tran: No because `containerRef` is never updated. After React renders a component into DOM, it updates `containerRef.current` property, not the ref itself.

modules/oxide-components/src/main/ts/keynav/keyboard/NavigationTypes.ts
```diff
@@ -0,0 +1,11 @@
+import * as FlowType from './FlowType';
+import * as KeyingType from './KeyingType';
+import * as SpecialType from './SpecialType';
+import * as TabbingType from './TabbingType';
```
TheSpyder, line 4: This is one of the areas where pieces of Alloy were left out that I believe we should keep. Not in this PR, but `execution` in particular will remove the need for special-case handling of things like enter in an input component. They all have their purpose.
https://github.com/tinymce/tinymce/blob/5dfffd8f62ec7cebc5b4b4c64e7aa3d755995b65/modules/alloy/src/main/ts/ephox/alloy/behaviour/keyboard/KeyboardBranches.ts#L10-L17

modules/oxide-components/src/main/ts/keynav/KeyboardNavigationHooks.ts
TheSpyder: Fair enough. I'm still learning.

modules/oxide-components/src/main/ts/keynav/keyboard/FocusManagers.ts
```diff
@@ -0,0 +1,26 @@
...
+  set: (component: SugarElement<HTMLElement>, focusee: SugarElement<HTMLElement>) => void;
+}
+
+const dom = (): FocusManager => {
+  const get = (component: SugarElement<HTMLElement>) => Focus.search(component);
+
+  const set = (_component: SugarElement<HTMLElement>, focusee: SugarElement<HTMLElement>) => {
+    focusee.dom.focus();
+    // TODO: May not need to dispatch a focus event
+    focusee.dom.dispatchEvent(new window.Event('Focus'));
```
TheSpyder: I'm quite confident a normal `HTMLElement` will dispatch this. Seems like something copied from Alloy that we don't really need.

- [x] remove (or maybe just comment it out for now)

modules/oxide-components/src/main/ts/keynav/keyboard/Keys.ts
TheSpyder: There's also a `todo` comment in `KeyMatch.ts`

yarn.lock
TheSpyder, line 1: How did this change so much when you only added internal dependencies, not external? It doesn't even look like upgrades, they're mostly new additions.

TheSpyder (approved):
I might just take a pass at simplifying this later. It was cherry-picked from Alloy, and trying to follow up on why the bits that are missing are important, then deciding whether to include those or rewrite things, will hold up other tasks.

modules/oxide-components/src/main/ts/keynav/keyboard/NavigationTypes.ts
tiny-ben-tran: We definitely need a follow-up ticket or leave a comment to make sure we will port them over in the future.

tiny-ben-tran:
> I might just take a pass at simplifying this later. It was cherry-picked from Alloy, and trying to follow up on why the bits that are missing are important, then deciding whether to include those or rewrite things, will hold up other tasks.

@TheSpyder are you suggesting we should get it right in this PR instead of leaving for later?

TheSpyder:
@tiny-ben-tran as I said in my first comment I would prefer to get things right the first time rather than start with tech debt. We don't get many chances to build something new that we plan to use for 5-10 years.

But I also don't have a photographic memory of Alloy so trying to fix this in PR would take days if not weeks. At least once it's merged the other projects will be converted to it and we'll have more of a real world use-case to verify changes against.

modules/oxide-components/src/main/ts/keynav/keyboard/Keys.ts
spocke: We need to update agar to support key otherwise tests that does anything with the keyboard on our plugins will fail.
Currently agar does its simulated events using keyCode not key. However I started the work on fixing that by adding this lookup table. https://github.com/tinymce/tinymce/blob/main/modules/agar/src/main/ts/ephox/agar/keyboard/Keycodes.ts it's currently only used for simulating input in textareas and input elements using the Keyboard.type function but if we expand that to include Keyboard.keydown/keystroke etc then we can simulate key and then we can use key for anything tinymce releated.

TheSpyder: ah that was part of the holdup, thank you for the reminder

modules/oxide-components/src/main/ts/keynav/KeyboardNavigationHooks.ts
hamza0867: @tiny-ben-tran I'm afraid we run into the stale closure problem because the dependency array is not updated as expected

modules/oxide-components/src/main/ts/keynav/keyboard/NavigationTypes.ts
hamza0867: Or we can port them as we need them, instead of just going through them and porting them. Isn't this how we did it with comments and revision history ?

yarn.lock
hamza0867: This is weird indeed. And by the looks of it, those seem to be storybook dependencies.

modules/oxide-components/src/main/ts/keynav/keyboard/Keys.ts
hamza0867: So what should we do in this case ? Add a TODO comment and Jira ticket to update agar ?

kimwoodfield (approved):

modules/oxide-components/src/main/ts/keynav/keyboard/Keys.ts
TheSpyder: yes

modules/oxide-components/src/main/ts/keynav/keyboard/NavigationTypes.ts
TheSpyder: The problem is that where somebody needs a capability, they might try to implement it instead of asking if there's an Alloy solution that would be appropriate.

It's a follow-up ticket either way, not something to address now.
- [x] create ticket

modules/oxide-components/src/main/ts/keynav/keyboard/Keys.ts
hamza0867: Look what I found: [this ticket](https://ephocks.atlassian.net/browse/TINY-8724) and [this comment](https://ephocks.atlassian.net/browse/TINY-11429?focusedCommentId=142202). I think we should reopen [TINY-8742](https://ephocks.atlassian.net/browse/TINY-8724) and mention it in the TODO comment. What do you think ?

modules/oxide-components/src/main/ts/keynav/keyboard/NavigationTypes.ts
hamza0867: Ticket created [here](https://ephocks.atlassian.net/browse/TINY-12533)

yarn.lock
TheSpyder: Discussed on slack

modules/oxide-components/src/main/ts/keynav/keyboard/Keys.ts
TheSpyder: Any time you see `done: shelved` it's 100% available to reopen. That process was just a big cleanup we did a couple of years ago.

PR #10461, opened 2025-07-09
TINY-12245: fix scroll behavior for an edge case

lorenzo-pomili:
Related Ticket: TINY-12245

Description of Changes:
- [x] avoid scroll to top when the editor is scrolled via "browser search" and the user clicks on the horizontal scroll bar
- [x] enforce a check on a FirerFox quirk that cause an error in the case described in the row below

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved an issue where clicking on a non-selectable element while a selection exists would cause the editor viewport to scroll to the top.
  * Improved handling to prevent unwanted scrolling when focusing the editor body in certain scenarios.
  * Enhanced stability when interacting with image selections to avoid potential errors.

### DIFF
--- a/.changes/unreleased/tinymce-TINY-12245-2025-07-14.yaml
+++ b/.changes/unreleased/tinymce-TINY-12245-2025-07-14.yaml
@@ -1,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: Now clicking on a non selectable element when there is already a selection doesn't scroll back to top
+time: 2025-07-14T09:15:42.517772534+02:00
+custom:
+    Issue: TINY-12245
--- a/.changes/unreleased/tinymce-TINY-12245-2025-07-14.yaml
+++ b/.changes/unreleased/tinymce-TINY-12245-2025-07-14.yaml
@@ -1,6 +1,6 @@
 project: tinymce
 kind: Fixed
-body: Now clicking on a non selectable element when there is already a selection doesn't scroll back to top
+body: Clicking on a non selectable element when the selection is off screen no longer scrolls to the selection
 time: 2025-07-14T09:15:42.517772534+02:00
 custom:
     Issue: TINY-12245
--- a/modules/tinymce/src/core/main/ts/util/Quirks.ts
+++ b/modules/tinymce/src/core/main/ts/util/Quirks.ts
@@ -150,7 +150,9 @@ const Quirks = (editor: Editor): Quirks => {
           rng = selection.getRng();
           // TINY-12245: this is needed to avoid the scroll back to the top when the content is scrolled, there is no selection and the user is clicking on a non selectable editor element
           // example content scrolled by browser search and user click on the horizontal scroll bar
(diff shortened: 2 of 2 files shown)


modules/tinymce/src/core/main/ts/util/Quirks.ts
```diff
@@ -260,7 +264,7 @@ const Quirks = (editor: Editor): Quirks => {
     editor.on('mousedown', (e) => {
       Optionals.lift2(Optional.from(e.clientX), Optional.from(e.clientY), (clientX, clientY) => {
         const caretPos = editor.getDoc().caretPositionFromPoint(clientX, clientY);
-        const img = caretPos?.offsetNode.childNodes[caretPos.offset - (caretPos.offset > 0 ? 1 : 0)] || caretPos?.offsetNode;
+        const img = caretPos?.offsetNode?.childNodes[caretPos.offset - (caretPos.offset > 0 ? 1 : 0)] || caretPos?.offsetNode;
 
         if (img && isEditableImage(img)) {
```
hamza0867: I'd suggest we make a small change here:

```suggestion
        if (Type.isNonNullable(img) && isEditableImage(img)) {
```

hamza0867 (approved):

.changes/unreleased/tinymce-TINY-12245-2025-07-14.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: Now clicking on a non selectable element when there is already a selection doesn't scroll back to top
```
TheSpyder:
```suggestion
body: Clicking on a non selectable element when the selection is off screen no longer scrolls to the selection
```

modules/tinymce/src/core/main/ts/util/Quirks.ts
TheSpyder: would `img ?? isEditableImage(img)` be the same thing? Sometimes I wonder if we should do that more often.

.changes/unreleased/tinymce-TINY-12245-2025-07-14.yaml
lorenzo-pomili: changed

modules/tinymce/src/core/main/ts/util/Quirks.ts
lorenzo-pomili: refactored using `Type.isNonNullable` because the `??` operator doesn't work in this case since it returns the right side if the left side is `null`/`undefined`, so in this case we would have `isEditableImage` called with `null`/`undefined`

TheSpyder: Right, I was thinking about something else. Apologies for the weird request.

TheSpyder (approved):

PR #10463, opened 2025-07-10
TINY-12497: Update split button main button width to auto

ArvinJ-H:
Related Ticket: 
TINY-12497

Description of Changes:
- Update split button main button width to auto

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/oxide/src/less/theme/components/toolbar-button/toolbar-split-button.less
+++ b/modules/oxide/src/less/theme/components/toolbar-button/toolbar-split-button.less
@@ -10,6 +10,7 @@
   .tox-split-button__main {
     border-radius: @toolbar-button-border-radius 0 0 @toolbar-button-border-radius;
     margin-right: 0;
+    width: auto;
 
     &:focus {
       border: @toolbar-button-border;


shanmen-tiny (approved):

kimwoodfield (approved):

TheSpyder (approved):

modules/oxide/src/less/theme/components/toolbar-button/toolbar-split-button.less
```diff
@@ -10,6 +10,7 @@
   .tox-split-button__main {
     border-radius: @toolbar-button-border-radius 0 0 @toolbar-button-border-radius;
     margin-right: 0;
+    width: auto;
```
tiny-ben-tran, line 13: Should there be a min-width and max-width? `tox-tbtn__select-label` should also have a max-width and handle text overflow? (check the convention)

ArvinJ-H: not necessarily, we shouldn't set a restriction on the button length, at least, currently we don't know what value they should be, user can set whatever strings they want, and different language will take up different amount of spaces, cut off the content would be a bad UX

tiny-ben-tran: I disagree that truncating overflow text on buttons is a bad UX. It is a standard UI convention that prevents layout breakage. If requires, tooltip can be used as a fallback.

ArvinJ-H: the button is currently behaving the same as normal button, so we will keep this as is for now

PR #10465, opened 2025-07-13
TINY-12114: Add visual regression testing with playwright

hamza0867:
Related Ticket: 

Description of Changes:
* Placeholder text

Pre-checks:
* [ ] Changelog entry added

Review:
* [ ] Milestone set


hamza0867:
* **#10465** <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10465?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10465?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10448** <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10448?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10447** <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10447?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `epic/EPIC-269`


<h2></h2>

modules/oxide-components/visual-regression-test.mjs
```diff
@@ -0,0 +1,50 @@
+import { exec, execSync } from 'child_process';
+import { setTimeout } from 'timers';
+
+// This script is used to run visual regression tests in a Docker container.
+
+process.chdir('../../');
```
TheSpyder: Why not just put the `Dockerfile` in this folder? We can move it and adjust the paths later if we decide to use it for other things.

modules/oxide-components/visual-regression-test.mjs
```diff
@@ -0,0 +1,50 @@
...
+
+execSync('docker logs --follow oxide-components-test-visual', {
+  stdio: ['ignore', 'inherit', 'inherit'] // Pipe the output to the parent process
+});
+
+// Terminate logs process after a reasonable timeout
+setTimeout(() => {
+  // Finishing the process after displaying the logs
+  process.exit(0);
+}, 5000);
```
TheSpyder: If you get rid of the `-d` you won't need the `logs`, timeout management or "remove if exists" code; it'll either end naturally or the user will use ctrl+c to stop it.

This gives up too much control for a dev-only tool.

modules/oxide-components/src/test/ts/visual.spec.ts-snapshots/button--button-with-counter-desktop-linux-desktop-linux.png
TheSpyder, line 1: These images are error messages?

Dockerfile.playwright-test
```diff
@@ -0,0 +1,18 @@
+# Build stage
+FROM public.ecr.aws/docker/library/node:20 AS build
+
+WORKDIR /tinymce
+
+COPY ./ ./
+
+RUN yarn
+RUN yarn build
```
TheSpyder: Wouldn't it be faster to bind mount the local directory 🤔 

And in either case, `yarn` could be run outside the docker before building it 🤔

Dockerfile.playwright-test
```diff
@@ -0,0 +1,18 @@
...
+
+# Test stage
+FROM mcr.microsoft.com/playwright:v1.53.1-noble AS test
```
kimwoodfield: Could we pull the image tags into build time args at the top of the file? That way we can bump versions in one spot.

modules/oxide-components/visual-regression-test.mjs
```diff
@@ -0,0 +1,50 @@
...
+    return result.length > 0;
+  } catch (error) {
+    return false; // If command fails, assume container doesn't exist
+  }
+}
+
+const exists = containerExists('oxide-components-test-visual');
+
+
+if(exists){
```
kimwoodfield:
```suggestion
if (exists){
```

modules/oxide-components/visual-regression-test.mjs
```diff
@@ -0,0 +1,50 @@
...
+  execSync('docker rm oxide-components-test-visual', {
```
kimwoodfield: If you attempt to remove a container that is still running, you will receive an error. It would be best to make sure that you've stopped it first: https://stackoverflow.com/questions/45357771/stop-and-remove-all-docker-containers

Dockerfile.playwright-test
hamza0867: I don't think this is possible in this case as we are using a multi stage build. Right ?

modules/oxide-components/src/test/ts/visual.spec.ts-snapshots/button--button-with-counter-desktop-linux-desktop-linux.png
hamza0867: Yeah I just noticed right after pushing. I will fix this.

hamza0867: Fixed

modules/oxide-components/visual-regression-test.mjs
hamza0867: I have the Dockerfile in the root of the monorepo because the image needs to have the other libraries as well so that it can build the tinymce project correctly inside the container.

modules/oxide-components/visual-regression-test.mjs
hamza0867: Indeed!

modules/oxide-components/visual-regression-test.mjs
hamza0867: Removed

modules/oxide-components/visual-regression-test.mjs
hamza0867: removed this code

Dockerfile.playwright-test
hamza0867: The problem with that is that there are some dependencies that compile differently depending on the OS, Mac is not the same as linux so it's preferable to install dependencies inside the container. That's what I know, I might be wrong.

Bind mounting the local directory means the container can write to the repository as well, that's not what I want. I just want to copy the files inside the container and then run the build in there.

That's why I am bind mounting the folder that contains the screenshots from playwright, because I want the container to save those files to the repo so they can be used in CI.

Dockerfile.playwright-test
kimwoodfield: Yep, it's totally possible even with multi-stage builds.

Here's an example:
```
ARG NODE_VERSION=20
ARG PLAYWRIGHT_VERSION=v1.53.1-noble

# Build stage
FROM public.ecr.aws/docker/library/node:${NODE_VERSION} AS build
```

Give it a try 😄

Dockerfile.playwright-test
TheSpyder: I appreciate that my request will be hard to meet, but the impact of not bind-mounting is local changes in development will require rebuilding the entire docker image. That's extremely slow compared to a bind mount which might eventually allow for a fast reload (or even hot reload) watch loop.

Don't be scared of the container writing to the repository. Git is enough to guard against that becoming a problem.

The only concern I agree needs resolving is the cross-platform dependencies. It's not clear to me why that's a problem, but I haven't tried this docker playwright setup myself yet.

Dockerfile.playwright-test
hamza0867: Oh I misunderstood you at first, I thought you mean I should put the `FROM base_image` instruction twice at the top. My bad. Done!

Dockerfile.playwright-test
hamza0867: I believe this now done! The Dockerfile is much simpler and it works using bind mounted volume

modules/oxide-components/visual-regression-test.mjs
hamza0867: It seems, indeed, I didn't need the Dockerfile to be in the root of the monorepo

kimwoodfield (approved):

Jenkinsfile
```diff
@@ -319,6 +319,8 @@ timestamps { alertWorseResult(
   processes['playwright'] = runPlaywrightPod(cacheName, 'playwright-tests') {
     exec('yarn -s --cwd modules/oxide-components test-ci')
     junit allowEmptyResults: true, testResults: 'modules/oxide-components/scratch/test-results.xml'
+    exec('yarn -s --cwd modules/oxide-components test-visual-ci')
+    junit allowEmptyResults: true, testResults: 'modules/oxide-components/scratch/test-results-visual.xml'
```
TheSpyder, line 323: I strongly suggest adding an archive step so we capture the images of the visual regression instead of just the XML report

This could be a new ticket/PR

hamza0867: I believe this is done. Please look at the next line of code.

TheSpyder: I think so. I suspect this will have problems if things fail, but at least you've acknowledged the intent to capture the data.

modules/oxide-components/src/test/ts/visual.spec.ts-snapshots/button--button-with-counter-desktop-chrome-linux-desktop-chrome-linux.png
TheSpyder, line 1: Do we really need browser-specific images? I thought things were mostly consistent.

Perhaps as a separate ticket, see if there's an easy way to share the test image across browsers with a way to override specific edge cases where necessary.

Dockerfile.playwright-test
TheSpyder: Thank you for sticking with me, I think it's a much better result 🙂

modules/oxide-components/src/test/ts/visual.spec.ts-snapshots/button--button-with-counter-desktop-chrome-linux-desktop-chrome-linux.png
hamza0867: There is a way actually. It needs some setting up but I think it's possible. Ticket created [here](https://ephocks.atlassian.net/browse/TINY-12564)

PR #10468, opened 2025-07-15
TINY-12289: Add a z-index to views, moving them above sink elements

TheSpyder:
Related Ticket: TINY-12289

Description of Changes:
* Added `z-index` to work around issues where tooltips show on top of a view. The proper fix, to make them hide correctly, will come later.
* Swapped from `margin-top` on view content to `padding-bottom` on the view toolbar (which just means using a single `padding` value)
* Applied those changes to suggested edits manually, since it doesn't leverage the tox view classes

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/oxide/src/less/theme/components/revisionhistory/revisionhistory.less
+++ b/modules/oxide/src/less/theme/components/revisionhistory/revisionhistory.less
@@ -47,7 +47,6 @@
     display: flex;
     flex: 1;
     height: 100%;
-    margin-top: 8px;
     overflow-x: auto;
     overflow-y: hidden;
     position: relative;
--- a/modules/oxide/src/less/theme/components/suggestededits/suggestededits.less
+++ b/modules/oxide/src/less/theme/components/suggestededits/suggestededits.less
@@ -28,6 +28,10 @@
 @suggestededits-card-resolved-background-color: @suggestededits-dark-transparent-color;
 @suggestededits-card-selected-border-color: @suggestededits-color-tint;
 
+// this file is imported above `view.less`, but we need a variable from it
+// TODO: replace `.tox-suggestededits__toolbar` with `.tox-view__toolbar`
+@view-header-padding: (@pad-sm + @keyboard-focus-outline-width);
+
 .tox {
   [data-mce-name="suggestededits"] .tox-icon .tox-icon--badge {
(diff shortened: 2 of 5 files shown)


modules/oxide/src/less/theme/components/revisionhistory/revisionhistory.less
```diff
@@ -47,7 +47,6 @@
     display: flex;
     flex: 1;
     height: 100%;
-    margin-top: 8px;
```
tiny-ben-tran, line 50: Is this change necessary? It removes a space between the toolbar and the body in the revisionhistory.

TheSpyder: @tiny-ben-tran I replaced it with an extra 8px of `padding-bottom` in the `@view-header-padding` variable of `view.less`. It's not visible in the diff, but that applies to the `.tox-view__header` class.

TheSpyder: As near as I could tell, adding padding to the header replicated the margin on the "body" of both suggested edits and revision history. And it makes more sense to me, if we want that gap between the toolbar and "body" in all cases.

Advanced Code is the only other plugin that uses this view and it doesn't have a toolbar.

modules/oxide/src/less/theme/components/suggestededits/suggestededits.less
```diff
@@ -28,6 +28,10 @@
 @suggestededits-card-resolved-background-color: @suggestededits-dark-transparent-color;
 @suggestededits-card-selected-border-color: @suggestededits-color-tint;
 
+// this file is imported above `view.less`, but we need a variable from it
+// TODO: replace `.tox-suggestededits__toolbar` with `.tox-view__toolbar`
```
ltrouton: Does this have a referenced JIra?

ltrouton (approved):

modules/oxide/src/less/theme/components/suggestededits/suggestededits.less
TheSpyder: No, I thought it could just wait for later UI tasks, but I should know better. I'll make one.

PR #10471, opened 2025-07-15
TINY-12528: persist default avatars between sessions

michalnieruchalski-tiugo:
Related Ticket: TINY-12528

Description of Changes:
* Avatar colors are now determined using a hash function based on the user ID, replacing the previous random selection method. This ensures that each user's default avatar consistently displays the same color across all sessions.
* A variant of `djb2` hash function has been used, ensuring good distribution between colors.
* Hash function has been selected based on empirical test, see
https://github.com/michalnieruchalski-tiugo/hash-benchmark/blob/main/report.md
(There's a [script](https://github.com/michalnieruchalski-tiugo/hash-benchmark/blob/main/index.mjs) inside this repo)

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

## Summary by CodeRabbit

* **New Features**
  * Default avatars now use a consistent color derived from the user ID, ensuring each user's avatar color remains stable and unique.

* **Bug Fixes**
  * Improved avatar generation to produce consistent default avatars based on user ID and name.

* **Tests**
  * Added tests to verify consistent generation of default avatars as SVG images for the same user information.

### DIFF
--- a/modules/tinymce/src/core/main/ts/lookup/DefaultAvatar.ts
+++ b/modules/tinymce/src/core/main/ts/lookup/DefaultAvatar.ts
@@ -37,7 +37,12 @@ const getFirstChar = (name: string): string => {
   }
 };
 
-// For a given string returns integer between 0 and maxValue (inclusive).
+/* For a given string returns integer between 0 and maxValue (inclusive).
+  This function is based on the djb2 hash algorithm reported by Dan Bernstein.
+  You can find more informations here: http://www.cse.yorku.ca/~oz/hash.html
+  The hashing algorithm is using bitwise operators to multiply by 32, and later to ensure a positive integer.
+  The result is then reduced to the range of 0 to maxValue using modulo operation.
+*/
 const djb2Hash = (key: string, maxValue: number) => {
   let hash = 5381;
   for (let i = 0; i < key.length; i++) {


michalnieruchalski-tiugo:
* **#10471** <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10471?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10471?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

modules/tinymce/src/core/main/ts/lookup/DefaultAvatar.ts
```diff
@@ -0,0 +1,71 @@
...
+// For a given string returns integer between 0 and maxValue (inclusive).
+const djb2Hash = (key: string, maxValue: number) => {
+  let hash = 5381;
+  for (let i = 0; i < key.length; i++) {
+    // eslint-disable-next-line no-bitwise
+    hash = ((hash << 5) + hash) + key.charCodeAt(i);
+  }
+  // eslint-disable-next-line no-bitwise
+  return (hash >>> 0) % (maxValue + 1);
+};
```
TheSpyder: Is this code copied from somewhere? We need to add an attribution comment if it was.

.changes/unreleased/tinymce-TINY-12532-2025-07-15.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Improved
+body: Made default avatars to be persisted between sessions.
```
TheSpyder:
```suggestion
body: Default avatars are now generated with a consistent color based on the username.
```
We can't use the terminology "persistent between sessions" as it sounds like we're storing state. "Consistently generated" is both more accurate to what we're doing and less scary for someone reading the changelog.

modules/tinymce/src/core/main/ts/lookup/DefaultAvatar.ts
michalnieruchalski-tiugo: It is a djb2 hashing function. I didn't copy it. I found C implementation: http://www.cse.yorku.ca/~oz/hash.html

and added some changes:
`% (maxValue + 1)` (keep the result between 0 and 25)
`(hash >>> 0)` (change sign bit to 0, to ensure positive number)

so that it would work well with javascript, and with our use-case.

So the idea of the algorithm is owned by Daniel J. Bernstein but this implementation is owned by me.

ztomaszyk (approved):

modules/tinymce/src/core/main/ts/lookup/DefaultAvatar.ts
TheSpyder: So include a link to that page in a function method comment and describe how you changed it.

Just because it's seemingly public information doesn't negate our need to provide attribution; the information may also be useful for somebody looking to maintain this code in 5-10 years.

michalnieruchalski-tiugo: I agree that it’s worth to add more information for future reference. I’ve added a link to this page, described how the function work and mentioned the author that originally came up with this algorithm :)

You seem to have a good understanding of licensing and attributions, so maybe there’s a place for me to learn something.
I’m curious about the attribution standards. Given the fact that I didn’t just copy-pasted someone’s code (different language, lack of modulo at the end) I don’t think I had to leave an attribution for person who implemented the C function. So I think you meant to provide attribution for a person who invented the algorithm (the idea behind the implementation)

We don’t typically attribute algorithm inventors like let’s say Tony Hoare for Quick Sort, or Dijkstra while working with graphs, right?

TheSpyder: Attribution is, at the core, acknowledging that we did not invent everything ourselves. There are legal frameworks around it in many cases but mostly it's just about being nice.

C and JS aren't that different - your code still looks fairly close to the C original.

Old algorithms were created by people who didn't want to restrict the use of their work. Modern researchers aren't always so generous. That doesn't necessarily apply here but better to be safe than sorry.

And the other point I hinted at, maintenance is a key consideration for us. Some of the code in TinyMCE is 20 years old; this code you've added looks nothing like how we develop the rest of the editor. Imagine a junior engineer being asked to adjust how the avatar colours work in 2035. Your comment will help them.

TheSpyder (approved):

PR #10473, opened 2025-07-16
TINY-12536: Loader catches and reports errors from `tinymce.init`

tiny-james:
Related Ticket: [TINY-12536](https://ephocks.atlassian.net/browse/TINY-12536), [TCLOUD-4737](https://ephocks.atlassian.net/browse/TCLOUD-4737)

Description of Changes:
* Previously mcagar’s Loader did not report errors thrown while running `tinymce.init` which means that any tests waiting for that TinyMCE instance have to timeout to fail. This was a problem for informant recently as in those tests TinyMCE is created in a before hook. As the error was not reported to informant, the before hook would timeout and skip the tests instead of failing them. This change instead attempts to catch errors thrown from `tinymce.init` and any rejected promises returned by it.

Pre-checks:
* [X] Changelog entry added
* [X] Tests have been added (if applicable) - I don't believe it is applicable here as it is a test component

Review:
* [x] Milestone set
* [X] Docs ticket created (if applicable) - not applicable

[TINY-12536]: https://ephocks.atlassian.net/browse/TINY-12536?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[TCLOUD-4737]: https://ephocks.atlassian.net/browse/TCLOUD-4737?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling during TinyMCE initialization, ensuring that any errors (both synchronous and asynchronous) are now properly captured and reported back to the caller. This enhances reliability and provides clearer feedback when initialization fails.

### DIFF
--- a/modules/mcagar/src/main/ts/ephox/mcagar/loader/Loader.ts
+++ b/modules/mcagar/src/main/ts/ephox/mcagar/loader/Loader.ts
@@ -1,5 +1,5 @@
 import { TestLogs } from '@ephox/agar';
-import { Arr, Fun, FutureResult, Global, Id, Optional, Result } from '@ephox/katamari';
+import { Arr, Fun, FutureResult, Global, Id, Optional, Result, Type } from '@ephox/katamari';
 import { Attribute, DomEvent, Insert, Remove, SelectorFilter, SugarBody, SugarElement, SugarHead, SugarShadowDom } from '@ephox/sugar';
 
 import { Editor } from '../alien/EditorTypes';
@@ -74,7 +74,8 @@ const setup = (callbacks: Callbacks, settings: Record<string, any>, elementOpt: 
   };
 
   // Agar v. ??? supports logging
-  const onFailure = (err: Error | string, logs?: TestLogs) => {
+  const onFailure = (errU: unknown, logs?: TestLogs) => {
+    const err = Type.isString(errU) || errU instanceof Error ? errU : String(errU);
     // eslint-disable-next-line no-console
     console.log('Tiny Loader error: ', err);
     // Do no teardown so that the failed test still shows the editor. Important for selection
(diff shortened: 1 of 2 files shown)


modules/mcagar/src/main/ts/ephox/mcagar/loader/Loader.ts
```diff
@@ -88,31 +89,34 @@ const setup = (callbacks: Callbacks, settings: Record<string, any>, elementOpt:
...
-          }, 100);
-        });
-
-        editor.once('SkinLoadError', (e) => {
-          callbacks.failure(e.message);
-        });
-      }
-    });
+    try {
+      Promise.resolve(tinymce.init({
```
TheSpyder, line 93: If it's inside a `try/catch` already, does the promise catch really add anything?

tiny-james: Yes, because the try/catch will only catch problems creating the Promise, it won't catch if the promise is rejected - which is where I expect most of the errors will come from since most of TinyMCE's startup is async.

tiny-james: Of course this still relies on TinyMCE itself catching errors and rejecting - I am not certain that this will 100% solve the problem as it depends on how well we've written the startup code. :man_shrugging:

tiny-james: It might be worth considering adding a [global unhandled rejection handler](https://developer.mozilla.org/en-US/docs/Web/API/Window/unhandledrejection_event) but I've never gotten them to work in the past so I figure this is a good first step.

tiny-james: > Because this event can leak data, [Promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) rejections that originate from a cross-origin script won't fire this event.

Ah, and there's the catch. I guess that explains why it didn't work when I tried it.

TheSpyder: I thought `tinymce.init()` returned nothing, turns out it returns a promise. But even then, you shouldn't actually need the `Promise.resolve` you can just `tinymce.init().catch()` 🤔

tiny-james: True, the types aren't available for the `tinymce` global in this file so it made it easier to use Promise.resolve to ensure the value is a promise.

tiny-james: Also it does handle some odd edge cases where something other than a promise gets returned - no idea how that'd actually happen but no reason not to code defensively in a commonly used test component. :-)

modules/mcagar/src/main/ts/ephox/mcagar/loader/Loader.ts
```diff
@@ -74,7 +74,8 @@ const setup = (callbacks: Callbacks, settings: Record<string, any>, elementOpt:
   };
 
   // Agar v. ??? supports logging
-  const onFailure = (err: Error | string, logs?: TestLogs) => {
+  const onFailure = (errU: unknown, logs?: TestLogs) => {
+    const err = Type.isString(errU) || errU instanceof Error ? errU : String(errU);
```
kimwoodfield, line 78: Would it help to also log the [stack](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/stack) when available? I wonder if it could help make it easier to diagnose newly surfaced failures (e.g. whether the throw came from internals vs test setup).

tiny-james: If you think it is needed I can add that. I was just trying to keep things minimal. :grin:

tiny-james: This line:
```
console.log('Tiny Loader error: ', err);
```
should log at least part of the stack. For example I tried this out in Firefox's console:
image

Do you think there is a gain to be had to log the stack specifically?

kimwoodfield: Thanks! I don’t feel strongly - just thinking about the debugging convenience. I thought that logging the stack might help capture more information but if it doesn't really benefit your task too much then I think it's totally fine to keep the PR minimal and skip it. Just wanted to capture the thought while we're here.

PR #10470, opened 2025-07-15
TINY-12479: convert button to transitional state

ztomaszyk:
Related Ticket: TINY-12479

Description of Changes:
* Added CSS custom properties for the button component when custom properties are enabled
* Updated button styling to use custom properties with fallbacks to existing LESS variables
* Added control line height variable in global custom properties

Pre-checks:
* [x] ~~Changelog entry added~~
* [x] ~~Tests have been added (if applicable)~~
* [x] Branch prefixed with `feature/`, `hotfix/` or `spike/`

Review:
* [x] Milestone set
* [x] ~~Docs ticket created (if applicable)~~

GitHub issues (if applicable):

## Summary by CodeRabbit

* **New Features**
  * Introduced CSS custom properties for button components to enable dynamic theming and easier customization of button styles and states.

* **Refactor**
  * Simplified and modularized button styling by replacing static variables with CSS custom properties, improving maintainability.
  * Updated button state styles (disabled, focus, hover, active, enabled) to use new CSS variables for consistent theming.

* **Style**
  * Renamed a global line-height CSS variable and added a new control-specific line-height variable for improved consistency across components.

PR #10464, opened 2025-07-10
TINY-12507: introduce reset layer

ztomaszyk:
Related Ticket: TINY-12507

Description of Changes:
Added CSS layer to the reset styles by wrapping all reset styles in `@layer reset {}`. This will guarantee that the reset is applied before all of the other, unlayered styles and will not fight for specificity with them. 

Pre-checks:
* [x] ~~Changelog entry added~~
* [x] ~~Tests have been added (if applicable)~~
* [x] Branch prefixed with `feature/`, `hotfix/` or `spike/`

Review:
* [x] Milestone set
* [x] ~~Docs ticket created (if applicable)~~

GitHub issues (if applicable):

## Summary by CodeRabbit

* **Style**
  * Enhanced CSS structure by grouping `.tox` reset styles into a dedicated layer for better style management without affecting the user interface.

PR #10462, opened 2025-07-09
TINY-12340: global CSS Custom Properties

ztomaszyk:
Related Ticket: TINY-12340

Description of Changes:
* Added feature flag for enabling CSS custom properties
* Created global custom properties for theming
* Reorganized global styling files for better structure
* Moved reset focus states to reset.less
* Updated variable references to use custom properties with fallbacks

Pre-checks:
* [x] ~~Changelog entry added~~
* [x] ~~Tests have been added (if applicable)~~
* [x] Branch prefixed with `feature/`, `hotfix/` or `spike/`

Review:
* [x] Milestone set
* [x] ~~Docs ticket created (if applicable)~~

GitHub issues (if applicable):

## Summary by CodeRabbit

* **New Features**
  * Introduced a global feature flag to enable advanced theming capabilities.
  * Added a comprehensive set of CSS custom properties for dynamic theming, including support for light and dark modes, color schemes, font settings, spacing, and responsive breakpoints.

* **Style**
  * Updated styles to use CSS custom properties for font, color, and z-index, allowing for easier customization and theme changes.
  * Adjusted and centralized theming variables for improved consistency across the UI.
  * Improved focus outline and border reset styles for better accessibility and browser consistency.

* **Refactor**
  * Reorganized import order in theme files for better structure and maintainability.
  * Removed redundant style declarations and streamlined focus reset rules.

PR #10478, opened 2025-07-16
TINY-12531: convert checkbox to transitional state

ztomaszyk:
Related Ticket: TINY-12531

Description of Changes:
* Added CSS custom properties for the checkbox component when custom properties are enabled
* Updated checkbox styling to use custom properties with fallbacks to existing LESS variables
* Updated hidden input implementation
* Added a new global custom property `--tox-private-control-height` for consistent sizing

Pre-checks:

Review:
* [x] Milestone set

## Summary by CodeRabbit

* **New Features**
  * Introduced new CSS custom properties for checkboxes, allowing for improved theming and customization.
  * Added a global custom property for control height to standardize sizing across components.

* **Style**
  * Updated checkbox styles to use CSS variables instead of hardcoded values or LESS variables.
  * Improved consistency and maintainability of checkbox appearance, including better support for different states and theming options.
  * Removed an unused checkbox box-shadow variable to simplify styling.

### DIFF
--- a/modules/oxide/src/less/theme/components/checkbox/checkbox.less
+++ b/modules/oxide/src/less/theme/components/checkbox/checkbox.less
@@ -37,8 +37,8 @@
     /* Hide from view but visible to screen readers */
     position: absolute;
     opacity: 0;
-    width: 0;
-    height: 0;
+    width: 1px;
+    height: 1px;
   }
   
   .tox-checkbox__icons {
--- a/modules/oxide/src/less/theme/components/checkbox/checkbox.less
+++ b/modules/oxide/src/less/theme/components/checkbox/checkbox.less
@@ -2,6 +2,19 @@
 // Checkbox
 //
 
+.tox-checkbox when (@custom-properties-enabled = true) {
+  --tox-private-checkbox-unselected-color: light-dark(
+    hsl(from var(--tox-private-color-black) h s l / 30%),
+    hsl(from var(--tox-private-color-white) h s l / 20%)
+  );
+  --tox-private-checkbox-selected-color: var(--tox-private-color-tint);
+  --tox-private-checkbox-indeterminate-color: var(--tox-private-checkbox-selected-color);
+
+  --tox-private-checkbox-border-radius: var(--tox-private-control-border-radius);
+  --tox-private-checkbox-focus-box-shadow: inset 0 0 0 1px var(--tox-private-keyboard-focus-outline-color);
+  --tox-private-checkbox-disabled-text-color: var(--tox-private-text-color-muted);
+}
+
 @checkbox-unselected-color: contrast(@background-color, fade(@color-black, 30%), fade(@color-white, 20%));
 @checkbox-selected-color: @color-tint;
 @checkbox-indeterminate-color: @checkbox-selected-color;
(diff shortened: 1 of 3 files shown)


zuzanna-antas:
* **#10478** <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10478?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#10499](https://github.com/tinymce/tinymce/pull/10499) <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10499?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#10500](https://github.com/tinymce/tinymce/pull/10500) <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10500?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10478?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10470** <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10470?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10464** <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10464?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10462** <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10462?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

MitchC1999 (approved):

michalnieruchalski-tiugo (approved):

modules/oxide/src/less/theme/components/checkbox/checkbox.less
```diff
@@ -2,144 +2,143 @@
...
+  min-width: var(--tox-private-control-height, 36px);
+  border-radius: var(--tox-private-checkbox-border-radius, @control-border-radius);
+  
   .tox-checkbox__input {
     /* Hide from view but visible to screen readers */
-    height: 1px;
-    overflow: hidden;
     position: absolute;
-    top: auto;
+    opacity: 0;
```
tiny-ben-tran: I find `opacity: 0` is good for something like a transparent layer (ie: modal).  I wonder if the solution stated on https://css-tricks.com/inclusively-hidden/ is a better approach

zuzanna-antas: In the original article linked in the css-tricks there is a mention that the `opacity: 0` is no longer problematic: 

> Regarding the use of opacity: 0 to hide content, this used to completely hide elements from the ChromeVox web extension screen reader.
> 
> Testing today (June 17, 2021) after a prompt by Rachel Leggett, it appears this is no longer a bug and opacity: 0 will no longer incorrectly completely hide content from the screen reader.
(from https://www.scottohara.me/blog/2017/04/14/inclusively-hidden.html)

modules/oxide/src/less/theme/components/checkbox/checkbox.less
```diff
@@ -2,144 +2,143 @@
...
 @checkbox-selected-color: @color-tint;
 @checkbox-indeterminate-color: @checkbox-selected-color;
 
 @checkbox-disabled-text-color: fade(@text-color, 50%);
 
 @checkbox-focus-box-shadow: inset 0 0 0 1px @button-background-color;
+@checkbox-unfocused-box-shadow: none;
 @checkbox-focus-border-radius: @control-border-radius;
 
-.tox {
```
TheSpyder, line 14: I would think we still need this `tox` wrapper? they're still used further down in this file 🤔

zuzanna-antas: What do you mean they are still used? The `tox` wrapper doesn't do much here tbh. I would say it is redundant with the `tox` prefix we add to each class. Without it we have a little bit simpler selectors.  

Before the layers, `.tox *:not(svg):not(rect)` would win with `.tox-checkbox`, BUT it would lose with `.tox .tox-checkbox`. So the wrapper was added to fix the specificity issues with our CSS reset. With layers it is no longer the case. Styles defined in the reset layer will lose with styles declared in the next layers (and with unlayered styles) no matter the specificity.

TheSpyder: Lower down, there are still `.tox[dir=rtl]` and `.tox:not([dir=rtl])` styles.

What I meant it could be helpful with is making sure our styles are only applied within our root component. Not for specificity, but to prevent accidents if the class is used outside the root for some reason.

zuzanna-antas: For `[dir=rtl]` it's a different case because we do need to check that on the `.tox` element.
I get treating the wrapper as a guard but I don't think we need to use it this way. We namespace our classes by adding a `tox` prefix - this also works as a guard. We know that we are styling our components, not something from the outside. 
I am also thinking about the future, where we might add some detached components - will they be wrapped in the `.tox` element as well?

TheSpyder: If we continue to rely on the `.tox` element for RTL checks then everything will need to be wrapped yes. I don't know if that's the best idea or not - something to consider with this project.

zuzanna-antas: There is no media query for checking RTL, we have to explicitly set the attribute on the wrapper and check it. But why other stuff need to be wrapped as well because of that? 

I was thinking about introducing another layer `tox-rtl-overrides` and apply it on top - for organization and overrides will always win specificity fight.

spocke: I'm not sure we need the `.tox .tox-something` namespacing. It was only there to override the reset rules of `.tox` it needed a higher specificity. The `.tox[dir="rtl") .tox-something` already has a higher specificity than say just `.tox-something` or `.tox-something .tox-somethingelse` so the RTL rule will be applied in the same way as before it's just that the layers now deal with the specificity conflict problems of the reset rules if I understood everything correctly.

TheSpyder: Which is why I said above
> Not for specificity, but to prevent accidents if the class is used outside the root for some reason.

If we require the `tox` root for RTL management, and we don't make it clear somehow that all oxide components must be inside a `tox` root, then at some point we're going to have an RTL bug when a component is accidentally loaded outside the root.

michalnieruchalski-tiugo: If there won't be a `.tox` element wrapping the whole UI, then `.tox:not([dir=rtl])` styles won't be applied. This is a problem, and it's better to break things early by not applying the styles at all, than wait until the developer will notice subtle problems with margins, padding and so on.

~~But, let's take a step back. What do we gain by having `.tox[dir=rtl]`, why can't it be just `[dir=rtl]`. 
Why do we have to specify `dir` only on the `.tox` root element?~~
EDIT: I believe we just don't want to inherit dir from the integrator's page.

spocke: Since we will be feeding the css custom properties variables in `.tox` class nothing will really render correctly unless it's inside that element.

zuzanna-antas: Added a `tox` wrapper back

TheSpyder: Thank you. It can be hard to sort these things out over PR comments, maybe at some point we should sit down to discuss in a meeting on zoom or when we're all in the same office next month.

PR #10481, opened 2025-07-18
TINY-12545: Update image plugin decorative image check to align with a11y plugin

ArvinJ-H:
Related Ticket: 
TINY-12545

Description of Changes:
- Update image plugin decorative image check to align with a11y plugin

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/tinymce/src/plugins/image/test/ts/browser/core/ImageDataTest.ts
```diff
@@ -685,4 +685,144 @@ describe('browser.tinymce.plugins.image.core.ImageDataTest', () => {
...
   });
+
+  it('TINY-12545: Read decorative image with only alt=""', () => {
+    const image = createHtml('<img src="some.gif" alt="">');
+    const data = readFromImage(image);
+    assertModel({
+      src: 'some.gif',
+      alt: '',
+      title: '',
+      width: '',
```
kimwoodfield: We could remove quite a few LOC from the file by defining the object we are passing to assertModel once with empty values and then just spreading it in place, then we just provide the values we want to update on top of them. It would also make it easier to read which values are being modified across each scenario

kimwoodfield (approved):

modules/tinymce/src/plugins/image/test/ts/browser/core/ImageDataTest.ts
ArvinJ-H: `ImageData` asks for all the field, so they gotta be there

kimwoodfield: They will still be there. I'm on my phone so it's hard to format but essentially you'd want your object you're passing to assertModel:

const assertModelObj = {
  width: '',
  height: '',
  Etc.
}

Then when you need to pass that object to assertModel you can do:

...assertModelObj,
alt: 'foo'

ArvinJ-H: ah yeah I see, updated

kimwoodfield: Much clearer!

ltrouton (approved):

PR #10497, opened 2025-07-28
TINY-12468: Add tests for flow type navigation

hamza0867:
# Add tests for flow type keyboard navigation 

Related Ticket: TINY-12468

Description of Changes:
* Added proper JSDoc comments to keyboard navigation hooks and components
* Created FlowType stories to demonstrate and test keyboard navigation functionality
* Updated KeyMatch and Keys to use modern keyboard event properties (code, key) alongside legacy 'which'
* Fixed click event handling in FlowType to use the standard click() method
* Enhanced Storybook configuration to better handle TypeScript documentation

Pre-checks:
* [x] ~Changelog entry added~

Review:
* ~[x] Milestone set~
* ~[x] Docs ticket created (if applicable)~

## Summary by CodeRabbit

- New Features
  - Added interactive Storybook demos for flow keyboard navigation (with and without cycling).
  - Improved key handling to recognize key/code across browsers.
  - Default “execute” now triggers a native click for more consistent behavior.

- Documentation
  - Enhanced Storybook docs with TypeScript prop extraction, focused defaults, and canvas styling.

- Tests
  - Added browser-based keyboard navigation tests; visual tests now skip tagged stories.

- Chores
  - Updated build/test scripts and configurations to support browser testing and monorepo builds.

### DIFF
--- a/modules/oxide-components/package.json
+++ b/modules/oxide-components/package.json
@@ -39,8 +39,10 @@
     "test-visual-local": "node ./visual-regression-test.mjs",
     "test-visual-local-update": "node ./visual-regression-test.mjs update",
     "test-ci": "vitest --watch=false --reporter=junit --outputFile=scratch/test-results.xml",
-    "test-storybook": "vitest --project storybook",
-    "ci": "yarn lint && yarn build-storybook && yarn build && yarn test"
+    "test-browser-manual": "vitest --watch --project browser --browser.headless false",
+    "test-browser-headless": "vitest run --project browser",
+    "test-storybook": "vitest run --project storybook",
+    "ci": "yarn lint && yarn --cwd ../../ lerna run build --scope @tinymce/oxide-components --include-dependencies && yarn test"
   },
   "peerDependencies": {
     "react": "^18.3.1",
@@ -73,7 +75,8 @@
     "storybook": "^9.0.10",
     "terser": "^5.28.1",
     "vite": "^6.3.5",
-    "vitest": "^3.1.4"
+    "vitest": "^3.1.4",
+    "vitest-browser-react": "^1.0.1"
   },
   "workspaces": {
     "nohoist": [
--- a/modules/oxide-components/src/main/ts/keynav/keyboard/KeyMatch.ts
+++ b/modules/oxide-components/src/main/ts/keynav/keyboard/KeyMatch.ts
@@ -1,17 +1,18 @@
 import { Arr } from '@ephox/katamari';
 
+import type { Key } from './Keys';
+
 export type KeyMatcher = (evt: KeyboardEvent) => boolean;
 
-const inSet = (keys: ReadonlyArray<number>): KeyMatcher => (event: KeyboardEvent) => {
-  return Arr.contains(keys, event.which);
+const inSet = (keys: ReadonlyArray<Key>): KeyMatcher => (event: KeyboardEvent) => {
+  return Arr.exists(keys, (key) => key.code === event.code || key.key === event.key || key.which === event.which);
 };
 
 const and = (preds: KeyMatcher[]): KeyMatcher => (event: KeyboardEvent) => Arr.forall(preds, (pred) => pred(event));
 
-const is = (key: number): KeyMatcher => (event: KeyboardEvent) => {
-  // TODO: 'which' is deprecated and ie is not longer supported
-  // should `key` be used instead?
-  return event.which === key;
+const is = (key: Key): KeyMatcher => (event: KeyboardEvent) => {
+  // use key but fallback to which to support @ephox/agar
+  return event.key === key.key || event.code === key.code || event.which === key.which;
 };
 
 const isShift = (event: KeyboardEvent): boolean => {
--- a/modules/oxide-components/src/main/ts/keynav/keyboard/Keys.ts
+++ b/modules/oxide-components/src/main/ts/keynav/keyboard/Keys.ts
@@ -1,15 +1,24 @@
 // TODO: This file would need to be updated once we work on https://ephocks.atlassian.net/browse/TINY-8724
 
-const TAB: ReadonlyArray<number> = [ 9 ];
-const ENTER: ReadonlyArray<number> = [ 13 ];
-const SHIFT: ReadonlyArray<number> = [ 16 ];
-const END: ReadonlyArray<number> = [ 35 ];
-const LEFT: ReadonlyArray<number> = [ 37 ];
-const UP: ReadonlyArray<number> = [ 38 ];
-const RIGHT: ReadonlyArray<number> = [ 39 ];
-const DOWN: ReadonlyArray<number> = [ 40 ];
-const SPACE: ReadonlyArray<number> = [ 32 ];
-const ESCAPE: ReadonlyArray<number> = [ 27 ];
+export interface Key {
+  readonly code: string;
+  readonly key: string;
+  readonly which: number;
+}
+
+const TAB: ReadonlyArray<Key> = [{ code: 'Tab', key: 'Tab', which: 9 }];
+const ENTER: ReadonlyArray<Key> = [{ code: 'Enter', key: 'Enter', which: 13 }];
+const SHIFT: ReadonlyArray<Key> = [
+  { code: 'ShiftLeft', key: 'Shift', which: 16 },
+  { code: 'ShiftRight', key: 'Shift', which: 16 },
+];
+const END: ReadonlyArray<Key> = [{ code: 'End', key: 'End', which: 35 }];
+const LEFT: ReadonlyArray<Key> = [{ code: 'ArrowLeft', key: 'ArrowLeft', which: 37 }];
+const UP: ReadonlyArray<Key> = [{ code: 'ArrowUp', key: 'ArrowUp', which: 38 }];
+const RIGHT: ReadonlyArray<Key> = [{ code: 'ArrowRight', key: 'ArrowRight', which: 39 }];
+const DOWN: ReadonlyArray<Key> = [{ code: 'ArrowDown', key: 'ArrowDown', which: 40 }];
+const SPACE: ReadonlyArray<Key> = [{ code: 'Space', key: ' ', which: 32 }];
+const ESCAPE: ReadonlyArray<Key> = [{ code: 'Escape', key: 'Escape', which: 27 }];
 
 export {
   TAB,
--- a/modules/oxide-components/src/test/ts/browser/keynav/KeynavFlowTypeTest.spec.tsx
+++ b/modules/oxide-components/src/test/ts/browser/keynav/KeynavFlowTypeTest.spec.tsx
@@ -1,0 +1,246 @@
+import { Optional } from '@ephox/katamari';
+import { Focus, SelectorFind, SugarElement } from '@ephox/sugar';
+import { userEvent } from '@vitest/browser/context';
+import { useFlowKeyNavigation } from 'oxide-components/keynav/KeyboardNavigationHooks';
+import { forwardRef, useRef } from 'react';
+import { beforeEach, describe, expect, test } from 'vitest';
+import { render } from 'vitest-browser-react';
+
+const styles = `.stay:focus {
+    background-color: #cadbee;
+  }
+  .skip:focus {
+    background-color: red;
+  }
+`;
+
+const store: string[] = [];
+
+const clearStore = (store: string[]) => {
+  store.splice(0, store.length);
+};
+
+interface ItemProps {
+  classes: string[];
+  name: string;
+  store: string[];
+}
+
+const Item = ({ classes, name, store }: ItemProps) =>
+  (
+    <span
+      className={classes.join(' ')}
+      style={{
+        display: 'inline-block',
+        width: '20px',
+        height: '20px',
+        margin: '2px',
+        border: '1px solid ' + (classes.includes('stay') ? 'blue' : 'yellow')
+      }}
+      tabIndex={-1}
+      onClick={() => store.push('item.execute: ' + name)}
+    />
+  );
+
+const Container = forwardRef<HTMLDivElement, { store: string[]; testId: string }>(({ store, testId }, ref) => (
+  <div
+    data-testid={testId}
+    ref={ref}
+    className="flow-keying-test"
+    style={{
+      background: 'white',
+      width: '200px',
+      height: '200px'
+    }}
+  >
+    <Item classes={[ 'stay', 'one' ]} name="one" store={store} />
+    <Item classes={[ 'stay', 'two' ]} name="two" store={store} />
+    <Item classes={[ 'skip', 'three' ]} name="three" store={store} />
+    <Item classes={[ 'skip', 'four' ]} name="four" store={store} />
+    <Item classes={[ 'stay', 'five' ]} name="five" store={store} />
+  </div>
+));
+
+const sequence = async (doc: Document, key: string, identifiers: Array<{ label: string; selector: string }>) => {
+  for (let i = 0; i < identifiers.length; i++) {
+    await userEvent.keyboard(`{${key}}`);
+    const element = doc.querySelector(identifiers[i].selector);
+    await expect.element(element, {
+      message: 'Focus should move from ' + (i > 0 ? identifiers[i - 1].label : '(start)') + ' to ' + identifiers[i].label
+    }).toHaveFocus();
+  }
+};
+
+const targets = {
+  one: { label: 'one', selector: '.one' },
+  two: { label: 'two', selector: '.two' },
+  five: { label: 'five', selector: '.five' }
+};
+
+describe('KeynavFowTypeTest', () => {
+
+  beforeEach(() => {
+    clearStore(store);
+  });
+
+  test('FlowKeying with cycles', async () => {
+
+    const FlowKeyingWithCycles = () => {
+      const ref = useRef<HTMLDivElement>(null);
+      useFlowKeyNavigation({
+        containerRef: ref,
+        selector: '.stay',
+        escape: () => {
+          store.push('flow.onEscape');
+          return Optional.some(true);
+        },
+        cycles: true,
+      });
+      return (
+        <Container ref={ref} store={store} testId='container' />
+      );
+    };
+
+    const { getByTestId, baseElement } = render(<FlowKeyingWithCycles />, {
+      wrapper: ({ children }) => (
+        <>
+          <style>
+            {styles}
+          </style>
+          <div className='tox'>
+            {children}
+          </div>
+        </>
+      )
+    });
+    const doc = baseElement.ownerDocument;
+    const container = getByTestId('container');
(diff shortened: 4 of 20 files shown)


hamza0867:
* **#10497** <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10497?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10497?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

TheSpyder:
Are there Alloy tests we could take some inspiration from for these?

hamza0867:
These are straight ports from the Alloy tests

modules/oxide-components/src/main/ts/keynav/keyboard/FlowTypeDemo.tsx
```diff
@@ -0,0 +1,74 @@
...
+): Optional<boolean> => {
+  focused.dom.click();
+  return Optional.some(true);
+
+};
+
+export const FlowTypeDemo: React.FC<Omit<FlowKeyingProps, 'containerRef'>> = ({
+  selector,
+  execute = defaultExecute,
+  escape = Fun.constant(Optional.none()),
```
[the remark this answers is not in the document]
hamza0867: This was discussed in a PR before. We don't need to worry about this.

modules/oxide-components/package.json
```diff
@@ -29,7 +29,7 @@
   "types": "lib/src/main/ts/main.d.ts",
   "scripts": {
     "dev": "vite",
-    "build": "yarn build-storybook && tsc -b && vite build",
+    "build": "tsc -b ../agar && yarn build-storybook && tsc -b && vite build",
```
TheSpyder: Dependencies should be managed by the overall build system, not specific to a module.

But it looks like you have a circular dependency; if `tsc -b` can't run before `storybook build`, but storybook depends on agar, that's a big problem.

hamza0867: It is a problem because agar is a workspace dependency. Usually dependencies grabbed from npm would be ready for consumption already. So this seems like an edge case. What do you think ?

hamza0867: One other thing is, I think we would not have hit this issue in the premium monorepo, because nx can analyze the dependency graph and build the dependencies in the correct order, so agar would be properly built instead of just calling `tsc -b ../agar` here in the script for building `oxide-components`. Does that make sense ?

TheSpyder: Having it as a workspace dependency isn't the problem. Once it's added to `devDependencies` here, agar should build before it no matter what build system we're using.

This monorepo is designed to build with a single `tsc -b` at the base of the project, but the components interrupt that with the storybook requirement.

modules/oxide-components/src/main/ts/keynav/keyboard/FlowType.stories.tsx
```diff
@@ -0,0 +1,289 @@
...
+  >
+    <Item classes={[ 'stay', 'one' ]} name="one" store={store} />
+    <Item classes={[ 'stay', 'two' ]} name="two" store={store} />
+    <Item classes={[ 'skip', 'three' ]} name="three" store={store} />
+    <Item classes={[ 'skip', 'four' ]} name="four" store={store} />
+    <Item classes={[ 'stay', 'five' ]} name="five" store={store} />
+  </div>
+));
+
+const sequence = async (doc: Document, key: string, identifiers: Array<{ label: string; selector: string }>) => {
```
spocke: I wonder if we should start having utility modules.. This seems like a generic function that we might want to reuse for other tests.

modules/oxide-components/src/test/ts/visual.spec.ts
```diff
@@ -6,7 +6,7 @@ import storybook from '../../../storybook-static/index.json' with { type: 'json'
 
 // Only run tests on stories, not other documentation pages.
 const stories = Object.values(storybook.entries).filter(
-  (e) => e.type === 'story'
+  (e) => e.type === 'story' && !e.tags.includes('skip-visual-testing')
```
spocke, line 9: I think we should still use katamari for consistency. If we ever decide to move away from it to native things then we should switch it all.

modules/oxide-components/src/main/ts/keynav/keyboard/Keys.ts
```diff
@@ -1,15 +1,21 @@
...
-const DOWN: ReadonlyArray<number> = [ 40 ];
-const SPACE: ReadonlyArray<number> = [ 32 ];
-const ESCAPE: ReadonlyArray<number> = [ 27 ];
+export interface Key {
+  code: string;
+  key: string;
+  which: number;
+}
+
+const TAB: ReadonlyArray<Key> = [{ code: 'Tab', key: 'Tab', which: 9 }];
```
spocke, line 9: Since all of these are just single array items does it even make sense to have them as an array. I see that the code has inSet etc. But if the set is only one we don't really need a set. This could be a later cleanup in some other PR but I think we should get rid of it.

modules/oxide-components/src/main/ts/keynav/keyboard/FlowTypeDemo.tsx
```diff
@@ -0,0 +1,74 @@
+
```
spocke: Maybe we should have a lint rule for whitespace at the beginning of the file. I'm surprised that we don't have that.

modules/oxide-components/src/main/ts/keynav/keyboard/FlowType.stories.tsx
```diff
@@ -0,0 +1,289 @@
...
+          ]
+        );
+      });
+
+      await step('Test execute', async () => {
+        await userEvent.keyboard('{Enter}');
+        const executeAssertion = () => {
+          store.assertEq('Check that execute has fired on the right target', [ 'item.execute: one' ]);
+        };
+        await expect(executeAssertion).not.toThrow();
```
spocke: This doesn't combine well. I think maybe we could expose a `store.get()` function so that you instead can do:
```await expect(store.get(), 'Check that execute has fired on the right target').toEqual([ 'item.execute: one' ]);```

File: modules/oxide-components/src/main/ts/keynav/keyboard/KeyMatch.ts [hunk 8]
```diff
@@ -1,17 +1,19 @@
...
+const inSet = (keys: ReadonlyArray<Key>): KeyMatcher => (event: KeyboardEvent) => {
+  return Arr.exists(keys, (key) => key.code === event.code || key.key === event.key || key.which === event.which);
 };
 
 const and = (preds: KeyMatcher[]): KeyMatcher => (event: KeyboardEvent) => Arr.forall(preds, (pred) => pred(event));
 
-const is = (key: number): KeyMatcher => (event: KeyboardEvent) => {
+const is = (key: Key): KeyMatcher => (event: KeyboardEvent) => {
   // TODO: 'which' is deprecated and ie is not longer supported
   // should `key` be used instead?
```
spocke: Should this comment be changed. We can get rid of `which` if we fix agar.

modules/oxide-components/src/main/ts/keynav/keyboard/FlowType.stories.tsx
```diff
@@ -0,0 +1,289 @@
...
+        store.clear();
+      });
+
+      await step('Test escape', async () => {
```
spocke: There doesn't seem to be a way to skip a step or focus in on running just one step. Will this be a problem as we scale this up.

modules/oxide-components/src/main/ts/keynav/keyboard/FlowType.stories.tsx
```diff
@@ -0,0 +1,289 @@
...
+    });
+
+    await step('Test escape', async () => {
+      await userEvent.keyboard('{Escape}');
+      store.assertEq('Check that escape handler has fired', [ 'flow.onEscape' ]);
+    });
+  },
+};
+
+export const FlowKeyingWithoutCycles: Story = {
```
spocke: How would one split this up generally if we have one export point with multiple stories and want each story to be it's own module. Is the structure `FlowType.stories.tsx` then `stories/FlowKeyingWithoutCycles.tsx` or do one have `FlowKeyingWithoutCycles.stories.tsx` with just one story in it. I kind of want to see what the structure would be here if we had hundreds of tests.

modules/oxide-components/src/main/ts/keynav/keyboard/FlowTypeDemo.tsx
spocke, line 1: Why is this demo file separate from the stories. Is there a change we accidentally include this in the bundle. How do we make sure we separate concerns here testing, demos and production code?

modules/oxide-components/src/main/ts/keynav/keyboard/Keys.ts
```diff
@@ -1,15 +1,21 @@
...
-const ENTER: ReadonlyArray<number> = [ 13 ];
-const SHIFT: ReadonlyArray<number> = [ 16 ];
-const END: ReadonlyArray<number> = [ 35 ];
-const LEFT: ReadonlyArray<number> = [ 37 ];
-const UP: ReadonlyArray<number> = [ 38 ];
-const RIGHT: ReadonlyArray<number> = [ 39 ];
-const DOWN: ReadonlyArray<number> = [ 40 ];
-const SPACE: ReadonlyArray<number> = [ 32 ];
-const ESCAPE: ReadonlyArray<number> = [ 27 ];
+export interface Key {
```
spocke, line 3: Annotate with readonly if this is supposed to be immutable.

modules/oxide-components/src/main/ts/keynav/keyboard/FlowType.stories.tsx
hamza0867: Indeed!

modules/oxide-components/package.json
hamza0867: Removed the agar dependency all together, there was no need for it 🤷‍♂️

modules/oxide-components/src/main/ts/keynav/keyboard/FlowType.stories.tsx
hamza0867: Indeed. Seeing how simple the implementation was in this case, I just changed the store to be an array of strings

modules/oxide-components/src/main/ts/keynav/keyboard/Keys.ts
hamza0867: For now yes. But I think there are keys that would be different depending on the OS so those would require having an array. I think an example is the meta key.

modules/oxide-components/src/main/ts/keynav/keyboard/FlowTypeDemo.tsx
hamza0867: I changed the structure so that it's more clear. But maybe we should use vitest for testing instead of storybook.

modules/oxide-components/package.json
hamza0867: But I bumped into the same problem with `katamari`. And the solution was indeed using Lerna to properly build the project and its dependencies in the correct order

modules/oxide-components/src/test/ts/browser/defaultavatar/DefaultAvatarTest.spec.ts
```diff
@@ -1,8 +1,7 @@
 import { Optional, Fun } from '@ephox/katamari';
+import * as DefaultAvatar from 'oxide-components/defaultavatar/DefaultAvatar.ts';
```
michalnieruchalski-tiugo: Awesome! That's much better

modules/oxide-components/package.json
TheSpyder: I still think the lerna command should be at the project root not here, but I guess it's a start

modules/oxide-components/src/test/ts/visual.spec.ts
hamza0867: The problem with using katamari in this file specifically (playwright runs this file, this doesn't affect the code in `src/main`) is that playwright struggles to resolve katamari because playwright runs this file in nodejs and expects katamari to be a cjs module but it's compiled to an es module but doesn't specify `"type":"module"` in `package.json`. And this is the only file we need for visual regression. I think in this case we can skip using katamari. But I agree that it is preferable to use katamari in general as it has better type support and type inference.

modules/oxide-components/src/main/ts/keynav/keyboard/FlowType.stories.tsx
hamza0867: Yes stories can be imported and then exported. So the way to do it would be as you described and then for example: `FlowType.stories.tsx` would import stories from the `stories` folder and then export them.

modules/oxide-components/src/main/ts/keynav/keyboard/Keys.ts
```diff
@@ -1,15 +1,21 @@
...
+const TAB: ReadonlyArray<Key> = [{ code: 'Tab', key: 'Tab', which: 9 }];
+const ENTER: ReadonlyArray<Key> = [{ code: 'Enter', key: 'Enter', which: 13 }];
+const SHIFT: ReadonlyArray<Key> = [{ code: 'Shift', key: 'Shift', which: 16 }];
+const END: ReadonlyArray<Key> = [{ code: 'End', key: 'End', which: 35 }];
+const LEFT: ReadonlyArray<Key> = [{ code: 'ArrowLeft', key: 'ArrowLeft', which: 37 }];
+const UP: ReadonlyArray<Key> = [{ code: 'ArrowUp', key: 'ArrowUp', which: 38 }];
+const RIGHT: ReadonlyArray<Key> = [{ code: 'ArrowRight', key: 'ArrowRight', which: 39 }];
+const DOWN: ReadonlyArray<Key> = [{ code: 'ArrowDown', key: 'ArrowDown', which: 40 }];
+const SPACE: ReadonlyArray<Key> = [{ code: 'Space', key: 'Space', which: 32 }];
+const ESCAPE: ReadonlyArray<Key> = [{ code: 'Escape', key: ' ', which: 27 }];
```
spocke: Why is `key` a space character it should be the key name so `Escape`. If you press escape here you see that it produces a `Escape` when the escape key is pressed. https://w3c.github.io/uievents/tools/key-event-viewer.html

Also not sure we need `code` we `key` should be the right value to check. Code is the physical key regardless of layout. Might not matter much for these keys but it does matter when it comes to letter keys.

modules/oxide-components/src/test/ts/browser/keynav/KeynavFlowTypeTest.spec.tsx
```diff
@@ -0,0 +1,245 @@
...
+          </style>
+          <div className='tox'>
+            {children}
+          </div>
+        </>
+      )
+    });
+    const doc = baseElement.ownerDocument;
+    const container = getByTestId('container');
+    const firstChild: HTMLElement | null = container.element().querySelector('.one');
```
spocke: Don't we want o blow up if this fails?

modules/oxide-components/src/test/ts/visual.spec.ts
spocke: Ok makes sense. Thanks for clearifying. I think if we are to make them proper modules we might need to also need to add `.js` to all imports within the module unless node js fixed that.

spocke (approved):

modules/oxide-components/src/test/ts/browser/keynav/KeynavFlowTypeTest.spec.tsx
hamza0867: It will blow up 2 lines down the line, but this is more like the `getOrDie` we use in tests that should never fail

modules/oxide-components/tsconfig.test.json
```diff
@@ -0,0 +1,39 @@
...
+  "compilerOptions": {
+    "tsBuildInfoFile": "./node_modules/.tmp/tsconfig.node.tsbuildinfo",
+    "target": "ES2022",
+    "lib": [
+      "ES2022",
+      "DOM"
+    ],
+    "types": [
+      "node"
+    ],
```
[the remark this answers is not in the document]
hamza0867: describe, it and expect are imported from Vitest and are not globals

modules/oxide-components/src/test/ts/browser/keynav/KeynavFlowTypeTest.spec.tsx
spocke: Is the `one` expected to always exist. Then this isn't safe enough. Then we should have a querySelectorOrDie abstraction or something on top of this. If I alter `one` to `oneX` then it fails on focusing not on finding the element is that expected?

hamza0867: We can use this then: `expect(firstChild).not.toBeNull();`. What do you think ?

hamza0867: I ended up using Sugar api with `getOrDie`

modules/oxide-components/src/test/ts/visual.spec.ts
```diff
@@ -6,7 +6,7 @@ import storybook from '../../../storybook-static/index.json' with { type: 'json'
...
 );
```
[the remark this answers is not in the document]
hamza0867: tags are always present. C.f: https://github.com/storybookjs/storybook/blob/next/code/core/src/csf/story.ts#L28

modules/oxide-components/src/test/ts/browser/keynav/KeynavFlowTypeTest.spec.tsx
spocke: I'm starting to think that we might need to wrap that Locator API. We spent hard time avoiding null in the code base this opens the door for that and through that unsafe patterns or hard to understand errors. It would have also been useful to be able to extend that Locator with our custom things if we wanted to.
Getting the locator is fine since it throws an exception if you try to find a ID that doesn't exist. It's the querying part later that is a bit of a risk since that might return null. If you manually need to add an expect every time you query something then that code should be a re-usable thing something that tries to get and throws an error if it doesn't get that thing.

It would have been nice if you could do:
```
    const container = getByTestId('container');
    const firstChild = container.querySelectorOrDie<HTMLElement>('.one', 'Failed to find one');
```

Or like this if we returned an Optional:
```
    const container = getByTestId('container');
    const firstChild = container.querySelector<HTMLElement>('.one').getOrDie('Failed to find one');
```

But that means we would have to wrap everything basically. But maybe that is a good thing.

We should strive towards using concrete values or optionals and avoid null.

hamza0867: We can extend them quite easily: https://vitest.dev/guide/browser/locators.html#custom-locators

PR #10501, opened 2025-07-30
TINY-12504: Issue with encoded links.

HAFRMO:
Related Ticket: TINY-12504

Description of Changes:
Replace `&amp;`in links with `&`. This is done at `mceInsertLink` level to ensure it is done in all cases.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved link insertion so that ampersand characters (`&`) in URLs are no longer encoded as `&amp;` in the link's `href` attribute.

* **Tests**
  * Enhanced test coverage and organization for link insertion, including new tests to verify correct handling of ampersands in URLs.


.changes/unreleased/tinymce-TINY-12504-2025-07-29.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: mceInsertLink command replaces `&amp;` encoding with `&` in href.
```
TheSpyder: Needs to be past tense, otherwise it sounds like the bug is that the replacement happens 🙂 

```suggestion
body: Links inserted with `&amp;` encoding are now decoded to `&` before inserting
```

I'm also not convinced this is the correct fix - _every_ link will be impacted by this, fixing only pasted links would be done in [SmartPaste](https://github.com/tinymce/tinymce/blob/bab20d5a0e2f719d78d71757461af23b2ec3f8fa/modules/tinymce/src/core/main/ts/paste/SmartPaste.ts#L47) - but maybe that's ok.

HAFRMO: Yeah, `&amp;` would break for any link anyway and is the one most likely to appear by mistake

hamza0867 (approved):

modules/tinymce/src/core/main/ts/api/commands/LinkCommands.ts
```diff
@@ -13,7 +13,7 @@ export const registerCommands = (editor: Editor): void => {
 
     if (Type.isObject(linkDetails) && Type.isString(linkDetails.href)) {
       // Spaces are never valid in URLs and it's a very common mistake for people to make so we fix it here.
-      linkDetails.href = linkDetails.href.replace(/ /g, '%20');
+      linkDetails.href = linkDetails.href.replace(/ /g, '%20').replace(/&amp;/g, '&');
```
spocke, line 16: Is this really the right spot. Since the link dialog and the pasting a link acts differently I think it's somewhere else this should go. Maybe entity decode the clipboard contents before paste.
If you look in the paste/SmartPaste.ts it looks like it just takes the HTML as is and then uses that as the link instead of first entity decoding it if it comes form HTML. That means other HTML like things like < and > etc that needs to be encoded might also fail.

- [x] Consider different approach

spocke: One could argue that maybe the space fix here should be when input is made not in the low level command but it has been like for a long time.

spocke (approved):

modules/tinymce/src/core/main/ts/api/commands/LinkCommands.ts
HAFRMO: Yes, it's the presence of the space check that makes me feel that this is an appropriate place to put it. If it'd not been there I'd agree with you that it would not be the right place and instead move it a bit higher up in the stack where it comes from the paste.

PR #10517, opened 2025-08-05
TINY-12589: Added more abilities to the parser.

HAFRMO:
Related Ticket: TINY-12589

Description of Changes:
As the new fullpage plugin will allow the user to change certain values within the header we may need to use the domparser to prevent anything funny happening.
At the moment the parser deliberately only parses the body and discarding everything else, including the very header that the fullpage plugin wants to check. This PR adds a flag to make it parse and return the entirety of the html.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

## Summary by CodeRabbit

* **New Features**
  * Optional parser mode to preserve full documents (keeps head and elements outside body when enabled).
* **Bug Fixes**
  * Sanitization updated to allow the html element, improving handling of full HTML documents.
* **Tests**
  * Added tests validating full-document parsing with and without the new parser mode.
* **Chores**
  * Added an unreleased changelog entry for this improvement.

### DIFF
--- a/.changes/unreleased/tinymce-TINY-12589-2025-07-04.yaml
+++ b/.changes/unreleased/tinymce-TINY-12589-2025-07-04.yaml
@@ -1,0 +1,6 @@
+project: tinymce
+kind: Added
+body: Added parser option `useDocumentNotBody` to allow parser to include header and other elements outside of the body.
+time: 2025-07-04T07:22:38.998267+02:00
+custom:
+    Issue: TINY-12589
--- a/.changes/unreleased/tinymce-TINY-12589-2025-07-04.yaml
+++ b/.changes/unreleased/tinymce-TINY-12589-2025-07-04.yaml
@@ -1,6 +1,6 @@
 project: tinymce
 kind: Added
-body: Added parser option `useDocumentNotBody` to allow parser to include header and other elements outside of the body.
+body: Added option to set `root_body` to `#document` to allow parser to include header and other elements outside of the body element.
 time: 2025-07-04T07:22:38.998267+02:00
 custom:
     Issue: TINY-12589
--- a/.changes/unreleased/tinymce-TINY-12589-2025-07-04.yaml
+++ b/.changes/unreleased/tinymce-TINY-12589-2025-07-04.yaml
@@ -1,6 +1,6 @@
 project: tinymce
 kind: Improved
-body: The editor `DomParser` API can now parse full documents when the API's `root_name` option is set to `#document`.
+body: "The editor `DomParser` API can now parse full documents when the API's `root_name` option is set to `#document`."
 time: 2025-07-04T07:22:38.998267+02:00
 custom:
     Issue: TINY-12589
--- a/.changes/unreleased/tinymce-TINY-12589-2025-07-04.yaml
+++ b/.changes/unreleased/tinymce-TINY-12589-2025-07-04.yaml
@@ -1,6 +1,6 @@
 project: tinymce
-kind: Added
-body: Added option to set `root_body` to `#document` to allow parser to include header and other elements outside of the body element.
+kind: Improved
+body: The editor `DomParser` API can now parse full documents when the API's `root_name` option is set to `#document`.
 time: 2025-07-04T07:22:38.998267+02:00
 custom:
     Issue: TINY-12589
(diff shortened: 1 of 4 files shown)


spocke:
The parser used to be able to parse a full document by doing this in v5:

```
tinymce.html.DomParser({
   root_name: '#document'
}).parse('<html xmlns="foo"><head><title>X</title></head><body>X</body></html>');
```

Is it possible to get back to that then we don't need to add a new option and it would also parse the attributes out of the HTML element.

The fragment structure is then also firstChild being `html` not `head`.

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
```diff
@@ -318,15 +320,20 @@
     const makeWrap = () => {
       if (format === 'xhtml') {
         // If parsing XHTML then the content must contain the xmlns declaration, see https://www.w3.org/TR/xhtml1/normative.html#strict
-        return `<html xmlns="http://www.w3.org/1999/xhtml"><head></head><body>${content}</body></html>`;
+        if (useDocumentNotBody) {
+          return `<html xmlns="http://www.w3.org/1999/xhtml">${content}</html>`;
+        } else {
+          return `<html xmlns="http://www.w3.org/1999/xhtml"><head></head><body>${content}</body></html>`;
+        }
```
TheSpyder: I don't think format `xhtml` is common anymore, so does this change even make much difference to the feature?

Perhaps the `xmlns` attribute should be optionally included in the checks below if the `<html` path is used.

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
```diff
@@ -80,6 +81,7 @@
   sandbox_iframes_exclusions?: string[];
   sanitize?: boolean;
   validate?: boolean;
+  useDocumentNotBody?: boolean;
```
spocke: I don't think we need this new option if we make it work like pre v5 where it would parse in a full document context if you set the root_name to `#document` then this becomes more of a hidden feature.

HAFRMO: Sure.

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
HAFRMO: You mean making it from its own if-statement to putting it into the two below?

TheSpyder: Yeah, the fact that it's xhtml should only control whether the `xmlns` attribute is added to the `<html` tag. The rest of the checks below still make sense whether xhtml or not.

.changes/unreleased/tinymce-TINY-12589-2025-07-04.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Added
+body: Added option to set `root_body` to `#document` to allow parser to include header and other elements outside of the body element.
```
TheSpyder: Just to clarify that this is an API, not an editor config option. And correcting the option name.
```suggestion
kind: Improved
body: The editor `DomParser` API can now parse full documents when the API's `root_name` option is set to `#document`.
```

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
```diff
@@ -309,24 +311,27 @@
...
     // special element then we need to wrap it so the internal content is handled appropriately.
     const isSpecialRoot = Obj.has(schema.getSpecialElements(), rootName.toLowerCase());
     const content = isSpecialRoot ? `<${rootName}>${html}</${rootName}>` : html;
     const makeWrap = () => {
-      if (format === 'xhtml') {
-        // If parsing XHTML then the content must contain the xmlns declaration, see https://www.w3.org/TR/xhtml1/normative.html#strict
-        return `<html xmlns="http://www.w3.org/1999/xhtml"><head></head><body>${content}</body></html>`;
-      } else if (/^[\s]*<head/i.test(html) || /^[\s]*<html/i.test(html) || /^[\s]*<!DOCTYPE/i.test(html)) {
-        return `<html>${content}</html>`;
+      const isxhtml = format === 'xhtml';
```
spocke: You could move this up and also use it on line 315 or maybe even refactor it out so that the format is a boolean all the way. Unless that becomes a big refactor our side of the scope of this PR.

PR #10536, opened 2025-08-14
TINY-12346: Create IconButton component

hamza0867:
Related Ticket: TINY-12346

Description of Changes:

- Enhanced Button component with variant support (primary, secondary, outlined, naked)
- Added proper className handling in Button component
- Created new IconButton component with icon rendering capabilities
- Updated Button stories to showcase different variants
- Added IconButton stories for all variants
- Reorganized storybook structure by moving Button to components/Button
- Updated visual test snapshots to reflect new components and variants

Pre-checks:

- [x] ~Changelog entry added~
- [x] ~Tests have been added (if applicable)~
- [x] Branch prefixed with `feature/`, `hotfix/`or `spike/`

Review:

- [x] Milestone set

## Summary by CodeRabbit

- **New Features**
    - Added Icon and IconButton components.
    - Button gained variant support: primary, secondary, outlined, naked.
- **Documentation**
    - New Storybook stories for Button and IconButton with pseudo-state previews (hover/active/focus).
- **Refactor**
    - Button component reorganized and reworked while preserving its public API/usage.
- **Chores**
    - Upgraded Storybook to 9.0.17 and added pseudo-states addon.


hamza0867:
* **#10547** <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10547?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10541** <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10541?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10540** <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10540?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10539** <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10539?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10536** <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10536?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10536?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

modules/oxide-components/src/main/ts/button/Button.component.tsx
```diff
@@ -1,20 +1,39 @@
...
+      return [ 'tox-button' ];
+  }
+};
+
 export interface ButtonProps extends ButtonHTMLAttributes<HTMLButtonElement> {
+  readonly variant?: 'primary' | 'secondary' | 'outlined' | 'naked';
 }
 
 /** Primary UI component for user interaction */
 export const Button: React.FC<ButtonProps> = ({
```
kimwoodfield: I have a few ideas on how we can extend this component's API while still keeping it flexible:

- `disabled` - Simple boolean flag, and surely we already have some styles or something for this.
- `loading` - Another boolean flag. This one would be useful if clicking the button triggers a network request or something else asynchronous. We can add a spinner or loading pattern and set the button to disabled while the value provided to `loading` is `true`.
- `size` - Having a few default sizes would definitely help keep things consistent and we could have each one as a union type like you've done for `variant`.
- `startIcon` & `endIcon` - This would allow us to include an icon/svg within the button and it's placement. Super useful. 

Not something that we need to update as part of this PR, but let's keep thinking of ways to improve these components.

kimwoodfield (approved):
A few CodeRabbit suggestions which I think could be worth considering, but it's a good start. Thanks for adding this.

modules/oxide-components/src/main/ts/button/Button.component.tsx
hamza0867: disabled already exists because it is an html attribute of a button element already. As for the other suggestions, I would talk to @nie-dzwiedz about them 👍

modules/oxide-components/src/main/ts/iconbutton/IconButton.component.tsx
```diff
@@ -0,0 +1,25 @@
...
+export interface IconButtonProps extends IconProps, Omit<ButtonProps, 'children'> {
+}
+
+export const IconButton: React.FC<IconButtonProps> = ({ icon, resolver, ...props }) => {
+  return (
+    <Button className={classes([ 'tox-button--icon' ])} {...props}>
+      <Icon icon={icon} resolver={resolver} />
+    </Button>
+  );
+};
```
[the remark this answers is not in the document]
TheSpyder: could we omit `className` from `IconButtonProps` on line 15 instead?

TheSpyder (approved):

modules/oxide-components/src/main/ts/button/Button.component.tsx
spocke: But do we want to expose all the possible HTML attributes you can set on a button here or should we be restrictive about what props we expose. I think we should be restrictive. 

So that would mean that the props type doesn't extend from ButtonHtmlAttributes but it's all custom props mapping to html attributes.

If we want to have a generic put any attribute in as a prop then that should be an internal component that we don't expose like a `CommonButton` or something. The benefit of that it's becomes a internal library concern of how we deal with raw attributes.

I think this is super important to get right that the API surface of the `oxide-components` should be something we control something some what strict. I don't think you should be able to set arbitrary classes on a button for example that should be an internal class concern. If you are setting custom classes on things then that is a bespoke button not a generic oxide button that we use all over the UI.

modules/oxide-components/src/main/ts/iconbutton/IconButton.component.tsx
```diff
@@ -0,0 +1,25 @@
+import type { ButtonProps } from '../button/Button.component';
+import { Button } from '../main';
+import { classes } from '../utils/Styles';
+
+interface IconProps {
+  /** The name of the icon */
+  icon: string;
+  /** The function to resolve the icon name to an html string */
+  resolver: (icon: string) => string;
```
spocke: Could you add a TODO comment that this resolver would eventually default to being handled by the Universe/Backstage or what ever we will call the context.

modules/oxide-components/src/main/ts/iconbutton/IconButton.component.tsx
```diff
@@ -0,0 +1,25 @@
...
+}
+
+const Icon = ({ icon, resolver }: IconProps) => {
```
spocke: This is it's own component it should be in it's own file IconComponent but not sure if it should be a re-usable thing maybe we need to have a set of internal components as well. Something that we don't expose out but are re-used between components?
So maybe we have three sets of component types common, internal and bespoke we have them as separate story categories but only common and bespoke is exposed in the entry points of the library.

modules/oxide-components/src/test/ts/visual.spec.ts-snapshots/components-iconbutton--outlined-desktop-firefox-linux-desktop-firefox-linux.png
spocke, line 1: We need to figure out how to handle the png blobs here in some external storage. Could we log a ticket for that might need help from Seb or someone.

This is around ~300k of uncompressable binary data that is being added to this PR.

Once we have say hundreds of these and we change the skin then it will start to add up real fast and we don't know how often that will happen. I think we need to fix this before it becomes a problem.

spocke (approved):

modules/oxide-components/src/main/ts/iconbutton/IconButton.component.tsx
kimwoodfield: Yeah I agree `Icon` should be separate too. I think it would be nice to have a folder within oxide-components that contains all of our icons as individual SVG React components so it would be like our own icon pack in component form. Something very similar to this: https://www.npmjs.com/package/react-icons

modules/oxide-components/src/main/ts/iconbutton/IconButton.component.tsx
```diff
@@ -0,0 +1,25 @@
...
+  return <span className={classes([ 'tox-icon' ])} dangerouslySetInnerHTML={{ __html: resolver(icon) }} />;
+};
```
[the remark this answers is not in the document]
hamza0867: We are aware of the dangers of `dangerouslySetInnerHTML` but in this case the result of the resolver comes from an internal library, not user input.

As for the aria-label, that would be set on the button using the icon, not the icon itself.

modules/oxide-components/src/main/ts/iconbutton/IconButton.component.tsx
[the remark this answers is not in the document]
hamza0867: @TheSpyder I think we might want to merge instead, in case we need to apply additional classes on top of the default `tox-button--icon`. I can see this happening in new projects. Unless we want to remove this ability and apply only `tox-button--icon`. What do you think ?

modules/oxide-components/src/main/ts/iconbutton/IconButton.stories.tsx
```diff
@@ -0,0 +1,55 @@
...
+    layout: 'centered',
+  },
+  // This component will have an automatically generated Autodocs entry: https://storybook.js.org/docs/writing-docs/autodocs
+  tags: [ 'autodocs' ],
+  // Use `fn` to spy on the onClick arg, which will appear in the actions panel once invoked: https://storybook.js.org/docs/essentials/actions#action-args
+  args: {
+    icon: 'leftArrow',
+    resolver: Fun.constant(resolvedIcon)
+  },
+} satisfies Meta<typeof IconButton>;
```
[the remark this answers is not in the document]
hamza0867: Handling a11y will come in later PRs when we add translation and icon packs

modules/oxide-components/.storybook/main.ts
```diff
@@ -15,6 +15,7 @@ const config: StorybookConfig = {
     getAbsolutePath("@storybook/addon-a11y"),
     getAbsolutePath("@storybook/addon-vitest"),
     getAbsolutePath("@storybook/addon-docs"),
+    getAbsolutePath("storybook-addon-pseudo-states")
   ],
```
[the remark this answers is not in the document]
hamza0867: The `parameters.pseudo` blocks are working quite fine. Where did you get this information from ?

modules/oxide-components/src/test/ts/visual.spec.ts-snapshots/components-iconbutton--outlined-desktop-firefox-linux-desktop-firefox-linux.png
hamza0867: A discussion started around this topic in this [thread](https://tiugotech.slack.com/archives/C04KXE6QUTX/p1755365079294339)

modules/oxide-components/src/main/ts/iconbutton/IconButton.component.tsx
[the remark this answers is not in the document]
TheSpyder: @hamza0867 that makes sense for generic buttons, for specific icon buttons I think we should retain tight control over the class name.

At least for now; if we change our minds it's easy to allow merging on the property. If we allow merge now and want to block it later that will cause compile errors wherever it was used.

PR #10539, opened 2025-08-16
TINY-12555: Create Textarea component

hamza0867:
Related Ticket: 

Description of Changes:
* Placeholder text

Pre-checks:
* [ ] Changelog entry added

Review:
* [ ] Milestone set


hamza0867:
* **#10547** <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10547?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10541** <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10541?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10540** <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10540?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10539** <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10539?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10539?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10536** <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10536?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

hamza0867:
@coderabbitai review

modules/oxide-components/src/main/ts/components/textarea/Textarea.component.tsx
```diff
@@ -0,0 +1,103 @@
...
+      // In mathematical terms: newRows = (scrollHeight / lineHeight) such as minRows <= newRows <= maxRows
+      const newRows = Math.min(Math.max(minRows, Math.floor(scrollHeight / lineHeight)), maxRows);
+
+      if (newRows !== rows) {
+        setRows(newRows);
+      }
+
+      clone.parentNode?.removeChild(clone);
+    }
+  }, [ rows, value, computeMinRows, computeMaxRows ]);
```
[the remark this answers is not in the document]
hamza0867: 1. I want to create a new clone every time so it lives in the Dom only for the time when we are running this effect.
2. We can't have a textarea without a parent, because that comes from the Dom, not a created element in js.
3. That would go against the rules of hooks. Since `row` is being used in the hook then it should be in the dependency array.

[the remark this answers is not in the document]
hamza0867: Alright! I'm convinced. I removed it from the dependency array

modules/oxide-components/src/main/ts/components/textarea/Textarea.component.tsx
```diff
@@ -0,0 +1,104 @@
+import { Type } from '@ephox/katamari';
+import { forwardRef, useCallback, useLayoutEffect, useRef, useState, type MutableRefObject, type TextareaHTMLAttributes } from 'react';
+
+import { classes } from '../../utils/Styles';
+
+export type Height = {
```
kimwoodfield: Do we need this added complexity? Having both rows and px will mean we need conversion logic. Could we simplify it by using only a single unit of measurement?

modules/oxide-components/src/main/ts/components/textarea/Textarea.component.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+  useLayoutEffect(() => {
+    if (textareaRef.current) {
+      const minRows = computeMinRows();
+      const maxRows = computeMaxRows();
+
+      const textarea = textareaRef.current;
+
+      // Create a hidden clone to measure content
+      const clone = textarea.cloneNode() as HTMLTextAreaElement;
```
kimwoodfield: Are we doing this for every height calculation? Creating and destroying DOM elements is expensive and I feel like we could run the risk of introducing memory leaks. It also feels a bit wrong to be interacting with elements in this way via React.

modules/oxide-components/src/main/ts/components/textarea/Textarea.component.tsx
```diff
@@ -0,0 +1,104 @@
...
+};
+
+const defaultMaxHeight = {
+  unit: 'rows',
+  value: 4
+};
+
+export const Textarea = forwardRef<HTMLTextAreaElement, TextAreaProps>(({ maxHeight = defaultMaxHeight, minHeight = defaultMinHeight, onChange, value, ...rest }, ref) => {
+  const textareaRef: MutableRefObject<HTMLTextAreaElement | null> = useRef(null);
+  const computeMaxRows = useCallback(() => {
```
kimwoodfield: This `computeMaxRows` as well as `computeMinRows` feels a bit complicated. We're also using a few magic numbers like `20` without explaining why.

kimwoodfield:
Thanks for the work on this one @hamza0867! 
While thorough, it feels a bit complex for what we need. We have a simpler implementation in the `suggestededits` plugin that might be worth checking out. The dual unit system, DOM cloning and complex state management feels like overkill for what should be a straightforward TextArea component. Could we look at simplifying this? I'm thinking about long term maintainability and keeping things simpler from the start will help us manage complexity as these components grow. Just my two cents 😅

modules/oxide-components/src/main/ts/components/textarea/Textarea.component.tsx
hamza0867: Good observation. We got rid of the cloning for now.

modules/oxide-components/src/main/ts/components/textarea/Textarea.component.tsx
hamza0867: We do indeed! Using only a single unit of measurement will still require conversion logic. It's just that it would have to be done by the caller instead. Using rows as a unit results in a nicer UI (where the lines are not cut because they fit within the rows) and converting from px to rows might not be intuitive so we let the component handle it to make it easier for the caller.

modules/oxide-components/src/main/ts/components/textarea/Textarea.component.tsx
```diff
@@ -0,0 +1,87 @@
...
+      return Infinity;
+    }
+    if (maxHeight.unit === 'rows') {
+      return Math.max(maxHeight.value, 1);
+    }
+    if (Type.isNonNullable(textareaRef.current)) {
+      const lineHeight = parseInt(window.getComputedStyle(textareaRef.current).lineHeight, 10) || 20;
+      return Math.max(Math.floor(maxHeight.value / lineHeight), 1);
+    }
+    return 4;
```
michalnieruchalski-tiugo: (should) 4 here is a magic number. I think we should define a descriptive variable for this value.
`DEFAULT_MAX_HEIGHT_IN_ROWS` or similar.

modules/oxide-components/src/main/ts/components/textarea/Textarea.component.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+  const computeMinRows = useCallback(() => {
+    if (!minHeight) {
+      return 1;
+    }
+    if (minHeight.unit === 'rows') {
+      return Math.max(minHeight.value, 1);
+    }
+    if (Type.isNonNullable(textareaRef.current)) {
+      const lineHeight = parseInt(window.getComputedStyle(textareaRef.current).lineHeight, 10) || 20;
```
michalnieruchalski-tiugo: `20` seems like another magic number that we should take care of

modules/oxide-components/src/main/ts/components/textarea/Textarea.component.tsx
```diff
@@ -0,0 +1,87 @@
...
+      // Measure the content height
+      const scrollHeight = textarea.scrollHeight;
+      const lineHeight = parseInt(window.getComputedStyle(textarea).lineHeight, 10) || 20;
+      // In mathematical terms: newRows = (scrollHeight / lineHeight) such as minRows <= newRows <= maxRows
+      const newRows = Math.min(Math.max(minRows, Math.floor(scrollHeight / lineHeight)), maxRows);
+      textarea.rows = newRows;
+    }
+  }, [ value, computeMinRows, computeMaxRows ]);
+
+  return <textarea className={classes([ 'tox-textarea' ])} value={value} {...rest} ref={(el) => {
```
michalnieruchalski-tiugo: (question) Do you think we could move `computeMinRows` and `computeMaxRows` out of this file? Like, to introduce a custom hook for that?

I like the idea and functionality of this component, but I would feel overwhelmed when I first open it

modules/oxide-components/src/main/ts/components/textarea/Textarea.stories.tsx
```diff
@@ -0,0 +1,42 @@
...
+  }
+};
+
+export const StartsWith2Rows: Story = {
+  render: () => {
+    const [ value, setValue ] = useState('initial value');
+    const onInput = (event: React.FormEvent<HTMLTextAreaElement>) => {
+      setValue(event.currentTarget.value);
+    };
+    return <Textarea value={value} onInput={onInput} minHeight={{
```
michalnieruchalski-tiugo: (could) Maybe we could default the minHeight and maxHeight to rows? So:
`<Textarea minHeight={2} maxHeight={2}` />
instead of 
`<Textarea minHeight={{ unit: 'rows', value: 2 }} maxHeight={{ unit: 'rows', value: 2 }}` />

It would be a bit easier to use this component. What do you think?

michalnieruchalski-tiugo (approved):
Handy component with autoresize, good work.

modules/oxide-components/src/main/ts/components/textarea/Textarea.component.tsx
hamza0867: Indeed. Good suggestion!

modules/oxide-components/src/main/ts/components/textarea/Textarea.stories.tsx
hamza0867: You are correct, it would be easier but it can be confusing, because people generally use pixels for height, so they might expect the value there to be in pixels. This way it's clear what the unit is.

modules/oxide-components/src/main/ts/components/textarea/Textarea.utils.ts
```diff
@@ -0,0 +1,38 @@
...
+}
+
+const computeMaxRows = ({ maxHeight, textarea }: ComputeMaxRowsProps): number => {
+  if (maxHeight.unit === 'rows') {
+    return Math.max(maxHeight.value, 1);
+  }
+  const lineHeight = parseInt(window.getComputedStyle(textarea).lineHeight, 10);
+  const maxRows = Math.floor(maxHeight.value / lineHeight);
+  return Math.max(maxRows, 1);
+};
```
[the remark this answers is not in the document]
hamza0867: This does not solve the fundamental issue which is relying on an arbitrary default value, like the `16` you used there or the `1.2`

TheSpyder:
@kimwoodfield perhaps somewhere in the middle

[edit] removed incorrect comment

modules/oxide-components/src/main/ts/components/textarea/Textarea.utils.ts
```diff
@@ -0,0 +1,38 @@
...
+  minHeight: Height;
+  textarea: HTMLTextAreaElement;
+}
+const computeMinRows = ({ minHeight, textarea }: ComputeMinRowsProps): number => {
+  if (minHeight.unit === 'rows') {
+    return Math.max(minHeight.value, 1);
+  }
+  const lineHeight = parseInt(window.getComputedStyle(textarea).lineHeight, 10);
+  return Math.max(Math.floor(minHeight.value / lineHeight), 1);
+};
```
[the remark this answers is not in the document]
hamza0867: Indeed. Good observation! I fixed it in the updated code!

modules/oxide-components/src/main/ts/components/textarea/Textarea.utils.ts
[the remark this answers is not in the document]
hamza0867: Now that's smart. I ended up using the option 1!

modules/oxide-components/src/main/ts/components/textarea/Textarea.component.tsx
```diff
@@ -0,0 +1,69 @@
...
+    textareaRef.current = el;
+    if (ref) {
+      if (typeof ref === 'function') {
+        ref(el);
+      } else if (Type.isNonNullable(ref)) {
+        ref.current = el;
+      }
+    }
+  }}
+  />;
```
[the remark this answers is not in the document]
hamza0867: The whole purpose of this component is to be controlled, if someone wants to use an uncontrolled version of this component, they might as well use the standard textarea component. That being said, maybe I should force the user of this component to always pass a value through types ?

modules/oxide-components/src/main/ts/components/textarea/Textarea.utils.ts
```diff
@@ -0,0 +1,63 @@
...
+  singleRowHeight: number;
+}
+
+const computeSingleRowHeight = (textarea: HTMLTextAreaElement): number => {
+  const originalRows = textarea.rows;
+  textarea.rows = 1;
+  const res = textarea.scrollHeight;
+  textarea.rows = originalRows;
+  return res;
+};
```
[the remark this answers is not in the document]
hamza0867: Can the scrollHeight of a textarea be 0 ? A textarea will always have some space and be visible. Right ?

modules/oxide-components/src/main/ts/components/textarea/Textarea.utils.ts
```diff
@@ -0,0 +1,63 @@
...
+
+const computeMaxRows = ({ maxHeight, singleRowHeight }: ComputeMaxRowsProps): number => {
+  if (maxHeight.unit === 'rows') {
+    return Math.max(maxHeight.value, 1);
+  }
+
+  const maxRows = Math.floor(maxHeight.value / singleRowHeight);
+  return Math.max(maxRows, 1);
+};
```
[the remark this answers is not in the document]
hamza0867: The singleRowHeight can never be 0, it is the height of the textarea when it has exactly one row. Don't you agree ?

modules/oxide-components/src/main/ts/components/textarea/Textarea.utils.ts
```diff
@@ -0,0 +1,63 @@
...
+
+const resizeTextarea = ({ minRows, maxRows, singleRowHeight, textarea }: ComputeNewRowsProps): void => {
+  textarea.rows = minRows;
+
+  // Measure the content height
+  const { scrollHeight } = textarea;
+  // In mathematical terms: newRows = (clientHeight / singleRowHeight) such as minRows <= newRows <= maxRows
+  const newRows = Math.min(Math.max(minRows, Math.ceil(scrollHeight / singleRowHeight)), maxRows);
+  textarea.rows = newRows;
+};
```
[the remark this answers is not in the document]
hamza0867: Good observation! I thought about this already. But I think in this case it is the responsibility of the caller to provide sane values.

modules/oxide-components/src/main/ts/components/textarea/Textarea.utils.ts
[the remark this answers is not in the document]
hamza0867: Indeed! I don't expect this Textarea component to ever be used in scenarios where it might initially be hidden.

modules/oxide-components/src/main/ts/components/textarea/Textarea.utils.ts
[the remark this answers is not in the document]
hamza0867: I would. Yes!

modules/oxide-components/src/main/ts/components/textarea/Textarea.component.tsx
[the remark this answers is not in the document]
hamza0867: Done!

modules/oxide-components/src/main/ts/components/textarea/Textarea.component.tsx
```diff
@@ -0,0 +1,70 @@
...
+  rows: _ignoredRows,
+  ...rest }, ref) => {
+  const textareaRef: MutableRefObject<HTMLTextAreaElement | null> = useRef(null);
+
+  // Here we use the useState hook instead of useMemo, because computing the singleRowHeight
+  // requires the textareaRef.current to be available, but useMemo is computed before the render,
+  //  so we set it to 1 initially and update it in the useLayoutEffect
+  const [ singleRowHeight, setSingleRowHeight ] = useState(1);
+
+  useLayoutEffect(() => {
```
kimwoodfield: It looks like we only fire this effect once on mount, but what would happen if the styles for the text inside of it changed, would the textarea be able to respond? I wonder if it would be worth us listening for changes to styles in that dependency array. Maybe it's a change for later but just a thought.

modules/oxide-components/src/main/ts/components/textarea/Textarea.component.tsx
kimwoodfield: Ah right, that makes sense! Handling that conversion will mean potentially avoiding a number of different implementations of it when we go to use it. And good point about the cleaner line breaks when using rows 👍🏻

modules/oxide-components/src/main/ts/components/textarea/Textarea.component.tsx
hamza0867: I did think of that, but I don't have an answer for it yet. Also, I believe this would be a rare case. So I say we'll tackle when we need to.

modules/oxide-components/src/main/ts/components/textarea/Textarea.component.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+  return <textarea {...rest} className={`${classes([ 'tox-textarea' ])} ${className}`} value={value} ref={(el) => {
+    textareaRef.current = el;
+    if (ref) {
+      if (typeof ref === 'function') {
+        ref(el);
+      } else if (Type.isNonNullable(ref)) {
+        ref.current = el;
+      }
+    }
```
TheSpyder: What's this code doing? It doesn't seem specific to textareas. Isn't `ref` always a function? Or is this a sneaky hack that I should consider a code smell?

modules/oxide-components/src/main/ts/components/textarea/Textarea.stories.tsx
TheSpyder: Pixels aren't supported at the moment and for textareas in particular you gave a very good reason to use rows instead of pixels. Unless you see a possibility that we will need pixels, perhaps make it `minRows` and `maxRows` 🤔

modules/oxide-components/vitest.setup.ts
```diff
@@ -0,0 +1,2 @@
+// @ts-expect-error the bundler handles this just fine but tsc is not happy with it
+import '@tinymce/oxide/build/skins/ui/default/skin.css';
```
TheSpyder: isn't there any other way to handle this import? Or adjust it in a way that makes TS happy?

Maybe just make it JS 🤷‍♂️

modules/oxide-components/src/main/ts/components/textarea/Textarea.utils.ts
spocke, line 1: Can we please follow the naming standards we have for these on other modules. So `TextareaUtils.ts` would be the correct naming.

- [x] Follow the naming conventions we currently have

modules/oxide-components/src/main/ts/components/textarea/Textarea.component.tsx
```diff
@@ -0,0 +1,71 @@
...
+  value: 4
+};
+
+export interface TextAreaProps extends TextareaHTMLAttributes<HTMLTextAreaElement> {
+  readonly maxHeight?: Height;
+  readonly minHeight?: Height;
+  readonly value: string;
+}
+
+export const Textarea = forwardRef<HTMLTextAreaElement, TextAreaProps>(({
```
spocke: This needs to be called something else or the auto resizing needs to be optional. If we want this to be a common textarea component then it needs to be able to be static like the ones we currently have in other parts of the editor like dialogs etc.
Maybe this should be called `AutoResizeingTextarea` or something since it's auto resizing to the contents of the textarea.

modules/oxide-components/src/main/ts/components/textarea/Textarea.utils.ts
```diff
@@ -0,0 +1,63 @@
+export interface Height {
+  unit: 'rows' | 'px';
```
spocke: These should be readonly unless they are supposed to be mutable.

modules/oxide-components/src/main/ts/components/textarea/Textarea.utils.ts
```diff
@@ -0,0 +1,63 @@
+export interface Height {
```
spocke: It's a bit weird to have a type defined in a utils module that is public API. Maybe make a new module like `TextareaTypes.ts` or something and stick shared typed into that.

modules/oxide-components/src/main/ts/components/textarea/Textarea.utils.ts
```diff
@@ -0,0 +1,63 @@
...
+  unit: 'rows' | 'px';
+  value: number;
+};
+
+interface ComputeMaxRowsProps {
+  maxHeight: Height;
```
spocke: Readonly on all properties that are not mutable in all interfaces.

modules/oxide-components/src/main/ts/components/textarea/Textarea.component.tsx
kimwoodfield: In React, a forwarded `ref` can be a function or an object. I believe we are just attempting to handle either of those cases. It may also help ensure the compatibility between React and Preact, since this is a React project but will be used in plugins using Preact. Feel free to correct me if I'm wrong @hamza0867

modules/oxide-components/src/main/ts/components/textarea/Textarea.utils.ts
```diff
@@ -0,0 +1,63 @@
...
+  minRows: number;
+  maxRows: number;
+  singleRowHeight: number;
+  textarea: HTMLTextAreaElement;
+}
+
+const resizeTextarea = ({ minRows, maxRows, singleRowHeight, textarea }: ComputeNewRowsProps): void => {
+  textarea.rows = minRows;
+
+  // Measure the content height
```
spocke: Not sure that this comment adds much value this code block does more than just measuring so it's already incorrect. It's better to document things with code. So maybe move the computation out into a function that way it can be atomically tested the move the measurement into it's own function it would look a bit silly with just a `textarea.scrollHeight` but it would then describe to the user what it being measured.
Comments should only be used for edge cases or things that needs explaining, functions forces you to keep the docs up to date since you wouldn't change the code without changing the name but a comment might not go under the same treatment so they are more risky in terms of maintenance.

modules/oxide-components/src/test/ts/browser/components/Textarea.spec.tsx
```diff
@@ -0,0 +1,203 @@
+import { userEvent } from '@vitest/browser/context';
+import { Textarea } from 'oxide-components/components/textarea/Textarea.component';
+import type { Height } from 'oxide-components/components/textarea/Textarea.utils';
+import { classes } from 'oxide-components/utils/Styles';
+import { useState } from 'react';
+import { describe, expect, it } from 'vitest';
+import { render } from 'vitest-browser-react';
+
+describe('Textarea', () => {
+
```
spocke: I know I'm not supposed to nit pick on whitespace things but I wonder why there is so much blank lines between these types of things. I wonder if there is some eslint rule we can enable since usually we don't have this space if you look at other tests etc.

modules/oxide-components/src/main/ts/components/textarea/Textarea.component.tsx
```diff
@@ -0,0 +1,71 @@
...
+  unit: 'rows',
+  value: 1
+};
+
+const defaultMaxHeight: Height = {
+  unit: 'rows',
+  value: 4
+};
+
+export interface TextAreaProps extends TextareaHTMLAttributes<HTMLTextAreaElement> {
```
spocke: Do we want to expose all the properties of a textarea or do we want them to be locked down to the ones that we support. I think it should be locked down, I will open a chat thread on this. We need to decide on what type of API we want to build here.

- [x] Discuss what type of API we want to have a locked down API vs open API

modules/oxide-components/src/main/ts/components/textarea/Textarea.utils.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  const maxRows = Math.floor(maxHeight.value / singleRowHeight);
+  return Math.max(maxRows, 1);
+};
+
+interface ComputeMinRowsProps {
+  minHeight: Height;
+  singleRowHeight: number;
+}
+const computeMinRows = ({ minHeight, singleRowHeight }: ComputeMinRowsProps): number => {
```
spocke: I think this should have separate unit tests.

modules/oxide-components/src/test/ts/browser/components/Textarea.spec.tsx
```diff
@@ -0,0 +1,203 @@
+import { userEvent } from '@vitest/browser/context';
+import { Textarea } from 'oxide-components/components/textarea/Textarea.component';
+import type { Height } from 'oxide-components/components/textarea/Textarea.utils';
+import { classes } from 'oxide-components/utils/Styles';
+import { useState } from 'react';
+import { describe, expect, it } from 'vitest';
+import { render } from 'vitest-browser-react';
+
+describe('Textarea', () => {
```
spocke: Could we name this `browser.TextAreaTest` to follow the conventions we have elsewhere or is that some how not ideal if so why?

modules/oxide-components/src/main/ts/components/textarea/Textarea.utils.ts
kimwoodfield: I agree with @spocke on this one, I also don't think we need the `component` in the filename for `Textarea.component.tsx`. This file already lives under the `textarea` directory. `stories` is fine because that's a convention for Storybook.

modules/oxide-components/src/main/ts/components/textarea/Textarea.component.tsx
hamza0867: Indeed @kimwoodfield.

modules/oxide-components/src/main/ts/components/textarea/Textarea.utils.ts
hamza0867: Done!

modules/oxide-components/vitest.setup.ts
hamza0867: Indeed I just made it JS

modules/oxide-components/src/main/ts/components/autoresizingtextarea/AutoResizingTextarea.stories.tsx
```diff
@@ -0,0 +1,48 @@
...
+export const Example: Story = {
+  args: {
+    value: 'initial value',
+  },
+  render: (args) => {
+    const [ value, setValue ] = useState(args.value);
+    const onInput = (event: React.FormEvent<HTMLTextAreaElement>) => {
+      setValue(event.currentTarget.value);
+    };
+    return <Textarea {...args} value={value} onInput={onInput} />;
```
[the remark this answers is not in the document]
hamza0867: This does typecheck just fine!

modules/oxide-components/src/main/ts/components/autoresizingtextarea/AutoResizingTextarea.tsx
```diff
@@ -0,0 +1,72 @@
...
+  useLayoutEffect(() => {
+    if (textareaRef.current) {
+      resizeTextarea({
+        textarea: textareaRef.current,
+        minRows,
+        maxRows,
+        singleRowHeight
+      });
+    }
+  }, [ value, minRows, maxRows, singleRowHeight ]);
```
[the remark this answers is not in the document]
hamza0867: Good observation! But as of now this is getting out of scope, maybe worth for a follow up ticket. CC: @spocke @TheSpyder

modules/oxide-components/src/main/ts/components/autoresizingtextarea/AutoResizingTextareaUtils.ts
```diff
@@ -0,0 +1,59 @@
...
+};
+
+const computeMaxRows = ({ maxHeight, singleRowHeight }: ComputeMaxRowsProps): number => {
+  if (maxHeight.unit === 'rows') {
+    return Math.max(maxHeight.value, 1);
+  }
+
+  const maxRows = Math.floor(maxHeight.value / singleRowHeight);
+  return Math.max(maxRows, 1);
+};
```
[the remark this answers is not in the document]
hamza0867: Indeed but this is a bit overboard defensive programming I think. Providing the correct value in this case is the responsibility of the caller

modules/oxide-components/src/main/ts/components/autoresizingtextarea/AutoResizingTextareaUtils.ts
```diff
@@ -0,0 +1,59 @@
...
+  readonly singleRowHeight: number;
+}
+const computeMinRows = ({ minHeight, singleRowHeight }: ComputeMinRowsProps): number => {
+  if (minHeight.unit === 'rows') {
+    return Math.max(minHeight.value, 1);
+  }
+
+  const minRows = Math.ceil(minHeight.value / singleRowHeight);
+  return Math.max(minRows, 1);
+};
```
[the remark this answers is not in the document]
hamza0867: same as the previous comment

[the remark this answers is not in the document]
hamza0867: I meant same as [this comment](https://app.graphite.dev/github/pr/tinymce/tinymce/10539/TINY-12555-Create-Textarea-component#comment-PRRC_kwDOAAciD86Ic1o0)

modules/oxide-components/src/main/ts/components/iconbutton/IconButton.tsx
```diff
@@ -1,6 +1,6 @@
 import { Icon, type IconProps } from '../../internal/icon/Icon.component';
 import { classes } from '../../utils/Styles';
-import { Button, type ButtonProps } from '../button/Button.component';
+import { Button, type ButtonProps } from '../button/Button';
```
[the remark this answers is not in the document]
hamza0867: Indeed! Fixed now

modules/oxide-components/src/main/ts/components/textarea/Textarea.utils.ts
hamza0867: I created a follow up ticket for this: https://ephocks.atlassian.net/browse/TINY-12773

modules/oxide-components/src/main/ts/components/autoresizingtextarea/AutoResizingTextareaUtils.ts
```diff
@@ -0,0 +1,59 @@
...
+  readonly singleRowHeight: number;
+}
+
+const computeSingleRowHeight = (textarea: HTMLTextAreaElement): number => {
+  const originalRows = textarea.rows;
+  textarea.rows = 1;
+  const res = textarea.scrollHeight;
+  textarea.rows = originalRows;
+  return res;
+};
```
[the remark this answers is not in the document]
hamza0867: Good suggestion.  I implemented it in the new version. As for the selection, this function would be called when the textarea component first mounts, so there would not be time for the user to set selection before this function is called.

modules/oxide-components/src/main/ts/components/textarea/Textarea.component.tsx
hamza0867: Discussion happened and we agreed on having a locked down api

modules/oxide-components/src/main/ts/components/textarea/Textarea.component.tsx
TheSpyder: But having it here implies we need to do this for every component? That seems like a weird restriction.

I'm starting to question whether we should try to bridge between react and preact if there are these sorts of pitfalls people can fall into.

hamza0867: It's just that the ref can be either a function or an object (both in react and preact, they're not different in this regard) so we have to handle both cases

TheSpyder: I just struggle to believe that this is a normal thing to do in react components 🤔

PR #10542, opened 2025-08-18
TINY-12529: Added a LESS variable mapping for secondary button hover text color customization

kimwoodfield:
Related Ticket: [TINY-12529](https://ephocks.atlassian.net/browse/TINY-12529)

Description of Changes:
* Added a mapping variable to make it easier to customize hover text colors in secondary buttons. 
* Before this, we had to either change all text colors or break the pattern in our skins where we only use variables. 
* Now, skins can just override one variable to change the hover text color for secondary buttons.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

[TINY-12529]: https://ephocks.atlassian.net/browse/TINY-12529?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

- Refactor
  - Improved internal theming for the secondary outlined button’s hover text color by introducing an intermediate mapping, making overrides and customization across themes easier.
  - Maintains current visuals and behavior; this change enhances maintainability and consistency of design tokens with no impact on end-users.

### DIFF
--- a/modules/oxide/src/less/theme/components/button/button-secondary-outlined.less
+++ b/modules/oxide/src/less/theme/components/button/button-secondary-outlined.less
@@ -4,7 +4,7 @@
 // For base button styles see button.less
 //
 
-@button-secondary-outline-hover-text-color-mapping: var(--tox-private-button-secondary-outline-hover-text-color, @text-color);
+@button-secondary-outline-hover-text-color: @text-color;
 
 @layer tox-atomic-component {
   .tox {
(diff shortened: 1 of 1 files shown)


modules/oxide/src/less/theme/components/button/button-secondary-outlined.less
```diff
@@ -59,7 +61,7 @@
       &:hover:not(:disabled) {
         background-color: var(--tox-private-button-secondary-outline-hover-background-color, @button-secondary-background-color);
         border-color: var(--tox-private-button-secondary-outline-hover-border-color, @button-secondary-hover-background-color);
-        color: var(--tox-private-button-secondary-outline-hover-text-color, @text-color);
+        color: @button-secondary-outline-hover-text-color-mapping;
```
TheSpyder: I thought during the oxide project the `var()` declarations had to go inside `when (@custom-properties-enabled = true)`. Or am I misunderstanding something? We should check with @ztomaszyk 🤔

zuzanna-antas: The `var()` can be used outside `when (@custom-properties-enabled = true)` block but only with a fallback to the less variable or fixed value. However, all the CSS Custom Properties need to be declared inside the guarded block.
So when the feature flag is not enabled Custom Properties are not defined and the fallback value is used instead.

modules/oxide/src/less/theme/components/button/button-secondary-outlined.less
```diff
@@ -4,6 +4,8 @@
 // For base button styles see button.less
 //
 
+@button-secondary-outline-hover-text-color-mapping: var(--tox-private-button-secondary-outline-hover-text-color, @text-color);
```
zuzanna-antas: I would avoid mixing CSS Custom Properties and LESS variables and simply do: 

```suggestion
@button-secondary-outline-hover-text-color: @text-color;
```

then for the style declaration instead of the fallback to the `@text-color` we can have a fallback to this new LESS variable.

modules/oxide/src/less/theme/components/button/button-secondary-outlined.less
zuzanna-antas:
```suggestion
        color: var(--tox-private-button-secondary-outline-hover-text-color, @button-secondary-outline-hover-text-color);
```

ztomaszyk (approved):

PR #10546, opened 2025-08-20
TINY-12744: Update TypeScript and Terser to target ES2022 

kimwoodfield:
Related Ticket: [TINY-12744](https://ephocks.atlassian.net/browse/TINY-12744)

Description of Changes:
* TypeScript will now output modern JavaScript syntax (or at least from `es2022`) in the compiled code instead of transforming it down to older syntax, so we are maintaining broad browser support while taking advantage of ES2022 features.
* This keeps our configuration consistent since we were already using ES2022 features during development via `lib`, now our output matches that capability.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

[TINY-12744]: https://ephocks.atlassian.net/browse/TINY-12744?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

- Chores
  - Updated JavaScript compilation and minification targets to ES2022 across the app, aligning build outputs.
  - Impact: Smaller, more efficient bundles and improved performance on modern browsers; older browser support may be reduced.

- Tests
  - Removed the silver (phantomjs/browser/webdriver) test configuration from automated test settings.

### DIFF
--- a/tsconfig.shared.json
+++ b/tsconfig.shared.json
@@ -1,8 +1,8 @@
 {
   "compilerOptions": {
     "moduleResolution": "node",
-    "target": "es2018",
-    "module": "es2015",
+    "target": "es2022",
+    "module": "es2022",
     "lib": ["es2022", "dom"],
     "importHelpers": true,
     "declaration": true,
--- a/modules/tinymce/Gruntfile.js
+++ b/modules/tinymce/Gruntfile.js
@@ -199,7 +199,7 @@ module.exports = function (grunt) {
     terser: Object.assign(
       {
         options: {
-          ecma: 2018,
+          ecma: 2022,
           output: {
             ascii_only: true
           },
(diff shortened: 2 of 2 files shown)


spocke (approved):

ltrouton (approved):

tsconfig.shared.json
```diff
@@ -1,8 +1,8 @@
 {
   "compilerOptions": {
     "moduleResolution": "node",
-    "target": "es2018",
-    "module": "es2015",
+    "target": "es2022",
+    "module": "es2022",
```
ArvinJ-H, line 5: can we have `esnext` as the module?

kimwoodfield: I didn't experiment with esnext, so theoretically it's possible. However, the "module" setting controls how our imports/exports are emitted and updating to esnext would mean opting into the latest and possibly experimental JS features. It's likely not all browsers would support those features and opting for es2022 which has reported support and compatibility with modern browsers feels like the safer bet.

TheSpyder: We bundle, so we could do anything we want in theory, but for safety I'd rather keep these values to what is compatible with our supported browsers.

Issue #9910, opened 2024-10-08
[loadScript] should use the root `nonce` attribute to load rest of the scripts


felixmosh:
## 📝 Provide a description of the new feature or improvement
Given a `nonce` to the `tinymce.min.js`, the inner `loadScript` function should use it for rest of the loaded script, as it does for the "base url".

### 🫶 What is the motivation?
To improve a CSP compatability, user can specify a nonce which will be used to load rest of tinymce scripts.
This is the way that Google Maps does it, if you specify a nonce to the entry point, it gets used to load rest of the files.

### 🔗 What is the consequence of not having this feature?
Without it, it is much dificult to make tinymce CSP compatible without allowinf the entire cdn js domain, which exposes the user to potential CSP bypasses.

### 🚦 How important would you rate the requested feature or improvement?
*Choose from nice-to-have, important or critical.*
important
---

felixmosh:
Ping

PR #6137, opened 2020-10-24
Add the possibility to add a nonce attribute for script and css loading.

rotdrop:
Related Ticket:

Description of Changes:
* Add an optional "nonce" option to the the script- and stylesheet-loaders in order to allow the use in
  conjunction with CSP policies using the "nonce" feature.

Pre-checks:
* [ ] Changelog entry added
* [ ] Tests have been added (if applicable)
* [x] Branch prefixed with `feature/` for new features (if applicable)
* [ ] License headers added on new files (if applicable)

Review:
* [ ] Milestone set
* [ ] Review comments resolved

GitHub issues (if applicable):

Issue #1901, opened 2014-09-15
Content Security Policy Violation (Compatibility Issue)


TinyMCE-User:
Description of problem: If using a strict Content-Security-Policy, TinyMCE does not work and creates an Javascript error.

Steps to reproduce:
1. The HTTP-header of the page must include CSP (X-Content-Security-Policy: allow 'self')
2. Apply TinyMCE to a textarea in the same page
3. A CSP violation is triggered (visible in debug console of Google Chrome)

Expected result: TinyMCE works

Actual result: Line 1274: return a = h(t, "string") ? o.doc.createElement(t) : t, o.setAttribs(a, n), r && (r.nodeType ? a.appendChild(r) : o.setHTML(a, r)), i ? a : e.appendChild(a)

Refused to execute Javascript URL because it violates the Content Security Policy directive script-src:'self'

> _**Legacy information imported from TinyMCE bug tracker:**_
> 
> _[#T7172](http://archive.tinymce.com/develop/bugtracker_view.php?id=7172) posted by [Blackbam](http://archive.tinymce.com/profile.php?name=Blackbam)_
> 
> _**Tags:** [csp]_
> _**Status:** Open_
> _**Resolution:** None_
> _**Attached URL:** none_

parmarg:
Any update on this?

spocke:
I added this to our internal backlog will see if we can schedule this in for the next sprint it's been in here for way to long.

spocke:
Had a go at this and it seems that this works:
Content-Security-Policy: default-src 'none'; script-src 'self' 'unsafe-inline'; connect-src 'self'; img-src 'self'; style-src 'self'; font-src 'self';

We need to be able to set styles inline since we use it to place dialogs, menus etc absolute.

maarten-v:
Is it also really necessary to use inline javascript? It would be great if we could get rid of the insecure "script-src unsafe-inline" setting.

spocke:
I pushed a change where it would not set the inline source to javascript: if the browser has CSP support.
So after that patch a CSP header like this would work:
Content-Security-Policy: default-src 'none'; script-src 'self'; connect-src 'self'; img-src 'self'; style-src 'self' 'unsafe-inline'; font-src 'self';

The UI requires us to position things at absolute positions with code so unsafe-inline for styles is still  required.

Daijobou:
Is this in 4.4.3 included? Actual I get still

[...]

PR #10550, opened 2025-08-21
Add support for a CSP nonce to script and style loading.

jeroen1602:
This adds an option to add a nonce to the places where a script/ link/ style tag is created.

The nonce is mostly needed to support loading styles using `mceDefaultStyles`, in `DOMUtils.addStyle`. As well as `InitContentBody.appendStyle`, and `StyleSheetLoader.loadRawCss`. But if for whatever reason someone wants to also load scripts and styles from external resources using a nonce instead of allowed hosts, then this is now also possible.

In the editor config you can decide to set a global nonce or a specific nonce for script loading and another one for style loading.

This allows a user to remove `'unsafe-inline'` from the `style-src` CSP policy and only keep it for the `style-src-attr` policy.

P.S. Wasn't completely sure if this would be appreciated so I haven't put the time in to create tests.

Related Ticket: 

Description of Changes:
* Placeholder text

Pre-checks:
* [X] Changelog entry added

Review:
* [ ] Milestone set

 - https://github.com/tinymce/tinymce/issues/9910
 - https://github.com/tinymce/tinymce/issues/1901
 - https://github.com/tinymce/tinymce/pull/6137

## Summary by CodeRabbit

- New Features
  - Added a csp_nonce editor option to support CSP nonces (single string or separate script/style values). Nonces are applied to dynamically loaded scripts, stylesheets, and inline/default styles to improve Content Security Policy compatibility.

- Documentation
  - Updated changelog with an Unreleased entry describing the new nonce option.


TheSpyder:
@coderabbitai ignore

TheSpyder:
@jeroen1602 This is a very interesting idea. I can't guarantee when we'll respond to it but please know that I've added it to our discussions around security best practices.

jeroen1602:
I did see in the spec that a nonce (should) only apply to inline `script` and `style` tags. So I should probably change it on places where a src is used and with the link tag that loads css.

https://www.w3.org/TR/CSP3/#match-element-to-source-list

PR #10556, opened 2025-08-26
TINY-12315: Fix expandable accordions in disabled editor

michalnieruchalski-tiugo:
Related Ticket: TINY-12315

Description of Changes:
* Accordions are now not toggable in disabled editor

Pre-checks:
* [x] Changelog entry added
* [x] Tests have been added (if applicable)
* [x] Branch prefixed with `feature/`, `hotfix/` or `spike/`

Review:
* [x] Milestone set
~~* [ ] Docs ticket created (if applicable)~~

GitHub issues (if applicable):

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented accordion sections from expanding or collapsing when the editor is disabled (TINY-12315); summary clicks now preserve disabled state.

* **Behavior**
  * Bookmark/link navigation from inside summaries still works in disabled mode without toggling accordions.
  * Modifier-key link clicks open links in a new tab/window as expected from summaries.

* **Tests**
  * Added regression browser tests verifying summary clicks, nested elements, and in-summary links do not toggle accordions when disabled.

* **Chores**
  * Added an unreleased changelog entry for TINY-12315.

PR #10563, opened 2025-08-27
TINY-12316: make toggle accordion button enabled in readonly mode

michalnieruchalski-tiugo:
Related Ticket: TINY-12316

Description of Changes:
* make toggle accordion button enabled in readonly mode
* restore initial content when entering design mode
* return initial content when getContent is used in readonly mode

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

## Summary by CodeRabbit

- Improvements
  - Accordions can now be toggled in readonly mode; open/closed state is preserved when switching between readonly and design. Toggle button enabled in readonly/design and disabled when editor is disabled. Toggling in readonly mode no longer creates undo levels; design mode still does.

- Tests
  - Added comprehensive browser tests covering readonly/design/disabled states, toolbar availability, toggle behavior, state preservation, and undo semantics.

- Documentation
  - Added unreleased changelog entries describing readonly accordion behavior.

### DIFF
--- a/modules/tinymce/src/core/main/ts/undo/Levels.ts
+++ b/modules/tinymce/src/core/main/ts/undo/Levels.ts
@@ -67,27 +67,32 @@ const getLevelContent = (level: NewUndoLevel): string => {
   return level.type === 'fragmented' ? level.fragments.join('') : level.content;
 };
 
-const getCleanLevelContent = (level: NewUndoLevel): string => {
+const getCleanLevelContent = (isReadonly: boolean, level: NewUndoLevel): string => {
   const elm = SugarElement.fromTag('body', lazyTempDocument());
   Html.set(elm, getLevelContent(level));
   Arr.each(SelectorFilter.descendants(elm, '*[data-mce-bogus]'), Remove.unwrap);
+
+  if (isReadonly) {
+    Arr.each(SelectorFilter.descendants(elm, 'details[open]'), Remove.unwrap);
+  }
+
   return Html.get(elm);
 };
 
 const hasEqualContent = (level1: NewUndoLevel, level2: NewUndoLevel): boolean =>
   getLevelContent(level1) === getLevelContent(level2);
 
-const hasEqualCleanedContent = (level1: NewUndoLevel, level2: NewUndoLevel): boolean =>
-  getCleanLevelContent(level1) === getCleanLevelContent(level2);
+const hasEqualCleanedContent = (isReadonly: boolean, level1: NewUndoLevel, level2: NewUndoLevel): boolean =>
+  getCleanLevelContent(isReadonly, level1) === getCleanLevelContent(isReadonly, level2);
 
 // Most of the time the contents is equal so it's faster to first check that using strings then fallback to a cleaned dom comparison
-const isEq = (level1: NewUndoLevel | undefined, level2: NewUndoLevel | undefined): boolean => {
+const isEq = (isReadonly: boolean, level1: NewUndoLevel | undefined, level2: NewUndoLevel | undefined): boolean => {
   if (!level1 || !level2) {
     return false;
   } else if (hasEqualContent(level1, level2)) {
     return true;
   } else {
-    return hasEqualCleanedContent(level1, level2);
+    return hasEqualCleanedContent(isReadonly, level1, level2);
   }
 };
 
--- a/modules/tinymce/src/core/main/ts/undo/Operations.ts
+++ b/modules/tinymce/src/core/main/ts/undo/Operations.ts
@@ -38,7 +38,7 @@ export const addUndoLevel = (
   }
 
   // Add undo level if needed
-  if (lastLevel && Levels.isEq(lastLevel, newLevel)) {
+  if (lastLevel && Levels.isEq(editor.readonly, lastLevel, newLevel)) {
     return null;
   }
 
@@ -147,7 +147,7 @@ export const reset = (undoManager: UndoManager): void => {
 
 export const hasUndo = (editor: Editor, undoManager: UndoManager, index: Index): boolean =>
   // Has undo levels or typing and content isn't the same as the initial level
-  index.get() > 0 || (undoManager.typing && undoManager.data[0] && !Levels.isEq(Levels.createFromEditor(editor), undoManager.data[0]));
+  index.get() > 0 || (undoManager.typing && undoManager.data[0] && !Levels.isEq(editor.readonly, Levels.createFromEditor(editor), undoManager.data[0]));
 
 export const hasRedo = (undoManager: UndoManager, index: Index): boolean =>
   index.get() < undoManager.data.length - 1 && !undoManager.typing;
--- a/modules/tinymce/src/core/main/ts/undo/Setup.ts
+++ b/modules/tinymce/src/core/main/ts/undo/Setup.ts
@@ -81,7 +81,7 @@ export const registerEvents = (editor: Editor, undoManager: UndoManager, locks: 
     }
 
     // Fire a TypingUndo event on the first character entered
-    if (isFirstTypedCharacter.get() && undoManager.typing && !Levels.isEq(Levels.createFromEditor(editor), undoManager.data[0])) {
+    if (isFirstTypedCharacter.get() && undoManager.typing && !Levels.isEq(editor.readonly, Levels.createFromEditor(editor), undoManager.data[0])) {
       if (!editor.isDirty()) {
         editor.setDirty(true);
       }
--- a/modules/tinymce/src/core/test/ts/browser/ReadOnlyModeTest.ts
+++ b/modules/tinymce/src/core/test/ts/browser/ReadOnlyModeTest.ts
@@ -1,4 +1,4 @@
-import { ApproxStructure, Mouse, UiFinder, Clipboard } from '@ephox/agar';
+import { ApproxStructure, Clipboard, Mouse, UiFinder } from '@ephox/agar';
 import { describe, it } from '@ephox/bedrock-client';
 import { PlatformDetection } from '@ephox/sand';
 import { Attribute, Class, Css, Scroll, SelectorFind, SugarBody, Traverse } from '@ephox/sugar';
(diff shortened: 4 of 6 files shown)


michalnieruchalski-tiugo:
* **#10563** <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10563?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10563?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10615** <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10615?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10591** <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10591?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10556** <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10556?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

modules/tinymce/src/plugins/accordion/main/ts/core/readonly/Readonly.ts
```diff
@@ -0,0 +1,57 @@
...
+import type { EditorEvent } from 'tinymce/core/api/util/EventDispatcher';
+
+import * as AstNodeAttributeUtils from './attribute/AstNodeAttributeUtils';
+import * as SugarAttributeUtils from './attribute/SugarAttributeUtils';
+
+const setup = (editor: Editor): void => {
+  // Prevent adding an undo level on ToggleAccordion in readonly mode
+  editor.on('BeforeAddUndo', (event) => {
+    if (editor.readonly) {
+      const originalEvent = event.originalEvent as unknown as EditorEvent<unknown> | undefined;
```
michalnieruchalski-tiugo: I'm aware that we have to do something about this ugly type workaround :D

```
editor.on('BeforeAddUndo', (event) => {
Here the event is: EditorEvent<AddUndoEvent>
Which is:

event: {
    originalEvent: Event | undefined;
    level: UndoLevel;
    lastLevel: UndoLevel | undefined;
} & {
    readonly type: string;
    readonly target: any;
    readonly isDefaultPrevented: () => boolean;
    readonly preventDefault: () => void;
    readonly isPropagationStopped: () => boolean;
    readonly stopPropagation: () => void;
    readonly isImmediatePropagationStopped: () => boolean;
    readonly stopImmediatePropagation: () => void;
}
```

`originalEvent` is suppose to represent an event that caused adding the undo level. In my opinion it has the wrong type and instead of `Event` it should be `EditorEvent<unknown>` .

I think the root cause of this mistake is in the `UndoManagerTypes.ts` where we have wrongly defined event as `EditorEvent<any>`:

https://github.com/tinymce/tinymce/blob/main/modules/tinymce/src/core/main/ts/undo/UndoManagerTypes.ts#L32

Because of how the `EditorEvent` is defined `EditorEvent<any>` is the same as `any`

Then we've made another mistake while implementing the UndoManager interface in `UndoManager.ts` 

https://github.com/tinymce/tinymce/blob/main/modules/tinymce/src/core/main/ts/api/UndoManager.ts#L52

we've defined type of event as `Event` and typescript didn't complain (because `EditorEvent<any>` is bascially `any`)

Now, why am I writing this instead of just fixing it 😅 
Because to change that would be a breaking change. Currently all of those are valid from typescript point of view:

```
editor.undoManager.add(undefined, 'some string');
editor.undoManager.add(undefined, 123);
editor.undoManager.add(undefined, { });
```

If I were to change `EditorEvent<any>` to `EditorEvent<unknown>` typescript would raise the error in all of those examples.

Let me know what do you think, and how can we possibly fix it / schedule to fix it.

michalnieruchalski-tiugo:
@coderabbitai review

modules/tinymce/src/plugins/accordion/main/ts/core/readonly/Readonly.ts
TheSpyder: What would those invalid `undoManager.add` calls result in? If the answer is an error, or an invalid state, then the type change is a bug fix not a breaking change.

modules/tinymce/src/plugins/accordion/main/ts/ui/Buttons.ts
```diff
@@ -18,7 +18,8 @@ const register = (editor: Editor): void => {
   editor.ui.registry.addToggleButton('accordiontoggle', {
     icon: 'accordion-toggle',
     tooltip: 'Toggle accordion',
-    onAction: () => editor.execCommand('ToggleAccordion')
+    onAction: () => editor.execCommand('ToggleAccordion'),
+    context: 'any'
```
TheSpyder, line 22: What's this doing? how is adding `any` different from having no context?

TheSpyder (approved):

modules/tinymce/src/plugins/accordion/main/ts/ui/Buttons.ts
michalnieruchalski-tiugo: Without context specified `accordiontoggle` button would be disabled in readonly mode.
With `context: any` it's always enabled.

modules/tinymce/src/plugins/accordion/main/ts/core/readonly/Readonly.ts
spocke: I'm not surprised that the events are wrong here since we added types to the code after it was created. Even if the API calls with invalid input like that works then it's against what we documented in API docs so I think it would still not be a breaking change in that sense.
It might also be that we typed this before there even was a `unknown` type it was introduced 2018 I think we added TS before that. But either way we should fix any type issues that would make this more fragile.

modules/tinymce/src/plugins/accordion/main/ts/core/readonly/Readonly.ts
```diff
@@ -0,0 +1,57 @@
...
+      const shouldStopEvent = isBlur(originalEvent) || (isExecCommand(originalEvent) && originalEvent.command === 'ToggleAccordion');
```
spocke: Can this expression be simplified maybe add a isToggleAccordionEvent predicate that does the second part and call that.

spocke (approved):

modules/tinymce/src/plugins/accordion/main/ts/core/readonly/Readonly.ts
michalnieruchalski-tiugo: I've created a separate PR to address this issue

modules/tinymce/src/plugins/accordion/main/ts/core/readonly/Readonly.ts
```diff
@@ -0,0 +1,60 @@
...
+import * as AstNodeAttributeUtils from './attribute/AstNodeAttributeUtils';
+import * as SugarAttributeUtils from './attribute/SugarAttributeUtils';
+
+const setup = (editor: Editor): void => {
+  // Prevent adding an undo level on ToggleAccordion in readonly mode
+  editor.on('BeforeAddUndo', (event) => {
+    if (
+      editor.readonly
+      && event.originalEvent
+      && (isBlurEvent(event.originalEvent) || isToggleAccordionEvent(event.originalEvent))
```
michalnieruchalski-tiugo: (question)
Initially it was only `ToggleAccordion` event.

Then I've noticed that in this scenario:
1. Open editor in readonly mode
2. Toggle accordion
3. Move focus outside of the editor (blur)

undo level is added, triggered by `blur` event. So I've added `isBlurEvent` check. But thinking about it:

Shouldn't we just ignore all the undo levels in readonly mode? 
Should this if statement be:
```
if (editor.readonly) {
    event.preventDefault();
}
```
(and possibly move it out of the accordion plugin) 
@spocke @TheSpyder

TheSpyder: Yes I did wonder why the `blur` fix was in the accordion plugin. Blocking undo would make sense in readonly (and disabled) modes.

TheSpyder (approved):

modules/tinymce/src/plugins/accordion/main/ts/core/readonly/Readonly.ts
spocke: Yeah, lets change it. The whole purpose of readonly is that it's readonly right. I guess there are many commands that still apply in readonly that we are currently not blocking. But no client could argue that undo levels should be produced in readonly mode that makes no sense.

And if you do a dom mutation while in readonly then the undo manager would pick it up when you leave readonly mode so one needs to be careful to make sure that no state change happened in the HTML when exiting readonly mode.

spocke: But I think this change might need it's own changelog since it's a behavioural change something about undo levels are no longer created while in readonly mode.

michalnieruchalski-tiugo: - [x] I'll add automated test to check leaving the readonly mode scenario.

If this is the direction we're going to, then this change shouldn't be in accordion plugin. It should be directly in `addUndoLevel` method.

modules/tinymce/src/plugins/accordion/main/ts/core/readonly/attribute/AttributeUtils.ts
```diff
@@ -0,0 +1,33 @@
...
+};
+
+const addTemporaryAttributes = <T>(adapter: Pick<AttributeOperationAdapter<T>, 'setAttribute' | 'hasAttribute'>, detailsElement: T): void => {
+  const { setAttribute, hasAttribute } = adapter;
+  setAttribute(detailsElement, accordionTemporaryOpenAttribute, hasAttribute(detailsElement, 'open') ? 'true' : 'false');
+};
+
+const restoreNormalState = <T>(adapter: AttributeOperationAdapter<T>, detailsElement: T): void => {
+  const { setAttribute, hasAttribute, getAttribute, removeAttribute } = adapter;
+  // At this point every <details> should have data-mce-open attribute. But I will ignore those that don't - just in case.
```
spocke:
```suggestion
  // At this point every <details> should have data-mce-open attribute. But we should ignore those that don't - just in case.
```

I think it's better to use "we" as the code is maintained by multiple people.

modules/tinymce/src/plugins/accordion/main/ts/core/readonly/attribute/AttributeUtils.ts
```diff
@@ -0,0 +1,33 @@
+import { accordionTemporaryOpenAttribute } from '../../Identifiers';
+
+export interface AttributeOperationAdapter<T> {
```
spocke, line 3: I see that you try to use the same code but for the ast and the regular dom. But I wonder if this abstraction is the same size as just duplicating the code for toggling those attributes?
Does this add extra complexity with little gain? If this module was heavy then I would get that it would give us more but right now it's just a couple of lines so to me it mostly adds complexity.
Or do you think this part of the code will grow so that it would make sense to have an abstraction for the two tree models?

PR #10569, opened 2025-08-28
TINY-12804: Added options for the parser previously deemed not needed.

HAFRMO:
Related Ticket: TINY-12804

Description of Changes:
Adding it by way of the args was previously not thought needed, but it was determined that it was due to some changes in approach.

Pre-checks:
* ~~[x] Changelog entry added~~
* ~~[x] Tests have been added (if applicable)~~

Review:
* [x] Milestone set

## Summary by CodeRabbit

* **New Features**
  * Added an opt-in setting to explicitly enforce full-document HTML parsing when a document context is specified. This improves consistency between fragment and full-document parsing while remaining backward compatible and requiring no changes to existing integrations.

### DIFF
--- a/modules/tinymce/src/core/main/ts/api/html/DomParser.ts
+++ b/modules/tinymce/src/core/main/ts/api/html/DomParser.ts
@@ -39,7 +39,6 @@ const makeMap = Tools.makeMap, extend = Tools.extend;
 export interface ParserArgs {
   getInner?: boolean | number;
   forced_root_block?: boolean | string;
-  root_name?: string;
   context?: string;
   isRootContent?: boolean;
   format?: string;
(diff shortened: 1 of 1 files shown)


modules/tinymce/src/core/main/ts/api/html/DomParser.ts
```diff
@@ -479,7 +480,7 @@ const DomParser = (settings: DomParserSettings = {}, schema = Schema()): DomPars
    */
   const parse = (html: string, args: ParserArgs = {}): AstNode => {
     const validate = defaultedSettings.validate;
-    const preferFullDocument = defaultedSettings.root_name === '#document';
+    const preferFullDocument = (args.root_name ?? defaultedSettings.root_name) === '#document';
```
spocke: We don't need a new option you should be able to check if `args.context` is set to `#document` then you do a full page parse. That was what the old parser did.

So I guess this:
```suggestion
    const preferFullDocument = (args.context ?? defaultedSettings.root_name) === '#document';
```

- [x] Use context instead of a new option for ParserArgs

spocke (approved):

modules/tinymce/src/core/main/ts/api/html/DomParser.ts
HAFRMO: I was concerned about this creating unintended side-effects since `context` is used in different places than `root_name`, it seems to change the validation steps.

spocke: The only thing we could mess up would be this plugin so I hope tests and QA will make sure it works correctly. It would only do something different if you pass in `#document` as the context and the only place we will do that is in the `fullpagehtml` plugin.

HAFRMO: I suppose that if it turns out to be a problem it won't be that hard to split it out.

PR #10575, opened 2025-09-01
TINY-12682: Upgrade TypeScript to 5.9.2

kimwoodfield:
Related Ticket: [TINY-12682](https://ephocks.atlassian.net/browse/TINY-12682)

Description of Changes:
* Bump `typescript` to `5.9.2`
* Turned off the `@stylistic/indent` eslint rule until a TypeScript issue is resolved which otherwise results in a stack overflow.
* Modified the chunk handling in `Http.ts` to explicitly convert each chunk to a new `Uint8Array` until [this](https://github.com/tinymce/tinymce/pull/10546) change is merged. This change was necessary due to [this](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-9.html#libdts-changes) recent change.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set


modules/oxide-components/eslint.config.ts
```diff
@@ -31,6 +31,20 @@ export default defineConfig(
...
+      rules: {
+        "@stylistic/indent": "off"
+      }
+    },
+    {
+      files: ["**/*.ts"],
+      rules: {
+        "@stylistic/indent": ["error", 2]
+      }
+    },
```
ltrouton: Just wanted to check if "@stylistic/indent" is already set for `.ts` and doesn't need to be included?

kimwoodfield: Yep - I removed that one because we don't need it 👍🏻

modules/oxide-components/eslint.config.ts
```diff
@@ -31,6 +31,14 @@ export default defineConfig(
...
       },
     },
+    // TINY-12801: Temporary workaround until issue with @stylistic/indent rule introduced in TypeScript 5.9.2 is resolved.
+    // See https://github.com/microsoft/TypeScript/issues/62188
+    {
+      files: ["**/*.tsx"],
+      rules: {
+        "@stylistic/indent": "off"
+      }
+    },
```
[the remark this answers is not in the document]
kimwoodfield: Thanks for the suggestion, but it seems the current position of the `.tsx` override is working fine - CI and linting are passing without needing to move it after the `storybookConfig`.

modules/oxide-components/eslint.config.ts
```diff
@@ -31,6 +31,14 @@ export default defineConfig(
         ],
       },
     },
+    // TINY-12801: Temporary workaround until issue with @stylistic/indent rule introduced in TypeScript 5.9.2 is resolved.
+    // See https://github.com/microsoft/TypeScript/issues/62188
+    {
+      files: ["**/*.tsx"],
+      rules: {
+        "@stylistic/indent": "off"
```
TheSpyder: What is 5.9.2 even getting us? I thought this was a critical lint rule and I'd rather avoid upgrading TS than turn it off.

TheSpyder (dismissed):

modules/agar/src/main/ts/ephox/agar/http/MockClient.ts
```diff
@@ -92,7 +92,11 @@ const handleBodyResponse = async (body: ReadableStream<Uint8Array>, port: Messag
           }
 
           if (!done) {
-            const buffer = chunk.buffer.slice(chunk.byteOffset, chunk.byteOffset + chunk.byteLength);
+            const sliced = chunk.buffer.slice(chunk.byteOffset, chunk.byteOffset + chunk.byteLength);
+            const buffer = sliced instanceof SharedArrayBuffer
+              ? new ArrayBuffer(sliced.byteLength).constructor.prototype.slice.call(sliced, 0)
+              : sliced;
```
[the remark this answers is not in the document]
kimwoodfield: I have already addressed this with a more robust solution that also handles the SharedArrayBuffer conversion properly and checks environment compatibility.

modules/oxide-components/eslint.config.ts
kimwoodfield: TypeScript [reverted the change](https://github.com/microsoft/TypeScript/pull/62425) which led to us needing to add this in 5.9.3 so I've removed this now

modules/agar/src/main/ts/ephox/agar/http/MockClient.ts
```diff
@@ -92,7 +92,17 @@ const handleBodyResponse = async (body: ReadableStream<Uint8Array>, port: Messag
...
+            const sliced = chunk.buffer.slice(chunk.byteOffset, chunk.byteOffset + chunk.byteLength);
+            const buffer = Optional
+              .from(typeof SharedArrayBuffer !== 'undefined' ? SharedArrayBuffer : null)
+              .filter(() => sliced instanceof SharedArrayBuffer)
+              .map(() => {
+                const newBuffer = new ArrayBuffer(sliced.byteLength);
+                new Uint8Array(newBuffer).set(new Uint8Array(sliced));
+                return newBuffer;
+              })
+              .getOr(sliced as ArrayBuffer);
```
TheSpyder: This is not really what Optionals were designed for. A simple if/else would likely suffice.

Looking at the description, the other PR that you reference as the reason for this was merged so is it even still necessary?

kimwoodfield: I've replaced the Optional approach with a simpler if/else approach. 

Our work to bump our target/module to ES2022 introduced the SharedArrayBuffer types as part of [SharedMemory in ES2017](https://github.com/microsoft/TypeScript/blob/main/src/lib/es2017.sharedmemory.d.ts), and TypeScript 5.9.3's type checking is now properly enforcing that type incompatibility that was always there but previously overlooked. The [MockedResponseBodyChunkMessage](https://github.com/tinymce/tinymce/blob/07e5e210e7d68a16d46ea94af56e2a46f47b6b18/modules/agar/src/main/ts/ephox/agar/http/Shared.ts#L45) always expects `buffer` to be of type `ArrayBuffer` so the conversion is necessary to make sure the behaviour remains the same as it was before even if SharedArrayBuffer is available.

package.json
```diff
@@ -86,11 +86,11 @@
...
     "stylelint-order": "^6.0.4",
     "terser": "^5.28.1",
-    "ts-loader": "^9.5.2",
+    "ts-loader": "^9.5.4",
     "tsconfig-paths-webpack-plugin": "^4.2.0",
     "tslib": "^2.8.1",
     "twemoji": "^14.0.2",
-    "typescript": "^5.7.3",
+    "typescript": "^5.9.2",
     "webpack": "^5.74.0",
```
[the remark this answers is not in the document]
kimwoodfield: All of these ranges are satisfied by TypeScript 5.9.3, which is what Yarn has hoisted (we can confirm that via `yarn why typescript`). The monorepo already has a unified TypeScript version. Since `lerna` is an external npm package, we cannot modify its dependencies, nor do we need to - the resolution is working correctly.

PR #10578, opened 2025-09-02
TINY-12584: add styles for printing 

ztomaszyk:
Related Ticket: TINY-12584

Description of Changes:
* Added print-specific CSS styles to hide UI elements when printing content
* Hides placeholder text, bogus elements, caret, selection outlines, and focus indicators during printing

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

## Summary by CodeRabbit

* **Bug Fixes**
  * Printouts are now clean and content-only. Editor artifacts such as placeholders, bogus element indicators, the visual caret, and selection/focus outlines are hidden when printing. This ensures printed documents mirror the final content without editor chrome. On-screen editing behavior is unchanged.

### DIFF
--- a/.changes/unreleased/tinymce-TINY-12584-2025-09-02.yaml
+++ b/.changes/unreleased/tinymce-TINY-12584-2025-09-02.yaml
@@ -1,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: UI elements like focus outlines and placeholders would be visible after printing.
+time: 2025-09-02T14:40:37.026118+02:00
+custom:
+  Issue: TINY-12584


zuzanna-antas:
* **#10578** <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10578?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10578?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

hamza0867 (approved):

modules/oxide/src/less/theme/content/print/print.less
TheSpyder, line 1: I thought many of these were already set to `@media screen` so they didn't display when printed. Isn't that more sustainable than using `@media print` to override their visibility 🤔

TheSpyder (approved):

modules/oxide/src/less/theme/content/print/print.less
zuzanna-antas: They are not inside `@media screen` unfortunately. The ticket was originally about the uploadcare placeholder but then I discovered that all of these UI elements are visible while printing. 
I can hunt all of these down and wrap them in the media query, but it was just easier to do the `@media print`. Not sure which approach is more sustainable though. 
1. `@media screen` approach -> we will need to remember about wrapping new UI elements inside this media query, but we don't have special overrides file for printing.
2. `@media print` approach -> we have special attributes like `data-mce-bogus` that we add to elements that should not appear in printing. We add them to UI elements, but we don't have to remember about printing styles while writing CSS for the UI. We have additional print overrides file that we need to maintain though.

TheSpyder: Oh, maybe I was thinking of something else. Perhaps log the swap to `@media screen` as a tech debt ticket, and stick to this approach for now.

PR #10586, opened 2025-09-09
TINY-12830: Fix error while splitting list-item into two using enter key

michalnieruchalski-tiugo:
Related Ticket: TINY-12830

Description of Changes:
* Fix error while splitting list-item into two using enter key

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/.changes/unreleased/tinymce-TINY-12830-2025-09-09.yaml
+++ b/.changes/unreleased/tinymce-TINY-12830-2025-09-09.yaml
@@ -1,7 +1,6 @@
 project: tinymce
 kind: Fixed
-body: Error was thrown on the console while splitting list-item into two using enter
-  key.
+body: Splitting list items by pressing enter did not move the cursor due to an error in the browser console.
 time: 2025-09-09T13:47:44.83911+02:00
 custom:
   Issue: TINY-12830


modules/tinymce/src/core/test/ts/browser/newline/NewlineInLiTest.ts
```diff
@@ -130,4 +130,12 @@ describe('browser.tinymce.core.newline.NewlineInLiTest', () => {
     TinyAssertions.assertContent(editor, expectedContent);
     TinyAssertions.assertCursor(editor, [ 0, 2, 0, 0, 0, 0 ], 0);
   });
+
+  it('TINY-12830: splitting li into two using enter should not throw an error', () => {
```
spocke: This doesn't really check if there is no error if you do `editor.execCommand('mceInsertNewline')` that would be roughly the same as pressing enter and that can be wrapped in a try assert that checks that it doesn't throw.
Not sure if we can but the try catch around the content actions in theory that should be sync but maybe make two tests if that works so that we test both the command and the keyboard one.
Another thing we can check is that if the selection is where we expect it to be since in the fiddles at least it looks like it's splitting the LI just fine but the caret remains on the row above it.

spocke: If the `keydown` throws there then we should capture the intent of that not throwing in the test with shouldNotThrow or what ever it's called in chai.

michalnieruchalski-tiugo: I've tried this test and it fails when the error is on the console.

What I can do is:
- [x] - Add another test testing `mceInsertNewline` event
- [x] - Ensure the selection after event in both test cases

spocke: Yeah it seems that we can't capture the exception on a keydown event dispatch. I guess that makes sense since it's the browsers DOM apis that does that event dispatching not the JS runtime.

michalnieruchalski-tiugo: You're right - we can't capture the exception on keydown event. However I was able to capture the exception thrown by executing mceInsertNewLine command - thank you for good suggestion.

spocke (approved):

.changes/unreleased/tinymce-TINY-12830-2025-09-09.yaml
```diff
@@ -0,0 +1,7 @@
+project: tinymce
+kind: Fixed
+body: Error was thrown on the console while splitting list-item into two using enter
+  key.
```
TheSpyder:
```suggestion
body: Splitting list items by pressing enter did not move the cursor due to an error in the browser console.
```

TheSpyder (approved):

PR #10591, opened 2025-09-10
TINY-12862: Normalize `open` attribute value for accordions

michalnieruchalski-tiugo:
Related Ticket: TINY-12862

Description of Changes:
* Normalize `open` attribute value for accordions

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

## Summary by CodeRabbit

- Changed
  - Accordion plugin now normalizes the open attribute on details elements to open="open" for consistent content, regardless of original value (e.g., boolean, "false", or other strings).
- Tests
  - Added browser tests verifying normalization behavior and ensuring non-details elements are unaffected.
- Documentation
  - Added unreleased changelog entry describing the open attribute normalization in the accordion plugin.


michalnieruchalski-tiugo:
* **#10563** <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10563?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10615** <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10615?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10591** <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10591?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10591?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10556** <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10556?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

modules/tinymce/src/plugins/accordion/main/ts/Plugin.ts
```diff
@@ -1,7 +1,7 @@
 import PluginManager from 'tinymce/core/api/PluginManager';
 
 import * as Commands from './api/Commands';
-import * as FilterContent from './core/FilterContent';
+import * as FilterContent from './core/FilterContent/FilterContent';
```
spocke: Maybe name the directory `content` ?

modules/tinymce/src/plugins/accordion/main/ts/core/Identifiers.ts
```diff
@@ -1,12 +1,14 @@
...
 const accordionTag = 'details';
+const openAttribute = 'open';
 const accordionDetailsClass = 'mce-accordion';
 const accordionSummaryClass = 'mce-accordion-summary';
 const accordionBodyWrapperClass = 'mce-accordion-body';
 const accordionBodyWrapperTag = 'div';
 
 export {
   accordionTag,
+  openAttribute,
```
TheSpyder: Do we really need a constant for this? The rest of the plugin doesn't use one. Consistency is very important in a codebase this size.

- [x] Either use the constant everywhere or delete this and just use the string `'open'` like everywhere else.

modules/tinymce/src/plugins/accordion/main/ts/core/content/Normalize.ts
```diff
@@ -0,0 +1,16 @@
+import { Arr } from '@ephox/katamari';
+
+import type AstNode from 'tinymce/core/api/html/Node';
+
+import { openAttribute } from '../Identifiers';
+
+const normalizeOpenAttribute = (node: AstNode): void => {
+  if (hasAttribute(node, openAttribute)) {
```
TheSpyder: You don't need to search the array manually, the `attr` function will return the value if the second argument isn't provided (jquery style)
```suggestion
  if (node.attr(openAttribute)) {
```

modules/tinymce/src/plugins/accordion/main/ts/core/Identifiers.ts
spocke: It's not likely going to change so I would recommend not using a constant. We do use constants for other things like data attributes or classes since it's easier to make typos on longer things and if you do want to change them it's in one spot. The data attributes are mostly internal so those we can usually change if we want to.

modules/tinymce/src/plugins/accordion/main/ts/core/content/Normalize.ts
spocke: It's a weird jQuery ish API I regret that I made it like that I should have used Java style. getAttribute and setAttribute and then we could have had a hasAttribute that just returns a boolean.

So it's a getter and setter in the same function depending on how many arguments you pass to it.

modules/tinymce/src/plugins/accordion/main/ts/core/content/FilterContent.ts
```diff
@@ -75,10 +77,12 @@ const setup = (editor: Editor): void => {
     // Purpose:
     // - add mce-accordion-summary class to summary node
     // - wrap details body in div and add mce-accordion-body class (TINY-9959 assists with Chrome selection issue)
+    // - Normalize accordion 'open' attribute value to open="open"
     parser.addNodeFilter(Identifiers.accordionTag, (nodes) => {
       // Using a traditional for loop here as we may have to iterate over many nodes and it is the most performant way of doing so
       for (let i = 0; i < nodes.length; i++) {
         const node = nodes[i];
+        Normalize.normalizeOpenAttribute(node);
```
spocke, line 85: Do we want to normalize all details elements or just the ones we care about? Should this be inside the if statement?

modules/tinymce/src/plugins/accordion/main/ts/core/Identifiers.ts
michalnieruchalski-tiugo: Yeah I agree it's not going to change. In general I try to avoid using untyped strings because (as you said) it's easier to make a typo.

It's just my preference, therefore I will remove this constant.

Also, we have a coderabbit in place right now, and it's very good at picking up the typos

modules/tinymce/src/plugins/accordion/main/ts/core/content/FilterContent.ts
michalnieruchalski-tiugo: I think all the accordions should be normalized - not only the ones created inside TinyMCE.

If you define the accordion without our class on it it'll still work well, and it's going to be toggable by our accordion plugin. And if it's toggable by the accordion plugin I need the `open` attribute to be normalized. (To prevent content change in readonly mode)

modules/tinymce/src/plugins/accordion/main/ts/core/content/Normalize.ts
michalnieruchalski-tiugo: The only problem I have with this is that I have to look into `attr` implementation to validate what will be returned in case when attribute is missing (it's `undefined`).

This if statement:
```
if (node.attr('open')) {
...
}
```
would not execute for `<details open=""` as empty string is falsy.

Anyway, since the `attr` method is rather hard to follow I would like to stick with `hasAttribute` and `setAttribute` helper functions. I hope you're fine with that.

TheSpyder: oh, right, good point. that's fine.

TheSpyder (approved):

PR #10595, opened 2025-09-16
TINY-12626: No longer expect the top of the editor to always be in view.

HAFRMO:
Related Ticket: TINY-12626

Description of Changes:
Previously we accidentally needed the top of the editor to be in view or we scrolled down to the center of it, which became a problem if the editor was very tall or the element containing it was relatively narrow as it would then scroll to the center whenever focus was gained even when the editor was plainly in view ( and might even already fill the screen )

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved an issue where focusing the editor could unexpectedly scroll the page to center the editor in certain browsers when the editor’s top was off-screen; focusing now preserves the page’s scroll position (TINY-12626).

* **Chores**
  * Added an unreleased changelog entry documenting the fix with timestamp and issue reference.

### DIFF
--- a/modules/tinymce/src/core/main/ts/focus/FocusController.ts
+++ b/modules/tinymce/src/core/main/ts/focus/FocusController.ts
@@ -144,6 +144,8 @@ const setup = (editorManager: EditorManager): void => {
 };
 
 export {
-  isEditorContentAreaElement, isEditorUIElement, isUIElement, setup
+  setup,
+  isEditorUIElement,
+  isEditorContentAreaElement,
+  isUIElement
 };
-
--- a/modules/tinymce/src/core/test/ts/browser/focus/FocusControllerTest.ts
+++ b/modules/tinymce/src/core/test/ts/browser/focus/FocusControllerTest.ts
@@ -115,7 +115,7 @@ describe('browser.tinymce.core.focus.FocusControllerTest', () => {
     });
 
     it('TINY-12017: Editor is inside the viewport once focused', async () => {
-      if (PlatformDetection.detect().browser.isSafari()) {
+      if (PlatformDetection.detect().browser.isSafari()) { // To be removed in TINY-13025
         return;
       }
 


modules/tinymce/src/core/main/ts/focus/FocusController.ts
```diff
@@ -77,7 +77,7 @@ const registerEvents = (editorManager: EditorManager, e: { editor: Editor }) =>
     const element = SugarElement.fromDom(editor.iframeElement);
     const op = OuterPosition.find(element);
     const viewportBounds = WindowVisualViewport.getBounds(window);
-    if (op.top < viewportBounds.y || op.top > (viewportBounds.bottom - minimumVisibility)) {
+    if ((op.top + Height.get(element) - minimumVisibility) < viewportBounds.y || op.top > (viewportBounds.bottom - minimumVisibility)) {
```
TheSpyder: Why do we need the element height here? Isn't that checking the _bottom_ of the editor?

The test moves the editor completely off screen, so it will pass either way, but the changelog and PR title talk about the _top_ of the editor.

TheSpyder (commented):
For the record this is a regression caused by #10319

modules/tinymce/src/core/main/ts/focus/FocusController.ts
HAFRMO: Previously we checked the top of the editor against the top of the viewport, and if it was outside of it the editor was considered 'out of view' even if the editor already filled the screen ( which, if the editor is taller than the view, isn't that difficult ), and the scrolling occurred. This change moves it to check if the bottom of the editor is out of the top of the screen instead, meaning that it won't scroll if the editor is in view anymore. The already existing check checks if the top of the editor is lower than the view, so it's now properly contained.

spocke: Looks correct to me the first one checks if the bottom of the editor is above the current viewport position. The second checks if the top is below the current viewport bottom position.
But unsure if we need the minimumVisibility not sure why that was added?

modules/tinymce/src/core/test/ts/browser/focus/FocusControllerTest.ts
```diff
@@ -118,5 +135,23 @@ describe('browser.tinymce.core.focus.FocusControllerTest', () => {
...
+    it('TINY-12626: Editor is inside the viewport when focused, regardless if it stars above or below the viewport', async () => {
+      const editor = hook.editor();
+
+      Scroll.to(0, 0, SugarDocument.getDocument());
+      assertEditorBodyIsOutsideViewport(editor);
+      editor.getBody().focus();
+      assertEditorBodyInsideViewport(editor);
+      editor.getBody().blur();
+      Scroll.to(0, 1000, SugarDocument.getDocument());
+      assertEditorBodyIsOutsideViewport(editor);
```
spocke: Could we split this assert into two separate functions so that it's above and below it would make this easier to understand.
Don't we also need a editor that is higher than the scroll viewport to test with? So maybe add a separate context with a separate editor hook and wrap the exiting ones in context as a "small editor".

We want to test if editor is above then scroll into view, if the editor is below then scroll into view, if editor is already in view don't scroll. That test should pass for both a editor smaller than the viewport and higher than the viewport.

modules/tinymce/src/core/main/ts/focus/FocusController.ts
TheSpyder: But the PR title and ticket talk about whether the top of the editor is in view, and we didn't make any changes to checks comparing the editor's top position.

Sounds like just a changelog description issue, not a problem in the code.

spocke: Yeah it's if top of editor and the editor is higher than the viewport height then it fails. We need to fix the changelog message.

michalnieruchalski-tiugo:
It’s hard to test this code. I’m using Google Chrome version that contains the fix. So I would have to downgrade my browser to test it properly.

The math included in this ticket looks correct. However it seems like browsers are more intelligent than our workaround. They don’t only check if the iframe is in the viewport, but they also take the selection inside the iframe into consideration. 

Why can’t we just get rid of this if statement? It would fix the TINY-12626 problem, and as for TINY-12017 the solution is to update the browser. 

If we decide to keep and improve the workaround, then we should disable it for versions with a proper fix included - and this would mean version sniffing.

@spocke @TheSpyder @HAFRMO

modules/tinymce/src/core/main/ts/focus/FocusController.ts
HAFRMO: Yes, we previously checked if the top of the editor was in frame, and if it wasn't assumed the editor as a whole was not in frame. We no longer demand that the top is always in view, as the changelog and the PR states. I'm not clear on what the changelog is getting wrong?

As for the minimumVisibility it is a check to make sure the editor is 'reasonably' visible, so it doesn't pass just because a single line of pixels are in frame but instead requires a certain amount of it.

modules/tinymce/src/core/test/ts/browser/focus/FocusControllerTest.ts
```diff
@@ -114,6 +115,10 @@ describe('browser.tinymce.core.focus.FocusControllerTest', () => {
     });
 
     it('TINY-12017: Editor is inside the viewport once focused', async () => {
+      if (PlatformDetection.detect().browser.isSafari()) {
```
michalnieruchalski-tiugo: We should make a ticket to enable this test in the future

modules/tinymce/src/core/main/ts/focus/FocusController.ts
```diff
@@ -163,8 +144,6 @@ const setup = (editorManager: EditorManager): void => {
 };
 
 export {
-  setup,
-  isEditorUIElement,
-  isEditorContentAreaElement,
-  isUIElement
+  isEditorContentAreaElement, isEditorUIElement, isUIElement, setup
```
spocke: Undo this change. The TS LSP sometimes messes this line up if you run organize imports.

modules/tinymce/src/core/test/ts/browser/focus/FocusControllerTest.ts
spocke: Yes, we need a comment here why it's disabled and a ticket that explains that in detail referenced in that comment.

spocke (approved):

PR #10596, opened 2025-09-16
TINY-12384: Added attribute to dragster

HAFRMO:
Related Ticket: TINY-12384

Description of Changes:
Added filter to serializer to prevent Dragster elements showing up in `editor.getContent()`-calls.
Was not able to create a proper test as it'd involve triggering dragster during editor.getContent-call.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

## Summary by CodeRabbit

- **Bug Fixes**
  - Serializer now excludes temporary drag-and-drop helper UI elements, preventing unintended markup in copy/paste, export, and save flows.
- **Chores**
  - Added an unreleased changelog entry recording the fix (TINY-12384).

### DIFF
--- a/.changes/unreleased/tinymce-TINY-12384-2025-09-16.yaml
+++ b/.changes/unreleased/tinymce-TINY-12384-2025-09-16.yaml
@@ -1,6 +1,6 @@
 project: tinymce
 kind: Fixed
-body: Filtering out Dragster elements in serializer ( elements with classes starting with `ephox-dragster-` )
+body: Some UI elements related to dragging elements were not properly filtered out when fetching content.
 time: 2025-09-16T07:19:26.916854+02:00
 custom:
     Issue: TINY-12384
--- a/.changes/unreleased/tinymce-TINY-12384-2025-09-16.yaml
+++ b/.changes/unreleased/tinymce-TINY-12384-2025-09-16.yaml
@@ -1,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: Filtering out Dragster elements in serializer ( elements with classes starting with `ephox-dragster-` )
+time: 2025-09-16T07:19:26.916854+02:00
+custom:
+    Issue: TINY-12384
--- a/modules/tinymce/src/core/main/ts/dom/DomSerializerFilters.ts
+++ b/modules/tinymce/src/core/main/ts/dom/DomSerializerFilters.ts
@@ -1,4 +1,4 @@
-import { Arr, Optional } from '@ephox/katamari';
+import { Arr, Optional, Strings } from '@ephox/katamari';
 
 import type DOMUtils from '../api/dom/DOMUtils';
 import type DomParser from '../api/html/DomParser';
@@ -23,7 +23,7 @@ const register = (htmlParser: DomParser, settings: DomSerializerSettings, dom: D
 
   htmlParser.addNodeFilter('div', (nodes) => {
     Arr.each(nodes, (node) => {
-      if (node.attributes && Arr.find(node.attributes, (attribute) => attribute.name === 'class' && attribute.value.includes('ephox-dragster-')).isSome()) {
+      if (node.attributes && Strings.contains(node.attr('class') || '', 'ephox-dragster')) {
         node.remove();
       }
     });
--- a/modules/tinymce/src/core/main/ts/dom/DomSerializerFilters.ts
+++ b/modules/tinymce/src/core/main/ts/dom/DomSerializerFilters.ts
@@ -1,4 +1,4 @@
-import { Arr, Optional, Strings } from '@ephox/katamari';
+import { Arr, Optional } from '@ephox/katamari';
 
 import type DOMUtils from '../api/dom/DOMUtils';
 import type DomParser from '../api/html/DomParser';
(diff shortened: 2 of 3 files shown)


modules/tinymce/src/core/main/ts/dom/DomSerializerFilters.ts
```diff
@@ -21,6 +21,14 @@ const register = (htmlParser: DomParser, settings: DomSerializerSettings, dom: D
     }
   });
 
+  htmlParser.addNodeFilter('div', (nodes) => {
+    Arr.each(nodes, (node) => {
+      if (node.attributes && Arr.find(node.attributes, (attribute) => attribute.name === 'class' && attribute.value.includes('ephox-dragster-')).isSome()) {
```
lorenzo-pomili: I think you can refactor this as: `Strings.contains(node.attr('class'), 'ephox-dragster')`

modules/tinymce/src/core/main/ts/dom/DomSerializerFilters.ts
```diff
@@ -21,6 +21,14 @@ const register = (htmlParser: DomParser, settings: DomSerializerSettings, dom: D
     }
   });
 
+  htmlParser.addNodeFilter('div', (nodes) => {
```
lorenzo-pomili: Have you tested that other elements like `img` don't produce the same result?

HAFRMO: Dragster doesn't generate images, so those should not occur?

TheSpyder: No but we do resize images.

spocke:
> I think there may be better ways to solve this.
> 
> When the dragster code was created, the content problem was avoided by adding the resize bars as siblings of the `body` tag instead of inside it. Could that solution work here as well?
> 
> Or perhaps add the MCE bogus attribute to the bars, which is already filtered out, instead of needing a custom filter for them?

Yes adding the bogus attribute is how we should handle this. Adding a option to add extra attributes to dragster handles. I think this is how we did it before for some other similar thing.
The thing is the filtering is one part of this the undo snapshots will not use parser filters they look for bogus attributes to filter out things from a raw innerHTML snapshot to be efficient. I will properly review this today have had to deal with the layers problem.

HAFRMO:
> > I think there may be better ways to solve this.
> > When the dragster code was created, the content problem was avoided by adding the resize bars as siblings of the `body` tag instead of inside it. Could that solution work here as well?
> > Or perhaps add the MCE bogus attribute to the bars, which is already filtered out, instead of needing a custom filter for them?
> 
> Yes adding the bogus attribute is how we should handle this. Adding a option to add extra attributes to dragster handles. I think this is how we did it before for some other similar thing. The thing is the filtering is one part of this the undo snapshots will not use parser filters they look for bogus attributes to filter out things from a raw innerHTML snapshot to be efficient. I will properly review this today have had to deal with the layers problem.

I did try to use the bogus attribute at first, that was my first go-to, but it did not work. Maybe I missed something when trying it. I'll give it another shot and see if my mind's cleared up a bit on it.

spocke:
> > > I think there may be better ways to solve this.
> > > When the dragster code was created, the content problem was avoided by adding the resize bars as siblings of the `body` tag instead of inside it. Could that solution work here as well?
> > > Or perhaps add the MCE bogus attribute to the bars, which is already filtered out, instead of needing a custom filter for them?
> > 
> > 
> > Yes adding the bogus attribute is how we should handle this. Adding a option to add extra attributes to dragster handles. I think this is how we did it before for some other similar thing. The thing is the filtering is one part of this the undo snapshots will not use parser filters they look for bogus attributes to filter out things from a raw innerHTML snapshot to be efficient. I will properly review this today have had to deal with the layers problem.
> 
> I did try to use the bogus attribute at first, that was my first go-to, but it did not work. Maybe I missed something when trying it. I'll give it another shot and see if my mind's cleared up a bit on it.

Here is a fiddle of that working: https://fiddle.tiny.cloud/cq9PsvAwzB/0 it sets the `data-mce-bougs` to `all` but works with setting it to just `1` as well. But the difference between `all` and `1` is that `all` will remove any children of that element. So setting `1` is more usual for things like spellchecker annotations where you want to keep the things inside the element and `all` is more for UI things where you don't want to keep any of it's children. They are oddly named I know but legacy code etc. 😄 

The resize bars already have this attribute since they do not appear in the output you can see that in the fiddle if you enable the raw HTML.

modules/tinymce/src/core/main/ts/dom/DomSerializerFilters.ts
spocke: The resize handles already use that specific `data-mce-bogus` attribute:
https://github.com/tinymce/tinymce/blob/main/modules/tinymce/src/core/main/ts/api/dom/ControlSelection.ts#L333

So we should use the same method here.

.changes/unreleased/tinymce-TINY-12384-2025-09-16.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: Filtering out Dragster elements in serializer ( elements with classes starting with `ephox-dragster-` )
```
spocke: No customer would know what dragster is. So maybe it should be more vague just mentioning that some UI elements for table resizing was not properly filtered out.

spocke (approved):
Unsure if we can make a test for this. The resize handles only appear on hover so might be hard to setup such a thing.

PR #10603, opened 2025-09-18
TINY-12819: Migrate yarn to bun

ArvinJ-H:
Related Ticket: 
TINY-12819
TINY-12658

Description of Changes:
* Placeholder text

Pre-checks:
* [ ] Changelog entry added

Review:
* [ ] Milestone set


package.json
```diff
@@ -1,24 +1,42 @@
 {
+  "name": "tinymce-monorepo",
```
ArvinJ-H, line 2: this is required by bun

package.json
```diff
@@ -78,7 +79,7 @@
     "less-plugin-preset-env": "^2.0.2",
     "load-grunt-tasks": "^5.1.0",
     "moxie-zip": "0.0.4",
-    "npm-run-all": "^4.1.5",
+    "npm-run-all2": "^7.0.2",
```
ArvinJ-H, line 82: this is to sync up with premium

Jenkinsfile
```diff
@@ -135,6 +123,37 @@ def runPlaywrightPod(String cacheName, String name, Closure body) {
...
+              # Install bun with retry logic
+              for i in 1 2 3; do
+                curl -fsSL https://bun.sh/install | bash && break || {
+                  echo "Bun install attempt $i failed, retrying in 5s...";
+                  sleep 5;
+                }
+              done
+
+              # Verify installation
+              export PATH="$HOME/.bun/bin:$PATH"
```
jscasca, line 149: well this is obviously not working if you have to do it again in line 399

Jenkinsfile
```diff
@@ -135,6 +123,37 @@ def runPlaywrightPod(String cacheName, String name, Closure body) {
         build: cacheName
       ) {
         container('playwright') {
+          // Install bun if not present (improved with retry logic)
```
jscasca, line 126: Wonder if there is a more efficient way to do this. The whole point of having containers is to avoid installing things every single run. Luckily we don't have to worry about time since the other tests take 10min or more but still...

Jenkinsfile
```diff
@@ -206,7 +202,8 @@ def runSeleniumPod(String cacheName, String name, String browser, String version
 def gitMerge(String primaryBranch) {
   if (env.BRANCH_NAME != primaryBranch) {
     echo "Merging ${primaryBranch} into this branch to run tests"
-    exec("git merge --no-commit --no-ff origin/${primaryBranch}")
+    // TBD: reenable soon
```
jscasca, line 205: Don't forget to re-enable this.

- [ ]  re-enable merge step

Jenkinsfile
ArvinJ-H: playwright container doesn't have bun at the moment, this will be a workaround in the meantime, i won't really worry about time to install, since it takes extra few seconds, but definally doesn't look good

ArvinJ-H:
> I think we need to migrate to storybook 10 as well

yeah... maybe another ticket?

package.json
```diff
@@ -34,9 +35,9 @@
     "headless-test": "grunt headless-auto",
     "headless-test-firefox": "grunt headless-auto --bedrock-browser=firefox-headless",
     "headless-test-manual": "grunt headless-manual",
-    "bedrock": "bedrock --chunk=5000 --customRoutes modules/tinymce/src/core/test/json/routes.json",
-    "test-one": "yarn tsc && yarn bedrock-auto -b chrome-headless -f",
-    "tinymce-dev": "yarn --cwd modules/tinymce dev"
+    "bedrock": "npx bedrock --chunk=5000 --customRoutes modules/tinymce/src/core/test/json/routes.json",
```
hamza0867: Why is this `npx` instead of `bun` ?

ArvinJ-H: loos like that one got missed, and `npm-run-all2`, updated

modules/oxide-components/vite.config.ts
```diff
@@ -40,5 +40,5 @@ export default defineConfig({
       },
     },
   },
-  plugins: [ react() ]
+  plugins: [ react() as any ]
```
ltrouton, line 43: Is this still needed?

modules/oxide-icons-default/package.json
```diff
@@ -10,7 +10,7 @@
   "main": "dist/js/icons.js",
   "scripts": {
     "build": "grunt",
-    "prepublishOnly": "yarn ci",
+    "prepublishOnly": "bun ci",
```
ltrouton, line 13: I am curious, does bun have the feature where we just have this as `"npm ci"` and then when `bun prepublishOnly` is run, bun aliases `npm` automatically? I am just wondering if that is more future proof and could be worth doing for all of the package.json files

yarn.lock
spocke, line 1: I guess this does mean that we will have no lock file for the OSS part. What would happen if we simply copy in the lock file from the premium side and then run `bun`to install packages in that sub repo would that then remove the unused dependencies. I would be nice if we could keep the lock file somehow but make the process of keeping these in sync something we could script instead of the horrible way we are dealing with this right now.

modules/oxide-components/src/main/ts/bespoke/tinymceai/tag/Tag.tsx
```diff
@@ -1,4 +1,4 @@
-import { forwardRef, type FunctionComponent } from 'react';
+import { forwardRef, type FunctionComponent, type MouseEvent, type FocusEvent } from 'react';
```
spocke, line 1: Why did these type changes popup? Did we bump eslint?

Jenkinsfile
```diff
@@ -135,6 +123,37 @@ def runPlaywrightPod(String cacheName, String name, Closure body) {
...
+              for i in 1 2 3; do
+                (apt-get update -qq && apt-get install -y -qq unzip curl) && break || {
+                  echo "Apt install attempt $i failed, retrying in 5s...";
+                  sleep 5;
+                }
+              done
+
+              # Install bun with retry logic
+              for i in 1 2 3; do
+                curl -fsSL https://bun.sh/install | bash && break || {
```
spocke, line 142: I wonder if this could be a source of failure or flakyness maybe there is some easier way dealing with this

bunfig.toml
```diff
@@ -0,0 +1,13 @@
+# Bun configuration file
+# Replaces .yarnrc settings
```
spocke, line 2: Not sure we need this comment once we merge this yarn will be gone

modules/oxide-components/src/main/ts/bespoke/tinymceai/error/ErrorMessage.tsx
```diff
@@ -1,6 +1,8 @@
-import { Bem, IconButton } from 'oxide-components/main';
 import React from 'react';
 
+import { IconButton } from '../../../components/iconbutton/IconButton';
```
spocke, line 3: What are these changed about, why are they needed when we update to bun? I thought we would just import the package lock into bun?

PR #10607, opened 2025-09-19
TINY-12857: Added elements to prevent excess newlines.

HAFRMO:
Related Ticket:  TINY-12857

Description of Changes:
Some items recieved several newlines instead of the one they should have.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/.changes/unreleased/tinymce-TINY-12857-2025-09-19.yaml
+++ b/.changes/unreleased/tinymce-TINY-12857-2025-09-19.yaml
@@ -1,0 +1,6 @@
+project: tinymce
+kind: Improved
+body: Some elements would be given an extra newline by the serializer when indented
+time: 2025-09-19T08:34:40.606198+02:00
+custom:
+    Issue: TINY-12857
--- a/modules/tinymce/src/core/test/ts/browser/html/DomParserTest.ts
+++ b/modules/tinymce/src/core/test/ts/browser/html/DomParserTest.ts
@@ -1944,5 +1944,21 @@ describe('browser.tinymce.core.html.DomParserTest', () => {
         expected: `<div><${elementName}><em>test</em></${elementName}></div>`
       }));
     });
+
+    it('TINY-12857: Serializer newlines properly', () => {
+      // eslint-disable-next-line max-len
+      const originalText = '<html><head><meta http-equiv="x-ua-compatible" content="ie=edge"><meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no"><link rel="canonical" href="https://www.tiny.cloud/docs/tinymce/latest/"><link rel="sitemap" type="application/xml" href="/gatsby_sitemap.xml"><meta property="og:title" content="My Website"><meta property="og:description" content="Check out my awesome site!"></head><body></body></html>';
+
+      const output = HtmlSerializer({
+        validate: false,
+        indent: true,
+        indent_before: 'head,html,body,meta,title,script,link,style',
+        indent_after: 'head,html,body,meta,title,script,link,style'
+      }, schema).serialize( DomParser({ root_name: '#document', sanitize: false }, schema).parse(originalText));
+
+      // eslint-disable-next-line max-len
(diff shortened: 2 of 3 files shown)


.changes/unreleased/tinymce-TINY-12857-2025-09-19.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Improved
+body: Some elements would be given an extra newline by the serializer when indented
```
spocke: I think we should also log this for the `fullpagehtml` plugin since that is where a user would hit this issue.

```suggestion
body: Some elements would be given an extra newline by the serializer when indented.
```

modules/tinymce/src/core/test/ts/browser/html/DomParserTest.ts
```diff
@@ -1944,5 +1944,21 @@ describe('browser.tinymce.core.html.DomParserTest', () => {
...
+        indent: true,
+        indent_before: 'head,html,body,meta,title,script,link,style',
+        indent_after: 'head,html,body,meta,title,script,link,style'
+      }, schema).serialize( DomParser({ root_name: '#document', sanitize: false }, schema).parse(originalText));
+
+      // eslint-disable-next-line max-len
+      const expected = '<head>\n<meta http-equiv="x-ua-compatible" content="ie=edge">\n<meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">\n<link rel="canonical" href="https://www.tiny.cloud/docs/tinymce/latest/">\n<link rel="sitemap" type="application/xml" href="/gatsby_sitemap.xml">\n<meta content="My Website">\n<meta content="Check out my awesome site!">\n</head>\n<body></body>';
+      assert.equal(output, expected);
+    });
   });
```
spocke, line 1970: I think it would be cleaner if you just did this with an array that you then joined with '' here and then in the expected you join the lines with '\n'. You could then get rid of the eslint rules to disable the long lines.

This also filters out the RDFa `property` not sure we need to have that in there since it we don't then you can just take middle chunk stick that into an array and re-use it in the expected case reducing the lines in the test case.

```suggestion
    it('TINY-12857: Serializer newlines properly', () => {
      const originalText = [
        '<html>',
        '<head>',
        '<meta http-equiv="x-ua-compatible" content="ie=edge">',
        '<meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">',
        '<link rel="canonical" href="https://www.tiny.cloud/docs/tinymce/latest/">',
        '<link rel="sitemap" type="application/xml" href="/gatsby_sitemap.xml">',
        '<meta property="og:title" content="My Website">',
        '<meta property="og:description" content="Check out my awesome site!">',
        '</head>',
        '<body>'
        '</body>',
        '</html>'
      ].join('');

      const output = HtmlSerializer({
        validate: false,
        indent: true,
        indent_before: 'head,html,body,meta,title,script,link,style',
        indent_after: 'head,html,body,meta,title,script,link,style'
      }, schema).serialize( DomParser({ root_name: '#document', sanitize: false }, schema).parse(originalText));

      const expected = [
        '<head>',
        '<meta http-equiv="x-ua-compatible" content="ie=edge">',
        '<meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">',
        '<link rel="canonical" href="https://www.tiny.cloud/docs/tinymce/latest/">',
        '<link rel="sitemap" type="application/xml" href="/gatsby_sitemap.xml">',
        '<meta content="My Website">',
        '<meta content="Check out my awesome site!">',
        '</head>',
        '<body></body>'
      ].join('\n');
    });
  });
```

spocke (approved):

modules/tinymce/src/core/test/ts/browser/html/DomParserTest.ts
TheSpyder: Or use a template string (although that's harder to get right when the test code is indented).

spocke: Yeah I've thought maybe we should built in a function into mcagar that removes whitespace at beginning of lines in a string since then you could use a template string with indentation. I've seen other libraries doing that.

PR #10612, opened 2025-09-22
TINY-12867: Remove `allow-same-origin` from sanbox spec

ArvinJ-H:
Related Ticket: 
TINY-12867

Description of Changes:
- Removed `allow-same-origin` from sanbox spec

Pre-checks:
* [ ] Changelog entry added

Review:
* [ ] Milestone set

### DIFF
--- a/modules/tinymce/src/themes/silver/main/ts/ui/dialog/IFrame.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/ui/dialog/IFrame.ts
@@ -133,7 +133,7 @@ const renderIFrame = (spec: IframeSpec, providersBackstage: Backstage.UiFactoryB
   const attributes = {
     ...spec.label.map<{ title?: string }>((title) => ({ title })).getOr({}),
     ...initialData.map((html) => ({ srcdoc: html })).getOr({}),
-    ...spec.sandboxed ? { sandbox: 'allow-scripts allow-same-origin' } : { }
+    ...spec.sandboxed ? { sandbox: 'allow-scripts' } : {}
   };
 
   const sourcing = getDynamicSource(initialData, spec.streamContent);


modules/sugar/src/main/ts/ephox/sugar/api/dom/Link.ts
```diff
@@ -59,7 +63,47 @@ const getPreventClicksOnLinksScript = (): string => {
   return `<script>(${fn.toString()})(${isMacOSOrIOS})</script>`;
 };
 
+const preventClicksOnLinks = (scope?: SugarElement<Document>): void => {
+  const doc = scope ?? SugarElement.fromDom(document);
```
spocke, line 67: There is a SugarDocument module.

modules/sugar/src/main/ts/ephox/sugar/api/dom/Link.ts
```diff
@@ -59,7 +63,47 @@ const getPreventClicksOnLinksScript = (): string => {
...
+
+  const existingHandler = (doc.dom as any)[blockHandlerKey] as ((e: MouseEvent) => void) | undefined;
```
spocke, line 69: There is no point in tracking this and re-attaching it since a contentDocument can just be created once for an iframe.
If you want to track the attached state then make this return a unbind that you manually call first.

modules/sugar/src/main/ts/ephox/sugar/api/dom/Link.ts
```diff
@@ -59,7 +63,47 @@ const getPreventClicksOnLinksScript = (): string => {
...
+        if (!isMetaKeyPressed) {
+          e.preventDefault();
+        }
+
+        return;
+      }
+    }
+  };
+
+  doc.dom.addEventListener('click', handler, false);
```
spocke, line 101: There is a DomEvent module use that and return the unbinder from this function that way you don't need to have some mutating state to track if it's attached or not.

modules/sugar/src/main/ts/ephox/sugar/api/dom/Link.ts
```diff
@@ -59,7 +63,47 @@ const getPreventClicksOnLinksScript = (): string => {
...
+
+  const existingHandler = (doc.dom as any)[blockHandlerKey] as ((e: MouseEvent) => void) | undefined;
+  if (existingHandler) {
+    doc.dom.removeEventListener('click', existingHandler, false);
+  }
+
+  const isMacOSOrIOS = Platform.isMacOS() || Platform.isiOS();
+
+  const handler = (e: MouseEvent) => {
+    for (let elm = e.target as Node | null; Type.isNonNullable(elm); elm = elm.parentNode) {
```
spocke, line 77: I would use `SelectorFind.closest<HTMLAnchorElement>(e.target, 'a[href]')` that way you get a HTMLElement and you don't need to unsafe case anything. Then you just grab the href from that and do the even prevent and scrolling etc.

spocke: I know the old code was messy but that was inlined into the iframe if we are moving it out we can clean it up.

PR #10627, opened 2025-10-01
TINY-12875: introduce draggable component

michalnieruchalski-tiugo:
Related Ticket: TINY-12875

Description of Changes:
* Add draggable component to oxide-comoponents

Pre-checks:

Review:
~~* [ ] Milestone set~~

## Summary by CodeRabbit

- **New Features**
  - Added a Draggable component with a nested Handle for pointer-driven dragging, with bounded, integer-rounded translations; exported for public use.
- **Documentation**
  - Added a Storybook example demonstrating Draggable usage.
- **Tests**
  - Added unit tests for drag calculations and browser tests verifying dragging, rounding, boundary clamping, and Handle usage constraints.
- **Chores**
  - Added utilities, types, context and updated test configurations and test-ignore patterns.

### DIFF
--- a/modules/oxide-components/src/main/ts/components/draggable/Draggable.stories.tsx
+++ b/modules/oxide-components/src/main/ts/components/draggable/Draggable.stories.tsx
@@ -11,7 +11,7 @@ const meta = {
     layout: 'centered',
   },
   // This component will have an automatically generated Autodocs entry: https://storybook.js.org/docs/writing-docs/autodocs
-  tags: [ 'autodocs' ],
+  tags: [ 'autodocs', 'skip-visual-testing' ],
 } satisfies Meta<typeof Draggable>;
 
 export default meta;


michalnieruchalski-tiugo:
* **#10627** <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10627?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10627?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

modules/oxide-components/src/main/ts/components/draggable/Draggable.tsx
```diff
@@ -0,0 +1,82 @@
...
+  const [ isDragging, setIsDragging ] = useState(false);
+  const handleRef = useRef<HTMLDivElement | null>(null);
+  const lastMousePositionRef = useRef<Position>({ x: 0, y: 0 });
+  const boundriesRef = useRef<Boundries>({ shiftX: { min: 0, max: 0 }, shiftY: { min: 0, max: 0 }});
+  const { setShift, draggableRef, shift } = useDraggable();
+
+  const onPointerDown = useCallback((event: React.PointerEvent) => {
+    if (handleRef.current === null || draggableRef.current === null) {
+      // If handleRef or draggableRef is not present then abort dragging
+      return;
```
TheSpyder, line 32: if we're aborting, should it also set dragging to false?

modules/oxide-components/src/main/ts/components/draggable/Draggable.tsx
```diff
@@ -0,0 +1,82 @@
...
+  const onPointerMove = useCallback((event: React.PointerEvent) => {
+    if (isDragging) {
+      const currentPointerPosition = { x: event.clientX, y: event.clientY };
+      const { deltaX, deltaY } = delta(lastMousePositionRef.current, currentPointerPosition);
+      lastMousePositionRef.current = currentPointerPosition;
+      const calculatedBoundries = boundriesRef.current;
+      setShift(({ x, y }) => ({
+        x: clamp(Math.floor(x + deltaX), calculatedBoundries.shiftX.min, calculatedBoundries.shiftX.max),
+        y: clamp(Math.floor(y + deltaY), calculatedBoundries.shiftY.min, calculatedBoundries.shiftY.max)
+      }));
```
TheSpyder: By clamping this way suspect you might be creating some weird drag behaviour. I think I recognise this pattern from some previous UX I've implemented.

Unfortunately I don't have time to find a proper example before I leave on holidays for a week, so hopefully this description is enough:
- drag something outside of the allowed boundaries
- move the mouse around while the mouse is still outside the boundary
- if the item you're dragging moves off the boundary while wiggling the mouse in this different location, that's a bug

The trick is to clamp the mouse position, not the dragged item position. This ensures the dragged item always sticks to the mouse but it still won't follow the mouse outside the boundaries.

- [x] check my theory, or ping me on slack and we can discuss when I return

TheSpyder (approved):

modules/oxide-components/src/test/ts/atomic/draggable/calculations.spec.ts
hamza0867, line 1: I believe this should be an atomic test. Maybe we should setup atomic tests to run on node instead of the browser ?

modules/oxide-components/src/main/ts/components/draggable/internals/types.ts
```diff
@@ -0,0 +1,29 @@
+import type { PropsWithChildren } from 'react';
+
+export interface DraggableProps extends PropsWithChildren { }
+
+export interface DraggableHandleProps extends PropsWithChildren { }
+
+export interface Shift {
+  x: number; y: number;
```
hamza0867: I think this is more readable:

```suggestion
  x: number;
  y: number;
```

modules/oxide-components/src/main/ts/components/draggable/internals/types.ts
```diff
@@ -0,0 +1,29 @@
...
+};
+
+export interface Position {
+  x: number; y: number;
```
hamza0867: Same as before

modules/oxide-components/src/test/ts/browser/components/Draggable.spec.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+  return { handle, draggable, draggableWrapper: draggableWrapper as HTMLElement };
+};
+
+/*
+  I would prefer to use React's 'act' function.
+  But I couldn't find a way to use 'act' with 'vitest-browser-react' library,
+  without having error message on the console.
+*/
+const tick = () => new Promise<void>((res) => setTimeout(() => res()));
```
hamza0867, line 93: What about `Waiter.pWait(0)` ?

modules/oxide-components/src/test/ts/browser/components/draggable/Draggable.spec.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+    assertTransform(draggableWrapper, { x: 2, y: 50 });
+  });
+
+  it('TINY-12875: Should not exceed border - left top', async () => {
+    const { handle, draggableWrapper } = await renderDraggable();
+    const rect = draggableWrapper.getBoundingClientRect();
+
+    // Try to move far beyond the window boundaries
+    const largeShift = [ -1 * window.innerWidth * 2, -1 * window.innerHeight * 2 ] as const;
```
hamza0867: Why not 
```suggestion
    const largeShift = [ window.innerWidth * (-2), window.innerHeight * (-2) ] as const;
```

hamza0867 (approved):

modules/oxide-components/src/main/ts/components/draggable/Draggable.tsx
```diff
@@ -0,0 +1,82 @@
...
+      onPointerUp={onPointerUp}
+      onPointerMove={onPointerMove}
+      onLostPointerCapture={onLostPointerCapture}
+      style={{
+        cursor: isDragging ? 'grabbing' : 'grab'
+      }}>
+      {children}
+    </div>
+  );
+};
```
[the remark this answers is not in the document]
michalnieruchalski-tiugo: Good point, we're planning to address that in future PR's

modules/oxide-components/src/test/ts/browser/components/Draggable.spec.tsx
michalnieruchalski-tiugo: I would have to add `@ephox/agar` as `devDependency` of `oxide-components`. Do you think it's worth it?

modules/oxide-components/src/test/ts/atomic/draggable/calculations.spec.ts
michalnieruchalski-tiugo: Yes! I did that, please review my setup :)

hamza0867: Well Done!

modules/oxide-components/src/test/ts/browser/components/Draggable.spec.tsx
hamza0867: Ok. That's fine then 👍

modules/oxide-components/src/main/ts/components/draggable/Draggable.tsx
michalnieruchalski-tiugo: It wouldn't hurt to add that but I think it's not needed. This method is where dragging starts, so we can assume that `isDragging` is `false` there.

modules/oxide-components/src/main/ts/components/draggable/Draggable.tsx
michalnieruchalski-tiugo: This was a very good advice. I followed it and implemented clamping the mouse. The experience feels much more natural.

I think it'll be good if you could try to play around with the dragging element and validate if the behavior you're worried about exists there.

modules/oxide-components/src/main/ts/components/draggable/Draggable.tsx
```diff
@@ -0,0 +1,81 @@
...
+    if (isDragging) {
+      const currentPointerPosition = {
+        x: clamp(Math.round(event.clientX), boundriesRef.current.x.min, boundriesRef.current.x.max),
+        y: clamp(Math.round(event.clientY), boundriesRef.current.y.min, boundriesRef.current.y.max)
+      };
+      const { deltaX, deltaY } = delta(lastMousePositionRef.current, currentPointerPosition);
+      lastMousePositionRef.current = currentPointerPosition;
+      setShift(({ x, y }) => ({ x: x + deltaX, y: y + deltaY }));
+    }
+  }, [ isDragging, setShift ]);
```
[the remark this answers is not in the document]
michalnieruchalski-tiugo: I don't get this comment and the math in **Example scenario** is just wrong.

* Mouse at (600, 100) is clamped to (475, 100)
* Mouse moves to (550, 200), clamped to (475, 200)
* Delta is (0, 100), so the element moves down 
* But the actual mouse moved left (from 600 to 550) and down—the element's movement doesn't match the user's intent

Yes - the left movement is ignored. If we would follow it the element would be moved outside of boundry.

modules/oxide-components/src/main/ts/components/draggable/Draggable.tsx
TheSpyder: I'm glad, it's one of those cases where you don't realise how much complexity goes into a seemingly simple thing - when a lot of people will only describe the dragging as "it feels wrong" without being able to articulate why.

Unfortunately I can't really point to a specific tutorial to follow, I've spent quite a bit of time looking at and studying UX both in and out of browsers. There is some cool maths involved, such as this clamping example, to make things feel natural.

Is there an easy demo where I can play with it?

modules/oxide-components/src/main/ts/components/draggable/Draggable.tsx
TheSpyder: Only if the previous drag events happened in the order you expect. If it doesn't hurt, being defensive can help avoid bugs when your assumptions no longer hold.

PR #10632, opened 2025-10-03
INT-3373: Focus and caret placement order change.

HAFRMO:
Related Ticket: INT-3373

Description of Changes:
When focus was given before moving the caret chrome would scroll to the top of the editor in some situations ( editor was not focused, caret was not in view at the time, the content was large enough to require a scroll ). Moving the caret first, then setting focus solves the issue.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

## Summary by CodeRabbit

- Bug Fixes
  - Prevents unexpected page/editor scrolling in Chromium when dragging content over the editor, even if another element is focused.

- Tests
  - Added an automated test to ensure no scroll occurs during dragover in the described scenario.

- Chores
  - Added an unreleased changelog entry documenting the fix.

### DIFF
--- a/.changes/unreleased/tinymce-INT-3373-2025-10-02.yaml
+++ b/.changes/unreleased/tinymce-INT-3373-2025-10-02.yaml
@@ -1,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: Chromium browsers would in certain situations scroll the editor unexpectedly when dragging content over the editor.
+time: 2025-10-02T15:42:09.368737+02:00
+custom:
+    Issue: INT-3373
--- a/modules/tinymce/src/core/test/ts/browser/paste/DragDropSummaryTest.ts
+++ b/modules/tinymce/src/core/test/ts/browser/paste/DragDropSummaryTest.ts
@@ -1,6 +1,7 @@
 import { Waiter } from '@ephox/agar';
 import { describe, it } from '@ephox/bedrock-client';
 import { TinyAssertions, TinyHooks, TinySelections } from '@ephox/wrap-mcagar';
+import { assert } from 'chai';
 
 import type Editor from 'tinymce/core/api/Editor';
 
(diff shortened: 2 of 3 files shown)


michalnieruchalski-tiugo (approved):
(question) How come is this ticket integration? INT- prefix instead of TINY- ?

HAFRMO:
> (question) How come is this ticket integration? INT- prefix instead of TINY- ?

Oh, that is very easy. It seems I was dealing with an INT- ticket, then needed to make this one and did not check that I was in the correct project. And by that point we realized we just thought it'd be easier to just finish it than remake it.
So no meaning behind it.

.changes/unreleased/tinymce-INT-3373-2025-10-02.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: Chromium browsers would in certain situations scroll the editor unexpectedly when dragging content over the editor.
+time: 2025-10-02T15:42:09.368737+02:00
+custom:
+    Issue: INT-3373
```
spocke, line 6: Should this really be a INT issue. The bug is not in the integrations but in the tinymce core. We only have one other ticket like this in the changelog. Can we log at TINY ticket and then remake this PR with that ticket?
We might need to check this with someone not sure if this is the procedure we want.

modules/tinymce/src/core/test/ts/browser/paste/DragDropSummaryTest.ts
```diff
@@ -56,5 +57,26 @@ describe('browser.tinymce.core.paste.DragDropSummaryTest', () => {
...
+
+    editor.setContent('<p>Text</p>'.repeat(1000));
+    const div = document.createElement('div');
+    div.setAttribute('contentEditable', 'true');
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    editor.getContainer().parentElement!.append(div);
+
+    const scrollAmount = 5000;
+    editor.getWin().scrollTo(0, scrollAmount);
+    await Waiter.pTryUntil('Scroll to point', () => assert.equal(editor.getWin().scrollY, scrollAmount));
```
spocke, line 72: Would this still be needed if you used behavior instant?

.changes/unreleased/tinymce-INT-3373-2025-10-02.yaml
HAFRMO: Since this need to be in today, no. But probably what we should do next time.

modules/tinymce/src/core/test/ts/browser/paste/DragDropSummaryTest.ts
HAFRMO: Still good to check.

PR #10656, opened 2025-10-14
TINY-12827: Moved in spinner

HAFRMO:
Related Ticket: TINY-12827

Description of Changes:
Moved spinner over

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

## Summary by CodeRabbit

* **New Features**
  * Added an animated Spinner component to the UI library and exported it for use across apps.
* **Documentation**
  * Added a Storybook entry showcasing the Spinner.
* **Style**
  * Added TinyAI-themed styles and included the spinner in the theme stylesheet.

### DIFF
--- a/modules/oxide/src/less/theme/theme.less
+++ b/modules/oxide/src/less/theme/theme.less
@@ -71,6 +71,7 @@
 @import 'components/slider/slider';
 @import 'components/source-code/source-code';
 @import 'components/spinner/spinner';
+@import 'components/tinyai/spinner';
 @import 'components/statusbar/statusbar';
 @import 'components/tag/tag';
 @import 'components/throbber/throbber';


modules/oxide-components/src/main/ts/bespoke/tinyai/TinyAiUiAttributes.ts
hamza0867, line 1: This file shouldn't exist. You should use the `classes` utility function. It should have the class name we need

modules/oxide-components/src/main/ts/bespoke/tinyai/spinner/Spinner.tsx
```diff
@@ -0,0 +1,15 @@
+import type { FunctionComponent } from 'react';
+
+import * as TinyAiUiAttributes from '../TinyAiUiAttributes';
+
+export const Spinner: FunctionComponent = () =>
+  <div className={TinyAiUiAttributes.classes.aiSpinner}>
```
hamza0867: So in here you should instead use the classes utility function

spocke (approved):
Don't we also need to export the component. Now it's just internal to the storybook right?

modules/oxide/src/less/theme/components/tinyai/tinyai.less
```diff
@@ -0,0 +1,4 @@
+.tox-ai__spinner {
```
spocke: This needs to be properly prefixed with `.tox` also it's a separate component shouldn't this be a `spinner.less` file in the `tinyai` directory.
The spinner here is it's own component so following BEM it should not be an element class but a block class so `.tox-ai-spinner`.

- [x] Properly prefix and move this to the right file and maybe change the name

zuzanna-antas: I agree, it's a separate component and it should be in a separate file. As in oxide-components it's in the `bespoke` components I think it can be in the tinyai directory. Or maybe simply next to the `ai-chat.less` in the ai-chat dir?

PR #10661, opened 2025-10-16
TINY-13071: Add initial InlineToolbar component with context

kimwoodfield:
Related Ticket: [TINY-13071](https://ephocks.atlassian.net/browse/TINY-13071)

Description of Changes:
* First pass at the new `InlineToolbar` component which will be used in an upcoming feature. Includes basic scaffolding and behaviour. More to come in follow-up PRs.
* Still to do: advanced positioning logic, keyboard navigation and more Storybook examples.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

[TINY-13071]: https://ephocks.atlassian.net/browse/TINY-13071?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/oxide-components/src/main/ts/components/inlinetoolbar/InlineToolbar.tsx
+++ b/modules/oxide-components/src/main/ts/components/inlinetoolbar/InlineToolbar.tsx
@@ -24,7 +24,7 @@ const InlineToolbarContext = createContext<InlineToolbarContextValue | null>(nul
 
 const useInlineToolbarContext = () => {
   const context = useContext(InlineToolbarContext);
-  if (!context) {
+  if (!Type.isNonNullable(context)) {
     throw new Error('useInlineToolbarContext must be used within an InlineToolbarProvider');
   }
   return context;
@@ -108,25 +108,12 @@ const Toolbar: FC<ToolbarProps> = ({
     const handleClickOutside = (event: MouseEvent) => {
       if (
         isOpen &&
-          Type.isNonNullable(toolbarRef.current) &&
-          Type.isNonNullable(triggerRef.current)
+        Type.isNonNullable(toolbarRef.current) &&
+        Type.isNonNullable(triggerRef.current) &&
+        event.target instanceof Node
       ) {
-        const toolbarRect = toolbarRef.current.getBoundingClientRect();
-        const triggerRect = triggerRef.current.getBoundingClientRect();
-        const { clientX, clientY } = event;
-
-        const clickedToolbar =
-          clientX >= toolbarRect.left &&
-          clientX <= toolbarRect.right &&
-          clientY >= toolbarRect.top &&
-          clientY <= toolbarRect.bottom;
-
-        const clickedTrigger =
-          clientX >= triggerRect.left &&
-          clientX <= triggerRect.right &&
-          clientY >= triggerRect.top &&
-          clientY <= triggerRect.bottom;
-
+        const clickedToolbar = toolbarRef.current.contains(event.target);
+        const clickedTrigger = triggerRef.current.contains(event.target);
         if (!clickedToolbar && !clickedTrigger) {
           close();
         }
@@ -134,7 +121,8 @@ const Toolbar: FC<ToolbarProps> = ({
     };
     document.addEventListener('mousedown', handleClickOutside);
     return () => document.removeEventListener('mousedown', handleClickOutside);
-  }, [ isOpen, close, toolbarRef, triggerRef, persistent ]);
+    /* eslint-disable-next-line react-hooks/exhaustive-deps -- toolbarRef/triggerRef are stable ref objects and don't need to be in deps list. */
+  }, [ isOpen, close, persistent ]);
 
   const getPosition = (sink: HTMLDivElement, trigger: HTMLDivElement) => {
     const sinkRect = sink.getBoundingClientRect();


modules/oxide-components/src/test/ts/browser/components/InlineToolbar.spec.tsx
```diff
@@ -0,0 +1,131 @@
...
+    const { getByTestId } = render(
+      <Fragment>
+        <div ref={sinkRef} className="tox" style={{ position: 'relative' }} />
+        <InlineToolbar sinkRef={sinkRef}>
+          <InlineToolbar.Trigger>
+            <div data-testid={triggerTestId}>Click Me</div>
+          </InlineToolbar.Trigger>
+          <InlineToolbar.Toolbar>
+            <div data-testid={toolbarTestId}>Toolbar Content</div>
+          </InlineToolbar.Toolbar>
```
tiny-ben-tran, line 40: IMO using <Trigger> for opening/closing a toolbar/dropdown is not great. Instead, the toolbar should have a prop like `isOpen`, so the toolbar can be used like this:
```
<InlineToolbar isOpen={false} sinkRef={sinkRef}>
       <div> toolbar content</div>
</InlineToolbar>
```
The implementation and component interface will be much simpler.

MitchC1999: The point of this component is for it to manage it's own state, without needing to re-render the parent. Having it open with a boolean property is essentially the same as only rendering the component in a conditional. Like this
```
isOpen ? <InlineToolbar /> : null
```
This still puts the burden of state on the parent, which we should avoid with these reusable components. If we continue with that, then this component becomes nothing more than a popup container for toolbar buttons.
Eventually this component should support external triggers, but it should still be able to open/close based on the trigger being clicked/focused, just like the core context toolbar.

tiny-ben-tran: > manage it's own state, without needing to re-render the parent. Having it open with a boolean property is essentially the same as only rendering the component in a conditional

This can be achieved in a different way that doesn't need to have a trigger or require using `context` .  I also see that all sub-components here are tightly coupled as it needs the inlinetoolbar context to work. It reduces the reusability of the components.  I don't mind this pattern when it is only being used in one place ie: suggestededits plugin. However, I would go against using this pattern in a UI library. 

> Eventually this component should support external triggers
Why not think about that now, then later? We will have to support opening/closing the inlinetoolbar at the cursor very soon, right

MitchC1999: The subcomponents are part of the whole, sharing the context. This pattern makes this component much easier to use, and more customizable. This component supports opening/closing by clicking on or outside the trigger component, hence the need to share context. What benefit do we gain from a reusable component library if we have to reimplement the same state management every time we use a component, which itself ends up being a basic wrapper we could have written bespoke? If we aren't thinking of making these components plug and play/self-contained, what's the point of this library?

modules/oxide-components/src/main/ts/components/inlinetoolbar/InlineToolbar.tsx
```diff
@@ -0,0 +1,179 @@
...
+  if (!context) {
+    throw new Error('useInlineToolbarContext must be used within an InlineToolbarProvider');
+  }
+  return context;
+};
+
+const InlineToolbar: FC<InlineToolbarProps> & {
+  Trigger: FC<TriggerProps>;
+  Toolbar: FC<ToolbarProps>;
+} = ({
```
hamza0867: Similar to the `Draggable` component we made. I'd rather we call this the `Root` component and export `Root` and `Trigger` and `Toolbar` as three components. It makes it easier to read types and work with them

modules/oxide-components/src/main/ts/components/inlinetoolbar/InlineToolbar.tsx
```diff
@@ -0,0 +1,179 @@
...
+  children,
+  sinkRef,
+  persistent = false
+}) => {
+  const [ isOpen, setIsOpen ] = useState(false);
+  const triggerRef = useRef<HTMLDivElement>(null);
+  const toolbarRef = useRef<HTMLDivElement>(null);
+
+  const open = () => setIsOpen(true);
+  const close = () => setIsOpen(false);
```
hamza0867: Small thing, but this can be put in a `useCallback`.

modules/oxide-components/src/main/ts/components/inlinetoolbar/InlineToolbar.tsx
```diff
@@ -0,0 +1,179 @@
...
+
+  const context: InlineToolbarContextValue = {
+    isOpen,
+    open,
+    close,
+    triggerRef,
+    toolbarRef,
+    sinkRef,
+    persistent
+  };
```
hamza0867: Similar, this can be put in `useMemo`

modules/oxide-components/src/main/ts/components/inlinetoolbar/InlineToolbar.tsx
```diff
@@ -0,0 +1,179 @@
...
+  useEffect(() => {
+    const handleKeyDown = (event: KeyboardEvent) => {
+      if (KeyMatch.inSet(Keys.ESCAPE)(event) && isOpen) {
+        close();
+      }
+    };
+
+    document.addEventListener('keydown', handleKeyDown);
+    return () => document.removeEventListener('keydown', handleKeyDown);
+  }, [ isOpen, close ]);
```
hamza0867: You can use the `useSpecialKeyNavigation` from the `KeyboardNavigationHooks` and define `onEscape`

modules/oxide-components/src/main/ts/components/inlinetoolbar/InlineToolbar.tsx
```diff
@@ -0,0 +1,179 @@
...
+    if (persistent) {
+      return;
+    }
+
+    const handleClickOutside = (event: MouseEvent) => {
+      if (
+        isOpen &&
+                Type.isNonNullable(toolbarRef.current) &&
+                Type.isNonNullable(triggerRef.current)
+      ) {
```
hamza0867, line 114: The indentation looks weird here.

modules/oxide-components/src/main/ts/main.ts
```diff
@@ -2,6 +2,7 @@ import { AutoResizingTextarea } from './components/autoresizingtextarea/AutoResi
 import { Button } from './components/button/Button';
 import { Draggable } from './components/draggable/Draggable';
 import { IconButton } from './components/iconbutton/IconButton';
+import { InlineToolbar } from './components/inlinetoolbar/InlineToolbar';
```
hamza0867: So if we follow, what I mentioned earlier about using `Root` we can here `import * as InlineToolbar from '...'` and then `export InlineToolbar`

modules/oxide-components/src/test/ts/browser/components/InlineToolbar.spec.tsx
hamza0867: I will have to agree with Mitch here. This way makes it much easier to use this component. That being said, do we plan to use this for the editor toolbar or context menu toolbar ? Or is this a new concept that is used only in suggested edits for now ?

MitchC1999: I would hope this could be easily adjusted to work as a replacement for the core editor context toolbar. This concept was first explored in suggested edits with the Dropdown component, which I was able to use for every dropdown in the plugin. Its usage is based on radix ui which I found really easy to use, but we needed our own version for obvious reasons. That dropdown imitates the core collection menus, with the same classes, and those are reused in a number of core components themselves. This is like a horizontal version of that, but it opens from clicking within a node, like how the editor context toolbar works.

modules/oxide-components/src/main/ts/components/inlinetoolbar/InlineToolbar.stories.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+import { Button } from '../button/Button';
+
+import { InlineToolbar } from './InlineToolbar';
+
+const meta = {
+  title: 'components/InlineToolbar',
+  component: InlineToolbar,
+  parameters: {
+    // Optional parameter to center the component in the Canvas. More info: https://storybook.js.org/docs/configure/story-layout
```
spocke: I see that we copy this over and over. I wonder if we should I think it just becomes noisy if you don't know what center is then read the docs. Maybe something we need to discuss in general.

modules/oxide-components/src/main/ts/components/inlinetoolbar/InlineToolbar.tsx
```diff
@@ -0,0 +1,179 @@
...
+    toolbarRef,
+    sinkRef,
+    triggerRef,
+    close,
+    persistent
+  } = useInlineToolbarContext();
+
+  useEffect(() => {
+    const handleKeyDown = (event: KeyboardEvent) => {
+      if (KeyMatch.inSet(Keys.ESCAPE)(event) && isOpen) {
```
spocke: Is this to only check if the event is an escape key. Then I think we should maybe have a separate predicate then use this partially applied thing.

modules/oxide-components/src/main/ts/components/inlinetoolbar/InlineToolbar.tsx
```diff
@@ -0,0 +1,179 @@
...
+        close();
+      }
+    };
+
+    document.addEventListener('keydown', handleKeyDown);
```
spocke: Is this to much to register the escape handler on the whole document?

modules/oxide-components/src/main/ts/components/inlinetoolbar/InlineToolbar.tsx
```diff
@@ -0,0 +1,179 @@
...
+                    clientY >= toolbarRect.top &&
+                    clientY <= toolbarRect.bottom;
+
+        const clickedTrigger =
+                    clientX >= triggerRect.left &&
+                    clientX <= triggerRect.right &&
+                    clientY >= triggerRect.top &&
+                    clientY <= triggerRect.bottom;
+
+        if (!clickedToolbar && !clickedTrigger) {
```
spocke, line 117: Couldn't this be done with toolbarRef.current.contains(event.target) or something?
I also wonder if this needs to be a general concept. We will be clicking away lots of things. Drop down menus, toolbars etc.

spocke (approved):

modules/oxide-components/src/main/ts/components/inlinetoolbar/InlineToolbar.tsx
```diff
@@ -0,0 +1,179 @@
...
+  InlineToolbarProps,
+  TriggerProps,
+  ToolbarProps
+} from './InlineToolbarTypes';
+
+const InlineToolbarContext = createContext<InlineToolbarContextValue | null>(null);
+
+const useInlineToolbarContext = () => {
+  const context = useContext(InlineToolbarContext);
+  if (!context) {
```
spocke: Should we use Type.isNonNullable or something here?

modules/oxide-components/src/test/ts/browser/components/InlineToolbar.spec.tsx
tiny-ben-tran: I misunderstood how the component was to be used with the `<Trigger>` part. I agree that the state management should be contained inside `<InlineToolbar>` to avoid re-rendering of the parent. 

However, subcomponents should not be tightly coupled by context so for instance, `<Toolbar>` can be used outside of `InlineToolbarContext`.  Here are my 2 cents on how to expose context and make subcomponents more generic. 
```
<InlineToolbar>
<InlineToolbarContext.Consumer>
	// Rename triggerRef to anchorRef to reflect the relationship between the button and toolbar
  ({anchorRef, open} => (
    <Button ref={anchorRef} onClick={open}/>
  ))
</InlineToolbarContext.Consumer>
<InlineToolbarContext.Consumer>
	({  isOpen,
	    toolbarRef,
	    sinkRef,
	    triggerRef,
	    close,
	    persistent
	} => (
		<Toolbar {context as props and other props}>
			<>Toolbar content</>
		</Toolbar>
	))
	</InlineToolbarContext.Consumer>
))
</InlineToolbar>
```
In this approach, `<Trigger>` is no longer needed since Button can access to `open` function directly. `<Toolbar>` becomes more generic as what it needs can be passed through props.

I know `Context.Consumer` is an old way of accessing context and is not recommended by React docs but I think it is helpful in this case. 

We should aim to make UI components as generic as possible so they can be reused in as many places as possible, especially simple and small UI components. Using `context` to share data has made development easier but everything has a price.

kimwoodfield: Thanks everyone for the thoughtful discussion! I think we're debating valid architectural preferences rather than a technical issue since:
- The current implementation works and has test coverage
- It follows an existing pattern in suggested edits
- Multiple reviewers have different but valid perspectives

I'm going to resolve this conversation and suggest that we discuss component architecture conventions in our next team sync. We can establish team standards there and refactor if needed. 

For now, I need to unblock and move forward with the next stories. Happy to revisit in a follow-up PR once we have team alignment on patterns.

modules/oxide-components/src/main/ts/components/inlinetoolbar/InlineToolbar.tsx
kimwoodfield: Good point, there's no sense in recreating them on each render. Fixed.

modules/oxide-components/src/main/ts/components/inlinetoolbar/InlineToolbar.tsx
kimwoodfield: Nice suggestion, that's definitely an improvement. Fixed.

modules/oxide-components/src/main/ts/components/inlinetoolbar/InlineToolbar.tsx
kimwoodfield: Agreed, I am not sure what happened there. Fixed.

modules/oxide-components/src/main/ts/components/inlinetoolbar/InlineToolbar.stories.tsx
kimwoodfield: Yep, it's just a leftover default Storybook comment. I've removed it.

modules/oxide-components/src/main/ts/components/inlinetoolbar/InlineToolbar.tsx
kimwoodfield: Refactored to use `useSpecialKeyNavigation` hook instead.

modules/oxide-components/src/main/ts/components/inlinetoolbar/InlineToolbar.tsx
kimwoodfield: Good catch, I've refactored to focus the toolbar on open and use `useSpecialKeyNavigation`. 
Now it listens only on the focused element.

modules/oxide-components/src/main/ts/components/inlinetoolbar/InlineToolbar.tsx
kimwoodfield: Good point, switched to `contains` which is much cleaner as you say. Thanks for the recommendation

modules/oxide-components/src/main/ts/components/inlinetoolbar/InlineToolbar.tsx
kimwoodfield: Done!

PR #10662, opened 2025-10-16
TINY-13084: Added new agar http mocking

spocke:
Related Ticket: TINY-13084

Description of Changes:

- Added new `Http` module to `agar` that enables mocking of http requests
- Added new `agar-sw` package that is just the service worker that is served by `bedrock`

Note: This branch fails to build since `bedrock`​ needs to be updated first

Pre-checks:

Review:

- [x] Milestone set

## Summary by CodeRabbit

- **New Features**
    - Service worker HTTP mocking with client control, start/stop messaging, and a test hook; consolidated public HTTP exports and response helpers (JSON, text, blob, chunked).
    - Client-side mock manager, HTTP routing (path params, splat, dynamic status), stateful handlers, streaming responses, and file upload support.
- **Tests**
    - Browser test suites covering mocking flows, streaming, status variations, stateful routes, and uploads.
- **Chores**
    - New module packaging, build/lint/tsconfig/project references, scripts, .gitignore and CI script updates.


spocke:
* **#10662** <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10662?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10662?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10663** <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10663?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

modules/agar-sw/src/main/ts/core/Shared.ts
```diff
@@ -0,0 +1,72 @@
+// Keep this file in sync between the `agar` and `agar-sw` cross importing provded too painful
```
TheSpyder: can we script it at least?

That might give us an opportunity to add automation like failing the build if they're different.

TheSpyder (approved):
I haven't looked at the details too closely, but we can tweak things later if they're wrong.

modules/agar-sw/src/main/ts/core/Shared.ts
spocke: Yes, I can add a simple compare in the Gruntfile for the `agar-sw` project good idea. I wanted these to be shared between the two projects but TS really makes this painful since you can't even import a DOM based dependency into a WebWorker project without it starting to conflict on it's global types. 😞

TheSpyder: Right, we'd need some sort of _third_ project with no lib dependencies for `agar` and `agar-sw` to import. Very painful.

It's almost like implicit dependencies are annoying and `dom-globals` had it right all along...

spocke: Yeah having a project just to have the common types would be very annoying.

modules/agar-sw/src/main/ts/core/ServiceWorker.ts
```diff
@@ -0,0 +1,152 @@
...
+    if (reqUrl.origin !== self.location.origin) {
+      return;
+    }
+
+    if (isMockingTargetRequest(reqUrl.pathname)) {
+      event.respondWith(handleRequest(event.clientId, event.request));
+    }
+  });
+
+  self.addEventListener('message', (event) => {
```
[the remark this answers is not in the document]
spocke: This is used in a service worked used for local testing not something that will be shipped to production. We can't really verify the origin since the point of this message is to accept request for any page that has loaded that service worker and start mocking.

modules/agar/src/main/ts/ephox/agar/http/Shared.ts
```diff
@@ -0,0 +1,74 @@
...
+  readonly buffer: ArrayBuffer;
+}
+
+export interface MockedResponseBodyDoneMessage {
+  readonly type: 'AGAR_MOCKED_RESPONSE_BODY_DONE';
+}
+
+export const mockPrefix = '/custom/';
+
+const isMessageOfType = <T extends { type: string }>(type: T['type']) => (data: unknown): data is T =>
```
michalnieruchalski-tiugo, line 54: You could use [discriminating unions](https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html#discriminating-unions) since all the types have type property.

```
type MockedMessage = MockingStartMessage | MockingStartedMessage ... | MockedResponseBodyDoneMessage
```

Then your helper functions would be:
```
export const isMockingStartMessage = (message: MockedMessage) => message.type === 'AGAR_MOCKING_START'
```
and typescript will know that in this case:
```
if (isMockingStartMessage(a)) {
  // here `a` is of type `MockingStartMessage`
}
```

michalnieruchalski-tiugo (approved):

modules/agar/src/main/ts/ephox/agar/http/Shared.ts
spocke: Oh so TS will infer the `param is T` in that case. Will try that.

spocke: No wait that wouldn't work in this case the type signature is `<T>(data: unknown) => data is T` we don't know that data is a message of one if the types in the union we don't know what it is so this is a validation and a type predicate in the same function.

PR #10664, opened 2025-10-16
TINY-13044: Add initial position to floating sidebar

michalnieruchalski-tiugo:
Related Ticket: TINY-13044

Description of Changes:
* Add `initialPosition` property to `FloatingSidebar`. It allows you to set the initialPosition of sidebar (before dragging)
* It has `origin` property. `topleft` means top left corner has to be in specified point. `topright` measn top right corner has to be in specified point and so on...

Pre-checks:
~~* [ ] Changelog entry added~~
* [x] Tests have been added (if applicable)
* [x] Branch prefixed with `feature/`, `hotfix/` or `spike/`

Review:
~~* [ ] Milestone set~~
~~* [ ] Docs ticket created (if applicable)~~

GitHub issues (if applicable):

## Summary by CodeRabbit

* **New Features**
  * Added configurable initial positioning for the floating sidebar. Users can now specify x, y coordinates and select from four origin points (top-left, top-right, bottom-left, bottom-right) to control sidebar placement.

PR #10657, opened 2025-10-14
TINY-13052: implement FloatingSidebar component

michalnieruchalski-tiugo:
Related Ticket: TINY-13052

Description of Changes

- `Draggable` now exposes `ref` to element with `transform` property
- `Draggable` now accepts additional props that can be applied to element with `transform` property
- `Draggable` now follows `Draggable.Root` pattern
- Implement `FloatingSidebar` component, using `Draggable` under the hood
- Add necessery styles to `oxide`

Pre-checks:
~~* [ ] Changelog entry added~~

- [x] Tests have been added (if applicable)
- [x] Branch prefixed with `feature/`, `hotfix/` or `spike/`

Review:
~~* [ ] Milestone set~~
~~* [ ] Docs ticket created (if applicable)~~

GitHub issues (if applicable):

## Summary by CodeRabbit

- **New Features**
  - FloatingSidebar component with draggable header, open/close controls, configurable height; Storybook example added.

- **Updates**
  - Draggable API reorganized into Root and Handle for clearer composition.
  - Draggable props now accept standard div attributes and an optional popover mode.
  - Module exports updated to include FloatingSidebar.

- **Styling**
  - Theme styles for FloatingSidebar (responsive sizing, shadow, layout) and CSS typing for its height variable.

- **Tests**
  - Browser tests and drag utilities added; test build includes new test paths.

PR #10669, opened 2025-10-20
TINY-13102: Allow clickable elements in Draggable.Handle

michalnieruchalski-tiugo:
Related Ticket: TINY-13102

Description of Changes:
* Allow clickable elements in `Draggable.Handle` by setting `setPointerCapture` on `event.target` instead of on `handlerRef`
* `event.target` element has to be inside `Draggable.Handle` (we wouldn't capture the event otherwise)
* If both `pointerdown` and `pointerup` events are called on the same element, then `click` event is executed on element

Pre-checks:

Review:
~~* [ ] Milestone set~~

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where buttons and interactive elements inside draggable handles could not be clicked without interfering with the drag functionality.

* **Tests**
  * Added test coverage to verify that interactive elements within draggable handles function correctly.

### DIFF
--- a/modules/oxide-components/src/test/ts/browser/components/Draggable.spec.tsx
+++ b/modules/oxide-components/src/test/ts/browser/components/Draggable.spec.tsx
@@ -218,7 +218,7 @@ describe('browser.draggable.Draggable', () => {
     );
 
     const { getByTestId } = render(TestElementWithButton, { wrapper: Wrapper });
-    const button = getByTestId('button-in-handle').element() as HTMLButtonElement;
+    const button = getByTestId('button-in-handle').element();
     await expect.element(button).toBeDefined();
 
     await userEvent.click(button);


michalnieruchalski-tiugo:
* **#10669** <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10669?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10669?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10664** <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10664?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10657** <a href="https://app.graphite.dev/github/pr/tinymce/tinymce/10657?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

ztomaszyk (approved):

modules/oxide-components/src/test/ts/browser/components/Draggable.spec.tsx
```diff
@@ -197,4 +198,31 @@ describe('browser.draggable.Draggable', () => {
...
+            <div data-testid={draggableHandleTestId} style={{ width: '100%', height: 50, backgroundColor: 'black' }}>
+              <button data-testid="button-in-handle" onClick={handleClick}>Click me</button>
+            </div>
+          </Draggable.Handle>
+        </div>
+      </Draggable.Root>
+    );
+
+    const { getByTestId } = render(TestElementWithButton, { wrapper: Wrapper });
+    const button = getByTestId('button-in-handle').element() as HTMLButtonElement;
```
spocke: I wonder if we should abstract stuff like this away to avoid the magic cast I guess technically it would still be a magic cast but not using the `foo as bar` pattern.

spocke: Not in this ticket but maybe something we need to start think about if we start having lots and lots of these things.

michalnieruchalski-tiugo: If `getByTestId` would be a generic then we could assume the type. But it's returning an `Element`. So we will need special cast anyway.

Anyway I've just noticed it's actually not needed here. So I'm going to remove the cast

PR #10672, opened 2025-10-21
TINY-13068: Add CSS `anchor()` for position anchoring

ArvinJ-H:
Related Ticket: 
TINY-13068

Description of Changes:

Using CSS `anchor()` for position anchor the contextToolbar

1. Generates unique anchor name for each toolbar instance
2. Sets `anchor-name` CSS property on trigger element
3. Links toolbar to anchor via `position-anchor` property
4. Calculates position using `anchor()` and `anchor-size()` CSS functions
5. Applies transform for center/bottom alignment
6. Enables automatic flipping when toolbar would overflow viewport

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/oxide-components/src/main/ts/components/inlinetoolbar/InlineToolbar.tsx
+++ b/modules/oxide-components/src/main/ts/components/inlinetoolbar/InlineToolbar.tsx
@@ -184,9 +184,7 @@ const Toolbar: FC<ToolbarProps> = ({
     onMouseDown?.(event);
   }, [ onMouseDown ]);
 
-  const toolbarClasses = [ 'tox-inline-toolbar', className ]
-    .filter(Boolean)
-    .join(' ');
+  const toolbarClasses = `tox-inline-toolbar${Type.isNonNullable(className) ? ` ${className}` : ''}`;
 
   return (
     isOpen &&


modules/oxide-components/src/main/ts/components/inlinetoolbar/InlineToolbar.stories.tsx
```diff
@@ -101,9 +132,14 @@ Contains the toolbar content (buttons, text, etc.). Renders as a portal into the
 export default meta;
 type Story = StoryObj<typeof meta>;
 
-// TODO: Add more stories (TINY-13065)
-
```
ArvinJ-H, line 21: @kimwoodfield we still need this?

kimwoodfield: @ArvinJ-H  Yes, let's keep that until we have finished documenting both positioning and keyboard navigation to close that.

modules/oxide-components/src/main/ts/components/inlinetoolbar/InlineToolbar.tsx
```diff
@@ -85,58 +122,98 @@ const Toolbar: FC<ToolbarProps> = ({
...
-    const top = triggerRect.bottom - sinkRect.top;
-    const left = triggerRect.left - sinkRect.left;
+  useEffect(() => {
+    if (!isOpen || !triggerRef.current || !toolbarRef.current) {
+      return;
+    }
+
+    const trigger = triggerRef.current;
+    const toolbar = toolbarRef.current;
+    const anchorElement = (trigger.firstElementChild as HTMLElement) ?? trigger;
```
kimwoodfield, line 153: Do we need `detectAnchorPosition` at all? For diffs, could we just default to below and left and let `position-try-fallbacks` automatically flip when needed?

modules/oxide-components/src/main/ts/components/inlinetoolbar/InlineToolbar.tsx
```diff
@@ -85,58 +122,98 @@ const Toolbar: FC<ToolbarProps> = ({
...
+
+    const transform = getTransformValue(isCenter, isBottom);
+    if (transform) {
+      toolbar.style.setProperty('transform', transform);
+    }
+
+    toolbar.style.setProperty('position-try-fallbacks', 'flip-block, flip-inline, flip-block flip-inline');
+
+    return () => {
+      anchorElement.style.removeProperty('anchor-name');
```
kimwoodfield:
```suggestion
  anchorElement.style.removeProperty('anchor-name');
  
  Arr.each(['position-anchor', 'top', 'left', 'transform', 'position-try-fallbacks'], (property) => {
    toolbar.style.removeProperty(property);
  });
```

We could iterate over the properties to reduce the repetition.

modules/oxide-components/src/main/ts/components/inlinetoolbar/InlineToolbar.stories.tsx
```diff
@@ -75,9 +76,39 @@ Contains the toolbar content (buttons, text, etc.). Renders as a portal into the
...
+## Positioning Anchoring Support
+
+- ✅ **Chrome 125+**
+- ✅ **Edge 125+**
+- ✅ **Safari 26+**
+- ❌ **Firefox** (not yet supported)
+
+## Default Behavior
+
+By default, the InlineToolbar anchor itself to the **bottom left** of the trigger element (left alignment). It automatically detects whether to position above or below, and left/center/right based on the trigger's position in the viewport.
```
kimwoodfield:
```suggestion
By default, the InlineToolbar anchors to the bottom left of the trigger element. It can position above/below or left/center/right based on the trigger element's inline positioning styles. The position-try-fallbacks CSS property automatically flips the toolbar when it would overflow the viewport.
```

modules/oxide/src/less/theme/components/inline-toolbar/inline-toolbar.less
```diff
@@ -0,0 +1,26 @@
...
+    // CSS variable for gap between anchor and toolbar
+    --inline-toolbar-gap: 6px;
+
+    // CSS Anchor Positioning (required for anchor() to work)
+    position: absolute;
+
+    // Enable automatic flip when toolbar would overflow viewport, very new, not recongnized by stylelint yet
+    // stylelint-disable-next-line property-no-unknown -- CSS Anchor Positioning spec (Chrome 125+)
+    position-try-fallbacks: flip-block, flip-inline, flip-block flip-inline;
+    z-index: 9999;
```
MitchC1999: Is there a smaller number we could use? I'm imagining we could support dropdowns from this toolbar at some point, where they would need to be above this.

MitchC1999 (approved):

modules/oxide/src/less/theme/components/inline-toolbar/inline-toolbar.less
ArvinJ-H: yeah I remember this came up before, I'll have a look and see what we did for other `z-index`

modules/oxide-components/src/main/ts/components/inlinetoolbar/InlineToolbar.tsx
```diff
@@ -85,58 +122,102 @@ const Toolbar: FC<ToolbarProps> = ({
...
+    /* eslint-disable-next-line react-hooks/exhaustive-deps -- triggerRef/toolbarRef are stable ref objects */
+  }, [ anchorName, isOpen ]);
+
+  const handleMouseDown = useCallback<MouseEventHandler<HTMLDivElement>>((event) => {
+    event.preventDefault();
+    onMouseDown?.(event);
+  }, [ onMouseDown ]);
+
+  const toolbarClasses = [ 'tox-inline-toolbar', className ]
+    .filter(Boolean)
```
ltrouton: What is this filter for? More curious as I haven't really seen this technique before

modules/oxide-components/src/main/ts/components/inlinetoolbar/InlineToolbar.tsx
```diff
@@ -85,58 +122,102 @@ const Toolbar: FC<ToolbarProps> = ({
...
+
+    toolbar.style.setProperty('position-try-fallbacks', 'flip-block, flip-inline, flip-block flip-inline');
+
+    return () => {
+  anchorElement.style.removeProperty('anchor-name');
+  
+  Arr.each(['position-anchor', 'top', 'left', 'transform', 'position-try-fallbacks'], (property) => {
+    toolbar.style.removeProperty(property);
+  });
+      toolbar.style.removeProperty('position-anchor');
```
ltrouton: Is there some duplication here - does the Arr call above already delete the necessary properties?

tiny-ben-tran: Should `CSS` lib from `Sugar` be used to manipulate elements styles?

modules/oxide-components/src/main/ts/components/inlinetoolbar/InlineToolbar.tsx
```diff
@@ -85,58 +122,102 @@ const Toolbar: FC<ToolbarProps> = ({
...
+    const trigger = triggerRef.current;
+    const toolbar = toolbarRef.current;
+    const anchorElement = (trigger.firstElementChild as HTMLElement) ?? trigger;
+
+    anchorElement.style.setProperty('anchor-name', anchorName);
+    toolbar.style.setProperty('position-anchor', anchorName);
+
+    const gap = toolbar.ownerDocument?.defaultView
+      ? toolbar.ownerDocument.defaultView.getComputedStyle(toolbar).getPropertyValue('--inline-toolbar-gap') || '6px'
+      : '6px';
```
ltrouton: I believe we should be able to use `Css` from sugar for this rather than manually doing `.getComputedStyle(toolbar).getPropertyValue`. 

I guess this applies to the whole file where we would typically use Sugar for setting/removing Css. Or are we generally avoiding using Sugar for oxide components?

modules/oxide-components/src/main/ts/components/inlinetoolbar/InlineToolbar.tsx
```diff
@@ -60,21 +79,39 @@ const Root: FC<InlineToolbarProps> = ({
...
-const Trigger: FC<TriggerProps> = ({ children }) => {
+const Trigger: FC<TriggerProps> = ({
+  children,
+  onClick,
+  onMouseDown,
+  ...rest
+}) => {
   const { open, triggerRef } = useInlineToolbarContext();
+  const handleClick = useCallback<MouseEventHandler<HTMLDivElement>>((event) => {
+    open();
```
ltrouton, line 73: Just want to double check we don't have to call `event.preventDefault()` for the click handler. Just noticed we do it for the other mouse handlers

modules/oxide-components/src/main/ts/components/inlinetoolbar/InlineToolbar.stories.tsx
kimwoodfield: I've pushed a commit to add it back.

modules/oxide/src/less/theme/components/inline-toolbar/inline-toolbar.less
kimwoodfield: Nice catch @MitchC1999! I've updated it to use `@z-index-menu` from Oxide (1150): https://github.com/tinymce/tinymce/blob/main/modules/oxide/src/less/theme/globals/global-variables.less#L75

This keeps it consistent with other menus/dropdowns but let me know if you think it could be adjusted further

modules/oxide-components/src/main/ts/components/inlinetoolbar/InlineToolbar.tsx
kimwoodfield: Good spotting! There was a suggestion committed directly which borked things and also didn't include the correct import. I've fixed that.

modules/oxide-components/src/main/ts/components/inlinetoolbar/InlineToolbar.tsx
kimwoodfield: It's typically used to remove falsy values from an array. In our use case, it will prevent us from rendering something like `className="tox-inline-toolbar undefined"` in the DOM.

modules/oxide-components/src/main/ts/components/inlinetoolbar/InlineToolbar.tsx
TheSpyder: It would be nice (later, not now) if the elements we are positioning don't need to contain position code at all. It should be either coordinated by a parent or global layout system, or delegated to a callout function like `useSpecialKeyNavigation`.

That way after we add the required JS fallback we do it in one place not 50.

modules/oxide/src/less/theme/components/inline-toolbar/inline-toolbar.less
TheSpyder: We should, at some point, dump all of this z-index mess - along with the need for a sink and separate root - and use `popover`. It's already widely supported, unlike anchors. I might start a thread in slack about that.

modules/oxide-components/src/main/ts/components/inlinetoolbar/InlineToolbar.tsx
kimwoodfield: I can see that we've already used it in a few plugins which use React and `@ephox/sugar` is already listed as a dependency, so I don't see any compelling reason not to make use of it. I've updated this file to start using `Css` from Sugar. Thanks for raising it!

modules/oxide-components/src/main/ts/components/inlinetoolbar/InlineToolbar.tsx
kimwoodfield: It's a good point - I've pushed a commit to begin using it and responded to a similar comment here: https://github.com/tinymce/tinymce/pull/10672#discussion_r2450170690

modules/oxide-components/src/main/ts/components/inlinetoolbar/InlineToolbar.tsx
kimwoodfield: Nah this one is fine. The other mouse handlers serve different purposes. The mouseDown preventDefault prevents unwanted side effects like text selection and focus changes. The click should propagate normally to trigger the open action since there's no default div click behaviour to prevent

modules/oxide-components/src/main/ts/components/inlinetoolbar/InlineToolbar.tsx
kimwoodfield: I've pushed a commit to simplify this.

modules/oxide-components/.storybook/main.ts
```diff
@@ -25,8 +26,13 @@ const config: StorybookConfig = {
     defaultName: "Documentation"
   },
   viteFinal: (config) => {
-    config.server ??= {}
+    config.server ??= {};
     config.server.allowedHosts = ['host.docker.internal'];
+    config.server.fs ??= {};
+    config.server.fs.allow = [
+      // Allow the modules directory (parent of oxide-components)
```
TheSpyder: why was it necessary to add this?

modules/oxide-components/src/main/ts/components/inlinetoolbar/InlineToolbar.tsx
TheSpyder: That's a code smell, possibly an anti-pattern. Either change the `className` type so it can't be undefined, or use a ternary to append it to the array only when necessary (this is a pattern I think I've seen elsewhere).

There may even be code elsewhere in oxide-components that handles this? I'm sure I've seen examples where the class name is required to match a list imported from oxide.

modules/oxide-components/src/main/ts/components/inlinetoolbar/InlineToolbarTypes.tsx
```diff
@@ -16,10 +16,6 @@ export interface InlineToolbarContextValue {
   readonly persistent: boolean;
 }
 
-export interface ToolbarProps {
-  readonly children: ReactNode;
-}
+export interface ToolbarProps extends HTMLAttributes<HTMLDivElement> {}
```
TheSpyder: That's a very big expansion in allowed properties, from a restricted list to "anything in the div element". Is it really necessary?

TheSpyder (approved):

modules/oxide-components/.storybook/main.ts
kimwoodfield: It was probably something Arvin thought was needed but I removed it and it looks like Vite/Storybook handles the Oxide imports fine without it. Since it works without it, I've removed it.

modules/oxide-components/src/main/ts/components/inlinetoolbar/InlineToolbarTypes.tsx
kimwoodfield: Definitely not necessary. I've updated that by restricting it to only the props we explicitly need (e.g. children, className, etc.) rather than just accepting all HTMLAttributes. Thanks for catching it.

modules/oxide-components/src/main/ts/components/inlinetoolbar/InlineToolbar.tsx
kimwoodfield: Fair point - switched to template literal for clarity.

TheSpyder: nested template literals are like nested ternaries, they break my brain and should be banned 🤣

PR #10683, opened 2025-10-24
TINY-13066: Add keyboard navigation and toolbar groups to InlineToolbar

kimwoodfield:
Related Ticket: [TINY-13066](https://ephocks.atlassian.net/browse/TINY-13066)

Description of Changes:
* Added keyboard navigation to the InlineToolbar component.
* Created an additional `InlineToolbar.Group` component for grouping toolbar buttons. 
* `Tab` to navigate between toolbar groups.
* `arrow keys` to navigate within a group.
* `Enter` to execute the focused button.
* `Escape` to close the toolbar (unless `persistent={true}`).
* Updated all the Storybook stories and documentation.
* Included tests for the keyboard navigation behaviour.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

[TINY-13066]: https://ephocks.atlassian.net/browse/TINY-13066?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/oxide-components/src/main/ts/components/inlinetoolbar/InlineToolbar.stories.tsx
+++ b/modules/oxide-components/src/main/ts/components/inlinetoolbar/InlineToolbar.stories.tsx
@@ -180,7 +180,7 @@ export const Persistent: Story = {
   },
   render: () => {
     return (
-      <div className='tox' style={{ position: 'relative' as const }}>
+      <div className='tox' style={{ position: 'relative' }}>
         <InlineToolbar.Root persistent={true}>
           <InlineToolbar.Trigger>
             <div style={{ backgroundColor: 'blue', padding: '10px' }}>
@@ -324,7 +324,7 @@ export const Corners: Story = {
           {triggerPositions.map((pos) => (
             <InlineToolbar.Root key={pos.id} persistent={false}>
               <InlineToolbar.Trigger>
-                <div style={{ position: 'absolute' as const, display: 'inline-flex', ...pos.style }}>
+                <div style={{ position: 'absolute', display: 'inline-flex', ...pos.style }}>
                   <Button>{pos.label}</Button>
                 </div>
               </InlineToolbar.Trigger>


ArvinJ-H (approved):

modules/oxide-components/src/main/ts/components/inlinetoolbar/InlineToolbar.stories.tsx
```diff
@@ -60,20 +62,33 @@ The provider component that manages toolbar state.
...
+### Group
+Groups related buttons together for keyboard navigation. Buttons within a group are navigated with arrow keys, while Tab moves between groups.
 
 ## Behaviour
 - Click trigger to open the toolbar
-- Press Escape to close the toolbar
+- Press **Escape** to close the toolbar (unless \`persistent=true\`)
+- Press **Tab** to navigate to the first button in the next group
+- Press **Shift+Tab** to navigate to the first button in the previous group
+- Press **Arrow keys** to navigate between buttons within the current group
```
TheSpyder, line 75: Funny story, we're planning to drop this behaviour in the near future. In modern toolbars arrows just move left and right regardless of visual group; tab is not captured by the toolbar to move between groups.

But this does match our other UI for now so that's fine.

modules/oxide-components/src/main/ts/components/inlinetoolbar/InlineToolbar.stories.tsx
```diff
@@ -163,7 +180,7 @@ export const Persistent: Story = {
   },
   render: () => {
     return (
-      <div className="tox" style={{ position: 'relative' }}>
+      <div className='tox' style={{ position: 'relative' as const }}>
```
TheSpyder: why is this required here and not on similar lines below 🤔

modules/oxide-components/src/main/ts/components/inlinetoolbar/InlineToolbar.tsx
```diff
@@ -101,22 +102,60 @@ const Toolbar: FC<ToolbarProps> = ({
...
     const element = toolbarRef.current;
     if (Type.isNonNullable(element)) {
-      isOpen ? element.showPopover() : element.hidePopover();
+      if (isOpen) {
+        element.showPopover();
+        // Defer focus to next event loop tick to ensure
+        // it runs after Popover API's focus management
+        setTimeout(() => {
+          element.focus();
+        }, 0);
```
TheSpyder, line 116: I'd prefer not to do this, but it probably depends where the toolbar is added to the DOM. I suppose so long as it's in the sink we need this but I'm hoping to change that soon.

TheSpyder (approved):

modules/oxide-components/src/main/ts/components/inlinetoolbar/InlineToolbar.stories.tsx
kimwoodfield: It's not required, it was just mistakenly auto-completed when updating the className to use single quotes. No harm in leaving it but better to keep it consistent. Fixed.

modules/oxide-components/src/main/ts/components/inlinetoolbar/InlineToolbar.stories.tsx
kimwoodfield: It's funny that you mention that because that's how I had originally expected it to work until I dug into the existing implementation... will be good to align it with modern standards!

modules/oxide-components/src/main/ts/components/inlinetoolbar/InlineToolbar.tsx
kimwoodfield: Yeah not ideal, the Popover API's internal focus management runs when `showPopover` is called and overrides our manual `focus` without the setTimeout. Tried a few other approaches and they all had the same issue. Happy to revisit that later and come up with a more elegant solution.

TheSpyder: What I meant was ideally we'd change where we attach the toolbar so that the normal popover focus management operates correctly with our new toolbar. But yes we can revisit later.

PR #10686, opened 2025-10-27
TINY-13059: Add ContextToolbar component to oxide-components

kimwoodfield:
Related Ticket: [TINY-13059](https://ephocks.atlassian.net/browse/TINY-13059)

Description of Changes:

- Built a new ContextToolbar React component for oxide-components that provides a generic, reusable toolbar anchored to trigger elements. This is designed to work with a variety of use cases and plugins.
- The component uses CSS anchor positioning for automatic placement with viewport-aware flipping, the Popover API for layering and focus management, and includes keyboard navigation support (tabs between groups, arrow keys to navigate within groups, enter to activate and escape to close). It's split into subcomponents (Root, Trigger, Toolbar and Group), following existing patterns from the codebase and other components.
- Storybook documentation covers usage examples, positioning behaviour, keyboard navigation and the `persistent` mode option.

Pre-checks:

- [x] Changelog entry added

Review:

- [x] Milestone set

## Summary by CodeRabbit

- **New Features**
  - Introduced a Context Toolbar (Root, Trigger, Toolbar, Group) with anchored toolbars, persistent/dismissible modes, focus management, and keyboard navigation (Tab, Arrow, Escape).
- **Documentation**
  - Added comprehensive Storybook examples demonstrating usage, variants, and nine-position anchoring.
- **Tests**
  - Added browser tests for rendering, interactions, keyboard navigation, persistence, and accessibility.
- **Style**
  - Added theme styles and layout rules for the Context Toolbar.
- **Chores**
  - Exported ContextToolbar from the package public API.

[TINY-13059]: https://ephocks.atlassian.net/browse/TINY-13059?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/oxide-components/src/main/ts/components/contexttoolbar/ContextToolbar.stories.tsx
+++ b/modules/oxide-components/src/main/ts/components/contexttoolbar/ContextToolbar.stories.tsx
@@ -62,7 +62,7 @@ The provider component that manages toolbar state.
 Wraps the element that opens the toolbar when clicked.
 
 ### Toolbar
-Contains the toolbar content and groups. Uses the Popover API for automatic layering and positioning.
+Contains the toolbar content and groups. Uses the Popover API (popover='manual') for layering and positioning, with manual control over dismissal behavior.
 
 ### Group
 Groups related buttons together for keyboard navigation. Buttons within a group are navigated with arrow keys, while Tab moves between groups.
@@ -85,8 +85,8 @@ Groups related buttons together for keyboard navigation. Buttons within a group 
   - \`Enter\`: Execute focused button
   - \`Escape\`: Close the toolbar (unless \`persistent=true\`)
 - **Focus Management**: Toolbar automatically receives focus when opened, with the first button in the first group focused
-- **Click Outside**: Click outside the toolbar to dismiss it (unless \`persistent=true\`)
-- **Persistent Mode**: Use \`persistent=true\` for toolbars that require explicit dismissal (e.g. forms, critical actions)
+- **Click Outside**: Click outside the toolbar to dismiss it (unless \`persistent=true\`). The component handles click-outside detection.
+- **Persistent Mode**: Use \`persistent=true\` for toolbars that require explicit dismissal (e.g. forms, critical actions). Disables both Escape key and click-outside dismissal.
 - **ARIA**: Toolbar uses \`role="toolbar"\` and groups use \`role="group"\` for proper accessibility semantics
 - **Toolbar Groups**: Matches TinyMCE core context toolbar behavior for consistency
 
--- a/modules/oxide-components/src/main/ts/components/contexttoolbar/ContextToolbar.tsx
+++ b/modules/oxide-components/src/main/ts/components/contexttoolbar/ContextToolbar.tsx
@@ -93,8 +93,6 @@ const Toolbar: FC<ToolbarProps> = ({
     persistent
   } = useContextToolbarContext();
 
-  const popoverMode = persistent ? 'manual' : 'auto';
-
   useEffect(() => {
     const element = toolbarRef.current;
     if (Type.isNonNullable(element)) {
@@ -120,23 +118,7 @@ const Toolbar: FC<ToolbarProps> = ({
     };
   }, [ isOpen, toolbarRef ]);
 
-  // Listen for popover auto-dismiss (e.g. when user presses Escape)
-  // and sync our state
-  useEffect(() => {
-    const element = toolbarRef.current;
-    if (Type.isNonNullable(element)) {
-      const handleToggle = (event: Event) => {
-        const toggleEvent = event as ToggleEvent;
-        if (toggleEvent.newState === 'closed' && isOpen) {
-          close();
-        }
-      };
-      element.addEventListener('toggle', handleToggle);
-      return () => element.removeEventListener('toggle', handleToggle);
-    }
-  }, [ isOpen, close, toolbarRef ]);
-
-  // Only allow Escape to close if not `persistent={true}`
+  // Handle Escape key to close (unless persistent={true})
   KeyboardNavigationHooks.useSpecialKeyNavigation({
     containerRef: toolbarRef,
     onEscape: persistent ? undefined : close,
@@ -224,7 +206,7 @@ const Toolbar: FC<ToolbarProps> = ({
     <div
       ref={toolbarRef}
       // @ts-expect-error - TODO: Remove this expect error once we've upgraded to React 19+ (TINY-13129)
-      popover={popoverMode}
+      popover='manual'
       tabIndex={-1}
       className='tox-context-toolbar'
       style={{
--- a/modules/oxide-components/src/test/ts/browser/components/ContextToolbar.spec.tsx
+++ b/modules/oxide-components/src/test/ts/browser/components/ContextToolbar.spec.tsx
@@ -157,6 +157,37 @@ describe('browser.ContextToolbar.ContextToolbar', () => {
     await expect.element(toolbar).toBeVisible();
   });
 
+  it('TINY-13066: Should not close toolbar on click outside when persistent={true}', async () => {
+    const { getByTestId } = render(
+      <Fragment>
+        <div className='tox' style={{ position: 'relative' }}>
+          <ContextToolbar.Root persistent={true}>
+            <ContextToolbar.Trigger>
+              <div data-testid={triggerTestId}>Click Me</div>
+            </ContextToolbar.Trigger>
+            <ContextToolbar.Toolbar>
+              <ContextToolbar.Group>
+                <div data-testid={toolbarTestId}>Toolbar Content</div>
+              </ContextToolbar.Group>
+            </ContextToolbar.Toolbar>
+          </ContextToolbar.Root>
+        </div>
+      </Fragment>,
+      { wrapper: Wrapper }
+    );
+
+    const trigger = getByTestId(triggerTestId);
+    await trigger.click();
+
+    const toolbar = getByTestId(toolbarTestId);
+    await expect.element(toolbar).toBeVisible();
+
+    // Click outside
+    await page.elementLocator(document.body).click({ position: { x: 0, y: 0 }});
+
+    await expect.element(toolbar).toBeVisible();
+  });
+
   it('TINY-13066: Should navigate between groups with Tab key', async () => {
     const { getByTestId } = render(
       <Fragment>


kimwoodfield:
The changes in this PR have already been reviewed by the wider team as individual PRs.

modules/oxide-components/src/main/ts/components/contexttoolbar/ContextToolbar.tsx
```diff
@@ -0,0 +1,271 @@
...
+    <div
+      ref={triggerRef}
+      onClick={handleClick}
+      onMouseDown={handleMouseDown}
+      {...rest}
+    >
+      {children}
+    </div>
+  );
+};
```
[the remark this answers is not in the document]
kimwoodfield: The trigger is a wrapper and consumers should provide keyboard accessible triggers like buttons and links. The component documentation already shows it using buttons. Adding this would create edge cases without solving core issues (like non-focusable elements).

modules/oxide-components/src/main/ts/components/contexttoolbar/ContextToolbarTypes.tsx
```diff
@@ -0,0 +1,34 @@
...
+  readonly onMouseDown?: MouseEventHandler<HTMLDivElement>;
+}
+
+export interface TriggerProps {
+  readonly children: ReactNode;
+  readonly onClick?: MouseEventHandler<HTMLDivElement>;
+  readonly onMouseDown?: MouseEventHandler<HTMLDivElement>;
+  readonly className?: string;
+  readonly style?: CSSProperties;
+}
```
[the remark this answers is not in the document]
kimwoodfield: Same explanation as here: https://github.com/tinymce/tinymce/pull/10686#discussion_r2464648964

modules/oxide-components/src/main/ts/components/contexttoolbar/ContextToolbar.stories.tsx
```diff
@@ -0,0 +1,343 @@
...
+
+## Accessibility
+- **Keyboard Navigation**: 
+  - \`Tab\` / \`Shift+Tab\`: Navigate between toolbar groups (cyclic)
+  - \`Arrow Left\` / \`Arrow Up\`: Navigate to previous button in group
+  - \`Arrow Right\` / \`Arrow Down\`: Navigate to next button in group
+  - \`Enter\`: Execute focused button
+  - \`Escape\`: Close the toolbar (unless \`persistent=true\`)
+- **Focus Management**: Toolbar automatically receives focus when opened, with the first button in the first group focused
+- **Click Outside**: Click outside the toolbar to dismiss it (unless \`persistent=true\`)
```
spocke: I think this might need a predicate since it depends on where you click say you click in other parts of the editor UI then you might not want this to auto close but if you click on other non editor elements on the page you want it to close.

modules/oxide-components/src/main/ts/components/contexttoolbar/ContextToolbar.stories.tsx
```diff
@@ -0,0 +1,343 @@
...
+        >
+          {triggerPositions.map((pos) => (
+            <ContextToolbar.Root key={pos.id} persistent={false}>
+              <ContextToolbar.Trigger>
+                <div style={{ position: 'absolute', display: 'inline-flex', ...pos.style }}>
+                  <Button>{pos.label}</Button>
+                </div>
+              </ContextToolbar.Trigger>
+              <ContextToolbar.Toolbar>
+                <ContextToolbar.Group>
```
spocke, line 332: Would we at this point add support for name and labels for the groups or is that scope creep. https://www.tiny.cloud/docs/tinymce/latest/contexttoolbar/#add-labels-and-groups-for-context-toolbar-buttons

modules/oxide-components/src/main/ts/components/contexttoolbar/ContextToolbar.tsx
```diff
@@ -0,0 +1,279 @@
...
+  type MouseEventHandler
+} from 'react';
+
+import * as KeyboardNavigationHooks from '../../keynav/KeyboardNavigationHooks';
+
+import type { ContextToolbarContextValue, ContextToolbarProps, GroupProps, ToolbarProps, TriggerProps } from './ContextToolbarTypes';
+
+const ContextToolbarContext = createContext<ContextToolbarContextValue | null>(null);
+
+const useContextToolbarContext = () => {
```
spocke, line 21: Should this be spit out into a ContextToolbarHook module?

modules/oxide-components/src/main/ts/components/contexttoolbar/ContextToolbar.tsx
```diff
@@ -0,0 +1,279 @@
...
+
+    const sugarAnchor = SugarElement.fromDom(anchorElement);
+    const sugarToolbar = SugarElement.fromDom(toolbar);
+
+    Css.set(sugarAnchor, 'anchor-name', anchorName);
+    Css.set(sugarToolbar, 'position-anchor', anchorName);
+
+    const gap = Type.isNonNullable(toolbar.ownerDocument?.defaultView)
+      ? Css.get(sugarToolbar, '--context-toolbar-gap') || '6px'
+      : '6px';
```
spocke: Avoid using || as the fallback for falsy values use proper Type predicates. Also maybe split this out into a separate function and also when will the toolbar here now have a default view if that is the case then something is really wrong right?

modules/oxide-components/src/main/ts/components/contexttoolbar/ContextToolbar.tsx
```diff
@@ -0,0 +1,279 @@
...
+  const anchorName = useMemo(() => `--${Id.generate('context-toolbar')}`, []);
+
+  useEffect(() => {
+    if (!isOpen || !Type.isNonNullable(triggerRef.current) || !Type.isNonNullable(toolbarRef.current)) {
+      return;
+    }
+
+    const trigger = triggerRef.current;
+    const toolbar = toolbarRef.current;
+    const anchorElement = (trigger.firstElementChild as HTMLElement) ?? trigger;
```
spocke: So what happens here if the first element child is an SVG?

modules/oxide-components/src/main/ts/components/contexttoolbar/ContextToolbar.tsx
```diff
@@ -0,0 +1,279 @@
...
+    return () => {
+      Css.remove(sugarAnchor, 'anchor-name');
+      Arr.each([ 'position-anchor', 'top', 'left', 'position-try-fallbacks' ], (property) => {
+        Css.remove(sugarToolbar, property);
+      });
+    };
+  }, [ anchorName, isOpen, triggerRef, toolbarRef ]);
+
+  const handleMouseDown = useCallback<MouseEventHandler<HTMLDivElement>>((event) => {
+    event.preventDefault();
```
spocke: Why do we need to prevent default here?

modules/oxide-components/src/test/ts/browser/components/ContextToolbar.spec.tsx
```diff
@@ -0,0 +1,270 @@
...
+    await expect.element(button2).toHaveFocus();
+
+    await userEvent.keyboard('{ArrowRight}');
+    await expect.element(button3).toHaveFocus();
+
+    await userEvent.keyboard('{ArrowLeft}');
+    await expect.element(button2).toHaveFocus();
+  });
+
+  it('TINY-13066: Should execute button on Enter key', async () => {
```
spocke, line 270: We also trigger on space could you add a test for that as well.

modules/oxide/src/less/theme/components/context-toolbar/context-toolbar.less
```diff
@@ -0,0 +1,42 @@
+//
+// Context Toolbar
+//
+.tox {
+  .tox-context-toolbar {
+    // CSS variable for gap between anchor and toolbar
+    --context-toolbar-gap: 6px;
```
spocke: I'm not sure we should grab the gap from the variable like this instead I think you could just get the computed gap value. Since any variable would be public API.

modules/oxide/src/less/theme/components/context-toolbar/context-toolbar.less
```diff
@@ -0,0 +1,42 @@
...
+
+    // CSS Anchor Positioning (required for anchor() to work)
+    position: absolute;
+
+    // Enable automatic flip when toolbar would overflow viewport, very new, not recongnized by stylelint yet
+    // stylelint-disable-next-line property-no-unknown -- CSS Anchor Positioning spec (Chrome 125+)
+    position-try-fallbacks: flip-block, flip-inline, flip-block flip-inline;
+    z-index: @z-index-menu;
+
+    background-color: #ffffff;
```
spocke, line 14: I guess all these needs to be some variable maybe @ztomaszyk can assist.

modules/oxide-components/src/main/ts/components/contexttoolbar/ContextToolbarTypes.tsx
```diff
@@ -0,0 +1,34 @@
...
+  readonly open: () => void;
+  readonly close: () => void;
+  readonly triggerRef: RefObject<HTMLDivElement>;
+  readonly toolbarRef: RefObject<HTMLDivElement>;
+  readonly persistent: boolean;
+}
+
+export interface ToolbarProps {
+  readonly children: ReactNode;
+  readonly className?: string;
```
spocke: I don't think we should expose className or style outside the component. I think all the components we have in `oxide-components` should be styled with `oxide` then any alterations in say size or what ever should be properties similar to `bridge` internally inside the `oxide-components` we can have class and style props but not the library API surface.

modules/oxide-components/src/main/ts/components/contexttoolbar/ContextToolbar.tsx
```diff
@@ -0,0 +1,281 @@
...
+  } = useContextToolbarContext();
+
+  const popoverMode = persistent ? 'manual' : 'auto';
+
+  useEffect(() => {
+    const element = toolbarRef.current;
+    if (Type.isNonNullable(element)) {
+      if (isOpen) {
+        element.showPopover();
+        // Defer focus to next event loop tick to ensure
```
spocke: Are we sure that it will always be just one tick or is it up to the browser to pick when this will happen. If so then this might flake in tests. Is there really no way of blocking auto focus of things in popups? Could we prevent default some event?

modules/oxide-components/src/main/ts/components/contexttoolbar/ContextToolbar.tsx
```diff
@@ -0,0 +1,281 @@
...
+}) => {
+  const {
+    isOpen,
+    toolbarRef,
+    triggerRef,
+    close,
+    persistent
+  } = useContextToolbarContext();
+
+  const popoverMode = persistent ? 'manual' : 'auto';
```
spocke: I think we might need to always have manual here so we have more control.

modules/oxide-components/src/main/ts/components/contexttoolbar/ContextToolbar.stories.tsx
kimwoodfield: I see your point but adding a predicate would add complexity for a hypothetical use case that we don't have yet. If a future consumer needs more nuanced "outside" logic with persistent={false} then we can add it but for now the simple behaviour and persistent mode covers all our needs.

modules/oxide-components/src/main/ts/components/contexttoolbar/ContextToolbar.stories.tsx
kimwoodfield: It's a good idea for the future but I think it's beyond the scope of this PR. Despite the name, we're building a generic toolbar component for now, not replicating every feature from core.

modules/oxide-components/src/main/ts/components/contexttoolbar/ContextToolbar.tsx
kimwoodfield: Yes, but not in this PR. I've created a ticket for us to work on that separately.

modules/oxide-components/src/main/ts/components/contexttoolbar/ContextToolbar.tsx
kimwoodfield: Right, good point. Thanks for raising that. I've pushed a fix.

modules/oxide-components/src/main/ts/components/contexttoolbar/ContextToolbar.tsx
kimwoodfield: I've updated it to use instanceof to check if firstElementChild is an HTMLElement before using it. Otherwise it falls back to the trigger if it's an SVG or other non-HTML element.

modules/oxide-components/src/main/ts/components/contexttoolbar/ContextToolbar.tsx
kimwoodfield: Good catch, I think that was a leftover from when I was fighting with the focus management. I've removed it.

modules/oxide-components/src/test/ts/browser/components/ContextToolbar.spec.tsx
kimwoodfield: Good point, I've added a test for that one.

modules/oxide/src/less/theme/components/context-toolbar/context-toolbar.less
kimwoodfield: Right, it's not really configurable so we don't need it. I've removed it.

modules/oxide-components/src/main/ts/components/contexttoolbar/ContextToolbarTypes.tsx
kimwoodfield: Makes sense. I've removed className and style from ToolbarProps and TriggerProps. Consumers can still style their children (buttons, etc.) but not the toolbar wrapper itself. I've kept the event handlers (onClick, onMouseDown) since those are needed for interactions.

modules/oxide-components/src/main/ts/components/contexttoolbar/ContextToolbar.tsx
kimwoodfield: Fair point. I've replaced it with queueMicrotask to improve the reliability over setTimeout. From what I can tell, showPopover focuses the popover and queueMicrotask is running reliably after that. Tests are stable. Happy to try a different approach if you have one in mind

modules/oxide-components/src/main/ts/components/contexttoolbar/ContextToolbar.tsx
kimwoodfield: With the dynamic approach, we get auto-dismiss for free when persistent={false}. Always using manual means we'd have to code that ourselves. Is there a specific issue you're seeing?

spocke: Yes, we have zero control without control then we might always just use manual.

I think generally we close UI things if the user clicks away from the editor UI not clicking away anywhere on the page.

Maybe not the best demo but in this demo place caret inside editor, then open a dialog vs clicking outside the on the rest of the page. https://fiddle.tiny.cloud/glTqRD6Fla/0

kimwoodfield: Right, I updated it to always use manual mode. For persistent use cases, use persistent={true} to keep it open until we explicitly dismiss it. For simpler cases that should auto-dismiss on click outside, persistent={false} handles that.

modules/oxide/src/less/theme/components/context-toolbar/context-toolbar.less
kimwoodfield: As discussed, I've created a ticket for that one and added it to our tech debt list.

modules/oxide-components/src/main/ts/components/contexttoolbar/ContextToolbar.tsx
```diff
@@ -0,0 +1,270 @@
...
+
+    Css.set(sugarToolbar, 'position-try-fallbacks', 'flip-block, flip-inline, flip-block flip-inline');
+
+    return () => {
+      Css.remove(sugarAnchor, 'anchor-name');
+      Arr.each([ 'position-anchor', 'top', 'left', 'position-try-fallbacks' ], (property) => {
+        Css.remove(sugarToolbar, property);
+      });
+    };
+  }, [ anchorName, isOpen, triggerRef, toolbarRef ]);
```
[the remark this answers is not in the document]
kimwoodfield: We are already aware that Firefox does not support it yet. We will address that later. For now, we're okay with the limitation.

PR #10704, opened 2025-10-31
TINY-13077: Add anchorRef and open prop support to ContextToolbar

kimwoodfield:
Related Ticket: [TINY-13077](https://ephocks.atlassian.net/browse/TINY-13077)

Description of Changes:
* Added support for anchorRef and open props on the ContextToolbar component.
* `open` support so that we can still programatically control when the toolbar stays open/closed when there is no trigger.
* `anchorRef` so we can optionally anchor to a specific element.
* `onOpenChange` callback prop so Escape key and click-outside handlers can properly close controlled toolbars by updating parent state.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

[TINY-13077]: https://ephocks.atlassian.net/browse/TINY-13077?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Optional anchorRef on Root to anchor the toolbar to an external element; supports auto-open on mount and works alongside existing Trigger usage.

* **Bug Fixes**
  * Improved positioning, layout and outside-click handling to consider the anchor; enhanced cleanup of positioning-related styles.

* **Documentation**
  * Story demonstrating external anchoring, visibility control, and auto-open behavior.

* **Tests**
  * Browser test validating anchor positioning, auto-open, focus, click behavior, and persistence.

### DIFF
--- a/modules/oxide-components/src/main/ts/components/contexttoolbar/ContextToolbar.tsx
+++ b/modules/oxide-components/src/main/ts/components/contexttoolbar/ContextToolbar.tsx
@@ -47,27 +47,24 @@ const Root: FC<ContextToolbarProps> = ({
 
   const getAnchorElement = useCallback((): HTMLElement | null => {
     // Prefer anchorRef.current if provided
-    if (Type.isNonNullable(anchorRef?.current)) {
-      return anchorRef.current;
-    }
-
-    if (Type.isNonNullable(triggerRef.current)) {
-      const firstChild = triggerRef.current.firstElementChild;
-      if (Type.isNonNullable(firstChild)) {
-        const sugarFirstChild = SugarElement.fromDom(firstChild);
-        if (SugarNode.isHTMLElement(sugarFirstChild)) {
-          return sugarFirstChild.dom;
-        }
-      }
-      return triggerRef.current;
-    }
-    return null;
+    return Optional.from(anchorRef?.current)
+      .orThunk(() => {
+        // Otherwise, try to get first child of trigger, or fall back to trigger itself
+        return Optional.from(triggerRef.current)
+          .map(SugarElement.fromDom)
+          .bind(Traverse.firstChild)
+          .filter(SugarNode.isHTMLElement)
+          .map((child) => child.dom)
+          .orThunk(() => Optional.from(triggerRef.current));
+      })
+      .getOrNull();
   }, [ anchorRef, triggerRef ]);
 
   // Auto-open when mounting with anchorRef (no Trigger)
+  // Since component remounts when anchorRef changes, we can read it directly
   useEffect(() => {
     const anchor = getAnchorElement();
-    if (Type.isNonNullable(anchor)) {
+    if (Type.isNonNullable(anchor) && !Type.isNonNullable(triggerRef.current)) {
       // Use requestAnimationFrame to ensure anchor is ready and allow Trigger to mount first
       const rafId = window.requestAnimationFrame(() => {
         if (!Type.isNonNullable(triggerRef.current)) {


modules/oxide-components/src/main/ts/components/contexttoolbar/ContextToolbar.stories.tsx
```diff
@@ -57,6 +57,13 @@ The provider component that manages toolbar state.
 
 **Props:**
 - \`persistent\`: (optional, default: \`false\`). If true, the toolbar stays open when clicking outside of it.
+- \`open\`: (optional). When provided, controls the toolbar visibility in controlled mode. If not provided, the toolbar manages its own state (uncontrolled mode).
+- \`onOpenChange\`: (optional). Callback function called when toolbar open state changes. Required when using controlled mode (\`open\` prop) to handle state updates from Escape key and click-outside handlers.
+- \`anchorRef\`: (optional). A ref to an external element to anchor the toolbar to. When provided, the toolbar positions relative to this element instead of the Trigger component.
+
+**Important:** \`open\` and \`anchorRef\` are mutually exclusive with using the \`Trigger\` component. Use either:
+- **Uncontrolled mode**: Use \`Trigger\` component (no \`open\` prop)
```
tiny-ben-tran: It would be nice if it could log a warning message when either Uncontrolled or Controlled mode rule is not met. Even nicer if it could be done at Typescript level

modules/oxide-components/src/main/ts/components/contexttoolbar/ContextToolbar.tsx
```diff
@@ -166,37 +186,42 @@ const Toolbar: FC<ToolbarProps> = ({
...
-    const anchorElement = trigger.firstElementChild instanceof window.HTMLElement
-      ? trigger.firstElementChild
-      : trigger;
+    const anchorElement = Optional.from(anchorRef?.current)
+      .orThunk(() =>
+        Optional
+          .from(triggerOrAnchor.firstElementChild)
+          .filter((child) => child instanceof window.HTMLElement)
+      )
+      .getOr(triggerOrAnchor);
```
tiny-ben-tran: I think we should prevent updating/changing the mode after init. Then the logic of handling the anchor element should be much simplier. The component should also throw an error when the mode is to be updated

modules/oxide-components/src/main/ts/components/contexttoolbar/ContextToolbar.tsx
```diff
@@ -166,37 +204,42 @@ const Toolbar: FC<ToolbarProps> = ({
...
     const topValue = `calc(anchor(${anchorName} bottom) + ${defaultToolbarGap})`;
-    const leftValue = `anchor(${anchorName} left)`;
 
     Css.set(sugarToolbar, 'top', topValue);
-    Css.set(sugarToolbar, 'left', leftValue);
-
+    Css.set(sugarToolbar, 'justify-self', 'anchor-center');
     Css.set(sugarToolbar, 'position-try-fallbacks', 'flip-block, flip-inline, flip-block flip-inline');
 
     return () => {
```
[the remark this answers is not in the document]
kimwoodfield: We already have an upcoming ticket for handling JS-based positioning fallback for Firefox / non-Anchor browsers.

modules/oxide-components/src/main/ts/components/contexttoolbar/ContextToolbar.tsx
kimwoodfield: Done! 👍🏻

modules/oxide-components/src/main/ts/components/contexttoolbar/ContextToolbar.tsx
```diff
@@ -89,6 +126,7 @@ const Toolbar: FC<ToolbarProps> = ({
     isOpen,
     toolbarRef,
     triggerRef,
+    anchorRef,
```
tiny-ben-tran: From what I can see, these two refs are the same for the Toolbar, as an anchor ref. I think the toolbar only needs to know one `anchorRef` and the mode (for getting the right DOM element from ref).

kimwoodfield: Good point, I've added a `getAnchorElement` function to the Root component that abstracts the anchor selection so the Toolbar will now use that instead of choosing between refs.

shanmen-tiny (approved):

modules/oxide-components/src/main/ts/components/contexttoolbar/ContextToolbar.tsx
```diff
@@ -30,24 +30,77 @@ const defaultToolbarGap = '6px';
...
+    }
+  }, [ onOpenChange ]);
+
+  const getAnchorElement = useCallback((): HTMLElement | null => {
+    if (Type.isNonNullable(anchorRef?.current)) {
+      return anchorRef.current;
+    }
+
+    if (Type.isNonNullable(triggerRef.current)) {
+      const firstChild = triggerRef.current.firstElementChild;
```
spocke: Any particular reason we couldn't use sugar here and then use the SugarNode.isHTMLElement predicate.

modules/oxide-components/src/test/ts/browser/components/ContextToolbar.spec.tsx
```diff
@@ -330,4 +331,312 @@ describe('browser.ContextToolbar.ContextToolbar', () => {
...
+    }
+
+    // Verify clicking anchor doesn't close toolbar (works regardless of Anchor API support)
+    await anchor.click();
+    await expect.element(button).toBeVisible();
+
+    // Verify toolbar is still visible after clicking anchor
+    const toolbarAfterClick = container.querySelector('.tox-context-toolbar');
+    expect(toolbarAfterClick).toBeTruthy();
+    if (toolbarAfterClick instanceof window.Element) {
```
spocke: This should already be a element is this a null check if so then we should use Type.isNonNullable or something.

modules/oxide-components/src/main/ts/components/contexttoolbar/ContextToolbar.tsx
kimwoodfield: Yes, we can do it. I've updated it to use the SugarNode predicate.

modules/oxide-components/src/test/ts/browser/components/ContextToolbar.spec.tsx
kimwoodfield: I've removed this code so it is no longer needed.

modules/oxide-components/src/main/ts/components/contexttoolbar/ContextToolbar.tsx
```diff
@@ -30,24 +31,91 @@ const defaultToolbarGap = '6px';
...
+    return null;
+  }, [ anchorElement, triggerRef ]);
+
+  // Sync anchorRef.current to state after every render to catch ref changes
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  useLayoutEffect(() => {
+    const current = anchorRef?.current ?? null;
+    setAnchorElement((prev) => {
+      // Only update if value actually changed to avoid unnecessary re-renders
+      return prev !== current ? current : prev;
```
TheSpyder: If `prev === current` I highly doubt returning `current` will be any different to `prev` 🤔

kimwoodfield: I've removed that code.

PR #10708, opened 2025-11-04
TINY-13143: Update UI context field to support multiple entries

tiny-ben-tran:
Related Ticket: TINY-13143

Description of Changes:

- Updated the logic of checking UI context to handle multiple entries
- Changed `contentType` to string array

Pre-checks:

- [x] Changelog entry added

Review:

- [ ] Milestone set
- [ ] Docs ticket created (if applicable)

## Summary by CodeRabbit

- **New Features**
  - UI component context now accepts multiple comma-separated values; components evaluate all listed contexts, short-circuit when disabled, and use the existing design-mode fallback.

- **Tests**
  - Expanded test coverage for multi-context handling: full/partial matches, ignored empty/unknown entries, mixed results, and updated mode-related expectations to reflect array-based context checks.

### DIFF
--- a/modules/tinymce/src/themes/silver/test/ts/browser/editor/backstage/BackstageCheckContextTest.ts
+++ b/modules/tinymce/src/themes/silver/test/ts/browser/editor/backstage/BackstageCheckContextTest.ts
@@ -72,28 +72,28 @@ describe('browser.tinymce.themes.silver.editor.backstage.BackstageSinkTest', () 
             editorReadOnly: true
           });
 
-          assert.deepEqual(lazyBackstages().popup.shared.providers.checkUiComponentContext('mode:design'), { contextType: [ 'mode' ], shouldDisable: false });
-          assert.deepEqual(lazyBackstages().popup.shared.providers.checkUiComponentContext('mode:readonly'), { contextType: [ 'mode' ], shouldDisable: true });
-          assert.deepEqual(lazyBackstages().popup.shared.providers.checkUiComponentContext('any'), { contextType: [ 'any' ], shouldDisable: false });
-          assert.deepEqual(lazyBackstages().popup.shared.providers.checkUiComponentContext('mode:!design'), { contextType: [ 'mode' ], shouldDisable: true });
-          assert.deepEqual(lazyBackstages().popup.shared.providers.checkUiComponentContext('mode:!readonly'), { contextType: [ 'mode' ], shouldDisable: false });
-          assert.deepEqual(lazyBackstages().popup.shared.providers.checkUiComponentContext('mode:!random'), { contextType: [ 'mode' ], shouldDisable: false });
+          assert.deepEqual(lazyBackstages().popup.shared.providers.checkUiComponentContext('mode:design'), { contextTypes: [ 'mode' ], shouldDisable: false });
+          assert.deepEqual(lazyBackstages().popup.shared.providers.checkUiComponentContext('mode:readonly'), { contextTypes: [ 'mode' ], shouldDisable: true });
+          assert.deepEqual(lazyBackstages().popup.shared.providers.checkUiComponentContext('any'), { contextTypes: [ 'any' ], shouldDisable: false });
+          assert.deepEqual(lazyBackstages().popup.shared.providers.checkUiComponentContext('mode:!design'), { contextTypes: [ 'mode' ], shouldDisable: true });
+          assert.deepEqual(lazyBackstages().popup.shared.providers.checkUiComponentContext('mode:!readonly'), { contextTypes: [ 'mode' ], shouldDisable: false });
+          assert.deepEqual(lazyBackstages().popup.shared.providers.checkUiComponentContext('mode:!random'), { contextTypes: [ 'mode' ], shouldDisable: false });
 
           editor.mode.set('readonly');
-          assert.deepEqual(lazyBackstages().popup.shared.providers.checkUiComponentContext('mode:design'), { contextType: [ 'mode' ], shouldDisable: true });
-          assert.deepEqual(lazyBackstages().popup.shared.providers.checkUiComponentContext('mode:readonly'), { contextType: [ 'mode' ], shouldDisable: false });
-          assert.deepEqual(lazyBackstages().popup.shared.providers.checkUiComponentContext('mode:!design'), { contextType: [ 'mode' ], shouldDisable: false });
-          assert.deepEqual(lazyBackstages().popup.shared.providers.checkUiComponentContext('mode:!readonly'), { contextType: [ 'mode' ], shouldDisable: true });
-          assert.deepEqual(lazyBackstages().popup.shared.providers.checkUiComponentContext('mode:!random'), { contextType: [ 'mode' ], shouldDisable: false });
+          assert.deepEqual(lazyBackstages().popup.shared.providers.checkUiComponentContext('mode:design'), { contextTypes: [ 'mode' ], shouldDisable: true });
+          assert.deepEqual(lazyBackstages().popup.shared.providers.checkUiComponentContext('mode:readonly'), { contextTypes: [ 'mode' ], shouldDisable: false });
+          assert.deepEqual(lazyBackstages().popup.shared.providers.checkUiComponentContext('mode:!design'), { contextTypes: [ 'mode' ], shouldDisable: false });
+          assert.deepEqual(lazyBackstages().popup.shared.providers.checkUiComponentContext('mode:!readonly'), { contextTypes: [ 'mode' ], shouldDisable: true });
+          assert.deepEqual(lazyBackstages().popup.shared.providers.checkUiComponentContext('mode:!random'), { contextTypes: [ 'mode' ], shouldDisable: false });
 
           editor.mode.set('testmode');
-          assert.deepEqual(lazyBackstages().popup.shared.providers.checkUiComponentContext('mode:design'), { contextType: [ 'mode' ], shouldDisable: true });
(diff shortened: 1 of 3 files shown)


modules/tinymce/src/themes/silver/main/ts/backstage/Backstage.ts
```diff
@@ -69,17 +82,17 @@ const init = (lazySinks: { popup: () => Result<AlloyComponent, string>; dialog:
...
+        Arr.filter(specContext.split(','), (spec) => spec.trim() !== ''),
+        (spec) => enabledInContextFn(spec.trim(), contexts)
+      );
+
+      const isAllEnabled = Arr.forall(contextResults, (spec) => spec.result === true);
+
       return {
-        contextType: key,
-        shouldDisable: !enabledInContext
+        contextType: Arr.map(contextResults, (al) => al.key).join(','),
```
ltrouton: I am a little bit worried about how this could affect current uses of `contextType` in other files. Could it be safer to change `contextType` to an array instead then update the other files to iterate over the array?

modules/tinymce/src/themes/silver/test/ts/browser/editor/backstage/BackstageCheckContextTest.ts
```diff
@@ -91,6 +91,23 @@ describe('browser.tinymce.themes.silver.editor.backstage.BackstageSinkTest', ()
...
+
+          assert.deepEqual(lazyBackstages().popup.shared.providers.checkUiComponentContext('context_1:test,context_2:test'), { contextType: 'context_1,context_2', shouldDisable: false });
+
+          assert.deepEqual(lazyBackstages().popup.shared.providers.checkUiComponentContext('context_1:test,context_2:test2'), { contextType: 'context_1,context_2', shouldDisable: true });
+
+          // Ignore empty entry
+          assert.deepEqual(lazyBackstages().popup.shared.providers.checkUiComponentContext('context_1:test,,context_2:test2,'), { contextType: 'context_1,context_2', shouldDisable: true });
+
+          // context_3 is not registered so it is reasonable to expect shouldDisabled:true
+          // However, because of the current logic, any unfound context is assumed as mode:design
```
ltrouton: I personally don't mind the fallback to `mode:design`. Although, I do think it needs to be updated slightly if we are supporting multiple contexts. Could we refactor the code so that contexts that don't exist are first filtered out. If there are no contexts to iterate over, then can fallback to `mode:design'.

modules/tinymce/src/themes/silver/main/ts/backstage/Backstage.ts
tiny-ben-tran: I don't think it is necessary to change the type of `contextType` since its only usage [here](https://github.com/tinymce/tinymce/blob/f1d1a8148fa411f9942255b23b016d1ed8e10440/modules/tinymce/src/themes/silver/main/ts/UiState.ts#L74) and in a couple of tests are for a single entry.

ltrouton: Where you point to UiState.ts, could having a comma separated string interfere with the predicate check?

modules/tinymce/src/themes/silver/main/ts/backstage/Backstage.ts
```diff
@@ -69,17 +82,17 @@ const init = (lazySinks: { popup: () => Result<AlloyComponent, string>; dialog:
...
-      const [ key, value = '' ] = specContext.split(':');
       const contexts = editor.ui.registry.getAll().contexts;
-      const enabledInContext = Obj.get(contexts, key)
-        .fold(
-          // Fallback to 'mode:design' if key is not found
-          () => Obj.get(contexts, 'mode').map((pred) => pred('design')).getOr(false),
-          (pred) => value.charAt(0) === '!' ? !pred(value.slice(1)) : pred(value)
-        );
+      const contextResults = Arr.map(
+        Arr.filter(specContext.split(','), (spec) => spec.trim() !== ''),
```
TheSpyder: The AI was correct about there being a whitespace concern, but if you think about it we use this same pattern in many other places (that's likely why it was chosen for this feature). Find the common function we use for all of them and use it here, please. I think it's called `explode` or something like that.

modules/tinymce/src/themes/silver/main/ts/backstage/Backstage.ts
TheSpyder: if `contextType` is only used once, wouldn't that make changing the type _easier_?

TheSpyder: Another way to approach this would be to make the internal value converted from the comma-separated string to an array, and have `checkUiComponentContext` be called once for each element in the array.

modules/tinymce/src/themes/silver/test/ts/browser/editor/backstage/BackstageCheckContextTest.ts
tiny-ben-tran: Done!

modules/tinymce/src/themes/silver/main/ts/backstage/Backstage.ts
tiny-ben-tran: > Where you point to UiState.ts, could having a comma separated string interfere with the predicate check?

@ltrouton I see. I will change `contentType` type to a string array.

> Another way to approach this would be to make the internal value converted from the comma-separated string to an array, and have checkUiComponentContext be called once for each element in the array.

@TheSpyder I am not sure if that approach is ideal because it will touch many files that call ` providerBackstage.checkUiComponentContext(spec.context).shouldDisable`.

modules/tinymce/src/themes/silver/main/ts/backstage/Backstage.ts
```diff
@@ -65,21 +79,24 @@ const init = (lazySinks: { popup: () => Result<AlloyComponent, string>; dialog:
...
-        .fold(
-          // Fallback to 'mode:design' if key is not found
-          () => Obj.get(contexts, 'mode').map((pred) => pred('design')).getOr(false),
-          (pred) => value.charAt(0) === '!' ? !pred(value.slice(1)) : pred(value)
-        );
+      const contextResults = Arr.map(
+        Arr.filter(Tools.explode(specContext), Strings.isNotEmpty),
+        (spec) => enabledInContextFn(spec, contexts)
+      );
+      const matchedContextResults = Arr.filter(contextResults, (c) => Type.isNonNullable(c.result));
```
TheSpyder: uh, use optionals please not `null` and `filter` like this (There should be a `filterMap` for array somewhere that discards the `Optional.none()` results)

- [x] avoid null

modules/tinymce/src/themes/silver/main/ts/backstage/Backstage.ts
```diff
@@ -55,6 +56,19 @@ const init = (lazySinks: { popup: () => Result<AlloyComponent, string>; dialog:
   const contextMenuState = Cell(false);
   const toolbar = HeaderBackstage(editor);
 
+  const enabledInContextFn = (specContext: string, contexts: Record<string, (args: string) => boolean>) => {
```
TheSpyder: missing return type (why isn't this a lint error?)

TheSpyder (approved):

modules/tinymce/src/themes/silver/main/ts/backstage/Backstage.ts
```diff
@@ -65,21 +79,24 @@ const init = (lazySinks: { popup: () => Result<AlloyComponent, string>; dialog:
     checkUiComponentContext: (specContext: string) => {
       if (Options.isDisabled(editor)) {
         return {
-          contextType: 'disabled',
+          contextType: [ 'disabled' ],
```
kimwoodfield: I think we should update contextType to contextTypes to reflect the change. Before it was always a single context type, where now it's a list of them.

modules/tinymce/src/themes/silver/main/ts/backstage/Backstage.ts
tiny-ben-tran: It appears that `@typescript-eslint/explicit-function-return-type` is turned off by default. Maybe because a lot of places need a return type.

modules/tinymce/src/themes/silver/main/ts/backstage/Backstage.ts
tiny-ben-tran: I used `Optionals.cat` instead of `filterMap` because the function is from another library `fs-pt/Array`.

modules/tinymce/src/themes/silver/main/ts/backstage/Backstage.ts
TheSpyder: Hmmm I thought we turned that on. Maybe we left it off for core because it has so much legacy code.

modules/tinymce/src/themes/silver/main/ts/backstage/Backstage.ts
TheSpyder: Ah it's `findMap` we have in core. We did create `filterMap` in other places so I was a bit confused. 

One day I should check whether we have other cases of `Optionals.cat(Arr.map(...))` which would make it worth adding the filter map function 😂

kimwoodfield (approved):

modules/tinymce/src/themes/silver/main/ts/backstage/Backstage.ts
ltrouton: I think that is a good idea 👍

tiny-ben-tran:
We decided to close this PR pending further investigation/discussion

PR #10712, opened 2025-11-05
TINY-12912: Moved user prompt bubble to core

HAFRMO:
Related Ticket: TINY-12912

Description of Changes:
Moved user prompt bubble

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

## Summary by CodeRabbit

* **New Features**
  * Added a UserPromptBubble component for displaying user prompts.
  * Added a Storybook example to preview the component.

* **Chores**
  * Exposed the new component via the package public API.
  * Updated TypeScript configuration to support local path aliases for module imports.

### DIFF
--- a/modules/oxide-components/src/main/ts/bespoke/tinyai/bubbles/UserPromptBubble.stories.tsx
+++ b/modules/oxide-components/src/main/ts/bespoke/tinyai/bubbles/UserPromptBubble.stories.tsx
@@ -1,8 +1,13 @@
-import type { Meta, StoryObj } from '@storybook/react-vite';
+import type { Meta, ReactRenderer, StoryObj } from '@storybook/react-vite';
+import { classes } from 'oxide-components/utils/Styles';
+import type { PartialStoryFn } from 'storybook/internal/csf';
 
 import { UserPromptBubble } from './UserPromptBubble';
 
 const meta = {
+  decorators: [
+    (Story: PartialStoryFn<ReactRenderer>): JSX.Element => <div className={classes([ 'tox-ai' ])}><Story /></div>
+  ],
   title: 'bespoke/tinyai/UserPromptBubble',
   component: UserPromptBubble,
   parameters: {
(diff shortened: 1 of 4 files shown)


modules/oxide-components/src/main/ts/bespoke/tinyai/bubbles/UserPromptBubble.stories.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+import { UserPromptBubble } from './UserPromptBubble';
+
+const meta = {
+  title: 'bespoke/tinyai/UserPromptBubble',
+  component: UserPromptBubble,
+  parameters: {
+    layout: 'centered',
+  },
+  tags: [ 'autodocs', 'skip-visual-testing' ],
```
hamza0867: We do need visual tests for this component as it doesn't have much logic, but rather defines visuals

modules/oxide-components/src/main/ts/bespoke/tinyai/bubbles/UserPromptBubble.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+interface UserPromptProps {
+  readonly prompt: string;
+}
+
+export const UserPromptBubble: FunctionComponent<UserPromptProps> = ({ prompt }) => {
+  return (
+    <div
+      tabIndex={-1}
+      className={classes([ 'tox-ai__user-prompt__text' ])}
```
hamza0867: I'd rather we use the Bem helper function here instead of the `classes` helper function

modules/oxide-components/src/main/ts/bespoke/tinyai/bubbles/UserPromptBubble.tsx
```diff
@@ -0,0 +1,18 @@
+import type { FunctionComponent } from 'react';
+
+import * as Bem from '../../../utils/Bem';
```
TheSpyder: We should probably use this more often
```suggestion
import * as Bem from 'oxide-components/utils/Bem';
```

HAFRMO: That doesn't work though, since oxide-components is not a reference within itself. Is it something that should be setup?

TheSpyder: Oh, it was only set up in demos and tests. Every other module only has one `tsconfig` so it works everywhere.
https://github.com/tinymce/tinymce/blob/ef4d5fe0fef16d71b89d1766cfd9dc55d52dc996/modules/oxide-components/tsconfig.demo.json#L26-L28

Yes, it should work everywhere, e.g.
https://github.com/tinymce/tinymce/blob/ba6994f39b09ecd2c28cef68c3e8a85a4815e235/modules/katamari/tsconfig.json#L8-L12

TheSpyder: I admit most projects only use it in demos and tests. But it should still work everywhere, and the point where we need a triple `../` I think is time to start using it.

PR #10715, opened 2025-11-06
TINY-13230: Added selector helpers to Bem module

spocke:
Related Ticket: TINY-13230

Description of Changes:
* Add more BEM helpers to build selectors

Pre-checks:

Review:
* [x] Milestone set

## Summary by CodeRabbit

* **New Features**
  * Added utilities to generate CSS selectors for component blocks and elements, representing modifiers as selector classes.
* **Improvements**
  * Block and element class outputs now consistently use space-separated modifier classes for predictable formatting.
  * Updated public APIs to use the new modifier-joining behavior.
* **Tests**
  * Added tests for selector generation covering modifiers, order-invariance, and falsey modifier handling.
* **Documentation**
  * Added JSDoc comments for public utilities.

### DIFF
--- a/modules/oxide-components/src/main/ts/utils/Bem.ts
+++ b/modules/oxide-components/src/main/ts/utils/Bem.ts
@@ -35,7 +35,9 @@ const applyModifiers = <M extends string>(joiner: string, base: string, modifier
 /**
  * Creates a BEM block class name string.
  *
- * @example Bem.block('tox-button', { 'active': true, 'naked': false });
+ * @example
+ * Bem.block('tox-button', { 'active': true, 'naked': false });
+ * // Returns: 'tox-button tox-button--active'
  * @param block The BEM block name
  * @param modifiers Optional modifiers for the block
  * @returns The BEM block class name string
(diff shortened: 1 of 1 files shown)


modules/oxide-components/src/main/ts/utils/Bem.ts
```diff
@@ -27,21 +27,38 @@ type ModifiersForBlock<B extends BEMBlocks> = Extract<ValidBlockModifiers, [B, a
...
   block: B,
   element: E,
   modifiers?: ModifierRecord<M>
 ): string => {
   const base = `${block}__${element}`;
-  return Type.isUndefined(modifiers) ? base : applyModifiers(base, modifiers);
+  return Type.isUndefined(modifiers) ? base : applyModifiers(' ', base, modifiers);
+};
+
+export const blockSelector = <B extends BEMBlocks, M extends ModifiersForBlock<B> = never>(
```
TheSpyder, line 80: I wonder if we need clearer names in this file 🤔  I had to look up some usages to understand that `Bem.block()` is used to generate class names and `Bem.blockSelector()` is to generate class selectors that find those class names.

TheSpyder (approved):

modules/oxide-components/src/main/ts/utils/Bem.ts
spocke: We could name them `Bem.blockClasses` and `Bem.elementClasses` but the name is weird since plural to me means an array and we want a string.

We could name them `Bem.blockClassName` and `Bem.elementClassName` then it's more specifically generating a string but then it becomes very long.

I think I prefer the shorter name even if it might be confusing since it's easier to embed into components less indentation etc.

TheSpyder: At least add method documentation so it's clearer without hunting through where these things are used 🙂

spocke: Added API docs now. I think we should also add a section on BEM in general in the `oxide-components` repo. How one should think them using that and these functions with examples.

PR #10716, opened 2025-11-07
TINY-12315: Preventing accordions to be clicked when disabled.

HAFRMO:
Related Ticket: TINY-12315

Description of Changes:
Preventing both using the command as well as using the mouse

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

## Summary by CodeRabbit

* **Bug Fixes**
  * Accordions now respect the editor's disabled state — toggle actions and related clicks are blocked when disabled.

* **Tests**
  * Added browser tests verifying accordions cannot be toggled while the editor is disabled and that behavior is correct when switching states.

* **Chores**
  * Added a changelog entry documenting the fix.


modules/tinymce/src/core/main/ts/mode/Disabled.ts
```diff
@@ -99,6 +104,10 @@ const processDisabledEvents = (editor: Editor, e: Event): void => {
         window.open(href, '_blank', 'rel=noopener noreferrer,menubar=yes,toolbar=yes,location=yes,status=yes,resizable=yes,scrollbars=yes');
       }
     });
+
+    if (editor.hasPlugin('accordion') && hasAccordion(editor, elm)) {
+      e.preventDefault();
+    }
```
TheSpyder: Why even have the plugin check? The toggle is happening at the DOM level, so the autoresize bug will happen with or without the plungin.

Also, this needs to be better combined with the anchor click code
- [x] remove hasPlugin
- [x] Don't run this check if the click was handled by the anchor href code

HAFRMO: Because I thought it best to tie all of the accordion actions to accordion plugin, and have whatever default actions we have for details elements untouched if someone haven't added the accordion plugin.
The click is sensible though.

spocke: The function is self is a bit weird it doesn't dispatch any events except `copy` but if we where to add `click` to allowed events then it wouldn't work due to the branching on links. Also since this is in the branch of the meta key check would you still be able to toggle accordions by holding down ctrl or cmd?
I think we should block them all even things we don't add our self. Because the click it self mutates the dom and since we are in disabled mode the hole idea is to not mutate the dom.

modules/tinymce/src/plugins/accordion/main/ts/api/Options.ts
```diff
@@ -0,0 +1,12 @@
+
```
spocke: Remove

modules/tinymce/src/plugins/accordion/test/ts/browser/AccordionPluginTest.ts
```diff
@@ -167,6 +168,29 @@ describe('browser.tinymce.plugins.accordion.AccordionPluginTest', () => {
     TinyAssertions.assertCursor(editor, [ 0, 0, 0 ], 0);
   });
 
+  it('TINY-12316: Toggle an accordion element not possible when disabled', () => {
```
spocke, line 171: I think you can make a test for the default prevent as well. Just do a Mouse.click on the details element then you have a `editor.on('click', ..)` that captures the event object and you check is `isDefaultPrevented` is true on it.

spocke (approved):

modules/tinymce/src/plugins/accordion/main/ts/api/Options.ts
HAFRMO: But this is needed to prevent the command from running while disabled?

spocke: The empty line at the start of the file is not needed.

spocke: I guess we should maybe have a lint rule for that.

HAFRMO: Oh, my apologizes, I thought you meant the whole file. I'll be happy to remove the empty line.

modules/tinymce/src/core/main/ts/mode/Disabled.ts
```diff
@@ -87,9 +92,11 @@ const processDisabledEvents = (editor: Editor, e: Event): void => {
       b) open the link using default browser behaviour
   */
   if (isClickEvent(e) && !VK.metaKeyPressed(e)) {
+    let isHandled = false;
```
ltrouton: Can we do `getAnchorHrefOpt.fold`​ instead to not require the `let`​ variable?

modules/tinymce/src/core/main/ts/mode/Disabled.ts
```diff
@@ -99,6 +106,10 @@ const processDisabledEvents = (editor: Editor, e: Event): void => {
         window.open(href, '_blank', 'rel=noopener noreferrer,menubar=yes,toolbar=yes,location=yes,status=yes,resizable=yes,scrollbars=yes');
       }
     });
+
+    if (!isHandled && hasAccordion(editor, elm)) {
```
ltrouton: Would calling `e.preventDefault()`​ potentially twice actually be an issue?

modules/tinymce/src/plugins/accordion/test/ts/browser/AccordionPluginTest.ts
```diff
@@ -167,6 +168,34 @@ describe('browser.tinymce.plugins.accordion.AccordionPluginTest', () => {
...
```
ltrouton, line 171: Can we add some comments or a line between each block ofassertions. It is hard to follow when it is all one block of assertions

modules/tinymce/src/core/main/ts/mode/Disabled.ts
HAFRMO: I think the desire is to avoid doing the `hasAccordion`-check twice.

modules/tinymce/src/plugins/accordion/test/ts/browser/AccordionPluginTest.ts
HAFRMO: Sure.

modules/tinymce/src/core/main/ts/mode/Disabled.ts
spocke: You can just refactor this to not have the test like moving out the anchor option into a variable and doing isNone on that and prevent for example.

PR #10724, opened 2025-11-12
TINY-13133: Add ai styles

shanmen-tiny:
Related Ticket: TINY-13133

Description of Changes:
* Add AI skin and content styles

Pre-checks:

Review:
* [x] Milestone set

### DIFF
--- a/modules/oxide/src/less/theme/content-ui.less
+++ b/modules/oxide/src/less/theme/content-ui.less
@@ -3,7 +3,7 @@
 //
 @import 'globals/global-variables';
 @import 'globals/animations';
-import 'content/diff/diff';
+@import 'content/diff/diff';
 @import 'content/ai-preview/ai-preview';
 @import 'content/anchor/anchor';
 @import 'content/comments/comments';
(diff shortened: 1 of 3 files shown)


modules/oxide/src/less/theme/content/ai-preview/ai-preview.less
```diff
@@ -0,0 +1,54 @@
+@import '../diff/diff.less';
```
spocke: Do we have other files that import things like this. I think all of them are imported at the root lessfile?

spocke (approved):

modules/oxide/src/less/theme/content/ai-preview/ai-preview.less
shanmen-tiny: Yeah, only here at the moment, the diffs are shared between two plugins. I wanted to make sure that if we fix things, it would be applied to both places. However, I'm not sure if that's a good idea. Should I add to the root instead?

shanmen-tiny: Moved to the root lessfile 👍

TheSpyder (approved):

PR #10728, opened 2025-11-13
TINY-13095: update toggle css to transitional stage

ztomaszyk:
Related Ticket: TINY-13095

Description of Changes:
* Updated the toggle component to use CSS custom properties and follow CSS transitional stage guidelines 
* Add fallback values for CSS Custom Properties
* Added conditional loading of custom properties based on the `@custom-properties-enabled` flag
* Improved variable organization for toggle states (checked, unchecked)

Pre-checks:

Review:
* [x] Milestone set

## Summary by CodeRabbit

* **Style**
  * Enhanced toggle component styling to support theme customization using CSS custom properties with fallback values for improved design flexibility.

### DIFF
--- a/modules/oxide/src/less/theme/components/toggle/toggle.less
+++ b/modules/oxide/src/less/theme/components/toggle/toggle.less
@@ -1,19 +1,26 @@
 .tox {
-  .tox-toggle {
-    // private variables scoped to tox-toggle class
-    --tox-private-slider-background-color: @background-color;
-    --tox-private-slider-border-color: .bg-luma-checker(hsl(from @color-white h s calc(l - 11)), hsla(from @color-white h s l / 15%))[@result]; // to be figured out after the spike with conditionals in CSS
-    --tox-private-slider-handle-background-color: @text-color;
+  .tox-toggle when (@custom-properties-enabled = true) {
+    --tox-private-slider-background-color: var(--tox-private-background-color);
+    --tox-private-slider-border-color: light-dark(
+      hsl(from var(--tox-private-color-white) h s calc(l - 11)), 
+      hsl(from var(--tox-private-color-white) h s l / 15%));
+    --tox-private-slider-handle-background-color: var(--tox-private-text-color);
 
+    --tox-private-slider-checked-background-color: var(--tox-private-color-tint);
+    --tox-private-slider-checked-border-color: var(--tox-private-color-tint);
+    --tox-private-slider-checked-handle-background-color: var(--tox-private-color-white);
+  }
+
+  .tox-toggle {
     display: flex;
     align-items: center;
-    gap: @pad-xs;
-    padding: @pad-xs;
-    font-weight: @font-weight-normal;
-    line-height: 24px; // the same as in button and textfield. Button component is referencing the textfield variable so maybe it should be a global API? The textfield has a fixed value for line-height
+    gap: var(--tox-private-pad-xs, @pad-xs);
+    padding: var(--tox-private-pad-xs, @pad-xs);
+    font-weight: var(--tox-private-font-weight-normal, @font-weight-normal);
+    line-height: var(--tox-private-control-line-height, 24px);
     white-space: nowrap;
-    background-color: @background-color;
-    color: @text-color;
+    background-color: var(--tox-private-background-color, @background-color);
+    color: var(--tox-private-text-color, @text-color);
 
     input {
       position: absolute;
@@ -32,8 +39,10 @@
     border-radius: 34px; 
     width: 28px;
     height: 16px;
-    background-color: var(--tox-private-slider-background-color);
-    border: 1px solid var(--tox-private-slider-border-color);
+    background-color: var(--tox-private-slider-background-color, @background-color);
+    border: 1px solid var(
+      --tox-private-slider-border-color,
+      .bg-luma-checker(darken(@color-white, 11%), fade(@color-white, 15%))[@result]);
   }
 
   .tox-toggle__slider::before {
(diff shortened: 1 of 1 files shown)


zuzanna-antas:
* **#10728** <a href="https://app.graphite.com/github/pr/tinymce/tinymce/10728?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/tinymce/tinymce/10728?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

modules/oxide/src/less/theme/components/toggle/toggle.less
```diff
@@ -1,19 +1,26 @@
...
-  .tox-toggle {
-    // private variables scoped to tox-toggle class
-    --tox-private-slider-background-color: @background-color;
-    --tox-private-slider-border-color: .bg-luma-checker(hsl(from @color-white h s calc(l - 11)), hsla(from @color-white h s l / 15%))[@result]; // to be figured out after the spike with conditionals in CSS
-    --tox-private-slider-handle-background-color: @text-color;
+  .tox-toggle when (@custom-properties-enabled = true) {
+    --tox-private-slider-background-color: var(--tox-private-background-color);
+    --tox-private-slider-border-color: light-dark(
+      hsl(from var(--tox-private-color-white) h s calc(l - 11)), 
+      hsl(from var(--tox-private-color-white) h s l / 15%));
```
ArvinJ-H, line 6: As mentioned previously, there were compile issue with newer LESS, specifically on `from...` syntax. Moving this around won't solve it. It doesn’t matter that it’s inside a CSS variable. It doesn’t matter that it’s inside `.bg-luma-checker(...)`. As long as Less tries to treat that as a color and do math on it, it’ll blow up with `invalid type`.

`hsl(...) / hsla(...)` are parsed as Less color functions, but `from @color-white h s l...` is not a valid Less signature.

TheSpyder: I found a reference that suggests doing this to bypass the Less compiler and emit it as a string. I haven't checked if it works.
```suggestion
      ~"hsl(from var(--tox-private-color-white) h s calc(l - 11))", 
      ~"hsl(from var(--tox-private-color-white) h s l / 15%))";
```

TheSpyder (approved):

modules/oxide/src/less/theme/components/toggle/toggle.less
ArvinJ-H: yeah, if we need to dump the css here, while keeping less, it'll have to be the bypass, perhaps something like this?

```suggestion
  --tox-private-slider-border-color: ~"light-dark(
    hsl(from var(--tox-private-color-white) h s calc(l - 11)), 
    hsl(from var(--tox-private-color-white) h s l / 15%)
  )";
```

spocke: Will this we resolved by bumping less. I'm starting to feel that the issues we have here with modern css is maybe something we need to work on sooner than later.

It will be very very easy to break things otherwise and if we have no tests it might go to prod until we notice.

TheSpyder: I think bumping less is what causes this one? we're using a very old version at this point

ArvinJ-H: Yes, it's from bumping less.

This might not be an issue if we using CSS directly, tho I'm not sure if it's the best combination of syntax to use 🤔

zuzanna-antas: We are using this syntax in other files and it works just fine. From the error log Arvin send in the ticket only the toggle.less fails. And I think I know whats happening - in other files the modern CSS syntax is used behind a feature flag. This means if the flag is not enabled, the part inside the `when {}` guard is omitted by the compiler. So it sort of hides the issue. I was able to replicate the error having the FF enabled and less bumped to 4.4.0 (currently we have 4.2.0). It is weird tho that on the [LESS playground](https://lesscss.org/less-preview/#eyJjb2RlIjoiXG5AYmFja2dyb3VuZC1jb2xvcjogI2ZmZmZmZjtcbkBjb2xvci13aGl0ZTogI2ZmZmZmZjtcblxuLmJnLWx1bWEtY2hlY2tlcihAY29sb3ItbGlnaHQsIEBjb2xvci1kYXJrKSB7XG4gIEByZXN1bHQ6IGlmKChsdW1hKEBiYWNrZ3JvdW5kLWNvbG9yKSA+IDUwJSksIEBjb2xvci1saWdodCwgQGNvbG9yLWRhcmspO1xufVxuXG5kaXYge1xuICAtLWNvbG9yLXdoaXRlOiAjZmZmZmZmO1xuICBcbiAgLS11c2luZy1sZXNzLWFuZC1sdW1hLWNoZWNrZXI6IC5iZy1sdW1hLWNoZWNrZXIoaHNsKGZyb20gQGNvbG9yLXdoaXRlIGggcyBjYWxjKGwgLSAxMSkpLCBoc2xhKGZyb20gQGNvbG9yLXdoaXRlIGggcyBsIC8gMTUlKSlbQHJlc3VsdF07XG4gIC0tdXNpbmctY3NzLXByb3BlcnRpZXM6IGxpZ2h0LWRhcmsoaHNsKGZyb20gdmFyKC0tY29sb3Itd2hpdGUpIGggcyBjYWxjKGwgLSAxMSkpLCBoc2woZnJvbSB2YXIoLS1jb2xvci13aGl0ZSkgaCBzIGwgLyAxNSUpKTtcbn1cblxuIiwiYWN0aXZlVmVyc2lvbiI6IjQuNC4yIiwibWF0aCI6InBhcmVucy1kaXZpc2lvbiIsInN0cmljdFVuaXRzIjpmYWxzZX0=) it compiles just fine for 4.4.0 

As for using a string value `~"hsl(from ....)"` - this is a valid workaround, we are using it for example for CSS min and max functions in other places. I think it is maybe the best solution if we want to keep modern CSS for CSS Custom Properties and not mixing them with less color functions? We will remove less entirely at the end so it's best not to entangle CSS Custom Props with less too much.

TheSpyder: Interesting, the playground does fail if I change the math mode 🤔  I don't know if it's with the same error.

zuzanna-antas: I looked up the math modes and I think changing our math mode is better solution than putting each color function in a string. 

It was never about hsl or relative colors not being correct less code, but about less trying to do math where it shouldn't. Relative colors don't cause a problem. However, passing an alpha channel after `/` causes an error when the math mode is set to `always`. Why? Less does math eagerly. It doesn't require calc() function like CSS does so it tries to do a math operation here. 
In case of relative colors it throws an error but it can also cause silent errors where it just compiles to the wrong value. For example: `rgb(10 10 10 / 10%)`  would compile to `rgb(10, 10, 1)`.

I think we should change math mode to `parens-division`. It would mean we need to wrap division operations inside `()`, but we don't have to wrap modern CSS in strings. Doing math more intentionally. More CSS style 😄

TheSpyder: I agree, is there scope to do that in this PR?

zuzanna-antas: I cerated a separate ticket for changing math mode https://ephocks.atlassian.net/browse/TINY-13317

I am not closing this PR though as the changes are still relevant for following transitional stage guidelines.

PR #10734, opened 2025-11-18
TINY-13224: Fix font-size not applied to a new list item after pressing enter key

tiny-ben-tran:
Related Ticket: TINY-13224

Description of Changes:
* Fixed font-size not applied to a new list item when pressing Enter key

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where font-size styling was not being copied to newly created list items. When pressing Enter in a list item with custom font sizing, the new list item will now correctly inherit and preserve the font-size styling from the original list item.

### DIFF
--- a/.changes/unreleased/tinymce-TINY-13224-2025-11-18.yaml
+++ b/.changes/unreleased/tinymce-TINY-13224-2025-11-18.yaml
@@ -1,6 +1,6 @@
 project: tinymce
 kind: Fixed
-body: '`font-size` on `li` elements was not copied to new list items.'
+body: 'The `font-size` style on `li` elements was not copied to new list items.'
 time: 2025-11-18T11:45:16.117483+10:30
 custom:
     Issue: TINY-13224


modules/tinymce/src/core/main/ts/newline/ReduceNestedFonts.ts
```diff
@@ -9,12 +9,10 @@ const reduceFontStyleNesting = (block: Element, node: Element): void => {
   const hasFontSize = (element: SugarElement<Node>): element is SugarElement<Element> =>
     SugarNode.isElement(element) && Css.getRaw(element, 'font-size').isSome();
 
-  const elementsWithFontSize = [
-    ...(hasFontSize(nodeSugar) ? [ nodeSugar ] : []),
-    ...PredicateFilter.ancestors<Element>(nodeSugar, hasFontSize, isEndBlock)
-  ];
+  const ancestorsWithFontSize = PredicateFilter.ancestors<Element>(nodeSugar, hasFontSize, isEndBlock);
```
tiny-ben-tran: @michalnieruchalski-tiugo, can you review this change? My understanding is that it should only remove the font-size of the targeted element's ancestors.

.changes/unreleased/tinymce-TINY-13224-2025-11-18.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: '`font-size` was not applied to a new list item created when pressing Enter key.'
```
TheSpyder:
```suggestion
body: '`font-size` on `li` elements was not copied to new list items.'
```

modules/tinymce/src/core/main/ts/newline/ReduceNestedFonts.ts
TheSpyder: Sort of. The old implementation was `closest()` (i.e. it checks itself as well as the ancestors), but we never needed that method on the `*Filter` modules (only `*Find` and `*Exists`). We could add it if we need to.

However since the first element was dropped before using the array, it's not clear what the actual goal was and this change might be correct.

TheSpyder (approved):

modules/tinymce/src/core/main/ts/newline/ReduceNestedFonts.ts
tiny-ben-tran: I assume the goal is to remove the duplicated font-size of all ancestors besides the closest one to the targeted element, including itself (hence `slice(1)`). 

It is better to undo this change rather than risk causing another regression.

.changes/unreleased/tinymce-TINY-13224-2025-11-18.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: '`font-size` on `li` elements was not copied to new list items.'
```
spocke: Is it just me or does it looks weird to have back ticks items as the first word.

```suggestion
body: 'The `font-size` style on `li` elements was not copied to new list items when making new lines.'
```

spocke (approved):

.changes/unreleased/tinymce-TINY-13224-2025-11-18.yaml
TheSpyder: I don't think we need `when making new lines`, but yes good point.

spocke: Then it doesn't tell you how to trigger the bug but I guess we could fill that out in release notes.

modules/tinymce/src/core/main/ts/newline/ReduceNestedFonts.ts
michalnieruchalski-tiugo: Yes, the goal was to remove the duplicated font-size of all ancestors besides the closest one to the targeted element, including itself, so:

```
<main>
  <div style="font-size: 15px;">
    <section style="font-size: 14px;">
    </section>
  </div>
</main>
```
If `nodeSugar` would be `main` tag, then the result should be
```
<main>
  <div style="font-size: 15px;">
    <section>
    </section>
  </div>
</main>
```

But in this scenario:
```
<main style="font-size: 16px;">
  <div style="font-size: 15px;">
    <section style="font-size: 14px;">
    </section>
  </div>
</main>
```

the result should be
```
<main style="font-size: 16px;">
  <div>
    <section>
    </section>
  </div>
</main>
```

So It's good that you've reverted this change, as the first scenario would be broken

PR #10736, opened 2025-11-19
TINY-12985: create dropdown component

ztomaszyk:
Related Ticket: TINY-12985

Description of Changes:

- Added a new Dropdown component that provides a button and an anchored container
- Implemented positioning logic for the dropdown content with support for different sides (top, bottom, left, right) and alignments (start, center, end)
- Created Storybook stories to demonstrate all positioning combinations
- Exported the Dropdown component from the main module

Pre-checks:

Review:

- [x] ~~Milestone set~~ (epic branch)

### DIFF
--- a/modules/oxide-components/src/main/ts/components/dropdown/Dropdown.tsx
+++ b/modules/oxide-components/src/main/ts/components/dropdown/Dropdown.tsx
@@ -7,8 +7,8 @@ import { DropdownContext, useDropdown } from './internals/Context';
 import * as PositioningUtils from './internals/PositioningUtils';
 
 export interface DropdownProps extends PropsWithChildren {
-  side?: 'top' | 'bottom' | 'left' | 'right';
-  align?: 'start' | 'center' | 'end';
+  readonly side?: 'top' | 'bottom' | 'left' | 'right';
+  readonly align?: 'start' | 'center' | 'end';
 }
 
 // margin/gap between the trigger button and anchored container
@@ -26,10 +26,10 @@ const Content: FC<PropsWithChildren> = ({ children, ...props }) => {
     if ((event as ToggleEvent).newState === 'open' && triggerRef.current && contentRef.current) {
       const documentRect = document.documentElement.getBoundingClientRect();
       const anchorRect = triggerRef.current.getBoundingClientRect();
-      const contentRect = contentRef.current.getBoundingClientRect();
+      const anchoredContainerRect = contentRef.current.getBoundingClientRect();
 
       // using document rect as a boundry, but maybe it should be the Editor area?
-      setPositioningStyles(PositioningUtils.getPositionStyles(anchorRect, contentRect, side, align, GAP, documentRect));
+      setPositioningStyles(PositioningUtils.getPositionStyles({ anchorRect, anchoredContainerRect, side, align, gap: GAP, boundaryRect: documentRect }));
     }
   }, [ contentRef, triggerRef, align, side ]);
 
--- a/modules/oxide-components/src/main/ts/components/dropdown/internals/Context.ts
+++ b/modules/oxide-components/src/main/ts/components/dropdown/internals/Context.ts
@@ -3,8 +3,8 @@ import { createContext, useContext } from 'react';
 export interface DropdownState {
   readonly popoverId: string;
   readonly triggerRef: React.RefObject<HTMLButtonElement>;
-  side: 'top' | 'bottom' | 'left' | 'right';
-  align: 'start' | 'center' | 'end';
+  readonly side: 'top' | 'bottom' | 'left' | 'right';
+  readonly align: 'start' | 'center' | 'end';
 }
 
 const DropdownContext = createContext<DropdownState | null>(null);
--- a/modules/oxide-components/src/main/ts/components/dropdown/internals/PositioningUtils.ts
+++ b/modules/oxide-components/src/main/ts/components/dropdown/internals/PositioningUtils.ts
@@ -16,14 +16,23 @@ const getMaxWidth = (side: 'left' | 'right', anchorRect: DOMRect, boundaryRect: 
   }
 };
 
-const getPositionStyles = (
-  anchorRect: DOMRect,
-  anchoredContainerRect: DOMRect,
-  side: 'top' | 'bottom' | 'left' | 'right',
-  align: 'start' | 'center' | 'end',
-  gap: number,
-  boundaryRect: DOMRect = document.documentElement.getBoundingClientRect(),
-): CSSProperties => {
+interface GetPositionStylesProps {
+  readonly anchorRect: DOMRect;
+  readonly anchoredContainerRect: DOMRect;
+  readonly side: 'top' | 'bottom' | 'left' | 'right';
+  readonly align: 'start' | 'center' | 'end';
+  readonly gap: number;
+  readonly boundaryRect?: DOMRect;
+}
+
+const getPositionStyles = ( {
+  anchorRect,
+  anchoredContainerRect,
+  side,
+  align,
+  gap,
+  boundaryRect = document.documentElement.getBoundingClientRect()
+}: GetPositionStylesProps): CSSProperties => {
 
   if (side === 'top' || side === 'bottom') {
     const maxHeight = getMaxHeight(side, anchorRect, boundaryRect);


zuzanna-antas:
* **#10736** <a href="https://app.graphite.com/github/pr/tinymce/tinymce/10736?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/tinymce/tinymce/10736?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `epic/TINY-12378`


<h2></h2>

modules/oxide-components/src/main/ts/components/dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,77 @@
...
+import { DropdownContext, useDropdown } from './internals/context';
+import { getPositionStyles } from './internals/positioningUtils';
+
+export interface DropdownProps extends PropsWithChildren {
+  side?: 'top' | 'bottom' | 'left' | 'right';
+  align?: 'start' | 'center' | 'end';
+}
+
+// margin/gap between the trigger button and anchored container
+const GAP = 8;
```
spocke, line 15: This seems like something we want to control from the css side of things. Can this be extracted from some style like margin or something? Or that something we fix when we use proper CSS anchors?

modules/oxide-components/src/main/ts/components/dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+const Content: FC<PropsWithChildren> = ({ children, ...props }) => {
+  const { popoverId, triggerRef, side, align } = useDropdown();
+  const contentRef = useRef<HTMLDivElement>(null);
+
+  const [ positioningStyles, setPositioningStyles ] = useState({});
+
+  // this can be later replaced with CSS anchor positioning
+  const updatePosition = useCallback((event: Event) => {
+    if ((event as ToggleEvent).newState === 'open' && triggerRef.current && contentRef.current) {
```
spocke, line 26: Crappy that the addEventListener doesn't produce the ToggleEvent type so we need this magic cast.

modules/oxide-components/src/main/ts/components/dropdown/internals/Context.ts
spocke, line 1: Can we camelcase this file for consistency with the rest of the code base.

modules/oxide-components/src/main/ts/components/dropdown/internals/PositioningUtils.ts
spocke, line 1: Can we camel case this for consistency with the rest of the code base.

modules/oxide-components/src/main/ts/components/dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,77 @@
...
+  }}>
+    {children}
+  </div>;
+};
+
+const TriggerButton: FC<PropsWithChildren> = ({ children }) => {
+  const { popoverId, triggerRef } = useDropdown();
+
+  // @ts-expect-error - TODO: Remove this expect error once we've upgraded to React 19+
+  return <Button variant='secondary' popovertarget={popoverId} popovertargetaction={'toggle'} ref={triggerRef} >{children}</Button>;
```
spocke: Would we want to some how expose variants etc on this. Or is there some way to make a generic tigger that can contain a button?

spocke:
```suggestion
  return <Button variant='secondary' popovertarget={popoverId} popovertargetaction={'toggle'} ref={triggerRef}>{children}</Button>;
```

modules/oxide-components/src/main/ts/components/dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,77 @@
+import { useCallback, useEffect, useId, useRef, useState, type FC, type PropsWithChildren } from 'react';
+
+import { Button } from '../button/Button';
+
+import { DropdownContext, useDropdown } from './internals/context';
+import { getPositionStyles } from './internals/positioningUtils';
```
spocke: Can the NS import this?

modules/oxide-components/src/main/ts/components/dropdown/Dropdown.tsx
zuzanna-antas: Yeah, I think I can set a margin in CSS instead. Should work fine 👍

modules/oxide-components/src/main/ts/components/dropdown/Dropdown.tsx
zuzanna-antas: Yeah, I don't like it either :(

modules/oxide-components/src/main/ts/components/dropdown/Dropdown.tsx
zuzanna-antas: I think we can expose them the same as we do in IconButton. I wanted to do generic trigger but it has to be a button - the `popovertarget` attribute has to be set on `<button>` or `<input>` element. Can't be a div that contains a button for example.

modules/oxide-components/src/main/ts/components/dropdown/Dropdown.tsx
hamza0867: You can look at the way I did it [here](https://github.com/tinymce/tinymce-premium/blob/epic/TINY-12378/plugins/tinyai/src/main/ts/ui-react/prompt/ModelSelectionButton.tsx#L119). You won't need type casting.

modules/oxide-components/src/main/ts/components/dropdown/Dropdown.tsx
hamza0867: I don't think this should always be a button, instead to should be only a wrapper that attaches the correct props. You can take some inspiration from [here](https://github.com/tinymce/tinymce-premium/blob/epic/TINY-12378/plugins/tinyai/src/main/ts/ui-react/components/Tooltip.tsx#L27-L62). Or we can set the popover api and then spread the props to the Button component

modules/oxide-components/src/main/ts/components/dropdown/Dropdown.stories.tsx
```diff
@@ -0,0 +1,240 @@
+import type { Meta, StoryObj } from '@storybook/react-vite';
+
+import * as Dropdown from './Dropdown';
+
+// More on how to set up stories at: https://storybook.js.org/docs/writing-stories#default-export
```
hamza0867: No need for this comment any more

modules/oxide-components/src/main/ts/components/dropdown/Dropdown.stories.tsx
```diff
@@ -0,0 +1,240 @@
...
+  parameters: {
+    layout: 'fullscreen',
+    docs: {
+      description: {
+        component: `A dropdown component. Contains a button and an anchored container that can be used for creating menus.`
+      }
+    }
+  },
+  // This component will have an automatically generated Autodocs entry: https://storybook.js.org/docs/writing-docs/autodocs
+  tags: [ 'autodocs', 'skip-visual-testing' ],
```
hamza0867: Do we need to skip visual tests here ? I think it might be worth to visually test a few stories at least. What do you think ?

modules/oxide-components/src/main/ts/components/dropdown/Dropdown.tsx
TheSpyder: Providing the types you need with generics is just type casting without the `as` keyword. Still worth doing, it looks cleaner, but please don't assume it's any safer.

modules/oxide-components/src/main/ts/components/dropdown/Dropdown.tsx
TheSpyder: That seems like a question for the ARIA standard. Or, as Zuzanna said in the other comment, it's because `popovertarget` requires a button.

modules/oxide-components/src/main/ts/components/dropdown/Dropdown.tsx
hamza0867: > Still worth doing, it looks cleaner, but please don't assume it's any safer.

Exactly my point 👍

zuzanna-antas: In my opinion it's hidden type casting, I don't think it's any better. Moreover, the newest typescript produces ToggleEvent correctly so maybe instead of changing that I will add a todo to replace `Event` with `ToggleEvent` after we update ts?

modules/oxide-components/src/main/ts/components/dropdown/Dropdown.stories.tsx
zuzanna-antas: The thing is - button needs to be clicked first so the tests make sense. I added custom ones now. Can you take a look?

modules/oxide-components/src/main/ts/components/dropdown/Dropdown.tsx
zuzanna-antas: Unfortunately, it can't be a margin set in CSS. It calculates the position wrong then. Also, I am not sure if this should be configurable or not

modules/oxide-components/src/main/ts/components/dropdown/internals/positioningUtils.ts
```diff
@@ -0,0 +1,77 @@
...
+};
+
+const getPositionStyles = (
+  anchorRect: DOMRect,
+  anchoredContainerRect: DOMRect,
+  side: 'top' | 'bottom' | 'left' | 'right',
+  align: 'start' | 'center' | 'end',
+  gap: number,
+  boundaryRect: DOMRect = document.documentElement.getBoundingClientRect(),
+): CSSProperties => {
```
hamza0867: This function seems to take too many arguments, can we instead change it to take one object argument with 6 properties instead?

modules/oxide-components/src/main/ts/components/dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+import { Bem } from '../../main';
+import { Button, type ButtonProps } from '../button/Button';
+
+import { DropdownContext, useDropdown } from './internals/Context';
+import * as PositioningUtils from './internals/PositioningUtils';
+
+export interface DropdownProps extends PropsWithChildren {
+  side?: 'top' | 'bottom' | 'left' | 'right';
+  align?: 'start' | 'center' | 'end';
```
hamza0867: readonly ?

modules/oxide-components/src/main/ts/components/dropdown/internals/context.ts
```diff
@@ -0,0 +1,20 @@
+import { createContext, useContext } from 'react';
+
+export interface DropdownState {
+  readonly popoverId: string;
+  readonly triggerRef: React.RefObject<HTMLButtonElement>;
+  side: 'top' | 'bottom' | 'left' | 'right';
+  align: 'start' | 'center' | 'end';
```
hamza0867: readonly?

PR #10744, opened 2025-11-24
TINY-13107: Lazy render FloatingSidebar content

michalnieruchalski-tiugo:
Related Ticket: TINY-13107

Description of Changes:
* `FloatingSidebar` content is now lazy rendered, so we render the content once the popover is opened
* This fixes the problem with `AutoResizingTextarea` math being wrong, as now it's being initialized when it's visible

Pre-checks:

Review:
~~* [ ] Milestone set~~

### DIFF
--- a/modules/oxide-components/src/main/ts/components/floatingsidebar/FloatingSidebar.tsx
+++ b/modules/oxide-components/src/main/ts/components/floatingsidebar/FloatingSidebar.tsx
@@ -37,7 +37,7 @@ const transformToCss = (position: InitialPosition): CssPosition => {
 const Root = forwardRef<Ref, FloatingSidebarProps>(({ isOpen = true, height = 600, children, ...props }, ref) => {
   const elementRef = useRef<HTMLDivElement | null>(null);
   const initialPosition = transformToCss(props.initialPosition ?? { x: 0, y: 0, origin: 'topleft' });
-  // The idea is to postpone the rendering of children until the sidebar is opened.
+  // The idea is to render the children after the sidebar has opened.
   // All the children should render once the sidebar is opened.
   const [ shouldRenderChildren, setShouldRenderChildren ] = useState(false);
 


michalnieruchalski-tiugo:
* **#10744** <a href="https://app.graphite.com/github/pr/tinymce/tinymce/10744?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/tinymce/tinymce/10744?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `epic/TINY-12378`


<h2></h2>

modules/oxide-components/src/main/ts/components/autoresizingtextarea/AutoResizingTextarea.tsx
```diff
@@ -47,6 +47,23 @@ export const AutoResizingTextarea = forwardRef<HTMLTextAreaElement, AutoResizing
     }
   }, []);
 
+  // Sometimes this component is rendered hidden (e.g. inside a closed popover), then scrollHeight is 0
+  // and size calculations are wrong. So we use ResizeObserver to detect when the textarea becomes visible
+  useEffect(() => {
+    const observer = new window.ResizeObserver(() => {
```
TheSpyder: The parent control already knows when it's being shown and hidden, using a resize observer on a child to monitor for that change is creating a side effect the parent can't influence. That's a recipe for weird bugs.

Maybe we need a prop or something to manage this? Or perhaps the `toggle`​ event can be captured so it specifically responds to the intended change?

michalnieruchalski-tiugo: `AutoResizingTextarea` should not know where it's being used. It makes measurements based on it's size so I have to make sure it has size, before making the measurements.

One alternative approach that I can think of is that we could expose `recalculate` imperative method out of this component and call it after the popover is opened.

Listening to `toggle` event from within the component seems as problematic as the `ResizeObserver`.

WDYT?

TheSpyder: I don't mean that the textarea should know where it's used, I mean to say that when the parent visibility changes that should trigger the `AutoResizingTextarea`​ function to be evaluated again - at which point if the `scrollHeight`​ has changed it can re-render.

I'm assuming the property change on the parent doesn't re-evaluate the textarea due to refs or some other side effect React isn't keeping track of.

modules/oxide-components/src/main/ts/components/autoresizingtextarea/AutoResizingTextarea.tsx
```diff
@@ -75,10 +85,11 @@ export const AutoResizingTextarea = forwardRef<HTMLTextAreaElement, AutoResizing
     ref={(el) => {
       textareaRef.current = el;
       if (ref) {
+        const elementWithRecalculate = !Type.isNull(el) ? Object.assign(el, { recalculate }) : null;
```
michalnieruchalski-tiugo: @hamza0867 

Do you have any opinions on how to do that correctly? I want to expose the reference for `textarea` element, but I also want to expose `recalculate` function. Should I use imperative handle, and expose the dom reference as a separate property, like:

```
useImperativeHandle(ref, () => {
   return {
     recalculate: () => { ... },
     nativeRef: textareaRef
   };
});
```

modules/oxide-components/src/main/ts/components/autoresizingtextarea/AutoResizingTextarea.tsx
spocke: Couldn't we simply render the sidebar when the popover is shown and empty it when it's hidden. That way it would always re-render every time you open the popver?

spocke: Or would that blink then would it have time to first render and show a empty popover then render in the content?

michalnieruchalski-tiugo: I changed the approach, right now content of the FloatingSidebar is being lazy rendered. So we'll only render content after the popover is opened.

modules/oxide-components/src/main/ts/components/floatingsidebar/FloatingSidebar.tsx
```diff
@@ -37,24 +37,30 @@ const transformToCss = (position: InitialPosition): CssPosition => {
 const Root = forwardRef<Ref, FloatingSidebarProps>(({ isOpen = true, height = 600, children, ...props }, ref) => {
   const elementRef = useRef<HTMLDivElement | null>(null);
   const initialPosition = transformToCss(props.initialPosition ?? { x: 0, y: 0, origin: 'topleft' });
+  // The idea is to postpone the rendering of children until the sidebar is opened.
```
hamza0867: Postpone is a bit of a strong word it in this case I would say.

```suggestion
  // The idea is to render the children after the sidebar has opened.
```

TheSpyder (approved):

PR #10747, opened 2025-11-24
TINY-13197: Fix styles not applied to the list item when its content is wrapped inside a block element

tiny-ben-tran:
Related Ticket: TINY-13197

Description of Changes:
* Updated the logic of apply/remove inline formats for list items to ensure that LI is included even when the content is wrapped inside a block element

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where inline formatting (e.g., bold, color) was not applied to list item markers when a list item's content is wrapped inside block elements.

* **Tests**
  * Added browser tests covering applying/removing inline formats on list items with block-wrapped content, including single/multiple and partial-selection scenarios.

* **Changelog**
  * Added an unreleased changelog entry documenting the fix (TINY-13197).

### DIFF
--- a/.changes/unreleased/tinymce-TINY-13197-2025-11-25.yaml
+++ b/.changes/unreleased/tinymce-TINY-13197-2025-11-25.yaml
@@ -1,6 +1,6 @@
 project: tinymce
 kind: Fixed
-body: Inline formats were not applied to the marker of a list item when its content is wrapped inside a block element.
+body: Inline formats were not applied to the marker of a list item when its content was a single block element.
 time: 2025-11-25T11:50:27.304549+10:30
 custom:
     Issue: TINY-13197


modules/tinymce/src/core/test/ts/browser/fmt/ListItemFormatTest.ts
```diff
@@ -372,6 +476,8 @@ describe('browser.tinymce.core.fmt.ListItemFormatTest', () => {
           expected: '<ul><li style="text-decoration: underline;">abc</li></ul>',
         })
       );
+
+      // it('TINY-13197: Remove all format')
```
ltrouton:
```suggestion
```

.changes/unreleased/tinymce-TINY-13197-2025-11-25.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: Inline formats were not applied to the marker of a list item when its content is wrapped inside a block element.
```
TheSpyder:
```suggestion
body: Inline formats were not applied to the marker of a list item when its content was a single block element.
```

modules/tinymce/src/core/test/ts/browser/fmt/ListItemFormatTest.ts
TheSpyder: Are you going to implement this? If not delete it please

modules/tinymce/src/core/main/ts/fmt/ListItemFormat.ts
```diff
@@ -32,25 +32,44 @@ const isRngEndAtEndOfElement = (rng: Range, elm: Element) => {
...
 const getFullySelectedBlocks = (selection: EditorSelection) => {
   const blocks = selection.getSelectedBlocks();
   const rng = selection.getRng();
 
   if (selection.isCollapsed()) {
     return [];
-  } if (blocks.length === 1) {
+  }
+
+  // TINY-13197: The first block element returned by getSelectedBlocks() is not LI if the content is wrapped inside a block element, even when the content is fully selected. The following logic finds and replaces the first block with its parent LI
```
TheSpyder: Don't we have a line length lint restriction anymore? 😅

modules/tinymce/src/core/main/ts/fmt/ListItemFormat.ts
```diff
@@ -32,25 +32,44 @@ const isRngEndAtEndOfElement = (rng: Range, elm: Element) => {
 
 const isEditableListItem = (dom: DOMUtils) => (elm: Element) => NodeType.isListItem(elm) && dom.isEditable(elm);
 
+const findParentListItem = (selection: EditorSelection, el: Element) => Optional.from(selection.dom.getParent(el, NodeType.isListItem));
```
MitchC1999: Could you use `Traverse.parent`?

modules/tinymce/src/core/main/ts/fmt/ListItemFormat.ts
```diff
@@ -32,25 +32,44 @@ const isRngEndAtEndOfElement = (rng: Range, elm: Element) => {
...
     return [];
-  } if (blocks.length === 1) {
+  }
+
+  // TINY-13197: The first block element returned by getSelectedBlocks() is not LI if the content is wrapped inside a block element, even when the content is fully selected. The following logic finds and replaces the first block with its parent LI
+  if (blocks.length > 0 && isRngStartAtStartOfElement(rng, blocks[0]) && !NodeType.isListItem(blocks[0])) {
+    findParentListItem(selection, blocks[0]).each((listItem) => {
+      blocks[0] = listItem;
+    });
+  }
```
MitchC1999:
```suggestion
  Arr.head(blocks)
    .filter((node) => isRngStartAtStartOfElement(rng, node) && !NodeType.isListItem(node))
    .bind((node) => findParentListItem(selection, node))
    .each((listItem) => {
      blocks[0] = listItem;
    });
  }
```

modules/tinymce/src/core/main/ts/fmt/ListItemFormat.ts
```diff
@@ -32,25 +32,44 @@ const isRngEndAtEndOfElement = (rng: Range, elm: Element) => {
...
-export const getPartiallySelectedListItems = (selection: EditorSelection): Element[] =>
-  Arr.filter(selection.getSelectedBlocks(), isEditableListItem(selection.dom));
+export const getPartiallySelectedListItems = (selection: EditorSelection): Element[] => {
+  const blocks = selection.getSelectedBlocks();
+  // TINY-13197 The first block element returned by getSelectedBlocks() is not LI if the content is wrapped inside a block element, even when the content is fully selected. The following logic finds and replaces the first block with its parent LI
+  if (blocks.length > 0 && !NodeType.isListItem(blocks[0])) {
+    findParentListItem(selection, blocks[0]).each((listItem) => {
+      blocks[0] = listItem;
+    });
+  }
```
MitchC1999: Similarly, you could use `Arr.head`, `filter`, and `bind` here.

modules/tinymce/src/core/main/ts/fmt/ListItemFormat.ts
TheSpyder: I also suggest adding some clarity around this - this only needs to edit the first block because when there are multiple list items selected, the second and subsequent _do_ return the `li` as the selected block. So really this is just normalising the selection (and one way to implement this would've been dealing with this in the selection normalisation code instead of the list code).

modules/tinymce/src/core/main/ts/fmt/ListItemFormat.ts
```diff
@@ -32,25 +32,44 @@ const isRngEndAtEndOfElement = (rng: Range, elm: Element) => {
...
 
   if (selection.isCollapsed()) {
     return [];
-  } if (blocks.length === 1) {
+  }
+
+  // TINY-13197: The first block element returned by getSelectedBlocks() is not LI if the content is wrapped inside a block element, even when the content is fully selected. The following logic finds and replaces the first block with its parent LI
+  if (blocks.length > 0 && isRngStartAtStartOfElement(rng, blocks[0]) && !NodeType.isListItem(blocks[0])) {
+    findParentListItem(selection, blocks[0]).each((listItem) => {
+      blocks[0] = listItem;
```
TheSpyder: You're not checking if the block is the only child of the list item. What happens if there is a list with two paragraphs in it, and only the second is selected?

modules/tinymce/src/core/main/ts/fmt/ListItemFormat.ts
TheSpyder: de-dupe with the fully selected code please. Looks like the only difference is the start of element check.

modules/tinymce/src/core/main/ts/fmt/ListItemFormat.ts
ltrouton: I am not a massive fan of the idea of mutating the original `blocks` variable

Would it simpler to just do a map over all of the blocks e.g.
```suggestion
 const normalisedBlocks = Arr.map(blocks, (block) => {
 	if (!NodeType.isListItem(block) {
 		return findParentListItem(selection, block).getOr(block);
 	} else {
 	  return block;
 	}
 })l
```

ltrouton (approved):

modules/tinymce/src/core/main/ts/fmt/ListItemFormat.ts
ltrouton: I see Mitch has another suggestion which might work as well

modules/tinymce/src/core/main/ts/fmt/ListItemFormat.ts
tiny-ben-tran: > Don't we have a line length lint restriction anymore? 😅

The max line length is set to 260 at the moment. It seems really long 😄 


> I also suggest adding some clarity around this - this only needs to edit the first block because when there are multiple list items selected, the second and subsequent _do_ return the `li` as the selected block. So really this is just normalising the selection (and one way to implement this would've been dealing with this in the selection normalisation code instead of the list code).

This is correct. I didn't really want to deal with selection normalisation code because I am afraid of introducing a regression.

modules/tinymce/src/core/main/ts/fmt/ListItemFormat.ts
tiny-ben-tran: Maybe. I opted for `selection.dom.parent` because it takes a selector for list items

modules/tinymce/src/core/main/ts/fmt/ListItemFormat.ts
tiny-ben-tran: > What happens if there is a list with two paragraphs in it, and only the second is selected?

I assume the list item style should not be updated, as observed with LI with content separated by `<br>`.

tiny-ben-tran: What if both paragraphs are selected?

TheSpyder: If both paragraphs are selected it should update (because then all the content is selected).

I asked because this code looks like it will update the list item if only the second paragraph is selected. And I mean two `p`​ tags, not two lines separated by `br`​ tag. It's a rare case, but I've seen weirder.

modules/tinymce/src/core/main/ts/fmt/ListItemFormat.ts
spocke: `getParent`​ is weird it more like `PredicateFind.closest`​ so `Traverse.parent`​ would be wrong if the intent is to treverse up until you find the LI.

modules/tinymce/src/core/main/ts/fmt/ListItemFormat.ts
tiny-ben-tran: `isRngStartAtStartOfElement` check ensures that the list item is updated only when all content is selected.

Generally, I replicate how the formatting works for text content (without block element). Hopefully it is good enough

modules/tinymce/src/core/main/ts/fmt/ListItemFormat.ts
tiny-ben-tran: The intention is to traverse up until LI is found so using `getParent` is more appropriate

tiny-ben-tran:
@ltrouton @MitchC1999 I refactored the code based on the idea in your suggestions.

modules/tinymce/src/core/main/ts/fmt/ListItemFormat.ts
```diff
@@ -32,25 +32,44 @@ const isRngEndAtEndOfElement = (rng: Range, elm: Element) => {
...
 
-const getFullySelectedBlocks = (selection: EditorSelection) => {
-  const blocks = selection.getSelectedBlocks();
-  const rng = selection.getRng();
+// TINY-13197: If the content is wrapped inside a block element, the first block returned by getSelectedBlocks() is not LI, even when the content is fully selected.
+// However, the second and subsequent do return LI as the selected block so only the first block needs to be adjusted
+const normalizeFirstBlockIf = (selection: EditorSelection, blocks: Element[], pred: (block: Element) => boolean) =>
+  Arr.map(blocks, (block, i) => {
+    if (i === 0 && pred(block)) {
+      return Optional.from(selection.dom.getParent(block, NodeType.isListItem)).getOr(block);
```
TheSpyder: I think I'm going to start suggesting this more often in response to `Optional.from(...).getOr()`​. Optionals still have their place but the language has fine tools for this now.

```suggestion
      return selection.dom.getParent(block, NodeType.isListItem) ?? block;
```

modules/tinymce/src/core/main/ts/fmt/ListItemFormat.ts
```diff
@@ -32,25 +32,44 @@ const isRngEndAtEndOfElement = (rng: Range, elm: Element) => {
...
   }
 };
 
 export const getFullySelectedListItems = (selection: EditorSelection): Element[] =>
   Arr.filter(getFullySelectedBlocks(selection), isEditableListItem(selection.dom));
 
-export const getPartiallySelectedListItems = (selection: EditorSelection): Element[] =>
-  Arr.filter(selection.getSelectedBlocks(), isEditableListItem(selection.dom));
+export const getPartiallySelectedListItems = (selection: EditorSelection): Element[] => {
+  const blocks = normalizeFirstBlockIf(selection, selection.getSelectedBlocks(), (el) => !NodeType.isListItem(el));
```
TheSpyder: You're passing the `selection`​ in, you don't need to pass `selection.getSelectedBlocks`​ as well :thinking_face:

modules/tinymce/src/core/main/ts/fmt/ListItemFormat.ts
TheSpyder: Yes, but that's only checking if the first result in the list of selected blocks is selected. If the list item has two paragraphs, and only the second is selected, `getSelectedBlocks()`​ returns the second paragraph as the first result.

And then I think your code will apply the format to the list item.

- [x] Please write a test for this, if nothing else

tiny-ben-tran: I believe [this test](https://github.com/tinymce/tinymce/pull/10747/files#diff-dea049579947e24be80725ef2e0e1eb3fe9d0b949bdaceee24ab00b94eab0940R289) covers that case

modules/tinymce/src/core/main/ts/fmt/ListItemFormat.ts
tiny-ben-tran: I renamed the function and removed passing `selection.getSelectedBlocks`

modules/tinymce/src/core/main/ts/fmt/ListItemFormat.ts
TheSpyder: Can you please give me the name instead of a link? github permalinks don't work very well when I've marked a file as viewed, and even if I unmark it I just get the file highlighted when clicking on the link.

TheSpyder: ah I found it, `Should only apply color to the selected blocks and not update the first LI as it is not fully selected`​

yep that's fine sorry for the hassle

tiny-ben-tran: The test name: `TINY-13197: Should only apply color to the selected blocks and not update the first LI as it is not fully selected`

PR #10748, opened 2025-11-25
TINY-13024: Add popover props and improve anchor behaviour

shanmen-tiny:
Related Ticket: TINY-13204

Description of Changes:
* Add a  new `popover` prop to the `ContextToolbar` component, allowing the toolbar to operate in two distinct modes:
	- **Popover mode (`popover={true}`)** — uses the native Popover API (`showPopover` / `hidePopover`) to render the toolbar in the top layer.
	- **Anchored mode (`popover={false}`)** — positions the toolbar relative to an anchor element using CSS anchors and remains inside its container (required for iframe overlays).

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/oxide-components/src/main/ts/components/contexttoolbar/ContextToolbar.stories.tsx
+++ b/modules/oxide-components/src/main/ts/components/contexttoolbar/ContextToolbar.stories.tsx
@@ -57,7 +57,7 @@ The provider component that manages toolbar state.
 
 **Props:**
 - \`persistent\`: (optional, default: \`false\`). If true, the toolbar stays open when clicking outside of it.
-- \`popover\`: (optional, default: \`false\`). Enables Native Popover API mode. When \`true\`, toolbar uses \`showPopover()\` and \`hidePopover()\` to appear in the top layer. When \`false\`, the toolbar behaves as a regular absolutely-positioned element, typically used when anchoring inside an iframe overlay where Popover cannot work.
+- \`usePopover\`: (optional, default: \`false\`). Enables Native Popover API mode. When \`true\`, toolbar uses \`showPopover()\` and \`hidePopover()\` to appear in the top layer. When \`false\`, the toolbar behaves as a regular absolutely-positioned element, typically used when anchoring inside an iframe overlay where Popover cannot work.
 - \`anchorRef\`: (optional). A ref to an external element to anchor the toolbar to. When provided, the toolbar positions relative to this element instead of the Trigger component. The toolbar auto-opens on mount if no Trigger component is present.
 
 **Important:**


modules/oxide-components/src/main/ts/components/contexttoolbar/ContextToolbar.stories.tsx
```diff
@@ -57,18 +57,19 @@ The provider component that manages toolbar state.
 
 **Props:**
 - \`persistent\`: (optional, default: \`false\`). If true, the toolbar stays open when clicking outside of it.
+- \`popover\`: (optional, default: \`false\`). Enables Native Popover API mode. When \`true\`, toolbar uses \`showPopover()\` and \`hidePopover()\` to appear in the top layer. When \`false\`, the toolbar behaves as a regular absolutely-positioned element, typically used when anchoring inside an iframe overlay where Popover cannot work.
```
kimwoodfield:
```suggestion
- \`usePopover\`: (optional, default: \`false\`). Enables Native Popover API mode. When \`true\`, toolbar uses \`showPopover()\` and \`hidePopover()\` to appear in the top layer. When \`false\`, the toolbar behaves as a regular absolutely-positioned element, typically used when anchoring inside an iframe overlay where Popover cannot work.
```

kimwoodfield (approved):
Makes sense and the example from the Story works great 👍🏻

modules/oxide-components/src/main/ts/components/contexttoolbar/ContextToolbar.tsx
```diff
@@ -144,33 +146,40 @@ const Toolbar: FC<ToolbarProps> = ({
...
+    }
+
+    if (usePopover) {
+      element.showPopover();
+    }
+    // Defer focus using queueMicrotask to ensure it runs after
+    // the Popover API's internal focus management is complete
+    window.queueMicrotask(() => {
+      const sugarElement = SugarElement.fromDom(element);
+      const firstGroup = SelectorFind.descendant(sugarElement, '.tox-toolbar__group');
```
hamza0867, line 172: Maybe we should use the `Bem.selector`​ helper here, instead of the string `.tox-toolbar__group`​

modules/oxide-components/src/main/ts/components/contexttoolbar/ContextToolbar.tsx
```diff
@@ -144,33 +146,40 @@ const Toolbar: FC<ToolbarProps> = ({
...
+      return;
+    }
+    if (!isOpen) {
+      if (usePopover) {
         element.hidePopover();
       }
-    };
-  }, [ isOpen, toolbarRef ]);
+      return;
+    }
```
TheSpyder, line 163: Hmm I'm a bit sad to lose popover. But I can see the problem, a movable toolbar that uses CSS anchors is incompatible with popover (unless the bounds of the movement are also in popover, and that's no better than `position: fixed`​).

I think what we'll find, long term, is that a "context toolbar" should _never_ use popover and the anchor elements should be controlled by the toolbar directly instead of patched in with a ref.

But let's get this in for now.

modules/oxide-components/src/main/ts/components/contexttoolbar/ContextToolbar.tsx
TheSpyder: The whole thing needs to be checked based on Piotr's review

TheSpyder (approved):

PR #10756, opened 2025-11-27
TINY-13236: create menu component and add hoverable behavior to dropdown

ztomaszyk:
Related Ticket: TINY-13236

Description of Changes:
* Added `gap` and `triggersOnHover` props to the Dropdown component
* Implemented hover behavior for dropdowns that opens on hover and closes on mouse leave
* Created a new Menu component with support for simple menu items, toggle menu items, and submenus
* Added keyboard navigation support for the Menu component
* Added utility function to hide all open dropdowns
* Added stories for the Menu component to showcase its functionality

Pre-checks:
* [x] ~~Tests have been added (if applicable)~~ (Added a separate ticket)

Review:
* [x] Milestone set (epic branch)

### DIFF
--- a/modules/oxide-components/src/main/ts/components/menu/components/Item.tsx
+++ b/modules/oxide-components/src/main/ts/components/menu/components/Item.tsx
@@ -13,8 +13,6 @@ export const Item = forwardRef<HTMLButtonElement, MenuItemProps>(({ autoFocus = 
   });
   const id = useId();
 
-  const [ itemIcon, setItemIcon ] = useState<JSX.Element>();
-
   useEffect(() => {
     setState((prevState) => ({ ...prevState, enabled }));
   }, [ enabled ]);
@@ -33,15 +31,9 @@ export const Item = forwardRef<HTMLButtonElement, MenuItemProps>(({ autoFocus = 
     }
   }, [ onSetup, api ]);
 
-  useEffect(() => {
-    if (Type.isNonNullable(icon)) {
-      if (Type.isString(icon)) {
-        setItemIcon(<Icon icon={icon} resolver={iconResolver}/>);
-      } else {
-        setItemIcon(icon);
-      }
-    }
-  }, [ icon, iconResolver ]);
+  const itemIcon = Type.isString(icon)
+    ? <Icon icon={icon} resolver={iconResolver} />
+    : icon;
 
   return (
     <button
@@ -63,9 +55,7 @@ export const Item = forwardRef<HTMLButtonElement, MenuItemProps>(({ autoFocus = 
       autoFocus={autoFocus}
       aria-keyshortcuts={shortcut}
     >
-      <div className={Bem.element('tox-collection', 'item-icon')}>
-        {itemIcon}
-      </div>
+      {itemIcon && <div className={Bem.element('tox-collection', 'item-icon')}>{itemIcon}</div>}
       <div className={Bem.element('tox-collection', 'item-label')}>{children}</div>
       {shortcut && <div className={Bem.element('tox-collection', 'item-accessory')}>{shortcut}</div>}
     </button>
--- a/modules/oxide-components/src/main/ts/components/menu/components/SubmenuItem.tsx
+++ b/modules/oxide-components/src/main/ts/components/menu/components/SubmenuItem.tsx
@@ -14,8 +14,6 @@ export const SubmenuItem = forwardRef<HTMLButtonElement, SubmenuProps>(({ autoFo
   });
   const id = useId();
 
-  const [ itemIcon, setItemIcon ] = useState<JSX.Element>();
-
   useEffect(() => {
     setState((prevState) => ({ ...prevState, enabled }));
   }, [ enabled ]);
@@ -35,15 +33,9 @@ export const SubmenuItem = forwardRef<HTMLButtonElement, SubmenuProps>(({ autoFo
     return Fun.noop;
   }, [ onSetup, api ]);
 
-  useEffect(() => {
-    if (Type.isNonNullable(icon)) {
-      if (Type.isString(icon)) {
-        setItemIcon(<Icon icon={icon} resolver={iconResolver}/>);
-      } else {
-        setItemIcon(icon);
-      }
-    }
-  }, [ icon, iconResolver ]);
+  const itemIcon = Type.isString(icon)
+    ? <Icon icon={icon} resolver={iconResolver} />
+    : icon;
 
   return (
     <Dropdown.Root side={submenusSide} align={'start'} triggerEvents={[ 'click', 'hover' ]}>
@@ -66,9 +58,7 @@ export const SubmenuItem = forwardRef<HTMLButtonElement, SubmenuProps>(({ autoFo
           })}
           autoFocus={autoFocus}
         >
-          <div className={Bem.element('tox-collection', 'item-icon')}>
-            {itemIcon}
-          </div>
+          {itemIcon && <div className={Bem.element('tox-collection', 'item-icon')}>{itemIcon}</div>}
           <div className={Bem.element('tox-collection', 'item-label')}>{children}</div>
           <div className={Bem.element('tox-collection', 'item-caret')}>
             {iconResolver && <Icon resolver={iconResolver} icon={'chevron-right'}></Icon>}
--- a/modules/oxide-components/src/main/ts/components/menu/components/ToggleItem.tsx
+++ b/modules/oxide-components/src/main/ts/components/menu/components/ToggleItem.tsx
@@ -14,8 +14,6 @@ export const ToggleItem = forwardRef<HTMLButtonElement, ToggleMenuItemProps>(({ 
   });
   const id = useId();
 
-  const [ itemIcon, setItemIcon ] = useState<JSX.Element>();
-
   useEffect(() => {
     setState((prevState) => ({ ...prevState, enabled, active }));
   }, [ enabled, active ]);
@@ -39,15 +37,9 @@ export const ToggleItem = forwardRef<HTMLButtonElement, ToggleMenuItemProps>(({ 
     return Fun.noop;
   }, [ onSetup, api ]);
 
-  useEffect(() => {
-    if (Type.isNonNullable(icon)) {
-      if (Type.isString(icon)) {
-        setItemIcon(<Icon icon={icon} resolver={iconResolver}/>);
-      } else {
-        setItemIcon(icon);
-      }
-    }
-  }, [ icon, iconResolver ]);
+  const itemIcon = Type.isString(icon)
+    ? <Icon icon={icon} resolver={iconResolver} />
+    : icon;
 
   return (
     <button
@@ -71,9 +63,7 @@ export const ToggleItem = forwardRef<HTMLButtonElement, ToggleMenuItemProps>(({ 
       autoFocus={autoFocus}
       aria-keyshortcuts={shortcut}
     >
-      <div className={Bem.element('tox-collection', 'item-icon')}>
-        {itemIcon}
-      </div>
+      {itemIcon && <div className={Bem.element('tox-collection', 'item-icon')}>{itemIcon}</div>}
       <div className={Bem.element('tox-collection', 'item-label')}>{children}</div>
       {shortcut && <div className={Bem.element('tox-collection', 'item-accessory')}>{shortcut}</div>}
       <div className={Bem.element('tox-collection', 'item-checkmark')} >
--- a/modules/oxide-components/src/test/ts/browser/components/__snapshots__/Menu.spec.tsx.snap
+++ b/modules/oxide-components/src/test/ts/browser/components/__snapshots__/Menu.spec.tsx.snap
@@ -21,9 +21,6 @@ exports[`browser.MenuTest > Should be able to open submenus > 1. Before open sub
           tabindex="-1"
         >
           <div
-            class="tox-collection__item-icon"
-          />
-          <div
             class="tox-collection__item-label"
           >
             Menu item 1
@@ -38,9 +35,6 @@ exports[`browser.MenuTest > Should be able to open submenus > 1. Before open sub
           role="menuitemcheckbox"
           tabindex="-1"
         >
-          <div
-            class="tox-collection__item-icon"
-          />
           <div
             class="tox-collection__item-label"
           >
@@ -213,9 +207,6 @@ exports[`browser.MenuTest > Should be able to open submenus > 2. After opening s
           tabindex="-1"
         >
           <div
-            class="tox-collection__item-icon"
-          />
-          <div
             class="tox-collection__item-label"
           >
             Menu item 1
@@ -230,9 +221,6 @@ exports[`browser.MenuTest > Should be able to open submenus > 2. After opening s
           role="menuitemcheckbox"
           tabindex="-1"
         >
-          <div
-            class="tox-collection__item-icon"
-          />
           <div
             class="tox-collection__item-label"
           >
@@ -394,9 +382,6 @@ exports[`browser.MenuTest > Should be able to open submenus > 2. After opening s
                 tabindex="-1"
               >
                 <div
-                  class="tox-collection__item-icon"
-                />
-                <div
                   class="tox-collection__item-label"
                 >
                   Nested menu item 1
@@ -411,9 +396,6 @@ exports[`browser.MenuTest > Should be able to open submenus > 2. After opening s
                 role="menuitemcheckbox"
                 tabindex="-1"
               >
-                <div
-                  class="tox-collection__item-icon"
-                />
                 <div
                   class="tox-collection__item-label"
                 >
@@ -506,9 +488,6 @@ exports[`browser.MenuTest > Should be able to render using MenuRenderer > 1. Bef
           tabindex="-1"
         >
           <div
-            class="tox-collection__item-icon"
-          />
-          <div
             class="tox-collection__item-label"
           >
             Menu item 1
@@ -523,9 +502,6 @@ exports[`browser.MenuTest > Should be able to render using MenuRenderer > 1. Bef
           role="menuitemcheckbox"
           tabindex="-1"
         >
-          <div
-            class="tox-collection__item-icon"
-          />
           <div
             class="tox-collection__item-label"
           >
@@ -597,9 +573,6 @@ exports[`browser.MenuTest > Should be able to render using MenuRenderer > 1. Bef
           style="box-sizing: border-box; width: 100%;"
           tabindex="-1"
         >
-          <div
-            class="tox-collection__item-icon"
-          />
           <div
             class="tox-collection__item-label"
           >
@@ -694,9 +667,6 @@ exports[`browser.MenuTest > Should be able to render using MenuRenderer > 2. Aft
           tabindex="-1"
         >
           <div
-            class="tox-collection__item-icon"
-          />
-          <div
             class="tox-collection__item-label"
           >
             Menu item 1
@@ -711,9 +681,6 @@ exports[`browser.MenuTest > Should be able to render using MenuRenderer > 2. Aft
           role="menuitemcheckbox"
           tabindex="-1"
         >
-          <div
-            class="tox-collection__item-icon"
-          />
           <div
             class="tox-collection__item-label"
           >
@@ -785,9 +752,6 @@ exports[`browser.MenuTest > Should be able to render using MenuRenderer > 2. Aft
           style="box-sizing: border-box; width: 100%;"
           tabindex="-1"
         >
-          <div
-            class="tox-collection__item-icon"
-          />
           <div
             class="tox-collection__item-label"
           >
@@ -871,9 +835,6 @@ exports[`browser.MenuTest > Should be able to render using MenuRenderer > 2. Aft
                 tabindex="-1"
               >
                 <div
-                  class="tox-collection__item-icon"
-                />
-                <div
                   class="tox-collection__item-label"
                 >
                   Nested menu item 1
@@ -888,9 +849,6 @@ exports[`browser.MenuTest > Should be able to render using MenuRenderer > 2. Aft
                 role="menuitemcheckbox"
                 tabindex="-1"
               >
-                <div
-                  class="tox-collection__item-icon"
-                />
                 <div
                   class="tox-collection__item-label"
                 >


zuzanna-antas:
* **#10816** <a href="https://app.graphite.com/github/pr/tinymce/tinymce/10816?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10783** <a href="https://app.graphite.com/github/pr/tinymce/tinymce/10783?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10766** <a href="https://app.graphite.com/github/pr/tinymce/tinymce/10766?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10756** <a href="https://app.graphite.com/github/pr/tinymce/tinymce/10756?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/tinymce/tinymce/10756?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `epic/TINY-12378`


<h2></h2>

modules/oxide-components/src/main/ts/components/dropdown/Dropdown.tsx
```diff
@@ -1,22 +1,19 @@
...
 
-export interface DropdownProps extends PropsWithChildren {
-  readonly side?: 'top' | 'bottom' | 'left' | 'right';
-  readonly align?: 'start' | 'center' | 'end';
-}
-
-// margin/gap between the trigger button and anchored container
-const GAP = 8;
+const isInDropdownContent = (contentRef: React.RefObject<HTMLDivElement>, node: Node): boolean => {
+  return contentRef.current?.contains(node) || false;
```
spocke: Could this be less ninja like a ternary with maybe a katmari Type.isXXX check?

modules/oxide-components/src/main/ts/utils/Dropdowns.ts
```diff
@@ -0,0 +1,5 @@
+export const hideAll = (): void => {
+  document.querySelectorAll('.tox-dropdown-content:popover-open').forEach((popover) => {
```
spocke: We can use sugar, katmari etc. So it would be.

```typescript
Arr.each(SelectorAll.decendants<HTMLElement>('.tox-dropdown-content:popover-open'), (el) => el.hidePopover());
```

If we start dealing with these popovers often maybe we should have a popover module in Sugar that has functions hide `hidePopover(el)` since then this could have been:

```typescript
Arr.each(SelectorAll.decendants<HTMLElement>('.tox-dropdown-content:popover-open'), SugarPopover.hidePopover);
```

spocke (approved):

modules/oxide-components/src/main/ts/components/menu/internals/Types.ts
```diff
@@ -0,0 +1,72 @@
...
+  readonly items: MenuItem[];
+  readonly onSetup?: (api: CommonMenuItemInstanceApi) => (api: CommonMenuItemInstanceApi) => void;
+}
+
+export type MenuItem = ToggleMenuItem | SimpleMenuItem | Submenu;
+
+// Components Props
+export interface ToggleMenuItemProps extends ToggleMenuItem {
+  id: string;
+  iconResolver: (icon: string) => string;
```
spocke, line 37: This really needs to go. We need that global universe context provider. I think we have a ticket for it.

modules/oxide-components/src/main/ts/components/menu/Menu.tsx
```diff
@@ -0,0 +1,244 @@
...
+      setState((prev) => ({ ...prev, enabled: newEnabled }));
+    }
+  }), [ state ]);
+
+  useEffect(() => {
+    if (onSetup) {
+      const teardown = onSetup(api);
+      return () => teardown(api);
+    }
+    return Fun.noop;
```
spocke: Do we even need to return something here?

modules/oxide-components/src/main/ts/components/menu/Menu.tsx
```diff
@@ -0,0 +1,244 @@
...
+    selector: '.tox-collection__item:not([aria-disabled="true"])',
+    allowHorizontal: false,
+    cycles: false
+  });
+  useEffect(() => {
+    if (autoFocus && firstItemRef.current) {
+      firstItemRef.current.focus();
+    }
+  }, [ autoFocus ]);
+  const itemsWithId = useMemo(() => Arr.map(items, (itemProps) => ({ ...itemProps, id: Id.generate('menu-item') })), [ items ]);
```
spocke: There is a built in id generator in react I think we used that for other things.

modules/oxide-components/src/main/ts/components/menu/Menu.tsx
```diff
@@ -0,0 +1,244 @@
+import { Arr, Fun, Id } from '@ephox/katamari';
+import { forwardRef, useEffect, useMemo, useRef, useState, type FC } from 'react';
+
+import { Icon } from '../../internal/icon/Icon.component';
+import { Bem, Dropdown, KeyboardNavigationHooks } from '../../main';
```
spocke: I don't think we should import from `main`​ use relative imports for these different things. I think we have a lint rule if this was setup like other libraries. We should really fix that make this `api/Main.ts`​ instead.

modules/oxide-components/src/main/ts/components/menu/Menu.stories.tsx
```diff
@@ -0,0 +1,193 @@
...
+      story: {
+        inline: false,
+        iframeHeight: 700
+      }
+    }
+  },
+  render: (args: MenuProps): JSX.Element => (<Menu items={args.items} iconResolver={args.iconResolver}></Menu>)
+};
+
+export const MenuInADropdown: Story = {
```
michalnieruchalski-tiugo, line 167: There's a bug in this story. Every time I click on the Icon the menu is first rendered in the top left corner of the page, and then it moves where it's supposed to. It's causing flickering effect. We can sync if you can't replicate it

modules/oxide-components/src/main/ts/components/dropdown/Dropdown.tsx
```diff
@@ -45,24 +42,70 @@ const Content: FC<PropsWithChildren> = ({ children, ...props }) => {
...
+  const props = {
+    popovertarget: popoverId,
+    popovertargetaction: 'toggle',
+    ref: triggerRef,
+    children,
+    ...triggersOnHover && onHoverTriggerProps,
+    ...args
+  };
+
+  return (<>{ args.variant ? <Button { ...props } variant={args.variant} /> : <button { ...props }/> }</>);
```
michalnieruchalski-tiugo: I can see that you needed a native button inside the Menu, rather than our styled `Button`
I would advise us to take a step back, maybe we should decouple `Dropdown.Trigger` from the `Button` underneath. 

We could define Dropdown as follows:
```
<Dropdown.Root>
  <Dropdown.Trigger>
      <Button variant={'secondary'}></Button> // or <button, or <IconButton (doesn't matter)
  </Dropdown.Trigger>
   <Dropdown.Content>
      <Menu items={args.items} iconResolver={args.iconResolver}></Menu>
   </Dropdown.Content>
</Dropdown.Root>
```

`Dropdown.Trigger` could have event listeners, and it's the responsibility of developer to provide what element should be used as a trigger.

It appears that this is how radix works for example https://www.radix-ui.com/themes/docs/components/dropdown-menu

Another example would be https://headlessui.com/react/menu#menu-button 
They have `MenuButton` component, but they provide other ways to customize the trigger.

Conditional rendering based on `variant` property seems a bit messy in my opinion.

modules/oxide-components/src/main/ts/components/dropdown/Dropdown.tsx
```diff
@@ -45,24 +42,70 @@ const Content: FC<PropsWithChildren> = ({ children, ...props }) => {
...
-const Root: FC<DropdownProps> = ({ children, side = 'top', align = 'start' }) => {
+export interface DropdownProps extends PropsWithChildren {
+  readonly side?: 'top' | 'bottom' | 'left' | 'right';
+  readonly align?: 'start' | 'center' | 'end';
+  // margin/gap between the trigger button and anchored container
+  readonly gap?: number;
+  readonly triggersOnHover?: boolean;
+}
+
+const Root: FC<DropdownProps> = ({ children, side = 'top', align = 'start', gap = 8, triggersOnHover = false }) => {
```
michalnieruchalski-tiugo: (could)
How about `triggerEvent: 'click' | 'hover' = 'click'` ? It will be easier to add other events, like dblclick or contextmenu :)

modules/oxide-components/src/main/ts/components/menu/Menu.tsx
```diff
@@ -0,0 +1,244 @@
...
+
+import type { CommonMenuItemInstanceApi, MenuProps, SimpleMenuItemProps, SubmenuProps, ToggleMenuItemInstanceApi, ToggleMenuItemProps } from './internals/Types';
+
+const SubmenuItem = forwardRef<HTMLDivElement, SubmenuProps>(({ id, enabled = true, text, icon, iconResolver, onSetup, items, submenusSide }, ref) => {
```
michalnieruchalski-tiugo: I find this file hard to swallow. I would prefer to split it into smaller chunks. For me `SubmenuItem` `SimpleMenuItem` and `ToggleMenuItem` each should be in separate file.

When I open `Menu.tsx` I want to see how Menu component works.

It's just a preference of course, so I'll be happy to hear how others feel.

modules/oxide-components/src/main/ts/components/menu/Menu.tsx
michalnieruchalski-tiugo: There's `useId` https://react.dev/reference/react/useId but it's a hook so I don't know if it fits here

modules/oxide-components/src/main/ts/components/menu/Menu.tsx
```diff
@@ -0,0 +1,244 @@
...
+  useEffect(() => {
+    if (autoFocus && firstItemRef.current) {
+      firstItemRef.current.focus();
+    }
+  }, [ autoFocus ]);
+  const itemsWithId = useMemo(() => Arr.map(items, (itemProps) => ({ ...itemProps, id: Id.generate('menu-item') })), [ items ]);
+  return (
+    <div ref={ref} role='menu' className={[ Bem.block('tox-menu'), Bem.block('tox-collection', { list: true }) ].join(' ')}>
+      <div className={Bem.element('tox-collection', 'group')}>
+        {
```
michalnieruchalski-tiugo: In the current implementation, developer has to provide a javascript object with menu definition.
so it's 

```
<Menu item={here object with javascript definition}`></Menu>
```

I don't feel like this is the path we should follow. I would expect it to be:
```
<Menu>
   <MenuItem select={{ doSomething }}>Option one</MenuItem>
   <MenuItem select={{ doSomething }}>Option two</MenuItem>
   <Menu.Divider></Menu.Divider>
   <Submenu>
        <Submenu.Button>Option three</Submenu.Button>
        <MenuItem ...
        <MenuItem ...
   </Submenu>
</Menu>
```

So to use `jsx` to define the `Menu` not some custom format.
I've checked how other UI frameworks handle it, see:

https://www.radix-ui.com/themes/docs/components/dropdown-menu
https://headlessui.com/react/menu
https://mui.com/material-ui/react-select/

With current approach developer can only achieve what's possible through the already defined API, so it's less flexible. Moreover, adding new elements (like `Divider`) requires changes in `Menu.tsx` this map function can grow into something big and messy. 

For me it's open-closed principle. With `jsx` approach you can add new functionalities to `Menu` by defining new components and you don't need to modify the existing ones, whereas here you always have to update `Menu` component.

michalnieruchalski-tiugo: I can see that we wanted to support our previous format of defining menus but maybe we should have some kind of `adapter` function? Do we really want to keep defining menus that way?

modules/oxide-components/src/main/ts/components/dropdown/Dropdown.tsx
TheSpyder: agreed

modules/oxide-components/src/main/ts/components/menu/Menu.tsx
zuzanna-antas: I am not really sure, the current implementation works more like an API for menus. You right that when we will need to add new type of elements we will need to change the menu component. On the other hand, we restrict the usage to these item types only (not a bad thing IMO).   
What do you think @hamza0867? You suggested that approach :)

modules/oxide-components/src/main/ts/components/menu/Menu.tsx
hamza0867: Yes we do otherwise we get an eslint error for inconsistent return types

modules/oxide-components/src/main/ts/components/menu/Menu.stories.tsx
```diff
@@ -0,0 +1,193 @@
...
+    <defs></defs>
+    <g stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
+        <path d="M18.1679497,5.4452998 C18.4743022,4.98577112 19.0951715,4.86159725 19.5547002,5.16794971 C20.0142289,5.47430216 20.1384028,6.09517151 19.8320503,6.5547002 L11.8320503,18.5547002 C11.4831227,19.0780915 10.7433669,19.1531818 10.2963845,18.7105809 L5.29919894,13.7623796 C4.90675595,13.3737835 4.90363744,12.7406262 5.29223356,12.3481832 C5.68082968,11.9557402 6.31398698,11.9526217 6.70642997,12.3412178 L10.8411868,16.4354442 L18.1679497,5.4452998 Z" fill="#000000" fill-rule="nonzero"></path>
+    </g>
+</svg>` ],
+    [ 'item', `<?xml version="1.0" encoding="UTF-8"?>
+<svg width="24px" height="24px" viewBox="0 0 24 24" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <!-- Generator: Sketch 51.3 (57544) - http://www.bohemiancoding.com/sketch -->
+    <title>icon-emoji</title>
+    <desc>Created with Sketch.</desc>
```
hamza0867, line 53: Is this something we are using in the editor ?

modules/oxide-components/src/main/ts/components/menu/internals/Types.ts
hamza0867: Agreed, we need take care of that sooner rather than later

modules/oxide-components/src/main/ts/components/dropdown/Dropdown.tsx
zuzanna-antas: I think this is fine as is tbh. A check with a fallback to false. How would you rewrite it with Type.is? I think it adds boilerplate but the outcome is the same.

modules/oxide-components/src/main/ts/components/dropdown/Dropdown.tsx
```diff
@@ -1,72 +1,136 @@
...
+  const updateToggleState = useCallback((event: ToggleEvent) => {
+    setIsOpen(event.newState === 'open');
+  }, [ setIsOpen ]);
+
   // this can be later replaced with CSS anchor positioning
-  const updatePosition = useCallback((event: Event) => {
+  const updatePosition = useCallback((event: ToggleEvent) => {
     // TODO: remove type casting after updating TypeScript. In the newest version addEventListener correctly produces ToggleEvent
-    if ((event as ToggleEvent).newState === 'open' && triggerRef.current && contentRef.current) {
+    if (event.newState === 'open' && triggerRef.current && contentRef.current) {
```
hamza0867: I prefer using `Type.isNonNullable`​ is such cases

modules/oxide-components/src/main/ts/components/dropdown/Dropdown.tsx
zuzanna-antas: Done, I like it better than before. Thanks!

modules/oxide-components/src/main/ts/components/menu/Menu.stories.tsx
zuzanna-antas: Not sure, I copied emoji icon here 😅

modules/oxide-components/src/main/ts/components/menu/Menu.stories.tsx
zuzanna-antas: I can't replicate that :(

modules/oxide-components/src/main/ts/components/menu/Menu.tsx
zuzanna-antas: `useId`​ hook can't be used inside `useMemo`​thats why its katamari function here

modules/oxide-components/src/main/ts/components/menu/internals/Types.ts
zuzanna-antas: Yeah, it's a workaround for now

modules/oxide-components/src/main/ts/utils/Dropdowns.ts
zuzanna-antas: I will add a ticket to add a popover to Sugar then 👍

modules/oxide-components/src/main/ts/components/menu/components/SimpleMenuItem.tsx
```diff
@@ -0,0 +1,60 @@
+import { forwardRef, useEffect, useId, useMemo, useState } from 'react';
+
+import { Icon } from '../../../internal/icon/Icon.component';
```
michalnieruchalski-tiugo: I think this import is wrong. It does not resolve on my machine

modules/oxide-components/src/main/ts/components/menu/components/SubmenuItem.tsx
```diff
@@ -0,0 +1,71 @@
+import { Fun } from '@ephox/katamari';
+import { forwardRef, useEffect, useId, useMemo, useState } from 'react';
+
+import { Icon } from '../../../internal/icon/Icon.component';
```
michalnieruchalski-tiugo: same

modules/oxide-components/src/main/ts/components/menu/components/ToggleMenuItem.tsx
```diff
@@ -0,0 +1,72 @@
+import { Fun } from '@ephox/katamari';
+import { forwardRef, useEffect, useId, useMemo, useState } from 'react';
+
+import { Icon } from '../../../internal/icon/Icon.component';
```
michalnieruchalski-tiugo: same

modules/oxide-components/src/main/ts/components/menu/Menu.stories.tsx
michalnieruchalski-tiugo: Me neither. I think it's fixed now

modules/oxide-components/src/main/ts/components/dropdown/Dropdown.tsx
```diff
@@ -1,72 +1,161 @@
...
+export interface DropdownProps extends PropsWithChildren {
+  readonly side?: 'top' | 'bottom' | 'left' | 'right';
+  readonly align?: 'start' | 'center' | 'end';
+  // margin/gap between the trigger button and anchored container
+  readonly gap?: number;
+  readonly triggerEvent?: 'click' | 'hover' | 'both';
+}
+
+const Root: FC<DropdownProps> = ({ children, side = 'top', align = 'start', gap = 8, triggerEvent = 'click' }) => {
+  const triggerRef = useRef<HTMLElement | undefined>();
```
michalnieruchalski-tiugo, line 137: Why these two useRefs are typed in a different way? 

  
In one empty value is represented by null, in the other one it's represented by undefined

modules/oxide-components/src/main/ts/components/dropdown/Dropdown.tsx
```diff
@@ -1,72 +1,161 @@
...
+  const contentRef = useRef<HTMLDivElement>(null);
+  const [ isOpen, setIsOpen ] = useState(false);
+
+  // debounced hide popover function on mouse leave (used when triggersOnHover is enabled)
```
michalnieruchalski-tiugo: `triggersOnHover` no longer exists. We can either remove this comment or update it. What do you think?

modules/oxide-components/src/main/ts/components/dropdown/Dropdown.tsx
```diff
@@ -1,72 +1,161 @@
...
 
 import { DropdownContext, useDropdown } from './internals/Context';
 import * as PositioningUtils from './internals/PositioningUtils';
 
-export interface DropdownProps extends PropsWithChildren {
-  readonly side?: 'top' | 'bottom' | 'left' | 'right';
-  readonly align?: 'start' | 'center' | 'end';
-}
+const isInDropdownContent = (contentRef: React.RefObject<HTMLDivElement>, node: Node): boolean => {
+  return contentRef.current?.contains(node) ?? false;
```
michalnieruchalski-tiugo, line 10: We could use some Sugar method (I think it would be `findIn`) . It's up to you

modules/oxide-components/src/main/ts/components/dropdown/Dropdown.tsx
```diff
@@ -1,72 +1,161 @@
...
 
-const Root: FC<DropdownProps> = ({ children, side = 'top', align = 'start' }) => {
-  const popoverId = useId();
-  const triggerRef = useRef<HTMLButtonElement>(null);
+export interface DropdownProps extends PropsWithChildren {
+  readonly side?: 'top' | 'bottom' | 'left' | 'right';
+  readonly align?: 'start' | 'center' | 'end';
+  // margin/gap between the trigger button and anchored container
+  readonly gap?: number;
+  readonly triggerEvent?: 'click' | 'hover' | 'both';
```
michalnieruchalski-tiugo: We have agreed to move away from boolean, to possibly support more events in the future. Value `both` suggests there are only two events. So if anything, this could be called `all` - but I don't like it either.

I would suggest:

```suggestion
  readonly triggerEvent?: ('click' | 'hover') | Array<'click' | 'hover>;
```

So you can set that to click, hover or click or hover by using an array.  


```
<Dropdown triggerEvent='click'
<Dropdown triggerEvent='hover'
<Dropdown triggerEvent={['click', 'hover']}
```

modules/oxide-components/src/main/ts/components/dropdown/Dropdown.tsx
```diff
@@ -1,72 +1,161 @@
...
-  // @ts-expect-error - TODO: Remove this expect error once we've upgraded to React 19+
-  return <Button popovertarget={popoverId} popovertargetaction={'toggle'} ref={triggerRef} {...args}>{children}</Button>;
+const Trigger: FC<PropsWithChildren> = ({ children }) => {
+  const { triggerRef, contentRef, triggerEvent, debouncedHideHoverablePopover, isOpen } = useDropdown();
+
+  let child = Children.toArray(children)[0];
+  if (!isValidElement(child)) {
+    return null;
+  }
+  child = child as ReactElement;
```
michalnieruchalski-tiugo, line 89: Do you need this cast?

I think because `isValidElement`is a type guard, child will be `ReactElement`anyway

modules/oxide-components/src/main/ts/components/dropdown/Dropdown.tsx
```diff
@@ -1,72 +1,161 @@
...
+    };
+
+    element.addEventListener('toggle', onToggle);
 
     return () => {
-      element.removeEventListener('toggle', updatePosition);
+      element.removeEventListener('toggle', onToggle);
     };
-  }, [ contentRef, updatePosition ]);
+  }, [ contentRef, triggerRef, updatePosition, updateToggleState ]);
```
michalnieruchalski-tiugo: Do you need `updatePosition` as a dependency?

modules/oxide-components/src/main/ts/components/menu/MenuRenderer.tsx
```diff
@@ -0,0 +1,62 @@
...
+        }
+        if (itemProps.type === 'menuitem') {
+          return (<SimpleMenuItem
+            iconResolver={iconResolver}
+            key={itemProps.id}
+            {...itemProps}
+          />);
+        }
+        if (itemProps.type === 'submenu') {
+          return (<SubmenuItem
```
michalnieruchalski-tiugo: SubmenuItem does not have `items` property exposed and should not receive such input. Is it a problem now? Of course not, but It can lead to some problems later on. So Just to be safe I would deconstruct itemProps  
  


```
const { id, items, ...props } = itemProps;
<Submenu {...props}
  {render({items ...
```

modules/oxide-components/src/main/ts/main.ts
```diff
@@ -32,6 +37,11 @@ export {
   KeyboardNavigationTypes,
   Spinner,
   Tag,
+  SimpleMenuItem,
```
michalnieruchalski-tiugo: It's hard to see that these components are related.  
I think we should follow the pattern from `Dropdown` component.  
So it would be

```
Menu.Root
Menu.Item
Menu.ToggleItem
Menu.Sub
```

This is what Radix is doing and I like their approach, it looks super clean 

https://www.radix-ui.com/themes/docs/components/dropdown-menu

modules/oxide-components/src/main/ts/components/menu/components/SimpleMenuItem.tsx
```diff
@@ -0,0 +1,60 @@
...
+import * as Bem from '../../../utils/Bem';
+import type { CommonMenuItemInstanceApi, SimpleMenuItemProps } from '../internals/Types';
+
+export const SimpleMenuItem = forwardRef<HTMLButtonElement, SimpleMenuItemProps>(({ autoFocus = false, enabled = true, onSetup, text, icon, iconResolver, shortcut, onAction }, ref) => {
```
michalnieruchalski-tiugo: If it's simple/plain/regular then I would prefer to call it `MenuItem`, rather than `SimpleMenuItem`

modules/oxide-components/src/main/ts/components/menu/components/SimpleMenuItem.tsx
```diff
@@ -0,0 +1,60 @@
...
+  useEffect(() => {
+    if (onSetup) {
+      const teardown = onSetup(api);
+      return () => teardown(api);
+    }
+  }, [ onSetup, api ]);
+
+  return (
+    <button
+      id={useId()}
```
michalnieruchalski-tiugo: useId is a hook and it can't be used in loops or if statements. It's not if statement nor loop but it could be as it's easy to overlook in the jsx.

I would advise us to use hooks at the top level of the component, in my opinion it's easier to keep track of that.

modules/oxide-components/src/main/ts/components/menu/components/Item.tsx
```diff
@@ -0,0 +1,60 @@
...
+      onClick={() => onAction(api)}
+      className={Bem.element('tox-collection', 'item', {
+        'active': state.focused || state.hovered,
+        'state-disabled': !state.enabled,
+      })}
+      ref={ref}
+      autoFocus={autoFocus}
+      aria-keyshortcuts={shortcut}
+    >
+      <div className={Bem.element('tox-collection', 'item-icon')}>
```
michalnieruchalski-tiugo: We agreed to use `jsx` format to define our menus, however I still don't think we're flexible enough.

This is how this component can be used:

```
<SimpleMenuItem text='Option one' />
```

In my opinion it should be:

```
<SimpleMenuItem>Option one</SimpleMenuItem>
```

It's more flexible, and you can do more with that, the content is not limited to text only. 

As for handling icons we should use props for that. So instead:  


```
<SimpleMenuItem text='Option one' icon="arrow-left"'>
```

It should be  


```
<SimpleMenuItem
  icon={<Icon type="arrow-left" />}
>Option one</SimpleMenuItem>
```



IMO: Content should be flexible. The designer should be able to design multiline options, or option that has more icons than just one.   
  
If some pattern is popular, like icons/shortcuts we can use props. If it becomes too messy, maybe we should define separate component, like `IconMenuItem`

modules/oxide-components/src/main/ts/components/menu/components/SubmenuItem.tsx
```diff
@@ -0,0 +1,71 @@
...
+  useEffect(() => {
+    if (onSetup) {
+      const teardown = onSetup(api);
+      return () => teardown(api);
+    }
+    return Fun.noop;
+  }, [ onSetup, api ]);
+
+  return (
+    <Dropdown.Root side={submenusSide} align={'start'} triggerEvent={'both'}>
```
michalnieruchalski-tiugo: Same as with SimpleMenuItem I think this component is not composable enough with icon, shortcut, text

modules/oxide-components/src/main/ts/components/menu/components/ToggleItem.tsx
```diff
@@ -0,0 +1,72 @@
...
+      className={Bem.element('tox-collection', 'item', {
+        'enabled': state.active,
+        'active': state.focused || state.hovered,
+        'state-disabled': !state.enabled,
+      })}
+      ref={ref}
+      autoFocus={autoFocus}
+      aria-keyshortcuts={shortcut}
+    >
+      <div className={Bem.element('tox-collection', 'item-icon')}>
```
michalnieruchalski-tiugo: Same as with SimpleMenuItem I think this component is not composable enough with icon, shortcut, text

modules/oxide-components/src/main/ts/components/dropdown/Dropdown.tsx
zuzanna-antas: Unfortunately, without it the child is ReactElement or ReactPortal. According to the `isValidElement`[ docs ](https://18.react.dev/reference/react/isValidElement)the guard should work, but there must be something wrong with the types on React side. Still worth doing the check though. 

Docs:

> **Only** [**JSX tags**](https://18.react.dev/learn/writing-markup-with-jsx) **and objects returned by** [**`createElement`**](https://18.react.dev/reference/react/createElement) **are considered to be React elements.** For example, even though a number like `42` is a valid React _node_ (and can be returned from a component), it is not a valid React element. **Arrays and portals created with** [`createPortal`](https://18.react.dev/reference/react-dom/createPortal) **are also** **_not_** **considered to be React elements.**

modules/oxide-components/src/main/ts/components/dropdown/Dropdown.tsx
zuzanna-antas: It is because the trigger has to be mutable ref. So the types of this ref objects are different. We have:

- `triggerRef: React.MutableRefObject<HTMLElement | undefined>;`
- `contentRef: React.RefObject<HTMLDivElement>;`

We use mutation in the cloneElement function. Without this we would not be able to change the `triggerRef.current` value [in line 116](https://app.graphite.com/github/pr/tinymce/tinymce/10756/TINY-13236-create-menu-component-and-add-hoverable-behavior-to-dropdown#lines-modules/oxide-components/src/main/ts/components/dropdown/Dropdown.tsx-R116-R116). React is more strongly typed than preact I guess. You need to explicitly make the ref object mutable.  If you know how to solve it without mutation let me know, but I think it's necessary evil in this case.

modules/oxide-components/src/main/ts/components/menu/components/SimpleMenuItem.tsx
zuzanna-antas: Yeah, the path changed here: https://app.graphite.com/github/pr/tinymce/tinymce/10762. Will fix that :+1:

modules/oxide-components/src/main/ts/components/dropdown/Dropdown.tsx
zuzanna-antas: Which method do you mean? I can't find `findIn` in Sugar

modules/oxide-components/src/main/ts/components/dropdown/Dropdown.tsx
zuzanna-antas: Good catch, forgot to remove it :+1:

modules/oxide-components/src/main/ts/components/dropdown/Dropdown.tsx
michalnieruchalski-tiugo: But what would happen if you would only remove the line with type cast, will you get any typescript errors?

modules/oxide-components/src/main/ts/components/dropdown/Dropdown.tsx
michalnieruchalski-tiugo: Sorry, my bad, it's not findIn

```
Has.descendant()
```

spocke: That would be way slower since it traverses manually. Compare.contains is the sugar equivalent.

modules/oxide-components/src/main/ts/components/dropdown/Dropdown.tsx
zuzanna-antas: Yes, there is a ts error then

modules/oxide-components/src/main/ts/components/dropdown/Dropdown.tsx
zuzanna-antas: Sugar `Compare.contains` does exactly the same under the hood (`Node.contains()`). The only difference is it returns false when the elements are the same. With `Node.contains()` function the node contains itself also so it returns true in the same case - this is what we need here.

modules/oxide-components/src/main/ts/components/menu/components/Item.tsx
zuzanna-antas: Done, you can take a look :smile:

As discussed: 

- I went with using `children` instead of `text` property  - that way we can add additional formatting to the item label
- `icon` property now accepts string or `JSX.Element`(an `Icon` component)
- In `Submenu` I wanted to have the same pattern so I replaced existing use of `children` to `submenuContent` prop which accept a `JSX.Element`. The `children` are now placed as a label same as in `Item` and `ToggleItem`. To `submenuContent` you can pass another Menu or for example ColorPicker, Table widget - we don't have these components yet, but we do have such cases in tinymce menus.

modules/oxide-components/src/main/ts/components/dropdown/Dropdown.tsx
michalnieruchalski-tiugo: I'm fine with that

modules/oxide-components/src/main/ts/components/menu/components/Item.tsx
michalnieruchalski-tiugo: Perfect, I'm happy with the API we're having now. Good job

modules/oxide-components/src/main/ts/components/menu/components/Item.tsx
```diff
@@ -0,0 +1,73 @@
...
+        {itemIcon}
```
michalnieruchalski-tiugo: I don't think we need to use state here. We could simply:

```
const itemIcon = Type.isString(icon)
  ? <Item icon={icon} resolver={iconResolver} />
  : icon;

<div className={Bem.element('tox-collection', 'item-icon')}>
  {itemIcon}
</div>
```

michalnieruchalski-tiugo: Also I think we should not render \<div class="item-icon" when `itemIcon` is null ( same as for shortcut )

PR #10759, opened 2025-11-28
TINY-13362: Add a slight top-border to the `tinyai__bottom-bar`

kimwoodfield:
Related Ticket: [TINY-13362](https://ephocks.atlassian.net/browse/TINY-13362)

Description of Changes:
* Add a `border-top` to the bottom bar in the TinyMCE AI plugin so that there is separation between the iframe content and that bottom bar to match the design.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

[TINY-13362]: https://ephocks.atlassian.net/browse/TINY-13362?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/oxide/src/less/theme/content/ai-preview/ai-preview.less
+++ b/modules/oxide/src/less/theme/content/ai-preview/ai-preview.less
@@ -10,7 +10,6 @@
 @tinymceai-selected-box-shadow: 0 -2px 0 0 @tinymceai-selected-outline-color inset, 0 -2px 0 0 @tinymceai-selected-outline-color;
 
 @tinymceai-dim-overlay-color: #222f3e80;
-@tinymceai-bottom-bar-border-color: #E3E3E3;
 
 .diff-content(
   tinymceai,
@@ -54,5 +53,5 @@ div[tinymceai-data-pending-diff="true"], span[tinymceai-data-pending-diff="true"
   padding: 12px @pad-md;
   background: @color-white;
   z-index: 2;
-  border-top: 1px solid @tinymceai-bottom-bar-border-color;
+  border-top: 1px solid hsl(from var(--tox-color-white, @color-white) h s calc(l - 11));
 }


modules/oxide/src/less/theme/content/ai-preview/ai-preview.less
```diff
@@ -10,6 +10,7 @@
 @tinymceai-selected-box-shadow: 0 -2px 0 0 @tinymceai-selected-outline-color inset, 0 -2px 0 0 @tinymceai-selected-outline-color;
 
 @tinymceai-dim-overlay-color: #222f3e80;
+@tinyai-bottom-bar-border-color: #E3E3E3;
```
ArvinJ-H: I believe this name needs update?

modules/oxide/src/less/theme/content/ai-preview/ai-preview.less
```diff
@@ -10,6 +10,7 @@
 @tinymceai-selected-box-shadow: 0 -2px 0 0 @tinymceai-selected-outline-color inset, 0 -2px 0 0 @tinymceai-selected-outline-color;
 
 @tinymceai-dim-overlay-color: #222f3e80;
+@tinymceai-bottom-bar-border-color: #E3E3E3;
```
zuzanna-antas: We should be careful with introducing new LESS variables as they become part of the skinning API. That means removing them later is a breaking change. In this case I think we can apply the color directly in the styles. 
I would also reconsider the rest of variables introduced in this file - do we want them as part of the sinning API? We can take another stab before releasing (that's out of scope for this PR for sure)

Additionally, we can make it adjust to skin changes. Instead of the fixed value I would do a color calculation to achieve that. 
- In LESS that would be: `darken(@background-color, 11%)`. 
- And in CSS: `hsl(from var(--tox-private-background-color, @background-color) h s calc(l - 11))`.

We should try to use CSS functions instead of less as it will be easier to transition fully to CSS later.

kimwoodfield: Noted, thanks for the information. I've gone with the CSS approach, using a combination of `hsl(from ... )` and `calc` as well as created a ticket for us to swing back for a review of the existing variables in this file before the release.

PR #10762, opened 2025-11-30
TINY-13316: Export Icon component from oxide-components

kimwoodfield:
Related Ticket: [TINY-13316](https://ephocks.atlassian.net/browse/TINY-13316)

Description of Changes:
* There are use cases where we might want to use the Icon component on its own rather than having to use it with the IconButton component. This PR exports it from oxide-components so it can be used freely.
* Also added accompanying Story (with documentation) and a simple test. Also added `readonly` to its props.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

[TINY-13316]: https://ephocks.atlassian.net/browse/TINY-13316?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/oxide-components/vitest.config.ts
+++ b/modules/oxide-components/vitest.config.ts
@@ -24,7 +24,7 @@ export default defineConfig({
       },
       {
         optimizeDeps: {
-          include: [ 'react/jsx-dev-runtime', 'react/jsx-runtime' ],
+          include: [ 'react/jsx-dev-runtime' ],
         },
         test: {
           name: 'browser',


modules/oxide-components/src/test/ts/browser/components/Icon.spec.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { Icon } from 'oxide-components/components/icon/Icon';
+import { Fragment } from 'react/jsx-runtime';
+import { describe, expect, it, vi } from 'vitest';
+import { render } from 'vitest-browser-react';
+
+const chevronDownTestIconId = 'chevron-down-icon';
+const chevronUpTestIconId = 'chevron-up-icon';
+const customTestIconId = 'custom-icon-icon';
+
+describe('browser.Icon', () => {
```
spocke: browser.components.IconTest maybe to follow the other convention?

modules/oxide-components/src/main/ts/components/icon/Icon.stories.tsx
```diff
@@ -0,0 +1,81 @@
...
+};
+
+<Icon icon="chevron-down" resolver={iconResolver} dataTestId="chevron-down-icon" />
+\`\`\`
+
+## Props
+
+- \`icon\` (string): The name of the icon to render.
+- \`resolver\` (function): A function that takes an icon name and returns the corresponding SVG string.
+- \`dataTestId\` (string, optional): An optional \`data-testid\` attribute for testing purposes.
```
hamza0867: This is a little bit dangerous as this documentation now requires manual updating whenver we change the api.  Doesn't this conflict with the autodocs storybook plugin ?

modules/oxide-components/src/main/ts/components/icon/IconTypes.tsx
```diff
@@ -0,0 +1,14 @@
...
+  /** The function to resolve the icon name to an html string.
+   *
+   * This would eventually default to retrieving the icon from the editor's registry.
+   * (name: string) => editor.ui.registry.getAll().icons[name] ?? 'temporary-placeholder'
+   *
+   * Returns the html string representation of the icon.
+   */
+  readonly resolver: (icon: string) => string;
+  /** Optional data-testid attribute for testing purposes */
+  readonly dataTestId?: string;
```
hamza0867: I don't think we should add dataTestId to all of our components. We can instead spread the rest of the props to the span. What do you think @spocke @TheSpyder ?

kimwoodfield: Yeah that's fine, I'm happy to go with that too if the others agree

modules/oxide-components/src/main/ts/components/icon/Icon.stories.tsx
kimwoodfield: I didn't know about the autodocs plugin and that it could automatically generate the props table for us. Thanks for letting me know, I'll remove this section in that case.

modules/oxide-components/src/test/ts/browser/components/Icon.spec.tsx
kimwoodfield: It doesn't look like the other component tests follow that convention. The current style is: `browser.icon.Icon`. I made another commit to update them all to include `browser.component.<component_name>`

modules/oxide-components/src/main/ts/components/icon/IconTypes.tsx
TheSpyder: Having a unique way to identify UI elements is very helpful

https://github.com/tinymce/tinymce/blob/2b65edeeeda247c329ef785d2d48553c1932ef55/modules/tinymce/src/themes/silver/main/ts/ui/toolbar/button/ToolbarButtons.ts#L135

But I'm not sure the icon needs that, only the components directly, and it tends to be based on the name not a dedicated property.

So I agree this isn't necessary.

kimwoodfield: Done!

modules/oxide-components/src/test/ts/browser/components/Icon.spec.tsx
spocke: I think it's just to make sure we make unique names for them for junit reports. But for other tests in other parts of the code base we kind of follow that convention kind of Java ish name spacing that maps to the directory structure.

PR #10782, opened 2025-12-10
TINY-13420: add error handler for unallowed file upload

lorenzo-pomili:
Related Ticket: TINY-13420

Description of Changes:
- add `errorHandler` to the dropzone
- manage dropzone error for `link` plugin
- manage dropzone error for `image` plugin

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

## Summary by CodeRabbit

* **New Features**
  * Added a configurable error handler for file drop/upload components to allow custom handling of invalid uploads.

* **Improvements**
  * Shows explicit feedback when dropped files are unsupported in image and link dialogs.
  * Upload-related errors are now surfaced via the dialog alert mechanism for clearer user notifications.

### DIFF
--- a/modules/tinymce/src/themes/silver/main/ts/ui/dialog/Dropzone.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/ui/dialog/Dropzone.ts
@@ -65,7 +65,7 @@ export const renderDropZone = (spec: DropZoneSpec, providersBackstage: UiFactory
       Representing.setValue(component, filteredFiles);
       AlloyTriggers.emitWith(component, formChangeEvent, { name: spec.name });
       if (filteredFiles.length === 0) {
-        spec.onInvalidFiles();
+        spec.onInvalidFiles(component.element);
       }
     }
   };
--- a/modules/bridge/src/main/ts/ephox/bridge/components/dialog/Dropzone.ts
+++ b/modules/bridge/src/main/ts/ephox/bridge/components/dialog/Dropzone.ts
@@ -1,5 +1,6 @@
 import { FieldSchema, StructureSchema, ValueType } from '@ephox/boulder';
 import { Fun, type Optional, type Result } from '@ephox/katamari';
+import type { SugarElement } from '@ephox/sugar';
 
 import { type FormComponentWithLabel, type FormComponentWithLabelSpec, formComponentWithLabelFields } from './FormComponent';
 
@@ -10,7 +11,7 @@ export interface DropZoneSpec extends FormComponentWithLabelSpec {
   buttonLabel?: string;
   allowedFileTypes?: string;
   allowedFileExtensions?: string[];
-  onInvalidFiles?: () => void;
+  onInvalidFiles?: (el: SugarElement<HTMLElement>) => void;
 }
 
 export interface DropZone extends FormComponentWithLabel {
@@ -20,7 +21,7 @@ export interface DropZone extends FormComponentWithLabel {
   buttonLabel: Optional<string>;
   allowedFileTypes: Optional<string>;
   allowedFileExtensions: Optional<string[]>;
-  onInvalidFiles: () => void;
+  onInvalidFiles: (el: SugarElement<HTMLElement>) => void;
 }
 
 const dropZoneFields = formComponentWithLabelFields.concat([
--- a/modules/tinymce/src/plugins/link/main/ts/ui/Dialog.ts
+++ b/modules/tinymce/src/plugins/link/main/ts/ui/Dialog.ts
@@ -1,4 +1,5 @@
 import { Arr, Fun, Optional, Optionals } from '@ephox/katamari';
+import { Focus, type SugarElement } from '@ephox/sugar';
 
 import type Editor from 'tinymce/core/api/Editor';
 import type { BlobInfo } from 'tinymce/core/api/file/BlobCache';
@@ -162,7 +163,7 @@ const makeDialogBody = (
   catalogs: LinkDialogCatalog,
   hasUploadPanel: boolean,
   fileTypes: DocumentsFileTypes[],
-  onInvalidFiles: () => void
+  onInvalidFiles: (el: SugarElement<HTMLElement>) => void
 ): Dialog.PanelSpec | Dialog.TabPanelSpec => {
 
   const generalPanelItems = Arr.flatten<Dialog.BodyComponentSpec>([
@@ -247,7 +248,7 @@ const makeDialog = (settings: LinkDialogInfo, onSubmit: (api: Dialog.DialogInsta
     catalogs,
     settings.hasUploadPanel,
     Options.getDocumentsFileTypes(editor),
-    () => helpers.alertErr('All inserted files have unallowed extensions', Fun.noop)
+    (el) => helpers.alertErr('All inserted files have unallowed extensions', () => Focus.focus(el))
   );
   return {
     title: 'Insert/Edit Link',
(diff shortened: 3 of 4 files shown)


.changes/unreleased/tinymce-TINY-13420-2025-12-10-3.yaml
TheSpyder, line 1: I think it's fine for image and link to be in the same changelog entry

- [x] remove

.changes/unreleased/tinymce-TINY-13420-2025-12-10-2.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Improved
+body: '`link` dialog gives feedback for unsupported files error.'
```
TheSpyder:
```suggestion
body: 'The file upload feature of `link` and `image` dialogs now provide feedback when an unsupported file type is selected.'
```

.changes/unreleased/tinymce-TINY-13420-2025-12-10.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Added
+body: 'New option for `dropzone` dialog components: errorHandler.'
```
TheSpyder:
```suggestion
body: 'New `errorHandler` option for `dropzone` dialog components.'
```

modules/tinymce/src/plugins/image/main/ts/ui/Dialog.ts
```diff
@@ -279,14 +279,14 @@ const closeHandler = (state: ImageDialogState) => (): void => {
   state.open = false;
 };
 
-const makeDialogBody = (info: ImageDialogInfo): DialogType.TabPanelSpec | DialogType.PanelSpec => {
+const makeDialogBody = (info: ImageDialogInfo, errorHandler: (error: string) => void): DialogType.TabPanelSpec | DialogType.PanelSpec => {
```
TheSpyder: The reason we have the `info` object is to avoid adding parameters every time. Link doesn't do this because it's smaller, but you can already see the `makeDialogBody` function in that plugin starting to become unweildy.

- [x] move error handler into ImageDialogInfo

modules/tinymce/src/themes/silver/main/ts/ui/dialog/Dropzone.ts
```diff
@@ -61,8 +61,15 @@ export const renderDropZone = (spec: DropZoneSpec, providersBackstage: UiFactory
...
-      Representing.setValue(component, filterByExtension(files, providersBackstage, spec.allowedFileExtensions));
-      AlloyTriggers.emitWith(component, formChangeEvent, { name: spec.name });
+      const filteredFiles = filterByExtension(files, providersBackstage, spec.allowedFileExtensions);
+      if (filteredFiles.length === 0) {
+        spec.errorHandler?.each((errorHandler) => {
+          errorHandler('All inserted files have unallowed extensions');
+        });
+      } else {
+        Representing.setValue(component, filteredFiles);
+        AlloyTriggers.emitWith(component, formChangeEvent, { name: spec.name });
```
TheSpyder: You need to call these lines even if it's zero, otherwise the dialog might save with a stale value

- [x] set every time

.changes/unreleased/tinymce-TINY-13420-2025-12-10-3.yaml
lorenzo-pomili: done

.changes/unreleased/tinymce-TINY-13420-2025-12-10-2.yaml
lorenzo-pomili: changed

.changes/unreleased/tinymce-TINY-13420-2025-12-10.yaml
lorenzo-pomili: changed

modules/tinymce/src/themes/silver/main/ts/ui/dialog/Dropzone.ts
lorenzo-pomili: fixed

modules/tinymce/src/plugins/image/main/ts/ui/Dialog.ts
lorenzo-pomili: we generate the `info` using the function `collect`, that function seems to generate all the `info` from the `editor` to add the error handler I can pass it as a second parameter of reimplement it inside `collect`.

since add it as a param seems similar to the current situation and I didn't want to duplicate that function I moved inside the `collect` and removed from the `helpers` let me know if it is ok or if I should manage it in a different way

EDIT: commit https://github.com/tinymce/tinymce/pull/10782/commits/bbe47d2afcf96715ffb2377011681666a05df81a

modules/tinymce/src/themes/silver/main/ts/ui/dialog/Dropzone.ts
```diff
@@ -61,8 +61,14 @@ export const renderDropZone = (spec: DropZoneSpec, providersBackstage: UiFactory
 
   const handleFiles = (component: AlloyComponent, files: FileList | null | undefined) => {
     if (files) {
-      Representing.setValue(component, filterByExtension(files, providersBackstage, spec.allowedFileExtensions));
+      const filteredFiles = filterByExtension(files, providersBackstage, spec.allowedFileExtensions);
+      Representing.setValue(component, filteredFiles);
       AlloyTriggers.emitWith(component, formChangeEvent, { name: spec.name });
+      if (filteredFiles.length === 0) {
+        spec.errorHandler?.each((errorHandler) => {
+          errorHandler('All inserted files have unallowed extensions');
```
spocke: I don't think we should hard code a string like this. Instead I think the component should have a `onInvalidFiles` that passes in an array of invalid files then it's up to the link and image plugin to emit the error. Also this only accepts drops and picks of a single file, not sure why we added it as an array maybe we thought it would accept multiple ones in the future but right now we just act on the first file.

The `on<something>` is also what we use as a pattern for other things like `onAction` 

Another alternative would be to have a invalidFilesMessage property then display the message in the drop zone as some red text instead of a alert message. But that needs to be checked with design etc.

spocke (approved):

modules/tinymce/src/themes/silver/main/ts/ui/dialog/Dropzone.ts
lorenzo-pomili: I refactored it ( https://github.com/tinymce/tinymce/pull/10782/commits/3770a62c8a1dca7940d94c95ca42efbbcacb85af ), let me know if it is what you meant, about the interface of the error I used the alert since is the same thing that we do when there is an error from the server, we could maybe change both but I'm not sure having different behaviour is worth

modules/tinymce/src/plugins/image/main/ts/ui/Dialog.ts
TheSpyder: That's a good idea, since both helpers and info are based on the editor instance

It's a little bit of a weird place to have a helper function in an `info` object, but the alternative is to pass `helpers` into `UploadTab.makeTab` which I'm not sure is the correct answer either

lorenzo-pomili:
closed in favour of the PR on release, since this should go into a patch: https://github.com/tinymce/tinymce/pull/10796

TheSpyder (approved):

PR #10802, opened 2025-12-17
TINY-13375: Add styles and api needed for deleting conversation

hamza0867:
Related ticket: TINY-13375

# Enhanced Error Message Component with Removable Option

## Description of Changes:

- Enhanced `ErrorMessage` component to support a removable option with an "X" button
- Added styling for the new error message layout with flexbox for proper alignment
- Updated CSS to improve hover and focus states for chat history items
- Added cursor pointer to chat history items for better UX
- Implemented new `parentElements` utility function in `Traverse.ts` that returns only HTML elements (not all nodes)

## Pre-checks:

- [x] Changelog entry added

## Review:

- [x] Milestone set

### DIFF
--- a/modules/agar/src/main/ts/ephox/agar/http/ResponseHelpers.ts
+++ b/modules/agar/src/main/ts/ephox/agar/http/ResponseHelpers.ts
@@ -1,5 +1,7 @@
 export const makeResponse = (body: BodyInit | null, init: ResponseInit): Response => new window.Response(body, init);
 
+export const emptyResponse = (init: ResponseInit = {}): Response => makeResponse(null, init);
+
 export const jsonResponse = (data: any, init: ResponseInit = {}): Response =>
   makeResponse(JSON.stringify(data), {
     ...init,
--- a/modules/oxide-components/src/main/ts/bespoke/tinymceai/error/ErrorMessage.tsx
+++ b/modules/oxide-components/src/main/ts/bespoke/tinymceai/error/ErrorMessage.tsx
@@ -1,13 +1,35 @@
+import { Bem, IconButton } from 'oxide-components/main';
 import React from 'react';
 
-import { classes } from '../../../utils/Styles';
-
-interface ErrorMessageProps {
-  message: string;
+interface BaseProps {
+  readonly message: string;
 }
 
-const ErrorMessage: React.FC<ErrorMessageProps> = ({ message }) => {
-  return <div className={classes([ 'tox-ai__error-message' ])} role="alert" aria-live="polite">{message}</div>;
+interface NonRemoveableErrorMessageProps extends BaseProps {
+  readonly removable?: false;
+}
+
+interface RemoveableErrorMessageProps extends BaseProps {
+  readonly removable: true;
+  readonly onRemove: () => void;
+}
+
+export type ErrorMessageProps = NonRemoveableErrorMessageProps | RemoveableErrorMessageProps;
+
+const ErrorMessage: React.FC<ErrorMessageProps> = (props) => {
+  const { message, removable } = props;
+  return (
+    <div className={Bem.block('tox-ai-error')} role="alert" aria-live="polite">
+      <div className={Bem.element('tox-ai-error', 'message')}>
+        {message}
+      </div>
+      {removable && (
+        <div className={Bem.element('tox-ai-error', 'icon')}>
+          <IconButton variant='naked' icon='close' onClick={props.onRemove} />
+        </div>
+      )}
+    </div>
+  );
 };
 
 export default ErrorMessage;
--- a/modules/agar/src/main/ts/ephox/agar/api/Http.ts
+++ b/modules/agar/src/main/ts/ephox/agar/api/Http.ts
@@ -2,5 +2,5 @@ export type { HttpHandler } from '../http/HttpHandler';
 export { get, post, put, del, patch } from '../http/HttpHandler';
 export type { MockHttpHook, MockHttpHookConfig } from '../http/MockHttpHook';
 export { mockHttpHook } from '../http/MockHttpHook';
-export { makeResponse, jsonResponse, textResponse, blobResponse, chunkedResponse } from '../http/ResponseHelpers';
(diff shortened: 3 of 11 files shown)


hamza0867:
* **#10802** <a href="https://app.graphite.com/github/pr/tinymce/tinymce/10802?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/tinymce/tinymce/10802?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `epic/TINY-12378`


<h2></h2>

modules/agar/src/main/ts/ephox/agar/http/ResponseHelpers.ts
```diff
@@ -1,5 +1,7 @@
 export const makeResponse = (body: BodyInit | null, init: ResponseInit): Response => new window.Response(body, init);
 
+export const emptyResponse = (init: ResponseInit = {}): Response => makeResponse(null, init);
```
kimwoodfield, line 3: Why do we need this?

modules/oxide-components/src/main/ts/bespoke/tinymceai/error/ErrorMessage.tsx
```diff
@@ -1,13 +1,35 @@
+import { Bem, IconButton } from 'oxide-components/main';
 import React from 'react';
 
-import { classes } from '../../../utils/Styles';
+interface BaseProps {
+  readonly message: string;
+}
+
+interface NonRemoveableErrorMessageProps extends BaseProps {
```
kimwoodfield, line 8: Why do we need both types? Couldn't `removable` just be an optional prop on `ErrorMessageProps`? If we want it to default to being removable I think we could just default to `const { message, removable = true } = props;` but you might have to double check if that's valid

modules/sugar/src/main/ts/ephox/sugar/api/search/Traverse.ts
```diff
@@ -67,6 +67,27 @@ const parents = (element: SugarElement<Node>, isRoot?: (e: SugarElement<Node>) =
   return ret;
 };
 
+const parentElements = (element: SugarElement<HTMLElement>, isRoot?: (e: SugarElement<Node>) => boolean): SugarElement<HTMLElement>[] => {
```
kimwoodfield: Did we not already have an API for doing this? 🤔

modules/sugar/src/main/ts/ephox/sugar/api/search/Traverse.ts
```diff
@@ -67,6 +67,27 @@ const parents = (element: SugarElement<Node>, isRoot?: (e: SugarElement<Node>) =
...
+  const stop = Type.isFunction(isRoot) ? isRoot : Fun.never;
+
+  // This is used a *lot* so it needs to be performant, not recursive
+  let dom: Node = element.dom;
+  const ret: SugarElement<HTMLElement>[] = [];
+
+  while (dom.parentElement !== null && dom.parentElement !== undefined) {
```
kimwoodfield: Why not use Katamari's Type module for this? We have both `isNull` and `isUndefined` that you could use with logical NOT if you don't want to use `isNonNullable`

kimwoodfield (approved):

modules/sugar/src/main/ts/ephox/sugar/api/search/Traverse.ts
hamza0867: We have one that returns an array of `Node`s NOT `HTMLElement`s. It's right above this one :smile:

modules/sugar/src/main/ts/ephox/sugar/api/search/Traverse.ts
hamza0867: Because I just copied the code from the above version. I want to keep them consistent.

modules/agar/src/main/ts/ephox/agar/http/ResponseHelpers.ts
hamza0867: Because the delete api returns a `204` response that has no content. I need this for mocking the delete api for tests

modules/oxide-components/src/main/ts/bespoke/tinymceai/error/ErrorMessage.tsx
hamza0867: Good queestion. Because we need to tie the `removable` with `onRemove`. If we keep everything in the same type but optional then we can theoretically end up in a situation like this:

```
{
message: 'some string',
removeable: true
}
```

And that would be accepted by the compiler. So I used different types so that the case above would be caught by the compiler as an error. Does that make sense ?

kimwoodfield: Yep, makes sense - thanks!

modules/sugar/src/main/ts/ephox/sugar/api/search/Traverse.ts
spocke: This is more like parentHTMLElements since the return type is not Element[] however this is consistent with the parentElement that also return a HTMLElement.

Also you can always make this by just doing `Arr.filter(Traverse.parents(..), SugarNode.isHTMLElement)` you could instead of duplicating the code do that it's a small perf diff but you normally don't have like 1000 depth anyway.

HTMLElement and just Element can be nested inside each other SVGs can have embedded HTML for example and SVG elements is not HTMLElements.

Also why does it take a element as it's input it would be more generic if you could pass in a node.

spocke (approved):

modules/sugar/src/main/ts/ephox/sugar/api/search/Traverse.ts
spocke: Also if we add this core library level API this probably needs tests in sugar.

spocke (approved):

modules/sugar/src/main/ts/ephox/sugar/api/search/Traverse.ts
hamza0867: It turns out it wasn't needed

PR #10803, opened 2025-12-17
TINY-13276: Improvement for lists delete

HAFRMO:
Related Ticket: TINY-13276

Description of Changes:
Allowed users to prevent keydown events for the list delete functionality.
To allow the user to preventdefault the event it had to be placed in the init-event since otherwise it would be located before the user's chance to preventdefault it.
Other delete functionality had to be moved as to be placed behind the moved delete functionality to prevent reintroducing a bug.
We need some sort of method to prevent all of these order

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed list editing so Backspace/Delete keyboard events can be prevented in selection scenarios, preserving content and selection.

* **Tests**
  * Added/adjusted tests verifying deletion prevention behavior and selection preservation during list editing.

* **Chores**
  * Added an unreleased changelog entry referencing the fix.


.changes/unreleased/tinymce-TINY-13276-2025-12-17.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Improved
+body: Certain deletion-events in lists could not be properly prevented.
```
TheSpyder:
```suggestion
kind: Fixed
body: Deletion keyboard events in lists could not be prevented.
```

TheSpyder (approved):

modules/tinymce/src/plugins/lists/test/ts/webdriver/DeleteTest.ts
```diff
@@ -61,4 +61,43 @@ describe('webdriver.tinymce.plugins.lists.DeleteTest', () => {
...
+      const expectedContent = '<ol>' +
+        '<li>' +
+        'List 1' +
+        '</li>' +
+        '</ol>';
+
+      editor.setContent(initialContent);
+      editor.undoManager.add();
+      TinySelections.setSelection(editor, [ 0, 0, 0 ], 2, [ 0, 0, 0 ], 5);
+      await RealKeys.pSendKeysOn('iframe => body', [ RealKeys.backspace() ]);
```
spocke, line 98: We dont need realkeys for this. You can check if the event was prevented by the isDefaultPrevented() on the event or make sure that the operation never happened.

We should avoid having more webdriver tests that will break. If you make a non web driver test then we can stick that into some other list delete related test file.

HAFRMO: This one does check that the event didn't happen by checking that no change occurred on the content. As for not making it a webdriver, trying the fake keys seems to not work right, not sure why.

PR #10805, opened 2025-12-18
TINY-13027: Update dependencies

MitchC1999:
Related Ticket: TINY-13027, TINY-12738

Description of Changes:
* Updating dependencies
* Major fast-check update includes numerous required changes to tests
  * Removal of `asciiString` and `hexaString` apis
  * Arbitrary min and max are now set as properties on constraints object
  * Introduction of testing for prototype pollution inside string arbitraries resulted in flakes
* Avoided oxide-components updates to avoid potential conflicts with ongoing work
* Avoided updating emojilib due to large breaking changes affecting distributed files
* Type errors from enum comparisons, now using `toString()`
* Lerna update now requires nx.json config

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [x] Milestone set

## Summary by CodeRabbit

* **Chores**
  * Updated dev dependencies (notably fast-check → v4.4.0) and tooling; adjusted ignore and project config for better environment handling and caching.
  * Modernized test generators (object-based options, new ASCII-safe and hex generators) and added prototype-pollution safeguards.
  * Simplified lint config and related runtime path resolution.

* **Bug Fixes**
  * Fixed a string concatenation error in tests and strengthened some cache-related test assertions.

### DIFF
--- a/modules/snooker/src/test/ts/browser/TableSizeTest.ts
+++ b/modules/snooker/src/test/ts/browser/TableSizeTest.ts
@@ -24,7 +24,7 @@ describe('TableSizeTest', () => {
     });
 
     it('tables with widths should be detected as percent or pixel', () => {
-      fc.assert(fc.property(fc.integer({ min: 100, max: 1000 }), fc.float({ min: 1, max: 100 }), (pixel, percent) => {
+      fc.assert(fc.property(fc.integer({ min: 100, max: 1000 }), fc.float({ min: 1, max: 100, noNaN: true, noDefaultInfinity: true }), (pixel, percent) => {
         const pixelTable = SugarElement.fromHtml<HTMLTableElement>(pixelTableHtml.replace('400px', pixel + 'px'));
         const percentageTable = SugarElement.fromHtml<HTMLTableElement>(percentTableHtml.replace('80%', percent + '%'));
 


modules/katamari/src/test/ts/atomic/api/str/ContainsTest.ts
```diff
@@ -38,7 +38,7 @@ describe('atomic.katamari.api.str.ContainsTest', () => {
   it('A string added to a string (at the start) must be contained within it', () => {
     fc.assert(fc.property(
       fc.string(),
-      fc.string(1, 40),
+      fc.string({ minLength: 1, maxLength: 40 }),
       (str, contents) => {
         const r = contents = str;
```
TheSpyder: The bot pointed out this has been incorrectly guaranteed to pass, probably since it was created. I think it's unlikely to fail but we should still fix the test

```suggestion
        const r = contents + str;
```

eslint.config.ts
```diff
@@ -1,15 +1,20 @@
...
+import { fileURLToPath } from 'url';
+import { dirname } from 'path';
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = dirname(__filename);
 
 export default defineConfig(
   [
-    tinyPlugin.configs.editor,
+    ...tinyPlugin.configs.editor,
```
TheSpyder, line 12: why was the spread necessary?

modules/katamari/src/test/ts/atomic/api/arr/ObjKeysTest.ts
```diff
@@ -22,7 +22,12 @@ describe('atomic.katamari.api.arr.ObjKeysTest', () => {
   });
 
   it('only returns elements that are in the input', () => {
-    fc.assert(fc.property(fc.dictionary(fc.asciiString(), fc.integer()), (obj) => {
+    fc.assert(fc.property(fc.dictionary(
+      fc.string({ unit: 'binary-ascii' }).filter((s) => s !== '__proto__'), // Avoid testing prototype pollution
+      fc.integer(),
+      { noNullPrototype: true }
+    ),
```
TheSpyder: Should this be using `arbAsciiDict`?

modules/tinymce/src/themes/silver/main/ts/Render.ts
```diff
@@ -491,7 +491,7 @@ const setup = (editor: Editor, setupForTheme: ThemeRenderSetup): RenderInfo => {
     editor.addQueryStateHandler('ToggleToolbarDrawer', () => OuterContainer.isToolbarDrawerToggled(outerContainer));
 
     editor.on('blur', () => {
-      if (Options.getToolbarMode(editor) === Options.ToolbarMode.floating && OuterContainer.isToolbarDrawerToggled(outerContainer)) {
+      if (Options.getToolbarMode(editor) === Options.ToolbarMode.floating.toString() && OuterContainer.isToolbarDrawerToggled(outerContainer)) {
```
TheSpyder: I wonder why this was required. `getToolbarMode()` should be defined as having a return type that matches `Options.ToolbarMode.floating`.

nx.json
```diff
@@ -0,0 +1,12 @@
...
+  "namedInputs": {
+    "tiny-product": [
+      "{projectRoot}/**/*",
+      "!{projectRoot}/node_modules/**/*"
+    ],
+    "tiny-library": [
+      "{projectRoot}/**/*",
+      "!{projectRoot}/node_modules/**/*"
+    ]
+  }
```
TheSpyder, line 11: Keeping these in sync with the premium nx build could be a pain. I suppose eventually we're going to combine the two and then we can simplify the open source build away from lerna/nx.

modules/tinymce/src/plugins/autolink/test/ts/browser/AutoLinkPluginTest.ts
```diff
@@ -20,6 +20,10 @@ describe('browser.tinymce.plugins.autolink.AutoLinkPluginTest', () => {
     allow_unsafe_link_target: true
   }, [ Plugin ], true);
 
+  // https://fast-check.dev/docs/migration-guide/from-3.x-to-4.x/#hexa-or-hexastring
```
spocke, line 23: We keep having this over and over again maybe we should build it into agar.

eslint.config.ts
kimwoodfield: Because `tinyPlugin.configs.editor` is an array so without spreading we created a nested array which broke ESLint's flat config parser causing the following error: `TypeError: Key "languageOptions": Key "globals": Expected an object.`:

Screenshot 2025-12-19 at 8 46 09 am

kimwoodfield (approved):

modules/tinymce/src/themes/silver/main/ts/Render.ts
MitchC1999: This was failing https://typescript-eslint.io/rules/no-unsafe-enum-comparison/
I think `Options.ToolbarMode.floating` was evaluated to type `"floating"` or `ToolbarMode.floating`, rather than `ToolbarMode`. It doesn't seem to be failing any more after reverting this though so I'll revert it.

eslint.config.ts
TheSpyder: Ah, the old "typescript didn't actually type check this" problem. Good to know.

modules/tinymce/src/plugins/autolink/test/ts/browser/AutoLinkPluginTest.ts
TheSpyder: It's a weird choice by fast-check to no longer provide it for sure.

Although I wonder how often we _actually_ need a hex string when an arbitrary ascii string would suffice.

MitchC1999: I'll make a note to look at adding this in a future hip sprint

PR #10701, opened 2025-10-31
TINY-13045: Include scrollbars in draggable boundaries

michalnieruchalski-tiugo:
Related Ticket: TINY-13045

Description of Changes:
* Include scrollbars in draggable boundaries, so that draggable element can't go below scrollbar

Pre-checks:
~~* [ ] Changelog entry added~~
~~* [ ] Tests have been added (if applicable)~~
* [x] Branch prefixed with `feature/`, `hotfix/` or `spike/`

Review:
~~* [ ] Milestone set~~
~~* [ ] Docs ticket created (if applicable)~~

GitHub issues (if applicable):

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed naming inconsistencies in draggable component boundary handling.
  * Improved viewport dimension detection for more accurate boundary constraints during drag operations.

* **Tests**
  * Updated test suite to validate boundary calculation refinements.

PR #10809, opened 2025-12-19
TINY-13548: add positioning config options to draggable

michalnieruchalski-tiugo:
Related Ticket: TINY-13548

Description of Changes:

`Draggable` component uses `transform` property to position itself while dragging. When dragging ends (pointerup) we used to translate `transform` into `top` and `left` properties. We do that to later wrap `top` and `left` values with css `min` to ensure that the element stays on the page on window resize.

In this PR I introduced new prop for `Draggable` component called `origin`, with values `top-left | top-right | bottom-left | bottom-right`. This property allows us to modify the behavior described above. So for example:

`<Draggable origin="top-right" />` will use css `top` and `right` properties to position itself on the page.

This option enables us to provide more natural feeling while resizing the window with elements that are initially near the right edge of the viewport. Especially now, that we allow partial overflow ( see https://github.com/tinymce/tinymce/pull/10795 )

Draggable:

- Add new `origin` property with default value of `top-left`
- Update documentation
- Where needed, run tests for multiple origin values

FloatingSidebar:

- Add new `origin` property with default value of `top-right`
- Change type of `initialPosition` property
- Remove obsolete tests
- Remove obsolete story
- Adjust tests
- Update documentation

Pre-checks:

~~\- [ ] Changelog entry added~~

Review:

~~\- [ ] Milestone set~~
~~\- [ ] Docs ticket created (if applicable)~~


michalnieruchalski-tiugo:
* **#10819** <a href="https://app.graphite.com/github/pr/tinymce/tinymce/10819?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10818** <a href="https://app.graphite.com/github/pr/tinymce/tinymce/10818?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10815** <a href="https://app.graphite.com/github/pr/tinymce/tinymce/10815?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10809** <a href="https://app.graphite.com/github/pr/tinymce/tinymce/10809?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/tinymce/tinymce/10809?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `epic/TINY-12378`


<h2></h2>

hamza0867 (approved):

modules/oxide-components/src/main/ts/components/draggable/Draggable.tsx
```diff
@@ -37,16 +37,17 @@ const Handle: FC<DraggableHandleProps> = ({ children }) => {
...
-  const { setShift, draggableRef, isDragging, setIsDragging, setPosition, allowedOverflow } = useDraggable();
+  const { setShift, draggableRef, isDragging, setIsDragging, setPosition, allowedOverflow, origin } = useDraggable();
 
   const stopDragging = useCallback(() => {
     setIsDragging(false);
     setShift({ x: 0, y: 0 });
     if (draggableRef.current !== null) {
       const rect = draggableRef.current.getBoundingClientRect();
-      setPosition({ x: Math.round(rect.left), y: Math.round(rect.top) });
+      const viewport = { width: document.documentElement.clientWidth, height: document.documentElement.clientHeight };
```
spocke, line 47: I think window.innerHeight would be the same value and shorter.

modules/oxide-components/src/main/ts/components/draggable/internals/calculations.ts
```diff
@@ -1,9 +1,22 @@
...
 
+/**
+ * Calculates element position relative to viewport based on origin point.
+ * Returns distance from left/right edge horizontally and top/bottom edge vertically.
+ */
+const position = (element: Position & Size, viewport: Size, origin: Origin): Position => {
+  const elementRight = element.x + element.width;
+  const elementBottom = element.y + element.height;
+  return {
+    x: Math.round(origin === 'top-left' || origin === 'bottom-left' ? element.x : viewport.width - elementRight),
```
spocke: Could we make these constants predicate checks instead so `Origin.isTopLeft` that way you won't need constants and you would also see where that value if kind of defined since that if a function in a module?

modules/oxide-components/src/main/ts/components/draggable/internals/styles.ts
```diff
@@ -1,24 +1,38 @@
...
 
-import type { AllowedOverflow, CssPosition, CssSize, Position, Shift } from './types';
+import type { AllowedOverflow, Origin, CssPosition, CssSize, Position, Shift } from './types';
 
 const getTransform = (shift: Shift): string => `translate3d(${shift.x}px, ${shift.y}px, 0px)`;
 
-const normalizePosition = (position: CssPosition | Position): CssPosition =>
-  'x' in position ? { top: `${position.y}px`, left: `${position.x}px` } : position;
+const normalizePosition = (position: CssPosition | Position): CssPosition => ({
+  x: typeof position.x === 'number' ? `${position.x}px` : position.x,
```
spocke: Don't we have a `Type.isNumber` ?

modules/oxide-components/src/test/ts/atomic/draggable/calculations.spec.ts
```diff
@@ -136,4 +136,59 @@ describe('browser.draggable.calculations', () => {
       });
     });
   });
+
+  describe('position', () => {
+    it.each([
```
spocke: I don't really like these loopy things for generating tests since say there is a bug with the top-left only then you need to comment out all the other tests to only focus on that when doing TDD then remember to uncomment it before you commit.

It can easily be refactored into this:

```
it('should calculate position for bottom-right origin', () => testOriginPosition({
  origin: 'bottom-right' as const,
  element: { x: 100, y: 200, width: 50, height: 50 },
  viewport: { width: 1000, height: 1000 },
  expected: { x: 850, y: 750 }
}));
```

The only downside is that it becomes a bit more repetition in the label string but now you can just to `it.only`, `it.skip` on it etc.

modules/oxide-components/src/main/ts/components/draggable/Draggable.tsx
michalnieruchalski-tiugo: That will not work as innerHeight and innerWidth include scrollbar.  
See: https://github.com/tinymce/tinymce/pull/10701

modules/oxide-components/src/main/ts/components/draggable/internals/calculations.ts
michalnieruchalski-tiugo: Will do.

Why would it be better though? The way I see it origin is not a string. It's a type with 4 possible values (like an enum).

modules/oxide-components/src/test/ts/atomic/draggable/calculations.spec.ts
michalnieruchalski-tiugo: agreed

PR #10810, opened 2025-12-19
TINY-13213: Prevent list element selection from being overzealous

HAFRMO:
Related Ticket: TINY-13213

Description of Changes:
If we are dealing with a non-collapsed range we want to ignore the forced root block as it causes the selection to step too far up and into parent elements ( it keeps going up divs, if you have a structure with several divs I'd just end up selecting everything after ).
The original issue that caused the check appears to have something to do with empty selection ranges, so this one attempts to side-step that issue by skipping that check if the selection is not collapsed.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

## Release Notes

**Bug Fixes**
* Improved handling of list operations when the text selection is collapsed, ensuring correct list context is maintained.
* Fixed issue when applying ordered list formatting to unordered lists, preserving the correct structure and selection state.

### DIFF
--- a/modules/tinymce/src/core/main/ts/lists/lists/Selection.ts
+++ b/modules/tinymce/src/core/main/ts/lists/lists/Selection.ts
@@ -14,7 +14,7 @@ const listSelector = listNames.join(',');
 const getParentList = (editor: Editor, node?: Node): HTMLElement | null => {
   const selectionStart = node || editor.selection.getStart(true);
 
-  return editor.dom.getParent(selectionStart, listSelector, getClosestListHost(editor, selectionStart));
+  return editor.dom.getParent(selectionStart, listSelector, getClosestListHost(editor, selectionStart, editor.selection.isCollapsed()));
 };
 
 const isParentListSelected = (parentList: HTMLElement | null, selectedBlocks: Element[]): boolean =>
@@ -36,9 +36,9 @@ const getSelectedSubLists = (editor: Editor): HTMLElement[] => {
   }
 };
 
-const findParentListItemsNodes = (editor: Editor, elms: Element[]): Element[] => {
+const findParentListItemsNodes = (editor: Editor, elms: Element[], isCollapsed: boolean): Element[] => {
   const listItemsElms = Tools.map(elms, (elm) => {
-    const parentLi = editor.dom.getParent(elm, 'li,dd,dt', getClosestListHost(editor, elm));
+    const parentLi = editor.dom.getParent(elm, 'li,dd,dt', getClosestListHost(editor, elm, isCollapsed));
 
     return parentLi ? parentLi : elm;
   });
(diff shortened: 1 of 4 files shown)


modules/tinymce/src/core/test/ts/browser/lists/ApplyTest.ts
```diff
@@ -22,6 +22,24 @@ describe('browser.tinymce.core.lists.ApplyTest', () => {
     base_url: '/project/tinymce/js/tinymce'
   });
 
+  const forcedhook = TinyHooks.bddSetupLight<Editor>({
```
spocke, line 25: I don't think we should have two hooks in the same context use a separate context. Each hook sets up it's own editor and it might be confusing to have two.

modules/tinymce/src/core/main/ts/lists/lists/Selection.ts
```diff
@@ -62,10 +62,10 @@ const getClosestEditingHost = (editor: Editor, elm: Element): HTMLElement => {
 const isListHost = (schema: Schema, node: Node): boolean =>
   !NodeType.isListNode(node) && !NodeType.isListItemNode(node) && Arr.exists(listNames, (listName) => schema.isValidChild(node.nodeName, listName));
 
-const getClosestListHost = (editor: Editor, elm: Node): HTMLElement => {
+const getClosestListHost = (editor: Editor, elm: Node, isCollapsed: boolean): HTMLElement => {
```
spocke, line 65: Since all call sites of this function simply gets the collapsed state of the current selection maybe it's better to just check that state inside this function instead of passing it into the function.

spocke (approved):

modules/tinymce/src/core/main/ts/lists/lists/Selection.ts
```diff
@@ -77,7 +77,7 @@ const isListInsideAnLiWithFirstAndLastNotListElement = (list: SugarElement<Node>
   );
 
 const findLastParentListNode = (editor: Editor, elm: Element): Optional<HTMLOListElement | HTMLUListElement> => {
-  const parentLists = editor.dom.getParents<HTMLOListElement | HTMLUListElement>(elm, 'ol,ul', getClosestListHost(editor, elm));
+  const parentLists = editor.dom.getParents<HTMLOListElement | HTMLUListElement>(elm, 'ol,ul', getClosestListHost(editor, elm, true));
```
lorenzo-pomili, line 80: why is this `true` and no `editor.selection.isCollapsed()`?

modules/tinymce/src/core/main/ts/lists/lists/Selection.ts
TheSpyder: Separating the parameters can be helpful for testing, but in this case it doesn't seem to provide any benefit (other than the `true` case lorenzo picked up, if that's valid then this is fine).

modules/tinymce/src/core/test/ts/browser/lists/ApplyTest.ts
TheSpyder: Agreed. Multiple editors on the page are a valid use case, but we need to keep our tests isolated unless we're specifically testing multiple editors.

modules/tinymce/src/core/main/ts/lists/lists/Selection.ts
HAFRMO: I'm not sure how much we want to pass around the range if we're just going to use it to perform a single is-collapsed-check, it feels better to just pass in the boolean. Particularly since it's not like we can get the element from the range either, it's different ones.

modules/tinymce/src/core/main/ts/lists/lists/Selection.ts
HAFRMO: I believe this was intentional. But I don't remember why.

PR #10811, opened 2025-12-19
TINY-13520: fix failing test

michalnieruchalski-tiugo:
Related Ticket: TINY-13520

Description of Changes:

- Adjust assertion now that we support parts of the sidebar being out of the viewport

Pre-checks:

~~- [ ] Changelog entry added~~
- [x] Tests have been added (if applicable)
- [x] Branch prefixed with `feature/`, `hotfix/` or `spike/`

Review:

~~- [ ] Milestone set~~
~~- [ ] Docs ticket created (if applicable)~~

GitHub issues (if applicable):

PR #10812, opened 2025-12-20
TINY-13374: Add styles and api needed for renaming a conversation

hamza0867:
Related Ticket: TINY-13374

Description of Changes:
* Updated the Spinner component to use Bem utility instead of classes utility
* Removed fill color from SVG path in Spinner component and moved it to CSS
* Fixed dependency arrays in useTabKeyNavigation and useFlowKeyNavigation hooks
* Added type import for SugarElement in TabbingType.ts
* Added styling for spinner in AI chat component
* Added styles for edit mode in chat history list items

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/oxide-components/src/main/ts/keynav/KeyboardNavigationHooks.ts
+++ b/modules/oxide-components/src/main/ts/keynav/KeyboardNavigationHooks.ts
@@ -38,8 +38,7 @@ export const useTabKeyNavigation = (props: TabKeyingProps): void => {
     } else {
       return Fun.noop;
     }
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, []);
+  }, [ props ]);
 };
 
 export interface FlowKeyingProps extends BaseProps, FlowType.FlowConfig { }
@@ -54,8 +53,7 @@ export const useFlowKeyNavigation = (props: FlowKeyingProps): void => {
     } else {
       return Fun.noop;
     }
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, []);
+  }, [ props ]);
 };
 
 export interface SpecialKeyingProps extends BaseProps, SpecialType.SpecialConfig { }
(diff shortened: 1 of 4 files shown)


hamza0867:
* **#10812** <a href="https://app.graphite.com/github/pr/tinymce/tinymce/10812?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/tinymce/tinymce/10812?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `epic/TINY-12378`


<h2></h2>

hamza0867:
CI is failing until #10811 is merged

modules/oxide/src/less/theme/components/ai-chat/ai-chat.less
```diff
@@ -239,14 +243,31 @@
...
+    cursor: pointer;
+
     &:hover, &:focus {
       // TODO: Replace with a proper global variable once we have it setup
       background-color: #f0f0f0;
     }
-    border-radius: var(--tox-private-control-border-radius, @control-border-radius);
-    cursor: pointer;
+    &.tox-ai-chat-history-list__item-edit-title {
+      background-color: #f0f0f0;
```
spocke: This needs to be derived from some color to work with darkmode and possible other skins.

hamza0867: That's very weird, it seems I forgot to update that specific value to a variable. Good catch!

modules/oxide-components/src/main/ts/keynav/KeyboardNavigationHooks.ts
```diff
@@ -54,8 +53,7 @@ export const useFlowKeyNavigation = (props: FlowKeyingProps): void => {
     } else {
       return Fun.noop;
     }
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, []);
+  }, [ props ]);
```
kimwoodfield, line 56: Do we need to watch the entire props object? Or could we just pass the specific props we need? I know `containerRef` won't be trackable but are we relying on specific things in the call to `FlowType.create`?

hamza0867: Ideally we should spread the objects that we need to watch and only watch those but that would make the code less readable and I doubt we would hit problems with the current approach as generally when using these hooks we only provide the bare minimum we need. The other way around (where we provide more than needed) would be more problematic, but practically this never happens

PR #10815, opened 2025-12-23
TINY-13520: fix incorect css formula

michalnieruchalski-tiugo:
Related Ticket: TINY-13520

Description of Changes:

- Fix incorrect positioning css formula
- Add unit tests for this function

### 

Pre-checks:

Review:
~~* [ ] Milestone set~~

### DIFF
--- a/modules/oxide-components/src/main/ts/components/draggable/internals/styles.ts
+++ b/modules/oxide-components/src/main/ts/components/draggable/internals/styles.ts
@@ -18,23 +18,23 @@ const getPosition = (position: CssPosition, allowedOverflow: AllowedOverflow, or
       let y = position.y;
 
       if (allowedOverflow.horizontal < 1) {
-        let visibleWidth: string | 0;
+        let visibleWidth: string;
         if (allowedOverflow.horizontal > 0) {
           const multiplier = (1 - allowedOverflow.horizontal).toFixed(2);
           visibleWidth = `${width} * ${multiplier}`;
         } else {
-          visibleWidth = width;
+          visibleWidth = `${width}`;
         }
         x = `min( ${position.x}, calc( 100% - ${visibleWidth} ) )`;
       }
 
       if (allowedOverflow.vertical < 1) {
-        let visibleHeight: string | 0;
+        let visibleHeight: string;
         if (allowedOverflow.vertical > 0) {
           const multiplier = (1 - allowedOverflow.vertical).toFixed(2);
           visibleHeight = `${height} * ${multiplier}`;
         } else {
-          visibleHeight = height;
+          visibleHeight = `${height}`;
         }
         y = `min( ${position.y}, calc( 100% - ${visibleHeight} ) )`;
       }


michalnieruchalski-tiugo:
* **#10819** <a href="https://app.graphite.com/github/pr/tinymce/tinymce/10819?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10818** <a href="https://app.graphite.com/github/pr/tinymce/tinymce/10818?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10815** <a href="https://app.graphite.com/github/pr/tinymce/tinymce/10815?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/tinymce/tinymce/10815?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10809** <a href="https://app.graphite.com/github/pr/tinymce/tinymce/10809?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `epic/TINY-12378`


<h2></h2>

modules/oxide-components/src/main/ts/components/draggable/internals/styles.ts
```diff
@@ -12,10 +12,21 @@ const normalizePosition = (position: CssPosition | Position): CssPosition => ({
...
     Fun.constant(position),
-    ({ width, height }) => ({
-      x: `min(${position.x}, calc(100% - (${width} * ${1 - allowedOverflow.horizontal}))`,
-      y: `min(${position.y}, calc(100% - (${height} * ${1 - allowedOverflow.vertical}))`
-    })
+    ({ width, height }) => {
+      let x = position.x;
+      let y = position.y;
+      if (allowedOverflow.horizontal < 1) {
+        const elementVisibleWidth = allowedOverflow.horizontal > 0
```
spocke: Would we simplify this having to much logic in the template string makes it harder to read and also it's in a ternary so if the multiplier part could be added to a variable it would be less dense and maybe also provide better meaning.

spocke (approved):

modules/oxide-components/src/main/ts/components/draggable/internals/styles.ts
```diff
@@ -13,10 +13,34 @@ const normalizePosition = (position: CssPosition | Position): CssPosition => ({
...
-    ({ width, height }) => ({
-      x: `min(${position.x}, calc(100% - (${width} * ${1 - allowedOverflow.horizontal}))`,
-      y: `min(${position.y}, calc(100% - (${height} * ${1 - allowedOverflow.vertical}))`
-    })
+    ({ width, height }) => {
+      let x = position.x;
+      let y = position.y;
+
+      if (allowedOverflow.horizontal < 1) {
+        let visibleWidth: string | 0;
```
spocke: Could we avoid this weird type by converting the width to a string when you make it a visibleString sure it has some runtime cost but that should not be that heavy it's not like we have thousands of these at the same time. So `visibleWidth = String(width)`

spocke (approved):

modules/oxide-components/src/main/ts/components/draggable/internals/styles.ts
michalnieruchalski-tiugo: This type is accurate it's not a hack. But yes we could avoid that with a String cast.

PR #10823, opened 2026-01-02
TINY-13560: Add PreviewLoader styles to oxide

michalnieruchalski-tiugo:
Related Ticket: TINY-13560

Description of Changes:

- Add `previewloader` styles to oxide

Pre-checks:

~~- [ ] Changelog entry added~~
~~- [ ] Tests have been added (if applicable)~~

Review:

~~- [ ] Milestone set~~
~~- [ ] Docs ticket created (if applicable)~~


michalnieruchalski-tiugo:
* **#10823** <a href="https://app.graphite.com/github/pr/tinymce/tinymce/10823?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/tinymce/tinymce/10823?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10822** <a href="https://app.graphite.com/github/pr/tinymce/tinymce/10822?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `epic/TINY-12378`


<h2></h2>

modules/oxide-components/src/main/ts/bespoke/tinymceai/previewloader/PreviewLoader.stories.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+import { Button } from '../../../components/button/Button';
+
+import { PreviewLoader } from './PreviewLoader';
+
+const meta = {
+  title: 'bespoke/tinymceai/PreviewLoader',
+  component: PreviewLoader,
+  parameters: {
+    // Optional parameter to center the component in the Canvas. More info: https://storybook.js.org/docs/configure/story-layout
```
spocke: Could we remove these boilerplate comments.

modules/oxide-components/src/main/ts/bespoke/tinymceai/previewloader/PreviewLoader.tsx
```diff
@@ -0,0 +1,22 @@
+import type { FunctionComponent, HTMLAttributes, PropsWithChildren, ReactNode } from 'react';
+
+import * as Bem from '../../../utils/Bem';
+import { Spinner } from '../spinner/Spinner';
+
+interface PreviewLoaderProps extends PropsWithChildren, HTMLAttributes<HTMLDivElement> {
```
spocke: Why does this inherit all props from HTMLDivElement I think we should only include the things we need in this case next to nothing.

modules/oxide-components/src/main/ts/bespoke/tinymceai/previewloader/PreviewLoader.tsx
```diff
@@ -0,0 +1,22 @@
...
+  text?: ReactNode;
```
spocke: Should this be readonly?

modules/oxide/src/less/theme/content/ai-preview/ai-preview.less
```diff
@@ -55,3 +55,15 @@ div[tinymceai-data-pending-diff="true"], span[tinymceai-data-pending-diff="true"
   z-index: 2;
   border-top: 1px solid hsl(from var(--tox-color-white, @color-white) h s calc(l - 11));
 }
+
+.tox .tox-tinymceai__previewloader-position {
+  width: fit-content;
+  height: fit-content;
+  position: fixed;
+  z-index: 1;
```
spocke: why was this needed?

spocke (approved):

modules/oxide-components/src/main/ts/bespoke/tinymceai/previewloader/PreviewLoader.stories.tsx
michalnieruchalski-tiugo: Yes, it was copy pasted

modules/oxide-components/src/main/ts/bespoke/tinymceai/previewloader/PreviewLoader.tsx
michalnieruchalski-tiugo: I only need `className` to be honest. The only problem is next time I will want add some attribute I would have to edit this component.  


If having minimal props is what we do here I will change it

modules/oxide/src/less/theme/content/ai-preview/ai-preview.less
michalnieruchalski-tiugo: There's another element with position fixed, so loader has to be on top of that

modules/oxide-components/src/main/ts/bespoke/tinymceai/previewloader/PreviewLoader.tsx
spocke: Yes, but we do control what is in there and it doesn't clash with new attributes that might be added to the spec for our custom ones. Why do you need the className exposed all classes should be as much as possible be internal concerns of the components.

modules/oxide-components/src/main/ts/bespoke/tinymceai/previewloader/PreviewLoader.tsx
hamza0867: Also having the prop called `text` when it's type is `ReactNode` is a bit confusing. Shouldn't it be more something like `previewLoaderContent` ?

hamza0867:
@coderabbitai review

modules/oxide-components/src/main/ts/bespoke/tinymceai/previewloader/PreviewLoader.tsx
michalnieruchalski-tiugo: It's supposed to represent text that will be rendered next to the spinner. I could make it   
  
text?: string  
  
But then underline and bold wouldn't be possible and I feel like we should support it.

`previewLoaderContent` seems a bit too verbose to me, it could be `content`but then it's weird to have children and content.

How about `label?: ReactNode`

spocke: But do we need rich text there?

michalnieruchalski-tiugo: No. It's the same conversation that we had for MenuItems. I don't care what `text` is, all I care is where to render it. My default is to keep it flexible

michalnieruchalski-tiugo: From developer perspective `<PreviewLoader text="Hello World" />` handy syntax is the same either way as ReactNode is:

```
type ReactNode =
    | ReactElement
    | string
    | number
    ...
```

We can make it more strict and make it more flexible once needed, or we can make it more flexible from the beginning. For labels I would prefer to have jsx possibilities by default

spocke: Then it should be called content or something since text to be is plain text so a string.
Menu items are a bit different in that we want to have complex things inside of them. For the abort button that should be outside the spinner right?

michalnieruchalski-tiugo: Abort button will be rendered in { children }. 

It can't be called content as IMO children is content.   
  
I agree text: ReactNode is weird, then I think it can be "label" or "labelContent" but I prefer to keep it short "label".  
  
If you have strong fillings against "label" then I can make this a text: string. It's super simple component I don't want us to spend too much time on it.

spocke: Why do we need text that is not a child here? Wouldn't children be enoug?

If I look at the design here the text and abort button could just be children if we need some special margin on the text then maybe that should be it's own component then some form of PreviewLabel or something.

michalnieruchalski-tiugo: The reason is margin. Initially I wanted it to only accept { children } as you're suggesting now.  
  
So it's spinner - text - button but margin between spinner and text is different than between text and button :sweat_smile:

spocke: I think this one will likely never be generic enough and it's just a wrapper for other things so I think this one can live in the `tinymceai` project instead. If we are to generalize this then it would likely be some form of generic popup container.  
  
So keep the CSS in this one and make another PR on the premium side with the component.

spocke (approved):

PR #10829, opened 2026-01-07
TINY-13450: Add basic and styled accordion component

ArvinJ-H:
Related Ticket: 
TINY-13450

Description of Changes:
- added accordion component
- added accordion story
- added tests

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

### DIFF
--- a/modules/oxide-components/src/main/ts/components/accordion/Accordion.tsx
+++ b/modules/oxide-components/src/main/ts/components/accordion/Accordion.tsx
@@ -189,7 +189,8 @@ const Item: FC<AccordionItemProps> = ({
     'icon-end': iconPosition === 'end'
   });
   const contentClassName = Bem.element('tox-accordion', 'content', {
-    expanded: isExpanded
+    expanded: isExpanded,
+    collapsed: !isExpanded
   });
 
   const iconElement = (


modules/oxide-components/src/main/ts/components/accordion/Accordion.tsx
```diff
@@ -0,0 +1,313 @@
...
+    const itemRefs = Array.from(getItemRefs().entries());
+    const currentIndex = Arr.findIndex(itemRefs, ([ itemId ]) => itemId === id);
+
+    if (currentIndex.isNone()) {
+      return;
+    }
+
+    const index = currentIndex.getOrDie();
+
+    switch (event.key) {
```
spocke: Shouldn't we use the keyboard navigation hooks we already have here?

modules/oxide-components/src/main/ts/components/accordion/Accordion.tsx
```diff
@@ -0,0 +1,313 @@
...
+        event.preventDefault();
+        if (itemRefs.length > 0) {
+          const lastRef = itemRefs[itemRefs.length - 1][1];
+          lastRef?.focus();
+        }
+        break;
+    }
+  }, [ disabled, toggleItem, id, getItemRefs ]);
+
+  const headerClassName = `tox-accordion__header${isExpanded ? ' tox-accordion__header--expanded' : ''}${disabled ? ' tox-accordion__header--disabled' : ''}${iconPosition === 'end' ? ' tox-accordion__header--icon-end' : ''}`;
```
spocke: Please use the Bem helper here to make this type safe and also easier to read. The Bem helper should be used anytime we have `className` basically inside the components.

I think this would be something like:

```
const headerClassName = Bem.element('tox-accordion', 'header', {
  expanded: isExpanded,
  disabled,
  'icon-end': iconPosition === 'end'
});
```

spocke (approved):

PR #10833, opened 2026-01-11
TINY-13379: Add content_id editor option

hamza0867:
Related Ticket: TINY-13379

Description of Changes:

- Added a new `content_id` option to TinyMCE core
- Implemented processor to validate that the option value is a string
- Added getter function `getContentId()` to retrieve the option value

Pre-checks:

- [x] Changelog entry added
- [x] ~~Tests have been added (if applicable)~~

Review:

- [x] ~~Milestone set~~

### DIFF
--- a/modules/tinymce/src/core/main/ts/api/Options.ts
+++ b/modules/tinymce/src/core/main/ts/api/Options.ts
@@ -921,6 +921,10 @@ const register = (editor: Editor): void => {
     default: 'Anonymous'
   });
 
+  registerOption('content_id', {
+    processor: 'string',
+  });
+
   registerOption('fetch_users', {
     processor: (value) => {
       if (value === undefined) {
--- a/.changes/unreleased/tinymce-TINY-13379-2026-01-12.yaml
+++ b/.changes/unreleased/tinymce-TINY-13379-2026-01-12.yaml
@@ -1,0 +1,6 @@
+project: tinymce
+kind: Added
+body: Editor option `content_id` that sets the id of the edited document.
+time: 2026-01-12T11:11:49.041709+01:00
+custom:
+    Issue: TINY-13379


hamza0867:
* **#10833** <a href="https://app.graphite.com/github/pr/tinymce/tinymce/10833?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/tinymce/tinymce/10833?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `epic/TINY-12378`


<h2></h2>

modules/tinymce/src/core/main/ts/api/Options.ts
```diff
@@ -921,6 +921,18 @@ const register = (editor: Editor): void => {
...
   });
 
+  registerOption('content_id', {
+    processor: (value) => {
+      if (value === undefined) {
+        return { valid: true, value: undefined };
+      }
+      if (Type.isString(value)) {
+        return { valid: true, value };
+      }
```
lorenzo-pomili: how about?
```suggestion
      if (value === undefined || Type.isString(value)) {
        return { valid: true, value };
      }
```

modules/tinymce/src/core/main/ts/api/Options.ts
```diff
@@ -921,6 +921,18 @@ const register = (editor: Editor): void => {
     default: 'Anonymous'
   });
 
+  registerOption('content_id', {
+    processor: (value) => {
```
spocke: Why not just use the built in processor for `string` ?

modules/tinymce/src/core/main/ts/api/Options.ts
```diff
@@ -1225,5 +1238,6 @@ export {
   getListMaxDepth,
   getDocumentsFileTypes,
   getFetchUsers,
-  getUserId
+  getUserId,
+  getContentId
```
spocke: If we are just registering an option and we never use it in the core why do we need to have this export nothing is using it?

spocke (changes requested):
I think we need a test at least that checks if you can define the `content_id` and if it's a string you can get it out using `editor.options.get('content_id')`.

spocke:
Also it needs a changelog since this is new public API and generic beyond the AI plugin.

modules/tinymce/src/core/main/ts/api/Options.ts
hamza0867: Because `undefined` is a valid value. If we use the built-in string processor then it wouldn't accept `undefined` as a value. See the definition of `fetch_users` a few lines below. What do you think ?

spocke: Why do we need undefined to be a valid value, if you don't want it then omit it from the options to tinymce.init?

hamza0867: That works indeed!

Issue #8649, opened 2023-04-14
directionality plugin does not let you set the current direction in content


andrewnicols:
**What is the current behavior? Describe the bug**

When working with the directionality plugin, the editor assumes that the current direction of the editor matches the final destination of the content.

It should be possible to set the direction attribute, even when the _current_ page direction matches the intended direction.

The content may be intended to exist within another directionality.

For example you may be placing some content into a div:

```
<div dir="rtl">
    <p>Some RTL content is here.</p>
    <div>
        <!-- This is where we will place our editor content -->
    </div>
</div>
```

When you create the TinyMCE editor, it may be in a separate location (such as a modal). If that is the case, then the parent direction may be "ltr" instead, and therefore when you set the "ltr" attribute, TinyMCE is _unsetting_ it.

**Please provide the steps to reproduce and if possible a minimal demo of the problem via [fiddle.tiny.cloud](https://fiddle.tiny.cloud/) or similar.**

1. Given your _current_ page has a direction of `ltr`
2. And you enter some content which you wish to force a direction of `ltr`
3. When I press the `LTR` directionality button
4. Then the `dir="ltr"` attribute should be set on the element

**What is the expected behavior?**

The `dir="ltr"` attribute should be set if you have explicitly pressed the button

**Which versions of TinyMCE, and which browser / OS are affected by this issue? Did this work in previous versions of TinyMCE?**
6.3.2 Chrome MacOS

It has not worked correctly in the older versions I have looked at.

The code behind that behaviour is here: <https://github.com/tinymce/tinymce/blob/develop/modules/tinymce/src/plugins/directionality/main/ts/core/Direction.ts#L30>

I think the ideal solution here is:

1. the buttons should have no active state if there is no attribute on the current, or any parent, element _within_ the editor
2. when pressing the button it should set the state
3. if you push the LTR button in an LTR setting, it should:
   - toggle on
   - toggle off
4. if you push the RTL button in an LTR setting, it should:
    - toggle on

TheSpyder:
Your solution makes sense from an engineer perspective, but not a user perspective. If the content is visibly LTR then the LTR button should be active, even if there is no explicit direction.

[...]

PR #10837, opened 2026-01-13
TINY-13337: Always apply directionality to blocks

MitchC1999:
Related Ticket: TINY-13337

Description of Changes:
* Always apply/unapply directionality to blocks
* Active state of directionality buttons now reflect directionality as set on parent blocks
* This means that neither directionality button is active by default, when there is no directionality style or attribute set on the parent block.
* Clicking these buttons toggles the directionality set on the parent block.
* Directionality inline style is only set on the parent block if already set, or if required to override CSS to apply desired directionality.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

GitHub issues (if applicable): [#8649](https://github.com/tinymce/tinymce/issues/8649)

## Summary by CodeRabbit

* **Bug Fixes**
  * Direction toggle now applies/removes direction on the selected block and shows active state only when the block itself has direction set; buttons respect editable context.
  * Improved detection of a block’s effective direction so toolbar state matches selection.

* **Tests**
  * Expanded browser tests for toggling behavior, nested lists, parent/child interactions, stylesheet-driven direction, and attribute/style combinations.

### DIFF
--- a/.changes/unreleased/tinymce-TINY-13337-2026-01-13.yaml
+++ b/.changes/unreleased/tinymce-TINY-13337-2026-01-13.yaml
@@ -1,6 +1,6 @@
 project: tinymce
 kind: Improved
-body: Directionality plugin buttons only appear active when directionality is toggled on selected block.
+body: Directionality buttons now only appear active when directionality is set on the selected block.
 time: 2026-01-13T13:12:06.721743+10:00
 custom:
     Issue: TINY-13337
--- a/.changes/unreleased/tinymce-TINY-13337-2026-01-14.yaml
+++ b/.changes/unreleased/tinymce-TINY-13337-2026-01-14.yaml
@@ -1,6 +1,6 @@
 project: tinymce
 kind: Improved
-body: Directionality plugin now always toggles directionality attribute on selected blocks.
+body: Directionality buttons now always toggle the directionality attribute on selected blocks.
 time: 2026-01-13T13:12:06.721743+10:00
 custom:
     Issue: TINY-13337


kimwoodfield (approved):

.changes/unreleased/tinymce-TINY-13337-2026-01-13.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Improved
+body: Always apply directionality to blocks.
```
TheSpyder: From my understanding of the request, this is a more accurate description?

```suggestion
body: Simplified the directionality plugin. It no longer considers inherited `dir` attributes, and when activated it will always apply to the nearest block element of the cursor.
```

modules/tinymce/src/plugins/directionality/main/ts/ui/Buttons.ts
```diff
@@ -1,35 +1,36 @@
...
-  editor.on('NodeChange', nodeChangeHandler);
-  api.setEnabled(editor.selection.isEditable());
+const register = (editor: Editor): void => {
+  const setupHandler = (dir: 'ltr' | 'rtl') => (api: Toolbar.ToolbarToggleButtonInstanceApi) => {
+    const nodeChangeHandler = (e: EditorEvent<NodeChangeEvent>) => {
+      const block = editor.dom.isBlock(e.element) ? e.element : editor.dom.getParent(e.element, editor.dom.isBlock);
+      if (block) {
+        const direction = editor.dom.getStyle(block, 'direction') || editor.dom.getAttrib(block, 'dir');
+        api.setActive(direction === dir);
+      }
```
TheSpyder, line 13: I don't know if we can update `Direction.getDirection()` with this new logic, it doesn't seem to be used in very many places, but I suppose that's a bit risky.

Perhaps refactor it into the plugin's `Direction` core module instead? And then this code will look a lot more like it did before?

TheSpyder (approved):

modules/tinymce/src/plugins/directionality/test/ts/browser/DirectionStyleTest.ts
```diff
@@ -26,91 +26,119 @@ describe('browser.tinymce.plugins.directionality.DirectionStyleTest', () => {
...
+      'ltr',
+      '<p dir="ltr" style="direction: ltr;">Lorem ipsum</p>'
+    ));
+
+    it('TINY-9314: Applying rtl with dir="ltr" and style direction: ltr', () => testDirectionStyle(
       hook.editor(),
       '<p dir="ltr" style="direction: ltr;">Lorem ipsum</p>',
       'rtl',
-      '<p dir="rtl">Lorem ipsum</p>'
+      '<p dir="rtl" style="direction: rtl;">Lorem ipsum</p>'
```
spocke, line 68: This is a change in behavior but is this because the style has higher precedence over the dir attribute. Maybe we should then always do the style not set both like this.

spocke (approved):

modules/tinymce/src/plugins/directionality/main/ts/ui/Buttons.ts
MitchC1999: Direction in Sugar returns the calculated direction, which is essential for knowing when we need to add the direction as an inline style if the dir attribute isn't enough. And in this case we don't want the button to be activated just from the calculated direction, only the inline style or dir attribute.

modules/tinymce/src/plugins/directionality/test/ts/browser/DirectionStyleTest.ts
MitchC1999: Yes, precedence goes inline style > stylesheet > dir attribute. I think we still want to apply only the dir attribute if thats all thats required, but we were already only setting the style if it was necessary.

.changes/unreleased/tinymce-TINY-13337-2026-01-13.yaml
MitchC1999: How about "Directionality plugin now always toggles directionality on blocks, including default directionality"?

TheSpyder: Apologies for not responding earlier, unless things are flagged for re-review I don't tend to see them.

I like your proposal, but I'd still like to mention in the changelog that the buttons will no longer toggle on for inherited direction. Perhaps we need two changelog entries instead of trying to cram both changes into one?

MitchC1999: Good idea

.changes/unreleased/tinymce-TINY-13337-2026-01-13.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Improved
+body: Directionality plugin buttons only appear active when directionality is toggled on selected block.
```
TheSpyder:
```suggestion
body: Directionality buttons now only appear active when directionality is set on the selected block.
```

.changes/unreleased/tinymce-TINY-13337-2026-01-14.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Improved
+body: Directionality plugin now always toggles directionality attribute on selected blocks.
```
TheSpyder:
```suggestion
body: Directionality buttons now always toggle the directionality attribute on selected blocks.
```

Issue #8939, opened 2023-07-27
BUG: `mceEditable` image not working as it did on previous versions 5.x


batata004:
Hi,

This bug only started to happen on version 6.x - it does not happen on all previous versions. If I have a parent element with the class `mceNonEditable`, on previous versions I could edit all elements that have the class `mceEditable` inside, even if it was an image. However, in the 6.x version, if the element that has the class `mceEditable` is an `img` the user cant select it, cant resize (using the blue squares).

You can check the bug at:

https://fiddle.tiny.cloud/PDiaab

Using exacly the same code, but on version 5.x the bug does not happen:

https://fiddle.tiny.cloud/QDiaab

Try to click the image, you will see in the first link nothing happens, on the second link you can select the image and even resize it using the blue squares on the corners.

TheSpyder:
We've been making some changes to the noneditable code in the last few months to improve templates. Using the fiddle version picker, it seems to work on 6.3.2 but appears to be a regression in 6.4.0. Can you confirm?

batata004:
> We've been making some changes to the noneditable code in the last few months to improve templates. Using the fiddle version picker, it seems to work on 6.3.2 but appears to be a regression in 6.4.0. Can you confirm?

You are correct. Until the version 6.3.2 it still works, then on 6.4.0 it gets broken up to the latest stable version.

TheSpyder:
Oh I forgot - yes, `noneditable` is now a core feature:
https://www.tiny.cloud/docs/tinymce/6/migration-from-5x/#removed-plugins

Note that the configuration options changed as a result of this move.

batata004:
@TheSpyder thank you for your attention. One final question: on versions prior to 6.x I could simply import the `noneditable` plugin and the user could only edit elements with class `mceEditable`. But sometimes, for admin reasons, I would not import the `noneditable` plugin in order for me to be able to change everything in the page (not only the `mceEditable` elements). Now, that the `noneditable` is a core feature, I ask you: how do I disable this plugin without having to remove all `mceEditable` and `mceNonEditable` class from all elements?

TheSpyder:
@batata004 good question. Does changing `noneditable_class` and `editable_class` work?
https://www.tiny.cloud/docs/tinymce/6/content-behavior-options/#noneditable_class
https://www.tiny.cloud/docs/tinymce/6/content-behavior-options/#editable_class

We lost the central documentation page for the feature when the plugin was removed, I've flagged that with our docs team.

[...]

PR #10838, opened 2026-01-13
TINY-10121: New option to control noneditable content

soritaheng:
Related Ticket: TINY-10121

Description of Changes:
* New `allow_noneditable` option to allow disabling the noneditable feature in core 
* Option is enabled by default and there is no change in the current behaviour 

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

GitHub issues (if applicable): https://github.com/tinymce/tinymce/issues/8939

## Summary by CodeRabbit

* **New Features**
  * Added an allow_noneditable option (enabled by default) to control handling and initialization of noneditable content in the editor.

* **Tests**
  * Reorganized test suite and added coverage for noneditable content injection, regexp handling, editable/noneditable interactions, and behavior with allow_noneditable disabled.

### DIFF
--- a/.changes/unreleased/tinymce-TINY-10121-2026-01-13.yaml
+++ b/.changes/unreleased/tinymce-TINY-10121-2026-01-13.yaml
@@ -1,6 +1,6 @@
 project: tinymce
 kind: Added
-body: 'New `allow_noneditable` option to control noneditable content.'
+body: 'The noneditable feature can now be disabled with the new `allow_noneditable` option.'
 time: 2026-01-13T13:57:48.402737+10:00
 custom:
     Issue: TINY-10121


spocke (commented):
I wonder if this is what we want here the customer could just configure a more unique class that doesn't match `mceEditable` and `mceNonEditable` or we could allow them to be configured to empty strings and then they wouldn't act on them. The noneditable feature does three things checks for editable class, the non editable class and some regexp. The regexp is optional and has no default.

The editable and non editable classes have defaults and the defaults are what could be problematic.

Is it worth adding a new option for this case that has been around for many years now. Seems like a edge case to me.

soritaheng:
@spocke Yes, changing the class names was suggested as a workaround in the original GitHub issue. From my understanding, adding this option is a matter of being backwards compatible with when the feature was a plugin. In that case, a user could choose not to load the plugin, and the default classes would not take effect. So, this option could be helpful when the feature is frequently used but needs to be opted out of in some specific cases. But let me know if the current behavior is more preferable.

.changes/unreleased/tinymce-TINY-10121-2026-01-13.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Added
+body: 'New `allow_noneditable` option to control noneditable content.'
```
TheSpyder:
```suggestion
body: 'The noneditable feature can now be disabled with the new `allow_noneditable` option.'
```

It's hard to describe this properly without using a double negative, I think this will do

tiny-ben-tran (approved):

.changes/unreleased/tinymce-TINY-10121-2026-01-13.yaml
soritaheng: Thanks Andy!

PR #10840, opened 2026-01-13
TINY-13387: add more tests for menu

ztomaszyk:
Related Ticket: TINY-13387

Description of Changes:
- Added comprehensive test coverage for Menu components
- Implemented keyboard navigation tests for Menu component
- Created new test files for MenuItem, MenuToggleItem, and SubmenuItem components
- Added tests for item enabling/disabling, toggling active states, and submenu interactions
- Fixed import paths for vitest browser utilities

Pre-checks:

Review:
* [x] Milestone set

## Summary by CodeRabbit

* **Refactor**
  * Improved internal state handling in menu components to stabilize enabled/active checks and avoid unnecessary API recreation.

* **Tests**
  * Added extensive browser tests covering keyboard navigation, item actions, toggle behavior, submenu hover interactions, and enabled/disabled scenarios.

### DIFF
--- a/modules/oxide-components/src/test/ts/browser/components/Menu.spec.tsx
+++ b/modules/oxide-components/src/test/ts/browser/components/Menu.spec.tsx
@@ -42,7 +42,7 @@ const waitForElementText = async (getByText: (text: string) => Locator, text: st
   await expect.poll(() => getByText(text).element()).toBeVisible();
 };
 
-const expectActiveElementToHaveText = async (text: string) => {
+const pAssertActiveElementText = async (text: string) => {
   await expect.poll(() => document.activeElement?.textContent).toBe(text);
 };
 
@@ -201,19 +201,19 @@ describe('browser.MenuTest', () => {
       const { getByText } = render(<TestComponent />, { wrapper });
       await waitForElementText(getByText, 'Item 1');
 
-      await expectActiveElementToHaveText('Item 1');
+      await pAssertActiveElementText('Item 1');
 
       await userEvent.keyboard('{ArrowDown}');
-      await expectActiveElementToHaveText('Item 2');
+      await pAssertActiveElementText('Item 2');
 
       await userEvent.keyboard('{ArrowDown}');
-      await expectActiveElementToHaveText('Item 3');
+      await pAssertActiveElementText('Item 3');
 
       await userEvent.keyboard('{ArrowUp}');
-      await expectActiveElementToHaveText('Item 2');
+      await pAssertActiveElementText('Item 2');
 
       await userEvent.keyboard('{ArrowUp}');
-      await expectActiveElementToHaveText('Item 1');
+      await pAssertActiveElementText('Item 1');
     });
 
     it('Should skip disabled items when navigating', async () => {
@@ -232,10 +232,10 @@ describe('browser.MenuTest', () => {
       const { getByText } = render(<TestComponent />, { wrapper });
       await waitForElementText(getByText, 'Item 1');
 
-      await expectActiveElementToHaveText('Item 1');
+      await pAssertActiveElementText('Item 1');
 
       await userEvent.keyboard('{ArrowDown}');
-      await expectActiveElementToHaveText('Item 3');
+      await pAssertActiveElementText('Item 3');
     });
   });
 });


zuzanna-antas:
* **#10840** <a href="https://app.graphite.com/github/pr/tinymce/tinymce/10840?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/tinymce/tinymce/10840?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

modules/oxide-components/src/main/ts/components/menu/components/Item.tsx
```diff
@@ -12,17 +12,22 @@ export const Item = forwardRef<HTMLButtonElement, MenuItemProps>(({ autoFocus =
...
+
+  useEffect(() => {
+    stateRef.current = state;
+  }, [ state ]);
 
   useEffect(() => {
     setState((prevState) => ({ ...prevState, enabled }));
   }, [ enabled ]);
 
   const api: CommonMenuItemInstanceApi = useMemo(() => ({
```
spocke, line 25: I wonder if it even make sense to do `useMemo` here creating an object is not that computationally heavy and looking up the memo and dealing with that is likely to be as slow as just creating the object on every re-render.

modules/oxide-components/src/test/ts/browser/components/Menu.spec.tsx
```diff
@@ -177,4 +177,61 @@ describe('browser.MenuTest', () => {
...
+              <Menu.Item onAction={vi.fn()}>Item 3</Menu.Item>
+            </Menu.Root>
+          </UniverseProvider>
+        );
+      };
+
+      const { getByText } = render(<TestComponent />, { wrapper });
+      await waitForElementText(getByText, 'Item 1');
+
+      await expect.poll(() => document.activeElement?.textContent).toBe('Item 1');
```
spocke: This could be a re-usable function like ``pAssertActiveElementText('Item 1`)`` I've been wondering if we should start making a mini library inside the tests for this with re-usable things.

modules/oxide-components/src/test/ts/browser/components/SubmenuItem.spec.tsx
```diff
@@ -0,0 +1,151 @@
...
+        </UniverseProvider>
+      );
+    };
+
+    const { getByText } = render(<TestComponent />, { wrapper });
+    await waitForElementText(getByText, 'Submenu');
+
+    await userEvent.hover(getByText('Submenu'));
+
+    await new Promise((resolve) => setTimeout(resolve, 300));
```
spocke, line 147: What is this wait for, why 300?

modules/oxide-components/src/main/ts/components/menu/components/Item.tsx
TheSpyder: It depends how the `useEffect` below operates. It lists `api` as a dependency, and the values in this object are functions. So making a new one might cause the effect to run every time.

I don't think the effect needs to list `api` as a dependency, though, since it's such a simple object.

modules/oxide-components/src/main/ts/components/menu/components/ToggleItem.tsx
```diff
@@ -13,21 +13,26 @@ export const ToggleItem = forwardRef<HTMLButtonElement, ToggleMenuItemProps>(({
     hovered: false,
   });
   const id = useId();
+  const stateRef = useRef(state);
+
+  useEffect(() => {
+    stateRef.current = state;
+  }, [ state ]);
```
TheSpyder, line 20: So is the idea to not cause a re-render when the state changes, since the `isEnabled` function will be used by other parts of the system?

... of course now that I've written it, of course that's why. Recreating the API instance won't update the one held by the menu (unless the menu is also recreated).

TheSpyder (approved):

modules/oxide-components/src/main/ts/components/menu/components/ToggleItem.tsx
zuzanna-antas: Yes, thats exaclty why. I noticed this in tests where I wanted to reference an api and it had all old values because the component was using a new api instance object.

modules/oxide-components/src/main/ts/components/menu/components/Item.tsx
zuzanna-antas: Not including `api` in dependency array breaks eslint rule. `useMemo` is used to not create new object on every rerender

modules/oxide-components/src/test/ts/browser/components/SubmenuItem.spec.tsx
zuzanna-antas: I am testing that the submenu is in fact not opened when you hover over a disabled menu item. I can't do the assertion immediately as it could mean that it just didn't opened **yet.** Thats why there is a wait. Maybe this test doesn't make much sense tho.

Issue #1491, opened 2014-02-24
paste_word_valid_elements Strips "strong" tags unless "b" is marked valid


TinyMCE-User:
Description of problem: Using paste_word_valid_elements: "strong" will strip all STRONG tags from the pasted content, but "strong,b" will leave them intact. 

Steps to reproduce:
1. Init with paste_word_valid_elements: "strong"
2. Paste something bold

Expected result: Bold font style (wrapped in P > STRONG tags)

Actual result: Normal font style (usually in a P tag)

Notes: I can't reproduce this easily in a Fiddle because I'm not using actual Word, so I have to force Word detection on my local copy. 
> _**Legacy information imported from TinyMCE bug tracker:**_
>
> _[#T6736](http://archive.tinymce.com/develop/bugtracker_view.php?id=6736) posted by [ahamilton9](http://archive.tinymce.com/profile.php?name=ahamilton9)_

> _**Tags:** [firefox msie safari chrome opera]_
> _**Status:** Open_
> _**Resolution:** None_
> _**Attached URL:** none_

blackcathikari:
Hi, we've investigated this issue and cannot reproduce it in TinyMCE v4.4.0. Thanks, the TinyMCE Team.

PR #10849, opened 2026-01-15
TINY-12819: Migrate yarn to bun

ArvinJ-H:
Related Ticket: TINY-12819

Description of Changes:

**Package Manager Swap**

- Migrate package manager from yarn to bun across the monorepo
- Replace all yarn commands with bun equivalents in scripts, CI, and docs
- Add `bunfig.toml`, `bun.lock`; remove `yarn.lock`, `.yarnrc`

**bunfig.toml**

- `frozenLockfile = true` (camelCase required; kebab-case is silently ignored by bun)
- `exact = true`, `linker = "hoisted"`, `bun = false` (webpack ~2.5x slower under bun runtime)

**CI/Build**

- Jenkinsfile: custom ECR containers, bun install in playwright pod, re-enable gitMerge
- Dockerfile.playwright-test: install bun + unzip
- Add `scripts/lerna-changie-hook.sh` for versioning lifecycle
- Sync visual regression test scripts and playwright config with main (yarn to bun swap only)

**Docs**

- Update [CONTRIBUTING.md](http://CONTRIBUTING.md), module READMEs, and oxide-components docs

**Notes**

- `lerna.json` uses `npmClient: "bun"` (not officially supported but works in practice)
- Playwright container installs bun at runtime (TODO: bake into custom ECR image)
- Premium: PR #1491

Pre-checks:

- [x] Changelog entry added

Review:

- [x] Milestone set

## Release Notes

- **Chores**
    - Migrated monorepo build toolchain from Yarn to Bun package manager
    - Updated configuration and dependency management files across all modules
    - Removed Yarn-specific configuration files and lockfiles
    - Updated CI/CD pipeline infrastructure for new package manager
- **Bug Fixes**
    - Fixed potential runtime error in iframe coordinate scaling computation by adding optional chaining
- **Documentation**
    - Updated development and testing instructions across all modules to use Bun commands


CONTRIBUTING.md
```diff
@@ -29,86 +29,85 @@ $ git clone https://github.com/tinymce/tinymce.git
```
TheSpyder, line 29: <p class='graphite__hidden'><i>[Re: line +9]</i></p>
The bot can't add a comment here but this should no longer mention yarn. Also, NPM did sort of work for a long time but this change probably breaks it again?
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/tinymce/tinymce/10849?utm_source=unchanged-line-comment">Graphite</a>.</p>

package.json
```diff
@@ -1,42 +1,43 @@
 {
+  "name": "tinymce-monorepo",
```
TheSpyder, line 2: I have a vague memory that setting a name here causes issues with nx in the premium builds. Please make sure this works in both of our build environments.

CONTRIBUTING.md
ArvinJ-H: oh right, my global search was case sensitive...  
  
I'll have a test with NPM very soon with Luke

package.json
ArvinJ-H: this name space was added at the beginning of bun ticket, the premium had a bit of on and off hiccup but i don't remember there were any issue with premium build.  
  
Now the core is relatively stable now I'll start rebasing premium changed.

ArvinJ-H: premium CI resolved

.dockerignore
```diff
@@ -4,7 +4,7 @@
 *.sublime-workspace
 
 package-lock.json
-yarn-error.log*
+bun-error.log*
```
TheSpyder: I highly doubt bun creates files like this, maybe just delete the line (also from gitignore)

Jenkinsfile
```diff
@@ -1,13 +1,17 @@
...
+@Field String ciRegistry = "${ciAccountId}.dkr.ecr.us-east-2.amazonaws.com"
+
 standardProperties()
 
 def runBedrockTest(String name, String command, Boolean runAll, int retry = 0, int timeout = 0) {
   def bedrockCmd = command + (runAll ? " --ignore-lerna-changed=true" : "")
   echo "Running Bedrock cmd: ${bedrockCmd}"
   def testStatus = sh(script: bedrockCmd, returnStatus: true)
-  junit allowEmptyResults: true, testResults: "scratch/TEST-${name}.xml"
+  junit allowEmptyResults: true, testResults: 'scratch/TEST-*.xml'
```
TheSpyder: This is reverting to an old version of the line, we now run multiple bedrock instances in the same codebase. So each one needs to only pull in a single XML file, otherwise we'll get duplicate test results.

modules/oxide-components/Dockerfile.playwright-test
```diff
@@ -3,6 +3,10 @@ ARG PLAYWRIGHT_VERSION=1.58.2-noble
 # Test stage
 FROM mcr.microsoft.com/playwright:v${PLAYWRIGHT_VERSION} AS test
 
+RUN apt-get update -qq && apt-get install -y -qq unzip && \
+    curl -fsSL https://bun.sh/install | bash -s "bun-v1.3.11"
+ENV PATH="/root/.bun/bin:${PATH}"
```
TheSpyder: I'm changing the docker image in #11015 to not use yarn, so we won't need this

modules/oxide-components/package.json
```diff
@@ -40,9 +40,9 @@
     "test-visual-local": "node ./visual-regression-test.mjs",
     "test-visual-local-update": "node ./visual-regression-test.mjs update",
     "test-ci": "vitest --watch=false --reporter=default --reporter=junit --outputFile=scratch/test-results.xml",
-    "test-browser-manual": "vitest --watch --browser.headless false",
-    "test-browser-headless": "vitest --watch",
-    "ci": "yarn lint && yarn --cwd ../../ lerna run build --scope @tinymce/oxide-components --include-dependencies && yarn test"
+    "test-browser-manual": "vitest --watch --project browser --browser.headless false",
+    "test-browser-headless": "vitest run --project browser",
```
TheSpyder: These are reverting back to old lines

modules/oxide-components/package.json
```diff
@@ -40,9 +40,9 @@
...
+    "ci": "bun lint && lerna run --stream build --scope @tinymce/oxide-components --include-dependencies && bun run test"
```
TheSpyder: This lerna command won't work, I'm also adjusting it in #11015 to use `npm exec` instead of `yarn`

modules/oxide-components/package.json
```diff
@@ -79,15 +79,6 @@
...
   },
-  "workspaces": {
-    "nohoist": [
-      "**/storybook",
-      "**/@storybook/**",
-      "**vite**",
-      "**/@vitejs/**",
-      "**/@vitest/**"
-    ]
-  },
```
TheSpyder, line 90: @hamza0867 can you remember why were these set to nohoist, and does it matter that they're removed in this bun PR?

modules/oxide-components/package.json
```diff
@@ -29,8 +29,8 @@
   "types": "lib/src/main/ts/main.d.ts",
   "scripts": {
     "dev": "vite",
-    "build": "npm run build-storybook && tsc -b && vite build",
-    "lint": "eslint --max-warnings=0 src/**/*.ts?",
+    "build": "bun build-storybook && tsc -b && vite build",
+    "lint": "eslint --max-warnings=0 'src/**/*.ts' 'src/**/*.tsx'",
```
TheSpyder, line 33: I prefer * over two lists. I thought we'd fixed all of these mistakes. Perhaps lint fixes should be a separate PR to reduce the noise in this one.

```suggestion
    "lint": "eslint --max-warnings=0 'src/**/*.ts*'",
```

package.json
```diff
@@ -1,42 +1,43 @@
...
+    "oxide-start": "bun run -F @tinymce/oxide start",
+    "oxide-ci": "bun run -F @tinymce/oxide ci",
+    "agar-sw-build": "bun run -F @ephox/agar-sw build",
     "tsc": "tsc -b",
     "watch": "tsc -b -w",
     "eslint": "lerna exec --stream --no-bail -- eslint --max-warnings=0 src/**/*.ts",
-    "start": "yarn -s tinymce-grunt start",
-    "dev": "npm-run-all agar-sw-build oxide-icons-build oxide-build -p tsc \"tinymce-grunt dev\"",
+    "start": "bun run tinymce-grunt start",
+    "dev": "npm-run-all2 oxide-icons-build oxide-build -p tsc \"tinymce-grunt dev\"",
```
TheSpyder: you've lost `agar-sw-build` (not sure what that is).

package.json
```diff
@@ -1,42 +1,43 @@
...
+    "ci": "npm-run-all2 -p oxide-ci oxide-icons-ci oxide-components-ci -p tsc \"tinymce-grunt dev\"",
     "ci-seq": "run-s agar-sw-build oxide-ci oxide-icons-ci oxide-components-ci tsc \"tinymce-grunt dev\"",
-    "ci-all": "npm-run-all -p changie-merge eslint ci -s tinymce-rollup",
-    "ci-all-seq": "npm-run-all -p eslint ci-seq -s tinymce-rollup",
+    "ci-all": "npm-run-all2 -p changie-merge eslint ci -s tinymce-rollup",
+    "ci-all-seq": "npm-run-all2 -p eslint ci-seq -s tinymce-rollup",
     "local-ci": "run-s -s ci-all test",
     "test": "run-s headless-test browser-test",
-    "build": "npm-run-all -p oxide-icons-ci oxide-ci oxide-components-ci -s tinymce-grunt",
+    "build": "npm-run-all2 -p oxide-icons-ci oxide-ci oxide-components-ci -s tinymce-grunt",
```
TheSpyder: The package changes to `npm-run-all2` but the script is aliased to both for compatibility

https://github.com/bcomnes/npm-run-all2/blob/b3290872caba4a78786c1a854561bb7dd6bd1804/package.json#L9-L10

- [x] revert all `2` additions

tsconfig.json
```diff
@@ -29,6 +29,7 @@
   ],
 
   "compilerOptions": {
+    "baseUrl": ".",
```
TheSpyder: Is this still necessary?

TheSpyder (approved):

Jenkinsfile
```diff
@@ -80,6 +84,108 @@ def runBrowserTests(String name, String browser, String platform, String bucket,
   runBedrockTest(name, bedrockCommand, runAll)
 }
 
+def runTestPod(String cacheName, String name, String testname, String browser, String provider, String platform, String version, String bucket, String buckets, Boolean runAll) {
```
TheSpyder: Why so many additions? Is this a bad merge or something?

modules/oxide-components/package.json
hamza0867: Yes, I do remember. Because yarn v1 had a bug where it installed dependencies in the wrong place, also sometimes it would not dedupe deps correctly. So I made these nohoist to have the repo kind of self contained. But I believe this should be fine for bun. I still have to test it and confirm

CONTRIBUTING.md
```diff
@@ -29,86 +29,85 @@ $ git clone https://github.com/tinymce/tinymce.git
 
 ### Install dependencies
 
-* `yarn`
+* `bun`
```
hamza0867:
```suggestion
* `bun i`
```

CONTRIBUTING.md
```diff
@@ -29,86 +29,85 @@ $ git clone https://github.com/tinymce/tinymce.git
...
 
-* `yarn build`
+* `bun run build`
 
 This will produce an editor build in `modules/tinymce/js`, with distribution zips in `modules/tinymce/dist/tinymce_[number].zip`.
 
 ## Developing TinyMCE
 
-* `yarn dev`
+* `bun dev`
```
hamza0867: This might work but it's better to use `bun run dev` to make it explicit we are calling the package.json script not a bun internal script

modules/oxide-components/package.json
hamza0867: I tested it and it works :+1:

hamza0867 (approved):

modules/oxide-components/package.json
ArvinJ-H: - [x] log a follow up ticket

CONTRIBUTING.md
ArvinJ-H: this does work currently, but yeah better be more explicit

Jenkinsfile
ArvinJ-H: something else, removed

modules/oxide-components/package.json
ArvinJ-H: https://ephocks.atlassian.net/browse/TINY-14254

modules/oxide-components/package.json
```diff
@@ -42,7 +42,7 @@
     "test-ci": "vitest --watch=false --reporter=default --reporter=junit --outputFile=scratch/test-results.xml",
     "test-browser-manual": "vitest --watch --browser.headless false",
     "test-browser-headless": "vitest --watch",
-    "ci": "npm exec --prefix ../../ lerna -- run --scope @tinymce/oxide-components --include-dependencies build && npm-run-all -p lint test"
+    "ci": "lerna run --scope @tinymce/oxide-components --include-dependencies build && npm-run-all -p lint test"
```
TheSpyder: The use of `npm exec` here was so that the command could be used while nested inside the monorepo, where `lerna` either isn't available or picks up the wrong root folder (I don't remember which).

Jenkinsfile
```diff
@@ -94,13 +98,14 @@ def cleanBuildName(String name) {
 
 def props
 
-def cacheName = "cache_${BUILD_TAG}"
+// def cacheName = "cache_${BUILD_TAG}"
+def cacheName = "${env.JOB_NAME.split('/').last()}_${env.BUILD_NUMBER}"
```
TheSpyder: What was this used for? Should we revert it before merge?

Jenkinsfile
```diff
@@ -286,12 +290,25 @@ timestamps { notifyStatusChange(
           container('playwright') {
             // 20 min: playwright unit + visual tests average ~8 min, max observed ~10m37s
             timeout(time: 20, unit: 'MINUTES') {
-              exec('yarn -s --cwd modules/oxide-components test-ci')
+              // Install bun in the playwright container (public image doesn't include it)
```
TheSpyder: I haven't looked at how bun manages `node_modules`, but I wonder.... if `npm exec` and/or `npm run` worked in a bun-managed repo we wouldn't need to do this.

ArvinJ-H: changed it to npm

modules/oxide-components/package.json
ArvinJ-H: It seems:
- Core CI: `bun ci-all-seq` → `ci-seq` → `oxide-components-ci`
- Premium CI: `make ci runs ${NX_RUNMANY} lint,build` 

They both work fine, and only core is calling that `ci` command directly?

TheSpyder: When it runs isn't what I was driving at, the point was that `lerna run --scope @tinymce/oxide-components` probably won't work if run in the `modules/oxide-components` directory. The `npm exec` command, and `yarn` before it, made sure lerna executed from the repository root.

ArvinJ-H: ah okay, that would be an issue, I'll revert this back

TheSpyder: Yeah it's a bit weird and perhaps should be coordinated by the repository CI script, but that's out of scope of this change.

kimwoodfield (approved):
As discussed, let's get this in so that we have ample time to jump on any unforseen issues that might pop up.

PR #10854, opened 2026-01-21
TINY-13391: New options to set default table header rows and columns

soritaheng:
Related Ticket: TINY-13391

Description of Changes:
* New `table_default_header_rows` and `table_default_header_cols` options to enable setting table headers by default
* Extended `assertTableStructureWithSizes` in `TableTestUtils` to support checking for header types 

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

GitHub issues (if applicable): https://github.com/tinymce/tinymce/discussions/10721

## Summary by CodeRabbit

* **New Features**
  * Added table_default_header_rows and table_default_header_cols; newly inserted tables now use editor defaults and support configurable header types.

* **Tests**
  * Added browser tests validating header defaults across header types and combinations of header rows/columns.

* **Refactor**
  * Updated table test utilities to accept and validate header type alongside header row/column settings.

### DIFF
--- a/.changes/unreleased/tinymce-TINY-13391-2026-01-21.yaml
+++ b/.changes/unreleased/tinymce-TINY-13391-2026-01-21.yaml
@@ -1,6 +1,6 @@
 project: tinymce
 kind: Added
-body: New `table_default_header_rows` and `table_default_header_cols` options to set default table headers
+body: New `table_default_header_rows` and `table_default_header_cols` options to set the default header size for new tables
 time: 2026-01-21T11:33:46.893719+10:00
 custom:
     Issue: TINY-13391


.changes/unreleased/tinymce-TINY-13391-2026-01-21.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Added
+body: New `table_default_header_rows` and `table_default_header_cols` options to set default table headers
```
TheSpyder:
```suggestion
body: New `table_default_header_rows` and `table_default_header_cols` options to set the default header size for new tables
```

modules/tinymce/src/models/dom/test/ts/browser/table/TableDefaultHeaderTest.ts
```diff
@@ -0,0 +1,34 @@
...
+    Arr.range(3, (headerRows) =>
+      Arr.range(3, (headerCols) => {
+        context(`${headerRows} header rows, ${headerCols} header cols with ${headerType}`, () => {
+          const hook = TinyHooks.bddSetup<Editor>({
+            base_url: '/project/tinymce/js/tinymce',
+            table_default_header_rows: headerRows,
+            table_default_header_cols: headerCols,
+            table_header_type: headerType,
+            table_use_colgroups: false
+          }, []);
```
TheSpyder: By calling `Options.defaultHeaderRows()` every time a table is inserted you've actually made this a dynamic property, which means it's possible to use `editor.options.set()` to update the config. It's not widely supported, many options can't work this way, but it's convenient when it works.

I think the code will be cleaner with a single `context` (or perhaps you'll need one context per `headerType`, I'm not sure if fhat's dynamic) then use the loops to generate just the `it` block with the rows/cols update inside.

The tests will also complete much faster as you aren't setting up and tearing down the editor 27 times - but you may find you then need a `beforeEach()` block to reset the editor content in between tests. I don't recall if we added that in the framework.

TheSpyder (approved):

modules/tinymce/src/models/dom/test/ts/browser/table/TableDefaultHeaderTest.ts
soritaheng: @TheSpyder Thanks for pointing that out! I've updated the test to set the options dynamically: https://github.com/tinymce/tinymce/pull/10854/commits/6091b52dfba8c1f00fff174f57f03c766f2af269

TheSpyder: Excellent, thank you. Did it make much difference to the test time?

soritaheng: Yes, definitely. The original test took around 8s. Now it takes 0.3s.

modules/tinymce/src/models/dom/test/ts/module/table/TableTestUtils.ts
```diff
@@ -210,42 +212,71 @@ const assertTableStructureWithSizes = (
...
-
-    const colGroup = s.element('colgroup', {
-      children: Arr.range(cols, () =>
-        s.element('col', {})
-      )
-    });
+  // In header types 'sectionCells' and 'cells', every cell in a header row is `th`.
+  // In header types 'sectionCells' and 'section', every header row is in the `thead`.
+  assertTableStructure(editor, ApproxStructure.build((s, str) => {
+    const colGroup = useColGroups
```
spocke: Multiline ternary maybe better to break this out some how maybe make a new function or the parts of the ternary into variables since this is not a perf sensitive part of the code.

modules/tinymce/src/models/dom/test/ts/module/table/TableTestUtils.ts
```diff
@@ -210,42 +212,71 @@ const assertTableStructureWithSizes = (
...
+    };
+
+    const headerRowCols = Arr.range(cols, () => s.element('th', colFields));
+
+    const bodyRowCols = [
+      ...Arr.range(options.headerCols, () => s.element('th', colFields)),
+      ...Arr.range(cols - options.headerCols, () => s.element('td', colFields))
+    ];
+
+    const tableHead = (headerType !== 'cells') && options.headerRows > 0
```
spocke: This becomes very complex fast if you have bunch of boolean logic then also a ternary that is multiline see if it can be cleaned up for readability. Since these are just substructures I think you should be able to break this out into fuincitons like createTableHead that returns a option or something then you convert that option to an array. So `const tableHead = createTableHead(headerType, options).toArray();` then that function can use if/else instead of a multi line ternary.

soritaheng: @spocke Thanks for the suggestions! I restructured the code some more to remove the multi line ternary operators in this commit: https://github.com/tinymce/tinymce/pull/10854/commits/a2c3f6c312853e92b967016db0cc1de7e9304e37 

I decided to keep the new helper functions in scope for now, but let me know if they should be more general and placed outside of the `assertTableStructureWithSizes` function.

PR #10860, opened 2026-01-27
TINY-13458: Refactor card component and add skeleton variant

ArvinJ-H:
Related Ticket: 
TINY-13458

Description of Changes:
- Add `loading?:` boolean prop to `Card.Root` that displays skeleton content when `true`
- Disable interactions (onClick, onKeyDown, onFocus) during loading
- Add `aria-busy` attribute for skeleton

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

## Summary by CodeRabbit

* **New Features**
  * Card gains a loading state with a reusable skeleton placeholder and a new demo story.

* **Style**
  * Updated card padding, hover shadow, and skeleton visual styles.

* **Tests**
  * Added tests for skeleton rendering, loading behavior, accessibility, and interaction during loading.

* **Refactor**
  * Card list split into controlled/uncontrolled variants with clearer runtime selection.

* **Chores**
  * Added project configuration file.

### DIFF
--- a/modules/oxide-components/src/test/ts/browser/components/Card.spec.tsx
+++ b/modules/oxide-components/src/test/ts/browser/components/Card.spec.tsx
@@ -646,32 +646,33 @@ describe('browser.components.CardTest', () => {
 
   describe('Skeleton Tests', () => {
     it('TINY-13458: Should render skeleton with default props', async () => {
-      const { container } = render(<Card.Skeleton />, { wrapper });
+      const { container } = render(<Card.Root loading />, { wrapper });
       const skeleton = container.querySelector('.tox-card.tox-skeleton');
       expect(skeleton).toBeTruthy();
       expect(skeleton?.querySelectorAll(Bem.elementSelector('tox-skeleton', 'line')).length).toBe(2); // 1 body + 1 actions
     });
 
     it('TINY-13458: Should render skeleton with custom line count', async () => {
-      const { container } = render(<Card.Skeleton lines={2} />, { wrapper });
+      const { container } = render(<Card.Root loading />, { wrapper });
       const skeleton = container.querySelector('.tox-card.tox-skeleton');
-      expect(skeleton?.querySelectorAll(Bem.elementSelector('tox-skeleton', 'line')).length).toBe(3); // 2 body + 1 actions
+      // The skeleton always renders 2 lines: 1 for body, 1 for actions (hardcoded)
+      expect(skeleton?.querySelectorAll(Bem.elementSelector('tox-skeleton', 'line')).length).toBe(2);
     });
 
     it('TINY-13458: Should be hidden from accessibility tree', async () => {
-      const { container } = render(<Card.Skeleton />, { wrapper });
+      const { container } = render(<Card.Root loading />, { wrapper });
       const skeleton = container.querySelector('.tox-card.tox-skeleton');
-      expect(skeleton?.getAttribute('aria-hidden')).toBe('true');
+      expect(skeleton?.getAttribute('aria-busy')).toBe('true');
     });
 
     it('TINY-13458: Should have tox-skeleton class for animation inheritance', async () => {
-      const { container } = render(<Card.Skeleton />, { wrapper });
+      const { container } = render(<Card.Root loading />, { wrapper });
       const skeleton = container.querySelector('.tox-card');
       expect(skeleton?.className).toContain('tox-skeleton');
     });
 
     it('TINY-13458: Should have proper card structure with body and actions', async () => {
-      const { container } = render(<Card.Skeleton />, { wrapper });
+      const { container } = render(<Card.Root loading />, { wrapper });
       const body = container.querySelector('.tox-card__body');
       const actions = container.querySelector('.tox-card__actions');
       expect(body).toBeTruthy();


modules/oxide-components/src/main/ts/components/card/Card.tsx
```diff
@@ -177,12 +174,32 @@ const Highlight: FC<CardHighlightProps> = ({ children, type }) => {
...
+};
+
 export {
   Root,
   Header,
   Body,
   Actions,
-  Highlight
+  Highlight,
+  Skeleton
```
kimwoodfield: Should `Skeleton` be exported? The `loading` prop covers individual card loading states, but Skeleton could be useful for placeholder cards while fetching lists at some point. 
If we don't need it externally then we could make it an implementation detail.

modules/oxide-components/src/main/ts/components/card/Card.tsx
```diff
@@ -177,12 +174,32 @@ const Highlight: FC<CardHighlightProps> = ({ children, type }) => {
...
+const Skeleton: FC<CardSkeletonProps> = ({ className, lines = 1 }) => {
+  const cardClassName = Bem.block('tox-card')
+    + ' tox-skeleton'
+    + (Type.isNonNullable(className) ? ` ${className}` : '');
+
+  return (
+    <div className={cardClassName} aria-hidden="true">
+      <div className={Bem.element('tox-card', 'body')}>
+        {Arr.range(lines, (i) => (
+          <div key={i} className="tox-skeleton__line" style={{ width: '100%' }} />
```
kimwoodfield: This looks very similar to the `skeletonContent` we defined on L110. 
We could have Root use the Skeleton component internally for consistency.

modules/oxide-components/src/main/ts/components/card/Card.tsx
```diff
@@ -37,19 +38,20 @@ export interface CardHighlightProps extends PropsWithChildren {
   readonly type: CardHighlightType;
 }
 
-/**
- * Card Root component.
- * Container for a card with support for selection states.
- * Must be used within a CardList for proper keyboard navigation.
- */
+export interface CardSkeletonProps {
+  readonly className?: string;
```
spocke: I don't think we should have arbitrary class names as props. Those should be inside the component. But I see what we already exposed that though the whole component. Something to discuss I guess.

modules/oxide-components/src/main/ts/components/card/Card.tsx
```diff
@@ -103,28 +103,38 @@ const Root: FC<CardRootProps> = ({
   const cardClassName = Bem.block('tox-card', {
     'selected': isFocused || selected,
     'has-decision': hasDecision
-  }) + (Type.isNonNullable(className) ? ` ${className}` : '');
+  })
+    + (loading ? ' tox-skeleton' : '')
```
spocke, line 107: Could we add a comment here since this is not following BEM so this is a tech depth the tox-skeleton should be a modifier to the card if it's supposed to be used that way so `tox-card--skeleton` I guess. Maybe we need to discuss how we track these things. I think we should deprecate all classes that is not following that format and then switch to a BEM format and remove the weird ones in v9.

I wonder if we should even tag these with actual code so we have like a `Bem.invalid(<class name>)` that simply checks if the class name is valid then just returns what is passed in we could then have find by reference in the editor to find all instances of these violations.

modules/oxide-components/src/main/ts/components/card/Card.tsx
spocke: The class name here is a element so use `Bem.element('tox-skeleton', 'line')`

modules/oxide-components/src/test/ts/browser/components/Card.spec.tsx
```diff
@@ -643,4 +643,139 @@ describe('browser.components.CardTest', () => {
       expect(asFragment()).toMatchSnapshot('CardList with multiple cards');
     });
   });
+
+  describe('Skeleton Tests', () => {
+    it('TINY-13458: Should render skeleton with default props', async () => {
+      const { container } = render(<Card.Skeleton />, { wrapper });
+      const skeleton = container.querySelector('.tox-card.tox-skeleton');
+      expect(skeleton).toBeTruthy();
+      expect(skeleton?.querySelectorAll('.tox-skeleton__line').length).toBe(2); // 1 body + 1 actions
```
spocke: You could use the `Bem.elementSelector` for this then you get type checking. That goes for all the selectors here except the non BEM standard one where the card and the skeleton is combined.

modules/oxide-components/src/main/ts/components/card/Card.tsx
ArvinJ-H: yeah the `tox-skeleton` already exist as a standalone, didn't make much sense to duplicat this, since its used for suggested edit also.  
  
I feel top level `tox-skeleton` should be kept, similar to other css classes. otherwise we'll be duplicating the css library, on top of what already rather complex and duplicated library?

modules/oxide-components/src/main/ts/components/card/Card.tsx
ArvinJ-H: yeah true, its not being used anywhere else, i think i exported it as one of the variant, just like every other one. I've made it internal only

PR #10877, opened 2026-02-05
TINY-13324: fix missing styles in new `li` with a sub-list inside

lorenzo-pomili:
Related Ticket: TINY-13324

Description of Changes:
the problem was that `trimInlineElementsOnLeftSideOfBlock` was removing some styles from the new block if it was a `li` with inside a sub-list, to fix it:
- now if we are in a case where the new block is a `li` with some styles + a sub-list we just trim the empty text inside it
- `NewListUtils.ts` now instead of creating a `&nbsp` in our case it creates a bogus br

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed formatting (e.g., font-size) persistence when creating new list items that contain sublists; improved handling of empty lines in lists by inserting a bogus line break to preserve Enter/caret behavior.

* **Tests**
  * Added and updated tests covering new list/line handling, styling preservation, and Enter key scenarios.

### DIFF
--- a/modules/tinymce/src/core/main/ts/newline/InsertBlock.ts
+++ b/modules/tinymce/src/core/main/ts/newline/InsertBlock.ts
@@ -13,6 +13,7 @@ import { findPreviousBr, isAfterBr } from '../caret/CaretBr';
 import * as CaretContainer from '../caret/CaretContainer';
 import CaretPosition from '../caret/CaretPosition';
 import { isAfterTable } from '../caret/CaretPositionPredicates';
+import { isList, isListItem } from '../dom/ElementType';
 import * as NodeType from '../dom/NodeType';
 import * as NormalizeRange from '../selection/NormalizeRange';
 import { isWhitespaceText } from '../text/Whitespace';
@@ -290,10 +291,9 @@ const insert = (editor: Editor, evt?: EditorEvent<KeyboardEvent>): void => {
     return true;
   };
 
-  const isInsideLiBeforeAList = (newBlock: Element) => {
-    const isList = (e: SugarElement) => /^(ul|ol|dl)$/.test(SugarNode.name(e));
-    const nextSibling = newBlock.firstChild?.nextSibling;
-    return NodeType.isListItem(newBlock) && nextSibling && isList(SugarElement.fromDom(nextSibling));
+  const isInsideLiBeforeAList = (newBlock: SugarElement<Element>) => {
+    const nextSibling = Traverse.firstChild(newBlock).bind(Traverse.nextSibling);
+    return isListItem(newBlock) && nextSibling.exists(isList);
   };
 
   const trimEmptySpacesInLeftLeaf = (newBlock: Element) => {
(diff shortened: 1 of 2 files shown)


.changes/unreleased/tinymce-TINY-13324-2026-02-05.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: Some formatting were not applied to items in multilevel list.
```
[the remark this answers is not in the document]
lorenzo-pomili: changed, have a check

modules/tinymce/src/core/main/ts/newline/InsertBlock.ts
```diff
@@ -289,6 +290,20 @@ const insert = (editor: Editor, evt?: EditorEvent<KeyboardEvent>): void => {
     return true;
   };
 
+  const isInsideLiBeforeAList = (newBlock: Element) => {
+    const isList = (e: SugarElement) => /^(ul|ol|dl)$/.test(SugarNode.name(e));
```
spocke: We have this function in the ElementType module there is a wanting about not using them but I think lists is not something we are going to configure though the schema. If we are to switch to using the schema for that as well we want to find all spots of this usage so having one function only would make it easier to spot than having it spread out here and there.

One could also add a isList to the NodeType that is using raw nodes then one doesn't need to lift the dom element into a sugar element just to see what type it is.

modules/tinymce/src/core/main/ts/newline/InsertBlock.ts
```diff
@@ -289,6 +290,20 @@ const insert = (editor: Editor, evt?: EditorEvent<KeyboardEvent>): void => {
...
+    const nextSibling = newBlock.firstChild?.nextSibling;
```
spocke: If we are mixing in sugar here then I think one should convert newBlock to a sugar element then use the traverse functions to get to the right spot then so the predicate checks as a exists call.

modules/tinymce/src/core/main/ts/newline/InsertBlock.ts
lorenzo-pomili: removed and used the one in `ElementType`, I also changed the checks in `NewLineUtils.ts` using `ElementType`

modules/tinymce/src/core/main/ts/newline/InsertBlock.ts
lorenzo-pomili: refactored now, is all SugarElement

modules/tinymce/src/core/main/ts/newline/InsertBlock.ts
```diff
@@ -289,6 +290,20 @@ const insert = (editor: Editor, evt?: EditorEvent<KeyboardEvent>): void => {
...
+    return NodeType.isListItem(newBlock) && nextSibling && isList(SugarElement.fromDom(nextSibling));
+  };
+
+  const trimEmptySpacesInLeftLeaf = (newBlock: Element) => {
+    const leaf = DomDescent.toLeaf(SugarElement.fromDom(newBlock), 0).element;
+
+    if (SugarNode.isText(leaf) && dom.isEmpty(leaf.dom)) {
+      leaf.dom.remove();
+    }
+  };
```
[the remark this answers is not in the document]
lorenzo-pomili: this code has been changed, have a check

PR #10880, opened 2026-02-05
TINY-13665: Add styles needed for resuming a conversation

hamza0867:
Related Ticket: TINY-13665

Description of Changes:

- Added cursor: pointer style to anchor tags with the tox-tag class within the AI chat component
- Add support for target prop in Tag when it is used as a link
- Fix circular dependency warning

Pre-checks:

- [x] Changelog entry added

Review:

- [x] Milestone set

## Summary by CodeRabbit

* **New Features**
  * Added support for customizable link targets in tags, allowing links to open in new windows or tabs.

* **Style**
  * Tag links now display a pointer cursor on hover for improved user experience.

### DIFF
--- a/modules/oxide-components/src/main/ts/bespoke/tinymceai/tag/Tag.tsx
+++ b/modules/oxide-components/src/main/ts/bespoke/tinymceai/tag/Tag.tsx
@@ -21,6 +21,7 @@ interface NonLinkTagProps {
 interface LinkTagProps {
   readonly link: true;
   readonly href: string;
+  readonly target?: string;
 }
 
 interface NonClosableProps {
@@ -40,6 +41,8 @@ export const Tag = forwardRef<HTMLDivElement | HTMLAnchorElement, TagProps>((pro
   const { label, icon, closeable, ariaLabel, link, ...rest } = props;
   const disabled = closeable && props.disabled === true;
   const href = link ? props.href : undefined;
+  const target = link ? (props.target ?? '_blank') : undefined;
+  const rel = link && target === '_blank' ? 'noopener noreferrer' : undefined;
 
   const content = (
     <>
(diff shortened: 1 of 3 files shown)


hamza0867:
* **#10880** <a href="https://app.graphite.com/github/pr/tinymce/tinymce/10880?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/tinymce/tinymce/10880?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

hamza0867:
@coderabbit review

modules/oxide-components/src/main/ts/bespoke/tinymceai/tag/Tag.tsx
```diff
@@ -40,6 +41,8 @@ export const Tag = forwardRef<HTMLDivElement | HTMLAnchorElement, TagProps>((pro
   const { label, icon, closeable, ariaLabel, link, ...rest } = props;
   const disabled = closeable && props.disabled === true;
   const href = link ? props.href : undefined;
+  const target = link ? props.target : '_blank';
```
spocke: THis doesn't look right is link if false it defaults to _blank ? Did you mean to check the props.target?

modules/oxide/src/less/theme/components/ai-chat/ai-chat.less
```diff
@@ -31,6 +31,9 @@
     flex-direction: column;
     gap: inherit;
     align-items: end;
+    a.tox-tag {
```
spocke: Is this really the right spot shouldn't we always set cursor: pointer on the `a.tox-tag` not just in this spot?

modules/oxide-components/src/main/ts/bespoke/tinymceai/tag/Tag.tsx
hamza0867: Good catch. Fixed!

modules/oxide/src/less/theme/components/ai-chat/ai-chat.less
hamza0867: Right. I moved this rule to the tag.less file. Now all `a.tox-tag` will have `cursor: pointer`

hamza0867:
@coderabbitai adding href and target in the spread of the props causes a type error because at the point of the spread we don't know if link is true or not. If link is false then there is no href prop and no target prop. In the case where link is true, then either the target prop is defined and that is the one used, if the target prop is not provided then we set target to '_blank' and it won't be overridden by the spread of the rest props because the rest does not contain a target prop. Does that make sense? I am making a mistake somewhere ?

michalnieruchalski-tiugo (approved):

PR #10895, opened 2026-02-13
TINY-13685: Improve Accordion text spacing and focus

kimwoodfield:
Related Ticket: [TINY-13685](https://ephocks.atlassian.net/browse/TINY-13685)

Description of Changes:
* Addressed an inconsistency between the implementation and design of the Accordion component, where there was too much spacing between the Accordion header and description once it was expanded.
* Adjusted the focus, such that the Accordion header does not receive focus once the Accordion is expanded, but the entire Accordion item itself.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

[TINY-13685]: https://ephocks.atlassian.net/browse/TINY-13685?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Refined keyboard navigation and focus management: root and item focus behavior improved, Enter/Space toggles, Escape refocuses, navigation skips interactive descendants, and mouse interactions better preserve focus.

* **Style**
  * Enhanced focus outlines and visual indicators, including forced-colors accessibility support and adjusted item/header backgrounds, padding, and expanded-state visuals.

* **Tests**
  * Updated keyboard/navigation tests to target item-level focus and interactions.

### DIFF
--- a/modules/oxide-components/src/main/ts/components/accordion/Accordion.tsx
+++ b/modules/oxide-components/src/main/ts/components/accordion/Accordion.tsx
@@ -144,6 +144,7 @@ const Root: FC<AccordionRootProps> = ({
 
   return (
     <AccordionContext.Provider value={contextValue}>
+      {/* TODO TINY-13777: Revisit tab order */}
       <div ref={containerRef} className="tox-accordion" tabIndex={0}>
         {children}
       </div>


modules/oxide-components/src/main/ts/components/accordion/Accordion.tsx
```diff
@@ -183,6 +203,84 @@ const Item: FC<AccordionItemProps> = ({
...
+      currentTarget.focus();
+      const isInteractive = target.matches(INTERACTIVE_SELECTOR) ||
+        target.hasAttribute('contenteditable') ||
+        target.getAttribute('role') === 'textbox';
+
+      if (!isInteractive) {
+        e.stopPropagation();
+      }
+    }
+  }, []);
```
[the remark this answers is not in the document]
kimwoodfield: The current behaviour is intentional - Escape always returns focus to the accordion item so that users can navigate between items with arrow keys. Without this, there's no keyboard-only way to return to accordion-level navigation once you've tabbed into the content.

modules/oxide-components/src/main/ts/components/accordion/Accordion.tsx
```diff
@@ -107,19 +120,31 @@ const Root: FC<AccordionRootProps> = ({
...
+
       focused.dom.click();
       return Optional.some(true);
     }
   });
 
   return (
     <AccordionContext.Provider value={contextValue}>
-      <div ref={containerRef} className="tox-accordion">
+      <div ref={containerRef} className="tox-accordion" tabIndex={0}>
```
spocke, line 148: I think ideally we should have -1 here and we should control tab order not the browser.

modules/oxide-components/src/main/ts/components/accordion/Accordion.tsx
```diff
@@ -183,6 +208,82 @@ const Item: FC<AccordionItemProps> = ({
...
+    const target = e.target as HTMLElement;
+    const currentTarget = e.currentTarget as HTMLElement;
+
+    if (target === currentTarget) {
+      e.preventDefault();
+      if (!disabled) {
+        toggleItem(id);
+      }
+    } else {
+      const isInteractiveTarget = target.matches(INTERACTIVE_SELECTOR) ||
```
spocke, line 256: I really think we should start using sugar for all DOM related things. Maybe not in this PR.

Since here for example you have a `boolean | Element | null` type. In sugar land this would have been `Selector.is(target, INTERACTIVE_SELECTOR) || SelectorExists.closest(target, INTERACTIVE_SELECTOR)` and then the type would be just `boolean` also that captures intent better.

modules/oxide-components/src/main/ts/components/accordion/Accordion.tsx
kimwoodfield: I agree that it can be improved, but the main goal of this ticket is to just fix the spacing between the header and description without messing with the focus styles too much. I've created a follow-up ticket to come back and take another look at this: https://ephocks.atlassian.net/browse/TINY-13777

modules/oxide-components/src/main/ts/components/accordion/Accordion.tsx
kimwoodfield: Yeah, I can see that being a nice approach.

PR #10904, opened 2026-02-17
TINY-13470: Add SegmentedControl component

kimwoodfield:
Related Ticket: [TINY-13470](https://ephocks.atlassian.net/browse/TINY-13470)

Description of Changes:
* Introduced a SegmentedControl component which functions similarly to the existing ToggleSwitch component with an updated design.

Screenshot 2026-02-18 at 8 37 22 am

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

[TINY-13470]: https://ephocks.atlassian.net/browse/TINY-13470?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Added a SegmentedControl UI with keyboard navigation, autofocus and ARIA accessibility.

* **Removals**
  * ToggleSwitch component and its Storybook stories removed.
  * Public API now exposes SegmentedControl instead of ToggleSwitch.

* **Documentation**
  * Storybook examples added demonstrating multiple SegmentedControl states and usages.

* **Tests**
  * Browser tests added to validate interactions, accessibility, disabled states, and multi-option behavior.

* **Updates**
  * Theme styling swapped to include segmented control styles and remove toggle styling.


modules/oxide-components/src/main/ts/components/segmentedcontrol/SegmentedControl.tsx
```diff
@@ -0,0 +1,147 @@
...
+        onClick();
+      }
+    };
+
+    const handleRightSegmentClick = () => {
+      if (!disabled && !checked) {
+        onClick();
+      }
+    };
+
```
michalnieruchalski-tiugo: Could we use replace these two handlers with hooks from https://github.com/tinymce/tinymce/blob/main/modules/oxide-components/src/main/ts/keynav/KeyboardNavigationHooks.ts

?

modules/oxide-components/src/main/ts/components/segmentedcontrol/SegmentedControl.tsx
```diff
@@ -0,0 +1,147 @@
...
+  type HTMLAttributes,
+  useRef,
+  useEffect
+} from 'react';
+
+import * as Bem from '../../utils/Bem';
+
+interface SegmentedControlProps extends Omit<HTMLAttributes<HTMLDivElement>, 'onClick'> {
+  readonly leftLabel: string;
+  readonly rightLabel: string;
```
hamza0867: I suggest we type these as a ReactNode instead, so we can put things like html in there or other components for fancier labels if we want

modules/oxide-components/src/main/ts/components/segmentedcontrol/SegmentedControl.tsx
```diff
@@ -0,0 +1,147 @@
...
+  readonly checked: boolean;
```
michalnieruchalski-tiugo: In my opinion boolean flag cannot be used to accurately represent the state of this component. This was fine for the Toggle, but not for SegmentedControl.

For example, why does `checked=true` mean the right button is checked, not the other way around?

In my opinion this component is more like a Radio Group, with two Radio Buttons, than Checkbox/Toggle.

Here's the example implementation of similar component in Radix https://www.radix-ui.com/themes/docs/components/segmented-control

I would argue that it should be implemented like a Radio Group, with each option knowing it's `value`.

modules/oxide-components/src/main/ts/components/segmentedcontrol/SegmentedControl.tsx
```diff
@@ -0,0 +1,147 @@
...
+    checked,
+    onClick,
+    onKeyDown,
+    disabled,
+    ...rest
+  }, ref) => {
+    const leftSegmentRef = useRef<HTMLSpanElement>(null);
+    const rightSegmentRef = useRef<HTMLSpanElement>(null);
+    const shouldFocusLeft = useRef(false);
+    const shouldFocusRight = useRef(false);
```
hamza0867: Can we have both of these true at the same time ? If not, then maybe this should be turned into a union ?

modules/oxide-components/src/main/ts/components/segmentedcontrol/SegmentedControl.tsx
```diff
@@ -0,0 +1,147 @@
...
+        onKeyDown(e);
+      }
+    };
+
+    const getTabIndex = (isActive: boolean): number => {
+      if (disabled) {
+        return -1;
+      }
+      return isActive ? 0 : -1;
+    };
```
hamza0867: This seems to be using the roving tabindex technique. Instead we might want to use the keyboard navigation hooks we already have.

hamza0867:
This seems to be missing behaviour tests (not to be confused with visual tests).  
  
Edit: tests were added

modules/oxide-components/src/main/ts/components/segmentedcontrol/SegmentedControl.tsx
kimwoodfield: Yes, good point. I haven't really seen that UI before but I like that it's not limited to just two segments. I'll refactor it to follow that compound pattern.

modules/oxide-components/src/main/ts/components/segmentedcontrol/SegmentedControl.tsx
kimwoodfield: Those are no longer apart of the API with the new pattern.

modules/oxide-components/src/main/ts/components/segmentedcontrol/SegmentedControl.tsx
kimwoodfield: These were removed with the most recent changes.

kimwoodfield:
> This seems to be missing behaviour tests (not to be confused with visual tests)

@hamza0867  I didn't prepare any initially as there hadn't been any done for the ToggleSwitch component. I've included some as part of the latest change.

modules/oxide-components/src/main/ts/components/segmentedcontrol/SegmentedControl.tsx
```diff
@@ -0,0 +1,157 @@
...
+  readonly onChange: (value: string) => void;
+  readonly disabled?: boolean;
+  readonly children: ReactNode;
+}
+
+interface SegmentedControlOptionProps {
+  readonly value: string;
+  readonly disabled?: boolean;
+  readonly children: ReactNode;
+}
```
hamza0867: A nitpick:

```suggestion
interface PropsWithChildren<SegmentedControlOptionProps> {
  readonly value: string;
  readonly disabled?: boolean;
}
```

Same for other children accepting component

modules/oxide-components/src/main/ts/components/segmentedcontrol/SegmentedControl.tsx
```diff
@@ -0,0 +1,148 @@
...
+  useContext,
+  useEffect,
+  useRef
+} from 'react';
+
+import * as KeyboardNavigationHooks from '../../keynav/KeyboardNavigationHooks';
+import * as Bem from '../../utils/Bem';
+
+interface SegmentedControlContextValue {
+  readonly value: string;
```
michalnieruchalski-tiugo, line 18: I would say, as long as values are unique we shouldn't care weather it's string, number or an object, or something else.  


It's up to you if you want to change that or not. I'm happy with what we have now

michalnieruchalski-tiugo (approved):

modules/oxide-components/src/main/ts/components/segmentedcontrol/SegmentedControl.tsx
```diff
@@ -0,0 +1,148 @@
...
+    }, [ ref ]);
+
+    KeyboardNavigationHooks.useFlowKeyNavigation({
+      containerRef,
+      selector: '[role="radio"]',
+      allowHorizontal: true,
+      allowVertical: false,
+      cycles: true,
+      execute: (focused: SugarElement<HTMLElement>) => {
+        const optionValue = focused.dom.getAttribute('data-value');
```
spocke: Why not sure the sugar APIs `Attribute.get(focused, 'data-value')` ?

spocke (approved):

modules/oxide-components/src/main/ts/components/segmentedcontrol/SegmentedControl.tsx
kimwoodfield: I think keeping it as a string for now is the right choice. It's sufficient for our current use case and we can avoid adding any immediate complexity. For sure we can add it later if and when it's needed

modules/oxide-components/src/main/ts/components/segmentedcontrol/SegmentedControl.tsx
kimwoodfield: Yes, I missed those. Thanks for spotting them

PR #10912, opened 2026-02-20
TINY-12918: Restore focus to dialogs when unblocked.

MitchC1999:
Related Ticket: TINY-12918

Description of Changes:
* Dialogs would lose focus when blocked because all components are disabled, and we don’t restore focus when the dialog is unblocked.
* In a regular editor, or with a modal dialog, the dialog doesn’t disappear when focus is lost. With an inline (persistent) dialog in an inline editor, the dialog is only visible when either it is focused or the editor is focused. Since neither are focused when the dialog is unblocked, the dialog disappears.
* The solution is to automatically restore focus when unblocking by default on all dialogs.
* When a dialog is opened, the first focusable element is automatically focused, so this solution is just applying the same logic when a dialog is unblocked.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed dialogs (including inline/persistent ones) not restoring focus after being blocked and unblocked, improving keyboard navigation and accessibility across affected dialogs.

* **Tests**
  * Added an automated browser test asserting that interactive elements in persistent inline dialogs retain focus after block/unblock cycles to prevent regressions.


modules/tinymce/src/themes/silver/test/ts/browser/window/SilverInlineDialogTest.ts
```diff
@@ -267,4 +267,61 @@ describe('browser.tinymce.themes.silver.window.SilverInlineDialogTest', () => {
...
+    await FocusTools.pTryOnSelector(
+      'Focus should be on the block button',
+      SugarDocument.getDocument(),
+      'button[data-mce-name="Block"]'
+    );
+
+    // Click the block button
+    Mouse.trueClickOn(SugarBody.body(), 'button[data-mce-name="Block"]');
+
+    // Wait for the dialog to be unblocked (50ms + some buffer)
```
spocke: Isn't there a proper state we can wait for here like checking if a item in the dialog is no longer disabled then move on. These wait for a while, slows down test runs and potentially adds flakes if we can avoid them it would be better.

MitchC1999: Yea I'll have a look. This test was entirely generated by claude, goes to show how efficient AI code really is 🤔

.changes/unreleased/tinymce-TINY-12918-2026-02-20.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: Dialogs would not restore focus after being unblocked.
```
TheSpyder:
```suggestion
body: Dialogs did not retain focus when using the block/unblock API.
```

modules/tinymce/src/themes/silver/test/ts/browser/window/SilverInlineDialogTest.ts
TheSpyder: The test can call block and unblock itself, it has the `api` object, instead of using buttons in the dialog to do that. Then use a "wait for selector" call (or I think there's a focus-specific waiter?) to drive the assertions instead of timing loops.

modules/tinymce/src/themes/silver/main/ts/ui/window/SilverDialogEvents.ts
```diff
@@ -46,12 +69,15 @@ const initCommonEvents = <A, S extends EventSpec<A>>(fireApiEvent: FireApiFunc<A
     AlloyTriggers.emit(self, formCloseEvent);
   }),
 
-  AlloyEvents.run<FormUnblockEvent>(formUnblockEvent, (_c, _se) => extras.onUnblock()),
+  AlloyEvents.run<FormUnblockEvent>(formUnblockEvent, (component, _se) => {
+    extras.onUnblock();
+    focusFirstTabbable(getSink, component);
```
TheSpyder, line 74: It would be nice if this could restore focus from before it was blocked, but this is still an improvement over not having focus at all

MitchC1999: In the case of the AI plugin the dialog redials while blocked, and the previously focused component, the prompt input, becomes disabled. Since we can't assume the same component would be focusable or still exist, I'm just applying the same logic as when the dialog opens.

TheSpyder: We could still try, and fall back to the first component if it can't be restored to the same component.

But this is not essential which is why I approved it.

modules/tinymce/src/themes/silver/test/ts/browser/window/SilverInlineDialogTest.ts
```diff
@@ -267,4 +267,31 @@ describe('browser.tinymce.themes.silver.window.SilverInlineDialogTest', () => {
...
+    await FocusTools.pTryOnSelector(
+      'Focus should be on the input when dialog opens',
+      SugarDocument.getDocument(),
+      'input'
+    );
+
+    // Block the dialog for a short time to simulate an async action, then unblock it
+    api.block('Blocking dialog for test');
+    await Waiter.pWait(10);
+    api.unblock();
```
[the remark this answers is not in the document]
MitchC1999: Document focus moves slightly asynchronously. When the dialog blocks, focus leaves shortly afterward, which is why there is a wait before unblocking, to allow for the focus to leave.

PR #10923, opened 2026-02-24
TINY-13700: deprecate `exportpdf_converter_options`

lorenzo-pomili:
Related Ticket: TINY-13700

Description of Changes:
deprecation for TINY-13700, is the changelog is needed or should be included in the premium change?

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [x] Milestone set

## Summary by CodeRabbit

* **Chores**
  * Marked the 'exportpdf_converter_options' configuration option as deprecated. Users will receive warnings when using this option.

### DIFF
--- a/modules/tinymce/src/core/main/ts/Deprecations.ts
+++ b/modules/tinymce/src/core/main/ts/Deprecations.ts
@@ -18,7 +18,7 @@ const removedOptions = (
   'template_cdate_classes,template_mdate_classes,template_selected_content_classes,template_preview_replace_values,template_replace_values,templates,template_cdate_format,template_mdate_format'
 ).split(',');
 
-const deprecatedOptions: string[] = [ 'content_css_cors' ];
+const deprecatedOptions: string[] = [ 'content_css_cors', 'exportpdf_converter_options' ];
 
 const removedPlugins = 'bbcode,colorpicker,contextmenu,fullpage,legacyoutput,spellchecker,template,textcolor,rtc'.split(',');
 


modules/tinymce/src/core/main/ts/Deprecations.ts
```diff
@@ -18,7 +18,7 @@ const removedOptions = (
   'template_cdate_classes,template_mdate_classes,template_selected_content_classes,template_preview_replace_values,template_replace_values,templates,template_cdate_format,template_mdate_format'
 ).split(',');
 
-const deprecatedOptions: string[] = [ 'content_css_cors' ];
+const deprecatedOptions: string[] = [ 'content_css_cors', 'exportpdf_converter_options' ];
```
spocke, line 21: I wonder if this deprecation should be in that plugin instead. But there is a benefit of having it all in one spot as well.

lorenzo-pomili: I used this because looking for `deprecate` in the merged PR and I found the PR that deprecated `content_css_cors`, maybe having it in the plugin could allow us to have a more custom error, but having this centralized means that in the next version we should be able to check which options should we remove more easily

TheSpyder: I think we usually put premium deprecations here as well. But as discussed in the other PR I'd rather not rename the option at all.

lorenzo-pomili:
this is no more needed

PR #10924, opened 2026-02-24
TINY-13854: add prevent default on closing a tag

ztomaszyk:
Related Ticket: TINY-13854

Description of Changes:
* Added preventDefault() call to the Tag component's close button click handler to prevent triggering href opening 

Pre-checks:

Review:
* [x] Milestone set

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed tag close button behavior to prevent unintended default actions when closing tags.

### DIFF
--- a/modules/oxide-components/src/main/ts/bespoke/tinymceai/tag/Tag.tsx
+++ b/modules/oxide-components/src/main/ts/bespoke/tinymceai/tag/Tag.tsx
@@ -62,7 +62,10 @@ export const Tag = forwardRef<HTMLDivElement | HTMLAnchorElement, TagProps>((pro
       <span className={Bem.element('tox-tag', 'label')}>{label}</span>
       {closeable && (
         <span className={Bem.element('tox-tag', 'close')}>
-          <IconButton icon='source-close' variant='naked' disabled={disabled} onClick={props.onClose} aria-label={ariaLabel} />
+          <IconButton icon='source-close' variant='naked' disabled={disabled} aria-label={ariaLabel} onClick={(e) => {
+            e.preventDefault();
+            props.onClose();
+          }} />
         </span>
       )}
     </>


zuzanna-antas:
* **#10924** <a href="https://app.graphite.com/github/pr/tinymce/tinymce/10924?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/tinymce/tinymce/10924?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

modules/oxide-components/src/main/ts/bespoke/tinymceai/tag/Tag.tsx
```diff
@@ -62,7 +62,10 @@ export const Tag = forwardRef<HTMLDivElement | HTMLAnchorElement, TagProps>((pro
       <span className={Bem.element('tox-tag', 'label')}>{label}</span>
       {closeable && (
         <span className={Bem.element('tox-tag', 'close')}>
-          <IconButton icon='source-close' variant='naked' disabled={disabled} onClick={props.onClose} aria-label={ariaLabel} />
+          <IconButton icon='source-close' variant='naked' disabled={disabled} aria-label={ariaLabel} onClick={(e) => {
+            e.preventDefault();
```
TheSpyder, line 66: Why is the link around the whole tag in the first place? Shouldn't it be just around the label?

Won't this still cause the cursor to be a hand rather than a pointer when hovering over the button?

zuzanna-antas: > Why is the link around the whole tag in the first place? Shouldn't it be just around the label?

When the tag has a link the whole thing is an `a` element instead of a `div` so I guess that's by design. 

> Won't this still cause the cursor to be a hand rather than a pointer when hovering over the button?

All buttons have cursor pointer in tiny. There is also a style for link tags to have a cursor pointer so it's a pointer for both close button and the whole tag in this case.

TheSpyder: It's "by design" that I am questioning. I don't think the whole thing needs to be a clickable link. But without UX context it's hard to say that for sure.

Toolbar buttons don't use a pointer, but I'd forgotten dialogs buttons do. It's a decision I disagreed with so long ago that I forgot and I don't use our dialogs much anymore :melting_face:

PR #10942, opened 2026-03-04
TINY-13937: Fix SegmentedControl to use the first option as tabstop

kimwoodfield:
Related Ticket: [TINY-13937](https://ephocks.atlassian.net/browse/TINY-13937)

Description of Changes:
* Updated SegmentedControl to always put the first option in the tab order instead of the selected option. This provides a consistent tabstop for external keyboard navigation and follows the standard roving tabindex patterns.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

[TINY-13937]: https://ephocks.atlassian.net/browse/TINY-13937?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced keyboard navigation behavior in segmented controls to properly set the first non-disabled option as the default focus point.

* **Tests**
  * Updated test cases to validate focus behavior, tab order, and handling of disabled options in segmented controls.

### DIFF
--- a/modules/oxide-components/src/main/ts/components/segmentedcontrol/SegmentedControl.tsx
+++ b/modules/oxide-components/src/main/ts/components/segmentedcontrol/SegmentedControl.tsx
@@ -1,4 +1,4 @@
-import { Optional, Type } from '@ephox/katamari';
+import { Arr, Optional, Type } from '@ephox/katamari';
 import { Attribute, type SugarElement } from '@ephox/sugar';
 import {
   createContext,
@@ -66,14 +66,13 @@ const Root = forwardRef<HTMLDivElement, SegmentedControlRootProps>(
 
     const firstOptionValue = useMemo(() => {
       const childArray = Children.toArray(children);
-      const firstNonDisabledOption = childArray.find(
-        (child): child is ReactElement<SegmentedControlOptionProps> =>
-          isValidElement(child) &&
-          typeof child.type !== 'string' &&
-          !disabled &&
-          !child.props.disabled
+
+      const validOptions = Arr.filter(childArray, (child): child is ReactElement<SegmentedControlOptionProps> =>
+        isValidElement(child) && typeof child.type !== 'string'
       );
-      return firstNonDisabledOption?.props.value ?? null;
+
+      const firstNonDisabledOption = Arr.find(validOptions, (option) => !disabled && !option.props.disabled);
+      return firstNonDisabledOption.map((option) => option.props.value).getOrNull();
     }, [ children, disabled ]);
 
     KeyboardNavigationHooks.useFlowKeyNavigation({


modules/oxide-components/src/main/ts/components/segmentedcontrol/SegmentedControl.tsx
```diff
@@ -59,6 +64,17 @@ const Root = forwardRef<HTMLDivElement, SegmentedControlRootProps>(
...
 
+    const firstOptionValue = useMemo(() => {
+      const childArray = Children.toArray(children);
+
+      const validOptions = Arr.filter(childArray, (child): child is ReactElement<SegmentedControlOptionProps> =>
+        isValidElement(child) && typeof child.type !== 'string'
+      );
+
+      const firstNonDisabledOption = Arr.find(validOptions, (option) => !disabled && !option.props.disabled);
+      return firstNonDisabledOption.map((option) => option.props.value).getOrNull();
```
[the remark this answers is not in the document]
kimwoodfield: The type guard is checking that the children are valid React elements, and the component only accepts `SegmentedControl.Option` children which should always have a `value` prop. If someone passes the wrong children then that is a usage error. We shouldn't need to add defensive code for API misuse.

modules/oxide-components/src/main/ts/components/segmentedcontrol/SegmentedControl.tsx
```diff
@@ -109,17 +126,19 @@ const Option = forwardRef<HTMLSpanElement, SegmentedControlOptionProps>((
...
   const isDisabled = groupDisabled || optionDisabled;
+  const isFirstOption = firstOptionValue === optionValue;
 
   const getTabIndex = (): number => {
     if (isDisabled) {
       return -1;
     }
-    return isActive ? 0 : -1;
+    return isFirstOption ? 0 : -1;
   };
```
TheSpyder, line 142: Could the tab index be made a prop instead of adding the first value to the context? The parent code would then be responsible for passing `0` or `-1`. Or if you prefer the boolean `isFirst` could be the prop.

By the same thinking `isActive` should be a prop as well but that's likely a much bigger change...

spocke: I don't think we should ever use `0` that is telling the browser to do it's thing. But maybe we can keep it for now and log a ticket later to remove it since -1 is what we should be using and then manage all keyboard navigation by our selfs at least until we do the big switch away from the tab key.

michalnieruchalski-tiugo: Yeah I'm not sure about that as well. Usually we set tabindex to -1 and handle the behavior with javascript hooks. Which is more flexible, but has its own tradeoffs.  
  
For example, I had to implement useTabStop function that relied heavily on css classes used inside component.

kimwoodfield: I'll keep the current approach for now and I have logged a ticket for us to revisit the tabindex strategy as part of that bigger switch @spocke mentioned: https://ephocks.atlassian.net/browse/TINY-13960

TheSpyder: Classes are how keyboard navigation is managed in Alloy, but there's a system-level handler it doesn't require manual tabstop management in each component.

PR #10950, opened 2026-03-05
TINY-9655: Prevent valid iframe, script, style, and meta elements from being removed by DomPurify

MitchC1999:
Related Ticket: TINY-9655

Description of Changes:
* Script, style, and meta elements would incorrectly be removed by DomPurify when considered valid in the schema.
* Iframe elements with children would incorrectly be removed by DomPurify.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserve valid <script> elements during sanitization instead of stripping their content when allowed by the schema.
  * Keep iframes with valid sources while removing unsafe inner content or scripting from iframe contents.

* **Chores**
  * Added changelog entries documenting these fixes (TINY-9655).

### DIFF
--- a/modules/tinymce/src/core/main/ts/html/Sanitization.ts
+++ b/modules/tinymce/src/core/main/ts/html/Sanitization.ts
@@ -1,11 +1,12 @@
 import { Arr, Fun, Obj, Optional, Strings, Type } from '@ephox/katamari';
-import { Attribute, Html, NodeTypes, Remove, Replication, SugarElement, SugarNode } from '@ephox/sugar';
+import { Attribute, Html, NodeTypes, Remove, Replication, SugarElement } from '@ephox/sugar';
 import createDompurify, { type Config, type DOMPurify, type UponSanitizeAttributeHookEvent, type UponSanitizeElementHookEvent } from 'dompurify';
 
 import type { DomParserSettings } from '../api/html/DomParser';
 import type Schema from '../api/html/Schema';
 import Tools from '../api/util/Tools';
 import * as URI from '../api/util/URI';
+import * as ElementType from '../dom/ElementType';
 import * as NodeType from '../dom/NodeType';
 
 import * as KeepHtmlComments from './KeepHtmlComments';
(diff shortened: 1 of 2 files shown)


modules/tinymce/src/core/main/ts/html/Sanitization.ts
```diff
@@ -168,6 +183,18 @@ const filterAttributes = (ele: Element, settings: DomParserSettings, schema: Sch
   }
 };
 
+const restoreValidContent = (node: Node) => {
+  // Construct the sugar element wrapper
+  const element = SugarElement.fromDom(node) as SugarElement<Element>;
+
+  if (SugarNode.isTag('script')(element)) {
```
spocke: Add this partially applied function to ElementType there is a warning in that file that says we shouldn't add new things to that and use the schema but that is more for categories of things a script will always be a script you shouldn't be able to configure what element types are scripts in the schema.

modules/tinymce/src/core/test/ts/browser/content/insert/InsertContentTest.ts
```diff
@@ -881,7 +881,7 @@ describe('browser.tinymce.core.content.insert.InsertContentTest', () => {
       const editor = hook.editor();
       editor.setContent('<p>initial</p>');
       TinySelections.setCursor(editor, [ 0 ], 0);
-      editor.insertContent('<!--\ufeff><iframe onload=alert(document.domain)>-></body>-->');
+      editor.insertContent('<!--\ufeff><script onload=alert(document.domain)>-></body>-->');
```
spocke, line 884: So scripts also cause this. Maybe we should have two sets of tests one for iframe and one for script?

MitchC1999: These were swapped to scripts because they cause the same behaviour and now we keep the iframes after they've been sanitized. DomPurify was detecting the body as a child of the iframe and incorrectly removing the whole iframe, rather than just its contents and the onload attribute. These tests were unintentionally reliant on the behaviour this ticket is addressing. Its fine to use scripts here since they're already considered invalid in the default schema.

If this is an issue I'm happy to make changes, but I spent ages down the rabbit hole of this particular test to see what was expected and couldn't find an easy solution, just that we want the content after the zwnbsp to still be sanitized.

modules/tinymce/src/core/main/ts/html/Sanitization.ts
```diff
@@ -55,6 +55,21 @@ const processNode = (node: Node, settings: DomParserSettings, schema: Schema, sc
...
+    const shouldKeepContent = lcTagName === 'script' && schema.isValid('script');
+    if (isHTMLElement && shouldKeepContent) {
+      Attribute.set(element, 'data-mce-tmp', Html.get(element));
+    }
+
+    // TINY-9655: Clear innerHTML of script and iframe tags to prevent DOMPurify from removing them entirely
+    const shouldClearContent = lcTagName === 'iframe' && schema.isValid('iframe');
+    if (isHTMLElement && (shouldKeepContent || shouldClearContent)) {
+      Html.set(element, '');
+    }
```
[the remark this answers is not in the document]
MitchC1999: We don't ever want to restore iframe children, we just want to prevent them from being removed when they have children.

kimwoodfield (approved):

PR #10976, opened 2026-03-11
TINY-14034: Added new context property to sidebar spec

spocke:
Related Ticket: TINY-14034

Description of Changes:
* Added new `context` property to sidebar spec

Pre-checks:
* [x] Changelog entry added

- [ ] Don't merge until we figure out what to do here

Review:
* [x] Milestone set

## Summary by CodeRabbit

* **New Features**
  * Sidebars can include a context property so panels and their toggle buttons enable/disable based on editor context.

* **Tests**
  * Added tests and a helper to verify context-driven enable/disable behavior for sidebar toggle buttons.

* **Documentation**
  * Changelog updated with an entry describing the new sidebar context capability (TINY-14034).

### DIFF
--- a/modules/tinymce/src/themes/silver/main/ts/ui/sidebar/Sidebar.ts
+++ b/modules/tinymce/src/themes/silver/main/ts/ui/sidebar/Sidebar.ts
@@ -28,10 +28,11 @@ const setup = (editor: Editor): void => {
   // Setup each registered sidebar
   Arr.each(Obj.keys(sidebars), (name) => {
     const spec = sidebars[name];
+    const bridged: BridgeSidebar.Sidebar = StructureSchema.getOrDie(BridgeSidebar.createSidebar(spec));
     const isActive = () => Optionals.is(Optional.from(editor.queryCommandValue('ToggleSidebar')), name);
     editor.ui.registry.addToggleButton(name, {
-      icon: spec.icon,
-      tooltip: spec.tooltip,
+      ...(bridged.icon.map((icon) => ({ icon })).getOr({})),
+      ...(bridged.tooltip.map((tooltip) => ({ tooltip })).getOr({})),
       onAction: (buttonApi) => {
         editor.execCommand('ToggleSidebar', false, name);
         buttonApi.setActive(isActive());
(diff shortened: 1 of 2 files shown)


modules/bridge/src/main/ts/ephox/bridge/components/sidebar/Sidebar.ts
```diff
@@ -13,6 +13,7 @@ export interface SidebarSpec {
   onShow?: (api: SidebarInstanceApi) => void;
   onSetup?: (api: SidebarInstanceApi) => (api: SidebarInstanceApi) => void;
   onHide?: (api: SidebarInstanceApi) => void;
+  context?: string;
```
TheSpyder, line 16: The `context` property in other parts of the UI defines when a button can be enabled. This seems more like an "applies context" feature?

I don't want to rush this, if we aren't using it in 8.4 the next minor release won't be too far away.

spocke: This is the exact same thing. All sidebars register buttons this context controls the disabled state of that button like all the other buttons.

Peiole have asked for the button to be disabled while you are doing a review we can't disable the button if there is no api for that.

TheSpyder: This again? You can't stop sidebars from hiding. There's an API to swap sidebars, and if another plugin has a sidebar registered the button to activate that is going to hide the current one. End of story.

spocke: When the UI is in readonly mode you can. All buttons are gray in the toolbar except the review button or the chat button depending on what you actioned.

spocke: But I pulled this out of the 8.4 now but then we need to tell people why we can't disable all toolbar buttons. I think this sidebar API is poorly designed. I think we should separate the actual button part from the sidebar part.

So something like this in the spec:
```
editor.ui.registry.addSidebar('foo', {
  toolbarButton:  {
     tooltip: 'Toggle foo'
     context: 'some'
  },
  onShow: () => ...
  onHide: () => ...
})
```

Becausse if we want toggling of sidebars though other methods like a vertical button bar then this toolbar should be distinct from that. I think this was some form of quick fix in the apollo project since we never implemented those sidebar toggle buttons v4 had.

PR #10978, opened 2026-03-11
TINY-13913: Added minimum line height.

HAFRMO:
Related Ticket: TINY-13913

Description of Changes:
Added a line height for the line with the icon on it.

Pre-checks:
* ~~[ ]~~ Changelog entry added

Review:
* [x] Milestone set

## Summary by CodeRabbit

* **Style**
  * Standardized AI chat response status icon sizing by introducing a new configurable height variable with a private fallback, replacing fixed sizing.
  * Applied the new sizing to both icon height and line-height to ensure consistent visual alignment, spacing, and responsive behavior across the AI chat interface.

### DIFF
--- a/modules/oxide/src/less/theme/components/ai-chat/ai-chat.less
+++ b/modules/oxide/src/less/theme/components/ai-chat/ai-chat.less
@@ -12,7 +12,7 @@
 }
 
 .tox .tox-ai {
-  --tox-private-tinymceai-response-status-icon-height: max(24px, calc(var(--tox-private-font-size-sm) + 4px))
+  --tox-private-tinymceai-response-status-icon-height: max(24px, calc(var(--tox-private-font-size-sm) + 4px));
 
   .tox-expandable-box {
     &:focus-visible:not(:disabled) {


modules/oxide/src/less/theme/components/ai-chat/ai-chat.less
```diff
@@ -152,8 +153,8 @@
     display: flex;
     color: var(--tox-private-text-color-muted,@text-color-muted);
     font-size: var(--tox-private-font-size-sm, @font-size-sm);
-    line-height: calc(var(--tox-private-font-size-sm, @font-size-sm) + 4px);
     gap: var(--tox-private-pad-xs, @pad-xs);
+    line-height: var(--response-status-icon-height);
```
spocke: This needs to use both the less variable and the actual variable follow the same fallback pattern we have for other things. Right @ztomaszyk

HAFRMO: Isn't that what the assignment of the variable is doing? Or were you thinking of something else?

spocke: Compare with the other lines in the file.

modules/oxide/src/less/theme/components/ai-chat/ai-chat.less
```diff
@@ -11,6 +11,7 @@
 }
 
 .tox .tox-ai {
+  --response-status-icon-height: max(24px, calc(var(--tox-private-font-size-sm, @font-size-sm) + 4px));
```
spocke: This uses a mixture of less vairables and css custom properties. We should only use the less variables then optionally opt into the custom properties. The name of this variable should pre prefixed the same as the other ones.

So at the top of file something liket his:

```
@tinymceai-response-status-icon-height: max(24px, calc(var(@font-size-sm, @font-size-sm) + 4px))
```

Then in the `.tox-ai` maybe use the same calculation but this time with variables:

```
--tox-private-tinymceai-response-status-icon-height: max(24px, calc(var(--tox-private-font-size-sm) + 4px))
```

Then at the call site it would be use the CSS custom property first then fallback to the LESS variable:

```
line-height: var(--tox-private-tinymceai-response-status-icon-height, @tinymceai-response-status-icon-height);
```

But I think @ztomaszyk is more of an expert here but the general idea is that we use CSS custom properties first then fallback to less variables when the custom property doesn't exist and define both in that optional transitional thing.

modules/oxide/src/less/theme/components/ai-chat/ai-chat.less
```diff
@@ -1,6 +1,7 @@
 @tinymceai-border-width: 1px;
 @tinymceai-focus-border-width: 2px;
 @tinymceai-context-more-max-width: 320px;
+@tinymceai-response-status-icon-height: max(24px, calc(var(@font-size-sm, @font-size-sm) + 4px));
```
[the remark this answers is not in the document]
hamza0867: @HAFRMO you didn’t fix this issue before merging

PR #11021, opened 2026-03-23
TINY-13949: fix focus lost on tag delete

ztomaszyk:
Related Ticket: TINY-13949

Description of Changes:

- Changed Tag component keyboard handler from `onKeyUp` to `onKeyDown` for Backspace/Delete key handling
- Added `requestAnimationFrame` delay to tag removal to allow focus navigation to complete first
- Added BACKSPACE and DELETE key constants to the Keys module
- Enhanced TabbingType with `focusPreviousOnDelete` configuration option to enable automatic focus movement when delete keys are pressed on matching elements
- Modified keydown rules to conditionally include delete key handling based on the new configuration

Pre-checks:

- [x] ~~Changelog entry added~~ (in related PR)
- [x] Tests have been added (if applicable) (in related PR)

Review:

- [x] Milestone set

## Summary by CodeRabbit

- **New Features**
    - Closable tags can be dismissed with Backspace or Delete (now fires on key-down for more responsive keyboard interaction)
    - Added explicit Backspace/Delete keyboard recognition for navigation
    - Tab-key navigation API now exposes a stable "goBackwards" action to programmatically move focus backward for keyboard users

### DIFF
--- a/modules/oxide-components/src/main/ts/bespoke/tinymceai/tag/Tag.tsx
+++ b/modules/oxide-components/src/main/ts/bespoke/tinymceai/tag/Tag.tsx
@@ -48,7 +48,7 @@ export const Tag = forwardRef<HTMLDivElement | HTMLAnchorElement, TagProps>((pro
   const focusable = Type.isNullable(focusableProp) || closeable ? true : props.focusable;
   const sharedAttrs = {
     className: Bem.block('tox-tag'),
-    onKeyUp: (e: React.KeyboardEvent<HTMLDivElement | HTMLAnchorElement>) => {
+    onKeyDown: (e: React.KeyboardEvent<HTMLDivElement | HTMLAnchorElement>) => {
       if (closeable && [ 'Backspace', 'Delete' ].includes(e.key) && !disabled) {
         props.onClose();
       }
--- a/modules/oxide-components/src/main/ts/keynav/KeyboardNavigationHooks.ts
+++ b/modules/oxide-components/src/main/ts/keynav/KeyboardNavigationHooks.ts
@@ -1,6 +1,6 @@
 import { Fun } from '@ephox/katamari';
 import { SugarElement } from '@ephox/sugar';
-import { useEffect, type RefObject } from 'react';
+import { useCallback, useEffect, useRef, type RefObject } from 'react';
 
 import * as EscapingType from './keyboard/escaping/EscapingType';
 import * as ExecutingType from './keyboard/execution/ExecutionType';
@@ -28,17 +28,28 @@ const bindEvents = (container: HTMLElement, handlers: KeyingType.Handlers) => {
 
 export interface TabKeyingProps extends BaseProps, TabbingType.TabbingConfig { }
 
-export const useTabKeyNavigation = (props: TabKeyingProps): void => {
+export interface TabKeyNavigationApi {
+  readonly goBackwards: () => void;
+}
+
+export const useTabKeyNavigation = (props: TabKeyingProps): TabKeyNavigationApi => {
+  const goBackwardsRef = useRef<() => void>(Fun.noop);
+
   useEffect(() => {
     const { containerRef } = props;
 
     if (containerRef.current) {
       const handlers = TabbingType.create(SugarElement.fromDom(containerRef.current), props);
+      goBackwardsRef.current = handlers.goBackwards;
       return bindEvents(containerRef.current, handlers);
     } else {
+      goBackwardsRef.current = Fun.noop;
       return Fun.noop;
     }
   }, [ props ]);
+
+  const goBackwardsCallback = useCallback(() => goBackwardsRef.current(), []);
+  return { goBackwards: goBackwardsCallback };
 };
 
 export interface FlowKeyingProps extends BaseProps, FlowType.FlowConfig { }
--- a/modules/oxide-components/src/main/ts/keynav/keyboard/TabbingType.ts
+++ b/modules/oxide-components/src/main/ts/keynav/keyboard/TabbingType.ts
@@ -3,7 +3,8 @@ import { Compare, Height, SelectorFilter, SelectorFind, type SugarElement, Trave
 
 import * as ArrNavigation from '../navigation/ArrNavigation';
 
-import type { GeneralKeyingConfig } from './KeyingModeTypes';
+import * as FocusManagers from './FocusManagers';
+import { FocusInsideModes, type GeneralKeyingConfig } from './KeyingModeTypes';
 import * as KeyingType from './KeyingType';
 import * as KeyMatch from './KeyMatch';
 import * as KeyRules from './KeyRules';
@@ -119,7 +120,11 @@ const getKeyupRules = Fun.constant([
   KeyRules.rule(KeyMatch.inSet(Keys.TAB), goFromPseudoTabstop),
 ]);
 
-const create = (source: SugarElement<HTMLElement>, config: TabbingConfig): KeyingType.Handlers => {
+export interface TabbingHandlers extends KeyingType.Handlers {
+  readonly goBackwards: () => void;
+}
+
+const create = (source: SugarElement<HTMLElement>, config: TabbingConfig): TabbingHandlers => {
   const partialConfig: Required<TabbingConfig> = {
     execute: Fun.constant(Optional.none()),
     escape: Fun.constant(Optional.none()),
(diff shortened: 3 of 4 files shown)


zuzanna-antas:
* **#11021** <a href="https://app.graphite.com/github/pr/tinymce/tinymce/11021?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/tinymce/tinymce/11021?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

modules/oxide-components/src/main/ts/keynav/keyboard/TabbingType.ts
```diff
@@ -20,6 +20,7 @@ export interface TabbingConfig {
   readonly cyclic?: boolean;
   readonly focusIn?: boolean;
   readonly closest?: boolean;
+  readonly focusPreviousOnDelete?: string;
```
hamza0867: I don’t think this is how it should be done. Moving the focus from the current tag to the previous one needs to be handled by the plugin, not the TabbingType keyboard nav hook

hamza0867 (changes requested):
I don’t believe it is the hook’s responsibility to move the focus after an element is deleted. I understand that we might want to handle the keyboard on keydown instead of up, but it’s the plugin that needs to move the focus when an element is about to be deleted

modules/oxide-components/src/main/ts/bespoke/tinymceai/tag/Tag.tsx
```diff
@@ -48,9 +48,12 @@ export const Tag = forwardRef<HTMLDivElement | HTMLAnchorElement, TagProps>((pro
...
   const sharedAttrs = {
     className: Bem.block('tox-tag'),
-    onKeyUp: (e: React.KeyboardEvent<HTMLDivElement | HTMLAnchorElement>) => {
+    onKeyDown: (e: React.KeyboardEvent<HTMLDivElement | HTMLAnchorElement>) => {
       if (closeable && [ 'Backspace', 'Delete' ].includes(e.key) && !disabled) {
-        props.onClose();
+        // Defer removal so the keynav goBackwards handler can move focus first
+        window.requestAnimationFrame(() => {
+          props.onClose();
+        });
```
TheSpyder: We have too much async in our UI already. Now that 8.4 is done I'm going to start cracking down on this, so you'll have to find another way.

modules/oxide-components/src/main/ts/keynav/keyboard/TabbingType.ts
TheSpyder: Sort of. I agree this setting is the wrong way, but it must remain the responsibility of the focus management code to manage focus.

The plugin should catch the tag close event, remove it from the UI and at that point direct the focus management code to move in the direction it wants - in other words the fix is still within oxide-components, but what we need is a way to call the "go backwards" function.

modules/oxide-components/src/main/ts/keynav/keyboard/TabbingType.ts
```diff
@@ -107,13 +108,21 @@ const execute = (component: SugarElement<HTMLElement>, _event: Event, tabbingCon
...
   KeyRules.rule(KeyMatch.inSet(Keys.TAB), goForwards),
   KeyRules.rule(KeyMatch.inSet(Keys.ENTER), execute),
   KeyRules.rule(KeyMatch.inSet(Keys.SPACE), execute),
   KeyRules.rule(KeyMatch.inSet(Keys.ESCAPE), escape),
-]);
+  ...(config.focusPreviousOnDelete ? [
+    KeyRules.rule(KeyMatch.and([
+      KeyMatch.inSet(DELETE_OR_BACKSPACE),
+      (event) => event.target instanceof window.HTMLElement && event.target.matches(config.focusPreviousOnDelete)
+    ]), goBackwards),
```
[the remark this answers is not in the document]
TheSpyder: We don't support running our UI in an iframe separate to the editor.

TheSpyder (changes requested):

modules/oxide-components/src/main/ts/keynav/keyboard/TabbingType.ts
hamza0867: My bad, that's what I meant, I just didn't word it properly. The only thing I would add is that we want to move the focus before we remove the element from the DOM, so after the click on remove but before removing the element from the DOM so that we don't end up with focus moving to body at some point

TheSpyder: We'd move the focus in the click event and let React handle removing the tag from the DOM.

modules/oxide-components/src/main/ts/bespoke/tinymceai/tag/Tag.tsx
spocke: We should just need the call to happen on keydown then do all management of this in the plugin. It's a plugin concern to deal with what the delete key does in this case. Moving focus is a plugin concern. So in the plugin on this even you move focus to where it need to go using a state in zustand store or directly by a dom ref if the element doesn't re-render.

ztomaszyk:
@TheSpyder & @hamza0867, I changed the approach and updated related PR. You can take another look, thanks for review!

modules/oxide-components/src/main/ts/keynav/KeyboardNavigationHooks.ts
```diff
@@ -28,17 +28,28 @@ const bindEvents = (container: HTMLElement, handlers: KeyingType.Handlers) => {
...
       const handlers = TabbingType.create(SugarElement.fromDom(containerRef.current), props);
+      goBackwardsRef.current = handlers.goBackwards;
       return bindEvents(containerRef.current, handlers);
     } else {
+      goBackwardsRef.current = Fun.noop;
       return Fun.noop;
     }
   }, [ props ]);
+
+  const goBackwardsCallback = useCallback(() => goBackwardsRef.current(), []);
```
hamza0867, line 51: You already have the ref that you can return,  why use the `useCallback` here ? especially with the empty deps array, it seems to me this will be stale once the props change (admittedly they shouldn't change often).  
  
What do you think @coderabbitio

modules/oxide-components/src/main/ts/keynav/keyboard/TabbingType.ts
```diff
@@ -107,19 +108,23 @@ const execute = (component: SugarElement<HTMLElement>, _event: Event, tabbingCon
...
 
-const getKeydownRules = Fun.constant([
+const getKeydownRules = (_component: SugarElement<HTMLElement>, _event: KeyboardEvent, _config: FullTabbingConfig) => [
   KeyRules.rule(KeyMatch.and([ KeyMatch.isShift, KeyMatch.inSet(Keys.TAB) ]), goBackwards),
   KeyRules.rule(KeyMatch.inSet(Keys.TAB), goForwards),
   KeyRules.rule(KeyMatch.inSet(Keys.ENTER), execute),
   KeyRules.rule(KeyMatch.inSet(Keys.SPACE), execute),
   KeyRules.rule(KeyMatch.inSet(Keys.ESCAPE), escape),
-]);
+];
```
hamza0867: This is equivalent to the version from before the change

modules/oxide-components/src/main/ts/keynav/KeyboardNavigationHooks.ts
```diff
@@ -28,17 +28,28 @@ const bindEvents = (container: HTMLElement, handlers: KeyingType.Handlers) => {
 
 export interface TabKeyingProps extends BaseProps, TabbingType.TabbingConfig { }
 
-export const useTabKeyNavigation = (props: TabKeyingProps): void => {
+export interface TabKeyNavigationApi {
+  readonly goBackwards: () => void;
```
hamza0867, line 32: If we're adding this, then we would have to add other apis as well, like goForwards, and then we would need to do the same for other keyboard navigation types (flow type, menu type ...etc)

hamza0867 (approved):
I think a simpler and more coherent approach would have been to do this completely in the plugin, move the focus in the onRemove callback. Now, to be consistent we would have to add the same api to other keyboard nav types.

modules/oxide-components/src/main/ts/keynav/KeyboardNavigationHooks.ts
zuzanna-antas: Created ticket https://ephocks.atlassian.net/browse/TINY-14184

modules/oxide-components/src/main/ts/keynav/keyboard/TabbingType.ts
zuzanna-antas: Went with this approach and created a ticket to expose goForward as well and add the APIs to other hooks -  https://ephocks.atlassian.net/browse/TINY-14184

> The plugin should catch the tag close event, remove it from the UI and at that point direct the focus management code to move in the direction it wants - in other words the fix is still within oxide-components, but what we need is a way to call the "go backwards" function.

spocke (approved):

modules/oxide-components/src/main/ts/keynav/KeyboardNavigationHooks.ts
TheSpyder: React, often making hard things easy but always making easy things complicated.

TheSpyder (approved):
The `goBackwards` implementation should absolutely be in the specific keyboard type, because what backwards means depends on how the focus flows.

The issues with how we call `goBackwards` seems like more of an architectural issue with our choice of focus management.

This is still an improvement.

PR #11027, opened 2026-03-24
TINY-14070: Prevent lists from detecting hosts too early.

HAFRMO:
Related Ticket: TINY-14070

Description of Changes:
The list detection would prematurely detect divs in a list-item as a list host rather than being in a list. Added a check that if it is in a list find said list, unless we're in a table inside of that list.

This issue seems to have existed for a while but was made easier to run into with the weakening of the forcedrootblock-check.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed cases where certain combinations of divs inside lists caused unexpected behavior when disabling lists, improving list toggle reliability.

* **Tests**
  * Added browser tests validating cursor and range scenarios to ensure list items are correctly removed or preserved when lists are toggled.


modules/tinymce/src/core/main/ts/lists/lists/Selection.ts
```diff
@@ -62,10 +63,31 @@ const getClosestEditingHost = (editor: Editor, elm: Element): HTMLElement => {
 const isListHost = (schema: Schema, node: Node): boolean =>
   !NodeType.isListNode(node) && !NodeType.isListItemNode(node) && Arr.exists(listNames, (listName) => schema.isValidChild(node.nodeName, listName));
 
+const requireLiElementFirst = (parentBlocks: HTMLElement[]): boolean => {
+  for (let index = 0; index < parentBlocks.length; index++) {
```
spocke: Can be done with Arr.findMap to avoid the imperative loop.

modules/tinymce/src/core/main/ts/lists/lists/Selection.ts
```diff
@@ -62,10 +63,31 @@ const getClosestEditingHost = (editor: Editor, elm: Element): HTMLElement => {
...
+  return false;
+};
+
 const getClosestListHost = (editor: Editor, elm: Node, isCollapsed: boolean): HTMLElement => {
   const parentBlocks = editor.dom.getParents<HTMLElement>(elm, editor.dom.isBlock);
+  let foundListBlock = !requireLiElementFirst(parentBlocks);
   const isNotForcedRootBlock = (elm: HTMLElement) => elm.nodeName.toLowerCase() !== Options.getForcedRootBlock(editor);
-  const parentBlock = Arr.find(parentBlocks, (elm) => (!isCollapsed || isNotForcedRootBlock(elm)) && isListHost(editor.schema, elm));
+  const checkListRequirement = (element: HTMLElement) => {
+    foundListBlock = foundListBlock || isListItem(SugarElement.fromDom(element)) || isList(SugarElement.fromDom(element));
```
spocke: Why locally mutating this just use return here instead?

.changes/unreleased/tinymce-TINY-14070-2026-03-24.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Improved
```
spocke: Isn't this a bug?

modules/tinymce/src/core/main/ts/lists/lists/Selection.ts
```diff
@@ -62,10 +63,31 @@ const getClosestEditingHost = (editor: Editor, elm: Element): HTMLElement => {
...
   !NodeType.isListNode(node) && !NodeType.isListItemNode(node) && Arr.exists(listNames, (listName) => schema.isValidChild(node.nodeName, listName));
 
+const requireLiElementFirst = (parentBlocks: HTMLElement[]): boolean => {
+  for (let index = 0; index < parentBlocks.length; index++) {
+    const element = parentBlocks[index];
+    if (isListItem(SugarElement.fromDom(element))) {
+      return true;
+    }
+
+    if (NodeType.isTableCellNode(element)) {
```
spocke, line 72: The tests has nothing with tables in them is this covered by some other test?

spocke (approved):

modules/tinymce/src/core/main/ts/lists/lists/Selection.ts
HAFRMO: Because if you had the setup `list -> table -> cursor` it'd go out of the table, which was detected by already existing tests.

modules/tinymce/src/core/main/ts/lists/lists/Selection.ts
HAFRMO: Because once we have found the list any subsequent entries should be considered valid ( for this particular check ) so we need to remember that.

modules/tinymce/src/core/main/ts/lists/lists/Selection.ts
```diff
@@ -62,10 +63,32 @@ const getClosestEditingHost = (editor: Editor, elm: Element): HTMLElement => {
...
+      return Optional.some(false);
+    }
+
+    return Optional.none();
+  });
+
+  return result.getOr(false);
+};
```
lorenzo-pomili, line 80:
```suggestion
const requireLiElementFirst = (parentBlocks: HTMLElement[]): boolean => 
  Arr.exists(parentBlocks, (element) => isListItem(SugarElement.fromDom(element)));
```

maybe I'm missing something but this couldn't be wrote in this form?

lorenzo-pomili (approved):

modules/tinymce/src/core/main/ts/lists/lists/Selection.ts
HAFRMO: No, that causes trouble when dealing with tables, unfortunately ( I also thought I could do that one at first ). Basically what we want to find is 'is there a list item as a parent, unless we're in a table in that list element'.
So if we have something like `div > list` or `div > list > div > div > div` we want to find the list element, but if it is `div > list > table > div` we don't want to.

modules/tinymce/src/core/main/ts/lists/lists/Selection.ts
TheSpyder: Can you at least make it a bit more explicit, instead of a one liner. Something like

```
if (Type.isUndefined(foundListBlock)) {
// set foundListBlock
}
return foundListBlock
```

TheSpyder (approved):

PR #11035, opened 2026-03-26
TINY-14140: Update release libs script to handle updating private module dependencies

ltrouton:
Related Ticket: TINY-14140

Description of Changes:

After running `release-libs.sh` to bump library versions, `yarn` fails in the workspace. The cause is that **tinymce's** `package.json` **dependency ranges are not updated to reflect the new library versions.** For example, if `@ephox/agar` is bumped from `9.x` to `10.0.0`, tinymce's `package.json` still declares `"@ephox/agar": "^9.0.0"`. In a yarn workspace, `10.0.0` does not satisfy `^9.0.0`, so yarn cannot resolve the dependency and fails. This is because we have `private: false` set in for the version section in our lerna.json

**Pre-April 2024:** lerna.json had no `private: false`. When `yarn lerna version` ran, it included all packages - including tinymce.

**April 2024 (**TINY-10602 **\- "Cleaning up tech debt"):** `private: false` was added to lerna.json's `command.version` section to "Disable lerna version on TinyMCE". However, it has an unintended consequence: `private: false` maps to lerna's `--no-private` CLI flag, which **excludes private packages entirely** - not just from versioning, but from all processing including dependency range updates.

Since we want to keep `private: false` so we don't bump and tag the private modules, I have added a  cript that updates the private modules dependencies section to be correct. What the script does can be see here: https://github.com/tinymce/tinymce/commit/89dff305c002fa38b69d8ad158dcf54e89593cee. I have also tested against the premium branch with the version updates to ensure nothing breaks

Worth noting that this problem would be **entirely eliminated** by migrating from Yarn Classic (v1) to a package manager that supports the `workspace:` protocol: Yarn Berry (v2+), pnpm, or Bun.

After merging this, I should have everything in order to version and tag the libraries for the release.

Pre-checks:

- [x] Changelog entry added

Review:

- [x] Milestone set


ltrouton:
* **#11035** <a href="https://app.graphite.com/github/pr/tinymce/tinymce/11035?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/tinymce/tinymce/11035?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `release/8`


<h2></h2>

bin/update-private-deps.sh
spocke, line 1: When bash scripts start becoming this complex one should consider using typescript. Benefits that you can do way more complex things in TS more easily, you have proper type checking, it's a proper language etc etc. Maybe when we switch to bun we can more easily write these types of things in TS instead.

spocke (approved):

bin/update-private-deps.sh
ltrouton: Agreed. A little while a go I created a tech debt ticket to convert some of our shell scripts to typescript. [https://ephocks.atlassian.net/browse/TINY-13966](https://ephocks.atlassian.net/browse/TINY-13966)

I think `zx` could be a good tool to use as well the helps with scripting with typescript

kimwoodfield (approved):

shanmen-tiny (approved):

PR #11041, opened 2026-03-27
TINY-14072: Tooltip overflow detection

HAFRMO:
Related Ticket: TINY-14072

Description of Changes:
Check if the content of the tooltip trigger is overflowing. Only if it is show the tooltip, given the right setting ( defaults to always showing )

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

## Summary by CodeRabbit

* **New Features**
  * Added optional showCondition for tooltip triggers with values "always" (default) and "overflow" to control when tooltip content is rendered; triggers now detect overflow and avoid attaching interactive handlers when rendering is suppressed.

* **Documentation**
  * Storybook updated with new args and an "OverflowingTooltip" example demonstrating overflow-driven tooltip behavior.

* **Chores**
  * Added unreleased changelog entry documenting the change.


modules/oxide-components/src/main/ts/components/tooltip/Tooltip.tsx
```diff
@@ -1,31 +1,72 @@
...
+  setRenderComponents: Fun.noop,
   contentRef: { current: null },
   triggerRef: { current: null },
 };
 
 const TooltipContext = createContext<TooltipState>(defaultState);
 
-interface TriggerInternalProps extends PropsWithChildren<HTMLAttributes<HTMLElement>> {}
+interface TriggerSpecificProps {
+  readonly showCondition?: 'always' | 'overflow';
```
hamza0867, line 16: 1 . This should rather be a Root level prop, not a Trigger prop.   
2\. It should also be a public prop not an internal prop

modules/oxide-components/src/main/ts/components/tooltip/Tooltip.tsx
```diff
@@ -1,31 +1,72 @@
...
+interface TriggerInternalProps extends PropsWithChildren<HTMLAttributes<HTMLElement>>, TriggerSpecificProps { }
+
+const TriggerImpl = forwardRef<HTMLElement, TriggerInternalProps>(({ children, showCondition, ...props }, ref) => {
+  const { renderComponents, setIsOpen, triggerRef, setRenderComponents } = useContext(TooltipContext);
+
+  const shouldRender = () => {
+    if (showCondition === 'overflow') {
+      const trigger = triggerRef.current;
+
+      return !!(trigger && (trigger.offsetHeight < trigger.scrollHeight || trigger.offsetWidth < trigger.scrollWidth));
```
hamza0867: We should have a separate function that takes a DOM element and returns if the text content overflows. Also shouldn't we only compare the widths ? The height is not taken into when deciding if text overflows and will need an ellipsis as far as I can tell (keep in mind, we are not using the line-clamp property).

modules/oxide-components/src/main/ts/components/tooltip/Tooltip.tsx
```diff
@@ -1,31 +1,72 @@
...
+      }
+      setRenderComponents(rerender);
+    };
+
+    window.addEventListener('resize', handleResize);
+
+    return () => {
+      window.removeEventListener('resize', handleResize);
+    };
+  });
```
hamza0867: We need dependency array for these effects, we don't want to run this all the time

modules/oxide-components/src/main/ts/components/tooltip/Tooltip.stories.tsx
```diff
@@ -15,8 +15,8 @@ const meta = {
   title: 'components/Tooltip',
   component: (props) =>
     <Tooltip.Root>
-      <Tooltip.Trigger>
-        <div title='hover' style={{ border: '1px solid #000' }}>Hover Me</div>
+      <Tooltip.Trigger showCondition={props.alwaysShow ? 'always' : 'overflow'}>
+        <div title='hover' style={{ border: '1px solid #000', maxWidth: '200px', maxHeight: '200px', overflow: 'hidden' }}>{props.oversizeContent ? 'Hover Me, but Big'.repeat(50) : 'Hover Me'}</div>
```
hamza0867: That's not really what we want to do ? We really want to detect if the `text-overflow: ellipsis` rule will be triggered, not the `overflow: hidden` rule.

hamza0867 (changes requested):
On top of the comments I left, we are also missing tests. Please add tests.

modules/oxide-components/src/main/ts/components/tooltip/Tooltip.tsx
HAFRMO: I was under the impression that effects only run once per element if they don't have dependency arrays, rather than run every time?

hamza0867: https://react.dev/reference/react/useEffect#useeffect

modules/oxide-components/src/main/ts/components/tooltip/Tooltip.tsx
```diff
@@ -1,31 +1,73 @@
...
 
 const TriggerImpl = forwardRef<HTMLElement, TriggerInternalProps>(({ children, ...props }, ref) => {
-  const { setIsOpen, triggerRef } = useContext(TooltipContext);
+  const { renderComponents, setIsOpen, showCondition, triggerRef, setRenderComponents } = useContext(TooltipContext);
+
+  const shouldRender = () => {
+    if (showCondition === 'overflow') {
+      const trigger = triggerRef.current;
+
+      return !!(trigger && isOverflowing(trigger));
```
spocke: Use `Type.isNonNullable` here instead of !!.

modules/oxide-components/src/main/ts/components/tooltip/Tooltip.tsx
spocke: Why wasn't this a lint failure?

modules/oxide-components/src/main/ts/components/tooltip/Tooltip.tsx
```diff
@@ -148,9 +203,14 @@ const Content = forwardRef<HTMLDivElement, ContentProps>(({ text }, ref) => {
   );
 });
 
-const Root: FC<PropsWithChildren> = ({ children }) => {
+interface RootProps extends PropsWithChildren {
```
spocke: Type should be at top of file ideally.

modules/oxide-components/src/main/ts/components/tooltip/Tooltip.tsx
HAFRMO: Ah, I mixed up 'no array' and 'empty array', thank you for correcting it.

modules/oxide-components/src/main/ts/components/tooltip/Tooltip.tsx
```diff
@@ -1,35 +1,81 @@
...
+
+      return Type.isNonNullable(trigger) && isOverflowing(trigger);
+    }
+
+    return true;
+  };
+
+  useLayoutEffect(() => {
+    const rerender = shouldRender();
+    if (renderComponents === rerender) {
```
hamza0867: Are these booleans ? They don't look like it from their names

HAFRMO: Yes, they are. And what would you prefer in terms of naming?

hamza0867: `@coderabbitai` what do you suggest ?

.changes/unreleased/tinymce-TINY-14072-2026-03-18.yaml
hamza0867, line 1: Are you sure we need this file ? This library is not released and the change is not reflected in the tiny project. I don’t think we need to have a changelog. What do you think @spocke ?

spocke: The changelog should go into the premium repo if it only affects the ai plugin. Also it should be more general that this, these properties are not options for the editor it would make no sense to anyone outside the company consuming this changelog.

The changelogs on the product parts is not for us it's for the integrators. The changelogs on libraries are for us internally and can have this level of detail. We don't have a changelog for oxide-components not sure we need one to be honest since it's not consumed outside the monorepo.

So please remove this and make a separate PR with just the change that the user would see for the `tinymceai` plugin in the premium repo.

modules/oxide-components/src/main/ts/components/tooltip/Tooltip.tsx
```diff
@@ -1,19 +1,87 @@
...
+  delayForHide: 100,
+  setIsOpen: Fun.noop,
+  showCondition: 'always',
+  setRenderComponents: Fun.noop,
+  contentRef: { current: null },
+  triggerRef: { current: null },
+  popupAnchor: '--tooltip-anchor' // this is wrong, but I don't have time to rewrite context today
+};
+
+const TooltipContext = createContext<TooltipState>(defaultState);
```
hamza0867: Why did you move this from the context file ?

HAFRMO: Didn't move in from there, it was here initially and then a merge returned them. Moved them back out.

spocke (approved):

spocke:
This would need a separate AI changelog PR to explain what the new behavior is here.

hamza0867 (approved):

modules/oxide-components/src/main/ts/components/tooltip/internals/Context.tsx
```diff
@@ -2,11 +2,14 @@ import { createContext, useContext } from 'react';
 
 export interface TooltipState {
   readonly isOpen: boolean;
+  readonly shouldRenderComponents: boolean;
```
hamza0867, line 5: I wouldn't call this a state. This might be part of the context but not the state

modules/oxide-components/src/main/ts/components/tooltip/internals/Context.tsx
```diff
@@ -2,11 +2,14 @@ import { createContext, useContext } from 'react';
...
   readonly delayForShow: number;
   readonly delayForHide: number;
   readonly setIsOpen: (isOpen: boolean) => void;
+  readonly setRenderComponents: (isOpen: boolean) => void;
```
hamza0867, line 9: Same for these two

modules/oxide-components/src/main/ts/components/tooltip/internals/Context.tsx
HAFRMO: So where do you want to put it? This is what the context consists of.

hamza0867: I would say we should then rename the interface to reflect this meaning from State to Context

modules/oxide-components/src/main/ts/components/tooltip/Tooltip.tsx
```diff
@@ -1,19 +1,72 @@
...
 
 import { useTooltip, TooltipContext } from './internals/Context';
 
-interface TriggerInternalProps extends PropsWithChildren<HTMLAttributes<HTMLElement>> {}
+interface RootProps extends PropsWithChildren {
+  readonly showCondition?: 'always' | 'overflow';
+}
+
+interface TriggerProps {
+  readonly checkChildren?: boolean;
```
hamza0867, line 20: I don't think this is needed. At least it should be true by default.

PR #11042, opened 2026-03-27
TINY-14149: Prevent deleting the editor body.

HAFRMO:
Related Ticket: TINY-14149

Description of Changes:
Some situations could accidentally delete the whole body for the editor.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

## Bug Fixes
- Fixed an issue where certain selections would delete the editor body, preventing normal editor operations.

### DIFF
--- a/modules/tinymce/src/core/main/ts/delete/DeleteElement.ts
+++ b/modules/tinymce/src/core/main/ts/delete/DeleteElement.ts
@@ -142,10 +142,8 @@ const deleteElement = (
 ): void => {
   // Existing delete logic
   const afterDeletePos = findCaretPosOutsideElmAfterDelete(forward, editor.getBody(), elm.dom);
-
-  const isOutside = NodeType.matchNodeNames([ 'body', 'html' ])(elm.dom);
   const parentBlock = PredicateFind.ancestor(elm, Fun.curry(isBlock, editor), eqRawNode(editor.getBody()));
-  const normalizedAfterDeletePos = (isOutside ? Traverse.firstChild(elm) : Optional.from(elm)).bind((child) => deleteNormalized(child, afterDeletePos, editor.schema, isInlineElement(editor, elm)));
+  const normalizedAfterDeletePos = deleteNormalized(elm, afterDeletePos, editor.schema, isInlineElement(editor, elm));
 
   if (editor.dom.isEmpty(editor.getBody())) {
     editor.setContent('');
--- a/modules/tinymce/src/core/main/ts/delete/DeleteElement.ts
+++ b/modules/tinymce/src/core/main/ts/delete/DeleteElement.ts
@@ -142,8 +142,10 @@ const deleteElement = (
 ): void => {
   // Existing delete logic
   const afterDeletePos = findCaretPosOutsideElmAfterDelete(forward, editor.getBody(), elm.dom);
+
+  const isOutside = NodeType.matchNodeNames([ 'body', 'html' ])(elm.dom);
   const parentBlock = PredicateFind.ancestor(elm, Fun.curry(isBlock, editor), eqRawNode(editor.getBody()));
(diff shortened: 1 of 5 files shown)


modules/tinymce/src/core/main/ts/delete/DeleteElement.ts
```diff
@@ -142,8 +142,10 @@ const deleteElement = (
 ): void => {
   // Existing delete logic
   const afterDeletePos = findCaretPosOutsideElmAfterDelete(forward, editor.getBody(), elm.dom);
+
+  const isOutside = NodeType.matchNodeNames([ 'body', 'html' ])(elm.dom);
   const parentBlock = PredicateFind.ancestor(elm, Fun.curry(isBlock, editor), eqRawNode(editor.getBody()));
-  const normalizedAfterDeletePos = deleteNormalized(elm, afterDeletePos, editor.schema, isInlineElement(editor, elm));
+  const normalizedAfterDeletePos = (isOutside ? Traverse.firstChild(elm) : Optional.from(elm)).bind((child) => deleteNormalized(child, afterDeletePos, editor.schema, isInlineElement(editor, elm)));
```
spocke: I don't think this code is correct or even needed to change. So please revert.

modules/tinymce/src/core/test/ts/browser/delete/DeleteElementTest.ts
```diff
@@ -239,4 +239,16 @@ describe('browser.tinymce.core.delete.DeleteElementTest', () => {
...
   });
+
+  it('TINY-14149: Delete when the body is selected', () => {
+    const editor: Editor = hook.editor();
+    editor.setContent('<table><tbody><tr><td>Test</td></tr></tbody></table><p></p><p>Content</p>');
+    const range = editor.getDoc().createRange();
+    range.setStart(editor.getBody(), 0);
+    range.setEnd(editor.getBody(), 1);
+    editor.selection.setRng(range);
+    deleteElementPath(editor, false, [ ]);
```
spocke: This path is not correct it should be [ 0 ] if you are to delete the table. Also not sure that this test is even needed since the bug is in SelectedHTMLElementDelete the bug is that we get the parent element when we delete not the selected element and that fix is correct.

spocke (changes requested):

modules/tinymce/src/core/main/ts/delete/DeleteElement.ts
HAFRMO: This code is to ensure that even if we are given the body as an item to delete we don't do that. And it works as intended, as far as I could tell? What did I get wrong?

modules/tinymce/src/core/test/ts/browser/delete/DeleteElementTest.ts
HAFRMO: This is a test to make sure that even if we are given the body as a subject, as the SelectedHTMLElementDelete did do, we don't delete it.

modules/tinymce/src/core/main/ts/delete/DeleteElement.ts
spocke: We should never call it with the body that is the bug and already fixed in the other module. The test is testing a scenario that will never happen.

modules/tinymce/src/core/test/ts/browser/delete/DeleteElementTest.ts
spocke: But you fixed that bug so this will never happen now.

HAFRMO: Yes, but I thought it'd be good to add the check there too, to prevent any other delete-element from accidentally doing the same thing. But if you think that's not needed I can remove it.

spocke (approved):

hamza0867 (approved):

PR #11054, opened 2026-04-02
TINY-9655: Prevent valid iframe, script, and style elements from being removed by DomPurify

MitchC1999:
Related Ticket: [TINY-9655](https://ephocks.atlassian.net/browse/TINY-9655)

Description of Changes:
* Script and style elements would incorrectly be removed by DomPurify when considered valid in the schema.
* Iframe elements with children would incorrectly be removed by DomPurify.
* Reimplement https://github.com/tinymce/tinymce/pull/10950
* Which was reverted in https://github.com/tinymce/tinymce/pull/11038
* Remove `sanitize: false` from tests which were updated when the last dompurify patch was removed
  * One test would now fail from missing the `allow_html_in_comments` option which is only applicable when sanitize is true

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

[TINY-9655]: https://ephocks.atlassian.net/browse/TINY-9655?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


modules/tinymce/src/core/main/ts/html/Sanitization.ts
```diff
@@ -55,6 +56,20 @@ const processNode = (node: Node, settings: DomParserSettings, schema: Schema, sc
...
+    // TINY-9655: Preserve the content of script tags if they are valid elements in the schema
+    const shouldKeepContent = ElementType.isScript(element) && schema.isValid('script');
+    if (shouldKeepContent) {
+      Optional.from(TextContent.get(element)).each((content) => Attribute.set(element, 'data-mce-tmp', content));
+    }
+
+    // TINY-9655: Clear innerHTML of script and iframe tags to prevent DOMPurify from removing them entirely
+    const shouldClearContent = ElementType.isIframe(element) && schema.isValid('iframe');
+    if (shouldKeepContent || shouldClearContent) {
+      Html.set(element, '');
```
spocke: Maybe we should use `Remove.empty` here to capture intent but this is fine as well.

modules/tinymce/src/core/main/ts/html/Sanitization.ts
```diff
@@ -168,6 +183,18 @@ const filterAttributes = (ele: Element, settings: DomParserSettings, schema: Sch
   }
 };
 
+const restoreValidContent = (node: Node) => {
+  // Construct the sugar element wrapper
+  const element = SugarElement.fromDom(node) as SugarElement<Element>;
+
+  if (ElementType.isScript(element)) {
+    Optional.from(Attribute.get(element, 'data-mce-tmp')).each((content) => {
```
spocke: There is a `Attribute.getOpt` bummer that there isn't a `TextContent.getOpt`.

modules/tinymce/src/core/test/ts/browser/dom/SerializerTest.ts
```diff
@@ -434,231 +410,256 @@ describe('browser.tinymce.core.dom.SerializerTest', () => {
     );
   });
 
+  it('Whitespace preserve in pre', () => {
```
spocke: Why was these tests moved?

modules/tinymce/src/core/test/ts/browser/html/SanitizationTest.ts
```diff
@@ -19,7 +19,7 @@ describe('browser.tinymce.core.html.SanitizationTest', () => {
 
     it('Sanitize iframe HTML', () => testHtmlSanitizer({
```
spocke, line 20: Could we add a iframe case here as well where it had content so we are not just using the one in the media plugin if we ever rewrite that plugin. I see there is one for the secuirty issue just a test that has a iframe with say some text in it as a separate test with some valid HTML in it would cover normal use.

spocke (approved):

modules/tinymce/src/core/test/ts/browser/dom/SerializerTest.ts
MitchC1999: Unintentional

kimwoodfield (approved):

PR #11056, opened 2026-04-02
TINY-13938: manage data uri with BOM

lorenzo-pomili:
Related Ticket: TINY-13938

Description of Changes:
the problem was that the string here was prefixed with `%EF%BB%BF` which is Byte order mark ([MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar), [Wikipedia](https://en.wikipedia.org/wiki/Byte_order_mark)), since this seems something that is put only as prefix I checked only if there is at the start of the string

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed errors when processing Data URIs containing non‑Latin1 characters (including UTF‑8 BOM).
* **Tests**
  * Added tests verifying safe handling of Data URIs with BOM and other non‑Latin1 characters.
* **Changelog**
  * Added an unreleased changelog entry recording the fix (TINY-13938).


modules/tinymce/src/core/main/ts/file/BlobCacheUtils.ts
```diff
@@ -10,12 +10,15 @@ const uniqueId = (prefix?: string): string => {
   return (prefix || 'blobid') + (count++);
 };
 
-const processDataUri = (dataUri: string, base64Only: boolean, generateBlobInfo: (base64: string, type: string) => Optional<BlobInfo>): Optional<BlobInfo> => {
+const withoutByteOrderMark = (str: string) =>
```
spocke: I think we should look into using the text encoder API here so that we can get handle higher byte ranges maybe the text encoder API also handles any BOM characters. Could we investigate that, I added another case that fails to the ticket.

lorenzo-pomili: changed

modules/tinymce/src/core/main/ts/file/BlobCacheUtils.ts
```diff
@@ -10,12 +10,14 @@ const uniqueId = (prefix?: string): string => {
   return (prefix || 'blobid') + (count++);
 };
 
-const processDataUri = (dataUri: string, base64Only: boolean, generateBlobInfo: (base64: string, type: string) => Optional<BlobInfo>): Optional<BlobInfo> => {
+const withoutByteOrderMark = (str: string): string => String.fromCharCode(...(new window.TextEncoder().encode(str)));
```
spocke: I'm not sure about this spreading arguments like that might be breaking JS runtimes even. Maybe we need to loop here.

modules/tinymce/src/core/test/ts/atomic/file/BlobCacheUtilsTest.ts
```diff
@@ -0,0 +1,15 @@
...
+
+import { processDataUri } from 'tinymce/core/file/BlobCacheUtils';
+
+describe('atomic.tinymce.core.file.BlobCacheUtilsTest', () => {
+  it('TINY-13938: processDataUri should not throw error with URIs with Byte Order Mark (BOM)', () => {
+    assert.doesNotThrow(() => processDataUri('data:image/svg+xml,%EF%BB%BF%3Csvg', false, Fun.constant(Optional.none())));
+  });
+
+  it('TINY-13938: processDataUri should not throw error with URIs with not Latin1 characters', () => {
+    assert.doesNotThrow(() => processDataUri('data:image/svg+xml,%EF%BB%BF%3Csvg犬', false, Fun.constant(Optional.none())));
```
spocke: Could we also assert the output data that this produces so that this doesn't break things in unexpected ways.

modules/tinymce/src/core/main/ts/file/BlobCacheUtils.ts
spocke: This does more than just remove the BOM so maybe call it `utf8Encode` or something.

spocke (approved):

.changes/unreleased/tinymce-TINY-13938-2026-04-02.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: URIs prefixed with BOM were returning an error.
```
spocke: We need to update this maybe mention that URIs with non ascii range characters was throwing an error.

spocke (approved):

modules/tinymce/src/core/main/ts/file/BlobCacheUtils.ts
lorenzo-pomili: changed to `Arr.foldl` and renamed to `utf8Encode`

.changes/unreleased/tinymce-TINY-13938-2026-04-02.yaml
lorenzo-pomili: changed

modules/tinymce/src/core/test/ts/atomic/file/BlobCacheUtilsTest.ts
lorenzo-pomili: added the assertion

modules/tinymce/src/core/test/ts/atomic/file/BlobCacheUtilsTest.ts
```diff
@@ -0,0 +1,15 @@
...
+  });
```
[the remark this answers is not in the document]
lorenzo-pomili: changed have a check

HAFRMO (approved):

PR #11072, opened 2026-04-16
TINY-7552: Optimise loading all emoticons

MitchC1999:
Related Ticket: TINY-7552

Description of Changes:
* New TabPanel `dynamicHeight` optional property
  * Defaulted to `true` to keep current behaviour in all TabPanel dialogs
  * When set to `false`, height is not dynamically calculated by rendering all tabs before loading
* Removed cap of 300 on the amount of emoticons shown in the emoticons dialog `All` tab
* Set `dynamicHeight` to `false` in the emoticons dialog `TabPanel` component
* Emoticons dialog now loads all emoticons in the 'All' tab without any significant performance issues.
* Performance improvements demonstrated in [this Jira comment](https://ephocks.atlassian.net/browse/TINY-7552?focusedCommentId=172540)

Pre-checks:
* [x] Changelog entry added
* [x] Tests (Not applicable)

Review:
* [x] Milestone set
* [x] Docs ticket created (not applicable)

## Summary by CodeRabbit

* **New Features**
  * Tab panels now support configurable dynamic-height behavior for more flexible, per-tab layouts.

* **Improved**
  * Emoticons dialog’s “All” tab now shows all emoticons (no 300-item cap).
  * Emoticon rendering optimized for smoother, more responsive browsing.


.changes/unreleased/tinymce-TINY-7552-2026-04-16.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Improved
+body: Removed cap on emoticons dialog 'All' tab and optimised rendering performance.
```
TheSpyder:
```suggestion
body: Removed 300 item limit on the emoticons dialog 'All' tab by optimising render performance.
```

modules/tinymce/src/themes/silver/main/ts/ui/dialog/TabPanel.ts
```diff
@@ -93,7 +93,9 @@ export const renderTabPanel = (spec: TabPanelSpec, dialogData: Dialog.DialogData
   });
 
   // Assign fixed height or variable height to the tabs
-  const tabMode = DialogTabHeight.smartMode(allTabs);
+  const tabMode = spec.dynamicHeight
+    ? DialogTabHeight.smartMode(allTabs)
+    : DialogTabHeight.naiveMode(allTabs);
```
TheSpyder: I'm sure you realised but this doesn't set a fixed height, it allows the dialog to change heights for each tab. This is visually quite silly as a user experience with tabs on the side of the dialog, which is why we never used it (and probably should've deleted it as an option). Please check whether swapping to the individual tabs (especially short lists such as "user defined" [in our documentation](https://www.tiny.cloud/docs/tinymce/latest/emoticons/)) maintain the dialog height correctly.

TheSpyder (approved):

modules/tinymce/src/themes/silver/main/ts/ui/dialog/TabPanel.ts
```diff
@@ -93,7 +93,9 @@ export const renderTabPanel = (spec: TabPanelSpec, dialogData: Dialog.DialogData
   });
 
   // Assign fixed height or variable height to the tabs
-  const tabMode = DialogTabHeight.smartMode(allTabs);
+  const tabMode = spec.dynamicHeight
+    ? DialogTabHeight.naiveMode(allTabs) // Height is determined by the content of the tab, which may cause layout shifts when switching tabs.
```
ltrouton, line 97: Will this cap the dialog size at the default max height? Could we force the dialog to just be the fixed default max height so it doesn't change between tabs. It's been a while since I have worked with dialogs so might be asking silly questions but just wanted to check

MitchC1999: I think that would be a better solution. That was my original thought when I saw this function, but it's not built for that. It doesn't affect the emoticons dialog since it always has the same height, but I don't think we'd want any dialogs to change height like this anyway. Since this naive mode was never used, I'll update it to just use the default max height from the initial tab, so it's never recalculated. This won't affect any current behaviour since it will be off by default, and the new property could be something like `staticHeight`, where it only calculates height once from the initial tab.

MitchC1999: Thinking about it more, I'm not sure if we would want to always use the max height. Unless by default max height you mean the max height across all tabs, which is what the smart mode does. If we were to cap it at the max height of the initial tab, it might be weird for the content to then become scrollable when switching to a tab with greater height. Then the height isn't able to be set, and appears different across dialogs.
At least in this case, where the dialog stretches to fit the current tab, it has a consistent behaviour and works for any configuration.

modules/tinymce/src/themes/silver/main/ts/ui/dialog/TabPanel.ts
MitchC1999: We had a discussion in standup and have agreed that using `dynamicHeight: false`, the default behaviour, is enough for customers who don't want their dialogs to change height. This new property is opt-in, and self-explanatory, so any customer using it shouldn't be surprised when the height changes. It was probably the case that this mode was never used because of that behaviour, but as the solution to the emoticons issue, this behaviour is a non-issue as that particular dialog won't resize, even with custom emoticons.

TheSpyder: It's opt-in for custom dialogs, that's fine, but it's not opt-in for emoticons. I was just asking you to manually check emoticons, which it sounds like you did.

MitchC1999: Yea sorry, there were other questions in standup about the effect on custom dialogs. Emoticons is fine with this config.

kimwoodfield (approved):

ltrouton (approved):

Issue #7832, opened 2022-05-17
Background color under style, getting converted to lowercase


bkadwal:
**What is the current behavior? Describe the bug**

Passing the background color as a variable (camel case) under style attribute is converted into lowercase.
 <td class="stack" style="background-color:${backgroundColor}; "> <td> 
resulted in 
 <td class="stack" style="background-color:${backgroundcolor}; "> <td> 
 tinymce is changing the case.
 When this data is saved, it results in no background color.

**What is the expected behavior?**
it should not impact casing.

bkadwal:
Can you please suggest.

Issue #4204, opened 2018-02-05
Keep inline styles case, eg. Margin for outlook emails


alexpineda:
**Do you want to request a *feature* or report a *bug*?**
feature

**What is the current behavior?**
Currently inline styles are normalized to lowercase letters

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via [fiddle.tinymce.com](http://fiddle.tinymce.com/) or similar.**
Set source code to:
```html
<div style="width:100%"> <div style="width:50px; Margin: 0 auto;">hello</div></div>
```

Result:
```html
<div style="width:100%"> <div style="width:50px; margin: 0 auto;">hello</div></div>
```

**What is the expected behavior?**
Have an option to keep case of inline styles, due to use in email clients.

**Which versions of TinyMCE, and which browser / OS are affected by this issue? Did this work in previous versions of TinyMCE?**
I know at least from 4.5.7 to 4.7.6 this is the behavior.

fyrkant:
Are you sure the casing of `margin` does a difference? From a quick google I found this that seems to say that outlook doesn't support margin at all: https://stackoverflow.com/questions/20347803/margin-in-html-not-working-in-outlook

bago:
AFAIK Margin vs margin WAS something needed with outlook.com (hotmail) until few months ago. AFAIK this is not valid anymore in any email client. Unfortunately many email frameworks, blogs, tools around keep suggesting you need to use Margin if you want to use margin in email, so people keep trying ;-)

@alexpineda can you tell if you checked the uppercase vs lowercase margin issue for real? Can you provide which client/client version and which html code let you reproduce the issue?

thedrjim:
This issue has been closed due to inactivity. Tiny will soon be working with members of the community to close other old issues on our GitHub repositories.
Please note the official support window for TinyMCE 4 is ending on December 31, 2020. If you haven't already upgraded to TinyMCE 5 it is highly encouraged that you do so. Upgrading will give you access to the latest releases and updates made to our WYSIWYG editor focused on bringing the latest UX and developer features into the editing experience. To assist with upgrading the editor we have put together a migration guide you can use here - https://www.tiny.cloud/blog/how-to-migrate-from-tinymce-4-to-tinymce-5/
If you have any questions or concerns, you can contact me at community-qa@tiny.cloud. You can also join the discussion on our mailing list at the following link: community-qa+subscribe@tiny.cloud.

PR #11082, opened 2026-04-20
TINY-11524: Retain user case for CSS property names and color values in Styles parser

ArvinJ-H:
Related Ticket: TINY-11524

Description of Changes:
* Stop lowercasing CSS property names and color values in `Styles.parse`. `color: RED`, `color: var(--MyVar)`, `color: {{MyVar}}`, `Margin-Top: 20px` now round-trip unchanged.
* Internal `behavior:` XSS guard and legacy `font-weight: 700 → bold` substitution route through a local lowercase so mixed-case names still trigger them.

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

GitHub issues:
* https://github.com/tinymce/tinymce/issues/7832
* https://github.com/tinymce/tinymce/issues/4204

## Summary by CodeRabbit

* **Bug Fixes**
  * Styles parsing now preserves user-provided casing for custom CSS properties and for many color values in style attributes; standard property names remain normalized and font-weight shorthand still normalizes.
* **Tests**
  * Added/updated tests covering casing preservation, shorthand compression, font-weight normalization, and XSS/behavior guarding.

### DIFF
--- a/modules/tinymce/src/core/main/ts/api/html/Styles.ts
+++ b/modules/tinymce/src/core/main/ts/api/html/Styles.ts
@@ -246,8 +246,10 @@ const Styles = (settings: StylesSettings = {}, schema?: Schema): Styles => {
             // Decode escaped sequences like \65 -> e
             name = decodeHexSequences(name);
             value = decodeHexSequences(value);
-            // Lowercase form for internal comparisons and storage; custom properties keep user case
-            const lowerName = name.toLowerCase();
+            // Custom properties (--*) keep user case; standard names normalize to lowercase
+            if (!name.startsWith('--')) {
+              name = name.toLowerCase();
+            }
 
             // Skip properties with double quotes and sequences like \" \' in their names
             // See 'mXSS Attacks: Attacking well-secured Web-Applications by using innerHTML Mutations'
@@ -257,12 +259,12 @@ const Styles = (settings: StylesSettings = {}, schema?: Schema): Styles => {
             }
 
             // Don't allow behavior name or expression/comments within the values
-            if (!settings.allow_script_urls && (lowerName === 'behavior' || /expression\s*\(|\/\*|\*\//.test(value))) {
+            if (!settings.allow_script_urls && (name === 'behavior' || /expression\s*\(|\/\*|\*\//.test(value))) {
               continue;
             }
 
             // Opera will produce 700 instead of bold in their style values
-            if (lowerName === 'font-weight' && value === '700') {
+            if (name === 'font-weight' && value === '700') {
               value = 'bold';
             }
 
@@ -275,7 +277,7 @@ const Styles = (settings: StylesSettings = {}, schema?: Schema): Styles => {
 
             // Convert URLs and force them into url('value') format
             value = value.replace(urlOrStrRegExp, processUrl);
-            styles[name.startsWith('--') ? name : lowerName] = isEncoded ? decode(value, true) : value;
+            styles[name] = isEncoded ? decode(value, true) : value;
           }
         }
         // Compress the styles to reduce it's size for example IE will expand styles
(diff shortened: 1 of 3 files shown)


modules/tinymce/src/core/main/ts/api/html/Styles.ts
```diff
@@ -255,15 +257,13 @@ const Styles = (settings: StylesSettings = {}, schema?: Schema): Styles => {
...
 
             // Opera will produce 700 instead of bold in their style values
-            if (name === 'font-weight' && value === '700') {
+            if (lowerName === 'font-weight' && value === '700') {
               value = 'bold';
-            } else if (name === 'color' || name === 'background-color') { // Lowercase colors like RED
-              value = value.toLowerCase();
             }
 
             // Convert RGB colors to HEX
```
spocke, line 271: <p class='graphite__hidden'><i>[Re: line +278]</i></p>
I'm not so sure we want to not normalize the property names. Since with this change parsing and expecting the keys to be lowercase would no longer work also you would be able to parse multiple values that define the same thing into the object. For example: `tinymce.html.Styles().parse('font-Weight: bold; font-weight: bold')` or `tinymce.html.Styles().parse('font-Weight: bold')['font-weight']` I think we shoul duse the lowerName name here unless the name starts with `--` because then it's a custom property definition.
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/tinymce/tinymce/11082?utm_source=unchanged-line-comment">Graphite</a>.</p>

spocke (changes requested):

modules/tinymce/src/core/main/ts/api/html/Styles.ts
spocke: So I didn't realize the implications that this would have if we are to change the case on non custom property things. That could in theory be a breaking change.

TheSpyder: I agree. I know we've had a lot of complaints about the decision to force lowercase, and Spocke himself logged the ticket to make this change, but I don't think we have enough tests to cover the edge cases this might break.

There are so many places we do style name lookups, are we sure they all use browser APIs where normalisation is done for us?

ArvinJ-H: I had a look what downstream consumers and found several lookup sites that assume lowercase keys: `ElementUtils.compare` format merging, schema `valid_styles/invalid_styles` serialize, `paste_webkit_styles` filter, `StyleUtils non-inheritable` set. All breaks under raw case-preserved storage.

We can exception `--`  to address one of the GH issue. would be a one-line fix at `Styles.ts:278` (ie. lower case everything except custom vars.

`styles[name.startsWith('--') ? name : lowerName] = isEncoded ? decode(value, true) : value;`

Original jira ticket does mention 2 GH issues, this closes GH #7832 (`var(--MyVar)`, templating tokens) but not GH #4204 (`Margin-Top:` round-trip for Outlook email).

spocke: I think the rule here is that we want to support the HTML and CSS specs and they now allow CSS custom properties to be case sensitive those didn't exist when this code was introduced. The templates strings like `{{fooBar}}` is not standard CSS so not sure we should support that. The browsers just retain the style attribute value as is it doesn't try to normalize it old IE did that is why it was a mess thats why we have the re-boxing of values etc in the serializer.

So I think the safe thing to do here is retain case on the values only and on the keys only if the `--` prefix is there.

spocke (approved):

modules/tinymce/src/core/main/ts/api/html/Styles.ts
```diff
@@ -239,13 +239,17 @@ const Styles = (settings: StylesSettings = {}, schema?: Schema): Styles => {
...
           let value = matches[2].replace(trimRightRegExp, '');
 
           if (name && value) {
             // Decode escaped sequences like \65 -> e
             name = decodeHexSequences(name);
             value = decodeHexSequences(value);
+            // Custom properties (--*) keep user case; standard names normalize to lowercase
+            if (!name.startsWith('--')) {
+              name = name.toLowerCase();
+            }
```
ArvinJ-H, line 252: @spocke The `toLowerCase` is still happening, its just a moved after custom variable exception check

spocke: Yes that fine then we scope it to only things where case really matters and also scoping it to something that is less likely to be used often. 👍

ltrouton (approved):

kimwoodfield (approved):

PR #11085, opened 2026-04-22
TINY-13341: Collapse dropdown menu on tab 

soritaheng:
Related Ticket: TINY-13341

Description of Changes:
 * Change Tab behaviour in dropdown menus to close the menu rather than cycle through its items, aligning with accessibility guidelines.                                            
  * Remove the `moveOnTab` option from the `alloy` module and add `onTab` and `onShiftTab` handlers to `TieredMenuSpec`.                        
  * Update existing tests and add new tests to verify the new behaviour.  

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

## Summary by CodeRabbit

* **Changed**
  * Pressing Tab or Shift+Tab now closes open dropdowns and tiered menus, then moves focus to the next or previous focusable element.
  * Keyboard navigation within menus now relies on arrow keys for item movement while Tab/Shift+Tab are used to exit menus and advance overall focus.


modules/alloy/src/main/ts/ephox/alloy/dropdown/DropdownUtils.ts
```diff
@@ -66,6 +68,16 @@ const openF = (
 
   const getLazySink = getSink(component, detail);
 
+  // When Tab or Shift+Tab bubbles up from inside an open menu, close the menu,
+  // refocus the trigger, and re-emit the keydown so the trigger's parent
+  // (toolbar, menubar, etc.) handles it via its own Keying config.
+  const onTabOutOfMenu = (_tmenu: AlloyComponent, se: NativeSimulatedEvent<KeyboardEvent>): Optional<boolean> => {
+    Focusing.focus(component);
+    AlloyTriggers.emitWith(component, 'keydown', { raw: se.event.raw });
+    Sandboxing.close(sandbox);
```
TheSpyder, line 77: I was hoping that we could just ignore the tab event inside the menu and have it operate the same as if tab had been pressed with the menu closed.

Sandbox close likely has some side effects, does it work better if you do that first?

soritaheng: @TheSpyder I tried letting the tab event bubble up without custom handling, but since the menu is in a sink that is rendered outside the trigger button's container, the browser would shift focus to the next/previous element in the tab order as determined from that sink. This is especially a problem in dialogs, where the focus would shift outside the dialog when a menu is closed via tab. The focus is also not guaranteed to always land in the editor's body. For example, in the self-hosted setup, it would land on an upgrade link first. In order to have the focus via tabbing to behave as though the menu was never opened, I went with this workaround, which follows the one introduced here: https://github.com/tinymce/tinymce/blob/12db8365273cd2d9798e1bee0344199492eabbad/modules/tinymce/src/themes/silver/main/ts/ui/dropdown/CommonDropdown.ts#L91-L116 But please let me know if there are other approaches I should explore. 

A side effect I found from calling `Sandboxing.close` is it would set the `aria-expanded="false"` attribute to the trigger button. I tried testing with VoiceOver on Mac by calling it first and moving the focus after, but there doesn't seem to be a noticeable effect in the screen reader announcement, and the behaviour appears to work the same.

TheSpyder: I thought Alloy overrode the tab order? If it's using the browser native behaviour then yeah I suppose this approach makes sense.

soritaheng: @TheSpyder Yes, containers like toolbar or dialog could set 'cyclic' mode for the keying and control the tab behaviour. But the container in this case is the sandbox sink, so the programmatic focus on the trigger button lets its parent container to take control of the tabbing again.

.changes/unreleased/tinymce-TINY-13341-2026-04-22.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Improved
+body: 'Pressing Tab or Shift+Tab now closes open dropdown menus and moves focus to next or previous focusable element.'
```
TheSpyder:
```suggestion
body: 'Pressing Tab or Shift+Tab on an open menu now closes it and moves focus to the next or previous focusable element.'
```

TheSpyder (approved):

kimwoodfield (approved):

shanmen-tiny (approved):

MitchC1999 (approved):

PR #11088, opened 2026-04-23
TINY-13783: confirmation dialog

lorenzo-pomili:
Related Ticket: TINY-13783

Description of Changes:
the idea here is to implement a component that can be used by the plugins, toggling it in the way that the plugin preferes

Pre-checks:
* [x] ~Changelog entry added~

Review:
* [x] Milestone set

## Summary by CodeRabbit

* **New Features**
  * Added a Confirmation dialog with async confirm/cancel handlers, including a loading spinner and disabled actions while confirming.
  * Added a programmatic hook-based trigger for opening the confirmation.
  * Added an imperative host-based alternative trigger for opening the confirmation.
* **Documentation**
  * Added Storybook examples covering default, success, and error flows, including both trigger approaches.


modules/oxide-components/src/main/ts/components/confirmation/Confirmation.tsx
```diff
@@ -0,0 +1,84 @@
...
+  <Spinner type="circle" />
+</div>;
+
+export const Confirmation: FC<ConfirmationProps> = (({
+  title,
+  text,
+  buttonName,
+  cancelBtnName = 'Cancel',
+  onConfirm,
+  onCancel
```
TheSpyder, line 20: Why have two? The existing confirm dialog uses a single callback that takes a boolean, that should be fine here as well.

For bonus points, you could make the callback return type `Promise<void> | void` to support scenarios where the callback doesn't need to be async.

modules/oxide-components/src/main/ts/components/confirmation/Confirmation.tsx
```diff
@@ -0,0 +1,84 @@
...
+  const onCancelHandler = () => {
+    // eslint-disable-next-line @typescript-eslint/no-floating-promises
+    onCancel().finally(() => {
+      dialogRef.current?.close();
+    });
+  };
+
+  return <dialog ref={dialogRef} style={{ margin: 'auto' }}>
+    <div className={Bem.block('tox-dialog-wrap')}>
+      {confirming ? <Loading /> :
```
TheSpyder: I don't think we should hide the whole dialog contents while in the loading state. That seems quite visually jarring. Could we disable the components and make the spinner an overlay instead?

modules/oxide-components/src/main/ts/components/confirmation/Confirmation.tsx
lorenzo-pomili: about having the 2 handlers, it's needed if we want to let the caller manage the show/hide

about the `Promise<void> | void` I'll refactor it

TheSpyder: How does having separate handlers help with show/hide? It's just a boolean branch in one handler instead of two handlers.

spocke: I think it might be better to have these as two separate things since you might not care about the cancel some times say you use it for deleting a item then you don't really care if the user hit cancel since only the confirm is what would actually delete the thing. If it's just one callback you would have to have a if statement that just checks if the value is true.

lorenzo-pomili: oh, wait if you mean that we can pass an action to the callback (ex: 'cancel' | 'confirm') instead of having 2 callback, yes it could be an alternative I don't have any strong opinion on this

TheSpyder: Our existing confirm dialog has a single callback, and it's been fine :man-shrugging:  
  
Having two simplifies the "don't care about cancel" case to complicate the "care about both" case :man-shrugging:

hamza0867:
There should be some visual tests

modules/oxide-components/src/main/ts/components/confirmation/Confirmation.stories.tsx
```diff
@@ -0,0 +1,99 @@
...
+        component: `
+The Alert component displays warning/error feedback with optional action and remove controls.
+
+## Props
+- \`message\`: string content shown in the alert body
+- \`severity\`: \`error\` | \`warning\`
+- \`actions\`: optional slot for one or more action buttons/components
+- \`removable\`: optional close affordance (requires \`onRemove\`)
+- \`closeAriaLabel\`: optional accessible label for the close button (defaults to \`Close\`)
+        `
```
hamza0867: This was copied from somehwere else, and I don't even believe we need this description section for this component

modules/oxide-components/src/main/ts/components/confirmation/Confirmation.tsx
hamza0867: Or we can disable the buttons and add a spinner to the confirm button the same way we do in comments

TheSpyder: yeah whether the spinner is an overlay or just in the button, keeping the layout is essential.

modules/oxide-components/src/main/ts/components/confirmation/Confirmation.tsx
lorenzo-pomili: I tried to refactor but the problem is where the components are initialized:
- for `windowManager.confirm`, is the editor itself to manage the show/hide, and it does it also defining [what happen on cancel](https://github.com/tinymce/tinymce/blob/main/modules/tinymce/src/themes/silver/main/ts/ui/dialog/ConfirmDialog.ts#L66)
- for this new dialog is the plugin that manage the show/hide, so we need the `onCancel` to hide the dialog or know when the dialog is hidden

the problem of making it manage by the editor is that we would need to import the oxide-components alongside with React.

I'm trying to refactor a bit to separate better the show/hide manage and the rest but I'm not sure it's possible to do that

modules/oxide-components/src/main/ts/components/confirmation/Confirmation.stories.tsx
lorenzo-pomili: removed

modules/oxide-components/src/main/ts/components/confirmation/Confirmation.tsx
lorenzo-pomili: ok, I refactored it alongside with the implementation in the plugin, now:
- the plugin create a boilerplate that manage the show/hide -> this because at the moment centralize it in the editor would mean import oxide-components alongside with React in the editor
- the `onCancel` is used only to know when the dialog should be hidden

if you think this approach is ok, I'll finish to write some tests, refactor some usage in the plugin and open the PR

modules/oxide-components/src/main/ts/components/confirmation/Confirmation.tsx
lorenzo-pomili: changed, have a check if it was your idea

modules/oxide-components/src/main/ts/components/confirmation/Confirmation.tsx
TheSpyder: Sorry for not responding yesterday. And I know I'm being hard on this task but we're building important foundational components here.

I was not saying this dialog should be managed by the editor, but the way it's used should be similar (with a single callback instead of two). A single function can still work because it will be called for _both_ confirm and cancel.

However to your point about hiding the dialog, I see no reason why the hide/show logic needs to be in the plugin. It can return an empty component just as the plugin does in your other PR?

modules/oxide-components/src/main/ts/components/confirmation/Confirmation.tsx
```diff
@@ -0,0 +1,70 @@
...
+      setConfirming(false);
+    });
+  };
+
+  const onCancelHandler = () => {
+    // eslint-disable-next-line @typescript-eslint/no-floating-promises
+    onCancel();
+  };
+
+  return <div className={Bem.block('tox-dialog-wrap')}>
```
TheSpyder, line 39: You didn't have to roll all the way back to a div, you could've just said to me that yes, a `ref` and an `effect` is how `dialog` elements are deployed in React

https://stackoverflow.com/a/76567796/7546

modules/oxide-components/src/main/ts/components/confirmation/Confirmation.tsx
lorenzo-pomili: you are right about being hard on this, I want to make something easy to mantain too, sorry for the delay in the responses on my side too.

> I was not saying this dialog should be managed by the editor, but the way it's used should be similar (with a single callback instead of two). A single function can still work because it will be called for both confirm and cancel.

I did the boilerplate thing because otherwise I wasn't sure how to trigger the `show` of the dialog.

My idea is this the boilerplate it's the equivalent of [silver ConfirmDialog](https://github.com/tinymce/tinymce/blob/main/modules/tinymce/src/themes/silver/main/ts/ui/dialog/ConfirmDialog.ts) that manage some defaults (like the show/hide) and register the action that is the real interface to use this confirm dialog.

I was talking about manage it in the editor because that would "hide" the `onCancel` like it happens in silver ConfirmDialog, so I was thinking that until we find a way to centralize the boilerplate, the bolerplate (`ConfirmationDialog.tsx`) should be a just a copy paste that do the minimal amout of things.

> However to your point about hiding the dialog, I see no reason why the hide/show logic needs to be in the plugin. It can return an empty component just as the plugin does in your other PR?

I tried to do it but I have this states:

1. the plugin registers put dialog in the DOM and register the command -> the command registration is here, because we don't have the editor in the component and we don't have the component in core
2. some action call the command -> this needs to tell to the component to show somehow, setting a component prop to true or an external flag that render or not the component

now the problem is that if I don't have the `onCancel` I don't know when the modal will be hide so I cannot set the prop or the flag to false and in that way I cannot trigger the show again

we could pass a show prop as a Cell to the component to be possible to modify it both from the component and from the plugin, but it seems a little hacky to me

modules/oxide-components/src/main/ts/components/confirmation/Confirmation.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+  const onClick: MouseEventHandler<HTMLButtonElement> = (e) => {
+    e.stopPropagation();
+    e.preventDefault();
+    setConfirming(true);
+    // eslint-disable-next-line @typescript-eslint/no-floating-promises
+    onConfirm().finally(() => {
+      setConfirming(false);
+    });
+  };
```
hamza0867: We should wrap this up in a useCallback to stabilize the ref

modules/oxide-components/src/main/ts/components/confirmation/Confirmation.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+  const onCancelHandler = () => {
+    // eslint-disable-next-line @typescript-eslint/no-floating-promises
+    onCancel();
+  };
```
hamza0867: This looks weird, you are creating a new function that does nothing but call another function. I believe this is to change the type from returning a Promise to return void. In this case I suggest you add the type annotation so that it's easier to spot

modules/oxide-components/src/main/ts/components/confirmation/Confirmation.tsx
```diff
@@ -0,0 +1,70 @@
+import { useState, type FC, type MouseEventHandler } from 'react';
+
+import { Spinner } from '../../bespoke/tinymceai/spinner/Spinner';
+import * as Bem from '../../utils/Bem';
+import { Button } from '../button/Button';
+
+export interface ConfirmationProps {
+  readonly title: string;
+  readonly text: string;
+  readonly buttonName: string;
```
hamza0867, line 9: Shouldn't this one be optional as well ?

modules/oxide-components/src/main/ts/components/confirmation/Confirmation.tsx
```diff
@@ -0,0 +1,70 @@
+import { useState, type FC, type MouseEventHandler } from 'react';
+
+import { Spinner } from '../../bespoke/tinymceai/spinner/Spinner';
+import * as Bem from '../../utils/Bem';
+import { Button } from '../button/Button';
+
+export interface ConfirmationProps {
+  readonly title: string;
```
hamza0867: This should be optional as well

modules/oxide-components/src/main/ts/components/confirmation/Confirmation.tsx
hamza0867: I don't mind having two different callbacks. But if the goal is to "re-implement the alloy confirmation dialog but allow async callbacks" then we should go for one callback. That being said, if we are willing, I'd rather have two different callbacks instead of one.

About the management of show/hide and even the api itself. Instead of exposing a component, we can expose a hook that manages this component. This would provide an api even closer to the one we have now. You can look at this example for inspiration: https://ant.design/components/modal#modal-demo-confirm

modules/oxide-components/src/main/ts/components/confirmation/Confirmation.tsx
hamza0867: Agreed

modules/oxide-components/src/main/ts/components/confirmation/Confirmation.tsx
lorenzo-pomili: I don't have clear how do `ant` attach it to the DOM, which is the problem that the boilerplate solve, my current implementation is using a command to call the dialog:
```
editor.execCommand('confirmAction', false, {
      buttonName: 'Ok',
      cancelBtnName: 'Cancel',
      onConfirm: () => deleteConversation(chat.id).catch(universe.errorReporter.historyError),
      text: 'text test',
      title: 'title test'
    });
```
but it was just the faster way that I found to call it, it's ok to have an hook instead but it would require the boilerplate anyway

modules/oxide-components/src/main/ts/components/confirmation/Confirmation.tsx
lorenzo-pomili: reverted to dialog

modules/oxide-components/src/main/ts/components/confirmation/Confirmation.tsx
lorenzo-pomili: wrapped

modules/oxide-components/src/main/ts/components/confirmation/Confirmation.tsx
lorenzo-pomili: I moved it directly inside the `onClick` let me know if it is ok for you

modules/oxide-components/src/main/ts/components/confirmation/Confirmation.tsx
lorenzo-pomili: since now we use an hook as the API to use this component I made all the props mandatory

modules/oxide-components/src/main/ts/components/confirmation/Confirmation.tsx
lorenzo-pomili: since now we use an hook as the API to use this component I made all the props mandatory

modules/oxide-components/src/main/ts/components/confirmation/Confirmation.stories.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+const render = (args: ConfirmationProps): JSX.Element => {
+  const [ container, setContainer ] = useState<HTMLDivElement | null>(null);
+  const confirmationHook = useConfirmation(container);
+
+  return <div style={{ width: '480px' }}>
+    <div ref={setContainer}></div>
+    <button style={{ cursor: 'pointer' }} onClick={() =>
+      confirmationHook(args.text, args.onConfirm)
+    }>show</button>
```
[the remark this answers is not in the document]
lorenzo-pomili: changed, have a check

modules/oxide-components/src/main/ts/components/confirmation/Confirmation.tsx
```diff
@@ -0,0 +1,68 @@
...
+  const [ confirming, setConfirming ] = useState(false);
+
+  const onClick: MouseEventHandler<HTMLButtonElement> = useCallback((e) => {
+    e.stopPropagation();
+    e.preventDefault();
+    setConfirming(true);
+    // eslint-disable-next-line @typescript-eslint/no-floating-promises
+    onConfirm().finally(() => {
+      setConfirming(false);
+    });
```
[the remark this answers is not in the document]
lorenzo-pomili: changed, have a check

modules/oxide-components/src/main/ts/components/confirmation/Confirmation.tsx
```diff
@@ -0,0 +1,68 @@
...
+              aria-label={buttonName}
+            >{confirming ? <Spinner type="circle" /> : buttonName}</Button>
+            <Button
+              variant='secondary'
+              onClick={() => {
+                // eslint-disable-next-line @typescript-eslint/no-floating-promises
+                onCancel();
+              }}
+              aria-label={cancelBtnName}
+            >{cancelBtnName}</Button>
```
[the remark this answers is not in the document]
lorenzo-pomili: changed, have a check

modules/oxide-components/src/main/ts/components/confirmation/internals/ConfirmationHook.tsx
```diff
@@ -0,0 +1,45 @@
...
+  const [ root, setRoot ] = useState<Root | null>(null);
+  const [ request, setRequest ] = useState<Request | null>(null);
+
+  useEffect(() => {
+    if (container) {
+      const r = createRoot(container);
+      setRoot(r);
+      return () => r.unmount();
+    }
+  }, [ container ]);
```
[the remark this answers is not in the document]
lorenzo-pomili: changed, have a check

modules/oxide-components/src/main/ts/components/confirmation/internals/ConfirmationHook.tsx
```diff
@@ -0,0 +1,45 @@
...
+  useEffect(() => {
+    if (root) {
+      root.render(
+        request
+          ? <Confirmation
+            title=''
+            text={request.text}
+            buttonName='Yes'
+            cancelBtnName='No'
+            onConfirm={() => request.onConfirm().finally(() => setRequest(null))}
```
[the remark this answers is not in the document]
lorenzo-pomili: removed title, have a check

modules/oxide-components/src/main/ts/components/confirmation/internals/ConfirmationHook.tsx
```diff
@@ -0,0 +1,47 @@
...
+              setRequest(null);
+              return Promise.resolve();
+            }}
+          />
+          : null
+      );
+    }
+  }, [ root, request ]);
+
+  return useCallback((text: string, onConfirm: () => Promise<void>) => {
```
spocke: Shouldn't we also expose a onCancel?

modules/oxide-components/src/main/ts/components/confirmation/internals/ConfirmationHook.tsx
```diff
@@ -0,0 +1,47 @@
...
+  onConfirm: () => Promise<void>;
+}
+
+export const useConfirmation = (container: HTMLElement | null): (text: string, onConfirm: () => Promise<void>) => void => {
+  const [ root, setRoot ] = useState<Root | null>(null);
+  const [ request, setRequest ] = useState<Request | null>(null);
+
+  useEffect(() => {
+    if (container) {
+      const r = createRoot(container);
```
spocke: Is this really the best appraoch here having two react render trees. I think this can instead be modeled with having the div that you attached to the app be something reactive. So that is the thing that re-rendes the dialog based on confirm calls. Maybe using a imperative API.  
  
So where you want the confirm to render you create a `<ConfirmHost ref={confirmRef} />` that would then render the confirm element based on internal state exposed though an API on the ref.

So the imperative API could the be `confirmRef.current?.confirm({ text: 'Hello world', onConfirm: () => ...`}); so the confirmRef is what we expose on the global we could wrap it in a object that hides the ref as well make a funciton that takes a ref but returns an interface that does guards and calls the refs imperative API.

Calling that would then internally in that component either render null or the confirm dialog depending on an internal state. In theory the host could have multiple confirms at the same time if we wanted to and track all of them. We could then stick this into a sink element together with other sink like elements.

So for example:

```
<Sink>
  <ConfirmHost ref={confirmRef} />
  <AlertHost ref={alertRef} />
  <DialogHost ref={dialogRef} />
</Sink>
```

modules/oxide-components/src/main/ts/components/confirmation/Confirmation.tsx
```diff
@@ -0,0 +1,66 @@
...
+}) => {
+  const [ confirming, setConfirming ] = useState(false);
+
+  const onClick: MouseEventHandler<HTMLButtonElement> = useCallback((e) => {
+    e.stopPropagation();
+    e.preventDefault();
+    setConfirming(true);
+    void onConfirm()
+      // eslint-disable-next-line no-console
+      .catch(console.error)
```
hamza0867: I don't think it is the responsability of the confirmation dialog to report/handle errors. That should be done inside the `onConfirm` callback

modules/oxide-components/src/main/ts/components/confirmation/Confirmation.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+export interface ConfirmationProps {
+  readonly text: string;
+  readonly buttonName: string;
+  readonly cancelBtnName: string;
+  readonly onConfirm: () => Promise<void>;
+  readonly onCancel: () => Promise<void>;
+}
+
+export const Confirmation: FC<ConfirmationProps> = (({
```
hamza0867: I think there is an extra parenthesis here:

```suggestion
export const Confirmation: FC<ConfirmationProps> = ({
```

modules/oxide-components/src/main/ts/components/confirmation/Confirmation.tsx
```diff
@@ -0,0 +1,66 @@
...
+                void onCancel().catch(console.error);
+              }}
+              aria-label={cancelBtnName}
+            >{cancelBtnName}</Button>
+          </div>
+        </div>
+      </dialog>
+    </div>
+  </div>;
+});
```
hamza0867: It would make this one, unnecessary as well:

```suggestion
};
```

modules/oxide-components/src/main/ts/components/confirmation/internals/ConfirmationHook.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { createRoot, type Root } from 'react-dom/client';
+
+import { Confirmation } from '../Confirmation';
+
+interface Request {
+  text: string;
+  onConfirm: () => Promise<void>;
+}
+
+export const useConfirmation = (container: HTMLElement | null): (text: string, onConfirm: () => Promise<void>) => void => {
```
hamza0867: I would change this api to return an object with a property `confirm` like this:

```suggestion
export const useConfirmation = (container: HTMLElement | null): { confirm: (text: string, onConfirm: () => Promise<void>) } => void => {
```

modules/oxide-components/src/main/ts/components/confirmation/internals/ConfirmationHook.tsx
hamza0867: I agree with @spocke here. If this is the approach we want to take then we should render in a Portal. It can be discussed

modules/oxide-components/src/main/ts/components/confirmation/Confirmation.tsx
TheSpyder: I mind having two callbacks, as I've been trying to say for weeks, for example if we keep it as one the eventual use of this in the editor won't break the public API

https://www.tiny.cloud/docs/tinymce/latest/apis/tinymce.windowmanager/#confirm

lorenzo-pomili: @TheSpyder @spocke @hamza0867 since at the moment what we expose at the plugins is the hook, could it make sense to have the minimal interface for it so:
```
{
  text: string;
  onConfirm: () => Promise<void>;
}
```
and if we need in the future add `onCancel` etc.?

spocke: Did you see my comment in the other PR on this that we could model this as a component with a imperative api instead of a hook.

spocke: My mistake I was thinking of this one: https://github.com/tinymce/tinymce/pull/11088#discussion_r3296723874

lorenzo-pomili: sorry, I was checking another PR, anyway I'll try to implement the solution in your other comment

modules/oxide-components/src/main/ts/components/confirmation/internals/ConfirmationHook.tsx
lorenzo-pomili: done

modules/oxide-components/src/main/ts/components/confirmation/Confirmation.tsx
lorenzo-pomili: @spocke  I added a new `ConfirmationHost` that I think is what you meant in your comment, however I don't think this is the best solution as [React documentation](https://react.dev/reference/react/useImperativeHandle) (scroll till the bottom of the page) use ref could be not a good idea in this case, in my opinion the best option would be:
- the `Plugin` import a component that wraps `<Confirmation>` and register a command
- when we will have the opportunity to register this in `windowManager` we remove the export of the wrapper to break all the instances of the wrapper
- at this point refactor should be easy from:
```
editor.execCommand('confirmAction', false, {
    onConfirm: onConfirmFunction,
    text: modalText
});
```
to: 
```
editor.windowManager.asyncConfirm(modalText, onConfirmFunction);
```

modules/oxide-components/src/main/ts/components/confirmation/Confirmation.tsx
lorenzo-pomili: changed

modules/oxide-components/src/main/ts/components/confirmation/Confirmation.tsx
lorenzo-pomili: changed

modules/oxide-components/src/main/ts/components/confirmation/internals/ConfirmationHook.tsx
lorenzo-pomili: changed

modules/oxide-components/src/main/ts/components/confirmation/Confirmation.stories.tsx
```diff
@@ -0,0 +1,116 @@
+import { Obj } from '@ephox/katamari';
+import type { Meta, StoryObj } from '@storybook/react-vite';
+import { getAll as getAllIcons } from '@tinymce/oxide-icons-default';
+import { useRef, useState } from 'react';
+
+import { UniverseProvider } from '../../main';
+
+import { ConfirmationHost, type ConfirmationHostHandle } from './internals/ConfimrationHost';
```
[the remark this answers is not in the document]
lorenzo-pomili: name changed

modules/oxide-components/src/main/ts/components/confirmation/Confirmation.tsx
lorenzo-pomili: @spocke @TheSpyder @hamza0867 now we have an example of the `confirmationHook` (https://github.com/tinymce/tinymce/pull/11088/changes#diff-2f633ad48e2b08f00409f16d1e5577aa728f49cfe4a5fce1664179efebf77be4R21-R31) and one of the `confirmationHost` (https://github.com/tinymce/tinymce/pull/11088/changes#diff-2f633ad48e2b08f00409f16d1e5577aa728f49cfe4a5fce1664179efebf77be4R86-R91) 

let me know if you are happy with one of the interface so I can solve the various comments, clean-up and open the PR

modules/oxide-components/src/main/ts/components/confirmation/internals/ConfirmationHost.tsx
```diff
@@ -0,0 +1,31 @@
...
+interface ConfirmOptions {
+  readonly text: string;
+  readonly onConfirm: () => Promise<void>;
+}
+
+export interface ConfirmationHostHandle {
+  confirm: (options: ConfirmOptions) => void;
+}
+
+export const ConfirmationHost = forwardRef<ConfirmationHostHandle>((_props, ref) => {
```
spocke, line 14: Nice that became a lot cleaner than I thought. I vote for this one over the hook since it's then just like any other component.

lorenzo-pomili: ok, thanks, I'll go with that

modules/oxide-components/src/main/ts/components/confirmation/Confirmation.tsx
lorenzo-pomili: ok, there is the implementation with the ConfirmationHost, have a check

modules/oxide-components/src/main/ts/components/confirmation/internals/ConfirmationHook.tsx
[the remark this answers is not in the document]
lorenzo-pomili: I moved both in `ConfirmationHost` and they are static now, have a check

modules/oxide-components/src/main/ts/components/confirmation/internals/ConfirmationHook.tsx
lorenzo-pomili: there is a comment about not exposing it: https://github.com/tinymce/tinymce/pull/11088#discussion_r3315446680

modules/oxide-components/src/main/ts/components/confirmation/Confirmation.tsx
lorenzo-pomili: the caller could intercept the error in the definition of their `onConfirm`, this is just a fallback in case it isn't managed

hamza0867: Still, I don't think it should be here as it would hide errors since they are now caught. No need for this fallback

hamza0867 (approved):

modules/oxide-components/src/main/ts/components/confirmation/Confirmation.tsx
lorenzo-pomili: removed

modules/oxide-components/src/main/ts/components/confirmation/internals/ConfirmationHost.tsx
```diff
@@ -0,0 +1,31 @@
...
+    text={request.text}
+    buttonName='Yes'
+    cancelBtnName='No'
+    onConfirm={() => request.onConfirm().finally(() => setRequest(null))}
+    onCancel={() => {
+      setRequest(null);
+      return Promise.resolve();
+    }}
+  /> : null;
+});
```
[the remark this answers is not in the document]
lorenzo-pomili: changed have a check

modules/oxide-components/src/main/ts/bespoke/tinymceai/spinner/Spinner.tsx
```diff
@@ -1,13 +1,16 @@
+import type * as CSS from 'csstype';
```
TheSpyder: In other cases we've done `import type { Property } from 'csstype';`, I think I'd prefer us to be consistent

TheSpyder (approved):

modules/oxide-components/src/main/ts/bespoke/tinymceai/spinner/Spinner.tsx
lorenzo-pomili: changed

PR #11095, opened 2026-04-27
TINY-14318: AutoResizingTextarea no longer over-grows when first rendered inside a collapsed accordion

hamza0867:
Related Ticket: TINY-14318

Description of Changes:

- Fixed `AutoResizingTextarea` incorrectly measuring `singleRowHeight` as `0` when the component is initially mounted inside a hidden ancestor (e.g. a `display: none` container such as a collapsed accordion)
- Added an `isVisible` utility that checks whether a textarea has layout by inspecting `offsetParent` and `getClientRects`, and used it to guard both the height measurement and the resize logic
- Introduced a `ResizeObserver` in the `useLayoutEffect` to defer the `singleRowHeight` measurement until the textarea becomes visible, disconnecting once a valid measurement is obtained to avoid spurious `ResizeObserver loop` warnings
- Ensured `computeSingleRowHeight` returns a minimum of `1` rather than a potentially invalid `0` when `scrollHeight` is unavailable
- Added a browser test covering the case where a textarea is mounted hidden and then revealed, verifying it correctly sizes to 1 row and remains at 1 row after typing a short line of text

Pre-checks:

Review:

- [x] Milestone set

## Summary by CodeRabbit

* **Bug Fixes**
  * Auto-resizing textarea now defers measuring until visible and guarantees a minimum single-row height, preventing incorrect sizing and spurious resize updates when mounted hidden.

* **Tests**
  * Added a browser test that verifies the textarea resolves to one row when revealed and stays stable after typing, improving regression coverage.

### DIFF
--- a/modules/oxide-components/src/main/ts/components/autoresizingtextarea/AutoResizingTextarea.tsx
+++ b/modules/oxide-components/src/main/ts/components/autoresizingtextarea/AutoResizingTextarea.tsx
@@ -1,4 +1,5 @@
-import { Type } from '@ephox/katamari';
+import { Cell, Singleton, Type } from '@ephox/katamari';
+import { SugarElement, Visibility } from '@ephox/sugar';
 import { forwardRef, useLayoutEffect, useMemo, useRef, useState, type MutableRefObject } from 'react';
 
 import * as Bem from '../../utils/Bem';
@@ -36,15 +37,42 @@ export const AutoResizingTextarea = forwardRef<HTMLTextAreaElement, AutoResizing
 }, ref) => {
   const textareaRef: MutableRefObject<HTMLTextAreaElement | null> = useRef(null);
 
-  // Here we use the useState hook instead of useMemo, because computing the singleRowHeight
-  // requires the textareaRef.current to be available, but useMemo is computed before the render,
-  //  so we set it to 1 initially and update it in the useLayoutEffect
+  // Initial value of 1 is a placeholder; the real value is measured once the textarea
+  // actually has layout (it may mount inside a `display: none` ancestor — e.g. a collapsed
+  // accordion — where `scrollHeight` reads 0). A ResizeObserver re-measures when layout returns.
   const [ singleRowHeight, setSingleRowHeight ] = useState(1);
 
   useLayoutEffect(() => {
-    if (textareaRef.current) {
-      setSingleRowHeight(computeSingleRowHeight(textareaRef.current));
+    const textarea = textareaRef.current;
+    if (Type.isNullable(textarea)) {
+      return;
     }
+
+    const measured = Cell(false);
+    const observer = Singleton.value<InstanceType<typeof window.ResizeObserver>>();
+
+    const tryCompute = () => {
+      if (measured.get() || !Visibility.isVisible(SugarElement.fromDom(textarea))) {
+        return;
+      }
+      const value = computeSingleRowHeight(textarea);
+      if (value > 1) {
+        measured.set(true);
+        setSingleRowHeight(value);
+        // Once measured, we're done — stop observing so subsequent `rows`
+        // mutations from `resizeTextarea` don't re-fire this callback (which
+        // would surface as a `ResizeObserver loop` warning).
+        observer.on((obs) => obs.disconnect());
+        observer.clear();
+      }
+    };
+
+    tryCompute();
+    if (!measured.get()) {
+      observer.set(new window.ResizeObserver(tryCompute));
+      observer.on((obs) => obs.observe(textarea));
+    }
+    return () => observer.on((obs) => obs.disconnect());
   }, []);
 
   // The minRows and maxRows only need to be computed once per component instance, so they are in the useMemo hook
--- a/modules/oxide-components/src/main/ts/components/autoresizingtextarea/AutoResizingTextareaUtils.ts
+++ b/modules/oxide-components/src/main/ts/components/autoresizingtextarea/AutoResizingTextareaUtils.ts
@@ -1,8 +1,22 @@
+import { SugarElement, Visibility } from '@ephox/sugar';
+
 import type { Height } from './AutoResizingTextareaTypes';
 
 interface ComputeMaxRowsProps {
   readonly maxHeight: Height;
   readonly singleRowHeight: number;
+}
+
+interface ComputeMinRowsProps {
+  readonly minHeight: Height;
+  readonly singleRowHeight: number;
+}
+
+interface ComputeNewRowsProps {
+  readonly minRows: number;
+  readonly maxRows: number;
+  readonly singleRowHeight: number;
+  readonly textarea: HTMLTextAreaElement;
 }
 
 const computeSingleRowHeight = (textarea: HTMLTextAreaElement): number => {
@@ -13,7 +27,7 @@ const computeSingleRowHeight = (textarea: HTMLTextAreaElement): number => {
   const res = textarea.scrollHeight;
   textarea.rows = originalRows;
   textarea.value = originalValue;
-  return res;
+  return Math.max(res, 1);
 };
 
 const computeMaxRows = ({ maxHeight, singleRowHeight }: ComputeMaxRowsProps): number => {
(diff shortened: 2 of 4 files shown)


hamza0867:
* **#11096** <a href="https://app.graphite.com/github/pr/tinymce/tinymce/11096?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#11095** <a href="https://app.graphite.com/github/pr/tinymce/tinymce/11095?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/tinymce/tinymce/11095?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

modules/oxide-components/src/main/ts/components/autoresizingtextarea/AutoResizingTextarea.tsx
```diff
@@ -36,15 +36,41 @@ export const AutoResizingTextarea = forwardRef<HTMLTextAreaElement, AutoResizing
...
+  // Initial value of 1 is a placeholder; the real value is measured once the textarea
+  // actually has layout (it may mount inside a `display: none` ancestor — e.g. a collapsed
+  // accordion — where `scrollHeight` reads 0). A ResizeObserver re-measures when layout returns.
   const [ singleRowHeight, setSingleRowHeight ] = useState(1);
 
   useLayoutEffect(() => {
-    if (textareaRef.current) {
-      setSingleRowHeight(computeSingleRowHeight(textareaRef.current));
+    const textarea = textareaRef.current;
+    if (!textarea) {
```
kimwoodfield:
```suggestion
    if (!Type.isNonNullable(textarea)) {
```

modules/oxide-components/src/test/ts/browser/components/AutoResizingTextarea.spec.tsx
```diff
@@ -236,4 +236,57 @@ describe('browser.components.AutoResizingTextareaTest', () => {
       message: 'Textarea rows should be initially resolved to 1'
     }).toHaveAttribute('rows', `${1}`);
   });
+
+  it('TINY-14318: Should size to 1 row when typed in a textarea that was initially mounted hidden', async () => {
```
kimwoodfield, line 240: @michalnieruchalski-tiugo Should we consider including an additional test case to cover the popover with 0 height scenario?

modules/oxide-components/src/test/ts/browser/components/AutoResizingTextarea.spec.tsx
```diff
@@ -236,4 +236,57 @@ describe('browser.components.AutoResizingTextareaTest', () => {
...
+        </>
+      );
+    };
+
+    const { getByTestId } = render(
+      <TestComponent />,
+      {
+        wrapper: ({ children }) => {
+          return (
+            <div className={classes([ 'tox' ])}>
```
hamza0867: Should use Bem helpers instead

modules/oxide-components/src/main/ts/components/autoresizingtextarea/AutoResizingTextarea.tsx
```diff
@@ -36,15 +36,41 @@ export const AutoResizingTextarea = forwardRef<HTMLTextAreaElement, AutoResizing
...
+        return;
+      }
+      const value = computeSingleRowHeight(textarea);
+      if (value > 1) {
+        measured = true;
+        setSingleRowHeight(value);
+        // Once measured, we're done — stop observing so subsequent `rows`
+        // mutations from `resizeTextarea` don't re-fire this callback (which
+        // would surface as a `ResizeObserver loop` warning).
+        observer?.disconnect();
```
TheSpyder: I suspect a singleton work better here? It's designed to paper over uninitialised variable state with an optional. I doubt we have a pre-built ResizeObserver singleton but it should be easy to build one, or perhaps add a new singleton type for it.

Especially since the only thing you need is to clear the singleton contents, which has no effect if the contents is `none`.

modules/oxide-components/src/main/ts/components/autoresizingtextarea/AutoResizingTextareaUtils.ts
```diff
@@ -45,7 +49,15 @@ interface ComputeNewRowsProps {
   readonly textarea: HTMLTextAreaElement;
 }
 
+// A textarea inside a `display: none` ancestor has no layout — `scrollHeight` reads as 0,
+// which would make the resize formula produce garbage. Detect that and skip until layout returns.
+const isVisible = (textarea: HTMLTextAreaElement): boolean =>
+  textarea.offsetParent !== null || textarea.getClientRects().length > 0;
```
TheSpyder: If the problematic condition is a zero scrollheight... why not check for that?

Or perhaps leverage how we've done it in sugar for a decade or so.

https://github.com/tinymce/tinymce/blob/947db4e6c8f865ce5b6990e4d4a07bb94d55c9a4/modules/sugar/src/main/ts/ephox/sugar/api/view/Visibility.ts#L28-L29

TheSpyder (approved):

modules/oxide-components/src/main/ts/components/autoresizingtextarea/AutoResizingTextareaUtils.ts
```diff
@@ -13,7 +27,11 @@ const computeSingleRowHeight = (textarea: HTMLTextAreaElement): number => {
   const res = textarea.scrollHeight;
   textarea.rows = originalRows;
   textarea.value = originalValue;
-  return res;
+  if (res >= 1) {
+    return res;
+  } else {
+    return 1;
+  }
```
spocke: `return res >= 1 ? res : 1;` or `return Math.max(res, 1)` .

modules/oxide-components/src/main/ts/components/dropdown/Dropdown.tsx
```diff
@@ -1,5 +1,5 @@
 import { Id, Type } from '@ephox/katamari';
-import { Children, cloneElement, forwardRef, isValidElement, useCallback, useMemo, useRef, useState, type FC, type HTMLAttributes, type MouseEvent, type PropsWithChildren, type ReactElement, type KeyboardEvent, useLayoutEffect } from 'react';
+import { Children, cloneElement, forwardRef, isValidElement, useCallback, useLayoutEffect, useMemo, useRef, useState, type FC, type HTMLAttributes, type KeyboardEvent, type MouseEvent, type PropsWithChildren, type ReactElement } from 'react';
```
spocke: This line is comming very long should we indent it?

modules/oxide-components/src/main/ts/components/autoresizingtextarea/AutoResizingTextarea.tsx
```diff
@@ -36,15 +37,42 @@ export const AutoResizingTextarea = forwardRef<HTMLTextAreaElement, AutoResizing
...
   useLayoutEffect(() => {
-    if (textareaRef.current) {
-      setSingleRowHeight(computeSingleRowHeight(textareaRef.current));
+    const textarea = textareaRef.current;
+    if (Type.isNullable(textarea)) {
+      return;
+    }
+
+    const measured = Cell(false);
+    const observer = Singleton.value<InstanceType<typeof window.ResizeObserver>>();
```
spocke, line 52: This is a pretty interesting way of detecting when a element is visible. I wonder is there any lag time for the resize observer on all browsers like will it trigger after a render tick then one might see a flicker or is it always instant.

spocke (approved):

modules/oxide-components/src/main/ts/components/dropdown/Dropdown.tsx
hamza0867: If we do then it will have each element imported in a separate line. I just want to confirm we are okay with that. I am okay with it.

modules/oxide-components/src/main/ts/components/autoresizingtextarea/AutoResizingTextarea.tsx
hamza0867: It depends, on who the observer is, for the human eye it's pretty instantanious. Also the automatic tests ran fine :man-shrugging:

modules/oxide-components/src/test/ts/browser/components/AutoResizingTextarea.spec.tsx
michalnieruchalski-tiugo: We could, but it's basically the same cause of the problem - no height on component mount. I'll leave it up to Hamza to decide.

kimwoodfield (approved):

hamza0867:
### Merge activity

* **May 8, 8:47 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/tinymce/tinymce/11095).
* **May 8, 8:47 AM UTC**: @hamza0867 merged this pull request with [Graphite](https://app.graphite.com/github/pr/tinymce/tinymce/11095).

PR #11101, opened 2026-05-06
TINY-13886: fix quirk for selection in `li`

lorenzo-pomili:
Related Ticket: TINY-13886

Description of Changes:
In Chrome in a `LI` that contains a block element and where the first child is an inline element
clicking on the right side of the first child the carret goes at the start of the element instead that in the end of it
issue: https://issues.chromium.org/issues/40767343

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where clicking on the right side of content within list items could sometimes fail to place the caret correctly; clicks now reliably position the caret at the expected location across affected scenarios.

* **Tests**
  * Added browser tests covering ordered, unordered, inline-with-block-children, and nested list scenarios to verify caret placement for right-side clicks.

* **Changelog**
  * Added an unreleased changelog entry documenting the fix (TINY-13886).


modules/tinymce/src/plugins/accordion/test/ts/browser/QuirksTest.ts
```diff
@@ -31,4 +32,48 @@
...
+    const rng = editor.getDoc().createRange();
+    rng.setStart(textNode, 0);
+    rng.setEnd(textNode, textNode.data.length);
+    const rect = rng.getClientRects()[0];
+
+    const mouseEvent = {
+      target: li as EventTarget,
+      clientX: rect.right + 100,
+      clientY: rect.top + rect.height / 2
+    } as MouseEvent
```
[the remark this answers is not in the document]
lorenzo-pomili: changed, have a check

modules/tinymce/src/core/main/ts/util/Quirks.ts
```diff
@@ -707,6 +709,32 @@ const Quirks = (editor: Editor): Quirks => {
...
+  const clickAfterEl = (clientX: number, clientY: number, rect: NodeClientRect): boolean =>
+    clientX >= rect.right && clientY >= rect.top && clientY <= rect.bottom;
+
+  const fixInLISelection = () => {
+    editor.on('mousedown', (e) => {
+      const target = SugarElement.fromDom(e.target);
+      if (isListItem(target) && hasBlockChildren(target)) {
+        const firstChildOpt = Traverse.firstChild(target);
+        firstChildOpt.each((firstChild) => {
+          if (clickAfterEl(e.clientX, e.clientY, getClientRects([ firstChild.dom ])[0])) {
```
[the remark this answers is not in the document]
lorenzo-pomili: changed, have a check

modules/tinymce/src/plugins/accordion/test/ts/browser/QuirksTest.ts
```diff
@@ -31,4 +32,48 @@ describe('browser.tinymce.plugins.accordion.QuirksTest', () => {
...
+    editor.dispatch('mousedown', mouseEvent);
+    editor.dispatch('click', mouseEvent);
+    editor.dispatch('mouseup', mouseEvent);
+
+    TinyAssertions.assertCursor(editor, [ 0, 0, 0 ], textNode.length);
```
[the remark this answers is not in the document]
lorenzo-pomili: changed, have a check

modules/tinymce/src/core/main/ts/util/Quirks.ts
```diff
@@ -707,6 +709,32 @@ const Quirks = (editor: Editor): Quirks => {
...
+            editor.selection.setCursorLocation(firstChild.dom, firstChild.dom.nodeValue?.length ?? 0);
```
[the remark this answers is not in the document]
lorenzo-pomili: changed, have a check

modules/tinymce/src/core/main/ts/util/Quirks.ts
```diff
@@ -707,6 +709,34 @@ const Quirks = (editor: Editor): Quirks => {
...
+  const clickAfterEl = (clientX: number, clientY: number, rect: NodeClientRect): boolean =>
+    clientX >= rect.right && clientY >= rect.top && clientY <= rect.bottom;
+
+  const fixInLISelection = () => {
+    editor.on('mousedown', (e) => {
+      const target = SugarElement.fromDom(e.target);
+      if (isListItem(target) && hasBlockChildren(target)) {
+        const firstChildOpt = Traverse.firstChild(target);
+        firstChildOpt.each((firstChild) => {
+          if (Arr.get(getClientRects([ firstChild.dom ]), 0).exists((rect) => clickAfterEl(e.clientX, e.clientY, rect))) {
```
spocke: I think this needs to be a bit more advanced since the content might be `<li><b>foo</b>bar<ul>` so N children. I think you should make a range from the first child up to the first block and then find a matching client rect in that.

modules/tinymce/src/core/main/ts/util/Quirks.ts
```diff
@@ -707,6 +709,34 @@ const Quirks = (editor: Editor): Quirks => {
...
+            const firstChildText = (SugarNode.isText(firstChild) ? firstChild : PredicateFind.descendant(firstChild, SugarNode.isText).getOr(firstChild)).dom;
```
spocke: Not sure if this is advanced enough either since the structure might be `<li><ul>` so we need to traverse back from the first block to the first suitable caret position I think we can use `CaretFinder.prevPosition` for that and from that you can get a range and just set it with `editor.selection.setRng`.

modules/tinymce/src/core/main/ts/util/Quirks.ts
```diff
@@ -746,6 +776,7 @@ const Quirks = (editor: Editor): Quirks => {
     // All browsers
     removeBlockQuoteOnBackSpace();
     emptyEditorWhenDeleting();
+    fixInLISelection();
```
spocke: This should only run on Chrome.

- [x] Make sure patch only runs on Chrome

modules/tinymce/src/core/main/ts/util/Quirks.ts
```diff
@@ -707,6 +709,34 @@ const Quirks = (editor: Editor): Quirks => {
     });
   };
 
+  /**
+   * In Chrome in a `LI` that contains a block element and where the first child is an inline element
+   * clicking on the right side of the first child the carret goes at the start of the element instead that in the end of it
+   * issue: https://issues.chromium.org/issues/40767343
+  **/
+  const hasBlockChildren = (target: SugarElement<any>): boolean =>
```
spocke: This is HTMLElement or Element at least not `any`

spocke (changes requested):

modules/tinymce/src/core/main/ts/util/Quirks.ts
lorenzo-pomili: done

modules/tinymce/src/core/main/ts/util/Quirks.ts
lorenzo-pomili: changed

modules/tinymce/src/core/main/ts/util/Quirks.ts
lorenzo-pomili: I changed the logic, now it looks for the first inline element before the block and put the caret after it

modules/tinymce/src/core/main/ts/util/Quirks.ts
```diff
@@ -707,6 +709,64 @@ const Quirks = (editor: Editor): Quirks => {
     });
   };
 
+  const hasBlockChildren = (target: SugarElement<Node>): boolean =>
```
spocke: I don't think you need this one since you already have the firstBlockChild is that returns some then there is blocks. So you could just do a `firstBlockChild(target).each((firstBlock) => {})` inside the if statement then inside that you do all the handling. You could just wrap all the contents in from the start of the target to the before the first block then get the client rect of that so like:

```typescript
const rng = new Range();
rng.setStartBefore(target.firstChild);
rng.setEndBefore(firstBlock.dom);
Arr.get(rng.getClientRects(), 0).each((firstRect) => { ... })
```

I don't think there will be that many rects between the first node and to the first block unless the whole document is like one LI or something weird.

spocke (changes requested):

modules/tinymce/src/core/main/ts/util/Quirks.ts
```diff
@@ -707,6 +709,64 @@ const Quirks = (editor: Editor): Quirks => {
...
+    Arr.exists(Traverse.children(target), (child) => SugarNode.isElement(child) ? Css.get(child, 'display') === 'block' : false);
```
TheSpyder: If `find` and `has` are duplicate conditions, please don't repeat the search. Just start with `find` and use the `none` case for "does not have"

lorenzo-pomili: I removed `hasBlockChildren`

modules/tinymce/src/core/main/ts/util/Quirks.ts
lorenzo-pomili: I removed `hasBlockChildren`, but I don't get the changes about how to calculate the rect, at the moment we need to intercept the last inline element to check if the click was in the correct place so at the right of the last inline element before the first block element, with your change I'd have to get the `lastInlineBeforeBlock` anyway

spocke: If we check the last child before the block is there a risk that it breaks if say a br is be before that element or if the text is wrapping maybe I'm over thinking it. Like is that enough or do we need to check the lines before the block.

modules/tinymce/src/core/main/ts/util/Quirks.ts
```diff
@@ -707,6 +709,59 @@ const Quirks = (editor: Editor): Quirks => {
...
+
+  const fixInLISelection = () => {
+    editor.on('mousedown', (e) => {
+      const target = SugarElement.fromDom(e.target);
+      if (isListItem(target)) {
+        firstBlockChild(target).bind((block) => {
+          const prevSibling = Traverse.prevSibling(block);
+          return prevSibling.exists(isValidSibling)
+            ? prevSibling
+            : prevSibling.bind((el) => Traverse.prevSibling(el).or(prevSibling));
```
TheSpyder: Is the fallback always specifically the second previous sibling? Or should this search as far as it can?

modules/tinymce/src/core/main/ts/util/Quirks.ts
```diff
@@ -707,6 +709,59 @@ const Quirks = (editor: Editor): Quirks => {
     });
   };
 
+  const firstBlockChild = (target: SugarElement<Node>) =>
+    Arr.find(Traverse.children(target), (child) => SugarNode.isElement(child) ? Css.get(child, 'display') === 'block' : false);
```
TheSpyder: You can save a little bit of overhead by using `PredicateFind.child`. This will wrap target child nodes in `Sugar` objects on demand instead of doing them all up front and searching the result.

```suggestion
    PredicateFind.child(target, (child) => SugarNode.isElement(child) ? Css.get(child, 'display') === 'block' : false);
```

modules/tinymce/src/core/main/ts/util/Quirks.ts
```diff
@@ -707,6 +709,59 @@ const Quirks = (editor: Editor): Quirks => {
...
+   * <li>a
+   * <div>b</div></li>
+   * ```
+   * since if the indentation of the HTML has a new line it creates a fake child in the `li` that is an empty text
+  **/
+  const isValidSibling = (el: SugarElement<Node>): boolean => {
+    if (SugarNode.isText(el)) {
+      return el.dom.nodeValue !== null && Strings.isNotEmpty(el.dom.nodeValue.trim());
+    }
+    return true;
```
TheSpyder: This concept is wrapped up in `SugarText.getOption()` where it will look something like `SugarText.getOption(el).exists(Strings.isNotEmpty)`

We don't use Sugar as much as we used to, but when we did a lot of common patterns were turned into composable functions :slightly_smiling_face:

.changes/unreleased/tinymce-TINY-13886-2026-05-06.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: Selection clicking at the right of an `li` could fail in some cases.
```
TheSpyder: Should this perhaps mention that it's a workaround?

`Workaround for a chrome bug selecting text in list items` for example

TheSpyder (approved):

modules/tinymce/src/core/main/ts/util/Quirks.ts
lorenzo-pomili: changed

modules/tinymce/src/core/main/ts/util/Quirks.ts
lorenzo-pomili: the case with BR wasn't managed correctly (I added a test case for it), I changed the mechanism now:
- I look for the first block or new line
- I get the last `prevSiblings` that creates a rect (this is needed to exclude the new lines created by the indentation)
- if the click in on the right side of this last element I put the selection at the end of it

modules/tinymce/src/core/main/ts/util/Quirks.ts
lorenzo-pomili: I changed the mechanism, check this comment for more information: https://github.com/tinymce/tinymce/pull/11101#discussion_r3241512389

modules/tinymce/src/core/main/ts/util/Quirks.ts
lorenzo-pomili: changed, check this comment for more information: https://github.com/tinymce/tinymce/pull/11101#discussion_r3241512389

.changes/unreleased/tinymce-TINY-13886-2026-05-06.yaml
lorenzo-pomili: changed, have a check

TheSpyder: Better, but now it needs a grammar update

.changes/unreleased/tinymce-TINY-13886-2026-05-06.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: In Chrome, selection clicking at the right of an `li` could fail in some cases; this workaround solves it.
```
TheSpyder:
```suggestion
body: Worked around a Chromium bug where clicking on the right of an `li` could fail.
```

TheSpyder (approved):

.changes/unreleased/tinymce-TINY-13886-2026-05-06.yaml
lorenzo-pomili: fixed

.changes/unreleased/tinymce-TINY-13886-2026-05-06.yaml
lorenzo-pomili: fixed

modules/tinymce/src/core/main/ts/util/Quirks.ts
```diff
@@ -707,6 +709,53 @@ const Quirks = (editor: Editor): Quirks => {
...
 
+  /**
+   * this is needed to manage the difference between
+   * ```
+   * <li>a<div>b</div></li>
+   * ```
+   * and
+   * ```
+   * <li>a
+   * <div>b</div></li>
```
spocke: Could we make this:

```suggestion
   * <li>a<br><div>b</div></li>
```

modules/tinymce/src/core/main/ts/util/Quirks.ts
```diff
@@ -707,6 +709,53 @@ const Quirks = (editor: Editor): Quirks => {
...
+   * ```
+   * since if the indentation of the HTML has a new line it creates a fake child in the `li` that is an empty text
+   * it's check it trying to get the rects and if it can't it means that it's the false unwanted new line
+  **/
+  const isValidSibling = (el: SugarElement<Node>): boolean =>
+    getClientRects([ el.dom ]).length > 0;
+
+  const firstBlockChildOrNewLine = (target: SugarElement<Node>) =>
+    PredicateFind.child(target, (child) => SugarNode.isTag('br')(child) || SugarNode.isElement(child) && Css.get(child, 'display') === 'block');
```
spocke: The ElementType.isBr can be used here or the NodeType.isBr(child.dom) I think those are fine to use since a BR will always be a BR so those are not schema specified.

modules/tinymce/src/core/main/ts/util/Quirks.ts
```diff
@@ -707,6 +709,53 @@ const Quirks = (editor: Editor): Quirks => {
...
+   * clicking on the right side of the first child the carret goes at the start of the element instead that in the end of it
+   * issue: https://issues.chromium.org/issues/40767343
+  **/
+
+  const fixInLISelection = () => {
+    editor.on('mousedown', (e) => {
+      const target = SugarElement.fromDom(e.target);
+      if (isListItem(target)) {
+        firstBlockChildOrNewLine(target).map(Traverse.prevSiblings).bind((prevSiblings) =>
+          Arr.findLastIndex(prevSiblings, isValidSibling).bind((lastI) => Arr.get(prevSiblings, lastI))
```
spocke, line 750: Oh that is super annoying that we don't have a findLast in katamari. Maybe log a ticket to add that seems like a useful thing to have.

modules/tinymce/src/core/main/ts/util/Quirks.ts
```diff
@@ -707,6 +709,53 @@ const Quirks = (editor: Editor): Quirks => {
...
+        ).each((lastInlineBeforeBlock) => {
+          if (Arr.get(getClientRects([ lastInlineBeforeBlock.dom ]), 0).exists((rect) => clickAfterEl(e.clientX, e.clientY, rect))) {
+            const rng = editor.dom.createRng();
+            rng.setStartAfter(lastInlineBeforeBlock.dom);
+            rng.collapse(true);
+            editor.selection.setRng(rng);
+            e.preventDefault();
```
spocke: I think I would traverse this to find a better spot so something like this:

```suggestion
            CaretFinder.prevPosition(target.dom, CaretPosition.after(lastInlineBeforeBlock.dom)).each((pos) => {
              e.preventDefault();
              editor.selection.setRng(pos.toRange());
            });
            
```

modules/tinymce/src/core/main/ts/util/Quirks.ts
```diff
@@ -755,6 +804,7 @@ const Quirks = (editor: Editor): Quirks => {
 
     // WebKit
     if (isWebKit) {
+      fixInLISelection();
```
spocke: Isn't it just Chrome or is Safari also affected?

spocke (approved):

modules/tinymce/src/core/main/ts/util/Quirks.ts
lorenzo-pomili: created `TINY-14425`

modules/tinymce/src/core/main/ts/util/Quirks.ts
lorenzo-pomili: refactored

modules/tinymce/src/core/main/ts/util/Quirks.ts
lorenzo-pomili: actually that new line is exactly the difference between the first example in the comment and this example, for this case indent the HTML in a different way creates a fake child in the parent `li`

modules/tinymce/src/core/main/ts/util/Quirks.ts
lorenzo-pomili: this seems not to work in some cases (for example the cases with the tag `img`)

modules/tinymce/src/core/main/ts/util/Quirks.ts
lorenzo-pomili: ok, Safari seems to have the same bug

lorenzo-pomili: ok, I was wrong the bug is not in Safari

lorenzo-pomili: ok, now it isn't applied to Safari

modules/tinymce/src/core/main/ts/util/Quirks.ts
spocke: I want us to run some code that isn't browser defined behaviour so that we know that it does the right thing. Setting the selection after the last inline in not really a valid caret position say the last one was a `b` then the caret would be at `<li><b>foo</b>|<div>...` the valid caret position in that case is `<li><b>foo|</b><div>...` so we would rely on browser normalization. The caret finder it will walk the dom until it finds a spot so it it looks like this: `<li><b><i>foo</i></b><div>|...` and I run the CaretFinder.prevPosition it would walk backwards traversing the tree until it find the right spot so in this case this `<li><b><i>foo|</i></b><div>...`.

So I altered the code to use the block as the start point I think that is what I got wrong in the suggestion and that seems to work:

```
  const fixInLISelection = () => {
    editor.on('mousedown', (e) => {
      const target = SugarElement.fromDom(e.target);
      if (isListItem(target)) {
        firstBlockChildOrNewLine(target).each((firstBlock) => {
          const prevSiblings = Traverse.prevSiblings(firstBlock);

          Arr.findLastIndex(prevSiblings, isValidSibling).bind((lastI) => Arr.get(prevSiblings, lastI))
            .each((lastInlineBeforeBlock) => {
              if (Arr.get(getClientRects([ lastInlineBeforeBlock.dom ]), 0).exists((rect) => clickAfterEl(e.clientX, e.clientY, rect))) {
                CaretFinder.prevPosition(target.dom, CaretPosition(firstBlock.dom, 0)).each((pos) => {
                  e.preventDefault();
                  editor.selection.setRng(pos.toRange());
                });
              }
            });
        });
      }
    });
  };
```

However I ran into a bug while testing that out if you don't have focus set to the editor then try to place the caret after the LI then the focus is not moved to the editor. We need to fix that since otherwise this fix introduces a new bug. I think we can fix that by manually shifting focus into the editor but that needs to work for iframe more and inline mode and in shadow dom.

modules/tinymce/src/core/main/ts/util/Quirks.ts
spocke: I thought all white space before/after blocks are trimmed at parse time? 🤔

lorenzo-pomili: I suppose they should but I notice that it creates this strange empty `child` and this code is needed to manage that case

modules/tinymce/src/core/main/ts/util/Quirks.ts
spocke: I think `editor.focus()` is all we need to call after the `e.preventDefault()` call since that handles all complexities around focus of iframe, inline, editable fields etc.

lorenzo-pomili: I was fixing the tests, now I'll try to also add the editor.focus, thanks

lorenzo-pomili: yes, it seems to solve the issue before the editor is focussed

lorenzo-pomili: fixed

modules/tinymce/src/core/main/ts/util/Quirks.ts
lorenzo-pomili: ok, the problem is HTML too, basically the `previousSibling` in that case gets us a different case depending if it has a new line or not, if you try to click on them you can notice that if it has not new line the `previousSibling` is the first `span` and if you click on the one with new line you get an empty test, here the fiddle: https://fiddle.tiny.cloud/727T4ZlKRr/0

lorenzo-pomili: you can see it better here: https://fiddle.tiny.cloud/727T4ZlKRr/1

spocke: That gets converted to a space if you run `tinymce.activeEditor.getBody().innerHTML` on that fiddle there are no newlines.

lorenzo-pomili: do you mean that is fiddle to give that behaviour?
because also just writing a plain html in a file gives the same result:
```
<!DOCTYPE html>
<html>
  <body>
    <div>
      <div id="debug_div">
      <ol><li><span style="border: 2px solid red;">abc</span>
      <span style="display: block;">def</span></li></ol>
      <ol><li><span style="border: 2px solid red;">abc</span><span style="display: block;">def</span></li></ol>
    </div>
  </body>
  <script>
  document.getElementById('debug_div').addEventListener('mousedown', (e) => {
    console.log('e: ', e.target.childNodes, e.target.children, e.target.children[1]?.previousSibling);
  });
  </script>
</html>
```

lorenzo-pomili: the problem is that `childNodes` and traversing them via `prevSibling/nextSibling` includes whitespaces in the source markup so we need that check and to represent it in the comment we can use the current version or:
- `<li>a\n<div>b</div></li>`, but using `\n` would be a valid way to represent this case only if it's used with something like `setContent`
- `<li>a<br><div>b</div></li>` with a `<br>` is a different case, we already intercept it with `firstBlockChildOrNewLine`

lorenzo-pomili: ok, now I understood have a check

PR #11105, opened 2026-05-11
TINY-14296: fix bogus content on print

lorenzo-pomili:
Related Ticket: TINY-14296

Description of Changes:
The problem was that we were removing all kind of bogus element but we should remove the `data-mce-bogus=all` and unwrap the others as you can see in [ContentCleanup.ts](https://github.com/tinymce/tinymce/blob/main/modules/tinymce/src/core/main/ts/content/ContentCleanup.ts)

Pre-checks:
* [x] Changelog entry added

Review:
* [x] Milestone set

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed print behavior so placeholder/bogus elements remain hidden when printing.
  * Restored previously suppressed content so valid elements now appear correctly in printed output.

[Review Change Stack](https://app.coderabbit.ai/change-stack/tinymce/tinymce/pull/11105)

### DIFF
--- a/.changes/unreleased/tinymce-TINY-14296-2026-05-11.yaml
+++ b/.changes/unreleased/tinymce-TINY-14296-2026-05-11.yaml
@@ -1,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: The content of bogus was always hidden in print, even when it shouldn't be.
+time: 2026-05-11T13:06:17.044648272+02:00
+custom:
+    Issue: TINY-14296
--- a/.changes/unreleased/tinymce-TINY-14296-2026-05-11.yaml
+++ b/.changes/unreleased/tinymce-TINY-14296-2026-05-11.yaml
@@ -1,6 +1,6 @@
 project: tinymce
 kind: Fixed
-body: The content of bogus was always hidden in print, even when it shouldn't be.
+body: Some elements were hidden when printed even if the content should be shown.
 time: 2026-05-11T13:06:17.044648272+02:00
 custom:
     Issue: TINY-14296
--- a/.changes/unreleased/tinymce-TINY-14296-2026-05-11.yaml
+++ b/.changes/unreleased/tinymce-TINY-14296-2026-05-11.yaml
@@ -1,6 +1,6 @@
 project: tinymce
 kind: Fixed
-body: Some elements were hidden when printed even if the content should be shown.
+body: Some annotated text was hidden when printed even though it should have been visible.
 time: 2026-05-11T13:06:17.044648272+02:00
 custom:
     Issue: TINY-14296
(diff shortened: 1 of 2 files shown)


modules/oxide/src/less/theme/content/print/print.less
```diff
@@ -1,11 +1,14 @@
...
     .mce-placeholder,
-    [data-mce-bogus=all],
-    [data-mce-bogus] {
+    [data-mce-bogus=all] {
       display: none !important;
     }
 
+    [data-mce-bogus] {
+      display: contents !important;
+    }
```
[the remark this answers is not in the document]
lorenzo-pomili: changed, have a check

.changes/unreleased/tinymce-TINY-14296-2026-05-11.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: The content of bogus was always hidden in print, even when it shouldn't be.
```
TheSpyder: This only describes the technical change. Please describe the user-facing change, users don't know what "bogus" is.

TheSpyder (changes requested):

.changes/unreleased/tinymce-TINY-14296-2026-05-11.yaml
lorenzo-pomili: changed

TheSpyder (approved):

.changes/unreleased/tinymce-TINY-14296-2026-05-11.yaml
```diff
@@ -0,0 +1,6 @@
+project: tinymce
+kind: Fixed
+body: Some elements were hidden when printed even if the content should be shown.
```
spocke: I wonder if this is to abstract reading that I don't know what element is now visible in printed. The JIRA is about spellchecking markers and those are text annotations. So maybe we should have that in there:
`Some annotated text was hidden when printed even though it should have been visible.`

spocke: Also with this does this means that the red squiggly is printed I assume not since are background only right?

spocke (approved):

.changes/unreleased/tinymce-TINY-14296-2026-05-11.yaml
lorenzo-pomili: changed.

> Also with this does this means that the red squiggly is printed I assume not since are background only right?

from my tests it's not printed, the `display: contents` should do exactly what we do with the `getContent` unwrap the element, so we should have the correct behaviour for every bogus element

